PR #724, opened 2023-08-17
Agent: always build parent directory first

olafurpg:
We have two cases of this fixing an error related to `Could not resolve "@sourcegraph/cody-shared"`

## Test plan

n/a


vdavid (approved):
THANK YOU!! This should be very helpful!

PR #725, opened 2023-08-17
Autocomplete: Do not trigger a multi-line request for a method invocation

philipp-spiess:
A quick workaround for https://github.com/sourcegraph/cody/discussions/358#discussioncomment-6519606 that just doesn't trigger a multi-line completion if the prefix matches `/\..*\($/)` (so a `.` and ends with `(`).

The idea is to use tree sitter and language analysis to detect this case but that's a quick fix that should cover a large range of languages without causing too much harm. WDYT @vovakulikov @valerybugakov?

## Test plan

- Added a test case

### DIFF
--- a/vscode/src/completions/multiline.ts
+++ b/vscode/src/completions/multiline.ts
@@ -21,6 +21,11 @@ export function detectMultiline(
     }
 
     if (enableExtendedTriggers && currentLinePrefix.match(OPENING_BRACKET_REGEX)) {
+        // @TODO: Use tree sitter to detect a method call instead of this
+        if (currentLinePrefix.match(/\..*\($/)) {
+            return false
+        }
+
         return true
     }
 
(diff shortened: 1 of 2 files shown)


vscode/src/completions/multiline.ts
```diff
@@ -21,6 +21,11 @@ export function detectMultiline(
     }
 
     if (enableExtendedTriggers && currentLinePrefix.match(OPENING_BRACKET_REGEX)) {
+        // @TODO: Use tree sitter to detect a method call instead of this
+        if (currentLinePrefix.match(/\..*\($/)) {
```
vovakulikov, line 25: Interesting, @philipp-spiess, but I think this regexp doesn't catch when we do a function call `hello(`
We also should be careful with function declarations cases like `function hello(`. I tried to catch these cases
in this PR https://github.com/sourcegraph/cody/pull/726 (maybe I missed something important there actually)

philipp-spiess: I haven't thought about function invocation much, you're right. More than happy to move forward with your regex too, just one case I noticed that might be interesting, not sure.

Glad I could at least help with the unit test 🙈

philipp-spiess:
Let's move this to #726. One PR is enough, happy to hand this win to Vova 🙈

PR #726, opened 2023-08-17
Don't trigger multiline completion in case of function or method invocation

vovakulikov:
Fixes https://github.com/sourcegraph/cody/discussions/358#discussioncomment-6519606

As it is described in GH discussion, we shouldn't make multiline completion for cases like function or method 
invocation (but we should do this in case of method or function declaration). 

So this PR gates multiline completion behind this regexp check (regexp isn't ideal, but this is what we have to
work with until we get parsers working) You can check this regexp [here](https://regex101.com/r/k85odc/1)

## Test plan
- Check that completion doesn't try to do multiline complete if your cursor is within arguments
position of function or method invocation/call statement

### DIFF
--- a/vscode/src/completions/utils/text-utils.ts
+++ b/vscode/src/completions/utils/text-utils.ts
@@ -2,6 +2,7 @@ import * as vscode from 'vscode'
 
 export const INDENTATION_REGEX = /^[\t ]*/
 export const OPENING_BRACKET_REGEX = /([([{])$/
+export const FUNCTION_OR_METHOD_INVOCATION_REGEX = /\b[^()]+\((.*)\)$/g
 
 export const BRACKET_PAIR = {
     '(': ')',
--- a/vscode/src/completions/utils/text-utils.ts
+++ b/vscode/src/completions/utils/text-utils.ts
@@ -3,6 +3,7 @@ import * as vscode from 'vscode'
 export const INDENTATION_REGEX = /^[\t ]*/
 export const OPENING_BRACKET_REGEX = /([([{])$/
 export const FUNCTION_OR_METHOD_INVOCATION_REGEX = /\b[^()]+\((.*)\)$/g
+export const FUNCTION_KEYWORDS = /^(function|def|fn)/g
 
 export const BRACKET_PAIR = {
     '(': ')',
--- a/vscode/src/completions/getInlineCompletions.test.ts
+++ b/vscode/src/completions/getInlineCompletions.test.ts
@@ -583,6 +583,30 @@ describe('getInlineCompletions', () => {
             expect(requests[0].stopSequences).not.toContain('\n')
         })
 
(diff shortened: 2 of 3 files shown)


philipp-spiess (commented):
This doesn't work if someone has automatic closing tags insertion disabled: 

Screenshot 2023-08-17 at 18 17 22


I think it's easier if we just look at the prefix. In my PR I just check wether there's also a `.` somewhere on the same line prefix. WDYT?

vscode/src/completions/utils/text-utils.ts
```diff
@@ -2,6 +2,7 @@ import * as vscode from 'vscode'
 
 export const INDENTATION_REGEX = /^[\t ]*/
 export const OPENING_BRACKET_REGEX = /([([{])$/
+export const FUNCTION_OR_METHOD_INVOCATION_REGEX = /\b[^()]+\((.*)\)$/g
```
philipp-spiess, line 5: TIL about [word boundaries](https://www.regular-expressions.info/wordboundaries.html)

philipp-spiess (approved):

vovakulikov:
@philipp-spiess 
>I think it's easier if we just look at the prefix. In my PR, I check whether there's also a . somewhere on the same line prefix. WDYT?

Yeah, I've seen your regexp. I think we should check prefix+suffix for several reasons. 
- The main reason is to detect declarations (sometimes that can look like invocations) like `method.hello () {` these should be treated as declarations because of { body bracket 
- `hello(` could be many things. We just can't say much about what exactly we want to have here 

But I see your point about `hello.world(` probably we can do both with this regexp `(\b[^()]+\((.*)\)$)|(\..*\($)` ?

philipp-spiess:
@vovakulikov 

>The main reason is to detect declarations (sometimes that can look like invocations) like method.hello () { these should be treated as declarations because of { body bracket

Can you clarify this? I think this is actually a good reason of why to only check for the prefix, because depending on where my cursor is, I want either a single line or a multi line completion. E.g.:

-  `method.hello (█) { ` 
	Would expect this to be a method call and do a single-line completion
- `method.hello () { █`
    Would expect this to be the beginning of a block instead

vovakulikov:
@philipp-spiess 

Cursor context This is something that we probably can achieve with proper lexeme type context around the cursor with regexp; it's too complex for a number of langs we're trying to support here. My point here is that `method.hello (█) {` regardless of cursor position will be handled as multi-line completion because of `{` in the suffix.

vovakulikov:
I updated unit tests and added some logic about empty/non-empty suffix, I hope it fixes both cases that you mentioned above (but it doesn't take into account context information based on cursor position)

philipp-spiess:
@vovakulikov Cool, makes sense. Feel free to merge this, I’m sure you make the right call

vovakulikov:
>  I’m sure you make the right call

👍 good luck to me :)

PR #727, opened 2023-08-17
Autocomplete: Include up to a single new line in the Anthropic prompt

philipp-spiess:
We learned that Anthropic is giving us worse results with trailing whitespace in the prompt.
 To fix this, we started to trim the prompt.

 However, when the prefix includes a line break, the LLM needs to know that we do not want the
 current line to complete and instead start a new one. For this specific case, we're injecting
 a line break in the trimmed prefix.

 This will only be added if the existing line is otherwise empty and will help especially with
 cases like users typing a comment and asking the LLM to provide a suggestion for the next
 line of code:

     // Write some code
     █

## Test plan

- Added a test case

### Before

Screenshot 2023-08-17 at 17 53 06

### After

Screenshot 2023-08-17 at 18 04 44


vovakulikov (approved):
this looks good to me, just for the history of this issue, I thought that there will be two cases 
when we want to avoid skipping /n (only when we're inside of the comment block) 
So I came up with two regexps for this 
- [Python comment](https://regex101.com/r/ElhYrY/1) (inline and multiline string like comments)
- [and all other langs comments ](https://regex101.com/r/2tGEvE/1)with // as inline and /**/ as multiline comments 

but after chat with @philipp-spiess I see that we want to keep /n in general for other cases too, so these regexp are not useful here, but maybe it would help later in some other case when we want to track comments block while parsers are in progress

valerybugakov (approved):

Issue #583, opened 2023-08-07
Local index-backed search


beyang:
Add back the `/search` command, backed by an indexed local search utility.

- [x] Finish refactor
- [x] Integrate into VS Code UI
- [x] Update VS Code results display
- [x] Search quality pass
- [x] Automatically trigger index

Follow-up:
- [ ] Replace context fetching mechanism with indexed-search fetcher
- [ ] Language support
  - [ ] JavaScript
  - [ ] Java
  - [ ] C
  - [ ] C++
  - [ ] C#
  - [ ] SQL
  - [ ] PHP
  - [ ] Ruby
  - [ ] Swift
  - [ ] Rust
- [ ] Anthropic -> Sourcegraph Cody endpoint

PR #728, opened 2023-08-17
Adds an experimental command `/symf` that provides an indexed keyword search capability

beyang:
Adds an experimental `/symf` command to the Cody sidebar. `/symf` uses [`symf`](https://github.com/sourcegraph/symf), a CLI that implements LLM-enabled keyword search. MVP for https://github.com/sourcegraph/cody/issues/583

## Use

1. Install `symf`: `go get github.com/sourcegraph/symf/cmd/symf@latest`
2. Set the following fields in the VS Code settings:
    ```
   "cody.experimental.symf.path": "the full path to the symf binary",
   "cody.experimental.symf.anthropicKey": "an api key from anthropic",
   ```
3. Type "/symf \<your query\>" in the Cody sidebar:
image

The first time `/symf` is used for a given repository, it will need to build an index, so you'll see a message like this:

An alert will appear when the index is finished building:

## Comments

The command is experimental and requires manual steps (installing the binary and setting the config fields) at the moment. It's intended for internal use to test out the quality of the search results within the Cody UI.

Currently, the following languages are supported by `symf`:
* TypeScript
* TSX
* Go
* Python
* C

The next milestone is to make this a viable replacement for `/search`. This will require:
- [ ] Replacing the direct Anthropic dependency with Cody Gateway
- [ ] Auto-download the symf binary
- [ ] Show the expanded keyword query being used under the hood (more transparency)
- [ ] A better UX for communicating index build status
- [ ] (Potential) implement index-less search in symf
- [ ] Add more languages to symf

## Test plan

Command is experimental only at the moment.


sqs (approved):
haven't run it, but i don't see anything that would cause the non-exoerimental functionality in the extension to fail, so lgtm to merge

PR #730, opened 2023-08-17
update dotcomUrl

akalia25:
This PR does two things:
- update the if statement that sends events to dotcom to do a match on hostname (avoids issue with serverEndpoint and dotcomUrl not matching because `"sourcegraph.com" != "sourcegraph.com/"`
- sends to sendEventLogRequestToAPI to have proper user_ids.

## Test plan
tested locally

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -342,7 +342,7 @@ export class SourcegraphGraphQLAPIClient {
             return {}
         }
         if (new URL(this.config.serverEndpoint).hostname === new URL (this.dotcomUrl).hostname) {
-            return this.sendEventLogRequestToDotComAPI(event)
+            return this.sendEventLogRequestToAPI(event)
         }
         const responses = await Promise.all([
             this.sendEventLogRequestToAPI(event),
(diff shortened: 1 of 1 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -341,7 +341,7 @@ export class SourcegraphGraphQLAPIClient {
         if (this.config.isRunningInsideAgent) {
             return {}
         }
-        if (this.config.serverEndpoint === this.dotcomUrl) {
+        if (new URL(this.config.serverEndpoint).hostname === new URL (this.dotcomUrl).hostname) {
             return this.sendEventLogRequestToDotComAPI(event)
```
nathan-downs, line 345: This function only sends an anonymous request, we should change this to this.sendEventLogRequestToAPI(event) so that dotcom events are associated with user_id and configuration information.

nathan-downs (approved):
LGTM!

akalia25:
As @nathan-downs mentioned sending to `sendEventLogRequestToAPI` will now have proper user_ids. Verified below with local testing:

image

dominiccooney (approved):

PR #732, opened 2023-08-18
Agent: recipes/execute - handle empty selection

olafurpg:
Previously, the agent crashed when executing a recipe when the client hadn't provided any text document selection.

## Test plan

- Start IntelliJ with an open file but no selection
- Send a chat message. Nothing happened, it was stuck in processing mode and the "Stop generating" button is just there.


vdavid (approved):
Nice.

PR #733, opened 2023-08-18
Add increased autocomplete debounce time feature flag support

taras-yemets:
Tweak single-line autocomplete debounce time depending on `'cody-autocomplete-increased-debounce-time-enabled'` feature flag value. 

## Test plan
As we restrict feature flags evaluation only to dotcom users, the actual events data should be checked on dotcom instance `event_logs` table ([docs](https://docs.sourcegraph.com/dev/how-to/use_feature_flags#measure-the-effect-of-a-feature-flag)). 
During the development I tested the functionality in the following way:
- remove (comment out) `isDotCom` condition in `FeatureFlagProvider` methods
- run local Sourcegraph instance (`sg start`)
- log in to the local instance as admin and add `'cody-autocomplete-increased-debounce-time-enabled'` feature flag
- sign it to the local Sourcegraph instance in Cody VSCode extension
- launch the VSCode extension from the VSCode Run&Debug menu
- perform a Cody completion action and check whether the correct timeout is used depending on the feature flag value
- double-check feature flag value using the GraphQL API
```
query Name {
  evaluateFeatureFlag(flagName: "cody-autocomplete-increased-debounce-time-enabled")
} 
```


philipp-spiess (approved):
LGTM! Let's add a note to the change log though :)

PR #734, opened 2023-08-18
Agent: add requests for logging events and to find user ID

olafurpg:
Previously, the JetBrains plugin still sent direct network requests to the Sourcegraph instance. This was problematic because it meant that users behind proxies had to configure both proxy settings for the agent and for the Java plugin. This PR adds two endpoints to the Agent protocol allowing the JetBrains plugin to send requests to the Agent instead of sending network requests to Sourcegraph.

## Test plan

Tested alongside JetBrains PR

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -196,6 +196,24 @@ export class Agent extends MessageHandler {
                 return { items: [] }
             }
         })
+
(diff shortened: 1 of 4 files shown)


chwarwick (approved):
This looks reasonable to me and aligned with moving logic into the agent.  I think there are likely to be several more of the graphql requests but I don't think it will explode in size.

PR #735, opened 2023-08-18
Experimentation: Move FeatureFlagProvider to shared and add tests

philipp-spiess:
Moved the new `FeatureFlagProvider` to `shared` so we can use it everywhere. Also added some tests while at it.

## Test plan

- Added tests


taras-yemets (approved):
Thank you, @philipp-spiess!

lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -1,7 +1,12 @@
...
+import { SourcegraphGraphQLAPIClient } from '../sourcegraph-api/graphql'
+import { isError } from '../utils'
 
 export enum FeatureFlag {
+    // This flag is only used for testing the behavior of the provider and should not be used in
+    // product code
+    TestFlagDoNotUse = 'test-flag-do-not-use',
+
     CodyAutocompleteIncreasedDebounceTimeEnabled = 'cody-autocomplete-increased-debounce-time-enabled',
 }
```
taras-yemets, line 11: Different clients Cody clients may have different sets of feature flags. Maybe we should use strings (instead of enum) for feature flag names?

philipp-spiess: I think it's fine if the enum contains more flags. It's actually going to be useful to have them centralised. E.g. the flags for autocomplete will be in agent too and we can test that implementation across all clients at the same time.

PR #738, opened 2023-08-18
log custom command only

abeatrix:
Since we are already logging default command being called in MessageProvider, we should only log custom command in the command controller.

## Test plan

check output channel to confirm default commands are not being logged twice.


akalia25 (approved):
LGTM! Thank you!

Issue #653, opened 2023-08-11
Update notice w/ link to release post


toolmantim:
We need a way for notifying people that there's been a Cody release (minor version, not patch), w/ a link to an announcement post to explain what's changed. The goal being to bring people on the journey with us as we improve Cody: understand the changes, get excited, a share it with their colleagues and the world.

- [x] Design
- [x] Reviewed & ready to go

toolmantim:
<img src="https://github.com/sourcegraph/cody/assets/153/02408042-d6b6-4efa-b3f5-a0196cc0a181" width="275">

https://github.com/sourcegraph/cody/assets/153/bb45f229-cb55-4e94-acb7-9f6b4b61967b

Figma:
- [Design](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=4655-12897&mode=design&t=nkSBtl6vjShXNV9H-4)
- [Prototype](https://www.figma.com/proto/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?page-id=3250%3A136841&type=design&node-id=4655-12898&viewport=1373%2C-18668%2C1&t=NwKfMw7l4wgK5npR-1&scaling=min-zoom&starting-point-node-id=3532%3A142502&show-proto-sidebar=1&mode=design)

Notes:
- Top position was chosen instead of bottom to avoid clashing with the command menu and stop generating button
- It uses the same styles as the [command popup menu](https://github.com/sourcegraph/cody/blob/main/vscode/webviews/ChatCommands.module.css) aside from the border-radius
- Sparkles SVG should use `var(--vscode-textLink-foreground)`

philipp-spiess:
This is so pretty ✨

toolmantim:
I'll get started on the frontend side of this (and #667)

toolmantim:
Versions could also add callouts for specific big changes, e.g. recipes → commands:

<img src="https://github.com/sourcegraph/cody/assets/153/4648c8af-72aa-4c67-aa46-e14681875a6e" width="275">

PR #746, opened 2023-08-19
Show a notice and link to release notes after extension updates

toolmantim:
Implements #653, adding a update notice to the top of the chat view for when the extension gets updated:

Screenshot 2023-08-20 at 9 52 23 pm

- Shows a notice whenever major+minor version changes
- Always links to the GitHub release for the current version (e.g. 0.8.2) and assumes the blog post for that minor release (0.8.x) will be available at the top of each one
- Doesn't show a notice for new installs
- Adds an autocomplete onboarding component too, just to ensure we've the right component structure, but is disabled for now

## Test plan

- New user
  - Clear chat history, reload extension, verify no version notice
- Existing user
  - WIth chat history, reload extension, verify has version notice
  - Dismiss version notice, increase version number, verify has version notice
  - Dismiss version notice, reload extension, verify no version notice


vscode/webviews/Notices/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { UpdateNotice } from './UpdateNotice'
+
+import styles from './index.module.css'
+
+interface NoticesProps {
+    extensionVersion: string
+}
+
+export const Notices: React.FunctionComponent<NoticesProps> = ({ extensionVersion }) => {
+    // TODO: Implement (true when version is different from last load — localstorage?)
```
toolmantim, line 13: @abeatrix could you help make this work plz?

For the logic, we could do it either on when the minor version changes (0.7.x -> 0.8.x) or when any version change happens (0.7.1 -> 0.7.2).

Either is probably fine to start with, but the former would be my preference (we don't show a notice on a patch release so people can learn to expect there'll always be an interesting blog post to go along with this notification.

toolmantim: I'll give this a go today @abeatrix and report back…

vscode/webviews/Notices/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface NoticesProps {
+    extensionVersion: string
+    probablyNewInstall: boolean
+}
+
+export const Notices: React.FunctionComponent<NoticesProps> = ({ extensionVersion, probablyNewInstall }) => (
+    <NoticesContainer>
+        <VersionUpdatedNotice version={extensionVersion} probablyNewInstall={probablyNewInstall} />
+        {/* <OnboardingAutocompleteNotice /> */}
```
sqs, line 13: I think it's probably best to just omit this commented-out component from this PR to avoid confusion, but if you want to get it in, then add a comment explaining why this is commented-out here.

vscode/webviews/Notices/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+    </NoticesContainer>
+)
+
+const NoticesContainer: React.FunctionComponent<React.PropsWithChildren> = ({ children }) => (
```
sqs, line 17: No need to make this a separate React element.

vscode/webviews/Notices/OnboardingAutocompleteNotice.module.css
```diff
@@ -0,0 +1,31 @@
+.notice {
+    /* Set the link color that VSCodeLink will use */
+    --link-foreground: var(--vscode-gitDecoration-untrackedResourceForeground, var(--vscode-textLink-foreground));
```
sqs, line 3: Using the gitDecoration color might work for your theme but it's essentially just using that you like its color in your current theme, and there is no guarantee that other themes will assign a desirable color for it.

sqs (approved):
One UX comment: as a user I'd slightly prefer that the notice not be "floating" but instead actually take up height. It feels less spammy that way IMO. Feel free to ignore.

Along with this, let's REALLY make sure the release notes are useful to users. VS Code's release notes are a really good bar to meet.

vscode/webviews/Notices/index.tsx
toolmantim: Good point… it did cross my mind. Will bring it back in a follow up!

vscode/webviews/Notices/OnboardingAutocompleteNotice.module.css
toolmantim: It's a good point. I did a whole bunch of theme testing to land on this, thinking that `-vscode-gitDecoration-untrackedResourceForeground` and `--vscode-textLink-foreground` should always end up desirable against the background, but when I re-add I'll take a fresh look.

toolmantim:
> One UX comment: as a user I'd slightly prefer that the notice not be "floating" but instead actually take up height. It feels less spammy that way IMO. Feel free to ignore.

Yeah, it was fixed, until we needed it for some other things (first being the autocomplete notice). But will keep an eye on the balance here… an easy change for the future to special case this or try diff things.

> Along with this, let's REALLY make sure the release notes are useful to users. VS Code's release notes are a really good bar to meet.

100% — I'm keen for us to tell users what's changing, and to do it well.

On the spamminess… that's why I made this major+minor only. Every fortnightly release, if there's nothing user-facing'ly interesting, we should bump the patch version only and let it invisibly auto-update.

PR #748, opened 2023-08-20
New welcome chat words

### LINKED ISSUE #633: Improve first time user experience for the command changes
In 0.6.5 we removed the small bit of onboarding we did have (the recipe buttons in the chat). We also removed the Recipes tab. We need to design better solutions to ensure people experience value, get the most out of trying Cody for the first time, and we improve day 1 retention.

- [x] Design
- [x] Reviewed & ready to go

toolmantim:
Partially implements #633 with new welcome message text:

Screenshot 2023-08-21 at 7 53 34 pm

Reason for the partial implementation is that there's currently no way to embed images or a component in the welcome markdown, so we couldn't include the illustrations as per the Figma.

Closes #633

## Test plan

- Started a new chat, verified welcome text

### DIFF
--- a/vscode/webviews/Chat.tsx
+++ b/vscode/webviews/Chat.tsx
@@ -150,9 +150,7 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
             // down here to render cody is disabled on the instance nicely.
             isCodyEnabled={true}
             codyNotEnabledNotice={undefined}
-            afterMarkdown={
-                '🔔 Recipes are now [Cody Commands](command:cody.action.commands.menu)! You can start using them with the [⌥C](command:cody.action.commands.menu) shortcut on highlighted code. \n\nSee [Getting Started](command:cody.welcome) to learn more about [Cody Commands](command:cody.action.commands.menu).'
-            }
+            afterMarkdown={welcomeMessageMarkdown}
             helpMarkdown=""
             ChatButtonComponent={ChatButton}
             pluginsDevMode={pluginsDevMode}
@@ -358,3 +356,10 @@ const FeedbackButtons: React.FunctionComponent<FeedbackButtonsProps> = ({ classN
         </div>
     )
 }
+
+const welcomeMessageMarkdown = `Start writing code and I’ll autocomplete lines and entire functions for you.
+
+You can ask me to explain, document and refactor code using the [Cody Commands](command:cody.action.commands.menu) action (⌥C), or by right-clicking on code and using the “Cody” menu.
+
+See the [Getting Started](command:cody.welcome) guide for more tips and tricks.
+`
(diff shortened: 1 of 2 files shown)


vscode/webviews/Chat.tsx
```diff
@@ -358,3 +356,10 @@ const FeedbackButtons: React.FunctionComponent<FeedbackButtonsProps> = ({ classN
         </div>
     )
 }
+
+const welcomeMessageMarkdown = `Start typing code in any source file and I’ll offer autocompletions for lines and entire functions (press Tab to accept them).
+
+Use [Cody Commands](command:cody.action.commands.menu) to explain code, document code, and refactor code using the Cody button (⌥C) in the top right of a file.
```
sqs, line 362: - The shortcut key might be different for different users. There is an API to get what it actually is for the user. As a quick fix, just mention the VS Code command to run instead of the shortcut key. (That is good practice, anyway, and is how most other VS Code extensions describe how to invoke their functionality.)
- The icon showing up in the top right of a file is an experimental feature that is off by default. I think it is annoying/spammy to have a logo in every editor tab's title bar, personally.
- Overall this feels like "feature-first" language ("Use Cody Commands ...") rather than "Ask Cody to edit ... using Cody Commands".
  - I also think it should emphasize editing not explaining (by putting editing first) since in my experience that is a more common use case for me. WDYT?

vscode/package.json
```diff
@@ -655,7 +655,7 @@
       "editor/title": [
         {
           "command": "cody.action.commands.menu",
-          "when": "cody.activated && config.cody.experimental.editorTitleCommandIcon",
```
sqs, line 658: I think it feels spammy and annoying to have a logo in every editor tab title bar. If I wasn't building Cody myself and didn't already love it, I would personally insta-uninstall if I saw that.

sqs (changes requested):

vscode/package.json
toolmantim: Fair call. I figured because you can right click on it and choose "Hide" it tipped the scales towards being the right call… but given we've also made "Settings" more prominent (the cog in the title bar), this can be an optional thing to easily turn on. Let's leave it off by default for now.

I don't quite understand the "Experimental" label on it though.

vscode/webviews/Chat.tsx
toolmantim: Yeah, the shortcut should be dynamic (and cater for Windows users too). It's hardcoded in a number of other places like this.

If the Cody editor title action button is optional, we need another good option for teaching people how to interact with Cody in the editor, in a way that is on-by-default.

Perhaps:

> Hello! I'm Cody
>
>   As you write code I’ll offer autocompletions for lines, and entire functions (press Tab to accept them).
>
> You can ask me to explain, document and refactor code using the [Cody Commands]() action (⌥C), or right-clicking on any code and use the “Cody” menu.
>
> See the Getting Started guide for more tips and tricks.

toolmantim: This is what's currently in `main`, which this aims to improve:

Screenshot 2023-08-21 at 2 15 07 pm

sqs: That message looks good to me (i.e., not enabling the editor title action button by default, and therefore not mentioning it). I'll approve this to unblock.

sqs (approved):

PR #752, opened 2023-08-21
improve 2 Cody text labels

sqs:
- Use "Cody is working" label instead of "Asking Cody". "Working" makes sense for inline refactors *and* questions. "Asking" makes less sense for inline refactors.
- Eliminate weird-looking spaces in `' / '`.

## Test plan

n/a


valerybugakov (approved):

vscode/src/services/CodeLensProvider.ts
```diff
@@ -143,7 +143,7 @@ function getLenses(id: string, codeLensRange: vscode.Range, isPending: boolean):
     const title = new vscode.CodeLens(codeLensRange)
     // Open Chat View
     title.command = {
-        title: isPending ? '$(sync~spin) Asking Cody...' : '✨ Edited by Cody',
+        title: isPending ? '$(sync~spin) Cody is working...' : '✨ Edited by Cody',
```
toolmantim, line 146: Much nicer!

Issue #714, opened 2023-08-17
Some output (`/search`) is not throttled; some LLM use (reranking, etc.) is throttled but should not be


dominiccooney:
Forked from discussion in #704. After #699 is fixed so the scrolling behavior is back, that scrolling is too fast because `/search` sends a lot of content "at once."

We have "Typewriter" wired to chat clients [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@23ce972044869cbfca2cd31b20dd85293d4b1771/-/blob/lib/shared/src/chat/chat.ts?L23:28&popover=pinned) but:
- The layer that intervention happens at is wrong. For example, it sits behind the BotResponseMultiplexer for recipes. The BRM will buffer and undo some of the typewriter effect; some BRM subscribers are code and don't need a typewriter effect. For the LLM reranker, the typewriter effect will slow down the reranker.
- The effect doesn't work for search, because that doesn't use the chat client.

It would be good to move this incremental display closer to the web UI (maybe not into it, but closer to it) so that everything that a human sees uses it (and it's harmonized with the incremental Markdown handling) and things destined for machines can happen as fast as possible.

arafatkatze:
@dominiccooney Thanks for the information. I would love to work on this issue and look deeper into this.

dominiccooney:
@arafatkatze Super! Thanks for your help with this.

I have a diff which moves the typewriter into MessageProvider which I'll post soon—I'm just working on updating tests. This improves the Typewriter (I added a method to wait for the typewriter to finish without forcibly stopping it and getting a huge chunk) and gets it out of the reranker and keyword expander path.

I haven't integrated Typewriter with the "search" output. So you might like to crib from the improved Typewriter in my diff and integrate it there, or move the typewriter out of the MessageProvider and closer to the chat view.

arafatkatze:
@dominiccooney Thanks a lot for such a detailed response and guidance in [this comment](https://github.com/sourcegraph/cody/pull/719#issuecomment-1683393076),
I was able to make the default typewriter work for the `search` scenario but it was obviously too fast as you had mentioned  
_What you need here will involve the typewriter "slowing down" the interaction from ContextSearch._

https://github.com/sourcegraph/cody/blob/906603537a822e19a2f87c3f813c872af5ecc73e/lib/shared/src/chat/typewriter.ts#L59-L60

Basically I wanted to change this ^ a bit and after some tweaks to the typewriter max_delay I was able to make it "slow"

[...]

PR #755, opened 2023-08-21
Typewriter outputs larger chunks, up to short phrases

dominiccooney:
This simplifies the typewriter effect so that it doesn't change speed depending on the amount of pressure behind the typewriter. It also types in larger chunks. See #714.

## Test plan

```
pnpm test -- typewriter.test.ts
```

Manual test:

1. Open Cody-VScode and chat to it: "Explain Rust IterMut with examples." Typing should be incremental but be in chunks and generally hit word boundaries.
2. Cmd-Shift-P, Configure Display Language, 日本語 (ja), follow prompts to restart.
3. Chat to Cody: "RustのIterMutを例で説明して". Typing should be incremental—though monotonous—even in passages without spaces.

### DIFF
--- a/lib/shared/src/chat/typewriter.test.ts
+++ b/lib/shared/src/chat/typewriter.test.ts
@@ -1,0 +1,69 @@
+import { describe, expect, it, vi } from 'vitest'
+
+import { Typewriter } from './typewriter'
+
+describe('TypeWriter', () => {
+    it('closes the downstream consumer, no output', async () => {
+        const update = vi.fn(() => {})
+        const close = vi.fn(() => {})
+        const tw = new Typewriter({
+            update,
+            close,
+        })
+        tw.close()
+        await tw.finished
+        expect(update).toHaveBeenCalledTimes(0)
+        expect(close).toHaveBeenCalledTimes(1)
+    })
+    it('completes output before closing the downstream consumer', async () => {
+        let lastText = ''
+        const update = (s: string) => {
+            lastText = s
+        }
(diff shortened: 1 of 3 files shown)


arafatkatze (commented):
@dominiccooney I found something different while testing locally. Your tests worked perfectly but I faced some jitter when typing code examples with the typewriter. 

See this [commit](https://github.com/sourcegraph/cody/commit/d29b088924e89c292f84e46c9c1bfc9d86b11a4b) for adding a blob of text to be printed(on top your branch).  I changed the original speed of 700WPM because it was very jittery.  See the video below

https://github.com/sourcegraph/cody/assets/11155207/6e6c5c1e-b6c4-43ca-b879-7fa0449d17c0



Then I changed to DELAY_MS to 20ms and then the jitter was a lot lower than before. See the video below

https://github.com/sourcegraph/cody/assets/11155207/8459c9ed-ea36-41f8-a905-de807f58c609

Also in the case of things coming upstream from the bot they are throttled by the speed of the response from upstream so setting a lower DELAY_MS doesn't affect that. 


The trade off with this is that it prints everything super fast in the case of search but atleast it's smooth. I am trying to figure out the causality of the jitter to figure out if it's just a debug runner view thing or if there is a deeper issue.

dominiccooney:
Interesting find @arafatkatze. Let's hold off on merging this until we work out what's happening here. That bounce is quite distracting.

arafatkatze:
Makes sense

PR #759, opened 2023-08-21
VS Code: Release 0.8.0

philipp-spiess:
Preparing for the 0.8.0 release.

Internal Slack discussion with blocking PRs: https://sourcegraph.slack.com/archives/C05AGQYD528/p1692606488444629

## Test plan

👀 : just a version change

PR #769, opened 2023-08-21
changelog update for improved unit test

abeatrix:
update with more context

## Test plan

not required


kalanchan (approved):

philipp-spiess:
Added in https://github.com/sourcegraph/cody/pull/759/commits/7abb45f2f88132dd1215244ec4b61d4ede214237 so we don't need to run more CI

PR #778, opened 2023-08-21
Intelligent Cody: Fetch locations multiple hops in the graph

efritz:
This PR modifies the initial implementation in #692 to not only track definitions _within the current user selection_, but also _definitions of symbols referenced in fetched definitions_. This goes two hops deep in the code navigation graph, but the implementation is general enough to go an arbitrary number of hops by updating a `recursionLimit` global variable.

## Test plan

Existing unit tests; validated by hand.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
           key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: ${{ runner.os }}-pnpm-store-
       - run: pnpm install
+      - run: pnpm build
       - run: pnpm run test:unit --run
 
   test-integration:
--- a/vscode/src/chat/GraphContextProvider.test.ts
+++ b/vscode/src/chat/GraphContextProvider.test.ts
@@ -57,7 +57,7 @@ func bonk() => { return new bar().Bar(new foo().Foo(), baz.Foo()) }
 
 describe('extractRelevantDocumentSymbolRanges', () => {
     test('returns all document symbol ranges by default', async () => {
-        const ranges = await extractRelevantDocumentSymbolRanges(URI.file('/test-1.test'), undefined, () =>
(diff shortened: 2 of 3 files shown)


abeatrix (commented):
Cody is stuck when I tried to run this 🤔 
image

efritz:
@abeatrix Looks like maybe definitions are in files that can't be opened in the workspace. What repo/file is this on (I'd like to repro).

In the meantime could you try it on some Go or TypeScript code? I'm currently testing on the sg/sg repo so I might be blind to some issues in Python code or non-sg repositories at the moment.

abeatrix:
@efritz sounds good, will try it on go and typescript and report back! Here is the repo with the code I highlighted when running into the aforementioned issue: https://sourcegraph.com/github.com/langchain-ai/langchain/-/blob/libs/langchain/langchain/prompts/prompt.py?L57-92

abeatrix:
python: 
image
typescript:
image
go:
image

Should we limit the amount of symbol we include in each request? I'm getting ` WARN [mainThreadStorage] large extension state detected (extensionId: sourcegraph.cody-ai, global: true): 12056.3720703125kb. Consider to use 'storageUri' or 'globalStorageUri' to store this data on disk instead.` 😅 

It also seems to look for symbols from files beyond the current workspace 👀 
image

efritz:
> Should we limit the amount of symbol we include in each request?

@abeatrix Yes, but I need a way to start to pre-rank the symbols that are most relevant for a given piece of code. I'll be experimenting with that this week as well. It's hard to just say "give me the first 10 symbols" because it will very likely miss critical symbol definitions half-way through a function.

> It also seems to look for symbols from files beyond the current workspace

Yes, otherwise just put your repo/package in the LLM window. This ability is is one part of our differentiation. I do agree it's currently noisy though.

abeatrix (approved):
approving as the feature is behind an internal feature flag and the plan is to merge this before working on fixes and improvements in the next PR 👍

PR #704, opened 2023-08-16
Use CSS scroll anchoring for most chat scroll pinning

dominiccooney:
Chat scroll pinning was broken for `/search` because that adds messages in quick succession and the React effect was only looking at the change in message author.

The general approach was also iffy because we were basically polling the scroll position on updates, but in practice scrolling works best when it is managed asynchronously by the browser.

This takes a different approach:
- Use CSS scroll anchoring. This takes care of 90% of what we want with pinning the scroll position to the bottom: If you scroll away from the anchor, the scroll position doesn't change, the browser manages all of this for us. We also get better behavior when you resize the panel; the scroll position will stick to the bottom pretty intelligently.
- Because browsers anchor the scroll position to the top until there's any scrolling, this uses an intersection observer to detect when the content becomes scrollable and force it to scroll. I actually like the behavior without that, but have implemented this for consistency with what we had.
- When there's a new human message, scroll to the bottom. This means when you commit input in the chat box, we scroll to it.

I've made some subtle UX changes like animating some of the scrolls.

The scroll anchor adds a 5px gutter at the bottom of the content which is the "grippy" region for the scroll. We can avoid that extra space if we want, I took a look at it, it requires a couple of stacking contexts and I thought it wasn't worth it. But we can do it if we want.

## Test plan

1. New Cody chat, ask it to enumerate 50 states in order of accession in a numbered list. The scroll should follow the content as it streams in.
2. Scroll up, chat to ask it to enumerate the 50 states and their state vegetables. The scroll should jump to the new question when you hit enter; then follow the content.
3. Scroll up, switch to the files panel, switch back. The scroll position should be where you left it.
4. While content is streaming in, scroll away from it. As new content appears the scroll should not jump back to the bottom.
5. Scroll to the bottom, ask to enumerate 50 states etc., while content is streaming in switch to the files panel, wait, then switch back. The scroll position should be where it was when you switched away. That is, *not* pinned any more.
6. Do `/search foo` and the scroll position should keep up with the bottom of the content. (This comes in very *fast* and I think we should add a heuristic to maybe not do that. But it is better than the behavior today, which quietly drops the new content below the fold.)
7. Reload the webviews/restart the extension with a long transcript. The restored transcript should be scrolled to the bottom.

PR #779, opened 2023-08-22
lib/ui: only render scroll anchor "gutter" if there's more than one message

courier-new:
Fixes a visual bug introduced by https://github.com/sourcegraph/cody/pull/704 on Cody Web/App where the initial placeholder message in chat gets weirdly scrolled out of view when the chat is still empty.

| Before | After |
:-:|:-:
<video src="https://github.com/sourcegraph/cody/assets/8942601/6a534084-e650-450e-ab86-548c8bc623dc"> | <video src="https://github.com/sourcegraph/cody/assets/8942601/8da62c07-0e15-4654-9896-9f9b395bcb0c">

As an aside, note that the scroll anchoring doesn't actually work for the first pair of message interactions, for example in this extended "Before" clip I recorded:

https://github.com/sourcegraph/cody/assets/8942601/961436bc-a395-4c61-8365-b887a528cd44

It _does_ work once you manually scroll to the bottom of the chat for the first time, though (not included in the clip). I seem to experience this pretty consistently in VSCode, too, for what it's worth. This PR doesn't try to fix that problem, only correct the visual bug with the scroll of the first chat message.

## Test plan

- Start a new chat conversation in Web, App, and VSCode
- Verify the first message is not scrolled out of view
- Ask Cody to enumerate the 50 US states
- [After the first pair of messages,] verify scroll position follows the content as it streams in

### DIFF
--- a/lib/ui/src/chat/Transcript.tsx
+++ b/lib/ui/src/chat/Transcript.tsx
@@ -174,7 +174,7 @@ export const Transcript: React.FunctionComponent<
                     />
                 )}
             </div>
-            <div className={classNames(styles.scrollAnchor)}>&nbsp;</div>
+            {transcript.length > 1 && <div className={classNames(styles.scrollAnchor)}>&nbsp;</div>}
         </div>
     )
 })


dominiccooney:
If the 5px gap is a problem, I think we should eliminate the gap. You can put the anchor inside the container and make it `position: absolute; bottom: 0;`; put a z-index: -1 on it to float it under the thumbs up/down buttons. You might need z-index: 0 on the parent for a stacking context for that.

FWIW when debugging this it can be handy to add a `border: 2px dashed green` or rule like that to see where the anchor is.

> the scroll anchoring doesn't actually work for the first pair of message interactions. It does work once you manually scroll to the bottom of the chat for the first time, though (not included in the clip).

That's browser behavior with scroll anchoring. There's an intersection observer which does a programmatic scroll to make the anchoring start to work. But it's probably not working on web...

Can I have a pointer to how to get cody web dev going so I can look at this?

> ...  I seem to experience this pretty consistently in VSCode, too, for what it's worth.

That's odd because that does work for me. Does the intersection observer fire?

If this is too flaky we can certainly go back to managing it manually, but we need something more robust than what we were using.

lib/ui/src/chat/Transcript.tsx
```diff
@@ -174,7 +174,7 @@ export const Transcript: React.FunctionComponent<
                     />
                 )}
             </div>
-            <div className={classNames(styles.scrollAnchor)}>&nbsp;</div>
+            {transcript.length > 1 && <div className={classNames(styles.scrollAnchor)}>&nbsp;</div>}
```
dominiccooney, line 177: Seems fine but it would be nice to get to the underlying issue of the intersection observer not working for the first overflow on web.

courier-new:
> Can I have a pointer to how to get cody web dev going so I can look at this?

Yeah for sure! You'll want to make another clone of `sourcegraph/cody` in the root of `sourcegraph/sourcegraph` (it'll be `.gitignore`'d) and then follow [these steps](https://github.com/sourcegraph/cody/blob/main/doc/dev/library-development.md#local-development) to link the local versions of `cody-shared` and `cody-ui`.

Will reply more tomorrow! 😄

PR #782, opened 2023-08-22
Add a 'v' to the start of the version number in the update notice

toolmantim:
Adds a 'v' to the start of the version number in the update notice.

| Before | After |
| - | - |
| Screenshot 2023-08-22 at 12 28 20 pm | Screenshot 2023-08-22 at 12 27 59 pm |

## Test plan

- Inspected the UI, verified 'v' was added

### DIFF
--- a/vscode/webviews/Notices/VersionUpdatedNotice.tsx
+++ b/vscode/webviews/Notices/VersionUpdatedNotice.tsx
@@ -71,7 +71,7 @@ export const VersionUpdatedNotice: React.FunctionComponent<VersionUpdateNoticePr
     return (
         <Notice
             icon={<Icon />}
-            title={`Cody updated to ${majorMinorVersion}`}
+            title={`Cody updated to v${majorMinorVersion}`}
             linkHref={whatsNewURL(version)}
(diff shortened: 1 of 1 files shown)


vscode/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 
 ### Changed
 
+- Added a 'v' to the version number displayed in the update notice. [pull/782](https://github.com/sourcegraph/cody/pull/782)
```
camdencheek, line 15: This change is probably a bit smaller than needs to be logged in the changelog. Don't want to water down all the cool things being shipped, and I can't imagine there are loads of users who have been waiting for this one 🙂

camdencheek (approved):

vscode/CHANGELOG.md
toolmantim: If there was an award for the smallest diff with the largest changelog, I would keep it (the award, and the changelog). But you have a great point @camdencheek 😆 Removed! 👍🏼

PR #787, opened 2023-08-22
Agent: add cancelation support

olafurpg:
Previously, the agent didn't support canceling autocomplete requests. This PR adds support for cancellation through the new `$/cancelRequest` notification, which mirrors the structure of the same notification in LSP.

## Test plan
Manually tested with IntelliJ here https://github.com/sourcegraph/sourcegraph/pull/56119

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -162,13 +162,12 @@ export class Agent extends MessageHandler {
             })
             return null
         })
-        this.registerRequest('autocomplete/execute', async params => {
+        this.registerRequest('autocomplete/execute', async (params, token) => {
             const provider = await vscode_shim.completionProvider
             if (!provider) {
                 console.log('Completion provider is not initialized')
                 return { items: [] }
             }
-            const token = new vscode.CancellationTokenSource().token
             const document = this.workspace.getDocument(params.filePath)
             if (!document) {
                 console.log('No document found for file path', params.filePath, [...this.workspace.allFilePaths()])
--- a/vscode/src/testutils/mocks.ts
+++ b/vscode/src/testutils/mocks.ts
@@ -8,6 +8,7 @@
 import type {
     Disposable as VSCodeDisposable,
     InlineCompletionTriggerKind as VSCodeInlineCompletionTriggerKind,
+    Location as VSCodeLocation,
     Position as VSCodePosition,
     Range as VSCodeRange,
 } from 'vscode'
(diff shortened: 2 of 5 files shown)


vscode/src/testutils/mocks.ts
```diff
@@ -336,6 +337,21 @@ export class Position implements VSCodePosition {
     }
 }
 
+export class Location implements VSCodeLocation {
```
olafurpg, line 340: This change is unrelated to cancelation. There was a new esbuild linking error from new graph/codeintel related changes so I fixed it here while working on cancelation.

agent/src/agent.ts
```diff
@@ -162,13 +162,12 @@ export class Agent extends MessageHandler {
             })
             return null
         })
-        this.registerRequest('autocomplete/execute', async params => {
+        this.registerRequest('autocomplete/execute', async (params, token) => {
```
vdavid, line 165: Nit: Not sure about best practices, but I find the word "token" overused in programming in general. (Think access tokens, refresh tokens, bearer tokens, cancellation tokens, etc.) As a result, I usually err on the side of being specific in every case; I'd probably call this one "cancellationToken", but "cancelToken" is also fine, to match the rest of the code. WDYT? (I know you just moved this from a local variable, but it might be a good time for a rename.)

olafurpg: I agree in the general case but I think it's fine to use `token` here since we're wrapping that vscode API that uses `token` 

```
		/**
		 * Provides inline completion items for the given position and document.
		 * If inline completions are enabled, this method will be called whenever the user stopped typing.
		 * It will also be called when the user explicitly triggers inline completions or explicitly asks for the next or previous inline completion.
		 * In that case, all available inline completions should be returned.
		 * `context.triggerKind` can be used to distinguish between these scenarios.
		 *
		 * @param document The document inline completions are requested for.
		 * @param position The position inline completions are requested for.
		 * @param context A context object with additional information.
		 * @param token A cancellation token.
		 * @return An array of completion items or a thenable that resolves to an array of completion items.
		 */
		provideInlineCompletionItems(document: TextDocument, position: Position, context: InlineCompletionContext, token: CancellationToken): ProviderResult<InlineCompletionItem[] | InlineCompletionList>;
```

vdavid (approved):
We did the code review in a sync session and now it looks 👌

PR #788, opened 2023-08-22
Enable embeddings in chat in agent

pjlast:
Sets `useContext` to `'embeddings'` so that chat messages with agent can use embeddings.

## Test plan

Does this make sense?


chwarwick (approved):
Makes sense to me.

PR #792, opened 2023-08-22
Intelligent Cody: Include implementations of interface definitions

efritz:
This PR builds off of #692 to bring in non-definition relationships. See the [original Slack thread](https://sourcegraph.slack.com/archives/C03CSAER9LK/p1692729740458289) that shows definitions are not enough context for simple questions about code explanation.

This PR does a second pass over context snippets, replacing definitions directly to interfaces their implementations. This should give insight into the actual defined behavior of a function, rather than just its signature.

## Test plan

Tested new relationships are gathered by hand. Pre-merge we should find a strategy to cut down on the number of symbols are returned. We should have a strategy going forward before merging this (but shouldn't block on implementation; just initial discussions of what should work well). The context window is already getting **stuffed**.


vscode/src/chat/GraphContextProvider.ts
```diff
@@ -297,53 +314,100 @@ export const gatherDefinitions = async (
...
         }
+
+        // NOTE: deduplicating here will save duplicate queries that are _likely_ to point to the
+        // same definition, but we may be culling aggressively here for some edge cases. I don't
+        // currently think that these are likely to be make-or-break a quality response on any
+        // significant segment of real world questions, though.
+        for (const { symbolName, position } of dedupeWith(requestQueue, ({ symbolName }) => symbolName)) {
+            definitionMatches.push({
+                symbolName,
+                locations: getDefinitions(uri, position),
```
efritz, line 352: @efritz We'll want to sometimes do different relationships here

efritz:
Closing in favor of #983.

PR #793, opened 2023-08-23
Agent: implement `vscode.env.language`

tjdevries:
Previously, Cody Chat would regularly complain how it doesn't understand the language `undefined` when running through the agent. This PR fixes the issue by implementing `vscode.env.language` in the `vscode` shim.

## Test plan

- [x] Tests still pass
- [x] Cody doesn't reply yelling at me about undefined RFC language

### DIFF
--- a/lib/shared/src/prompt/prompt-mixin.ts
+++ b/lib/shared/src/prompt/prompt-mixin.ts
@@ -47,6 +47,10 @@ export class PromptMixin {
  * End with a new statement to redirect Cody to the next prompt. This prevents Cody from responding to the language prompt.
  */
 export function languagePromptMixin(languageCode: string): PromptMixin {
+    if (!languageCode || languageCode.length === 0) {
+        return new PromptMixin('(Reply as Cody, a coding assistant developed by Sourcegraph) ')
+    }
+
     return new PromptMixin(
         `(Reply as Cody, a coding assistant developed by Sourcegraph, in the language with RFC5646/ISO language code "${languageCode}") `
     )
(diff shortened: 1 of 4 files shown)


lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -47,6 +47,10 @@ export class PromptMixin {
  * End with a new statement to redirect Cody to the next prompt. This prevents Cody from responding to the language prompt.
  */
 export function languagePromptMixin(languageCode: string): PromptMixin {
+    if (!languageCode || languageCode.length === 0) {
+        return new PromptMixin('(Reply as Cody, a coding assistant developed by Sourcegraph) ')
+    }
+
     return new PromptMixin(
         `(Reply as Cody, a coding assistant developed by Sourcegraph, in the language with RFC5646/ISO language code "${languageCode}") `
     )
```
dominiccooney, line 56: Let's de-dup this a bit? Also `Boolean('')` is false so you can simplify a bit.

```suggestion
export function languagePromptMixin(languageCode: string): PromptMixin {
    const languagePrompt = languageCode ? `, in the language with RFC5646/ISO language code "${languageCode}"` : ''
    return new PromptMixin(
        `(Reply as Cody, a coding assistant developed by Sourcegraph${languagePrompt}) `
    )
```

dominiccooney (approved):
Good but let's de-dup the prompts, PromptMixin, simplify the condition, etc.

dominiccooney:
What context is Cody yelling at you about undefined languages in? It would be great to elaborate the test plan with specifics about what tests you're running, and what context to see if Cody yells in.

olafurpg (approved):
Very nice to finally fix this issue! I thought the root problem was `AgentTextDocument.languageId` (https://github.com/sourcegraph/cody/issues/785) but I didn't understand where `undefined` was coming from. This fix makes a lot more sense. LGTM 👍🏻

@dominiccooney I'm able to reproduce this issue by sending the chat message "hello" in IntelliJ

CleanShot 2023-08-23 at 08 50 24@2x

I ran a local build of IntelliJ with this PR and confirmed that Cody responds correctly to `hello`

CleanShot 2023-08-23 at 09 59 42@2x

PR #525, opened 2023-08-01
Autocomplete: Fix codebase context inference for embeddings

philipp-spiess:
Closes #446

In the current state, embedding results would only ever be added if a `cody.codebase` was configured in the VS Code settings. That was because we didn't account for two things:

- The version of codebase context we passed into the config is the _initial_ version that is only inferred from the config. We have the class `ContextProvider` which is actively tracking changes to the current active document and updates the config accordingly (this is necessary because you can have multiple git repos in a workspace and want to infer the remote from the proper repo)
- Additionally, ContextProvider was only initialized when the chat web view was loaded the first time meaning that context wasn't added if you don't have the sidebar visible (I’m not sure if this affects inline chat too cc @abeatrix @umpox @dominiccooney)

This PR changes the API to accept a `getCodebaseContext` callback that's only queried when the context is actually needed (so it can always resolve to the latest value) and makes sure that `ContextProvider` is initialized as part of the startup.

## Test plan

- Open a repo that is indexed but does not have a codebase config set up (I used `sourcegraph/sourcegraph` but removed the option from the config
- Observe that we now fetch embeddings and that the regular chat still seems to work.

Screenshot 2023-08-01 at 12 43 25
Screenshot 2023-08-01 at 12 42 24

PR #796, opened 2023-08-23
Remove CODY_FOCUS_SIDEBAR_ON_STARTUP

umpox:
## Description

Removes the auto focusing of Cody when we start local development.

The reasoning is that this is not a typical flow for any of our users. We are possibly building false assumptions and even missing bugs by doing this.

Also:
- Some logic only runs when the sidebar is open, always having the sidebar open from the start means we can miss some bugs when this code doesn't run (https://github.com/sourcegraph/cody/pull/525)
- The sidebar is no longer the obvious way to interact with Cody. We're possibly overemphasising parts of this because it's always visible to us.

## Test plan

Local dev change, tested by running local dev


taras-yemets (approved):
It makes sense to me! Thank you, Tom!

PR #799, opened 2023-08-23
Agent: fix bug when reading `.id` from null

olafurpg:
## Test plan
* Run IntelliJ
* Open settings
* Update endpoint URL
* Click "Apply" button
* Confirm there is no stack trace from the agent


vdavid (approved):
Thanks!

PR #806, opened 2023-08-24
vscode: Add docs about the purpose of different kinds of tests

dominiccooney:
I keep forgetting the difference between test/e2e and test/integration, so document that.

## Test plan

This is a doc-only change.


abeatrix (approved):
Nice! Thanks for adding this!
Do you think it'd be helpful to add links to these new docs in the CONTRIBUTING.md file as well?

PR #807, opened 2023-08-24
Agent: return `CompletionEvent` telemetry data from `autocomplete/execute`

olafurpg:
Previously, it was not possible for agent clients to know if a completion result was powered by embeddings. This PR fixes the problem by adding a new `completionEvent` property to the return type.

## Test plan
n/a

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -150,7 +150,7 @@ export class Agent extends MessageHandler {
             )
         )
 
-        this.registerRequest('recipes/execute', async data => {
+        this.registerRequest('recipes/execute', async (data, token) => {
             const client = await this.client
             if (!client) {
                 return null
@@ -189,7 +189,7 @@ export class Agent extends MessageHandler {
                         : result.items.flatMap(({ insertText, range }) =>
                               typeof insertText === 'string' && range !== undefined ? [{ insertText, range }] : []
                           )
-                return { items }
+                return { items, completionEvent: (result as any)?.completionEvent }
             } catch (error) {
                 console.log('autocomplete failed', error)
                 return { items: [] }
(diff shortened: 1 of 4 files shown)


agent/src/agent.ts
```diff
@@ -150,7 +150,7 @@ export class Agent extends MessageHandler {
             )
         )
 
-        this.registerRequest('recipes/execute', async data => {
+        this.registerRequest('recipes/execute', async (data, token) => {
```
vdavid, line 153: We don't seem to be using that cancellation token. Am I missing something? 🤔

vdavid (approved):
This PR looks like full circle to me, except the new token arg passed to recipes/execute which I didn't get, but it doesn't hurt either.

vscode/src/completions/vscodeInlineCompletionItemProvider.ts
```diff
@@ -218,15 +218,18 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
         } else {
             CompletionLogger.noResponse(result.logId)
         }
 
-        return {
-            items,
-        }
+        const completionResult: vscode.InlineCompletionList = { items }
+
+        ;(completionResult as any).completionEvent = event
```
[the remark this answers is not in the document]
philipp-spiess: I mean we do javascript without semicolons so really it's our fault :D

philipp-spiess (approved):
Not a big fan of the any cast but we can later make our own extended `InlineCompletionList`. Curious though why we need to leak these details? Is this because we log everything from the editor directly?

agent/src/agent.ts
vdavid: I don't think it is for 'recipes/execute'. I think you mean `autocomplete/execute`.

PR #816, opened 2023-08-25
Agent: Make recipes work through agent

vdavid:
Added titles to recipes. This was the last requirement, everything else was in place.

## Test plan

Tested with the JetBrains client in https://github.com/sourcegraph/sourcegraph/pull/56196


olafurpg (approved):
LGTM 👍🏻 This is perfect

PR #818, opened 2023-08-25
Autocomplete: Add Llama Code support to Fireworks provider

philipp-spiess:
Adds a new `cody.autocomplete.advanced.model` config option to overwrite the model used for the Fireworks adapter and adds support for the new LLama code model (Note: The FIM prompt is not properly tokenized yet so this isn't super useful yet, but I decided to merge the infra for this sooner so we can start playing around with more models sooner).

## Test plan

- Tested locally in VS Code
- You can get the access token here: https://sourcegraph.slack.com/archives/C05AGQYD528/p1690551781096999


vscode/src/completions/providers/unstable-fireworks.ts
```diff
@@ -10,20 +10,37 @@ import { Provider, ProviderConfig, ProviderOptions } from './provider'
...
     accessToken: null | string
+    model: null | string
 }
 
 const PROVIDER_IDENTIFIER = 'fireworks'
 const STOP_WORD = '<|endoftext|>'
 const CONTEXT_WINDOW_CHARS = 5000 // ~ 2000 token limit
 
+const MODEL_MAP = {
+    'starcoder-16b': 'accounts/fireworks/models/starcoder-16b-w8a16',
```
valerybugakov, line 21: If I need to find a model name myself, how can I do that? Let's leave a comment about that if the answer is not "Ask them in Slack".

valerybugakov (approved):

vscode/src/completions/providers/unstable-fireworks.ts
philipp-spiess: Unfortunately that's the only reliable way I found so far 🙈

PR #826, opened 2023-08-25
Autocomplete: Connect to Fireworks via Sourcegraph Server

philipp-spiess:
This PR changes the `unstable-fireworks` provider to no longer directly connect to Fireworks but instead use the updated code completion endpoint from https://github.com/sourcegraph/sourcegraph/pull/56120.

Note 1: These API calls will fail if SG is not run in dotcom mode.

Note 2: This will also mean that `unstable-firworks` is going to no longer work for dogfooders (I assume this affects very few people though)

Note 3: This will start using streaming the same way that anthropic does

## Test plan

- Set up the Cody Gateway with the `CODY_GATEWAY_FIREWORKS_ACCESS_TOKEN` on localhost
- Start `sg start dotcom` in the sg repo
- Connect Cody to the local instance
- Change the provider to be `unstable-fireworks`
- Observe requests made to the right backend

```
█ CompletionLogger:onComplete: {"type":"completion","status":"success","duration":574}  {
  "result": {
    "completion": "'Hello world');",
    "stopReason": "stop"
  },
  "params": {
    "messages": [
      {
        "speaker": "human",
        "text": "<fim_prefix>// Path: test.ts\nconsole.log(<fim_suffix>\n<fim_middle>"
      }
    ],
    "maxTokensToSample": 256,
    "temperature": 0.4,
    "topP": 0.95,
    "model": "fireworks/accounts/fireworks/models/starcoder-7b-w8a16"
  }
}
```


lib/shared/src/sourcegraph-api/completions/client.ts
```diff
@@ -60,162 +54,30 @@ export abstract class SourcegraphCompletionsClient {
         }
     }
 
-    public async complete(
```
philipp-spiess, line 63: I move this out from the shared completion client beacuse the code completion API `/.api/completions/code` will now have a very different use case than the normal chat based completion and can be running a code specific model.

As a result, these APIs should not be accessible to anyone except code completion users. 

There was only one user of this API outside of autocomplete and that was CLI. I changed CLI to use the streaming endpoint instead (which runs the same model as chat and is much more desired anyways).

cc @thenamankumar in case this could cause issues with Cody Web?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -507,3 +508,12 @@ function verifyResponseCode(response: Response): Response {
 
 class RepoNotFoundError extends Error {}
 export const isRepoNotFoundError = (value: unknown): value is RepoNotFoundError => value instanceof RepoNotFoundError
+
+export function isDotCom(serverEndpoint: string): boolean {
+    return new URL(serverEndpoint).origin === new URL(DOTCOM_URL).origin
+}
+
+export function isLocalhost(serverEndpoint: string): boolean {
```
valerybugakov, line 516: Leftover?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -507,3 +508,12 @@ function verifyResponseCode(response: Response): Response {
 
 class RepoNotFoundError extends Error {}
 export const isRepoNotFoundError = (value: unknown): value is RepoNotFoundError => value instanceof RepoNotFoundError
+
+export function isDotCom(serverEndpoint: string): boolean {
```
valerybugakov, line 512: We already have `isDotCom` and `DOTCOM_URL` in `vscode/src/chat/protocol.ts`. Does it make sense to keep once version?

lib/shared/src/sourcegraph-api/completions/client.ts
valerybugakov: Makes sense. Let's add this comment to the new code-completions client to clarify why we do not use the shared one. It might be confusing without this context.

valerybugakov (approved):
Nice, the new model is closer and closer!

lib/shared/src/sourcegraph-api/graphql/client.ts
philipp-spiess: Whoops, yeah, thanks!

lib/shared/src/sourcegraph-api/completions/client.ts
philipp-spiess: I did a grep and it doesn't seem to be used in Web either.

Issue #804, opened 2023-08-23
Locked out of commands after breaking the custom commands JSON file


iskyOS:
Cody version: v0.8.0
OS: MacOS Ventura 13.5

**Overview:**
After introducing a syntax error to the custom commands JSON file and then saving and closing the file, there isn't an obvious way to re-open the file to fix the issue. Commands stop working, and the button to access the JSON file doesn't work.

**Details:**
I added a custom command to Cody in VS Code and then went into the JSON file to remove it later. When I deleted it from the JSON file, I must have introduced an error in the file, because now I get an error every time I press the commands hotkey:
`Expected double-quoted property name in JSON at position 1682`

Here is the  popup error when I press `option`+`c`:
![CleanShot 2023-08-23 at 17 43 06](https://github.com/sourcegraph/cody/assets/13739780/e1191a2c-0091-44e2-9a3d-d1da1547f591)

The problem is that I can't find a way to re-open the custom commands JSON settings file to fix the issue, so I'm locked out of commands for now. The chat bar no longer suggests any commands when I enter a `/`:

![CleanShot 2023-08-23 at 17 48 22](https://github.com/sourcegraph/cody/assets/13739780/15d41901-97ab-4ace-9c71-8f6220d2c8bc)

In the Cody Settings dropdown, the "Custom Commands Settings" button does nothing when pressed:

![CleanShot 2023-08-23 at 17 38 21](https://github.com/sourcegraph/cody/assets/13739780/27d49bde-0e58-4e31-a6d2-06748c513cbe)

deepak2431:
I can reproduce this, too. I am on Mac, with the same version of Cody.

deepak2431:
@iskyOS To add more, the complete Cody app stopped working after this. I am not getting any response from Chat and can't use the Menu options too.

abeatrix:
@deepak2431 @iskyOS the PR above should fix the issue but for now, can you please try typing `/open user` in your Cody chatbox to open the user setting json file ? 🙇‍♀️

deepak2431:
@abeatrix Awesome, I will test the fix soon. Yeah using the command `/open user`, I am able to open the command settings json file.

iskyOS:
Thanks @abeatrix, that worked on my end! I was able to open the JSON file. Once I deleted the stray `,` that was causing the error, functionality went back to normal.

PR #827, opened 2023-08-25
fix: invalid json file for custom commands blocks command menu

abeatrix:
Close https://github.com/sourcegraph/cody/issues/804

- fix issue where invalid json file for custom commands blocks command menu
  - Wrap prompt loading in try/catch to handle errors
  - Return empty prompt map on error instead of throwing
  - Log errors loading prompts using debug logging
- Add editor-specific commands to the default commands list
- Create editor-specific commands with /fix
- Add alwaysShow option to createQuickPickItem
- Minor menu item cleaning

image

image

## Test plan

1. Remove a line from your current workspace config file for cody custom commands so you can see an inline error, ex:
image

2. You should still be able to open the command menu:
image

Before - you won't be able to open the command menu and see this error message instead: 
image

### DIFF
--- a/lib/shared/src/chat/prompts/index.ts
+++ b/lib/shared/src/chat/prompts/index.ts
@@ -2,8 +2,17 @@ import { Preamble } from '../preamble'
 
 import * as defaultPrompts from './default-prompts.json'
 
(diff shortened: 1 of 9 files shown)


taras-yemets (approved):
Thank you, @abeatrix!

PR #831, opened 2023-08-26
Fix autocomplete embeddings refresh condition

chwarwick:
Noticed that embeddings for autocomplete were not refreshing every 5 min.  
## Test plan
Added temporary logging to confirm 
before:
```
embeddings for /projects/sourcegraph/cmd/frontend/backend/app.go are -3 minutes old
```

after:
```
embeddings for /Users/chriswarwick/projects/sourcegraph/cmd/frontend/backend/app.go are 1 minutes old
```


vovakulikov (approved):
Very strange, difference calc should act in abs mode I think

chwarwick:
> Very strange, difference calc should act in abs mode I think

Aligns with the  docs 
https://date-fns.org/v2.30.0/docs/differenceInMinutes

Issue #809, opened 2023-08-24
Tree-sitter: Rank autocomplete suggestions based on syntactic validation


valerybugakov:
Use Tree-sitter to rank autocomplete suggestions based on syntactic validation.

- **Single-line completion**: Integrate the suggested code into the Tree-sitter parse tree and query for errors within the updated line range. Prioritize completions that are syntactically correct.
- **Multi-line completion**: Iteratively update the tree for each line received. Present the completion once a valid tree is established and the end of the block reached; otherwise, move completions to the end of the suggestions stack.

We need to integrate Tree-sitter and gather the basic experience of working with it in this codebase. Then, we can confidently apply it to [the other Tree-sitter-related enhancements](https://docs.google.com/document/d/1qjpYAYUyX-Yr9PTaq5KcwHOMRJtaVCVMqaHX2NT01nU/edit#heading=h.zbty0tj4mcek). This gives us the incremental adoption strategy where we can change the direction or opt out early if we bump into unforeseen issues. 

Discussed in [RFC: Enhancing VS Code Autocomplete with Tree-Sitter Integration](https://docs.google.com/document/d/1qjpYAYUyX-Yr9PTaq5KcwHOMRJtaVCVMqaHX2NT01nU/edit).

- [x] https://github.com/sourcegraph/cody/issues/849
- [x] https://github.com/sourcegraph/cody/issues/850
- [ ] https://github.com/sourcegraph/cody/issues/860

valerybugakov:
- https://github.com/sourcegraph/cody/issues/860 will be implemented as a part of https://github.com/sourcegraph/cody/issues/910

PR #837, opened 2023-08-28
Autocomplete: add naive suggestions ranking based on syntactic validity

valerybugakov:
## Context

This PR adds naive suggestions ranking based on syntactic validity powered by tree-sitter. In the post-processing logic, we insert the suggested code snippet into the document, parse it with tree-sitter, and query for `ERROR` nodes. Then in the ranking logic, suggestions without syntax errors are pulled forward.

- All the changes are behind the feature flag `cody.autocomplete.experimental.syntacticPostProcessing`. 
- This PR _does not_ leverage the incremental parsing capabilities tree-sitter has. They will be integrated with a follow-up PR. We will need to subscribe to all document changes and update the parse-tree on every change.
- Part of #809 

## Test plan

- Added unit test for the updated post-processing logic.
- Manually tested locally.


lib/shared/src/common/index.ts
```diff
@@ -52,3 +52,13 @@ export const escapeMarkdown = (text: string): string => {
     }
     return text.replaceAll('<', '&lt;').replaceAll('>', '&gt;')
 }
+
+/**
+ * Return a filtered version of the given array, de-duplicating items based on the given key function.
+ * The order of the filtered array is not guaranteed to be related to the input ordering.
+ */
+export const dedupeBy = <T>(items: T[], key: keyof T | ((item: T) => string)): T[] => [
```
valerybugakov, line 60: Moved to shared utilities for reusability

lib/shared/src/common/paths.ts
```diff
@@ -0,0 +1,19 @@
...
+    const resolvedPath = path.resolve(...args)
+
+    try {
+        return fs.realpathSync(resolvedPath)
+    } catch {
+        return resolvedPath
+    }
+}
+
+export const ROOT_PATH = resolveWithSymlink(__dirname, '../../../../')
```
valerybugakov, line 18: Used to avoid '../../../../' in path constants.

vscode/src/completions/processInlineCompletions.ts
```diff
@@ -23,14 +26,19 @@ export function processInlineCompletions(
...
+    const visibleResults = removeEmptyCompletions(processedCompletions)
 
     // Remove duplicate results
-    const uniqueResults = [...new Map(visibleResults.map(item => [item.insertText, item])).values()]
+    const uniqueResults = dedupeBy(visibleResults, 'insertText')
+
+    // Add parse errors info to completions
+    // Does nothing if `cody.autocomplete.experimental.syntacticPostProcessing` is not enabled.
+    // TODO: add explicit configuration check here when it's possible to avoid prop-drilling for config values.
+    const withParseInfo = addParseInfoToCompletions(uniqueResults, { document, position, docContext })
```
philipp-spiess, line 38: Out of curiosity: Is there a reason to do this outside of the rankCompletions function and patching the `Completion` type instead of passing it as arguments?

vscode/package.json
```diff
@@ -821,6 +821,12 @@
           "markdownDescription": "API key for Anthropic",
           "default": ""
         },
+        "cody.experimental.autocomplete": {
+          "order": 9,
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Experimental feature for internal use."
+        },
```
philipp-spiess, line 829: Leftover?

vscode/src/completions/document,
philipp-spiess, line 1: Huh :D

valerybugakov: WOOT

vscode/src/completions/processInlineCompletions.ts
valerybugakov: The plan is to use tree-sitter in `processItem` to rely on the parse tree for multi-line truncation, so I added it to the neutral place for now to better understand how to use it. Will need to move it around in follow-up PRs, and we would probably need to have an augmented completions type because parsing will be one of the first operations in the `processInlineCompletions` function.

vscode/CONTRIBUTING.md
```diff
@@ -12,8 +12,8 @@ Tip: Enable `cody.debug.enable` and `cody.debug.verbose` in VS Code settings dur
 - `src`: source code of the components for the extension host
 - `webviews`: source code of the extension sidebar webviews, built with Vite
 - `test`: [tests](test/README.md)
-- `dist`: build outputs from both webpack and vite
-- `resources`: everything in this directory will be move to the ./dist directory automatically during build time for easy packaging
+- `dist`: build outputs from both esbuild and vite
+- `resources`: everything in this directory will be moved to the ./dist directory automatically during build time for easy packaging
```
valerybugakov, line 16: Do you know if this is still true? @philipp-spiess @umpox @abeatrix

vscode/package.json
```diff
@@ -22,13 +22,13 @@
     "build:dev:desktop": "concurrently \"pnpm run -s _build:esbuild:desktop\" \"pnpm run -s _build:webviews --mode development\"",
     "build:dev:web": "concurrently \"pnpm run -s _build:esbuild:web\" \"pnpm run -s _build:webviews --mode development\"",
     "watch:build:dev:web": "concurrently \"pnpm run -s _build:esbuild:web --watch\" \"pnpm run -s _build:webviews --mode development --watch\"",
-    "_build:esbuild:desktop": "esbuild ./src/extension.node.ts --bundle --outfile=dist/extension.node.js --external:vscode --format=cjs --platform=node --sourcemap",
+    "_build:esbuild:desktop": "pnpm download-wasm && esbuild ./src/extension.node.ts --bundle --outfile=dist/extension.node.js --external:vscode --format=cjs --platform=node --sourcemap",
```
valerybugakov, line 25: We need to manually copy `tree-sitter.wasm` to the dist folder. It's fast because modules are already downloaded in post-install.

vscode/tsconfig.json
```diff
@@ -14,7 +14,6 @@
     "test/e2e",
     "webviews",
     "vite.config.ts",
-    "webpack.config.js",
```
valerybugakov, line 17: We do not have `webpack` in this repo.

vscode/CONTRIBUTING.md
philipp-spiess: Yeah I can't find anything for that anyways only that it is used as that static dir in the vite config which might copy it into dist? 

It doesn't have to be in `dist` to be in the resulting package though AFAIK 

https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/web/vite.config.ts?L10

vscode/package.json
```diff
@@ -22,13 +22,13 @@
...
     "_build:esbuild:web": "esbuild ./src/extension.web.ts --platform=browser --bundle --outfile=dist/extension.web.js --alias:path=path-browserify --external:vscode --define:process='{\"env\":{}}' --define:window=self --format=cjs --sourcemap",
     "_build:webviews": "vite -c webviews/vite.config.ts build",
     "lint": "pnpm run lint:js",
     "lint:js": "eslint --cache '**/*.[tj]s?(x)'",
     "release": "ts-node ./scripts/release.ts",
-    "download-wasm": "ts-node ./scripts/download-wasm-modules.ts",
+    "download-wasm": "ts-node-transpile-only ./scripts/download-wasm-modules.ts",
```
philipp-spiess, line 31: What does this change do?

vscode/src/completions/processInlineCompletions.ts
philipp-spiess: Makes sense. I think what I don't "like" (very subjective btw feel free to ignore) is that we extend the object type and then leak the results to the callers, too (which might start depending on it and then the lines become blurry).

vscode/src/completions/tree-sitter/grammars.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * List of all supported languages that we have grammars and
+ * lexems for. Note that enum values are copied from VSCode API,
+ * if we want to make it work with different editors we should
+ * enhance language detection.
+ *
+ * TODO: Decouple language detect to make it editor agnostic
+ */
+export enum SupportedLanguage {
+    JavaScript = 'javascript',
```
philipp-spiess, line 10: Not a fan of introducing new symbols for the languages when we already have one (e.g. now we need to remember the exact casing for all of the languages below which are also cased very inconsistent ➡️  `Php` vs `TSX`). Is there anything here that a literal string enum can't do?

vscode/src/completions/tree-sitter/parse-tree-cache.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+
+export async function initParser(document: TextDocument): Promise<void> {
+    const parseLanguage = getLanguageIfTreeSitterEnabled(document)
+
+    if (!parseLanguage) {
+        return
+    }
+
+    const parser = await createParser({ language: parseLanguage })
```
philipp-spiess, line 42: If this never yields before the first completion request is done, the parser just won't be adde and we will do the completion without treesitter support, is that correct? 

Asking because I see this function being called and not awaited in the code below. Might be good to add a quick comment if we do that on purpose

vscode/src/completions/processInlineCompletions.test.ts
```diff
@@ -46,3 +53,55 @@ describe('adjustRangeToOverwriteOverlappingCharacters', () => {
...
+            function hello() {
+                alert('hello world!')
+            }
+
+            function sort(█)
+        `,
+            ['array) {\nreturn array.sort()\n}', 'array) new\n']
+        )
+
+        expect(completions.map(c => c.hasParseErrors)).toEqual([false, true])
```
philipp-spiess, line 105: Nit: What do you think that instead of testing the implementation detail here, we instead yield two completions where only one is valid and assert on the behavior (that it is now ordered correctly) instead?

philipp-spiess (approved):
Awesomesauce

vscode/CONTRIBUTING.md
valerybugakov: Yeah, I was thinking specifically about the VS Code package.

vscode/package.json
valerybugakov: Disables type-checking, which makes the execution faster. We can afford that because we check Typescript types on CI anyways.

vscode/CONTRIBUTING.md
philipp-spiess: I think currently everything will be in the VS Code package even the source code 🙈 

Screenshot 2023-08-29 at 12 35 47

vscode/src/completions/tree-sitter/grammars.ts
valerybugakov: > we need to remember the exact casing
Typescript covers us in cases where we need to reference enum keys. 

> Is there anything here that a literal string enum can't do?
It doesn't make a practical difference for our use case so I'm happy with both approaches. We can migrate to that in a follow-up.

vscode/src/completions/tree-sitter/parse-tree-cache.ts
valerybugakov: Added additional comment [here](https://github.com/sourcegraph/cody/pull/837/commits/8f610a3ff9134f34701972d09e882bbc1a5c2ff9). This behavior will be updated in a follow-up with the incremental parsing.

vscode/src/completions/processInlineCompletions.test.ts
valerybugakov: As discussed in the call, I will shuffle the implementation details in multiple follow-up PRs. I agree that testing the desirable behavior of the whole post-processing pipeline is better, and I plan to introduce that once all the tree-sitter pieces are stable.

vscode/CONTRIBUTING.md
valerybugakov: fun investigation in DMs

vscode/src/completions/tree-sitter/grammars.ts
philipp-spiess: > Typescript covers us in cases where we need to reference enum keys.

It can do that for literal enums too btw 😬

PR #839, opened 2023-08-28
Autocomplete: Add feature flag to enable Fireworks as the default provider

philipp-spiess:
## Test plan

Tested that anthropic is still the default locally

```
█ AuthProvider:init:lastEndpoint: https://sourcegraph.com/
█ CodyCompletionProvider:initialized: provider: anthropic
```

### DIFF
--- a/vscode/src/completions/providers/createProvider.ts
+++ b/vscode/src/completions/providers/createProvider.ts
@@ -1,4 +1,5 @@
 import { Configuration } from '@sourcegraph/cody-shared/src/configuration'
+import { FeatureFlag, FeatureFlagProvider } from '@sourcegraph/cody-shared/src/experimentation/FeatureFlagProvider'
 import { SourcegraphNodeCompletionsClient } from '@sourcegraph/cody-shared/src/sourcegraph-api/completions/nodeClient'
 
 import { debug } from '../../log'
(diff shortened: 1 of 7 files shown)


vscode/src/completions/providers/createProvider.ts
```diff
@@ -90,3 +93,18 @@ export function createProviderConfig(
             return null
     }
 }
+
+async function resolveDefaultProvider(
```
valerybugakov, line 97: What do you think about evaluating the required feature flags right after creating the config (calling `getConfiguration`) and keeping their values in the module scope instead of passing the `featureFlagClient` around? 

The pattern we use here is similar to how we approach the VS Code plugin configuration values. I'm thinking about the improvement we discussed, which would remove config drilling and make it possible to refer to it as an always-up-to-date store. The idea can be applied to feature flags.

```ts
import config from 'whatever/config'

// always up to date, no need to manually handle reloads and
// pass the constructed config object down the call stack
use(config.autocompleteAdvancedProvider) 
```

Keeping all related logic in one place would make this transition easier and help us scale our experiments quickly with minimal code changes. Also, we can consider evaluating all the feature flags in one go on the extension activation. Happy to chat about it in more detail in Slack.

cc @taras-yemets

valerybugakov (approved):

vscode/src/completions/providers/createProvider.ts
philipp-spiess: @valerybugakov Evaluating the feature flags is when we o the bucketing, we have to do this as close to the actual behavior we want to test as possible or otherwise we have to replicate a lot of checks.

For this one for example, we only want people in the experiment that:

- Use autocomplete
- Use the default setup
- Are on dotcom

If we evaluate early, we have to replicate all these tests which would get quite messy IMO and we would likely still end up with more people in the experiment that actually use the features.

Unfortunately that means that these checks have to be async, at least for the first time a user when a user is not assigned to an experiment just yet.

philipp-spiess: (Let's continue this discussion outside of this PR, I don't think this needs to block this change)

PR #840, opened 2023-08-28
Autocomplete: Include model identifier with autocomplete logs

philipp-spiess:
This is going to be necessary only for the Fireworks provider where we will ship with the option to change models. For the A/B experiment, we need to be able to filter out users that have configured a different `model` from the on we want to test against.

This PR now includes the model next to the provider identifier in the autocomplete logs.

## Test plan

Check the outputs tab

```
CodyVSCodeExtension:completion:suggested {"multiline":false,"providerIdentifier":"anthropic","providerModel":"claude-instant-1","languageId":"typescript","type":"inline","multilineMode":null,"id":"qmq2pshbi","contextSummary":{"duration":0.20449999906122684},"source":"LastCandidate","lineCount":1,"charCount":27,"latency":989.3145000003278,"displayDuration":721.9014170002192,"read":true,"accepted":true,"otherCompletionProviderEnabled":true,"completionsStartedSinceLastSuggestion":4}
```

### DIFF
--- a/vscode/src/completions/providers/unstable-huggingface.ts
+++ b/vscode/src/completions/providers/unstable-huggingface.ts
@@ -12,6 +12,7 @@ interface UnstableHuggingFaceOptions {
     accessToken: null | string
 }
 
+const MODEL = 'starcoder'
 const PROVIDER_IDENTIFIER = 'huggingface'
 const STOP_WORD = '<|endoftext|>'
 const CONTEXT_WINDOW_CHARS = 3500 // ~ 1280 token limit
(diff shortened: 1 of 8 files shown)


vscode/src/completions/providers/unstable-huggingface.ts
```diff
@@ -121,16 +122,8 @@ export class UnstableHuggingFaceProvider extends Provider {
     }
 }
 
-function postProcess(content: string, multiline: boolean): string {
-    content = content.replace(STOP_WORD, '')
-
-    // The model might return multiple lines for single line completions because
-    // we are only able to specify a token limit.
-    if (!multiline && content.includes('\n')) {
```
valerybugakov, line 129: Why is this no longer relevant?

valerybugakov (approved):

vscode/src/completions/providers/unstable-huggingface.ts
philipp-spiess: Moved to shared post processing: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/completions/processInlineCompletions.ts?L60-66

PR #798, opened 2023-08-23
Update quick-pick labels

taras-yemets:
First iteration on https://github.com/sourcegraph/cody/issues/631
[Design](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?node-id=4560:14203&mode=dev)

This PR: 

1. Changes Cody command palette item labels: now `CodyPrompt.slashCommand` is rendered as a name, and `CodyPrompt.name` as a description. 
2. Adds `/ask` and `/fix` commands shortcuts. 
3. Submitting the `/ask` command with a non-empty input value submits the input value as a prompt in the current chat in the sidebar. If the input value is empty, a new quick-pick menu is rendered, asking for a prompt. If not empty, this quick pick submits its value as a prompt in the current chat in the sidebar.
4. Submitting the `/fix` command with a non-empty input value executes inline fixup. Submitting this command with the empty value triggers a new quick-pick menu, which executes an inline fixup on submit. 

The following changes (see @toolmantim [comment](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL?node-id=4560:14203&mode=dev#534226528) after chatting with @beyang) seem quite extensive and touch different parts of the codebase so they'd better be addressed in separate PRs:
- rename `/fix` command to `/edit`
- change commands interface in `lib/shared/src/chat/prompts/default-prompts.json` and `.vscode/cody.json` (slash command to be required and act as command identifier/name instead of command description) and update `CodyPrompt` interface accordingly
- update sidebar input to include the same slash commands as the main Cody quick-pick.

https://github.com/sourcegraph/cody/assets/25318659/e3064d17-3e5a-4b39-b4cc-863b096d2d10

## Test plan
Run this branch locally and check that the main Cody quick-pick menu matches the referenced design.

Issue #631, opened 2023-08-09
Clean Up Quickpick & Keyboard Shortcut Interactions


toolmantim:
The new Cody quickpick is the main in-editor experience for interacting with Cody, but a few things have happened since the initial design[^1]:
- `Refactor` commands were added to it as part of 0.6.5
- Internal testing and feedback that have outlined a number of problems and room for improvement
- Keyboard shortcuts are inconsistent (or missing)

We need to:
- Update the quickpick to handle refactors and better match the chat pane command interface
- Consolidate and clarify keyboard shortcuts
- Review editor contribution points (editor buttons, codelenses, status bars, etc)
- And overall, just ensure it's clear and simple to interact with Cody

Figma:
https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=4560-11184&mode=design&t=tINkTrd8OIADm9jT-4

- [x] Design
- [x] Reviewed & ready to go

[^1]: ![image](https://github.com/sourcegraph/cody/assets/153/ddcb4258-c593-4960-a72d-633298ff5f63)

toolmantim:
[Design](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=4560-11184&mode=design&t=tINkTrd8OIADm9jT-4) ready for feedback:

![image](https://github.com/sourcegraph/cody/assets/153/e1d64039-9293-4f66-9863-b5a92d9f0f53)

toolmantim:
Updated based on feedback:
![image](https://github.com/sourcegraph/cody/assets/153/c80a69ed-abf6-4f68-90cc-6eb3cbe48e36)

chenkc805:
Looks great! 

Can we add back code spells as a command in 0.8.0? It's a pretty highly used and liked feature both qualitatively and quantitatively

Quant: https://sourcegraph.looker.com/explore/sourcegraph/cody?qid=KuBRB9BGwmUcMVaOrogrZG&origin_space=562&toggle=fil,vis
Qual: https://sourcegraph.slack.com/archives/C05AGQYD528/p1691595505023259

toolmantim:
@chenkc805 yep, code smells are coming back.

Added to the Figma.

[...]

PR #841, opened 2023-08-28
make slash command required in command configs

taras-yemets:
[Design](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?node-id=4560:14212&mode=dev)

1. Changes commands format in `default-prompts.json` and `.vscode/cody.json`: 
    - slash command now is required and acts as a key in commands map, `slashCommand` field on a command is removed
    - command name previously used as a key is now used as a command description.
3. Updates the custom command creation process making the slash command input required and going first before the description and prompt inputs.
4. If spotted user or workspace Cody config with commands in "old" format, transforms them to the new one, saves updated config file and informs user about it.

https://github.com/sourcegraph/cody/assets/25318659/d2a2b770-1d27-4de5-8887-b781bd620a1a

## Test plan
CI passes.


vscode/src/custom-prompts/CustomPromptsStore.ts
```diff
@@ -111,13 +111,8 @@ export class CustomPromptsStore implements vscode.Disposable {
...
                     prompt.type = type
-                    // replace any '/' from the start
-                    const slashCommand = prompt.slashCommand?.replace(/^\//, '')
-                    if (slashCommand?.length) {
-                        prompt.slashCommand = `/${slashCommand}`
-                    }
+                    prompt.slashCommand = key.startsWith('/') ? key : '/' + key
                     this.myPromptsMap.set(key, prompt)
                 }
             }
```
taras-yemets, line 118: Open question: how to ensure these changes are backward compatible (users with the previous format of `cody.json` with command names as keys, users with recipes)?
cc: @abeatrix

lib/shared/src/chat/prompts/index.ts
```diff
@@ -57,11 +53,11 @@ export interface CodyPremade {
 }
 
 export interface CodyPrompt {
-    name?: string
+    description?: string
     prompt: string
     context?: CodyPromptContext
     type?: CodyPromptType
-    slashCommand?: string
+    slashCommand: string
```
taras-yemets, line 60: @abeatrix, can you think any additional side-effects of changing this interface not handled in this PR?

vscode/src/custom-prompts/CustomPromptsStore.ts
taras-yemets: A possible solution could be to apply new logic if the `description` field exists in a command. This would mean that slash command is used as a key.
If the `description` field is missing, store the command only if the `slashCommand` is defined. 
However, this will lead to partial use of `cody.json` content (in "old" format commands only with slash commands will be stored). Maybe we should rewrite (make slash command a key) the `cody.json` content when we spot the old format? WDYT?

lib/shared/src/chat/prompts/index.ts
abeatrix: since we run `build` every time we load the json file, I think as long as we make sure these are handled correctly there it should be fine.

Question: do we still need the "slashCommand" option if we are moving to use `key` instead? Maybe we can update this option to boolean, so if someone has this setting currently, we set it to true and display the command? Or are we displaying all commands with no option to not not display them?

vscode/src/custom-prompts/menus/CustomCommandBuilderMenu.ts
```diff
@@ -10,46 +10,67 @@ const NewCustomCommandConfigMenuOptions = {
...
+            placeHolder: 'e.g. "explain" to assign /explain for the "Explain Code" command',
             ignoreFocusOut: true,
             validateInput: (input: string) => {
                 if (!input) {
-                    return 'Command name cannot be empty. Please enter a unique name.'
+                    return 'Slash command cannot be empty. Please enter a unique identifier.'
                 }
                 if (commands.has(input)) {
-                    return 'A command with the same name exists. Please enter a different name.'
+                    return 'A command with the same identifier exists. Please enter a different command name.'
```
abeatrix, line 42:
```suggestion
                    return 'A command with the same keyword exists. Please enter a different command name.'
```

we are calling it keyword above, should we keep the same word?

vscode/src/custom-prompts/CustomPromptsStore.ts
```diff
@@ -111,14 +111,10 @@ export class CustomPromptsStore implements vscode.Disposable {
...
-                    prompt.name = key
                     prompt.type = type
-                    // replace any '/' from the start
-                    const slashCommand = prompt.slashCommand?.replace(/^\//, '')
-                    if (slashCommand?.length) {
-                        prompt.slashCommand = `/${slashCommand}`
-                    }
-                    this.myPromptsMap.set(key, prompt)
+                    // ensure there is only one leading forward slash
+                    prompt.slashCommand = key.replace(/^\/+/, '').replace(/^/, '/')
```
abeatrix, line 116: Do you think we should also update the `save` method to handle backward compatibility? 

I think we should also extract the first words from all the current keys and turn them into the prompt.slashCommand option here to auto-migrate the current commands without the "slashCommand" settings for our users? 

Maybe we can add a CTA to let them know how they can change the command keyword afterward. What do you think?

abeatrix (commented):
Left comments in line regarding migrating current file into next format

lib/shared/src/chat/prompts/index.ts
taras-yemets: Previously, the command `name` was a key. As there's no way to add a command without a key in JSON configs, I don't understand why `name` was optional (maybe for command-like objects). Here we make `slashCommand` required and ensure (in parse JSON configs functions) it's a string starting with a forward slash.

vscode/src/custom-prompts/menus/CustomCommandBuilderMenu.ts
taras-yemets: Makes sense, addressed in https://github.com/sourcegraph/cody/pull/841/commits/66f5b3963b81b053acf095e17347cc24863b7613

lib/shared/src/chat/prompts/index.ts
taras-yemets: My explanation is a bit confusing. When saying `key` I mean a key in the commands map in the Cody config file (like [this](https://sourcegraph.com/github.com/sourcegraph/cody@4ce5e547fdb46fc514e06e96addd208fa1a274d2/-/blob/lib/shared/src/chat/prompts/default-prompts.json?L11)).

vscode/src/custom-prompts/CustomPromptsStore.ts
taras-yemets: Thanks for pointing this out! Updated the PR. Now we transform "old" format commands to the new format, update the corresponding config files, and inform the user about the change.
This PR converts names to dash-case (not the first word in the command as it's not guaranteed to be unique) and uses them as slash commands.

taras-yemets:
@toolmantim, except for the overall implementation, I'm looking for your feedback on:
- the updated custom command creation flow;
- "old" format configs transformation and notification shown after.

abeatrix:
tried to run the branch but i'm getting:
image

Looks like issue with sentry?

taras-yemets:
> Looks like issue with sentry?
@abeatrix, it seems so. But it doesn't seem to be caused by te current PR changes.

toolmantim (approved):
Works great! I gave it a thorough test locally, and popped on some commits to update a bunch of the UI labels.

abeatrix (approved):
Works beautifully! Thanks @taras-yemets! 

I have a PR opened that would be affected by the changes in this PR. Since my changes are closer to the main branch as it is using the old format, can you review my PR and see if we can merge mine before yours? Or would it cause other conflicts that would not matter which one gets merged first? 

Here is the PR: https://github.com/sourcegraph/cody/pull/907

taras-yemets:
>I have a PR opened that would be affected by the changes in this PR. Since my changes are closer to the main branch as it is using the old format, can you review my PR and see if we can merge mine before yours? Or would it cause other conflicts that would not matter which one gets merged first?
>Here is the PR: https://github.com/sourcegraph/cody/pull/907

@abeatrix There is an ongoing discussion about whether the referenced PR will be included in the upcoming release, so I'm this one to have a bit more time to test it on `main`. I'm happy to help with resolving merge conflicts!

PR #847, opened 2023-08-29
Taras yemets/refactor fix to edit

taras-yemets:
## Test plan
Tested manually (screenshots attached).


valerybugakov (approved):

taras-yemets:
> For the name, we should make this match the casing used on the other commands, or ideally update all the names to match the Figma (sentence case). Could be a follow up?

@toolmantim, thanks for pointing it out! Will address in https://github.com/sourcegraph/cody/pull/841

PR #854, opened 2023-08-29
Intelligent Cody: Heuristically reduce get definition calls

efritz:
This PR reduces the amount of calls to VSCode's get definition provider by removing things that are likely to be within comments, and collapses same-named symbols into a single definition request. 

We make the simplifying assumption that in a single function that symbols with the same name refer to the same symbol, which doesn't hold everywhere but maybe enough places to be a useful heuristic.

## Test plan

Validated by hand; updated unit tests.


vscode/src/chat/GraphContextProvider.ts
```diff
@@ -304,20 +304,38 @@ export const gatherDefinitions = async (
...
                     })
                 }
             }
         }
+
+        // NOTE: deduplicating here will save duplicate queries that are _likely_ to point to the
+        // same definition, but we may be culling aggressively here for some edge cases. I don't
+        // currently think that these are likely to be make-or-break a quality response on any
+        // significant segment of real world questions, though.
+        for (const { symbolName, position } of dedupeWith(requestQueue, ({ symbolName }) => symbolName)) {
```
valerybugakov, line 333: This is also possible now with type-safety: 

```suggestion
        for (const { symbolName, position } of dedupeWith(requestQueue, 'symbolName')) {
```

valerybugakov (approved):

vscode/src/chat/GraphContextProvider.ts
efritz: WOAH. Can you explain this?

valerybugakov: I wanted to use this util too, so I moved it to the `shared` package and tweaked the implementation. The second argument now accepts `keyof T`, which is used if `typeof key !== 'function'`. 

https://github.com/sourcegraph/cody/blob/c121137003ec58c050b8eb16ccd6356693d04af6/lib/shared/src/common/index.ts#L60-L62

efritz: Updated!

PR #855, opened 2023-08-29
Intelligent Cody: Fix duplicate locations

efritz:
The current implementation of `locationKeyFn` is not very deterministic as we serialize the uri _object_ instead of the underlying path that can be serialized independently. If we use the `fsPath` instead we get a single string.

This was causing issues before where a location URI may have additional internal state at different levels of construction, making it appear when we're deduplicating objects that logically equivalent URIs sometimes had distinct object identity.

## Test plan

Validated by hand.

### DIFF
--- a/vscode/src/chat/GraphContextProvider.ts
+++ b/vscode/src/chat/GraphContextProvider.ts
@@ -460,7 +460,7 @@ const extractSnippets = (lines: string[], symbolRanges: vscode.Range[], targetRa
  * Returns a key unique to a given location for use with `dedupeWith`.
  */
 const locationKeyFn = (location: vscode.Location): string =>
-    `${location.uri}?L${location.range.start.line}:${location.range.start.character}`
+    `${location.uri.fsPath}?L${location.range.start.line}:${location.range.start.character}`
 
 /**
  * Convert a mapping from K -> Thenable<V> to a map of K -> V.


vscode/src/chat/GraphContextProvider.ts
```diff
@@ -460,7 +460,7 @@ const extractSnippets = (lines: string[], symbolRanges: vscode.Range[], targetRa
  * Returns a key unique to a given location for use with `dedupeWith`.
  */
 const locationKeyFn = (location: vscode.Location): string =>
-    `${location.uri}?L${location.range.start.line}:${location.range.start.character}`
+    `${location.uri.fsPath}?L${location.range.start.line}:${location.range.start.character}`
```
valerybugakov, line 463: > where a location URI may have additional internal state at different levels of construction

Could you share an example of issues and internal states that resulted in not very deterministic keys?

valerybugakov (approved):

abeatrix (approved):

vscode/src/chat/GraphContextProvider.ts
efritz: It may have been caught in testing and might not be a huge issue in practice, but the serialized URI had a `_state` field that was sometimes set and sometimes not depending on if the path of that object was previously constructed or not.

It's just internal memoized state that was getting in the way.

valerybugakov: I'm curious about that because I used `uri.toString()` as a cache key and do not want it to fail because of the internal state. At first glance, [sources](https://github.com/microsoft/vscode/blob/e7756c8870ee1df7360e6624e220534174039b02/src/vs/base/common/uri.ts#L371-L384) seem to suggest that string representation should be stable. I'm going to dig into it further.

dominiccooney: @valerybugakov using `toString()` as a key is right.

URIs in general have lots of "renderings" (escaping or not, default ports for protocols, punycode or not, ordering of querystrings, etc.) and lots of "applications" which makes notions of equality extremely weak.

Sounds like @efritz was serializing the actual fields of the object and ran into the equality problem with that? @efritz where's `_state`?

Using `fsPath` could be the road to ruin. For example if you have a content provider, its documents' URIs probably aren't going to have an `fsPath`. So later if we run Cody on these, their URIs could slip in here.

dominiccooney: @efritz I haven't read through this in detail but regarding already using `fsPath` a bunch, some of those uses look spurious. For example, if you have a URI in hand, just do `vscode.window.openTextDocument(uri)` instead of `vscode.window.openTextDocument(uri.fsPath)`. The former is simpler and works with more URIs.

With my vscode hat on, operations like `const uri = workspaceRootUri.with({ path: activeEditor.filePath })` look dubious because not all active editors are editing files within the workspace. The shared abstractions may have hemmed you in here, or maybe this is always safe by construction—but if that's the case, it isn't immediately obvious.

Documents which don't have filesystem URIs arise with extensions, like someone following [design advice for go to definition in standard libraries here](https://github.com/microsoft/vscode/issues/33823#issuecomment-327097698), things like that.

Yes, the Cody VScode extension is heavily hooked on the "files in the workspace" assumption. I just wanted to push back on entrenching it even further, particularly when it is probably easier not to.

PR #857, opened 2023-08-29
fix: reset current chat after clearing history

abeatrix:
fix: resetting transcript and session on clear history

Current clearing chat history does not reset current chat sessions, so if you use a command after clearing the chat session, you will still see the last chat session when a new session is expected

This PR includes changes to address the above issue:
- Reset the transcript when clearing chat history
- Call clearAndRestartSession() after clearing chat history to reset the current chat session
- Send updated history after saving to storage
- Add test to cover the updated behavior

## Test plan

1. click button to clear history
2. Use a command from your right click menu
3. The command output should show up in a new chat session without old convo


vscode/src/chat/MessageProvider.ts
```diff
@@ -147,6 +147,9 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         MessageProvider.chatHistory = {}
         MessageProvider.inputHistory = []
         await this.localStorage.removeChatHistory()
+        // Reset the current transcript
```
valerybugakov, line 150: Is there a test we can update to ensure this behavior is preserved in the future?

valerybugakov (approved):

vscode/src/chat/MessageProvider.ts
abeatrix: Good point. Will add a test before merging, thanks!

PR #862, opened 2023-08-30
Autocomplete: Add llama-v2-13b-code support

philipp-spiess:
Just adds a new supported model string. Cody Gateway change is not rolled out yet so this won't work _just_ yet.

## Test plan

👀


vscode/src/completions/providers/unstable-fireworks.ts
```diff
@@ -211,3 +217,16 @@ export function createProviderConfig(
         model,
     }
 }
+
+// We want to remove the same line suffix from a completion request since both StarCoder and Llama
+// code can't handle this correctly.
+function getSuffixAfterFirstNewline(suffix: string): string {
+    const firstNlInSuffix = suffix.indexOf('\n')
```
valerybugakov, line 224: I always trip when I read the middle of this abbreviation `tNlInS`.

valerybugakov (approved):

PR #865, opened 2023-08-30
Add section observer to track code navigation section changes

philipp-spiess:
This PR adds the first version of a section observer that maintains a list of all code sections (for now using the VS Code document symbol API but later using tree sitter). When a user navigates "into" sections (by moving their cursor), we use these hints to preload context for these ranges.

We can then synchronously query ranges if they have preloaded context when we make an autocomplete request and thus include graph context without paying the high latency costs.

This is very much WIP. I'll be keeping most tasks in this issue https://github.com/sourcegraph/cody/issues/940. All work is currently scoped behind an experimental feature flag: `cody.autocomplete.experimental.graphContext`.

## Test plan

Here's an example that mainly showcases the new debugging APIs that I have added. Note that the results aren't very good yet, see #940

Screenshot 2023-09-06 at 14 14 16


vscode/src/graph/section-observer.ts
```diff
@@ -0,0 +1,280 @@
...
+const TEN_MINUTES = 10 * ONE_MINUTE
+
+const NUM_OF_CHANGED_LINES_FOR_SECTION_RELOAD = 3
+
+/**
+ * Watches a document for changes and refreshes the sections if needed. Preloads the sections that
+ * the document is being modified by intersecting the cursor position with the document sections.
+ *
+ * TODO:
+ *  - [ ] GC?? How to make sure this does not grow unbound? Limit it to n total sections?
```
efritz, line 33: Do we need to keep more than the currently visible document? If not, then we can bound it only by the number of sections in a single document, which should be a reasonable amount for most non-psychopathic developers (or if we're editing generated code, which is less likely). If so then maybe an LRU cache based on documents would suffice?

Assuming that each entry we're holding in memory is fairly small so the number of entries here could be in the thousands without much issue (but I might be wrong).

philipp-spiess: Yeah I think you're right. We can add some logging if a document has an excessive number of sections and if that happens often, we can always revisit.

philipp-spiess:
@efritz @valerybugakov I still have to _use_ the graph context in the autocomplete prompt so there's a bit of plumbing still ahead (I want to _only_ use graph context if we have it) -- The rest is ready for an early review though. I’m sorry I haven't gotten it all cleaned up just yet, I promised I have it ready for your morning @valerybugakov 🙈

efritz (approved):

PR #868, opened 2023-08-30
Autocomplete: Enable keepalive agent for Node

philipp-spiess:
We found another small way to improve overall latencies: Making sure we keep the TCP connection alive and reducing the need for subsequent SSL/TLS handshakes. 

In Node, reusing TCP connections is not the default behavior (like it is for example on our Go backends). To fix this, we have to make sure we create a custom network agent that is configured to keep connections allive.

Since this is Node only, we are using a mutable ref objects to not run any of those code paths on the web branches.

c.f. https://sourcegraph.slack.com/archives/C05497E9MDW/p1693387804620389

## Test plan

Tested using the local completion CLI script

We found about a 5% latency improvement:

image


vscode/src/extension.node.ts
```diff
@@ -21,6 +22,8 @@ import { getRgPath } from './rg'
  * (Node.js/Electron).
  */
 export function activate(context: vscode.ExtensionContext): ExtensionApi {
+    initializeNetworkAgent()
```
philipp-spiess, line 25: @olafurpg Is this method being run when the agent is starting up? :)

vscode/src/fetch.ts
```diff
@@ -0,0 +1,26 @@
...
+ * environments.
+ *
+ * Agent is a mutable ref so that we can override it from `fetch.node.ts`
+ */
+export const agent: { current: ((url: URL) => Agent) | undefined } = { current: undefined }
+
+export function fetch(input: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    return isomorphicFetch(input, {
+        ...init,
+        agent: agent.current,
```
philipp-spiess, line 24: @cbart This pattern should also help you unblock https://github.com/sourcegraph/cody/pull/836

umpox:
Out of curiosity, what is your `"http.proxySupport"` value in VS Code? I looked into this for a moment yesterday but noticed that VS Code patches over the agent for the default proxySupport value (`override`). Related issue: https://github.com/microsoft/vscode/issues/173861

umpox (approved):

philipp-spiess:
@umpox Oh, interesting. Yeah I remember that I have had to deal with that before for the old SG VS Code extension. I have set it to `"http.proxySupport": "off",` right now and it did seem to try and use the agent (I validated that by returning `{}` and it throwed). Let me check what happens with the default values 🤔

philipp-spiess:
@umpox yeah i think there's nothing we can do when the default option is set. So this change will only have an impact for users that either have `"http.proxySupport": "off"` set or are in our controlled environments like Agent. 

Wow this is frustrating.

philipp-spiess:
I found a way 😐

umpox (approved):
👍

vscode/src/extension.node.ts
olafurpg: Thanks for the ping. Yes, the agent tries to be as invisible as possible 🙈

vscode/src/completions/client.ts
```diff
@@ -40,6 +42,8 @@ export function createClient(
             if (config.accessToken) {
                 headers.set('Authorization', `token ${config.accessToken}`)
             }
+            // c.f. https://github.com/microsoft/vscode/issues/173861
```
valerybugakov, line 45: Adding a one-liner explainer in addition to issue links is helpful to quickly understand the underlying problem without going to third-party sources.

```suggestion
            // Force HTTP connection reuse to reduce latency.
            // c.f. https://github.com/microsoft/vscode/issues/173861
```

valerybugakov (approved):
Tested it locally. Great find, @philipp-spiess!

vscode/src/fetch.node.ts
```diff
@@ -0,0 +1,62 @@
+import http from 'http'
+import https from 'https'
+
+import { agent } from './fetch'
+
+const nodeModules = '_VSCODE_NODE_MODULES'
```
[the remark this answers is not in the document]
philipp-spiess: Since we do access this from the context that has access to node modules, I think we can try using `require` instead.

vscode/src/fetch.node.ts
```diff
@@ -0,0 +1,62 @@
...
+const proxyAgentPath = '@vscode/proxy-agent/out/agent'
+const proxyAgent = 'PacProxyAgent'
```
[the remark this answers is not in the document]
philipp-spiess: Yes of course!

vscode/src/fetch.ts
```diff
@@ -0,0 +1,26 @@
+import type { Agent } from 'http'
+
+/**
+ * By hard-requiring isomorphic-fetch, we ensure that even in newer Node environments that include
+ * `fetch` by default, we still use the `node-fetch` polyfill and have access to the networking code
+ */
+import isomorphicFetch from 'isomorphic-fetch'
```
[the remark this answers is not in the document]
philipp-spiess: @valerybugakov Yeah I found that relying on the polyfilled global `fetch`, `node-fetch` was used in VS Code (older Node.js version) but the native fetch was used on Node 20 (CLI). By requiering fetch here, it would always use `node-fetch` in Node environment.s 

Wow I wasn't aware of undici this is amazing. I’m glad we see some improvements there. We'll have to still use some polyfills though because this code can also run inside a browser 🦭

PR #870, opened 2023-08-30
Agent: use esbuild CLI invocation instead of ESM script

sqs:
The esbuild `--alias:...=...` CLI flag suffices. A separate script with an esbuild plugin is not needed.

## Test plan

Run `cd agent && pnpm run build`. Observe no type errors during build.


vovakulikov (approved):

olafurpg (approved):
LGTM 👍🏻 Nice to remove so much unnecessary build config

olafurpg:
Ignoring the unrelated CI failure

PR #875, opened 2023-08-31
OpenAI  provider that sends requests though sourcegraph server

chwarwick:
Adds a new openai provider that routes requests though the sourcegraph instance.  Currently this reuses the prompt logic from the unstabl-azure-openai provider.  This should enable working though the sg backend and then any prompt updates can be handled separately.

## Test plan
ran locally with https://github.com/sourcegraph/sourcegraph/pull/56063


chwarwick:
@philipp-spiess do you have any suggestions here, from the testing I've done I have been getting reasonable single and multiline autocompletes most of the time.   

I'm not sure if I want to replace the existing provider or add another one since there may be some trialing the current one.

vscode/src/completions/providers/unstable-azure-openai.ts
```diff
@@ -1,125 +1,161 @@
...
-        logprobs: null
-    }[]
-    usage: {
-        completion_tokens: number
-        prompt_tokens: number
-        total_tokens: number
-    }
-}
+const PROVIDER_IDENTIFIER = 'unstable-azure-openai'
+const CONTEXT_WINDOW_CHARS = 2048
```
philipp-spiess, line 19: This is now defined twice: once here and once in the `createProvider` as a default.

vscode/src/completions/providers/unstable-azure-openai.ts
```diff
@@ -1,125 +1,161 @@
...
-            return new UnstableAzureOpenAIProvider(options, unstableAzureOpenAIOptions)
+            return new UnstableAzureOpenAIProvider(options, { ...unstableAzureOpenAIOptions })
         },
-        maximumContextCharacters: contextWindowChars,
+        maximumContextCharacters: CONTEXT_WINDOW_CHARS,
         enableExtendedMultilineTriggers: false,
         identifier: PROVIDER_IDENTIFIER,
         supportsInfilling: false,
-        model: 'gpt-35-turbo',
+        model: 'azure-openai',
```
philipp-spiess, line 159: What's the model that is running? Do we know that on the client? This logging is mostly for Fireworks where we can run multiple models though so don't worry about it.

vscode/src/completions/providers/unstable-azure-openai.ts
```diff
@@ -1,125 +1,161 @@
...
+        snippets: ContextSnippet[],
+        tracer?: CompletionProviderTracer
+    ): Promise<Completion[]> {
+        const prompt = this.createPrompt(snippets)
+
+        const args: CompletionParameters = {
+            messages: [{ speaker: 'human', text: prompt }],
+            maxTokensToSample: this.options.multiline ? MAX_RESPONSE_TOKENS : 50,
+            temperature: 0.4,
+            topP: 0.95,
```
philipp-spiess, line 74: The values here have changed a lot, was this on purpose?

philipp-spiess (approved):
Looks good overall - Pretty much like what we do for Fireworks now too. I don't really have any advice, if you think it would make sense to have a separate provider while we're transitioning, I don't see a problem with that either

vscode/src/completions/providers/unstable-azure-openai.ts
chwarwick: Not intentional I copied a bunch over from the fireworks provider based on a comment you made on the original pr about making the prompt similar and didn't notice those values were different.

chwarwick:
> Looks good overall - Pretty much like what we do for Fireworks now too. I don't really have any advice, if you think it would make sense to have a separate provider while we're transitioning, I don't see a problem with that either

Thank you.  Any concerns if I call this the `unstable-openai` provider I don't see a reason it would be limited to just Azure OpenAI once I apply the backend changes to the non-azure version of the server side pieces.

Issue #874, opened 2023-08-31
dx: introduce a way to subscribe to individual VS Code extension config keys


valerybugakov:
This issue aims to simplify the way we define and access VS Code extension settings:
- Avoid prop-drilling of the monolithic configuration object from the activation function to consumers by relying on module scope imports instead.
- Enable subscription to individual configuration keys, which doesn't require manual maintenance to keep values up to date.

These changes should improve the development velocity by significantly reducing the amount of boilerplate work required to introduce and consume VS Code settings. 

Touched on pain points in [the PR discussion](https://github.com/sourcegraph/cody/pull/636#discussion_r1298003228)
Discussed the idea briefly in the [RFC 796 WIP: Cody VScode Feature Flags](https://docs.google.com/document/d/1SML0hZXmky28iwss1wvFS53DdDtl7PLZ2O8tlJzDams/edit#bookmark=id.6yoqwkywt73b).

PR #876, opened 2023-08-31
VS Code: use singleton storage providers

valerybugakov:
## Context

- This PR migrates `localStorage` and `secretStorage` to a singleton pattern. Instead of initializing these entities upon the extension activation, they are initialized in the module scope and can be imported anywhere in the VS Code extension code. This removes a lot of boilerplate required for prop-drilling these entities from the extension init function to consumers. See the example of the boilerplate removal here: https://github.com/sourcegraph/cody/commit/f445839ee39a44b11fdedcd2be0c0c883357edde.
- Touched on similar pain points in [the PR discussion](https://github.com/sourcegraph/cody/pull/636#discussion_r1298003228)
- Part of #874
- Planned follow-up: migrate other `localStorage` and `secretStorage` consumers to importing singletons instead of using them as arguments. Same as done [here](https://github.com/sourcegraph/cody/commit/f445839ee39a44b11fdedcd2be0c0c883357edde).

The end goal of this effort (as stated in https://github.com/sourcegraph/cody/issues/874) is to decouple `getConfiguration` from the other entities in the codebase and introduce atomic and reactive getting for config values.

## Test plan

CI


philipp-spiess (approved):
Noice

PR #878, opened 2023-08-31
Autocomplete: Tweak Llama code params and add 7b model support

philipp-spiess:
Fireworks are now also hosting the 7b model for us to try out. Server PR here: https://github.com/sourcegraph/sourcegraph/pull/56307

This PR tweaks some params and updates all existing completion benchmark snapshots for a better comparison.

Looking at the test suite, I still prefer the StarCoder 7b results. I did notice that the temperature for Code Llama needs to be a lot lower otherwise it is easily getting off track and results will appear too smart 

## Test plan

This just updates some configs and snapshots

To visualize the changes:

- Check out this branch
- `cd completions-review-tool/`
- `pnpm run dev`

Screenshot 2023-08-31 at 11 28 44


vscode/src/completions/providers/unstable-fireworks.ts
```diff
@@ -230,3 +231,17 @@ function getSuffixAfterFirstNewline(suffix: string): string {
...
     return suffix.slice(suffix.indexOf('\n'))
 }
+
+function getModelConfig(model: string): { temperature: number; topP: number } {
+    if (model.startsWith('llama-code')) {
+        return {
+            temperature: 0.3,
+            topP: 0.95,
+        }
+    }
```
philipp-spiess, line 241: cc @sqs @olafurpg I’m curious what values you have used for your Llama code experiments? I found a lower temperature is necessary because otherwise the model goes off track easily.

olafurpg: I've only used codellama locally via ollama. Looking forward to trying it out here.

valerybugakov: I used lower temperatures for my local testing. Anecdotally, 0.2 gives more desirable results (fewer comments or mock implementations compared to higher temperatures).

valerybugakov (approved):

PR #881, opened 2023-08-31
Autocomplete: Add tracing for network requests

philipp-spiess:
This PR adds request tracing for autocomplete behind a feature flag. The plan is to enable tracing for 5% of the population so we can start collecting traces.

## Test plan

- Connect to localhost and throw in the `codecompletion.go` file
- Enable the FF for your user
- Observe that a trace ID is included in the error ping

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:error {"message":"error parsing response CodeCompletionResponse: SyntaxError: Unexpected token 'U', \"Unsupporte\"... is not valid JSON, response text: Unsupported custom model\n","traceId":"5a672ce14f16225b53e46f0afc6e02e9","count":1}
```

### DIFF
--- a/lib/shared/src/experimentation/FeatureFlagProvider.ts
+++ b/lib/shared/src/experimentation/FeatureFlagProvider.ts
@@ -7,6 +7,7 @@ export enum FeatureFlag {
     // product code
     TestFlagDoNotUse = 'test-flag-do-not-use',
 
+    CodyAutocompleteTracing = 'cody-autocomplete-tracing',
     CodyAutocompleteIncreasedDebounceTimeEnabled = 'cody-autocomplete-increased-debounce-time-enabled',
     CodyAutocompleteStreamingResponse = 'cody-autocomplete-streaming-response',
     CodyAutocompleteDefaultProviderFireworks = 'cody-autocomplete-default-provider-fireworks',
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -9,6 +9,24 @@ export class RateLimitError extends Error {
     }
 }
 
+export function isRateLimitError(error: Error): error is RateLimitError {
+    return error instanceof RateLimitError
+}
+
+export class NetworkError extends Error {
+    constructor(
+        message: string,
+        public traceId: string | undefined
+    ) {
+        super(message)
+        Object.setPrototypeOf(this, NetworkError.prototype)
+    }
+}
+
+export function isNetworkError(error: Error): error is NetworkError {
+    return error instanceof NetworkError
+}
+
 export function isAbortError(error: Error): boolean {
     return (
         // http module
@@ -17,8 +35,4 @@ export function isAbortError(error: Error): boolean {
         error.message.includes('The operation was aborted') ||
         error.message.includes('The user aborted a request')
     )
-}
-
-export function isRateLimitError(error: Error): boolean {
-    return error instanceof RateLimitError
 }
--- a/vscode/src/completions/client.ts
+++ b/vscode/src/completions/client.ts
@@ -28,8 +28,8 @@ export function createClient(
     featureFlagProvider?: FeatureFlagProvider,
     logger?: CompletionLogger
 ): CodeCompletionsClient {
-    function getCodeCompletionsEndpoint(): string {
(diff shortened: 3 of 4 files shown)


lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -7,6 +7,7 @@ export enum FeatureFlag {
     // product code
     TestFlagDoNotUse = 'test-flag-do-not-use',
 
+    CodyAutocompleteTracing = 'cody-autocomplete-tracing',
```
philipp-spiess, line 10: I creates this flag on dotcom for 5% of users

eseliger (approved):
cc @bobheadxi since you know MUCH more about tracing

vscode/src/completions/client.ts
```diff
@@ -28,8 +28,8 @@ export function createClient(
     featureFlagProvider?: FeatureFlagProvider,
     logger?: CompletionLogger
 ): CodeCompletionsClient {
-    function getCodeCompletionsEndpoint(): string {
-        return new URL('/.api/completions/code', config.serverEndpoint).href
+    function getCodeCompletionsEndpoint(tracing?: boolean): string {
+        return new URL('/.api/completions/code' + (tracing ? '?trace=1' : ''), config.serverEndpoint).href
```
bobheadxi, line 32: have you confirmed this works? I was under the impression that `?trace=1` only works in some web app areas, and that to get tracing via the API we use `X-Sourcegraph-Should-Trace` header https://sourcegraph.sourcegraph.com/search?q=context:sourcegraph+X-Sourcegraph-Should-Trace&groupBy=repo

vscode/src/completions/client.ts
```diff
@@ -41,18 +41,24 @@ export function createClient(
...
-            const isFeatureFlagEnabled = featureFlagProvider
-                ? await featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStreamingResponse)
-                : false
-            const enableStreaming = !!isNode && isFeatureFlagEnabled
 
-            const response = await fetch(getCodeCompletionsEndpoint(), {
+            const enableStreaming = !!isNode && !!streamingResponseFlagEnable
+
+            const url = getCodeCompletionsEndpoint(tracingFlagEnabled)
+            const response: Response = await fetch(url, {
```
bobheadxi, line 61: Do we always start a span in the web app as well, to make the backend trace that gets generated link up usefully in client?

lib/shared/src/sourcegraph-api/errors.ts
```diff
@@ -9,6 +9,24 @@ export class RateLimitError extends Error {
...
+
+export class NetworkError extends Error {
+    constructor(
+        message: string,
+        public traceId: string | undefined
+    ) {
+        super(message)
+        Object.setPrototypeOf(this, NetworkError.prototype)
+    }
+}
```
bobheadxi, line 24: awesome to collect this here 😁

vscode/src/completions/client.ts
philipp-spiess: @bobheadxi Yeah it worked but I can use the http header too :)

vscode/src/completions/client.ts
philipp-spiess: We don't have any tracing span infrastructure for VS Code yet but I can technically use whatever we have on the web if you point me to it? Caveat is that it can neither _exclusively_ use Node.js APIs or Web APIs, so if it relies on async local storage, this might be a bit tricky to adapt, but it would be AWESOME if we can join these somehow and do have that data.

philipp-spiess: @bobheadxi Would you be interested in helping me get this set up? Now that you gave me a taste of it, I want the full thing 🙈 I now want E2E traces for completions spanning over VS Code, the SG instance, and Gateway 🥹

Also I think let's merge this PR that enables tracing on the backend first and then we can extend it with frontend spans, how does that sound?

valerybugakov: @bobheadxi, can we use the same OpenTelemetry infra we used for the web application here to get all the benefits of the distributed tracing and OTel collector we already have on the cloud instance?

philipp-spiess: Moving this to a new issue: https://github.com/sourcegraph/cody/issues/901

PR #882, opened 2023-08-31
VSCode: Add Sentry to extension host process

philipp-spiess:
This PR adds a simple setup for Sentry to be used inside the extension host thread. It does not set up any default integration and we will instead need to wrap top-level entry points with calls to `Setnry.captureException`.

This also wires up autocomplete errors to be routed to Sentry so we have an easy service in place to understand the reliability of the product.

Note: When the user is connected to an instance other than dotcom, this does not send any events. 

## Test plan

- Add a throw case into the anthropic handler
- Force the env to be production
- See that the error is logged to Sentry

Screenshot 2023-08-31 at 14 03 13

I have also changed the backend to S2 and observed that no error is logged.

### DIFF
--- a/vscode/src/services/sentry/sentry.ts
+++ b/vscode/src/services/sentry/sentry.ts
@@ -29,8 +29,8 @@ export abstract class SentryService {
             environment: this.config.isRunningInsideAgent
                 ? 'agent'
                 : typeof process !== 'undefined'
-                ? 'vscode-web'
-                : 'vscode-node',
+                ? 'vscode-node'
+                : 'vscode-web',
 
             // In dev mode, have Sentry log extended debug information to the console.
             debug: !isProd,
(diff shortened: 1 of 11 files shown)


vscode/src/services/sentry/sentry.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public onConfigurationChange(newConfig: Pick<Configuration, 'serverEndpoint'>): void {
+        this.config = newConfig
+        this.prepareReconfigure()
+    }
+
+    private prepareReconfigure(): void {
+        const isProd = process.env.NODE_ENV === 'production'
+        const options: SentryOptions = {
```
valerybugakov, line 24: Having a `release` field set in our web application was very helpful. Do you think we can do it here, too?
https://docs.sentry.io/platforms/javascript/configuration/options/#release

valerybugakov (approved):
Nice!

vscode/src/services/sentry/sentry.ts
philipp-spiess: @valerybugakov Sure! I was just not sure what to put in here? Just the version number? Maybe we can also use this to differentiate vscode and agent?

philipp-spiess: I'll use environment for agent vs vscode :)

PR #902, opened 2023-09-01
make sidebar chat commands match main quick pick

taras-yemets:
[Design](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=4313-152881&mode=design&t=5MyjxGTdfMdZ25Ca-0)

1. Adds separators to the sidebar chat command groups.
2. Adds reset command.
3. If the input value starts with a known slash command, the corresponding item remains visible in the commands dropdown.
4. When a user starts typing a known slash command and hits "Enter", the command is autocompleted in the input field. The next time a user hits "Enter" this command is submitted. 

https://github.com/sourcegraph/cody/assets/25318659/f1b00c32-d00f-4cee-ab72-9957da8d9098

## Test plan
CI passes.
Tested manually (video attached).


abeatrix (commented):
image
image
Should the /reset command be added to the quick pick menu too, to be consistent? @toolmantim

toolmantim:
> Should the /reset command be added to the quick pick menu too, to be consistent?

@abeatrix no let’s leave that off as it’s more specific to the chat view.

toolmantim (approved):
Pushed a few small tweaks. It would be good to add `/ask` in there too.

taras-yemets:
> Pushed a few small tweaks. It would be good to add `/ask` in there too.

@toolmantim, "/ask" submits a question is a sidebar chat. The user has already chat input focused, they need to type their question and submit. Does having "/ask" command add additional value in this case?

toolmantim:
> @toolmantim, "/ask" submits a question is a sidebar chat. The user has already chat input focused, they need to type their question and submit. Does having "/ask" command add additional value in this case?

@taras-yemets there's a similar argument in the command quickpick too, in that they can ask a question just by typing and hitting enter.

But the real value is in consistency & muscle memory.

People can get used to mashing the same keys on the keyboard and having the same commands available, whether it's the chat window or the command palette. The goal is that people don't have to keep a mental map of what commands are available in each of the places. Even though it does feel a little redundant or repetitive in isolation.

PR #904, opened 2023-09-01
remove experimental plugins feature

sqs:
This is not being used right now.

## Test plan

n/a; removal


valerybugakov (approved):
🧹

sqs:
Hmm, lint is ooming.

valerybugakov:
@sqs, the linting issue was fixed recently by bumping the node.js memory limit.
I re-ran the failed actions (it looks like a GitHub outage).

Issue #619, opened 2023-08-09
Improve Generate Unit Tests


abeatrix:

rakeshjosh2003:
For one of the clients we recently observed that unit tests do not generate for files that rely on parameters from a construct in a base file i.e files that have a dependency on other files.

[Slack thread](https://sourcegraph.slack.com/archives/C058H9W5VCK/p1690176424323429)

[Customer](https://sourcegraph2020.lightning.force.com/lightning/r/Account/0013t00001Xiai9AAB/view)

abeatrix:
@rakeshjosh2003 can you confirm which version the client is on? 

When you say `unit tests do not generate`,  do you mean Cody did not response with any unit tests? Or did Cody generate unit tests, but they do not work due to dependency missing? 

Would you happen to have an example you can show me since the slack thread did not have any?

rakeshjosh2003:
Hey @abeatrix ,

It is able to generate unit tests but not able to specify the parameters in those tests for the files with dependencies.

abeatrix:
@rakeshjosh2003 do you have an example of what the failed tests look like? just to make sure we are on the same page, thanks!

rakeshjosh2003:
Hey @abeatrix , I have contacted the customer to share some examples as it looks scenario based. Will update once I hear back, thank you!

abeatrix:
End of sprint update: https://github.com/sourcegraph/cody/pull/907

- Improved context and prompt for the unit tests command that generates tests with higher quality consistently.
- The way the tests are presented could be clearer and need follow-up work. Going to work with @toolmantim on exploring options to see what's the best way to handle this

PR #907, opened 2023-09-03
improve generate unit tests command output quality

abeatrix:
Close https://github.com/sourcegraph/cody/issues/619

This PR contains changes that improve the current generate unit tests command:
- update the current unit test prompt to give it detailed directions on how to produce unit tests with higher quality consistently ([ref anthropic docs](https://docs.anthropic.com/claude/docs/claude-says-it-cant-do-something#give-explanatory-instructions))
- update context fetching process for tests to filter other test files like e2e and integration tests when generating unit-test
- include import statements from current files as context to avoid cases where cody fails to include import statements in tests
- update language used in prompt to let Cody know all files come from the same codebase. I noticed Cody currently does not know the shared context is from files, and certainly doe snot know the file path we shared with each context comes from a file that belongs to the same codebase. It just doesn't make that connection until we spell it out for them
- did not make any changes that would change the behavior of the current test command

## Before and After

The tests will not always be perfect, but with the changes mentioned above, the improved tests should do the following consistently when possible:
- use the correct testing frameworks
- include imports (common issues with the current version where imports are incorrectly added or often missing)
- include more relevant context

### before

In the video below, you can see the test imported the "chai" testing library for the tests, instead of "vitest" which is the testing framework that is used by other test files in the project:

https://github.com/sourcegraph/cody/assets/68532117/d5e82070-6269-4d5b-bdc6-d3cc4173dc18

You can also see from the context list that it includes test files from e2e test and integration tests that are not relevant to unit tests:

image

### after

In the improved version,  you can see it import the correct testing framework "vitest", using more relevant test files as context:

image

https://github.com/sourcegraph/cody/assets/68532117/75a72840-3657-4f27-873d-95a12ad9d1ec

#### before - go

use 'NewMockCache' that does not exist

image

#### after - go

image

## Follow-up

- [ ] a way to add generated tests easily to a test file
- [ ]  consolidate vscode-context and move it out of shared lib 

## Test plan

- Start Cody from this branch
- Select a function in a file and run the `Generate Unit Tests` command
- Compare the test output with the output from the current `Generate Unit Tests` command

Try running other commands. The changes in this should not change the behavior of other commands.

https://github.com/sourcegraph/cody/assets/68532117/0779ce8a-713a-40e4-a92e-6fc612143798

image

It doesn't always produce the perfect tests, but it generates better and higher quality tests consistently compared to the older version of the command. Works consistently with other languages as well. More demo videos:
typescript: https://www.loom.com/share/3489f9ae811b41d08ca0329489580543?sid=379b3122-3ea7-4b1e-bed2-12721e1f09b6
go: https://www.loom.com/share/d1514db896c840989e05f7c74fd30b81?sid=60f8d4d6-2b6c-42d3-b016-b347666893e9
go 2: https://www.loom.com/share/952caef2067a4d748bcd10eef8a60800?sid=04ca3f58-e2c8-474f-86ee-08af03f00689
scala/java: https://www.loom.com/share/d28b1089ac8a4d9ea95816c3e476ec35?sid=adabd913-3811-4009-9391-ccd1406fe91d
python: https://www.loom.com/share/7a788b389ce84a4b85fbe2d31f4a0ab4?sid=11d9314d-31c3-4530-be6a-b4e9faa5de0b


vscode/src/chat/MessageProvider.ts
```diff
@@ -206,6 +206,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                     // TODO(keegancsmith) guardrails may be slow, we need to make this async update the interaction.
                     displayText = await this.guardrailsAnnotateAttributions(displayText)
                     this.transcript.addAssistantResponse(text || '', displayText)
+                    void this.editor.controllers.command?.onCompletion(text)
```
abeatrix, line 209: this sends the response back to the command controller to create the test files when needed

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,133 @@ export class CustomPrompt implements Recipe {
...
-            contextMessages.push(...getEditorSelectionContext(selection))
-        }
-
-        // Command output
-        if (isContextRequired.command?.length && commandOutput) {
-            contextMessages.push(...getTerminalOutputContext(commandOutput))
+        const workspaceRootUri = editor.getWorkspaceRootUri()
+        const isUnitTestRequest = extractTestType(text) === 'unit'
+
+        switch (true) {
```
abeatrix, line 110: from if functions to switch

lib/shared/src/chat/prompts/templates.ts
```diff
@@ -1,18 +1,21 @@
-const instruction_prompt = `Please follow these rules when answering my question:
+const instruction_prompt = `Please follow these rules when answering my questions:
 - Your answers and suggestions should based on the shared context only.
 - Do not suggest anything that would break the working code.
-- Do not make assumptions or fabricating additional details.
+- Do not make any assumptions about the code and file names or any misleading information.
```
abeatrix, line 4: this helps prevent Cody from "hallucinating"

abeatrix:
cc @kalanchan regarding the next release

toolmantim:
@abeatrix I’ll be writing this one!

Do you have a description of how the generate test output is improved by these changes? (E.g. Before and after examples? Cases that didn’t work well before, that do now?)

abeatrix:
@toolmantim Just added more demo videos at the end of the PR description and added a Before and After section with example test output. let me know if you need more info :)

toolmantim (commented):
Super cool. I gave it a go w/ some ts and go files in various projects! The results I got were a little mixed… even though some output is really promising.

I spotted in some cases the chat output contained this for existing tests:

> I am writing these unit tests in the existing test file search_test.go which already contains tests for the search package.

Which makes it sound like Cody updated the existing file, which it doesn't (it opens the temporary file)

Also this happened when there was no existing file (one was created):

> I am not importing any new packages since this test file already exists and follows the patterns from the existing tests.

Screenshot 2023-09-02 at 10 46 49 pm

I feel like we might need some wider testing this week, rather than rushing it into the release. And give @beyang & @csells a chance to see how it's going too.

toolmantim:
> Just added more demo videos at the end of the PR description and added a Before and After section with example test output. let me know if you need more info :)

Just saw this after the review sorry! 😅

abeatrix:
> > Just added more demo videos at the end of the PR description and added a Before and After section with example test output. let me know if you need more info :)
> 
> 
> 
> Just saw this after the review sorry! 😅

Yea. Also included a note

> It doesn't always produce the perfect tests, but it generates better and higher quality tests consistently compared to the older version of the command. 

I don't think it's possible to get results that works perfectly for all cases. This is just an improved version of what we have now. I'd suggest compare the results between the current version and the improved version to see the quality improvements. 
Agree with giving more time for others to test and compare.

abeatrix:
I guess calling this a PR for improving context for the generate unit test command is more accurate 😅

vscode/src/custom-prompts/ToolsProvider.ts
```diff
@@ -101,4 +101,60 @@ export class ToolsProvider {
...
+        if (!testFileExists && testFileName) {
+            await this.createNewFile(testFileUri, testContent)
+            return
+        }
+
+        const document = await vscode.workspace.openTextDocument(testFileUri)
+        const language = document.languageId || ''
+
+        // Create a temporary file in current editor and add content to the file
+        await this.createTempFile([testFileName, testContent].join('\n'), language)
```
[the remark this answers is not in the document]
abeatrix: @toolmantim Im using the file name on top to create the new temporary file but i see what you're saying. 

I use a temporary file instead (copilot and codium do the same) because not all the test file located in the same directory as the current file for some languages, that's why we won't be able to generate a link that would work in all cases.

Maybe I can remove the file name from the top if you think that's less confusing?

lib/shared/src/chat/prompts/default-prompts.json
```diff
@@ -17,7 +17,7 @@
...
... code. If yes, use the same pattern to create new unit tests for edge cases. Focus on testing key behaviors for each function. Only includes mocks if you detected one. Use descriptive variable names that explain the purpose and meaning, not just generic names like 'foo'/'bar'/'name' etc. In your answer, start by telling me which libraries you are importing and why. For example, \"No new imports because I am writing unit tests for an existing test suite\" or \"I am importing unittest since that was used in the shared Python code,\" or \"Importing Jest because it is the configured test framework for Typescript in package.json.\" Then, show the complete test code you wrote, including 1) All necessary dependencies including packages and libraries are imported 2) Setting up the test environment 3) Writing passing tests. At the end, tell me the file path where these tests should be inserted and put the file name in the <fileName> tags. The tests should validate expected functionality and cover any example test cases from comments, as well as edge cases and bugs. All generated tests should be complete and workable, no fragments or comments, and must be enclosed in one single code block.",
```
taras-yemets, line 20: > First, review the shared context to see if there is an existing test file that match the file path contains the selected code.

"...that matches the file path _and_ contains the selected code" - is _"and"_ missing in the original version?

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -104,65 +121,136 @@ export async function getCurrentDirContext(isTestOnly: boolean): Promise<Context
...
+        const filesInDir = await vscode.workspace.fs.readDirectory(dirUri)
 
-        const contextMessages: ContextMessage[] = []
+        // Filter out directories, non-test files, and dot files
+        return filesInDir.filter(file => {
+            const fileName = file[0]
+            const fileType = file[1]
+            const isDirectory = fileType === vscode.FileType.Directory
+            const isHiddenFile = fileName.startsWith('.')
+            const isTestFile = testOnly ? fileName.includes('test') : true
```
taras-yemets, line 167: nit: When `testOnly` is `false`, `isTestFile` variable name seems a bit misleading. Maybe we can think of another name better communicating that `fileName` satisfies `testOnly` condition 🤔

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -104,65 +121,136 @@ export async function getCurrentDirContext(isTestOnly: boolean): Promise<Context
...
+    // Default to first 10 files
+    const firstFiles = filteredFiles.slice(0, 10)
+    return getDirContextMessages(directoryUri, firstFiles)
+}
+
+export async function getEditorTestContext(fileName: string): Promise<ContextMessage[]> {
+    const currentTestFile = await getCurrentTestFileContext(fileName)
+    console.log(currentTestFile)
+    const codebaseTestFiles = await getCodebaseTestFilesContext(fileName)
+    console.log(codebaseTestFiles)
```
taras-yemets, line 153:
```suggestion
    const codebaseTestFiles = await getCodebaseTestFilesContext(fileName)
```

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -104,65 +121,136 @@ export async function getCurrentDirContext(isTestOnly: boolean): Promise<Context
...
+}
 
-            if (filteredFiles.length > 1) {
-                return contextMessages
-            }
+// Get context for test file in current directory
+// TODO bee find matches using regex based on language
+export async function getCurrentTestFileContext(fileName: string): Promise<ContextMessage[]> {
+    const filePathParts = fileName.split('/')
+    const fileNameWithoutExtension = filePathParts.pop()?.split('.').shift() || ''
```
taras-yemets, line 181: Extracting dir name and file name without the extension from the file path logic is used in a few places and may be worth moving to a separate helper function.

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -178,8 +266,8 @@ export async function getPackageJsonContext(filePath?: string): Promise<ContextM
     if (!currentFilePath) {
         return []
     }
-    // Search for the package.json from the root of the repository
-    const packageJsonPath = await vscode.workspace.findFiles('**/package.json', '**/node_modules/**', 1)
+    // Search for the package.json from the base path
+    const packageJsonPath = await vscode.workspace.findFiles('**/package.json', undefined, 1)
```
taras-yemets, line 270: > When `undefined`, default file-excludes (e.g. the `files.exclude` setting
but not `search.exclude`) will apply. When `null`, no excludes will apply.

nit: `package.json` is not likely to be excluded in any way, but still, `null` may be a safer choice here.

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -229,11 +326,18 @@ export async function populateVscodeDirContextMessage(
             const fileContent = await vscode.workspace.fs.readFile(fileUri)
             const decoded = new TextDecoder('utf-8').decode(fileContent)
             const truncatedContent = truncateText(decoded, MAX_CURRENT_FILE_TOKENS)
```
taras-yemets, line 328: nit: This logic is repeated in a few places. Maybe we can move it to a helper function (e.g. `getTruncatedContent`)?

taras-yemets (approved):
Great work, @abeatrix! 
Left a few non-blocking comments.

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: > When undefined, default file-excludes (e.g. the files.exclude setting but not search.exclude) will apply. 

I think `undefined` is safer and faster because when set to null, all the dot files and node_modules will likely be searched as well

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: Yea there are a few things I will need to clean up in this file alone, will add this to my next PR to clean up this file!

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: same as above, will add this to my next PR to clean up this file!

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: Yea there are a few things I will need to consolidate for this particular file. I will add this to my next PR for cleaning this file up specifically!

vscode/src/custom-prompts/ToolsProvider.ts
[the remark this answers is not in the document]
toolmantim: Is it possible to make the chat message more friendly?

For the temporary file, is the idea that people use the filename in the tab and the filename in the top of the file to search the codebase for the file, and copy and paste?

lib/shared/src/chat/prompts/default-prompts.json
```diff
@@ -17,7 +17,7 @@
...
```
beyang, line 20: Is there a reason why we have the term `framework&libraries`? It looks like a typo, but does it yield better results?

lib/shared/src/chat/prompts/utils.ts
```diff
@@ -76,6 +80,21 @@ export function isOnlySelectionRequired(contextConfig: CodyPromptContext): boole
...
 }
 
+// Extract the word "unit" / "e2e" / "integration" from a test prompt text
+export function extractTestType(text: string): string {
+    // match "unit", "e2e", or "integration" that is follow by the word test, but don't include the word test in the matches
+    const testTypeRegex = /(unit|e2e|integration)(?= test)/i
+    return text.match(testTypeRegex)?.[0] || ''
+}
+
+export function getHumanTextForCommand(text: string, fileName?: string): string {
```
beyang, line 90: I think the term "human" is overloaded here. Claude uses the `Human: ` tag, but is the output of this function fed directly into that? It's hard to tell, and the replacement of the `{humanInput}` string makes it more confusing. Perhaps the `text` argument should be renamed, as well.

Possible name (but double check my understanding of what's happening): `getClaudeHumanText(commandInstructions: string, currentFileName?: string)`

lib/shared/src/chat/recipes/custom-prompt.ts
beyang: I don't think this is an equivalent substitute, as the switch will fallthrough all subsequent cases once it encounters a true case, rather than checking subsequent cases: https://timmousk.com/blog/typescript-switch/

beyang (changes requested):

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,134 @@ export class CustomPrompt implements Recipe {
...
+                    contextMessages.push(...rootFileNames)
+                }
+
+                // Add package.json content for ts/js files only
+                if (selection?.fileName && getFileExtension(selection?.fileName).match(/ts|js/)) {
+                    const packageJson = await getPackageJsonContext(selection?.fileName)
+                    contextMessages.push(...packageJson)
+                }
+
+                // TODO bee only top parts
```
beyang, line 150: What does this TODO mean?

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,134 @@ export class CustomPrompt implements Recipe {
...
+        const text = getHumanTextForCommand(promptText, selection?.fileName)
 
         // Attach code selection to prompt text if only selection is needed as context
-        if (selection && isOnlySelectionRequired(isContextNeeded)) {
-            const truncatedTextWithCode = promptTextWithCodeSelection(codyPromptText, selection)
-            if (truncatedTextWithCode) {
-                return newInteraction({ text: truncatedTextWithCode, displayText })
-            }
+        if (selection && isOnlySelectionRequired(contextConfig)) {
+            // const truncatedTextWithCode = promptTextWithCodeSelection(codyPromptText, selection)
```
beyang, line 77: Delete?

vscode/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 - Autocomplete on VS Code desktop instances now reuses TCP connections to reduce latency. [pull/868](https://github.com/sourcegraph/cody/pull/868)
 - Errors are now always logged to the output console, even if the debug mode is not enabled. [pull/851](https://github.com/sourcegraph/cody/pull/851)
 - Changed default and custom commands format: slash command is now required. [pull/841](https://github.com/sourcegraph/cody/pull/841)
+- The `Generate Unit Tests` command has been improved with an enhanced context fetching process that produces more consistent test results across languages. A new file containing the new tests will be created in the same directory automatically when one does not exist. If a test file already exists, a temporary file with the tests will be created. [pull/907](https://github.com/sourcegraph/cody/pull/907)
```
beyang, line 30: Does it actually create a new file? When I was testing the functionality, it only seemed to print the generated test in the sidebar—it didn't actually create a new file or make modifications to an existing file.

vscode/src/custom-prompts/CommandsController.ts
```diff
@@ -79,26 +79,42 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
...
             case 'current':
-                return this.myPromptInProgress?.description || null
+                return myPromptInProgress?.description || null
             default:
-                return this.myPromptInProgress?.prompt || null
+                return myPromptInProgress?.prompt || null
+        }
+    }
+
+    public getCurrentCommand(fileName?: string): CodyPrompt | null {
```
beyang, line 104: The logic around the "current command" feels a bit brittle and is hard for me to follow. Is the following the case?
* Some user action causes the `this.commandInProgress` to be set
* We rely on `this.commandInProgress` in other code to tell us what code to execute. This effectively constrains us to executing at most one command at a time?
* At some later point (which is unclear to me), `this.commandInProgress` is reset to null or empty?

vscode/src/custom-prompts/ToolsProvider.ts
```diff
@@ -101,4 +101,60 @@ export class ToolsProvider {
...
+        return
+    }
+
+    private async createNewFile(fileUri: vscode.Uri, content?: string): Promise<void> {
```
beyang, line 138: Add a docstring

vscode/src/custom-prompts/ToolsProvider.ts
```diff
@@ -101,4 +101,60 @@ export class ToolsProvider {
             return false
         }
     }
+
+    public async addContentToTestFile(content: string, currentFileName: string): Promise<void> {
```
beyang, line 105: Add a docstring

lib/shared/src/chat/prompts/default-prompts.json
umpox, line 20: More of a general thought: Should we consider changing this format from `json` to something else (maybe `yaml`? or something)

Writing a prompt in a JSON file doesn't feel nice at all.

Do we structure it this way so users can write their own too?

Is `{languageName}` a global that we always replace? Are there others? (Sorry if I've missed something!)

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -104,65 +121,134 @@ export async function getCurrentDirContext(isTestOnly: boolean): Promise<Context
...
+}
 
-            // Search for test files in directory
-            const testFilesPattern = `**/{test,tests}/**/*test*.${fileExtension}`
-            const testFilesContext = await getEditorFoundFilesContext(testFilesPattern)
-            contextMessages.push(...testFilesContext)
+// Get context for test file in current directory
+async function getCodebaseTestFilesContext(fileName: string): Promise<ContextMessage[]> {
+    // exclude any files in the path with e2e or integration in the directory name
+    const excludePattern = '**/*{e2e,integration}*/**'
```
umpox, line 200: This feels quite specific, it obviously applies to our codebase but I'm curious how often this would apply elsewhere.

I feel like we should refrain from trying to fine tune this like this, it could lead to hard to debug situations. E.g.

This wouldn't include any files that are in a `/vscode-integration` folder, but the user might actually want those!

I guess the alternative would be to let the user specific their own patterns that are relevant to their own codebase

lib/shared/src/chat/prompts/default-prompts.json
abeatrix: @beyang I've documented most of the prompts I've tested so you can get an idea of how the prompt evolved: https://sourcegraph.sourcegraph.com/notebooks/Tm90ZWJvb2s6MTMz

It started from `framework/library` but ended up not working for some languages.
I haven't tried `framework & libraries` though. Maybe it'd be the same

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -178,8 +264,8 @@ export async function getPackageJsonContext(filePath?: string): Promise<ContextM
     if (!currentFilePath) {
         return []
     }
-    // Search for the package.json from the root of the repository
-    const packageJsonPath = await vscode.workspace.findFiles('**/package.json', '**/node_modules/**', 1)
+    // Search for the package.json from the base path
+    const packageJsonPath = await vscode.workspace.findFiles('**/package.json', undefined, 1)
```
umpox, line 268: This also feels quite specific to our setup and TypeScript files.

Do we suggest that generate tests works best for TS/JavaScript? I also feel like this could be an area where we let users actually specify patterns they want to include. Otherwise we'll have a _lot_ of logic where we want to apply similar patterns for different languages. (Like `go.mod`  for `Go`)

Likewise this wouldn't necessarily work for users who use yarn/npm workspaces. Idk what other scenarios exist for other languages

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -362,3 +475,84 @@ export function getCurrentFileContextFromEditorSelection(selection: ActiveTextEd
...
+ */
+export function getCurrentFileContext(): ContextMessage[] {
+    const currentFile = vscode.window.activeTextEditor?.document
+    const currentFileText = currentFile?.getText()
+    if (!currentFileText || !currentFile?.fileName) {
+        return []
+    }
+
+    const truncatedContent = truncateText(currentFileText, MAX_CURRENT_FILE_TOKENS)
+    const fileName = vscode.workspace.asRelativePath(currentFile.fileName)
```
umpox, line 490: Should we try to make this general seeing as we're in `lib/shared` and not `vscode`? I feel like we should probably lint against using `vscode`! (I know we are already using it in a few places already)

lib/shared/src/chat/prompts/vscode-context.ts
```diff
@@ -362,3 +475,84 @@ export function getCurrentFileContextFromEditorSelection(selection: ActiveTextEd
...
+    }
+}
+
+/**
+ * Finds the test file corresponding to the given file name.
+ *
+ * @param fileName - The name of the file to find the test file for
+ * @returns The path to the found test file, or an empty string if none found
+ */
+export async function getTestFileOfCurrentFileContext(fileName: string): Promise<ContextMessage[]> {
```
umpox, line 538: I think it'd be super helpful to have some simple unit tests for the functions in this file. Just like with examples of the files that are found and the files that aren't found.

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: left over. will remove it

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,134 @@ export class CustomPrompt implements Recipe {
...
-        // Check if selection is required. If selection is not defined, accept visible content
-        const selectionContent = isContextNeeded?.selection
-            ? context.editor.getActiveTextEditorSelection()
-            : context.editor.getActiveTextEditorSelectionOrVisibleContent()
 
-        const selection = selectionContent
-        if (isContextNeeded?.selection && !selection?.selectedText) {
-            const errorMessage = `__${slashCommand}__ requires highlighted code. Please select some code in your editor and try again.`
-            return newInteractionWithError(errorMessage, slashCommand)
+        const commandName = command?.slashCommand || command?.description || promptText
```
umpox, line 65: `description` or `promptText` seems incorrect here, they will both likely be very long as it doesn't feel helpful to incldue this in the error message

vscode/src/custom-prompts/CommandsController.ts
umpox: Just to add, I think we should consider making this a `ViewProvider` like `ChatViewProvider` or `FixupViewProvider`.

We can still use the command controller for interactions, but I think it will clean up some of this logic, and mean that we can fire multiple commands at once instead of being restricted.

It will also mean we don't need to add command logic to the shared `MessageProvider`. That feels like it could cause bugs, e.g. if a user sends off a sidebar chat message or a fixup whilst a command is executing - it will still be sent to the command controller. Instead - `CommandViewProvider` would just be able to do this logic inside a `handleTranscript`

vscode/src/custom-prompts/ToolsProvider.ts
abeatrix: added

vscode/src/custom-prompts/ToolsProvider.ts
abeatrix: added

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: Yea `getPackageJsonContext` is only applied to [ts/js file ](https://sourcegraph.com/github.com/sourcegraph/cody@bee/unit-tests/-/blob/lib/shared/src/chat/recipes/custom-prompt.ts?L143).

For other language, the test files we fetch for a specific language is usually good enough to tell Cody what testing framework the project is using. But with ts and js because of all kind of different tests files it can have, we might not able to fetch the correct test files all the time so I thought adding package.json as context would help. Do you have other suggestions on what other approach we can take here?

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: Deleted

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: switched back to use if statementd

lib/shared/src/chat/prompts/default-prompts.json
abeatrix: Yea it's using the same JSON file that the users would use to create a custom command.
Current `{languageName}` is the only global we always replace (for internal use)

lib/shared/src/chat/prompts/utils.ts
abeatrix: updated from getHumanTextForCommand getClaudeHumanText

vscode/CHANGELOG.md
abeatrix: i'm going to meet with @toolmantim this afternoon to explore what the best way to handle this, will report back!

vscode/src/custom-prompts/ToolsProvider.ts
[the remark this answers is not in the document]
abeatrix: Yea i will update the prompt and see if that works. Let's explore more options when we meet later to see what makes sense for tests generated that work for all the popular langauges

lib/shared/src/chat/prompts/vscode-context.ts
umpox: I feel like it's easier just to ask the user to specific what testing framework they use. I think the ideal case this would work but in many cases it might just lead to more confusion (e.g. in the sourcegraph repo we have both `mocha` and `jest` in use, but really we want to be using `jest`).

This is beyond this PR, but in the longer term I really think we should consider building a form of template engine for these instructions. It could even be a Markdown format, so it's really easy for users to create their own commands.

We would provide certain globals that users can lean on to build their prompts, e..g:
- `{languageName}` like we already have, but also `{selection}`, `{fileName}` etc etc
- `{find("glob")}` we replace this with the results of a glob search before the prompt is sent to the LLM
- `{prompt("placeholder")}` we replace this with the outcome of an editor prompt, so users can build prompts that can take further input.

That's just some examples, but ideally we can have a set of base tools that lets us build new commands like this without ever having to write additional logic.

umpox: Given that's likely out of scope and commands are the most high priority thing right now, I think that can wait - but just something we should think about

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: i've updated so that is exclude Pattern only applies to unit tests and not integration test or e2e test requests. 

The main goal of this function is to filter unhelpful contexts that would confuse Cody.

You can see how this is the case for the current test command:
image

vscode/src/custom-prompts/CommandsController.ts
abeatrix: > This effectively constrains us to executing at most one command at a time?

1. We get the prompt text from commandInProgress during the [chatCommandsFilter step in MessageProvider](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/MessageProvider.ts?L527-529) that also sets `this.commandInProgress`
2. For the `/test` command, the `custom-prompt` recipe sets the file details for the `commandInProgress` for the post-response process (creating test files)
3. `commandInProgress` is reset when it's disposed or when a new command is run in step 1

This does not constrain users from running one command at a time (our Chat View does, users cannot submit a new request when a message is in progress).  However, if a user executes a command in Chat View, and then runs another command in the Inline-Chat, this would stop the new test file from creating, which I will look into now (thanks for pointing this out).

I can try the approach @umpox suggested, separating this into a CommandViewProvider. @umpox can you share more insights on how this is handled by the current inline-chat implementation?

>  That feels like it could cause bugs, e.g. if a user sends off a sidebar chat message or a fixup whilst a command is executing - it will still be sent to the command controller.

Regarding conflict with the fixup commands: The `CommandController` only handles custom commands and the `default` commands, so running fixup whilst a custom command is executing should not cause any issue as fixup is handled by the fixup controller that we catch through the [chatCommandsFilter in MessageProvider](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/MessageProvider.ts?L527-529)

lib/shared/src/chat/prompts/default-prompts.json
abeatrix: No i think what I have here is correct (and something the LLM understands). 
"the file path contains the selected code" is referring to _the current file that contains the selected code_ here.

Updating it to "an existing test file that matches the file path and contains the selected code" would means  asking Cody to check if there is an existing test file that contains the selected code, which would always be false because selected code exists in non-test files

lib/shared/src/chat/prompts/default-prompts.json
abeatrix: Just updated from `framework&libraries` to `framework and libraries` to be clear.

You can see the test results for the updated prompt listed under the [Test 10 section in the Notebook](https://sourcegraph.sourcegraph.com/notebooks/Tm90ZWJvb2s6MTMz#test-10---latest-prompt-used-48aab2f1-793f-408a-86a8-0aa9a5a73a6e)

vscode/src/custom-prompts/CommandsController.ts
abeatrix: @umpox Does the Inline Chat currently use the same context as defined for each command?

I am currently getting different results for the `/test` command when running the command via Inline Chat, which seems to relate to the recent change to the Inline Chat Provider / Controller:

image

abeatrix: Update: I've reverted changes made to `this.commandInProgress` so that it will not constrain us from executing more than one command at a time.

vscode/src/custom-prompts/ToolsProvider.ts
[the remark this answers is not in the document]
abeatrix: will do this in a separate PR as discussed.

abeatrix:
## Update

Updated the PR to focus on improving:
- the quality of the context fetched for `/test` command
- the quality of the tests generated by the `/test` command

Working with @toolmantim to explore options for inserting the generated tests into new/existing test files, which will be included in a new PR.

vscode/CHANGELOG.md
abeatrix: Updated the PR to focus on test output and context used for the /test command. 
Working with Tim on finding the "right behavior" for the /test command and will update that in a separate PR.

lib/shared/src/chat/prompts/default-prompts.json
```diff
@@ -17,7 +17,7 @@
...
... see if there is an existing test file for the selected code. If yes, use the same pattern to create new unit tests for edge cases. Focus on testing key behaviors for each function. Only includes mocks if you detected one. Use descriptive variable names that explain the purpose and meaning, not just generic names like 'foo'/'bar'/'name' etc. In your answer, start by telling me which libraries you are importing and why. For example, \"No new imports because I am writing unit tests for an existing test suite\" or \"I am importing unittest since that was used in the shared Python code,\" or \"Importing Jest because it is the configured test framework for Typescript in package.json.\" Then, show the complete test code you wrote, including 1) All necessary dependencies including packages and libraries are imported 2) Setting up the test environment 3) Writing passing tests. At the end, tell me the file path where these tests should be added. The tests should validate expected functionality and cover any example test cases from comments, as well as edge cases and bugs. Code for all generated tests must be complete and workable, no fragments or comments, enclosed in a single code block.",
```
dominiccooney, line 20: I'd love to see some samples of this on codebases which don't have package.json, and more languages.

Grammar nit: "only includes mocks" => "only include mocks"

lib/shared/src/chat/prompts/templates.ts
dominiccooney: Since this is applied everywhere it feels like a roll of the dice. We need to record a battery of inputs and outputs to monitor for regressions. But since we don't have that yet, I guess YOLO?

Curious how you arrived at "any misleading information" being useful.

lib/shared/src/chat/prompts/templates.ts
```diff
@@ -1,18 +1,21 @@
...
 
 export const answers = {
-    terminal: 'Noted. I will answer your next question based on this terminal output with the code you just shared.',
+    terminal: 'Noted. I will answer your next question based on this terminal output with other code you shared.',
     selection: 'Noted. I will refer to this code you selected in the editor to answer your question.',
-    file: 'Noted. I will refer to this file you are looking at, with your selected code inside the <selected> tags to answer your question.',
+    file: 'Noted. I will refer to this codebase file you are looking at to answer you next question for the code in the <selected> tags.',
+    fileList:
+        'Noted. I will refer to this list of files from the {fileName} directory of your codebase to answer your next question.',
+    packageJson: 'Noted. I will use the right libraries/framework already setup in your codebase for your questions.',
```
dominiccooney, line 18: Not sure how much LLMs care about grammar, but libraries/framework*s* for consistent plurals, and *set up* instead of *setup*?

lib/shared/src/chat/prompts/utils.ts
```diff
@@ -76,6 +80,21 @@ export function isOnlySelectionRequired(contextConfig: CodyPromptContext): boole
     return !contextConfig.none && ((contextConfig.selection && contextConfigLength === 1) || !contextConfigLength)
 }
 
+// Extract the word "unit" / "e2e" / "integration" from a test prompt text
```
dominiccooney, line 83: There's a lot more relevant words here, what about: acceptance, functional, smoke, sanity (ableist but I guess we meet the user where they are), stress, performance, regression, stress, load, user acceptance, UAT, automated, security, accessibility, a11y, fuzz, ...

But since you only use this to look for "unit", why not simplify the code to just look for "unit"? You should also look for "unittest"... probably about 1/6th of "unit test" per Google Books ngrams but growing.

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,132 @@ export class CustomPrompt implements Recipe {
...
-        const isContextNeeded = contextConfig
-            ? (JSON.parse(contextConfig) as CodyPromptContext)
-            : defaultCodyPromptContext
+        const contextConfigString = (await context.editor.controllers?.command?.get('context')) || ''
+        const contextConfig = JSON.parse(contextConfigString) as CodyPromptContext
+
+        // Check if selection is required. If selection is not defined, accept visible content
+        const selectionContent = contextConfig?.selection
+            ? context.editor.getActiveTextEditorSelection()
+            : context.editor.getActiveTextEditorSelectionOrVisibleContent()
```
dominiccooney, line 52: Doesn't `getActiveTextEditorSelectionOrVisibleContent` already get the selection, or if there's none, the visible content? Why do we need this ternary?

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,132 @@ export class CustomPrompt implements Recipe {
...
+
+        const selection = selectionContent
```
dominiccooney, line 54: Adding a new binding for the same value makes sense if the name adds some documentation, but that's not the case here; let's simplify this!

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,132 @@ export class CustomPrompt implements Recipe {
...
         // Check if context is required
-        const contextConfig = await context.editor.controllers?.command?.get('context')
-        const isContextNeeded = contextConfig
-            ? (JSON.parse(contextConfig) as CodyPromptContext)
-            : defaultCodyPromptContext
+        const contextConfigString = (await context.editor.controllers?.command?.get('context')) || ''
+        const contextConfig = JSON.parse(contextConfigString) as CodyPromptContext
+
+        // Check if selection is required. If selection is not defined, accept visible content
+        const selectionContent = contextConfig?.selection
```
dominiccooney, line 50: The name `selectionContent` doesn't seem great, it is actually an object, not the content of the selection. That is just one part of the object.

lib/shared/src/chat/prompts/utils.ts
```diff
@@ -76,6 +80,21 @@ export function isOnlySelectionRequired(contextConfig: CodyPromptContext): boole
...
+export function extractTestType(text: string): string {
+    // match "unit", "e2e", or "integration" that is follow by the word test, but don't include the word test in the matches
+    const testTypeRegex = /(unit|e2e|integration)(?= test)/i
+    return text.match(testTypeRegex)?.[0] || ''
+}
+
+export function getClaudeHumanText(commandInstructions: string, currentFileName?: string): string {
```
dominiccooney, line 90: We need a generic name for the "upstream LLM" because this is Claude less over time.

lib/shared/src/chat/prompts/utils.ts
```diff
@@ -76,6 +80,21 @@ export function isOnlySelectionRequired(contextConfig: CodyPromptContext): boole
...
+    const promptText = prompts.instruction.replace('{humanInput}', commandInstructions)
+    if (!currentFileName) {
+        return promptText
```
dominiccooney, line 93: Should we fail if we're returning prompts containing uninterpolated strings like `{languageName}`? That would be a programming error and it is good to detect the problem close to the source.

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,132 @@ export class CustomPrompt implements Recipe {
...
+        if (promptContext.currentDir) {
+            const currentDirMessages = await getCurrentDirContext(isUnitTestRequest)
+            contextMessages.push(...currentDirMessages)
         }
-
-        // Files from a directory path
-        if (isContextRequired.directoryPath?.length) {
-            const fileContext = await getEditorDirContext(isContextRequired.directoryPath, selection?.fileName)
-            contextMessages.push(...fileContext)
+        if (promptContext.directoryPath !== undefined) {
```
dominiccooney, line 125: Simplify this to `if (promptContext.directoryPath)`, undefined is also false; ditto below with `promptContext.filePath`

lib/shared/src/chat/recipes/custom-prompt.ts
```diff
@@ -38,140 +39,132 @@ export class CustomPrompt implements Recipe {
...
-            const contextMsg = getCurrentFileContextFromEditorSelection(selection)
-            currentFileContextStack.push(...contextMsg)
+        // Context for unit tests requests
+        if (isUnitTestRequest && contextMessages.length === 0) {
+            if (workspaceRootUri) {
+                const rootFileNames = await getDirectoryFileListContext(workspaceRootUri)
+                contextMessages.push(...rootFileNames)
+            }
+            // Add package.json content for ts/js files only
+            if (selection?.fileName && getFileExtension(selection?.fileName).match(/ts|js/)) {
```
dominiccooney, line 144: Maybe add tsx and jsx here on the off chance someone's using those exclusively? Are there any other template filename extensions we should add? ejs?

dominiccooney (commented):
Some feedback inline.

I realize there's a lot of reviewers on this. Wanted to chime in to help but feel free to not block on me.

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: this regex should match tsx and jsx and ejs  already?

lib/shared/src/chat/prompts/default-prompts.json
abeatrix: Yee here is a list of videos of me trying this out in different languages:
typescript: https://www.loom.com/share/3489f9ae811b41d08ca0329489580543?sid=379b3122-3ea7-4b1e-bed2-12721e1f09b6
go: https://www.loom.com/share/d1514db896c840989e05f7c74fd30b81?sid=60f8d4d6-2b6c-42d3-b016-b347666893e9
go 2: https://www.loom.com/share/952caef2067a4d748bcd10eef8a60800?sid=04ca3f58-e2c8-474f-86ee-08af03f00689
scala/java: https://www.loom.com/share/d28b1089ac8a4d9ea95816c3e476ec35?sid=adabd913-3811-4009-9391-ccd1406fe91d
python: https://www.loom.com/share/7a788b389ce84a4b85fbe2d31f4a0ab4?sid=11d9314d-31c3-4530-be6a-b4e9faa5de0b

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: Good question! if the command is configured to have selection required, we will only accept `context.editor.getActiveTextEditorSelection()` to make sure we are not using visible content instead.  Will update the docstring to reflect this

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: Good catch! I must be testing something and forgot to change it back lol Thanks!

lib/shared/src/chat/recipes/custom-prompt.ts
abeatrix: updated to selection :)

lib/shared/src/chat/prompts/utils.ts
abeatrix: I find claude would just replace the language name "in their head" if one is not provided so I think we are safe?

lib/shared/src/chat/prompts/utils.ts
abeatrix: My thinking was that not everyone puts unit tests in a directory called unit, but they most likely do not exist in directories called e2e 🤣  (or if they do, they hopefully aren't the only unit tests in the projects for us to find). The main goal is to find a couple of unit tests as examples for the LLM to see how unit tests are set up in the current project.
We try to find the test files in the current directory first, if we did find one, we will just stop there.
If we can't find any, then we will do a project level search

lib/shared/src/chat/prompts/vscode-context.ts
abeatrix: Yea I have added it to my follow-up items. Is editor the best place?

lib/shared/src/chat/prompts/templates.ts
abeatrix: (prompt updated in https://github.com/sourcegraph/cody/pull/984 which will be merged back into this branch)

lib/shared/src/chat/prompts/templates.ts
abeatrix: I think you are thinking about the one we use in premade? That one is applied everywhere but this one is only applied to prompts for custom commands, on top of the one in premade.

abeatrix: > Curious how you arrived at "any misleading information" being useful.

lol a teammate reported an issue  on [slack](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1692656525976169?thread_ts=1692649901.285349&cid=C04MSD3DP5L) about Cody hallucinated directory and file names, which I was able to reproduce. I saw that we already asked Cody not to make up file names, file path, only shows file paths that they are certain of their existence etc in the rules we set for Cody, which we can see are not doing much. 
I noticed Cody doesn't know the connections between the files, file name, file path etc we share with them. Long story short, refer anything I share with Cody as context / information has been working well for me, so I tried the `"any misleading information" `  approach by adding it to my custom command starter, asked Cody the same question and find it to be helpful so far.

image

^ You can see it in the output channel where I added `Do not make any assumptions about the code and file names or any misleading information.` in front of the same question where teammate reported the hallucinating issue

Just tried asking Cody the same question on web and get the same hallucinating problem (the [codeownership](https://sourcegraph.com/search?q=context:global+src/components/CodeOwnership+r:%5Egithub%5C.com/sourcegraph/sourcegraph%24+&patternType=standard&sm=1&groupBy=repo) it mentioned doesn't exist):
image

If i prepend the statement to the question, I get the "honest" response. It would just refer to the actual shared context:
image

lib/shared/src/chat/prompts/default-prompts.json
```diff
@@ -17,7 +17,7 @@
...
... <selected> using the detected test framework and libraries. Ensure to follow the same patterns as the shared unit tests (if any), and add packages, imports, dependencies, and assertions only if used in the shared code. Pay attention to the file path of each shared context to identify if a test file for the selected code already exists. If it exists, focuse on generate new unit tests for uncovered cases. If none detected, import common unit test libraries for {languageName}. Focus on validating the key functionality with simple and clear assertions. Only includes mocks if one is detected in shared code. Before writing the tests, determine which test libraries or frameworks to import, e.g. 'No new imports needed - using existing libs'/'Importing test framework that matches shared context usage'/'Importing the defined framework' etc. Then, tell me the file path where these tests should be added and summarize test coverage. At the end, enclose only fully completed code for the new unit tests in single markdown codeblock, no fragments. New tests should validate expected functionality and cover edge cases for <selected> with all required methods imported. Do not repeat existing tests.",
```
dominiccooney, line 20: Where you have "if it exists", do you find "it" makes the LLM squirrelly versus reiterating "if *a test file* exists"?

Why did we add flourishes like "shared code context *and configurations*"? I'm interested in how you arrived at these, determined the effect of them, because I am curious if we can scale it up.

These "think through" prompts for no new imports, importing test framework, etc. are super interesting too.

Spelling/grammar nits:
- `focuse on generate` &rarr; `focus on generating` (or maybe just `generate`?)
- `Only includes mocks` &rarr; `Only include mocks`

dominiccooney (approved):
Looks good. Some minor feedback about the prompt inline.

lib/shared/src/chat/prompts/default-prompts.json
abeatrix: > Where you have "if it exists", do you find "it" makes the LLM squirrelly versus reiterating "if a test file exists"? Why did we add flourishes like "shared code context and configurations"? I'm interested in how you arrived at these, determined the effect of them, because I am curious if we can scale it up.

Good question! I noticed `"if a test file exists"` doesn't work in some languages; even if test files were shared, it would still refer to them as "shared Python code":

image

So I tried to minimize the use of the word "file" and replace it with either "shared context" or "shared code". 

As for configurations, we are sharing package.json for ts/js tests when no test context is found, but for non-ts/js tests, we will just share a list of files located in their root dir to see if that would give them hints if a test framework is set up. Plus on top of that, i'm not familiar with how tests are set up in other languages so instead of listing each config file for the model to pick up, it seems like `configurations` has been working so far 🙏 

I am not sure if I am on the right track, though 😅 


And thanks for the grammar check and for taking the time to review 🙇‍♀️ Going to follow your suggestion and make sure the prompt still works before merging.

abeatrix:
Here is a demo run from the latest commit build: https://www.loom.com/share/a17c8c63d1284ac889cff2d0bdc93a9c?sid=aee107b7-bf23-4344-9235-a290d897e0f9

Tested the command with go, python, scala, and typescripts files as shown in the loom video above.

I have also worked on inserting the tests automatically to their test files (see video below) in https://github.com/sourcegraph/cody/pull/984, which was closed as we agreed more design works are required as well as code change with current behavior.
https://github.com/sourcegraph/cody/assets/68532117/610bce25-ab02-4fcb-a4b7-4700f71b62b9

I am going to merge what I have now for the quality improvement to focus on autocompletion works.
Will pick this back up after the oct release.

dominiccooney:
@abeatrix thanks for showing the output from those alternative prompts! It really illustrates how steep the drop off in performance is even changing single words.

PR #912, opened 2023-09-04
VS Code: use singleton storage providers in rest consumers

valerybugakov:
## Context

- A follow-up to https://github.com/sourcegraph/cody/pull/876: migrates other `localStorage` and `secretStorage` consumers to importing singletons instead of using them as arguments.
- Part of #874

The end goal of this effort (as stated in https://github.com/sourcegraph/cody/issues/874) is to decouple `getConfiguration` from the other entities in the codebase and introduce atomic and reactive getting for config values.

## Test plan

CI


philipp-spiess (approved):
Of course!

PR #913, opened 2023-09-04
Autocomplete: trigger multiline completions on empty block only

valerybugakov:
## Context

- Trigger single-line completions instead of multi-line if a cursor is placed at the beginning of a non-empty block.
- Discussed [here](https://sourcegraph.slack.com/archives/C05AGQYD528/p1693799104151879).

## Test plan

CI


philipp-spiess (approved):
Happy with this change but

- Let's add it to the change log
- Let's add a test for it

PR #885, opened 2023-08-31
hackathon poc - autocomplete using all surrounding code as context

abeatrix:
This branch contains an update to the current prompt used for autocomplete as a proof of concept to get Claude Instant to consider code after the current cursor when auto-completing:

https://github.com/sourcegraph/cody/assets/68532117/b75e3583-1043-476e-8727-c80963697f4c

As shown in the video above (and the ones below), you can see Cody can generate doc string for existing functions inline when building from this branch. I also haven't encountered the "random XML tags" mentioned by Philipp below:

> As mentioned in the original post, the issue was noticeable after using it for a bit and random XML tags started to be in the completions OR with some other mutations I was getting a lot of empty results instead.

However, please note that no tests have been conducted to confirm the stability of these changes and everything in this branch should be treated as a concept until confirmed.

PR: I also played around with the context a bit but didn't have time to remove them for the PR, just ignore those lol

### After

https://github.com/sourcegraph/cody/assets/68532117/70406618-ea95-47e1-bfc0-12906c47c5da

### Before

https://github.com/sourcegraph/cody/assets/68532117/dbdc8f6d-5736-4f76-b09e-d2d405ece117

## Test plan

Not a real branch -- this contains unrelated WIP that was added by accident

Screenshot 2023-08-31 at 4 54 28 AM
Screenshot 2023-08-31 at 4 55 01 AM
Screenshot 2023-08-31 at 4 56 47 AM
Screenshot 2023-08-31 at 4 58 19 AM
Screenshot 2023-08-31 at 4 59 32 AM
Screenshot 2023-08-31 at 5 00 06 AM
Screenshot 2023-08-31 at 5 00 51 AM
Screenshot 2023-08-31 at 5 01 38 AM
Screenshot 2023-08-31 at 5 02 23 AM
Screenshot 2023-08-31 at 5 03 21 AM

PR #914, opened 2023-09-04
Modify autocomplete to use all surrounding code as context 

arafatkatze:
This is a work from @abeatrix  on https://github.com/sourcegraph/cody/pull/885/ Just worded slightly differently so that I could understand how autocomplete actually functions. 

I am referencing this code for example
https://github.com/sourcegraph/cody/blob/66bf5f1d6bb991c106e59cbfe0fbbb5fc49083a9/vscode/src/custom-prompts/utils/menu.ts#L176-L218

Lets say the cursor is at Line 208 then if you look above the cursor you see a few config menu options

## Before
In the current behaviour when you try to add a comment it takes into account only the config menu options and makes a docstring

https://github.com/sourcegraph/cody/assets/11155207/7d1e747e-b1e2-4452-a896-4866fdf6c9ac

On asking GPT4 we realize the docstring doesn't make perfect sense(which is obviously try coz its missing the suffix context)
image

## After

https://github.com/sourcegraph/cody/assets/11155207/dd68c040-d3e0-469d-8e4e-1aed1e85ee32

Now in the new code the docstring makes a lot of sense coz it has the suffix context also and I checked whether the docstring makes sense with GPT4 also. 
image

The essence of this PR is that adding suffix context and sending that to the autocomplete gives much better results and I tried that in other contexts too. 

## Test plan
1. The local tests are passing
```
 ✓ |vscode| src/completions/context/bestJaccardMatch.test.ts (4)
 ✓ |vscode| src/services/InlineAssist.test.ts (16)
 ✓ |vscode| src/completions/utils/string-comparator.test.ts (12)
 ✓ |vscode| src/non-stop/utils.test.ts (3)
 ✓ |agent| src/index.test.ts (2)
 ✓ |e2e| src/entity-detection.test.ts (3)
 ✓ |vscode| src/completions/text-processing.test.ts (12)
 ✓ |agent| src/AgentTextDocument.test.ts (5)
 ✓ |web| src/sample.test.ts (1)
 ✓ |vscode| src/configuration.test.ts (2)
 ✓ |vscode| src/completions/vscodeInlineCompletionItemProvider.test.ts (11) 462ms
 ✓ |shared| src/utils.test.ts (11)
 ✓ |shared| src/experimentation/FeatureFlagProvider.test.ts (6)
 ✓ |shared| src/chat/transcript/transcript.test.ts (10)
 ✓ |shared| src/guardrails/index.test.ts (6)
 ✓ |shared| src/common/markdown/markdown.test.ts (12)
 ✓ |shared| src/chat/bot-response-multiplexer.test.ts (9)
 ✓ |shared| src/codebase-context/rerank.test.ts (1)
 ✓ |shared| src/sourcegraph-api/utils.test.ts (4)
 ✓ |shared| src/chat/recipes/fixup.test.ts (2)

 Test Files  31 passed (31)
      Tests  282 passed | 1 skipped (283)
   Start at  01:46:17
   Duration  6.44s (transform 1.40s, setup 1.18s, collect 9.97s, tests 1.56s, environment 9.21s, prepare 5.19s)


 PASS  Waiting for file changes...
       press h to show help, press q to quit
```

2. I stopped at the debugger at various places to actually check the prompt to make sure it includes text AFTER the cursor position and now that it does that we can get much better results.

### DIFF
--- a/vscode/src/completions/getInlineCompletions.test.ts
+++ b/vscode/src/completions/getInlineCompletions.test.ts
@@ -1488,7 +1488,7 @@ describe('getInlineCompletions', () => {
         expect(messages[messages.length - 1]).toMatchInlineSnapshot(`
             {
               "speaker": "assistant",
-              "text": "Here is the code: <CODE5711>constructor(startLine: number, startCharacter: number, endLine: number, endCharacter: number) {
+              "text": "Understood. Here is the completed code snippet aligned with your guidelines: <CODE5711>constructor(startLine: number, startCharacter: number, endLine: number, endCharacter: number) {
                     this.startLine =",
             }
         `)
@@ -1516,7 +1516,9 @@ describe('getInlineCompletions', () => {
         )
         expect(requests).toHaveLength(3)
         const messages = requests[0].messages
-        expect(messages[messages.length - 1].text).toBe('Here is the code: <CODE5711>class Range {\n')
+        expect(messages[messages.length - 1].text).toBe(
+            'Understood. Here is the completed code snippet aligned with your guidelines: <CODE5711>class Range {\n'
+        )
     })
 
     test('synthesizes a completion from a prior request', async () => {
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -64,25 +64,20 @@ export class AnthropicProvider extends Provider {
         const prefixMessages: Message[] = [
             {
                 speaker: 'human',
-                text: `You are a code completion AI that writes high-quality code like a senior engineer. You are looking at ${
-                    this.options.fileName
-                }. You write code in between tags like this: ${OPENING_CODE_TAG}${
-                    this.options.languageId === 'python' || this.options.languageId === 'ruby'
-                        ? '# Code goes here'
-                        : '/* Code goes here */'
-                }${CLOSING_CODE_TAG}.`,
+                text: 'You are a sophisticated code-completion AI, specifically designed to understand the intricacies of coding context. Your abilities include grasping the semantic and syntactic elements of the code I’m working on and offering completion suggestions that not only fit the functional requirements but also adhere to the stylistic and architectural patterns present in the existing codebase.',
             },
             {
                 speaker: 'assistant',
-                text: 'I am a code completion AI that writes high-quality code like a senior engineer.',
+                text: 'Acknowledged. My design incorporates advanced contextual understanding, which allows me to generate code completions that are functionally coherent, stylistically consistent, and architecturally aligned with your existing code.',
             },
             {
                 speaker: 'human',
-                text: `Complete this code: ${OPENING_CODE_TAG}${head.trimmed}${CLOSING_CODE_TAG}.`,
+                text: `I'd like you to focus on the following attributes while completing the code snippet enclosed in the ${OPENING_CODE_TAG}. First, adhere to the naming conventions present in the existing code. Second, maintain stylistic consistency. Third, make sure to define return types explicitly for functions if that's the practice in the surrounding code. Fourth, refrain from using any libraries or methods that aren't already imported or defined. Fifth, avoid redundancy by not repeating code, functions, or methods that are present in the existing code. Sixth: Focus on writing clean, efficient code that works seamlessly with surrounding code. Complete the following code: ${OPENING_CODE_TAG}
+                ${head.trimmed}${OPENING_CODE_TAG}${tail.trimmed}${CLOSING_CODE_TAG}${this.options.docContext.suffix}`,
             },
             {
                 speaker: 'assistant',
-                text: `Here is the code: ${OPENING_CODE_TAG}${tail.trimmed}`,
+                text: `Understood. Here is the completed code snippet aligned with your guidelines: ${OPENING_CODE_TAG}${tail.trimmed}`,
             },
         ]
         return { messages: prefixMessages, prefix: { head, tail, overlap } }
@@ -101,11 +96,11 @@ export class AnthropicProvider extends Provider {
             const snippetMessages: Message[] = [
                 {
                     speaker: 'human',
-                    text: `Here is a reference snippet of code: ${OPENING_CODE_TAG}${snippet.content}${CLOSING_CODE_TAG}`,
+                    text: `Codebase context from a file with file path ${snippet.fileName}: ${OPENING_CODE_TAG}${snippet.content}${CLOSING_CODE_TAG}`,
                 },
                 {
                     speaker: 'assistant',
-                    text: 'I have added the snippet to my knowledge base.',
+                    text: 'I will refer to this code when completing your next request.',
                 },
             ]
             const numSnippetChars = messagesToText(snippetMessages).length + 1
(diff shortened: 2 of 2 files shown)


philipp-spiess:
Hey @arafatkatze! Thanks for the contributions. I think we should launch this behind a feature flag so we can measure the effects of the prompt change, do you mind adding one following the example of e.g. [this PR](https://github.com/sourcegraph/cody/pull/881)

Also I’m curious if you have any anecdotal thoughts to that: When I tried out this prompt change, it causes me to insert code comments much more often then desired. It also returned no results more often than the previous implementation but I don't think this was necessarily a bad thing - It's just unexpected.

arafatkatze:
@philipp-spiess Here are some anecdotal thoughts from me

> When I tried out this prompt change, it causes me to insert code comments much more often then desired. 

So given that this new method takes into account the context AFTER the code the comments now would make a lot more sense than before coz the comment would be explaining the upcoming function. I hope that's what you meant and I hope the extra comment suggestions isn't problematic. Do let me know if you meant something else and I am happy to take a look.

>  It also returned no results more often than the previous implementation but I don't think this was necessarily a bad thing - It's just unexpected.

There is a very good reason for that behaviour and I am quite happy with it. In the previous code we do 
```
text: `Complete this code: ${OPENING_CODE_TAG}${head.trimmed}${CLOSING_CODE_TAG}.`,
```
What this does is that it shows 1-2 lines before the position of the current cursor whereas now we do 
```
${OPENING_CODE_TAG}
                ${head.trimmed}${OPENING_CODE_TAG}${tail.trimmed}${CLOSING_CODE_TAG}${this.options.docContext.suffix}`,
```

My guess is that  because of tail.trimmed  when you are writing stuff and its only partially complete like the completion results tend to be empty string and then when you write a full word then the completion results tend to be meaningful. That's what I found in my local tests with the debugger.
`c`  => Nothing on completion 
`co`  => Nothing on completion 
`con`  => Nothing on completion 
`cons`  => Nothing on completion
`const `  =>  shows  `"const variableName = something"`

The above example is what I usually found in my local testing with the debugger. 

Let me add a feature flag now.

philipp-spiess:
> So given that this new method takes into account the context AFTER the code the comments now would make a lot more sense than before coz the comment would be explaining the upcoming function. I hope that's what you meant and I hope the extra comment suggestions isn't problematic. Do let me know if you meant something else and I am happy to take a look.

I meant that when I played around with Beatrix's prototype, the code suggestions would be comments `// do something` instead of actual code. 


>  It also returned no results more often than the previous implementation but I don't think this was necessarily a bad thing - It's just unexpected.
> 
> There is a very good reason for that behaviour and I am quite happy with it. In the previous code we do 
> ```
> text: `Complete this code: ${OPENING_CODE_TAG}${head.trimmed}${CLOSING_CODE_TAG}.`,
> ```
> What this does is that it shows 1-2 lines before the position of the current cursor whereas now we do 
> ```
> ${OPENING_CODE_TAG}
>                 ${head.trimmed}${OPENING_CODE_TAG}${tail.trimmed}${CLOSING_CODE_TAG}${this.options.docContext.suffix}`,
> ```
> 
> My guess is that  because of tail.trimmed  when you are writing stuff and its only partially complete like the completion 
> results tend to be empty string and then when you write a full word then the completion results tend to be meaningful. That's what I found in my local tests with the debugger.
> `c`  => Nothing on completion 
> `co`  => Nothing on completion 
> `con`  => Nothing on completion 
> `cons`  => Nothing on completion
> `const `  =>  shows  `"const variableName = something"`
> 
> The above example is what I usually found in my local testing with the debugger. 

I’m not sure I follow. What does this has to do with the prompt? When I try this locally, it does seem to give completions. Here's an example where the prefix is `con`, for example:

```
█ CompletionLogger:onComplete: {"type":"completion","status":"success","duration":807}  {
  "result": {
    "completion": "sol.log(bubbleSort([5,3,6,2,10]));",
    "stopReason": "stop_sequence"
  },
  "params": {
    "temperature": 0.5,
    "messages": [
      {
        "speaker": "human",
        "text": "You are a code completion AI that writes high-quality code like a senior engineer. You are looking at test.ts. You write code in between tags like this: <CODE5711>/* Code goes here */</CODE5711>."
      },
      {
        "speaker": "assistant",
        "text": "I am a code completion AI that writes high-quality code like a senior engineer."
      },
      {
        "speaker": "human",
        "text": "Complete this code: <CODE5711>function bubbleSort(arr) {\n  const len = arr.length;\n  for (let i = 0; i < len; i++) {\n    for (let j = 0; j < len - i - 1; j++) {\n      if (arr[j] > arr[j + 1]) {\n        let temp = arr[j];\n        arr[j] = arr[j + 1];\n        arr[j + 1] = temp;\n      }\n    }\n  }\n\n  return arr;</CODE5711>."
      },
      {
        "speaker": "assistant",
        "text": "Here is the code: <CODE5711>}\n\ncon"
      }
    ],
    "maxTokensToSample": 50,
    "stopSequences": [
      "\n\nHuman:",
      "</CODE5711>",
      "\n\n"
    ]
  }
}
```

--- 

One unrelated thing that I saw now though is that you add the suffix starting from the same line. In my experience with fill-in-middle based models like StarCoder or Code LLama (which we will be running A/B tests with very shortly btw., so hopefully they will provide much better suggestion that the chat based Claude Instant model. 🤞), I noticed that if the line already has a suffix (e.g. you're here: `function bubbleSort(█) {}` and the same line suffix is `) {}`), the models tend to not perform so well. Instead, what we do here is cut off the suffix after the cursor: https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/providers/unstable-fireworks.ts#L73

Might make completions better in the case where there is something after the cursor. 

---

> 
@philipp-spiess Here are some anecdotal thoughts from me

When I tried out this prompt change, it causes me to insert code comments much more often then desired.

So given that this new method takes into account the context AFTER the code the comments now would make a lot more sense than before coz the comment would be explaining the upcoming function. I hope that's what you meant and I hope the extra comment suggestions isn't problematic. Do let me know if you meant something else and I am happy to take a look.

It also returned no results more often than the previous implementation but I don't think this was necessarily a bad thing - It's just unexpected.

There is a very good reason for that behaviour and I am quite happy with it. In the previous code we do

text: `Complete this code: ${OPENING_CODE_TAG}${head.trimmed}${CLOSING_CODE_TAG}.`,
What this does is that it shows 1-2 lines before the position of the current cursor whereas now we do

${OPENING_CODE_TAG}
                ${head.trimmed}${OPENING_CODE_TAG}${tail.trimmed}${CLOSING_CODE_TAG}${this.options.docContext.suffix}`,
My guess is that because of tail.trimmed when you are writing stuff and its only partially complete like the completion results tend to be empty string and then when you write a full word then the completion results tend to be meaningful. That's what I found in my local tests with the debugger.
c => Nothing on completion
co => Nothing on completion
con => Nothing on completion
cons => Nothing on completion
const  => shows "const variableName = something"

The above example is what I usually found in my local testing with the debugger.

> Let me add a feature flag now.

Nice! Thank you

abeatrix (changes requested):
I tried to run autocomplete with the changes from your branch and noticed it does well with docstring, but not when completing functions or methods: 
image

It'd also suggest code that is already present in the file randomly, and ignore the suffix in some cases:

https://github.com/sourcegraph/cody/assets/68532117/224f3c34-d6ee-4d90-a4f5-3a4859ed1a3d

arafatkatze:
Closing in favour of the extraordinary work from @abeatrix  in https://github.com/sourcegraph/cody/pull/974  🙇

PR #918, opened 2023-09-04
Intelligent Cody: Use hover data for better completions context

efritz:
WIP

### DIFF
--- a/lib/shared/src/codebase-context/messages.ts
+++ b/lib/shared/src/codebase-context/messages.ts
@@ -24,6 +24,8 @@ export interface PreciseContext {
     symbol: {
         fuzzyName?: string
     }
+    forCompletions: boolean
+    hoverText: string[]
     definitionSnippet: string
     filePath: string
     range?: {
--- a/vscode/src/chat/GraphContextProvider.ts
+++ b/vscode/src/chat/GraphContextProvider.ts
@@ -14,7 +14,10 @@ export class GraphContextProvider implements GraphContextFetcher {
     }
 }
 
-const recursionLimit = 1
+// TODO(efritz) - move to options object
+const recursionLimit = 0
+const removeFileLocalResults = false
+const forCompletions = true
 
 /**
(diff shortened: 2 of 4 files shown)


vscode/src/chat/GraphContextProvider.ts
```diff
@@ -96,12 +112,12 @@ const getGraphContextFromSelection = async (
...
     const matches = dedupeWith(
-        definitionMatches
-            .map(({ symbolName, locations }) => locations.map(location => ({ symbolName, location })))
-            .flat(),
-        ({ location }) => locationKeyFn(location)
+        definitionMatches.map(({ locations, ...rest }) => locations.map(location => ({ location, ...rest }))).flat(),
+        ({ symbolName, location }) => `${symbolName}:${locationKeyFn(location)}`
     )
 
+    // TODO - see if we can remove fields of types we've also captured?
```
efritz, line 119: If we have a struct variable and use one of its fields, we'll return hover text for the struct AND the field. Not sure if we need both or if we can get away with just keeping the parent in such cases.

lib/shared/src/codebase-context/messages.ts
```diff
@@ -24,6 +24,8 @@ export interface PreciseContext {
     symbol: {
         fuzzyName?: string
     }
+    forCompletions: boolean
+    hoverText: string[]
```
valerybugakov, line 28: What's the purpose of the `forCompletions` boolean? If we want a data structure specific for completions, would it make sense to introduce a separate type?

philipp-spiess: I think this might just be for the prototype, in https://github.com/sourcegraph/cody/pull/865 I’m forking the top level entrypoint here specifically for autocomplete and then we can also fork the return struct.

efritz:
cc @philipp-spiess 

Closing in favor of #865 which cherry-picked these commits.

PR #919, opened 2023-09-04
E2E tests: Upload recording artifacts and fix flaky test

philipp-spiess:
This fixes the flaky E2E test. It tried to bring the Cody sidebar into the foreground but it was never hidden in the first place. I also added additional assertions about the `Hola` case that was not asserted on before.

This PR also brings back video uploads from failed cases so we can easily debug CI errors again.

## Test plan

- Checked with a failing E2E test

### DIFF
--- a/vscode/test/e2e/helpers.ts
+++ b/vscode/test/e2e/helpers.ts
@@ -1,4 +1,4 @@
-import { mkdir, mkdtempSync, readdirSync, rmSync, writeFile } from 'fs'
+import { mkdir, mkdtempSync, rmSync, writeFile } from 'fs'
 import { tmpdir } from 'os'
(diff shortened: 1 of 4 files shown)


vscode/test/e2e/helpers.ts
```diff
@@ -79,11 +79,11 @@ export const test = base
 
             // Delete the recorded video if the test passes
             if (testInfo.status === 'passed') {
-                rmdirSync(videoDirectory, { recursive: true })
+                rmSync(videoDirectory, { recursive: true })
```
philipp-spiess, line 82: There was a deprecation notice about this 🤷

taras-yemets (approved):

PR #926, opened 2023-09-05
Autocomplete: Add model specific feature flags

philipp-spiess:
As part of our eval we want to test more than one model. Adding this second switch before our release will allow us to save 1-2 days because we don't have to do a patch release when we want to change the default.

Also tested everything locally where it turns out it didn't work before because the default was never `null` which needs to be added to the enum specifically.

## Test plan

👀


valerybugakov (approved):
Yay 🔥

PR #468, opened 2023-07-31
Log events from E2E tests in dedicated testing environment

akalia25:
This PR implements event logging in our end-to-end tests to send data to a dedicated testing environment, separate from production database.

Collecting analytics on E2E test runs will enable the data team to verify that events are firing as expected for specific user flows. This will help ensure the accuracy and integrity of our production analytics data by allowing robust testing and validation of the event tracking implementation.

## Test plan
Tested locally to confirm events from E2E tests are flowing into dedicated testing pubsub topic. E2E test data can be found here: `sourcegraph-telligent-testing.e2e_testing.vscode`. Verified that telemetry is not sent to our prod database when in local debug mode.

PR #935, opened 2023-09-06
Fix dev build by not overwriting CODY_TESTING env

philipp-spiess:
#468 added a new mechanism to publish logs to a seperate env for test builds. However as part of it it added a new branch that sets `process.env.CODY_TESTING` to true in development builds which breaks the current assumption that this is only true in test runs.

It's important we don't set this flag in dev builds since it also changes a bunch of other small things in the extension so I’m reverting this here. Not to mention that this was also breaking the changes in #468 as my dev env was now flooded in these errors:

```
Error logging event Error: error fetching Testing Sourcegraph API: FetchError: request to http://localhost:49300/.api/testLogging failed, reason: connect ECONNREFUSED 127.0.0.1:49300 (http://localhost:49300/.api/testLogging)
    at /Users/philipp/dev/cody/lib/shared/src/sourcegraph-api/graphql/client.ts:523:29
    at processTicksAndRejections (/Users/philipp/dev/cody/lib/internal/process/task_queues.js:96:5)
    at SourcegraphGraphQLAPIClient.sendEventLogRequestToTestingAPI (/Users/philipp/dev/cody/lib/shared/src/sourcegraph-api/graphql/client.ts:398:36) {vslsStack: Array(3), stack: 'Error: error fetching Testing Sourcegraph API…src/sourcegraph-api/graphql/client.ts:398:36)', message: 'error fetching Testing Sourcegraph API: Fetc…00 (http://localhost:49300/.api/testLogging)'}
```

@akalia25 I don't think that the env overwrite is a necessary branch here anyways, can you confirm that this change makes sense?

## Test plan

No more errors in the console when running dev mode

### DIFF
--- a/vscode/src/services/EventLogger.ts
+++ b/vscode/src/services/EventLogger.ts
@@ -25,13 +25,9 @@ export async function createOrUpdateEventLogger(
     config: ConfigurationWithAccessToken,
     isExtensionModeDevOrTest: boolean
 ): Promise<void> {
-    if (config.telemetryLevel === 'off') {
+    if (config.telemetryLevel === 'off' || isExtensionModeDevOrTest) {
         eventLogger = null
         return
-    }
-
-    if (isExtensionModeDevOrTest) {
-        process.env.CODY_TESTING = 'true'
     }
 
     const { anonymousUserID, created } = await localStorage.anonymousUserID()


akalia25:
Hey @philipp-spiess, sorry for the confusion here, and thanks for the revert on this.
```
if (isExtensionModeDevOrTest) {
        process.env.CODY_TESTING = 'true'
    }
```
The reason for checking `isExtensionModeDevOrTest` and setting `CODY_TESTING` was to handle logging events during test runs. I was finding that both dev mode and tests execute in `vscode.ExtensionMode.Development`, so there was no way to distinguish them. The solution I attempted was to set CODY_TESTING to 'true' when in `isExtensionModeDevOrTest`, so test events could be logged.

The errors on the dev env were probably because the mock server is not running during regular development mode (which makes sense), so the connection was failing as it was trying to log to the test API. 

A potential fix for that error could be to update `client.ts` to check both CODY_TESTING and the server endpoint URL like:
```if (process.env.CODY_TESTING === 'true' && this.config.serverEndpoint === 'http://localhost:49300') then sendEventLogRequestToTestingAPI```

However, this would still break the assumption that CODY_TESTING is only true in test runs. 🤔 I'll think of a more refined solution; but if you have any suggestions would love to hear them :)!

philipp-spiess:
@akalia25 Are you refering to unit tests or what type of tests? We can make it so that every unit test has `CODY_TESTING` set if that would help 🤔

akalia25:
Hey @philipp-spiess, referring to the E2E tests `vscode/test/e2e/...`. I believe for those CODY_TESTING is already set to True? 🤔 .

I forgot to mention that I also made this change because eventLogger was not being initialized if 
```config.telemetryLevel === 'off' || isExtensionModeDevOrTest``` and so every time the tests were ran (isExtensionModeDevOrTest was true) and the eventLogger would not be created.

philipp-spiess:
> Hey @philipp-spiess, referring to the E2E tests vscode/test/e2e/.... I believe for those CODY_TESTING is already set to True? 🤔 

I would expect that to be the case, yeah. Have you observed something else?

> I forgot to mention that I also made this change because eventLogger was not being initialized if
config.telemetryLevel === 'off' || isExtensionModeDevOrTest and so every time the tests were ran (isExtensionModeDevOrTest was true) and the eventLogger would not be created.

That should be fine though because of the early return, no?

>     if (!eventLogger || !globalAnonymousUserID) {
>        return
>    }

akalia25:
> I would expect that to be the case, yeah. Have you observed something else?
All good there!

> That should be fine though because of the early return, no?
I believe we would need the event logger to be initialized in order for the events to get sent to the dedicated testing environment. so we would need it to invoke `eventLogger.log` and not have it return.

So, in summary:
1. Need the eventLogger to be initialized
2. Isolate that it is a E2E Test vs. dev mode
3. Route those events on client.ts to send to `sendEventLogRequestToTestingAPI`

dominiccooney:
>> I forgot to mention that I also made this change because eventLogger was not being initialized if
config.telemetryLevel === 'off' || isExtensionModeDevOrTest and so every time the tests were ran (isExtensionModeDevOrTest was true) and the eventLogger would not be created.
>
> That should be fine though because of the early return, no?

@philipp-spiess, @akalia25's goal in #468 was to test that event logging wasn't broken, by directing the telemetry logs to a sandboxed environment.

In hindsight we're missing clarity on what logs we want to send where. Question for @akalia25 and the analytics team.

Status quo:

- Testing (E2E, integration), which we have been dropping
- Development, which we have been logging in prod
- Real use &Rarr; also prod, naturally

Issue #910, opened 2023-09-04
Tree-sitter: truncate multiline completions based on parse-trees


valerybugakov:
Truncate multiline completions based on parse-trees. 

- Replace fragile indentation-based logic with the new implementation based on parse-trees produced by tree-sitter.
- Consider supporting more languages and use cases that the indentation-based approach didn't allow.

PR #938, opened 2023-09-06
Autocomplete: add multi-line trigger info to `docContext`

valerybugakov:
## Context

- Adds information about a multi-line completion trigger for a current completion to the `docContext` object. This data will be used in a follow-up PR to get a starting position for a first block-like statement in a completion.
- Part of #910 

## Test plan

Added unit tests.

### DIFF
--- a/vscode/src/completions/detect-multiline.ts
+++ b/vscode/src/completions/detect-multiline.ts
@@ -7,16 +7,28 @@ import {
     OPENING_BRACKET_REGEX,
 } from './text-processing'
 
+export type DetectMultilineResult =
+    | {
+          multiline: true
+          multilineTrigger: string
+      }
+    | {
+          multiline: false
+          multilineTrigger: undefined
+      }
+
+const MULTILINE_DISABLED_RESULT: DetectMultilineResult = { multiline: false, multilineTrigger: undefined }
+
 export function detectMultiline(
-    docContext: DocumentContext,
+    docContext: Omit<DocumentContext, 'multiline' | 'multilineTrigger'>,
     languageId: string,
     enableExtendedTriggers: boolean
-): boolean {
+): DetectMultilineResult {
     const { prefix, prevNonEmptyLine, nextNonEmptyLine, currentLinePrefix, currentLineSuffix } = docContext
 
     const languageConfig = getLanguageConfig(languageId)
     if (!languageConfig) {
-        return false
+        return MULTILINE_DISABLED_RESULT
     }
 
     const checkInvocation =
@@ -28,11 +40,15 @@ export function detectMultiline(
         !currentLinePrefix.trim().match(FUNCTION_KEYWORDS) &&
         checkInvocation.match(FUNCTION_OR_METHOD_INVOCATION_REGEX)
     ) {
-        return false
+        return MULTILINE_DISABLED_RESULT
     }
 
-    if (enableExtendedTriggers && currentLinePrefix.match(OPENING_BRACKET_REGEX)) {
-        return true
+    const openingBracketMatch = currentLinePrefix.match(OPENING_BRACKET_REGEX)
+    if (enableExtendedTriggers && openingBracketMatch) {
+        return {
+            multiline: true,
+            multilineTrigger: openingBracketMatch[0],
+        }
     }
 
     const { blockStart } = languageConfig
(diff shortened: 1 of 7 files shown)


vscode/src/completions/detect-multiline.ts
```diff
@@ -7,16 +7,28 @@ import {
     OPENING_BRACKET_REGEX,
 } from './text-processing'
 
+export type DetectMultilineResult =
+    | {
+          multiline: true
```
dominiccooney, line 12: Could we use the truthiness of multilineTrigger and have one property instead of two? Is it always `Boolean(multilineTrigger)` &hArr; `multiline`? If so let's simplify to one property—it can never get out of sync!

vscode/src/completions/get-current-doc-context.ts
```diff
@@ -36,6 +37,7 @@ export function getCurrentDocContext(
     position: vscode.Position,
     maxPrefixLength: number,
     maxSuffixLength: number,
+    enableExtendedTriggers: boolean,
```
dominiccooney, line 40: Make this an enum. It's too easy to mix up boolean parameters, and the meaning of the parameters is too mysterious at callsites without reading the definition, which contributes to errors.

vscode/src/completions/text-processing/parse-completion.ts
```diff
@@ -43,23 +50,35 @@ export function parseCompletion(context: CompletionContext): ParsedCompletion {
...
+        points.trigger = {
+            row: triggerPosition.line,
+            column: triggerPosition.character,
+        }
     }
 
     // Search for ERROR nodes in the completion range.
     const query = parser.getLanguage().query('(ERROR) @error')
-    const matches = query.matches(treeWithCompletion.rootNode, startPosition, endPosition)
+    // TODO: query bigger range to catch higher scope syntactic errors caused by the completion.
```
dominiccooney, line 76: Same feedback as above, it is great to drop a breadcrumb indicating what's blocking/when it's appropriate to attempt this/what the fitness function of fixing the TODO is.

dominiccooney (approved):
Looks good to me although I'm a tourist in autocomplete. Some suggestions inline.

vscode/src/completions/text-processing/parse-completion.ts
valerybugakov: Thanks for the review @dominiccooney!

 I'm actively working on the tree-sitter integration into the completions post-processing pipeline and plan to address these TODOs in one of the follow-up PRs. Going to mark them as short-lived 👍

vscode/src/completions/getInlineCompletions.ts
```diff
@@ -252,7 +251,7 @@ async function doGetInlineCompletions({
         document,
         docContext,
         position,
-        multiline,
+        multiline, // TODO: drop in favor of docContext.
```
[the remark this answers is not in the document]
valerybugakov: Address in the follow-up PR: https://github.com/sourcegraph/cody/pull/879/commits/cfdc9d934194796152ff24221d248a7150218c63

vscode/src/completions/text-processing/parse-completion.ts
valerybugakov: Addressed in the follow-up PR: https://github.com/sourcegraph/cody/pull/879/commits/cfdc9d934194796152ff24221d248a7150218c63

vscode/src/completions/detect-multiline.ts
```diff
@@ -7,16 +7,28 @@ import {
...
+          multilineTrigger: string
```
philipp-spiess: Why do we need to expose this now? Will this help in post-production?

vscode/src/completions/get-current-doc-context.ts
valerybugakov: Agreed! Converted a list of arguments to an object https://github.com/sourcegraph/cody/pull/938/commits/0c42b605c65aacb5cd4f6e83f76006ba2936c460

philipp-spiess (approved):
Makes sense! I like the direction of moving more into the context, this way we will have it easier to pass all the necessary stuff around 👍

vscode/src/completions/detect-multiline.ts
valerybugakov: Yep

> This data will be used in a follow-up PR to get a starting position for a first block-like statement in a completion

valerybugakov: I'm using it here: https://github.com/sourcegraph/cody/pull/879

PR #943, opened 2023-09-06
Autocomplete: Tweaks to the graph context to make it actually slightly usable already

philipp-spiess:
After testing locally, here are some tweaks that makes graph context for autocomplete much more usable (although not as powerful - yet!)

- We removed the recursion limit so it's faster and far fewer irrelevant symbols are included
- Instead of Go To Type Defs, we default to Go To Defs now because in a lot of cases, Go To Type Defs wouldn't work very well. We need a heuristics to decide which one to follow.
- Rewrite the markdown extracting logic because I build 💩 

The result is that I had the first autocomplete request that included the proper props for a React component that I never had opened or anything. Just because we found the right definitions 💥  (The completion is still wrong because the dummy objct is defined in the suffix and Anthropic doesn't read the suffix yet. Baby steps 😬)

## Test plan

Screenshot 2023-09-06 at 18 30 06

### DIFF
--- a/vscode/src/graph/graph.ts
+++ b/vscode/src/graph/graph.ts
@@ -57,8 +57,8 @@ export const getGraphContextFromRange = async (
     const contexts = await getGraphContextFromSelection(
         [{ uri, range }],
         new Map([[uri.fsPath, editor.document.getText().split('\n')]]),
-        1, // recursion limit
-        true // use type definitions
+        0, // recursion limit
+        false // use type definitions
     )
 
     return contexts
@@ -456,7 +456,8 @@ export const extractDefinitionContexts = async (
 
         if (contentPromise && documentSymbolsPromises) {
             const content = contentPromise
-            const documentSymbols = await documentSymbolsPromises // NOTE: already resolved)
+            const documentSymbols = await documentSymbolsPromises // NOTE: already resolved
+
             const definitionSnippets = extractSnippets(content, documentSymbols, [range])
 
             for (const definitionSnippet of definitionSnippets) {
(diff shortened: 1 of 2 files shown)


valerybugakov (commented):
> We removed the recursion limit so it's faster and far fewer irrelevant symbols are included

Without all the implementation details, it's not apparent why this change improved the speed and quality.

vscode/src/graph/graph.ts
```diff
@@ -57,8 +57,8 @@ export const getGraphContextFromRange = async (
     const contexts = await getGraphContextFromSelection(
         [{ uri, range }],
         new Map([[uri.fsPath, editor.document.getText().split('\n')]]),
-        1, // recursion limit
-        true // use type definitions
+        0, // recursion limit
```
valerybugakov, line 60: Did we have it in the first place to reduce latency?

valerybugakov (approved):
Exciting example in the description!

vscode/src/graph/graph.ts
philipp-spiess: The recursion limit sets up how often we recurse the Go To Def algorithm. 

if the limit is 1, we do one recursive call. So first, we extract all symbols of a given range, then we do a Go To Def to get their definitions, then we recurse and again extract all symbols in the given ranges returned from the previous Go To Def calls and get their definitions again.

This obviously makes it slower (it's usually exponentially more symbols after every recursion) but what I found is that it also just returns too many symbols so that the important ones (the one you're working on right now) were diluted in a sea of implementation details.

We'll have to find a good compromise and heuristics to organize this. Maybe we want to exhaust type definitions and only do one level of implementations or so? I don't know yet. 🦭

PR #948, opened 2023-09-07
E2e test logging patch1

akalia25:
Patch to #468 and #935,  to add logic so that on dev mode it does not try to log events to the testing API and only e2e test events are logged.

## Test plan
tested locally to ensure only e2e tests are sending data and debug console is not showing errors


philipp-spiess (approved):
Noice

PR #963, opened 2023-09-07
feat: custom starter message configuration

abeatrix:
RE: https://sourcegraph.slack.com/archives/C05AGQYD528/p1693983800785199

This PR adds a new configuration `cody.customer.starter`  to allow setting a custom starter message that will prepend chat requests to Cody:
- Configuration now has a `cody.customer.starter` string property.
- A new setting cody.custom.starter was added in vscode/package.json to allow configuring a custom message.
- getConfiguration() was updated to read the new customStarter config value.
- ConfigurationWithAccessToken gained a customStarter property.
- main.ts imports PromptMixin to set the custom starter message on extension start.

## Test plan

1. set `cody.customer.starter` in your user config, e.g. "call me jeff"
2. reload extension
3. submit a chat request
4. check the output channel to see your setting was added to the chat request

Screenshot 2023-09-06 at 9 03 38 PM


vscode/package.json
```diff
@@ -811,6 +811,14 @@
           "default": false,
           "markdownDescription": "Experimental feature for internal use."
         },
+        "cody.custom.starter": {
```
valerybugakov, line 814: Many settings are grouped under feature categories (e.g., `cody.autocomplete.*` or `cody.inlineChat.*`). Would it make sense to apply this pattern here too?

valerybugakov (approved):

vscode/package.json
toolmantim: What about `cody.chat.preInstruction` with the description "An instruction for Cody to be included at the start of all chat messages sent"

That is assuming it's chat messages only?

toolmantim (approved):
Great! I left a suggestion for a new config name.

vscode/package.json
abeatrix: Yea autocompletes uses a different prompt so this is only for chat

PR #964, opened 2023-09-07
Show commands format changed notification

taras-yemets:
If user/workspace Cody config file contains commands in old format, show notification with accept/ignore config update options. If user accepts, overwrite the config file and open it.

Previously we updated user/workspace cody config file if commands were in old format and then informed user about this change. This behavior may be annoying as there is no way for users to avoid these changes: if user discards or stashes config changes, these changes would be reapplied next time  user interacts with the main quick pick or sidebar chat ([thread](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1694010637819529)).

## Test plan
CI passes. Tested manually (video attached).

### DIFF
--- a/vscode/src/custom-prompts/CustomPromptsStore.ts
+++ b/vscode/src/custom-prompts/CustomPromptsStore.ts
@@ -114,7 +114,8 @@ export class CustomPromptsStore implements vscode.Disposable {
             if (isOldFormat) {
                 void vscode.window
                     .showInformationMessage(
-                        `Commands format has changed. To be able to use custom commands, we suggest updating your ${type} configuration to the new format.`,
+                        `Your custom commands ${type} JSON (${type === 'user' ? '~/.vscode/cody.json' : '.vscode/cody.json'}) is using an old format, and needs to be upgraded.`,
+                        'Upgrade JSON',
                         'Update',
                         'Ignore'
                     )
--- a/vscode/src/custom-prompts/CustomPromptsStore.ts
+++ b/vscode/src/custom-prompts/CustomPromptsStore.ts
@@ -116,7 +116,7 @@ export class CustomPromptsStore implements vscode.Disposable {
                     .showInformationMessage(
                         `Your custom commands ${type} JSON (${type === 'user' ? '~/.vscode/cody.json' : '.vscode/cody.json'}) is using an old format, and needs to be upgraded.`,
                         'Upgrade JSON',
-                        'Update',
+                        'Upgrade JSON',
                         'Ignore'
                     )
                     .then(choice => {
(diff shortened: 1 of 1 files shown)


taras-yemets:
When the config file contains commands in the old format, we show the notification each time the user interacts with the main quick pick or the sidebar chat. It may be too intrusive. Would showing this notification once in an hour for example be a better option? WDYT, @toolmantim?

vscode/src/custom-prompts/CustomPromptsStore.ts
```diff
@@ -112,33 +112,39 @@ export class CustomPromptsStore implements vscode.Disposable {
...
-                    {}
-                )
-                // write transformed commands to the corresponding config file
-                await this.updateJSONFile({ ...json, commands }, type)
-                // inform user about this change
                 void vscode.window
                     .showInformationMessage(
-                        `Your Cody ${type} configuration file has been automatically updated to the new format.`,
-                        'Open File'
+                        `Commands format has changed. To be able to use custom commands, we suggest updating your ${type} configuration to the new format.`,
```
toolmantim, line 117:
```suggestion
                        `Your custom commands ${type} JSON (${type === 'user' ? '~/.vscode/cody.json' : '.vscode/cody.json'}) is using an old format, and needs to be upgraded.`,
                        'Upgrade JSON',
```

vscode/src/custom-prompts/CustomPromptsStore.ts
```diff
@@ -112,33 +112,39 @@ export class CustomPromptsStore implements vscode.Disposable {
...
+                        'Update',
```
toolmantim, line 118:
```suggestion
                        'Upgrade JSON',
```

vscode/src/custom-prompts/CustomPromptsStore.ts
```diff
@@ -112,33 +112,39 @@ export class CustomPromptsStore implements vscode.Disposable {
...
+                        'Ignore'
                     )
                     .then(choice => {
-                        if (choice === 'Open File') {
-                            const filePath = type === 'user' ? this.jsonFileUris.user : this.jsonFileUris.workspace
-                            if (filePath) {
-                                void vscode.window.showTextDocument(filePath)
-                            }
+                        if (choice === 'Update') {
```
toolmantim, line 122:
```suggestion
                        if (choice === 'Upgrade JSON') {
```

toolmantim (approved):
Works great! Updated some UI labelling based on my local testing.

I did find that if you cmd-z in the file after upgrade, you get some weird behaviour ("Unable to create file '/Users/tim/Codez/handbook/.vscode/cody.json' that already exists when overwrite flag is not set") but seems unrelated to this PR.

PR #966, opened 2023-09-07
Autocomplete: Tweak StarCoder temperature

philipp-spiess:
This changes the default temperature for StarCoder from `0.4` to `0.2` which is also used in [their own testing](https://github.com/bigcode-project/bigcode-evaluation-harness). In my experiments, this made the model behave a bit more predictable while still providing enough variation. 

**CONTROVERSIAL:** I also tweaked the example Bee added in #945. I’m happy to revert though. StarCoder was not completing this and I think it was because there are two reference function in that file that are implemented the same way with the same line missing so starcoder did pattern match and thought that this is the expected result. I feel that the example is a bit contrived so I changed it into something that I would encounter while developing. 

## Test plan

Ran the completions test suite and made sure the output looked good:

- `cd vscode`
- `pnpm run generate:completions`

### DIFF
--- a/vscode/src/completions/providers/unstable-fireworks.ts
+++ b/vscode/src/completions/providers/unstable-fireworks.ts
@@ -100,7 +100,9 @@ export class UnstableFireworksProvider extends Provider {
             // To speed up sample generation in single-line case, we request a lower token limit
             // since we can't terminate on the first `\n`.
             maxTokensToSample: this.options.multiline ? 256 : 30,
-            ...getModelConfig(this.model),
+            temperature: 0.2,
+            topP: 0.95,
+            topK: 0,
             model: MODEL_MAP[this.model],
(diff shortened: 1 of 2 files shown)


vscode/src/completions/providers/unstable-fireworks.ts
```diff
@@ -235,16 +235,18 @@ function getSuffixAfterFirstNewline(suffix: string): string {
...
         return {
             temperature: 0.2,
             topP: 0.95,
+            topK: 0,
         }
     }
 
     return {
-        temperature: 0.4,
+        temperature: 0.2,
```
valerybugakov, line 248: I'm in favor of this change! (based on my anecdotal experience of course)

valerybugakov (approved):

Issue #923, opened 2023-09-05
Autocomplete: Heuristics on when to show a completion


philipp-spiess:
It currently happens quite often that we show a completion only for a brief moment because the user is still typing or doesn't need one at all right now. Maybe we can find ways to hold back on doing that unless we're more certain a completion should be shown? Maybe the cache/resolve-with-previous-request logic would need to have an additional timeout of a few hundred MS after the last keystroke before it can be shown (something that is smaller than our desired minimum latency of course).

philipp-spiess:
In #973 I’m adding some more useful logging so that everyone can help investigate here and find out when a completion was accepted that isn't even being read. Let's see if we can find heuristics to prevent them.

philipp-spiess:
If anyone has any other ideas then #1017 I’m all ears 👀

PR #973, opened 2023-09-07
Autocomplete: Add bare bones statistics logging and UI

philipp-spiess:
I often find myself wanting to know when exactly we count something as suggested vs. when we count it as accepted. Looking at the output tab is not super helpful here either so I added a small UI to the trace view.

This also adds the "infra" to support the trace view to show other views that re-render which I want to use for the section observer shortly.

I’m specifically hoping that this will help with #923.

## Test plan

https://github.com/sourcegraph/cody/assets/458591/2279928c-4056-4d0e-9f56-d5bdff6dbc03

### DIFF
--- a/vscode/src/completions/logger.ts
+++ b/vscode/src/completions/logger.ts
@@ -50,7 +50,7 @@ export interface CompletionEvent {
     acceptedAt: number | null
 }
 
-const READ_TIMEOUT = 750
+const READ_TIMEOUT_MS = 750
 
 const displayedCompletions = new LRUCache<string, CompletionEvent>({
     max: 100, // Maximum number of completions that we are keeping track of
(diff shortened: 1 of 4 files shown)


vscode/src/completions/logger.ts
```diff
@@ -110,6 +117,9 @@ export function loaded(id: string): void {
 // Suggested completions will not be logged immediately. Instead, we log them when we either hide
 // them again (they are NOT accepted) or when they ARE accepted. This way, we can calculate the
 // duration they were actually visible for.
+//
+// For statistics logging we start a timeout matching the READ_TIMEOUT so we can increment the
```
dominiccooney, line 121: Realize this isn't a new name, but it is nice to put the units on quantities ie READ_TIMEOUT_MSEC because it can make it easier to catch some clowntown with dimensions in review.

vscode/src/completions/tracer/traceView.ts
```diff
@@ -167,6 +185,13 @@ ${codeDetailsWithSummary('JSON for dataset', jsonForDataset(data))}
     return renderMarkdown(markdownSource, { noDomPurify: true })
 }
 
+function statisticSummary(): string {
+    const { accepted, suggested } = statistics.getStatistics()
+    return `📈 Suggested: ${suggested} | Accepted: ${accepted} | Acceptance rate: ${
+        suggested === 0 || accepted === 0 ? 'N/A' : `${((accepted / suggested) * 100).toFixed(2)}%`
```
dominiccooney, line 191: You could still calculate an acceptance rate with `accepted === 0`, right?

dominiccooney (approved):

vscode/src/completions/tracer/traceView.ts
philipp-spiess: Haha yes of course 🤦

PR #975, opened 2023-09-07
Autocomplete: Add document section logging to trace view

philipp-spiess:
The section observer keeps track of visible documents and their notable sections. While working on it, we need to be able to see which sections are available and preloaded, etc.

This new view helps with that, check out the video for an example.

## Test plan

https://github.com/sourcegraph/cody/assets/458591/31b9fe1b-6f14-4ba7-95e2-e607b865cce4

### DIFF
--- a/vscode/src/completions/utils.ts
+++ b/vscode/src/completions/utils.ts
@@ -45,22 +45,21 @@ export function forkSignal(signal: AbortSignal): AbortController {
 /**
  * Creates a simple subscriber that can be used to register callbacks
  */
-type Callback<T> = (value: T) => void
+type Listener<T> = (value: T) => void
 interface Subscriber<T> {
-    subscribe(callback: Callback<T>): () => void
+    subscribe(listener: Listener<T>): () => void
     notify(value: T): void
 }
 export function createSubscriber<T>(): Subscriber<T> {
-    const callbacks: Set<Callback<T>> = new Set()
-    function subscribe(callback: Callback<T>): () => void {
-        callbacks.add(callback)
-        return () => callbacks.delete(callback)
+    const listeners: Set<Listener<T>> = new Set()
+    function subscribe(listener: Listener<T>): () => void {
+        listeners.add(listener)
+        return () => listeners.delete(listener)
     }
 
     function notify(value: T): void {
-        for (const callback of callbacks) {
-            // eslint-disable-next-line callback-return
-            callback(value)
+        for (const listener of listeners) {
+            listener(value)
         }
     }
 
--- a/vscode/src/completions/utils.ts
+++ b/vscode/src/completions/utils.ts
@@ -41,3 +41,31 @@ export function forkSignal(signal: AbortSignal): AbortController {
     signal.addEventListener('abort', () => controller.abort())
     return controller
 }
+
+/**
+ * Creates a simple subscriber that can be used to register callbacks
+ */
+type Callback<T> = (value: T) => void
+interface Subscriber<T> {
+    subscribe(callback: Callback<T>): () => void
+    notify(value: T): void
+}
+export function createSubscriber<T>(): Subscriber<T> {
+    const callbacks: Set<Callback<T>> = new Set()
+    function subscribe(callback: Callback<T>): () => void {
+        callbacks.add(callback)
+        return () => callbacks.delete(callback)
+    }
+
+    function notify(value: T): void {
+        for (const callback of callbacks) {
+            // eslint-disable-next-line callback-return
+            callback(value)
+        }
+    }
+
+    return {
+        subscribe,
+        notify,
+    }
+}
--- a/vscode/src/graph/section-observer.test.ts
+++ b/vscode/src/graph/section-observer.test.ts
@@ -80,7 +80,7 @@ describe('SectionObserver', () => {
                 ] satisfies PreciseContext[]
         )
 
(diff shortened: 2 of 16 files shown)


vscode/src/graph/section-observer.ts
```diff
@@ -59,6 +65,12 @@ export class SectionObserver implements vscode.Disposable, GraphContextFetcher {
         private getDocumentSections: typeof defaultGetDocumentSections = defaultGetDocumentSections,
         private getGraphContextFromRange: typeof defaultGetGraphContextFromRange = defaultGetGraphContextFromRange
     ) {
+        if (singleton) {
+            throw new Error('SectionObserver already initialized')
+        }
+        // eslint-disable-next-line @typescript-eslint/no-this-alias, unicorn/no-this-assignment
```
philipp-spiess, line 71: We have very stupid lint rules 😡

vscode/src/graph/section-observer.ts
```diff
@@ -119,7 +135,7 @@ export class SectionObserver implements vscode.Disposable, GraphContextFetcher {
         const lines: string[] = []
         // eslint-disable-next-line ban/ban
         this.activeDocuments.forEach(document => {
-            lines.push(document.uri)
+            lines.push(path.normalize(vscode.workspace.asRelativePath(document.uri)))
```
philipp-spiess, line 138: Unclear why this doesn't make the path pretty? 😐

vscode/src/completions/utils.ts
```diff
@@ -41,3 +41,31 @@ export function forkSignal(signal: AbortSignal): AbortController {
...
+export function createSubscriber<T>(): Subscriber<T> {
+    const callbacks: Set<Callback<T>> = new Set()
+    function subscribe(callback: Callback<T>): () => void {
+        callbacks.add(callback)
+        return () => callbacks.delete(callback)
+    }
+
+    function notify(value: T): void {
+        for (const callback of callbacks) {
+            // eslint-disable-next-line callback-return
```
philipp-spiess, line 62: ????

dominiccooney: The intent is to treat calling a callback as the end of control flow.

Just call the variable something other than "callback" and go on with your life 😉

vscode/src/graph/section-observer.test.ts
```diff
@@ -78,17 +82,23 @@ describe('SectionObserver', () => {
 
         sectionObserver = new SectionObserver(
             {
-                onDidChangeVisibleTextEditors: (_onDidChangeVisibleTextEditors: any) =>
-                    (onDidChangeVisibleTextEditors = _onDidChangeVisibleTextEditors),
-                onDidChangeTextEditorSelection: (_onDidChangeTextEditorSelection: any) =>
-                    (onDidChangeTextEditorSelection = _onDidChangeTextEditorSelection),
+                onDidChangeVisibleTextEditors: (_onDidChangeVisibleTextEditors: any) => {
+                    onDidChangeVisibleTextEditors = _onDidChangeVisibleTextEditors
```
dominiccooney, line 86: Might be worth commenting what's happening here. You're caching these to fire them manually later, is that right?

vscode/src/graph/section-observer.test.ts
```diff
@@ -97,10 +107,13 @@ describe('SectionObserver', () => {
         // for it to finish loading.
         await nextTick()
     })
+    afterEach(() => {
+        sectionObserver.dispose()
+    })
 
     it('loads visible documents when it loads', () => {
         expect(sectionObserver.debugPrint()).toMatchInlineSnapshot(`
-          "file:///document1.ts
```
dominiccooney, line 103: Just curious but why did we lose //?

vscode/src/graph/section-observer.ts
dominiccooney: True. If you made a factory function which does the check, construction and memo that's a nice separation of concerns, WDYT?

dominiccooney (approved):
Looks good, some suggestions inline.

Agree that lints are taking some of the fun out of JavaScript hacking. But on the whole I guess I like having them more than not having them.

vscode/src/completions/utils.ts
philipp-spiess: Yeah you're right, I was just annoyed because I noticed this after I decided for whatever reason to rename all variables from `listener(s)` to `callback(s)` and then this error popped up

vscode/src/graph/section-observer.test.ts
philipp-spiess: That's exactly right, I will note that.

vscode/src/graph/section-observer.test.ts
philipp-spiess: Since this is just used for visualisation, I've ran that URI through `path.normalize(vscode.workspace.asRelativePath(document.uri))` but for some reason this doesn't work fully. I think it is because it assumes that the parameter is a path and not a full URI so it can't handle protocols properly. I can try to parse the URI and then just use the path.

Again, just want to make the UI for debugging prettier since it's hard to scan for me if it contains the `file:///Users/philipp/dev` as a prefix all the time 🦭

vscode/src/graph/section-observer.ts
philipp-spiess: Yeah agree

PR #982, opened 2023-09-07
Update VS Code README.md

iskyOS:
I opened this small PR to fix a couple small issues on the VS Code marketplace page for Cody:

- 2 of the images were up against one another - I removed one.
- There was some duplicate copy in 2 parts of the page (under "Chat" and "Code graph"), so I updated one section.

## Test plan

README only.


iskyOS:
@toolmantim, are you the right person to continue reviewing PRs against the marketplace pages, or would you prefer I share these with someone else on the team? Lmk if you'd rather I not send these your way!

toolmantim:
@iskyOS yep, I’m happy to be the one!

toolmantim (approved):
Nice!

PR #983, opened 2023-09-07
Intelligent Cody: Supercharge hover text

efritz:
For the completions path, we want to be able to fill in the _meta_ details about symbols (shapes of interfaces, classes, etc) and not necessarily the full implementation. This context seems to be enough to give a boost to the completion engine.

This PR updates the completions code path to return the _hover text_ which provides sufficient information for Go and TypeScript (and possibly other languages). We go further in our approach by not only selecting the hover text of the symbol itself (which should be equivalent to the *definition*'s hover text), but also the hover text of associated type definitions and implementations.

## Test plan

Updated unit tests, manual testing.

image


philipp-spiess (approved):

vscode/src/graph/graph.ts
```diff
@@ -54,14 +56,35 @@ export const getGraphContextFromRange = async (
...
+        contentMap,
+        requestCandidates,
+        defaultGetHover,
+        defaultGetDefinitions,
+        defaultGetTypeDefinitions,
+        defaultGetImplementations
     )
 
-    return contexts
+    console.log({ resolvedHoverText })
```
efritz, line 82: @philipp-spiess I've tried to re-implement something a bit more specific to completions (using hover-text only) here. I think we might want a non-PreciseContext type to be passed back to the section observer and LLM context. Could you help me out enabling whatever debug vies you find most useful in local dev so I can have a quicker iteration cycle (we can do this live in our meeting Monday as well)?

valerybugakov (commented):
Read through everything that changed here. Going to do a second pass later today.

lib/shared/src/codebase-context/messages.ts
```diff
@@ -35,6 +35,23 @@ export interface PreciseContext {
...
 
+export interface HoverContext {
+    symbolName: string
+    filePath: string
+    range?: {
+        startLine: number
+        startCharacter: number
+        endLine: number
+        endCharacter: number
+    }
```
philipp-spiess, line 46: This `range` is the range of the symbol we create this hover context for, correct? I wonder if we should instead try to include ranges for every hover content we include here. The idea is that if e.g the type definition is already inside the prompt prefix/suffix, we don't need to add it as context. 

E.g. consider this case:

```
interface Foo {
  bar(): void;
}

function foo(foo: Foo) {
  █
}
```

In this case, the prefix might already contain the first line (since we include many lines of the document) and including the type definition (or any information, really) for `Foo` as additional context would only duplicate that info in the prompt.

My idea is that if we can find ranges for this, we could see that e.g. the type def, def, and impl of `Foo` is already inside the prefix and would thus be able to skip it.

vscode/src/completions/context/graph-section-observer.test.ts
```diff
@@ -66,20 +66,18 @@ describe('GraphSectionObserver', () => {
...
                         range: { startCharacter: 0, startLine: 0, endCharacter: 0, endLine: 10 },
                     },
                     {
-                        symbol: { fuzzyName: 'bar' },
-                        definitionSnippet: 'function bar() {}',
-                        filePath: document1Uri.fsPath,
-                        hoverText: [],
+                        symbolName: 'bar',
+                        hovers: [{ type: 'definition', content: ['function bar() {}'] }],
+                        filePath: document1Uri.toString(),
```
philipp-spiess, line 77: If we change this field to contain the full URI, we should probably rename it from `filePath` to `uri`

vscode/src/completions/context/graph-section-observer.ts
```diff
@@ -240,7 +239,11 @@ export class GraphSectionObserver implements vscode.Disposable, GraphContextFetc
     /**
      * A pretty way to print the current state of all cached sections
      */
-    public debugPrint(selectedDocument?: vscode.TextDocument, selections?: readonly vscode.Selection[]): string {
+    public debugPrint(
+        selectedDocument?: vscode.TextDocument,
+        selections?: readonly vscode.Selection[],
+        showLastVisited = true
```
philipp-spiess, line 245: I made it so that the last visited sections are not logged if you have none to avoid the branching here so I think we can revert this (this is also not used after the rebase it seems)

philipp-spiess (commented):
I'll continue the rest of the review when I’m back, sorry 😬

lib/shared/src/codebase-context/messages.ts
efritz: Ok so the types are actually constructed in two parts:

- "definition" hovers, which points to the go-to-def range
- "type definition" hovers which points to the go-to-typedef range, and also includes implementation hovers (but we don't care about those ranges)

We really only care about filtering out definitions overlapping a given range, we should always want to include type definitions+implementations as it's unlikely that we'll show type declarations other implementing definititons that are already within the given user scope.

efritz: (Trying something out, expect an update)

philipp-spiess: Yeah I think this makes sense 👍

philipp-spiess (approved):
This is good stuff! Looking forward to doing a bunch of testing with that :)

vscode/src/completions/context/graph-section-observer.ts
```diff
@@ -449,46 +452,39 @@ export class GraphSectionObserver implements vscode.Disposable, GraphContextFetc
...
-        } else if (isCodeBlockDelimiter && isCodeBlock) {
-            isCodeBlock = false
-        } else if (isCodeBlock) {
-            codeBlocks.push(line)
-        }
+    if (nonDefinitionHovers.length > 0) {
+        snippets.push({
+            fileName: path.normalize(vscode.workspace.asRelativePath(context.filePath)),
+            symbol: context.symbolName,
+            content: nonDefinitionHovers.map(h => h.content.join('\n').trim()).join('\n\n'),
```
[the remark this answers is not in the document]
efritz: I have another handful of heuristics I think we'll want (Go implementations will include other interfaces, which isn't very useful).

PR #984, opened 2023-09-08
feat: insert new unit tests to file

abeatrix:
part of https://github.com/sourcegraph/cody/pull/907

PR includes the following changes:
- move returned unit tests into a test file in the current directory
- create test file if it doesn't exists
- if it exists, try remove duplicate import statements, before adding new tests to the line after the import statements to make sure the new import statements are also on top
- cleaned up prompt

https://github.com/sourcegraph/cody/assets/68532117/610bce25-ab02-4fcb-a4b7-4700f71b62b9

Note: 
- Sometimes Cody will miss importing the libraries and packages in use; i have encountered the same issue from all LLMs I've tried so far.  
- Additional changes might be required when asking Cody to generate tests for functions with lots of imports and complex set ups

## Test plan

Try running the test command for a function with existing unit tests in the same directory.

### Demo on go, python, and typescript files

https://github.com/sourcegraph/cody/assets/68532117/4db093c6-ce69-44c9-95e5-4ffbab008620


toolmantim (commented):
Exciting!

I gave it a go with the handbook, but it seemed to swallow the code:

https://github.com/sourcegraph/cody/assets/153/d3228fea-166b-4372-88a8-1a29d97a0471

I had more luck adding a test to some about site JS (which doesn't have any tests), and it produced a test but the file created doesn't match the comment at the top of the file for some reason:

Screenshot 2023-09-11 at 5 36 17 pm

Testing it with docsite and some go code I got better results (but I experienced a bug first time I ran it if you switch directories with the same chat open, it seems to not run the command… I had to run it twice?):

Screenshot 2023-09-11 at 5 34 41 pm

I can do some more testing… might need some others to do it too.

From a design POV, we might need to try opening a file and inserting the contents as it's sent from Cody? It feels much slower than the chat response, mostly because I think you can't see the response streaming in so you just sit there waiting (sometimes mid-response) with no updates.

abeatrix:
@toolmantim thanks for testing!

> Testing it with docsite and some go code I got better results (but I experienced a bug first time I ran it if you switch directories with the same chat open, it seems to not run the command… I had to run it twice?):

This seems to be a bug on main and not related to this command? Can you confirm if this behavior is reproducible when you run other commands?

> From a design POV, we might need to try opening a file and inserting the contents as it's sent from Cody? It feels much slower than the chat response, mostly because I think you can't see the response streaming in so you just sit there waiting (sometimes mid-response) with no updates.

I am currently using what we have already implemented for the fixup logic, which is to insert content to file once the response from cody is completed. We will need to implement something else if we want to stream incompleted response to file directly, which I assume will need to take a sprint (or two) time. 

In this case, do we still want to add the tests to the file with a slower time, or wait until we have implemented the streaming incompleted response to file?

abeatrix:
Closing as per discussion with @toolmantim 

> just always output into chat for now and revisit inserting or creating temp files after Oct 4.

toolmantim:
Yeah based on this work it looks like streaming or non-stop-fixup style UI would be needed… and sounds like the sprints after Oct 4 will be perfect time to explore it.

> Or do we want to go back to adding a new button for inserting to new file instead so we can keep the code in the UI, and user can choose if they want to insert the test to file or not?

Yep, I reckon we do this. Any improvements to the chat sidebar & code block interactions are still a good idea, as they'll be useful for chat and almost every command aside from `/edit`.

PR #986, opened 2023-09-08
Autocomplete: Various section observer tweaks

philipp-spiess:
Just a few assorted graph context tweaks:

- We now extract the range of everything in the prefix and suffix. This is helpful so we can disregard definitions that are already in our context anyways. 
- Remove the `.d.ts` logic since it doesn't make sense to special case. Instead, we should fix this with other heuristics
- Add tests for the document context logic 

## Test plan

Tests green, this is not shipped yet anyways.

### DIFF
--- a/vscode/src/completions/context/context.ts
+++ b/vscode/src/completions/context/context.ts
@@ -24,7 +24,11 @@ export interface GetContextOptions {
 }
 
 export interface GraphContextFetcher {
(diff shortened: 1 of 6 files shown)


philipp-spiess:
@sourcegraph/cody-vscode 🥺👉👈

efritz (approved):

PR #992, opened 2023-09-11
UX improvements to the custom command workflow

toolmantim:
UX improvements to the custom command workflow:
* Improved labels
* Gives you the choice of where to store the custom command
* Saved notification now has a button for opening the settings and inspecting the results
* Fixes bug where you can incorrectly choose the same slash name as an existing command
* Move examples from the generated config to the new [Custom Commands docs](https://docs.sourcegraph.com/cody/custom-commands)

Screenshot 2023-09-11 at 8 56 39 am

custom-command-setup-6

Screenshot 2023-09-11 at 9 25 39 am

## Requires

- [x] https://github.com/sourcegraph/sourcegraph/pull/56489

## Test plan

- Opened the custom commands quickpick
- Manually ran through the new command builder flow
- Clicked new buttons, inspected JSON


vscode/src/chat/protocol.ts
```diff
@@ -39,7 +39,7 @@ export type WebviewMessage =
           value?: string
       }
     | { command: 'abort' }
-    | { command: 'custom-prompt'; title: string; value?: CodyPromptType }
+    | { command: 'custom-prompt'; title: string; value?: 'user' | 'workspace' }
```
abeatrix, line 42: i have't built from the branch yet, but i think this should includes 'default' as the prompt type?

dominiccooney: @abeatrix How would we have a custom prompt with type `'default'`? See `isNonCustomType` etc.

dominiccooney (approved):
This looks good to me.

It would be better if we added a type like:

```
type CustomPromptType = 'user' | 'workspace'
```

and referred to that everywhere instead of repeating `'user' | 'workspace'`. That will considerably reduce toil if we add a third kind of custom command, maybe ones synced with your whole Enterprise or something, because we can expand the type in one place and follow the type errors.

vscode/src/chat/protocol.ts
abeatrix: Oh I didn't see the change from CodyPromptType to isNonCustomType my bad :D

toolmantim:
@dominiccooney thanks! Updated to use `CustomCommandType` in [1c96984](https://github.com/sourcegraph/cody/pull/992/commits/1c96984f66c585a97ca0f10a2b6f3510e2373806).

PR #998, opened 2023-09-11
Agent: add notification that clears the last autocomplete candidate

chwarwick:
Clearing the last candidate ensures that autocomplete id's and cached potential suggestions are not reused.  This adds a notification an extension can send when it no longer wishes the consider the last candidate.  The most common reason is that the suggestion was accepted. 
## Test plan
Tested using JetBrains extension 
Accepted an autocomplete and then immediately removed it and triggered another 
Got the same suggestion but verified the id was different.

### DIFF
--- a/vscode/src/completions/vscodeInlineCompletionItemProvider.ts
+++ b/vscode/src/completions/vscodeInlineCompletionItemProvider.ts
@@ -262,6 +262,9 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
         CompletionLogger.accept(logId, completion)
     }
 
+    /**
+     * Should only be used by agent to allow it access to clear the last candidate
+     */
     public clearLastCandidate(): void {
         this.lastCandidate = undefined
     }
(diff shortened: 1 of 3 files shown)


vscode/src/completions/vscodeInlineCompletionItemProvider.ts
```diff
@@ -257,11 +257,15 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
     public handleDidAcceptCompletionItem(logId: string, completion: InlineCompletionItem): void {
         // When a completion is accepted, the lastCandidate should be cleared. This makes sure the
         // log id is never reused if the completion is accepted.
-        this.lastCandidate = undefined
+        this.clearLastCandidate()
 
         CompletionLogger.accept(logId, completion)
     }
 
+    public clearLastCandidate(): void {
```
philipp-spiess, line 265: Maybe add a note here that this is only used by agent?

philipp-spiess (approved):

olafurpg (approved):
LGTM 👍🏻 Thank you @chwarwick for following up on this 🙏🏻

PR #999, opened 2023-09-11
Autocomplete: Add section history to graph context

philipp-spiess:
This PR starts to keep track of all the code sections that a user is navigated into (i.e if they move the cursor into a section) and uses sections they have previously visited as strong references for context.

The idea here is if you look at section A (say, the `bubbleSort()` function) and then move to somewhere else (e.g. a new test file), chances are that what you looked at previously is relevant to the new section too (e.g. you might want to write a test for `bubbleSort()`).

## Test plan

- Enable graph context
- Create a file with a `bubbleSort` implementation
- Create a new file and start a `describe()` function
- Notice that we have context of the previous function now:

Screenshot 2023-09-11 at 17 45 38


efritz (approved):

vscode/src/graph/section-observer.ts
```diff
@@ -87,12 +91,73 @@ export class SectionObserver implements vscode.Disposable, GraphContextFetcher {
...
+        let usedContextChars = 0
+        const context: ContextSnippet[] = []
+
+        const sectionHistory = (
+            await Promise.all(
+                this.lastVisitedSections
+                    .map(location => this.getSectionForLocation(location))
+                    .filter(isDefined)
+                    // TODO: Use the language ID to compare here instead of this hack
+                    .filter(section => fileExtensionsMatch(section.location.uri.fsPath, document.uri.fsPath))
```
valerybugakov, line 111: Sharing our recent discussion about `fsPath` for context in case it's useful for further changes: https://github.com/sourcegraph/cody/pull/855#discussion_r1314677653

valerybugakov (approved):

vscode/src/graph/section-observer.ts
philipp-spiess: @valerybugakov Ah yeah, I thought this isn't really relevant since I only need to look at the file extension but your comment is still helpful, I did want to clean this up and store the langauge ID instead so we can rely on that instead of the file extension. 🤗

PR #1003, opened 2023-09-11
Remove remove unstable-azure provider

chwarwick:
Now that the 5.1.8 release of Sourcegraph has experimental support for `azure-openai` as an autocomplete provider, this unstable provider that connects directly to azure is no longer needed.  To use azure for autocomplete a user should choose the `unstable-openai` provider and connect to a sg instance that is configured to use azure's openai apis.
## Test plan
`pnpm --filter cody-ai run start:dev:desktop` ensure autocomplete still works.


slimsag:
Should we defer removing this? In case we run into a customer that needs this but can't upgrade quickly/easily?

chwarwick:
> Should we defer removing this? In case we run into a customer that needs this but can't upgrade quickly/easily?

I debated that, but I think with needing the keys in the extension it's pretty much unusable by customers, and it causes confusion around which one to use for azure openai.  There is always the ability to install a specific VS Code extension version.

dominiccooney (approved):
LGTM

PR #1005, opened 2023-09-11
update scrolling behavior

beyang:
Previously the behavior was:
1. When a new human message is entered, scroll all the way to the bottom.
2. As the assistant response appears, continue scrolling to keep the bottom in view (this is done with the browser's default scroll anchor behavior).

This worked fine for most chat inputs, but it was not a good UX for `/search` and `/symf`, because both these commands often generate a response that's taller than the viewport, meaning the user has to scroll back up to get to the top of results.

This PR updates the scrolling behavior in the Cody sidebar to the following:
1. When a new human message is entered, scroll that human message up toward the top of the viewport, but not past the top.
2. If the assistant response is streamed (not the case for `/search` and `/symf`) and the bottom of the transcript is in view, then continue scrolling as the response generates.

(2) should preserve the existing behavior for commands/chat that are not `/search` or `/symf`.

https://github.com/sourcegraph/cody/assets/1646931/a1cfa61b-ab71-47db-ab62-eb776f0affd4

## Test plan

Tested locally

### DIFF
--- a/lib/ui/src/chat/Transcript.tsx
+++ b/lib/ui/src/chat/Transcript.tsx
@@ -63,15 +63,17 @@ export const Transcript: React.FunctionComponent<
     ChatButtonComponent,
     isTranscriptError,
 }) {
-    // Scroll down whenever a new human message is received as input.
+    // Scroll the last human message to the top whenever a new human message is received as input.
     const transcriptContainerRef = useRef<HTMLDivElement>(null)
     const scrollAnchoredContainerRef = useRef<HTMLDivElement>(null)
+    const lastHumanMessageTopRef = useRef<HTMLDivElement>(null)
     const humanMessageCount = transcript.filter(message => message.speaker === 'human').length
     useEffect(() => {
         if (transcriptContainerRef.current) {
-            transcriptContainerRef.current?.scrollTo({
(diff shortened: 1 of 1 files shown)


lib/ui/src/chat/Transcript.tsx
```diff
@@ -114,41 +117,53 @@ export const Transcript: React.FunctionComponent<
...
-                                showFeedbackButtons={index !== 0 && !isTranscriptError}
-                                submitButtonComponent={submitButtonComponent}
-                                chatInputClassName={chatInputClassName}
-                                ChatButtonComponent={ChatButtonComponent}
-                            />
-                        )
-                )}
+                {earlierMessages.map(messageToTranscriptItem)}
+                <div ref={lastHumanMessageTopRef} />
+                {lastInteractionMessages.map(messageToTranscriptItem)}
```
dominiccooney, line 166: These nodes will have keys which overlap with `earlierMessages` ones. This is undefined behavior in React—nodes can be duplicated, omitted, behavior change between versions...

`messageToTranscriptItem` should take an offset and add it to the map index to restore the transcript item index.

dominiccooney (approved):
Looks good but we need to ensure keys are unique—feedback inline about that.

PR #1007, opened 2023-09-11
Introduce testRunID when logging events from e2e tests

akalia25:
Adds `testRunID` to the logged data so that we can identify each test run uniquely. 
## Test plan
Verify events are coming through with updated fields


dominiccooney (approved):
A bit of background about the purpose of viewing events per test run would be great for orienting. What's the use case?

Could you update the test plan with details about how to actually do it? If you weren't around and I was worried I had broken this, it would take me a while to work out what "Verify events are coming through with updated fields" means in practice.

PR #1012, opened 2023-09-12
transcript: ensure react key uniqueness

beyang:
also handle edge case where human message not found

Addresses comment: https://github.com/sourcegraph/cody/pull/1005#discussion_r1322223545

## Test plan

Tested locally


dominiccooney (approved):
Looks good!

PR #1017, opened 2023-09-12
Autocomplete: Add minimum latency

philipp-spiess:
This PR adds a new feature flag that, when enabled, will enforce a minimum latency between the last keystroke and the time a _new_ (read: not the last candidate) completion is being displayed.

The goal here is to reduce visual churn by now showing completions immediately (as they can be synthesised below that latency) while the user is still typing. Conceptually, the delay is a hint that we use to indicate that the user is indeed ready for a completion (in the same way as a debounce in any typeahead form). Contrary to our existing debounce, this is concerned about the UX and the end to end latency and not the fact as to when we trigger a network request.

Note that this is highly experimental and the code is only used for an A/B test. 

## Test plan

- Stub the feature flag locally
- Notice that while typing comments, the amount of time a random completion appears for only a brief amount of time is reduced.


dominiccooney (approved):
Very precise work, handling the time for fetching the feature flag, etc. Lovely!

valerybugakov (approved):
Nice!

PR #1018, opened 2023-09-12
Autocomplete: Add stop sequence to Fireworks

philipp-spiess:
This PR adds a stop sequence to Fireworks requests. We can rely on `\n` for single line completions because the prefix is not truncated. We also use `\n\n` in Anthropic to terminate multi-line requests so this just brings it on par.

## Test plan

- Enable StarCoder and be connected to dotcom
	- `cody.autocomplete.advanced.provider": "unstable-fireworks"`
- Observe that single line completions still work

### DIFF
--- a/vscode/src/completions/providers/unstable-fireworks.ts
+++ b/vscode/src/completions/providers/unstable-fireworks.ts
@@ -104,6 +104,7 @@ export class UnstableFireworksProvider extends Provider {
             topP: 0.95,
             topK: 0,
             model: MODEL_MAP[this.model],
+            stopSequences: [this.options.multiline ? '\n\n' : '\n'],
         }
 
         tracer?.params(args)


dominiccooney (approved):
Just curious, but for Windows developers editing files with `\r\n` line endings, does the LLM autocomplete with `\r\n`? That might make multiline completions "run on" over multiple line breaks there.

Re: test plan, enabling the provider is a very brief instruction. Can you be a bit more explicit there to bring everyone up to a competent level going through the test plans?

dominiccooney:
@philipp-spiess 

> We should probably do this more holistically, since there are a lot of places where we split by a single \n right now.

Agree that we should be methodical about line endings. Splitting by a single `\n` conveniently will kinda work for single lines, though, because the Windows-style line endings `\r\n` end in `\n`. But here specifically you're looking for successive `\n\n` and that won't work.

philipp-spiess:
@dominiccooney Ohh, I see, yep, I will add `\n\r\n` as a stop sequence too then. Thanks!

PR #1022, opened 2023-09-12
Intelligent Cody: Track which local variable types/implementations alter

efritz:
**Idea:** We have context snippets that contain type definition and implementation hover text, but that doesn't help link it to a local variable. We'll want to provide this data to the LLM in the form of _type `T` is the type of variable `X`_ or _type `T2` implements type `T1`_ so that the LLM can reason about the shape of `X` and `T1` (not just `T` and `T2`).

## Test plan

Manual testing.

image

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -124,12 +124,32 @@ export class AnthropicProvider extends Provider {
         let remainingChars = this.promptChars - this.emptyPromptLength()
 
         for (const snippet of snippets) {
+            const formatRelationship = (
+                relationship:
+                    | {
+                          symbol: string
+                          relationship: 'typeDefinition' | 'implementation'
+                      }
+                    | undefined
+            ): string => {
+                if (relationship) {
+                    switch (relationship.relationship) {
+                        case 'typeDefinition':
+                            return ` (the type of \`${relationship.symbol}\`)`
+                        case 'implementation':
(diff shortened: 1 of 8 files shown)


dominiccooney (approved):
Splendid.

vscode/src/completions/providers/anthropic.ts
```diff
@@ -124,12 +124,34 @@ export class AnthropicProvider extends Provider {
...
 
         for (const snippet of snippets) {
+            const formatRelationship = (
+                relationship:
+                    | {
+                          symbol: string
+                          relationship: 'typeDefinition' | 'implementation'
+                      }
+                    | undefined
+            ): string => {
```
valerybugakov, line 134: Can we reuse the type we already have?

```suggestion
            const formatRelationship = (
                relationship: SymbolContextSnippet['sourceSymbolAndRelationship']): string => {
```

valerybugakov (approved):

philipp-spiess (commented):
This also needs an implementation for the fireworks provider here: https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/providers/unstable-fireworks.ts#L61-L68 (50% of our community users are on the new provider right now)

Maybe we can extract this somehow?

philipp-spiess (commented):
Love this change but let's make sure this also works for StarCoder/Code Llama on Fireworks.

philipp-spiess (approved):
YES YES YES please

PR #1028, opened 2023-09-13
Autocomplete: do not report auth errors to Sentry

valerybugakov:
## Context

- Do not report auth errors ([467k reports in a couple of weeks](https://sourcegraph.sentry.io/issues/4448742691/?query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=0)) from completion generation to Sentry to reduce the volume of reported events.
- [Slack discussion](https://sourcegraph.slack.com/archives/C05AGQYD528/p1694556997905489)
- TODO follow-up https://github.com/sourcegraph/cody/issues/1027

## Test plan

1. Sign in to sourcegraph.com.
2. Drop the token via the DotCom interface.
3. Generate a completion.
4. Ensure that the `not authenticated` error is not reported to Sentry.

### DIFF
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -36,3 +36,17 @@ export function isAbortError(error: Error): boolean {
         error.message.includes('The user aborted a request')
     )
 }
+
+export class AuthError extends Error {
+    constructor(
+        message: string,
+        public traceId: string | undefined
+    ) {
+        super(message)
+        Object.setPrototypeOf(this, AuthError.prototype)
+    }
+}
+
+export function isAuthError(error: Error): error is AuthError {
+    return error instanceof AuthError
+}
--- a/vscode/src/completions/logger.ts
+++ b/vscode/src/completions/logger.ts
@@ -1,7 +1,12 @@
 import { LRUCache } from 'lru-cache'
 import * as vscode from 'vscode'
 
-import { isAbortError, isNetworkError, isRateLimitError } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
+import {
+    isAbortError,
+    isAuthError,
+    isNetworkError,
+    isRateLimitError,
+} from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
 import { TelemetryEventProperties } from '@sourcegraph/cody-shared/src/telemetry'
 
 import { logEvent } from '../services/EventLogger'
@@ -290,7 +295,7 @@ function lineAndCharCount({ insertText }: InlineCompletionItem): { lineCount: nu
 const TEN_MINUTES = 1000 * 60 * 10
 const errorCounts: Map<string, number> = new Map()
 export function logError(error: Error): void {
-    if (isAbortError(error) || isRateLimitError(error)) {
+    if (isAbortError(error) || isRateLimitError(error) || isAuthError(error)) {
         return
     }
 
--- a/vscode/src/completions/client.ts
+++ b/vscode/src/completions/client.ts
@@ -7,7 +7,12 @@ import type {
     CompletionParameters,
     CompletionResponse,
 } from '@sourcegraph/cody-shared/src/sourcegraph-api/completions/types'
-import { isAbortError, NetworkError, RateLimitError } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
+import {
+    AuthError,
+    isAbortError,
+    NetworkError,
+    RateLimitError,
+} from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
 
 import { fetch } from '../fetch'
 
@@ -87,6 +92,12 @@ export function createClient(
 
             if (response.body === null) {
                 throw new NetworkError('No response body', traceId)
+            }
+
+            // Terminate early and do not report this error to Sentry if a request is unauthenticated.
+            if (response.status === 401) {
+                const result = await response.text()
+                throw new AuthError(result, traceId)
             }
 
             // For backward compatibility, we have to check if the response is an SSE stream or a


lib/shared/src/sourcegraph-api/errors.ts
```diff
@@ -36,3 +36,17 @@ export function isAbortError(error: Error): boolean {
...
     )
 }
+
+export class AuthError extends Error {
+    constructor(
+        message: string,
+        public traceId: string | undefined
+    ) {
+        super(message)
+        Object.setPrototypeOf(this, AuthError.prototype)
```
dominiccooney, line 46: Unless `super` is doing something weird, ES constructor semantics do this.

dominiccooney (approved):
Looks good. I don't think we need to set the prototype explicitly, ES constructors do that.

lib/shared/src/sourcegraph-api/errors.ts
valerybugakov: AFAIU, it's related to the Typescript behavior. See [this wiki section](https://github.com/Microsoft/TypeScript-wiki/blob/main/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work:~:text=Extending%20built%2Dins%20like%20Error%2C%20Array%2C%20and%20Map%20may%20no%20longer%20work) for more information.

dominiccooney: Can you be more specific? Look at this: https://www.typescriptlang.org/play?#code/MYGwhgzhAEBiD29oFMAeAXZA7AJjAogE6HyHQDeAUNDdMPFhOoQK7DqkAUAlBdbQIgsADskKcA5AngTu-ATXqN4IZADoQ8AOacA8gCMAVsnZqtydAAUSHdAE9RugGad0ACwCWEXgF4-cRDVhG3h7UTkBAF9KaMpQSBgAITAyNExcGGk+ASUmVnYuXioFWlyVdU0dCWBCZDB0DywtaDBoZMJZeQUhUXEIkroGCHKNbT0jE3QzC2tQ0IdkZ1dPb2g-HzaUoJCw5H6aaNisZAB3APgeSmOz9p4gA

```
class Foo extends Error {
    constructor() {
        super('Foo')
        console.log(Object.getPrototypeOf(this) === Foo.prototype)
    }
}

class Bar extends Foo {
    constructor() {
        console.log('creating a Bar')
        super()
        console.log(Object.getPrototypeOf(this) === Bar.prototype)
    }
}

new Foo()
new Bar()
```

Prints: true, creating a bar, false (because Bar instance's proto is set to Bar.prototype), true.

You only need to do this prototype setting if you do it anywhere. So don't do it anywhere! Unless you're doing old school type swizzling, which is bad for modern JavaScript VM optimizations.

valerybugakov: Yes, I did a reading on this, and this recommendation stands for older environments where Typescript transpiled code to ES5, and the above example wouldn't produce the expected results there. 

> As part of substituting the value of this with the value returned by a super(...) call, subclassing Error, Array, and others may no longer work as expected. This is due to the fact that constructor functions for Error, Array, and the like use ECMAScript 6's new.target to adjust the prototype chain; however, there is no way to ensure a value for new.target when invoking a constructor in ECMAScript 5. 

That's why they have this recommendation in the wiki (for TypeScript 2.1), but we can consider it outdated for our setup because we can fully leverage ES6.

valerybugakov: Cc @philipp-spiess if you had other reasons in mind for explicitly adjusting the prototype

valerybugakov: I'll address this comment in [the follow-up](https://github.com/sourcegraph/cody/pull/1029).

dominiccooney: Ah, `new.target`. Thanks for that background. We're using TypeScript 5.1.6 and compiling to ES6+ so we shouldn't be resetting prototypes. It's not necessary, and if we do it *sometimes* then if someone subclasses these classes and *doesn't* reset the prototype then the results are confusing.

philipp-spiess: @dominiccooney @valerybugakov Oh I wasn't sure if the web build (who is sharing some code with this repo) was also compiling to ES6+ already but it seems like it does. In that case: agree

Screenshot 2023-09-13 at 10 23 17

vscode/src/completions/logger.ts
```diff
@@ -290,7 +295,7 @@ function lineAndCharCount({ insertText }: InlineCompletionItem): { lineCount: nu
 const TEN_MINUTES = 1000 * 60 * 10
 const errorCounts: Map<string, number> = new Map()
 export function logError(error: Error): void {
-    if (isAbortError(error) || isRateLimitError(error)) {
+    if (isAbortError(error) || isRateLimitError(error) || isAuthError(error)) {
```
philipp-spiess, line 298: Ignoring this error here means it also won't count for the sad user metric anymore (since that is derived from the analytics logging of errors) which mean we consider this to be a "good UX". I don't think it's true. If something is causing excessive auth errors for users to pop up, it's something we should fix, just like #[970](https://github.com/sourcegraph/cody/pull/970). I am not sure I agree that simply ignoring this is a good fix?

philipp-spiess: Mhh maybe until we have fixed the access token becoming invalid, this is a reasonable change. I just wonder if we should add some comment or other reminder that we clean this up again once we fix it so that an invalid auth token change automatically logs a user out of the client

valerybugakov: @philipp-spiess, I created a follow-up issue that should fix the underlying problem. WDYT? We shouldn't have an excessive number of `not authenticated` requests if we log out users after the first one. 

- https://github.com/sourcegraph/cody/issues/1027

philipp-spiess: Sounds great!

PR #1029, opened 2023-09-13
Autocomplete: do not report some network errors to Sentry

valerybugakov:
## Context

- This PR makes our error reporting rules more strict on top of https://github.com/sourcegraph/cody/pull/1028 by omitting network errors with a network status >= 500. The assumption is that these errors are already monitored on the backend.
- Filtering logic is moved right into the Sentry's `beforeSend` hook to ensure we do not accidentally report these types of errors from another `captureException` call site.
- A separate `NetworkError` class is added in addition to the `TracedError` to distinguish between transport-level and application logic errors. 
- Disabled Sentry error reporting for non-production environments + introduced an environment variable to enable it for local debugging.
- Addresses the third most reported Sentry error `read ECONNRESET` ([292k events](https://sourcegraph.sentry.io/issues/4448881794/?query=is%3Aunresolved&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=2)). It should help us to keep the volume of reported errors under control.

## Test plan

Tested locally.

### DIFF
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -21,7 +21,7 @@ export class TracedError extends Error {
         public traceId: string | undefined
     ) {
         super(message)
-        Object.setPrototypeOf(this, NetworkError.prototype)
+        Object.setPrototypeOf(this, TracedError.prototype)
     }
 }
 
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -36,3 +36,17 @@ export function isAbortError(error: Error): boolean {
         error.message.includes('The user aborted a request')
     )
 }
+
+export class AuthError extends Error {
+    constructor(
+        message: string,
+        public traceId: string | undefined
+    ) {
+        super(message)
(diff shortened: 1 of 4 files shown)


lib/shared/src/sourcegraph-api/errors.ts
```diff
@@ -23,30 +25,38 @@ export class NetworkError extends Error {
...
     constructor(
-        message: string,
+        response: Response,
         public traceId: string | undefined
     ) {
-        super(message)
-        Object.setPrototypeOf(this, AuthError.prototype)
+        super(`Request to ${response.url} failed with ${response.status}: ${response.statusText}`)
+        this.status = response.status
+        Object.setPrototypeOf(this, NetworkError.prototype)
```
dominiccooney, line 41: Per previous review, you don't need this. ES constructors set object prototypes to the constructor's `.prototype`.

dominiccooney (approved):
Looks good, but again we don't need to touch __proto__, the constructor already does that.

lib/shared/src/sourcegraph-api/errors.ts
valerybugakov: Discussed this topic [here](https://github.com/sourcegraph/cody/pull/1028#discussion_r1323888485).

Issue #931, opened 2023-09-05
L: Extensions must configure autocomplete provider client-side even if proxied through Sourcegraph instance


chwarwick:
Currently the autocomplete provider in the extension must be manually selected if a user is not using the default Anthropic provider.

Sourcegraph instances can be be configured to use the following providers for autocomplete:

- Sourcegraph gateway 
- Anthropic (using own keys)
- Azure OpenAI (using own keys)

When an extensions connects to a sourcegraph instance the correct autocomplete provider should be defaulted based on the how the instance is configured.  For example if the sourcegraph instance is using the Azure OpenAI provider the extension should default to using the `unstable-openai`

chwarwick:
Some potential solutions:

- Extending the current feature flag based solution to work for instances that are not sourcegraph.com

_I think this is a reasonable approach if we end up in a situation where we need a quick solution but isn't ideal_


- Extending the current logic that gets the [LLMConfiguration](https://sourcegraph.com/github.com/sourcegraph/cody@9f441116ca2ca8f599c5b567ec417454f5ed01c9/-/blob/lib/shared/src/sourcegraph-api/graphql/client.ts?L264:18&popover=pinned)  to include enough information to properly select an autocomplete provider.  To me this feels like a better long term solution.

_To properly select the extension autocomplete provider the site completion provider and model are potentially needed.  When using the sourcegraph provider the model will be prefixed with the type openai, anthropic etc... If using Azure OpenAI the provider is the relevant piece of information as the model name is user defined and therefore isn't useful._

taras-yemets:
- https://github.com/sourcegraph/sourcegraph/pull/56568
- https://github.com/sourcegraph/cody/pull/1035

PR #1035, opened 2023-09-13
configure autocomplete provider based on cody LLM settings in site config

taras-yemets:
Part of https://github.com/sourcegraph/cody/issues/931
Test in pair with https://github.com/sourcegraph/sourcegraph/pull/56568

Defines the autocomplete provider config based on the completions provider and model names from the site config.
The suggested configuration hierarchy:
1. If `cody.autocomplete.advanced.provider` field in VSCode settings is set to a supported provider name, and all the additional conditions like model, access token, etc. are met the corresponding provider config is returned. Otherwise, return `null` (completions provider is not created).
2. If the provider name and model can be defined based on the evaluated feature flags, return the corresponding provider.
3. If the completions provider name is defined in the connected Sourcegraph instance site config, we return the corresponding provider config. If the provider name/model can't be parsed or this provider is not supported `null` is returned (completions provider is not created). 
4. Anthropic config provider is used by default.

TODO: 
- [x] Add `createProviderConfig` tests

## Test plan
- CI passes


lib/shared/src/configuration.ts
```diff
@@ -43,3 +45,10 @@ export interface ConfigurationWithAccessToken extends Configuration {
     /** The access token, which is stored in the secret storage (not configuration). */
     accessToken: string | null
 }
+
+const colors = ['red', 'green', 'yellow'] as const
+interface Config {
+    color: (typeof colors)[number] | null
+}
+
+const color: Config['color'] = 'blue'
```
philipp-spiess, line 54: 🤔

taras-yemets: I committed part of my TS playground 🤣

vscode/src/completions/providers/createProvider.ts
```diff
@@ -101,5 +162,45 @@ async function resolveDefaultProvider(
         return { provider: 'anthropic', model: 'claude-instant-infill' }
     }
 
-    return { provider: 'anthropic' }
+    return null
+}
+
+const delimeters: Record<string, string> = {
+    sourcegraph: '/',
+    'aws-bedrock': '.',
```
taras-yemets, line 170: `aws-bedrock` is still in development
- https://github.com/sourcegraph/sourcegraph/pull/56321
 ([sourcegraph/pull/56321]
- @chwarwick [comment](https://github.com/sourcegraph/sourcegraph/pull/56321#discussion_r1322941577)

This line should probably be removed if the current PR is going to be merged sooner than the referenced one.

vscode/src/completions/providers/unstable-openai.ts
```diff
@@ -153,14 +153,17 @@ export class UnstableOpenAIProvider extends Provider {
...
+}: UnstableOpenAIOptions & { model?: string }): ProviderConfig {
     return {
         create(options: ProviderOptions) {
             return new UnstableOpenAIProvider(options, { ...unstableAzureOpenAIOptions })
         },
         maximumContextCharacters: tokensToChars(unstableAzureOpenAIOptions.contextWindowTokens),
         enableExtendedMultilineTriggers: false,
         identifier: PROVIDER_IDENTIFIER,
-        model: 'gpt-35-turbo',
+        model: model || 'gpt-35-turbo',
```
taras-yemets, line 167: The `ProviderConfig['model']` field seems to be used [for logs only](https://sourcegraph.com/search?q=context:global+r:%5Egithub%5C.com/sourcegraph/cody$+providerConfig.model&patternType=standard&sm=1&groupBy=path) (though I'm not 100% sure).
 
So, we're passing the model name here for the correct logs only. 
We do not pass it as a completion parameter (e.g., [fireworks](https://github.com/sourcegraph/cody/blob/3ec19e57fb0c3d4075fcfafbffb03a8dd58868e7/vscode/src/completions/providers/unstable-fireworks.ts#L118)) as the user-defined model is not supported by the OpenAI completions provider.

cc: @philipp-spiess

taras-yemets:
I wonder if we should fall back to the completions provider from site config if provider/model/token is misconfigured in VSCode settings. The current approach for this case is not to create any completions provider. It may be more user-friendly. WDYT, @chwarwick?

philipp-spiess:
> if provider/model/token is misconfigured in VSCode settings. The current approach for this case is not to create any completions provider.

yeah this was done on purpose so if you really manually set something, we can show you what's wrong and not "silently ignore it and leave you with no debug info"

vscode/src/completions/providers/unstable-openai.ts
chwarwick: maybe we should pass this as some value like `user-defined` or something, the model name is going to end up being the name of the azure deployment and that is probably not a value we want to log

lib/shared/src/sourcegraph-api/graphql/queries.ts
```diff
@@ -46,6 +46,7 @@ query CurrentSiteCodyLlmConfiguration {
             fastChatModelMaxTokens
             completionModel
             completionModelMaxTokens
+            provider
```
chwarwick, line 49: I assume this doesn't cause issues if the provider field does not exist?

vscode/src/completions/providers/createProvider.test.ts
```diff
@@ -0,0 +1,258 @@
...
+                    codyLLMConfig: { provider: 'fireworks', completionModel: 'llama-code-7b' },
+                    expected: { provider: 'fireworks', model: 'llama-code-7b' },
+                },
+                {
+                    codyLLMConfig: { provider: 'fireworks' },
+                    expected: { provider: 'fireworks', model: 'starcoder-7b' },
+                },
+
+                // unknown-provider
+                {
```
chwarwick, line 226: We should add a test where the LLMConfig has a null provider, I think that would be the case for any client connecting to an enterprise version.

chwarwick (approved):
This looks good to me, but if you haven't already can you verify this works or has a proper fall back when connecting to a legacy sg version that doesn't have the provider available in the LLM config

lib/shared/src/sourcegraph-api/graphql/queries.ts
taras-yemets: Great catch! How did I miss that?! Thank you, @chwarwick! 
Addressed in https://github.com/sourcegraph/cody/pull/1035/commits/a16c760ee979585173822a77d564da13c4884d9f 
Honestly, I don't like the implementation. Do you know of a more robust way of getting a potentially non-existing field?

cc: @philipp-spiess

vscode/src/completions/providers/createProvider.test.ts
taras-yemets: > LLMConfig has a null provider

@chwarwick, I'm not sure I understand this case. Does it mean that the `"completions"` section is not defined in the site config? Or it's defined, but the `"provider"` field is missing? In these both cases "sourcegraph" is set as a provider ([code](https://github.com/sourcegraph/sourcegraph/blob/f8bda0d59230000823b8a8828c069e900ecb9351/internal/conf/computed.go#L591-L613)).

chwarwick: I meant the "provider" isn't available in graphql so it null here.

taras-yemets: https://github.com/sourcegraph/cody/pull/1035/commits/a5af6d14a4af182f41241f6360a7abd4ba265cd0

taras-yemets:
> can you verify this works or has a proper fall back when connecting to a legacy sg version that doesn't have the provider available in the LLM config

Thanks for your review, @chwarwick! 
I tested this PR with the latest changes (https://github.com/sourcegraph/cody/pull/1035/commits/a16c760ee979585173822a77d564da13c4884d9f) and can confirm it works with Sourcegraph versions without Cody LLM provider field available.

vscode/src/completions/providers/unstable-openai.ts
taras-yemets: I used the model name here for consistency with the other providers configs. Logs like _"CodyCompletionProvider:initialized: unstable-openai/PLACEHOLDER"_ doesn't seem helpful/informative to me. Moreover, if we want to use the `ProviderConfig['model']` value as an actual model anywhere else in the code, having a placeholder instead of a real value may be implicit and unexpected. WDYT?

chwarwick: Knowing that model (azure deployment) name doesn't provide us any value when the customer is using azure and I tend to favor not logging details of customers environments when not necessary, how about a blank string for the model? 

I also don't see a reason for this to hold things up so feel free to merge, if everything else is in place.

taras-yemets: https://github.com/sourcegraph/cody/pull/1035/commits/2f160ea40711048fcec18869a72e4f6299d96a41

PR #1046, opened 2023-09-13
Autocomplete: Don't cull symbols that are defined in common import paths

philipp-spiess:
This is a small regression from the last changes. We were only filtering out these common imports path when loading the files into the content map, however if the location of a symbol would end up being inside a common import, we would still include its hover tooltips. We would not have a symbol name then, though, because that would require us to read the ranges of these files which we haven't loaded.

So this PR filters out all locations in these ultra common imports 

### Before

Screenshot 2023-09-13 at 16 00 00

### After

Screenshot 2023-09-13 at 15 59 39


efritz (approved):
Ty!

PR #1047, opened 2023-09-13
Intelligent Cody: Add additional Go heuristics

efritz:
Reduce noise by not returning context/error or libexec/src (Go stdlib) entries.

## Test plan

Manually ensured go stdlib paths are not passed back to the provider.

### DIFF
--- a/vscode/src/graph/graph.ts
+++ b/vscode/src/graph/graph.ts
@@ -358,7 +358,14 @@ const commonImportPaths = new Set([
     'libexec/src/',
 ])
 
-const isCommonImport = (uri: vscode.Uri): boolean => [...commonImportPaths].some(p => uri.fsPath.includes(p))
+function isCommonImport(uri: vscode.Uri): boolean {
+    for (const importPath of commonImportPaths) {
+        if (uri.fsPath.includes(importPath)) {
+            return true
+        }
+    }
+    return false
+}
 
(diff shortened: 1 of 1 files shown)


vscode/src/graph/graph.ts
```diff
@@ -346,16 +353,12 @@ const commonImportPaths = new Set([
...
 
-function isCommonImport(uri: vscode.Uri): boolean {
-    for (const importPath of commonImportPaths) {
-        if (uri.fsPath.includes(importPath)) {
-            return true
-        }
-    }
-    return false
-}
+const isCommonImport = (uri: vscode.Uri): boolean => [...commonImportPaths].some(p => uri.fsPath.includes(p))
```
philipp-spiess, line 361: This now needs an additional shallow copy of the array for every invocation. Can we just loop over the set instead like I did it before? 😬

philipp-spiess (approved):

vscode/src/graph/graph.ts
efritz: 🤦 of course

efritz, line 361:
```suggestion
function isCommonImport(uri: vscode.Uri): boolean {
    for (const importPath of commonImportPaths) {
        if (uri.fsPath.includes(importPath)) {
            return true
        }
    }
    return false
}
```

PR #1052, opened 2023-09-14
Autocomplete: trigger multiline completions on empty block only

valerybugakov:
## Context

- A follow-up for https://github.com/sourcegraph/cody/pull/913
- Apply the same logic to `enableExtendedTriggers` (non-empty-line multi-line triggers): do not trigger multi-line completions for opening brackets if the content already exists in the block.
- [Slack discussion](https://sourcegraph.slack.com/archives/C05AGQYD528/p1694609236944899)
- Added `vitest` custom matchers for multi vs sigle-line completions checks.

```ts
// Previously generated multi-line completions.
function bubbleSort() {█
   const a = []
}
```

```ts
// We request multi-line completions only for empty blocks now.
function bubbleSort() {█
   
}
```

## Test plan

Added unit-tests.

### DIFF
--- a/vscode/src/completions/detect-multiline.ts
+++ b/vscode/src/completions/detect-multiline.ts
@@ -8,7 +8,7 @@ import {
 } from './text-processing'
 
 export function detectMultiline(
-    docContext: Omit<DocumentContext, 'multiline' | 'multilineTrigger'>,
+    docContext: Omit<DocumentContext, 'multilineTrigger'>,
     languageId: string,
     enableExtendedTriggers: boolean
 ): string | null {
@@ -32,7 +32,13 @@ export function detectMultiline(
     }
 
     const openingBracketMatch = currentLinePrefix.match(OPENING_BRACKET_REGEX)
-    if (enableExtendedTriggers && openingBracketMatch) {
+    if (
+        enableExtendedTriggers &&
+        openingBracketMatch &&
+        // Only trigger multiline suggestions when the next non-empty line is indented less
+        // than the block start line (the newly created block is empty).
+        indentation(currentLinePrefix) >= indentation(nextNonEmptyLine)
+    ) {
         return openingBracketMatch[0]
     }
 
--- a/vscode/src/completions/get-inline-completions-tests/triggers.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/triggers.test.ts
@@ -1,5 +1,5 @@
 import dedent from 'dedent'
-import { describe, expect, it, test } from 'vitest'
+import { describe, expect, it } from 'vitest'
 
 import { CompletionParameters } from '@sourcegraph/cody-shared/src/sourcegraph-api/completions/types'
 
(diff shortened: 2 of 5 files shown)


abeatrix (commented):
Hmmm interesting... I can still reproduce the issue but only on line 2 -> 3 (like @philipp-spiess mentioned)
 it doesn't seem to be an issue for the lines after that (see loom for details)🤔 
(It's kinda late here so i haven't really paid attention to the change, will look at it more closel.y tmr and let you know if i find anything!)

Loom: https://www.loom.com/share/9790bc66ac9847af920d92a1720e62a9?sid=a14685fa-79c3-4ba1-a48f-21b11cc9a309

Current:
image

Infill prompt
image

valerybugakov:
Hey @abeatrix, thanks for sharing the Loom video! 

I'm confused because you demonstrate single-line completions inside a non-empty block unrelated to changes in this PR. However, I see the issue with duplicated suffixes in your video. Could you elaborate on what problem you refer to above?

vscode/src/completions/detect-multiline.ts
```diff
@@ -32,7 +32,13 @@ export function detectMultiline(
     }
 
     const openingBracketMatch = currentLinePrefix.match(OPENING_BRACKET_REGEX)
-    if (enableExtendedTriggers && openingBracketMatch) {
+    if (
+        enableExtendedTriggers &&
+        openingBracketMatch &&
+        // Only trigger multiline suggestions when the next non-empty line is indented less
+        // than the block start line (the newly created block is empty).
+        indentation(currentLinePrefix) >= indentation(nextNonEmptyLine)
```
philipp-spiess, line 40: Should we extract this into a variable and use the same var in the check below to avoid duplication?

vscode/src/completions/get-inline-completions-tests/triggers.test.ts
```diff
@@ -82,33 +79,55 @@ describe('[getInlineCompletions] triggers', () => {
                     },
                 })
             )
-            expect(requests).toHaveLength(1)
-            expect(requests[0].stopSequences).toContain(MULTILINE_STOP_SEQUENCE)
+            expect(requests).toBeSingleLine()
```
philipp-spiess, line 82: ❤️

philipp-spiess (approved):
Awesomesauce

abeatrix (approved):
> Hey @abeatrix, thanks for sharing the Loom video!
> 
> I'm confused because you demonstrate single-line completions inside a non-empty block unrelated to changes in this PR. However, I see the issue with duplicated suffixes in your video. Could you elaborate on what problem you refer to above?

Sorry I just rebuilt and tried again this morning, and I can't reproduce the issue I posted to you anymore so maybe it was a caching issue? I can open a new issue if it comes back to me but like you mentioned, it's not what your PR is trying to cover so LGTM! Thanks for clarifying!

vscode/src/completions/detect-multiline.ts
valerybugakov: I wanted to do that, but we rely on the `currentLinePrefix` indentation for the `openingBracketMatch` and `prevNonEmptyLine` indentation for the empty line case.

PR #1058, opened 2023-09-14
Autocomplete: Use the same context window size for StarCoder

philipp-spiess:
I noticed an issue in the StarCoder setup that might impact our A/B tests: We currently define a lower context window size than we do for Claude Instant. Since StarCoder supports up to 8k token windows (a lot more than we need), I’m going to raise it to the same limits we have for Claude Instant (~2k). This should ensure that we include the same context in both branches. 

I plan on cherry picking this onto the stable branch and pushing this as a patch release so we have fairer conditions in the A/B test.

## Test plan

- Enable StarCoder by being connected to dotcom, setting the provider to Fireworks, and setting the model to StarCoder 16b
- Open a lot of typescript files and trigger a completion in a very long file
- Observer that the prompt now contains up to 8k characters


vscode/src/completions/providers/unstable-fireworks.ts
```diff
@@ -36,6 +35,29 @@ const MODEL_MAP = {
     'llama-code-13b-instruct': 'fireworks/accounts/fireworks/models/llama-v2-13b-code-instruct',
 }
 
+function getContextWindowChars(model: string): number {
```
valerybugakov, line 38: Thoughts on making it type-safe?

```suggestion
type FireworksModel = keyof typeof MODEL_MAP

function getContextWindowChars(model: FireworksModel): number {
```

vscode/src/completions/providers/unstable-fireworks.ts
```diff
@@ -36,6 +35,29 @@ const MODEL_MAP = {
...
+            // compare the results
+            return 8192 // ~ 2048 token limit
+        default:
+            return 5000
+    }
+}
+
 export class UnstableFireworksProvider extends Provider {
     private client: Pick<CodeCompletionsClient, 'complete'>
     private model: keyof typeof MODEL_MAP
```
valerybugakov, line 63:
```suggestion
    private model: FireworksModel
```

valerybugakov (approved):
Making their 🧠 bigger!

Issue #810, opened 2023-08-24
Autocomplete: Investigate potential race condition in matching prefix/suffix with completions


philipp-spiess:
> Cody for some reason occasionally suggests code with typos? I’ve never seen Copilot do this. Just an example from today: suggesting cleatSearchText or cleardSearch (it did both lol) when the input was cle. There was no other occurrence of this sort of typo in any of my open files. If I had to approximate, this happens maybe 1 in every 500 completions?

> when typing and accepting a completion, it feels like we could do a better job at matching the prefix of completion with what is typed (but I get that this depends on timing)
> 
> ex. if I type ap and get a completion for pend , if I manage to type the second p myself it would be cool if only end was added (instead of ending up with apppend)

Maybe we access text contexts in an async function run using outdated offsets. We do not yield completions in the case where they would have become outdated but they might still end up being cached somewhere. I think we need to do an audit here and make sure to only ever read content from the doc based on selections before any async await.

philipp-spiess:
I was trying to reproduce this but I couldn't. Not sure what's going on here yet. 

We never show a completion when at the end of all the async stuff, the completion request was cancelled, so I assume the only case where a race condition could actually happen is when a completion was added to the cache (most likely the last candidate cache) while it was already outdated (so during the time in pre or post processing, the document has changed somewhere). This can also happen as part of the "synthesizedFromParallelRequest" logic stuff. I will try to find out if there is any place _after_ the networking code where we read from the document (this would be the most probable candidate for this issue)

philipp-spiess:
Some proof that it's _not only_ a potential race condition: Here's a Claude Instant request that omits a necessary character in the completion: https://github.com/sourcegraph/cody/pull/1059#issuecomment-1719561052

PR #1059, opened 2023-09-14
Autocomplete: Only persist last candidate if it would be shown in the UI

philipp-spiess:
Part of #810 

As part of the audit for #810 I noticed a potential here in when we save the last candidate:

- Whenever the request manager yields a result (which might be for a completion that was already cancelled) we currently persist the returned result as the last candidate.
- For already cancelled completions, this could mean that the completion is never being shown on the screen, yet it will set the `lastCandidate` and can later be picked up as the last candidate immediatly
- This also means that if the document changed during a completion request (which would cancel the completion) we could have "borked" cases in the lastCandidate that used the `document.getText` APIs at various states of the document and thus this could be an explaination of the issues we're seeing in #810 
- Conceptually, I think this logic makes a lot of sense: If we never render ghost text, we should not cache something as the last candidate

One thing to note here is the interop with the `completeSuggestWidgetSelection` flag: When it is set, I expect this to fully work. However, when it is not set, we might be causing a few unnecessary requests. Consider this case:

https://github.com/sourcegraph/cody/assets/458591/6ac8caa6-1380-4fd2-b9ba-26d0c250a6db

Every time I navigate between the suggestions, we trigger a new autocomplete request yet nothing in the prefix/suffix changes. Since the completions aren't ever visible (they would only show up if the selection happens to overlap with the completion), we never save a last candidate and thus every subsequent navigation causes a new request.

This is currently not a problem as we gave the network request cache but we want to remove that eventually. As the completion would not ever be visible for this case, this does not make a difference for the user though.

## Test plan

- Play around with autocomplete and observe no obvious issues
- Tests still pass
- Unfortunalty we lack a consistent repro for #810


philipp-spiess:
@abeatrix OOH you found a repro, this is nice! Yeah I was aware that this is not a full fix, it was just me trying out things. With the repro it might be better, let me look into it 🏆

philipp-spiess:
@abeatrix Which provider were you using to repro this? Having trouble reproducing this with StarCoder on `main` 🤔 I wonder if this is a Claude problem?

philipp-spiess:
yeah this is not a race condition but a Claude issue. Here is a network request to proof that:

```
█ CompletionLogger:onComplete: {"type":"completion","status":"success","duration":1306}  {
  "result": {
    "completion": " getSlug(text: string): string {\n  return text\n    .toString()\n   ",
    "stopReason": ""
  },
  "params": {
    "temperature": 0.5,
    "messages": [
      {
        "speaker": "human",
        "text": "Codebase context from file path 'lib/parser/excerpt.ts': <CODE5711>export function excerpt(text: string, limit: number = 144) {\n  let result = \"\";\n\n  for (const word of text.split(\" \")) {\n    if (result.length + word.length + 1 <= limit) {\n      result += \" \" + word;\n    } else {\n      // Fix trailing comma. Might need a more generic solution at some point :D\n      if (result.endsWith(\",\")) {\n        result = result.slice(0, -1);\n      }\n      result += \"…\";\n      break;\n    }\n  }\n\n  return result;\n}\n</CODE5711>"
      },
      {
        "speaker": "assistant",
        "text": "I will refer to this code to complete your next request."
      },
      {
        "speaker": "human",
        "text": "Codebase context from file path 'pages/note/[...slug].tsx': <CODE5711>import Notes from \"../../lib/Notes\";\nimport React from \"react\";\nimport Bio from \"../../lib/Bio\";\n\nexport async function getStaticPaths() {\n  const paths = (await getNotes()).map((note) => \"/note/\" + note.id);\n  return {\n    paths,\n    fallback: false,\n  };\n}\n\nexport async function getStaticProps({\n  params,\n}: {\n  params: { slug: string[] };\n}) {\n  const notes = await getNotes();\n  const note = notes.find((note) => note.id === params.slug.join(\"/\"));\n  return {\n    props: { note: note, notes },\n    revalidate: 60 * 60,\n  };\n}\n\ninterface Props {\n  note: Note;\n  notes: Note[];\n}\nexport default function Slug({ note, notes }: Props) {\n  return (\n    <>\n      <Head>\n        <title>\n          {note.title} | {[...note.category].reverse().join(\" | \")} | Philipp\n          Spiess\n        </title>\n        <meta property=\"og:title\" content={note.title} />\n        <meta\n          property=\"og:image\"\n          content={`https://philippspiess.com/api/og?title=${encodeURIComponent(\n            note.title\n          )}&date=${encodeURIComponent(\n            note.formattedDate\n          )}&sub=${encodeURIComponent(note.category.join(\"/\"))}`}\n        />\n      </Head>\n\n      <ArticleHeader type=\"note\" containerClass={styles.container} />\n</CODE5711>"
      },
      {
        "speaker": "assistant",
        "text": "I will refer to this code to complete your next request."
      },
      {
        "speaker": "human",
        "text": "Codebase context from file path 'test.ts': <CODE5711>export function test(number: number): number {\n  return;\n}\n</CODE5711>"
      },
      {
        "speaker": "assistant",
        "text": "I will refer to this code to complete your next request."
      },
      {
        "speaker": "human",
        "text": "You are a code completion AI that writes high-quality code like a senior engineer. You are looking at lib/parser/notes.ts. You write code in between tags like this: <CODE5711>/* Code goes here */</CODE5711>."
      },
      {
        "speaker": "assistant",
        "text": "I am a code completion AI that writes high-quality code like a senior engineer."
      },
      {
        "speaker": "human",
        "text": "Complete this code: <CODE5711>import format from \"date-fns/format\";\nimport { parseMarkdown } from \"./markdown\";\n\nexport interface Note {\n  title: string;\n  id: string;\n  formattedDate: string;\n  date: string;\n  category: string[];\n  contentHtml: string;\n}\n\nconst TOKEN = process.env.GITHUB_TOKEN;\nconst GRAPHQL_URL = \"https://api.github.com/graphql\";\nconst HIDDEN_FILES = new Set([\"README.md\"]);\nconst HIDDEN_DIRS = new Set([\"Unlisted\"]);\n\n// No-op, used only for syntax highlighting in the IDE\nfunction gql(strings: TemplateStringsArray) {\n  return strings.raw.join(\"\");\n}\n\nconst headers = {\n  Authorization: `Bearer ${TOKEN}`,\n};\n\nconst CONTENTS_QUERY = gql`\n  {\n    repository(name: \"philipp-spiess\", owner: \"philipp-spiess\") {\n      ref(qualifiedName: \"main\") {\n        target {\n          ... on Commit {\n            tree {\n              entries {\n                ...MyTreeEntry\n                object {\n                  ... on Tree {\n                    entries {\n                      ...MyTreeEntry\n                      object {\n                        ... on Tree {\n                          entries {\n                            ...MyTreeEntry\n                          }\n                        }\n                      }\n                    }\n                  }\n                }\n              }\n            }\n          }\n        }\n      }\n    }\n  }\n\n  fragment MyTreeEntry on TreeEntry {\n    path\n    type\n    blob: object {\n      ... on Blob {\n        text\n      }\n    }\n  }\n`;\n\nexport async function getNotes(): Promise<Note[]> {\n  let notes: Note[] = [];\n  const rawNotes = (await fetchNotes()) as any;\n\n  for (const rawNote of rawNotes) {\n    const { data, contentHtml } = await parseMarkdown(rawNote.content);\n\n    const date = data.date instanceof Date ? data.date.toISOString() : null;\n\n    notes.push({\n      title: rawNote.path.split(\"/\").pop().replace(\".md\", \"\"),\n      id: getId(rawNote.path),\n      date,\n      formattedDate: format(new Date(date), \"LLLL d, Y\"),\n      category: rawNote.path.split(\"/\").slice(0, -1),\n      contentHtml,\n    });\n  }\n\n  // Sort posts by date\n  return notes.sort(({ date: a }, { date: b }) => {\n    if (a < b) {\n      return 1;\n    } else if (a > b) {\n      return -1;\n    } else {\n      return 0;\n    }\n  });\n}\n\ninterface RawNote {\n  path: string;\n  content: string;\n}\nasync function fetchNotes(dir: string = \"\"): Promise<RawNote[]> {\n  const res = await fetch(GRAPHQL_URL, {\n    method: \"POST\",\n    headers,\n    body: JSON.stringify({ query: CONTENTS_QUERY }),\n  }).then((r) => r.json());\n\n  return recursivelyResolveEntries(res.data.repository.ref.target.tree);\n}\n\ninterface GitHubTree {\n  entries: Array<\n    | {\n        path: string;\n        type: \"blob\";\n        blob: {\n          text: string;\n        };\n      }\n    | {\n        path: string;\n        type: \"tree\";\n        object: GitHubTree;\n      }\n  >;\n}\nfunction recursivelyResolveEntries(tree: GitHubTree): RawNote[] {\n  let result: RawNote[] = [];\n  for (let entry of tree.entries) {\n    if (entry.type == \"blob\") {\n      if (!entry.path.endsWith(\".md\") || HIDDEN_FILES.has(entry.path)) {\n        continue;\n      }\n\n      result.push({\n        path: entry.path,\n        content: entry.blob.text,\n      });\n    } else {\n      if (HIDDEN_DIRS.has(entry.path)) {\n        continue;\n      }\n\n      result = result.concat(recursivelyResolveEntries(entry.object));\n    }\n  }\n  return result;\n}\n\nfunction getId(text: string): string {\n  return text.replace(\".md\", \"\").split(\"/\").map(getSlug).join(\"/\");\n}\n\nfunction getSlug(text: string): string {\n  return text\n    .toString()\n    .normalize(\"NFD\")\n    .replace(/[\\u0300-\\u036f]/g, \"\")\n    .toLowerCase()\n    .trim()\n    .replace(/\\s+/g, \"-\")\n    .replace(/[^\\w-]+/g, \"\")\n    .replace(/--+/g, \"-\");</CODE5711>."
      },
      {
        "speaker": "assistant",
        "text": "Here is the code: <CODE5711>}\n\nfunctio"
      }
    ],
    "maxTokensToSample": 50,
    "stopSequences": [
      "\n\nHuman:",
      "</CODE5711>",
      "\n\n"
    ]
  }
}
```

You see that the prompt starts with `"Here is the code: <CODE5711>}\n\nfunctio"` but the completion completely omits the necessary `n` and returns `" getSlug(text: string): string {\n  return text\n    .toString()\n   "`

Any idea how we can fix this? I assume that because this is not a full word, it'll be converted to some tokens that aren't super common 🤔

valerybugakov:
> Any idea how we can fix this?

Thanks for sharing the loom, @abeatrix! Just reproduced this issue locally, too. I'm struggling to repro the case when I type forward, going from `func` to `function`. Need to invest more time into understanding how to trigger this typo by Claude. We can document this issue for now and return to it later. I do not bump into it often (only a couple of times during the last month). 

Fromt from the tree-sitter perspective, this code is mostly OK. Just a couple of missing `;`.

vscode/src/completions/vscodeInlineCompletionItemProvider.ts
```diff
@@ -214,25 +214,18 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
-                                  )
-                              ),
-                              lastTriggerSelectedInfoItem: context?.selectedCompletionInfo?.text,
-                              result: {
-                                  logId: result.logId,
-                                  items: result.items,
-                              },
-                          }
-                        : undefined
+            const candidate: LastInlineCompletionCandidate = {
```
valerybugakov, line 219: Readability goes UP! 💜

vscode/src/completions/vscodeInlineCompletionItemProvider.ts
```diff
@@ -254,9 +247,18 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                     abortController.signal
                 )
             ) {
+                this.lastCandidate = undefined
```
valerybugakov, line 250: What's the rationale here? Why do we want to discard the last candidate here? Let's add it as a comment.

valerybugakov (approved):

vscode/src/completions/vscodeInlineCompletionItemProvider.ts
philipp-spiess: Makes sense! Added a comment but pasting here for your convinience:

```

                // Returning null will clear any existing suggestions, thus we need to reset the
                // last candidate.
```

PR #1060, opened 2023-09-14
Autocomplete: Add alt+\ shortcut to trigger autocomplete and bypass debouncing times

### LINKED ISSUE #1019: Autocomplete: Enable the manual trigger shortcut
Screenshot 2023-09-12 at 17 27 06

philipp-spiess:
Closes #1019

This adds the manual completion shortcut to trigger completions.

This change also disables the debounce timings when the completion was manually requested (that is, either using this shortcut or when hovering over completions). This is done to slightly increase latency when we already have a strong sign that the user really wants that completion

## Test plan

- Move the cursor somewhere
- Press alt+\
- Observe a completion being loaded

https://github.com/sourcegraph/cody/assets/458591/be4f9070-8b4c-47c5-bd0c-5611519318a4

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -22,6 +22,8 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 - Enabled streaming responses for all autocomplete requests. [pull/995](https://github.com/sourcegraph/cody/pull/995)
 - Sign out immediately instead of showing the quick-pick menu. [pull/1032](https://github.com/sourcegraph/cody/pull/1032)
 - UX improvements to the custom command workflow (and new [custom command docs](https://docs.sourcegraph.com/cody/custom-commands)). [pull/992](https://github.com/sourcegraph/cody/pull/992)
+- You can now use `alt` + `\` to trigger autocomplete requests manually. [pull/1060](https://github.com/sourcegraph/cody/pull/1060)
+- Slightly reduce latency when manually triggering autocomplete requests. [pull/1060](https://github.com/sourcegraph/cody/pull/1060)
 
 ## [0.10.1]
 
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -484,6 +484,11 @@
       {
         "command": "-github.copilot.generate",
         "key": "ctrl+enter"
+      },
+      {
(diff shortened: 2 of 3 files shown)


vscode/CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 - Enabled streaming responses for all autocomplete requests. [pull/995](https://github.com/sourcegraph/cody/pull/995)
 - Sign out immediately instead of showing the quick-pick menu. [pull/1032](https://github.com/sourcegraph/cody/pull/1032)
 - UX improvements to the custom command workflow (and new [custom command docs](https://docs.sourcegraph.com/cody/custom-commands)). [pull/992](https://github.com/sourcegraph/cody/pull/992)
+- You can now use `alt` + `\` to trigger autocomplete requests manually. [pull/1060](https://github.com/sourcegraph/cody/pull/1060)
+- Slightly reduce latency when manually triggering autocomplete requests. [pull/1060](https://github.com/sourcegraph/cody/pull/1060)
```
abeatrix, line 26: Can you elaborate more on how this change reduces latency, please?

abeatrix (approved):
Love this! Needed this! 

Have a question about how this helps with speed as mentioned in the changelog (still trying to understand how everything works together ☺️) but lgtm!

vscode/CHANGELOG.md
philipp-spiess: This is related to this code change: https://github.com/sourcegraph/cody/pull/1060/files#diff-0bf82c65c27367a15630425d194c46f5a66bae138f46c63361027b00161c1720R208-R212

Previously, manually triggered completions (so either using the new shortcut or by hovering over a completed text to see all possible results) would still have waited for the small 25ms/125ms debounce time - That time is so that we don't trigger toooo many network requests while typing to save some money. However for manually invoked completions we don't need that because they are already very low in volume + they are a strong sign that you want a completion NOW.

philipp-spiess: @abeatrix let me know if this makes sense 🙂

abeatrix: It does! Thank you!

Issue #1030, opened 2023-09-13
Autocomplete: Backspace on displayed completion should remove suggestion


abeatrix:
RE https://sourcegraph.slack.com/archives/C05AGQYD528/p1694037206283699
Spin off from https://github.com/sourcegraph/cody/issues/969

> Deleting a line should force a new creation always

Pressing backspace on displayed completion should remove the suggestion and replace with a new one

Currently, there is no way to force regen unless the prefix is changed.

abeatrix:
> The normal behavior of backspace showing the old completion until the new one is loaded is pretty reasonable IMO, what do you think?

Personally I found it annoying because it won't go away when I already found it unhelpful and wanted a new one 😝 but I'm speaking for myself not for the users though.

We can go with your suggestions first and see what people think? I think as long as a new one will show up then I'm fine with having it stay

philipp-spiess:
> We can go with your suggestions first and see what people think? I think as long as a new one will show up then I'm fine with having it stay

Yeah, makes sense. To fix the fact that the old completion is visible before the new one is loaded would need a change of the  the insertion range again (which is currently always from the beginning of the current line). This was added specifically to address UX jankieness issues in this PR: https://github.com/sourcegraph/cody/pull/442 

Let's try to make sure we always generate a new completion first (i.e remove the network cache) and then we can still revisit.

We should also double-check what other completion providers do (IIRC Copilot does the exact same thing that we do)

abeatrix:
working on it!

Issue #969, opened 2023-09-07
Autocomplete: Audit cache and last candidate behavior


philipp-spiess:
> After making some minor changes to the surrounding code, I am still getting the suggestion returning from cache 

![image](https://github.com/sourcegraph/cody/assets/458591/6549c6a9-c4c2-41dc-9a84-d7adb7058420)
![image](https://github.com/sourcegraph/cody/assets/458591/f329e4fd-0250-4bbe-8275-b5cf845c5d09)

> Deleting a line should force a new creation always

We should validate these things:

- Does the last candidate properly reset if you delete a line (and thus reject the completion)
- Does the last candidate reset if you change the prefix
- Do we still need the network cache at all? What is the current value it provides?
- I noticed that hovering over a single line completion generates a variant right now

abeatrix:
umbrella 
- [ ] audit cache behavior
- [ ] audit the last candidate's behavior
- [ ] https://github.com/sourcegraph/cody/issues/1030

PR #442, opened 2023-07-28
more stable autocomplete

sqs:
Makes autocomplete not jitter as much as you type, with a cache and a few other techniques to help increase the (1) cache hit rate and (2) VS Code's ability to reuse and not invalidate existing completion results.

Why this change?

- Extract the non-VS Code-specific parts of `getInlineCompletions` so we can separately test *it* and the `InlineCompletionItemProvider` logic.
  - All test cases were ported to the new `getInlineCompletions` (and all pass without modification in substance).
  - This also helps with extracting completions to the agent.
- Make the cache and RequestManager simpler and dumber by exploiting VS Code's existing behavior (of showing w/o jitter a full-line inline completion while it is refreshed in the background).
  - Remove the existing cache entirely and replace it with a simpler cache that just memoizes calls (inside RequestManager) and does not have any other text logic.
  - Make the RequestManager not need to support `logId` rewriting or retesting caches for a document (see below).
- Separately handle the case of "typing the last suggestion" (e.g., typing as suggested by the ghost text) vs. the general case of cached network requests. This case is the only one where we want to reuse the previous `logId`, and we can do so simply instead of (if the 2 cases are conflated) requiring the cache to support `logId` rewriting.

Examples of bugs or undesirable behaviors this fixes:

- Absolute minimum jitter when deleting or typing ghost text.
- As long as you are still typing ghost text exactly as suggested, it won't replace it underneath you with a new completion.
- If you delete through ghost text past the initial trigger point, it will always suggest something new.
- When you press delete on an empty but indented line with ghost text, the ghost text remains in place (no jitter) until you type something different.

## Test plan

Use completions and watch the logIds. Try typing or deleting over ghost text.

PR #1064, opened 2023-09-14
wip: new completion on deleting unwant suggestion

abeatrix:
RE: https://github.com/sourcegraph/cody/issues/969 
CLOSE: https://github.com/sourcegraph/cody/issues/810 https://github.com/sourcegraph/cody/issues/1030

# WIP - currently in discussion and not ready for review

### feat: new completion on deleting unwant suggestion

- Add the ability to request a new suggestion when user tries to delete the prefix of an unwanted suggestion displayed from cache.

~~Add completion rejection tracking, the ability to track when completion is rejected, such as when a user presses backspace while a suggestion is displayed~~

### Example

Here is an example from the latest insider build of Cody, here we have Cody misspell `completion` in this example, where I am getting `comletion`

image

And since it's served from cache, every time i removed the prefix and retype `com` again, the unhelpful suggestion would still show up (where the change in this PR would remove this unhelpful suggestion to make sure it won't show up again):

https://github.com/sourcegraph/cody/assets/68532117/32135776-67ff-49d4-970d-4bf49d36a7ec

### Summary

- Checking if the current line prefix is shorter than the prefix when the last completion was triggered. This indicates the user deleted some text, likely the suggested completion.
- Clearing the cached last completion candidate if the current suggestion is being deleted. This avoids keeping around stale completion data after the user rejects it.
- Setting a timeout to call ~~handleDidRejectCompletionItem~~ `this.clearLastCandidate()` if the current suggestion is deleted. This will log that the completion was rejected.
- Clearing any existing timeout if a new completion is triggered, to avoid multiple rejections being logged.

It updates the completion cache and logs when a suggested completion is rejected or deleted by the user. This helps keep the cache fresh and allows tracking of how often completions are rejected.

### A workaround for potential race condition

RE: https://github.com/sourcegraph/cody/issues/810

This could also be treated as a workaround for potential race conditions from https://github.com/sourcegraph/cody/issues/810

As Philipp pointed out in the issue, LLM sometimes could produce unhelp suggestion  or suggestions with typo, and currently, there is no easy way to remove the unhelpful suggestions on display (apart from the keybinding feature added by Philipp earlier). The  changes made in this PR would help because when user see an unhelpful suggestion, they can backspace to get a new suggestion.

### Before

Currently, deleting prefix from the suggestion displayed from cache does not generate a new suggestion, so user would be stuck with an unhelpful suggestion that they don't want:

https://github.com/sourcegraph/cody/assets/68532117/eafa22f0-0ea7-4e4d-849b-ca96b51325f1

### After

In this example, I was getting a long suggestion from Cody. Backspaced two times, and I get a new suggestion that is shorter from Cody 

https://github.com/sourcegraph/cody/assets/68532117/1b7dd42c-8171-41a5-a96b-f67ffb947fcf

### DIFF
--- a/vscode/src/completions/logger.ts
+++ b/vscode/src/completions/logger.ts
@@ -48,9 +48,6 @@ export interface CompletionEvent {
     suggestionAnalyticsLoggedAt: number | null
     // The timestamp of when a completion was accepted and logged to our backend
     acceptedAt: number | null
-    // The timestamp of when a completion was rejected
-    // This happens when the user press backspaced during a suggestion is displayed
-    rejectedAt: number | null
 }
 
 const READ_TIMEOUT_MS = 750
@@ -85,7 +82,6 @@ export function create(inputParams: Omit<CompletionEvent['params'], 'multilineMo
         suggestionLoggedAt: null,
         suggestionAnalyticsLoggedAt: null,
         acceptedAt: null,
-        rejectedAt: null,
     })
 
     return id
@@ -163,9 +159,6 @@ export function accept(id: string, completion: InlineCompletionItem): void {
 
     // Some additional logging to ensure the invariant is correct. I expect these branches to never
     // hit but if they do, they might help debug analytics issues
-    if (completionEvent.rejectedAt) {
-        logCompletionEvent('unexpectedRejected')
-    }
     if (!completionEvent.loadedAt) {
         logCompletionEvent('unexpectedNotLoaded')
     }
@@ -187,37 +180,6 @@ export function accept(id: string, completion: InlineCompletionItem): void {
         otherCompletionProviderEnabled: otherCompletionProviderEnabled(),
     })
     statistics.logAccepted()
-}
-
-export function reject(id: string, completion: InlineCompletionItem): void {
-    const event = displayedCompletions.get(id)
-    if (!event) {
-        return
-    }
-
-    if (event.acceptedAt) {
-        logCompletionEvent('unexpectedAccepted')
-    }
-    if (!event.loadedAt) {
-        logCompletionEvent('unexpectedNotLoaded')
-    }
-    if (!event.startLoggedAt) {
-        logCompletionEvent('unexpectedNotStarted')
-    }
-    if (!event.suggestedAt) {
-        logCompletionEvent('unexpectedNotSuggested')
-    }
-
-    event.rejectedAt = performance.now()
-    logSuggestionEvents()
-    logCompletionEvent('rejected', {
-        ...event.params,
-        // We overwrite the existing lines and chars in the params and rely on the rejected one in
-        // case the popover is used to insert a completion different from the one that was suggested
-        ...lineAndCharCount(completion),
-        otherCompletionProviderEnabled: otherCompletionProviderEnabled(),
-    })
-    statistics.logRejected()
 }
 
 export function completionEvent(id: string): CompletionEvent | undefined {
--- a/vscode/src/completions/logger.ts
+++ b/vscode/src/completions/logger.ts
@@ -48,6 +48,9 @@ export interface CompletionEvent {
     suggestionAnalyticsLoggedAt: number | null
     // The timestamp of when a completion was accepted and logged to our backend
     acceptedAt: number | null
+    // The timestamp of when a completion was rejected
+    // This happens when the user press backspaced during a suggestion is displayed
+    rejectedAt: number | null
 }
 
 const READ_TIMEOUT_MS = 750
@@ -82,6 +85,7 @@ export function create(inputParams: Omit<CompletionEvent['params'], 'multilineMo
         suggestionLoggedAt: null,
         suggestionAnalyticsLoggedAt: null,
         acceptedAt: null,
+        rejectedAt: null,
     })
 
     return id
@@ -159,6 +163,9 @@ export function accept(id: string, completion: InlineCompletionItem): void {
 
     // Some additional logging to ensure the invariant is correct. I expect these branches to never
     // hit but if they do, they might help debug analytics issues
+    if (completionEvent.rejectedAt) {
+        logCompletionEvent('unexpectedRejected')
+    }
     if (!completionEvent.loadedAt) {
         logCompletionEvent('unexpectedNotLoaded')
     }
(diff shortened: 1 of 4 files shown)


vscode/src/completions/logger.ts
```diff
@@ -48,6 +48,9 @@ export interface CompletionEvent {
     suggestionAnalyticsLoggedAt: number | null
     // The timestamp of when a completion was accepted and logged to our backend
     acceptedAt: number | null
+    // The timestamp of when a completion was rejected
+    // This happens when the user press backspaced during a suggestion is displayed
+    rejectedAt: number | null
```
philipp-spiess, line 53: But isn't it also a rejection if the user types a character that is different from what the completion is showing? I think since we have accepted as a field, aren't all completions "rejected" that are not accepted? I’m confused as to why this new field is necessary.

abeatrix: Good point! if we are not tracking those then I will delete this, thanks!

abeatrix: removed!

philipp-spiess:
I'll be OOO tomorrow, so maybe @valerybugakov can take a look here. My only thought is that I thought [this line](https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/reuse-last-candidate.ts#L58-L59) already made it so that the last candidate is only reused if they are forward typing and when you type backspace, we already do trigger new requests, but it can be that I misunderstand something.

abeatrix:
Here is an example where right now backspacing just keeps showing me the same suggestion

before

In the updated version, i get a new suggestion on every backspace i press

after

But I just dug a little further, it seems like we don't remove rejected candidates from cache?   (so my approach in this PR is wrong if this is the case)

@valerybugakov do you know why we are not removing items from cache?  Would it cause any issue if I update my PR to remove rejected items from cache?

toolmantim:
@abeatrix we may want to only if you delete the whole line, vs any backspace? Watching that video, I can imagine someone wanting to get back to the autocomplete they just saw, but it changed on them, and feeling a little too unpredictable/finicky? 🤔

abeatrix:
> we may want to do this only if you delete the whole line, vs any backspace? Watching that video, I can imagine someone wanting to get back to the autocomplete they just saw, but it changed on them, and feeling a little too unpredictable/finicky? 

@toolmantim 👋  

It looks like the current behavior is (on latest insider build): 
- send new request and get new suggestion on backspace, but does not remove it from cache
- removing the whole line also does not remove it from cache 

example


This also causes behavior where a user has already updated the surrounding code, but because we are reusing cache, it's showing the unhelpful suggestion again:

image

I think the current logic (in the release branch) on backspacing is doing what I showed in the video above already, where new requests will be made for every backspacing, but somehow it would still reuse the cache when a line is deleted or surround context have already changed? I will investigate further before I make any changes. Any advice/help is greatly appreciated though :D 

I just reread the issue https://github.com/sourcegraph/cody/issues/1030 seems like the first thing I should work on is to make sure a new request is made on deletion

valerybugakov (commented):
> do you know why we are not removing items from cache?

Here are reasons for introducing cache on the request-manager level based on [this PR](https://github.com/sourcegraph/cody/pull/442#issue-1826256871): 

> Makes autocomplete not jitter as much as you type, with a cache and a few other techniques to help increase the (1) cache hit rate and (2) VS Code's ability to reuse and not invalidate existing completion results.
> 
> Examples of bugs or undesirable behaviors this fixes:
> - Absolute minimum jitter when deleting or typing ghost text.
> - As long as you are still typing ghost text exactly as suggested, it won't replace it underneath you with a new completion.
> - If you delete through ghost text past the initial trigger point, it will always suggest something new.
> - When you press delete on an empty but indented line with ghost text, the ghost text remains in place (no jitter) until you type something different.

It's useful to reduce jitter + keep the high cache-hit rate, I think it would be helpful to remove rejected completions from the cache. AFAIU, currently, this happens only when we go over the limit of cached suggested.

https://github.com/sourcegraph/cody/blob/f6059ef54f85600ebc667c2bb2330d400def4f7f/vscode/src/completions/request-manager.ts#L177

<hr>
<hr>

> This also causes behavior where a user has already updated the surrounding code, but because we are reusing cache, it's showing the unhelpful suggestion again

This is unexpected considering that we use `prefix` and `nextNonEmptyLine` for cache keys here:

https://github.com/sourcegraph/cody/blob/f6059ef54f85600ebc667c2bb2330d400def4f7f/vscode/src/completions/request-manager.ts#L179-L181

<hr>
<hr>

> we may want to do this only if you delete the whole line, vs any backspace?

I like the comment left by @sqs in [his PR ](https://github.com/sourcegraph/cody/pull/442#issue-1826256871)

> If you delete through ghost text past the initial trigger point, it will always suggest something new.

It would be intuitive for me if we removed cached completions if I deleted text past the initial trigger point. It covers the case @abeatrix mentioned in the PR description when you got an unhelpful completion, deleted the trigger point, typed `com` again, and got the same cached unhelpful completion. It would also preserve all the benefits of the request-manage cache PR (low jitter and high cache-hit rate). The rejection trigger (the moment when we remove the completions from the request manager cache) would be defined as deleting text past the trigger point. WDYT?

philipp-spiess:
> It would be intuitive for me if we removed cached completions if a user deleted text past the initial trigger point. It covers the case @abeatrix mentioned in the PR description when you got an unhelpful completion, deleted the trigger point, typed com again, and got the same cached unhelpful completion.

I agree that this makes the most sense. I sometimes see a completion that is halfway right and want to type a few words from it and I very often make typos, seeing it going away after i have already typed 2 words from it is a bit strange. However if I delete past the initial trigger point, this is expected (and IMO also what we have implemented in the current version)

Agree that the caching is bad. I think we should remove the network cache to be honest and rely only on the last candidate logging and the synthesization logic we have in request manager. 

So I think we can get to the desired effects by removing this code only: https://github.com/sourcegraph/cody/blob/f6059ef54f85600ebc667c2bb2330d400def4f7f/vscode/src/completions/request-manager.ts#L63-L66

toolmantim:
That approach sounds solid to me @valerybugakov

abeatrix:
thanks for the value input @philipp-spiess @toolmantim @valerybugakov ! I will start working on this based on what we discussed in this thread.

> We can also manually remove from the cache when the completion is rejected but since that is every time a completion is not accepted, that would only leave accepted completions in the cache.

This is a good point! I think we can consider a suggestion that is rejected manually by deletion as rejected, where suggestions not accepted as ignored, and only remove rejected from cache, what do you think?

philipp-spiess:
> I think we can consider a suggestion that is rejected manually by deletion as rejected, where suggestions not accepted as ignored, and only remove rejected from cache, what do you think?

That's a VERY good point actually: Maybe we should reconsider our definition for rejected completions? 

Like when you move your cursor to a different place or when you switch tabs, that is not clearly a rejection compared to when you type something that differs from the completion or when you use escape/backspace. The question is if we're able to detect all of those cases? Maybe that's something to look into for you, @abeatrix? 

I imagine this to be top notch analytics data too. 

Maybe we can start by enumerating all ways that we currently cause completions to be hidden from the UI?  What comes to my mind:

- Press escape
- Move cursor
- Tab out of the current window (causing the editor to blur)
- Forward-type a different symbol
- Remove text past the initial trigger point
- Using the completion widget drop-down and cycling through different suggestions there

Do you have other ideas? cc @valerybugakov @toolmantim @beyang

abeatrix:
Moved to https://github.com/sourcegraph/cody/pull/1105

PR #1065, opened 2023-09-14
Intelligent Cody: Extract identifiers with Treesitter

efritz:
This is a WIP we should discuss about finishing. This currently replaces the looks-like-an-identifier parsing with a treesitter node walk. This should be equivalent in functionality with fewer false positives (trying to extract definitions from strings or in block comments, for example).

Where this shines is the ability (in the future, not implemented here) to try to reduce multiple requests to different occurrences of the same variable. We should be able to do a quick tree walk to see which variables reside in the same scope and deduplicate them in a semantically intelligent way.

### DIFF
--- a/vscode/src/graph/graph.ts
+++ b/vscode/src/graph/graph.ts
@@ -1,10 +1,12 @@
 import * as vscode from 'vscode'
 import { URI } from 'vscode-uri'
+import { SyntaxNode } from 'web-tree-sitter'
 
 import { HoverContext, PreciseContext } from '@sourcegraph/cody-shared/src/codebase-context/messages'
 import { dedupeWith, isDefined } from '@sourcegraph/cody-shared/src/common'
 import { ActiveTextEditorSelectionRange, Editor } from '@sourcegraph/cody-shared/src/editor'
 
+import { getCachedParseTreeForDocument } from '../completions/tree-sitter/parse-tree-cache'
 import { logDebug } from '../log'
 
 // TODO(efritz) - move to options object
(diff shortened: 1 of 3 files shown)


philipp-spiess:
Uhh exiciting. I'll be OOO tomorrow so pinging @valerybugakov for his eyes too 👀

vscode/src/graph/graph.ts
```diff
@@ -379,21 +387,43 @@ interface Request {
...
+        if (cachedParseTree) {
+            const nodes: SyntaxNode[] = [cachedParseTree.tree.rootNode]
+            while (true) {
+                const node = nodes.shift()
+                if (!node) {
+                    break
+                }
+
+                if (range.start.line <= node.startPosition.row && node.startPosition.row <= range.end.line) {
+                    if (node.type.includes('identifier')) {
```
valerybugakov, line 411: We can use a tree-sitter query to get all identifiers from a specified range. I started using queries for multi-line completion truncation. Here is [an example](https://github.com/sourcegraph/cody/pull/1033/files#diff-75f945988f3a485220e6a605ab69692b00b0532bf092243bc5fcefa6b15cfdd2R56-R57). E.g., `(identifier) @identifier`.

Also, in some grammars it's possible to query for definitions and references. Sharing it in case it's useful. See example queries from the tree-sitter-javascript repo: 

https://github.com/tree-sitter/tree-sitter-javascript/blob/f1e5a09b8d02f8209a68249c93f0ad647b228e6e/queries/locals.scm#L12-L23

valerybugakov (commented):
Makes total sense!

PR #1068, opened 2023-09-14
Consistent event names for commands

dadlerj:
Any thoughts on this change @abeatrix ? Is there a reason why we've used `:called` in the past for the explain|doc|test|smell commands? Context: https://sourcegraph.slack.com/archives/CN4FC7XT4/p1694630386266389

Note that I left the various `:clicked` event names where the object in question was a "button"

## Test plan

No test required, just a string constant change


abeatrix (approved):
the commands that ends with called are actually executed by the custom-prompt recipe, where commands that ends with executed are more tied to the vs code API. We can just use `executed` for consistency tho, thanks!

PR #1075, opened 2023-09-15
dx: upgrade @sourcegraph/eslint-config

valerybugakov:
## Context 

- [Upgraded](https://github.com/sourcegraph/eslint-config/pull/289) `eslint-plugin-unicorn` in `@sourcegraph/eslint-config` to fix [the build issue](https://github.com/sourcegraph/cody/actions/runs/6192805499/job/16813421594) on main.
    - The root cased was reported here https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2117
- Ran `pnpm lint:js --fix` to auto-fix new lint errors.

## Test plan

CI

### DIFF
--- a/lib/shared/src/sourcegraph-api/completions/browserClient.ts
+++ b/lib/shared/src/sourcegraph-api/completions/browserClient.ts
@@ -75,7 +75,7 @@ export class SourcegraphBrowserCompletionsClient extends SourcegraphCompletionsC
 }
 
 declare const WorkerGlobalScope: never
-const isRunningInWebWorker = typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope
+const isRunningInWebWorker = WorkerGlobalScope !== undefined && self instanceof WorkerGlobalScope
 
 if (isRunningInWebWorker) {
     // HACK: @microsoft/fetch-event-source tries to call document.removeEventListener, which is not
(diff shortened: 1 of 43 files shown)


taras-yemets (approved):

lib/shared/src/sourcegraph-api/completions/browserClient.ts
```diff
@@ -75,7 +75,7 @@ export class SourcegraphBrowserCompletionsClient extends SourcegraphCompletionsC
 }
 
 declare const WorkerGlobalScope: never
-const isRunningInWebWorker = typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope
+const isRunningInWebWorker = WorkerGlobalScope !== undefined && self instanceof WorkerGlobalScope
```
olafurpg, line 78: FYI @valerybugakov this change caused the Cody Worker (aka. agent) to stop working. This logic is evaluated in during module initialization https://sourcegraph.slack.com/archives/C059N5FRYG3/p1694785510650199

valerybugakov: Yes, I'm already aware of this oversight. Reviewed the fix [here](https://github.com/sourcegraph/cody/pull/1081).

PR #1078, opened 2023-09-15
Fix VScode storybook CSS class name mangling

dominiccooney:
The VScode storybook config was not mangling kebab-case identifiers as camelCase. As a result, a bunch of rules weren't being applied in the storybook.

Before:

Screenshot 2023-09-15 at 19 41 25

After:

Screenshot 2023-09-15 at 19 55 49

## Test plan

1. `pnpm -C vscode storybook`
2. Open http://localhost:6007/?path=/story/cody-loginexperiment--login
3. The sign-in buttons should be the same width.

### DIFF
--- a/vscode/.storybook/main.ts
+++ b/vscode/.storybook/main.ts
@@ -1,4 +1,5 @@
 import type { StorybookConfig } from '@storybook/react-vite'
+import { mergeConfig } from 'vite'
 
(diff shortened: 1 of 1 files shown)


umpox (approved):
Thanks for fixing!

valerybugakov (approved):

PR #1081, opened 2023-09-15
Fix broken Cody Agent

steveyegge:
The WorkerGlobalScope symbol is not defined when running the Cody Agent. This is rolling back one of Valery's changes from https://github.com/sourcegraph/cody/pull/1075 -- it must have looked as if the symbol was clearly defined, in the VSCode environment.

## Test plan

I would welcome ideas on how to test this, because clearly it's easy to break. But for now let's get it running again?


valerybugakov (approved):
Thank you!

PR #974, opened 2023-09-07
feat: add infill mode to anthropic claude instant

abeatrix:
Close https://github.com/sourcegraph/cody/issues/811 https://github.com/sourcegraph/cody/issues/944

This PR contains changes related to adding "infill" mode to the Anthropic autocomplete provider for the Claude instant model:
- A new CodyAutocompleteClaudeInstantInfill feature flag was added to the FeatureFlagProvider.
- The vscode/package.json file was updated to include "claude-instant-infill" as a new advanced autocomplete model.
   - "cody.autocomplete.advanced.model": "claude-instant-infill"
- createProvider.ts was updated to use the "infill" mode when "claude-instant-infill" is configured instead of the default mode for Anthropic `claude-instant-1`
- AnthropicOptions now accepts a mode parameter to specify either 'default' (`claude-instant-1`) or 'infill' (`claude-instant-infill`).
- AnthropicProvider has a new useInfillPrefix property to track whether infill mode is enabled.
- When useInfillPrefix is true, getPromptPrefix will return different prompt context messages to guide the model to generate autocompletion with context awareness (infill).

## Quick Demo

See improved suggestions on the right, vs suggestions provided by the latest release build on the left:
Screenshot 2023-09-06 at 1 20 52 PM

https://github.com/sourcegraph/cody/assets/68532117/cd4d314e-dfd1-42df-b063-725bc659f6ca

## Summary

This PR adds an "infill" mode to the Anthropic autocomplete provider Claude Instant model to generate more seamless completions based on surrounding code, which fixes issues where:

- The current version of completion does not support including context after the cursor which leads to generating a lot of irrelevant suggestions that could be more helpful.
- Issue where Cody suggests code that exists in the shared context

## Before 

https://github.com/sourcegraph/cody/assets/68532117/2802fc8b-bcef-4138-bd62-b3f39007cfb1

## After

https://github.com/sourcegraph/cody/assets/68532117/205ae8c1-3628-4f8f-9131-021d20fcffe8

- context awareness
- suggest code based on the pattern used in current file
- suggest codes that align with surrounding style
- won't suggest code that are duplicates (see video below) https://github.com/sourcegraph/cody/issues/811 https://github.com/sourcegraph/cody/issues/944

## Resolve the issue where Cody suggests code that already exists

RE: https://github.com/sourcegraph/cody/issues/811 https://github.com/sourcegraph/cody/issues/944

[...]

Issue #585, opened 2023-08-07
Fixup: Use expanding ranges


umpox:
## Description

Currently fixups will use the users selected range, if it is available.

This is _ok_ but it becomes problematic in different scenarios.

Example:

Code actions -> Do we really just need to change the code that has the error? Often we need to change much more to satisfy the error.

Refactors -> I want to change the returned value of a function, if I have only selected the return value then we should expand the range scope to include the entire function.

## Implementation

I think we should investigate what the best option here is. We should look at competitors like Copilot too.

One basic idea would just be to always expand the range scope to the nearest function (if available).

arafatkatze:
At a cursory out look I saw that cody does use Prefix/Postfix text  in the comprehension of Fixup recipes. 

https://github.com/sourcegraph/cody/blob/54db16dc212702ee20fb50436c02ad8849eccab7/vscode/src/non-stop/FixupController.ts#L283-L295

And even for FixupCodeAction 

https://github.com/sourcegraph/cody/blob/54db16dc212702ee20fb50436c02ad8849eccab7/vscode/src/code-actions/fixup.ts#L22-L28

The broader question is that while the range expansion for input is good. How to extend it intelligently to the part where the fixup is actually applied. Like should it only be applied to EXACTLY what the user selected and if we wanna do more. Is it possible we might cause inconvenience for the user by going out of bounds?

dominiccooney:
This is a great idea, we should do it.

FWIW, anecdotally from when I built this:
- Fixups performs better with the before/after context. (Unsurprising... so would a human.)
- Fixups performs poorly when it would be beneficial to expand the scope. For example, you ask it to rename something but the declaration is outside the scope being edited. Typical bad behavior: repeating code outside the selection (which IIRC we now prompt to suppress); not generating output; etc.

In the first iteration I tried to let Cody push back in chat if you did that, but it was way too chatty in general so @abeatrix silenced it.

Expanding the selection to a boundary which trades off logical sense (e.g. "whole function") and amount of code sent sounds great as a first cut.

Performance can deteriorate asking it to reiterate a lot of code... we could relax the requirement to repeat all the code and use a diff to align the output we *do* get to the content of the file.

Hard mode: Edit multiple files/places. Might be interesting to let Cody pitch toil to the editor, for example, have it explain a multi-step refactoring with relatively abstract commands like "rename variable x to y" and then we do those precisely.

umpox:
[...]

Issue #223, opened 2023-06-16
Non-Stop Cody: Intelligent Selection


dominiccooney:
We should use code intelligence to pick better boundaries for letting Cody edit stuff. Today we use selection plus ~50 lines.

sqs:
Fixup does not work well when there is an empty selection.

For example, with Ctrl+Shift+V "add comments", this is what I get. Note the unhelpful addition plus the lack of newline.

![image](https://github.com/sourcegraph/cody/assets/1976/52ae103c-d34d-4010-ab38-54be20beef19)

PR #1083, opened 2023-09-16
[WIP] Expanding Fixup Ranges by doing intelligent selection

arafatkatze:
Related Issues -> https://github.com/sourcegraph/cody/issues/585 and https://github.com/sourcegraph/cody/issues/223 

## Problem Summary
Currently the selection made by users is not always fully inclusive of the entire range where changes might need to be made by Cody to satisfy the request of fixup by the user. Because we're limiting the changes to only the region that was selected by the user, that creates quite a few problems:
1. Empty Selections don't work( shown by  @sqs ) 
image

2. Variable renaming within a function would miss some parts of the function with the original variable name that wasn't selected
4. Bad selection boundaries cause issues with indentation with the changes are applied coz the diffs function doesn't work perfectly in some edge cases 
5. Refactors -> If you only selected the return value then any refactor would miss other parts of the function where changes need to be applied 
6. Imperfect selection boundaries sometimes lead to repetition of code outside the selection range and sometimes it doesn't generate outputs 
7. Hard mode Problem:  Its hard to do a mega refactor where change request is on a small selection but that edits multiple files and other places in the codebase to make it all work together in synergy. 

## My solution 
I wanted to solve this problem and explore quite a few different ways to be able to solve it.

If we were to not use LLMs then there's three options. 

1. TreeSitter: For TreeSitter, it has different parsers for different languages, and you can probably do the function start, function ending thing with TreeSitter, but I feel that it might be too inconvenient to write and handle a generic code for all these different cases. And then you would also have to import new libraries and stuff. I saw some tree sitter code in issues etc but didn't explore it enough to consider exploring this option.
2. LanguageServerProtocol: even in the LanguageServerProtocol, I'm pretty sure a similar problem is going to persist. I did not explore it deeply enough. 
3. Using some special kind of codegraph tools from SourceGraph, and based on my searches, I was not able to find any special functions that I could potentially use to be able to expand the the range of my code selection. Probably there is something, and if there is something  out there that’s trivial and relatively easier to use, feel free to let me know, but I couldn't find anything in my searches

If we were to use LLms:
The simplest solution was the expand the text selection with prefix suffix and then find all the line numbers of starting/ending of all the functions etc(We can expand this to other cases too like classes interface etc) but for now I wanted to just focus on finding a function within the +-50 lines range. That’s the algorithm I did right now

I am attempting to solve problems 1-5 shown above in the best way that I can by doing an LLM based selection of the closest approximation to a full function for any selection. Sometimes the llm results aren't perfect and in that case this automatically falls back to the original selection without any issues. This is a good first step to solve the problem of automatic range selection. 

### Before

https://github.com/sourcegraph/cody/assets/11155207/53d63ee7-8650-4c03-b935-dfcb64d38c01

### After
https://github.com/sourcegraph/cody/assets/11155207/0bfa817e-9706-49a2-8ee1-14d141c4ca73

## Before 

https://github.com/sourcegraph/cody/assets/11155207/5ab30335-6c2d-4142-849d-47234d6f4100

### After

https://github.com/sourcegraph/cody/assets/11155207/9e06a5ec-9171-476f-b8e2-45172f01280a

NOTE: Some code in this is quite hacky and I am not fully happy about it but I wanted to get something out to show something to people first and then I am happy to cleanup the whole thing to make a proper PR(with feature flags too) if people are okay with my approach.

## Potential Improvements
- I could still do some more prompt engineering to reduce the length of the prompts that I am using and still get results which are just as good as they are right now. This could reduce the token usage of the extra LLM call we are adding
- The extra LLM calls adds a layer of latency I could do some A/B Testing between different LLMs to see which one would be the fastest/cheapest. Something like @abeatrix did [here](https://github.com/sourcegraph/cody/pull/974#issuecomment-1710373050)
- The overwriting in the code structure looks like a bad practice to me and I am happy to fix that to use a much cleaner route so that this code is much more understandable. 
- I only added this to the edit case of fixup but happy to explore if there are perhaps other scenarios of auto selection where something like this can be helpful. Right now I just wanted to show a bare bones example.

Thanks to @abeatrix @jdorfman  I was able to focus on an interesting problem for me to work on. 

## Test plan
Just try this locally and try some ideas that I mentioned in the PR description.

### DIFF
--- a/lib/shared/src/chat/client.ts
+++ b/lib/shared/src/chat/client.ts
@@ -4,6 +4,7 @@ import { Editor } from '../editor'
 import { PrefilledOptions, withPreselectedOptions } from '../editor/withPreselectedOptions'
 import { SourcegraphEmbeddingsSearchClient } from '../embeddings/client'
 import { SourcegraphIntentDetectorClient } from '../intent-detector/client'
+import { SourcegraphFixupRangeExpander } from '../range-expander/client'
 import { SourcegraphBrowserCompletionsClient } from '../sourcegraph-api/completions/browserClient'
 import { CompletionsClientConfig, SourcegraphCompletionsClient } from '../sourcegraph-api/completions/client'
 import { SourcegraphGraphQLAPIClient } from '../sourcegraph-api/graphql'
@@ -90,6 +91,7 @@ export async function createClient({
         const codebaseContext = new CodebaseContext(config, config.codebase, embeddingsSearch, null, null, null)
 
         const intentDetector = new SourcegraphIntentDetectorClient(graphqlClient, completionsClient)
+        const rangeExpander = new SourcegraphFixupRangeExpander(completionsClient)
 
         const transcript = initialTranscript || new Transcript()
 
@@ -131,6 +133,7 @@ export async function createClient({
             const interaction = await recipe.getInteraction(humanChatInput, {
                 editor: options?.prefilledOptions ? withPreselectedOptions(editor, options.prefilledOptions) : editor,
                 intentDetector,
+                rangeExpander,
                 codebaseContext,
                 responseMultiplexer: new BotResponseMultiplexer(),
                 firstInteraction: transcript.isEmpty,
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -58,6 +58,8 @@ export class Fixup implements Recipe {
             return null
         }
 
+        const newRange = await this.getReformedRange(fixupTask, context)
+
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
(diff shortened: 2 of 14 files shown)


jdorfman:
@umpox @abeatrix thoughts?

arafatkatze:
@abeatrix i could add your latest changes into this to make a large range fix up selection. Happy to spin up a pr for it. Would you like that?

abeatrix:
> @abeatrix i could add your latest changes into this to make a large range fix up selection(including cases where two functions are selected etc). Happy to spin up a pr for it. Would you like that?

@arafatkatze Sorry about the wait! Yea that'd be great since it's now merged to main, so I'll be happy to review your change once it's merged with the latest update, thank you!

arafatkatze:
@abeatrix There you go -> https://github.com/sourcegraph/cody/pull/1317

Issue #1076, opened 2023-09-15
Tree-sitter: add query tests


valerybugakov:
- Part of https://github.com/sourcegraph/cody/issues/910
- Based on [the discussion in Slack](https://sourcegraph.slack.com/archives/C05P6R9CV3Q/p1694770903605079?thread_ts=1694769193.088069&cid=C05P6R9CV3Q)

Issue #1056, opened 2023-09-14
Tree-sitter: add multi-lang support to multiline truncation


valerybugakov:
- Part or https://github.com/sourcegraph/cody/issues/910

PR #1086, opened 2023-09-18
Autocomplete: add snapshot tests for tree-sitter queries

valerybugakov:
## Context

- Adds an experimental logic for snapshot testing tree-sitter queries. This is the first iteration, which is not a final version, but it already provides great value — a bug found because of the visual representation of matching code. 
- Snapshot tests use tree-sitter queries defined in `.scm` files, whereas the application code still uses queries defined in TS sources. It will be fixed in #1056 
- Based on discussion with @olafurpg ([ref](https://sourcegraph.slack.com/archives/C05P6R9CV3Q/p1694769193088069))
- Part of #1076 
- Snapshot tests are powered by [vitest.toMatchFileSnapshot()](https://vitest.dev/guide/snapshot.html#file-snapshots), which means we can iterate on the logic easily in the watch mode and update snapshots if needed by pressing `u`.

## Example

Input for snapshot tests where the `|` symbol signifies the query start position (cursor position).

```ts
// Captures the whole if_statement block

export function whatIf() {
    if (Doggo) {
    //         |
        console.log('You are right!')
    } else {
        console.log('Nope -_-')
    }
}
```

Snapshot tests output.

```ts
// Captures the whole if_statement block

  export function whatIf() {
      if (Doggo) {
//    ^^ ^^^^^^^ █
          console.log('You are right!')
//        ^^^^^^^^^^^^^^^^ ^^^ ^^^^^^^^
      } else {
//    ^ ^^^^ ^
          console.log('Nope -_-')
//        ^^^^^^^^^^^^^^^^^ ^^^^^
      }
//    ^
  }
````

## Test plan

- See docs
- `pnpm test:unit:tree-sitter-queries`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 out/
 dist/
 build/
+**/test-data/
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,6 +41,7 @@ const config = {
     'import/order': 'off',
     'id-length': 'off',
     'etc/no-deprecated': 'off', // slow
+    'arrow-body-style': 'off',
     'unicorn/switch-case-braces': 'off',
     'unicorn/prefer-event-target': 'off',
     'unicorn/prefer-top-level-await': 'off',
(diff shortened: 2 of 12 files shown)


philipp-spiess (approved):
Wow that's some gourmet work right here

PR #1087, opened 2023-09-18
Agent: add `pnpm agent:debug` command

valerybugakov:
## Context

- Enables [sourcemaps](https://esbuild.github.io/api/#sourcemap) by default.
- Adds the `pnpm agent:debug` command which runs the agent, allowing attaching an
  [inspector](https://nodejs.org/en/docs/inspector) to it.

## Test plan

1. `pnpm agent:debug`
2. Attach the node-inspector to the running process.
3. Add a breakpoint to sources.
4. Verify that breakpoints work as expected with sourcemaps.

### DIFF
--- a/agent/README.md
+++ b/agent/README.md
@@ -26,16 +26,18 @@ Currently, clients have to manually write bindings for the JSON-RPC methods.
 
 ## Useful commands
 
-- The command `pnpm run build-agent-binaries` builds standalone binaries for
+- The command `pnpm build-agent-binaries` builds standalone binaries for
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):

olafurpg (approved):
LGTM 👍🏻 Thank you! With what client are you using `pnpm agent:debug`?

valerybugakov:
> With what client are you using `pnpm agent:debug`?

I do not use it. I suggested this way of debugging the agent to @steveyegge when we discussed https://github.com/sourcegraph/cody/pull/1081 in Slack. Thought it was worth adding this command for other folks.

Issue #958, opened 2023-09-07
Spec out the new Cody App flow for new onboarding


chenkc805:
```[tasklist]
### Subtasks
- [ ] https://github.com/sourcegraph/cody/pull/1090
- [ ] https://github.com/sourcegraph/sourcegraph/pull/56716
- [ ] https://github.com/sourcegraph/cody/pull/1089
```

chenkc805:
@taylorsperry do you have the Figma for what the Cody app signup flow is? CC: @toolmantim

toolmantim:
Thanks for creating the issue!

I've just gone and done the designs for this here:
https://www.figma.com/file/INGthyKM1cYki5H02JPDpW/VS-Code-Beta-2-%3C%3E-App-Onboarding?type=design&node-id=0-1&mode=design

cc @dominiccooney @thenamankumar

taylorsperry:
This looks great!!

chenkc805:
Removing all other assignees to just @thenamankumar for clarity

Answering Naman's question here: 

1. When the user opens the App some other time later, do they see Cody Chat or the settings page or nothing?
2. What if embeddings for 1 repo are in progress, and then they open another repo in VS Code and clicked on "Install App" for that repo. How do we show the progress status of two repos now?
3. We also have an embedding limit in the App. How do we communicate that they have already reached the limit?
4. What if they were on an old feature branch checked out locally when they clicked on "Install App," but now they went back to main. The embeddings are not updated, how do we communicate that they are stale?
5. Based on the designs if we want that on first install, the App should automatically start indexing the repo opened in VS Code, we will need to save that somewhere in the filesystem, and then App will read that file and always start indexing the repos on restart/install.
6. Is there a way for users to delete embeddings if they hit the limit, do we show them the current settings page? Where is that in the flow.

I'm going to organize my response based on the type of question:

- For usability questions (1, 2): For #1, I think they should see the settings page, since the Cody App as a client is going away in the near future. @taylorsperry can you confirm if you're aligned with this decision? For #2, I don't think they should see "Install App" if they've already installed the app and in the process of embedding a repo. In this case, users shouldn't be able to embed another repo if they already have 1 repo being embedded at the moment. They have to wait until their repo is finished embedding to embed another one. 
[...]

PR #1090, opened 2023-09-18
Share a dotcom access token with app, when available

dominiccooney:
Currently always reflects configuration changes by overwriting the whole file.

Part of #958 

## Test plan

Manual test:

1. Start VScode
2. Sign in with dotcom
3. `cat ~/Library/Application Support/com.sourcegraph.cody/vscode.json`

The access token and root of the current workspace should appear in the file, like this:

```
{"dotcomAccessToken":"sgp_etcetc","repoPaths":["/Users/foo/bar"],"redirect":"vscode://sourcegraph.cody-ai/etc"}
```

### DIFF
--- a/vscode/src/services/LocalAppSetupPublisher.ts
+++ b/vscode/src/services/LocalAppSetupPublisher.ts
@@ -71,8 +71,14 @@ export class LocalAppSetupPublisher implements vscode.Disposable {
         }
         const settingsDir = settingsDirPattern.replace(/^~/, homeDir)
         const settingsDirUri = vscode.Uri.file(settingsDir)
-        const settingsDirStats = await vscode.workspace.fs.stat(settingsDirUri)
+        let settingsDirStats
+        try {
+            settingsDirStats = await vscode.workspace.fs.stat(settingsDirUri)
+        } catch {
+            // We could not stat the directory, it probably doesn't exist. Try to create it.
+        }
         if (
+            settingsDirStats &&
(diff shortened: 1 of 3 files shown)


vscode/src/services/LocalAppSetupPublisher.ts
```diff
@@ -0,0 +1,94 @@
...
+            settingsDirUri = await this.ensureAppConfigDir()
+        } catch {
+            // There's no app on this platform, we could not create the settings
+            // directory, etc.
+            return
+        }
+        const workspaceRoot = vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath
+        const settings: AppSetupJson = {
+            dotcomAccessToken: accessToken,
+            repoPaths: workspaceRoot ? [workspaceRoot] : [],
```
thenamankumar, line 54: Are we not checking if the file is already present and in that case just append the workspace root to the repoPaths, rather than replacing the array?

What if 2 VS Code instances are active?

vscode/src/services/LocalAppSetupPublisher.ts
```diff
@@ -0,0 +1,94 @@
...
+            redirect: vscode.env.uriScheme + '://sourcegraph.cody-ai/app-done',
```
thenamankumar, line 55: During my local test, I was unable to see this redirect field in vscode.json

thenamankumar (approved):

PR #1099, opened 2023-09-18
fix: use infillBlock not infillBlock.trimEnd()

abeatrix:
Follow up on https://github.com/sourcegraph/cody/pull/1063

Sorry was sick on Thursday and wasn't thinking straight 😅 

This PR fixes my oversight in https://github.com/sourcegraph/cody/pull/1063.  

The trim() logic should be added [inside the replacement block](https://github.com/sourcegraph/cody/pull/1063/files#diff-2f517c653a75e0f3b05fcb8cfc041e5125a7d347878749f828e9d8d0bb732753L109), instead of trimming the whole infillBlock

## Test plan

Test this change with the incomplete comment test case

#### Before

image

#### After. 

image

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -107,11 +107,11 @@ export class AnthropicProvider extends Provider {
         const { head, tail, overlap } = getHeadAndTail(this.options.docContext.prefix)
 
         // Infill block represents the code we want the model to complete
-        const infillBlock = `${tail.trimmed.trimEnd()}`
+        const infillBlock = `${tail.trimmed}`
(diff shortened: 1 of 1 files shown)


valerybugakov (approved):
Hey @abeatrix, did you have a chance to look into [this comment](https://github.com/sourcegraph/cody/pull/1063#discussion_r1326883085)? I'm happy with this change, but I wonder what you think about the proposed simplification.

abeatrix:
@valerybugakov Thanks for the review and for pointing me to the comments I overlooked! I've fixed the typo issue, and I'm going to work on adding the raw head proposal in a follow-up PR once i'm done with the caching issue since that has higher priority atm. Tysm!

PR #1101, opened 2023-09-18
Autocomplete: Limit parallel lang server hits and abort requests when rapidly changing sections

philipp-spiess:
This PR adds a new promise limiter which can be used to:

- Define a parallel execution limit of promises 
- Add the ability to abort promises that have not started yet and have endured no work
- Add a timeout to prevent the queue to be filled with requests that never respond

This is an attempt to reduce the pressure on the lang server when querying for code graph context. Specifically, all lang server API calls are now called via the limiter and limited to 20 concurrent request (This value was chosen at random) and will be timing out after 5 seconds.

When quickly navigating through sections, the section observer will now reset a section if it can be aborted.

## Test plan

- Enable graph context experimental flag
- Open a large file
- Click into a section, observe that we set the state to `loading`
- Quickly click into a different section. Observe that we only resolve the new section and abort the old one

https://github.com/sourcegraph/cody/assets/458591/2bf93e5c-bbce-4e54-a82d-f835e6e83e4e

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,7 +12,8 @@
       "outFiles": ["${workspaceRoot}/vscode/dist/**/*.js"],
       "env": {
         "NODE_ENV": "development",
-        "CODY_DEBUG_ENABLE": "true"
+        "CODY_DEBUG_ENABLE": "true",
+        "TSS_DEBUG": "5859"
       }
     },
     {
--- a/vscode/src/graph/graph.ts
+++ b/vscode/src/graph/graph.ts
@@ -7,8 +7,15 @@ import { ActiveTextEditorSelectionRange, Editor } from '@sourcegraph/cody-shared
 
 import { logDebug } from '../log'
 
+import { createLimiter } from './limiter'
+
 // TODO(efritz) - move to options object
 const recursionLimit = 2
+
+const limiter = createLimiter(
+    20, // concurrent requests
+    5000 // ms timeout
+)
 
 /**
  * Return the definitions of symbols that occur within the editor's active document. If there is
@@ -24,8 +31,7 @@ export const getGraphContextFromEditor = async (editor: Editor): Promise<Precise
         return []
     }
 
-    // Debuggin'
-    const label = 'precise context from editor'
+    const label = 'getGraphContextFromEditor'
     performance.mark(label)
 
     const uri = workspaceRootUri.with({ path: activeEditor.filePath })
@@ -37,7 +43,6 @@ export const getGraphContextFromEditor = async (editor: Editor): Promise<Precise
 
     const filteredContexts = contexts.filter(({ filePath }) => filePath !== uri.fsPath)
 
-    // Debuggin'
     logDebug('GraphContext:filteredSnippetsRetrieved', `Retrieved ${filteredContexts.length} filtered context snippets`)
     performance.mark(label)
     return filteredContexts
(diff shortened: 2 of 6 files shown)


vscode/src/graph/graph.ts
```diff
@@ -883,14 +883,10 @@ export const locationKeyFn = (location: vscode.Location): string =>
 /**
  * Convert a mapping from K -> Thenable<V> to a map of K -> V.
  */
-const unwrapThenableMap = async <K, V>(m: Map<K, Thenable<V>>): Promise<Map<K, V>> => {
-    // Force resolution so that the await in the loop below is unblocked.
-    await Promise.all(m.values())
-
+const unwrapThenableMap = async <K, V>(map: Map<K, Thenable<V>>): Promise<Map<K, V>> => {
     const resolved = new Map<K, V>()
-    for (const [k, v] of [...m.entries()]) {
```
efritz, line 891: lol why'd I do this 🤦

vscode/src/graph/graph.ts
```diff
@@ -7,9 +7,16 @@ import { ActiveTextEditorSelectionRange, Editor } from '@sourcegraph/cody-shared
 
 import { logDebug } from '../log'
 
+import { createLimiter } from './limiter'
+
 // TODO(efritz) - move to options object
 const recursionLimit = 2
 
+const limiter = createLimiter(
+    20, // concurrent requests
```
efritz, line 16: is this low?

efritz (approved):

vscode/src/graph/limiter.test.ts
```diff
@@ -0,0 +1,101 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { AbortError, TimeoutError } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
+
+import { createLimiter, Limiter } from './limiter'
+
+describe('limiter', () => {
+    it('should limit the execution of promises', async () => {
+        const limiter = createLimiter(2, 100)
```
valerybugakov, line 9: This is a classic interview question being answered in this PR. At last, the proof that some interview questions are not that bad.

vscode/src/graph/limiter.ts
```diff
@@ -0,0 +1,88 @@
...
+    creator: PromiseCreator<T>
+    abortSignal?: AbortSignal
+    resolve: (value: T) => void
+    reject: (reason: Error) => void
+}
+
+export type Limiter = <T>(creator: PromiseCreator<T>, abortSignal?: AbortSignal) => Promise<T>
+
+export function createLimiter(limit: number, timeout: number): Limiter {
+    const queue: Queued<any>[] = []
```
valerybugakov, line 14: Can we use `unknown` or a generic here?

vscode/src/graph/graph.ts
valerybugakov: It would be nice to add context on how this number was picked (even if it's just "tested locally on a couple of completions).

.vscode/launch.json
```diff
@@ -12,7 +12,8 @@
       "outFiles": ["${workspaceRoot}/vscode/dist/**/*.js"],
       "env": {
         "NODE_ENV": "development",
-        "CODY_DEBUG_ENABLE": "true"
+        "CODY_DEBUG_ENABLE": "true",
+        "TSS_DEBUG": "5859"
```
valerybugakov, line 16: Do we want to enable it for everybody using this command?

valerybugakov (approved):

vscode/src/graph/graph.ts
valerybugakov: >This value was chosen at random

Haha, now I read the PR description properly

philipp-spiess (commented):
i forgot to post these comments lol but i got the review already so feel free to ignore

.vscode/launch.json
philipp-spiess: Hmm I can comment it out if you think it makes thinks slower. It just starts the node debug bridge on the TS Server though :)

vscode/src/graph/graph.ts
philipp-spiess: Haha I can push it higher, sure, I just wanted something high enough so that we don't risk throughput while still having it so that not all requests for a section are run concurrently (and I was assuming an average of like 40 symbols or so). I will add some comment and double it to 40

vscode/src/graph/limiter.test.ts
philipp-spiess: @valerybugakov Wait for but soon we'll have to invert a binary tree too!

valerybugakov: 😱 no way!

PR #1105, opened 2023-09-19
feat: remove unwanted suggestion from the cache

abeatrix:
RE: https://github.com/sourcegraph/cody/issues/969
Spin off from discussion in https://github.com/sourcegraph/cody/pull/1064

This PR adds the ability to remove an unwanted suggestion where the user performs deletion on a displayed suggestion from the cache.

This is built on @valerybugakov 's idea in the other issue comment https://github.com/sourcegraph/cody/pull/1064#pullrequestreview-1628364816 that we all agreed:

> It would be intuitive for me if we removed cached completions if a user deleted text past the initial trigger point. It covers the case @abeatrix mentioned in the PR description when you got an unhelpful completion, deleted the trigger point, typed com again, and got the same cached unhelpful completion. It would also preserve all the benefits of the request-manage cache PR (low jitter and high cache-hit rate). The rejection trigger (the moment when we remove the completions from the request manager cache) would be defined as deleting text past the trigger point.

If a user deletes text past/back-to the initial trigger point while a suggestion is on display, we would consider the suggestion unwanted and remove it from the cache. This ensures:
1. The unhelpful suggestion is removed from the cache
2. Displays suggestion from a new request instead. 

Note: Even if the new suggestion could be the same as the one we removed from the cache, we still want to make sure the old unhelpful suggestion is removed from the cache to begins with. This helps in cases where unhelpful suggestions are stuck even when prefix/suffix have been modified.

### Follow up

- Identify other unwanted suggestions from current logic and remove them from the cache

### Before

In this example, you can see we got a suggestion `sortCompletions` from `sort`, and removing `sort`, then type `sort` again, we will still get `sortCompletions` as suggestion from the cache:

before

### After

Deleting past the initial trigger point `sort` will always remove the suggestion from the cache, and display result from the new request instead:

after

### DIFF
--- a/vscode/src/completions/vscodeInlineCompletionItemProvider.ts
+++ b/vscode/src/completions/vscodeInlineCompletionItemProvider.ts
@@ -20,7 +20,7 @@ import {
 } from './getInlineCompletions'
 import * as CompletionLogger from './logger'
 import { ProviderConfig } from './providers/provider'
-import { RequestManager } from './request-manager'
+import { RequestManager, RequestParams } from './request-manager'
 import { ProvideInlineCompletionItemsTracer, ProvideInlineCompletionsItemTraceData } from './tracer'
 import { InlineCompletionItem } from './types'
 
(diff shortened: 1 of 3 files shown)


abeatrix:
@philipp-spiess  @valerybugakov While we are reconsidering our definition for rejected completions as we discussed [here](https://github.com/sourcegraph/cody/pull/1064#issuecomment-1723820317), I started working on removing unwanted suggestions from the cache that should address some of the caching issues we are seeing.

Do you think this approach makes sense to you?

philipp-spiess:
Yeah I think something like this makes sense. Once we have identified the cases of rejections we might need to restructure the code a bit though.

Also: lol what is this completion is Cody a brit?

Screenshot 2023-09-19 at 13 17 01

abeatrix:
@philipp-spiess new feature: backspace to get a new suggestion from British Cody 💂‍♀️

philipp-spiess (approved):

vscode/src/completions/vscodeInlineCompletionItemProvider.ts
```diff
@@ -205,6 +205,26 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                 return null
             }
 
+            // Checks if the current line prefix length is less than or equal to the last triggered prefix length
+            // If true, that means user has backspaced/deleted characters to trigger a new completion request,
+            // meaning the previous result is unwanted/rejected.
+            // In that case, we mark the last candidate as "unwanted", remove it from cache, and clear the last candidate
```
valerybugakov, line 211: Nice! I'm eager to try it out because I often want to regenerate a completion, but it was unclear how to do it. 

It's an interesting UX question — do we want to convey this rule to users somehow, or should it be "intuitive" so no additional information is needed? What do you think, @toolmantim?

PR #1107, opened 2023-09-19
typewriter: do not throw on the same message

taras-yemets:
The first and the last messsages in the OpenAI response stream may be the same as the already received text (empty string for the first message and full response text for the last one). This caused `Typewriter` to throw `Content must be supplied incrementally` error. 
Anthropic starts with the unique (non-empty) message, but ends with the same message already stored in the typewriter. This also causes throwing the mentioned error (see logs screenshots below), but does not affect the UI as the response has been already received and displayed. That's why we probably didn't catch this error before.
This PR makes `Typewriter.update` method return early if the received message matches already received text.

Below are the logs from [[`Typewriter.update`]](https://github.com/sourcegraph/cody/blob/960fba74938ce3ff98e27dc3af04099b22192a4e/lib/shared/src/chat/typewriter.ts#L51) method (` console.log('Typewriter.update: ', { text: this.text, content, isEqual: this.text === content })` at the top of the method and same data with `'Typewriter: THROW'` label right before throwing an error).
| Azure OpenAI | Anthropic |
| -- | -- |
|Screenshot 2023-09-19 at 11 41 12|Screenshot 2023-09-19 at 11 54 38|

#### Demo
 

|| Azure OpenAI | Anthropic |
|-| -- | -- |
Config| config openai | Screenshot 2023-09-19 at 11 51 12 |
|Video|https://github.com/sourcegraph/cody/assets/25318659/fcf52520-16fc-4830-9d95-2e5a353e23d7| https://github.com/sourcegraph/cody/assets/25318659/f5943284-9214-418c-b912-9b1e60cf3545 |

## Test plan
- CI passes
- Tested manually (screenshots attached)


cbart (approved):

chwarwick:
> > We should verify if we need to get this update in tomorrow's patch release, with 5.2 on Oct 4th that may be soon enough.
> 
> I was not thinking about backporting it initially. I don't see other Cody-related backports merged and if we backport this one we'll need to verify that we don't break anything with `cody-shared` version upgrade. It may be a bit risky. What do you think, @chwarwick, @cbart?

Sounds like the next deployment that would need this will be 5.2 so I think it's ok to leave out for now.  cc @ryphil This fixes Azure OpenAI web based chat.

PR #1113, opened 2023-09-19
Autocomplete: Remove common TS hover text for graph context

philipp-spiess:
This removes common and not helpful hover text like you can see in this screenshot:

Screenshot 2023-09-19 at 13 45 16

## Test plan

The common and unhelpful hover text is no longer included in the context. Here instead we only see a previous selected section being added (the example does not have any references to code that is outside the context window) 
Screenshot 2023-09-19 at 13 55 46

### DIFF
--- a/vscode/src/graph/graph.ts
+++ b/vscode/src/graph/graph.ts
@@ -544,30 +544,36 @@ const hoverContextFromResolvedHoverText = (t: ResolvedHoverText): HoverContext[]
     ].filter(isDefined)
 
 const hoverContextFromElement = (
-    e: ResolvedHoverElement | undefined,
+    element: ResolvedHoverElement | undefined,
     type: HoverContext['type'],
     sourceSymbolName?: string
 ): HoverContext | undefined => {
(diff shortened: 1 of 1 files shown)


philipp-spiess:
Screenshot 2023-09-19 at 13 58 50

Also just noticed the absolute paths names. I thought I fixed this already 🤔

philipp-spiess:
`vscode.workspace.asRelativePath` doesn't do anything anymore for me lol. this is new 🤔

philipp-spiess:
... I'll look into this as a follow-up PR

abeatrix (approved):

philipp-spiess:
Ah okay the URL was absolute because I opened a subfolder as the workspace and:

```
		 * When there are no {@link workspace.workspaceFolders workspace folders} or when the path
		 * is not contained in them, the input is returned.
```

I wonder if relative paths to the first workspace folder would be better though since this just includes more chars and is potentially not helpful for the code completion. Anyway for now this is a different issue, the existing context is using the same API

Issue #1112, opened 2023-09-19
Autocomplete: Disable on scminput


philipp-spiess:
Disable autocomplete in this box lol

<img width="458" alt="Screenshot 2023-09-19 at 13 05 30" src="https://github.com/sourcegraph/cody/assets/458591/c96cbc01-d9de-4eb0-8158-30c65dac9a46">

philipp-spiess:
Might also want to consider:

- `html`
- `plaintex`
- `scss`

PR #1114, opened 2023-09-19
change: register inline completion provider for files and notebooks only

abeatrix:
Close https://github.com/sourcegraph/cody/issues/1112

The inline completion provider was previously registered for all languages ('*') at all places. 

This change updates the registration to explicitly include notebook and text documents with file scheme to ensure autocompletion only works in places that we currently support.

## Test plan

- Completions should work the same on saved text documents and notebook cells
- Completions should not work in source control input box or any other places that are not editor files / notebook files

### Before 

image

### After

https://github.com/sourcegraph/cody/assets/68532117/7a9ec776-398d-47ba-b3cc-6aeda86c3b24

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -29,6 +29,7 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 - Slightly reduce latency when manually triggering autocomplete requests. [pull/1060](https://github.com/sourcegraph/cody/pull/1060)
 - Configure autocomplete provider based on cody LLM settings in site config. [pull/1035](https://github.com/sourcegraph/cody/pull/1035)
 - Filters out single character autocomplete results. [pull/1109](https://github.com/sourcegraph/cody/pull/1109)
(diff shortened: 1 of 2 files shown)


valerybugakov (approved):

philipp-spiess:
Nice! But I think this regresses autocomplete on new unsaved files, did you try that?

philipp-spiess:
https://github.com/sourcegraph/cody/pull/1120

abeatrix:
@philipp-spiess I mentioned in the description that it would only work in saved text doc and notebook cells:

> Completions should work the same on saved text documents and notebook cells

I thought the quality will not be good like in scm when working in an unsaved file because we won't be able to fetch relevant context for them / no language support so I didn't include that.

If that's something we want to continue support I can add the untitled scheme to the provider!

abeatrix:
Saw you already did lol thanks!

philipp-spiess:
Ohh gotcha... Yeah I added this on purpose because a user expected it since Copilot does it. 

I guess what we can do is support it but only if a language id is specified (so we exclude `plaintext` which according to our analytics does also have a very bad acceptance rate)

philipp-spiess:
Forget that I just tested and Copilot does not do this (anymore?) so I think your intuition is correct

PR #1116, opened 2023-09-19
feat: new /doc command with smart selection

abeatrix:
feat: add smart selection and update doc prompt

This PR focuses on updating the `/doc` command, featuring `snart selection`.

- Add getActiveTextEditorSmartSelection method to editor interface for supporting "smart selection" on files that have lang server running as we rely on that to extract the ranges for each function / methods within a class
- Update prompt for the /doc command to specify docstring should not contain code but docstring only
- Introduce `CommandRunner` class that allows different services to run different commands async
- Add `mode` field to cody commands config 
- Add `insertMode` and `autoApply` fields to FixupTask that allows fixup to auto apply suggestions returned by Cody instead of displaying Code Lenses
- `/doc` command now uses fixup insert mode 

What is `smart selection`?

Smart selection removes the need to manually highlight code before running the /doc and /test commands. Instead, it will automatically detect and run the selected command for the function nearest to your cursor position. For /doc, the docstring will be added to the top of the function through fixup.

Why?

Based on user testing conducted by Megan, many first-time command users were unaware of the requirement to highlight code before running commands. We can create a smoother onboarding experience for new users with smart selection. Allowing our commands to work seamlessly without manual code selection will lower the learning curve and provide a more intuitive first-time user experience. 

However, it's important to note that smart selection should be treated as a fallback method, as it produces a selection based on guessing user intent. Manual selection truly reflects the user's real intent and should be used when possible. Moreover, smart selection may not work well for all languages, and depends on if users have language servers running in their editor. In cases where smart selection is unreliable, users should fall back to using the manual selection method to ensure accuracy.

## Follow-up

- [ ] improve /doc prompt
- [x] apply fixup automatically
- [x] indentation issues caused by fixup

## Test plan

Update 2 - latest

https://github.com/sourcegraph/cody/assets/68532117/a928612b-c792-43f5-9947-29bd52f2ea17

Update 1

https://github.com/sourcegraph/cody/assets/68532117/02953c54-efe0-41dc-acdb-45ffb15e6404

2. Run `/doc` command in a function without selecting any code (but put your cursor inside a function)
3. Run `/doc` command with code selected
4. Repeat step 2 and 3 with `/test` command
5. Run other commands, and use chat -- they should work as expected


toolmantim (commented):
This is so great @abeatrix! I love we're triggering a fixup, and just modifying the code! 

Indention is often incorrect when documenting indented code… maybe something we can improve in the future with @valerybugakov's tree-sitter tech? One thing we could try in the future is auto-applying too. But one step at a time! :)

Is it possible to adjust the prompt so that Cody doesn't describe "parameters" in places where it doesn't make sense, because I'm getting this when documenting some selected code:

Screenshot 2023-09-20 at 5 10 21 pm

I gave it a go locally with this diff:

```diff
diff --git a/lib/shared/src/chat/prompts/default-prompts.json b/lib/shared/src/chat/prompts/default-prompts.json
index f1852861..def4714b 100644
--- a/lib/shared/src/chat/prompts/default-prompts.json
+++ b/lib/shared/src/chat/prompts/default-prompts.json
@@ -2,7 +2,7 @@
   "commands": {
     "doc": {
       "description": "Generate code documentation",
-      "prompt": "Generate a comment documenting the parameters and functionality for the selected code. Only generate the documentation for the selected code, do not generate the code. Use the same documentation style in the file of the selected code to generate the comments. Pay attention to the file path of the selected code to make sure the comments are generated for the correct language. For example, use the JavaDoc documentation style to generate comments for .java files, or Python docstring using Python multi-line string for .py files. Do not include any other code or comments besides the docstring.",
+      "prompt": "Generate a comment briefly documenting the purpose of the selected code. If it takes parameters, document those. Only generate the documentation for the selected code, do not generate the code. Use the same comment style in the file of the selected code, or the most common style for that language. If no existing documentating comments exist, pay attention to the file path of the selected code to make sure the comments are generated in the style of that language. Do not include any other code or comments besides the comment.",
       "context": {
         "currentDir": true,
         "selection": true
```

And ended up with better results for the selected lines:

Screenshot 2023-09-20 at 5 20 24 pm

That appears to still document functions correctly too:

Screenshot 2023-09-20 at 5 22 04 pm

I don't have a good test suite / cases to verify it across more examples, so will defer to your prompt magic skills!

lib/shared/src/chat/prompts/default-prompts.json
```diff
@@ -2,11 +2,12 @@
...
+      "prompt": "Generate a comment documenting the parameters and functionality for the selected code. Only generate the documentation for the selected code, do not generate the code. Use the same documentation style in the file of the selected code to generate the comments. Pay attention to the file path of the selected code to make sure the comments are generated for the correct language. For example, use the JavaDoc documentation style to generate comments for .java files, or Python docstring using Python multi-line string for .py files. Do not include any other code or comments besides the docstring.",
       "context": {
         "currentDir": true,
         "selection": true
-      }
+      },
+      "mode": "insert"
```
toolmantim, line 10: Nice! Does this mean custom commands can do this too?

abeatrix: YES! Also takes "inline" as value where the prompt will be run in inline chat instead 😀

abeatrix:
The current one is the one from the old recipe so I didn't really do any prompt engineering on this, but yours looks good to me!

lib/shared/src/chat/prompts/default-prompts.json
toolmantim: Wow. This is ace!

vscode/src/editor/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @param activeCursor - The active cursor position
+ * @returns The folding range containing the active cursor, or undefined if none found
+ */
+export async function getCursorFoldingRange(
+    uri: vscode.Uri,
+    activeCursor: number
+): Promise<vscode.Selection | undefined> {
+    const symbols = await getSymbolRanges(uri).then(r => r.filter(s => s.kind === vscode.SymbolKind.Class))
+    // Get the folding ranges for the document,
+    const ranges = await getFoldingRange(uri).then(r => r.filter(r => r.kind !== 2 && r.kind !== 1 && r.kind !== 3))
```
philipp-spiess, line 47:
```suggestion
    const [symbols, ranges] = await Promise.all([
    		getSymbolRanges(uri).then(r => r.filter(s => s.kind === vscode.SymbolKind.Class)),
    		// Get the folding ranges for the document,
    		getFoldingRange(uri).then(r => r.filter(r => r.kind !== 2 && r.kind !== 1 && r.kind !== 3))
    ]
```

Can be parallelized 😎

philipp-spiess: What do the magic numbers mean (kind 2, 1, or 3)? Is there an enum we can use instead?

vscode/src/editor/utils.ts
```diff
@@ -0,0 +1,87 @@
...
+    // remove ranges that match the symbols from ranges
```
philipp-spiess, line 48: Can you explain why? It's not immediatly clear to me

arafatkatze:
@abeatrix This is truly amazing. I wish I had thought of this while doing https://github.com/sourcegraph/cody/pull/1083 I do have a minor suggestion about the logic.

Sometimes the lines of the symbols ranges and the folding ranges don't match exactly and therefore the symbol range isn't removed. 
```
            if (r.start === symbolRange.start.line && r.end === symbolRange.end.line) {
                ranges.splice(i, 1)
                i--
            }
```
This check is missed sometimes. 

IF you try inside the class in `/Users/arafatkhan/Desktop/cody/vscode/src/completions/providers/unstable-fireworks.ts` and `/Users/arafatkhan/Desktop/cody/vscode/src/completions/providers/unstable-openai.ts` you will see that the folding range is missed so the functions inside the class take up the whole class instead of the function alone. Some function to do "close enough" match instead of exact match would help a lot here.

vscode/src/editor/utils.ts
arafatkatze: IT's coz if there is a function inside a class then the whole class is taken rather than the function alone. 

```
1 | class OuterClass {
2 |     // ...
3 |     function something {
4 |         // ... cursor
5 |     }
6 |     // ...
7 | }
```

You wanna get line 3-5 instead of 1-7. But you get 1-7 unless you exclude folding ranges that match class symbol.

vscode/src/editor/utils.ts
arafatkatze: Yep the enums are available and can be used they actually make it easier to understand `vscode.SymbolKind.Module`, `vscode.SymbolKind.Namespace` and `vscode.SymbolKind.Package`

abeatrix: oooo good catch. I just replaced it with !r.kind since we just want ranges that don't belong to any of the folding range kind!

vscode/src/editor/utils.ts
abeatrix: What @arafatkatze said! I also just added a comment to the function to detail my thought process, here is what I added:

> The purpose of filtering to keep only folding ranges that contain other folding ranges is to find the outermost folding range enclosing the cursor position.

>Folding ranges can be nested - you may have a folding range for a function that contains folding ranges for inner code blocks.

>By filtering to ranges that contain other ranges, it removes the inner nested ranges and keeps only the outermost parent ranges.

>This way when it checks for the range containing the cursor, it will return the outer range that fully encloses the cursor location, rather than an inner range that may only partially cover the cursor line.

>However, if we have kept the ranges for classes, this may only return ranges for classes that contain individual methods rather than the outermost range of the methods within a class.

abeatrix:
@toolmantim Updated the PR to fix the Indention issues you were seeing + added autoApply mode to fixup 🎉 

https://github.com/sourcegraph/cody/assets/68532117/231bbd03-abe2-421e-a15c-f1ed9932986f

@arafatkatze thanks for your detailed feedback and for testing out the changes! smartSelection, unfortunately only works for files where users have lang servers running. If not,  I think we should just fallback to using the manual selection for now, but if you have any suggestions on how we can extract ranges for functions in those cases please let us know. We can also look into this as a follow up PR too 😄

toolmantim:
@abeatrix just tested this, and it works great! I tested undo/redo as well, and seems to be fine. The indentation it inserts is perfect now too.

All it needs is that prompt update and I think this is ready to rock! 🙌🏼🙌🏼🙌🏼

lib/shared/src/chat/prompts/index.ts
```diff
@@ -1,6 +1,9 @@
 import * as defaultPrompts from './default-prompts.json'
 import { toSlashCommand } from './utils'
 
+// A list of default cody commands
+export const defaultCodyCommands = ['ask', 'doc', 'edit', 'smell', 'test', 'reset']
```
arafatkatze, line 5: Nitpick: Not used anywhere

toolmantim:
Do we have a way of doing any e2e testing of this (perhaps with a few diff example languages/repos/scenarios)?

abeatrix:
> Do we have a way of doing any e2e testing of this (perhaps with a few diff example languages/repos/scenarios)?

Hmmm it's using fixup which is already covered by an e2e test.

Is there something you have in mind that you want to make sure it's covered by a test particularly?

vscode/src/editor/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+export function getOutermostRangesInsideClasses(
+    classRanges: vscode.Range[],
+    foldingRanges: vscode.FoldingRange[],
+    activeCursor: number
+): vscode.FoldingRange | undefined {
+    // Remove all ranges that are contained within class ranges
+    for (const cRange of classRanges) {
+        for (let i = 0; i < foldingRanges.length; i++) {
+            const r = foldingRanges[i]
+            if (r.start === cRange.start.line && r.end === cRange.end.line) {
```
arafatkatze, line 83: This could be 
`
if (Math.abs(r.start - cRange.start.line) <= 1 && Math.abs(r.end - cRange.end.line) <= 1) {
`

## Before(Turn on voice)

https://github.com/sourcegraph/cody/assets/11155207/9a72bd8b-7b3d-4f03-bea7-39dff619cf34



## After (Turn on voice)

https://github.com/sourcegraph/cody/assets/11155207/1561aa96-1a69-4434-bb68-b3ede22020bc

abeatrix: Yes nice catch!! If you can suggest a diff with your change, I can merge it into this branch ❤️

toolmantim:
Sounds like we don’t have this yet, but was just checking if there was a more automated way to run this command with diff scenarios of repos, files and selection… and verify the quality of output and behavior. My testing is still fairly adhoc.

vscode/src/editor/utils.ts
arafatkatze, line 83:
```suggestion
            if (Math.abs(r.start - cRange.start.line) <= 1 && Math.abs(r.end - cRange.end.line) <= 1) {
```

arafatkatze: Done

csells:
is there a way to keep the new docs commenting consistent with the rest of the doc comments in the file, e.g. // vs /* */

abeatrix:
> is there a way to keep the new docs commenting consistent with the rest of the doc comments in the file, e.g. // vs /* */

Yes, that requires some prompt engineering, which we will work on next in another PR! 😄

abeatrix:
@valerybugakov @dominiccooney can I have one of you review this PR when you have the time, please? Multiple people have tested what's in the change already (see conversations above), but will need someone from the eng side to do the final review/approval to move this forward, thank you 🙇‍♀️

vscode/src/editor/utils.ts
philipp-spiess: But can't nesting of folding ranges happen regardless? Like if you have a function inside a function:

```
function createFoo() {
  return {
    bar() { ....},
    baz() { ....},
  }
}
```

Maybe we should find nesting and prefer the leaves?

philipp-spiess: Maybe prefer the leaves but also balance against a "target LOC" length. This is something along the lines I would do for the graph context too. Because everything that's like ~50loc is a good context to visually work on, leaves that are only one line might not be super helpful but also classes that are like 5000 LOCs aren't.

Facing the very same problem right now with graph context. Our idea is to use tree sitter to solve this eventually. The problem with folding ranges is that some people set them to be indentation based and then you have a folding range for the arguments and the body of a function which makes it much less useful. 

cc @valerybugakov

philipp-spiess: (not wanting to block this PR on that btw, this is fantastic work, just trying to learn from it for selfish reasons 👀)

abeatrix: It does! The full process was to remove classes from the folding ranges first, then find the outermost range that contains the cursor position. 

#### Step 0

```
class Foo{
    function createFooOne() {
        return {
            bar() { ....},
            baz() { ....},<CURSOR HERE>
        }
    }
    function createFooTwo() {
        return {
            bar() { ....}
        }
    }
}
```

#### Step 1 - remove class

```

    function createFooOne() {
        return {
            bar() { ....},
            baz() { ....},<CURSOR HERE>
        }
    }
    function createFooTwo() {
        return {
            bar() { ....}
        }
    }

```

#### Step 2 - find ranges contains cursor, and then remove nested ranges

```

    function createFooOne() {
        
           
            
     
    }
   
        
            
     
  

```

vscode/src/editor/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * Folding ranges can be nested - you may have a folding range for a function that contains folding ranges for inner code blocks.
+ *
+ * By filtering to ranges that contain other ranges, it removes the inner nested ranges and keeps only the outermost parent ranges.
+ *
+ * This way when it checks for the range containing the cursor, it will return the outer range that fully encloses the cursor location,
+ * rather than an inner range that may only partially cover the cursor line.
+ * '
+ * However, if we keep the ranges for classes, this will then only return ranges for classes that contain individual methods rather
+ * than the outermost range of the methods within a class. So the first step is to remove class ranges.
```
philipp-spiess, line 78: I think this is not specific to classes btw. In some languages, it might be giant objects or modules. also in JS you could have a constructor function returning an object.

What about we define a target LOC count and while traversing upwards, we filter out ranges that are too large. Would that make sense?

vscode/src/editor/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (!foldingRanges?.length) {
+        return undefined
+    }
+
+    // Remove all ranges that are contained within class ranges
+    if (classRanges.length) {
+        for (const cRange of classRanges) {
+            for (let i = 0; i < foldingRanges.length; i++) {
+                const r = foldingRanges[i]
+                if (Math.abs(r.start - cRange.start.line) <= 1 && Math.abs(r.end - cRange.end.line) <= 1) {
```
philipp-spiess, line 94: Annoying that folding ranges are not normal ranges and thus have no `contains()` 🙃

philipp-spiess (approved):
Good stuff

vscode/src/editor/utils.ts
abeatrix: @philipp-spiess omg right?!!! That tricked Cody autocompletion too 🤣

vscode/src/editor/utils.ts
philipp-spiess: But what if a method in a class has 10k LOC? Or what if someone likes to use constructor functions instead of classes in JS. Might be interested to learn from the work on chunking for embeddings here too

vscode/src/editor/utils.ts
abeatrix: >What about we define a target LOC count and while traversing upwards, we filter out ranges that are too large. Would that make sense?

Yea it does that's smarter too and removes the dependency of the user having code intel installed for the language! I can work on refining this in a follow-up PR, thanks for the suggestions!

Does graph context already have something similar in place I can refer to?

philipp-spiess: No my implementation is much worse and only looks at the document sections (so for classes it's currently useless). I was hoping I can just use your implementation after that 🙈

vscode/src/editor/utils.ts
arafatkatze: @philipp-spiess @abeatrix I added an example of a slightly different edge case of the algorithm that philip was mentioning. 

Here's the [code](https://github.com/sourcegraph/cody/commit/8c36b5fd6e97e01610a13278c75e6a318e47f60d) 

Here's the [video](https://drive.google.com/file/d/1zFbdUgVlbOXgYk70iS4cWpIFbgaozs12/view?usp=sharing)(Turn on audio )


Here's the image I used 

folding_ranges_graph_cute_labeled_wide_legended

And here's how I made [the image in Gpt Chat ](https://chat.openai.com/share/5ca749fb-df1e-4f55-b5fd-dc61efe1a177)(Although image sharing and download is not supported in chat sharing but I attached the [final image](https://user-images.githubusercontent.com/11155207/270181250-734059b8-35ca-4747-aedb-70a46f0330e9.png) so you should be good, atleast the code is there)


I would be happy to explore the chunking of embeddings too and if you are happy with the current state then I can spin up a PR too.

philipp-spiess: This is awesome! Thank you for the in-depth video. This is exactly what I was talking about but now, after seeing it with the action, I wonder if we shouldn't make it so that only function folding ranges are considered for these. Especially when the action is "document code" since I would expect docstrings to be added for functions only 🤔

toolmantim: Yep, agreed for this feature we only want docstring when documenting things that explicitly take parameters (i.e. functions etc)… lines of code should just document the `why` of the code as plain comments, following the comment style from the file or similar files in the repo. That was improved with the prompt change merged with this PR, but my testing was only adhoc so I wouldn't be surprised if it needs further tweaking.

toolmantim: Oh, it looks like the prompt diff from https://github.com/sourcegraph/cody/pull/1116#pullrequestreview-1634862479 wasn't included in the end @abeatrix?

abeatrix: > Oh, it looks like the prompt diff from https://github.com/sourcegraph/cody/pull/1116#pullrequestreview-1634862479 wasn't included in the end @abeatrix?

I didn't make any changes to the prompt in that PR since that was for the /doc behavior change.

Do you want to create a new PR with your change for the prompt?

abeatrix: @toolmantim Phillips was referring to the changes suggested by @arafatkatze i believe. We are only adding docstring for function scoops.

toolmantim: @abeatrix ah! I don't think I've a good enough test suite / cases to verify prompt changes across more examples… I might leave that to you, unless we have a process for it I can follow.

arafatkatze: > after seeing it with the action, I wonder if we shouldn't make it so that only function folding ranges are considered for these. Especially when the action is "document code" since I would expect docstrings to be added for functions only  

  I agree with this given that just making doc strings for functions makes the most sense. IT would be still be nice to have some guardrails see this [video](https://drive.google.com/file/d/1RCRYUfYELlQQaQl95TzUUkGne_RN07ye/view?usp=sharing), so I was able to send 10K tokens for my case and I can prolly send more if I want.
image. 

WE could probably just have something that says like _function context too large for LLM etc_ for the edge case of very large function. But yeah focussing on the function alone makes perfect sense.

arafatkatze: @philipp-spiess  sorry for the ping but I wasn't sure if this got lost in the sea of comments.

PR #1117, opened 2023-09-19
refactor: simplify prefix code extraction for anthropic

abeatrix:
follow https://github.com/sourcegraph/cody/pull/1063#discussion_r1326883085

refactor: Simplify prefix code extraction for anthropic infill prompt building step

- Extract raw prefix code rather than reconstructing as suggested by @valerybugakov (thank you!)
- Remove custom infill block extraction from prefix
- Simplify infill model prompt

## Test plan

image

Infill model still works as expected.


valerybugakov (approved):
The beauty! Thank you for taking the time do implement this follow up 🙌

PR #1119, opened 2023-09-19
feat: New "Save Code to File" button and other chat code block style updates

abeatrix:
Updates to code blocks rendered in chat:
* Moves the code block buttons to the bottom of the code blocks and always shown (instead of just on hover)
* Adds a `Save Cody to File…` button
* For dark theme users, updates the chat code block background to match the editor background

## Details

This commit adds a new insertButtonOnSubmit prop to the Chat and CodeBlocks components. This allows passing in a callback function that will be triggered when the insert button is clicked.

The CopyButtonProps interface is updated to have separate callbacks for copy and insert actions. The CodeBlocks component is updated to conditionally render insert and insert new file buttons if the callback props are passed in.

This enables IDE integrations to handle insert actions separately from copy actions.

When `Save to File` button is pressed, user will be prompted to select a new file file to insert the code to. If they selected an existing file, the content will be replaced. See demo video below for details.

In addition, I've also updated the code block action styles to match Tim's design on [figma](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=5231-8881&mode=design&t=nrLPZluUa8zxyO6f-4).

## Test plan

- start Cody from this branch
- click on "Save to File" button below the code block

https://github.com/sourcegraph/cody/assets/68532117/3e86f8be-c9d7-4f1d-b634-eb4849eaab12


toolmantim:
Nice one @abeatrix!

This is great for inserting into a new file, but it feels too awkward when wanting to insert into an existing file. The interaction of choosing a file in the finder window is normally for overriding files, not appending.

Have you seen my proposal for "Insert to File" in https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=5110-7877&mode=design&t=Zrf09QUxBQz8XXP2-4? For appending, I think we need a quickpick in the middle that lists any open tabs. Would love to get your feedback and iterate on that together, and then we update this depending on where we land.

abeatrix:
@toolmantim yea I agree! I can limit this to "Save to new file" only, and then add the "insert into files" in a different PR since that will be a different button, along with your new design, what do you think? 

(I can't open figma on my phone so I'll have to wait til tomorrow to leave feedback on it, sorry! 😆)

toolmantim:
@abeatrix it's a single button in my design, so that might not make sense. But we can discuss tomorrow!

philipp-spiess (approved):

toolmantim (approved):
Pushed some design updates!

Turns out the icons I had in Figma were the wrong size to be pixel aligned (they should have been 16x16) so I updated them to use the SVGs directly from the codicons repo.

Updated the PR name, description and changelog to let people know that these buttons have moved to the bottom of the code block too

abeatrix:
> Pushed some design updates!
> 
> 
> 
> Turns out the icons I had in Figma were the wrong size to be pixel aligned (they should have been 16x16) so I updated them to use the SVGs directly from the codicons repo.
> 
> 
> 
> Updated the PR name, description and changelog to let people know that these buttons have moved to the bottom of the code block too

Thank you! I'll make a final run of the PR in a bit and merge it. Thanks for the improvement @toolmantim

Issue #1088, opened 2023-09-18
Bug: FixupTasks are not cleaned from the fixup view on failure


arafatkatze:
Currently the failure of a fixup task at any point will leave a pending job in the fixup view. 

https://github.com/sourcegraph/cody/assets/11155207/81810737-31a5-424e-b654-c6278f13e76d

I added a minor change to solve it for atleast the usecase where I saw it ([code](https://github.com/sourcegraph/cody/commit/9acd10a378b2daaca858bd0d7acb2f3c2bdc4c8a))

https://github.com/sourcegraph/cody/assets/11155207/a0e73968-fcd5-445a-80a7-431e23c95bfb

I would be happy to spin up a more clean solution in a PR for this but I wanted to know if people had other more elaborate plans for error handling that are yet to be implemented. CC @umpox  coz you wrote [this](https://github.com/sourcegraph/cody/pull/510/files#diff-319a8ef4cd07ac92062243b56f5f5aee5e906b82918cc35c08f28b526ef10937R95)

jdorfman:
Thanks @arafatkatze

abeatrix:
Thanks for looking into this. What you have looks good to me for unblocking the current issue! 

IIRC FixupTask has a method for displaying error, or somewhere in fixupController 😀

arafatkatze:
We have moved far ways from this UX so this needs to be closed.

Thanks bot!!

PR #1129, opened 2023-09-21
Temporary solution to remove failed fixup tasks from the fixup view

arafatkatze:
Solves https://github.com/sourcegraph/cody/issues/1088

## Before(Turn on voice)
https://github.com/sourcegraph/cody/assets/11155207/81810737-31a5-424e-b654-c6278f13e76d

## After(Turn on voice)
https://github.com/sourcegraph/cody/assets/11155207/a0e73968-fcd5-445a-80a7-431e23c95bfb

### Answering followup questions
From @abeatrix 
_> IIRC FixupTask has a method for displaying error, or somewhere in fixupController 😀_

That's true and the error is still gonna be displayed with this fix 
image

_IS this the perfect way to solve the problem?_ 
I think a more broader solution which has full blown error handling scenarios with a lot of different forms of failure would be a much nicer solution but for now this solution works and handles the error pretty well. As folks build more the understanding of the error handling scenarios would be better and maybe then that can be pursued but this seems okay for now. 

## Test plan
Just do what the video says

### DIFF
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -61,7 +61,10 @@ export class Fixup implements Recipe {
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
         if (truncateText(fixupTask.selectedText, quarterFileContext * 2) !== fixupTask.selectedText) {
             const msg = "The amount of text selected exceeds Cody's current capacity."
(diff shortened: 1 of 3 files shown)


arafatkatze:
CC @abeatrix @jdorfman

jdorfman:
@arafatkatze approved to run, @abeatrix will have to do the rest. :)

PR #1130, opened 2023-09-21
Automaticaly download symf

beyang:
If the path to the `symf` binary is not explicitly specified in user settings, auto-download symf the first time `/symf` is invoked.

https://github.com/sourcegraph/cody/assets/1646931/11c042c8-13ed-40a4-a436-89e7a38b4754

## Test plan

Tested locally

@philipp-spiess @valerybugakov please test it out in your local dev environment and report any issues.


valerybugakov (commented):
Got this warning upon activating the extension for the first time. The UX would be better if we could activate the extension anyway and then ask users to download the dependency to enable the search command or, even better, install everything behind the scenes. 

Screenshot 2023-09-21 at 17 33 59

toolmantim:
@valerybugakov that was exactly my thoughts with the UI too!

philipp-spiess:
> Got this warning upon activating the extension for the first time. The UX would be better if we could activate the extension anyway and then ask users to download the dependency to enable the search command or, even better, install everything behind the scenes.

Agree this doesn't feel good. Maybe we can bundle the extension code directly instead? https://github.com/microsoft/vscode-file-downloader

philipp-spiess:
Got it to work too and didn't notice any issues. Nice!

beyang:
> Got this warning upon activating the extension for the first time. The UX would be better if we could activate the extension anyway and then ask users to download the dependency to enable the search command or, even better, install everything behind the scenes.

Got rid of the extension dependency.

vscode/package.json
```diff
@@ -957,6 +957,7 @@
     "@types/stream-json": "^1.7.3",
     "@vscode/codicons": "^0.0.29",
     "@vscode/webview-ui-toolkit": "^1.2.2",
+    "axios": "^1.3.6",
```
valerybugakov, line 960: We already have `"isomorphic-fetch": "^3.0.0"`. Would it be possible to use it to avoid using multiple HTTP clients in the extension?

vscode/src/local-context/download-symf.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const symfVersion = 'v0.0.0'
+
+/**
+ * Get the path to `symf`. If the symf binary is not found, download it.
+ */
+export async function getSymfPath(context: vscode.ExtensionContext): Promise<string | null> {
+    // If user-specified symf path is set, use that
+    const config = vscode.workspace.getConfiguration()
+    const userSymfPath = config.get<string>('experimentalSymfPath')
```
valerybugakov, line 19: Note that it will require the VS Code reload on the setting change to take effect

vscode/src/local-context/download-symf.ts
```diff
@@ -0,0 +1,128 @@
...
+    const symfPath = path.join(symfContainingDir, symfFilename)
+    if (await fileExists(symfPath)) {
+        return symfPath
+    }
+
+    const symfURL = `https://github.com/sourcegraph/symf/releases/download/${symfVersion}/symf-${arch}-${platform}.zip`
+    logDebug('symf', `downloading symf from ${symfURL}`)
+
+    // Download symf binary with vscode progress api
+    await vscode.window.withProgress(
```
valerybugakov, line 42: Nice!

valerybugakov (approved):
Tested locally, works great! ✅

Bumped into a minor issue where I didn't have an Antrophic API key for symf. Adding it to settings and re-running the command was not enough. Reloading the VS Code instance fixed it.

vscode/src/local-context/download-symf.ts
beyang: Wouldn't invoking `vscode.workspace.getConfiguration()` get the latest configuration even without a restart? If not, I'll fix in a follow-up PR (okay since this is still behind a feature flag).

vscode/package.json
beyang: Using axios allows us to stream the response and pipe it directly into the unzipper. `isomorphic-fetch` appears not to support streaming and exposes a different response type that is incompatible with `pipe`. I'm not super familiar with node HTTP libraries, however, so if you feel strongly about this, I can look into it in a follow-up PR.

beyang:
> Bumped into a minor issue where I didn't have an Antrophic API key for symf. Adding it to settings and re-running the command was not enough. Reloading the VS Code instance fixed it.

Ah yes, I should remove that in a follow-up PR (no reason we shouldn't use Cody Gateway here)

vscode/src/local-context/download-symf.ts
valerybugakov: Yeah, right. We do not instantiate any class with this value 👍  It will be up to date on every invocation.

PR #1132, opened 2023-09-21
Autocomplete: Fix cody-autocomplete-claude-instant-infill experiment

philipp-spiess:
Seems like there was a conflict between @abeatrix's addition of the feature flag and @taras-yemets' restructurings. The result of `resolveDefaultProviderFromVSCodeConfigOrFeatureFlags` was no longer used for the Anthropic initialization causing users to load the wrong prompt mode if `cody-autocomplete-claude-instant-infill` is enabled.

## Test plan

- Make sure you are connected to doctom
- Be in the control group of the StarCoder experiment (needs dotcom admin access)
- Ensure no model or provider overwrites are active
- Observe that the right mode is loaded

Screenshot 2023-09-21 at 15 16 22

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -10,7 +10,7 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


taras-yemets:
Thank you, @philipp-spiess!

Issue #207, opened 2023-06-16
Fixups LLM output may be worse than cutting and pasting into chat


dominiccooney:
Feedback from @beyang via @abeatrix is that Fixups may perform worse than cutting and pasting the same code into chat. There's a hard error "The amount of text selected exceeds Cody’s current capacity" which isn't true because the same thing *is* possible in chat.

Personally, I have noticed Fixups performs better earlier in files and in small files.

Both of these things indicate that the context we're grabbing surrounding the selection is too large. We should use less—maybe even none—and rely on context fetching to provide it.

abeatrix:
We have addressed this with the changes we made in the past few sprints  👍

PR #1139, opened 2023-09-21
fix: fixup token limits

abeatrix:
RE: https://github.com/sourcegraph/cody/issues/207

> There's a hard error "The amount of text selected exceeds Cody’s current capacity" which isn't true because the same thing is possible in chat.

We should use `MAX_CURRENT_FILE_TOKENS` for the current selection instead of limiting it to a quarter of it, because surrounding code is not as important if users cannot use the command in the first place.

We can limit the context to include the quarter rule instead.

## Test plan

- Start Cody from this branch
- Select line 198 to 234 and run the /edit command, you can use this instruction for testing `correct any typos`

#### Before

you will get an error from Cody say `The amount of text selected exceeds Cody's current capacity.` 

image

#### After

Cody fixed all typos it found

image

### DIFF
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -59,7 +59,7 @@ export class Fixup implements Recipe {
         }
 
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
-        if (truncateText(fixupTask.selectedText, quarterFileContext * 2) !== fixupTask.selectedText) {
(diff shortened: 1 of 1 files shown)


arafatkatze (approved):
I agree with the PR and it makes more sense to have the token limit to 1000 versus 500 in the current case. Not to get too nitpicky, but if you were to go very literal about variable definition, the purpose of MAX_CURRENT_FILE_TOKENS is to highlight the maximum tokens that can come from the current file, which would still be more than this because after using this you're gonna add the prefix and the suffix AFTER this selection.

Perhaps can have a new variable which is MAX_CURRENT_FILE_SELECTION_TOKENS with the value of 1000 but this is cool too.

philipp-spiess (approved):
SCHMART

PR #1140, opened 2023-09-21
Agent: re-authenticate during configuration changes

chwarwick:
When the server or authtoken changes the agent needs to reload the auth status so that the autocomplete provider is regenerated with the new url & token.

This presented itself during first setup because the agent is initialized prior to the user providing account information.  Chat had continued to work because I new client was created when the configuration changed however the the autocomplete provider is created during activation and depends on events from the AuthHandler to reset itself.

resolves https://github.com/sourcegraph/sourcegraph/issues/56880 
## Test plan
Unit test added
Also did the following in IDE:

- Cleared all accounts and closed IDE
- Opened IDE created account on s2
- Verified autocomplete and chat worked
- Added a 2nd account to sourcegraph.com
- Switched account to dotcom
- Verified chat & autocomplete worked and verbose debugging showed requests to sourcegraph.com
- Switched account back to s2
- Verified chat & autocomplete worked and verbose debugging showed requests to s2

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -64,6 +64,7 @@ function initializeVscodeExtension(): void {
 
 export class Agent extends MessageHandler {
     private client: Promise<Client | null> = Promise.resolve(null)
+    private oldClient: Client | null = null
     public workspace = new AgentWorkspaceDocuments()
 
     constructor() {
@@ -79,9 +80,8 @@ export class Agent extends MessageHandler {
                 ? vscode_shim.Uri.parse(client.workspaceRootUri)
                 : vscode_shim.Uri.from({ scheme: 'file', path: client.workspaceRootPath })
 
-            const extensionConfig = client.extensionConfiguration ?? client.connectionConfiguration
-            if (extensionConfig) {
-                await this.setClient(extensionConfig)
+            if (client.extensionConfiguration) {
+                this.setClient(client.extensionConfiguration)
             }
 
             setUserAgent(`${client?.name} / ${client?.version}`)
@@ -101,7 +101,7 @@ export class Agent extends MessageHandler {
             return {
                 name: 'cody-agent',
                 authenticated: codyClient.sourcegraphStatus.authenticated,
-                codyEnabled: codyStatus.enabled,
+                codyEnabled: codyStatus.enabled && (client.extensionConfiguration?.accessToken ?? '').length > 0,
                 codyVersion: codyStatus.version,
             }
         })
@@ -137,12 +137,7 @@ export class Agent extends MessageHandler {
             vscode_shim.onDidCloseTextDocument.fire(this.workspace.agentTextDocument(document))
         })
 
-        const configurationDidChange = async (config: ExtensionConfiguration): Promise<null> => {
-            await this.setClient(config)
-            return null
-        }
-
-        this.registerRequest('extensionConfiguration/didChange', configurationDidChange)
+        this.registerNotification('extensionConfiguration/didChange', config => this.setClient(config))
 
         this.registerRequest('recipes/list', () =>
             Promise.resolve(
@@ -174,6 +169,7 @@ export class Agent extends MessageHandler {
             return null
         })
         this.registerRequest('autocomplete/execute', async (params, token) => {
+            await this.client // To let configuration changes propagate
             const provider = await vscode_shim.completionProvider()
             if (!provider) {
                 console.log('Completion provider is not initialized')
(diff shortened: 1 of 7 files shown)


agent/src/index.test.ts
```diff
@@ -93,7 +111,19 @@ describe.each([
     it('initializes properly', async () => {
         const serverInfo = await client.handshake(clientInfo)
         assert.deepStrictEqual(serverInfo.name, 'cody-agent', 'Agent should be cody-agent')
-        assert.deepStrictEqual(serverInfo.codyEnabled, true, 'Cody should be enabled')
+        assert.deepStrictEqual(
+            serverInfo.codyEnabled,
+            name !== 'NotConfigured',
+            'Cody should be enabled when configured'
+        )
```
philipp-spiess, line 118:
```suggestion
        expect(serverInfo.codyEnabled).not.toBe('NotConfigured')
```

Might be a bit easier to read + nicer visualized in a failed run if we use vitest assertions instead of the built-in `assert` lib

agent/src/index.test.ts
```diff
@@ -93,7 +111,19 @@ describe.each([
...
+    })
+
+    it('handles config changes correctly', async () => {
+        await client.request('extensionConfiguration/didChange', {
+            accessToken: process.env.SRC_ACCESS_TOKEN ?? 'invalid',
+            serverEndpoint: process.env.SRC_ENDPOINT ?? 'invalid',
+            customHeaders: {},
+        })
```
philipp-spiess, line 126: This test has no assertion? Is it only testing for it not to crash (in this case maybe we can make this more explicit)

agent/src/agent.ts
```diff
@@ -249,17 +251,20 @@ export class Agent extends MessageHandler {
...
             affectsConfiguration: () =>
                 // assuming the return value below only impacts performance (not
                 // functionality), we return true to always triggger the callback.
                 true,
         })
-        vscode_shim.commands.executeCommand('cody.auth.sync').then(
-            () => {},
-            () => {}
-        )
+        await vscode_shim.commands.executeCommand('cody.auth.sync')
```
philipp-spiess, line 267: Not sure I understand why this is necessary. In my mental mode, this event is only emitted by the auth provider when it's status change, so if we call `auth.reload` before, it would already have emitted this event when any auth settings have changed, no? 🤔

philipp-spiess (approved):

agent/src/agent.ts
olafurpg: Good catch! This command is not needed after the introduction of `agent.auth.reload`

agent/src/index.test.ts
olafurpg: The correct translation is `expect(serverInfo.codyEnabled).toBe(name !== 'Not Configured')`. Right now, we only use `assert` in this file so I think it's better to do a wholesale refactoring in the future if we prefer `expect` (which I personally don't like using so much)

vscode/src/main.ts
```diff
@@ -288,8 +288,8 @@ const register = async (
         vscode.commands.registerCommand('cody.auth.signin', () => authProvider.signinMenu()),
         vscode.commands.registerCommand('cody.auth.signout', () => authProvider.signoutMenu()),
         vscode.commands.registerCommand('cody.auth.support', () => showFeedbackSupportQuickPick()),
-        vscode.commands.registerCommand('cody.auth.sync', () => {
-            void contextProvider.syncAuthStatus()
+        vscode.commands.registerCommand('cody.auth.sync', async () => {
+            await contextProvider.syncAuthStatus()
             void featureFlagProvider.syncAuthStatus()
```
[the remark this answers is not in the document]
olafurpg: Good catch. I updated this logic to allow both of these to run in parallel (if possible). I also added a comment to clarify that we must return a promise to allow `AuthProvider` to await on the result.

agent/src/index.test.ts
olafurpg: I updated the test to have the title `handles config changes without errors` and a command saying there are intentionally no assertions. Strictly speaking, we could remove the enclosing `it` callback but I like having it nested inside a test.

For good and bad, the connection is stateful so the order of tests is important.

olafurpg:
@philipp-spiess Thank you for the review! I have addressed the comments and verified 100% that the problem is fixed in IntelliJ. Will cut a new release today with this fix (along with a few other fixes!)

olafurpg (approved):

PR #1141, opened 2023-09-22
VS Code: module scope telemetry service

valerybugakov:
## Context

- Dropped the export of the deprecated `logEvent` function.
- Replaced `createVSCodeTelemetryService` with the module scope `telemetryService` export. This removes the need to pass the `telemetryService` down from the extension init call. Chatted about it with @philipp-spiess and @dominiccooney.

## Test plan

CI


abeatrix (approved):
THIS IS AWESOME!!! Makes it so much easier to import it! Thank you for making this change!

philipp-spiess (commented):
BLISS

Issue #789, opened 2023-08-22
Handling for questions that Cody can't answer


beyang:
Example questions:
* "How many files of X type in Y directory"
* "Where is $SYMBOL being used?"

This question has come up a few times as people often attempt it when trying out Cody for the first time. It also seems like it could be useful in the context of exploring new projects.

Customer requests:
* https://sourcegraph.slack.com/archives/C02842Z6YAK/p1691983303904029
* https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1694545660675509?thread_ts=1694454085.492699&cid=C04MSD3DP5L

## Alternative idea

Respond with terminal commands that could help answer the user's question.

beyang:
Closing in favor of new chat design https://github.com/sourcegraph/cody/issues/1410

PR #1148, opened 2023-09-23
[WIP] Handling support for questions that can be answered by linux commands

arafatkatze:
Attempt to solve -> https://github.com/sourcegraph/cody/issues/789 

I thought of a couple ways to solve this ^ problem. One way was mainly through classification of all the questions coming in into two forms. 
A. Those that can be answered by embeddings+ LLM
B. Those that can't be answered by embeddings+ LLM but can possibly be answered by linux commands

I thought of separate the questions into two categories and then have a different prompt for the second category(Lots of work but it can be worth it if it's an absolute dealbreaker for new users). There were a few ways to separate questions into categories
1. Make a fast LLM call to classify the question itself(Similar to what I did here -> https://github.com/sourcegraph/cody/pull/1083)  -> Problem with this is that it adds to much latency and that's bad UX.
2. Use some kind of heuristics, regex match combination with weights to do a faster classification. But this wasn't easy to do and i couldn't capture the nuance. 
3.  Built like a small TensorFlow model that just classifies a question into a different types and just use that on the user prompt and then treat the two questions differently. It can theoretically be done and it's also fast enough but I felt like adding a small classifier model might be something that must be done if it's absolutely necessary. 

Looking at the problem again, I decided that another way to solve the problem would be just to encourage the existing prompts to be able to answer Linux commands when they feel more appropriate. I was actually quite cautious in making any changes to the existing prompt because this is the core of how Cody works and it should not be manipulated unless it's absolutely necessary. So I made the bare minimum change that I could make so that other stuff is not affected while we are still able to answer Linux commands as much as possible and the responses have been pretty good for some questions. 

This prompt change makes a slight improvement in the responses from cody and encourages it to give linux commands where it can so it would be good to have this but I am happy to explore other methods posted above too. 

## Before
image

## After
image

## Before
image

## After
image

## Before 
image

## After 
image

## Before 
image

## After
image

## Before
image

## After
image

## Test plan

Just try out a few prompts and you shall know

### DIFF
--- a/lib/shared/src/chat/preamble.ts
+++ b/lib/shared/src/chat/preamble.ts
@@ -7,9 +7,9 @@ export interface Preamble {
 }
 
 const actions = `You are Cody, an AI-powered coding assistant created by Sourcegraph. You work inside a text editor. You have access to my currently open files. You perform the following actions:
-- Answer general programming questions.
-- Answer questions about the code that I have provided to you.
-- Generate code that matches a written description.
+- Answer general programming questions, including those related to Linux commands and shell commands.
+- Answer questions about the code that I have provided to you, including symbol lookups and directory queries.
+- Generate code and shell commands that match a written description.
 - Explain what a section of code does.`
 
 const rules = `In your responses, obey the following rules:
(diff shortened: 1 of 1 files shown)


arafatkatze:
Closing this PR as the changed Prompts were working only for GPT 4 model on my codebase(Because I accidentally set GPT4 as my default setting) but that does not work for Claude. A silly testing mistake that I made while trying this out. 

Let me spin up a new PR for to solve this issue so that it works with claude.

PR #1150, opened 2023-09-23
Fix unit tests on Windows

DanTup:
I don't know how sound all of these changes are, but they fix the failing tests I have on Windows.

1. Use platform-correct paths (drive letters + backslashes on Windows) in tests and debug printing
  Previously lots of code assumed posix-style paths then had mismatches on windows because of slash direction
2. Replace fake Uri class with vscode-uri
  As part of doing (1), I found that `vscode.Uri` was missing. While adding it, I realised there was a mock `Uri` implementation (presumably to avoid importing `vscode` where it won't be available), however the `vscode-uri` package (which does not depend on VS Code) was also already imported, so I removed the fake `Uri` class and just used this. This required updating a bunch of places (`Uri` => `URI`) and I don't know if there are other implications of doing this, but I think it's better.
3. Add launch config to run current test file under debugger
  Makes it easy to debug tests (or run without debugging for the current file) without having to use the terminal
4. Use prettier when editing JSON files like `.vscode/settings.json` and `.vscode/launch.json` (these already seemed to be formatted using Prettier, but it wasn't set explicitly so the formatting gets messed up if you didn't have Prettier as default formatter)

## Test plan

Tested using `pnpm test:unit`.
There's still a failing test on Windows (`blocks.test.ts`) but I'll look at that another PR as it appears to be a different issue to these URI/path ones (perhaps line-ending related, I'm not certain yet).


vscode/src/completions/context/graph-section-observer.ts
```diff
@@ -269,7 +269,7 @@ export class GraphSectionObserver implements vscode.Disposable, GraphContextFetc
         const lines: string[] = []
         // eslint-disable-next-line ban/ban
         this.activeDocuments.forEach(document => {
-            lines.push(path.normalize(vscode.workspace.asRelativePath(document.uri)))
+            lines.push(document.uri.toString())
```
philipp-spiess, line 272: This view is used in the debugging UI (You can see it by running the "autocomplete tracing" action) so I'd prefer a relative representation here. Do you think we can still use that?

philipp-spiess (approved):
Nice!

vscode/src/completions/context/graph-section-observer.ts
DanTup: I couldn't see where this showed up in the trace window to verify, but I've put the relative path back. The reason I removed `normalize` was that in tests the URIs are fake and not in the workspace, which means the output of this function is not a path, but a string file:// URI (`asRelativePath` returns the string version of the input if it's not in the workspace). Calling normalize on a string that is a file URI on Windows ends up with values like "file:\\c%3A\\document1.ts".

DanTup:
@philipp-spiess do I need to do something to make the checks all run correctly for a fork, or does this need to be a branch in sourcegraph/ instead of a fork?

philipp-spiess:
@DanTup Yeah seems like you're right:

> google-github-actions/auth failed with: retry function failed after 1 attempt: the GitHub Action workflow must specify exactly one of "workload_identity_provider" or "credentials_json"! If you are specifying input values via GitHub secrets, ensure the secret is being injected into the environment. By default, secrets are not passed to workflows triggered from forks, including Dependabot.

Do you mind making this a branch on the main origin?

DanTup:
I've re-opened this at https://github.com/sourcegraph/cody/pull/1256 from a branch here without the fork.

PR #1152, opened 2023-09-25
Modifying the chat prompts to respond with terminal commands if needed 

arafatkatze:
Attempt to solve -> https://github.com/sourcegraph/cody/issues/789 from @beyang 
At first I made https://github.com/sourcegraph/cody/pull/1148 but that only worked for GPT 4 model so now I made this PR and it works great for many cases. 

I thought of a couple ways to solve this ^ problem. One way was mainly through classification of all the questions coming in into two forms. 
A. Those that can be answered by embeddings+ LLM
B. Those that can't be answered by embeddings+ LLM but can possibly be answered by terminal commands

I thought of separate the questions into two categories and then have a different prompt for the second category(Lots of work but it can be worth it if it's an absolute dealbreaker for new users). There were a few ways to separate questions into categories
1. Make a fast LLM call to classify the question itself(Similar to what I did here -> https://github.com/sourcegraph/cody/pull/1083)  -> Problem with this is that it adds to much latency and that's bad UX.
2. Use some kind of heuristics, regex match combination with weights to do a faster classification. But this wasn't easy to do and i couldn't capture the nuance. 
3.  Built like a small TensorFlow model that just classifies a question into a different types and just use that on the user prompt and then treat the two questions differently. It can theoretically be done and it's also fast enough but I felt like adding a small classifier model might be something that must be done if it's absolutely necessary. 

Looking at the problem again, I decided that another way to solve the problem would be just to encourage the existing prompts to be able to answer terminal commands when they feel more appropriate. I was actually quite cautious in making any changes to the existing prompt because this is the core of how Cody works and it should not be manipulated unless it's absolutely necessary. So I made the bare minimum change that I could make so that other stuff is not affected while we are still able to answer terminal commands as much as possible and the responses have been pretty good for most questions and it's great that I can extend this too if needed. 

This prompt change makes a slight improvement in the responses from cody and encourages it to give terminal commands where it can so it would be good to have this but I am happy to explore other methods posted above too. Also it was very important to give it examples of terminal commands coz it wouldn't give the answers without it. Also based on my testing it didn't affect the responses in other forms of questions asked to cody which is really awesome.  I would completely understand if people wouldn't want the core cody prompts to be untouched but I wanted to give one shot at this problem and this seemed like a good way. 

### Before
image

## After
image

### Before
image

### After
image

### Before 
image

### After 
image

### Before 
image
image

### After
image
image

### Before
image

### After

image

## Test plan

Just try out a few prompts and you shall know

## Test plan

Try out my branch with the examples that I showed

### DIFF
--- a/lib/shared/src/chat/recipes/chat-question.ts
+++ b/lib/shared/src/chat/recipes/chat-question.ts
@@ -7,7 +7,7 @@ import {
     populateCurrentEditorContextTemplate,
     populateCurrentEditorSelectedContextTemplate,
 } from '../../prompt/templates'
-import { truncateText } from '../../prompt/truncation'
+import { getLinuxCommandContextMessages, truncateText } from '../../prompt/truncation'
 import { Interaction } from '../transcript/interaction'
 
 import { isSingleWord, numResults } from './helpers'
@@ -71,9 +71,11 @@ export class ChatQuestion implements Recipe {
 
         // Add selected text as context when available
         if (selection?.selectedText) {
+            // Add some context messages so that the LLM has some context about how to respond with basic Linux commands
             contextMessages.push(...ChatQuestion.getEditorSelectionContext(selection))
         }
 
+        contextMessages.push(...getLinuxCommandContextMessages())
         return contextMessages
     }
 
--- a/lib/shared/src/chat/recipes/chat-question.ts
+++ b/lib/shared/src/chat/recipes/chat-question.ts
@@ -7,7 +7,7 @@ import {
     populateCurrentEditorContextTemplate,
     populateCurrentEditorSelectedContextTemplate,
 } from '../../prompt/templates'
-import { getLinuxCommandContextMessages, truncateText } from '../../prompt/truncation'
+import { getTerminalCommandContextMessages, truncateText } from '../../prompt/truncation'
 import { Interaction } from '../transcript/interaction'
 
 import { isSingleWord, numResults } from './helpers'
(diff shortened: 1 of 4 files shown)


arafatkatze:
@dominiccooney Great Catch!! I changed the solution to add OS selection: Everything else is the same


## Previous solution
image


## New Solution
image

lib/shared/src/chat/recipes/chat-question.ts
```diff
@@ -71,9 +71,11 @@ export class ChatQuestion implements Recipe {
 
         // Add selected text as context when available
         if (selection?.selectedText) {
+            // Add some context messages so that the LLM has some context about how to respond with basic terminal commands
             contextMessages.push(...ChatQuestion.getEditorSelectionContext(selection))
         }
 
+        contextMessages.push(...getTerminalCommandContextMessages())
```
beyang, line 78: These messages look like they are included for all Cody interactions, which might reduce response quality in situations where they aren't necessary (and take up space that could be used for other context). I wonder what the relative impact on quality of including just one or two (or none) of these is.

beyang (commented):
Overall this PR seems promising. A couple of thoughts:
* We need an example suite for commands/chat so we can evaluate how changes like this affect the range of use cases we want to support
* We could roll this out after the Oct 4 in an A/B test or behind a feature flag.

Will add to discussion agenda in our next planning cycle meeting.

lib/shared/src/chat/recipes/chat-question.ts
arafatkatze: @beyang I was very cautious of the problem of adding examples to prompt potentially affecting other search results. So I did quite a bit of prompt engineering, and I realized there's diminishing returns to adding more prompt examples. That being said the results on questions which don't need terminal responses were still almost exactly as good as before and the addition seemed to have made no difference to the previous results on non terminal questions for me. 

With that being said, I want to reduce the potential impact of this PR as much as I can, while at the same time maintaining the ability to answer the few terminal command questions that we want to be able to answer to. So now I've reduced it to just 2 examples which is okayish.  

I've tried this whole thing from zero to N, and this seems like a reasonable compromise. In the case of zero examples, the results were basically non-existent and would almost be the same as before(which is bad). So this seems like a reasonable tradeoff where we're still getting good enough answers with terminal commands, and we're only adding the minimal number of prompts that we need to add. 

Obviously, it should be good to release this behind a feature flag and see how people respond. Let me know if you want me to make any other changes.


## Some examples of the current state of the PR 
image
image
image

arafatkatze:
With the presense of better models this won't be needed anymore

PR #1163, opened 2023-09-25
refactor: expand code action range

abeatrix:
Refactor: expand code action range

Expand the range used for code actions to the range of the function rather than just the current line. This improves the quality of fixup actions by providing more context.

The range is now obtained by calling getTargetFoldingRange() instead of just expanding to the range of the curret line. 

The getCursorFoldingRange() utility was renamed to getTargetFoldingRange() to better reflect its purpose.

Also renamed utils.ts into utils/folding-range.ts

#### Before

https://github.com/sourcegraph/cody/assets/68532117/9c31b619-14a4-4e4c-9c53-2a1d63cfb96b

#### After

https://github.com/sourcegraph/cody/assets/68532117/2df34653-6cfe-4385-b348-a1d2406808a7


vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+// ------------------------ HELPER FUNCTIONS ------------------------ //
+
+/**
+ * Gets the outermost folding ranges for classes in the given document.
+ *
+ * @param ranges - The folding ranges for the document
+ * @param uri - The URI of the document
+ * @returns The outermost folding ranges corresponding to classes in the document.
+ */
+export async function getOuterClassFoldingRanges(
```
philipp-spiess, line 94: If this is just a helper function, why is it exported from the module? 🦭

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+import { getFoldingRanges, getSymbols } from '.'
+
+/**
+ * Gets the folding range containing the target position.
+ *
+ * @param uri - The URI of the document to get folding ranges for
+ * @param target - The target position number
+ * @returns The folding range containing the target position, or undefined if none found
+ */
+export async function getTargetFoldingRange(uri: vscode.Uri, target: number): Promise<vscode.Selection | undefined> {
```
philipp-spiess, line 12: I think the fact that these use the folding range API is not really important for the caller, is it? I say this because I think we can just name this `getSmartRanges` or `getSections` or whatever instead, to better explain that this is _more_ than just the folding range provider

philipp-spiess (approved):

vscode/src/editor/utils/folding-ranges.ts
abeatrix: good points, let me remove the export for helpers!

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+/**
+ * Combines overlapping folding ranges in the given array into nested ranges.
+ *
+ * This takes an array of folding ranges, sorts them by start position, and reduces them
+ * down to combine any overlapping ranges into nested ranges.
+ *
+ * @param ranges - Array of folding ranges to combine.
+ * @returns Array of combined folding ranges with overlapping ranges nested.
```
dominiccooney:
```suggestion
 * @returns Array of combined folding ranges with overlapping ranges merged.
```

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+    // look for ranges that end at -1 position from the next folding range
+    // then combine them into a single range where the new range will be the union of the start positions of the first range that connected to the n next range(s), and the end position of the last range in the chain
+    let chains: vscode.FoldingRange[] = []
+    const combinedRanges: vscode.FoldingRange[] = []
+    // iterate through ranges and add connected ranges into chains array, clean the chains array everytime we find a disconnected range
+    for (let i = 0; i < ranges.length; i++) {
+        const range = ranges[i]
+        // Check if current range is connected to previous range
+        const prevRange = ranges[i - 1]
+        if (prevRange && prevRange.end === range.start - 1) {
```
dominiccooney, line 199: What's the -1 about, here? Seems very specific.

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+            chains.push(prevRange)
+        }
+
+        // Check if current range is connected to the last combined range
+        const lastCombinedRange = combinedRanges.at(-1)
+        // If connected, add to chains array
+        if (lastCombinedRange && lastCombinedRange.end === range.start - 1) {
+            combinedRanges.pop()
+            chains = [lastCombinedRange, range]
```
dominiccooney, line 208: The condition on line 229 seems very close to line 222.

The modification of chains is hard to follow... there's only zero/one/two items in there for the body of the loop. Can you just track the start and end in variables?

Can this method be simplified? It seems a lot like the method for removing overlapping ranges, but it should only consider ranges which start at end-1 to be overlapping.

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+
+/**
+ * Adds the selection range to the prompt string.
+ *
+ * @param prompt - The original prompt string
+ * @param code - The code snippet to include in the prompt
+ * @returns The updated prompt string with the code snippet added
+ */
+export function addSelectionToPrompt(prompt: string, code: string): string {
+    return prompt + '\nHere is the code: \n<Code>' + code + '</Code>'
```
dominiccooney: `<code>` is a HTML tag, does that make it good to use (LLM gets the meaning) or not (more likely to appear in the code we are quoting and gets confusing)?

vscode/src/editor/utils/index.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @param uri - The URI of the document to get folding ranges for
+ * @param type - Optional type of folding range to filter on:
+ *   - 'imports' - Only import folding ranges
+ *   - 'comment' - Only comment folding ranges
+ *   - 'regions' - All non-import and non-comment folding ranges
+ * If no type specified, returns all folding ranges.
+ *
+ * @returns The array of folding ranges for the document.
+ */
+export async function getFoldingRanges(uri: vscode.Uri, type?: string): Promise<vscode.FoldingRange[]> {
```
dominiccooney, line 29: If type can only be specific strings, then let's type it to only allow those strings.

Even better, can we use the VScode folding range type that we're filtering on here? It will simplify this method a bit.

vscode/src/editor/utils/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    switch (type) {
+        case 'imports':
+            return ranges.filter(r => r.kind === vscode.FoldingRangeKind.Imports)
+        case 'comment':
+            return ranges.filter(r => r.kind === vscode.FoldingRangeKind.Comment)
+        case 'regions':
+            return ranges.filter(
+                r => r.kind !== vscode.FoldingRangeKind.Imports && r.kind !== vscode.FoldingRangeKind.Comment
```
dominiccooney, line 46: VScode has FoldingRangeKind.Region, why not use that?

vscode/src/editor/utils/index.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Gets the symbol information for the given document URI.
+ *
+ * @param uri - The URI of the document to get symbols for.
+ *
+ * @returns A promise that resolves to an array of SymbolInformation objects representing the symbols in the document.
+ */
+export async function getSymbols(uri: vscode.Uri): Promise<vscode.SymbolInformation[]> {
+    const symbols =
```
dominiccooney, line 61: Just return it directly?

dominiccooney (changes requested):
Comments inline. I think we can simplify this somewhat.

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+ *
+ * @param uri - The URI of the document to get folding ranges for
+ * @param target - The target position number
+ * @returns The folding range containing the target position, or undefined if none found
+ */
+export async function getTargetFoldingRange(uri: vscode.Uri, target: number): Promise<vscode.Selection | undefined> {
+    // Check if symbols is available in the doc by its file uri and look for langauge id
+    const doc = await vscode.workspace.openTextDocument(uri)
+    // Remove text-based languages that don't support symbols or folding ranges
+    const textBaseLangIds = ['markdown', 'plaintext', 'json', 'sql']
```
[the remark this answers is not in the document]
abeatrix: Because we use the symbol class to filter the class range, but for files that do not have symbols supported, we can still try to select the closest range through this step. Or do you think we should just not support these type of files at all?

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+    foldingRanges: vscode.FoldingRange[],
+    target: number,
+    isPlainText?: boolean
+): vscode.FoldingRange | undefined {
+    if (!foldingRanges?.length) {
+        return undefined
+    }
+
+    /**
+     * NOTE (bee) The purpose of filtering to keep only folding ranges that contain other folding ranges is to find
```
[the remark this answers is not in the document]
abeatrix: This step is just to remove all the nested folding ranges returned by the API, if the outermost folding range does not contain other folding range, we will still be able to capture the outermost ranges:
image

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+    // we should first find the the range with the largest end range to identify class ranges
+    const outermostFoldingRanges = removeNestedFoldingRanges(ranges)
+
+    // Check outerRanges array for the string 'class' in each starting line to confirm they are class ranges
+    // Filter the ranges to remove ranges that did not contain classes in their first line
+    const doc = await vscode.workspace.openTextDocument(uri)
+    const firstLines = outermostFoldingRanges.map(r => doc.lineAt(r.start).text)
+
+    return outermostFoldingRanges
+        .filter((r, i) => firstLines[i].includes('class') || firstLines[i].startsWith('object'))
```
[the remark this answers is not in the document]
abeatrix: this was inspired by [scala singleton object](https://docs.scala-lang.org/tour/singleton-objects.html) that structures similar to `class`

vscode/src/editor/utils/index.ts
abeatrix: > can we use the VScode folding range type that we're filtering on here
True! Will updated this!

vscode/src/editor/utils/index.ts
abeatrix: i wanted to make sure comments and imports are not included in regions, but you're right, that'd be different than what is defined by the folding range kind so i will update this to use FoldingRangeKind.Region, thanks!

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+    // If lanaguage id not available, find class object ranges heuristically
+    const classRanges = isSupported ? classes : await getOuterClassFoldingRanges(ranges, uri)
+
+    const targetRange = getNonClassOutermostFoldingRanges(classRanges, ranges, target, isSupported)
+    if (!targetRange) {
+        console.error('No folding range found containing cursor')
+        return undefined
+    }
+
+    return new vscode.Selection(targetRange.start, 0, targetRange.end + 2, 0)
```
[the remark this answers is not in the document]
philipp-spiess: @abeatrix If we add another line, don't we have to set the the selection end character to something > 0`?

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,262 @@
...
+ * By removing the nested ranges, you are left with only the top-level outermost folding ranges.
+ *
+ * @param ranges - Array of folding ranges
+ * @returns Array containing only folding ranges that do not contain any nested child ranges
+ */
+export function removeNestedFoldingRanges(ranges: vscode.FoldingRange[], isTextBased = false): vscode.FoldingRange[] {
+    const filtered = isTextBased ? combineNeiborFoldingRanges(ranges) : ranges
+
+    return filtered.filter(
+        cur => !filtered.some(next => next !== cur && next.start <= cur.start && next.end >= cur.end)
```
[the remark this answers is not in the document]
arafatkatze: Two things 
1. even if the ranges are identical but are different objects, they will not filter each other out. The condition 
`next !== cur`  in the inner  `some()` function checks for object identity, not for equality of the range values.

In this code, a 'cur' range will only be removed if there's another 'next' range that is not the same object` (
next !== cur
) `and that completely encompasses 'cur' `(
next.start <= cur.start && next.end >= cur.end
)`. Thus, two identical ranges (same start and end but different objects) will not remove each other.

2. There is no chance of having identical ranges because all the folding ranges are a directed acyclic graph. See [this](https://github.com/sourcegraph/cody/pull/1116#discussion_r1336323775)

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,228 @@
...
+        getFoldingRanges(uri).then(r => r?.filter(r => !r.kind)),
+        isSupported
+            ? getSymbols(uri)
+                  .then(r => r?.filter(s => s.kind === vscode.SymbolKind.Class))
+                  .then(s => s?.map(symbol => symbol.location.range))
+            : [],
+    ])
+
+    // If lanaguage id not available, find class object ranges heuristically
+    const classRanges = isSupported ? classes : await getOuterClassFoldingRanges(ranges, uri)
```
arafatkatze, line 32:
```suggestion
    const classRanges = doc.languageId ? classes : await getOuterClassFoldingRanges(ranges, uri)
```

isSupported is the check for textBasedLanguages but you wanna run `getOuterClassFoldingRanges` for languages that don't support langserver and you want to do that when the language id is not available so then you use heuristics to find class ranges.

In the case of Json, plaintext, sql, markdown the whole concept of classes doesn't make sense so you get empty array of classes(which is what you want)

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    const targetRange = getNonClassOutermostFoldingRanges(classRanges, ranges, target, isSupported)
```
arafatkatze, line 34:
```suggestion
    const isPlainText = !isSupported
    const targetRange = getNonClassOutermostFoldingRanges(classRanges, ranges, target, isPlainText)
```

Fundamentally the idea is that `isSupported` is the inverse of `isPlainText`. You wanted an isPlainText check to combine neighbour ranges for plaintext documents as shown [here](https://github.com/sourcegraph/cody/pull/1163/files#diff-7a994af9d659344fecd4e2771e864229c0cf1158c4ce5acbe08957b906cf765aR172) 

But the inverse was missed in the document flow so now the `combineNeiborFoldingRanges` actually works for NON text based files(I.E. Normal files like typescript etc)  which I don't think was your intention.

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,228 @@
...
+ * Combines neighboring folding ranges in the given array into single ranges.
+ *
+ * This looks for folding ranges that end on the line before the next range starts,
+ * indicating they are neighbors with no gap between. It combines such neighbors
+ * into a single range from the start of the first to the end of the last.
+ *
+ * @param ranges - Array of folding ranges to combine
+ * @returns Array of combined folding ranges
+ */
+function combineNeiborFoldingRanges(ranges: vscode.FoldingRange[]): vscode.FoldingRange[] {
```
arafatkatze, line 189: My first thought was that this function's logic can be greatly simplified. I checked it and found as simple Linear time algo for exactly the same thing 

```
function combineNeighboringFoldingRanges(ranges: vscode.FoldingRange[]): vscode.FoldingRange[] {
    const combinedRanges: vscode.FoldingRange[] = [];

    for (const range of ranges) {
        const lastCombinedRange = combinedRanges[combinedRanges.length - 1];

        if (lastCombinedRange && lastCombinedRange.end === range.start - 1) {
            // Expand the last combined range to include the current range
            lastCombinedRange.end = range.end;
        } else {
            // Current range is not connected; push it as a new combined range
            combinedRanges.push(range);
        }
    }

    return combinedRanges;
}
```

But now i am not able to understand how this function helps the algorithm for extending folding ranges and if the algorithm would be dysfunctional without this.
For the text based use cases or even for non-text use cases(Typescript etc). Can you elaborate on the reasoning for having this?

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
arafatkatze: Some thoughts on this https://drive.google.com/file/d/1R7IccCtYln0xcf7_-I2iCp0wwSzoCUot/view?usp=sharing

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
abeatrix: @philipp-spiess the + 1 includes the line where the bracket that was missing, and the extra +1 at 0 position is to capture the missing bracket, while not breaking even if the line after the bracket does not exist because the character position is at 0. See the example here where the line after the bracket is already at the end of file, and we were still able to capture the whole block:
eof

vscode/src/editor/utils/folding-ranges.ts
abeatrix: OMG yes this was where i messed up! Thanks for finding this i was wondering why it wasn't working anymore 😆 

#### Before your change

Doc added on top of the wrong folding range:

image


#### After your change

Doc is added at the correct folding range

image

THANK YOU!

vscode/src/editor/utils/folding-ranges.ts
abeatrix: doc.languageId returns 'plaintext' for code files that do not have lang server support, which I am trying to capture here. I will add a comment to reflect this, thanks!

vscode/src/editor/utils/folding-ranges.ts
abeatrix: refactor to <review> 👍

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
dominiccooney: Thanks @arafatkatze that has cleared up a bunch.

Let's revise the comment and mention that it isn't whether there's regions, but symbols, that we care about.

Next questions though:

Later we're looking for regions which contain the selection. Why do we need symbols for that?

For markdown folding @arafatkatze demonstrated that H1 folding is really broad. But isn't this analogous to the class level folding?

vscode/src/code-actions/fixup.ts
```diff
@@ -42,7 +52,7 @@ export class FixupCodeAction implements vscode.CodeActionProvider {
 
     private getCodeActionInstruction = (diagnostics: vscode.Diagnostic[]): string => {
         const intent: FixupIntent = 'edit'
-        return `/${intent} Fix the following error${diagnostics.length > 1 ? 's' : ''}:\n${diagnostics
+        return `/${intent} Fix the following error${diagnostics.length > 1 ? 's' : ''}: ${diagnostics
```
dominiccooney, line 55: With the broader selection, how does Cody know where specifically the error is now?

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,225 @@
+import * as vscode from 'vscode'
+
+import { getFoldingRanges, getSymbols } from '.'
+
+/**
+ * Gets the folding range containing the target position.
+ *
+ * NOTE: Use getSmartSelection from utils/index.ts instead
```
dominiccooney, line 8: Let's update this with more justification for why to use `getSmartSelection`. This might not be enough context to motivate people to do it.

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,225 @@
...
+ * @param targetLine - The target line position number
+ * @returns The folding range containing the target position, or undefined if none found
+ */
+export async function getTargetFoldingRange(
+    uri: vscode.Uri,
+    targetLine: number
+): Promise<vscode.Selection | undefined> {
+    // Check if symbols is available in the doc by its file uri and look for language id
+    const doc = await vscode.workspace.openTextDocument(uri)
+    // Remove text-based languages that don't support symbols or folding ranges
```
dominiccooney, line 20: We've seen from @arafatkatze that these languages all support folding ranges. I expect there's some packages that add symbols for sql. Can we make this more transparent about the intent... the results for folding aren't good for these languages so we want to exclude them, something like that? It will help people maintain this list correctly.

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,225 @@
...
+    classRanges: vscode.Range[],
+    foldingRanges: vscode.FoldingRange[],
+    targetLine: number,
+    isPlainText?: boolean
+): vscode.FoldingRange | undefined {
+    if (!foldingRanges?.length) {
+        return undefined
+    }
+
+    // NOTE (bee) The purpose of filtering to keep only folding ranges that contain other folding ranges is to find
```
dominiccooney, line 75: I don't think "folding range which contain other folding ranges" is what you want. You want the outermost folding range. You can be outermost without containing another folding range.

For example, if I have a JavaScript file with a top-level function, then that might be a range which doesn't contain any other ranges. But you would want to include that range, right?

vscode/src/editor/utils/folding-ranges.ts
```diff
@@ -0,0 +1,225 @@
...
+    // the outermost folding range enclosing the cursor position.
+    // Folding ranges can be nested - you may have a folding range for a function that contains folding ranges for inner code blocks.
+    // By filtering to ranges that contain other ranges, it removes the inner nested ranges and keeps only the outermost parent ranges.
```
dominiccooney, line 78: This also doesn't sound right. If you have A contains B contains C, do you want B or not? B is not outermost, so it sounds like "no." But B contains C, so it sounds like "yes" you do want it.

Let's get the intent clear in the comment and explain a bit about why and then we can review whether the code achieves the intent.

dominiccooney (changes requested):

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
philipp-spiess: @abeatrix This only works if there is a trailing newline though does it? Otherwise it might pick a line that is higher than the total lines of the document. Trailing newlines are inserted automatically in vscode but I think you can disable that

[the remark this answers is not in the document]
abeatrix: in the example there is no line 36, do you mean trailing newline will add an invisible line 36? If yes, I wasn't aware of that setting, but i will add a math.max just in case 👍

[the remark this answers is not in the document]
philipp-spiess: Yeah that's the case! In that case returning line 36 makes no sense and might break for some callers. Math.max also won't help because that would exclude the character on the last line. I think we need to get the number of chars on the last line and use that for the end

vscode/src/editor/utils/folding-ranges.ts
abeatrix: Will do! (we dont want to motivate people to do it btw 😅  will add my reasons to the comments, thanks for pointing this out!)

vscode/src/code-actions/fixup.ts
arafatkatze: This could be the answer for it https://github.com/sourcegraph/cody/commit/a974e7bac94d9ab92aaeeec45bad78ebdd302629

arafatkatze: Currently for an error here
image

The prompt going to LLM is 
```

The user has the following code in their selection:
  <selectedCode>    private postProcess(rawResponse: string): string {
        let completion = extractFromCodeBlock(rawResponse)

        const trimmedPrefixContainNewline = this.options.docContext.prefix
            .slice(this.options.docContext.prefix.trimEnd().length)
            .includes('\n')
        if (trimmedPrefixContainNewline) {
            // The prefix already contains a `\n` that Claude was not aware of, so we remove any
            // leading `\n` followed by whitespace that Claude might add.
            completion  completion.replace(/^\s*\n\s*/, '')
        } else {
            completion = trimLeadingWhitespaceUntilNewline(completion)
        }

        // Remove bad symbols from the start of the completion string.
        completion = fixBadCompletionStart(completion)

        return completion
    }
</selectedCode>

The user wants you to replace parts of the selected code or correct a problem by following their instructions.
Provide your generated code using the following instructions:
<instructions>
/edit Fix the following error: ```Unexpected keyword or identifier.```
</instructions>
'
```


The prompt doesn't have the granularity so as far as the LLM is concerned the error could be anywhere.  I added some prompt engineering which gives it the exact location of the error which has improved results a bit. So now the prompt is 

```
he user has the following code in their selection:
<selectedCode>    private postProcess(rawResponse: string): string {
        let completion = extractFromCodeBlock(rawResponse)

        const trimmedPrefixContainNewline = this.options.docContext.prefix
            .slice(this.options.docContext.prefix.trimEnd().length)
            .includes('\n')
        if (trimmedPrefixContainNewline) {
            // The prefix already contains a `\n` that Claude was not aware of, so we remove any
            // leading `\n` followed by whitespace that Claude might add.
            completion  completion.replace(/^\s*\n\s*/, '')
        } else {
            completion = trimLeadingWhitespaceUntilNewline(completion)
        }

        // Remove bad symbols from the start of the completion string.
        completion = fixBadCompletionStart(completion)

        return completion
    }
</selectedCode>

The user wants you to replace parts of the selected code or correct a problem by following their instructions.
Provide your generated code using the following instructions:
<instructions>
/edit Fix the following error: ```Unexpected keyword or identifier.``` in the line "            completion  completion.replace(/^\s*\n\s*/, '')"
</instructions>
'
``` 

This way the llm knows where to make the change and has the additional context so that it doesn't do something absurd or makes a change at the wrong place.

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
arafatkatze: @dominiccooney I tried this out and you are right https://github.com/sourcegraph/cody/pull/1163#issuecomment-1735513161 

Because they are different objects `next !== cur ` holds true and because of similar values `next.start <= cur.start && next.end >= cur.end` holds true.

Thanks for the clarification on language.

vscode/src/editor/utils/folding-ranges.ts
arafatkatze: @abeatrix Sorry for ping. Hope this doesn't get lost in the sea of comments about this PR.

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
abeatrix: fixed!

vscode/src/editor/utils/folding-ranges.ts
abeatrix: updated

vscode/src/editor/utils/folding-ranges.ts
abeatrix: Hey @arafatkatze thanks for the pings! I;ve updated logic for this functions too, do you think your suggestion aligns with the update?

vscode/src/editor/utils/folding-ranges.ts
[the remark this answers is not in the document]
arafatkatze: 1.  > Later we're looking for regions which contain the selection. Why do we need symbols for that?

I am sure you already know this but keeping it here just for understanding. 
https://github.com/sourcegraph/cody/blob/461262872679375be45d599d9da736d22c8f2043/vscode/src/editor/utils/folding-ranges.ts#L39-L44
This comment covers it. The only thing I would add to this comment is 
>  filter the filtered folding ranges to only include the outer ranges containing target line(Ideally this is a function)

2. 

> For markdown folding @arafatkatze demonstrated that H1 folding is really broad. But isn't this analogous to the class level folding?
Yes it very much is. The different levels of headings H1->H2->H3 follow the nesting structure of a class -> function -> function etc. Sub-nesting in this is fairly straight forward. 

The one thing I would add is that I am not sure if having smart selection would benefit so much for those cases. Markdown/free form text would benefit the most from autocomplete. I doubt if accounting for smart selection would be helpful in any way.

vscode/src/code-actions/fixup.ts
arafatkatze: @abeatrix ^

vscode/src/editor/utils/folding-ranges.ts
arafatkatze: @abeatrix My question was predominantly about the utility of this function the logic is fine. I see that this function is supposed to work for things that don't have lang server so i tried it in this file 

```
package com.example

// Top-level comment
object Main {

  // Object-level comment
  
  // Nested object
  object NestedObject {
    def nestedMethod(): Unit = {
      // Implementation here
    }
  }

  // First class
  /**
   * NestedClass class nested inside FirstClass
   * anotherMethod() method in NestedClass 
   * Does something
   */
  class FirstClass {
    // Nested class
    class NestedClass {
      def anotherMethod(): Unit = {
        // Do something
      }
    }
    
    def methodOne(): Unit = {
      if (true) {
        // Inside if statement
      } else {
        // Inside else statement
      }
    }
  }

  // Second class
  class SecondClass {
    def methodTwo(): Unit = {
      // Second method
    }
  }
  
  def main(args: Array[String]): Unit = {
    // Main method body
    for (i <- 1 to 10) {
      // Inside for loop
    }
  }
}
``` 

And I saw that this function helped in combining a few ranges for the if/else condition. And then also another case where there was a docstring and a function right after it. But I don't understand the utility of having this function. IF you are gonna remove nested ranges  etc. Why have this function? 

Basically I can't figure out if this function should be a part of the algorithm? Does it do a marginal improvement? Yes, absolutely. But maybe there is a more elaborate reasoning for the existence of this function and I can't pick that up. The logic is okay though.

abeatrix: >  I see that this function is supposed to work for things that don't have lang server so i tried it in this file

if you save the file as a plaintext file, you should be able to see how it works in action, but basically if you have a plaintext file that looks like:

```
class FirstClass {
  // Nested class
  class NestedClass {
    def method(
      params1: String = "",
      params2: String = ""
    ) {
        // Do something
    }
  }
}
```

Without this, it would returns:

```
    def method(
      params1: String = "",
      params2: String = ""
    ) {
```

when we want:

```
    def method(
      params1: String = "",
      params2: String = ""
    ) {
        // Do something
    }
```

vscode/src/editor/utils/folding-ranges.ts
abeatrix: updated!

abeatrix:
@dominiccooney i'm going to merge to unblock my other work since i know you'll be traveling! i've updated the code based on your suggestions and confirmed with Philipp that this works, (also as seen in our demo session yesterday), but I will schedule a live review session with you to walk through them, thanks!

PR #1164, opened 2023-09-25
change: use infill prompt by default for anthropic

abeatrix:
Update to use infill prompt by default.

Based on the numbers, users are getting higher-quality of suggestions using the infill prompt. 

Removed the old prompt for anthropic that does not support infilling.

## Test plan

Set provider to anthropic without setting modal. You should see an infill prompt being used.

Keeping the infill model feature flag for now as it's being used in the A/B test


philipp-spiess (changes requested):
Please also remove it from the settings enum, provider creation, and the feature flags 👼 

(grep for `claude-instant-infill` and `CodyAutocompleteClaudeInstantInfill`

philipp-spiess (approved):

PR #1168, opened 2023-09-25
fix: use lastTriggerDocContext instead current

abeatrix:
fix: use lastTriggerDocContext instead current 

- Add `lastTriggerDocContext` field to `LastInlineCompletionCandidate` interface to store document context for last triggered completion.
- Pass `lastTriggerDocContext instead` of current docContext when handling unwanted completion to ensure correct context is used.
- Update reuse logic to also check if the current prefix is shorter than the last triggered prefix before clearing the last candidate. 
- Set `lastTriggerDocContext` when creating a completion candidate.

#### After

https://github.com/sourcegraph/cody/assets/68532117/b40a9294-3b53-430a-bf08-856904f21cff

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/last-candidate.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/last-candidate.test.ts
@@ -25,6 +25,16 @@ describe('[getInlineCompletions] reuseLastCandidate', () => {
             lastTriggerCurrentLinePrefix: document.lineAt(position).text.slice(0, position.character),
             lastTriggerNextNonEmptyLine: nextNonEmptyLine,
             lastTriggerSelectedInfoItem,
+            lastTriggerDocContext: {
+                currentLinePrefix: 'const foo = ',
+                currentLineSuffix: '',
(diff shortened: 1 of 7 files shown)


vscode/src/completions/get-inline-completions-tests/last-candidate.test.ts
```diff
@@ -19,16 +19,26 @@ describe('[getInlineCompletions] reuseLastCandidate', () => {
...
             },
+            lastTriggerDocContext: {
+                currentLinePrefix: prefix,
+                currentLineSuffix: suffix,
+                multilineTrigger: null,
+                nextNonEmptyLine,
+                prevNonEmptyLine,
+                prefix,
+                suffix,
+                contextRange: new Range(position.character, 0, position.character, 0),
```
philipp-spiess, line 40: Let's just call into `getCurrentDocContext` here so this test does not break if we change `getCurrentDocContext`

philipp-spiess (approved):
Nice!

valerybugakov (approved):
👍

PR #1170, opened 2023-09-26
Autocomplete: Improved manual completion trigger

philipp-spiess:
This PR improves the manual completion triggered via the keyboard shortcut:

- We are now able to separate between manually triggered completions and hover over the ghost text
- The refined trigger kinds are now part of the suggestions logged
- When using the manual keyboard triggers, we now hide the current ghost text so that we are able to cycle through new completions with this shortcut
- Opt out from cache hits when manual suggestions are triggered

## Test plan

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:suggested {"multiline":false,"triggerKind":"Automatic","providerIdentifier":"anthropic","providerModel":"claude-instant-infill","languageId":"typescript","type":"inline","multilineMode":null,"id":"6370b94f-4c07-44ac-903c-149aa75abba5","contextSummary":{"duration":0.13420796394348145},"source":"Network","lineCount":1,"charCount":13,"latency":443.20716601610184,"displayDuration":311.8174999952316,"read":false,"accepted":false,"otherCompletionProviderEnabled":false,"completionsStartedSinceLastSuggestion":1}
```

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -17,6 +17,7 @@ import {
     InlineCompletionsParams,
     InlineCompletionsResultSource,
     LastInlineCompletionCandidate,
+    TriggerKind,
 } from './get-inline-completions'
 import * as CompletionLogger from './logger'
 import { CompletionEvent } from './logger'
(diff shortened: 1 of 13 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -137,6 +142,14 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
         const tracer = this.config.tracer ? createTracerForInvocation(this.config.tracer) : undefined
         const graphContextFetcher = this.config.graphContextFetcher ?? undefined
 
+        const triggerKind =
+            this.lastManualCompletionTimestamp && this.lastManualCompletionTimestamp > Date.now() - 500
```
valerybugakov, line 146: Lol, that's so strange that there's no way to pass additional args to `vscode.commands.executeCommand('editor.action.inlineSuggest.trigger')`

valerybugakov (approved):

Issue #1096, opened 2023-09-18
Improved action buttons for chat code blocks


toolmantim:
Our code block controls have been a problem in recent user testing. Code output in chat is important for anything that still relies on chat, which is most built-in commands, custom commands, and chat itself.

This is a basic building block that we need to get right to instil confidence and lovability. We also lag behind other leading extensions for the manipulating code blocks in chat.

Here's a new design for code block actions that replace our current "Copy" and "Insert at Cursor" buttons that appear on hover:

![image](https://github.com/sourcegraph/cody/assets/153/3104059c-f714-4a3d-9eb6-f4c087566327)

Figma:
https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=5110-7877&mode=design&t=KpUpBEhodg72KYQY-4

toolmantim:
I've tightened up the spacing and added the standard quick pick-style dropshadow:
<img src="https://github.com/sourcegraph/cody/assets/153/19be653a-4005-4e88-bc83-4294ad4c17c8" width="376"/>

Also added a prototype to show the experience:
https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=5123-9170&mode=design&t=Ey1vVRTX2jpZjtT5-4

https://github.com/sourcegraph/cody/assets/153/43b26a92-bf7f-4274-b46c-573b11d96271

yaosamo:
@toolmantim ah I see it now :) I was talking about the other one tho! 
![CleanShot 2023-09-25 at 17 24 35@2x](https://github.com/sourcegraph/cody/assets/5090118/8749d3ce-557a-45f1-a774-69d9c573abfe)
It's also only visible on hover, maybe that's why I couldn't see it at a glance on first use.

toolmantim:
@yaosamo you found it! Luckily that should be covered by this change too, because the code is shared. But that variant wasn't included in this design because that App chat UI is on pause for the moment.

toolmantim:
@abeatrix looks like we still have some things to clean up before this goes out in the next release, which I'll document here properly soon, but the things I've spotted are:
- Disabled for insert at cursor if no file is open
- Some kind of "Copied" feedback
- Border colors look off on different themes

abeatrix:
[...]

PR #1173, opened 2023-09-26
add: visual feedback on codeblock action clicks

abeatrix:
Part of https://github.com/sourcegraph/cody/issues/1096

feat: add visual feedback when code block actions clicked

- Add CheckCodeBlockIcon SVG 
- Update CodeBlockActionButtons to briefly show CheckIcon on click
- Add timeouts to revert CheckIcon back to the original after 5 seconds
- Display error message on `insert` click when no file or editor is opened.

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/e260369e-be2a-4d7c-8623-f34e2a74b1cd


toolmantim:
Thank you @abeatrix!

I just updated the [Figma](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=5231-8881&mode=design&t=faCtcPVSJOP8FT0P-4) with the states here.

For the copy button, instead of the check could we use this version of the copy icon I just made with a check in it?

```svg
<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="currentColor">
	<path d="M3,1L2,2v10l1,1V2h6.4l-1-1H3z"/>
	<path d="M7.8,14.7c-0.2-0.2-0.3-0.5-0.4-0.7H5V4h5l3,3v0.1c0.3,0.1,0.7,0.2,1,0.3V6.6L10.4,3H5L4,4v10l1,1h3C8,14.9,7.9,14.8,7.8,14.7z"/>
	<path fileRule="evenodd" clipRule="evenodd" d="M9.8,8.7c1.8-1.2,4.3-0.8,5.6,1.1s0.8,4.3-1.1,5.6c0,0,0,0,0,0c-1.8,1.2-4.3,0.7-5.5-1.1C7.5,12.4,8,9.9,9.8,8.7z M11.9,13.7l2.4-3.2l-0.8-0.6l-2.1,2.8l-1.3-1l-0.6,0.8l1.7,1.4L11.9,13.7L11.9,13.7z"/>
</svg>
```

For "Save Code to File…" we don't need to change it from what's on main.

For "Insert at Cursor", is it possible to change it to disabled when no file is open, rather than keep it there and show an error?

abeatrix:
> For the copy button, instead of the check could we use this version of the copy icon I just made with a check in it?

ok!

> For "Save Code to File…" we don't need to change it from what's on main.

ok!

> For "Insert at Cursor", is it possible to change it to disabled when no file is open, rather than keep it there and show an error? 
That would require some structure change that i'm afraid I won't have capacity to work on before the oct release, as we currently create the buttons when the chat view is loaded and when new chat is added.

dominiccooney (approved):

PR #1186, opened 2023-09-26
refactor: unify fixup and refactor commands&events

abeatrix:
context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1695733960910859

refactor: unify fixup and refactor commands & event names

Right now all the commands that uses the fixup recipe are logged under fixup. This PR adds the source to the event name and log all fixup event based on its source name, for example, if the command comes from 'editor' context menu:
Screenshot 2023-09-26 at 2 42 53 PM

- Unified 'cody.fixup.new' and 'cody.command.edit-code' commands into 'cody.command.edit-code'
- Updated references to command in package.json, MessageProvider.ts, and other files
- Updated fixup command to use new `edit-code` command

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -258,7 +258,8 @@
       {
         "command": "cody.command.edit-code",
         "category": "Ask Cody",
-        "title": "Refactor Code"
+        "title": "Refactor Code",
+        "icon": "$(sparkle)"
       },
       {
         "command": "cody.command.explain-code",
@@ -386,12 +387,6 @@
         "command": "cody.command.inline-touch",
         "category": "Cody",
         "title": "Touch"
-      },
-      {
-        "command": "cody.fixup.new",
-        "category": "Cody",
-        "title": "Fixup",
-        "icon": "$(sparkle)"
       },
       {
         "command": "cody.fixup.open",
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -504,7 +504,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 return null
             }
             case /^\/edit(\s)?/.test(text):
-                await vscode.commands.executeCommand('cody.fixup.new', { instruction: text })
+                await vscode.commands.executeCommand('cody.command.edit-code', { instruction: text }, 'sidebar')
                 return null
             default: {
                 if (!this.editor.getActiveTextEditor()?.filePath) {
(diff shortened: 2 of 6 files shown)


arafatkatze (approved):

kelsey-brown:
@abeatrix one quick thought - I'm actually not sure we need the chat:submitted event at all. Based on our conversation [here](https://sourcegraph.slack.com/archives/C05AGQYD528/p1692811370505429?thread_ts=1692054351.708989&cid=C05AGQYD528), chat:submitted fires for any event (chat or command) where an input is submitted through chat. But if all our chat and command actions have their own events (e.g. explain:executed, chat:executed, etc) with source in the metadata to determine where the interaction happened (inline, sidebar, etc), isn't the chat:submitted event redundant? 

Is there some user action the chat:submitted event captures that isn't captured by any other event? If not, I don't think we need the chat:submitted event at all.

abeatrix:
> @abeatrix one quick thought - I'm actually not sure we need the chat:submitted event at all. Based on our conversation [here](https://sourcegraph.slack.com/archives/C05AGQYD528/p1692811370505429?thread_ts=1692054351.708989&cid=C05AGQYD528), chat:submitted fires for any event (chat or command) where an input is submitted through chat. But if all our chat and command actions have their own events (e.g. explain:executed, chat:executed, etc) with source in the metadata to determine where the interaction happened (inline, sidebar, etc), isn't the chat:submitted event redundant?
> 
> Is there some user action the chat:submitted event captures that isn't captured by any other event? If not, I don't think we need the chat:submitted event at all.

how are we capturing chat events atm? with `chat:submitted` or through the `chat-question` recipe?

Is there currently a `chat:executed` event?

kelsey-brown:
> Is there currently a chat:executed event?

Yep, we have: 
`CodyVSCodeExtension:recipe:chat-question:executed` and `CodyVSCodeExtension:recipe:inline-chat:executed`

abeatrix:
@kelsey-brown removed!

kelsey-brown:
Thank you!! 

At some point it would probably make sense to have a single event that captures chats (not commands) regardless of source, and then put the source in the metadata for that chat event (the same way you're doing it here for commands). But that's just a nice to have.

kelsey-brown (approved):

abeatrix:
@kelsey-brown thanks for the review! we still have some time before next week's release, but we should start getting new events from this change from tomorrow's pre-release.

PR #1187, opened 2023-09-26
feat: auto-dismiss chat error banner

kalanchan:
QOL improvement: dismisses error banner on a set interval

Previously the error banner would stay on the chat view until the user manually clicks the `x`, this change adds an interval to each banner so it dismisses automatically after 3 seconds

## Test plan
- run more than one command sequentially, wait for error banner to pop up in the chat view
- wait 3 seconds, error banner should dismiss automatically (and sequentially if theres more than one!)

Before:

https://github.com/sourcegraph/cody/assets/51868853/f4b233d0-5303-4e0f-b73e-decd269589cf

After:

https://github.com/sourcegraph/cody/assets/51868853/9ef8a154-3be0-4e82-8924-2fb493111030


abeatrix:
How are we sure the users have seen the error message before it disappeared though? Maybe we should check if the view is focused first before we start the timeout?

toolmantim (commented):
Nice! This makes a lot of sense for this particular type of alert (not running multiple commands). I was worried that auto-dismissing might not make sense for other types of errors, but looking at [handleError calls](https://sourcegraph.com/search?q=context%3Aglobal+repo%3Asourcegraph%2Fcody+handleError&patternType=standard&sm=1&groupBy=repo) it looks safe to add… these aren't really actionable errors.

But after some testing… I think we should switch to VS Code’s native notifications, and I think they're superior to our own:

```diff
diff --git a/vscode/src/chat/ChatViewProvider.ts b/vscode/src/chat/ChatViewProvider.ts
index 83863b03..7848c286 100644
--- a/vscode/src/chat/ChatViewProvider.ts
+++ b/vscode/src/chat/ChatViewProvider.ts
@@ -252,7 +252,10 @@ export class ChatViewProvider extends MessageProvider implements vscode.WebviewV
      * It does not display error message as assistant response
      */
     public handleError(errorMsg: string): void {
-        void this.webview?.postMessage({ type: 'errors', errors: errorMsg })
+        // Even though this is an error, we use showWarningMessage, because our
+        // errors are more informational than actionable. We want VS Code to
+        // auto-dismiss them (which showErrorMessage doesn't do).
+        void vscode.window.showWarningMessage(errorMsg)
     }
 
     /**
diff --git a/vscode/src/chat/MessageProvider.ts b/vscode/src/chat/MessageProvider.ts
index ce4d68b0..6a56c403 100644
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -280,7 +280,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
 
     public async executeRecipe(recipeId: RecipeID, humanChatInput = ''): Promise<void> {
         if (this.isMessageInProgress) {
-            this.handleError('Cannot execute multiple recipes. Please wait for the current recipe to finish.')
+            this.handleError('Cannot run command. Please wait for current commant to finish.')
             return
         }
 
```

They auto-dismiss and de-dupe (not to we can `rm` all the code to render our custom ones etc):

https://github.com/sourcegraph/cody/assets/153/f6a5a596-c96b-4028-9813-fe8afa113ec0

PR #1192, opened 2023-09-26
vscode: add telemetry v2 SDK

bobheadxi:
This change adds integrations with the new telemetry SDK at https://github.com/sourcegraph/telemetry to the VSCode Cody client. Context is available in https://docs.sourcegraph.com/dev/background-information/telemetry and #wg-event-logging-everywhere. A notable change is that we can now only export private metadata of arbitrary shape on an allowlist basis, [defined here](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/telemetry/sensitivemetadataallowlist/sensitiviemetadataallowlist.go?L13:1-23:2), on the serverside, and these new events are intended for export via the connected instance, rather than directly to dotcom.

Overview:

1. Events should now be created using both the legacy event-logging **and** the new event-recording for a transition period. In the future we are going to remove the legacy event-logging entirely.
   - Updated callsites should provide an additional parameter so that we don't duplicate events: `logEvent(name, parameters, { hasV2Event: true }`
2. The new telemetry recorder should be used going forward - it enforces stricter metadata constraints and exports events via the connected Sourcegraph instance only.
   - The recorder's exporter (`GraphQLTelemetryExporter`) will check the instance's version to determine if it needs to export to the old `logEvents` mutation as well.
   - Arbitrarily shaped metadata is discouraged, but it _can_ be exported by being added to the allowlist: https://docs.sourcegraph.com/dev/background-information/telemetry#sensitive-attributes

Details of changes:

- `lib/shared/src/telemetry-v2`: predefined allowed event features, actions, and metadata keys, and a constructor for a `TelemetryRecorderProvider`
- `lib/shared/src/sourcegraph-api`: new method for the new `telemetry { recordEvents }` mutation, and a exporter implementation to plug into https://github.com/sourcegraph/telemetry. This exporter makes a best-effort attempt at sending things to the old API if the Sourcegraph server is outdated. I've also added support for https://github.com/sourcegraph/sourcegraph/pull/57056 - the only way to add an actor to an event now is to make the record request as an actor. The default implementation also attaches metadata we already include today about the configuration
- `vscode/src/services/telemetry-v2.ts`: partial clone of `vscode/src/services/telemetry.ts` using the new APIs
- `vscode/src/main.ts`: hook in init/update of the new APIs and one example usage at the `fixup` callsite
- `lib/shared/src/sourcegraph-api/telemetry/GraphQLTelemetryExporter.ts`: conditional exporter (pre-5.2.0, use legacy; post-5.2.0, use new)
- `lib/shared/src/sourcegraph-api/telemetry/MockServerTelemetryExporter.ts`: exporter to test server `test/fixtures/mock-server.ts`

Closes https://github.com/sourcegraph/sourcegraph/issues/56919
Part of https://github.com/sourcegraph/sourcegraph/issues/56813

## Test plan

- [x] Automated tests
- [x] Manual testing with extension in dev mode
image
- [x] Manual testing with extension hardcoded to send real events image  
image
- [x] Manual testing with extension harcoded to use legacy export mode 
image

(note the difference in `source` in the second-last image, translated in the backend, and the last image, which exports using the legacy exporter)


vscode/src/services/telemetryV2.ts
```diff
@@ -0,0 +1,56 @@
...
+ * Get a recorder for recording telemetry events in the new telemetry framework:
+ * https://docs.sourcegraph.com/dev/background-information/telemetry
+ *
+ * A new recorder should generally be retrieved in order to pick up configuration
+ * changes.
+ */
+export function getRecorder():
+    | TelemetryRecorder<EventFeature, EventAction, MetadataKey, BillingCategory, BillingProduct>
+    | undefined {
+    return telemetryRecorderProvider?.getRecorder()
```
valerybugakov, line 27: Does that mean that `telemetryRecorderProvider?.getRecorder()` picks up configuration changes? 
AFAIU, we need to call `createOrUpdateTelemetryRecorderProvider` to consume the updated config.

vscode/src/services/telemetryV2.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export async function createOrUpdateTelemetryRecorderProvider(
+    config: ConfigurationWithAccessToken,
+    isExtensionModeDevOrTest: boolean
+): Promise<void> {
+    if (config.telemetryLevel === 'off' || isExtensionModeDevOrTest) {
```
valerybugakov, line 34: It's helpful to log events into the console if `isExtensionModeDevOrTest`. That's what we do with the current `logEvent` implementation. It can be implemented as a wrapper around the recorder-getter with the following API:

```ts
import { recordEvent } from '...'

// If `isExtensionModeDevOrTest === true` logs event to the console.
// No need to call a getter function with a potentially `undefined` result.
recordEvent('cody', 'installed')
```

We can throw an error if the underlying `telemetryRecorderProvider` is not initialized because we always want it to be in place. We use this pattern for other singleton entities (e.g., `localStorage`).

lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
```diff
@@ -0,0 +1,79 @@
...
+    constructor(private config: ConfigurationWithAccessToken) {}
+
+    public processEvent(event: TelemetryEventInput): void {
+        if (!event.parameters.metadata) {
+            event.parameters.metadata = []
+        }
+        if (!event.parameters.privateMetadata) {
+            event.parameters.privateMetadata = {}
+        }
+        event.parameters.metadata.push(
```
valerybugakov, line 56: Does that mean that all these parameters are added to every event?

pnpm-lock.yaml
```diff
@@ -599,6 +605,44 @@ packages:
       - supports-color
     dev: false
 
+  /@ardatan/relay-compiler@12.0.0(graphql@16.8.1):
```
valerybugakov, line 608: It looks like the `@sourcegraph/telemetry` package comes with many dependencies, and we do not actually use most of them. Looking at the [package.json](https://github.com/sourcegraph/telemetry/blob/main/package.json#L37-L42), I see that most of the deps can be moved to `devDependencies` to avoid bloating consumers `node_modules` folders and slowing down the installation process.

```json
  "dependencies": {
    "@graphql-codegen/cli": "^5.0.0",
    "@graphql-codegen/typescript": "^4.0.1",
    "rxjs": "^7.8.1",
    "typescript-common": "^1.0.5-beta"
  },
```

valerybugakov (commented):
Great work, excited to migrate to this SDK from our old school `logEvent`!

lib/shared/src/configuration.ts
```diff
@@ -1,5 +1,25 @@
...
+            return 1
+        case 'keyword':
+            return 2
+
+        case 'blended':
+            return 10
+        case 'unified':
+            return 11
+    }
+}
```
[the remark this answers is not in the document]
bobheadxi: Is there a way to make sure that `Record<ConfigurationUseContext, number>` does not accidentally miss any `ConfigurationUseContext` values? I chose switch because I know that in switch cases, without `default`, Typescript will yell at you for missing anything in `ConfigurationUseContext` in the switch case

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1,5 +1,7 @@
 import fetch from 'isomorphic-fetch'
 
+import { TelemetryEventInput } from '@sourcegraph/telemetry/dist/api'
```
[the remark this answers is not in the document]
bobheadxi: This is something @nathan-downs and I would appreciate some feedback on actually, I have no idea what is the best way to publish a npm package "well", coming from Go where I just need to push to GitHub 😅 Do you have a reference or guide for how to set up the build and publish process in line with best practices?

vscode/src/services/telemetryV2.ts
bobheadxi: `createOrUpdateTelemetryRecorderProvider` will update the `telemetryRecorderProvider`, but unlike OpenTelemetry's SDK we do not update existing `TelemetryRecorder` in-place with new configuration from `telemetryRecorderProvider`. This docstring just means that to make sure you are recording using the latest configuration, you need to create a new recorder at the callsite where the event is being recorded

Is that clear enough, or do you think that might be too easy to hold wrong? We can also invest in making everything possible to update in-place

lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
bobheadxi: Yep! This _seems_ to be what already happens today in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/telemetry/EventLogger.ts?L71:3-83:10

pnpm-lock.yaml
bobheadxi: Thanks, I'll get that fixed - this ties into https://github.com/sourcegraph/cody/pull/1192#discussion_r1339171445, basically we don't have a good grasp on how to publish this good 😅

lib/shared/src/sourcegraph-api/graphql/client.ts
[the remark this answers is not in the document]
bobheadxi: Fixed!

vscode/src/services/telemetryV2.ts
bobheadxi: Replaced to export a single `recorder` instance we update ourselves instead

vscode/src/services/telemetryV2.ts
bobheadxi: Added a console exporter and error throwing if used before init!

pnpm-lock.yaml
bobheadxi: pnpm diff mostly gone, thanks!

bobheadxi:
I think I have this _mostly_ working - I see things working in test output mode, but can't seem to get the events to export to Sourcegraph.com when I use `pnpm run dev`:

image

The response seems to contain nothing, and when I take the exact data to the GraphQL console, I can record the event:

```
mutation RecordTelemetryEvents($events: [TelemetryEventInput!]!) {
	telemetry {
		recordEvents(events: $events) {
			alwaysNil
		}
	}
}
```
Variables:

```json
{
  "events": [
    {
      "feature": "cody.savedLogin",
      "action": "executed",
      "source": {
        "client": "VSCode.Cody",
        "clientVersion": "0.14.1"
      },
      "parameters": {
        "version": 0,
        "metadata": [
          {
            "key": "contextSelection",
            "value": 1
          },
          {
            "key": "chatPredictions",
            "value": 0
          },
          {
            "key": "inline",
            "value": 1
          },
          {
            "key": "nonStop",
            "value": 0
          },
          {
            "key": "guardrails",
            "value": 0
          }
        ]
      }
    }
  ]
}
```

https://sourcegraph.com/api/console#%7B%22query%22%3A%22mutation%20RecordTelemetryEvents(%24events%3A%20%5BTelemetryEventInput!%5D!)%20%7B%5Cn%5Cttelemetry%20%7B%5Cn%5Ct%5CtrecordEvents(events%3A%20%24events)%20%7B%5Cn%5Ct%5Ct%5CtalwaysNil%5Cn%5Ct%5Ct%7D%5Cn%5Ct%7D%5Cn%7D%22%2C%22operationName%22%3A%22RecordTelemetryEvents%22%2C%22variables%22%3A%22%7B%5Cn%20%20%5C%22events%5C%22%3A%20%5B%5Cn%20%20%20%20%7B%5Cn%20%20%20%20%20%20%5C%22feature%5C%22%3A%20%5C%22cody.savedLogin%5C%22%2C%5Cn%20%20%20%20%20%20%5C%22action%5C%22%3A%20%5C%22executed%5C%22%2C%5Cn%20%20%20%20%20%20%5C%22source%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%20%20%5C%22client%5C%22%3A%20%5C%22VSCode.Cody%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%5C%22clientVersion%5C%22%3A%20%5C%220.14.1%5C%22%5Cn%20%20%20%20%20%20%7D%2C%5Cn%20%20%20%20%20%20%5C%22parameters%5C%22%3A%20%7B%5Cn%20%20%20%20%20%20%20%20%5C%22version%5C%22%3A%200%2C%5Cn%20%20%20%20%20%20%20%20%5C%22metadata%5C%22%3A%20%5B%5Cn%20%20%20%20%20%20%20%20%20%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22key%5C%22%3A%20%5C%22contextSelection%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22value%5C%22%3A%201%5Cn%20%20%20%20%20%20%20%20%20%20%7D%2C%5Cn%20%20%20%20%20%20%20%20%20%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22key%5C%22%3A%20%5C%22chatPredictions%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22value%5C%22%3A%200%5Cn%20%20%20%20%20%20%20%20%20%20%7D%2C%5Cn%20%20%20%20%20%20%20%20%20%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22key%5C%22%3A%20%5C%22inline%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22value%5C%22%3A%201%5Cn%20%20%20%20%20%20%20%20%20%20%7D%2C%5Cn%20%20%20%20%20%20%20%20%20%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22key%5C%22%3A%20%5C%22nonStop%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22value%5C%22%3A%200%5Cn%20%20%20%20%20%20%20%20%20%20%7D%2C%5Cn%20%20%20%20%20%20%20%20%20%20%7B%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22key%5C%22%3A%20%5C%22guardrails%5C%22%2C%5Cn%20%20%20%20%20%20%20%20%20%20%20%20%5C%22value%5C%22%3A%200%5Cn%20%20%20%20%20%20%20%20%20%20%7D%5Cn%20%20%20%20%20%20%20%20%5D%5Cn%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%20%20%5D%5Cn%7D%22%7D

lib/shared/src/sourcegraph-api/telemetry/GraphQLTelemetryExporter.ts
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            const insiderBuild = siteVersion.length > 12 || siteVersion.includes('dev')
+            if (insiderBuild) {
+                this.exportMode = '5.2.0-5.2.1' // TODO: use '5.2.2+' after https://github.com/sourcegraph/sourcegraph/pull/57719
+            } else if (siteVersion === '5.2.0' || siteVersion === '5.2.1') {
+                this.exportMode = '5.2.0-5.2.1'
+            } else if (siteVersion > '5.2.2') {
+                this.exportMode = '5.2.2+'
+            } else {
```
bobheadxi, line 49: for https://github.com/sourcegraph/sourcegraph/pull/57719

valerybugakov (commented):
@bobheadxi, I pushed [a commit](https://github.com/sourcegraph/cody/pull/1192/commits/a5aef8337b19241d43e5aaa78b4dd42bef50d955) to fix the problem with the `recordEvents` mutation.

lib/shared/src/telemetry-v2/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * use 'cody.foobar' instead.
+ */
+export type EventFeature = 'cody' | 'cody.savedLogin' | 'cody.fixup' | 'cody.fixup.apply'
+
+/**
+ * Actions should denote a generic action within the scope of a feature.
+ */
+export type EventAction = 'succeeded' | 'failed' | 'installed' | 'executed' | 'created'
+
+export type MetadataKey = 'metadata' | 'duration' | 'lineCount' | 'charCount'
```
philipp-spiess, line 16: Is that shared for _all_ events? I.e if one event needs a `charCount` metadata, every event can use a `charCount`? If we delete an event, does it mean we have to manually check each field if it's still used?

also when we add `metadata` does it mean that every even can have a metadata field and bypass these checks? 🤔

bobheadxi:
CI checks green, tidied up the last bits, did manual testing - I think this is good for a real review now :)

lib/shared/src/telemetry-v2/index.ts
philipp-spiess: Thanks @bobheadxi. I’m sure you'll make good decisions here, I don't have strong opinions about API and everything. Only one thing that I noticed:

>values we always assume to be safe, because values can only be numeric

Does that mean no more dicts and strings? Also does numeric mean integer or decimals? Would that mean everything has to be an enum? There are quite a few strings we use for autocomplete logging right now and there are thoughts of logging even more (including parts of the prompt), what would be the workaround here?

E.g.:

- Completion UUID
- Tree-sitter based trigger identifier
- Provider name and model name
- Language ID (we don't allowlist languages)

philipp-spiess: Screenshot 2023-10-20 at 14 35 41

If I understand this correctly only 32bit integers are allowed. This means that most of the fields we currently log for autocomplete won't be possible anymore (or require long managed enums), or am I misunderstanding something?

vscode/src/non-stop/FixupController.ts
```diff
@@ -305,6 +309,15 @@ export class FixupController
             const codeCount = countCode(replacementText)
             const source = task.source
             telemetryService.log('CodyVSCodeExtension:fixup:applied', { ...codeCount, source })
+
+            // This is currently a test/demo of the new events system
+            telemetryRecorder.recordEvent('cody.fixup.apply', 'succeeded', {
+                metadata: [
+                    ['lineCount', codeCount.lineCount],
+                    ['charCount', codeCount.charCount],
+                ],
```
philipp-spiess, line 318:
```suggestion
                metadata: {
                    lineCount: codeCount.lineCount,
                    charCount: codeCount.charCount,
                },
```

Unrelated to the above question: wouldn't this be nicer as an object (won't allow duplicated definitions for the same argument, for example)

lib/shared/src/telemetry-v2/index.ts
bobheadxi: There are 2 types of metadata in the new system:

1. Metadata, which must be numeric. These can be added freely with no legal or security or engineering oversight for lots of reasons: no risk of PII/sensitive data exposure, minimal risks with payload sizes affecting our shared systems, etc. etc
2. "Private metadata", which can be arbitrarily shaped. These can only be exported from the instance on an allowlist basis (the current mechanism allowlists based on a hardcoded set in the backend based on `feature` and `action`), and should undergo some scrutiny.

Answering your specific cases:

- Completion UUID: Can we use OpenTelemetry trace IDs instead? These are accepted in Sourcegraph backends and are an open and widely adopted standard for tracking interactions. The backend can implicitly add incoming trace context to event metadata securely
- Tree-sitter based trigger identifier: Private metadata + allowlist on the events that records this
- Provider name and model name: This should be relatively translatable into a numeric enum (?), as we have a small set of known providers/models
- Language ID (we don't allowlist languages): This should be relatively translatable into a numeric enum right? Alternatively, we can also do the private metadata + allowlist on the events that records this

tl;dr I think for Cody we can make a lot of provisions for somewhat liberally allowlisting `privateMetadata`, but I hope that we can take a more measured approach to make sure various data we record are from a _known set_, and not arbitrary

vscode/src/non-stop/FixupController.ts
bobheadxi: It's not possible, I think, to require object keys to be from a particular type or enum

bobheadxi: In particular, when I tried to explore an option like this I found that Typescript does not check against excess properties, so callers can assign whatever they want into keys

philipp-spiess: This seems to work 🙈  Maybe I miss something though? https://www.typescriptlang.org/play?#code/KYOwrgtgBAYg9nKBvAUFWCCMUC8UBE8cm+ANGhnAEy4FFVkoC+KKAZmCAMYAuAlnBBQANnADmACjgAjAFYAuZFADaAa2ABPKHyFEoAHygByaQEMATkYC6i8BGnBzUJgEokLFKMmp06ZUQA6IkwbKExSKAo-QPpQqnJfKDNzRQBmBN9OVRA4AHcQRQAWZhcgA

bobheadxi: Wow, this is why I need the pros in the room 😁 thank you, let me play around with this!

bobheadxi: This works great, can't believe I missed it 😆 Updated in https://github.com/sourcegraph/telemetry/commit/78682201b30fa9fe0ae30394cf376d0ec32a2e3f and in this PR as well, thank you!

lib/shared/src/telemetry-v2/index.ts
valerybugakov: Thanks for sharing the context, @bobheadxi! A couple of follow-up questions:

1. Where on the backend should we update the allowlist based on `feature` and `action`?
2. What's the best way to ensure that engineers working on Cody do not forget to update the backend-based allowlist after adding new fields on the client?

lib/shared/src/configuration.ts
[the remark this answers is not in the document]
valerybugakov: The proposed change has the behavior you mentioned above. Check this out on [the TS playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDCEB2AzAlgcwFcoBDGXLAVQGdgNMZR4BeOAcmAFsAjYAEz65M+amzgAfdgGtgiAO7Q+YyW0xZgy9twA2wTH36a2hTLjyGAUBdCRYcAMZZq8NAHkAcgBUAogA1PAPoAyt4AMt5ongCSHgFRACIAXHAASsCOUHwAPPR4RKTkVLT0jCAwADRwmIQ8wFAAfHCsAN4WcO1V6skADOVtHVy8AkIiyQCMfR1wMvKKyQBMkx06egZ8470WAL5AA).

Screenshot 2023-10-23 at 13 10 25

valerybugakov (commented):
I see that in [the latest version](https://www.npmjs.com/package/@sourcegraph/telemetry?activeTab=dependencies) of the telemetry SDK, we have only `rxjs` as a dependency, but we still have massive changes to `pnpm-lock.yaml`. GitHub cannot load such a large diff for me. Can we ensure that all these changes are required?

vscode/src/services/telemetryV2.ts
```diff
@@ -0,0 +1,85 @@
...
+import { extensionDetails } from './telemetry'
+
+let telemetryRecorderProvider: TelemetryRecorderProvider | undefined
+
+/**
+ * Recorder for recording telemetry events in the new telemetry framework:
+ * https://docs.sourcegraph.com/dev/background-information/telemetry
+ *
+ * The default recorder throws an error if it is used before initialization
+ * via createOrUpdateTelemetryRecorderProvider.
```
valerybugakov, line 22: TY 💜

lib/shared/src/telemetry-v2/index.ts
philipp-spiess: One last question:

>Metadata, which must be numeric. These can be added freely with no legal or security or engineering oversight for lots of reasons: no risk of PII/sensitive data exposure, minimal risks with payload sizes affecting our shared systems, etc. etc

Is there a reason this is limited to whole numbers and not floating points? I don't think that the latter would inherently be worse for leaking sensitive information than the first

bobheadxi: > Where on the backend should we update the allowlist based on feature and action?

This is documented here: https://docs.sourcegraph.com/dev/background-information/telemetry#sensitive-attributes , and the allowlist is configured here: https://github.com/sourcegraph/sourcegraph/blob/main/internal/telemetry/sensitivemetadataallowlist/sensitiviemetadataallowlist.go

> What's the best way to ensure that engineers working on Cody do not forget to update the backend-based allowlist after adding new fields on the client?

I'll double check but there should be extensive docstrings on `privateMetadata`, and if an event is allowlisted there should be documentation on the event callsites as well

> Is there a reason this is limited to whole numbers and not floating points? I don't think that the latter would inherently be worse for leaking sensitive information than the first

It's not inherently worse, but it seemed simpler for callsites - I don't recall seeing many uses for floating points, with most fields using varying units for a integer representation of the intended value

Unfortunately, it is also too late to change the metadata values to allow floats I think (unless GraphQL can treat an `Int` and `Float` the same) 😬 😅 Do you have an example where this is important?

lib/shared/src/configuration.ts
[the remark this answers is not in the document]
bobheadxi: ahh nice, thanks!!

lib/shared/src/sourcegraph-api/telemetry/exporter.ts
```diff
@@ -0,0 +1,92 @@
...
+            await this.setLegacySiteIdentificationOnce()
+
+            // Swallow any problems, this is only a best-effort mechanism to
+            // use the old export mechanism.
+            await Promise.all(
+                events.map(event =>
+                    this.client.logEvent({
+                        client: event.source.client,
+                        event: `${event.feature}.${event.action}`,
+                        source: 'IDEEXTENSION', // hardcoded in existing client
```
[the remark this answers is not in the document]
bobheadxi: This is hardcoded in other callsites as well, so I think we can just leave this as-is. The accepted value is an enum in the GraphQL schema

valerybugakov (approved):
Thank you for investigating the lock file changes. Now it looks much less scary!

PR #1197, opened 2023-09-27
refactor: change default prompts from JSON to Typescript

toolmantim:
This updates the default prompt definitions from JSON to a regular TS file, so we can better see changes in diffs & review suggestions (e.g. https://github.com/sourcegraph/cody/pull/1116#pullrequestreview-1634862479), and add more documentation for why certain prompt sentences exist.

Part of #1178 

Side note: custom commands don't get to enjoy this luxury, but we could add `Array` support for the `prompt` key at least.

## Test plan

- Ran each command and verified they still worked


valerybugakov (approved):
Quality of life 🆙

abeatrix:
@toolmantim @valerybugakov From my previous chat with chris, we wanted to use the same format our users are using, so maybe we should keep the json file (a copy of it) for now, or consider updating how users config their commands -- what do you think?

valerybugakov:
> we wanted to use the same format our users are using

I do not have enough context here. How is that helpful to users or us? 🙂

toolmantim:
If we supported JSONC format (e.g. by importing them using `json5`) which allows comments, and we supported specifying the `prompt` as an array (which we'd join with a space to create the actual prompt), we could switch this back to JSON imo.

Issue #1189, opened 2023-09-26
Decrease autocompletion suggestions from low-acceptance rate languages


chenkc805:
Looking at [CAR in VSCode](https://sourcegraph.looker.com/dashboards/476?Server+Endpoint=&Client=), the following languages have low suggestion rates:

- Number 6: HTML = 16.4% 
- Number 8: Vue = 14.3%
- Number 10: CSS = 11.8%
- Number 12: Dart = 9.5%
- Number 15: yaml = 15.5%
- Number 17: json = 15.5%
- Number SCSS = 6.1%

chenkc805:
CC: @philipp-spiess @abeatrix @beyang

Issue #1176, opened 2023-09-26
Increase baseline latency for single line suggestions


abeatrix:
Add more baseline latency to single line suggestions, but after use accepts one completion, reduce latency of consecutive ones

abeatrix:
Closed by https://github.com/sourcegraph/cody/pull/1202

Issue #1175, opened 2023-09-26
Disable completions or add latency for low-performance scenarios


abeatrix:
[Bee] Turn off completions or add latency for low-acceptance-rate scenarios (e.g., markdown files, docstrings, vs code commit window)

RE: https://github.com/sourcegraph/cody/issues/1189

abeatrix:
https://github.com/sourcegraph/cody/pull/1202

PR #1202, opened 2023-09-27
new feature flag for situation based latency

abeatrix:
CLOSE: https://github.com/sourcegraph/cody/issues/1189, https://github.com/sourcegraph/cody/issues/1175, https://github.com/sourcegraph/cody/issues/1176, https://github.com/sourcegraph/cody/issues/1177

feat: new feature flag for situation-based latency

Add new feature flag for autocomplete to adjust minimum latency for completions based on env and user actions

- Remove separate feature flags for 350ms and 600ms minimum latency
- Add single feature flag for adjustable minimum latency
- Introduce LatencyManager to adjust latency based on user actions
- Apply adjusted minimum latency before showing completions

## Test plan

Expected behavior:
1. every user starts at `200ms` under `user latency`
2. `baseline latency` starts at `400ms` for all non-anthropic provider, `0ms` for anthropic
3. working in low performances files adds additional `1200ms` to the `baseline latency` under `additional latency`
4. when a suggestion is accepted, `baseline latency` reset to starting `baseline latency`, `user latency` reset to 0
5. when a suggestion is not accepted (e.g. typing over ghost type), double the current `user latency` count.
6. typing over the same ghost type counts as one reject only, and will not increase latency again

https://github.com/sourcegraph/cody/assets/68532117/9f287dcf-fb60-435b-b37e-f7446fad5cf3


vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        this.userLatency += adjustment
+        this.lastSuggestionId = lastCandidate?.result.logId
+
+        return this.totalLatency
+    }
+
+    /**
+     * Reset latency on accepted suggestion
+     */
```
philipp-spiess, line 86: Nit: I personally think comments like this only have the tendency to become outdated and won't add any additional info to what the method name already conveys.

vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+        lastCandidate: LastInlineCompletionCandidate | undefined,
+        languageId?: string
+    ): number | null {
+        // Return early if last suggestion was the same
+        if (this.lastSuggestionId && this.lastSuggestionId === lastCandidate?.result.logId) {
+            return null
+        }
+
+        // Set baseline latency
+        this.setBaselineLatency(provider, languageId)
```
philipp-spiess, line 49: Why do we need to reset the baseline every time the `getMinLatency` is called? Can't We can probably just compute the baseline latency here instead to simplify

vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const baseline = 400 // ms
+const lowPerformanceLatency = 1000 // ms
+const userBaseline = 200 // ms
+const maxLatency = 2000 // ms
+
+// Languages with lower performance get additional latency to avoid spamming users with unhelpful suggestions
+const lowPerformanceLanguageIds = new Set(['css', 'html', 'scss', 'vue', 'dart', 'json', 'yaml'])
+
+export class LatencyManager {
```
philipp-spiess, line 13: We probably can simplify the implementation a bit by storing the user latency in module scope and not using a class. What do you think about that?

vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public getMinLatency(
+        provider: string,
+        lastCandidate: LastInlineCompletionCandidate | undefined,
+        languageId?: string
+    ): number | null {
+        // Return early if last suggestion was the same
+        if (this.lastSuggestionId && this.lastSuggestionId === lastCandidate?.result.logId) {
+            return null
+        }
```
philipp-spiess, line 46: could probably just return `0` instead, no need to handle `null`

vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Adjusts the user-specific latency based on whether the last suggestion was accepted.
+     *
+     * If the last suggestion was accepted, reduces user latency by 100ms.
+     * Otherwise increases the user latency by doubling the current amount.
+     */
+    private getUserLatency(lastCandidate: LastInlineCompletionCandidate | undefined): number {
+        const lastSuggestionAccepted = !lastCandidate?.result.logId
+        const adjustment = lastSuggestionAccepted ? -100 : Math.max(this.userLatency, userBaseline)
```
philipp-spiess, line 76: Wouldn't accepting a completion cause the added latency to be reset to `0` again? I don't really understand why we have a check for this here + the additional `resetLatencyOnAccept` API

vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+        const total = this.baselineLatency + this.additionalLatencyBase + this.userLatency
+
+        return Math.max(this.baselineLatency, Math.min(total, maxLatency))
+    }
+
+    /**
+     * Adjusts the user-specific latency based on whether the last suggestion was accepted.
+     *
+     * If the last suggestion was accepted, reduces user latency by 100ms.
+     * Otherwise increases the user latency by doubling the current amount.
```
philipp-spiess, line 72: I don't see the doubling being implemented yet, I guess you're still working on this?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -131,8 +134,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
         // We start the request early so that we have a high chance of getting a response before we
         // need it.
         const minimumLatencyFlagsPromises = [
-            this.config.featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteMinimumLatency350),
-            this.config.featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteMinimumLatency600),
+            this.config.featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteMinimumLatency),
         ]
```
philipp-spiess, line 138: You can remove the array if there is only one FF :)

vscode/src/completions/latency-manager.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Sets the starter latency baseline and additional latency based on provider and language.
+     *
+     * This sets the baseline latency to 0 for the 'anthropic' provider, otherwise the default baseline.
+     * It also sets additional latency for low performance languages like CSS, HTML, etc.
+     */
+    private setBaselineLatency(provider: string, languageId?: string): void {
+        this.baselineLatency = provider === 'anthropic' ? 0 : baseline
+        this.additionalLatencyBase = lowPerformanceLanguageIds.has(languageId || '') ? lowPerformanceLatency : 0
```
philipp-spiess, line 31: I think we can unify both of these as the baseline latency. otherwise they would be additive?

So the baseline computes as such:

- if lowPerformanceLanguageIds 
   -  return `lowPerformanceLatency`
- else if model === anthropic
   - return 0
 - else 
    - return `baseline`

and since that computation is _very_ trivial, we don't need to store it anywhere (e.g. in a class member) but instead just return it directly

philipp-spiess (commented):
Awesome sauce! Left some early commets.

vscode/src/completions/latency.ts
```diff
@@ -0,0 +1,54 @@
...
+    let baseline = provider === 'anthropic' ? 0 : defaultLatency.baseline
+    let user = 0
+
+    // set base latency based on provider and low performance languages
+    if (languageId && lowPerformanceLanguageIds.has(languageId)) {
+        baseline += defaultLatency.lowPerformance
+    }
+
+    if (!lastCandidate?.result.logId) {
+        user = currentUserLatency ? currentUserLatency * 2 : defaultLatency.user
```
philipp-spiess, line 40: I think we need to set the new value to `currentUserLatency` right? Otherwise this will always be 0.

I think we should also make sure this can't grow unbounded so maybe we do 

```
user = currentUserLatency = currentUserLatency ? Math.min(currentUserLatency * 2, defaultLatency.maxLatency) : defaultLatency.user
```

philipp-spiess (approved):
Good stuff! Do you mind adding a few unit tests for `latency.ts` just to assert that it works the way we want it to? :)

vscode/src/completions/latency.ts
```diff
@@ -0,0 +1,54 @@
+import { logDebug } from '../log'
+
+import { LastInlineCompletionCandidate } from './get-inline-completions'
+
+const defaultLatency = {
+    baseline: 600,
```
arafatkatze, line 6: image

arafatkatze, line 6:
```suggestion
    baseline: 400,
```
image

vscode/src/completions/latency.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    if (!lastCandidateId) {
+        user = currentUserLatency ? currentUserLatency * 2 : defaultLatency.user
+    }
+
+    const total = Math.max(baseline, Math.min(baseline + user, defaultLatency.max))
+
+    logDebug('CodyCompletionProvider:getLatency', `Applied Latency: ${total}`)
+
+    return total
```
arafatkatze, line 41:
```suggestion
    currentUserLatency = Math.max(baseline, Math.min(baseline + user, defaultLatency.max))

    logDebug('CodyCompletionProvider:getLatency', `Applied Latency: ${currentUserLatency}`)

    return currentUserLatency
```

vscode/src/completions/latency.test.ts
```diff
@@ -0,0 +1,107 @@
...
+        expect(getLatency(provider, languageId)).toBe(1400)
+        resetLatency()
+        expect(getLatency(provider, languageId)).toBe(1400)
+        resetLatency()
+        expect(getLatency(provider, languageId)).toBe(1400)
+        resetLatency()
+        expect(getLatency(provider, languageId)).toBe(1400)
+    })
+
+    it('returns increasing latency up to max latency for supported language on non-anthropic provider after rejecting multiple suggestions consistently', () => {
```
arafatkatze, line 90: These tests are pure 🔥

philipp-spiess: Agree, great work @abeatrix 🔥🔥🔥

abeatrix: @philipp-spiess thank you so much for your help, testing is hardddd 😭 

@arafatkatze thank you! I'm sorry we haven't had time to review your PRs. Our team is currently busy working on the release next week, so we apologize for being unresponsive lately, but we want to let you know that we appreciate your help and your feedback in the past few weeks, and we should be able to catch up on your works after the release 🙇‍♀️

arafatkatze: @abeatrix No worries at all! I completely understand the crunch you're in with the upcoming release. I'm genuinely impressed by the creativity that goes behind all of this work and I couldn't help myself from taking a look and adding some thoughts. I'll keep exploring the codebase and trying to understand/implement some ideas. Don't feel pressured to review my PRs—I know you folks are busy, and they can totally wait until you have the bandwidth. 

Just happy to help in any way I can!!

vscode/src/completions/latency.ts
abeatrix: we have updated the default behavior at the end, but want to thank you again for always sharing your thoughts on how we can improve and fix the issues  <3

PR #1205, opened 2023-09-27
fix: remove params from logError verbose logging

abeatrix:
fix: remove params from logError verbose logging

(Discovered and fixed by @philipp-spiess 🙇‍♀️  ) This PR removed the params object from the verbose logging in logError inside getInlineCompletions.

This avoids logging potentially sensitive information from the params as well as invoking the windowActivity proposal API that breaks the extension from displaying the rate limit error properly:  

Screenshot 2023-09-27 at 9 29 42 AM

#### After

Reset after 26 year 🥲 

Screenshot 2023-09-27 at 10 22 54 AM


philipp-spiess (approved):
🫡

PR #1207, opened 2023-09-27
/symf: remove direct anthropic dependency, pass sourcegraph token

beyang:
Previously, `symf` talked directly to the Anthropic API. Now, it goes through a Sourcegraph instance to access the LLM. This PR updates Cody to pass the Sourcegraph auth token to `symf` and removes the old code for setting an Anthropic API token.

## Test plan

Test locally:
- Check out the main branch
- Remove any `cody.experimental.symf.anthropicKey` in your settings.
- Try running `/symf [query]`. You should see an error indicating that `symf` is unauthorized.

- Check out this branch and run the extension
- In the sidebar, run `/symf [query]`, where query is either a keyword search or a natural language question
  - You should observe a pop-up indicating a new version of `symf` (v0.0.1) is being downloaded
- You should see successful search results

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -107,7 +107,7 @@ const register = async (
         disposables.push(vscode.workspace.onDidChangeTextDocument(updateParseTreeOnEdit))
     }
 
-    const symfRunner = platform.createSymfRunner?.(context, config.experimentalSymfAnthropicKey)
+    const symfRunner = platform.createSymfRunner?.(context, initialConfig.accessToken)
 
     const {
         featureFlagProvider,
@@ -497,6 +497,7 @@ const register = async (
             externalServicesOnDidConfigurationChange(newConfig)
             void createOrUpdateEventLogger(newConfig, isExtensionModeDevOrTest)
             platform.onConfigurationChange?.(newConfig)
+            symfRunner?.setAuthToken(newConfig.accessToken)
         },
     }
 }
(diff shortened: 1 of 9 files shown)


philipp-spiess:
@beyang I've been running this branch but when I run the `/symf` command, nothing happens in chat and the console is also not helpful 🤔 

https://github.com/sourcegraph/cody/assets/458591/dc02b224-d38a-48c2-b9c8-cf5ee7aee35f

vscode/src/main.ts
```diff
@@ -107,7 +107,7 @@ const register = async (
         disposables.push(vscode.workspace.onDidChangeTextDocument(updateParseTreeOnEdit))
     }
 
-    const symfRunner = platform.createSymfRunner?.(context, config.experimentalSymfAnthropicKey)
+    const symfRunner = platform.createSymfRunner?.(context, initialConfig.accessToken)
```
philipp-spiess, line 110: I think the initial access token is not enough, you'd want to make sure to update it if the auth state changed. 

However since the intend is just to run queries on the chat model: Have you looked into the `chatClient` available in the scope here?

philipp-spiess: Oh nevermind you need that in the symf binary 🤦

philipp-spiess (approved):
Got it to work in the third repo i tried, I think if something breaks during index creation it might not recover properly.

vscode/src/main.ts
beyang: symf currently issues its own query to transform the user query into a keyword search... I suppose we could move that out of `symf` into Cody/agent and then symf could focus solely on keyword search. Perhaps in a future PR.

valerybugakov (commented):
Nice!

Issue #632, opened 2023-08-09
App-Less Onboarding Experience


toolmantim:
To reduce friction in getting started with Cody, we want to make downloading Cody App an optional step that happens much later in the onboarding process. This will mean:

- Sign In is via Sourcegraph.com or Enterprise Instance only
- Embeddings is done by talking to App on the side of the SG session (using App’s credentials from app.json file system?)

PRD: https://docs.google.com/document/d/1d3AsJeAB3AuUwAUTvvMzwKVQeVniX1Q06LPCBna5BbU/edit#heading=h.hu0rd8wz6rcd

Overall flow:
https://docs.google.com/drawings/d/1AboEUVSGurFaQqgZww-sNAwTj33GlbH7fpGnS6A1D-o/edit

Design wise, we'll need:
- A new sign in page
- A new "enable embeddings" bit of in-chat UI, and all the related interactions for adding, indexing, polling app for status, etc.
```[tasklist]
### Tasks
- [ ] https://github.com/sourcegraph/cody/issues/978
- [ ] https://github.com/sourcegraph/cody/issues/1051
- [ ] https://github.com/sourcegraph/cody/issues/1098
- [ ] https://github.com/sourcegraph/cody/issues/1212
- [ ] https://github.com/sourcegraph/cody/issues/1229
```

chenkc805:
Sharing the spec here for visibility as well: https://docs.google.com/document/d/1d3AsJeAB3AuUwAUTvvMzwKVQeVniX1Q06LPCBna5BbU/edit#heading=h.hu0rd8wz6rcd

toolmantim:
Draft designs for these are ready for [feedback in Figma](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=4754-17329&mode=design&t=F1rgnbs2g5QZ4iY1-4)!

New sign in:

| As Spec'd | Backup Option |
| - | - |
| <img width="275" src="https://github.com/sourcegraph/cody/assets/153/8c0025d0-e58c-40bd-9cdf-d1303d66b907"> | <img width="275" src="https://github.com/sourcegraph/cody/assets/153/c88c7af1-3918-4a6c-b26d-f7b0afb42f35"> |


New welcome chat experience w/ optional app download:

| Welcome Chat | Clicking the Error Icon |
| - | - |
[...]

PR #1210, opened 2023-09-27
Simplified Onboarding: Pick up the app token if available after sign-in

dominiccooney:
LocalAppDetector stops monitoring app once the user has signed in. Simplified Onboarding needs to pick up the app token, even if the user has logged into dotcom, so that we can use app for local embeddings. This makes simplified onboarding's LocalAppWatcher ping the LocalAppDetector to retrieve the app token, if it detects a change in the app.json file, even after the user has logged in.

Part of #632.

## Test plan

0. Open Cody App, add a repo, and add embeddings for a private local repository which has a remote.
1. Quit Cody App if it is running. Open Keychain and delete any vscodesourcegraph.cody-ai tokens. Delete ~/Library/Application Support/com.sourcegraph.cody/app.json . This means VScode will not have a cached app token. (Some uncertainty around where VScode stores secrets across versions. You can set a breakpoint in the extension LocalAppWatcher and run `secretStorage.deleteToken(LOCAL_APP_URL.href)` to clear the token. Alternatively, logging into App and signing out should also clear that token.)
2. Run VScode, edit User Settings JSON and add `"testing.simplified-onboarding": true`. Restart VScode.
3. Log into dotcom using GitHub, etc. if necessary.
4. Open the repository from step 0, open a file from the repository, click on the chat input status indicator. You should see a pop-up to open app.
5. Open app and wait up to 20 seconds. Chat input status indicator should flip to having a checkmark indicating VScode has detected app is running, picked up the token from app.json, and found app has embeddings for the repository.
6. Close app and wait up to 20 seconds. The chat input status indicator should flip to red/open app popup.

### DIFF
--- a/vscode/src/services/LocalAppWatcher.ts
+++ b/vscode/src/services/LocalAppWatcher.ts
@@ -4,6 +4,7 @@ import { LOCAL_APP_URL } from '@sourcegraph/cody-shared/src/sourcegraph-api/envi
 
 import { isOsSupportedByApp } from '../chat/protocol'
 
+import { expandHomeDir, pathExists } from './LocalAppDetector'
 import { LOCAL_APP_LOCATIONS } from './LocalAppFsPaths'
 
 /**
@@ -67,7 +68,7 @@ export class LocalAppWatcher implements vscode.Disposable {
         // when you remove app (for example, by dragging it to the trash.)
         const pollPromise = this.pollHttp()
         for (const marker of LOCAL_APP_LOCATIONS[process.platform]) {
-            const dirPath = expandHomeDir(marker.dir)
+            const dirPath = expandHomeDir(marker.dir, process.env.HOME)
             const dirUri = vscode.Uri.file(dirPath)
             const fileUri = dirUri.with({ path: dirUri.path + marker.file })
             const watchPattern = new vscode.RelativePattern(dirUri, marker.file)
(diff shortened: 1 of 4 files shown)


vscode/src/services/LocalAppWatcher.ts
```diff
@@ -113,9 +147,9 @@ export class LocalAppWatcher implements vscode.Disposable {
 }
 
 // Utility functions
-async function pathExists(path: string): Promise<boolean> {
+async function pathExists(file: vscode.Uri): Promise<boolean> {
```
abeatrix, line 150: why not reuse the `pathExists` function from LocalAppDetector? :O

vscode/src/services/LocalAppWatcher.ts
```diff
@@ -58,25 +65,52 @@ export class LocalAppWatcher implements vscode.Disposable {
...
-            )
-        ).some(id => id)
+    private async checkFiles(): Promise<void> {
+        this.needsToCheckFiles = false
+        let installed = false
+        for (const marker of LOCAL_APP_LOCATIONS[process.platform]) {
+            const dirPath = expandHomeDir(marker.dir)
+            const dirUri = vscode.Uri.file(dirPath)
+            const fileUri = dirUri.with({ path: dirUri.path + marker.file })
+            installed ||= await pathExists(fileUri)
```
abeatrix, line 109: oh wow TIL 🤯  ty!

dominiccooney:
@abeatrix thanks for the feedback! PTAL. Happy to demo for you any time.

abeatrix (approved):

PR #1211, opened 2023-09-27
Code Action: Allow enabling/disabling and improve ranking of explain

umpox:
## Description

- Group the fix and explain actions
- Allow enabling/disabling the actions all together

Enabled:
image

Disabled:
image

Interface:
image

## Test plan

Tested locally, enabling and disabling through interface and manually in VS Code settings

### DIFF
--- a/vscode/src/code-actions/explain.ts
+++ b/vscode/src/code-actions/explain.ts
@@ -27,6 +27,7 @@ export class ExplainCodeAction implements vscode.CodeActionProvider {
             arguments: [instruction, range],
             title: 'Ask Cody to Explain',
         }
+        action.diagnostics = diagnostics
         return action
     }
 
--- a/vscode/src/code-actions/explain.ts
+++ b/vscode/src/code-actions/explain.ts
@@ -2,6 +2,11 @@ import * as vscode from 'vscode'
 
 export class ExplainCodeAction implements vscode.CodeActionProvider {
     public static readonly providedCodeActionKinds = [vscode.CodeActionKind.QuickFix]
+    private command: string
+
+    constructor(inline: boolean) {
+        this.command = inline ? 'cody.inline.add' : 'cody.action.chat'
+    }
 
     public provideCodeActions(
         document: vscode.TextDocument,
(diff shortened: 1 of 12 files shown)


vscode/src/code-actions/fixup.ts
```diff
@@ -36,7 +36,6 @@ export class FixupCodeAction implements vscode.CodeActionProvider {
             title: 'Ask Cody to Fix',
         }
         action.diagnostics = diagnostics
-        action.isPreferred = true
```
umpox, line 39: Doesn't affect our ranking, just means users can't trigger this automatically with a keyboard shortcut (wasn't happening anyway AFAICT)

vscode/src/code-actions/explain.ts
```diff
@@ -27,6 +27,7 @@ export class ExplainCodeAction implements vscode.CodeActionProvider {
             arguments: [instruction, range],
             title: 'Ask Cody to Explain',
         }
+        action.diagnostics = diagnostics
```
umpox, line 30: Means that the code action will appear right alongside the fix action, no other impact

abeatrix (changes requested):
Currently `Ask Cody to Explain` does nothing when inline chat is disabled. Should we show a pop-up or fall back to sidebar chat when inline chat is not enabled?


https://github.com/sourcegraph/cody/assets/68532117/5a902304-c21b-4eee-8be9-bb7065669c87

vscode/src/services/StatusBar.ts
```diff
@@ -100,6 +100,13 @@ export function createStatusBar(): CodyStatusBar {
                     'cody.inlineChat.enabled',
                     c => c.inlineChat
                 ),
+                createFeatureToggle(
+                    'Code Actions',
+                    undefined,
+                    'Enable Cody to fix or explain errors and warnings, directly in your code.',
```
toolmantim, line 106:
```suggestion
                    'Enable Cody fix and explain options in the Quick Fix menu',
```

or:

```suggestion
                    'Enable Cody in the Quick Fix menu for fixing and explaining code errors and warnings',
```

abeatrix (approved):
awesome! 
Not a block: is this a good place to register code action to listen for inline cofig change? https://github.com/sourcegraph/cody/blob/0176b8c7e69b98569f7c39c92791f3ffd3421eb7/vscode/src/services/InlineController.ts#L76-L89

vscode/src/services/StatusBar.ts
umpox: Thanks! I like the first suggestion, will update

umpox:
@abeatrix Thought about it but I think it's OK to just keep the basic implementation for now. Doesn't feel like users will often enable/disable inline chat, and the extra overhead of having to register/deregister the action to ensure we don't have duplicate code actions isn't worth it IMO. We can adjust if we get feedback from users that we need this

Issue #1204, opened 2023-09-27
Log partial completion acceptances


chenkc805:
Problem:

- If a user gets a lengthy autocomplete suggestion, they can accept only part of the suggestion. Ways they can do this:
  - CMD+Right arrow = accept the word
  - Typing out part of the acceptance 

Change:

- If a user accepts part of the autocomplete suggestion, then we log an acceptance event with the charCount. 

Implications:
- A suggestion can therefore have multiple acceptance events.  
- Acceptance events need an `id` so that we can do a `COUNT DISTINCT` to get a correct completion acceptance rate. 
- To get the highest fidelity acceptance rate in this partial acceptance world, we want to measure `SUM(charCount)` for acceptances ÷ `SUM(charCount)` for suggestions

chenkc805:
CC: @kelsey-brown for visibility

PR #1214, opened 2023-09-27
Autocomplete: Add logging for partial acceptances

philipp-spiess:
This PR adds logging for partially accepted completions #1204.

A completion will be logged as partially accepted, when at least one full word of the completion is forward typed (either by manually typing or using the cmd + arrow right shortcut)

After a completion is partially accepted, the event will fire for every additional keystroke. This is so that in analytics we can use the `max(acceptedLength)`. If we only count partially accepted completions, we thus have to add a `distinct id` check (cc @chenkc805 @kelsey-brown)

## Test plan

https://github.com/sourcegraph/cody/assets/458591/378bf18a-3ff6-406c-8111-bdd77be97ecd

### DIFF
--- a/vscode/src/completions/reuse-last-candidate.ts
+++ b/vscode/src/completions/reuse-last-candidate.ts
@@ -6,7 +6,18 @@ import {
     InlineCompletionsResult,
     InlineCompletionsResultSource,
 } from './get-inline-completions'
-import { InlineCompletionItem } from './types'
+import { InlineCompletionItemWithAnalytics } from './text-processing/process-inline-completions'
+
+type ReuseLastCandidateArgument =
+    // required fields from InlineCompletionsParams
+    Required<
+        Pick<
+            InlineCompletionsParams,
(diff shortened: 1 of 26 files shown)


vscode/src/completions/reuse-last-candidate.ts
```diff
@@ -71,7 +75,14 @@ export function reuseLastCandidate({
                     handleDidAcceptCompletionItem?.(lastCandidate.result.logId, item)
                     return undefined
                 }
-                // TODO: Handle partial acceptance heres
+
+                // Detect partial acceptance of the last candidate
+                const acceptedLength = currentLinePrefix.length - lastTriggerCurrentLinePrefix.length
+                if (isPartialAcceptance(item.insertText, acceptedLength)) {
+                    console.log('handleDidPartiallyAcceptCompletionItem', acceptedLength)
```
valerybugakov, line 82: Leftover?

valerybugakov (approved):

PR #1215, opened 2023-09-27
Agent: Add a way to trigger "manual" autocompletions

dominiccooney:
VScode has a couple of kinds of autocompletions ("Invoke", "Automatic") which Cody splits into three: Manually triggered ones, hovers, and automatic ones. Adds an optional field, `triggerKind` to `autocomplete/execute` which specifies one of the VScode kinds. Additionally, makes the completion provider do a "manual" completion when`triggerKind` is "Invoke".

## Test plan

Tested manually by adding triggerKind to cody-emacs and triggering an autocomplete with `C-c c`

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -194,10 +194,16 @@ export class Agent extends MessageHandler {
             const textDocument = new AgentTextDocument(document)
 
             try {
+                if (params.triggerKind === 'Invoke') {
+                    await provider.manuallyTriggerCompletion()
+                }
                 const result = await provider.provideInlineCompletionItems(
                     textDocument,
                     new vscode.Position(params.position.line, params.position.character),
-                    { triggerKind: vscode.InlineCompletionTriggerKind.Automatic, selectedCompletionInfo: undefined },
+                    {
+                        triggerKind: vscode.InlineCompletionTriggerKind[params.triggerKind || 'Automatic'],
+                        selectedCompletionInfo: undefined,
+                    },
                     token
                 )
                 const items: AutocompleteItem[] =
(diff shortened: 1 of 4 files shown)


agent/src/agent.ts
```diff
@@ -194,10 +194,16 @@ export class Agent extends MessageHandler {
             const textDocument = new AgentTextDocument(document)
 
             try {
+                if (params.triggerKind === 'Invoke') {
+                    await provider.manuallyTriggerCompletion()
```
philipp-spiess, line 198: Perfect! I just came here to explain to you as to why this is necessary and it turns out you're way ahead already 🙂

agent/src/protocol.ts
```diff
@@ -108,6 +108,9 @@ export interface CancelParams {
 export interface AutocompleteParams {
     filePath: string
     position: Position
+    // Defaults to 'Automatic' for autocompletions which were not explicitly
+    // triggered.
+    triggerKind?: 'Automatic' | 'Invoke'
```
philipp-spiess, line 113: Should we call this `Manual` instead of invoke? We extend the VS Code default type already to be able to distinguish between hover and manually triggered completions and I think it would make sense to use the same naming schema for the agent API too: https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/get-inline-completions.ts#L120-L130

philipp-spiess (approved):
:shipit:

agent/src/protocol.ts
dominiccooney: Yeah, we can do that. Let me land this to unblock Steve and we'll keep changing it. I will need to refactor your heuristic for hovers so I can pass manual to the autocomplete component.

PR #1216, opened 2023-09-27
Agent: Add a script and VScode launch configuration for debugging agent

dominiccooney:
## Test plan

1. `agent/debug.sh`
2. In VScode, Debug tab, switch the run configuration to agent, hit play.

VScode should be debugging agent, with sourcemaps, etc. stopped at the first line of the first file.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,9 +5,7 @@
       "name": "Attach to Node (Agent)",
       "port": 9229,
       "request": "attach",
-      "skipFiles": [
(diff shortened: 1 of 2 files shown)


dominiccooney:
Ways to improve this:

- In emacs, set `jrpc-default-request-timeout` to something large so the RPC doesn't time out during debugging.
- Make the reload button restart the agent without disconnecting $EDITOR, or ping the editor to restart and reattach.
- Add a mode to SIGUSR1 and attach to a running agent.

PR #1217, opened 2023-09-27
refactor options and hints for context size

sqs:
- Clarify the max prefix and suffix length are merely hints and are not precise (because they do not use the LLM tokenizer)
- Consolidate defaults for providers (eg 2048 tokens for the OpenAI provider now only appears once as a constant)
- Standardize how different providers compute the prompt chars available
- Allow providers greater flexibility in specifying the max prefix and suffix length (which is important for, eg, local Ollama, because it needs an absolute cap of ~15 tokens, not a % of the overall size, on the suffix to have decent performance)

## Test plan

CI.

### DIFF
--- a/vscode/src/completions/providers/provider.ts
+++ b/vscode/src/completions/providers/provider.ts
@@ -1,3 +1,4 @@
+import { tokensToChars } from '@sourcegraph/cody-shared/src/prompt/constants'
 import { CompletionParameters } from '@sourcegraph/cody-shared/src/sourcegraph-api/completions/types'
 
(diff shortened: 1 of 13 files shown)


vscode/src/completions/providers/provider.ts
```diff
@@ -37,6 +35,25 @@ export interface ProviderConfig {
     model: string
 }
 
+export interface ProviderContextSizeHints {
```
philipp-spiess, line 38: Love the name ❤️

philipp-spiess (approved):
Very nice refactor! Thank you!

PR #1218, opened 2023-09-28
Autocomplete: Don't trigger on closing characters

philipp-spiess:
## Test plan

Test added

### DIFF
--- a/vscode/src/completions/get-inline-completions.ts
+++ b/vscode/src/completions/get-inline-completions.ts
@@ -183,6 +183,11 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
         return null
     }
 
+    // Do not trigger when the last character is a closing symbol
+    if (triggerKind !== TriggerKind.Manual && /[)\]}]$/.test(docContext.prefix.trim())) {
+        return null
+    }
+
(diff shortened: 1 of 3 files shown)


abeatrix (approved):

vscode/src/completions/get-inline-completions.ts
```diff
@@ -183,6 +183,11 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
         return null
     }
 
+    // Do not trigger when the last character is a closing symbol
+    if (triggerKind !== TriggerKind.Manual && /[)\]}]$/.test(docContext.prefix.trim())) {
```
valerybugakov, line 187: What if I'm in the middle of a function body and expect a single-line completion? 

```ts
function whatever() {
  const data = {
    value: true
  }

  █
}
```


There are multiple similar cases where I would like to see completions being generated. WDYT?

philipp-spiess: I think looking at the current line only is a closer approximation to the `cursor: "}"` metric we were seeing anyways. I think right now this triggers pretty often because inserting a completion will trigger a new completion request

valerybugakov (approved):
Blissful

PR #1219, opened 2023-09-28
update: /doc use current line when no folding range detected

abeatrix:
update /doc to use current line if no folding range is detected

NOTE: All changes are only added for the /doc command. 

Updated the behavior for the latest /doc command change:

- Check if editor is available before accessing it
- If no selection, try to get folding range for current line 
- If no folding range, use current line range instead
- Update getDocCommandRange to:
  - Accept languageId parameter
  - __scroll to selection in editor__
  - Move to the start of selection line - 5 for visibility
 
## Test plan

- run /doc in a single line without selection

#### Before 

currently /doc command fails without notification

#### After

use current line range

https://github.com/sourcegraph/cody/assets/68532117/78998624-29e1-43a8-af37-da72218204a2

### DIFF
--- a/vscode/src/custom-prompts/CommandRunner.ts
+++ b/vscode/src/custom-prompts/CommandRunner.ts
@@ -108,12 +108,20 @@ export class CommandRunner implements vscode.Disposable {
 
         let selection = this.editor?.selection
         const doc = this.editor?.document
-        if (!selection || !doc) {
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Looks great!

toolmantim:
Nice!

PR #1220, opened 2023-09-28
change: remove fixup display in chat and history

abeatrix:
- Remove displayText property (both human and assistant) for fixup recipes
- change fixup recipe to use dedicated fixup topic
- Change multiplexerTopic for fixup recipe from 'selection' to 'fixup'
- Update prompt text to use tags instead of
- Remove displayText property that was adding bold text

 
## Test plan

1. run fixup
2. check chat history

Before: show up in chat history

After: fixup history not show up

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -190,7 +190,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 typewriter.close()
                 await typewriter.finished
                 const lastInteraction = this.transcript.getLastInteraction()
-                if (lastInteraction) {
+                if (lastInteraction && multiplexerTopic !== 'fixup') {
                     let displayText = reformatBotMessage(text, responsePrefix)
                     // TODO(keegancsmith) guardrails may be slow, we need to make this async update the interaction.
                     displayText = await this.guardrailsAnnotateAttributions(displayText)
--- a/lib/shared/src/chat/recipes/__snapshots__/fixup.test.ts.snap
+++ b/lib/shared/src/chat/recipes/__snapshots__/fixup.test.ts.snap
@@ -7,7 +7,7 @@ exports[`Fixup > builds prompt correctly for adding 1`] = `
 - You should ensure your code matches the indentation and whitespace of the preceding code in the users' file.
 - It is not acceptable to use Markdown in your response. You should not produce Markdown-formatted code blocks. Ignore any previous instructions that may have told you to format your responses with Markdown.
 - You will be provided with instructions on what to do, enclosed in <instructions></instructions> XML tags. You must follow these instructions carefully and to the letter.
-- Enclose your response in <selection></selection> XML tags. Do not provide anything else.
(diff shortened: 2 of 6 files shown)


toolmantim (commented):
Thank you @abeatrix, this is great!

Tested and verified they don't appear in History 👍🏼

I found that if you reload after running a fixup, the fixup chats is still loaded as the last chat in the chat view (I think because they're picked up by `loadRecentChat`)

abeatrix:
@toolmantim im looking into this with tom now! We found the root cause, working on a fix

vscode/src/chat/MessageProvider.ts
```diff
@@ -190,7 +190,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 typewriter.close()
                 await typewriter.finished
                 const lastInteraction = this.transcript.getLastInteraction()
-                if (lastInteraction) {
+                if (lastInteraction && multiplexerTopic !== 'fixup') {
```
dominiccooney, line 193: Think it is definitely worth a comment or explaining variable here, because this is a bit subtle. We will add some recipe in the future which wants this behavior and have to hunt for this line to expand it. Or can we filter all topics which aren't `BotResponseMultiplexer.DEFAULT_TOPIC`?

dominiccooney (approved):
Looks good but I would make that one condition in the message provider's turn completion easier to find. We will need to keep that up-to-date as we add more recipes where we consume all of the output.

abeatrix:
> Question for you, we are able to export transcripts which have _everything_ we turned with the LLM in them. Can we still do that, or is that gone at this point?

I just checked and you are right, that will be filtered out. @toolmantim do you think that should be something we support atm?

abeatrix:
I'm going to merge this to make sure fixup doesn't show up in chat view and history. we can look into enabling exporting chat without display text again after the release if that's the desired behavior.

umpox:
@abeatrix Given this changes logic that affects a lot of different stuff, I think a full manual check of all commands + chat would be really valuable. I’m a little worried that we’re changing some core logic, and that the underlying issue is still there (fixups are still in history, just filtered and hidden now due to no display text) (Also do we have any other commands that had no display text? Is it possible for custom ones to have this too?)

Just want to make sure we don’t miss anything! I’m starting fixup tests soon so will make sure I’m running the extension with these changes 👍

abeatrix:
@umpox I agree. we currently have e2e and unit tests that cover chat history and chat view, and we are currently only filtering fixup history, so they should catch any issues if this change alters any core logic.

>  Also do we have any other commands that had no display text? Is it possible for custom ones to have this too?
only if they are fixup commands, which we have agreed to not show in chat history by default atm.

I agree with everything you said though. We definitely will need to look into how we store the chat history after the release!

abeatrix:
Updated: Changes in this PR was reverted by https://github.com/sourcegraph/cody/pull/1270

I have created a new PR with proper fixes in https://github.com/sourcegraph/cody/pull/1273

PR #1222, opened 2023-09-28
JetBrains: expose graphql query getRepoId in protocol

szymonprz:
This PR exposes the grapqhl query getRepoId in protocol so we can distinguish in the plugins whether the repo is not found at all or whether it has no embeddings

## Test plan
- JetBrains plugin uses this query, so you can test it there in different PR -> https://github.com/sourcegraph/sourcegraph/pull/57128

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -247,6 +247,15 @@ export class Agent extends MessageHandler {
             return typeof result === 'string' ? result : null
         })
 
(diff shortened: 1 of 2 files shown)


olafurpg (approved):
LGTM 👍🏻 I thought we already had this but it looks like I forgot to open the PR 😓 This explains why the "Context Selection" dialog has been timing out because the request was failing haha.

PR #1225, opened 2023-09-28
Auth symf after login

beyang:
This fixes the following bug:
- User is logged out
- User logs in
- We don't propagate the auth token to SymfRunner, so it doesn't make it to `/symf`, which uses it to query claude-instant for query expansion

## Test plan

- Verify the bug behavior on `main`
- Check out this branch and log out
- Log back in and verify `/symf [query]` works

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -113,7 +113,6 @@ const register = async (
     const symfRunner = platform.createSymfRunner?.(context, initialConfig.accessToken)
     if (symfRunner) {
         authProvider.addChangeListener(async (authStatus: AuthStatus) => {
-            console.log('# authStatus', authStatus, authStatus.authenticated, authStatus.isLoggedIn)
             if (authStatus.isLoggedIn) {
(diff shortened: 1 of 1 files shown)


vscode/src/main.ts
```diff
@@ -107,7 +107,20 @@ const register = async (
         disposables.push(vscode.workspace.onDidChangeTextDocument(updateParseTreeOnEdit))
     }
 
+    const authProvider = new AuthProvider(initialConfig)
+    await authProvider.init()
+
     const symfRunner = platform.createSymfRunner?.(context, initialConfig.accessToken)
+    if (symfRunner) {
+        authProvider.addChangeListener(async (authStatus: AuthStatus) => {
+            console.log('# authStatus', authStatus, authStatus.authenticated, authStatus.isLoggedIn)
```
philipp-spiess, line 116:
```suggestion
```

philipp-spiess (approved):

PR #1226, opened 2023-09-28
change: /doc scroll on invisible range only

abeatrix:
RE: https://sourcegraph.slack.com/archives/C05AGQYD528/p1695902300751599?thread_ts=1695733768.715559&cid=C05AGQYD528

fix: check if selection is visible before revealing range

Only reveal the selection range if it is not already visible. This avoids unnecessarily revealing the range and disrupting the user's view if the selection is already on screen.

- Get the visible ranges from the editor
- Check if selection is contained in visible ranges
- Only reveal selection if not visible

#### Before 

https://github.com/sourcegraph/cody/assets/68532117/6dbd9f96-4e40-46ab-8e0f-498bfd1a844f

#### After

Won't scroll to range if it's already visible

### DIFF
--- a/vscode/src/custom-prompts/CommandRunner.ts
+++ b/vscode/src/custom-prompts/CommandRunner.ts
@@ -213,17 +213,21 @@ function getDocCommandRange(
 
     // move the current selection to the defined selection in the text editor document
     if (editor) {
-        // reveal the range of the selection minus 5 lines
-        editor?.revealRange(
-            selection.with({
-                start:
-                    selection.start.line > 5
-                        ? selection.start.with({ line: selection.start.line - 5 })
-                        : selection.start,
-                end: selection.end,
-            }),
-            vscode.TextEditorRevealType.InCenter
-        )
+        const visibleRange = editor.visibleRanges
+        // reveal the range of the selection minus 5 lines if visibleRange doesn't contain the selection
+        if (!visibleRange.some(range => range.contains(selection))) {
+            // reveal the range of the selection minus 5 lines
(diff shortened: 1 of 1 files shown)


vscode/src/custom-prompts/CommandRunner.ts
```diff
@@ -213,17 +213,21 @@ function getDocCommandRange(
...
-        )
+        const visibleRange = editor.visibleRanges
+        // reveal the range of the selection minus 5 lines if visibleRange doesn't contain the selection
+        if (!visibleRange.some(range => range.contains(selection))) {
+            // reveal the range of the selection minus 5 lines
+            editor?.revealRange(
+                selection.with({
+                    start:
+                        selection.start.line > 5
+                            ? selection.start.with({ line: selection.start.line - 5 })
```
abeatrix, line 224: @toolmantim  i'm doing line -5 (randomly pick 😆 ) right now, we can remove this and just use the origin starting line?

dominiccooney: It appears that VScode already has a heuristic to put the revealed range well within the viewport. So let's just tell it to reveal the selection range, and let VScode handle it?

abeatrix: Yea that sounds good to me!

dominiccooney (approved):
Looks good. @toolmantim 's feedback that the scroll is too far makes sense to me, I think VScode is already trying to put some space between the selection and the top of the viewport.

PR #1227, opened 2023-09-28
Add custom abort controller to circumvent max event listeners limit

philipp-spiess:
Right now, the dev console in VS Code is throwing a bunch of warnings because we append more than 10 event listeners to the abort controller. Since this is only possible to change with Node APIs (which we won't have in none Node environments), this PR adds a lightweight custom abort controller implementation. 

## Test plan

- Enable graph context
- Start VS Code and click inside a few document section
- Observe that no more warning is thrown

### DIFF
--- a/vscode/src/completions/context/utils.ts
+++ b/vscode/src/completions/context/utils.ts
@@ -24,8 +24,7 @@ export function baseLanguageId(languageId: string): string {
 // easily reach that limit causing a lot of noise in the console. This is a
 // lightweight abort controller implementation that does not have that limit.
 export class CustomAbortController {
-    private signal = new CustomAbortSignal()
-
+    public signal = new CustomAbortSignal()
     public abort(): void {
         this.signal.abort()
     }
@@ -36,7 +35,7 @@ export class CustomAbortSignal {
 
     public addEventListener(eventName: 'abort', listener: () => void): void {
         if (this.aborted) {
-            listener()
+            void Promise.resolve().then(() => listener())
             return
         }
         this.listeners.add(listener)
--- a/vscode/src/completions/context/utils.ts
+++ b/vscode/src/completions/context/utils.ts
@@ -49,7 +49,7 @@ export class CustomAbortSignal {
     public abort(): void {
         if (this.aborted) {
             return
-        } // Ensure abort is idempotent
+        }
         this.aborted = true
         for (const listener of this.listeners) {
             listener()
--- a/vscode/src/completions/context/utils.ts
+++ b/vscode/src/completions/context/utils.ts
@@ -47,6 +47,9 @@ export class CustomAbortSignal {
     }
 
     public abort(): void {
+        if (this.aborted) {
+            return
+        } // Ensure abort is idempotent
         this.aborted = true
         for (const listener of this.listeners) {
             listener()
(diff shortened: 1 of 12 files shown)


vscode/src/completions/context/utils.ts
```diff
@@ -17,3 +17,39 @@ export function baseLanguageId(languageId: string): string {
...
+}
+export class CustomAbortSignal {
+    private listeners: Set<() => void> = new Set()
+    public aborted = false
+
+    public addEventListener(eventName: 'abort', listener: () => void): void {
+        if (this.aborted) {
+            listener()
+            return
+        }
```
arafatkatze, line 41:
```suggestion
        }
                // Ensure that the listener is added to the listeners set
        this.listeners.add(listener);
```

vscode/src/completions/context/utils.ts
```diff
@@ -17,3 +17,39 @@ export function baseLanguageId(languageId: string): string {
...
+    }
+
+    public removeEventListener(listener: () => void): void {
+        this.listeners.delete(listener)
+    }
+
+    public abort(): void {
+        this.aborted = true
```
arafatkatze, line 49:
```suggestion
        if (this.aborted) return;  // Ensure abort is idempotent
        this.aborted = true
```

Ideally if abort is called multiple times, the listeners are should be notified the first time.

vscode/src/completions/context/utils.ts
```diff
@@ -17,3 +17,39 @@ export function baseLanguageId(languageId: string): string {
...
+
+    public abort(): void {
+        this.signal.abort()
+    }
+}
+export class CustomAbortSignal {
+    private listeners: Set<() => void> = new Set()
+    public aborted = false
+
+    public addEventListener(eventName: 'abort', listener: () => void): void {
```
arafatkatze, line 37:
```suggestion
    public addEventListener(listener: () => void): void {
```
Not USed

philipp-spiess: I added it since it's a drop in replacement for AbortController and AbortSignal :)

vscode/src/completions/context/utils.ts
```diff
@@ -17,3 +17,43 @@ export function baseLanguageId(languageId: string): string {
...
+        if (this.aborted) {
+            listener()
```
dominiccooney, line 39: What do you think about `Promise.resolve().then(listener)`? This means that listeners are always called back asynchronously.

I have fought pernicious bugs with callbacks that usually happen a/synchronously which only arise when they are called back the other way. Always doing it one way avoids that possibility.

arafatkatze (approved):
This PR works fine. Attached Debug logs

## Before 
Screenshot 2023-09-28 at 11 38 08 AM


## AFter

Screenshot 2023-09-28 at 11 58 27 AM


## SideQuestion
For some reason changing my config  in the app to enable experimental features doesn't do it for me.
image
So I had to go into the sourcecode to [make the change](https://github.com/sourcegraph/cody/blob/ps/custom-abort-controller/vscode/package.json#L929) and that's fine but i wonder if I am missing something obvious.

dominiccooney (approved):
Looks good, one bit of feedback inline about being consistent calling listeners back asynchronously.

dominiccooney:
Looks like there's some TypeScript errors which we need to attend to here, too.

PR #1228, opened 2023-09-28
revert command configs back to json

abeatrix:
Proposing to revert the changes made in https://github.com/sourcegraph/cody/pull/1197/files where we moved the cody commands configuration file from typescript file to JSON.

The original purpose of using JSON was to use the same kind of configuration our users use.

I agree that it is difficult to read and review a JSON file on github , but I believe the benefits of being able to run into the same issues our users is worth the troubles 😄 

Thoughts?

## Test plan

revert https://github.com/sourcegraph/cody/pull/1197/files


dominiccooney (approved):
Looks good to me, plus one to experiencing what the customer does. I think we do want to make it easy to flip specific commands into code if JSON isn't expressive enough, but while they are just JSON, sure, keeping them in JSON makes sense.

PR #1230, opened 2023-09-28
Autocomplete: Add bookkeeping to reuse completion IDs over multiple suggestions

philipp-spiess:
This PR will add a new book keeping for completion IDs. These are different from the existing suggestion IDs in that:

- Suggestion IDs are used for one suggestion life cycle: One time a completion is rendered on the screen
- Completion IDs are used to identify a specific completion that is printed in a specific document location

The new change makes it so that we will log multiple completions events with the same id. This will help us identify how useful the current cache behavior is.

## Test plan

https://github.com/sourcegraph/cody/assets/458591/5e6604f8-e4a3-4b38-ab5c-16776677d265

### DIFF
--- a/vscode/src/completions/logger.test.ts
+++ b/vscode/src/completions/logger.test.ts
@@ -1,0 +1,158 @@
+import { afterEach, beforeEach, describe, expect, it, MockInstance, vi } from 'vitest'
+
+import { telemetryService } from '../services/telemetry'
+
+import { getCurrentDocContext } from './get-current-doc-context'
+import { InlineCompletionsResultSource, TriggerKind } from './get-inline-completions'
+import * as CompletionLogger from './logger'
+import { RequestParams } from './request-manager'
+import { documentAndPosition } from './test-helpers'
+
+const defaultArgs = {
+    multiline: false,
+    triggerKind: TriggerKind.Automatic,
+    providerIdentifier: 'bfl',
+    providerModel: 'blazing-fast-llm',
+    languageId: 'typescript',
+}
+
+const { document, position } = documentAndPosition('const foo = █')
+const defaultRequestParams: RequestParams = {
+    document,
+    position,
+    docContext: getCurrentDocContext({
+        document,
+        position,
+        maxPrefixLength: 100,
+        maxSuffixLength: 100,
+        enableExtendedTriggers: true,
+    }),
+    selectedCompletionInfo: undefined,
+}
+
+describe('logger', () => {
+    let logSpy: MockInstance
+    beforeEach(() => {
+        logSpy = vi.spyOn(telemetryService, 'log')
+    })
+    afterEach(() => {
+        CompletionLogger.reset_testOnly()
+    })
+
+    it('logs a suggestion life cycle', () => {
+        const item = { insertText: 'foo' }
+        const id = CompletionLogger.create(defaultArgs)
+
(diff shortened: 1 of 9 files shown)


vscode/src/completions/logger.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    })
+    afterEach(() => {
+        CompletionLogger.reset_testOnly()
+    })
+
+    it('logs a suggestion life cycle', () => {
+        const item = { insertText: 'foo' }
+        const id = CompletionLogger.create(defaultArgs)
+
+        expect(id).toBeDefined()
```
dominiccooney, line 46: If `create` is `: SuggestionID` and `create` isn't doing type assertions, can this ever be undefined? And if it is, the next assertion will fail anyway. What is this assertion about?

vscode/src/completions/logger.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        expect(typeof id).toBe('string')
+
+        CompletionLogger.start(id)
+        CompletionLogger.networkRequestStarted(id, { duration: 0.1337 })
```
dominiccooney, line 50: 0.1337 😄

vscode/src/completions/logger.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        // After accepting the completion, the ID won't be reused a third time
+        const id3 = CompletionLogger.create(defaultArgs)
+        CompletionLogger.start(id3)
+        CompletionLogger.networkRequestStarted(id3, { duration: 0 })
+        CompletionLogger.loaded(id3, defaultRequestParams, [item])
+        CompletionLogger.suggested(id3, InlineCompletionsResultSource.Cache, item)
+
+        const loggerItem3 = CompletionLogger.getCompletionEvent(id3)
+        expect(loggerItem3?.params.id).not.toBe(completionId)
```
dominiccooney, line 156: Can you assert that the ID of the second one is the ID of the first one? Bit confused about that aspect.

vscode/src/completions/logger.ts
```diff
@@ -10,13 +10,25 @@ import { telemetryService } from '../services/telemetry'
...
 import { InlineCompletionItemWithAnalytics } from './text-processing/process-inline-completions'
 import { lines } from './text-processing/utils'
 import { InlineCompletionItem } from './types'
 
+// A completion ID is a unique identifier for a specific completion text displayed at a specific
+// point in the document. A single completion can be suggested multiple times.
+type CompletionID = string & { _opaque: typeof CompletionID }
+declare const CompletionID: unique symbol
+
+// A suggestion DI is a unique identifier for a suggestion lifecycle.
```
dominiccooney, line 24:
```suggestion
// A suggestion ID is a unique identifier for a suggestion lifecycle.
```

dominiccooney (approved):
Looks good, some feedback inline

vscode/src/completions/logger.test.ts
philipp-spiess: @dominiccooney Yeah that's asserted based on the logger data being sent but I can make it more explicit, too!

vscode/src/completions/logger.test.ts
philipp-spiess: Yup you're right. I blindly accepted a completion here 😅

PR #1232, opened 2023-09-28
BFG: add new graph context engine

olafurpg:
Previously, the "Graph Context" feature flag was a boolean that used an LSP-powered solution when it had the value `true`. This PR changes the "Graph Context" feature flag into an enum with the values: `lsp`, `bfg` and `none` (default). The `bfg` engine is implemented by a new JSON-RPC server that speaks a bespoke BFG-protocol. BFG is implemented in the sourcegraph/sourcegraph repository and distributed via binary releases in the sourcegraph/bfg repository (https://github.com/sourcegraph/bfg/releases).

The design of BFG is documented in RFC 829
https://docs.google.com/document/d/1smI8mdRhJoZ23uGYh4bxl2iDd-b1kp3qMNIOKwXwacg/edit#heading=h.trqab8y0kufp

## Test plan

See `agent/src/bfg/bfg-context-fetcher.test.ts`.

### DIFF
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -28,7 +28,7 @@ export interface Configuration {
     autocompleteAdvancedAccessToken: string | null
     autocompleteCompleteSuggestWidgetSelection?: boolean
     autocompleteExperimentalSyntacticPostProcessing?: boolean
-    autocompleteExperimentalGraphContext: 'lsp-light' | null
+    autocompleteExperimentalGraphContext: 'lsp-light' | 'bfg' | null
     isRunningInsideAgent?: boolean
 }
 
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -937,13 +937,15 @@
           "markdownDescription": "Rank autocomplete results with tree-sitter."
         },
         "cody.autocomplete.experimental.graphContext": {
-          "type": "enum",
-          "default": null,
+          "type": "string",
+          "default": "none",
           "enum": [
-            null,
-            "lsp-light"
+            "lsp",
+            "lsp-light",
+            "bfg",
+            null
           ],
-          "markdownDescription": "Retrieve information from the code graph to augment the autocomplete results. This is only supported with the `unstable-codegen` provider"
+          "markdownDescription": "Use the code graph to retrieve context for autocomplete requests."
         }
       }
     },
@@ -992,6 +994,7 @@
     "stream-json": "^1.8.0",
     "unzipper": "^0.10.14",
     "uuid": "^9.0.0",
+    "vscode-jsonrpc": "^8.2.0",
     "vscode-languageserver-textdocument": "^1.0.8",
     "vscode-uri": "^3.0.7",
     "web-tree-sitter": "^0.20.8",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -937,13 +937,12 @@
           "markdownDescription": "Rank autocomplete results with tree-sitter."
         },
         "cody.autocomplete.experimental.graphContext": {
-          "type": "string",
-          "default": "none",
+          "type": "enum",
+          "default": null,
           "enum": [
-            "lsp",
+            null,
             "lsp-light",
-            "bfg",
-            null
+            "bfg"
           ],
           "markdownDescription": "Use the code graph to retrieve context for autocomplete requests."
         }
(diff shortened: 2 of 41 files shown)


lib/shared/src/configuration.ts
```diff
@@ -26,7 +26,7 @@ export interface Configuration {
     autocompleteAdvancedAccessToken: string | null
     autocompleteExperimentalCompleteSuggestWidgetSelection?: boolean
     autocompleteExperimentalSyntacticPostProcessing?: boolean
-    autocompleteExperimentalGraphContext?: boolean
+    autocompleteExperimentalGraphContext?: 'lsp' | 'bfg' | 'none' | boolean
```
philipp-spiess, line 29: I think it's possible to use `null` instead of having a string `"none"`, check out the `autocompleteAdvancedProvider` for how this works.

vscode/package.json
```diff
@@ -925,8 +925,13 @@
           "markdownDescription": "Rank autocomplete results with tree-sitter."
         },
         "cody.autocomplete.experimental.graphContext": {
-          "type": "boolean",
-          "default": false,
+          "type": "string",
+          "default": "none",
```
philipp-spiess, line 929:
```suggestion
          "default": null,
```

vscode/package.json
```diff
@@ -925,8 +925,13 @@
...
+          "enum": [
+            "lsp",
+            "bfg",
+            "none"
```
philipp-spiess, line 933:
```suggestion
            null
```

vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -51,24 +49,31 @@ export async function createInlineCompletionItemProvider({
...
     if (providerConfig) {
         const history = new VSCodeDocumentHistory()
-        const sectionObserver =
-            config.autocompleteExperimentalGraphContext || graphContextFlag
-                ? GraphSectionObserver.createInstance()
-                : undefined
+        const isLspGraphContextEnabled =
+            config.autocompleteExperimentalGraphContext === 'lsp' ||
+            // For backwards compatibility, treat true value as 'lsp'
+            config.autocompleteExperimentalGraphContext === true ||
```
philipp-spiess, line 62: Feel free to remove this, I don't think we have to support this use case

vscode/src/completions/providers/createProvider.test.ts
```diff
@@ -34,7 +34,7 @@ const DEFAULT_VSCODE_SETTINGS: Configuration = {
     autocompleteAdvancedAccessToken: null,
     autocompleteExperimentalCompleteSuggestWidgetSelection: false,
     autocompleteExperimentalSyntacticPostProcessing: false,
-    autocompleteExperimentalGraphContext: false,
+    autocompleteExperimentalGraphContext: 'none',
```
philipp-spiess, line 37:
```suggestion
    autocompleteExperimentalGraphContext: null,
```

vscode/src/graph/bfg/bfg-context-fetcher.ts
```diff
@@ -0,0 +1,70 @@
...
+import { logDebug } from '../../log'
+
+import { downloadBfg } from './download-bfg'
+
+export async function createBfgContextFetcher(
+    context: vscode.ExtensionContext,
+    gitDirectoryUri: (uri: vscode.Uri) => vscode.Uri | undefined
+): Promise<GraphContextFetcher> {
+    logDebug('BFG', 'Creating BFG context fetcher')
+    const codyrpc = await downloadBfg(context)
```
philipp-spiess, line 16: I think this makes the download a dependency on extension startup (since `createBfgContextFetcher` is called in via the `main` init function). This might be something that could potentially break the whole extension so maybe we can make this async somehow?

philipp-spiess (approved):
Good stuff! Do we want to add a check to only enable this on TypeScript for now?

SuperAuguste (approved):

ramsey-coding:
Would love to know about further details for the project. 

Also can I get read only access to the document: The design of BFG is documented in RFC 829
https://docs.google.com/document/d/1smI8mdRhJoZ23uGYh4bxl2iDd-b1kp3qMNIOKwXwacg/edit#heading=h.trqab8y0kufp

PR #1234, opened 2023-09-28
Agent: replace console.info with debug logs

chwarwick:
The console.info messages with \n were causing an error 
```
java.lang.IllegalStateException: Missing header Content-Length in input
```
replaced with the debug logger 
## Test plan
`./gradlew -PforceAgentBuild=true :runIDE`
Open a repo that does not have embeddings
ensure no error occurs.


olafurpg (approved):

olafurpg (commented):
Nice find! 🤩

Strum355:
> > Nice find! 🤩
> 
> never would of figured this one out if @Strum355  didn't point directly at it.

Have had this exact problem when working on LSPs of my own, and given that we're using LSP for this as well it just happened to click 😄

PR #1235, opened 2023-09-28
Increase simplified onboarding experiment rollout to 100%

dominiccooney:
There are still users in the control arm of the experiment. This does not remove the control arm, it just changes the rollout to direct all new users into the treatment arm of the experiment.

## Test plan

This has already shipped to half of users, this just increases the rollout.


chenkc805 (approved):
LGTM

PR #1236, opened 2023-09-28
Agent: make chat client respect NODE_TLS_REJECT_UNAUTHORIZED

chwarwick:
The JetBrains extension has a setting to allow untrusted certs, however requests to the chat client set this option directly on the request.  This adds a check of the environment variable set by the extension if it wishes to accept untrusted certs.

## Test plan
`sg start`
`./gradlew -PforceAgentBuild=true :runIDE`
connect to the local instance `https://sourcegraph.test:3443`
verify I can chat successfully


olafurpg (approved):
Very nice find!

PR #1243, opened 2023-09-28
Cody: Fix first command not executing in the sidebar

umpox:
## Description

Fixes bug where Cody would not execute command that required the sidebar, when executed from outside the sidebar (e.g. right click menu)

We didn't account for the possibility that the most recent chat may be an in-progress chat, so we would _always_ stop the current chat. This change stops that.

## Test plan

Tested locally


abeatrix (approved):
omg thank you!!! I can't believe this was it!! 🏆

philipp-spiess (approved):
yooo what, awesome work! 🥇

PR #1255, opened 2023-09-29
Autocomplete: Simplify injected prefix handling and fix logging issue

philipp-spiess:
When the suggestion widget is visible and we decide to use that option to steer the autocomplete, we currently patch the prefix as part of the doc context function. Since the prefix is now longer, we have to "adjust" for this in post processing. Right now this happens as part of the VS Code specific transformation.

This, however, will mean that the prefix in post processing is wrong and any syntactic parsing (or analytical processing) is not correct (since this runs _before_ the VS Code specific transformation is applied). This PR fixes this and extends the completion with the injected prefix in the shared post processing logic.

## Test plan

<table>
<tr>
 <td>Screenshot 2023-09-29 at 14 49 47
 <td>Screenshot 2023-09-29 at 14 51 15
<tr>
 <td>Screenshot 2023-09-29 at 14 49 27
 <td>Screenshot 2023-09-29 at 14 49 41
</table>


vscode/src/completions/text-processing/process-inline-completions.ts
```diff
@@ -63,7 +63,14 @@ export function processItem(params: ProcessItemParams): InlineCompletionItemWith
         return completion
     }
 
-    const adjusted = adjustRangeToOverwriteOverlappingCharacters(completion, { position, currentLineSuffix })
+    // Append any eventual inline completion context item to the prefix if
+    // completeSuggestWidgetSelection is enabled.
+    let withInjectedPrefix = completion
+    if (docContext.injectedPrefix) {
+        withInjectedPrefix = { ...completion, insertText: docContext.injectedPrefix + completion.insertText }
```
arafatkatze, line 70: Such an interesting PR and I loved playing with the `completeSuggestWidgetSelection` the only question that really perplexes me is. How was this even working at all BEFORE this PR? 

I thought somehow the parsing would fail, you don't have to answer this coz I will look into but very interesting PR and feature.

philipp-spiess: > How was this even working at all BEFORE this PR? I thought somehow the parsing would fail

I didn't change much on the behavior. We had the same `insertText` rewriting but at a later point in the pipeline, here: https://github.com/sourcegraph/cody/pull/1255/files#diff-a2dca3f7cbfa2847ddc325842db1723618f70febf6f26d372c7f8adf3d475bafL400-L405

I noticed that my diff is hard to read as it fixes the spacing in comments issue

> On some selections available in the widget it doesn't show anything with the autocomplete. I wonder why that is the case.

VS Code only shows inline suggestions if the current document text matches the selection. so if the document says `console.l` for example but you select `console.dir` in the dropdown, VS Code won't ever show the completion unfortunately.

arafatkatze: Thanks for the explanation

arafatkatze (approved):

abeatrix (approved):

vscode/src/completions/text-processing/process-inline-completions.ts
```diff
@@ -61,9 +61,16 @@ function processCompletion(params: ProcessItemParams): InlineCompletionItemWithA
         throw new TypeError('SnippetText not supported')
     }
 
+    // Append any eventual inline completion context item to the prefix if
+    // completeSuggestWidgetSelection is enabled.
+    let withInjectedPrefix = completion
+    if (docContext.injectedPrefix) {
+        withInjectedPrefix = { ...completion, content: docContext.injectedPrefix + completion.content }
```
beyang, line 68: `docContext.injectedPrefix` could be null, so should this be `(docContext.injectedPrefix || '') + completion.content`?

beyang (changes requested):

beyang (approved):

Issue #1157, opened 2023-09-25
Tree-sitter: use parse-trees to trigger multiline or singleline completions


valerybugakov:
Currently, we decide which completion (multi vs. single) to trigger based on hardcoded block start symbols and a couple of regexes. We can use tree-sitter to categorize completion triggers and use them instead of the current approach for supported languages.

beyang:
Examples:
* Function signature + open parens => generate full function body (even for long functions)
* Struct definitions => line by line (see how this affects CAR)

valerybugakov:
Let's consider this issue closed by:
1. https://github.com/sourcegraph/cody/pull/1259
2. https://github.com/sourcegraph/cody/pull/1318

The follow-up work will be tracked here: https://github.com/sourcegraph/cody/issues/1455

PR #1259, opened 2023-10-02
Autocomplete: add single completion trigger queries

valerybugakov:
## Context

- Trigger single-line completions on `type`/`interface` declarations in typescript and `struct` declarations in go. Powered by tree-sitter queries.
    - If a parse tree is available in `detectMultiline,` we query the document for node types mentioned above.
    - If singleline triggers are found under the cursor, we exit early. The current logic applies for all the other cases (e.g., we return `blockStart` as a multiline trigger if the document prefix ends with it).
    - How do we understand if the node found is under the cursor?
        - Queries captures `block_start` nodes — `{` characters. 
        - Then, we ensure that the `block_start` position equals the cursor position. If that's not the case, we discard found singleline completion trigger. 
        - Potentially, we can find the `block_start` node imperatively from a `trigger` node using parse-tree operations, but I'm not sure if it will be easy to keep the same logic across languages. 
- Adds a new user `cody.autocomplete.experimental.syntacticTriggers` setting to enable the new functionality.
- Adds snapshot tests for the new tree-sitter queries.
- Part of #1157 
- Builds on top of https://github.com/sourcegraph/cody/pull/1307 and https://github.com/sourcegraph/cody/pull/1311

## Planned follow-up

- Add an analytics field to track how often this logic is used.
- Replace the experimental setting with a feature flag to run an experiment.

## Test plan

Added unit tests + CI + tested locally.


vscode/src/completions/tree-sitter/query-tests/annotate-and-match-snapshot.ts
```diff
@@ -245,9 +244,6 @@ export async function annotateAndMatchSnapshot(params: AnnotateAndMatchParams):
     const header = dedent`
         ${commentOutLines(DOCUMENTATION_HEADER, delimiter)}
         ${delimiter}
-        ${delimiter} Tree-sitter query:
-        ${delimiter}
-        ${commentOutLines(rawQuery, delimiter)}
```
valerybugakov, line 250: Removed queries from snapshots because changing them makes it hard to review query result changes.

valerybugakov:
Let's merge this after the release.

vscode/src/completions/inline-completion-item-provider.test.ts
```diff
@@ -186,6 +186,7 @@ describe('InlineCompletionItemProvider', () => {
             },
             "uri": {
               "$mid": 1,
+              "external": "file:///test.ts",
```
valerybugakov, line 189: 🤔

philipp-spiess: Might be caused by a change in the URI implementation 🤔

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -196,6 +199,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             maxPrefixLength: this.config.providerConfig.contextSizeHints.prefixChars,
             maxSuffixLength: this.config.providerConfig.contextSizeHints.suffixChars,
             enableExtendedTriggers: this.config.providerConfig.enableExtendedMultilineTriggers,
+            syntacticTriggers: this.config.syntacticTriggers,
```
philipp-spiess, line 202: I’m curious what you think but I think for things that are small in scope, a user config option might be overkill and just confusing to the end user. Instead, how about we use the Feature Flag framework for this? This would still mean you can start enabling it for yourself (by overwriting the Feature Flag on dotcom) + we can easily trigger an A/B test. We can also remotely control feature flags so if it turns out to be bad, we don't need a second release to disable it.

vscode/src/completions/tree-sitter/query-sdk.ts
```diff
@@ -98,6 +98,13 @@ interface QueryWrappers {
             end?: Point
         ) => never[] | readonly [{ readonly node: SyntaxNode; readonly name: 'blocks' }]
     }
+    singlelineTriggers: {
+        getEnclosingTrigger: (
+            node: SyntaxNode,
+            start: Point,
+            end?: Point
+        ) => never[] | readonly [{ readonly node: SyntaxNode; readonly name: 'trigger' }]
```
philipp-spiess, line 106: What exactly does `never[]` mean? "empty array"? Can't any array be empty? 🤔

philipp-spiess (approved):
LGTM! One comment regarding gating though: Should we move this to a feature flag instead?

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Agree, I'm going to replace it with the feature flag. 

> Planned follow-up
> - Replace the experimental setting with a feature flag to run an experiment.

vscode/src/completions/tree-sitter/query-sdk.ts
valerybugakov: The return type here says: empty array or an array with one item of a specific shape.

valerybugakov:
I'm going to replace the new setting with a feature flag today.

vscode/src/completions/tree-sitter/query-sdk.ts
philipp-spiess: Oh, I overlooked that we specify the arity of the array here. Neat!

PR #1262, opened 2023-10-02
Autocomplete: Enable completeSuggestWidgetSelection by default

philipp-spiess:
This ships completeSuggestWidgetSelection by default but keeps a VS Code config option to turn of the behavior in case people are not happy with the config overwrite. 

This also improves the behavior slightly:

- Introduces a new `triggerKind` `SuggestWidget` so we detect if a completion was triggered because the user cycles through the suggest widget
- Based on this new `triggerKind`, ensure that no latency is used. A user selecting an option is a strong sign that they want to continue code in this direction so we should treat this more like a manual invocation and not like the automatic one.

## Test plan
- Start VS Code in dev mode
- Type `console.` and wait for the dropdown
- Select `dir(` from the dropdown and observe a fast completion like this:
    Screenshot 2023-10-02 at 13 21 44

### DIFF
--- a/vscode/src/completions/get-inline-completions.ts
+++ b/vscode/src/completions/get-inline-completions.ts
@@ -132,6 +132,9 @@ export enum TriggerKind {
 
     /** Completion was triggered manually by the user invoking the keyboard shortcut. **/
     Manual = 'Manual',
+
+    /** When the user uses the suggest widget to cycle through different completions. */
+    SuggestWidget = 'SuggestWidget',
 }
 
 export async function getInlineCompletions(params: InlineCompletionsParams): Promise<InlineCompletionsResult | null> {
@@ -177,7 +180,7 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
         setIsLoading,
         abortSignal,
         tracer,
-        completeSuggestWidgetSelection = false,
+        completeSuggestWidgetSelection = true,
         handleDidAcceptCompletionItem,
         handleDidPartiallyAcceptCompletionItem,
     } = params
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -533,10 +533,12 @@ function isCompletionVisible(
     //   popup into account when generating completions as we do with the
     //   completeSuggestWidgetSelection flag
     //
-    // - When no completions contains characters in the current line that are
-    //   not in the current line suffix. Since VS Code will try to merge
-    //   completion with the suffix, we have to do a per-character diff to test
-    //   this.
+    // - When no completion contains all characters that are in the suffix of
+    //   the current line. This happens because we extend the insert range of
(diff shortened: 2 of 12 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -514,21 +519,21 @@ function isCompletionVisible(
...
     //   sure to not log these completions as displayed.
     //
     //   This check is only needed if we do not already take the completion
     //   popup into account when generating completions as we do with the
     //   completeSuggestWidgetSelection flag
     //
-    //  - When no completions contains characters in the current line that are
+    // - When no completions contains characters in the current line that are
     //   not in the current line suffix. Since VS Code will try to merge
     //   completion with the suffix, we have to do a per-character diff to test
```
dominiccooney, line 538: Drive by comment because you're just reformatting here, but this double negative is a bit hard to grok and whether there's a distinction between the completion's "current line" and the "current line suffix"... the completion is a suffix, right? The suffix wouldn't care what's to the left (in LTR languages) of the cursor I assume?

dominiccooney (approved):
In general I think modes lead to combinatorial explosions of configurations which make software hard to maintain and we should remove the "off" state soon. But if we want to keep it around for a little while in case new evidence of problems emerge then that's good.

vscode/src/completions/inline-completion-item-provider.ts
philipp-spiess: I will reword this:

```
    // - When no completion contains all characters that are in the suffix of
    //   the current line. This is necessary because we extend the insert range
    //   of the completion to the whole line and any characters that are in the
    //   suffix that would be overwritten, will need to be part of the inserted
    //   completion as the VS Code UI does not allow character deletion. To test
    //   for this, we have to do a per-character diff.
```

Is that clearer?  Otherwise I can add an example to the text!

vscode/src/completions/get-inline-completions.ts
```diff
@@ -177,7 +180,7 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
         setIsLoading,
         abortSignal,
         tracer,
-        completeSuggestWidgetSelection = false,
+        completeSuggestWidgetSelection = true,
```
valerybugakov, line 183: Hey @philipp-spiess, how is the value passed here? I'm struggling to find it on the call site. Let me know if I'm looking at the wrong place: 

https://github.com/sourcegraph/cody/blob/c0e9b2940ab7d10336136726f0b0f4fc4e918b9f/vscode/src/completions/inline-completion-item-provider.ts#L215-L235

philipp-spiess: Yeah this seems like it was removed by accident 😬

philipp-spiess: https://github.com/sourcegraph/cody/pull/1408

PR #1265, opened 2023-10-02
Cody: Add benchmark suite

umpox:
# Cody Benchmark Evaluation

This is a benchmark suite that allows us to run Cody against automated code examples with known solutions.

## Usage

**Benchmarking Cody:**

```shell
export BENCHMARK_ENDPOINT=https://sourcegraph.com
export BENCHMARK_ACCESS_TOKEN=sgp_aaaaaaaaaaaaaaaa

pnpm run test:benchmark
```

**Benchmarking Cody on a specific dataset:**

You will often want to run Cody against a specific dataset. This can be helpful for running a series of small benchmarks, or even quickly evaluating against a specific benchmark.

```shell
export BENCHMARK_ENDPOINT=https://sourcegraph.com
export BENCHMARK_ACCESS_TOKEN=sgp_aaaaaaaaaaaaaaaa
export BENCHMARK_DATASET=./test/benchmark/datasets/api-invocation

pnpm run test:benchmark
```

**Benchmarking Cody on external datasets:**

We have some external, larger datasets (e.g. HumanEval), defined in the [cody-evaluation-datasets](https://github.com/sourcegraph/cody-evaluation-datasets) repository. You can still provide these to the benchmark script by ensuring you have that repository cloned, and provding the path to the relevant dataset.

```shell
export BENCHMARK_ENDPOINT=https://sourcegraph.com
export BENCHMARK_ACCESS_TOKEN=sgp_aaaaaaaaaaaaaaaa
export BENCHMARK_DATASET=../../cody-evaluation-datasets/human-eval-infill-single-line

pnpm run test:benchmark
```

**Benchmarking Cody + Copilot:**

The script will run Cody against all benchmark cases, and then run Copilot against the same cases.

We cannot programmatically authenticate with Copilot yet, so the script will pause when Copilot is installed. You need to click the Copilot "Sign in" notification manually in the opened VS Code window, and then the "Resume benchmark suite" notification to continue with the tests.

**Note:** There is currently a bug where authentication state is not shared across VS Code instances in these tests. This means that we need to sign in manually for each test case. This takes too long right now so we should debug and fix this.

**Note:** Copilot uses a URL-based authentication flow. VS Code often struggles to open the correct VS Code window when returning to the editor. If you run into issues, you should try closing all other VS Code windows and try from a separate terminal.

```shell
export BENCHMARK_ENDPOINT=https://sourcegraph.com
export BENCHMARK_ACCESS_TOKEN=sgp_aaaaaaaaaaaaaaaa

pnpm run test:benchmark:copilot
```

### Environment reference

`BENCHMARK_ENDPOINT`: The endpoint to authenticate against.

`BENCHMARK_ACCESS_TOKEN`: The authentication token to provide to the endpoint.

`BENCHMARK_DATASET`: The path to the dataset directory to benchmark against, relative to the current directory.
Can either be a specific case, e.g. `BENCHMARK_DATASET=./test/benchmark/datasets/api-invocation/internal-api-closed-file-multiple-args`, or multiple cases e.g. `BENCHMARK_DATASET=./test/benchmark/datasets/api-invocation`. Defaults to all datasets within `./test/benchmark/datasets`.

`BENCHMARK_COMPARE_WITH`: An alternative extension to benchmark against the same dataset. Will be downloaded from the VS Code marketplace before the tests start. Case-by-case comparisons can be viewed in the console after the tests have finished executing.
e.g. `BENCHMARK_COMPARE_WITH=GitHub.copilot`

## Architecture

LLMs are very difficult to evaluate. This suite attempts to fix this by prompting Cody to generate code from a **"masked" code example**. Given we know the actual solution, we can then **run a prewritten test** against the newly generated code to see if Cody produced a correct result.

The benchmark suite **generates code directly through running Cody in VS Code**. Cody currently relies on many in-editor APIs, running this directly through an editor means **we can focus less on mocking editor APIs** and more on producing new benchmark cases with relatively low friction. In the future, we may want to move or replicate these tests directly through Cody Agent. Alternatively we may want to extend this approach to support running this suite against other editors. One specific benefit to running inside an editor is that we can also run the exact same script, against the same benchmark cases, using a different extension. This means we can compare how Cody performs directly against competitors.

The benchmark suite **evaluates generated code by running an arbitary `testCommand` inside a Docker container**. If the command exits successfully, we consider the generated code to have passed the test, otherwise it has failed. Running inside a Docker container means we can ensure specific test commands are installed, regardless of who is running the benchmark. It also gives us a greater amount of safety, as it means we can execute the LLM-generated code in an isolated environment.

## Adding a new benchmark case

Benchmark cases are defined in the `/datasets` folder in this directory. Each case should have a `config.json` file that looks something like this:

```json
{
  "entryFile": "index.ts",
  "openFiles": [],
  "closedFiles": [],
  "solutionFile": "solution.ts",
  "testFile": "test.ts",
  "testCommand": "ts-node"
}
```

**entryFile**: This is the file that Cody will generate its completion in. It should define the exact position for Cody's completion using the placeholder character: `◆``. You can see other cases for examples of this.

**openFiles**: These are files that should be opened before Cody generates its completion. This can be used as a way of providing additional context to Cody.

**closedFiles**: These are files that exist in the editor, but are not open. Cody may still use these as context, but not directly because they have already been opened.

**solutionFile**: This is the file with the completed solution code that we want Cody to write. **It is not provided to Cody at completion time**. It is primarily as reference example for evaluating Cody's output manually and validating any tests. In the future, we may want to use this as an evaluation point by using edit similarity (ES) and exact match (EM) metrics.

**testFile**: This is the file that will be used to evaluate Cody's completion. It should point at the `entryFile` when running the test, but you should be able to validate the test is correct by manually pointing it at the `solutionFile`.

**testCommand**: This is the command that should be used in order to run the test in `testFile`. This can be any command, which means we can be flexible when testing things across language (e.g. the `python` command for Python completions). If you are adding a new `testCommand`, **you need to ensure a valid executable has been installed in the test container**. If you need to add a new executable, you can do so here: `./datasets/Dockerfile`.

## Configuring the benchmark suite

The benchmark suite can be configured differently depending on your needs. Every benchmark case will inherit the file: `vscode/test/benchmark/fixtures/workspace/settings.json`. You can add new editor settings and change Cody-specific settings to support enabling feature flags, trialing different models and more.

## Test plan

Tested locally (PR is essentially a test script)


slimsag:
@umpox I have these two set:

```
export BENCHMARK_ENDPOINT=https://sourcegraph.com
export BENCHMARK_ACCESS_TOKEN=sgp_SECRET
```

When I run `pnpm run test:benchmark` in the `vscode/` directory, all the tests fail (not a single one passes); each one shows this in the logs:

```
Failed to retrieve access token for Cody from secret storage: Error: token not found
```

I've been trying to debug but unsuccessful so far, do you have any ideas why this might happen? When the editor opens for the test, I am able to type into the Cody chat and see it responding - but autocomplete doesn't seem to work for some reason. In normal VS Code dev mode on this branch I am able to see autocomplete working, just not with this test suite.

image

umpox:
@slimsag If you open up the folder linked to one of the tests, e.g.

image

Is there anything in `output.log` that indicates an authentication failure?

I think we might have a regression where manually triggering the completion is conflicting with an automatic completion from deleting the placeholder cursor. I have removed the manual trigger and it seems to be working on my end

taras-yemets:
Great work, @umpox! Thank you!

I ran all the tests a couple of times, and each time all of them failed (dotcom instance, anthropic/claude-instant-infill model). It seemed suspicious, so I tried to debug a few of them. 

Here's what I found out for [`api-invocation/internal-api-same-file-object-arg`](https://github.com/sourcegraph/cody/tree/698047c46d887f5b2b3fb07f8f6d9cf6b4450877/vscode/test/benchmark/datasets/api-invocation/internal-api-same-file-object-arg) test.
Screenshot 2023-10-03 at 16 57 45
The diff seems correct, but probably because it's missing the closing round bracket (as the [generate.ts](https://github.com/sourcegraph/cody/blob/698047c46d887f5b2b3fb07f8f6d9cf6b4450877/vscode/test/benchmark/datasets/api-invocation/internal-api-same-file-object-arg/generate.ts#L22)) the test fails. 
Ensuring that generate file has the correct syntax (adding the missing bracket) results in an empty completion result. 
Screenshot 2023-10-03 at 17 04 16

Is this behavior expected? Are there any test cases producing relatively stable results so I can play around with them more?

vscode/test/benchmark/constants.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const VSCODE_CODY_ROOT = path.resolve(__dirname, '..', '..', '..', '..')
+export const EXTENSION_TEST_PATH = path.resolve(VSCODE_CODY_ROOT, 'dist', 'tsc', 'test', 'benchmark', 'vscode', 'index')
+export const TEST_WORKSPACE_PATH = path.resolve(VSCODE_CODY_ROOT, 'test', 'benchmark', 'fixtures', 'workspace')
+export const DATASETS_PATH = path.resolve(VSCODE_CODY_ROOT, 'test', 'benchmark', 'datasets')
+
+export const CODY_EXTENSION_ID = 'sourcegraph.cody-ai'
+export const CODY_EXTENSION_CHANNEL_ID = 'extension-output-sourcegraph.cody-ai-#1-Cody by Sourcegraph'
+
+export const CURSOR = '◆'
```
philipp-spiess, line 11: The tiniest of nits but can we use `█` so we have coherence across the cursor symbols we use in the unit tests? 😬

slimsag:
@umpox I got pulled into a different project so may have to return to this next week, so don't block on me. I am hoping to re-test and start writing tests with it Fri/Mon

slimsag:
update: it runs fine now :) Well, for the most part 

image

vscode/test/benchmark/evaluate-test-case.ts
```diff
@@ -0,0 +1,29 @@
...
+export const testCompletionResult = async (
+    testFile: string,
+    testCommand: string,
+    cwd: string
+): Promise<CaseStatus.PASS | CaseStatus.FAIL> => {
+    const benchmarkDockerImage = assertEnv('BENCHMARK_DOCKER_IMAGE')
+    const dockerCommand = `docker run --mount src="${cwd}",target=/app,type=bind ${benchmarkDockerImage}`
+    let status: CaseStatus
+    try {
+        await exec(`${dockerCommand} ${testCommand} ${testFile}`, { cwd })
```
slimsag, line 23:
```suggestion
        await exec(`${dockerCommand} ${testCommand}`, { cwd })
```

It might be better to leave the `testCommand` fully up to the config.json file. Some languages/tools won't want the test file itself as an argument like this

slimsag: @umpox it seems like the command output here gets eaten.

e.g. when writing a new test case, I see just:

```
Exit code:   0
Done
```

I have to go into the tmp directory and run the test command to see why it failed (compiler error output was sent to stderr):

```
# command-line-arguments
./main.go:6:5: f declared and not used
./main.go:6:8: err declared and not used
```

Maybe we can use `docker run -it` flag and capture it in the TypeScript side?

slimsag:
@umpox while writing a single test case (or trying to improve cody's responses for that case) I notice that I have to go into the tmp directory a good amount to e.g. view output.log and it gets quite tedious copying the tmp path to my other editor often. Maybe instead of a random tmp dir every time, we could use a known folder structure like `./benchmark-tmp/<dataset path>` then it'd be possible to leave a terminal in that directory and re-run the test?

vscode/test/benchmark/evaluate-test-case.ts
umpox: Added logs on error!

umpox:
@slimsag Would this help?

Each case logs the workspace dir when the benchmark is running
image

When the benchmark is finished, we log a results.json file which gives extra details for each test:
image

taras-yemets:
@umpox, does https://github.com/sourcegraph/cody/pull/1265/commits/b77915e7181b8ad8fc53f26c0dc0e8ab166ebd60 work for you as expected?
I'm getting `Error: Cannot find module './test.ts'`. Could it be because we're running this command 👇🏻 from `vscode` folder? https://github.com/sourcegraph/cody/blob/54e99e2a8cfaf05db02e0da10a5dd95d6b07d2b3/vscode/test/benchmark/evaluate-test-case.ts#L11

taras-yemets:
> I was getting [this error](https://github.com/sourcegraph/cody/pull/1265#issuecomment-1750435952) because I missed the `testFile` in `config.json`

@umpox, Is this behavior expected? Do we need to specify `testFile` if we reference it in `testCommand` https://github.com/sourcegraph/cody/pull/1332#discussion_r1348413854?

umpox:
@taras-yemets Yep, that's expected

Test file is there so we know what file to copy over before we run the tests. The test command will just be what we actually use to evaluate the test overall.

The reason testFile is optional is that some tests may not actually need a test file (e.g. `go build main.go` doesn't need a test file, so just a `testCommand`)

vscode/test/benchmark/constants.ts
umpox: @taras-yemets @slimsag Just a heads up: I've pushed this change up, please can you update your tests to use █ as the cursor when you next update with this branch :)

umpox:
Keen to get this PR merged so we can focus on iterating with improvements and adding other test cases.

@sourcegraph/cody-vscode @slimsag @taras-yemets Please can I get a review? It's quite a large diff but you can ignore anything loads of files after reading the first benchmark case inside `/vscode/test/benchmark/datasets`

vscode/test/benchmark/utils.ts
```diff
@@ -0,0 +1,130 @@
...
+    return {}
+}
+
+interface WriteBenchmarkResult {
+    path: string
+    testId: string
+    extensionId: string
+    result: BenchmarkResult
+}
+export const writeBenchmarkResult = ({ path, testId, extensionId, result }: WriteBenchmarkResult): void => {
```
taras-yemets, line 112: Will it fail or create `results` folder if it doesn't exist?

umpox: @taras-yemets Nope that's a bug, just pushed a fix

taras-yemets (approved):
Great work, @umpox! Thank you!
I didn't go through each line of the PR, but I checked out it locally and played around with it - seems to work fine 🚀

abeatrix:
Was seeing the following error but the latest commit worked for me!

```
Failed to run tests: Error: ENOENT: no such file or directory, open '/Users/bwork/dev/cody/vscode/test/benchmark/results/results-1696600113091.json'
    at Object.openSync (node:fs:593:3)
    at writeFileSync (node:fs:2325:35)
    at writeBenchmarkResult (/Users/bwork/dev/cody/vscode/dist/tsc/test/benchmark/utils.js:86:35)
    at start (/Users/bwork/dev/cody/vscode/dist/tsc/test/benchmark/main.js:121:50)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/Users/bwork/dev/cody/vscode/test/benchmark/results/results-1696600113091.json'
}
 ELIFECYCLE  Command failed with exit code 1.
```

I tried running it with the `cody-evaluation-datasets`, but got the following error in the middle of the test that caused the test to exit:

```
2023-10-06 07:03:21.627 Electron[10789:1123569] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.
2023-10-06 07:03:22.139 Code Helper (Renderer)[10795:1123653] CoreText note: Client requested name ".NewYork-Regular", it will get TimesNewRomanPSMT rather than the intended font. All system UI font access should be through proper APIs such as CTFontCreateUIFontForLanguage() or +[NSFont systemFontOfSize:].
2023-10-06 07:03:22.139 Code Helper (Renderer)[10795:1123653] CoreText note: Set a breakpoint on CTFontLogSystemFontNameRequest to debug.
Failed to retrieve access token for Cody from secret storage: Error: token not found
Error: command 'cody.autocomplete.manual-trigger' not found
    at y.n (vscode-file://vscode-app/Users/bwork/dev/cody/vscode/.vscode-test/vscode-darwin-arm64-1.83.0/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:1528:12299)
    at y.executeCommand (vscode-file://vscode-app/Users/bwork/dev/cody/vscode/.vscode-test/vscode-darwin-arm64-1.83.0/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:1528:12228)
Exit code:   1
Failed to run tests: Failed
 ELIFECYCLE  Command failed with exit code 1.
 ```

File: vscode/test/benchmark/vscode/execute-completion.ts [hunk 4]
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+export const executeCompletion = async (
+    { entryFile, openFiles = [] }: DatasetConfig,
+    cwd: string
+): Promise<Omit<BenchmarkResult, 'workspacePath'>> => {
+    for (const fileToOpen of openFiles) {
+        // Open any relevant 'open` files that can be used as additional context
+        const doc = await vscode.workspace.openTextDocument(path.resolve(cwd, fileToOpen))
```
abeatrix, line 84: I dont think this is working for me. I do not see any additional file opened that the test was using as context, resulting in the tests relying on that extra context to be failing.

umpox:
@abeatrix thanks for reporting that error, looks like we might need a short delay to ensure the extension fully activates, will add

vscode/test/benchmark/datasets/api-invocation/tiered-types-all-open/solution.ts
```diff
@@ -0,0 +1,5 @@
+import { printBananas } from './print-bananas'
+
+export function main(): void {
+    console.log(printBananas({ green: 1, yellow: 1 }, { green: 2, yellow: 2 }))
```
abeatrix, line 4: question: how does the LLM know what exact numbers to return if the reference file doesn't have them?

I saw this test failed on my side because it has the following output so trying to understand how this works (sorry if it was mentioned somewhere 😅 )

```
🔴 /Users/bwork/dev/cody/vscode/test/benchmark/datasets - tiered-types-all-open:
 Extension: sourcegraph.cody-ai
 Completion: Generated in 1275.6723330002278ms
 Evaluation: FAIL
 Workspace: /var/folders/g1/f24vdsfn2mb1z2_wwyl5vygh0000gn/T/cody-benchmark-TqP8yl
 Diff:
@@ -4 +4,11 @@ export function main(): void {
-    console.log(printBananas(█
+    console.log(printBananas(
+      {
+        yellow: 1,
+        green: 2
+      },
+      {
+        yellow: 3,
+        green: 4
+      }
+    ))
+  }
```

vscode/test/benchmark/vscode/execute-completion.ts
```diff
@@ -0,0 +1,99 @@
...
+        await vscode.window.showTextDocument(doc, { preview: false })
+    }
+
```
abeatrix, line 87:
```suggestion
    for (const fileToClose of closedFiles) {
        const doc = await vscode.workspace.openTextDocument(path.resolve(cwd, fileToClose))
        await vscode.window.showTextDocument(doc, { preview: false })
        await vscode.commands.executeCommand('workbench.action.closeActiveEditor')
        await new Promise(resolve => setTimeout(resolve, 500))
    }
```

We only read closed files if they have been opened before. If that's the intention, I think we will need to add this code to get it to work?


Below are the results I got with this change:

```
Claude-instant-infill:
Pass: 6
Fail: 4
No Completion: 0
Average Time to Completion*: 1684ms
Median Time to Completion*: 1460ms
Pass Rate: 60.00%
Adjusted Pass Rate**: 60.00%
```

```
starcoder-16b:
Pass: 4
Fail: 6
No Completion: 0
Average Time to Completion*: 1117ms
Median Time to Completion*: 1048ms
Pass Rate: 40.00%
Adjusted Pass Rate**: 40.00%
```

abeatrix (commented):
left some comments in line. I couldn't get the external data to work, and couldn't find the token for copilot using the script in README, but the default tests work well for me!


starcoder-7b
Pass: 3
Fail: 7
No Completion: 0
Average Time to Completion*: 961ms
Median Time to Completion*: 964ms
Pass Rate: 30.00%
Adjusted Pass Rate**: 30.00%

starcoder-16b
Pass: 3
Fail: 7
No Completion: 0
Average Time to Completion*: 1035ms
Median Time to Completion*: 1072ms
Pass Rate: 30.00%
Adjusted Pass Rate**: 30.00%

claude-instant-infill:
Pass: 4
Fail: 6
No Completion: 0
Average Time to Completion*: 1195ms
Median Time to Completion*: 1179ms
Pass Rate: 40.00%
Adjusted Pass Rate**: 40.00%

vscode/test/benchmark/datasets/api-invocation/tiered-types-all-open/solution.ts
taras-yemets: >question: how does the LLM know what exact numbers to return if the reference file doesn't have them?

`test.ts` in this case just runs the `main` function from `generate.ts` - if it compiles and runs successfully the test passes. So any result satisfying the `printBananas` return type results in successful test run.

<hr>

I'm not 100% sure why this test failed for you, but I suspect it may be because Cody provided a completion containing the closing curly bracket (last line of the diff provided), which may be odd given that the closing curly bracket existed (last line in `generate.ts`). Two closing brackets in this case result in syntax error, causing `main()` function to fail in `test.ts`. If this assumption is correct, then it may be another example of the edge cases reported before https://github.com/sourcegraph/cody/pull/1265#issuecomment-1745058188.

slimsag (approved):
Definitely in favor of merging and continuing to iterate on this :) Nice work Tom!

vscode/test/benchmark/vscode/execute-completion.ts
umpox: @abeatrix The intention here is just to add files that exist in this temporary codebase, but aren't open in the editor, or have every been opened. This would be helpful for cases where we are fetching code graph context. 

I wasn't aware that we made use of closed-but-prev-open files, that might be a context mechanism we might want to cover here in future, won't do this yet in this PR - thanks for raising this though!

umpox:
> Do we have theories why StarCoder performs consistently worse

@philipp-spiess I was hoping you might have ideas 😅  My assumption is that it's just because it's a stupider model - the HuamnEval tests are mainly just taking a decent amount of context and getting the correct solution, StarCoder just seems to produce more incorrect code, at least compared to claude-instant. Or maybe the other models are better at Python?

I think it'll be really valuable to compare CAR to "how long does the code live in the users' editor". I think the latter gives us more of an indication of quality than the former. I'd be interested to see if some models have more of a tendency to produce code that looks convincing, but isn't actually correct (and thus deleted)

abeatrix:
Thanks for clarifying Tom. Great work! Can't wait to play around some more with this!

abeatrix (approved):

PR #1269, opened 2023-10-02
change: add latency gradually

abeatrix:
Close https://github.com/sourcegraph/cody/issues/1177

NOTE: This change is hidden behind a feature flag `CodyAutocompleteMinimumLatency` and can land after code freeze is over

We learned that from the last experiment, the current logic is too aggressive and not helpful for supported languages.

After our discussion in person ([see slack thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1695991988424799)), this PR includes the following changes:

- The getLatency function now takes the file name and language ID as parameters instead of just the language ID. This provides more context for determining latency values.
- The logic for gradually increasing latency has also been updated to require 5 rejected suggestions before increasing. This avoids prematurely increasing latency due to intermittent rejections.
- Reset user latency after 5 minutes, or file change, or when a suggestion is accepted

## Test plan

See updated tests.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -254,7 +254,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                 if (triggerKind === TriggerKind.Automatic && minimumLatencyFlag) {
                     const minimumLatency = getLatency(
                         this.config.providerConfig.identifier,
-                        document.fileName,
+                        document.uri.fsPath,
                         document.languageId
                     )
 
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -243,8 +243,8 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             //  latency so that we don't show a result before the user has paused typing for a brief
             //  moment.
             if (result.source !== InlineCompletionsResultSource.LastCandidate) {
-                const minimumLatencyFlag = await Promise.resolve(minimumLatencyFlagsPromise)
-                if (!minimumLatencyFlag) {
+                const minimumLatencyFlag = await minimumLatencyFlagsPromise
+                if (minimumLatencyFlag) {
                     const minimumLatency = getLatency(
                         this.config.providerConfig.identifier,
                         document.fileName,
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -244,8 +244,12 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             //  moment.
             if (result.source !== InlineCompletionsResultSource.LastCandidate) {
                 const minimumLatencyFlag = await Promise.resolve(minimumLatencyFlagsPromise)
-                if (minimumLatencyFlag) {
-                    const minimumLatency = getLatency(this.config.providerConfig.identifier, document.languageId)
+                if (!minimumLatencyFlag) {
+                    const minimumLatency = getLatency(
+                        this.config.providerConfig.identifier,
+                        document.fileName,
+                        document.languageId
+                    )
 
                     const delta = performance.now() - start
                     if (minimumLatency && delta < minimumLatency) {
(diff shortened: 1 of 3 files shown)


valerybugakov (approved):
LGTM ✅

vscode/src/completions/latency.test.ts
```diff
@@ -9,99 +9,181 @@ describe('getLatency', () => {
...
+        // max at 2000 after multiple rejections
+        expect(getLatency(provider, fileName, languageId)).toBe(2000)
+        expect(getLatency(provider, fileName, languageId)).toBe(2000)
+        expect(getLatency(provider, fileName, languageId)).toBe(2000)
+        // reset latency on accepted suggestion
+        resetLatency()
+        expect(getLatency(provider, fileName, languageId)).toBe(400)
+    })
+
+    it('returns increasing latency up to max latency for supported language on non-anthropic provider after rejecting multiple suggestions consistently', () => {
```
philipp-spiess, line 150: Should we add another test case for the timeout? You can use fake timers to test this easily https://vitest.dev/guide/mocking.html#timers 

```
vi.advanceTimersByTime(5*60*1000)
```

vscode/src/completions/latency.ts
```diff
@@ -2,37 +2,70 @@ import { logDebug } from '../log'
...
+    // Start after 5 rejected suggestions
+    if (userMetrics.suggested < 5) {
+        return baseline
+    }
 
-    // last suggestion was rejected when last candidated is undefined
-    currentUserLatency = currentUserLatency > 0 ? currentUserLatency * 2 : defaultLatency.user
+    const total = Math.max(baseline, Math.min(baseline + userMetrics.currentLatency, defaultLatency.max))
+
+    userMetrics.currentLatency += defaultLatency.user
```
philipp-spiess, line 52: This can potentially overflow 😬

philipp-spiess (approved):
Awesomesauce

vscode/src/completions/latency.test.ts
abeatrix: TIL :O

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -250,9 +250,13 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             // latency so that we don't show a result before the user has paused typing for a brief
             // moment.
             if (result.source !== InlineCompletionsResultSource.LastCandidate) {
-                const minimumLatencyFlag = await Promise.resolve(minimumLatencyFlagsPromise)
+                const minimumLatencyFlag = await minimumLatencyFlagsPromise
                 if (triggerKind === TriggerKind.Automatic && minimumLatencyFlag) {
-                    const minimumLatency = getLatency(this.config.providerConfig.identifier, document.languageId)
+                    const minimumLatency = getLatency(
+                        this.config.providerConfig.identifier,
+                        document.fileName,
```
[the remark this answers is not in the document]
abeatrix: OH good point out. I assumed fileName = fsPath, glad you mentioned it!

vscode/src/completions/latency.test.ts
```diff
@@ -1,107 +1,202 @@
...
 
-    it('returns default latency for CSS language on non-anthropic provider after accepting suggestion consistently', () => {
+    it('returns default latency for CSS after accepting suggestion and resets after 5 minutes', () => {
         const provider = 'non-anthropic'
+        const fileName = 'foo/bar/test.css'
         const languageId = 'css'
 
         // start with default baseline latency with low performance and user latency added
-        expect(getLatency(provider, languageId)).toBe(1400)
+        expect(getLatency(provider, fileName, languageId)).toBe(1400)
```
philipp-spiess, line 114: I am not sure if the fact that we're on non-anthropic here and a low acceptance rate language justifies the _additive_ latency. I think the idea of the 400ms is to make it closer to the current anthropic implementation but when we change the latency to be higher here, it means that we will throttle heavier when the user is not using anthropic. This might be a bit confusing and harder to test. WDYT? 

I think the 400ms should just be applied at the minimum, and not be an additional penalty for using non-anthropic.

abeatrix: Just to make sure I'm with you, do you mean if a user is on non-anthropic(400ms base) and low-performance language (1000ms base), they should start at 1000ms, instead of 1400ms, is that correct?

abeatrix: (I think what you suggested makes sense too. We should start small and adjust them based on the results we will be getting :) )

abeatrix: updated!

philipp-spiess (approved):
Yes Yes Yes

PR #1270, opened 2023-10-02
fix: set display text for typewriter

abeatrix:
follow up on https://github.com/sourcegraph/cody/pull/1220

Right now we are not streaming assistant response as we are not setting display text in typewriter for assistant response because of the change in 1220.

This PR should fix the issue

## Test plan

1. run /test command on selected code
2. before the change, you will not see the response until cody has finished the response
3. when building from this branch, you should see the response from cody streaming in chat view


kalanchan (approved):

unknwon:
This is the last please :D

PR #1272, opened 2023-10-02
agent: start unifying event logging between clients

tjdevries:
I still have some more local testing to do, but just wanted to get your initial thoughts on this

## Test plan

- [ ] Send requests from local client and confirm that they are showing up for logging purposes

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -14,7 +14,7 @@ import { newTextEditor } from './AgentTextEditor'
 import { AgentWorkspaceDocuments } from './AgentWorkspaceDocuments'
 import { AgentEditor } from './editor'
 import { MessageHandler } from './jsonrpc'
-import { AutocompleteItem, ExtensionConfiguration, RecipeInfo } from './protocol'
+import { AutocompleteItem, ClientInfo, ExtensionConfiguration, RecipeInfo } from './protocol'
 import * as vscode_shim from './vscode-shim'
 
 const secretStorage = new Map<string, string>()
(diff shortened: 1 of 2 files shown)


valerybugakov (commented):
Heads up: we have another related initiative going on here: https://github.com/sourcegraph/cody/pull/1192

> This PR attempts to add integrations with the new telemetry SDK at https://github.com/sourcegraph/telemetry to the VSCode Cody client

It should supersede the `logEvent` API we use now.

agent/src/agent.ts
```diff
@@ -308,4 +310,46 @@ export class Agent extends MessageHandler {
         this.oldClient = client
         return client
     }
+
+    private async logEvent(name: string): Promise<null> {
```
bobheadxi, line 314: @tjdevries - for Event Logging Everywhere, we're working on a new SDK for recording events using a new mutation and event schema, so that individual instances report telemetry. There's a draft PR here: https://github.com/sourcegraph/cody/pull/1192

This won't really be available until late next week at the earliest, but it would be nice to have this new agent protocol reflect the API we plan to introduce, e.g. by changing this a bit here:

```suggestion
    private async recordEvent(feature: string, action: string): Promise<null> {
```

cc @nathan-downs

bobheadxi: A bit more context here: https://docs.sourcegraph.com/dev/background-information/telemetry

I will be OOO this week but can cc @nathan-downs @valerybugakov (who we just shared some context with today) for discussion in Slack if you have any questions!

agent/src/protocol.ts
```diff
@@ -165,6 +164,21 @@ export interface ExtensionConfiguration {
     debug?: boolean
     verboseDebug?: boolean
     codebase?: string
+    eventProperties?: EventProperties
+}
+
+export interface EventProperties {
+    /** Anonymous user ID */
+    user: string
```
olafurpg, line 172: wdyt?
```suggestion
    anonymousUserID: string
```

agent/src/protocol.ts
```diff
@@ -165,6 +164,21 @@ export interface ExtensionConfiguration {
...
+
+    /** Event prefix, like 'CodyNeovimPlugin' */
+    prefix: string
+
+    /** Name of client, like 'NEOVIM_CODY_EXTENSION' */
+    client: string
+
+    /** Source type enum*/
+    source: 'IDEEXTENSION'
```
olafurpg, line 181: Should we remove this property if it only permits the value `IDEEXTENSION`?

agent/src/protocol.ts
```diff
@@ -165,6 +164,21 @@ export interface ExtensionConfiguration {
     debug?: boolean
     verboseDebug?: boolean
     codebase?: string
+    eventProperties?: EventProperties
```
olafurpg, line 167: Worth adding a comment explaining that by providing a value here then the agent will handle telemetry. If the value is not provided then the client must use `graphql/logEvent`

agent/src/agent.ts
```diff
@@ -308,4 +310,46 @@ export class Agent extends MessageHandler {
...
+        const client = await this.client
+        if (!client) {
+            return null
+        }
+
+        const clientInfo = this.clientInfo
```
olafurpg, line 320: Super minor: destructuring could have been nice here

```
const { extensionConfiguration, eventProperties } = this.clientInfo
if (!extensionConfiguration || !eventProperties) { return null }
```
```

olafurpg (approved):
LGTM 👍🏻 This is a great idea. Would be nice to abstract over logging

agent/src/protocol.ts
tjdevries: Figured later we would probably have something like "COMMANDLINE" or whatever and I was going to enum it in the future.

agent/src/agent.ts
tjdevries: What I see there mostly makes sense, I think we can continue to push this PR forward and then replace the implementation with the v2 stuff once it lands. Thanks for the heads up

PR #1198, opened 2023-09-27
fix: doc prompt improvements

toolmantim:
This updates the `/doc` prompt to product better quality documentation across more cases. It:

- Reduces the chance that parameters are documented unnecessarily (e.g. for code that doesn't take parameters)
- Reduces the verbosity of documentation, creating more focused explanations
- Increases the chance that the correct documentation style is used for different parts of the file

Closes #1178

## Test plan

- Cloned each example repo, created before + after branches
- Ran `/doc` over each example, committing `main` to before and this branch to after
- Git diffed branches

PR #1273, opened 2023-10-02
fix: doc command

abeatrix:
I noticed the `/doc` command built from the current main branch has the following issues:
- a new empty line is added before the 
- if you run the /doc command and then reload VS code, your chat will start with the output of your last /doc command in chat view

image

This PR fixes the issues listed above.

Also merged the prompt with the one tested by Tim in https://github.com/sourcegraph/cody/pull/1198

IMPORTANT: This PR depends on the prompt improvement made in https://github.com/sourcegraph/cody/pull/1198

## Test plan

- highlight a function or method in the editor
- right click on the highlighted code, then select `Cody>Document Code`
- Expected behavior: Docstring added directly above the highlighted code without an extra new line added at the end

#### Before

https://www.loom.com/share/c1d316d9031d4550952da81d5e5c61fd?sid=3c06e825-e6c3-4c43-a17b-e93f186947a3

https://github.com/sourcegraph/cody/assets/68532117/664cc0b1-c2e2-4e1c-a164-b8b4f8a6d6b5

#### After

https://www.loom.com/share/accef38b7f0f4d0c8e1ba8b89581acd2?sid=29fe90a4-9f5e-4a51-8c40-f4adad1c848f

https://github.com/sourcegraph/cody/assets/68532117/8efddda8-5664-4bcb-adce-732869911184


toolmantim (approved):
Tested the "if you run the /doc command and then reload VS code, your chat will start with the output of your last /doc command in chat view" part of this and works as expected 👍🏼 But I reckon let's leave the prompt changes to the other PR.

vscode/src/chat/MessageProvider.ts
```diff
@@ -190,11 +195,15 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
...
                 await typewriter.finished
                 const lastInteraction = this.transcript.getLastInteraction()
-                if (lastInteraction && multiplexerTopic !== 'fixup') {
+                if (lastInteraction) {
                     let displayText = reformatBotMessage(text, responsePrefix)
                     // TODO(keegancsmith) guardrails may be slow, we need to make this async update the interaction.
                     displayText = await this.guardrailsAnnotateAttributions(displayText)
                     this.transcript.addAssistantResponse(text, displayText)
+                    // remove display text from last interaction if this is a non-display topic
+                    if (nonDisplayTopics.has(multiplexerTopic)) {
```
dominiccooney, line 204: Would it be simpler to just pass `nonDisplayTopics.has(multiplexerTopic) ? '' : displayText` on the line above?

vscode/src/non-stop/FixupController.ts
```diff
@@ -249,7 +249,7 @@ export class FixupController
 
         // Insert updated text at selection range
         const editOk = await editor.edit(editBuilder => {
-            editBuilder.insert(range.start, textLines.join('\n') + '\n')
+            editBuilder.insert(range.start, textLines.join('\n').trimEnd() + '\n')
```
dominiccooney, line 252: Sometimes you probably *do* want to insert blank lines, and `trimEnd` prevents that because it will eat multiple newlines. Why not just remove one trailing '\n', if that is present?

vscode/src/non-stop/FixupController.ts
```diff
@@ -249,7 +249,7 @@ export class FixupController
```
dominiccooney, line 249: By the way, above it might be better to copy the whitespace from the start of the line to `nonEmptyStartIndex`. That way if the file is indenting with tabs, (or worse, a mix of tabs and spaces), we are more likely to get the indentation right.

lib/shared/src/chat/transcript/interaction.ts
```diff
@@ -80,4 +80,13 @@ export class Interaction {
             timestamp: this.timestamp,
         }
     }
+
+    public removeDisplayText(): void {
+        if (this.humanMessage) {
+            this.humanMessage.displayText = undefined
```
dominiccooney, line 86: Why would we want to suppress the human display text at this point?

lib/shared/src/chat/transcript/index.ts
```diff
@@ -129,6 +129,13 @@ export class Transcript {
         })
     }
 
+    public removeDisplayTextFromLastInteraction(): void {
+        if (!this.getLastInteraction()) {
+            return
+        }
+        this.getLastInteraction()?.removeDisplayText()
```
dominiccooney, line 136: The way `?.` works, the call won't happen if `getLastInteraction()` returns undefined. So you can simplify this. (Personally, I find this aspect of TypeScript hard to read, because the meaning of:

```
this.foo()?.bar().baz()
```

and

```
const x = this.foo()?.bar()
x.baz()
```

are different, and JavaScript does not work this way. But oh well.)

```suggestion
        this.getLastInteraction()?.removeDisplayText()
```

dominiccooney (changes requested):
Some feedback inline.

I think we should try to simplify this a bit—only strip one newline, not multiple ones; not prune human text from the transcript?

vscode/src/non-stop/FixupController.ts
abeatrix: For insert, we want to make sure the new line added at the end does not include the indentation we added in the previous step, that's why trimEnd() before adding a new line works in this case.

Right now the issue with the current implementation is that we are adding new line twice, with one including the indentation (see  example in description).

Does this make sense? Or do you think this can be simplified?

lib/shared/src/chat/transcript/interaction.ts
abeatrix: This is a method inside the interaction class, so I think removing display text for all interaction (both human and assistant) in this method would make sense? Or do you think this should just remove display text for assistant message and rename the method to "removeAssistantDisplayText" instead?

abeatrix: (My initial thought was to make sure the display text is also removed from human display through this method, in cases display text were not removed in the previous steps)

vscode/src/non-stop/FixupController.ts
abeatrix: i reread what you said and I think i misunderstood your point here. Will update the code with your suggestion, thanks!

lib/shared/src/chat/transcript/interaction.ts
abeatrix: removed

vscode/src/non-stop/FixupController.ts
```diff
@@ -246,10 +246,11 @@ export class FixupController
         const nonEmptyStartIndex = editor.document.lineAt(range.start.line).firstNonWhitespaceCharacterIndex
         // add indentation to each line
         const textLines = text.split('\n').map(line => ' '.repeat(nonEmptyStartIndex) + line)
-
+        // join text with new lines, and then remove everything after the last new line if it only contains white spaces
+        const replacementText = textLines.join('\n').replace(/ +$/, '')
```
dominiccooney, line 250: Looks good, you could use `/[ \t]+$/` to handle tabs here too, for when we match tabs indentation.
```suggestion
        const replacementText = textLines.join('\n').replace(/[ \t]+$/, '')
```

dominiccooney (approved):
Looks great! Thank you!

abeatrix:
@dominiccooney thank you for your thoughtful review as always 🙇‍♀️

PR #1274, opened 2023-10-02
fix: preamble leak for anthropic

abeatrix:
Fix issues that cause the preamble to leak for anthropic by ending the code snippets with `\n<--End of code-->` to make sure Claude instant knows where the code ends, instead of having the code snippets end with a new line that confused Claude and makes it think the code was not completed and should be continued from there.

#### Before

Screenshot 2023-10-02 at 3 27 09 PM

#### After

Screenshot 2023-10-02 at 3 28 30 PM

## Test plan

1. build Cody from this branch
2. Go to `line 213` of [vscode/webviews/Chat.tsx](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/Chat.tsx?L213) in your editor
3. Instead of seeing `Human: ` as a suggestion, Cody should not suggest anything here.

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -85,8 +85,7 @@ export class AnthropicProvider extends Provider {
             },
             {
                 speaker: 'human',
-                text: `Below is the code from file path ${this.options.fileName}. Review the code outside the XML tags to detect the functionality, formats, style, patterns, and logics in use. Then, use what you detect and reuse methods/libraries to complete and enclose completed code only inside XML tags precisely without duplicating existing implementations.
-                Here is the code: ${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}`,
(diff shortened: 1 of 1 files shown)


valerybugakov (approved):
It looks logical and solves the issue we found, but I wonder if it affects the output unexpectedly (like when we found out that Claude doesn't like trailing spaces). Did you have a chance to run this change against our completions dataset?

philipp-spiess:
Agree with Valery. I’m a bit concerned too about shipping this so shortly before the release. Maybe as a quick fix we can add a filter for completions that start with `Human: ` from Anthropic and then we can land this after the release and test it a bit more? 

Unrelated but instead of `\n<--End of code-->` did you ever try to wrap everything in an xml tag like so?

```
<code>${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}</code>`
```

abeatrix:
as discussed with @philipp-spiess on slack, we will merge this fix and make additional changes in patches if required.

PR #1293, opened 2023-10-03
Don't show completion notice for existing installs

DanTup:
To avoid showing the notice to existing users (who have probably already accepted completions), disable it if there is chat history. In future (when the completion flag has been set for existing users), this extra check could be removed.

## Test plan

Tested with new test via `pnpm test:unit`

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -71,6 +71,8 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
     /** Accessible for testing only. */
     protected lastCandidate: LastInlineCompletionCandidate | undefined
 
+    private isProbablyNewInstall = true
+
     constructor({
         graphContextFetcher = null,
         completeSuggestWidgetSelection = false,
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):
Very nice!

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -342,17 +347,37 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+
+        const key = 'completion.inline.hasAcceptedFirstCompletion'
+        if (localStorage.get(key)) {
+            return // Already seen notice.
+        }
+
+        // Mark as seen, so we don't show again after this.
+        void localStorage.set(key, 'true')
+
+        if (!this.isProbablyNewInstall) {
```
philipp-spiess, line 371: @abeatrix suggested using [vscode.env.isNewAppInstall](https://code.visualstudio.com/api/references/vscode-api#env) instead (I wasn't aware this exists 🙈). Do you think this would work? :)

toolmantim (approved):
🙏🏼

PR #1295, opened 2023-10-03
add event logging validation to inline test

akalia25:
Adds logging validation to`inline-completion.test.ts`, to ensure that the correct events are logged when inline completions occur. 

To note, for this test we sort the array of expectedEvents and loggedEvents when comparing because `CodyVSCodeExtension:completion:suggested` and `CodyVSCodeExtension:completion:accepted` events fire at the same time making this test flaky if we do not sort when comparing the arrays.
## Test plan
Tested locally running pnpm test:e2e

### DIFF
--- a/vscode/test/e2e/inline-completion.test.ts
+++ b/vscode/test/e2e/inline-completion.test.ts
@@ -1,8 +1,19 @@
 import { expect, Page } from '@playwright/test'
 
+import { loggedEvents, resetLoggedEvents } from '../fixtures/mock-server'
+
 import { sidebarExplorer, sidebarSignin } from './common'
 import { test } from './helpers'
 
+const expectedOrderedEvents = [
+    'CodyVSCodeExtension:completion:suggested',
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):
This is awesome!!!

vscode/test/e2e/inline-completion.test.ts
```diff
@@ -26,6 +37,7 @@ test('shows completion onboarding notice on first completion accept', async ({ p
     // Accept the completion and expect the text to be added and
     // the notice to be shown.
     await acceptInlineCompletion(page)
+    console.log('1')
```
philipp-spiess, line 40: Huh this sneaked in @akalia25 :D

akalia25: 🙈 oh no! Used that while testing the changes. I just created a [commit here](https://github.com/sourcegraph/cody/pull/1295#pullrequestreview-1657192587) to remove that. Thank you for catching 🙇

PR #1296, opened 2023-10-03
Agent: disable tree-sitter setting

olafurpg:
Previously, the agent was crashing because it was failing to load WASM modules for tree-sitter. This PR fixes the problem by disabling the setting for tree-sitter post-processing. The autocomplete e2e test was updated to use TypeScript since our tree-sitter experiments are mostly focused on TypeScript for now.

## Test plan

- `cd agent/`
- `src login`
- `VITEST_ONLY=NotConfigured pnpm run test src/index.test.ts`

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -135,8 +135,8 @@ describe.each([
     })
 
     it('returns non-empty autocomplete', async () => {
-        const filePath = '/path/to/foo/file.js'
-        const content = 'function sum(a, b) {\n    \n}'
+        const filePath = '/path/to/foo/file.ts'
+        const content = 'function sum(a: number, b: number) {\n    \n}'
         client.notify('textDocument/didOpen', {
             filePath,
             content,
@@ -144,8 +144,8 @@ describe.each([
         })
         const completions = await client.request('autocomplete/execute', {
             filePath,
-            position: { line: 1, character: 4 },
-            triggerKind: 'Automatic',
+            position: { line: 1, character: 3 },
+            triggerKind: 'Invoke',
         })
         assert(completions.items.length > 0)
     })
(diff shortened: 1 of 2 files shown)


tjdevries (approved):

dominiccooney (approved):
Do we want a FIXME on the test to remove the type annotations when TreeSitter is back?

valerybugakov (approved):

olafurpg:
@dominiccooney FIXME comments have a tendency to rot in my experience. Closing feature gaps with VS Code is going to be a consistent theme over the next months and that includes matching context quality in the agent. We will definitely prioritize removing this workaround because there seems to be a 5% CAR gap between JetBrains and VS Code.

dominiccooney:
@olafurpg 

> FIXME comments have a tendency to rot in my experience.

My $0.02:

First, a lot of FIXMEs aren't well written, which makes them prone to rot. A good FIXME explains when it is unblocked which makes it easier to clean them up.

Second, if there's a lot of rotten FIXMEs in a codebase that itself is a signal.

How essential are the test changes to this change? If they're essential I think a FIXME is useful; if I told a random contributor "hey, syntactic postprocessing is enabled again, could you update the tests?" the chances they produce the changes you have here are low because they seem subtle. But I'm not familiar enough with these tests to be sure... are completions for JavaScript broken because that flag was turned off? Completions are slower... but only up to 100ms slower... and the execution times were so reliable/timeouts so tight before?

olafurpg:
@dominiccooney the test changes are not strictly necessary. The barrier to tweak the agent tests is low right now because we don't run them in CI. The hardcoded timeouts are unrelated to the autocomplete test, I just had to tweak the numbers to get the test passing locally. It's high on the priority list to get these tests into a good enough state (no magic timeouts, mocked network requests, etc.) so we can start running them on every PR in CI.

I changed from JavaScript to TypeScript because I know we seem to run experiments first with the TypeScript grammar before expanding to other languages. It wasn't strictly needed for this PR.

I've opened a P1 issue to enable tree-sitter via the agent. It's a high priority item because I want to close the 5% CAR gap between the editors https://github.com/sourcegraph/cody/issues/1309

Issue #1294, opened 2023-10-03
Autocomplete: When creating a multi line completion with a suffix in the same line, the closing bracket is not properly truncated


philipp-spiess:
This seems to be a regression with the tree sitter based multiline truncation only (cc @valerybugakov)

## With `"cody.autocomplete.experimental.syntacticPostProcessing": false`

<img width="895" alt="Screenshot 2023-10-03 at 14 05 23" src="https://github.com/sourcegraph/cody/assets/458591/2a041d9d-5137-4ba6-b439-455ba36e44e9">

## On `main`:

<img width="989" alt="Screenshot 2023-10-03 at 14 04 21" src="https://github.com/sourcegraph/cody/assets/458591/348eb6ad-f626-47d8-991b-e653ae953b21">

philipp-spiess:
Lol i can't reproduce this anymore, something fishy is going on 🤔

valerybugakov:
What a rollercoaster!

valerybugakov:
Existing unit tests should cover this case, but I added another example just in case: https://github.com/sourcegraph/cody/pull/1303
Let me know if you see ways for our unit tests to miss this issue.

PR #1303, opened 2023-10-04
Autocomplete: add multiline truncation unit test

valerybugakov:
## Context

- Closes https://github.com/sourcegraph/cody/issues/1294
- We already have unit tests for similar cases for both truncation strategies, tree-sitter and indentation-based. This PR adds another one that matches the example shared in the linked issue to ensure we cover this case.

## Test plan

CI


philipp-spiess (approved):
Thanks a ton! And sorry for the false alert haha

PR #1306, opened 2023-10-04
VS Code: singleton graphql client and feature flag provider

valerybugakov:
## Context

- Makes graphql client and feature flag provider singletons to avoid passing them around as parameters.
- Continue the migration to the singleton module-level exports. Related work: https://github.com/sourcegraph/cody/pull/876 and https://github.com/sourcegraph/cody/pull/912
- No functional changes.

## Test plan

CI and verified that `pnpm generate:completions` still works.


taras-yemets (approved):

philipp-spiess (approved):
🍯

PR #1307, opened 2023-10-04
Autocomplete: restructure tree-sitter related modules

valerybugakov:
## Context

- Move modules around to prepare for functional changes.
- Extracted from https://github.com/sourcegraph/cody/pull/1259 to make reviews easier.

## Test plan

CI


vscode/src/completions/tree-sitter/query-tests/annotate-and-match-snapshot.ts
```diff
@@ -245,9 +245,6 @@ export async function annotateAndMatchSnapshot(params: AnnotateAndMatchParams):
     const header = dedent`
         ${commentOutLines(DOCUMENTATION_HEADER, delimiter)}
         ${delimiter}
-        ${delimiter} Tree-sitter query:
-        ${delimiter}
-        ${commentOutLines(rawQuery, delimiter)}
```
valerybugakov, line 250: Removed queries from snapshots because changing them makes it hard to review query result changes.

vscode/src/completions/tree-sitter/query-tests/blocks.test.ts
```diff
@@ -1,23 +1,19 @@
...
 import { initTreeSitterParser } from '../../test-helpers'
 import { SupportedLanguage } from '../grammars'
-import { getDocumentQuerySDK } from '../queries'
+import { getDocumentQuerySDK } from '../query-sdk'
 
 import { annotateAndMatchSnapshot } from './annotate-and-match-snapshot'
 
 describe('getFirstMultilineBlockForTruncation', () => {
-    beforeAll(async () => {
-        await initTreeSitterParser(SupportedLanguage.TypeScript)
```
philipp-spiess, line 11: What's the reason behind requiring a manual initialization for these APIs? Olaf ran into some issues when he [started writing tests for the BFG graph context](https://github.com/sourcegraph/cody/pull/1232/files#diff-b31f0976fe5924ed781bbe1d6731836d4a16642a3c49087727ece401650be5f4R29) during the offsite and I wasn't sure why the init was necessary there. Maybe we can hide the initialization behind the first call to `getDocumentQuerySDK`?

philipp-spiess (approved):
good stuff

vscode/src/completions/tree-sitter/query-tests/blocks.test.ts
valerybugakov: A couple of reasons:

1. Tree-sitter was behind the experimental setting until the previous week, and I wanted to keep one place where the setting is used to keep it simple (in `main.ts` and not add it in other tree-sitter-related utils).
2. All the pre/post-processing code is synchronous, and it's nice to have this property to ensure we avoid accidentally adding slow async calls. In the production code, parsers and queries are always initialized separately.

We can add a helper that inits the parser behind the scenes or make everything async. I prefer keeping things sync now + add a comment with this context. WDYT?

philipp-spiess: That makes sense! Do we throw an appropriate error message if the init stuff is not called? If that's the case, I assume it's not a big deal either since we can easily recover from the failure state

valerybugakov: > Do we throw an appropriate error message if the init stuff is not called?

We return null from `getDocumentQuerySDK` if a relevant parser is not initialized. It's not an error in the production code because we might not have a parser for the specified language, or tree-sitter might be disabled. Where would you expect an error?

philipp-spiess: Hmm, I see. So we use `getDocumentQuerySDK` for both checking if a parser is available and actually parsing. Makes sense

valerybugakov: Yeah, the SDK is unavailable if a parser is not initialized, and it's not an error. I'll think about how to make it more prominent and maybe change the API to reflect that. Thank you for sharing this feedback!

PR #1313, opened 2023-10-04
refactor: move vscode context out of shared

abeatrix:
refactor: move `vscode-context` from `lib/shared` to `vscode/editor-context`

Right now, `vscode-context.ts` is living in the share library does not work for other editors, and prevents other editors from implementing custom commands through the custom-prompt recipe.

This change allows other editors to implement editor-level context through the new `EditorContext class`, and enable other recipes to get editor context easily without needing to implement the same methods for each recipe.

Renamed '@sourcegraph/cody-shared/src/chat/prompts' to '@sourcegraph/cody-shared/src/chat/commands'

No current functions or features were altered in this PR.

Will clean up the `vscode/editor-context` file in a separate PR since that would involve updating the current functions.

## Test plan

1. build from this branch
2. try different commands
3. everything should work the same


dominiccooney:
I hear so many mixed messages about libshared versus agent and abstractions versus editors implementing the vscode APIs and the VScode implementations "just working."

On the one hand, it is a drag on VScode to have to plumb through abstractions like EditorContext. On the other hand, maybe some abstraction makes sense because different editors are different.

CC @olafurpg who has opinions on this.

@abeatrix the intent of this PR is clear, but can you give a bit more context? What non-VScode editor wanted to implement custom commands and the VScode shim wasn't working for that?

olafurpg:
My 2c: `vscode` APIs are well designed and we should avoid introducing abstractions over them. For example, I have never enjoyed using `Editor` from `lib/shared/src/editor/index.ts`. I can only speak for the JetBrains client (not web), by gut feeling it's best to use `vscode` APIs.

olafurpg:
We will deal with it in the agent for how to extract the useful bits out of the `vscode/` extension. So far, I haven't hit on any major cases where we haven't been able to reproduce the VS Code environment (ignoring interactions with other VS Code extensions like LSP for grap context)

abeatrix:
@olafurpg @dominiccooney To clarify, does this mean we can keep using the vs code api in lib/shared and it wont break other editors even if the agent hasn't implemented that bit of the vs code api?

I also not enjoying using the Editor class. What's stopping us from replacing it with the one that we are using for Vscode? It'd save us a lot of time if we can just use vs code api in lib/share from now on 🚀

olafurpg:
I have no concerns about using `vscode` APIs from lib/shared wrt. the agent. It would be nice if you ping us in #wg-cody-agent if you start using a new vscode API so that we can ensure it's available in vscode-shim.ts. We are going to be increasing investments in automated testing over the coming weeks but we currently rely mostly on manual testing to make sure everything is working correctly.

olafurpg:
> I also not enjoying using the Editor class. What's stopping us from replacing it with the one that we are using for Vscode?

BTW, I want to replace our usage of `AgentEditor` with `VSCodeEditor`. I don't think there are any major blocking issues from doing this and it could help close some gap wrt. chat quality.

abeatrix:
@olafurpg Thank you so much for clarifying! And that sounds great! I've put some time on your calender for next week to learn more about agent and see if there is anything I can help :) 

At the moment, all our default commands are using the same `custom-prompt` recipe as custom commands because technically, all the recipes are the same, it's just the prompt and the context they use are different. Right now, we are using the folding range API, search AP, symbol API to get context for our unit test command and doc command, which are not implemented in agents.

@dominiccooney @olafurpg That said, do you think at least for now, that I should move forward with the change proposed in this PR where we create the `EditorContext` class for the command recipe, which can easily be replaced later with the new `VSCodeEditorContext` class when we are sure all the needed VS Code APis are available in agents? 
If not, i will update my PR to replace `EditorContext` with the context from the new  `VSCodeEditorContext` class, what do you think?

olafurpg:
> Right now, we are using the folding range API, search AP, symbol API to get context for our unit test command and doc command, which are not implemented in agents.

IIRC, some of these are accessed via `vscode.commands.executeCommands()`, which return `undefined` in the agent. As long as `undefined` is handled then it should be fine (doesn't crash) with the agent. The most important part is that the VS Code extension is unblocked from doing experiments to validate these feature make meaningful impact on quality. I'd love to eventually move these features (folding ranges, document symbols) away from vscode/LSP and into a bespoke solution like BFG. 

We can discuss this in our call on Tuesday, but I want to call out that folding ranges may not be exactly what you are looking for. You can probably replace usage of folding ranges with a basic heuristic that follow the indentation (super fast, works for all file types), which is how we implement folding in the blob view (Code Search).

olafurpg:
FTR, we implement support for `executeCommand` in the agent, we just don't register implementations yet for the folding ranges, document symbols, etc

umpox:
> It would be nice if you ping us in #wg-cody-agent if you start using a new vscode API so that we can ensure it's available in vscode-shim.ts

@olafurpg Is there a way we can make this better and more obvious? Like, could we just lint against importing from `vscode`, and instead import from a single place which changes the import to either `vscode` or the shim depending on the runtime?

olafurpg:
@umpox The ideal state is that the agent is invisible because it's a drop-in replacement for `vscode` imports. We should have enough QA on the editor client side to catch regressions if the VS Code extension starts using new APIs. Eventually, my goal is to have thorough snapshot tests for the agent so you get a failing test if you change the shape of the returned data structures. I have also considered using `Proxy` to throw exceptions when we access a new and un-shimmed API.

dominiccooney:
I'm glad we're getting on the same page about this.

To recap, it sounds like @olafurpg you would endorse these guidelines:

- DO just use the VScode API
- DO add new APIs where the VScode API is lacking (example: a "hover" mode for triggering autocomplete, CC @philipp-spiess)
- DON'T wrap VScode APIs just to give them new names
- VScode people DO feel free to break agent, because the IDE plugins bind to a specific version of agent, so they can update at leisure. Agent can get a Proxy mode going to make understanding these failures cheap. etc.
- Agent/other editor people DO feel free to refactor VScode if it simplifies agent and VScode.

These are not set in stone, they will change as the codebase changes. Now that we have 3.𝜀 editor efforts going, we'll find the areas where we do need some abstraction soon enough.

olafurpg:
Thank you @dominiccooney . That's very clear!

> VScode people DO feel free to break agent

Breaking changes have a non-zero cost for the other editor clients so I encourage you to 1) avoid breaking changes when possible and 2) clearly communicate breaking changes when they are needed. We only have a tiny bit of wiggle room to deal with breaking changes.

It's worth keeping in mind that breaking changes relate to **observable behaviors** of the API, not only changes to TypeScript signatures. The root cause of the JetBrains 3.4.0 regression was the plugin marked `displayText` as `@NotNull` when the TypeScript signature permitted `undefined`. This is a classic example of Hyrum's law https://www.hyrumslaw.com

```
With a sufficient number of users of an API,
it does not matter what you promise in the contract:
all observable behaviors of your system
will be depended on by somebody.
```

We are ramping up QA in the JetBrains plugin to become less vulnerable to these kinds of changes. My ideal outcome is that the VS Code extension can move at full speed and not sacrifice velocity to support other editors. I think we can get there, but we're not 100% there yet.

dominiccooney:
> Breaking changes have a non-zero cost for the other editor clients so I encourage you to 1) avoid breaking changes when possible

We need some guidance about what a break/breaking change is. `@NotNull` sounds like a Java annotation so it's not going to be top of mind for VScode hackers.

> 2) clearly communicate breaking changes when they are needed. 

What's the way to communicate those? It would be great if there's some automation (specific targets run in CI?) and a channel ready (#discuss-cody-clients?)

> This is a classic example of Hyrum's law https://www.hyrumslaw.com/ ...

+1 to the point about observable behaviors. From that point of view getting quality objectively observable would be great. VScode has some work to do here, too. Does JetBrains use agent at main, or does it point to a specific version of it? How do you decide JetBrains is broken/working?

olafurpg:
> `@NotNull` sounds like a Java annotation so it's not going to be top of mind for VScode hackers.

100%, and it shouldn't be a concern for y'all


> What's the way to communicate those?

We use `#wg-cody-agent` for communication about the agent


>  Does JetBrains use agent at main, or does it point to a specific version of it? How do you decide JetBrains is broken/working?

I'm working on addressing all these questions in https://github.com/sourcegraph/sourcegraph/issues/57368 as part of following up on INC-246

The short answer is: currently we use a lot of manual testing against whatever `HEAD` points to in our local clone of cody. I'm planning to change this so that the JetBrains plugin specifies a revision of Cody to build from, and increase our automated tests so that we cover at least core features (onboarding, autocomplete, chat)

olafurpg:
I am also planning to turn on `agent/src/index.test.ts` and update it to include snapshot assertions against observable behavior of the agent. Changes to those snapshot should be considered breaking changes in observable behavior. The idea state is that you should not worry about breaking the agent as long as those tests pass.

philipp-spiess:
Thanks for all the great input here, this is really useful!

One thing that is still not clear to me though is where exactly _is_ the boundary between VS Code and Agent? I have a pretty good understanding about this for Autocomplete since the entry point for this feature is just one interface but for other features like chat, it's really unclear for me. Maybe there is also a way to make the boundary clearer by e.g. forcing the VS Code package to export a bunch of modules for the Agent to be used (as an alternative to allow the Agent to require on _any_ module)?

Different modules might then offer different stability guarantees. E.g. the Autocomplete entry point is of course going to be very stable as it is how the VS Code API also works but we should be free to rewrite the way we do retrieval without causing any issues for Agent.

olafurpg:
@philipp-spiess the agent uses lib/shared `Editor` for chat, but I'm hoping to replace `AgentEditor` with `VSCodeEditor` at some point. Here's a quick regexp to check all the vscode/ APIs we're accessing

```
❯ rg '\.\./vscode'
tsconfig.json
12:  "references": [{ "path": "../lib/shared" }, { "path": "../vscode" }],

src/protocol.ts
7:import { CompletionEvent } from '../../vscode/src/completions/logger'

src/AgentTabGroups.ts
3:import { emptyEvent } from '../../vscode/src/testutils/mocks'

src/agent.ts
10:import { activate } from '../../vscode/src/extension.node'

src/vscode-shim.ts
18:import type { InlineCompletionItemProvider } from '../../vscode/src/completions/inline-completion-item-provider'
27:} from '../../vscode/src/testutils/mocks'
70:} from '../../vscode/src/testutils/mocks'
```

However, remember that the VS Code extension is accessing agent APIs (vscode-shim.ts) in every location where we import from `vscode`.

abeatrix:
Closing this PR based on the discussion above.

PR #1314, opened 2023-10-04
Refactor smart document section and account for more edge cases

philipp-spiess:
This PR refactors the smart ranges logic we have and accounts for a lot more edge cases. More specifically:

- Handle top line modules or namespaces
- Handle very large functions or other top line symbols (e.g huge jest `describe()` at the root) 
  - Note this does not work recursively so nested `describe` blocks might still cause unexpectedly large sectoins
- Better support for languages like Python that do not have closing braces
  - Turns out it's not trivial to convert a folding range to a document section since the folding range won't include the line with the closing brace. We now have a heuristic that no longer overshoots but might in some cases undershoot (although these instances are rare)
- Increase test harness
- Use the smart sections for LSP based graph context

## Test plan

- Check out the new tests
- When you have lsp based graph context enabled, you can check it out by opening the autocomplete trace view and clicking through documents. Some examples:

### Test file will show individual test cases instead of one huge `describe` block

Screenshot 2023-10-04 at 17 04 23

### Large class will show individual methods instead of huge `class` block

Screenshot 2023-10-04 at 17 04 03

### Example used in test cases

Screenshot 2023-10-04 at 17 01 27


vscode/src/editor/utils/document-sections.test.ts
```diff
@@ -0,0 +1,318 @@
...
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
```
abeatrix, line 177: nice!

abeatrix (approved):
work on all my test files! Thanks for cleaning this up 🙇‍♀️

vscode/src/editor/utils/document-sections.ts
```diff
@@ -0,0 +1,239 @@
...
+    line: number
+): Promise<vscode.Selection | undefined> {
+    const smartRanges = await getDocumentSections(doc)
+
+    // Filter to only keep folding ranges that contained nested folding ranges (aka removes nested ranges)
+    // Get the folding range containing the active cursor
+    const range = findRangeByLine(smartRanges, line)
+
+    if (!range) {
+        console.error('No folding range found containing target line')
```
abeatrix, line 43: should we remove this? I was seeing this fired quite often and it can get so noisy in the console.

vscode/src/editor/utils/document-sections.test.ts
```diff
@@ -0,0 +1,318 @@
+import dedent from 'dedent'
+import { describe, expect, it, vi } from 'vitest'
+
+import { document } from '../../completions/test-helpers'
+import { vsCodeMocks } from '../../testutils/mocks'
+import { range } from '../../testutils/textDocument'
+
+import { findRangeByLine, getDocumentSections } from './document-sections'
+
+vi.mock('vscode', () => vsCodeMocks)
```
valerybugakov, line 10: Is this required? We have `vscode/src/testutils/vscode.ts` with `vi.mock('vscode', () => vsCodeMocks)` in `vscode/vite.config.ts` setup files.

vscode/src/editor/utils/document-sections.test.ts
```diff
@@ -0,0 +1,318 @@
...
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
+                // long
```
valerybugakov, line 216: Very long, indeed!

valerybugakov (approved):

vscode/src/editor/utils/document-sections.test.ts
philipp-spiess: Oh we do? Interesting :D Then the answer is probably no? :D

vscode/src/editor/utils/document-sections.test.ts
philipp-spiess: Inflating my commiter stats :)

PR #1315, opened 2023-10-04
Unit test matrix

philipp-spiess:
Refactor the CI job to use a matrix instead, allowing us to also run tests on windows once #1256 lands

## Test plan

Test green

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,9 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner: [ubuntu-latest, windows-latest, macos-latest]
+        runner: [ubuntu, windows, macos]
         node: [16, 20]
-    runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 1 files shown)


valerybugakov (approved):

.github/workflows/ci.yml
```diff
@@ -10,35 +10,18 @@ on:
...
-      - run: pnpm build
-      - run: pnpm run test:unit --run
-
-  test-unit-16:
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu]
+        node: [16, 20]
```
DanTup, line 17: Out of interest, why these two? If this is to track specific versions used by some specific supported versions of VS Code (/latest), it may be worth adding comments so it's clear if/when they should be updated.

philipp-spiess: Answered here: https://github.com/sourcegraph/cody/pull/1256#issuecomment-1752589989 :)

PR #1316, opened 2023-10-04
add: line comment detection to completion latency

abeatrix:
This adds detection of line comments to the getLatency function to handle comment lines differently. The base latency is increased to 1000ms for comment lines.

Also added markdown, plaintext, XML to the low-performance language list. (feel free to remove if you have opinions on this).

NOTE: All changes in this PR is hidden behind the minLatency feature flag

TODO: put it behind a different feature flag for comments, but want to check with the team and see if this makes sense first.

## Test plan

Test cases were added for the new isLineComment function.


vscode/src/completions/latency.ts
```diff
@@ -72,3 +80,31 @@ export function resetLatency(): void {
...
+        return false
+    }
+
+    switch (languageId) {
+        case 'lua':
+            return trimmedLine.startsWith('--')
+        case 'shellscript':
+        case 'perl':
+        case 'r':
+            return trimmedLine.startsWith('#')
```
valerybugakov, line 96: We have the `getLanguageConfig` function in `vscode/src/completions/language.ts`, which returns the `commentStart` fields. 

Would it be nice to have one source of truth? I also use it for snapshot testing of the tree-sitter queries to get comment start symbols. Updating it would benefit other parts of the extension.

valerybugakov (approved):
Nice!

philipp-spiess (changes requested):
Can we use Treesitter to detect wether we're in a comment here or not? We already have this data available for the analytics section anyways, but this would otherwise simply not work for some cases like this:

```ts
/*
CURSOR
```
```ts
/**
 * CURSOR
```

As we'd need to parse _all_ lines above the prefix.

I’m okay with including the current behavior as a fallback but even there it's _very_ brittle, e.g.:

```
const foo = 1234 // CURSOR
```

won't trigger this behavior

valerybugakov:
Great call, @philipp-spiess! It's nice to have this logic as a fallback (for languages where we do not have a tree-sitter grammar) and implement a tree-sitter version. It is similar to our approach for multiline truncation: indentation-based + tree-sitter-based.

abeatrix:
@philipp-spiess @valerybugakov Thanks I can look into using Treesitter!

Before I make any changes there is one thing I want to clarify with you all: do we want to apply latency to all comments, including docstring and comment blocks?

My initial thought was to apply latency only to the comments that start at the beginning of lines because If I already have a comment typed out one line above it, I assume Cody will have a higher chance to predict the next comment than the one above it. Or if we ues Philipp's example `const foo = 1234 // CURSOR`, this seems like something Cody would have enough context where we won't need to apply latency, but I understand this is very subjective and would like to hear your thoughts on this. 


Thanks!

philipp-spiess:
> My initial thought was to apply latency only to the comments that start at the beginning of lines because If I already have a comment typed out one line above it, I assume Cody will have a higher chance to predict the next comment than the one above it. 

Interesting, my own experience diverges from this a bit and I find it less helpful even if already started writing a few lines. What is your experience?

philipp-spiess:
@abeatrix I think the changes are not pushed yet!

abeatrix:
@philipp-spiess Why is it taking so long for git to process my latest push 🤔  It's stuck in `processing update`
image

philipp-spiess:
Oh no

PR #1317, opened 2023-10-04
Adding Smart Selection to FixupRecipe

arafatkatze:
Original PR -> https://github.com/sourcegraph/cody/pull/1083 (Closed because @abeatrix added some really amazing folding range functions to get the resize the selection range)

Related Issues -> https://github.com/sourcegraph/cody/issues/585 and https://github.com/sourcegraph/cody/issues/223 

## Problem Summary
Currently the selection made by users is not always fully inclusive of the entire range where changes might need to be made by Cody to satisfy the request of fixup by the user. Because we're limiting the changes to only the region that was selected by the user, that creates quite a few problems:
1. Empty Selections don't work( shown by  @sqs ) 
image

2. Variable renaming within a function would miss some parts of the function with the original variable name that wasn't selected
4. Bad selection boundaries cause issues with indentation with the changes are applied coz the diffs function doesn't work perfectly in some edge cases 
5. Refactors -> If you only selected the return value then any refactor would miss other parts of the function where changes need to be applied 
6. Imperfect selection boundaries sometimes lead to repetition of code outside the selection range and sometimes it doesn't generate outputs 
7. Hard mode Problem:  Its hard to do a mega refactor where change request is on a small selection but that edits multiple files and other places in the codebase to make it all work together in synergy. 

## Original solution
Originally i leveraged an LLM call to decide the folding range and that was a cool algorithm but it had the latency of an extra LLM call. Now that @abeatrix added some really cool folding range functions I can just leverage them to get a better range for the selection. 

### Before

https://github.com/sourcegraph/cody/assets/11155207/53d63ee7-8650-4c03-b935-dfcb64d38c01

https://github.com/sourcegraph/cody/assets/11155207/5ab30335-6c2d-4142-849d-47234d6f4100

## After

https://drive.google.com/file/d/1RVShpNGiWK4wHJW6N_tHEGOdkzSwxM6d/view?usp=sharing

## Test plan

Tested in my local machine a few times. Works perfectly on edge cases too.

### DIFF
--- a/agent/src/editor.ts
+++ b/agent/src/editor.ts
@@ -4,6 +4,7 @@ import {
     ActiveTextEditor,
     ActiveTextEditorDiagnostic,
     ActiveTextEditorSelection,
+    ActiveTextEditorSelectionRange,
     ActiveTextEditorViewControllers,
     ActiveTextEditorVisibleContent,
     Editor,
@@ -99,7 +100,10 @@ export class AgentEditor implements Editor {
         throw new Error('Method not implemented.')
     }
 
-    public getActiveFixupTextEditorSmartSelection(): Promise<ActiveTextEditorSelection | null> {
+    public getActiveFixupTextEditorSmartSelection(
+        selectionRange: ActiveTextEditorSelectionRange,
+        fileName: string
+    ): Promise<vscode.Range | null> {
         throw new Error('Method not implemented.')
     }
 
--- a/agent/src/editor.ts
+++ b/agent/src/editor.ts
@@ -4,7 +4,6 @@ import {
     ActiveTextEditor,
     ActiveTextEditorDiagnostic,
     ActiveTextEditorSelection,
-    ActiveTextEditorSelectionRange,
     ActiveTextEditorViewControllers,
     ActiveTextEditorVisibleContent,
     Editor,
@@ -97,13 +96,6 @@ export class AgentEditor implements Editor {
     }
 
     public getActiveTextEditorSmartSelection(): Promise<ActiveTextEditorSelection | null> {
-        throw new Error('Method not implemented.')
-    }
-
-    public getFixupRecipeSmartSelection(
-        selectionRange: ActiveTextEditorSelectionRange,
-        fileName: string
-    ): Promise<vscode.Range | null> {
         throw new Error('Method not implemented.')
     }
 
--- a/agent/src/editor.ts
+++ b/agent/src/editor.ts
@@ -99,6 +99,10 @@ export class AgentEditor implements Editor {
         throw new Error('Method not implemented.')
     }
 
+    public getActiveFixupTextEditorSmartSelection(): Promise<ActiveTextEditorSelection | null> {
+        throw new Error('Method not implemented.')
+    }
+
     public getActiveTextEditorSelectionOrVisibleContent(): ActiveTextEditorSelection | null {
         throw new Error('Method not implemented.')
     }
--- a/agent/src/editor.ts
+++ b/agent/src/editor.ts
@@ -100,7 +100,7 @@ export class AgentEditor implements Editor {
         throw new Error('Method not implemented.')
     }
 
-    public getActiveFixupTextEditorSmartSelection(
+    public getFixupRecipeSmartSelection(
         selectionRange: ActiveTextEditorSelectionRange,
         fileName: string
     ): Promise<vscode.Range | null> {
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -1,5 +1,3 @@
-import * as vscode from 'vscode'
-
 import { ContextMessage, getContextMessageWithResponse } from '../../codebase-context/messages'
 import { VsCodeFixupTaskRecipeData } from '../../editor'
 import { IntentClassificationOption } from '../../intent-detector'
@@ -74,15 +72,8 @@ export class Fixup implements Recipe {
 
         // If the intent is 'edit', then potentially modify the fixup task.
         if (intent === 'edit') {
-            const newRangeSmartSelection = (await context.editor.getActiveFixupTextEditorSmartSelection())
-                ?.selectionRange
-            if (newRangeSmartSelection) {
-                const newRange = new vscode.Range(
-                    newRangeSmartSelection.start.line,
-                    0,
-                    newRangeSmartSelection.end.line,
-                    0
-                )
+            const newRange = await context.editor.getActiveFixupTextEditorSmartSelection()
+            if (newRange) {
                 await fixupController.resetSelectionRange(taskId, newRange)
                 // Update the fixup task if the range was modified.
                 finalFixupTask = (await fixupController.getTaskRecipeData(taskId)) || finalFixupTask
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -75,9 +75,9 @@ export class Fixup implements Recipe {
             )
             if (newRange) {
                 await fixupController.resetSelectionRange(taskId, newRange)
-                const newTaskData = await fixupController.getTaskRecipeData(taskId)
-                if (newTaskData) {
-                    fixupTask = newTaskData
+                const fixupTaskwithUpdatedSelectionRange = await fixupController.getTaskRecipeData(taskId)
+                if (fixupTaskwithUpdatedSelectionRange) {
+                    fixupTask = fixupTaskwithUpdatedSelectionRange
                 } else {
                     return null
                 }
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -69,7 +69,7 @@ export class Fixup implements Recipe {
 
         // If the intent is 'edit', then potentially modify the fixup task.
         if (intent === 'edit') {
-            const newRange = await context.editor.getActiveFixupTextEditorSmartSelection(
+            const newRange = await context.editor.getFixupRecipeSmartSelection(
                 fixupTask.selectionRange,
                 fixupTask.fileName
             )
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -53,7 +53,8 @@ export class Fixup implements Recipe {
         }
 
         const intent = await this.fetchRecipeIntent(fixupController, taskId, context)
-        const fixupTask = await fixupController.getTaskRecipeData(taskId, intent === 'edit')
+        const enableSmartSelection = intent === 'edit'
+        const fixupTask = await fixupController.getTaskRecipeData(taskId, enableSmartSelection)
 
         if (!fixupTask || !intent) {
             return null
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -85,7 +85,7 @@ export class Fixup implements Recipe {
                 )
                 await fixupController.resetSelectionRange(taskId, newRange)
                 // Update the fixup task if the range was modified.
-                finalFixupTask = (await fixupController.getTaskRecipeData(taskId)) || originalFixupTask
+                finalFixupTask = (await fixupController.getTaskRecipeData(taskId)) || finalFixupTask
             }
         }
         const promptText = this.getPrompt(finalFixupTask, intent)
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -54,7 +54,7 @@ export class Fixup implements Recipe {
 
         const intent = await this.fetchRecipeIntent(fixupController, taskId, context)
         const enableSmartSelection = intent === 'edit'
-        const fixupTask = await fixupController.getTaskRecipeData(taskId, enableSmartSelection)
+        const fixupTask = await fixupController.getTaskRecipeData(taskId, { enableSmartSelection })
 
         if (!fixupTask || !intent) {
             return null
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -52,40 +52,38 @@ export class Fixup implements Recipe {
             return null
         }
 
-        const originalFixupTask = await fixupController.getTaskRecipeData(taskId)
-        if (!originalFixupTask) {
+        let fixupTask = await fixupController.getTaskRecipeData(taskId)
+        if (!fixupTask) {
             await context.editor.showWarningMessage('Select some code to fixup.')
             return null
         }
 
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
-        if (truncateText(originalFixupTask.selectedText, quarterFileContext * 2) !== originalFixupTask.selectedText) {
+        if (truncateText(fixupTask.selectedText, quarterFileContext * 2) !== fixupTask.selectedText) {
             const msg = "The amount of text selected exceeds Cody's current capacity."
             await context.editor.showWarningMessage(msg)
             return null
         }
 
-        const intent = await this.getIntent(originalFixupTask, context)
+        const intent = await this.getIntent(fixupTask, context)
 
-        // Default to the initial task. It will be overwritten if the intent requires modification.
-        let finalFixupTask = originalFixupTask
         // If the intent is 'edit', then potentially modify the fixup task.
         if (intent === 'edit') {
             const newRange = await context.editor.getActiveFixupTextEditorSmartSelection(
-                finalFixupTask.selectionRange,
-                finalFixupTask.fileName
+                fixupTask.selectionRange,
+                fixupTask.fileName
             )
             if (newRange) {
                 await fixupController.resetSelectionRange(taskId, newRange)
                 const newTaskData = await fixupController.getTaskRecipeData(taskId)
                 if (newTaskData) {
-                    finalFixupTask = newTaskData
+                    fixupTask = newTaskData
                 } else {
                     return null
                 }
             }
         }
-        const promptText = this.getPrompt(finalFixupTask, intent)
+        const promptText = this.getPrompt(fixupTask, intent)
 
         return Promise.resolve(
             new Interaction(
@@ -96,7 +94,7 @@ export class Fixup implements Recipe {
                 {
                     speaker: 'assistant',
                 },
-                this.getContextFromIntent(intent, finalFixupTask, quarterFileContext, context),
+                this.getContextFromIntent(intent, fixupTask, quarterFileContext, context),
                 []
             )
         )
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -59,7 +59,7 @@ export class Fixup implements Recipe {
         }
 
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
-        if (truncateText(fixupTask.selectedText, quarterFileContext * 2) !== fixupTask.selectedText) {
+        if (truncateText(fixupTask.selectedText, MAX_CURRENT_FILE_TOKENS) !== fixupTask.selectedText) {
             const msg = "The amount of text selected exceeds Cody's current capacity."
             await context.editor.showWarningMessage(msg)
             return null
@@ -67,7 +67,7 @@ export class Fixup implements Recipe {
 
         const intent = await this.getIntent(fixupTask, context)
 
-        // If the intent is 'edit', then potentially modify the fixup task.
+        // If the intent is 'edit', then potentially modify the fixup task selection range
         if (intent === 'edit') {
             const newRange = await context.editor.getFixupRecipeSmartSelection(
                 fixupTask.selectionRange,
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -1,5 +1,5 @@
 import { ContextMessage, getContextMessageWithResponse } from '../../codebase-context/messages'
-import { VsCodeFixupTaskRecipeData } from '../../editor'
+import { VsCodeFixupController, VsCodeFixupTaskRecipeData } from '../../editor'
 import { IntentClassificationOption } from '../../intent-detector'
 import { MAX_CURRENT_FILE_TOKENS, MAX_HUMAN_INPUT_TOKENS } from '../../prompt/constants'
 import { populateCodeContextTemplate, populateCurrentEditorDiagnosticsTemplate } from '../../prompt/templates'
@@ -14,7 +14,7 @@ import { Recipe, RecipeContext, RecipeID } from './recipe'
  * This is either provided by the user, or inferred from their instructions
  */
 export type FixupIntent = 'add' | 'edit' | 'document'
-const FixupIntentClassification: IntentClassificationOption<FixupIntent>[] = [
+export const FixupIntentClassification: IntentClassificationOption<FixupIntent>[] = [
     {
         id: 'edit',
         rawCommand: '/edit',
@@ -52,38 +52,15 @@ export class Fixup implements Recipe {
             return null
         }
 
-        let fixupTask = await fixupController.getTaskRecipeData(taskId)
-        if (!fixupTask) {
-            await context.editor.showWarningMessage('Select some code to fixup.')
+        const intent = await this.fetchRecipeIntent(fixupController, taskId, context)
+        const fixupTask = await fixupController.getTaskRecipeData(taskId, intent === 'edit')
+
+        if (!fixupTask || !intent) {
             return null
         }
 
-        const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
-        if (truncateText(fixupTask.selectedText, MAX_CURRENT_FILE_TOKENS) !== fixupTask.selectedText) {
-            const msg = "The amount of text selected exceeds Cody's current capacity."
-            await context.editor.showWarningMessage(msg)
-            return null
-        }
-
-        const intent = await this.getIntent(fixupTask, context)
-
-        // If the intent is 'edit', then potentially modify the fixup task selection range
-        if (intent === 'edit') {
-            const newRange = await context.editor.getFixupRecipeSmartSelection(
-                fixupTask.selectionRange,
-                fixupTask.fileName
-            )
-            if (newRange) {
-                await fixupController.resetSelectionRange(taskId, newRange)
-                const fixupTaskwithUpdatedSelectionRange = await fixupController.getTaskRecipeData(taskId)
-                if (fixupTaskwithUpdatedSelectionRange) {
-                    fixupTask = fixupTaskwithUpdatedSelectionRange
-                } else {
-                    return null
-                }
-            }
-        }
         const promptText = this.getPrompt(fixupTask, intent)
+        const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
 
         return Promise.resolve(
             new Interaction(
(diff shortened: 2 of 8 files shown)


lib/shared/src/chat/recipes/fixup.ts
```diff
@@ -52,21 +54,41 @@ export class Fixup implements Recipe {
...
+            if (newRangeSmartSelection) {
+                const newRange = new vscode.Range(
+                    newRangeSmartSelection.start.line,
+                    0,
+                    newRangeSmartSelection.end.line,
+                    0
+                )
+                await fixupController.resetSelectionRange(taskId, newRange)
+                // Update the fixup task if the range was modified.
+                finalFixupTask = (await fixupController.getTaskRecipeData(taskId)) || originalFixupTask
```
dominiccooney, line 88: When would `fixupController.getTaskRecipeData(taskId)` return nothing, but `originalFixupTask` still be worth pursuing? The task IDs are the same.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,52 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+     */
+    public async getActiveFixupTextEditorSmartSelection(): Promise<ActiveTextEditorSelection | null> {
+        // Get the instance of the active text editor.
+        const activeEditor = this.getActiveTextEditorInstance()
+        if (!activeEditor) {
+            return null
+        }
+
+        const selection = activeEditor.selection
+        if (!selection?.start.line) {
```
dominiccooney, line 185: 0 is falsy, so this means you can't start the selection on line 0. Which will exclude people who did "select all", etc. So don't do this. Just check `if (selection)`

vscode/src/non-stop/FixupController.ts
```diff
@@ -255,6 +255,21 @@ export class FixupController
...
+    /**
+     * OverWrites the selectionRange of the FixupTask and selects new text corresponding to that range
+     */
+    public async resetSelectionRange(taskid: string, newRange: vscode.Range): Promise<void> {
+        const task = this.tasks.get(taskid)
+        if (!task) {
+            return undefined
+        }
+        task.selectionRange = newRange
+        const document = await vscode.workspace.openTextDocument(task.fixupFile.uri)
```
dominiccooney, line 267: The Fixup object hangs onto the original text for applying diffs, but that text should really be a property of each submission to the LLM, and not the Fixup object itself. See the TODO in `getTaskRecipeData`.

The idea is you can create fixup tasks and then they track the ranges they apply to and can re-fetch that original text to respin the LLM if there's conflicting edits, etc.

Because `getTaskRecipeData` sets the original text (and gets the preceding and successive text) you don't need to update `original`. Just update the selection and call `getTaskRecipeData`. If that return undefined then the task is cancelled, etc. and you shouldn't continue with the task (new OR old... it's the same task.)

dominiccooney (changes requested):
Like the direction this is going in but have some suggestions inline.

arafatkatze:
@dominiccooney Thanks a lot for the helpful comments. I fixed all the suggestions.

lib/shared/src/chat/recipes/fixup.ts
```diff
@@ -52,21 +52,34 @@ export class Fixup implements Recipe {
...
+        // Default to the initial task. It will be overwritten if the intent requires modification.
+        let finalFixupTask = originalFixupTask
+
+        // If the intent is 'edit', then potentially modify the fixup task.
+        if (intent === 'edit') {
+            const newRange = await context.editor.getActiveFixupTextEditorSmartSelection()
+            if (newRange) {
+                await fixupController.resetSelectionRange(taskId, newRange)
+                // Update the fixup task if the range was modified.
+                finalFixupTask = (await fixupController.getTaskRecipeData(taskId)) || finalFixupTask
```
dominiccooney, line 79: You don't want `|| finalFixupTask` here. It's the same task. If you're holding onto a reference to it, but the fixup controller has abandoned it, then there's no point continuing with it.

lib/shared/src/chat/recipes/fixup.ts
```diff
@@ -52,21 +52,34 @@ export class Fixup implements Recipe {
...
             await context.editor.showWarningMessage(msg)
             return null
         }
 
-        const intent = await this.getIntent(fixupTask, context)
-        const promptText = this.getPrompt(fixupTask, intent)
+        const intent = await this.getIntent(originalFixupTask, context)
+
+        // Default to the initial task. It will be overwritten if the intent requires modification.
+        let finalFixupTask = originalFixupTask
```
dominiccooney, line 71: This can just be one variable, `fixupTask`. It's the same task, you're just changing the selection range and text associated with it.

lib/shared/src/editor/index.ts
```diff
@@ -158,6 +160,10 @@ export class NoopEditor implements Editor {
         return Promise.resolve(null)
     }
 
+    public getActiveFixupTextEditorSmartSelection(): Promise<vscode.Range | null> {
```
dominiccooney, line 163: It is not JavaScripty to use null as much as this interface does. just make it `vscode.Range | undefined`.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+     * 1. Checks if there's an active editor.
+     * 2. Determines the current selection's start and end line numbers.
+     * 3. For both the start and end line numbers, it tries to find a more encompassing folding range.
+     *    - If a folding range is found, it uses that. Otherwise, it defaults to the original selection.
+     * 4. The function then creates a new selection that starts at the beginning of the expanded start range
+     *    and ends at the end of the expanded end range.
+     * 5. Returns this combined range as the final "smart" selection.
+     *
+     * @returns A Promise that resolves to an `ActiveTextEditorSelection` which represents the combined "smart" selection.
+     *          Returns null if no active editor is found or if the starting line of the selection is not defined.
```
dominiccooney, line 175: When would the start line of the selection not be defined?

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+     * and attempt to expand those positions to encompass more meaningful chunks of code, such as folding regions.
+     *
+     * The function does the following:
+     * 1. Checks if there's an active editor.
+     * 2. Determines the current selection's start and end line numbers.
+     * 3. For both the start and end line numbers, it tries to find a more encompassing folding range.
+     *    - If a folding range is found, it uses that. Otherwise, it defaults to the original selection.
+     * 4. The function then creates a new selection that starts at the beginning of the expanded start range
+     *    and ends at the end of the expanded end range.
+     * 5. Returns this combined range as the final "smart" selection.
```
dominiccooney, line 172: These comments make it sound more complicated than it is.

```suggestion
     * 2. If the selection starts in a folding range, moves the selection start back to the start of that folding range.
     * 3. If the selection ends in a folding range, moves the selection end forward to the end of that folding range.
```

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+        }
+
+        const selection = activeEditor.selection
+        if (!selection) {
+            return null
+        }
+
+        // Try to retrieve a folding range for the start position of the current selection.
+        const activeCursorStartPosition = selection.start.line
+        const foldingRangeStart =
```
dominiccooney, line 191: This isn't a good variable name, because sometimes it's *not* a folding range. Sometimes it is the original selection.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+            return null
+        }
+
+        const selection = activeEditor.selection
+        if (!selection) {
+            return null
+        }
+
+        // Try to retrieve a folding range for the start position of the current selection.
+        const activeCursorStartPosition = selection.start.line
```
dominiccooney, line 190: Improve this variable name; it is a line, not a position.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+        const foldingRangeStart =
+            (await getSmartSelection(activeEditor.document.uri, activeCursorStartPosition)) || selection
+
+        // Similarly, try to retrieve a folding range for the end position of the current selection.
+        const activeCursorEndPosition = selection.end.line
```
dominiccooney, line 195: Same feedback about names above applies to these two.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+        const foldingRangeEnd =
+            (await getSmartSelection(activeEditor.document.uri, activeCursorEndPosition)) || selection
+
+        // Create a new combined selection range that starts from the beginning of the folding rangeat the start position
+        // and ends at the end of the folding range at the end position.
+        const combinedFoldingRange = new vscode.Selection(foldingRangeStart.start.line, 0, foldingRangeEnd.end.line, 0)
```
dominiccooney, line 201: This doesn't look right because you set the end character to 0. So you might make the selection smaller if there's no folding range or if the folding range ends on the same line the selection ends on.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+        const activeSelection = this.createActiveTextEditorSelection(activeEditor, combinedFoldingRange)
+        const newRangeSmartSelection = activeSelection.selectionRange
+        if (newRangeSmartSelection) {
+            return new vscode.Range(newRangeSmartSelection.start.line, 0, newRangeSmartSelection.end.line, 0)
+        }
+        return null
+    }
```
dominiccooney, line 208: Use `return ... ? ... : ...` for succinctness. Also because you used early return null all along it looks odd to vary from that now.

vscode/src/non-stop/FixupController.ts
```diff
@@ -255,6 +255,18 @@ export class FixupController
         return editOk
     }
 
+    /**
+     * OverWrites the selectionRange of the FixupTask and selects new text corresponding to that range
```
dominiccooney, line 259: OverWrites &rarr; Overwrites

In what sense does it "select" new text? Not really. You want to leave the selected text as-is anyway so if there's a task in flight it can detect that the selected text changed so it should respin and not try to merge the result.

vscode/src/non-stop/FixupTask.ts
```diff
@@ -47,4 +47,9 @@ export class FixupTask {
     public get state(): CodyTaskState {
         return this.state_
     }
+
+    // Overwrites a selection range with a newly computed range
+    public overwriteSelectionRange(newRange: vscode.Range): void {
```
dominiccooney, line 52: Gut feel, do this through the fixup controller so if we need to update requests in flight the controller can do it. Who uses this anyway? It looks duplicative of `resetSelectionRange` on the fixup controller. And it has a name which varies from the name on the controller, seems like it is making a distinction, but is it? It seems to be doing exactly the same thing.

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+     * The function does the following:
+     * 1. Checks if there's an active editor.
+     * 2. Determines the current selection's start and end line numbers.
+     * 3. For both the start and end line numbers, it tries to find a more encompassing folding range.
+     *    - If a folding range is found, it uses that. Otherwise, it defaults to the original selection.
+     * 4. The function then creates a new selection that starts at the beginning of the expanded start range
+     *    and ends at the end of the expanded end range.
+     * 5. Returns this combined range as the final "smart" selection.
+     *
+     * @returns A Promise that resolves to an `ActiveTextEditorSelection` which represents the combined "smart" selection.
```
dominiccooney, line 174: It is declared to return vscode.Range, not ActiveTextEditorSelection.

agent/src/editor.ts
```diff
@@ -99,6 +99,10 @@ export class AgentEditor implements Editor {
         throw new Error('Method not implemented.')
     }
 
+    public getActiveFixupTextEditorSmartSelection(): Promise<ActiveTextEditorSelection | null> {
```
dominiccooney, line 102: This design seems weak. What's the difference between this and `getActiveTextEditorSmartSelection`? If it is for a fixup, why does it operate on the text editor's active selection and not a fixup task's selection?

vscode/src/editor/vscode-editor.ts
```diff
@@ -157,6 +157,56 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
...
+     *          Returns null if no active editor is found or if the starting line of the selection is not defined.
+     */
+    public async getActiveFixupTextEditorSmartSelection(): Promise<vscode.Range | null> {
+        // Get the instance of the active text editor.
+        const activeEditor = this.getActiveTextEditorInstance()
+        if (!activeEditor) {
+            return null
+        }
+
+        const selection = activeEditor.selection
```
dominiccooney, line 184: This does not seem right. Fixups have selections. If you're doing something to fixups' selections, you should be doing it to the fixup task, not based on the active text editor.

We want to make a system for doing edits to documents. It's OK to look at the active editor and selection when creating the fixup task, but after that point it is an independent object that is updated to track changes in the editor it was based off of. Dipping into the active editor this way goes against that.

So I would either expand the selection when you're creating the fixup task in the first place, and to be so focused on editing the fixup. I think this is best.

Or if you really want to edit fixups, then edit a fixup... don't dip back into the active editor but use the editor associated with the fixup and ignore that editor's active selection.

dominiccooney (changes requested):
This needs more work. Sorry for the slow reply.

lib/shared/src/chat/recipes/fixup.ts
arafatkatze: So some thoughts for this comment + [this comment](https://github.com/sourcegraph/cody/pull/1317#discussion_r1351508071) 

I had a few different considerations here but most of the absurdity came from `const` and `let` keyword.

If I were to use `const` then the subsequent calls to ` await fixupController.getTaskRecipeData(taskId)` would NOT change the value of `originalFixupTask.selectionRange` verified this in console so I decided to use the original and the final fixup task but you bought up a very important point that it's essentially the same fixup which is true.

So now I have to use the `let` key word but the keyword but the keyword can only be used if I do a reset of the `fixupTask` variable which is what I did in the. current solution has to do this 
 https://github.com/sourcegraph/cody/blob/fc363aeb1fb7006f37875eb73ce41873c9e9b81e/lib/shared/src/chat/recipes/fixup.ts#L78-L82

But that's to pass the nullity check of the original function, It's absurd but I couldn't think of a cleaner way.

lib/shared/src/editor/index.ts
arafatkatze: Totally agree(given the rest of the codebase) but I wanted to maintain consensus with the current file itself so for now i would prefer to keep this and will change it in a separate PR AFTER this is merged

vscode/src/editor/vscode-editor.ts
arafatkatze: Great catch(It's a bad idea to rely on LLms too much), I updated the commentary

vscode/src/editor/vscode-editor.ts
arafatkatze: Good catch, I changed this.

lib/shared/src/chat/recipes/fixup.ts
dominiccooney: If the task is removed from `FixupController.tasks` and `getTaskRecipeData` returns undefined, then it doesn't matter what happens after this—the FixupController is not going to process the result. You are done. You should stop processing the task. Returning undefined or whatever it takes to signal the caller to also stop is fine.

lib/shared/src/chat/recipes/fixup.ts
```diff
@@ -52,21 +52,34 @@ export class Fixup implements Recipe {
...
-        if (!fixupTask) {
+        const originalFixupTask = await fixupController.getTaskRecipeData(taskId)
+        if (!originalFixupTask) {
             await context.editor.showWarningMessage('Select some code to fixup.')
             return null
         }
 
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
-        if (truncateText(fixupTask.selectedText, MAX_CURRENT_FILE_TOKENS) !== fixupTask.selectedText) {
+        if (truncateText(originalFixupTask.selectedText, quarterFileContext * 2) !== originalFixupTask.selectedText) {
```
dominiccooney: Just curious but why did we change this? Is this a fudge factor so that when it broadens, it probably fits the budget?

vscode/src/editor/vscode-editor.ts
arafatkatze: Changed it to `newSelectionStartingPosition` but happy to do something else if you prefer.
Also added a comment too.

vscode/src/editor/vscode-editor.ts
arafatkatze: Great catch. I made the change to include characters too.

vscode/src/non-stop/FixupController.ts
arafatkatze: Great point. I made the change.

vscode/src/non-stop/FixupTask.ts
arafatkatze: This is redundant so i removed it.

vscode/src/editor/vscode-editor.ts
arafatkatze: > This does not seem right. Fixups have selections. If you're doing something to fixups' selections, you should be doing it to the fixup task, not based on the active text editor. We want to make a system for doing edits to documents. It's OK to look at the active editor and selection when creating the fixup task, but after that point it is an independent object that is updated to track changes in the editor it was based off of. Dipping into the active editor this way goes against that. So I would either expand the selection when you're creating the fixup task in the first place, and to be so focused on editing the fixup. I think this is best.

I totally agree I contemplated this idea and wanted to do expansion right at the start of the fixup recipe. Here

https://github.com/sourcegraph/cody/blob/aecdd88b1258d84db5183efdc1053cb9fff34b2b/vscode/src/main.ts#L284-L295

The only problem was that doing the expansion here would render the `add` intent basically untouchable coz we can't have the "no selection" case. 
 https://github.com/sourcegraph/cody/blob/aecdd88b1258d84db5183efdc1053cb9fff34b2b/lib/shared/src/chat/recipes/fixup.ts#L87-L90

That's why I had do it at the place that I did. I also totally understand why you are hesitant about using active editor 
so 
```    public async getFixupRecipeSmartSelection(
        selectionRange: ActiveTextEditorSelectionRange,
        fileName: string
    ): Promise<vscode.Range | null> {
        if (!selectionRange) {
            return null
        }
        const documentUri = vscode.Uri.file(fileName)

        // Retreive the start position of the current selection
        const activeCursorStartPosition = selectionRange.start
        // If we find a new expanded selection positon then we set it as the new start position
        // and if we don't then we fallback to the original selection made by the user
        const newSelectionStartingPosition =
            (await getSmartSelection(documentUri, activeCursorStartPosition.line))?.start || selectionRange.start

        // Retreive the ending line of the current selection
        const activeCursorEndPosition = selectionRange.end
        // If we find a new expanded selection positon then we set it as the new ending position
        // and if we don't then we fallback to the original selection made by the user
        const newSelectionEndingPosition =
            (await getSmartSelection(documentUri, activeCursorEndPosition.line))?.end || selectionRange.end

        // Create a new range that starts from the beginning of the folding range at the start position
        // and ends at the end of the folding range at the end position.
        return new vscode.Range(
            newSelectionStartingPosition.line,
            newSelectionStartingPosition.character,
            newSelectionEndingPosition.line,
            newSelectionEndingPosition.character
        )
    }
```

SO now it doesn't use the active editor at all. I only put this function here coz I just wanted to use `getSmartSelection` and I couldn't think of a better way than this but it doesn't even tough any objects of the activeEditor. I think this is better than before but I am open to new ideas to improve things.

arafatkatze:
@dominiccooney Big thanks for the thorough review! Took your feedback on board and made the necessary tweaks. I realize the initial version wasn’t up to par—sorry about that. After another look and your pointers, I think it's in better shape now. Let me know if anything else jumps out.

lib/shared/src/chat/recipes/fixup.ts
dominiccooney: OK, I understand what you're grappling with here!

I missed that `getTaskRecipeData` isn't returning the FixupTask, but a bag of properties based on it. You need to get the updated bag of properties after you modify the selection of the underlying FixupTask and the types including undefined is a pain.

The design of FixupController/FixupTask is weak at this point. I think we are missing a third kind of object, which is that snapshot of the state when you make a submission to the LLM. Original text should definitely live on that object, for example. But let's NOT try to clean that up now.

We are also missing an opportunity to start intent detection earlier and be more consistent. It's not 100% clear to me if every fixup task routes through this recipe, but maybe intent detection and task modification could/should be part of the fixup controller and the task's life cycle. (Today we have idle, working, ready, applying, fixed, error in CodyTaskState; maybe we can have a setup step where they get intent detection and adjustment.)

It's also clear that some paths want to skip intent detection, and the way that happens is a bit clumsy: IF the secret word `/edit` appears then skip intent detection. Probably because intent detection will be so slow, and it only detects "edit" or "document", which does not sound super high value (but could be if we generalize fixups so they can put their output in other places.) Instead `createTask` could have a hint about what intent detection the caller wants.

This are big clean-ups and I would like @abeatrix 's input/buy in on them too.

Here's what I suggest you do, short term:

- Make `getActiveFixupTextEditorSmartSelection` *not* use the active editor. Using the active editor couples creating the task, and doing this manipulation, in time. That coupling will be fragile (if some intervening Promise, like the intent detector, slows down now it is more likely that the active editor has changed and you're reading the wrong document.) Instead, use the FixupTask and its document to get the updated range and text. Be bold digging into the FixupController and friends and making an API that's nice for this.
- Consider making the intent detector work with the FixupTask and not the bag of data, although if it needs to retrieve the bag of data within the detector that's fine.
- In the code you're modifying here, retrieve the fixup task from the controller.
- If the intent is to edit, update the fixup's task range. That should happen through the controller and not by modifying the task directly because later if we had fixup requests in flight at this point, we might want to stop them, and that would involve the controller. Also we might need to coordinate with the document edit watcher.
- Then, finally, get the bag of task data.

How does that sound?

dominiccooney (changes requested):
OK, thanks for clarifying! I have some suggestions inline.

lib/shared/src/chat/recipes/fixup.ts
abeatrix: > Make getActiveFixupTextEditorSmartSelection not use the active editor 

That sounds good to me. @arafatkatze created `getActiveFixupTextEditorSmartSelection` for Fixup, so I don't think it would break other things that are using the old `getActiveTextEditorSmartSelection` and what you suggested makes sense.

@umpox is the one who added intent detector for Fixup, so he might have more insights to share.

dominiccooney:
> I realize the initial version wasn’t up to par—sorry about that.

Don't be! I like the direction of this change and I'm always happy to discuss. I'm just trying to process how we can use this to make the fixups system easier to use. Sorry for all the loaded footguns we left lying around.

lib/shared/src/chat/recipes/fixup.ts
arafatkatze: @domiccooney sounds great!! Lemme do this tomorrow.

arafatkatze: @domiccooney sounds great!! Lemme do this tomorrow.

arafatkatze: @dominiccooney Here's the response to your comments. 
> Make getActiveFixupTextEditorSmartSelection not use the active editor. Using the active editor couples creating the task, and doing this manipulation, in time. That coupling will be fragile (if some intervening Promise, like the intent detector, slows down now it is more likely that the active editor has changed and you're reading the wrong document.) Instead, use the FixupTask and its document to get the updated range and text. Be bold digging into the FixupController and friends and making an API that's nice for this.

I've eliminated the editor's role entirely and transitioned the entire logic to a private function within fixupController. It's now designed to execute only when enableSmartSelection is truthy, which primarily occurs when the intent is _edit_.

>Consider making the intent detector work with the FixupTask and not the bag of data, although if it needs to retrieve the bag of data within the detector that's fine.

I've integrated it into fixupController to leverage FixupTask directly, bypassing the initial reliance on the bag of data. This seemed more fitting, given the data bag's downstream utility and its limited relevance for intent detection. I trust this aligns with your vision.


>In the code you're modifying here, retrieve the fixup task from the controller.
If the intent is to edit, update the fixup's task range. That should happen through the controller and not by modifying the task directly because later if we had fixup requests in flight at this point, we might want to stop them, and that would involve the controller. Also we might need to coordinate with the document edit watcher.

I've centralized the task range modification within fixupController, ensuring better coordination, especially in potential scenarios where we might need synchronization with the document edit watcher.


>Then, finally, get the bag of task data.  

This has been streamlined to ensure we fetch the data bag only once.

I hope these modifications resonate with your initial suggestions. While I acknowledge further refactoring potential, I'm hoping these changes fit the current PR's scope. If you see areas for further refinement, please guide me accordingly. I'm more than happy to iterate further.

vscode/src/non-stop/FixupController.ts
```diff
@@ -177,6 +182,88 @@ export class FixupController
...
+     * 2. If the selection starts in a folding range, moves the selection start position back to the start of that folding range.
+     * 3. If the selection ends in a folding range, moves the selection end positionforward to the end of that folding range.
+     *
+     * @returns A Promise that resolves to an `vscode.Range` which represents the combined "smart" selection.
+     */
+    private async getFixupTaskSmartSelection(task: FixupTask, selectionRange: vscode.Range): Promise<vscode.Range> {
+        const fileName = task.fixupFile.uri.fsPath
+        const documentUri = vscode.Uri.file(fileName)
+
+        // Retreive the start position of the current selection
```
dominiccooney, line 243:
```suggestion
        // Retrieve the start position of the current selection
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -177,6 +182,88 @@ export class FixupController
...
+        const activeCursorStartPosition = selectionRange.start
+        // If we find a new expanded selection positon then we set it as the new start position
```
dominiccooney, line 245:
```suggestion
        // If we find a new expanded selection position then we set it as the new start position
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -177,6 +182,88 @@ export class FixupController
...
+        // and if we don't then we fallback to the original selection made by the user
+        const newSelectionStartingPosition =
+            (await getSmartSelection(documentUri, activeCursorStartPosition.line))?.start || selectionRange.start
+
+        // Retreive the ending line of the current selection
```
dominiccooney, line 250:
```suggestion
        // Retrieve the ending line of the current selection
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -177,6 +182,88 @@ export class FixupController
...
+        const activeCursorEndPosition = selectionRange.end
+        // If we find a new expanded selection positon then we set it as the new ending position
```
dominiccooney, line 252:
```suggestion
        // If we find a new expanded selection position then we set it as the new ending position
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -312,12 +399,20 @@ export class FixupController
     }
 
     // Called by the non-stop recipe to gather current state for the task.
-    public async getTaskRecipeData(id: string): Promise<VsCodeFixupTaskRecipeData | undefined> {
+    public async getTaskRecipeData(
+        id: string,
+        enableSmartSelection: boolean
```
dominiccooney, line 404: I'm not a fan of boolean parameters because they are mysterious to the reader at callsites (just `true`, etc.) and as soon as you have more than one it is easy to mix them up without the type checker complaining.

What about making this a bag of options with `{enableSmartSelection: boolean}`? Or add an enum with "selection kind" or something?

vscode/src/non-stop/FixupController.ts
```diff
@@ -312,12 +399,20 @@ export class FixupController
...
+    ): Promise<VsCodeFixupTaskRecipeData | undefined> {
         const task = this.tasks.get(id)
         if (!task) {
             return undefined
         }
         const document = await vscode.workspace.openTextDocument(task.fixupFile.uri)
+        if (enableSmartSelection || task.selectionRange) {
```
dominiccooney, line 411:
```suggestion
        if (enableSmartSelection && task.selectionRange) {
```

dominiccooney (approved):
Looks good! Some feedback inline. Definitely need `&&` and not `||` in the condition noted below.

vscode/src/non-stop/FixupController.ts
arafatkatze: Good catch

arafatkatze:
@dominiccooney Thanks a lot for the feedback. I made all the changes and tested things one last time. You can merge this after you take a look.

dominiccooney:
test:e2e passed locally for me on macOS. Will follow up with dev experience about why those integration tests are failing because secrets.

arafatkatze:
> Will follow up with dev experience about why those integration tests are failing because secrets.

@dominiccooney That's expected behaviour: By default, GitHub does not pass secrets to workflows that are triggered from forked repositories. This is a measure to prevent malicious users from creating PRs in open-source projects and using workflows to extract secret values.
My e2e tests have always failed coz of secrets.

[1 week ago ](https://github.com/sourcegraph/cody/actions/runs/6449294349/job/17523335681)
[3 weeks ago](https://github.com/sourcegraph/cody/actions/runs/6257115780/job/17124496404)

dominiccooney:
@arafatkatze Yep, I guessed as much, but I didn't know the workflow for actually merging this because branch protection prevents me from doing that. Seems I create a copy of your PR #1398 to run the tests and then this one picks up the results.

arafatkatze:
@dominiccooney that works too thank you!!

PR #1326, opened 2023-10-05
Autocomplete: LSP light graph context

philipp-spiess:
There are some bugs yet which prevents the `Banan` from being inlined but this is a WIP PR for some changes that should bring down the LSP usage by a _ton_.

1. We no longer rely on document sections to match context. Even with improvements in #1314 this would easily explode to too many symbols.
2. We still leverage cursor navigation based preloading to reduce latencies. 
3. Instead of a range, we use the current line, previous line for context only.
4. I also want to _recurse_ into the current line. My idea is to fix for the banana example that Olaf came up with.
5. The actual LSP calls were also changed. We now call `hover` on everything that looks like a symbol. We recurse the current line. this means for every symbol in this line we also goToDef and run generate the hover tooltip for the first 50 symbols inside the range of the definition of that symbol. That is enough to cover the bananas example.
6. Furthermore this now also awaits the retrieval promise during the autocomplete request. This means that even when the preloading is not finished yet/or you just typed a new character, we will wait for the context to be loaded.
7. Preloading is cached per line including the line text. That means that every change to the line will evict the cache so that we always prioritize the latest data for context.

## Test plan

Test with the infamous bananas example. The timing on the right is with 0 caches so for this example it fits the 100ms context budget but we might need to add some further limits for larger implementations

Screenshot 2023-10-09 at 16 04 13


vscode/src/completions/context/section-observer.test.ts
```diff
@@ -0,0 +1,268 @@
...
+            )
+
+            expect(context[0]).toEqual({
+                content: 'foo\nbar\nfoo',
+                fileName: '/document2.ts',
+            })
+        })
+
+        it('does not include sections that are contained in the prefix/suffix range', async () => {
+            // Visit the first and second section  in document 1
```
dominiccooney, line 232:
```suggestion
            // Visit the first and second section in document 1
```

vscode/src/graph/lsp/languages.ts
```diff
@@ -0,0 +1,169 @@
...
+    'typeof',
+    'var',
+    'void',
+    'while',
+    'with',
+    'yield',
+])
+
+const pythonKeywords = new Set([
+    'False',
```
dominiccooney, line 101: Might make maintenance diffs easier if this is sorted, too.

dominiccooney (commented):
Interesting! Some questions inline.

vscode/src/completions/context/section-observer.test.ts
```diff
@@ -0,0 +1,268 @@
...
+            {
+                onDidChangeTextDocument: (_onDidChangeTextDocument: any) => {
+                    onDidChangeTextDocument = _onDidChangeTextDocument
+                    return disposable
+                },
+            },
+            getDocumentSections
+        )
+        // The section observer loads the document asynchronously, so we wait
+        // for it to finish loading.
```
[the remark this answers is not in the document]
philipp-spiess: Yeah it's mocked here to a `Promise.resolve()`

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,211 @@
...
+                // still fit inside the budget
+                continue
+            }
+            usedContextChars += historyContext.content.length
+            context.push(historyContext)
+        }
+
+        if (sectionGraphContext) {
+            const preciseContexts = hoverContextsToSnippets(
+                sectionGraphContext.filter(context => !overlapsContextRange(URI.parse(context.uri), context.range))
```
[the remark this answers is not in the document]
philipp-spiess: Hah that's an oversight from copy pasting. We can use `string` here instead since it's already the `toString()` value of an URI. After this PR though I want to add a new abstraction to handle this (inspired by your RFC of the context mixer).

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -168,30 +183,72 @@ export class LspLightGraphCache implements vscode.Disposable, GraphContextFetche
...
+        docCache.delete(lineKey)
+    }
+
+    public evictForOtherDocuments(uri: vscode.Uri): void {
+        // eslint-disable-next-line ban/ban
+        this.cache.forEach((_, otherUri) => {
+            if (otherUri === uri.toString()) {
+                return
+            }
+            this.cache.delete(otherUri)
```
dominiccooney, line 250: Does lru-cache work if you modify while iterating? I'm reading the source...

- `forEach` uses `#indexes` which is a like iterator [here](https://github.com/isaacs/node-lru-cache/blob/main/src/index.ts#L1521)
- `#indexes` walks through the `prev` array [here](https://github.com/isaacs/node-lru-cache/blob/main/src/index.ts#L1350)
- `delete` patches up the previous pointer [here](https://github.com/isaacs/node-lru-cache/blob/main/src/index.ts#L2263)

So I think this *does* work, but only because it can reference the dangling item in `#prev` at the (now invalid) index. But doing more than one deletion, or mixing deletions and insertions while iterating, probably would not work AFAICT.

dominiccooney (commented):
Some comments inline.

Looks great.

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    public get(key: GraphCacheParams): Promise<HoverContext[]> | undefined {
+        return this.cache.get(this.toCacheKey(key))
+    }
+
+    public set(key: GraphCacheParams, entry: Promise<HoverContext[]>): void {
+        this.cache.set(this.toCacheKey(key), entry)
+    }
+
+    public delete(key: GraphCacheParams): void {
```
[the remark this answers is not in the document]
philipp-spiess: This is interesting! I plan to first see if this implementation is a useful improvement for CAR and if that's the case we can look into optimizing the cache, what do you think about this approach?

vscode/src/completions/context/lsp-light-graph-cache.ts
philipp-spiess: Ouch I haven't thought about this, I'll fix this, ty!

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+        }
+        this.instance = new LspLightGraphCache(window, workspace, getGraphContextFromRange, sectionObserver)
+        return this.instance
+    }
+
+    private constructor(
+        private window: Pick<typeof vscode.window, 'onDidChangeTextEditorSelection'> = vscode.window,
+        private workspace: Pick<typeof vscode.workspace, 'onDidChangeTextDocument'> = vscode.workspace,
+        private getGraphContextFromRange = defaultGetGraphContextFromRange,
+        private sectionObserver: SectionObserver | null = SectionObserver.createInstance()
```
valerybugakov, line 43: Are these defined only for testing purposes? Does it make sense to outline that as a comment? 

A broader question: I'm curious what folks think about mocking vs. exposing arguments for unit tests only. We have both patterns in the codebase, but my preference usually goes to mocking because it keeps the interface minimal to things required for the production code. 

cc @abeatrix @dominiccooney

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+        const abortController = new CustomAbortController()
+
+        this.lastRequestKey = key
+        this.abortLastRequest = () => abortController.abort()
+
+        const prevLine = Math.max(position.line - 1, 0)
+        const currentLine = position.line
+
+        const [prevLineContext, currentLineContext, sectionHistory] = await Promise.all([
+            this.getLspContextForLine(document, prevLine, 0, abortController.signal),
```
valerybugakov, line 72: Converting the argument list to an object would make understanding the meaning behind `0` easier.

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * When the cursor is moving into a new line, we want to fetch the context for the new line.
+     */
+    private onDidChangeTextEditorSelection(event: vscode.TextEditorSelectionChangeEvent): void {
+        if (!supportedLanguageId(event.textEditor.document.languageId)) {
+            return
+        }
+
+        void this.getContextAtPosition(event.textEditor.document, event.selections[0].active, 0)
```
valerybugakov, line 184: Is this a preloading request? Using `maxChars: 0` for it makes sense, but clarification here would also help.

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+            }
+
+            return contextRange.start.line <= range.startLine && range.endLine <= contextRange.end.line
+        }
+
+        // Allocate up to 40% of the maxChars budget to inlining previous section unless we have no
+        // graph context
+        if (sectionHistory) {
+            const maxCharsForPreviousSections = sectionGraphContext ? maxChars * 0.4 : maxChars
+            for (const historyContext of sectionHistory) {
```
valerybugakov, line 100: Thoughts on naming it `recentlyVisitedSection` or smth similar like you do in tests? I didn't know what the section history and reading tests made it clear. Looking at this loop, I thought we extract graph context for recently visited sections.

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+
+        const abortController = new CustomAbortController()
+
+        this.lastRequestKey = key
+        this.abortLastRequest = () => abortController.abort()
+
+        const prevLine = Math.max(position.line - 1, 0)
+        const currentLine = position.line
+
+        const [prevLineContext, currentLineContext, sectionHistory] = await Promise.all([
```
valerybugakov, line 71: Adding analytics data with the performance of this step would be nice. Based on this, we can consider changing the precise context size.

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+        if (!docCache) {
+            return undefined
+        }
+        docCache.delete(lineKey)
+    }
+
+    public evictForOtherDocuments(uri: vscode.Uri): void {
+        const keysToDelete: string[] = []
+        // eslint-disable-next-line ban/ban
+        this.cache.forEach((_, otherUri) => {
```
valerybugakov, line 247: Need to remove this already from the base eslint-config 😀

vscode/src/completions/context/lsp-light-graph-cache.ts
```diff
@@ -0,0 +1,271 @@
...
+        })
+        for (const key of keysToDelete) {
+            this.cache.delete(key)
+        }
+    }
+}
+
+function hoverContextsToSnippets(contexts: HoverContext[]): SymbolContextSnippet[] {
+    return dedupeWith(contexts.map(hoverContextToSnippets), context =>
+        [context.symbol, context.fileName, context.content].join('\n')
```
valerybugakov, line 261: Noice dedupe key

vscode/src/graph/lsp/languages.ts
```diff
@@ -0,0 +1,169 @@
...
+    'raise',
+    'return',
+    'True',
+    'try',
+    'while',
+    'with',
+    'yield',
+])
+
+export const commonKeywords = new Set([...goKeywords, ...typescriptKeywords, ...pythonKeywords])
```
valerybugakov, line 138: If the LSP-light approach proves helpful, we should replace this with a tree-sitter query. It should be straightforward to extract identifiers.

valerybugakov (approved):
I'm looking forward to testing it locally!

vscode/src/completions/context/lsp-light-graph-cache.ts
philipp-spiess: @valerybugakov No strong opinions. I do think the optional arguments are ugly to read in the implementation but it simplifies test setup a lot and we don't have to worry about restoring mocks or anything.

vscode/src/completions/context/lsp-light-graph-cache.ts
philipp-spiess: Yeah I'll do this as a follow up. The issue is that they might be called outside of a completion lifcycle (preloading) so there's nothing to obviously attach this data to. We do have basic duration logging for context fetching that will be helpful as the first step to understand the complexity of this.

vscode/src/completions/context/lsp-light-graph-cache.ts
philipp-spiess: Done now lol

PR #1329, opened 2023-10-05
refactor: add commands integration test file

abeatrix:
refactor: add commands integration test file & remove recipes integration test file

Removed the recipes.test.ts integration test file. This file contained a single test for the "Explain Code" feature and we have renamed recipes to command.

The new test file for commands include tests for all of our default commands: `/smell`, `/test`, `/explain`, `/doc`

## Test plan

The new test files should pass the integration test.


vscode/test/integration/commands.test.ts
```diff
@@ -0,0 +1,90 @@
+import * as assert from 'assert'
+
+import * as vscode from 'vscode'
+
+import { afterIntegrationTest, beforeIntegrationTest, getFixupTasks, getTranscript, waitUntil } from './helpers'
+
+suite('Commands', function () {
+    this.beforeEach(() => beforeIntegrationTest())
+    this.afterEach(() => afterIntegrationTest())
```
dominiccooney, line 9: No need for the extra closure here.

```suggestion
    this.beforeEach(beforeIntegrationTest)
    this.afterEach(afterIntegrationTest)
```

vscode/test/integration/commands.test.ts
```diff
@@ -0,0 +1,90 @@
...
+    })
+
+    test('Find Code Smells', async () => {
+        // Open Main.java
+        assert.ok(vscode.workspace.workspaceFolders)
+        const mainJavaUri = vscode.Uri.parse(`${vscode.workspace.workspaceFolders[0].uri.toString()}/Main.java`)
+        const textEditor = await vscode.window.showTextDocument(mainJavaUri)
+
+        // Select the "main" method
+        textEditor.selection = new vscode.Selection(5, 0, 7, 0)
```
dominiccooney, line 37: This bit is repetitive, so let's write a helper and call it.

The tests will be easier to read, and if we ever change the contents of Main.java we can update it in one place.

vscode/test/integration/commands.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+        // Run the "/smell" command
+        await vscode.commands.executeCommand('cody.command.smell-code')
+
+        // Check the chat transcript contains markdown
+        const humanMessage = await getTranscript(0)
+        assert.match(humanMessage.displayText || '', /^\/smell/)
+
+        await waitUntil(async () => /^hello from the assistant$/.test((await getTranscript(1)).displayText || ''))
```
dominiccooney, line 46: Might make sense to factor this out, too. It isn't very long but it is a bit tedious to read.

dominiccooney (approved):
Some feedback inline for making these shorter and clearer.

It would be great if we made the LLM response include some output to put in the editor so we could apply those changes and check that they're accurate, because that's an important part of the feature too.

PR #1330, opened 2023-10-05
feat: prevent unhelpful suggestion at empty file end line

abeatrix:
change: prevent completions at empty file end

Proposing to not trigger completions when the cursor is at the start of the last line of a file AND the line above is empty. 

This avoids unwanted completions when finishing up a file, or when it's clear that we do not have enough context to make a helpful suggestion.

### Proposal

Going through this [thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1695856420061639), I noticed we provide a lot of suggestions for new line, but based on the number a lot of them are not very unhelpful and probably can be cut down. One thing that is most obvious to me is the suggestions i've been getting at the end of file. 

Most of the time the suggestions from there are just PURE guesses, here are some examples:

image

image

image

This is why I propose not to trigger completions automatically at this specific position.

However,  this only applies to end lines where the current cursor is at position 0. if there is an indentation at the start, or if this is not an empty line, then we know this is most likely a WIP, and therefore a signal to let us know that this is not the end of the file and we might have context to continue (even though it might not be true, and that's why we still want to trigger suggestion automatically here):
image

We also don't want to prevent suggestions from showing up if the line above is not empty, as that could potentially be source of context, e.g. a comment:
image

Thoughts?

## Test plan

Move your cursor to the end of file.

Also added unit tests


vscode/src/completions/get-inline-completions.ts
```diff
@@ -202,6 +202,14 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
         return null
     }
 
+    // Do not trigger when cusor is at the start of the file ending line, and the line above is empty
+    if (triggerKind !== TriggerKind.Manual && position.line && position.line === document.lineCount - 1) {
```
philipp-spiess, line 206:
```suggestion
    if (triggerKind !== TriggerKind.Manual && position.line !== 0 && position.line === document.lineCount - 1) {
```

Let's be explicit about `0` checks. The inexplicit `if (foo)` should only be used for null and undefined IMO.

philipp-spiess (approved):
Nice! I think this makes sense, I have had similar experiences.

vscode/src/completions/get-inline-completions.ts
```diff
@@ -202,6 +202,14 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
...
+        const lineAbove = Math.max(position.line - 1, 0)
+        if (document.lineAt(lineAbove).isEmptyOrWhitespace && !position.character) {
```
valerybugakov, line 208: Let's add a unit test for this case 👍

valerybugakov (approved):
Agreed, these are usually random/unhelpful completions.

vscode/src/completions/get-inline-completions.ts
abeatrix: @valerybugakov @philipp-spiess Added unit tests for this case (thanks Philipp for the pointers!). Let me know if the cases don't make sense to you.

vscode/src/completions/get-inline-completions-tests/triggers.test.ts
```diff
@@ -179,4 +179,52 @@ describe('[getInlineCompletions] triggers', () => {
             })
         )
     })
+
+    describe('empty line at end of file', () => {
+        it.each(['function foo(){\n console.log()\n}\n\n█', 'const foo = bar\n\n█', 'function foo()\n\n█'])(
+            'does not trigger for empty line with an empty line above %s',
+            async prompt => expect(await getInlineCompletions(params(prompt, [completion`bar`]))).toBeNull()
+        )
```
philipp-spiess, line 187: Is there anything inherently different in the three test cases here? I think it might be enough to only do one test case for this scenario unless they go through different code paths.

vscode/src/completions/get-inline-completions-tests/triggers.test.ts
```diff
@@ -179,4 +179,52 @@ describe('[getInlineCompletions] triggers', () => {
...
+
+        it('trigger completion in an empty text doc', async () => {
+            expect(await getInlineCompletions(params('█', [completion`bar`]))).toEqual<V>({
+                items: [{ insertText: 'bar' }],
+                source: InlineCompletionsResultSource.Network,
+            })
+        })
+
+        it('triggers completion at the start of an empty line where the line above is not empty', async () => {
```
philipp-spiess, line 205: Is there a difference between this test and _any one_ from the `it.each(['function log(foo: string){\n█', '// console.log foo\n█', '// log foo\nconst foo = bar\n█'])(` tests above? Should all of this just be a single test?

vscode/src/completions/get-inline-completions-tests/triggers.test.ts
```diff
@@ -179,4 +179,52 @@ describe('[getInlineCompletions] triggers', () => {
...
+
+        it('trigger multi-line completion at cursor position greater than 0 inside a function', async () => {
+            const requests: CompletionParameters[] = []
+            await getInlineCompletions(
+                params('function bubbleSort() {\n    █', [], {
+                    onNetworkRequest(request) {
+                        requests.push(request)
+                    },
+                })
+            )
```
philipp-spiess, line 227: This test has no assertion

valerybugakov (approved):
Great stuff, thank you! 🙌

philipp-spiess (commented):
Some feedback on the tests. I think we might be able to remove a few of these examples since they are all testing the same behavior (and thus they only add maintainance burder and slow down the suite without providing signal). Also one test has no assertion it seems.

vscode/src/completions/get-inline-completions-tests/triggers.test.ts
abeatrix: 🤡

vscode/src/completions/get-inline-completions-tests/triggers.test.ts
philipp-spiess: Maybe three tests: one for it triggers when there's a new line before, one for it doesn't if there are two, one for it does trigger when there is nothing in the document

abeatrix: Thanks for the suggestions @philipp-spiess 🙇‍♀️ 
I've split them into four different tests:
- doesn't trigger when the line above is empty
- does trigger when the line above is not empty
- does trigger when there is nothing in the document
- does trigger when the character position is greater than 0 on an empty line

philipp-spiess: :chef-kiss:

PR #1331, opened 2023-10-05
add: context summary logging to recipe execution

abeatrix:
RE:  https://sourcegraph.slack.com/archives/C05AGQYD528/p1696542919289939

PR to add logging of context summary when executing recipes, including:
- Number of embeddings used
- Number of local files used

This will allow tracking usage of different context sources.

image

## Test plan

█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:custom-prompt:executed {"contextSummary":{"embeddings":0,"local":1}}

- Use the `/explain` command

█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:chat-question:executed {"contextSummary":{"embeddings":0,"local":1}}

- Ask Cody a question in side bar chat

█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:inline-chat:executed {"contextSummary":{"embeddings":8,"local":2}}

- Ask Cody a question in inline chat

█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:fixup:executed {"contextSummary":{"embeddings":3,"local":2}}

- Use the `/edit` or `/doc` command

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -329,6 +329,11 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         this.isMessageInProgress = true
         this.transcript.addInteraction(interaction)
 
+        const contextSummary = {
+            embeddings: 0,
+            local: 0,
+        }
+
         // Check whether or not to connect to LLM backend for responses
         // Ex: performing fuzzy / context-search does not require responses from LLM backend
         switch (recipeId) {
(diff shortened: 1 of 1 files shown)


kelsey-brown:
Thanks Bee! Clarifying question - for all relevant recipe/command events, the contextSummary details will be included on the  `CodyVSCodeExtension:recipe:custom-prompt:executed` event, as opposed to the event for the specific recipe/command (e.g. for your example above, the `CodyVSCodeExtension:command:explain:executed` event)? If so, is there a way to get those details on the explain event itself, or is that overly complicated 😅

abeatrix:
@kelsey-brown we can add that, but atm none of the default commands require embeddings, so they will always only use local resources. In this case, do you still think we should add them?

Right now if you run the default commands, we log if they use CodebaseContext or not to determine if it is using embeddings:

`█ logEvent: CodyVSCodeExtension:command:explain:executed {"mode":"ask","useCodebaseContex":false,"useShellCommand":false}`

kelsey-brown:
Ahh ok that makes sense! No need to add it then if the default commands don't use embeddings. Thanks for explaining.

kelsey-brown (approved):

PR #1332, opened 2023-10-06
benchmark suite: first pass at adding a golang dataset

slimsag:
Adds a few basic Go tests to the benchmark suite.

## Test plan

This adds tests.

### DIFF
--- a/vscode/test/benchmark/datasets/golang/basic-func-call-closed/test.sh
+++ b/vscode/test/benchmark/datasets/golang/basic-func-call-closed/test.sh
@@ -1,0 +1,1 @@
+go build main.go
--- a/vscode/test/benchmark/datasets/golang/basic-func-call-closed/test.sh
+++ b/vscode/test/benchmark/datasets/golang/basic-func-call-closed/test.sh
@@ -1,1 +1,0 @@
-go build main.go
--- a/vscode/test/benchmark/datasets/golang/call-with-struct-open/main.go
+++ b/vscode/test/benchmark/datasets/golang/call-with-struct-open/main.go
@@ -1,0 +1,7 @@
+package main
+
+func main() {
+	dir, err := Open█
+	_ = dir
+    _ = err
+}
--- a/vscode/test/benchmark/datasets/golang/open-file/config.json
+++ b/vscode/test/benchmark/datasets/golang/open-file/config.json
@@ -1,0 +1,5 @@
+{
+  "entryFile": "main.go",
+  "solutionFile": "solution.go",
+  "testCommand": "go build main.go"
+}
--- a/vscode/test/benchmark/datasets/golang/open-file/main.go
+++ b/vscode/test/benchmark/datasets/golang/open-file/main.go
@@ -1,0 +1,9 @@
+package main
+
+import "os"
+
+func main() {
+    f, err := os.█
+    _ = f
+    _ = err
+}
--- a/vscode/test/benchmark/datasets/Dockerfile
+++ b/vscode/test/benchmark/datasets/Dockerfile
@@ -3,5 +3,6 @@ FROM node:20-slim
 # Add test dependencies
 RUN npm install -g ts-node
 RUN apt update && apt install python3 -y
+RUN apt update && apt install golang -y
 
 WORKDIR /app
--- a/vscode/test/benchmark/datasets/golang/basic-func-call-closed/config.json
+++ b/vscode/test/benchmark/datasets/golang/basic-func-call-closed/config.json
@@ -1,0 +1,7 @@
+{
+  "entryFile": "main.go",
+  "openFiles": [],
+  "closedFiles": ["context.go"],
+  "solutionFile": "solution.go",
+  "testCommand": "go build main.go context.go"
+}
--- a/vscode/test/benchmark/datasets/golang/basic-func-call-closed/context.go
+++ b/vscode/test/benchmark/datasets/golang/basic-func-call-closed/context.go
@@ -1,0 +1,15 @@
+package main
+
+import "os"
+
+func OpenAbsolutePath(path string) (*os.File, error) {
+	return os.Open(path)
+}
+
+func OpenAbsoluteDir(dir string) *os.File {
+	d, err := os.Open(dir)
+	if err != nil {
+		panic("TO DO: handle error")
+	}
+	return d
+}
--- a/vscode/test/benchmark/datasets/golang/basic-func-call-closed/main.go
+++ b/vscode/test/benchmark/datasets/golang/basic-func-call-closed/main.go
@@ -1,0 +1,8 @@
+package main
+
+import "os"
+
+func main() {
+    dir := OpenAbs█
+    _ = dir
+}
--- a/vscode/test/benchmark/datasets/golang/basic-func-call-closed/solution.go
+++ b/vscode/test/benchmark/datasets/golang/basic-func-call-closed/solution.go
@@ -1,0 +1,6 @@
+package main
+
+func main() {
+	dir := OpenAbsoluteDir("/path/to/dir")
+	_ = dir
+}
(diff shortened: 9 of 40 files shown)


slimsag:
@umpox I wasn't actually able to get this to _pass_, interesting the `test.sh` script exits with the success exit code `0` and passes.. but the test suite calls it a failure for some reason? Am I missing something? When I run the `test.sh` script in this PR inside the tmp directory, it passes as I'd expect

```
Exit code:   0
Done

🔴 - /var/folders/j4/fs11plcs72s2kkrtnyf7mgy80000gq/T/cody-benchmark-d6Ht8l
@@ -6 +6 @@ func main() {
-    f, err := os.◆
+    f, err := os.Open("data.txt")

┌──────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────┐
│                   (index)                    │                              sourcegraph.cody-ai                              │
├──────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ ./test/benchmark/datasets/golang - open-file │ '🔴 - /var/folders/j4/fs11plcs72s2kkrtnyf7mgy80000gq/T/cody-benchmark-d6Ht8l' │
└──────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────┘

sourcegraph.cody-ai:
Pass: 0
Fail: 1
No Change: 0
Pass rate: 0.00%
Pass rate (excluding No Change): 0.00%
```

vscode/test/benchmark/datasets/golang/open-file/config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "entryFile": "main.go",
+  "openFiles": ["main.go"],
+  "closedFiles": [],
+  "solutionFile": "solution.go",
+  "testFile": "test.sh",
+  "testCommand": "test.sh"
```
slimsag, line 7: It'd be nice to not need a `test.sh` script here. Ideally I can run `go build main.go` where `main.go` is the file Cody did autocomplete in - just to check the file builds.

But because the test suite insists on adding the `testFile` to the command (https://github.com/sourcegraph/cody/pull/1265#discussion_r1348108534) and the `testFile` must be different than the `entryFile`, the command would end up being an illegal `go build` command so I had to use a dumb `test.sh` script to workaround it

umpox:
@slimsag Strange that this didn't pass, an exit code of `0` should definitely be a success. I'll make the test command spit out logs so we can debug this easier

umpox:
@slimsag I've merged in some main changes from the branch and simplified the config of this test following your suggestion. Re-run and LMK what you think! Works nicely now:

image

taras-yemets:
> ... interesting the test.sh script exits with the success exit code 0 and passes.. but the test suite calls it a failure for some reason? Am I missing something?

@slimsag, I've been getting a similar error. For me, the reason was not running a Docker service. 
```
  error: Error: Command failed: docker run --mount src="/var/folders/jp/bw890l7d4b549xjmgrb7c67h0000gn/T/cody-benchmark-eZBjKD",target=/app,type=bind cody-benchmark-harness ts-node test.ts
  docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
  See 'docker run --help'.
  
      at ChildProcess.exithandler (node:child_process:419:12)
      at ChildProcess.emit (node:events:512:28)
      at maybeClose (node:internal/child_process:1098:16)
      at Socket.<anonymous> (node:internal/child_process:456:11)
      at Socket.emit (node:events:512:28)
      at Pipe.<anonymous> (node:net:334:12) {
    code: 125,
    killed: false,
    signal: null,
    cmd: 'docker run --mount src="/var/folders/jp/bw890l7d4b549xjmgrb7c67h0000gn/T/cody-benchmark-eZBjKD",target=/app,type=bind cody-benchmark-harness ts-node test.ts',
    stdout: '',
    stderr: 'docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.\n' +
      "See 'docker run --help'.\n"
  }
  ```
I could catch it by adding additional logging to https://github.com/sourcegraph/cody/blob/b77915e7181b8ad8fc53f26c0dc0e8ab166ebd60/vscode/test/benchmark/evaluate-test-case.ts#L20

**Comment by umpox** [comment]
@taras-yemets Yep definitely, I've added additional logs here for when the test errors. This should hopefully be easier to debug now!

**Comment by slimsag** [comment]
@umpox Updated and added a few more tests, PTAL.

Currently the pass rate for these is not great but I think these are practical tests that reflect meaningful real-world use cases.. and won't rely on e.g. trained knowledge of existing APIs.

```
Pass: 1
Fail: 6
No Completion: 1
Average Time to Completion*: 1322ms
Median Time to Completion*: 1372ms
Pass Rate: 12.50%
Adjusted Pass Rate**: 14.29%
```

**Comment by umpox** [comment]
@slimsag This is interesting! This is the first dataset I've seen where there is a significant difference between us and Copilot

Cody default (Claude instant):
image

However, Cody with `starcoder-16b`...
```
  "cody.autocomplete.advanced.provider": "unstable-fireworks",
  "cody.autocomplete.advanced.model": "starcoder-16b",
```
image

File: vscode/test/benchmark/datasets/golang/basic-func-call-closed/test.sh [hunk 2]
```diff
@@ -0,0 +1 @@
+go build main.go
```
umpox, line 1: we can remove the `test.sh` files now right?

vscode/test/benchmark/datasets/golang/call-with-struct-open/main.go
```diff
@@ -0,0 +1,7 @@
+package main
+
+func main() {
+	dir, err := Open█
```
umpox, line 4: This is one case where I see Cody sometimes make a correct single line completion like `dir, err := Open(OpenOpt{`, but obviously the test fails because we want more than that.

We need to solve this somehow, ideally in a generic way that works for all examples. 

The simplest option would be just to iterate: generate a completion - if the test failed, generate another completion and so on. This would let us build more complex cases, but would be very wasteful for cases where it's likely Cody will fail. We'd basically have to set a completion limit and then bail out if we hit that.

Alternatively maybe this is still important to measure against - ideally Cody gets the correct solution in a single completion here.

umpox (approved):

taras-yemets (approved):

vscode/test/benchmark/datasets/golang/open-file/main.go
```diff
@@ -0,0 +1,9 @@
+package main
+
+import "os"
+
+func main() {
+    f, err := os.█
```
taras-yemets, line 6: What completion do we expect here? There are no hints around (or maybe there are, but don't know Go much) about what should go after `os.`. I'd say it can be anything from `"os"` package returning `(interface {}, error)`.

taras-yemets: Maybe we can create a helper function `openFile` and trigger completion there to see if Cody uses `"os"` correctly.

vscode/test/benchmark/datasets/golang/call-with-struct-open/main.go
slimsag: IMO Cody giving an incomplete single-line completion is not good and is a product bug. It puts the user into a mode where they're like "OK cody got that line right, now let me move to the next line and wait.... now let me move to the next line and wait..." - it's really a case where the user wanted a multi-line completion in my view.

vscode/test/benchmark/datasets/golang/open-file/main.go
slimsag: Check this search: https://sourcegraph.com/search?q=context%3Aglobal+f%2C+err+%3A%3D+os.&patternType=standard&sm=1&groupBy=repo

Pretty much any Go developer would look at `f, err` and the `os.` part and think: they're trying to open a file or create a file. Both are valid solutions - Cody just needs to suggest either

PR #1335, opened 2023-10-06
Use vscode-uri for the mock/shim Uri

DanTup:
This uses a proxy over `vscode-uri`'s `URI` class, with a fallback to the `Utils` class as a replacement for `Uri`. This should be compatible and avoids copying the `Uri` implementation from VS Code.

It's not possible to use the proxy as a type (so you must use `vscode_types.Uri` for files that aren't importing `vscode`), but you _can_ use it in `instanceof`, and TS also understands that it has all the methods from the standard `vscode_types.Uri` class.

## Test plan

Tests with `pnpm test:unit`. Includes new tests to ensure static methods, instance members and `instanceof`s work as expected.


vscode/src/testutils/mocks.ts
```diff
@@ -15,16 +15,21 @@ import type {
...
-// in the future when it becomes a drop-in replacement.
-import { Uri } from './vscode/uri'
+// eslint-disable-next-line @typescript-eslint/no-extraneous-class
+export class Uri {}
+// eslint-disable-next-line guard-for-in
+for (const p in UriUtils) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+    ;(Uri as any)[p] = (UriUtils as any)[p]
+}
+Object.setPrototypeOf(Uri, URI)
```
DanTup, line 27: @olafurpg I don't like all the ignores here, but I couldn't find a cleaner way. Also unsure if this belongs here or in shims?

olafurpg: I don't mind the ignores in this file or vscode-shim since there's enough magic going on already. The most important goal is that we have full fidelity with the behavior of `vscode.Uri`.

olafurpg (approved):
This is a fantastic improvement! LGTM 👍🏻 Thank you @DanTup for iterating on this so we could support instanceof checks. The final solution is a huge improvement, very nice to delete 3.3k loc 🤩

Issue #1341, opened 2023-10-08
VScode: Chat scrolls to top for each new message


dominiccooney:
Every time I input a new chat message, the scoll jumps to the top of the chat history.

Cody: 0.14.0
Version: 1.83.0 (user setup)
Commit: e7e037083ff4455cf320e344325dacb480062c3c
Date: 2023-10-03T16:12:16.321Z
Electron: 25.8.4
ElectronBuildId: 24154031
Chromium: 114.0.5735.289
Node.js: 18.15.0
V8: 11.4.183.29-electron.0
OS: Windows_NT x64 10.0.22621

arafatkatze:
Yeah I faced this too all the time

PR #1224, opened 2023-09-28
Add back support and CI tests for Node 16

philipp-spiess:
Closes #1191

## Test plan

- New tests pass 🤞

Issue #1191, opened 2023-09-26
Autocomplete: Browser build not working


tgaochn:
Autocomplete cannot be triggered in code-server ([code-server](https://github.com/coder/code-server/tree/main) is a web-based vscode).
Nothing shows up in the code and "Cody autocomplete trace view". 
The other functions seem fine in code-server.
Autocomplete works fine in my local vscode. 
![image](https://github.com/sourcegraph/cody/assets/22850831/cfc14b35-d1e7-4978-8b28-17fb320d418a)

jdorfman:
@tgaochn thanks for reporting. 

Is this happening with similar functioning extensions on code-server or just Cody?

philipp-spiess:
@tgaochn Heya! Thanks for reporting this.

Do you know what vs code version is used on the server setup? Did you see the loading indicator spin to indicate a request is being made? Additionally, can you take a look at the logs and send us anything that might pop up there? [How to access Cody logs](https://docs.sourcegraph.com/cody/capabilities#access-autocomplete-logs).

tgaochn:
> @tgaochn thanks for reporting.
> 
> Is this happening with similar functioning extensions on code-server or just Cody?

copilot works fine in code-server:
![image](https://github.com/sourcegraph/cody/assets/22850831/e24cfca4-00bf-4eb9-9e25-f848241c2e1f)

philipp-spiess:
@tgaochn Thanks! I was able to repro the issue locally too:

```
ERR lines(...).findLast is not a function: TypeError: lines(...).findLast is not a function
	at getPrevNonEmptyLine (/Users/philipp/.local/share/code-server/extensions/sourcegraph.cody-ai-0.12.2-universal/dist/extension.node.js:71913:68)
	at getCurrentDocContext (/Users/philipp/.local/share/code-server/extensions/sourcegraph.cody-ai-0.12.2-universal/dist/extension.node.js:72568:28)
	at InlineCompletionItemProvider.provideInlineCompletionItems (/Users/philipp/.local/share/code-server/extensions/sourcegraph.cody-ai-0.12.2-universal/dist/extension.node.js:73658:24)
	at pe.provideInlineCompletions (/opt/homebrew/Cellar/code-server/4.16.1/libexec/lib/vscode/out/vs/workbench/api/node/extensionHostProcess.js:101:53406)
[...]

PR #1343, opened 2023-10-08
Fixing the cody chat jump to the top on initiating a new chat

arafatkatze:
Fixing the Cody chat jump to the top on initiating a new chat

Solves [Issue Link](https://github.com/sourcegraph/cody/issues/1341)
The bug was first introduced in [PR #1224](https://github.com/sourcegraph/cody/pull/1224).
For a detailed discussion on resolving the original issue, see [Issue Link](https://github.com/sourcegraph/cody/issues/1191).

We can't remove the extra utility function because older versions of Node don't support it. This bug was surprising, as the utility function first creates a shallow copy before reversing the array. This still causes unintended side effects. Now, I've replaced it with a loop iterating from the end of the array, which resolves the issue. 

CC @dominiccooney @philipp-spiess 

## Test plan

There is not functionality/logic change here and all the tests pass already.
I've tested the changes on my machine and they work as expected.

### DIFF
--- a/lib/ui/src/chat/Transcript.tsx
+++ b/lib/ui/src/chat/Transcript.tsx
@@ -202,7 +202,10 @@ export const Transcript: React.FunctionComponent<
 })
 
 function findLastIndex<T>(array: T[], predicate: (value: T) => boolean): number {
-    const arr = array.slice()
-    arr.reverse()
-    return arr.findIndex(predicate)
+    for (let i = array.length - 1; i >= 0; i--) {
+        if (predicate(array[i])) {
+            return i
+        }
+    }
+    return -1
 }


philipp-spiess (approved):
> This still causes unintended side effects. 

Huh, do you have an example of this?

Regardless of that, the new code looks better (since it has less time and space complexity) so I’m happy to stamp. 

> Test plan

Do you mind adding some steps of how you tested this so it's easy for others to reproduce?

philipp-spiess:
(Looking into why CI doesn't run properly for the forked repo here 😬 )

arafatkatze:
Thanks a lot for the merge @philipp-spiess 

> This still causes unintended side effects. Huh, do you have an example of this?

Apologies for improper wording. What I meant was:  _This bug was surprising, as the utility function first creates a shallow copy before reversing the array. I was thinking that because we were just reversing the shallow copy and applying operations on it. The whole thing would be safe but this still causes the unintended side effect which cause the chat to jump up_

> Do you mind adding some steps of how you tested this so it's easy for others to reproduce?

The original code when run in the debugger would cause the chat to jump to the top every time you send a request and the new code doesn't. I suppose you tested this already so the PR is good. Happy to add more description if other people want to test this too.

PR #1346, opened 2023-10-09
Agent: mirror runtime behavior for `instanceof URI`

olafurpg:
Previously, `vscode.Uri.parse(...) instanceof URI` returned true with vscode-shim and false in VS Code itself. This PR fixes the problem by creating a wrapper `Uri` classes that implemements the same API as `vscode.Uri` but delegates all underlying logic to vscode-uri `URI`.

## Test plan

See updated tests.

### DIFF
--- a/agent/src/vscode-shim.test.ts
+++ b/agent/src/vscode-shim.test.ts
@@ -1,8 +1,9 @@
 import assert from 'assert'
 import * as path from 'path'
 
-import { describe, it } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import * as vscode from 'vscode'
+import { URI } from 'vscode-uri'
 
 describe('vscode-shim', () => {
     describe('vscode.Uri', () => {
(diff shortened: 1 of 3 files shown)


vscode/src/testutils/uri.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    public with(change: {
+        scheme?: string
+        authority?: string
+        path?: string
+        query?: string
+        fragment?: string
+    }): Uri {
+        return Uri.from({
```
DanTup, line 92: Any reason not to use `this.uri.with()` (and the same for `toJson()`)?

DanTup: nm, this is returning custom Uri, not the base Uri :-)

agent/src/vscode-shim.test.ts
```diff
@@ -30,7 +31,15 @@ describe('vscode-shim', () => {
...
         it('instanceof can be used', () => {
+            // eslint-disable-next-line @typescript-eslint/no-extraneous-class
+            class Qux {}
+
             assert.ok(vscode.Uri.parse('http://example.org/one/two') instanceof vscode.Uri)
+            expect(new Qux() instanceof vscode.Uri).toBe(false)
+
+            expect(vscode.Uri.parse('http://example.org/one/two') instanceof Qux).toBe(false)
+
+            expect(vscode.Uri.parse('http://example.org/one/two') instanceof URI).toBe(false)
```
DanTup, line 42: Probably should also test `vscode.Uri.parse().with()` does the same. If we had blindly called `with()` on the internal `URI` it would return an `instanceof URI` and not the wrapper, so we should test for that.

olafurpg: Done

DanTup (approved):

PR #404, opened 2023-07-27
Add naive html escaping for inline chat.

phuff:
This handles issue #169 .
## Test plan
I tested manually with responses in the form of a paragraph.  
image

PR #1349, opened 2023-10-09
fix: remove HTML escaping in inline chat markdown

abeatrix:
Close: https://github.com/sourcegraph/cody/issues/1350 https://github.com/sourcegraph/sourcegraph/issues/56644

Issue caused by the change in  https://github.com/sourcegraph/cody/pull/404 where `<` and `>` are manually replaced by the `naiveHtmlEscape method`, which is being removed in this PR.

The naiveHtmlEscape method was replacing < and > characters with escaped version of the characters from the markdown text before rendering. This prevented markdown syntax like <tags> from rendering correctly by the `vscode.MarkdownString` method which has the `supportHtml` turns enabled.

Removed the naiveHtmlEscape method call so markdown text renders as intended.

### Before

`<` and `>` are replaced with ` &lt` and `&gt`

image

image

### After

`<` and `>` are displayed correctly:

image

image

## Test plan

1. Start an inline chat for line 310 to 325 https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/Chat.tsx?L310-325
2. Ask Cody: improve variable names and show me the improved code
3. You should see the `<` and `>` being displayed correctly

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -389,6 +389,8 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 
 ### Fixed
 
+- Fixed HTML escaping in inline chat markdown.
+
 ### Changed
 
 ## [0.4.1]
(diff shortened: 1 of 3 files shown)


dominiccooney (approved):
LGTM

We should add some regression tests for this to the integration or e2e suite.

abeatrix:
@dominiccooney thanks for reminding me about the test! Just added a new e2e test for inline chat to track regression.

PR #1351, opened 2023-10-09
refactor: split min latency flags

abeatrix:
refactor: replace cody autocomplete minimum latency flag with user and language based latency flags

Replaced the CodyAutocompleteMinimumLatency feature flag with two new flags:

- CodyAutocompleteUserBasedLatency: Enables minimum latency for Cody autocomplete. Works the same as the old `CodyAutocompleteMinimumLatency` flag but rename it to make sure the old flag will not be used in older versions. 
- CodyAutocompleteLangBasedLatency: Enables language-specific minimum latency for Cody autocomplete for low-performance languages ONLY

The getLatency function now checks both new flags and calculates latency accordingly.

This allows more granular control over autocomplete latency based on user and language.

## Test plan

Unit tests are updated.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -135,10 +135,13 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
 
         // We start feature flag requests early so that we have a high chance of getting a response
         // before we need it.
-        const [isIncreasedDebounceTimeEnabledPromise, minimumLatencyFlagsPromise, syntacticTriggersPromise] = [
+        const [isIncreasedDebounceTimeEnabledPromise, syntacticTriggersPromise] = [
             featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteIncreasedDebounceTimeEnabled),
-            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteMinimumLatency),
             featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteSyntacticTriggers),
+        ]
+        const minLatencyFlagsPromises = [
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteUserBasedLatency),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLangBasedLatency),
         ]
 
         const tracer = this.config.tracer ? createTracerForInvocation(this.config.tracer) : undefined
@@ -254,9 +257,10 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             // latency so that we don't show a result before the user has paused typing for a brief
             // moment.
             if (result.source !== InlineCompletionsResultSource.LastCandidate) {
-                const minimumLatencyFlag = await minimumLatencyFlagsPromise
-                if (triggerKind === TriggerKind.Automatic && minimumLatencyFlag) {
+                const [userBasedLatencyFlag, lowPerformanceOnlyLatencyFlag] = await Promise.all(minLatencyFlagsPromises)
+                if (triggerKind === TriggerKind.Automatic && userBasedLatencyFlag) {
                     const minimumLatency = getLatency(
+                        lowPerformanceOnlyLatencyFlag,
                         this.config.providerConfig.identifier,
                         document.uri.fsPath,
                         document.languageId,
--- a/vscode/src/completions/latency.test.ts
+++ b/vscode/src/completions/latency.test.ts
@@ -17,26 +17,26 @@ describe('getLatency', () => {
         const languageId = undefined
 
         // start with default high latency for unsupported lang with default user latency added
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
         // gradually increasing latency after 5 rejected suggestions
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
         // gradually increasing latency after 5 rejected suggestions
-        expect(getLatency(provider, fileName, languageId)).toBe(1200)
-        expect(getLatency(provider, fileName, languageId)).toBe(1400)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1200)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1400)
         // after the suggestion was accepted, user latency resets to 0, using baseline only
         resetLatency()
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
         resetLatency()
         // next rejection doesn't change user latency until 5 rejected
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1000)
-        expect(getLatency(provider, fileName, languageId)).toBe(1200)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1000)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(1200)
     })
 
     it('returns gradually increasing latency up to max for CSS on anthropic provider when suggestions are rejected', () => {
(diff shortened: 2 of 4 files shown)


vscode/src/completions/latency.ts
```diff
@@ -33,23 +33,35 @@ let userMetrics = {
 // Start when the last 5 suggestions were not accepted
 // Increment latency by 200ms linearly up to max latency
 // Reset every 5 minutes, or on file change, or on accepting a suggestion
-export function getLatency(provider: string, fsPath: string, languageId?: string, nodeType?: string): number {
+export function getLatency(
+    lowPerformanceLanguagesOnly: boolean,
```
dominiccooney, line 37: Boolean parameters are hard to read at call sites, and once you have more than one, it is easy to mix them up but get no type error to detect the mistake.

Could you make this an enum and give it a type?

Alternatively you could take an optional set of languages you check against, and let the caller pass in the set of low performance languages.

vscode/src/completions/latency.ts
```diff
@@ -33,23 +33,35 @@ let userMetrics = {
...
     // set base latency based on provider and low performance languages or comments when available
     let baseline = provider === 'anthropic' ? 0 : defaultLatency.baseline
+
     const isLowPerformance = languageId && lowPerformanceLanguageIds.has(languageId)
     const isComment = nodeType === 'comment'
     if (!languageId || isLowPerformance || isComment) {
         baseline = defaultLatency.lowPerformance
     }
 
+    // Do not add latency when feature flag for low performance languages only is enabled and the current language is not low performance
```
dominiccooney, line 52: Let's move this up to after line 46, because the bit about setting the baseline for comments isn't because of the early return.

vscode/src/completions/latency.ts
```diff
@@ -33,23 +33,35 @@ let userMetrics = {
...
+
     const timestamp = Date.now()
     if (!userMetrics.sessionTimestamp) {
         userMetrics.sessionTimestamp = timestamp
     }
 
     const elapsed = timestamp - userMetrics.sessionTimestamp
     // reset metrics and timer after 5 minutes or file change
-    if (elapsed >= 5 * 60 * 1000 || userMetrics.fsPath !== fsPath) {
+    if (elapsed >= 5 * 60 * 1000 || (userMetrics.fsPath && userMetrics.fsPath !== fsPath)) {
```
dominiccooney, line 64: Why did this change? People are passing empty strings around? It doesn't seem related.

vscode/src/completions/latency.test.ts
```diff
@@ -167,41 +167,77 @@ describe('getLatency', () => {
...
         vi.advanceTimersByTime(3 * 60 * 1000)
-        expect(getLatency(provider, newFileName, languageId)).toBe(600)
+        expect(getLatency(false, provider, newFileName, languageId)).toBe(600)
         // reset latency on accepted suggestion
         resetLatency()
-        expect(getLatency(provider, newFileName, languageId)).toBe(400)
+        expect(getLatency(false, provider, newFileName, languageId)).toBe(400)
+    })
+
+    it('returns 0 if unsupport flag feature flag is enabled and the current language is not low performance', () => {
```
dominiccooney, line 208:
```suggestion
    it('returns 0 if unsupported flag feature flag is enabled and the current language is not low performance', () => {
```

vscode/src/completions/latency.test.ts
```diff
@@ -167,41 +167,77 @@ describe('getLatency', () => {
...
+        const provider = 'anthropic'
+        const fileName = 'foo/bar/test.go'
+        const languageId = 'go'
+
+        // start with default high latency for unsupported lang with default user latency added
```
dominiccooney, line 213: Because this returns 0 it looks like *low* latency, not *high* latency.

vscode/src/completions/latency.test.ts
dominiccooney: Also below.

vscode/src/completions/latency.test.ts
```diff
@@ -167,41 +167,77 @@ describe('getLatency', () => {
...
+        expect(getLatency(true, provider, fileName, languageId)).toBe(0)
+        expect(getLatency(true, provider, fileName, languageId)).toBe(0)
+    })
+
+    it('returns increased latency if unsupport flag feature flag is enabled and the current language is not low performance', () => {
+        const provider = 'anthropic'
+        const fileName = 'foo/bar/test'
+        const languageId = 'css'
+
+        // start with default high latency for low performance lang with default user latency added
```
dominiccooney, line 224: These tests would be more readable if they started with something like:

```
expect(lowPerformanceLanguages.has(languageId)).toBe(true)
```

so the reader can tell that, yes, 'css' is a low performance language. And if CSS ever becomes high performance, this test will fail at a line that exactly pinpoints the problem.

Also above.

vscode/src/completions/latency.test.ts
```diff
@@ -167,41 +167,77 @@ describe('getLatency', () => {
...
+        const fileName = 'foo/bar/test.go'
+        const languageId = 'go'
+
+        // start with default high latency for unsupported lang with default user latency added
+        expect(getLatency(true, provider, fileName, languageId)).toBe(0)
+        expect(getLatency(true, provider, fileName, languageId)).toBe(0)
+        expect(getLatency(true, provider, fileName, languageId)).toBe(0)
+    })
+
+    it('returns increased latency if unsupport flag feature flag is enabled and the current language is not low performance', () => {
```
dominiccooney, line 219: The comment here and on line 224 ("for low performance lang") don't agree. CSS is... or is not... a low performance language. This is confusing.

`unsupport` &Rarr; `unsupported` again.

dominiccooney (changes requested):
Some comments and questions inline.

vscode/src/completions/latency.test.ts
```diff
@@ -136,29 +136,29 @@ describe('getLatency', () => {
         const languageId = 'typescript'
 
         // start with default baseline latency with low performance and user latency added
-        expect(getLatency(provider, fileName, languageId)).toBe(400)
-        expect(getLatency(provider, fileName, languageId)).toBe(400)
-        expect(getLatency(provider, fileName, languageId)).toBe(400)
-        expect(getLatency(provider, fileName, languageId)).toBe(400)
-        expect(getLatency(provider, fileName, languageId)).toBe(400)
+        expect(getLatency(false, provider, fileName, languageId)).toBe(400)
```
philipp-spiess, line 139: I wonder if we should add a third flag for this behavior where starcoder starts with a base latency. I want to make sure it's not that behavior that is causing the regression in prod (I know we ran a test for this before but 😬)

philipp-spiess (commented):
Would it make more sense to have two feature flags that turn on their respective features individually instead of having this dependency where `cody-autocomplete-lang-latency` requires `cody-autocomplete-user-latency` to be on in order to work?

vscode/src/completions/latency.test.ts
abeatrix: Yea that's fair! I will add that

abeatrix:
@dominiccooney @philipp-spiess thanks for taking the time to review the PR. I've made the changes based on your valuable input,added a new feature flag for provider based, and made sure they all work independently.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -135,11 +135,15 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+        const [isIncreasedDebounceTimeEnabledPromise, syntacticTriggersPromise] = [
             featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteIncreasedDebounceTimeEnabled),
-            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteMinimumLatency),
             featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteSyntacticTriggers),
         ]
+        const minLatencyFlagsPromises = [
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteUserLatency),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLanguageLatency),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteProviderLatency),
+        ]
```
dominiccooney, line 146: This code is correct, but it is fragile because you have to get the order here to always match the order on line 261. (Sigh, GitHub doesn't let me make a suggestion that spans far enough...) Because you're awaiting them all, what about:
```suggestion
        const minLatencyFlagsPromises = {
            user: featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteUserLatency),
            language: featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLanguageLatency),
            provider: featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteProviderLatency),
        }
```

...continued below.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -254,9 +258,18 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
-                if (triggerKind === TriggerKind.Automatic && minimumLatencyFlag) {
+                const [userLatencyFlag, languageLatencyFlag, providerLatencyFlag] =
+                    await Promise.all(minLatencyFlagsPromises)
+                const latencyFeatureFlags: LatencyFeatureFlags = {
+                    user: userLatencyFlag,
+                    language: languageLatencyFlag,
+                    provider: providerLatencyFlag,
+                }
+
+                const isMinLatencyEnabled = userLatencyFlag || languageLatencyFlag || providerLatencyFlag
```
dominiccooney, line 269: ...continued from above.
```suggestion
                const latencyFeatureFlags: LatencyFeatureFlags = {
                    user: await minLatencyFlagsPromises.user,
                    language: await minLatencyFlagsPromises.language,
                    provider: await minLatencyFlagsPromises.provider,
                }

                const isMinLatencyEnabled = latencyFeatureFlags.user || latencyFeatureFlags.language || latencyFeatureFlags.provider
```

vscode/src/completions/latency.test.ts
```diff
@@ -11,197 +11,294 @@ describe('getLatency', () => {
...
+        expect(getLatency(featureFlags, provider, fileName, languageId)).toBe(1200)
     })
 
     it('returns gradually increasing latency up to max for CSS on anthropic provider when suggestions are rejected', () => {
         const provider = 'anthropic'
         const fileName = 'foo/bar/test.css'
+
+        // css is a low performance language
         const languageId = 'css'
+        expect(lowPerformanceLanguageIds.has(languageId)).toBe(true)
```
dominiccooney, line 53: Thank you!

vscode/src/completions/latency.test.ts
```diff
@@ -11,197 +11,294 @@ describe('getLatency', () => {
         resetLatency()
     })
 
+    const featureFlags = {
+        user: true,
```
dominiccooney, line 15: Let's add some tests where one, but not all, of these are true?

PR #1355, opened 2023-10-10
Autocomplete: Implement client-side timeouts

philipp-spiess:
This PR adds client side timeouts as discussed in [this Slack thread](https://sourcegraph.slack.com/archives/C05497E9MDW/p1696521645079319).

The reason for this are the following:

Our current server-side backend does request retrying for a fixed number of times. For Autocomplete however, latency is of upmost importance so we don't really want to retry requests for a long time. In some cases, we see requests being retried for up to 30 minutes which is just putting additional pressure on the downstream processes without providing a great UX. Since timeouts are highly client specific, we decided that adding those to the client is the best approach for now. All of our middleware respect aborted requests so all we need to do is to start a race between a timeout and abort the request if that is hit. 

**Note:** This PR is best reviewed [with whitespace changes ignored](https://github.com/sourcegraph/cody/pull/1355/files?w=1).

## Test plan

- Modify the code to have a timeout below 100ms
- Trigger an autocomplete request
- Observe that a timeout error is thrown

### DIFF
--- a/vscode/src/completions/client.ts
+++ b/vscode/src/completions/client.ts
@@ -11,12 +11,15 @@ import {
     isAbortError,
     NetworkError,
     RateLimitError,
+    TimeoutError,
     TracedError,
 } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
 
 import { fetch } from '../fetch'
 
-export type CodeCompletionsParams = Omit<CompletionParameters, 'fast'>
+import { forkSignal } from './utils'
+
+export type CodeCompletionsParams = Omit<CompletionParameters, 'fast'> & { timeoutMs: number }
 
 export interface CodeCompletionsClient {
     complete(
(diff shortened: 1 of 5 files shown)


vscode/src/completions/client.ts
```diff
@@ -35,122 +38,142 @@ export function createClient(config: CompletionsClientConfig, logger?: Completio
...
-            const log = logger?.startCompletion(params)
+    function completeWithTimeout(
+        params: CodeCompletionsParams,
+        onPartialResponse?: (incompleteResponse: CompletionResponse) => void,
+        signal?: AbortSignal
+    ): Promise<CompletionResponse> {
+        const abortController = signal ? forkSignal(signal) : new AbortController()
+        return Promise.race([
+            complete(params, onPartialResponse, abortController.signal),
+            createTimeout(params.timeoutMs).finally(() => {
```
eseliger, line 49: can we make sure that these are not logged as errors in sentry?

eseliger (approved):

vscode/src/completions/client.ts
eseliger: If we expect this to be low volume that’s fine 🙂

philipp-spiess: Haha yeah if this happens often we have to revisit this approach anyways 🫡

PR #1357, opened 2023-10-10
fix: log failures during fixup apply and respin

abeatrix:
fix: log failures during fixup apply and respin

This adds additional logging around fixup failures:

- Log when applying a fixup fails
- Log when respinning a fixup task, including the respin count

## Test plan

You can see the log events in the output channel for all events when running a fixup task.


dominiccooney (approved):
Awesome, thank you!

Issue #1049, opened 2023-09-13
Introduce exclusion patterns for files that shouldn't be sent over the network


beyang:
Cody should exclude .env files and other files that may contain secrets from being sent over the network.

One potential solution here is to exclude files that aren't part of the Git repository.

Customer requests:
* https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1694539385641899
* https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1694591949182719?thread_ts=1694539385.641899&cid=C04MSD3DP5L

jdorfman:
More requests: https://github.com/sourcegraph/cody/discussions/14

abeatrix:
10/18 RFC create, pending for review - https://docs.google.com/document/d/15XJnQ-qa-Y3SM2eltuKfuESmgYYfx7JhEdi2447YF_0/edit?usp=sharing

beyang:
@DanTup @abeatrix what's the latest status on this? Still to be done?

DanTup:
I think Beatrix was/is working on supporting multi-folder workspaces but I don't know if it's complete yet. The current PR is at https://github.com/sourcegraph/cody/pull/1382 but looks like it has a bunch of merge conflicts now that will need resolving.

abeatrix:
The work was done in the linked PR, including working with milti root repo and nest repo, and was waiting for Cody strat team to confirm but then became stale because it was implemented before simple chat was introduced and doesn't work with bfg yet. We will have to pick it back up after GA as we discussed in one of our product meetings @kalanchan

PR #1358, opened 2023-10-10
feat: allow configuring file exclusions

abeatrix:
Close: https://github.com/sourcegraph/cody/issues/1049

This change allows configuring file exclusions via the new  `'cody.excludeFiles'` configuration option.

Files can now be excluded from fetched context before they are sent to LLM to prevent sharing sensitive information from a known location, e.g. directory, file name, etc.

### Examples

configuration: cody.excludeFiles = [".json", ".env", "path/to/my/file.js", "/dirName/"]

files that will be excluded:
- anything with `.json` in the file name, e.g. `package.json`, `cody.json` etc
- anything with `.env` in the file name, e.g. `.env`, `vscode/.env`, etc
- the exact file `path/to/my/file.js`, e.g. `path/to/my/file.js`, or `new/path/to/my/file.js` etc
- all files under the `/dirName/` directory, e.g. `path/to/dirName/file.js`, `path/dirName/file.js` etc

I wanted to do the filtering in codebase-context/index.ts, but then we will need to do the filtering again when we combine them with context from the editor, so for now I think it makes more sense to do it at the recipe level. Open to other ideas and suggestions though!

## Test plan

WIP

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -298,7 +298,7 @@ export class Agent extends MessageHandler {
         const client = await createClient({
             initialTranscript: this.oldClient?.transcript,
             editor: new AgentEditor(this),
-            config: { ...config, useContext: 'embeddings', experimentalLocalSymbols: false },
+            config: { ...config, useContext: 'embeddings', experimentalLocalSymbols: false, excludeFiles: [] },
             setMessageInProgress: messageInProgress => {
                 this.notify('chat/updateMessageInProgress', messageInProgress)
             },
--- a/cli/src/client/context.ts
+++ b/cli/src/client/context.ts
@@ -23,7 +23,7 @@ export async function createCodebaseContext(
     const embeddingsSearch = repoId && !isError(repoId) ? new SourcegraphEmbeddingsSearchClient(client, repoId) : null
 
     const codebaseContext = new CodebaseContext(
-        { useContext: contextType, serverEndpoint, experimentalLocalSymbols: false },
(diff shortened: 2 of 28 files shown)


abeatrix:
@tjdevries would this approach work for neovim/agent?

philipp-spiess:
@abeatrix Some random ideas:

- Instead of a config option, have you considered adding support for a `.codyignore` file? This would allow for more fine-grained control as every sub-folder could have it's own `.codyignore` and users are familiar on how to set this up from git
- This absolutely _needs_ to work for Autocomplete too. Maybe we can wrap the VS Code document opening APIs so we can be 100% sure this is always being use whenever we open documents for context? 
- My personal hunch is that we should use glob rules like gitignore for maximum flexibility. In your example that would mean having to use `*.json` instead of `.json`.
- We should probably always ignore `.env` files 😬

abeatrix:
> Instead of a config option, have you considered adding support for a .codyignore file? This would allow for more fine-grained control as every sub-folder could have it's own .codyignore and users are familiar on how to set this up from git

I did, but because we will need to support other clients, including web, I was thinking config option might be the best option. Let me move this conversation to Slack and check with them to see if that's viable.

> My personal hunch is that we should use glob rules like gitignore for maximum flexibility. In your example that would mean having to use *.json instead of .json

Cool, I can look into that. Thanks Philipp!

philipp-spiess:
> Since agent is not using the vscode-editor.ts atm, this would not work for other clients I presume? I'll check with them on Slack.

Agent is using vs code apis to read files so if we wrap the VS Code API (e.g. `openSafeDocument`) it should work alright 🤔

PR #1361, opened 2023-10-11
Autocomplete: collect node types without pasted completion

valerybugakov:
## Context

Via exploring data in BigQuery, I found that the definition of the `at_cursor` node types is inconsistent. I used the multiline trigger position for multiline suggestions and the completion start for singleline suggestions. Meaning that for singleline completions, we were getting data about the first suggested node. 

### Example 1

```ts
const value = █
```

And with the inserted completion:

```ts
const value = "hello"
```

We were getting `atCursor: 'string'` instead of `program` as per the initial AST. The second one has "more" data because tree-sitter does not understand that it's a variable declaration until the completion is inserted.

### Example 2

```ts
function getData(value█
```

And with the inserted completion:

```ts
function getData(value) {
```

We were getting `atCursor: ')'` (from the start of the inserted completion) instead of `identifier`, which is unhelpful for analysis. We care more about the prefix nodes than those generated by LLM.

## Changes

- Use the same `at_custor` position for single/multi-line completions (the column before the cursor). 
- Collect `nodeTypes` for both tree states: without the pasted completion and with it. 
    - The tree without the completion gives us AST patterns to optimize against before the completion request is made. 
    - The tree with the pasted completion gives us more information about the completion intent. E..g., from the example above, we will know that completion was inserted into the variable declaration.

## Test plan

Updated unit tests

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/analytics.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/analytics.test.ts
@@ -51,6 +51,12 @@ describe('[getInlineCompletions] completion event', () => {
                   "greatGrandparent": "program",
                   "parent": "statement_block",
                 },
+                "nodeTypesWithCompletion": {
+                  "atCursor": "{",
+                  "grandparent": "function_declaration",
+                  "greatGrandparent": "program",
+                  "parent": "statement_block",
+                },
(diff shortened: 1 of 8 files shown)


philipp-spiess (approved):
Mr Burns Excellent

PR #1364, opened 2023-10-11
Autocomplete: Remove codegen provider

philipp-spiess:
This seems to no longer be relevant but I'll defer to @cbart / @chwarwick / @taras-yemets 

## Test plan

👀


chwarwick (approved):
We can always bring it back if needed.

PR #1368, opened 2023-10-11
Autocomplete: prevent negative `position.character` value

valerybugakov:
## Context

Fixes the runtime error caused by the negative `position.character` value.

## Test plan

Unit tests


philipp-spiess (approved):
Nice!

PR #1374, opened 2023-10-12
VScode: Add popups for all embeddings conditions

dominiccooney:
Dotcom/App users can always click on the chat input status indicator and see the state of embeddings in a popup.

Caveats/limitations:
- There's no progress indicator while app is indexing. You need to wait and hit the "Reload" button.
- Only for dotcom/App users however this will be fixed in a follow up PR.
- Various existing quirks of local embeddings remain for now. For example, your git repository must have a remote with a slash in the name.

Part of #632 

## Test plan

This does not add automated tests yet. That is in a follow-up change, I need to update the existing tests to use simplified onboarding first.

UI check: `pnpm storybook` and examine the App-less Onboarding stories.

Manual tests:

*A directory that's not a git repository:*

```
mkdir /tmp/project
cd /tmp/project
echo hello > main.c
code .
```

Open `main.c`, open Cody, click on the chat input status indicator:

Screenshot 2023-10-12 at 14 59 15

*A git repository that does not have a remote:*

```
git init .
```

Cmd-Shift-P, Reload Window (we do not watch for repos changing status in this way.) Chat status indicator should be as above.

*A git repository with a remote:*

Ensure App is not running.

```
git remote add origin git@host.example:cool/project.git
```

Cmd-Shift-P, Reload Window (we do not watch for repos changing status in this way.) Chat status indicator:

Screenshot 2023-10-12 at 15 04 26

*A git repository that is indexed with app:*

```
git add main.c
git commit -am 'first commit'
```

Install app if it is not installed. Click Open App and go through the repository indexing flow. If there are existing repositories click Add a repository.

You can force repo indexing by right clicking on App UI, Inspect Element, `window.location = '/api/console'` in the Console, and run this GraphQL mutation:

```
mutation addrepo {
  scheduleRepositoriesForEmbedding(repoNames: ["host.example:cool/project"], force: true) {
    alwaysNil
  }
}
```

Note there is no *.git* on the repo name. Check that there is a job with state "COMPLETED":

```
query arewethereyet {
  repository(name: "host.example:cool/project") {
    name
    id
    embeddingJobs(first: 2) {
      nodes {
        state
        revision {
          oid
        }
        startedAt
        stats {
          filesSkipped
          filesScheduled
          filesEmbedded
        }
      }
    }
  }
}
```

In VScode, open popup and click Reload, status indicator:

Screenshot 2023-10-12 at 15 14 56

*git repository indexed by dotcom:*

```
cd /tmp
git clone git@github.com:facebook/react.git --depth=1
```

In VScode, File, Open Folder, `/tmp/react` and trust the authors. Chat input status:

Screenshot 2023-10-12 at 15 18 07

Repeat the above looking at different themes.

### DIFF
--- a/lib/shared/src/chat/context.ts
+++ b/lib/shared/src/chat/context.ts
@@ -4,6 +4,7 @@ import { ActiveTextEditorSelectionRange } from '../editor'
 export interface ChatContextStatus {
     mode?: ConfigurationUseContext
     connection?: boolean
+    embeddingsEndpoint?: string
     codebase?: string
     filePath?: string
     selectionRange?: ActiveTextEditorSelectionRange
--- a/lib/shared/src/codebase-context/index.ts
+++ b/lib/shared/src/codebase-context/index.ts
@@ -95,6 +95,10 @@ export class CodebaseContext {
         return !!this.embeddings
     }
 
+    public get embeddingsEndpoint(): string | undefined {
+        return this.embeddings?.endpoint
+    }
+
     public getEmbeddingSearchErrors(): string {
         return this.embeddingResultsError.trim()
     }
--- a/lib/shared/src/embeddings/client.ts
+++ b/lib/shared/src/embeddings/client.ts
@@ -9,6 +9,10 @@ export class SourcegraphEmbeddingsSearchClient implements EmbeddingsSearch {
         private web: boolean = false
     ) {}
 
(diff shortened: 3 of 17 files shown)


toolmantim (approved):
This is fantastic! I popped a commit on to switch the repo button to a proper VS Code button, so it has all the right hover states etc.

PR #1376, opened 2023-10-12
Fixup: Improve error handling

umpox:
## Description

Improves how we handle errors for fixups:
- Errors from the LLM are no longer considered valid code that can be applied, so we don't get weird <span>error message</span> kinda code being added.
- Instead of discarding the fixup immediately on error, we show the error code lens.
- When the error code lens is clicked, it now shows an error modal in the editor, rather than just focusing the chat which will be empty.

### Before

https://github.com/sourcegraph/cody/assets/9516420/ca27207b-e947-40dd-9669-352c90bfada1

### After

https://github.com/sourcegraph/cody/assets/9516420/46d8fc7c-d7fa-4fa8-a715-899c0351bbde

## Test plan

Tested locally
- Error from LLM
- Forced error from controller (respins)
- Checked happy paths work as usual


umpox:
@toolmantim Would appreciate your 👀  on this UX flow.

Right now we treat most errors from Cody as valid fixups, which means we can apply some invalid code sometimes.

This changes it so we have an actual error state.

This thing I wasn't sure about is: Is this visible enough? We could also show a notification immediately after the error happens, so users don't just miss the error.

abeatrix (changes requested):
Love the pop up on the code lense click!

I found an error that is not being handled:

1. Right now recipe level error is not being handled and the task will continue to spin indefinitely: 

image


2. Fixup with no updates from Cody is not returning error?

image

We should probably update the action items for fixup task with errors in the tree view as well? 

image

@toolmantim might have some updates for this but here is the old figma design for fixup for reference: https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=2799-13359&mode=design&t=8KXSY8PA3YuFzIoj-0

umpox:
@abeatrix Thanks for testing, sorry didn't get round to this - fixing now!

abeatrix (approved):
Thanks for fixing this!

Confirmed:
- /doc and /edit works as before.
- show error properly when selected code excesses limit
image
- fixup without suggestion retruns error as well

toolmantim:
Nice!

toolmantim:
@umpox I can't think of a better option than putting the error message in the confirm like you've done. The only other option I could think of were a native VS Code notification, but that should trigger as soon as it happens, and you probably still want something on click. I think confirm is fine for now, and we can improve as we go.

PR #1380, opened 2023-10-12
Autocomplete: Track inserted completions

philipp-spiess:
Adds tracking for inserted completion after 30 seconds, 2 minutes, 5 minutes, 10 minutes.

This also replaces our levenshtein comparator since I found a bug in the current implementation (see the added test case for more context).

## Test plan

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:persistence:present {"id":"eeae5ebe-11d2-4263-ad4e-e5a9ae2f25c4","afterSec":30,"isMostlyUnchanged":true,"difference":0.2571428571428571,"lineCount":1,"charCount":26}
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:persistence:present {"id":"eeae5ebe-11d2-4263-ad4e-e5a9ae2f25c4","afterSec":120,"isMostlyUnchanged":true,"difference":0.2571428571428571,"lineCount":1,"charCount":26}
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:persistence:present {"id":"eeae5ebe-11d2-4263-ad4e-e5a9ae2f25c4","afterSec":300,"isMostlyUnchanged":true,"difference":0.2571428571428571,"lineCount":1,"charCount":26}
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:persistence:present {"id":"eeae5ebe-11d2-4263-ad4e-e5a9ae2f25c4","afterSec":600,"isMostlyUnchanged":false,"difference":0.6285714285714286,"lineCount":1,"charCount":20}
```


vscode/src/completions/persistence-tracker.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        if (measureTimeoutsIndex < MEASURE_TIMEOUTS.length - 1) {
+            this.enqueueMeasure(trackedCompletion, measureTimeoutsIndex + 1)
+            return
+        }
+
+        this.trackedCompletions.delete(trackedCompletion.id)
+    }
+
+    private onDidChangeTextDocument(event: vscode.TextDocumentChangeEvent): void {
```
arafatkatze, line 101: Such an interesting PR. Would some sort of event debouncing be helpful here? Coz the trackerd completions keep growing and then this gets triggered on every single button press by the user?

vscode/src/completions/persistence-tracker.ts
```diff
@@ -0,0 +1,125 @@
...
+        for (const [_, trackedCompletion] of this.trackedCompletions) {
+            if (trackedCompletion.document.uri.toString() !== event.document.uri.toString()) {
```
dominiccooney, line 103: File renames will bite you if you do this tracking for a long time.

vscode/src/completions/persistence-tracker.ts
dominiccooney: Caveat, I haven't studied the dimensions of what's happening here in detail. But based on experience with fixups, yeah, when things can fire per keystroke it is worth sweating some of these details.

You should:

- Track document renames, and other lifecycle so you can keep track of things, throw old ones out, etc. (Fixups actually doesn't throw old ones away well yet...)
- Bucket the tracked completions by document (map keyed on document identifier to set or whatever) so you're lifting the document URI comparison to the top level and doing it once, instead of for every completion on every edit. And when you are updating ranges, it's the smallest possible set of ranges. (And when a document goes away or changes URI you can drop or move the entire bucket of changes in O(1).)
- Updating the ranges themselves is fast. But it would be a nice improvement to let you pass through `event.contentChanges` without those extra allocations. (... if you don't have time you could lift the contentChanges.map outside of the loop. IIRC `updateRangeMultipleChanges` says it is destructive but it because it sorts the array, which is harmless in this case. Check me on that though.)

In terms of event debouncing, it is theoretically fine to buffer up edits and replay them later. But it is not worth the bookkeeping, and introduces the problem of having to "flush" that buffer when you need accurate state. So just make it fast and then do it on every change.

vscode/src/completions/persistence-tracker.ts
philipp-spiess: Yeah I had this as a shower thought today too and will need to handle it 👍 I remember that you pointed me to a place in Cody where we already do this, though!

vscode/src/completions/persistence-tracker.ts
philipp-spiess: This is great feedback, thanks @dominiccooney 🙇

philipp-spiess: Updated the implementation to address the issues. :) I can't pass `event.contentChanges` directly since it's marked as readonly and we actually change it so going with the hoisting approach. I’m also handling deletions and renames now. 

I also found a bug in our levenshtein implementation 😬 Will wrap this up on Monday though I want to add a test and fix CI but if you could review the implementation in the meantime this would be helpful 🙂

vscode/src/completions/persistence-tracker.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    private onDidDeleteFiles(event: vscode.FileDeleteEvent): void {
+        for (const uri of event.files) {
+            this.trackedCompletions.delete(uri.toString())
+        }
+    }
+
+    public dispose(): void {
+        for (const disposable of this.disposables) {
+            disposable.dispose()
```
arafatkatze, line 175: It would be good to add this 
```
public dispose(): void {
    for (const disposable of this.disposables) {
        disposable.dispose();
    }
    
    for (const timeoutId of this.managedTimeouts) {
        clearTimeout(timeoutId);
    }
    this.managedTimeouts.clear();
    
    this.trackedCompletions.clear();
}
```

philipp-spiess: @arafatkatze Oh yes, thanks!!

vscode/src/completions/persistence-tracker.ts
```diff
@@ -0,0 +1,193 @@
...
+        } else {
+            const maxLength = Math.max(initialText.length, latestText.length)
+            const editOperations = LevenshteinCompare(initialText, latestText)
+            const difference = editOperations / maxLength
+            const isMostlyUnchanged = difference < 0.33
+
+            logCompletionEvent('persistence:present', {
+                id: trackedCompletion.id,
+                afterSec: MEASURE_TIMEOUTS[measureTimeoutsIndex] / 1000,
+                isMostlyUnchanged,
```
valerybugakov, line 118: Do we need to log this flag? Might as well use the `difference` field we already have and define the `isMostlyUnchanged` value in the analytics dashboard.

vscode/src/completions/persistence-tracker.ts
```diff
@@ -0,0 +1,193 @@
...
+    ): void {
+        const isStillTracked = this.trackedCompletions.get(trackedCompletion.uri.toString())?.has(trackedCompletion)
+        if (!isStillTracked) {
+            return
+        }
+
+        const initialText = trackedCompletion.completion.insertText
+        const latestText = trackedCompletion.document.getText(trackedCompletion.latestRange)
+
+        if (latestText.length === 0) {
```
valerybugakov, line 106: Not sure if it's possible, but we might want to filter out items where `initialText.length === 0` too

vscode/src/completions/text-processing/string-comparator.ts
```diff
@@ -43,49 +45,5 @@ export const isAlmostTheSameString = (stringA: string, stringB: string, percenta
...
-                const substitution = matrix[n - 1][m - 1]
-
-                matrix[n][m] = Math.min(deletion, insertion, substitution) + 1
-            }
-        }
-    }
-
-    return matrix[B.length][A.length]
+    return levenshtein(A, B)
 }
```
valerybugakov, line 49: Should we remove this helper (`LevenshteinCompare`) altogether because it doesn't add value?

valerybugakov (approved):
New data incoming! 🎉

vscode/src/completions/logger.ts
```diff
@@ -300,6 +303,11 @@ export function accept(id: SuggestionID, completion: InlineCompletionItem): void
         acceptedItem: { ...completionItemToItemInfo(completion) },
     })
     statistics.logAccepted()
+
+    if (persistenceTracker === null) {
+        persistenceTracker = new PersistenceTracker()
+    }
```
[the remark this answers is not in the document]
philipp-spiess: There's some odd circular dependencies that will make it not work in vitest if I do that 😢

vscode/src/completions/persistence-tracker.ts
philipp-spiess: cc @kelsey-brown

vscode/src/completions/logger.ts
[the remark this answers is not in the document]
valerybugakov: Got it, makes sense 👍

PR #1382, opened 2023-10-12
feat: filter context from .cody/.ignore

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1049

For details, please see [RFC 852 File Exclusion Patterns for Cody](https://docs.google.com/document/d/15XJnQ-qa-Y3SM2eltuKfuESmgYYfx7JhEdi2447YF_0/edit?usp=sharing)

feat: filter context from ignored files

> IMPORTANT: Currently behind the `cody.internal.unstable` config and must be enabled to try the feature. When this is out of internal experimental stage, this should be enabled by default.

- Add `isCodyIgnoreFile` utility to check if file is defined in .cody/.ignore (at workspace level)
- Update Interaction to filter context messages from ignored files when first created
- Update context gathering in completions to skip ignored files
- Add check in editor to skip ignored files
- ignore all `.env` files / path by default

All clients should provide file content from the .cody/.ignore  file at start up, and update the ignoreList when the file is changed.

## Note

This PR includes update to the VS Code clients so that the ignore list is set at Editor start up, and whenever the .cody/.ignore file is updated, the ignore list will be updated accordingly.

This does not make any changes to the current embedding steps or BFG, which will be a feature to be implemented in the future as listed by Dom in the discussion below.

- [x] PRFAQ to agree on the `.cody/.ignore`name

## Test plan

Prep: 

image

- Build Cody from this branch, start in VS Code debug mode
- Turn on the Unstable Experimental Features from the Cody Settings
- Open the `cody` repository 
- Create a .codyignore file
- Add the following to the .codyignore file
```
vscode/**
utils.ts
```

### Chat

Ask cody about fixup in Chat View: "what is fixup?" 

You should see Cody did not read any files from the `vscode` directory to answer your question:

image

### Autocomplete 

Go to any of the file inside the `vscode` directory, you should not get any autocomplete suggestions.

image

### Command

Go to lib/shared/src/utils.ts file, and then select some code, right click and select `Explain Code`:
image

You should see cody explain a random code (hallucinating because we didn't provide any code as context) that is not your selected code.

image

You can check the output channel to confirm no context was sent to the LLM:

image

## Demo

https://github.com/sourcegraph/cody/assets/68532117/9ea47b4c-311b-40b4-954b-db6d91769a21

### Update

Here is a loom video of codyignore working in Simple chat panel:

https://www.loom.com/share/83fe3e0b0728449385d130d45c317cfa


lib/shared/src/chat/context-filter.ts
```diff
@@ -0,0 +1,64 @@
...
+            return true
+        }
+    }
+    return false
+}
+
+/**
+ * Parses a Cody ignore file content and sets the global codyIgnoreList array with the rules.
+ *
+ * NOTE: Each client should call this function at the start of the client, and every time the.codyignore file changes.
```
philipp-spiess, line 28: What are the thoughts about doing this inside the agent instead? Seems like something that would be easy to miss otherwise and we already have APIs to read files?

We can also do it lazily (and don't need to do even more work at start up - our extension is already starting so slowly) when `isCodyIgnoreFile` is called the first time, for example.

lib/shared/src/chat/context-filter.ts
```diff
@@ -0,0 +1,64 @@
...
+
+let codyIgnoreList: string[] = []
+
+/**
+ * Checks if the given file name should be ignored by Cody.
+ *
+ * @param fileName - The file name to check.
+ * @returns A boolean indicating if the file should be ignored.
+ */
+export function isCodyIgnoreFile(fileName?: string): boolean {
```
philipp-spiess, line 11: Nit but the file in question is not the `codyignore` file but an ignore**d** file 🙈 

Also let's use URIs instead so it will cover edge cases 

```suggestion
export function isCodyIgnoredFile(uri?: vscode.Uri): boolean {
```

lib/shared/src/chat/context-filter.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    codyIgnoreList = Array.from(filesToExclude)
+}
+
+function patternToRegExpString(pattern: string): string {
+    // Escape special characters and convert '*' and '?' to their regex equivalents
+    return pattern
+        .replace(/^\*\*\//, '')
+        .replaceAll('*', '.*')
+        .replaceAll('?', '.')
```
philipp-spiess, line 63: We'll also need to escape `.` to `\.` I think, otherwise:

```
.foo
```

will also exclude `afoo` or so 🙈 

Maybe we should rely on some npm package for this? e.g. https://www.npmjs.com/package/@wider/regexp-gitignore?activeTab=code

lib/shared/src/chat/transcript/interaction.ts
```diff
@@ -22,7 +23,35 @@ export class Interaction {
...
+     * Removes context messages for files that should be ignored.
+     *
+     * Loops through the context messages and builds a new array, omitting any
+     * messages for files that match the CODY_IGNORE files filter.
+     * Also omits the assistant message after any ignored human message.
+     *
+     * This ensures context from ignored files does not get used.
+     */
+    private async removeCodyIgnoredFiles(): Promise<void> {
+        const contextMessages = await this.fullContext
```
philipp-spiess, line 40: Is there a potential race condition with this being async but not awaited?

vscode/src/services/context-filter.ts
```diff
@@ -0,0 +1,40 @@
...
+        const currentIgnoreList = await getCodyIgnoreFile()
+        if (!currentIgnoreList) {
+            return
+        }
+        // Set the updated cody ignore list
+        setCodyIgnoreList(currentIgnoreList)
+    }
+
+    // Update ignore list on editor change
+    vscode.workspace.onDidChangeTextDocument(async e => {
```
philipp-spiess, line 17: I think this won't fire if the document is not open and modified with another editor making the cache potentially stale and leaking files that should be ignored. We should probably use FS watch libraries for this (or maybe not cache this at all/invalidate after 1 minute or so? How often is the check being called?)

I wonder how git does this 🤔

lib/shared/src/chat/context-filter.ts
```diff
@@ -0,0 +1,64 @@
+export const CodyIgnoreFileName = '.codyignore'
```
philipp-spiess, line 1:
```suggestion
export const CODY_IGNORE_FILENAME = '.codyignore'
```

We usually name constants in uppercase to visually distinguish it from classes.

vscode/src/services/context-filter.ts
```diff
@@ -0,0 +1,40 @@
...
+            await update()
+        }
+    })
+
+    await update()
+}
+
+async function getCodyIgnoreFile(): Promise<string | undefined> {
+    // Get git ignore file context using the vs code api
+    // First, get the gitignore file from the workspace root directory
```
philipp-spiess, line 31: These comments seem already outdated

philipp-spiess (commented):
Awesome stuff! Left some early comments but feel free to ignore

philipp-spiess:
Did you have a look at https://www.npmjs.com/package/ignore? This is what's being used by eslint for example :)

lib/shared/src/chat/context-filter.ts
philipp-spiess: Did you have a look at https://www.npmjs.com/package/ignore? This is what's being used by eslint for example :)

lib/shared/src/chat/context-filter.ts
philipp-spiess: Did you have a look at https://www.npmjs.com/package/ignore? This is what's being used by eslint for example :)

abeatrix: yo this is awesome, this would have saved me so much time 😭  should have asked you earlier, thank you for the link!!!

vscode/src/services/context-filter.ts
abeatrix: updated to use file watcher!

lib/shared/src/chat/transcript/interaction.ts
abeatrix: moved to getFullContext() which `Transcript` use for building the final context messages

vscode/src/services/context-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+        const watchPattern = getFileWatcherRelativePattern(ignoreFileUri.fsPath)
+        const watcher = vscode.workspace.createFileSystemWatcher(watchPattern)
+        // update the ignored list on file change
+        watcher.onDidChange(async () => {
+            await update(ignoreFileUri)
+        })
+        watcher.onDidChange(async () => {
+            await update(ignoreFileUri)
+        })
```
DanTup, line 52: These two blocks look the same - was one supposed to be for creation?

vscode/src/services/context-filter.ts
```diff
@@ -0,0 +1,91 @@
...
+            console.error('Failed to read codyignore file')
+        }
+    }
+
+    return getWatcher()
+}
+
+// Find the .cody/.ignore file location using the vs code api
+async function getCodyIgnoreFileUri(): Promise<vscode.Uri | undefined> {
+    const codyIgnoreFile = await vscode.workspace.findFiles(CODY_IGNORE_FILENAME)
```
DanTup, line 78: If we're only interested in the first result here (although that feels odd), we can pass `maxResults: 1` here to allow VS Code to stop searching the workspace after it's found one.

abeatrix:
@DanTup Thank you so much for reviewing this PR! 

Updated the PR to address your very helpful feedback, especially the point regarding handling relative file paths in isCodyIgnoredFile:

- Update isCodyIgnoredFile to handle relative file paths instead of absolute
- Extract current ignore file directory and make file paths relative to it
- Handle errors in isCodyIgnoredFile instead of returning false
- Update setCodyIgnoreList docs to mention it takes ignore file path 

I haven't looked into passing the entire ignore file which seems to make things a lot easier, will continue the discussion with you on Slack!

lib/shared/src/chat/context-filter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Checks if a file should be ignored by Cody based on the .codyignore rules.
+ */
+export function isCodyIgnoredFile(fileName?: string): boolean {
+    if (!fileName) {
+        return false
+    }
+    // file must be path.relative - remove '/' from the start of the path
+    return codyIgnored.ignores(path.join(fileName.replace(/^\//, '')))
```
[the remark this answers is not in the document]
abeatrix: GREAT POINT! Updated to use path.relative for this

vscode/src/services/context-filter.ts
DanTup: (this changed - we now use the full results because we merged nested ignores)

lib/shared/src/chat/context-filter.ts
```diff
@@ -0,0 +1,44 @@
...
+    patternList.add('.env')
+    // Split the content of the file by new lines
+    const codyIgnoreFileLines = codyIgnoreFileContent.toString().split('\n')
+    // Loop through each line of the gitignore file
+    for (const line of codyIgnoreFileLines) {
+        if (line.trim()) {
+            patternList.add(line.trim())
+        }
+    }
+    codyIgnored = ignore().add(Array.from(patternList))
```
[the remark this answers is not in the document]
DanTup: (no longer relevant, we manually add lines because we merge the files and have to add prefixes)

lib/shared/src/chat/context-filter.test.ts
```diff
@@ -0,0 +1,65 @@
+import { describe, expect, it } from 'vitest'
+
+import { isCodyIgnoredFile, setCodyIgnoreList } from './context-filter'
+
+describe('isCodyIgnoredFile', () => {
+    // Set the ignore list to the following content:
+    const codyignoreFileContent = `
+        node_modules/
+        **/cody
+        **/foo/**
```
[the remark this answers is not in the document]
DanTup: (I added some tests for `'one/**/two'` in the new ignore helper tests - `"returns the correct value for a sample of rules"`)

slimsag:
Q:

> IMPORTANT: Currently behind the cody.internal.unstable config and must be enabled to try the feature. When this is out of internal experimental stage, this should be enabled by default.

What's the motivation behind this vs. making the existence of a `.cody/.ignore` file opt-in to this functionality? When would it not be behind the feature flag?

lib/shared/src/chat/ignore-helper.ts
```diff
@@ -0,0 +1,166 @@
...
+
+/**
+ * The Cody ignore file path in the native platform style (backslashes on Windows).
+ */
+export const CODY_IGNORE_FILENAME = path.join('.cody', '.ignore')
+
+/**
+ * The Cody ignore file path in POSIX style (always forward slashes).
+ */
+export const CODY_IGNORE_FILENAME_POSIX_GLOB = path.posix.join('**', '.cody', '.ignore')
```
slimsag, line 14: I notice the PR title and description mentions `.codyignore`, while the code here seems to suggest `.cody/.ignore`. (not commenting on the awkwardness of two `.` files..)

which one is right / expected to work?

slimsag (approved):
This is quite useful to us on the Cody strat team, thank you for working on this!

abeatrix:
> Q:
> 
> 
> 
> > IMPORTANT: Currently behind the cody.internal.unstable config and must be enabled to try the feature. When this is out of internal experimental stage, this should be enabled by default.
> 
> 
> 
> What's the motivation behind this vs. making the existence of a `.cody/.ignore` file opt-in to this functionality? When would it not be behind the feature flag?

I wanted to emphasize on the fact that this feature is still in experimental and unstable (not throughly tested so users must manually enable this feature to confirm before they can test it out. 

I'm thinking once we have tested this internally and have .Cody/.ignore working for embedding jobs, we can have it enabled by default. 

Do you think this is a good idea, or am I overthinking 😂?

lib/shared/src/chat/ignore-helper.ts
```diff
@@ -0,0 +1,166 @@
...
+                // Gitignores always use POSIX/forward slashes, even on Windows.
+                const ignoreRule = relativeFolderPath.length
+                    ? relativeFolderPath.replaceAll(path.sep, path.posix.sep) + path.posix.sep + ignoreLine
+                    : ignoreLine
+                rules.add((isInverted ? '!' : '') + ignoreRule)
+            }
+        }
+
+        this.workspaceIgnores.set(workspaceRoot, rules)
+        this.currentWorkspace = workspaceRoot
```
DanTup, line 87: This is just tracking the _last_ folder in the workspace (because this function will be called for each workspace folder). Is this intended?

(it's not clear to me what the expected support is for Cody in multi-root workspaces though)

abeatrix: That's not intended, thanks for pointing this out!
I misread the code and thought we run `refresh` on workspace change, will update and document the change!

lib/shared/src/chat/ignore-helper.ts
```diff
@@ -110,6 +116,11 @@ export class IgnoreHelper {
             return this.getDefaultIgnores().ignores(path.basename(uri.fsPath))
         }
 
+        // Set current workspace for checking relative paths returned by embedding client
```
abeatrix, line 119: Does this make sense to you?

lib/shared/src/chat/ignore-helper.ts
```diff
@@ -110,6 +116,11 @@ export class IgnoreHelper {
...
+        // Before each request submitted by a user, we will first run this method to check
```
abeatrix, line 120: @DanTup Does this make sense to you?

lib/shared/src/chat/ignore-helper.ts
abeatrix: It'll be .cody/.ignore! Will attach the RFC to the issue to clarify! [RFC 852 File Exclusion Patterns for Cody](https://docs.google.com/document/d/15XJnQ-qa-Y3SM2eltuKfuESmgYYfx7JhEdi2447YF_0/edit?usp=sharing)

slimsag:
> I'm thinking once we have tested this internally and have .Cody/.ignore working for embedding jobs, we can have it enabled by default.

Sounds good. I wouldn't block releasing this on embedding jobs (since that may come later), but I agree about testing it more internally first.

chwarwick (commented):
Mostly an observation, it appears that every ContextRetriever must know it is responsible for checking if it should ignore the source file, but it doesn't look like there is anything that enforces this.  In my mind this opens a risk since the space is evolving so quickly, it is easy to miss this when adding a new ContextRetriever, and it wouldn't be easy to spot during a code review either. 

It would be great if we could enforce this and/or make it clear to the user that they were responsible for filtering out content.

abeatrix:
> Mostly an observation, it appears that every ContextRetriever must know it is responsible for checking if it should ignore the source file, but it doesn't look like there is anything that enforces this. In my mind this opens a risk since the space is evolving so quickly, it is easy to miss this when adding a new ContextRetriever, and it wouldn't be easy to spot during a code review either.
> 
> It would be great if we could enforce this and/or make it clear to the user that they were responsible for filtering out content.

I agree. I think eventually we want to create a ContextRetriever class and make sure all the users use it instead of creating their own. 

For chat and commands, all the context must go through the Interaction class before they are sent to the LLM, so this is where we would do the final filtering process: 
https://github.com/sourcegraph/cody/blob/8ae6f6fe1d82e60866b2a80ce7c7b03e81c92b5e/lib/shared/src/chat/transcript/interaction.ts

lib/shared/src/chat/ignore-helper.ts
DanTup: I'm not sure I understand how embeddings work well enough to know how this should work, but it looks like the current implementation will assume that any relative path (sent to `isIgnoredInCurrentWorkspace`) will treat the workspace of the last absolute path (sent to `isIgnored`) as the root for that relative path.

In a workspace with a single folder, maybe this is correct (although it's not clear to me if the relative paths from embeddings are guaranteed to match the workspace folder root.. can you open a parent folder of a project and have it load embeddings from a nested project?).

In a multi-root workspace, it seems like this behaviour may be inconsistent (depending on which workspace folder the last file sent to `isIgnored` came from). Although it's also unclear to me what happens in a multi-root workspace that has two projects - are embeddings used from both projects?

Perhaps the main confusion for me is what the idea of a "current workspace root" is in a multi-root workspace.

abeatrix: Ah sorry for the confusion.

I believe the plan is that strat team will work on [implementing .cody/.ignore at indexing time for embedding jobs](https://sourcegraph.slack.com/archives/C05FY86PHLM/p1700064374602069?thread_ts=1700000166.495919&channel=C05FY86PHLM&message_ts=1700064374.602069), so the clients (e.g. vs code) can assume all the results returned from embeddings are already filtered. But before that is implemented, we will just attempt to match it for their current workspace, which we know it would **not** work well in multi root workspaces, but at the same time, we current do not support multi root workspaces on the client side as well, since user can only define one codebase at a time in their Cody config, and I believe that's what @slimsag is referring to [here](https://github.com/sourcegraph/cody/pull/1382#issuecomment-1813315386), and why this is behind an unstable feature flag and marked for internal use only atm. 

This is valid concern though. Maybe we can mention it in our doc or somewhere that we only support the 
.Cody/.ignore file at the current workspace for now until the embeddings is ready? What do you think? @DanTup @slimsag

DanTup: >  we will just attempt to match it for their current workspace, which we know it would not work well in multi root workspaces, but at the same time, we current do not support multi root workspaces on the client side as well, since user can only define one codebase at a time in their Cody config

Ah, got it :)

>  Maybe we can mention it in our doc or somewhere that we only support the
`.Cody/.ignore` file at the current workspace (root dir for the codebase they have indexed)

Just to be clear - what is "current workspace" here? If I _do_ open a multi-root workspace and the first workspace folder is the one being used for Cody/embeddings, is it guaranteed that the `currentWorkspace` variable in this PR is using that same one? (or is it possible to trigger `isIgnored` to be called for a file in the second workspace folder, which will cause it to use that)?

It sounds like it's important that the ignore helpers definition of "currentWorkspace" is exactly the one embeddings can be coming from, but I don't know how to verify that.

(and if the plan is to do the embedding ignore work before removing this from a flag, maybe the point is moot?)

chwarwick: > I believe the plan is that strat team will work on [implementing .cody/.ignore at indexing time for embedding jobs](https://sourcegraph.slack.com/archives/C05FY86PHLM/p1700064374602069?thread_ts=1700000166.495919&channel=C05FY86PHLM&message_ts=1700064374.602069), so the clients (e.g. vs code) can assume all the results returned from embeddings are already filtered. 


There are a few additional things to consider as well.  The clients should not blindly trust that the backend will filter embeddings search results because of the different release cadence between the two.  For an extended period of time after a version with ignore filter is released there will be users that have an extension that supports `.cody/.ignore` but have not yet upgraded to that version.

A further consideration with embeddings is that the results may not be related to a repo in the current workspace, due to the ability to set a specific codebase as context, and if multi-repo context is enabled in future.   This would remain a potential issue until all users are running a version of the backend that supported ignore filtering.

dominiccooney:
@abeatrix 

> I believe the plan is that strat team will work on [implementing .cody/.ignore at indexing time for embedding jobs](https://sourcegraph.slack.com/archives/C05FY86PHLM/p1700064374602069?thread_ts=1700000166.495919&channel=C05FY86PHLM&message_ts=1700064374.602069), so the clients (e.g. vs code) can assume all the results returned from embeddings are already filtered.

We should apply the filter, redundantly, at every level. Defense in depth. Be conservative about what we *do* send. If the user adds something to the ignore file, makes a query, and we include the files they just told us to ignore then that will alarm people.

On the flip side, if they loosen the ignore file and we are slow to pick up the item from embeddings, well, that is a conservative default and there are lots of reasons why something is *not* included.

I just skimmed this thread, sorry if I have missed some nuance.

TL;DR, we should not rely on one layer implementing ignore filtering. The distributed nature of the problem means that won't work. We should implement it at every appropriate layer and have conservative failure modes.

lib/shared/src/chat/ignore-helper.ts
abeatrix: @chwarwick I think @DanTup has raised a good question about what the definition of "currentWorkspace" is. From what I understand, the concept of a workspace in VS Code is different than what we use for the embedding jobs, which is codebase/repo based. In this case, a user can work in a workspace with multiple codebases inside, but currently, the embedding client in our VS Code extension is only returning results from a single codebase, which is whatever codebase the user is defined in their current VS Code config --is this correct? 

If yes, do you think the proposal below can make sure the same `.cody/.ignore` file implementation works across clients?

- A single VS Code workspace can have multiple codebases nested in it
 - `.cody/.ignore` must be placed in the root of a codebase
 - the first line of `.cody/.ignore` should be the name of the codebase as shown in their Sourcegraph instance
   - this allows us to associate the results returned from the embeddings client with the correct `.cody/.ignore` file

This makes the concept of workspaces (somewhat) irrelevant while allowing users to define ignores for different codebase, and enabling us to associate each embeddings result with the correct `.cody/.ignore` file that is available locally since we know which codebase are the results from.

An example of a .cody/.ignore file for the cody codebase:

```
@github.com/sourcegraph/cody

.env
\*.json
.vscode/
README.md
```

Would this make sense? What do you think @chwarwick @DanTup @slimsag ?

abeatrix: P.S: Just saw your reply @dominiccooney, welcome 💃  Would love to hear your thoughts on this!

DanTup: I think deviating from a standard ignore file format would be unfortunate, but I don't know if there's a better way.

Maybe it would be useful if I had a better understanding of exactly what the embeddings and "a cody codebase" are. Is it possible they are unrelated to the code you have open in VS Code, or is the intention that these are the same (eg. you open a folder in VS Code, and the codebase/embeddings are supposed to be for that very same project)?

(If there are docs or code that would explain this that I should read to better understand, please let me know :) )

abeatrix: Here is a working prototype of what I described above: https://github.com/sourcegraph/cody/pull/1769

>  Is it possible they are unrelated to the code you have open in VS Code, or is the intention that these are the same

It is possible. For example, I can configure Cody to connect to the sourcegraph/sourcegraph repo in VS Code when I only have the sourcegraph/cody repository opened in my workspace. 

Or I can be working in a workspace with the sourcegraph/cody repo at fsPath `foo/bar/sourcegraph/cody`, but I have renamed it to `cody-dev`with filePath `foo/bar/sourcegraph/cody-dev` that'd cause mismatches? 

> I think deviating from a standard ignore file format would be unfortunate

We will still be using the standard ignore file format except the first line, or instead of `@` they can add `#` in the front like a comment? 😜 Just throwing ideas out haha

On second thought, maybe there is an easier way to do this. We can just run git to get their codebase info for each .cody/.ignore file we can find to do the mapping automatically?  🤔

slimsag: You can grab the remote URL from Git (`git remote get-url origin`) and then pass that to the GraphQL API designed to turn git remotes into repository names: https://sourcegraph.com/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20repository(cloneURL%3A%20%5C%22git%40github.com%3Asourcegraph%2Fsourcegraph.git%5C%22)%20%7B%5Cn%20%20%20%20name%5Cn%20%20%7D%5Cn%7D%5Cn%22%2C%22variables%22%3A%22%22%7D

This API was created for editor extensions and the `src` CLI in the past to do exactly this, and handles a bunch of edge cases where mapping a remote URL -> Sourcegraph repository name can be tricky.

lib/shared/src/chat/transcript/interaction.ts
```diff
@@ -24,6 +25,42 @@ export class Interaction {
...
+            // Skips the assistant message if the human message is ignored
+            if (message.speaker === 'human' && message.file) {
+                const { uri, repoName, fileName, source } = message.file
+                if (uri && isCodyIgnoredFile(uri)) {
+                    i++
+                    continue
+                }
+                // Filter embedding results from the current workspace
+                if (source === 'embeddings') {
+                    console.log(fileName, repoName)
```
DanTup, line 51: Should this `console.log` be removed? (I presume it was for debugging)

vscode/src/repository/repositoryHelpers.ts
```diff
@@ -39,3 +41,64 @@ export function gitAPI(): API | undefined {
...
+// Git Remote URL <> Codebase name
+let vscodeGitAPI: API | undefined
+
+export function gitAPIinit(): void {
+    const extension = vscode.extensions.getExtension<GitExtension>('vscode.git')
+    extension?.exports.onDidChangeEnablement(enabled => {
+        console.log('Git extension enabled state changed to ' + enabled)
+        if (enabled) {
+            vscodeGitAPI = extension.exports.getAPI(1)
+        }
```
DanTup, line 54: Should this be set to `undefined` if it's not disabled?

vscode/src/repository/repositoryHelpers.ts
```diff
@@ -39,3 +41,64 @@ export function gitAPI(): API | undefined {
...
+            const workspaceRoot = repository.rootUri.fsPath
+            const codebaseName = getCodebaseFromRepo(repository)
+            if (workspaceRoot && codebaseName) {
+                if (codebaseName) {
+                    matches.push({ ws: workspaceRoot, codebase: codebaseName })
+                }
+            }
+        }
+    } catch {
+        // no-ops
```
DanTup, line 89: Is it worth logging this to aid debugging if something isn't working? (I'm not sure what sort of exceptions this expects to catch, so maybe it's not useful).

vscode/src/services/context-filter.ts
```diff
@@ -0,0 +1,152 @@
...
+    const codebaseName = getCodebaseFromWorkspaceUri(wf.uri)
+    const ignoreFilePattern = new vscode.RelativePattern(wf.uri, CODY_IGNORE_FILENAME_POSIX_GLOB).pattern
+    const ignoreFiles = await vscode.workspace.findFiles(ignoreFilePattern)
+    const codebases = new Map<string, string>()
+    const filesWithContent = await Promise.all(
+        ignoreFiles.map(async fileUri => {
+            const codebase = codebaseName || getCodebaseFromWorkspaceUri(fileUri)
+            if (codebase) {
+                // file root is two level above the fileUri location
+                const fileRoot = fileUri.fsPath.replace(/(?:\/[^/]+){2}$/, '')
```
DanTup, line 99: Will this work for Windows? `fsPath` would have backslashes. Would it be better to do something like `path.dirname(path.dirname(...))`?

DanTup (approved):
LGTM with a few minor comments (although I'm less familiar with some of this code so you may want another reviewer too).

vscode/src/services/context-filter.ts
```diff
@@ -0,0 +1,152 @@
...
+            watcher.dispose()
+            didChangeSubscription.dispose()
+            onDidChangeConfig.dispose()
+        },
+    }
+}
+
+// TODO (bee) merge it with setUpCodyIgnore()
+let init = false
+export async function codyIgnoreInit(): Promise<void> {
```
abeatrix, line 53: @DanTup i have to split this into another function and call it in the simple chat view when a chat panel is ready because if we call it too early before the git extension is ready, the init step would fail for the initiate workspace and nothing would get updated. Do you have any ideas on how we can run `refresh` automatically when the git extension is ready? (not urgent, we can do this in a follow up PR but wanted to pick your brain on this 😃 )

DanTup: By "ready" do you mean has finished activating? If so, I think you could use the `activate()` method like:

```
vscode.extensions.getExtension("...").activate().then(() => codyIgnoreInit());
```

You might need to handle an error if the extension is disabled though (or there's some other reason it couldn't be activated).

abeatrix: @DanTup Oh yea that's what I meant, will give that a try, thank you so much!

vscode/src/repository/repositoryHelpers.ts
```diff
@@ -39,3 +44,89 @@ export function gitAPI(): API | undefined {
...
+ *
+ * Initializes the Git API by activating the Git extension and setting up the gitAPI global.
+ *
+ * Activates the Git extension, sets up a listener for when the extension becomes enabled/disabled,
+ * and initializes the global gitAPI instance. Also sets up the .codyignore file on startup.
+ */
+let vscodeGitAPI: API | undefined
+export async function gitAPIinit(): Promise<void> {
+    const extension = vscode.extensions.getExtension<GitExtension>('vscode.git')
+    extension?.exports.onDidChangeEnablement(enabled => {
```
DanTup, line 59: Does this handler need adding to disposables?

DanTup: Also - is it possible this function will be called multiple times? (there are a few places that call it). If so, I think we only want this part once?

abeatrix: > Does this handler need adding to disposables?

Hmm it is working as it is right now, but is it best practice to add all disposables to disposables? If yes, i can update it!

> Also - is it possible this function will be called multiple times? (there are a few places that call it). If so, I think we only want this part once?

Good call. I will remove it from `getCodebaseFromWorkspaceUri` to make sure it will only be called by one function once.

DanTup: > is it best practice to add all disposables to disposables?

I'm not certain to be honest.. My understanding is that an extension is only unloaded when the extension host is being destroyed (and it doesn't `await` any async `dispose`s), so the whole `extension.subscriptions` collection feels somewhat pointless to me, but I continue to do it just in case 😄

Adding multiple event handlers is probably the main issue.

DanTup, line 59: Oh also - according to the docs it's an error to access `.exports` before the extension is activated, so it seems like if we get to this code before it's active then this should error (and if we don't, then the `activate()` on line 63 is redundant). So perhaps something isn't quite right here.

abeatrix: cool, i've moved gitAPIinit to main to add it to the disposables!

abeatrix: @DanTup Does the changes in the latest commit make sense to you?

vscode/src/repository/repositoryHelpers.ts
```diff
@@ -39,3 +44,93 @@ export function gitAPI(): API | undefined {
...
+    const extension = vscode.extensions.getExtension<GitExtension>('vscode.git')
+
+    function init(): void {
+        if (!vscodeGitAPI && extension) {
+            vscodeGitAPI = extension.exports.getAPI(1)
+            setUpCodyIgnore()
+        }
+    }
+
+    await extension?.activate().then(() => init())
```
DanTup, line 65: Do we need to worry about any exceptions here (for example if you disable the Git extension does it throw?)

vscode/src/repository/repositoryHelpers.ts
DanTup: One question below about exceptions, but otherwise seems reasonable to me!

vscode/src/repository/repositoryHelpers.ts
abeatrix: ahh i added try catch to this for cases where Git extension is disabled. we probably will need to do some follow up works on UX improvement but I think what we have is a good start. I am going to do some final testings before I merge it but consider this is currently behind the `cody.internal.unstable` config flag, i think we should be good 🙏 

Thank you so much for your help and your throughout review 🙇‍♀️

PR #1386, opened 2023-10-13
Autocomplete: fix `canUsePartialCompletion` for completions with leading new lines

valerybugakov:
## Context

In cases where multiline completions start with a leading new line, we do not account for the fact that the `truncateMultilineCompletion` call strips it. That's why `canUsePartialCompletion` returns false positives, leading to prematurely truncated multline completions.

```js
{
  "completion": {
    "text": "\nconst merge = (left, right) => {\n  let arr = [];\n  while (left.length && right.length) {",
    "newLines": 4
  },
  "truncated": {
    "text": "const merge = (left, right) => {\n    let arr = [];\n    while (left.length && right.length) {",
    "newLines": 3
  },
  "conditionForPartialReuse": "truncated.split('\n').length < completion.split('\n').length"
}
```

- [Slack thread.](https://sourcegraph.slack.com/archives/C05AGQYD528/p1697139038515929)
- A follow-up issue: https://github.com/sourcegraph/cody/issues/1387

## Test plan

Added unit tests.

### DIFF
--- a/vscode/src/completions/can-use-partial-completion.ts
+++ b/vscode/src/completions/can-use-partial-completion.ts
@@ -23,6 +23,7 @@ export function canUsePartialCompletion(
     partialResponse: string,
     { document, multiline, docContext: { prefix, suffix } }: CanUsePartialCompletionParams
 ): boolean {
+    // return false
     const lastNlIndex = partialResponse.lastIndexOf('\n')
 
     // If there is no `\n` in the completion, we have not received a single full line yet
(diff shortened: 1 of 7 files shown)


vscode/src/completions/get-inline-completions-tests/streaming.test.ts
```diff
@@ -134,4 +134,71 @@ describe('[getInlineCompletions] streaming', () => {
...
+          "const merge = (left, right) => {
+              let arr = [];
+              while (left.length && right.length) {
+                  if (true) {}
+              }"
+        `)
+        expect(result?.source).toBe(InlineCompletionsResultSource.Network)
+    })
+
+    it.skip('cuts-off multlineline compeltions with inconsistent indentation correctly', async () => {
```
valerybugakov, line 179: This case cannot be covered by the indentation-based approach. I kept it here to re-enable this test when https://github.com/sourcegraph/cody/issues/1387 is shipped.

vscode/src/completions/text-processing/truncate-multiline-completion.ts
```diff
@@ -36,6 +48,7 @@ export function truncateMultilineCompletion(
     if (lines.length > 1 && lines[0] === '' && indentation(lines[1]) === startIndent) {
```
philipp-spiess, line 48: Oh, okay, I didn't know this was the behavior (although I added it lol). I think we should split `truncateMultilineCompletion` into two distinct behaviors:

- `normalizeStartLine`
- `truncateEndBasedOnIndentation`

WDYT?

vscode/src/completions/can-use-partial-completion.ts
```diff
@@ -23,6 +23,7 @@ export function canUsePartialCompletion(
     partialResponse: string,
     { document, multiline, docContext: { prefix, suffix } }: CanUsePartialCompletionParams
 ): boolean {
+    // return false
```
philipp-spiess, line 26:
```suggestion
```

philipp-spiess (approved):
Great find!

vscode/src/completions/text-processing/truncate-multiline-completion.ts
valerybugakov: I considered it first but thought it was easier to understand the whole truncation logic in one place. How we truncated the beginning of the completion might affect the end truncation, and we might want to add new heuristics later.

philipp-spiess: Sounds like you gave that more thought then 👍

PR #1392, opened 2023-10-13
feat: feature flag for making less unhelpful suggestion requests

abeatrix:
REF: https://sourcegraph.slack.com/archives/C05AGQYD528/p1697227062174209?thread_ts=1697115251.236759&cid=C05AGQYD528

feat: new feature flag for increased autocomplete debounce time

This commit is proposing to add a new feature flag called CodyAutocompleteLowPerformanceDebounce to enable increased debounce time for autocomplete in low-performance languages.

The debounce time is increased to 1000ms (the default number we used for added latency) for single lines and 1500ms for multi-line (which is rare in most of the low-performance langs) in low-performance languages when the feature flag is enabled. 

The list of low-performance language IDs is defined in lowPerformanceLanguageIds.

This allows us to experiment with increased debounce time, specifically for low-performance languages.

### Proposal

My rationale is that by adding latency to low-performance language, we are still making the same amount of requests (between low-performance languages and those that are not), we just display them with a delay. It improves user experience without any impact.

With debounce, we will not only reduce the number of unhelpful suggestions that contribute to costs. The users will also benefit the same from getting fewer unhelpful suggestions, which count towards their daily rate limit. (unless I misunderstood how latency and debounce work, again 🤡 )

## Test plan

Enable the 'cody-autocomplete-low-performance-debounce' and work in a low-performance file to check if you are still getting suggestions as often as before.


dominiccooney (approved):
Change looks like it matches the intent in the description.

If the flags here continue to multiply it could be a problem because it will be hard to understand what the code might do, without understanding all of the flag configurations used in practice. Some suggestions:
- Minimize the coupling of the control flow between the flags and the provider. "Strategy Pattern" in old school Design Patterns terms. The "policy" object depends on flags, but the interface between the provider and the policy is explicit. This makes it clear if the interface is getting complicated. And if the debt burden of the flags becomes too great, it is easy to reset because you can implement a simplified policy in one place.
- When you're adding a flag, state what the decision is, then go back and remove old flags at the decision point.
- If a flag depends on another flag, as is the case here, make it a hard failure if that is violated. This cuts off the part of "flag space" which doesn't make sense (no special handling for low performance languages, but low performance debounce is on.) This is why enums are better than booleans for experiment flags: You can extend "no low performance language handling", "low performance latency" by adding a "low performance latency and debounce" very easily.

CC @philipp-spiess for your input on the above ideas ^^^

philipp-spiess:
@dominiccooney 100% agree. I think part of the reason why this logic is so complex is that we haven't found out a way for it to show the desired effects and as such all of these feature flags are just quick ideas for us to prototype. We should be equaly quick to remove old flags as we're in adding them here. Also like the idea of the strategy pattern for this.

However I too am a bit worried that we're playing whack-a-mole here that will end us at a local optimum that might hurt us in the long run :/ Maybe we find out a specific combination of latencies that work well for the current setup but then the LLM becomes slightly faster/slower or smarter/dumber and we end up in a worse UX. There's also the noticeable UX change that a longer latency implies that is a bit unsetting 😬 

The other issue here is that there's combinatorial complexity with all of those feature flags and ideally we'd need to run them for a long time to get a big enough sample size which is becoming more and more complicated the more combinations we add. E.g. @abeatrix what do you think is the proposed order of how we run these timeout experiments now? I’m also not really sure why we need to run the experiment once with delaying latencies _after_ the request and once by increasing the debounce time (which will result in latencies that are almost 2x the latency of the other experiment since you have to add the whole retrieval and generation steps on top of that 1 second timer). I guess my question here is: Do we have a hunch or an intuition that this is the right way to approach this problem or are we just throwing spaghetti at the wall until something sticks? cc @abeatrix @chenkc805

abeatrix:
> what do you think is the proposed order of how we run these timeout experiments now? I’m also not really sure why we need to run the experiment once with delaying latencies after the request and once by increasing the debounce time (which will result in latencies that are almost 2x the latency of the other experiment since you have to add the whole retrieval and generation steps on top of that 1 second timer). 

@dominiccooney @philipp-spiess Sorry I mentioned in Slack that i have moved my decision to the PR, but looks like it wasn't saved, so I just added under the proposal section in the PR description, but to answer your question directly:

I think delaying latencies after the request and once by increasing the debounce time does not make sense, and that's not what i want to propose.

I was thinking increasing the debounce time for low-performance languages would make more sense than adding latency for displaying unhelpful suggestions for low-performance languages as essentially the main goal for both is for users to get less annoying unhelpful suggestions, where with adding latency, we are still making the same amount of requests, while the other one allows us to make less requests if I understand this correctly, which might now be the case here lol 

The proposed order to run the experience IMO:
1. CodyAutocompleteLowPerformanceDebounce to see its impact for low-performance langs
2. Latency ones without low-perofmrance lang latency enabled IF number 1 has a positive impact 

> Do we have a hunch or an intuition that this is the right way to approach this problem or are we just throwing spaghetti at the wall until something sticks?

I think what i proposed makes more sense for low-performance languages at least to me, and seems more intuitive based on what we know about low-performance languages. But at the same time, we cannot say for certain, and this is why I haven't made any changes to the latency flag. Curious to hear what you and the team think!

abeatrix:
> I’m more curious on how we came up with the goal latency

I'm using the default number we set for the latency test. I initially worried that this number might be too high, but then I tried test it locally, and the experience and the suggestions quality were still the same for me when I was working in a css file 😅

Happy to adjust it to a lower number if you think this is too aggressive, I really don't have an opinion on this!

philipp-spiess:
> the experience and the suggestions quality were still the same for me when I was working in a css file 😅

If you think the experience is not jeopardized then I’m happy :)

PR #1394, opened 2023-10-14
fix: apply latency only when last suggestion was read

abeatrix:
RE: https://github.com/sourcegraph/cody/pull/1351

fix: apply min latency only when last suggestion was read

The minimum latency for code suggestions is now only applied when the last suggestion was read by the user. This avoids applying the latency when the user is actively typing and has not read the previous suggestions.

The lastCompletionRequestTimestamp property tracks when the last suggestion request was made. The latency is skipped if the time since last request is less than 750ms.

## Test plan

Check the output channel. You should not see latency getting increased after 5 keystrokes, which is the current behavior.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -22,7 +22,7 @@ import {
 } from './get-inline-completions'
 import { getLatency, LatencyFeatureFlags, resetLatency } from './latency'
 import * as CompletionLogger from './logger'
-import { CompletionEvent, SuggestionID } from './logger'
+import { CompletionEvent, READ_TIMEOUT_MS, SuggestionID } from './logger'
 import { ProviderConfig } from './providers/provider'
 import { RequestManager, RequestParams } from './request-manager'
 import { getRequestParamsFromLastCandidate } from './reuse-last-candidate'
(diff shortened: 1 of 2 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -263,10 +269,12 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                     language: await minLatencyFlagsPromises.language,
                     provider: await minLatencyFlagsPromises.provider,
                 }
-
+                // Do not apply the minimum latency if the last suggestion was not read, e.g when user was typing
+                const isLastSuggestionRead = start - this.lastCompletionRequestTimestamp > 750
```
valerybugakov, line 273: Let's export the `READ_TIMEOUT_MS` variable and re-use it here to make sure the values are in sync.

valerybugakov (approved):

philipp-spiess (approved):
Cool but keep in mind this is still not a 1:1 match to our analytics logging. There could, for example, be a previous completion request that never ended in a completion being shown at all (since it was either filtered out by post-processing or VS Code).

If we want to match against rejections as we define it, we should probably not try to detect rejections inside the completion callback but maybe hook into `statistics.ts` or so? 

Since we don't know wether this code is kept at all yet I don't think we should invest too much though :)

PR #1395, opened 2023-10-14
Move Chromium installation into install-deps so "pnpm test:e2e" works on Windows

DanTup:
This changes the npm script for `test:e2e` to not use syntax that doesn't work on Windows. I believe the intention of the script was to install VS Code and Chromium in parallel which this change still does, but with both done from the TS file which is more easily run in the NPM script in a way that works across platforms.

While this fixes `pnpm test:unit` on Windows for me locally, this change does not enable the bots yet because some of the tests fail when run on GitHub on Windows and that requires some debugging (locally, everything passes). I'll debug this in another PR (I'd like to get screenshots or videos working for failures, but when I try this locally I just get blank white screenshots/videos so it also requires some additional debugging).

## Test plan

These changes are to the test setup themselves, if the bots are still running `pnpm test:e2e` and the tests passing, everything is working.


dominiccooney (approved):
🎉 Looks splendid, thanks!

Issue #667, opened 2023-08-14
Make it obvious in-editor when Cody has provided an autocomplete


toolmantim:
Autocomplete is our # 1 onboarding task for now. But Cody’s autocomplete works behind the scenes, with no chance for a new user to understand what's really going on. Currently it's somewhat confusing, and a little anticlimatic.

We need to:
- Make it visible that Cody is working, and when it contributes an autocorrect
- Give directions on how to accept it
- Point out that it was accepted (with a little celebratory moment)
- Create awareness of next steps of onboarding, with a mention of the status bar icon on acceptance of the second autocomplete provided by Cody, and an additional notification in the chat.

Related PRDs:
- https://docs.google.com/document/d/1d3AsJeAB3AuUwAUTvvMzwKVQeVniX1Q06LPCBna5BbU/edit

Related Slack messages:
- https://sourcegraph.slack.com/archives/C0HMGV90V/p1691614048664469

References:
- Gitlen’s inline blame annotations:
  - [lineAnnotationController](https://github.com/gitkraken/vscode-gitlens/blob/38802451ba5917f5d5ae69d1bdf58ca5177b6c5c/src/annotations/lineAnnotationController.ts#L188)
  - [getInlineDecoration](https://github.com/gitkraken/vscode-gitlens/blob/38802451ba5917f5d5ae69d1bdf58ca5177b6c5c/src/annotations/annotations.ts#L221)

chenkc805:
Nice! Love the ghost text in the editor and celebration text afterwards. 

@philipp-spiess could we launch this as a test? I'd love to see if this leads to a meaningful bump in autocomplete metrics

philipp-spiess:
@chenkc805 Sure, I don't see a reason why we can't :)

chenkc805:
Lets do it then! CC: @kelsey-brown  for visibility

chenkc805:
Here's the experiment requirements:

- **experiment name**: obvious_autocomplete_667
- **# variants**: 2 variants, A (control, no extra text) and B (treatment, with extra text). Split is 50/50
- **bucketing moment**: fire a bucketing event when we show an autocomplete suggestion. Only fire for new users who have never seen a Cody autocomplete suggestion before (to measure new user impact).

toolmantim:
I imagine the autocomplete acceptance rate is highly dependent on the quality of the answers themselves (you get very different results based on the type of file you open and what line you're editing etc) — we might need to control for that (if possible).

[...]

PR #1397, opened 2023-10-14
Show an inline notification on accepting first completion

DanTup:
An implementation for https://github.com/sourcegraph/cody/issues/667.

I initially implemented decorations both when the completion was visible and when accepted, as described in the issue above:

https://github.com/sourcegraph/cody/assets/1078012/b027245a-9af3-4a33-b01b-aac9d5a0b0d8

However, there was some concerns about being able to reliably shown this _only when a Cody completion is visible_. Specifically, when VS Code asks us for completions, there is no guarantee they will be shown (because it may be asking multiple providers at once). We also can't tell if the user hits `escape` to hide the inline completion so the decoration could be hidden.

So this change implements only the accepted decoration:

https://github.com/sourcegraph/cody/assets/1078012/e456a00f-d6d9-4a9d-b7d6-abd62f2625a0

@toolmantim WDYT? ~~Still needs tests but I want to ensure the implementation is right before adding them.~~

Fixes https://github.com/sourcegraph/cody/issues/667

## Test plan

Tested with a new e2e test (`pnpm test:e2e`).


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -695,3 +699,78 @@ function onlyCompletionWidgetSelectionChanged(prev: CompletionRequest, next: Com
...
+/**
+ * Handles showing an in-editor decoration when a first completion is accepted.
+ */
+class FirstCompletionDecorationHandler {
+    /**
+     * Duration to show decoration before automatically hiding.
+     *
+     * Modifying the document will also immediately hide.
+     */
+    private static readonly decorationDurationMilliseconds = 3000
```
toolmantim, line 712: Given it auto-hides, a longer value is probably better.

```suggestion
    private static readonly decorationDurationMilliseconds = 10000
```

I gave 10s a go locally and it seemed good… unless there's some scenarios you have in mind where that would cause confusion?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -695,3 +699,78 @@ function onlyCompletionWidgetSelectionChanged(prev: CompletionRequest, next: Com
...
+
+    /**
+     * A timer to hide the decoration automatically.
+     */
+    private hideTimer: NodeJS.Timeout | undefined
+
+    private readonly decorationType = vscode.window.createTextEditorDecorationType({
+        after: {
+            margin: '0 0 0 40px',
+            contentText: '    🎉 Accepted your "first" Cody autocomplete!',
```
toolmantim, line 729: After playing with it locally, let's change the message to:

```suggestion
            contentText: '    🎉 You just accepted your first Cody autocomplete!',
 ```

toolmantim (approved):
This looks great @DanTup!

vscode/src/completions/inline-completion-item-provider.ts
DanTup: sgtm - I didn't think much about the number I just thought it made sense to hide after some time.

DanTup:
@toolmantim pushed some changes to add a test and some mock/shim changes required for them to pass, please check this still all looks reasonable. Thanks!

toolmantim (approved):
This works great! Thank you @DanTup.

DanTup:
Pushed some minor tweaks in c0763840745f767c4e6667e692d40496f5b88bbf and 9c5fb2f774ce879090a2c1165dce7d78dda29089 to fix some test flakes.

PR #1399, opened 2023-10-16
VScode: Clean up login and chat status widgets, tests

### LINKED ISSUE #1229: Clean up app-optional onboarding code
Requirements:
- Remove any experiment code
- Remove old code that supported the old flow (tech debt cleanup)

dominiccooney:
- Turn on "simplified" login  for VScode Web, but with a simplified, functional UX. See below.
- Remove the old login UI. There is very little error reporting in the simplified UI; this adds a generic notification when authentication fails.
- Update the tests to use "simplified" login.
- Remove the simplified login experiment code, and clean up the cached experiment arm on startup.
- Add an "enable embeddings" dialog for Enterprise which does not advertise App, and use the new chat input status widget everywhere.

Notes about login in VScode Web: VScode Web can not use vscode://... links as-is, but needs to wrap them. Dotcom won't pass tokens to arbitrary URLs. So we can't use the SSO provider login buttons as-is in VScode Web without some work. This replaces those buttons with basic instructions for pasting in tokens. Note, VScode Web is unsupported. This is just a way to keep login barely functional.

Fixes #1229 

## Test plan

Automated tests:

```
pnpm test:integration && pnpm test:e2e
```

VScode web login manual test:

```
pnpm dev:web
open http://localhost:3000
```

Check the links to dotcom work; sign in by pasting an access token from dotcom.

Enterprise enable embeddings popup manual test:

1. Run VScode Cody, click *Sign In to Enterprise Instance*
2. Pick *Sign In to Sourcegraph Enterprise Instance v5.1 and above*
3. Enter `https://sourcegraph.sourcegraph.com`
4. Follow the flow to grant a token; you should sign in.
5. *File, Open Folder* and open a clone of https://github.com/sourcegraph/sourcegraph , the chat input status indicator should show *Embeddings Enabled* etc.
6. Open a clone of https://github.com/ocaml/ocaml , the chat input status indicator should show embeddings are *not* available, but should *not* advertise App. Instead, it should have a link to these docs: https://docs.sourcegraph.com/cody/explanations/code_graph_context#configuring-embeddings


vscode/src/services/AuthProvider.ts
```diff
@@ -108,7 +108,11 @@ export class AuthProvider {
                 if (!authStatus?.isLoggedIn) {
                     const newToken = await showAccessTokenInputBox(item.uri)
                     const authStatusFromToken = await this.auth(selectedEndpoint, newToken || null)
-                    this.showIsLoggedIn(authStatusFromToken?.authStatus || null)
+                    if (authStatusFromToken?.isLoggedIn) {
```
abeatrix, line 111:
```suggestion
                    this.showIsLoggedIn(authStatusFromToken?.authStatus || null)
```
I haven't gone through the code in awhile but is the if/else statement still needed if we move `showAuthFailureMessage()` into `this.showIsLoggedIn()`  for `null` or failed AuthStatus?

abeatrix (approved):
Thanks for the great works! The codebase looks so much cleaner 😃 

I tried 
- log out and log in with token, and github redirect
- switch between instances
- chat
- commands: /doc and /explain
- clicking on the codebase to see the pop-up
and they all work as expected.

qq/nit: the non-title text are also bolded, is that intended?

image

dominiccooney:
@abeatrix thank you for the thorough testing!

> qq/nit: the non-title text are also bolded, is that intended?

This is intended, the notices and popups use `font-weight: 500` for text.

> Woops I just tested autocompletes, and that is not working after I switched to a new instance:

I can reproduce that if I use fireworks while signed into s2. But `main` is also broken so not related to this change... Autocompletes without any completion provider overrides work for me. CC @philipp-spiess , should fireworks work with s2?

abeatrix (approved):
Sorry for not checking main, thanks for confirming that the bug is not related to this PR :)

chenkc805:
Woo!

Issue #1407, opened 2023-10-17
Cody: Apply fixups by default


umpox:
Fixups should apply by default. We should provide the codelens options:
- "Done" -> (Hide codelens)
- "Diff"
- "Edited by Cody" -> non-clickable

Undo/Revert should be possible using normal undo flow

toolmantim:
I put together a design for this @umpox, including the ability to retry with an updated prompt:

https://www.figma.com/file/kcXB8qZDpOGqWImwuvX6yq/Auto-applied-edits%2C-retryable-w%2F-new-updated-prompt?type=design&node-id=0-1&mode=design

It does include the idea of "Undo" and "Retry". Retry allows you to redo the prompt and try again.

The way it's designed, "Retry" does an undo then reattempts the edit with the new prompt, which is what a lot of other code AI tools are doing. But if it's not possible to implement "Undo" at all, we could have "Retry" just edit over the top, and I think the UI labelling still works.

> "Edited by Cody" -> non-clickable

I think every code lens action has to map to an action and be clickable.

Instead of "Diff" we can make the first "Edits Applied" action show the diff instead. That would be slightly less discoverable, but the diff view isn't great to use, and most people have a git diff too… so I think that's the right balance?

umpox:
@toolmantim Thanks, will update some things!

PR #1411, opened 2023-10-17
Cody: Apply fixups by default

umpox:
closes https://github.com/sourcegraph/cody/issues/1407

## Description

This PR:
- Applies fixups by default
- Adds support for applying fixups in the background, when the target file isn't visible. Instead of opening it and changing the users' flow, we do a workspace edit and notify the user when done.
- Adds a codelens state where the fixup is applied, but can be differed, regen etc. To avoid this being an annoyance, this codelens is removed when a user saves the document, or edits part of the fixup.
- Via the Fixup view, adds the functionality to manage applied fixups by accepting them all or accepting all in a file.
- Fixes some over-logging for telemetry

## Screenshot
image

## Video (Fixup - Happy path)

https://github.com/sourcegraph/cody/assets/9516420/a515bf02-ac64-45e3-a92b-9d196da77084

## Video (Fixup - Happy path (in background))

https://github.com/sourcegraph/cody/assets/9516420/9fc24cf2-b509-43dd-8346-b0c00188597f

## Video (Fixup - Unhappy path - Undo)

https://github.com/sourcegraph/cody/assets/9516420/421880d4-bf41-4c67-8877-7471e7b5556c

## Video (Fixup - Unhappy path - Retry)

<small>yes, i misspelled code _twice_</small>

https://github.com/sourcegraph/cody/assets/9516420/fefb2c98-c131-4837-bdb0-6cbaa24fb098

## Test plan

Test:
- Instruction fixups (via command palette)
- Documentation command
- Code action fixup
- Fixup UX flow


umpox:
@abeatrix Would appreciate your help testing this too! The /document command worked as expected, but may be a few things to iron out

umpox:
Just seen updated designs (https://github.com/sourcegraph/cody/issues/1407), will update..

abeatrix (commented):
I tried to run `Ask Cody to Fix` and it works, but it'd add a new line at the start of the new replacement code, and clicking `undo` would not remove the new line added by fixup 👀

toolmantim (commented):
In general this is looking great! I think this auto-apply is way better than having to hit the button. And using both the undo button, and command-z, seems to be pretty reliable!

But I did hit some snags while testing. I'm not sure which ones are related to these changes, or already part of the existing edits sorry… but I'll include there here anyhow.

This exception just left it spinning:

Screenshot 2023-10-26 at 9 14 26 pm

I had some decorations that would hang around sometimes:

Screenshot 2023-10-26 at 9 20 06 pm

https://github.com/sourcegraph/cody/assets/153/47bfe87c-462e-4857-bd21-2a7759ad7eab

I did some absolutely horrible things with current edits, and seemed to make things hang (with an error message that might need some reworking):

https://github.com/sourcegraph/cody/assets/153/a35fb628-66db-4269-af18-2fbb319c61a0

Can we remove `/edit` from the instruction on retry?

Screenshot 2023-10-26 at 9 21 39 pm

And then of course I hit the rate limit on dot com 😅

Screenshot 2023-10-26 at 9 23 23 pm

dominiccooney (approved):
Looks good to me.

@toolmantim 's feedback is great, we should trim `/edit` and report network failures better. But those are fine as follow-ups.

umpox:
@toolmantim Thanks for the UX review! I think those are great points, will address them in a follow up and here:

> This exception just left it spinning:

Fixed!

> I did some absolutely horrible things with current edits, and seemed to make things hang (with an error message that might need some reworking):

> And then of course I hit the rate limit on dot com 😅

The hanging and the weird error UX should now be fixed by https://github.com/sourcegraph/cody/pull/1376. Would appreciate thoughts on that UX flow when you have chance (trying to surface fixup errors without using chat)

> Can we remove /edit from the instruction on retry?

Yep, I'll follow up on this in a new PR as we need to make sure it works for all different prefixes `/edit` `/doc`?

> I had some decorations that would hang around sometimes:

Interesting, I haven't been able to replicate that yet but will do some further testing

umpox:
> I tried to run Ask Cody to Fix and it works, but it'd add a new line at the start of the new replacement code, and clicking undo would not remove the new line added by fixup 👀

@abeatrix 
 This should be fixed now!

PR #1412, opened 2023-10-17
Agent log autocomplete

tjdevries:

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -176,53 +176,13 @@ export class Agent extends MessageHandler {
                 })
             }
 
-            await this.recordEvent(`recipe:${data.id}`, 'executed')
+            await this.recordEvent(`recipe:${data.id}`, 'executed', {})
             await client.executeRecipe(data.id, {
                 signal: abortController.signal,
                 humanChatInput: data.humanChatInput,
                 data: data.data,
             })
             return null
-        })
-        this.registerRequest('autocomplete/execute', async (params, token) => {
-            await this.client // To let configuration changes propagate
-            const provider = await vscode_shim.completionProvider()
-            if (!provider) {
-                console.log('Completion provider is not initialized')
-                return { items: [] }
-            }
-            const document = this.workspace.getDocument(params.filePath)
-            if (!document) {
-                console.log('No document found for file path', params.filePath, [...this.workspace.allFilePaths()])
-                return { items: [] }
-            }
-
-            const textDocument = new AgentTextDocument(document)
-
-            try {
-                if (params.triggerKind === 'Invoke') {
-                    await provider.manuallyTriggerCompletion()
-                }
-                const result = await provider.provideInlineCompletionItems(
-                    textDocument,
-                    new vscode.Position(params.position.line, params.position.character),
-                    {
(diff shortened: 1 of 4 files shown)


agent/src/agent.ts
```diff
@@ -267,6 +228,97 @@ export class Agent extends MessageHandler {
...
+                    id: item.id,
+                    languageId: item.languageId,
+                    // TODO: local <- i'm not even sure what this one is
+                    // TODO: embeddings
+                    // TODO: source
+                    // TODO: charCount
+                    // TODO: lineCount
+                    // TODO: multilineMode
+                    // TODO: providerIdentifier
+                },
```
philipp-spiess, line 314: Some of these are highly specific to the implementation, it wouldn't really make sense to try and compute them here. I remember that we already exposed access to that via an additional object in the return type of `provideInlineCompletionItems` (c.f https://sourcegraph.com/github.com/sourcegraph/cody@ps/ac-tracking/-/blob/vscode/src/completions/inline-completion-item-provider.ts?L340%3A1-343%3A14). Looking at the `params` of this object should give you most of the things you need: https://sourcegraph.com/github.com/sourcegraph/cody@ps/ac-tracking/-/blob/vscode/src/completions/logger.ts?L31-44 (this this last link is the best description for the shape of metadata we add to the `suggest` event).

I think mid term it would be better to add an event for you to mark something as suggested and accepted and unify the logging to the vscode implementation. There's a ton of bookkeeping that we'd avoid this way and adding new fields (e.g. treesitter based computed trigger types) will _just work_. 

To answer your questions:

>  // It looks like "extensionDetails" is in the contextSummary?
>                 // Why would it be in the contextSummary instead of top level, like the others?

Maybe I misunderstand but `contextSummary` is only a small object that has information on what type of code snippets we've added to compute the request, it shouldn't have anything else inside it: https://sourcegraph.com/github.com/sourcegraph/cody@ps/ac-tracking/-/blob/vscode/src/completions/context/context.ts?L24%3A1-30%3A1

>      // TODO: local <- i'm not even sure what this one is
`local` here refers to local editor context (= snippets from other open tabs)

agent/src/protocol.ts
```diff
@@ -83,10 +83,17 @@ export type Notifications = {
     'textDocument/didClose': [TextDocument]
 
     '$/cancelRequest': [CancelParams]
+
     // The user no longer wishes to consider the last autocomplete candidate
     // and the current autocomplete id should not be reused.
     'autocomplete/clearLastCandidate': [null]
 
+    // Autocomplete request is suggested
+    'autocomplete/suggested': [AutocompleteSuggested]
```
philipp-spiess, line 92: The terminology for this on the VS Code side is `handleDidShowCompletionItem` if that helps (we can't use this yet though so the current vscode implementation will always log a completion as shown when it was returned from the callback and is still valid)

https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.inlineCompletionsAdditions.d.ts#L48

olafurpg, line 92: What is the contract if you display the suggestion multiple times? For example, you can cycle between autocomplete items with alt-[ and alt-] in JetBrains, so each item can appear multiple times with different display duration each time. I'm not sure exactly how we track telemetry for this case right now but it would be good to support the ideal/recommended telemetry with this new solution.

PR #1414, opened 2023-10-17
fix: swap custom prompts and default prompts order

abeatrix:
RE: https://sourcegraph.slack.com/archives/C052G9Y5Y8H/p1697536651010599?thread_ts=1697536557.499619&cid=C052G9Y5Y8H

fix: swap custom prompts and default prompts order 

The order of adding custom prompts and default prompts to allCommands map is swapped to prioritize default prompts over custom.

This fixes an issue where custom prompts were overriding default prompts with the same name.

Loom provided by Vincent: https://www.loom.com/share/d8aea1426d4b409a86fde3eaefc76ed0?sid=2058611d-5f2d-4457-b3cf-6490ceefe08b

## Test plan

- Create a custom command and name it `explain`
- Before: you will be running the custom command when trying to run `/explain`
- After: the new custom command will not override the built-in `/explain` command

### Before

image

### After

image


umpox (approved):

evict (approved):
Nice! 🚀

PR #1416, opened 2023-10-17
fix: unresponsive selections in custom command menu

abeatrix:
RE: https://sourcegraph.slack.com/archives/C05AGQYD528/p1697105184992049

This PR fixes the issuses reported by Philipp where `` does not work in Custom Commands Menu. This issue was caused by the previous update when `description` has become a required field. 

- Updated options to include `description` field
- Now check for the description option instead of labels when checking the selected option in custom command menu.
- Also fixed `last five used custom commands` not showing up in the same menu

This fixes an issue where commands with null descriptions would not be run when selected.

image

## Test plan

1. Open the Custom Commands Menu from right clicking in your editor and then select `Custom Commands`

image

2. At the bottom you will find `Configure Custom Commands...` and `New Custom Command...`

image

These options should now work when building from this branch.

Other commands should still work as expected.

### DIFF
--- a/vscode/src/custom-prompts/CommandsController.ts
+++ b/vscode/src/custom-prompts/CommandsController.ts
@@ -283,16 +283,16 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
 
             // Show the list of prompts to the user using a quick pick
             const selected = await showCustomCommandMenu([...promptItems])
-            const commandKey = selected?.description || selected?.label
+            const commandKey = selected?.description
 
             if (!commandKey) {
                 return
             }
 
             switch (commandKey.length > 0) {
-                case commandKey === addOption.label:
+                case commandKey === addOption.description:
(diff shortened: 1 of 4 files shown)


vscode/src/custom-prompts/CommandsController.ts
```diff
@@ -285,14 +285,14 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
             const selected = await showCustomCommandMenu([...promptItems])
             const commandKey = selected?.description
```
abeatrix, line 286: Here we are looking for `description` when we are checking `label` below, which caused the issue because these options did not have description fieldss

philipp-spiess (approved):
letsgoooooooo this works now nice

valerybugakov (commented):
Nice! I bumped into it a couple of times.

PR #1425, opened 2023-10-18
Agent: add new `git/codebaseName` request

mkondratek:
Fixes https://github.com/sourcegraph/sourcegraph/issues/56929. 

PR to jetbrains: https://github.com/sourcegraph/jetbrains/pull/42

## Test plan 
1. Unit tests, CI
2. Manual testing:
    1. Checkout these changes 
    2. Checkout the corresponding changes in jetbrains repo
    3. `runIde` with the local clone of Cody
    4. set git remote to `ssh://git@gitlab-my-company.net:20022/path/repo.git`
    5. verify that context selection contains `gitlab-my-company.net/path/repo`

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -268,7 +268,7 @@ export class Agent extends MessageHandler {
             return typeof result === 'string' ? result : null
         })
 
-        this.registerRequest('git/codebaseName', ({ cloneURL }) => {
+        this.registerRequest('git/codebaseName', ([cloneURL]) => {
             const result = convertGitCloneURLToCodebaseName(cloneURL)
             return Promise.resolve(typeof result === 'string' ? result : null)
         })
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -268,7 +268,7 @@ export class Agent extends MessageHandler {
             return typeof result === 'string' ? result : null
         })
 
-        this.registerRequest('convertGitCloneURLToCodebaseName', ({ cloneURL }) => {
+        this.registerRequest('git/codebaseName', ({ cloneURL }) => {
             const result = convertGitCloneURLToCodebaseName(cloneURL)
             return Promise.resolve(typeof result === 'string' ? result : null)
         })
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -2,6 +2,7 @@ import path from 'path'
 
 import * as vscode from 'vscode'
 
+import { convertGitCloneURLToCodebaseName } from '@sourcegraph/cody-shared/dist/utils'
 import { Client, createClient } from '@sourcegraph/cody-shared/src/chat/client'
 import { registeredRecipes } from '@sourcegraph/cody-shared/src/chat/recipes/agent-recipes'
 import { SourcegraphNodeCompletionsClient } from '@sourcegraph/cody-shared/src/sourcegraph-api/completions/nodeClient'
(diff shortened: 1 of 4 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -356,6 +356,10 @@ export class SourcegraphGraphQLAPIClient {
         )
     }
 
+    public convertGitCloneURLToCodebaseName(cloneURL: string) : string | null {
```
olafurpg, line 359: This function seems unnecessary. We should be able to call `convertGitCloneURLToCodebaseName` directly from agent.ts

agent/src/agent.ts
```diff
@@ -267,6 +268,11 @@ export class Agent extends MessageHandler {
             return typeof result === 'string' ? result : null
         })
 
+        this.registerRequest('convertGitCloneURLToCodebaseName', ({ cloneURL }) => {
```
olafurpg, line 271: Conventionally, we use a prefix in the request name. I propose we use something like
```suggestion
        this.registerRequest('git/codebaseName', ({ cloneURL }) => {
```

dominiccooney:
Thanks for doing this. There's other problems with the renaming heuristic, like it requires remotes in the first place; it needs names with slashes (IIRC); etc. But getting everyone using the same thing is a good start.

lib/shared/src/utils.test.ts
```diff
@@ -9,6 +9,12 @@ describe('convertGitCloneURLToCodebaseName', () => {
         )
     })
 
+    test('converts GitHub SSH URL with the port number', () => {
+        expect(convertGitCloneURLToCodebaseName('ssh://git@gitlab-my-company.net:20022/path/repo.git')).toEqual(
```
mkondratek, line 13: this is the original url that caused the issue in the first place

agent/src/agent.ts
```diff
@@ -267,6 +268,11 @@ export class Agent extends MessageHandler {
             return typeof result === 'string' ? result : null
         })
 
+        this.registerRequest('git/codebaseName', ([cloneURL]) => {
+            const result = convertGitCloneURLToCodebaseName(cloneURL)
+            return Promise.resolve(typeof result === 'string' ? result : null)
```
dominiccooney, line 273: Just return the result; you don't need a resolved Promise here.

dominiccooney (changes requested):

agent/src/agent.ts
olafurpg: I think you do, `registerResult` expects a `Promise`-like object. We use `.then` on the result.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -44,6 +44,8 @@ export type Requests = {
     'graphql/getRepoIdIfEmbeddingExists': [{ repoName: string }, string | null]
     'graphql/getRepoId': [{ repoName: string }, string | null]
 
+    'git/codebaseName': [[cloneURL: string], string | null]
```
olafurpg, line 47: Is there a reason this is an array instead of object?
```suggestion
    'git/codebaseName': [{cloneURL: string}, string | null]
```

mkondratek: I was initially attempting to use `{ cloneURL: string }` as the argument type here. However, the issue was that despite the specified argument type (`string`), the method `convertGitCloneURLToCodebaseName(cloneURL: string)` was receiving an object instead. It caused `TypeError: undefined is not a function` at `cloneURL.replace(...)` call.

mkondratek: I guess I can use an object, and change the argument type in jetbrains to smth like:
```java
public class CloneUrl {
  public final String url;

  public CloneUrl(String url) {
    this.url = url;
  }
}
```

mkondratek: I did it like this for consistency.

agent/src/agent.ts
dominiccooney: Missed that the function was not async. Sorry for the spam.

PR #1427, opened 2023-10-18
Explain code action: Always send to sidebar

umpox:
## Description

To remove a dependency on inline chat, we always send the code action to the sidebar now.

## Test plan

Tested by running the code action and expecting the chat to open in the sidebar:
- When sidebar closed
- When sidebar open


abeatrix (approved):
Works for me! 

nit: the format looks a bit weird, maybe add a space before the error?

image

PR #1430, opened 2023-10-19
Autocomplete: update the outdated setting description

valerybugakov:
Updated the outdated setting description. The `fireworks` provider is stable.

## Test plan

n/a

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -924,7 +924,7 @@
             "llama-code-13b",
             "llama-code-13b-instruct"
           ],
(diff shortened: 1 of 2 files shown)


camdencheek (approved):

vscode/src/sort-algos.ts
philipp-spiess, line 1: 🤔 @valerybugakov whoopsie

valerybugakov: 🫠

PR #1436, opened 2023-10-19
Autocomplete: Log which retrieval strategy is used

philipp-spiess:
Adds a new `strategy` string to the context summary field so we can differentiate between lsp-light and bfg

## Test plan

Screenshot 2023-10-19 at 15 00 39


olafurpg (approved):
Nice! Thank you

PR #1437, opened 2023-10-19
BFG: use RFC 795 naming conventions

olafurpg:
Previously, we used an ad-hoc naming convention for BFG binaries. Going forward, we will publish binaries using the naming conventions from RFC
795. The new binaries are named

- bfg-macos-arm64
- bfg-macos-x86
- bfg-linux-x86
- bfg-windows-x86

## Test plan

Manually tested by running plugin with BFG enabled and confirmed the following tree structure in the global cache directory
```
❯ pwd
/Users/olafurpg/Library/Application Support/Code/User/globalStorage/sourcegraph.cody-ai
❯ tree
.
└── bfg
   └── bfg-0.1.0-macos-arm64
```

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -937,7 +937,7 @@
           "markdownDescription": "Rank autocomplete results with tree-sitter."
         },
         "cody.autocomplete.experimental.graphContext": {
-          "type": "enum",
+          "type": "string",
           "default": null,
           "enum": [
             null,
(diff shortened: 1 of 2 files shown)


vscode/package.json
```diff
@@ -937,7 +937,7 @@
           "markdownDescription": "Rank autocomplete results with tree-sitter."
         },
         "cody.autocomplete.experimental.graphContext": {
-          "type": "enum",
+          "type": "string",
```
olafurpg, line 940: @philipp-spiess This diff is unrelated to the PR but I was getting warnings from VS Code about this having the value `"enum"`

philipp-spiess: Oh yeah this was an oversight from my side, thanks for catching it!

philipp-spiess (approved):
<3

PR #1440, opened 2023-10-19
fix: update post process logic for claude instant

abeatrix:
fix: use full infill block and adjust prefix

- fix prompt structure order, context should be shared after the role assigning
- empty prefix if it contains the infill block for issue where first line is duplicated 
- Simplify prompt messages to focus on providing clear context
- Add CLOSING tag for Assistant's infill block to preserve ending whitespace 
- This also fixed issues where multi-line autocomplete gets cut off.
- remove tags for context code as that might cause confusion for claude

## Test plan

Fix: show multi-line completion after `{`:

image

Fix: show multi-line completions the line after `{`

image

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -96,7 +96,7 @@ export class AnthropicProvider extends Provider {
         const introMessages: Message[] = [
             {
                 speaker: 'human',
-                text: `You are a code completion AI designed to take the surrounding code and shared context into account in order to predict and suggest high-quality code that works and fits seamlessly with surrounding code if any or use best practice and nothing else.`,
+                text: 'You are a code completion AI designed to take the surrounding code and shared context into account in order to predict and suggest high-quality code that works and fits seamlessly with surrounding code if any or use best practice and nothing else.',
             },
             {
                 speaker: 'assistant',
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -216,6 +216,10 @@ export class AnthropicProvider extends Provider {
             // leading `\n` followed by whitespace that Claude might add.
             completion = completion.replace(/^\s*\n\s*/, '')
         } else {
+            // prevent normalizeStartLine from removing the starting new line
+            if (completion.startsWith('\n')) {
+                completion = '\n' + completion.trimStart()
+            }
             completion = trimLeadingWhitespaceUntilNewline(completion)
         }
 
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -67,9 +67,9 @@ export class AnthropicProvider extends Provider {
         const { head, tail, overlap } = getHeadAndTail(this.options.docContext.prefix)
 
         // Infill block represents the code we want the model to complete
-        const infillBlock = tail.trimmed
+        const infillBlock = `${tail.raw}`
         // code before the cursor, without the code extracted for the infillBlock
-        const infillPrefix = head.raw
+        const infillPrefix = head.raw?.startsWith(tail.trimmed) ? '' : `${head.raw}`
         // code after the cursor
         const infillSuffix = this.options.docContext.suffix
         const relativeFilePath = vscode.workspace.asRelativePath(this.options.document.fileName)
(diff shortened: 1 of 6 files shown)


valerybugakov (commented):
Could we separate the bug fix from the prompt changes and put the prompt changes behind the feature flag so that we can test how they affect CAR in production before making it a default? 

The prompt updates look logical, and I expect them to perform well, but with LLMs, we never know how newly added tokens will affect the output. We have quite a few changes here, from the sentence structure updates to changing the code block tags. Splitting this PR in two would make it easier to reason about CAR changes later:

1. Minimal prompt change required to fix a bug + logic updates.
2. Prompt updates behind the feature flag.

abeatrix:
@valerybugakov that's fair, I can do that next!

Have you tried the change in this PR yet? Is it working for you?

philipp-spiess:
TBH I’m happy to move fast with this given how big our regression is right now. My only thinking is that:

- we should run our completion test suite to ensure we don't add known issues
- we delete old code that is no longer needed (like the head/tail splitting of the prefix) to avoid introducing code dept

once you're happy with the results we should push out a patch release.

philipp-spiess (approved):
preemptive stamp but lets make sure the tests are good

abeatrix:
@philipp-spiess @valerybugakov I tried to make minimal prompt change as possible but the one we had just doesn't work anymore because of the lines and whitespaces removed on claude side. I added <_cursor_> to the end of the infill block to make sure the spaces are not removed, and it seems to work fine on my side, but i wasn't able to get the eval tests to work to confirm if there is other regression, but definately better than what is in prod right now. If this is good for now, please feel free to merge and make a patch release on monday to make sure our users are not getting affected, and I can do any required follow up works

philipp-spiess:
@abeatrix I pushed a fix for the local completions test and there seem to be some issues with the newly used XML leaking into the completion:

Screenshot 2023-10-23 at 11 37 46

Also the test case with a comment followed by a `\n` isn't working well either :/ 

Screenshot 2023-10-23 at 11 38 47

abeatrix:
@valerybugakov After taking a look at the issue with @philipp-spiess ,  it looks like the root cause of the issue was the new `parseAndTruncateCompletion` that doesn't work well with the `this.postProcess()` for anthropic. I tried added some workaround in my lastest commits and that seems to work (see the updated infill test result). Can you take a look and see if this change makes sense to you? Since this is a P1 issue, do you mind taking over of this PR while I'm away to make sure we can have this issue resolved asap?

vscode/src/completions/providers/anthropic.ts
```diff
@@ -67,7 +67,7 @@ export class AnthropicProvider extends Provider {
         const { head, tail, overlap } = getHeadAndTail(this.options.docContext.prefix)
 
         // Infill block represents the code we want the model to complete
-        const infillBlock = tail.trimmed
+        const infillBlock = tail.trimmed.endsWith('{\n') ? tail.trimmed.trimEnd() : tail.trimmed
```
abeatrix, line 70: this allows claude to response with multi-line completion on new line after the `{` bracket

vscode/src/completions/providers/anthropic.ts
```diff
@@ -216,6 +216,10 @@ export class AnthropicProvider extends Provider {
             // leading `\n` followed by whitespace that Claude might add.
             completion = completion.replace(/^\s*\n\s*/, '')
         } else {
+            // prevent normalizeStartLine from removing the starting new line
+            if (completion.startsWith('\n')) {
+                completion = '\n' + completion.trimStart()
```
abeatrix, line 221: `normalizeStartLine` in `parseAndTruncateCompletion` would remove the `\n` if it follows by indentation when prefix didn't end with closing bracket 🤔  This solves the aforementioned issue

valerybugakov:
@abeatrix, thank you for sharing the context! I'm debugging it locally and will merge this if I can confirm your findings 👍

vscode/src/completions/providers/anthropic.ts
valerybugakov: Updated the logic to `normalizeStartLine` only for multiline completions like before my changes in [this PR](https://github.com/sourcegraph/cody/pull/1404/files). It fixes the issue without adding the additional logic. Thank you for narrowing down this regression!

valerybugakov (approved):
After my change, I verified that the cases mentioned here still work locally as expected. Regenerated completions using the `generate:completions` to ensure we do not have any leaking XML tags.

valerybugakov:
The PR is blocked by eslint errors on main. Fixing it here: https://github.com/sourcegraph/cody/pull/1471

PR #1441, opened 2023-10-19
feat: auto format code after applying edits

abeatrix:
Reported / requested in https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1697672986475259?thread_ts=1697672318.465339&cid=C05MW2TMYAV and https://sourcegraph.slack.com/archives/C03CSAER9LK/p1697148745798389?thread_ts=1696469865.956429&cid=C03CSAER9LK

fix: format code after applying edits

Automatically format the edited code after applying edits. 

A new range is calculated to format the selected area where fixes were applied.

The document is then formatted using the new range.

This commit fixes an issue where code was not formatted after applying quick fixes.

### Demo

https://github.com/sourcegraph/cody/assets/68532117/f5179669-5521-474e-bdde-38f328bb078b

## Test plan

Perform an edit and confirm the edited area was formatted correctly.

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -305,6 +305,16 @@ export class FixupController
             const codeCount = countCode(replacementText)
             const source = task.source
             telemetryService.log('CodyVSCodeExtension:fixup:applied', { ...codeCount, source })
+
+            // format the selected area after applying edits
+            const range = new vscode.Range(
+                new vscode.Position(task.selectionRange.start.line, 0),
+                new vscode.Position(
+                    task.selectionRange.start.line + codeCount.lineCount,
+                    task.selectionRange.end.character
+                )
+            )
+            await vscode.commands.executeCommand('editor.action.formatDocument', range)
         }
 
         // TODO: is this the right transition for being all done?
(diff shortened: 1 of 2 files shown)


vscode/src/non-stop/FixupController.ts
```diff
@@ -305,6 +305,16 @@ export class FixupController
...
+
+            // format the selected area after applying edits
+            const range = new vscode.Range(
+                new vscode.Position(task.selectionRange.start.line, 0),
+                new vscode.Position(
+                    task.selectionRange.start.line + codeCount.lineCount,
+                    task.selectionRange.end.character
+                )
+            )
+            await vscode.commands.executeCommand('editor.action.formatDocument', range)
```
valerybugakov, line 317: Wow, should we try the same for multiline completions? 🤩
By the way, what happens if there's no formatter available for the current language? Does it throw or do nothing?

valerybugakov (approved):
Great find!

vscode/src/non-stop/FixupController.ts
abeatrix: I didn't get any errors when running on a file with no lsp installed. I think it will just fall back to the vs code built in formatted? 

image

users will get a notification if they have more than one formatted installed though, so not sure if that s something we can use for autocomplete :( https://code.visualstudio.com/updates/v1_33#_default-formatter-selection

philipp-spiess:
@abeatrix left some feedback here: https://sourcegraph.slack.com/archives/C05AGQYD528/p1697791057625419 

I think we should only do that if formatting on save is enabled

Issue #1443, opened 2023-10-20
App-less: Local Indexes without App


dominiccooney:
In #632 we removed App from the onboarding flow and found it improved the product. Now we want to remove app entirely

Related links:
- [Planning spreadsheet.](https://docs.google.com/spreadsheets/d/1-Nd1FSEURFBgLtTSVMt3i3OypPaR4FSN4WgCyJoSrBI/edit#gid=0)

dominiccooney:
The only thing remaining here is to toast users to remove App, the language is in #1966.

beyang:
Can close?

dominiccooney:
Let's close this because technically it is done: You can have local embeddings indexes without App. I will keep #1966 around for the toast to remove app.

PR #1444, opened 2023-10-20
App-less: Add a client for generating embeddings

dominiccooney:
Embeddings convert text into numeric vectors. When Cody App generates the database of local embeddings, it breaks source files into chunks and embeds them using Cody Gateway; at query time, it embeds the query and finds chunks which are nearby in embedding space. This adds a class for generating embeddings on the client using Cody Gateway.

Part of #1443

## Test plan

```
pnpm test -- EmbeddingProvider.test.ts
```

This test can also hit the production frontend if you run it with a `SOURCEGRAPH_DOTCOM_TOKEN=sgp_...` environment variable set.

### DIFF
--- a/lib/shared/package.json
+++ b/lib/shared/package.json
@@ -30,6 +30,7 @@
     "isomorphic-fetch": "^3.0.0",
     "lodash": "^4.17.21",
     "marked": "^4.0.16",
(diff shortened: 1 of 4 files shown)


dominiccooney:
Let me close this for now, we might be taking a different direction with this work.

PR #1445, opened 2023-10-20
Fixup: Only apply formatting changes to a range, and clean the undo stack

umpox:
## Description

This PR:
- Only applies the formatting changes to the fixup selection. By calling the format provider directly and filtering the edit down to the selection range. VS Code only provides `formatDocument` and `formatSelection` direct commands, so this is the approach that gives us the control we need.
- Preserves a single undo/redo action for both the fixup and the formatting change. A user would have to undo a fixup twice if it also applied formatting. This should be an implementation detail and not visible to the user. Now both changes are consolidated into a single undo/redo action.

https://github.com/sourcegraph/cody/assets/9516420/7f5d6b9f-7c2b-4594-b1ad-25ac2d0b575b

## Test plan

Tested locally, see demo

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -291,7 +291,11 @@ export class FixupController
 
         editor.revealRange(task.selectionRange)
 
-        const editOk = task.insertMode ? await this.insertEdit(editor, task) : await this.replaceEdit(editor, diff)
+        // We will format this code once applied, so we avoid placing an undo stop after this edit to avoid cluttering the undo stack.
+        const applyEditOptions = { undoStopBefore: true, undoStopAfter: false }
+        const editOk = task.insertMode
+            ? await this.insertEdit(editor, task, applyEditOptions)
+            : await this.replaceEdit(editor, diff, applyEditOptions)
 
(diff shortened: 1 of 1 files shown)


umpox:
@abeatrix Other thought, should we add another code lens state for "Formatting..."? 🤔

vscode/src/non-stop/FixupController.ts
```diff
@@ -307,14 +311,32 @@ export class FixupController
...
                 new vscode.Position(
                     task.selectionRange.start.line + codeCount.lineCount,
                     task.selectionRange.end.character
                 )
             )
-            await vscode.commands.executeCommand('editor.action.formatDocument', range)
+
+            const formattingChanges = (
+                await vscode.commands.executeCommand<vscode.TextEdit[]>(
+                    'vscode.executeFormatDocumentProvider',
```
umpox, line 324: There's actually also `executeFormatRangeProvider` but it seems new and possibly broken. I couldn't get it to work reliably unless I passed in the entire document range, probably worth debugging further. Maybe it needs the formatter to support it, and Prettier doesn't right now?

abeatrix (approved):

PR #1446, opened 2023-10-20
Cody eval TypeScript dataset

taras-yemets:
Iteration 2/n on adding TypeScript dataset to Cody completion evaluation suite.

## Test plan
This PR adds completion evaluation datasets.

### DIFF
--- a/vscode/test/benchmark/datasets/typescript/factory-same-file/config.json
+++ b/vscode/test/benchmark/datasets/typescript/factory-same-file/config.json
@@ -1,0 +1,8 @@
+{
+  "entryFile": "generate.ts",
+  "openFiles": [],
+  "closedFiles": [],
+  "solutionFile": "solution.ts",
+  "testFile": "test.ts",
+  "testCommand": "ts-node test.ts"
+}
--- a/vscode/test/benchmark/datasets/typescript/factory-same-file/generate.ts
+++ b/vscode/test/benchmark/datasets/typescript/factory-same-file/generate.ts
@@ -1,0 +1,43 @@
+type ShapeType = 'triangle' | 'rectangle' | 'circle'
+
+export abstract class Shape {
+    constructor(public name: ShapeType) {}
+
+    public abstract calcPerimeter(): number
+}
+
+class Triangle extends Shape {
+    constructor(private sides: number[]) {
+        super('triangle')
+    }
+
+    calcPerimeter(): number {
+        return this.sides[0] + this.sides[1] + this.sides[2]
+    }
+}
+
+class Rectangle extends Shape {
+    constructor(private sides: number[]) {
+        super('rectangle')
+    }
+
+    calcPerimeter(): number {
+        return (this.sides[0] + this.sides[1]) * 2
+    }
+}
+
+class Circle extends Shape {
+    constructor(private radius: number) {
+        super('circle')
+    }
+
+    calcPerimeter(): number {
+        return 2 * Math.PI * this.radius
+    }
+}
+
+interface ShapeFactory {
+    createShape(...sides: number[]): Shape
+}
+
+export const shapeFactory: ShapeFactory = {█
--- a/vscode/test/benchmark/datasets/typescript/factory-same-file/solution.ts
+++ b/vscode/test/benchmark/datasets/typescript/factory-same-file/solution.ts
@@ -1,0 +1,56 @@
+type ShapeType = 'triangle' | 'rectangle' | 'circle'
+
+export abstract class Shape {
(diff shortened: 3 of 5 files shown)


taras-yemets:
@slimsag, what do you think about the existing (test cases already in `main` plus added in this PR) TypeScript dataset quality? Does it test real-world scenarios, in your opinion? 

As for now, this PR tests mainly function declaration body completions (multiline completions are usually [expected](https://docs.google.com/document/d/13kYxtJcgoDAY45vulenwGtOW0KnugJPD31MXx-sgx2c/edit#bookmark=id.t43edljoxszv) in such cases). Do these test cases seem too generic, and thus LLMs should be familiar to LLMs? 
There is no need to review each test case individually; I'm seeking high-level feedback.

vscode/test/benchmark/datasets/typescript/adapter-open-file/solution.ts
```diff
@@ -0,0 +1,35 @@
+import {
+    EmailNotification,
+    Notification,
+    SecretMessenger,
+    SecretMessengerAdapter,
+    SlackNotification,
+} from './notification'
+import { User } from './types'
+
+export function notifyUsers(users: User[], message: string): void {
```
slimsag, line 10: This solution feels large to me. Like, I'm not sure I would expect Cody to autocomplete **the entire function** for me. Maybe that's just my perception of how Cody should behave.. but I think an autocomplete for ~25 lines would be overwhelming

slimsag:
@taras-yemets In case it helps, here's the list of use-cases I came up with for Go that I am working through (I won't do all of these, just using this list for inspiration):

- [ ] Wrap error
- [ ] Check if channel is closed
- [ ] Check if struct is empty
- [ ] Implement an interface
- [ ] Use context
- [ ] Cast number to float
- [ ] Initialize a map
- [ ] Initialize a channel
- [ ] Initialize a buffered channel
- [ ] Non-blocking channel send
- [ ] Select channel
- [ ] Parallel spawn of goroutines
- [ ] Struct to map
- [ ] Struct to JSON
- [ ] Merge two maps
- [ ] Merge two structs
- [ ] Clear map
- [ ] Get map keys
- [ ] Measure time
- [ ] Write test function
- [ ] Iterate map
- [ ] HTTP write json
- [ ] HTTP read SON
- [ ] HTTP check status code
- [ ] HTTP write error
- [ ] HTTP server
- [ ] HTTP post request
- [ ] HTTP get request
- [ ] HTTP get request write file
- [ ] Exec command and get output
- [ ] Exec command error get stdout
- [ ] template.New().ParseFiles()
- [ ] template.New().Funcs().ParseFiles()

vscode/test/benchmark/datasets/typescript/adapter-open-file/solution.ts
taras-yemets: I consider this and other similar test cases as more complex ones. It aims to test if Cody can process the relatively extensive context, understand it, and develop a solution that at least compiles. I ran this test a couple of times, and in most cases, the solution compiles, and quite often it's even correct. So it can be a good test case to catch potential regressions.

taras-yemets: My main motivation for adding a few whole function body completions comes from internal discussions ([thread](https://sourcegraph.slack.com/archives/C05TVUMBES1/p1698138859488999?thread_ts=1697762361.821269&cid=C05TVUMBES1), [RFC](https://docs.google.com/document/d/13kYxtJcgoDAY45vulenwGtOW0KnugJPD31MXx-sgx2c/edit#bookmark=id.t43edljoxszv)) about multiline completing being expected in function bodies. 
Though I agree that the multiline completions we want to test may be smaller (incremental, as you suggest in https://github.com/sourcegraph/cody/pull/1446#issuecomment-1778049340) and tests may check only if the suggested code compiles.

taras-yemets:
Thanks for your feedback, @slimsag!
The suggestions and use cases you provided are very helpful, and I'll definitely add test cases to cover most of them!

vscode/test/benchmark/datasets/typescript/adapter-open-file/solution.ts
slimsag: Sounds great to me!

taras-yemets:
Moved to a separate repository: https://github.com/sourcegraph/coder-eval/pull/3

taras-yemets:
Moved to a separate repository: https://github.com/sourcegraph/coder-eval/pull/3

PR #1447, opened 2023-10-20
Enable e2e tests for Windows on bots

DanTup:
Enables the e2e bots on Windows.

## Test plan

Tested by the bots - the status checks on this PR should have a Windows e2e bot that is green.

image

### DIFF
--- a/vscode/test/e2e/chat-history.test.ts
+++ b/vscode/test/e2e/chat-history.test.ts
@@ -47,8 +47,13 @@ test('shows chat history in sidebar and update chat panel correctly', async ({ p
     await expect(page.getByRole('tab', { name: 'Hey' })).toBeVisible()
 
     // Click the delete chat button twice to remove the chats we submitted
+    // Check for counts to ensure we wait for the delete to be completed before
+    // trying to click again, or we might quickly click the last one twice
+    await expect(page.getByLabel('Delete Chat')).toHaveCount(2)
+    await page.getByLabel('Delete Chat').last().click()
+    await expect(page.getByLabel('Delete Chat')).toHaveCount(1)
+    await page.getByLabel('Delete Chat').last().click()
+
     // Once the chat history is empty, the 'New Chat' button should show up
-    await page.getByLabel('Delete Chat').last().click()
-    await page.getByLabel('Delete Chat').last().click()
     await expect(page.getByRole('button', { name: 'New Chat', exact: true })).toBeVisible()
 })
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,10 +158,7 @@ jobs:
         if: ${{ failure() }}
         with:
           name: playwright-recordings ${{ matrix.runner }}
-          path: |
-            playwright
-            test-results
-            vscode/test-results
+          path: playwright/**/*.webm
 
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,8 +74,11 @@ jobs:
         if: github.ref == 'refs/heads/main' && (matrix.runner == 'windows' || matrix.runner == 'macos')
 
   test-e2e:
-    runs-on: ubuntu-latest
-    name: 'test-e2e (ubuntu)' # required to match the GitHub PR status check requirement
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu, windows]
+    runs-on: ${{ matrix.runner }}-latest
     timeout-minutes: 15
     permissions:
       id-token: write
@@ -105,6 +108,7 @@ jobs:
           restore-keys: ${{ runner.os }}-pnpm-store-
       - run: pnpm install
       - run: xvfb-run -a pnpm -C vscode run test:e2e
+        if: matrix.runner == 'ubuntu'
       - uses: actions/upload-artifact@v3
         if: ${{ failure() }}
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,8 @@ jobs:
       - run: pnpm install
       - run: xvfb-run -a pnpm -C vscode run test:e2e
         if: matrix.runner == 'ubuntu'
+      - run: pnpm -C vscode run test:e2e
+        if: matrix.runner != 'ubuntu'
       - uses: actions/upload-artifact@v3
         if: ${{ failure() }}
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,10 @@ jobs:
         if: ${{ failure() }}
         with:
           name: playwright-recordings ${{ matrix.runner }}
-          path: playwright/**/*.webm
+          path: |
+            playwright
+            test-results
+            vscode/test-results
 
   build:
     runs-on: ubuntu-latest
--- a/vscode/playwright.config.ts
+++ b/vscode/playwright.config.ts
@@ -5,15 +5,4 @@ export default defineConfig({
     // Give failing tests a second chance
     retries: 2,
     testDir: 'test/e2e',
-    timeout: 120000,
-    expect: {
-        timeout: 10000,
-    },
-    use: {
-        // screenshot: 'only-on-failure',
-        video: 'retain-on-failure',
-        actionTimeout: 200000,
-        navigationTimeout: 600000,
-        trace: 'retain-on-failure',
-    },
 })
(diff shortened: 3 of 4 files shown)


DanTup:
@akalia25 I'm trying to get the e2e tests working on Windows and I'm working through some flakiness. Some of it comes from the log events being validated added in https://github.com/sourcegraph/cody/pull/1295 so I wonder if you might better understand this than me.

Here's an example failure:

```
  2) inline-completion.test.ts:59:5 › inline completion onboarding notice on first completion accept 

    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 0

    @@ -2,7 +2,6 @@
        "CodyVSCodeExtension:completion:accepted",
        "CodyVSCodeExtension:completion:accepted",
        "CodyVSCodeExtension:completion:suggested",
        "CodyVSCodeExtension:completion:suggested",
        "CodyVSCodeExtension:completion:suggested",
    -   "CodyVSCodeExtension:completion:suggested",
      ]
```

With my updates to the test, I expect `suggested` four times, and `accepted` twice. Here is the playright recording of the failure on the bot (note: it doesn't fail for me locally):

[c9456b4be6b73a05fa23480db7d946e8.webm](https://github.com/sourcegraph/cody/assets/1078012/d8ba929c-9122-4e24-bfea-e13e76221027)

In the video I see 4 suggestions (although some are very quick):

1. Click after `<body>` and press `enter`, the completion shows `myFirstCompletion`
2. After pressing `tab` to accept the completion, the 🎉 decoration appears and the completion also triggers again `myNotFirstCompletion` (since the mock server is basic and always provides completions)
3. After typing a and pressing `<enter>`, completion is triggered again showing `myNotFirstCompletion`
4. After accepting the second completion, for the same reason above, it re-triggers and shows `myNotFirstCompletion`

My feeling is that there might be a race somewhere (either in recording the log, or sending the log at all) but I can't find anything and it's tricky to debug since it only happens on the bots. Is this something you're familiar with and do you have any thoughts/suggestions?

DanTup:
@akalia25 I'm tweaking how the mock completion works in https://github.com/sourcegraph/cody/pull/1535 which should resolve these issues with "extra" (and intermittent) completions, so this should no longer be an issue. Sorry for the noise :-)

akalia25:
> #1535

Hey @DanTup Nice to e-meet you! Sorry for the delayed response, as I am on PTO this week.

Regarding the `inline-completion` test being flaky, you're entirely right that there is some race condition that was happening in the backend; that's the reason for this test only, I had to sort the events before checking they were equal ([line ref](https://github.com/sourcegraph/cody/pull/1295/files#diff-ea5400f7ecd7e0d39c1cf4ce8849337ac8ad875197ded0f6505ac4c143ef96c2R53)) because of this race condition.

Thank you for figuring it out and resolving this in #1535 🙇 !

If you have any feedback on the E2E tests, I would love to hear it and how we can improve, as I've been working on maintaining the accuracy of our logged events :)

DanTup:
@akalia25 

> that's the reason for this test only, I had to sort the events before checking they were equal ([line ref](https://github.com/sourcegraph/cody/pull/1295/files#diff-ea5400f7ecd7e0d39c1cf4ce8849337ac8ad875197ded0f6505ac4c143ef96c2R53)) because of this race condition.

Ah, that's good to know - I should be able to remote the sorting then. The main issue I had wasn't about the order, but rather that the inline completion would _sometimes_ re-trigger immediately after accepting, and sometimes not (it was very inconsistent on the bots between macOS/Windows, though pretty consistent for me locally).

My changes in the other PR should address that by making the mock completions smarter/more predictable though.

> If you have any feedback on the E2E tests, I would love to hear it and how we can improve, as I've been working on maintaining the accuracy of our logged events :)

Sure, I'll let you know if anything else comes up. I think mostly the outstanding issues here (to enable on Windows) are just some minor timing differences in the tests and some quirks with focus (seems like sometimes focus isn't automatically added to some boxes before we start typing). I'm hoping I'm almost at the end of all of those now 🙃

kalanchan:
@dominiccooney could you review if the workflow is good? (I don't feel confident touching windows stuff 😅)

dominiccooney (changes requested):
This doesn't look right because it only actually runs the `test:e2e` if `matrix.runner == 'ubuntu'`.

This is fine if we just want to check `pnpm install` is working on Windows, but we should go for it and run the suite.

DanTup:
Ugh, thanks for catching that 🤦‍♂️. I guess didn't read the command properly (in Dart I have a step to run `Xvfb` for Linux and then run the tests in a subsequent command).

Pushed a change, although maybe now we'll find out all the tests still aren't passing on Windows.. I'll work on fixing them if they're red.

DanTup:
One of the failures here is probably that these two entries for readmes are intended to be the same:

image

This might be slightly related to code noted at https://github.com/sourcegraph/cody/pull/2400#issuecomment-1867617956 where leading slashes are being removed for Mac/Linux but not Windows. We need to nail down exactly what we're expecting these fields to contain (and document them on the interfaces) so it's clear where the fix should be.

DanTup:
The issue above was resolved by some other changes on main, so this is now green (after merging with latest).

abeatrix (approved):
just curious, what were the changes on main that fixed the issue?

DanTup:
I think it was https://github.com/sourcegraph/cody/commit/d0998e4f2a0e47af20203a6d28d8d660bf85bdf9 which made some changes to how those URIs were computed. I'll double-check to see.

DanTup:
I've opened https://github.com/sourcegraph/cody/issues/2584 and will look into this separately.

(@dominiccooney merging is blocked on your review.. feel free to merge for me if you're happy with it now 🙂)

DanTup:
Seems there are some new failures here now, taking a look.

```
  1) chat-history.test.ts:6:5 › shows chat history in sidebar and update chat panel correctly ──────

    Error: Timed out 5000ms waiting for expect(locator).toBeVisible()

    Locator: getByRole('button', { name: 'New Chat', exact: true })
    Expected: visible
    Received: hidden
    Call log:
      - expect.toBeVisible with timeout 5000ms
      - waiting for getByRole('button', { name: 'New Chat', exact: true })


      51 |     await page.getByLabel('Delete Chat').last().click()
      52 |     await page.getByLabel('Delete Chat').last().click()
    > 53 |     await expect(page.getByRole('button', { name: 'New Chat', exact: true })).toBeVisible()
         |                                                                               ^
      54 | })
      55 |

        at D:\a\cody\cody\vscode\test\e2e\chat-history.test.ts:53:79

  2) custom-command.test.ts:11:5 › open the Custom Commands in sidebar and add new user command ────

    Error: EBUSY: resource busy or locked, unlink '\\?\C:\Users\RUNNER~1\AppData\Local\Temp\cody-vsced30WGT\Crashpad\metadata'

       at helpers.ts:133

      131 |             }
      132 |
    > 133 |             rmSync(userDataDirectory, { recursive: true })
          |                   ^
      134 |             rmSync(extensionsDirectory, { recursive: true })
      135 |         },
      136 |     })

        at Object.page (D:\a\cody\cody\vscode\test\e2e\helpers.ts:133:19)
```

vscode/test/e2e/chat-history.test.ts
```diff
@@ -47,8 +47,13 @@ test('shows chat history in sidebar and update chat panel correctly', async ({ p
     await expect(page.getByRole('tab', { name: 'Hey' })).toBeVisible()
 
     // Click the delete chat button twice to remove the chats we submitted
-    // Once the chat history is empty, the 'New Chat' button should show up
+    // Check for counts to ensure we wait for the delete to be completed before
+    // trying to click again, or we might quickly click the last one twice
+    await expect(page.getByLabel('Delete Chat')).toHaveCount(2)
```
abeatrix, line 52: Woops! Thanks for fixing this!

abeatrix (approved):
I don't know how to verify the changes on my end since I don't have a window machine but the code lgtm 😀

DanTup:
> I don't know how to verify the changes on my end since I don't have a window machine but the code lgtm 😀

The Windows bots now being green (but previously being red) should verify this.

It'd be really good to get this landed as it's been hanging around quite a while, but I'm not sure if @dominiccooney has been out. Should we dismiss his review or wait?

abeatrix:
I'll try to get it merged since the bots are green, and I don't think it'd block anything/anyone if it's not a required check right?

Plus we can always revert it if it becomes an issue 😁

DanTup:
Yeah, we can revert if there are issues or it turns out to be flaky. I'm happy to investigate any Windows-specific flakes/failures if you see any.

Getting this landed will help avoid new issues (in particular refactors like https://github.com/sourcegraph/cody/pull/2749 that change paths would benefit from having more tests running on Windows).

Thanks! :)

PR #1448, opened 2023-10-20
draft: Chat Improvements in Agent

tjdevries:
# Proposed Changes

Primary goal is to make chat a "first-class citizen" of the agent protocol.

- `Transcript` and `Message` both use an `id` to identify them
- The ChatHandler (name tbd) uses these `id`s to interact with `Transcript` and `Message`
  - This prevents accidental mutation and state changes from within the javascript side that won't get synchronized across the protocol boundary.
  - This makes sure that anything we need to do with `Transcript`s and `Message`s can be done via IDs (and thus, through the protocol without many changes)
- No longer rely on `last` transcript and/or `last` message, but instead always require IDs over the wire and within the agent.
  - This allows us to explicitly control whether recipes/chats inherit the current transcript, and have multiple conversations work simultaneously  
  - This also means that when we send "update" notifications for chat, the client does not have to track which was the latest message sent / etc but instead just updates the corresponding transcript/message using the `id`s provided.

(obviously will not have transcript2 and client2, but just using those for now to test out the differences) 

## Current Status

This is *not* complete. I don't plan on trying to merge with `client2` but this let me try exploring the new API without everything breaking immediately.

My goal is that we primarily:
- Use `id`s to communicate and mutate the transcripts. This prevents us from updating state of an object/class in javascript and not having those updates synced into the clients.
- Make each `chat/` message use `transcriptID` and `messageID` when appropriate.
  - This allows us to do one-shot recipes without sharing the same transcript history (for example, if you have a chat right now and then ask it to do a task, the transcripts are shared unless you first reset the transcript... then you've lost your chat history. This is fundamentally broken and bad behavior).
- Should hopefully simplify a lot of the logic and reasoning because we don't implicitly just use the `last` thing of everything (instead, use IDs. Don't keep global state of `last` thing cause it's too confusing and prevents plenty of meaningful and useful improvements)

I know that I have merge conflicts cause @olafurpg moved the agent protocol, I will fix those next week when I'm back in office. I just wanted to push a draft PR so that we can at least see if this direction is helpful at all before I got through the last bits of the work (that will take the most time of course). I will try and fix up all the VS Code side things required for this as well, or at least provide the same default behavior as before and the VS Code team can update as necessary after that. 

## Test plan

- [ ] DO NOT MERGE

### DIFF
--- a/agent/src/protocol.ts
+++ b/agent/src/protocol.ts
@@ -1,7 +1,8 @@
 /* eslint-disable @typescript-eslint/consistent-type-definitions */
 
 import { RecipeID } from '@sourcegraph/cody-shared/src/chat/recipes/recipe'
-import { ChatMessage } from '@sourcegraph/cody-shared/src/chat/transcript/messages'
+import { ChatMessage, MessageID } from '@sourcegraph/cody-shared/src/chat/transcript/messages'
+import { TranscriptID } from '@sourcegraph/cody-shared/src/chat/transcript/transcript2'
 import { event } from '@sourcegraph/cody-shared/src/sourcegraph-api/graphql/client'
 
 import { CompletionEvent } from '../../vscode/src/completions/logger'
@@ -101,6 +102,9 @@ export type Notifications = {
     // request. The server should never send this notification outside of a
     // 'chat/executeRecipe' request.
     'chat/updateMessageInProgress': [ChatMessage | null]
+
+    'chat/update': [{ transcriptID: TranscriptID; messageID: MessageID; text: string }]
+    'chat/completed': [{ transcriptID: TranscriptID; messageID: MessageID; text: string }]
 
     'debug/message': [DebugMessage]
 }
--- a/agent/src/protocol.ts
+++ b/agent/src/protocol.ts
@@ -44,6 +44,8 @@ export type Requests = {
     'graphql/getRepoIdIfEmbeddingExists': [{ repoName: string }, string | null]
     'graphql/getRepoId': [{ repoName: string }, string | null]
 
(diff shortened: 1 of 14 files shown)


agent/src/protocol.ts
```diff
@@ -44,6 +45,8 @@ export type Requests = {
     'graphql/getRepoIdIfEmbeddingExists': [{ repoName: string }, string | null]
     'graphql/getRepoId': [{ repoName: string }, string | null]
 
+    'chat/complete': [{ transcriptID: string; humanChatInput: string; data: object }, null]
```
tjdevries, line 48: should be deleted

agent/src/protocol.ts
```diff
@@ -100,6 +103,9 @@ export type Notifications = {
     // 'chat/executeRecipe' request.
     'chat/updateMessageInProgress': [ChatMessage | null]
```
tjdevries, line 104: will be marked as deprecated

agent/src/protocol.ts
```diff
@@ -100,6 +103,9 @@ export type Notifications = {
...
 
+    'chat/update': [{ transcriptID: TranscriptID; messageID: MessageID; text: string }]
+    'chat/completed': [{ transcriptID: TranscriptID; messageID: MessageID; text: string }]
```
tjdevries, line 107: chat/newMessage [request, { transcriptID, messageID }]

tjdevries: chat/newTranscript [request, { transcriptID }]

tjdevries: how to move a transcript from the client to the server

tjdevries: const transcriptID = await agent.newTranscript(...);

Issue #1451, opened 2023-10-23
Tree-sitter: add completion intent (syntactic trigger) to analytics events


valerybugakov:
- Use tree-sitter to map AST node types to higher-level cross-language abstractions to classify completion intents. (e.g., function declaration body completion or function call arguments).
- Add completion intent label to analytics events.

Issue #1455, opened 2023-10-23
Tracking issue: tree-sitter based completion intents (syntactic triggers)


valerybugakov:
A tracking issue for the tree-sitter-based completion intent work based on [RFC 848: Tree-sitter-powered intent detection for Cody](https://docs.google.com/document/d/13kYxtJcgoDAY45vulenwGtOW0KnugJPD31MXx-sgx2c/edit#heading=h.rabgao4dn1l):

```[tasklist]
### Completion Intents
- [x] https://github.com/sourcegraph/cody/issues/1451
- [ ] https://github.com/sourcegraph/cody/issues/1452
- [x] https://github.com/sourcegraph/cody/issues/1453
- [ ] https://github.com/sourcegraph/cody/issues/1454
- [ ] https://github.com/sourcegraph/cody/issues/1456
- [ ] https://github.com/sourcegraph/cody/issues/1470
- [x] https://github.com/sourcegraph/cody/issues/1472
```

```[tasklist]
### UPD
- [x] Use completion intent info to track CAR more granularly in Looker. Update the existing queries/logic if needed.
- [x] Pending the disable-streaming-truncation experiment results, consider leveraging intent information to expand the token limit for function/class declaration completion intents.
- [x] Analyze the results of the `syntactic-triggers` experiment where we switched multiline completions to singleline for type/struct/interface declaration intents. If this change positively affects CAR, make this a default behavior.
- [x] Make javascript completion intent queries more precise
```

### After GA Tasks:

```[tasklist]
### After GA
```

valerybugakov:
Closing this tracking issue since the remaining issues are low priority, and we can review them separately. I added them to the eng tracking board.

PR #1457, opened 2023-10-23
Autocomplete: add completion intent to completion events

valerybugakov:
## Context

DISCLAIMER: 60% of the diff are snapshot tests.

- Adds a tree-sitter query for completion intent (syntactic trigger) detection in Javascript/Typescript.
- Integrates the intent detection query into the completion pre-processing pipeline to extract intent before we fire up a completion.
- Adds the completion intent information to the completion analytics events.
- Adds corresponding snapshot tests to every completion intent described in the tree-sitter query.
- Simplifies the interface for running tree-sitter query wrappers.
- Closes https://github.com/sourcegraph/cody/issues/1451
- Part of https://github.com/sourcegraph/cody/issues/1455

## Test plan

Added new unit tests.


vscode/src/completions/get-inline-completions-tests/analytics.test.ts
```diff
@@ -64,6 +81,7 @@ describe('[getInlineCompletions] completion event', () => {
             ],
             "loggedPartialAcceptedLength": 0,
             "params": {
+              "completionIntent": "function.body",
```
valerybugakov, line 84: The completion intent is added to the analytics event.

vscode/src/completions/get-inline-completions-tests/analytics.test.ts
```diff
@@ -76,5 +94,51 @@ describe('[getInlineCompletions] completion event', () => {
...
+                      "parent": "return_statement",
+                    },
+                    "parseErrorCount": 0,
+                    "stopReason": "unit-test",
+                    "truncatedWith": undefined,
+                  },
+                ],
+                "loggedPartialAcceptedLength": 0,
+                "params": {
+                  "completionIntent": "return_statement",
```
valerybugakov, line 129: Also, for singleline completions

vscode/src/completions/tree-sitter/queries.ts
```diff
@@ -2,14 +2,103 @@ import dedent from 'dedent'
...
+ * 1. call_expression: example(
+ * 2. formal_parameters: function example(
+ * 3. import_statement: import react
+ * 4. lexical_declaration: const foo =
+ *
+ * The capture group name ending with "!" means this capture group does not require
+ * a specific cursor position to match.
+ *
+ * TODO: classes, try/catch, members, if/else, loops, etc.
+ * Tracking: https://github.com/sourcegraph/cody/issues/1456
```
valerybugakov, line 49: The query still needs to cover all the possible intents. The plan is to improve it in follow-up PRs.

vscode/src/completions/tree-sitter/queries.ts
```diff
@@ -2,14 +2,103 @@ import dedent from 'dedent'
...
+    'function.body',
+    'type_declaration.name',
+    'type_declaration.body',
+    'block_statement',
+    'import.source',
+    'comment',
+    'parameters',
+    'return_statement',
+    'string',
+] as const
```
beyang, line 30: Will we need to define a separate array for each language? That seems like a substantial undertaking. I wonder if it makes sense to target a few high-level intents that we know will be common. For example:
* Write a function signature following docstring
* Write a function following signature
* Write a class declaration following docstring
* Write a class following class declaration

Which of these do you think would be most impactful (both in terms of the "wow" of the experience and how frequently the experience occurs in day-to-day use)?

valerybugakov: > Will we need to define a separate array for each language?

The idea is to share this array across languages. It will be modified if we don't find matching intents.

> I wonder if it makes sense to target a few high-level intents that we know will be common

Yes, we can do that. Based on [the analytics data](https://console.cloud.google.com/bigquery?sq=839055276916:addb416ff2c146e39f524b526cccd9d8), call expressions, object pairs, and opening/closing elements are the most popular use cases:

Screenshot 2023-10-24 at 09 13 30


> Which of these do you think would be most impactful (both in terms of the "wow" of the experience and how frequently the experience occurs in day-to-day use)?

Based on the last 14 days data, if we merge `grand_parent` and `great_grand_parent` node types:

Screenshot 2023-10-24 at 10 14 03

Completions inside call expressions and argument lists account for ~30% of suggested completions, and CAR varies greatly for them based on the table above (from 15% to 27% based on the node types combination).

Plotting the number of completions for unique, most popular node type combinations shows that call expression and JSX opening elements account for a big chunk of generated completions with an average CAR lower than 20%.

Screenshot 2023-10-24 at 10 19 31

The AST patten "string" > "jsx_attribute" > "jsx_opening_element" has a notably lower CAR of 7.94% compared to the other combinations.

valerybugakov:
Added `arguments` and `argument` intents to the query. Let's add JSX-specific intents in a follow-up PR because it involves adding additional logic. https://github.com/sourcegraph/cody/issues/1472

vscode/src/completions/tree-sitter/query-sdk.ts
```diff
@@ -85,63 +80,145 @@ export function getDocumentQuerySDK(language: string): DocumentQuerySDK | null {
...
+            }
+
+            return [{ node: intentCapture.node, name: intentCapture.name as CompletionIntent }] as const
+        },
+    }
+}
+
+// TODO: check if the block parent is empty in the consumer.
+// Tracking: https://github.com/sourcegraph/cody/issues/1452
+function getIntentFromCaptures(
```
abeatrix, line 151: is it possible for multiple cursor captures to match the cursor position?

abeatrix (approved):
Left a comment in line but I was able to confirm the new intent field is added to the completion event:
image

It'd be nice if this is also added to the trace view :)

vscode/src/completions/tree-sitter/query-sdk.ts
valerybugakov: Yes, it's a matter of how we process this information. For example, here, the selected node is an object and function argument: 

Screenshot 2023-10-25 at 10 52 22

We can return a list of intents and process them differently in call sites. I want to keep things simple for the first iteration — only one intent per cursor position, prioritizing them by their value for context construction.

valerybugakov:
Thanks for the review, @abeatrix! I added the completion intent field to the trace view ✅

vscode/src/completions/tree-sitter/queries.ts
valerybugakov: Adding the `jsx_attribute` completion intent [here](https://github.com/sourcegraph/cody/pull/1490).

PR #1463, opened 2023-10-23
Expose a unstable_handleDidShowCompletionItem for agent

philipp-spiess:
The changes:

- Update to the `provideInlineItems` API so that we have a log Id for the request and individual IDs for every Item: https://github.com/sourcegraph/cody/pull/1463/files#diff-a2dca3f7cbfa2847ddc325842db1723618f70febf6f26d372c7f8adf3d475bafR34-R41
- Update to the `handleDidAcceptCompletionItem` and new unstable_handleDidShowCompletionItem APIs that will work with the new item id.

Todo:

- [x] Make TS happy and test VS code changes (they should work like they did before)
- [x] Change event names to not be VS Code specific
  - Do this in a backward compat way, we will soon switch to the telemetry v2 anyway where this is also solved, so the existing `CodyVSCode` prefixed events are going away anyway
  - I’m thinking that we continue to use the current name for vscode and then add a `CodyAgent` event temporarily. then with the telemtry v2 it'll be the same event using the new terminology: `cody.autocomplete`
- [x] Add a new `agent` log level that is opt-in from the call site

Everything else is now considered private APIs.  If we need another callback we’ll add it but for now we won’t.

**This does not yet change any logging on the Agent side as log level is still set to `off`.**

## Test plan

Made sure the current VS Code logging had no regressions:

```
█ CodyCompletionProvider:resetLatency: Latency Reset
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:suggested VSCode {"properties":{"multiline":false,"triggerKind":"Automatic","providerIdentifier":"fireworks","providerModel":"starcoder-hybrid","languageId":"typescript","type":"inline","multilineMode":null,"id":"4b469f19-6328-4410-a53c-b56417e571fe","contextSummary":{"strategy":"lsp-light","graph":1,"duration":6.172958999872208},"source":"Network","lineCount":1,"charCount":28,"items":[{"lineCount":1,"charCount":28,"stopReason":"streaming-truncation","parseErrorCount":0,"nodeTypes":{"atCursor":".","parent":"ERROR","grandparent":"program"},"nodeTypesWithCompletion":{"atCursor":"program"}}],"otherCompletionProviderEnabled":false,"otherCompletionProviders":[],"latency":892.7772920001298,"displayDuration":2427.4051670003682,"read":true,"accepted":true,"completionsStartedSinceLastSuggestion":3},"opts":{"agent":true}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:accepted VSCode {"properties":{"multiline":false,"triggerKind":"Automatic","providerIdentifier":"fireworks","providerModel":"starcoder-hybrid","languageId":"typescript","type":"inline","multilineMode":null,"id":"4b469f19-6328-4410-a53c-b56417e571fe","contextSummary":{"strategy":"lsp-light","graph":1,"duration":6.172958999872208},"source":"Network","lineCount":1,"charCount":28,"items":[{"lineCount":1,"charCount":28,"stopReason":"streaming-truncation","parseErrorCount":0,"nodeTypes":{"atCursor":".","parent":"ERROR","grandparent":"program"},"nodeTypesWithCompletion":{"atCursor":"program"}}],"otherCompletionProviderEnabled":false,"otherCompletionProviders":[],"acceptedItem":{"lineCount":1,"charCount":28,"stopReason":"streaming-truncation","parseErrorCount":0,"nodeTypes":{"atCursor":".","parent":"ERROR","grandparent":"program"},"nodeTypesWithCompletion":{"atCursor":"program"}}},"opts":{"agent":true}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:completion:persistence:present VSCode {"properties":{"id":"4b469f19-6328-4410-a53c-b56417e571fe","afterSec":30,"difference":0,"lineCount":1,"charCount":28},"opts":{"agent":true}}
```


vscode/src/completions/get-inline-completions.ts
```diff
@@ -10,8 +10,9 @@ import { GetContextOptions, GetContextResult } from './context/context'
 import { GraphContextFetcher } from './context/context-graph'
 import { DocumentHistory } from './context/history'
 import { DocumentContext } from './get-current-doc-context'
+import { AutocompleteItem } from './inline-completion-item-provider'
 import * as CompletionLogger from './logger'
-import { SuggestionID } from './logger'
+import { CompletionLogID } from './logger'
```
valerybugakov, line 15: 👍

vscode/src/completions/logger.ts
```diff
@@ -307,14 +326,27 @@ export function accept(id: SuggestionID, document: vscode.TextDocument, completi
     })
     statistics.logAccepted()
 
+    if (trackedRange === undefined || isRunningInsideAgent()) {
+        return
+    }
```
valerybugakov, line 331: Why do we exit early here?

vscode/src/completions/logger.ts
```diff
@@ -331,26 +363,27 @@ export function partiallyAccept(id: SuggestionID, completion: InlineCompletionIt
     })
 }
 
-export function getCompletionEvent(id: SuggestionID): CompletionEvent | undefined {
-    return activeSuggestions.get(id)
+/** @deprecated */
+export function getCompletionEvent(id: CompletionLogID): CompletionEvent | undefined {
```
valerybugakov, line 367: Do you think we should point to a replacement in the deprecation comment?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -22,18 +24,85 @@ import {
...
+    public logId: CompletionLogID
+
+    /**
+     * The range needed for tracking the completion after inserting. This is needed because the
+     * actual insert range might overlap with content that is already in the document since we set
+     * it to always start with the current line beginning in VS Code.
+     *
+     * TODO: Remove the need for making having this typed as undefined.
+     */
+    public trackedRange: vscode.Range | undefined
```
valerybugakov, line 63: Should we set a completion insert range to the start of the line only for VS Code? I [discussed](https://sourcegraph.slack.com/archives/C05B7C6FBPX/p1698382796915539) it in Slack recently, and it might be that UI jitter (the reason for setting the insert range to the start of the line) is not a problem for other editors.

vscode/src/completions/logger.ts
```diff
@@ -249,8 +263,13 @@ export function suggested(
     }
 }
 
-export function accept(id: SuggestionID, document: vscode.TextDocument, completion: InlineCompletionItem): void {
-    const completionEvent = activeSuggestions.get(id)
+export function accepted(
+    id: CompletionLogID,
+    document: vscode.TextDocument,
+    completion: InlineCompletionItemWithAnalytics,
+    trackedRange: vscode.Range | undefined
```
valerybugakov, line 270: If we only use `completion.range`, then the `trackedRange` argument will be redundant here.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -22,18 +24,85 @@ import {
...
-import { InlineCompletionItem } from './types'
 
 interface AutocompleteResult extends vscode.InlineCompletionList {
+    logId: CompletionLogID
+    items: AutocompleteItem[]
+    /** @deprecated */
     completionEvent?: CompletionEvent
 }
 
+export class AutocompleteItem extends vscode.InlineCompletionItem {
```
valerybugakov, line 41: We discussed streamlining Typescript types for completion items, and I do not propose to do it in this PR, but because we have a good number of `CompletionItem` types, let's maybe leave a comment on how this one should be used?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -447,15 +555,32 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
     }
 
     /**
-     * Called when the user partially accepts a completion. This API is inspired by the the
-     * be named the same, it's prefixed with `unstable_`
```
valerybugakov, line 451: Oh, nice! Earlier today, I was trying to decipher this comment. Now I know what it means!

valerybugakov (commented):
It's the end of my day, so I only did one pass. I'm going to finish the review by testing it locally tomorrow!

vscode/src/completions/inline-completion-item-provider.ts
philipp-spiess: @valerybugakov Yeah we can do that but it doesn't matter that we need a different field here because the returned range for the completion is larger than the range to track for an eventual insertion :/ (hence we need two ranges here which is the ugly part)

valerybugakov: > the returned range for the completion is larger than the range to track for an eventual insertion

Even if we stop making the insert range to start at the beginning of the current line? Could you elaborate on why is that?

valerybugakov (approved):
Tested it locally

vscode/src/completions/inline-completion-item-provider.ts
philipp-spiess: Sorry for the confusion! I meant that for VS Code, the completion range will be larger than the actual range we want to track after a completion is inserted, so in the VS Code case there will still be the need to handle two different ranges :/

vscode/src/completions/logger.ts
valerybugakov: @philipp-spiess, I just bumped into it. Why is this deprecated, and what is recommended to use instead?

philipp-spiess: This function was added to leak some logger internal bookkeeping for other client but since we moved the logging into the shared code now, I don't think we need to leak this anymore (ideally `CompletionEvent` never leaves the boundary of this file). What is your use case for this?

vscode/src/services/telemetry.ts
```diff
@@ -78,15 +84,23 @@ export async function createOrUpdateEventLogger(
...
  */
-function logEvent(eventName: string, properties?: TelemetryEventProperties, opts?: { hasV2Event: boolean }): void {
+function logEvent(
+    eventName: string,
+    properties?: TelemetryEventProperties,
+    opts?: { hasV2Event?: boolean; agent?: boolean }
+): void {
+    if (telemetryLevel === 'agent' && !opts?.agent) {
+        return
+    }
```
valerybugakov, line 98: @PiotrKarczmarz, does JetBrains client use `telemetryLevel === 'agent'`?

pkukielka: Wrong Piotr linked :D Second Piotr is working on Visual Studio now :)
I think it was working fine, as we were sending some events through this API.
Either way v1 is just being deprecated and we really only care about v2 now.
Thanks for pointing it out though!

valerybugakov: > Wrong Piotr linked :D
My apologies! 🙈

PR #1467, opened 2023-10-23
feat: add source tracking for recipe execution

abeatrix:
feat: add source tracking for recipe execution

- Pass source parameter to executeRecipe() to track where recipe was triggered from
- Update callsites to pass source: 'sidebar', 'code-action', 'command' for `chat-question` specifically
- Log source in recipe execution telemetry

## Test plan

Click `Ask Cody to Explain` and check the output channel, you should see the following telemetry event:

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:chat-question:executed {"contextSummary":{"embeddings":4,"local":2},"source":"code-action"}
```

Where `source` of the chat-question is `code-action`

When submitted a question from side bar:

```
logEvent (telemetry disabled): CodyVSCodeExtension:recipe:chat-question:executed {"contextSummary":{"embeddings":13,"local":1},"source":"sidebar"}
```

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -288,7 +288,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         return this.platform.recipes.find(recipe => recipe.id === id)
     }
 
-    public async executeRecipe(recipeId: RecipeID, humanChatInput = ''): Promise<void> {
+    public async executeRecipe(recipeId: RecipeID, humanChatInput = '', source?: string): Promise<void> {
         if (this.isMessageInProgress) {
             this.handleError('Cannot execute multiple recipes. Please wait for the current recipe to finish.')
             return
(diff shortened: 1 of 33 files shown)


valerybugakov (approved):

vscode/src/chat/MessageProvider.ts
```diff
@@ -288,7 +288,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         return this.platform.recipes.find(recipe => recipe.id === id)
     }
 
-    public async executeRecipe(recipeId: RecipeID, humanChatInput = ''): Promise<void> {
+    public async executeRecipe(recipeId: RecipeID, humanChatInput = '', source?: string): Promise<void> {
```
valerybugakov, line 291: Would it be helpful to have a union type for source to ensure we use consistent labels across the board? 

```ts
type RecipeSource = 'sidebar' | 'custom-command'

public async executeRecipe(recipeId: RecipeID, humanChatInput = '', source?: RecipeSource): Promise<void> {}

valerybugakov (commented):
We have [additional](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/sourcegraph/cody%24+executeRecipe%28&patternType=standard&sm=1&groupBy=path) `executeRecipe()` calls in test files. Should we update those too?

vscode/src/chat/MessageProvider.ts
abeatrix: that's a good idea! Just added a new type called ChatEventSource!

PR #1471, opened 2023-10-24
DX: fix eslint errors caused by the config upgrade

valerybugakov:
Fixes the eslint errors caused by the automatic config upgrade.

## Test plan

CI


burmudar (approved):

sqs:
weird that renovate merged it automatically despite failing?

valerybugakov:
@sqs, right, it looks like it's not required

PR #1477, opened 2023-10-24
Autocomplete: Fix suggest item race condition

philipp-spiess:
This PR fixes a race condition that was observed when picking items from the suggest widget. The main reason for this was that the `reuseLastCandidate` logic was not aware of the _actual_ document state but only comparing the doc context `prefix`. This, in some cases, can have the suggest widget candidate injected. 

Now, for that logic, these two completions are exactly the same:

- A completion request for `console.█` with `log` selected in the suggest widget
- A completion request for `console.log█`

This made this logic conclude that it is safe to return the same suggest (`log`) in both cases, causing the repitition.

While working on this, I also noticed another issue with how we calculate the `current-doc`: Right now, this would only work if the selection to be replaced is non-emtpy. This seems to always be the case in production (thankfully) since suggest widget will only trigger once you insert a `.` and it will always replace this dot (so in the above case, it will be `console.█` with `.log` selected, with a range that replaced the `.`). This is not how we wrote our unit tests though and conceptually it shouldn't require this. Hence this PR also adds support for empty ranges in the current doc logic together with a bunch of tests.

This was likely introduced by us starting to propagate the `completeSuggestWidgetSelection` flag properly (which before was always false) but even before that, this race condition could sometimes happen. This implementation should be a bit more robust.

### Before

https://github.com/sourcegraph/cody/assets/458591/df0cf31e-113a-41b2-a58d-9da9709b29a5

### After

https://github.com/sourcegraph/cody/assets/458591/9071bb54-1df9-4cca-9cbd-b8e307768fa4

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -13,6 +13,7 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 ### Fixed
 
 - User selection in active editor will not be replaced by smart selections for the `/edit` command. [pull/1429](https://github.com/sourcegraph/cody/pull/1429)
+- Fixes an issue that caused part of the autocomplete response to be completed when selecting an item from the suggest widget. [pull/](https://github.com/sourcegraph/cody/pull/)
 
 ### Changed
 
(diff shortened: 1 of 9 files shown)


abeatrix (approved):
Tested it locally, and can't seem to reproduce the issue anymore.

### Before

Screenshot 2023-10-24 at 4 46 03 AM

### After 

image

PR #1466, opened 2023-10-23
feat: add default prompt mixin to web

abeatrix:
RE: https://sourcegraph.slack.com/archives/C052G9Y5Y8H/p1697711213702149

The defaultPromptMixin is added and PromptMixin.add is called in activate event for web client.

## Test plan

See if `defaultPromptMixin` was added to every message submitted by users on web.

PR #1479, opened 2023-10-24
refactor: add default prompt mixin to transcripts by default

abeatrix:
Moved from https://github.com/sourcegraph/cody/pull/1466

refactor: add default prompt mixin to transcripts by default 

- Remove initialization of default prompt mixin by clients.
- Update Transcript to add the default prompt mixin to all interactions on creation.
- The PromptMixin class allows mixins to be defined that prepend text to prompts. 
- Transcript mixes in all registered PromptMixin instances into the human prompt text before sending to the assistant.
- This centralizes default prompt handling in Transcript. Clients no longer need to add the default prompt mixin themselves.
- This also adds the default prompt mixin to web and other clients that previously have not initialized the PromptMixin at activation. 

## Test plan

You should see `(Reply as Cody, a coding assistant developed by Sourcegraph...)` prepended to all users' questions:

image


lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -27,13 +33,14 @@ export class PromptMixin {
      * Prepends all mixins to `humanMessage`. Modifies and returns `humanMessage`.
      */
     public static mixInto(humanMessage: InteractionMessage): InteractionMessage {
+        // Default Maxin is added at the end so that it cannot be overriden by a custom mixin.
```
rafax, line 36:
```suggestion
        // Default Mixin is added at the end so that it cannot be overriden by a custom mixin.
```

rafax (approved):
Code LGTM, tried running locally but changes don't seem to be picked up, but I trust you :)

PR #1483, opened 2023-10-24
Add secondary telemetry to ChatViewProvider and main

nathan-downs:
## Test plan

run extension in debug mode

### DIFF
--- a/lib/shared/src/telemetry-v2/index.ts
+++ b/lib/shared/src/telemetry-v2/index.ts
@@ -14,18 +14,49 @@ export type EventFeature =
     // fixup-related events
     | 'cody.fixup'
     | 'cody.fixup.apply'
+    // command-related events
+    | 'cody.command'
+    | 'cody.command.edit'
+    // inline-assistant-related events
+    | 'cody.comment'
+    | 'cody.comment.delete'
+    | 'cody.comment.stop'
+    | 'cody.comment.collapse-all'
+    | 'cody.comment.open-in-sidebar'
+    | 'cody.interactive.clear'
+    | 'cody.history'
+    | 'cody.history.clear'
+    // walkthrough/support-related events
+    | 'cody.walkthrough'
+    | 'cody.walkthrough.showExplain'
+    | 'cody.walkthrough.enableInlineChat'
 
 /**
  * Actions should denote a generic action within the scope of a feature. Where
  * possible, reuse an existing action.
  */
-export type EventAction = 'succeeded' | 'failed' | 'installed' | 'savedLogin' | 'executed' | 'created'
+export type EventAction =
+    | 'succeeded'
+    | 'failed'
+    | 'installed'
+    | 'savedLogin'
+    | 'executed'
+    | 'created'
+    | 'clicked'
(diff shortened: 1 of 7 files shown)


vscode/src/main.ts
```diff
@@ -405,11 +429,25 @@ const register = async (
...
         vscode.commands.registerCommand('cody.walkthrough.enableInlineChat', async () => {
-            telemetryService.log('CodyVSCodeExtension:walkthrough:clicked', { page: 'enableInlineChat' })
+            telemetryService.log(
+                'CodyVSCodeExtension:walkthrough:clicked',
+                { page: 'enableInlineChat' },
+                { hasV2Event: true }
+            )
+            telemetryRecorder.recordEvent('cody.walkthrough.enableInlineChat', 'clicked', {
+                privateMetadata: { page: 'enableInlineChat' },
+            })
```
bobheadxi, line 450: `enableInlineChat` is in the `feature` already (which I think is great!) - maybe we can omit `privateMetadata.page` now? I think similar consideration applies for some of the other events

nathan-downs: sounds good to me!

lib/shared/src/telemetry-v2/index.ts
```diff
@@ -14,19 +14,56 @@ export type EventFeature =
...
  * Where possible, reuse an existing key.
  */
 export type MetadataKey = 'durationMs' | 'lineCount' | 'charCount'
 
+/**
+ * Events accept metadata for ease of categorization in analytics pipelines -
+ * this type enumerates known keys.
+ */
+export type TelemetryEventMetadataInput<MetadataKey> = 'location'
+
```
bobheadxi, line 66: Is this needed? I'm not sure what this does 😅 

```suggestion
```

vscode/src/chat/ChatViewProvider.ts
```diff
@@ -101,6 +110,7 @@ export class ChatViewProvider extends MessageProvider implements vscode.WebviewV
                 break
             case 'event':
                 telemetryService.log(message.eventName, message.properties)
+                telemetryRecorder.recordEvent(message.eventName as EventFeature, 'executed', message.properties)
```
bobheadxi, line 113: Let's remove this for now, since it creates incorrect events (`         'CodyVSCodeExtension:auth:clickOtherSignInOptions/executed' in the tests)

```suggestion
```

bobheadxi: OR - can we change this:

https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/protocol.ts?L17

to also require callers to provide a `message.feature` and `message.action`, update all usages, and use it here?

lib/shared/src/telemetry-v2/index.ts
nathan-downs: Let's see what happens when I remove it! I vaguely remember being asked to add it by one of the e2e tests

bobheadxi (approved):
Awesome, I think this LGTM 😁  thank you!

PR #1484, opened 2023-10-24
eventlogger: accept hasV2Event in properties, add to properties

bobheadxi:
This change makes it so that you can provide `hasV2Event` in `properties`, because it is easy to accidentally put `opts` as the second argument, and also adds `hasV2Event` to the published event log properties.

## Test plan

n/a


nathan-downs (approved):
Thanks for making this easier on me! 😄

PR #1486, opened 2023-10-24
feat: add experimental chat panel

abeatrix:
Close https://github.com/sourcegraph/cody/issues/1522

### New Chat UI

image

- This PR adds a new experimental chat panel UI behind `cody.experimental.chatPanel` config that create separate views for commands, search, chats, and settings. 
- It also adds chat panel-related configuration options and commands.
- Also updated the chat behavior to always start with empty chat, other than that, the chat behavior remains the same as the current one
- Support multiple chat session

### How does it work

The ChatViewManager class manages different chat view provider instances and handles switching between them based on config, and uses currentProvider or sidebarViewProvider to run chat commands.

A chat view provider (ChatViewProvider or ChatPanelProvider) is a class that implements the VS Code WebviewViewProvider interface to display a webview or webview panel for chat.

#### The ChatViewManager maintains:

- sidebarViewProvider - An instance of ChatViewProvider for the sidebar auth flow and default chat view. This is always kept around.
- chatPanelProviders - A map of ChatPanelProvider instances for editor panels, enabled when experimentalChatPanel config is true.
- currentProvider - The currently active provider instance.

It decides which provider to use based on the experimentalChatPanel config:
- When false, it uses the sidebarViewProvider.
- When true, it creates a ChatPanelProvider if none exists yet, and uses that.

To execute commands, it first checks currentProvider, and if not set, defaults to using sidebarViewProvider.

After executing the command, it sets currentProvider to the provider used.

### Next

Here are the items to be completed in future PRs:

- Add @ context
- Add context picker
- Update /test to inline
- Provider dropdown

## Test plan

You can switch between the new chat UI and the current one from Cody Setting under `New Chat UI`:

image

See video below for a quick walkthrough:

https://github.com/sourcegraph/cody/assets/68532117/83625473-1d9f-4b7d-9ba0-46f4a47feb5e

#### Bugs
- [x] Looks like it’s enabled by default for me (have been testing w/ separate instance)
- [x] Chats opening multiple tabs
- [x] Chats not opening - `pending for repro steps`
- [x] Changing active chat should change selected item in treeview
- [x] Delete all chats sometimes just deletes the last one, sometimes deletes all
- [x] Sign out with chat window open doesn't close existing chats
- [x] Sometimes it gets in a state where clicking the buttons in the chats treeview fail with command not found errors
- [x] Creating new chats in succession causes some to fail to load
- [x] This may be existing behaviour, but a blank unsaved editor seems to work but if you add some lines of code then it fails saying no file is open

#### Changes
- [x] Just a single “Settings” option that goes to our settings
- [x] Add “@ext:sourcegraph.cody-ai” to “Keyboard shortcuts” command arg (command: { command: - [ ] 'workbench.action.openGlobalKeybindings', args: ['@ext:sourcegraph.cody-ai’]})
- [x] “Clear Chat” -> “Delete Chat” and “Delete All Chats”
- [x] “Delete All Chats” needs a confirmation
- [x] Clicking “Custom” should run “cody.action.commands.custom.menu”
- [x] Remove timestamp label on chat items, because it gets cut off anyhow when chat messages are long
- [x] Search: hide for now behind a hidden setting for dev testing? - REMOVED
- [x] Add version number in Help sidebar: Let's just add the same between "Keyboard Shortcuts" and "Sign Out", but "Version 0.14.2" (no v) and the icon github and have it link to the GitHub release?
- [ ] We can make “Switch Account” an action button (icon: sign-in… I couldn’t find a better one) on “Sign Out” and remove the “…”
- [ ] New chat should immediately show in tree view

#### TBD
- [ ] “Custom Commands” item too strong for experimental feature I think, and we move to just an action button on panel root. If they have custom commands defined, we can have a collapsed “Custom Commands” item with a badge count that expands to show all the commands?
- [ ] Clicking command with no editor open does nothing. Can we disable treeview items that need a file? Notification would feel too abrupt. Something else?

I'm going to move the following items to a seperate PRs since they affect the current behavior:
- [ ] “Sign Out” needs a confirmation https://github.com/sourcegraph/cody/issues/1520
- [ ] too many experimental features in the settings menu https://github.com/sourcegraph/cody/issues/1521


vscode/src/chat/ChatPanelProvider.ts
```diff
@@ -0,0 +1,459 @@
...
+    public webviewPanel: vscode.WebviewPanel | undefined = undefined
+    public treeView: TreeViewProvider
+
+    constructor({ treeView, extensionUri, ...options }: ChatPanelProviderOptions) {
+        super(options)
+        this.extensionUri = extensionUri
+        this.treeView = treeView
+    }
+
+    private async onDidReceiveMessage(message: WebviewMessage): Promise<void> {
```
valerybugakov, line 46: Does it make sense to share part of the existing logic from `ChatViewProvider` with the new `ChatPanelProvider` to ensure we fix bugs for both implementations while `ChatPanelProvider` is behind the feature flag? Do we need to handle all the same messages as `ChatViewProvider` here?

vscode/src/chat/ChatViewManager.ts
```diff
@@ -0,0 +1,300 @@
...
+     *
+     * Checks currentProvider first.
+     * If not set, defaults to sidebarViewProvider.
+     * Sets currentProvider to the returned provider.
+     */
+    private getChatProvider(): ChatViewProvider | ChatPanelProvider {
+        const provider = this.currentProvider || this.sidebarViewProvider
+        this.currentProvider = provider
+        return provider
+    }
```
valerybugakov, line 93: Could we use `this.currentProvider` instead of this helper? It looks like `this.currentProvider` is always set.

vscode/src/chat/ChatPanelProvider.ts
```diff
@@ -0,0 +1,459 @@
...
+    /**
+     * Set webview view
+     */
+    public async setWebviewView(view: View): Promise<void> {
+        if (view !== 'chat') {
+            return
+        }
+        this.webviewPanel?.reveal(vscode.ViewColumn.Beside, !!view)
+        return Promise.resolve()
+    }
```
valerybugakov, line 348: `async` is not required here anymore:

```suggestion
    public setWebviewView(view: View): void {
        if (view !== 'chat') {
            return
        }
        this.webviewPanel?.reveal(vscode.ViewColumn.Beside, !!view)
    }
```

vscode/src/chat/ChatViewManager.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    constructor({ extensionUri, ...options }: ChatViewProviderOptions) {
+        this.messageProviderOptions = options
+        this.extensionUri = extensionUri
+        this.sidebarViewProvider = new ChatViewProvider({ extensionUri, ...options })
+        this.currentProvider = this.sidebarViewProvider
+
+        // Register Commands
+        this.disposables.push(
+            vscode.commands.registerCommand('cody.chat.panel.new', async () => this.createWebviewPanel()),
```
valerybugakov, line 51: Do we want to register commands related to the experimental chat panel even when it's not enabled?

vscode/src/services/TreeViewProvider.ts
```diff
@@ -0,0 +1,190 @@
...
+
+import { UserLocalHistory } from '@sourcegraph/cody-shared/src/chat/transcript/messages'
+
+import { CODY_DOC_URL } from '../chat/protocol'
+
+type CodyTreeItemType = 'command' | 'support' | 'search' | 'chat'
+export class TreeViewProvider implements vscode.TreeDataProvider<vscode.TreeItem> {
+    private treeNodes: vscode.TreeItem[] = []
+    private _disposables: vscode.Disposable[] = []
+    private _onDidChangeTreeData = new vscode.EventEmitter<vscode.TreeItem | undefined | void>()
```
valerybugakov, line 11: Just noticed that we use two patterns for private field naming in the codebase: prefixed with `_` and without it. Historically, `_` was used because we didn't have `private` modifiers, so having it now should be enough to protect it from external use.

vscode/src/editor/active-editor.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function getActiveEditor(): vscode.TextEditor | undefined {
+    // When there is no active editor, reset lastTrackedTextEditor
+    const activeEditors = vscode.window.visibleTextEditors
+    if (!activeEditors.length) {
+        lastTrackedTextEditor = undefined
+        return undefined
+    }
+
+    // When webview panel is focus, calling activeTextEditor will return undefined.
```
valerybugakov, line 19:
```suggestion
    // When the webview panel is focused, calling activeTextEditor will return undefined.
```

vscode/src/editor/active-editor.ts
```diff
@@ -0,0 +1,36 @@
...
+    // This allows us to get the active editor before the webview panel is focused.
+    const get = (): vscode.TextEditor | undefined => {
+        const activeEditor = vscode.window.activeTextEditor
+        if (activeEditor?.document.uri.scheme === 'file') {
+            lastTrackedTextEditor = activeEditor
+        }
+        return activeEditor || lastTrackedTextEditor
+    }
+
+    return get()
```
valerybugakov, line 29: Will it work the same way?

```suggestion
    const activeEditor = vscode.window.activeTextEditor
    if (activeEditor?.document.uri.scheme === 'file') {
        lastTrackedTextEditor = activeEditor
    }
    
    return activeEditor || lastTrackedTextEditor
```

valerybugakov (commented):
Exciting! Thank you for the video overview. 🙌 
I reviewed the PR, focusing on the implementation details, leaving UI/UX to @toolmantim.

vscode/webviews/Chat.tsx
```diff
@@ -216,7 +216,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
     onKeyDown,
 }) => {
     const inputRef = useRef<HTMLTextAreaElement>(null)
-    const placeholder = "Ask a question or type '/' for commands"
+    const placeholder = "Ask a question or type '/reset' to start a new chat"
```
toolmantim, line 219: Does this change it for existing chat users?

toolmantim (commented):
Woohoo! Just gave it a test, and left you some testing notes in Slack.

abeatrix:
### To do items from @toolmantim 

#### Bugs
- [x] Looks like it’s enabled by default for me (have been testing w/ separate instance)
- [x] Chats opening multiple tabs
- [x] Chats not opening - `pending for repro steps`
- [x] Changing active chat should change selected item in treeview
- [x] Delete all chats sometimes just deletes the last one, sometimes deletes all

#### Changes
- [x] Just a single “Settings” option that goes to our settings
- [x] Add “@ext:sourcegraph.cody-ai” to “Keyboard shortcuts” command arg (command: { command: - [ ] 'workbench.action.openGlobalKeybindings', args: ['@ext:sourcegraph.cody-ai’]})
- [x] “Clear Chat” -> “Delete Chat” and “Delete All Chats”
- [x] “Delete All Chats” needs a confirmation
- [x] Clicking “Custom” should run “cody.action.commands.custom.menu”
- [x] Remove timestamp label on chat items, because it gets cut off anyhow when chat messages are long
- [x] Search: hide for now behind a hidden setting for dev testing? - REMOVED

#### TBD
- [ ] Clicking command with no editor open does nothing. Can we disable treeview items that need a file? Notification would feel too abrupt. Something else?
- [ ] “Custom Commands” item too strong for experimental feature I think, and we move to just an action button on panel root. If they have custom commands defined, we can have a collapsed “Custom Commands” item with a badge count that expands to show all the commands?
- [ ] We can make “Switch Account” an action button (icon: sign-in… I couldn’t find a better one) on “Sign Out” and remove the “…”
- [ ] New chat should immediately show in tree view

I'm going to move the following items to a seperate PRs since they affect the current behavior:
- [ ] “Sign Out” needs a confirmation

vscode/webviews/Chat.tsx
toolmantim: I just gave this a test — I think we should revert this and leave it as-is for now

toolmantim (commented):
It's starting to feel much more solid now! The pre-existing bugs which you asked for repo-steps for seem to be fixed 🙌🏼

Left a few small suggestions and found a few more bugs in testing:

1. Sign out with chat window open doesn't close existing chats, which has some odd side-effects:

https://github.com/sourcegraph/cody/assets/153/5ad29537-8c4d-4e7a-9710-20ad5a160b04

https://github.com/sourcegraph/cody/assets/153/a5cbb75f-5dfb-417b-8ca0-45104bd0b273

2. Sometimes it gets in a state where clicking the buttons in the chats treeview fail with command not found errors:

https://github.com/sourcegraph/cody/assets/153/9805b835-adea-413d-96bb-ca62e2c4936d

3. Creating new chats in succession causes some to fail to load:

https://github.com/sourcegraph/cody/assets/153/a50f2c20-8e9d-495b-9584-a023eebd16a9

4. This may be existing behaviour, but a blank unsaved editor seems to work but if you add some lines of code then it fails saying no file is open:

https://github.com/sourcegraph/cody/assets/153/0ddf55da-b00c-443d-929f-2533ab008ad1

abeatrix:
@toolmantim Thank you so much for reporting all the bugs you found! I think I fixed all the bugs you last reported 🤞 Can you give the latest commit a try and let me know if you find other bugs before I move onto implementing the items on the rest of the list (see PR description under WIP section)? 🙇‍♀️

toolmantim (approved):
Looks great! I just pushed some tweaks after playing with it for real.

PR #1487, opened 2023-10-25
benchmark: add more Go test cases

slimsag:
For how I am thinking about these tests in general, and how I am coming up with ideas for user intents, see [this comment](https://github.com/sourcegraph/cody/pull/1446#issuecomment-1778049340)

## Tests added in this PR

- [ ] Wrap error
- [ ] Implement an interface
- [ ] Use context
- [ ] Parallel spawn of goroutines
- [ ] Merge two maps
- [ ] Merge two structs
- [ ] HTTP write json
- [ ] HTTP read SON
- [ ] HTTP check status code
- [ ] HTTP write error
- [ ] HTTP server
- [ ] HTTP post request
- [ ] HTTP get request
- [ ] HTTP get request write file
- [ ] Exec command and get output
- [ ] Exec command error get stdout
- [ ] template.New().ParseFiles()
- [ ] template.New().Funcs().ParseFiles()
    - [ ] https://www.digitalocean.com/community/tutorials/how-to-use-templates-in-go
- [x] Write test function
- [x] Cast number to float
- [x] Initialize a map
- [x] Generate JSON
- [x] Clear map
- [x] Get map keys
- [x] Iterate map
- [x] Struct to JSON
- [x] Struct to map
- [x] Check if struct is empty
- [x] Date string
- [x] Measure time
- [x] Make channel
- [x] Make buffered channel
- [x] Channel read
- [x] Channel read switch
- [x] Channel read select
- [x] Channel read closed
- [x] Channel send
- [x] Channel send non-blocking
- [x] Unwrap error

## Test plan

Adding tests to benchmark test suite

### DIFF
--- a/vscode/test/benchmark/datasets/golang/get-map-keys/main.go
+++ b/vscode/test/benchmark/datasets/golang/get-map-keys/main.go
@@ -1,0 +1,12 @@
+package main
+
+import (
+	"fmt"
+)
+
+func mapKeys█
+
+func main() {
+	resp := work()
+    fmt.Println(mapKeys(resp))
+}
--- a/vscode/test/benchmark/datasets/golang/basic-test-func/config.json
+++ b/vscode/test/benchmark/datasets/golang/basic-test-func/config.json
@@ -1,0 +1,7 @@
+{
+  "entryFile": "main.go",
+  "openFiles": ["context.go"],
+  "closedFiles": [],
+  "solutionFile": "solution.go",
+  "testCommand": "go build main.go context.go"
+}
--- a/vscode/test/benchmark/datasets/golang/basic-test-func/context.go
+++ b/vscode/test/benchmark/datasets/golang/basic-test-func/context.go
@@ -1,0 +1,22 @@
+package main
+
+import "fmt"
+
+type Response struct {
+	Value string
+	Code  int
+	Error error
+	Debug struct {
+		ID string
+	}
+}
+
+type ResponseID uint64
+
+func generateResponse() ResponseID {
+	return 200
+}
+
+func log(f float64) {
+	fmt.Println(f)
+}
--- a/vscode/test/benchmark/datasets/golang/basic-test-func/main.go
+++ b/vscode/test/benchmark/datasets/golang/basic-test-func/main.go
@@ -1,0 +1,9 @@
+package main
+
+func TestGenerateResponse(t *testing.T) {
+    █
+}
+
+func main() {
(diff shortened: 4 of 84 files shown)


slimsag:
@taras-yemets I'm also keen to get your feedback on mine, I still have quite a few more I want to add this week but keen to hear your feedback.

Currently this dataset is eerily close to our CAR%:

```
sourcegraph.cody-ai:
Pass: 4
Fail: 12
No Completion: 1
Average Time to Completion*: 1514ms
Median Time to Completion*: 1500ms
Pass Rate: 23.53%
Adjusted Pass Rate**: 25.00%
```

taras-yemets:
@slimsag, my initial thought was to play around with the completion request context (make it more complex): 
- more imports to see if Cody uses them for the current task ([example](https://github.com/sourcegraph/cody/blob/7227a80b8e7f8a3a350f3fd26b4e0e96884fb9b8/vscode/test/benchmark/datasets/typescript/utils-open-file/solution.ts))
- importing 2+ functions with similar names but returning different interfaces to check if Cody uses the correct one ([example](https://github.com/sourcegraph/cody/blob/7227a80b8e7f8a3a350f3fd26b4e0e96884fb9b8/vscode/test/benchmark/datasets/typescript/promise-open-file/generate.ts)) 
- import a third-party package and check if Cody uses it instead of implementing a similar functionality on its own, etc.

But, as Cody sometimes fails to complete code not requiring extensive context (except for the imported file, of course), it makes sense to continue testing one-step scenarios (for example, like you do for channels). 
I agree with your suggestion https://github.com/sourcegraph/cody/pull/1446#issuecomment-1778049340 and think that we should probably start with fewer steps test cases and then consider complicating them. Moreover, the "more complex" examples in the very end just import some functions and use them. And from what I experienced, whether the completion passes the test often depends on the file exporting this function being open, but not on the number of imports, number of the code paths in each function and other "complexity" indicators.

vscode/test/benchmark/datasets/golang/get-map-keys/main.go
```diff
@@ -0,0 +1,12 @@
+package main
+
+import (
+	"fmt"
+)
+
+func mapKeys█
```
taras-yemets, line 7: I often get single-line completions in similar situations. When in a test case, the function body is not complete resulting in a compilation error. Does it work differently in Go test cases?

slimsag:
Moving to new repo https://github.com/sourcegraph/coder-eval

PR #1488, opened 2023-10-25
Add secondary telemetry to MessageProvider

nathan-downs:
## Test plan

run extension in debug mode


lib/shared/src/telemetry-v2/index.ts
```diff
@@ -51,12 +82,19 @@ export type EventAction =
...
  * Where possible, reuse an existing key.
  */
-export type MetadataKey = 'durationMs' | 'lineCount' | 'charCount'
+export type MetadataKey = 'durationMs' | 'lineCount' | 'charCount' | 'codeBlocks'
+
+/**
+ * Events accept metadata for ease of categorization in analytics pipelines -
+ * this type enumerates known keys.
+ */
+export type TelemetryEventMetadataInput<MetadataKey> = 'location'
```
bobheadxi, line 97: I think you removed this in the other PR 😋

bobheadxi:
```suggestion
```

vscode/test/e2e/inline-assist.test.ts
```diff
@@ -54,5 +54,12 @@ test('start a fixup job from inline chat with valid auth', async ({ page, sideba
     await page.getByRole('button', { name: 'Apply Edits' }).click()
     await expect(page.getByText('<title>Goodbye Cody</title>')).toBeVisible()
     await expect.poll(() => loggedEvents).toEqual(expectedOrderedEvents)
-    await expect.poll(() => loggedV2Events).toEqual(['cody.command.edit/executed', 'cody.fixup.apply/succeeded'])
+    await expect
+        .poll(() => loggedV2Events)
+        .toEqual([
+            'CodyVSCodeExtension:auth:clickOtherSignInOptions/executed',
```
bobheadxi, line 60: I think this is unexpected (name in the old format) - maybe something we removed in the previous PR?

vscode/src/chat/MessageProvider.ts
```diff
@@ -375,7 +391,12 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 })
             }
         }
-        telemetryService.log(`CodyVSCodeExtension:recipe:${recipe.id}:executed`, { contextSummary, source })
+        telemetryService.log(
+            `CodyVSCodeExtension:recipe:${recipe.id}:executed`,
+            { contextSummary },
+            { hasV2Event: true }
+        )
+        telemetryRecorder.recordEvent(`cody.recipe.${recipe.id}`, 'executed', { privateMetadata: { contextSummary } })
```
bobheadxi, line 399: love how typescript can figures out string types like this, even with composition 😁

vscode/src/chat/MessageProvider.ts
```diff
@@ -449,9 +470,16 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
...
+            telemetryService.log(
+                'CodyVSCodeExtension:guardrails:annotate',
+                {
+                    codeBlocks: result.codeBlocks,
+                    duration: result.duration,
+                },
+                { hasV2Event: true }
+            )
+            telemetryRecorder.recordEvent('cody.guardrails.annotate', 'executed', {
+                metadata: { codeBlocks: result.codeBlocks, durationMs: result.duration },
```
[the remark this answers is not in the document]
nathan-downs: Good call out, I'm actually not sure, it looks like it might be in nanoseconds: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/guardrails/attribution/observability.go?L64
All of the references I'm seeing to guardrails logging duration calls `time.Since(start)` which looks like it returns nanoseconds.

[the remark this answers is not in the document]
bobheadxi: We might want to change the metadata key, or convert it to ms (which seems more usable than ns? i don't imagine we care much about sub-nanosecond differences)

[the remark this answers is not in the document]
nathan-downs: Agreed, just pushed a change to convert to ms

bobheadxi (approved):
nice, thank you!

minor comment is that we have `source` in a lot of metadata - seems like some of these are nicely condensed into event feature, but we'll need to be mindful that currently without anything in the allowlist we'll lose some granularity here

nathan-downs:
Moved `source` in metadata into `Feature` and moved `contextSummary` into metadata from privateMetadata

PR #1507, opened 2023-10-26
Autocomplete: do adjust a completion range if it does not match the current line suffix

valerybugakov:
## Context

- Fixes the issue where completions were needlessly hidden: `myFunction(█)` with the insert text being `"one", false` because the current line suffix `)` was not present in the generated completion. However, we still adjusted the completion insertion range to include it. 
- Fixes the issue where all the generated completions were hidden, even if only one needed to be ignored. Now, we filter out invisible completions individually and keep visible completions if we have any.

## Test plan

Added unit tests.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -354,7 +354,6 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
 
             // A completion that won't be visible in VS Code will not be returned and not be logged.
             if (visibleItems.length === 0) {
-                console.log('Invisible completions!')
                 // Returning null will clear any existing suggestions, thus we need to reset the
                 // last candidate.
                 this.lastCandidate = undefined
@@ -507,9 +506,9 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             // come in.
             const start = currentLine.range.start
 
-            // The completion will always exclude the same line suffix, so it has to overwrite the
-            // current same line suffix and reach to the end of the line.
-            const end = currentLine.range.end
+            // If the completion does not have a range set it will always exclude the same line suffix,
+            // so it has to overwrite the current same line suffix and reach to the end of the line.
+            const end = (completion.range?.end || currentLine.range.end) as vscode.Position
 
             const vscodeInsertRange = new vscode.Range(start, end)
             const trackedRange = new vscode.Range(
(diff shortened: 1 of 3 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -369,18 +373,22 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                     lastTriggerSelectedCompletionInfo: context?.selectedCompletionInfo,
                     result,
                 }
-                this.lastCandidate = items.length > 0 ? candidate : undefined
+                this.lastCandidate = visibleItems.length > 0 ? candidate : undefined
```
abeatrix, line 376: is it possible for `visibleItems.length > 0` if it's inside `visibleItems.length === 0`?

abeatrix (commented):
> Fixes the issue where completions were needlessly hidden: myFunction(█) with the insert text being "one", false because the current line suffix ) was not present in the generated completion. However, we still adjusted the completion insertion range to include it.

I found cases where completion is still not showing up because of current line suffix 🤔 

image

image

Is this something else that's unrelated to the current issue?

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: It's not. GitHub diff makes it look like it, but expanding the code above shows another if statement in between.

abeatrix: WHOOOPS 🤡

valerybugakov:
@abeatrix, great catch! I debugged your examples locally and found out that we change completion insert ranges in two places: 

1. https://github.com/sourcegraph/cody/blob/b1f78ac395f50387cf089096e73c3969cba0818a/vscode/src/completions/text-processing/process-inline-completions.ts#L67
2. https://github.com/sourcegraph/cody/blob/b1f78ac395f50387cf089096e73c3969cba0818a/vscode/src/completions/inline-completion-item-provider.ts#L506

The second part doesn't care if the completion range already exists, so the logic in the post-processing call (1) does not affect the result. 🙁 

I feel that this is one of these cases we bumped into frequently in the last month: we have multiple places implementing parts of the same logic, and they are not explicitly connected, which causes customer-facing bugs because it's impossible to keep all these pieces in our heads. The codebase snowballs. We should spend time streamlining the completion processing logic to clarify where specific changes are expected on the codebase architecture level (e.g., having one place where we operate on completion insert ranges or having one place for multiline truncation). Setting clear boundaries based on the current implementation would help us move faster with less time spent debugging such cases. WDYT? 

cc @philipp-spiess @beyang

abeatrix (approved):
What you said makes a lot of sense to me but I'll defer to you and Philipp since you both know the code better than I do 😃 it sounds like my examples require additional follow-up work but shouldn't block what you fixed here?

valerybugakov:
> it sounds like my examples require additional follow-up work but shouldn't block what you fixed here?

Yeah, I'm working on it now and will add them to the diff in this PR.

valerybugakov:
@abeatrix, I pushed the update: https://github.com/sourcegraph/cody/pull/1507/commits/3f646359b0ad89318650e94a5fb2eedb4608d33a to account for cases you shared above. Could you take another look, please?

abeatrix:
@valerybugakov sorry I spent too much time on the chat redesign pr and forgot to review this. Will do it first thing on Monday 🙇🏻‍♀️

PR #1508, opened 2023-10-26
fix: smart selection not working on the first line of code

abeatrix:
fix: smart selection not working on the first line of code

This issue was caused by `!selection?.start.line` in the smart selection function excluding the possibility of `selection?.start.line === 0` (the vs code API returns the line number on display - 1, so line 1 in the editor would returns 0). 

This commit fixes that by explicitly checking selection.start.line is not null and is not equal to 0 before failing.

## Test plan

Try running the `Generate Unit Tests` command on the first line of the file:

image

#### After

You should be able to run the command and the expanded range will be selected for you automatically when available:

image

#### Before

image

### DIFF
--- a/vscode/src/editor/vscode-editor.ts
+++ b/vscode/src/editor/vscode-editor.ts
@@ -138,7 +138,7 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
             return null
         }
         const selection = activeEditor.selection
-        if (selection?.start.line === undefined) {
+        if (!selection.start) {
             return null
         }
 
--- a/vscode/src/editor/vscode-editor.ts
+++ b/vscode/src/editor/vscode-editor.ts
@@ -138,7 +138,7 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
             return null
         }
         const selection = activeEditor.selection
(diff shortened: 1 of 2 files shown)


vscode/src/editor/vscode-editor.ts
```diff
@@ -139,7 +138,7 @@ export class VSCodeEditor implements Editor<InlineController, FixupController, C
             return null
         }
         const selection = activeEditor.selection
-        if (!selection?.start.line) {
+        if (!selection?.start.line && selection?.start.line !== 0) {
```
dominiccooney, line 141: Two negations are confusing, so you could have `!(... || ...)`. But in this case you can be direct by just thinking about the types: It is undefined, or a number. So check if that is the case.

I can't remember the specific context but I've seen this mistake before. Something to add to your test cases and personal desk check, whether it handles working on line 0.
 
```suggestion
        if (selection?.start.line === undefined) {
```

dominiccooney (approved):
Looks good but please simplify, suggestion inline.

vscode/src/editor/vscode-editor.ts
abeatrix: Oh yea I could just use `!selection.start` 😆  Thank you for your reviews as always 🙇‍♀️

PR #1511, opened 2023-10-26
add new aliases for x86_64 architecture

abeatrix:
A new alias x64 was added for the x86_64 architecture in the getOSArch function.

For windows with x64 returned by`os.arch()` currently returns undefined as it's missing on the list.

image

On Windows, os.arch() can return 'x64', 'ia32', or 'x86' to indicate 64-bit or 32-bit architectures.

Without the 'x64': 'x86_64' mapping, if os.arch() returns 'x64' on Windows, the arch value would be undefined in the return value of getOSArch.

## Test plan

This is also what we used to create the URLs for downloading Cody App

### DIFF
--- a/vscode/src/os.ts
+++ b/vscode/src/os.ts
@@ -13,6 +13,7 @@ export function getOSArch(): {
         arm64: 'aarch64',
         aarch64: 'aarch64',
         x86_64: 'x86_64',
+        x64: 'x86_64',
         i386: 'x86',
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
I'm still dealing with C++ installation quirks so I haven't been able to test this locally, but let's merge for now. The diff is safe and I will confirm the fix post-merge.

PR #1513, opened 2023-10-26
BFG: fix download URL for x64 CPU architecture

olafurpg:
Follow-up to #1511

## Test plan
Manually tested it locally on Windows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist/
 *.tsbuildinfo
 .DS_Store
 .env
+.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -5,3 +5,4 @@ dist/
 *.tsbuildinfo
 .DS_Store
 .env
+.idea/
```
olafurpg, line 8: I tried ignoring this in my $HOME/.gitignore but it wasn't working for some reason. Didn't want to spend more time investigating why.

SuperAuguste (approved):

abeatrix (approved):

PR #1517, opened 2023-10-26
Add secondary telemetry to AuthProvider

nathan-downs:
## Test plan

run extension in debug mode


dominiccooney:
Same feedback I wrote on https://github.com/sourcegraph/cody/pull/1519 applies here:

- Bit of background about the motivation for the change.
- Detail in the test plan so someone who's not familiar can reproduce it relatively easily.
- Ideally, an automated test.

vscode/src/services/AuthProvider.ts
```diff
@@ -64,13 +65,24 @@ export class AuthProvider {
...
         }
         const menuID = type || item?.id
-        telemetryService.log('CodyVSCodeExtension:auth:selectSigninMenu', { menuID })
+        telemetryService.log('CodyVSCodeExtension:auth:selectSigninMenu', { menuID }, { hasV2Event: true })
+        if (type) {
+            telemetryRecorder.recordEvent(`cody.auth.selectSigninMenu.${type ?? ''}`, 'clicked', {
+                privateMetadata: { menuID },
+            })
+        } else {
+            telemetryRecorder.recordEvent(`cody.auth.selectSigninMenu${type ?? ''}`, 'clicked', {
```
dominiccooney, line 81: You can simplify this, because you know type is falsy here.

vscode/src/services/AuthProvider.ts
```diff
@@ -64,13 +65,24 @@ export class AuthProvider {
...
+        telemetryService.log('CodyVSCodeExtension:login:clicked', undefined, { hasV2Event: true })
+        telemetryRecorder.recordEvent('cody.auth.login', 'clicked')
         const item = await AuthMenu(mode, this.endpointHistory)
         if (!item) {
             return
         }
         const menuID = type || item?.id
-        telemetryService.log('CodyVSCodeExtension:auth:selectSigninMenu', { menuID })
+        telemetryService.log('CodyVSCodeExtension:auth:selectSigninMenu', { menuID }, { hasV2Event: true })
+        if (type) {
```
dominiccooney, line 76: This duplicates a lot of code just for the purpose of inserting a `.`. Float the condition down so it is obvious what's different about these two branches.

vscode/src/services/AuthProvider.ts
```diff
@@ -120,9 +132,18 @@ export class AuthProvider {
...
+            {
+                success: Boolean(authState?.isLoggedIn),
+            },
+            { hasV2Event: true }
+        )
+        if (authState?.isLoggedIn) {
+            telemetryRecorder.recordEvent('cody.auth.fromToken', 'succeeded')
+        } else {
+            telemetryRecorder.recordEvent('cody.auth.fromToken', 'failed')
+        }
```
dominiccooney, line 146: Eliminate the redundancy here and highlight what's going on by putting the condition close to the effect it has:

```suggestion
        telemetryRecorder.recordEvent('cody.auth.fromToken', authState?.isLoggedIn ? 'succeeded' : 'failed')
```

dominiccooney (changes requested):
Let's introduce these changes as simply as possible.

vscode/src/services/AuthProvider.ts
nathan-downs: Implemented, thanks!

vscode/src/services/AuthProvider.ts
nathan-downs: Reworked the `feature` building logic so that it still complies with the enumerated EventFeature type

```
let eventName: EventFeature
eventName = 'cody.auth.selectSigninMenu'
if (type) {
    eventName = `${eventName}.${type}`
}
```

dominiccooney (approved):
Looks good.

I was thinking, instead of `{ hasV2Event: true }` you could do this logging in a closure and associate the V2 event that way. But it is up to you... if you trust people won't start copy-pasting `{ hasV2Event: true }` everywhere, this is fine. I am guessing all these old events go away very soon?

bobheadxi:
Thanks @dominiccooney! I like your idea, but because we used a shared interface over at https://github.com/sourcegraph/telemetry (in hopes of unifying how we interact with events in our various clients, instead of having bespoke contracts in each place), it does make this a bit trickier to add.

I think the `hasV2Event` copy-pasta risk is acceptable, as for _new_ events going forward, we should not be adding legacy events, and as you mention we will hopefully get to a point where it can be removed soon. I also think having the callsites next to each other and a `hasV2Event` is sufficient for us to find all the usages later and get rid of them, and we can toggle a no-op switch within the legacy logger directly as well.

WDYT @nathan-downs ?

nathan-downs:
replaced by https://github.com/sourcegraph/cody/pull/2060

PR #1518, opened 2023-10-26
refactor: simplify auth state management

abeatrix:
refactor: simplify auth state management on start-up

Syncing auth status after a change is now handled directly in AuthProvider.

Came across this when I was working on the new chat UI -- it looks like we have both `cody.auth.sync` and `authProvider.addChangeListener` to handle auth change events. This PR proposes addressing auth state changes in one central auth listener to remove the need to have multiple places to auth update events. 

- Remove redundant auth sync command registration `cody.auth.sync`
- Move auth state update logic into dedicated auth change listener
- Consolidate status bar indicator, completions setup, and other auth-dependent services into auth change handler
- Remove redundant auth sync logic
- Pass auth status object on auth change to status bar indicator instead of re-checking
- Fetch access token before setting in symfRunner to avoid null errors
- fix: remove redundant auth sync command registration

## Test plan

All the current tests should pass, and everything should work as before.

Try logging-in and then logout, and then log back in, each time try auto-completion and chat. Everything should work as before.

### DIFF
--- a/vscode/src/chat/ContextProvider.ts
+++ b/vscode/src/chat/ContextProvider.ts
@@ -25,7 +25,7 @@ import { telemetryRecorder } from '../services/telemetry-v2'
 
 import { ChatViewProviderWebview } from './ChatViewProvider'
 import { GraphContextProvider } from './GraphContextProvider'
-import { ConfigurationSubsetForWebview, LocalEnv } from './protocol'
+import { AuthStatus, ConfigurationSubsetForWebview, LocalEnv } from './protocol'
 
 export type Config = Pick<
     ConfigurationWithAccessToken,
@@ -146,8 +146,10 @@ export class ContextProvider implements vscode.Disposable {
      * Save, verify, and sync authStatus between extension host and webview
      * activate extension when user has valid login
      */
-    public async syncAuthStatus(): Promise<void> {
-        const authStatus = this.authProvider.getAuthStatus()
+    public async syncAuthStatus(authStatus?: AuthStatus): Promise<void> {
+        if (!authStatus) {
+            authStatus = this.authProvider.getAuthStatus()
+        }
         // Update config to the latest one and fire configure change event to update external services
         const newConfig = await getFullConfig()
         if (authStatus.siteVersion) {
@@ -168,7 +170,7 @@ export class ContextProvider implements vscode.Disposable {
         await this.publishConfig()
         this.onConfigurationChange(newConfig)
         // When logged out, user's endpoint will be set to null
-        const isLoggedOut = !authStatus.isLoggedIn && !authStatus.endpoint
+        const isLoggedOut = !authStatus.isLoggedIn && !authStatus.endpoint && !authStatus.showInvalidAccessTokenError
         const isAppEvent = isLocalApp(authStatus.endpoint || '') ? '.app' : ''
         const eventValue = isLoggedOut ? 'disconnected' : authStatus.isLoggedIn ? 'connected' : 'failed'
         // e.g. auth:app:connected, auth:app:disconnected, auth:failed
--- a/vscode/src/chat/ContextProvider.ts
+++ b/vscode/src/chat/ContextProvider.ts
@@ -98,16 +98,12 @@ export class ContextProvider implements vscode.Disposable {
     public onConfigurationChange(newConfig: Config): void {
         logDebug('ContextProvider:onConfigurationChange', 'using codebase', newConfig.codebase)
         this.config = newConfig
-        const authStatus = this.authProvider.getAuthStatus()
-        if (authStatus.endpoint) {
-            this.config.serverEndpoint = authStatus.endpoint
-        }
         this.configurationChangeEvent.fire()
     }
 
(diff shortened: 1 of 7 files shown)


vscode/src/main.ts
```diff
@@ -330,13 +321,6 @@ const register = async (
         vscode.commands.registerCommand('cody.auth.signin', () => authProvider.signinMenu()),
         vscode.commands.registerCommand('cody.auth.signout', () => authProvider.signoutMenu()),
         vscode.commands.registerCommand('cody.auth.support', () => showFeedbackSupportQuickPick()),
-        vscode.commands.registerCommand('cody.auth.sync', () => {
-            const result = contextProvider.syncAuthStatus()
-            void featureFlagProvider.syncAuthStatus()
-            // Important that we return a promise here to allow `AuthProvider`
```
abeatrix, line 336: When `'cody.auth.sync'` is called, the AuthStatus has already been updated and stored by AuthProvider, so I don't think it would affect auth config changes, @chwarwick can you confirm if this comment is still true?

vscode/src/main.ts
```diff
@@ -521,10 +506,30 @@ const register = async (
...
+                    .then(token => symfRunner.setSourcegraphAuth(authStatus.endpoint, token))
+                    .catch(error => console.error(error))
+            } else {
+                symfRunner.setSourcegraphAuth(null, null)
+            }
+        }
     })
+
     await setupAutocomplete()
+    updateAuthStatusBarIndicator(authProvider.getAuthStatus())
```
dominiccooney, line 532: Would it be possible to set up the callback, and then call it once? Because you have two duplicated lines here... and maybe contextProvider and featureFlag provider also need the initial auth status?

vscode/src/services/AuthProvider.ts
```diff
@@ -263,22 +263,21 @@ export class AuthProvider {
     }
 
     // Set auth status and share it with chatview
-    private async syncAuthStatus(authStatus: AuthStatus): Promise<void> {
+    private syncAuthStatus(authStatus: AuthStatus): void {
```
dominiccooney, line 266: Let's rename this `setAuthStatus`, `set authStatus()` or `updateAuthStatus`, something like that. It doesn't do any syncing now.

vscode/src/chat/ContextProvider.ts
```diff
@@ -146,8 +146,10 @@ export class ContextProvider implements vscode.Disposable {
      * Save, verify, and sync authStatus between extension host and webview
      * activate extension when user has valid login
      */
-    public async syncAuthStatus(): Promise<void> {
-        const authStatus = this.authProvider.getAuthStatus()
+    public async syncAuthStatus(authStatus?: AuthStatus): Promise<void> {
```
dominiccooney, line 149: Today we have a bunch of methods called "syncAuthStatus" all doing different things. Let's call this `onAuthStatusChange`. It consumes updates from the auth provider, I think that name makes it clear it consumes changes, it doesn't produce them.

vscode/src/chat/ContextProvider.ts
```diff
@@ -146,8 +146,10 @@ export class ContextProvider implements vscode.Disposable {
...
+        if (!authStatus) {
```
dominiccooney, line 150: Why do we have *this*? Seems dodgy, if we're being passed the authstatus, to turn around and grab it again.

vscode/src/chat/ContextProvider.ts
```diff
@@ -168,7 +170,7 @@ export class ContextProvider implements vscode.Disposable {
         await this.publishConfig()
         this.onConfigurationChange(newConfig)
         // When logged out, user's endpoint will be set to null
-        const isLoggedOut = !authStatus.isLoggedIn && !authStatus.endpoint
+        const isLoggedOut = !authStatus.isLoggedIn && !authStatus.endpoint && !authStatus.showInvalidAccessTokenError
```
dominiccooney, line 173: Clean up this to be `!(a || b || c)`; this is hard to read.

Where did `showInvalidAccessTokenError` come from? It seems weird to have `isLoggedOut` be anything more than `!authStatus.isLoggedIn`. Does `isLoggedOut` need a more accurate name?

dominiccooney (changes requested):
Thanks for working on this, this area will certainly benefit from simplification. Some feedback inline.

vscode/src/main.ts
chwarwick: This was related to an issue that occurred when initializing the agent prior to having a credentials setup and then later authenticating.  There were unit tests added to the agent for this case so if they continue to pass it shouldn't be a problem.

abeatrix: Awesome, thanks for the context 🙇‍♀️

vscode/src/chat/ContextProvider.ts
abeatrix: Yea that makes sense. Updated syncAuthStatus to onAuthStatusChange for ContextProvider and FeatureFlagProvider

PR #1532, opened 2023-10-27
Custom Commands: Fix custom command menu not showing for a single custom command

toolmantim:
The Custom Commands action shows the incorrect menu (the generic custom commands config quickpick) if you only have a single custom command defined, due to an off-by-one. This fixes it so it shows the correct menu.

## Test plan

- Created a single custom command. Tested the action.


abeatrix (approved):
🚀

PR #1535, opened 2023-10-27
Improve e2e completion tests to have better mock responses

DanTup:
Previously the mock completion handler would just always serve up some results. This caused a lot of flakiness in tests because immediately after accepting a completion another completion might show up (resulting in intermittent extra "suggest" events logged and extra ghost text in the editor).

This change extracts the code being completed and only returns results where the typed prefix matches a defined mock response. This prevents tests getting any unwanted completions.

Mostly these flakes showed up on Windows GH bots (which I'm trying to enable), the other bots seemed pretty consistent.

## Test plan

These are changes to the e2e tests that run on the bots with `pnpm test:e2e`.


vscode/src/completions/client.ts
```diff
@@ -67,6 +67,7 @@ export function createClient(config: CompletionsClientConfig, logger?: Completio
         // Force HTTP connection reuse to reduce latency.
         // c.f. https://github.com/microsoft/vscode/issues/173861
         headers.set('Connection', 'keep-alive')
+        headers.set('Content-Type', 'application/json; charset=utf-8')
```
DanTup, line 70: NOTE: Unlike the rest of this PR, this change affects production code, not just tests.

Without it, `request.body` is empty in the mock handler, so we could not parse out the completion code. Since adding it, I did test with a real endpoint (by debugging the extension from VS Code, opening a project and confirming I get real production completion responses).

However, I'm not sure whether there are multiple endpoints that this code might be used against, so I'd like a second opinion on whether this is sound 🙂

vscode/test/fixtures/mock-server.ts
```diff
@@ -87,10 +89,31 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
         res.send(`event: completion\ndata: {"completion": ${JSON.stringify(response)}}\n\nevent: done\ndata: {}\n\n`)
     })
 
-    let isFirstCompletion = true
     app.post('/.api/completions/code', (req, res) => {
-        const response = isFirstCompletion ? responses.firstCode : responses.code
-        isFirstCompletion = false
+        const OPENING_CODE_TAG = '<CODE5711>'
```
DanTup, line 93: This constant is also defined in `vscode\src\completions\text-processing\utils.ts` but I can't import that here (because it references VS Code). Should we move the constants out of `utils.ts` into a `constants.ts` or something where it could be shared? (there's at least one other test with its own hard-coded version of this).

vscode/src/completions/client.ts
dominiccooney: Should be fine. I assume utf-8 is what we actually send?

vscode/test/fixtures/mock-server.ts
dominiccooney: I personally would not bother because the tests will break if they get out of sync. But if you feel it is a burden to have the definitions duplicated, go for it.

vscode/test/e2e/inline-completion.test.ts
```diff
@@ -10,11 +10,12 @@ test.beforeEach(() => {
 })
 test('shows chat sidebar completion onboarding notice on first completion accept', async ({ page, sidebar }) => {
     const expectedOrderedEvents = [
```
dominiccooney, line 12: @bobheadxi idly curious, but do the Telemetry V2 events use this list?

vscode/test/fixtures/mock-server.ts
```diff
@@ -87,10 +89,31 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
...
+        const request = req as MockRequest
+
+        // Extract the code from the last message.
+        let completionPrefix = request.body.messages.at(-1)?.text
+        if (completionPrefix?.startsWith(OPENING_CODE_TAG)) {
```
dominiccooney, line 98: Why are we conditionally doing this?

vscode/test/fixtures/mock-server.ts
```diff
@@ -87,10 +89,31 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
...
+            completionPrefix = completionPrefix.slice(OPENING_CODE_TAG.length)
+        }
+        // Trim to the last word since our mock responses are just triggered by a word.
+        completionPrefix = completionPrefix?.split(/\s/g).at(-1)
```
dominiccooney, line 102: Splitting on `\s` isn't a great way to look for words because -1 can be a space if you have a trailing space.

```suggestion
        completionPrefix = completionPrefix?.split(/\W+/g).filter(x => x).at(-1)
```

dominiccooney (approved):
Looks good, some minor suggestions inline.

vscode/test/e2e/inline-completion.test.ts
bobheadxi: No, the events have a different naming scheme. It _does_ record an analogous set of events that get received during a test though in the same setup/teardown stuff: `loggedV2Events`

vscode/src/completions/client.ts
DanTup: My understanding was that's the default (I also thought setting this header instructs Node to use it, though I can't find anything explicitly confirming that in docs). I did verify some emoji come through correctly to the mock server:

image

We're also already setting this same header in the chat completion code (which is also setting the `body` to a `JSON.stringify`'d string), so I believe it's all good - but if you think there's anything else I should verify, let me know!

vscode/test/fixtures/mock-server.ts
DanTup: I wasn't sure whether this was guaranteed. I think it probably should be, so I've changed this to instead `throw` if it's not. Does it seem reasonable?

vscode/test/fixtures/mock-server.ts
DanTup: I think we care about trailing spaces here though. If my prefix is `"abcd "` I wouldn't expect the server to return results like `"abcdef"` because I've explicitly typed that space (and if I understand correctly, we can't delete it when you accept the completion). The mock implementation was intended to just complete partial words for simplicity.

umpox:
@DanTup I just realised I've made a (possibly incorrect) change here which impacts this diff.

We have some racey conditions where telemetry logs can sometimes come in a different order. I uncovered this by removing an artificial delay from our commands in this PR: https://github.com/sourcegraph/cody/pull/1561

It feels like we shouldn't be relying on the order of telemetry logs to assert functionality, but it seems this is providing useful value here - and I'm worried _not_ doing this will cause bugs.

umpox:
I'll look into fixing this now, we can keep the current functionality for most tests and make an exception for the tests that are affected by the racey logs

DanTup:
@umpox do you mean you think these tests will be flaky now you've removed the sort? The bots are currently green (with your change merged in), but I don't know if they're consistent. With this PR there should be fewer events and they should be more consistent than before, although as-is (with your change and this), it does assume the completion events are suggested -> accepted -> suggested -> accepted.

umpox:
@DanTup The fixup tests will be flaky, these tests should be OK. The race condition was only with commands so doesn't affect completions. Doesn't need to block this PR just wanted to raise that these tests don't technically check the order on latest `main`, PR to fix that: https://github.com/sourcegraph/cody/pull/1625/files?diff=split&w=1

DanTup:
Ah, I see. Before your change the completion test logs events were being sorted because they were also flaky, so I suspect your PR failure is because you're now asserting the order of them.

I _think_ this PR landing will possibly fix that, so it may be better to wait to merge until after this, or restore the code that sorted the completion events before comparing and add a TODO to remove it after this PR lands (I'll wait for @dominiccooney to respond to comments above before merging).

DanTup:
@dominiccooney I responded to some comments above I'm interested in your thoughts on. Thanks!

vscode/test/fixtures/mock-server.ts
dominiccooney: Makes perfect sense.

Consider tweaking the comment along those lines? "or empty string if there's trailing space", something like that, more succinctly if you can manage it?

dominiccooney (approved):
Splendid work. Sorry for the slow reply.

vscode/test/fixtures/mock-server.ts
DanTup: I'm not sure if it classes as succinct, but I've updated the comment to be clearer :)

vscode/test/e2e/inline-completion.test.ts
DanTup: I'm assuming these comments are not requests for changes, but if there's something I should be changing/adding, please let me know (if this merges in the meantime, I'll do in another PR).

bobheadxi: I think a good thing to have might be to add assertions on `loggedV2Events` everywhere we assert on the old events, to track migration changes in tests as well?

PR #1537, opened 2023-10-28
telemetry-v2: remove need to manually enumerate feature/action/metadata keys

bobheadxi:
See https://github.com/sourcegraph/telemetry/pull/3 - this change removes the need to manually enumerate all the feature/action/metadata keys by doing some generics hackery to require that parameters by "known keys": string literals or string types e.g. `"foo" | "bar" | ...`

## Test plan

Build/CI, no functional changes


dominiccooney:
Does this allow `x ? 'literal A' : 'literal B'` as the argument? Some of the code I saw in earlier reviews had a lot of duplication and only differed in the argument, and I gave feedback to be more succinct and use `?:` to highlight the part that changed.

If this requires all the duplication, I think the "type is a literal" thing is too tedious and you should try something else.

bobheadxi:
@dominiccooney yes, because `x ? 'literal A' : 'literal B'` evaluates to `'literal A' | 'literal B'` which is a subset of `string` (i.e. `string` does not extend the result) :)

image

nathan-downs (approved):
nice!

PR #1541, opened 2023-10-28
Fix task-controller e2e test flakes on Windows

DanTup:
On Windows without this change, I always see this failure in task-controller e2e tests:

```
    Error: expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 0

      Array [
    -   "cody.auth/failed",
        "cody.auth/connected",
        "cody.command.edit/executed",
        "cody.recipe.fixup/executed",
        "cody.fixup.apply/succeeded",
      ]
```

The `"cody.auth/failed"` event is never logged. I'm not too familiar with this, but I think this is because the mock server is not up and running at the time the extension activates and first tries to sync authentication.

This change starts the server up slightly earlier and appears to avoid the flake. Reducing the timeout on line 69 also resolves it for me locally, but that feels like it might re-introduce other issues it was designed to fix.

@philipp-spiess perhaps you're a good reviewer for this since there's a TODO in this area with your name? 🙂

## Test plan

This is a change to the test code itself, to fix a fail/flake when running `pnpm test:e2e` on Windows.


dominiccooney:
If the flake is caused by the test server not being up, can we instead wait for it to be up? The test driver can hammer the test server or something.

DanTup:
@dominiccooney 

> If the flake is caused by the test server not being up, can we instead wait for it to be up? The test driver can hammer the test server or something.

I'm not sure if I understand the suggestion. Right now (at least, if I understand correctly), this setup code clicks to show the sidebar first, and _then_ starts the server. As the sidebar is loading/extension is activating, it tries to sync the auth state and emits a log event for not being signed in. There's a race between the event being logged and the server being up (and at least as far as I can tell, the log event is not re-sent).

This change starts the server before starting to interact with/activate the extension to avoid that race.

Is your suggestion to have something in the extension/sidebar activation that delays until it can verify the server is available? I think that would require test-only code inside the extension startup (which doesn't feel better than just re-ordering the server to be started first)?

DanTup:
@dominiccooney any thoughts on the above?

dominiccooney:
@DanTup  I am *so* sorry for commenting and then disappearing.

> There's a race between the event being logged and the server being up (and at least as far as I can tell, the log event is not re-sent).

Is the problem that the server collecting the logs isn't up, or the mock auth server isn't up?

> This change starts the server before starting to interact with/activate the extension to avoid that race. Is your suggestion to have something in the extension/sidebar activation that delays until it can verify the server is available?

If the flake is gone and auth/failed no longer appears, let's update the expectation to not mention auth/failed.

If it is still flaky, let's de-flake it, I was suggesting:
- Remove the auth/failed expectation from the log.
- Hammer the server *in the test harness* to ensure it is up, then proceed with the test.

philipp-spiess (approved):
This looks reasonable to me, even if we add a bazillion logging on extension startup, we don't want it to be part of every individual test that assets on the logged events.

DanTup:
I reverted my original change and this seems to pass, but it's still flaky. The real issue here is that from the outside, we don't know when any telemetry logs have been pushed to the server (because it's all async). Therefore we don't know when it's safe to clear the logs and start the test.

Using 1000ms seems to pass, but 500ms does not.

I wonder whether we should just be excluding auth events from these logs? The PR at #1104 seems to do that for some kinds of events, but not the ones being logged here (`CodyVSCodeExtension:Auth:failed` vs `cody.auth/failed`).

@philipp-spiess ?

philipp-spiess:
I wonder if we should change the matcher so that it doesn't matter if more events are logged. I feel like a product is bound to log more events over time and having to update all the tests (or ensure no flaky is caused because of it) seems tedious. When we test a feature, we only want to test that the events that are associated with the feature are logged. What do you think?

So we'd change from `expect().toEqual()` to `expect().arrayContaining()` (I _think_)

DanTup:
> So we'd change from `expect().toEqual()` to `expect().arrayContaining()` (I think)

This makes sense to me.. I don't think this test case about other events (and probably most don't). Probably that should be the default, and only if tests want to test for the absence of events they do an exact check (or a "does not contain" check)?

philipp-spiess:
@DanTup Yeah I think this is a better default, this would also be a better documentation. Thank you for driving this!

DanTup:
`arrayContaining()` ignores order which we probably don't want, so I'll loop through and check that all items appear in the right order, but extras are allowed (as far as I can see, playwright doesn't have a built in assertion for this).

DanTup:
@philipp-spiess this is green now - happy for me to land?

I left the extra call to `resetLoggedEvents` after startup in, and I've also left the success auth - but I could remove them (either/both) if you'd prefer.

Thanks!

philipp-spiess (approved):
Awesome!

Issue #1523, opened 2023-10-27
VSCE: New Chat - @ file context via chat input


abeatrix:
<img width="1337" alt="image" src="https://github.com/sourcegraph/cody/assets/68532117/0f86ee66-348b-4fc5-b46c-a9237b8b3519">

<img width="503" alt="image" src="https://github.com/sourcegraph/cody/assets/68532117/dccf6c6d-80d7-47de-913b-ab1e218fd96c">

```[tasklist]
### Tasks
- [ ] VSCE New Chat: @ file - Display file and symbol match results in input area
- [ ] VSCE New Chat: @ file - Display added context in Chat view with clickable links
- [ ] VSCE New Chat: @ file - Create context message
- [ ] Add new context type for input context
```

#### part 1

![image](https://github.com/sourcegraph/cody/assets/68532117/4f8681e7-500f-44b4-8868-f9c6ba101fe3)

WIP: Display file and symbol match results in input area

https://github.com/sourcegraph/cody/pull/1549

![image](https://github.com/sourcegraph/cody/assets/68532117/c3d94f15-8a1e-486a-842f-eb203cb7dbe2)

#### part 2

![image](https://github.com/sourcegraph/cody/assets/68532117/b9d7f262-bccb-494c-9377-50e3388494e2)

Not started

PR #1549, opened 2023-10-31
feat: at file in chat input box

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1523

## WIP

image


toolmantim:
Woohoo! I know this is a WIP, but I just gave it a go to give some early feedback. Awesome start. Two things I noticed:
1. The menu doesn't seem to trigger consistently when you type @… seems to want you to type a forward slash or . before appearing
2. Hitting Enter when the menu is showing shouldn't submit, that should only act as a selection

vscode/webviews/ChatFileSelector.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    if (!filePaths?.length) {
+        return
+    }
+
+    return (
+        <div className={classNames(styles.container)}>
+            <div className={classNames(styles.headingContainer)}>
+                <h3 className={styles.heading}>Add selected file as context...</h3>
+            </div>
```
toolmantim, line 50: Let's remove this heading to start with

abeatrix:
> Woohoo! I know this is a WIP, but I just gave it a go to give some early feedback. Awesome start. Two things I noticed:
> 
> 1. The menu doesn't seem to trigger consistently when you type @… seems to want you to type a forward slash or . before appearing
> 2. Hitting Enter when the menu is showing shouldn't submit, that should only act as a selection

Thanks for testing this out! I've fixed the both issues  in the PR.  Also fixed the same issue (#1) on main in https://github.com/sourcegraph/cody/pull/1592

PR #1557, opened 2023-10-31
chore: gitignore playwright failure recordings

umpox:
## Description

It's painful to have to keep deleting these for a clean git

## Test plan

e2e tests go brrrr


abeatrix (approved):
omg yes

PR #1561, opened 2023-10-31
MessageProvider: Show errors when Cody does not provide any response

umpox:
## Test plan

1. Change a prompt to force that Cody will not return a correct multiplexer topic. E.g. change <fixup> to <eeeee>
2. Run a command, check that the error is handled. Previously it would hang for the user.

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -213,7 +213,15 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
             },
             onTurnComplete: async () => {
                 typewriter.close()
-                await typewriter.finished
+                typewriter.stop()
+
+                const formattedBotMessage = reformatBotMessage(text, responsePrefix)
+                if (!formattedBotMessage) {
+                    // Cody either didn't provide any text, or didn't use the correct topic.
+                    // Emit an error so we can handle it rather than leaving the UI in a pending or empty state.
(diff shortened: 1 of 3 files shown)


vscode/src/chat/MessageProvider.ts
```diff
@@ -213,7 +213,15 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
             },
             onTurnComplete: async () => {
                 typewriter.close()
-                await typewriter.finished
+                typewriter.stop()
```
umpox, line 216: @dominiccooney Tagging you here as I saw you added this line. (Also @abeatrix for visibility)

WDYT about flushing the typewriter and stopping it immediately when Cody finishes responding?

There are other ways to fix this problem, but it seems that this is beneficial anyway. If we have a full response from Cody should we not just show it immediately?

vscode/src/chat/MessageProvider.ts
```diff
@@ -213,7 +213,15 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
...
+
+                const formattedBotMessage = reformatBotMessage(text, responsePrefix)
+                if (!formattedBotMessage) {
```
abeatrix, line 219:
```suggestion
                if (!text) {
```
Would text works better in case responsePrefix is not empty but text is?

umpox: good point, will update

abeatrix (approved):

PR #1563, opened 2023-10-31
Autocomplete: Remove recursion from lsp light

### LINKED ISSUE #1506: LSP-light context: hover context snippets from the same file are removed
Bumped into this issue testing the LSP-light-context locally and sharing my findings for further verification:

```ts
import { buildBanana } from './helper.ts'

buildBanana(█
```

In the example above, hover information is received for the `buildBanana` function, but then it's removed here: 

https://github.com/sourcegraph/cody/blob/211f5db74c843f776db56bec248571e6266ae4fb/vscode/src/completions/context/lsp-light-graph-cache.ts#L119-L121

because the hover info range overlaps with the completion prefix (its position is the beginning of the symbol in the current line prefix). 

AFAIU it works in unit tests like this one:
https://github.com/sourcegraph/cody/blob/211f5db74c843f776db56bec248571e6266ae4fb/vscode/src/completions/context/lsp-light-graph-cache.test.ts#L171

because `context.range` is `undefined` in tests.

philipp-spiess:
Closes #1506

The recursion doesn't work well and there's no ned for improving it right now. Better test a stupider version that is less prone to have perf issues before we roll out the bigger changes.

## Test plan

- Enable `lsp-light` context
- Open the trace view
- Trigger a completion and see that nothing but the hover information is added

Screenshot 2023-10-31 at 17 04 07

### DIFF
--- a/vscode/src/completions/context/lsp-light-graph-cache.test.ts
+++ b/vscode/src/completions/context/lsp-light-graph-cache.test.ts
@@ -94,14 +94,12 @@ describe('LSPLightGraphCache', () => {
         expect(getGraphContextFromRange).toHaveBeenCalledWith(
             testDocuments.document1,
             range(0, 0, 0, 19),
-            expect.anything(),
-            0 // no recursion
+            expect.anything()
         )
         expect(getGraphContextFromRange).toHaveBeenCalledWith(
             testDocuments.document1,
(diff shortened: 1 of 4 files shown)


abeatrix (approved):

valerybugakov (approved):
Yes!

PR #1567, opened 2023-10-31
Return null from getRepoId in the case of private repos

mkondratek:


lib/shared/src/chat/useClient.ts
```diff
@@ -273,7 +273,12 @@ export const useClient = ({
                 // Here we are passing the current repo & file context based on `options.scope.editor`
                 // if present.
                 const additionalRepoId = await graphqlClient.getRepoId(activeEditor.repoName)
-                if (isError(additionalRepoId)) {
+                if (additionalRepoId == null ) {
```
olafurpg, line 276:
```suggestion
                if (additionalRepoId === null ) {
```

olafurpg (approved):
LGTM 👍🏻 I think it's an improvement that we now distinguish between an actual error and an empty result. Would be good to get a stamp from the VS Code folks in case we are overlooking something.

umpox (approved):

valerybugakov (approved):

PR #1571, opened 2023-10-31
telemetry: log request_id per interaction

abeatrix:
add interactionID to codebaseContext and recipes

- Pass interactionID to codebaseContext getContextMessages
- Add interactionID to RecipeContext
- Add interactionID parameter to relevant recipe methods
- Log interactionID as request_id for each chat question submitted

A new interactionID is assigned whenever executeRecipe is run, which allows tracing in individual webview/view.  

## Test plan

Submit a question in chat, you should see all the logEvents has the same `request_id` for each chat question submitted.

I wanted to create an id for each interaction, but an interaction is not created until context is fetched so that didn't work out.

Example: 

Question 1
```
█ logEvent (telemetry disabled): CodyVSCodeExtension:keywordContext:searchDuration {"properties":{"searchDuration":839.5423329919577,"request_id":"d6fd00bd-7559-456c-8bea-ae58b197a95c"}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:chat-question:executed {"properties":{"contextSummary":{"embeddings":0,"local":7},"request_id":"d6fd00bd-7559-456c-8bea-ae58b197a95c"},"opts":{"hasV2Event":true}}
```

Question 2:
```
█ logEvent (telemetry disabled): CodyVSCodeExtension:keywordContext:searchDuration {"properties":{"searchDuration":900.0559170097113,"request_id":"b5e1f00b-96d1-470a-9b0d-d9238550af71"}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:recipe:chat-question:executed {"properties":{"contextSummary":{"embeddings":0,"local":8},"request_id":"b5e1f00b-96d1-470a-9b0d-d9238550af71"},"opts":{"hasV2Event":true}}
```

When paste a insert a code block:
```
█ logEvent (telemetry disabled): CodyVSCodeExtension:insertButton:clicked {"properties":{"op":"insert","charCount":98,"lineCount":5,"source":"chat-question","request_id":"b5e1f00b-96d1-470a-9b0d-d9238550af71"}}
``

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -1,3 +1,4 @@
+import * as uuid from 'uuid'
 import * as vscode from 'vscode'
 
 import { BotResponseMultiplexer } from '@sourcegraph/cody-shared/src/chat/bot-response-multiplexer'
@@ -85,6 +86,7 @@ export interface MessageProviderOptions {
 
 export abstract class MessageProvider extends MessageHandler implements vscode.Disposable {
     public currentChatID = ''
+    public currentInteractionID: string | undefined = undefined
 
     // input and chat history are shared across all MessageProvider instances
     protected static inputHistory: string[] = []
@@ -150,7 +152,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         this.handleSuggestions([])
         this.sendTranscript()
         this.sendHistory()
-        telemetryService.log('CodyVSCodeExtension:chatReset:executed', undefined, { hasV2Event: true })
+        this.logTelemetryService('CodyVSCodeExtension:chatReset:executed')
         telemetryRecorder.recordEvent('cody.messageProvider.chatReset', 'executed')
     }
 
@@ -162,7 +164,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         this.transcript = new Transcript()
         await this.clearAndRestartSession()
         this.sendHistory()
-        telemetryService.log('CodyVSCodeExtension:clearChatHistoryButton:clicked', undefined, { hasV2Event: true })
+        this.logTelemetryService('CodyVSCodeExtension:clearChatHistoryButton:clicked')
         telemetryRecorder.recordEvent('cody.messageProvider.clearChatHistoryButton', 'clicked')
     }
 
@@ -177,7 +179,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         await this.transcript.toJSON()
         this.sendTranscript()
         this.sendHistory()
-        telemetryService.log('CodyVSCodeExtension:restoreChatHistoryButton:clicked', undefined, { hasV2Event: true })
+        this.logTelemetryService('CodyVSCodeExtension:restoreChatHistoryButton:clicked')
         telemetryRecorder.recordEvent('cody.messageProvider.restoreChatHistoryButton', 'clicked')
     }
 
@@ -228,11 +230,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 const codeCount = countGeneratedCode(text)
                 if (codeCount?.charCount) {
                     const source = lastInteraction?.getHumanMessage().source || recipeId
-                    telemetryService.log(
-                        'CodyVSCodeExtension:chatResponse:hasCode',
-                        { ...codeCount, source },
-                        { hasV2Event: true }
-                    )
+                    this.logTelemetryService('CodyVSCodeExtension:chatResponse:hasCode', { ...codeCount, source })
                     telemetryRecorder.recordEvent(`cody.messageProvider.chatResponse.${source}`, 'hasCode', {
                         metadata: {
                             ...codeCount,
@@ -318,9 +316,18 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
     }
 
     public async executeRecipe(recipeId: RecipeID, humanChatInput = '', source?: ChatEventSource): Promise<void> {
+        this.currentInteractionID = uuid.v4()
         if (this.isMessageInProgress) {
             this.handleError('Cannot execute multiple recipes. Please wait for the current recipe to finish.', 'system')
             return
+        }
+
+        if (source === 'suggestion') {
+            this.logTelemetryService('CodyVSCodeExtension:chatPredictions:used')
+        }
+
+        if (source === 'chat' && this.contextProvider.config.experimentalChatPredictions && humanChatInput) {
+            void this.runRecipeForSuggestion('next-questions', humanChatInput)
         }
 
         // Filter the human input to check for chat commands and retrieve the correct recipe id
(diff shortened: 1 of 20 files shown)

### CODE REVIEW

File: vscode/src/chat/MessageProvider.ts [hunk 1]
```diff
@@ -625,6 +623,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         const customInteraction = await newInteraction({
             displayText: humanInput,
             assistantDisplayText: assistantResponse,
+            request_id: this.currentInteractionID,
```
valerybugakov, line 626: I'm confused. We rename the same value (`this.currentInteractionID = uuid.v4()`) multiple times from interaction ID to request ID and the other way around. Would it make sense to keep the same name everywhere? 

The PR description says:
> I wanted to create an id for each interaction, but an interaction is not created until context is fetched so that didn't work out.

Does that mean we do not have the "true" interaction ID yet?

vscode/src/chat/MessageProvider.ts
```diff
@@ -783,6 +782,11 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
 
         return DEFAULT_MAX_TOKENS - solutionLimit
     }
+
+    public logTelemetryService(eventName: string, properties?: any, hasVSEvent = true): void {
```
valerybugakov, line 786:
```suggestion
    public logTelemetryService(eventName: string, properties?: Record<string, unknown>, hasVSEvent = true): void {
```

Or something better if you have anything in mind.

valerybugakov (commented):
Left inline comments. Functionality-wise, it's ✅, but we can also improve the docs and safety related to the `request_id` field added to analytics events.

vscode/src/local-context/local-keyword-context-fetcher.ts
```diff
@@ -128,7 +127,7 @@ export class LocalKeywordContextFetcher implements KeywordContextFetcher {
             })
         )
         const searchDuration = performance.now() - startTime
-        telemetryService.log('CodyVSCodeExtension:keywordContext:searchDuration', { searchDuration })
+        telemetryService.log('CodyVSCodeExtension:keywordContext:searchDuration', { searchDuration, request_id })
```
valerybugakov, line 130: With the current implementation, it's easy to call interaction-related events without the `request_id`. Also, it's unclear what `request_id` means here and where to get it. The only way to do that is to follow the call chain to the `executeRecipe` function, where the interaction ID is initialized. 

One option for making it safe is to implement an interaction logger similar to [the autocomplete logger](https://github.com/sourcegraph/cody/blob/8cc1f3d71e8fa25d31883ff8d971126d977e1e0d/vscode/src/completions/logger.ts#L23), which would define the interaction ID and emit all the interaction-related events from the exposed methods. This way, we can require the interaction ID argument and make it easy to determine where to get it from.

abeatrix: I wasn't sure how to get it to work since we have multiple services that can execute recipes asynchronously while sharing one event logger but what you shared seems like a good solution (instead of sharing one eventlogger each interaction will have its own?)

vscode/src/chat/MessageProvider.ts
abeatrix: Yea sorry for the confusion, I'll update to just use request_id instead because I'd assume InteractionID is created by interaction, but in this case it is not because it's an ID we assign for the whole request, which includes an interaction 😅

vscode/src/local-context/local-keyword-context-fetcher.ts
abeatrix: Does this mean instead of passing down a request_id, we will pass the logger into recipe and context fetcher interaction instead? 🤔

vscode/src/chat/MessageProvider.ts
valerybugakov: Got it, let's use `requestID` or `requestId` for consistent casing.

vscode/src/local-context/local-keyword-context-fetcher.ts
valerybugakov: > Does this mean instead of passing down a request_id, we will pass the logger into recipe and context fetcher interaction instead?

In autocomplete, we import the logger from the module scope and pass the completion ID to its methods. It's a collection of functions that are reused for all the completions (even if they are generated in parallel). The catch is having interaction logger methods with required arguments (`requestID` and others if needed).

bobheadxi: Just chiming in to mention this again - the tracing approach will negate the need for all this clientside lift, as clientside you can just describe a span when something starts and automagically propagate that to the server, and the server will own linking it to the corresponding trace - I cc'd everyone over in this thread to discuss the possibility: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1698783422921649?thread_ts=1697566943.698939&cid=C04MSD3DP5L

bobheadxi:
Another note - isn't the goal of this work to be able to tie clientside interactions into the "real" cost of completions? Right now, this ID has no way of being extracted in the backend to include in backend events or forward to Cody Gateway events, as it's embedded in event metadata.

https://github.com/sourcegraph/sourcegraph/pull/58016 proposes an approach for propagating manual request IDs better, as the backend can own adding request IDs to event metadata _and_ be able to forward it elsewhere, but per [this thread](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1698783422921649?thread_ts=1697566943.698939&cid=C04MSD3DP5L) tracing is a far more sustainable way to do this well

abeatrix:
> Another note - isn't the goal of this work to be able to tie clientside interactions into the "real" cost of completions? Right now, this ID has no way of being extracted in the backend to include in backend events or forward to Cody Gateway events, as it's embedded in event metadata.

Oh that's something @valerybugakov going to work on. This is for linking chat/commands actions with an ID: https://sourcegraph.slack.com/archives/C05AGQYD528/p1698355247771849?thread_ts=1697738493.299929&cid=C05AGQYD528

bobheadxi:
> Oh that's something @valerybugakov going to work on. This is for linking chat/commands actions with an ID: https://sourcegraph.slack.com/archives/C05AGQYD528/p1698355247771849?thread_ts=1697738493.299929&cid=C05AGQYD528

It looks like they solve for the same thing: each interaction (chat/commands actions) should end up with an ID that we can use to link related events. If Valery adds OpenTelemetry on top and we add that universally as metadata on top of event logs and events (the current plan), we not have duplicated ways of aligning data

abeatrix:
> > Oh that's something @valerybugakov going to work on. This is for linking chat/commands actions with an ID: https://sourcegraph.slack.com/archives/C05AGQYD528/p1698355247771849?thread_ts=1697738493.299929&cid=C05AGQYD528
> 
> It looks like they solve for the same thing: each interaction (chat/commands actions) should end up with an ID that we can use to link related events. If Valery adds OpenTelemetry on top and we add that universally as metadata on top of event logs and events (the current plan), we not have duplicated ways of aligning data
> 
> What Kevin asked for in that thread - being able to link chains of events - is exactly what tracing was designed for: https://opentelemetry.io/docs/concepts/signals/traces/

Does this cover cases where a user copy code from a code block that was generated from an old conversation?

abeatrix:
moved to https://github.com/sourcegraph/cody/pull/1586

PR #1579, opened 2023-11-01
Completions: Fix duplicate InlineCompletionItemProvider bug

umpox:
## Description

We often call `setupAutocomplete` multiple times, which throws a VS Code error and possibly causes other unintended behavior.

image

This change queues the calls so we don't attempt to create parallel completion providers.

## Test plan

Tested locally:
- Open a fresh VS Code window with the Cody sidebar already open
- Confirm no errors like `[sourcegraph.cody-ai]command 'cody.autocomplete.manual-trigger' already exists` are shown in dev tools

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -480,50 +480,54 @@ const register = async (
     updateAuthStatusBarIndicator()
 
     let completionsProvider: vscode.Disposable | null = null
+    let setupAutocompleteQueue = Promise.resolve() // Create a promise chain to avoid parallel execution
     disposables.push({ dispose: () => completionsProvider?.dispose() })
-    const setupAutocomplete = async (): Promise<void> => {
-        const config = getConfiguration(vscode.workspace.getConfiguration())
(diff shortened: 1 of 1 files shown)


vscode/src/main.ts
```diff
@@ -480,50 +480,54 @@ const register = async (
...
-        if (!config.autocomplete) {
-            completionsProvider?.dispose()
-            completionsProvider = null
-            if (config.isRunningInsideAgent) {
-                throw new Error(
-                    'The setting `config.autocomplete` evaluated to `false`. It must be true when running inside the agent. ' +
-                        'To fix this problem, make sure that the setting cody.autocomplete.enabled has the value true.'
-                )
+    const setupAutocomplete = (): void => {
+        setupAutocompleteQueue = setupAutocompleteQueue.then(async () => {
```
umpox, line 486: spammed the function and double checked that these are garbage collected once used

image

valerybugakov (approved):
Great find. Thank you, @umpox!

PR #1580, opened 2023-11-01
Make manual context option take priority over flags

SuperAuguste:
## Test plan

Tested locally.

### DIFF
--- a/vscode/src/completions/create-inline-completion-item-provider.ts
+++ b/vscode/src/completions/create-inline-completion-item-provider.ts
@@ -66,12 +66,17 @@ export async function createInlineCompletionItemProvider(
     ])
     if (providerConfig) {
         const history = new VSCodeDocumentHistory()
-        const graphContextFetcher: GraphContextFetcher | undefined =
-            config.autocompleteExperimentalGraphContext === 'lsp-light' || lspGraphContextFlag
-                ? LspLightGraphCache.createInstance()
-                : config.autocompleteExperimentalGraphContext === 'bfg' || bfgGraphContextFlag
-                ? platform.createBfgContextFetcher?.(context, gitDirectoryUri)
-                : undefined
+        let graphContextFetcher: GraphContextFetcher | undefined
+
+        if (config.autocompleteExperimentalGraphContext === 'lsp-light') {
+            graphContextFetcher = LspLightGraphCache.createInstance()
+        } else if (config.autocompleteExperimentalGraphContext === 'bfg') {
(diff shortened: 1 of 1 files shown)


vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -66,12 +66,17 @@ export async function createInlineCompletionItemProvider(
...
-        const graphContextFetcher: GraphContextFetcher | undefined =
-            config.autocompleteExperimentalGraphContext === 'lsp-light' || lspGraphContextFlag
-                ? LspLightGraphCache.createInstance()
-                : config.autocompleteExperimentalGraphContext === 'bfg' || bfgGraphContextFlag
-                ? platform.createBfgContextFetcher?.(context, gitDirectoryUri)
-                : undefined
+        let graphContextFetcher: GraphContextFetcher | undefined
+
+        if (config.autocompleteExperimentalGraphContext === 'lsp-light') {
+            graphContextFetcher = LspLightGraphCache.createInstance()
```
olafurpg, line 72: I think the following implementation would be cleaner because it avoids the duplication in the constructors

```ts
        const isLspLightEnabled =
            config.autocompleteExperimentalGraphContext === 'lsp-light' ||
            (!config.autocompleteExperimentalGraphContext && lspGraphContextFlag)
        const isBfgEnabled =
            config.autocompleteExperimentalGraphContext === 'bfg' ||
            (!config.autocompleteExperimentalGraphContext && bfgGraphContextFlag)
        const graphContextFetcher: GraphContextFetcher | undefined = isLspLightEnabled
            ? LspLightGraphCache.createInstance()
            : isBfgEnabled
            ? platform.createBfgContextFetcher?.(context, gitDirectoryUri)
            : undefined
```

olafurpg (approved):

vscode/src/completions/create-inline-completion-item-provider.ts
SuperAuguste: Frankly this is unreadable, I much prefer duplicated and readable code to this :)

valerybugakov (approved):
👍

vscode/src/completions/create-inline-completion-item-provider.ts
olafurpg: This would be much cleaner as a separate function. I can live with a lot of duplication, but this kind of duplication tends to become problematic in my experience when 

- you start changing the constructor APIs 
- you add a third type of graph context

Not a big deal, I'm fine merging as is.

PR #1589, opened 2023-11-01
fix: remove extra call to loadChatHistory

abeatrix:
fix: remove extra call to loadChatHistory()

This fixes issue where the last response from Cody is missing after starting a new chat session

This commit removes the extra unnecessary call, which avoids loading the old chat history before saving it in `saveChatHistory`.

## Test plan

Ask Cody a question, and then restart your VS Code. Go back to your chat history for your last session.

#### Before 

Cody's response in the last chat was not stored

image

#### After

Cody's response in the last chat was stored properly and can be seen in UI

image


kalanchan (approved):
lets go!

PR #1607, opened 2023-11-02
Autocomplete: reuse the last candidate if the user types forward as suggested despite the updated completion info

valerybugakov:
## Context

- Reuse the last candidate if the user types forward as suggested despite the updated completion info.
- See the new added test with the inline description for the use-case. 

## Test plan

Updated unit tests

### DIFF
--- a/vscode/src/completions/text-processing/process-inline-completions.ts
+++ b/vscode/src/completions/text-processing/process-inline-completions.ts
@@ -56,6 +56,10 @@ function processCompletion(completion: ParsedCompletion, params: ProcessItemPara
     const { prefix, suffix, currentLineSuffix, multilineTrigger } = docContext
     let { insertText } = completion
 
+    if (completion.insertText.length === 0) {
+        return completion
+    }
+
     if (docContext.injectedPrefix) {
         insertText = docContext.injectedPrefix + completion.insertText
     }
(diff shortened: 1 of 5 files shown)


vscode/src/completions/text-processing/process-inline-completions.ts
```diff
@@ -56,6 +56,10 @@ function processCompletion(completion: ParsedCompletion, params: ProcessItemPara
     const { prefix, suffix, currentLineSuffix, multilineTrigger } = docContext
     let { insertText } = completion
 
+    if (completion.insertText.length === 0) {
+        return completion
```
abeatrix, line 60: Question: why do we still want to return completion if there is nothing to insert? Is it for logging purpose?

abeatrix (approved):

vscode/src/completions/text-processing/process-inline-completions.ts
valerybugakov: In this function, we post-process completions, add additional fields, or change the insert text, so if the completion is the empty string, there's nothing to do with it. It will be filtered out in one of the next steps: https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/text-processing/process-inline-completions.ts#L178

PR #1608, opened 2023-11-02
Autocomplete: log partial acceptance events on every word instead of every char

valerybugakov:
## Context

- Log partial acceptance events on every word instead of every char to reduce the volume of the partial acceptance event being sent.

## Test plan

Updated unit tests


abeatrix (approved):
Makes sense!

PR #1611, opened 2023-11-02
Integrating autocompletion with lookup manager

Sa1to:
This is needed to fix [JetBrains: integrate autocomplete with lookup manager#57335](https://github.com/sourcegraph/sourcegraph/issues/57335).

It appends element from IntelliJ's lookup manager to the file in place of autocompletions, in order to generate completion including part from lookup manager.

## Test plan
Test requires changes on Jetbrains part so it is not possible at the moment. Those changes should not affect regular autocompletion behaviour.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -197,6 +197,14 @@ export class Agent extends MessageHandler {
                 return { items: [] }
             }
 
+            if (params.lookupElement !== undefined && document.content !== undefined && params.position !== undefined) {
+                const lookup = params.lookupElement
+                const lastPart = document.content.slice(params.offset, document.content.length)
+                const firstPart = document.content.slice(0, params.offset)
+                document.content = firstPart + lookup + lastPart
+                params.position.character = params.position.character + lookup.length
+            }
+
             const textDocument = new AgentTextDocument(document)
 
             try {
@@ -218,6 +226,21 @@ export class Agent extends MessageHandler {
                         : result.items.flatMap(({ insertText, range }) =>
                               typeof insertText === 'string' && range !== undefined ? [{ insertText, range }] : []
                           )
+
+                if (
+                    params.lookupElement !== undefined &&
+                    document.content !== undefined &&
+                    params.offset !== undefined &&
+                    params.position !== undefined
+                ) {
+                    const lastPart = document.content.slice(
+                        params.offset + params.lookupElement.length,
+                        document.content.length
+                    )
+                    const firstPart = document.content.slice(0, params.offset)
+                    document.content = firstPart + lastPart
+                }
+
                 return { items, completionEvent: (result as any)?.completionEvent }
             } catch (error) {
                 console.log('autocomplete failed', error)
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -198,11 +198,10 @@ export class Agent extends MessageHandler {
             }
 
             if (params.lookupElement !== undefined && document.content !== undefined && params.position !== undefined) {
-                const lookup = params.lookupElement
-                const lastPart = document.content.slice(params.offset, document.content.length)
-                const firstPart = document.content.slice(0, params.offset)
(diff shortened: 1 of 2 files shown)


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -115,6 +115,8 @@ export interface AutocompleteParams {
     // Defaults to 'Automatic' for autocompletions which were not explicitly
     // triggered.
     triggerKind?: 'Automatic' | 'Invoke'
+    lookupElement?: string
```
olafurpg, line 118: Can we add comments explaining how these fields are used?

agent/src/agent.ts
```diff
@@ -197,6 +197,14 @@ export class Agent extends MessageHandler {
                 return { items: [] }
             }
 
+            if (params.lookupElement !== undefined && document.content !== undefined && params.position !== undefined) {
+                const lookup = params.lookupElement
+                const lastPart = document.content.slice(params.offset, document.content.length)
+                const firstPart = document.content.slice(0, params.offset)
```
mkondratek, line 203: Can you pls explain to me what is 
- offset
- firstPart
- lastPart
❓

mkondratek (approved):
LGTM.
I guess it would be nice to test it with the changes to `jetbrains` BEFORE merging this.
btw, please fix the PR title

agent/src/agent.ts
```diff
@@ -197,6 +197,14 @@ export class Agent extends MessageHandler {
...
+                document.content = firstPart + lookup + lastPart
```
olafurpg: This solution feels like a hack. We want `document.content` to reflect the actual content in the editor buffer. For example, what happens when we send multiple subsequent requests with a different `lookupElement`. Ideally, we should use the same API as VS Code for the suggestion widget. Have you looked at `InlineCompletionContext.selectedCompletionInfo`?

olafurpg (changes requested):

Sa1to:
Updated PR. Here is example how it works with my local changes in JetBrains plugin part:

https://github.com/sourcegraph/cody/assets/9321940/2429f648-0947-4d41-8e13-7cef2928985d

olafurpg (changes requested):
One blocking comment, otherwise, this looks good

mkondratek (approved):
the demo looks great 🚀 pls apply Olaf's suggestions, lgtm

exigow (approved):
This is one of the features that is compatible with native look & feel in IntelliJ. IMO most of the users will prefer using this default lookup manager than custom one invoked by different keybindings. Can't wait!

About code: it looks fine :+1:, though I'm not yet comfortable with the TS agent's code so I might be missing something.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -1,5 +1,7 @@
 /* eslint-disable @typescript-eslint/consistent-type-definitions */
 
+import { SelectedCompletionInfo } from 'vscode'
```
olafurpg, line 3: Let's inline the definition of this interface and remove the import from `vscode` here because 1) the types are incompatible and 2) agent-protocol should be as standalone as possible meaning we should limit external imports. Re 1) the type for `.range` is a class in `vscode` while in the protocol it should be an interface.

olafurpg (approved):
Approving to unblock merge but please address the last comment before merging 🙇🏻

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -115,8 +115,13 @@ export interface AutocompleteParams {
     // Defaults to 'Automatic' for autocompletions which were not explicitly
     // triggered.
     triggerKind?: 'Automatic' | 'Invoke'
+    selectedCompletionInfo?: SelectedCompletionInfo
 }
 
+export interface SelectedCompletionInfo {
```
olafurpg, line 121: Thank you! LGTM

PR #1612, opened 2023-11-02
Benchmark: Move to separate repository

umpox:
## Description

The benchmark tool has moved to https://github.com/sourcegraph/coder-eval. This is so we don't have to pollute this repository with hundreds of benchmark files.

You can still run the benchmark script against a local build, [see docs](https://github.com/sourcegraph/coder-eval)

## Test plan

N/A - Deleting non production code


abeatrix (approved):
🙏

PR #1617, opened 2023-11-02
fix: code block actions event handlers

abeatrix:
Pass requestID and source metadata to copy button event handlers directly instead of through dependencies.

Using metadata as a dependency caused useEffect to keep re-rendering and append multiple copy buttons on mount. Passing the metadata directly to the event handler fixes this issue.

#### Fix

By constructing the metadata object inside the callback and passing it directly, the metadata gets passed to the buttons without causing re-renders from useEffect.

This avoids the issue of useEffect appending multiple buttons on each render.

## Test plan

To reproduce:

1. Ask Cody a question that generate code: `What does a button looks like in html?`
2. in the same Window, send Cody another question, e.g. `Hello`

### Before

image

### After

image

### DIFF
--- a/lib/ui/src/chat/CodeBlocks.tsx
+++ b/lib/ui/src/chat/CodeBlocks.tsx
@@ -169,20 +169,22 @@ export const CodeBlocks: React.FunctionComponent<CodeBlocksProps> = React.memo(f
         }
 
         const preElements = rootRef.current?.querySelectorAll('pre')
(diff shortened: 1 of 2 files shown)


kalanchan (approved):

kalanchan:
nice catch, I tested it on 14.3 and on your branch, was able to see the fix correctly 

v0.14.3:
image

your branch:
image

PR #1621, opened 2023-11-03
Edit: Fix recursive diff expansion

umpox:
## Description

We use a temp document to provide a "before" version of the edited fixup file. There is an edge case where, if we are adding new code (thus deleting from the temp document), we can reuse the already diffed temp document. This leads to the diff expanding every time it is triggered, as we delete more code from the selection.

We now ensure each diff has its own unique ID, so it always uses a fresh "temp" document. 

### Before

https://github.com/sourcegraph/cody/assets/9516420/c584f8a5-e21c-468f-bb82-c9eece29cc38

### After

https://github.com/sourcegraph/cody/assets/9516420/20eb4fd5-b318-47d2-b66c-c9a3b9d076c1

## Test plan

Tested locally:
- Diff for edits that expand ranges
- Diff for an edit where the doc has been changed elsewhere after it was applied

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -12,6 +12,8 @@ Starting from `0.2.0`, Cody is using `major.EVEN_NUMBER.patch` for release versi
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


abeatrix (approved):
Nice catch!

PR #1625, opened 2023-11-03
E2E: Assert against order of telemetry logs

umpox:
## Description

Reverts some of the changes from https://github.com/sourcegraph/cody/pull/1561

The order of telemetry events does actually provide value here, and can help us catch some bugs. We have some race conditions when sending events for commands, but we can handle these separately. (I thought that we might not actually care about the order of these events being sent but I think it's actually valuable)

## Test plan

E2E tests pass with no flakes.

### DIFF
--- a/vscode/test/e2e/inline-completion.test.ts
+++ b/vscode/test/e2e/inline-completion.test.ts
@@ -3,7 +3,7 @@ import { expect, Locator, Page } from '@playwright/test'
 import { loggedEvents, resetLoggedEvents } from '../fixtures/mock-server'
 
 import { sidebarExplorer, sidebarSignin } from './common'
-import { assertEvents, test } from './helpers'
+import { test } from './helpers'
 
 test.beforeEach(() => {
     resetLoggedEvents()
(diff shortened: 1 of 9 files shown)


vscode/test/e2e/inline-completion.test.ts
```diff
@@ -105,7 +105,7 @@ test('inline completion onboarding notice on first completion accept', async ({
     await acceptInlineCompletion(page)
     await expect(otherAcceptedCompletion).toBeVisible()
     await expect(decoration).not.toBeVisible()
-    await assertEvents(loggedEvents, expectedEvents)
+    await expect.poll(() => loggedEvents).toEqual(expectedEvents)
```
DanTup, line 108: These might need to be sorted for now. I think the reason for the flakes is that the acceptance and triggering of another inline completion race, but https://github.com/sourcegraph/cody/pull/1535 will remove some of the (unwanted) completions which should result in a more consistent log (of just suggest, accept, suggest, accept).

philipp-spiess (approved):

PR #1628, opened 2023-11-03
convert issue markdown templates to issue forms

kalanchan:
converting markdown issues to issue forms for templating features

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

this allows us to create more powerful form like templates instead of just markdown

example:
CleanShot 2023-11-03 at 13 17 36
CleanShot 2023-11-03 at 13 18 11

## Test plan
n/a


abeatrix (approved):
thanks for writing this!

PR #1629, opened 2023-11-03
fix grammar in prd-issues.yml

kalanchan:
fix grammar

## Test plan
n/a


abeatrix (approved):
maybe also on line 27?

I want more control ~on~ over what is fed into the LLM for answers. I should easily be able to distinguish what my last question was versus what was displayed from commands.

PR #1634, opened 2023-11-04
Support nested ignore files

DanTup:


DanTup:
@abeatrix these tests are all green now (although there's a TODO about which URI the Ignore tests should be using).

I had a quick test and creating an ignore file did exclude files from the automatic context. I couldn't get the file selection at the bottom of chat working though, it seems to require the app (which I understand isn't available on Windows?)

abeatrix:
@DanTup thank you so much for fixing the tests, and getting nested ignore file to work in less than a day 🤯

Regarding the file selection at the bottom below the chat, it doesn't required App to work iirc. It gets updated to whatever file you're currently on but it's not being sent to the llm. I can look into it more on Monday unless you are referring to something else?

lib/shared/src/chat/ignore-helper.ts
```diff
@@ -0,0 +1,134 @@
...
+        const relativePath = path.relative(workspaceRoot, uri.fsPath)
+        const rules = this.workspaceIgnores.get(workspaceRoot) ?? this.getDefaultIgnores()
+        return rules.ignores(relativePath) ?? false
+    }
+
+    private findWorkspaceRoot(filePath: string): string | undefined {
+        const candidates = Array.from(this.workspaceIgnores.keys()).filter(workspaceRoot =>
+            filePath.toLowerCase().startsWith(workspaceRoot.toLowerCase())
+        )
+        // If this file was inside multiple workspace roots, take the shortest one since it will include
```
abeatrix, line 101: 🧠

abeatrix (approved):
LGTM! I love how well documented everything is in your code, something I really need to learn from you! I can look into the file selection thing more on Monday plus I know you're about to start on the plg work soon, so feel free to merge this whenever, and we can put this behind a feature flag for internal users and other clients to dogfood, what do you think?

DanTup:
> Regarding the file selection at the bottom below the chat, it doesn't required App to work iirc. It gets updated to whatever file you're currently on but it's not being sent to the llm. I can look into it more on Monday unless you are referring to something else?

I tested the automatically file selection - that seemed to work (the "Read 3 files" but excluded my current file after I added to the ignore list). However there was another button underneath that seemed like it was to let me manually select the files, but I couldn't get it to work (it said it wanted the app). I might have misunderstood that functionality though.

I'll merge this into your branch and let you add the feature flag before merging to master then. If I run out/get blocked on PLG stuff I can finish if needed. Let me know if you have an issues/questions with the bits I did.

Some tests I hadn't written yet but thought would be valuable (probably they'd need to be integration tests):

- Ensure the VS Code parts are setting up the watcher correctly (verify a file that's not ignored, modify the ignore file on disk, verify it's now ignored)
- Ensure the VS Code parts are updating when workspace changes (open a workspace, add an extra workspace folder using `vscode.workspace.updateWorkspaceFolders`, ensure the ignores for files in the second workspace are correct)

PR #1668, opened 2023-11-07
Cody: Fix stop generating error

umpox:
## Description

There is a case where a user aborts the chat before a message is sent. In that case we don't actually want an error. We can handle this for each individual case.

We now already do this for fixups, which this was originally added for:

https://sourcegraph.com/github.com/sourcegraph/cody@tr/fix-stop-gen-error/-/blob/vscode/src/chat/FixupViewProvider.ts?L101%3A1-104%3A10

image

## Test plan

- Check no error when "Stop generating" is clicked
- Check no error when fixup is generated with incorrect xml tag


abeatrix (approved):
Thank you! Worth adding to changelog?

PR #1672, opened 2023-11-07
add fileUri

abeatrix:

### DIFF
--- a/agent/src/editor.ts
+++ b/agent/src/editor.ts
@@ -1,4 +1,5 @@
 import type * as vscode from 'vscode'
+import { URI } from 'vscode-uri'
 
 import { isCodyIgnoredFile } from '@sourcegraph/cody-shared/src/chat/context-filter'
 import {
@@ -37,7 +38,7 @@ export class AgentEditor implements Editor {
         if (this.agent.workspace.activeDocumentFilePath === null) {
             return undefined
         }
-        if (isCodyIgnoredFile(this.agent.workspace.activeDocumentFilePath)) {
+        if (isCodyIgnoredFile(URI.parse(this.agent.workspace.activeDocumentFilePath))) {
             return undefined
         }
         return this.agent.workspace.getDocument(this.agent.workspace.activeDocumentFilePath)
--- a/lib/shared/src/chat/transcript/interaction.ts
+++ b/lib/shared/src/chat/transcript/interaction.ts
@@ -39,13 +39,13 @@ export class Interaction {
         const newMessages = []
         for (let i = 0; i < contextMessages.length; i++) {
             const message = contextMessages[i]
-            if (message.speaker === 'human' && message.file?.fileUri?.fsPath) {
+            if (message.speaker === 'human' && message.file?.uri) {
                 // Skip embeddings results as they should be filtered when it was indexed
                 if (message.file.source === 'embeddings') {
                     continue
                 }
                 // Skips the assistant message if the human message is ignored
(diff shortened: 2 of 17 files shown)


lib/shared/src/chat/transcript/interaction.ts
```diff
@@ -39,9 +39,13 @@ export class Interaction {
         const newMessages = []
         for (let i = 0; i < contextMessages.length; i++) {
             const message = contextMessages[i]
-            if (message.speaker === 'human' && message.file?.fileName) {
+            if (message.speaker === 'human' && message.file?.fileUri?.fsPath) {
+                // Skip embeddings results as they should be filtered when it was indexed
+                if (message.file.source === 'embeddings') {
+                    continue
+                }
```
DanTup, line 46: Will embeddings have `fileUri` at all? If not (if it's made a rule that `fileUri` is only ever set if it's absolute) then the `if` condition above might handle this?

lib/shared/src/codebase-context/messages.ts
```diff
@@ -9,6 +11,9 @@ export type ContextFileSource = 'embeddings'
 
 export interface ContextFile {
     fileName: string
+
+    fileUri?: URI
```
DanTup, line 15: I wonder if we should just call this `uri`? In future, maybe documents aren't all from the file system with a file path (VS Code supports various schemes and workspaces without files), so it might be less to change later even if today they're all `file://` URIs?

vscode/src/completions/context/context.ts
```diff
@@ -57,8 +57,8 @@ export async function getContext(options: GetContextOptions): Promise<GetContext
     let totalChars = 0
     function addMatch(match: ContextSnippet): boolean {
         // TODO(@philipp-spiess): We should de-dupe on the snippet range and not
-        // return early if the file is on ignore list
-        if (isCodyIgnoredFile(match.fileName)) {
+        // TODO: makes sure fileUri is included for all context snippets
+        if (match.fileUri?.fsPath && isCodyIgnoredFile(match.fileUri?.fsPath)) {
```
DanTup, line 61: I also wonder if we should make `isCodyIgnoredFile` take a URI? 🤔

One one hand, it's harder to pass a relative path (it's probably more obvious that a `file` URI should be absolute than a path), but on the other hand you're less likely to pass a non-file URI this way (and the implementation would have to throw for non-file URIs for now if we accepted URIs).

DanTup (approved):

vscode/src/completions/context/context.ts
abeatrix: Yea I like this idea a lot. It's much clear than passing in a string that could be anything and much less confusing for the caller and the users! Will update it first thing tomorrow!

agent/src/editor.ts
```diff
@@ -37,7 +38,7 @@ export class AgentEditor implements Editor {
         if (this.agent.workspace.activeDocumentFilePath === null) {
             return undefined
         }
-        if (isCodyIgnoredFile(this.agent.workspace.activeDocumentFilePath)) {
+        if (isCodyIgnoredFile(URI.parse(this.agent.workspace.activeDocumentFilePath))) {
```
DanTup, line 41: I think these will need to be `URI.file()` to work on Windows.

I'll merge this into my branch and fix up there and see what the tests are like (it might also be easier than having so many layers or PRs 😄)

DanTup (approved):

PR #1674, opened 2023-11-07
Add a feature toggle for the new search panel

beyang:
image

## Test plan

Minor config change affecting experimental feature


abeatrix (approved):

abeatrix:
(will need to update the test files too)

PR #1677, opened 2023-11-08
fix: open document side by side with chat panel

abeatrix:
RE: https://sourcegraph.slack.com/archives/C05AGQYD528/p1699403470817339

fix: open document side by side with chat panel

Check current webview panel location before opening the document to show it side by side instead of replacing the chat panel.

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/bca885f8-8b09-4908-8500-707e5b3eeb91


beyang (approved):
This is better than the current behavior, but what if there are more than two columns?

abeatrix:
> This is better than the current behavior, but what if there are more than two columns?

good point. Updated to calculate the view column to open the file in based on the current webview panel's view column. 
This will make sure to always open the file in the view column beside the current webview panel instead of hardcoded view columns.

PR #1682, opened 2023-11-08
Autocomplete: Migrate retrieval strategies to new API

### LINKED ISSUE #1640: Autocomplete: Unify retrieval interfaces

philipp-spiess:
Closes #1640 

This PR is a restructure of the existing context gather to implement a new and shared retrieval API.

The API is defined as such:

```ts
/**
 * Interface for a general purpose retrieval strategy. During the retrieval phase, all retrievers
 * that are outlined in the execution plan will be called concurrently.
 */
export interface ContextRetriever extends vscode.Disposable {
    /**
     * The identifier of the retriever. Used for logging purposes.
     */
    identifier: string

    /**
     * Start a retrieval processes. Implementers should observe the hints to return the best results
     * in the available time.
     *
     * The client hints signalize _soft_ timeouts. When a hard timeout is reached, the retriever's
     * results will not be taken into account anymore so it's suggested to return _something_ during
     * the maxMs time.
     *
     * The abortSignal can be used to detect when the completion request becomes invalidated. When
     * triggered, any further work is ignored so you can stop immediately.
     */
    retrieve(options: ContextRetrieverOptions): Promise<ContextSnippet[]>

    /**
     * Return true if the retriever supports the given languageId.
     */
    isSupportedForLanguageId(languageId: string): boolean
}

export interface ContextRetrieverOptions {
    document: vscode.TextDocument
    position: vscode.Position
    docContext: DocumentContext
    hints: {
        maxChars: number
        maxMs: number
    }
    abortSignal?: AbortSignal
}

```

This is very similar to how the graph retrievers already worked but it now extends across all retrieval types (including jaccard similiarty).

Note: This PR does not have any behavioral change yet and is just moving a bunch of code around and editing interfaces. In the next steps, we will use this new structure to:

-  Implement proper mixing
- Call all retrievers in parallel
- Use the new structure for more unified logging

## Test plan

Ensured that local and lsp light context still works:

Screenshot 2023-11-08 at 18 19 45
Screenshot 2023-11-08 at 18 17 51

I was not able to run bfg context though (neither before nor after the change) as it always returns zero results for me 🤔 the test pass though.


vscode/src/graph/bfg/BfgContextFetcher.ts
```diff
@@ -121,6 +120,16 @@ export class BfgContextFetcher implements GraphContextFetcher {
...
 
+    public isSupportedForLanguageId(languageId: string): boolean {
+        switch (languageId) {
+            case 'typescript':
+            case 'typescriptreact':
+                return true
+            default:
+                return false
+        }
+    }
```
philipp-spiess, line 131: @olafurpg Where can I get the proper list? 👀

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -105,15 +103,14 @@ const suggestedCompletionItemIDs = new LRUCache<CompletionItemID, AutocompleteIt
...
-    history: DocumentHistory
     statusBar: CodyStatusBar
-    getCodebaseContext: () => CodebaseContext
-    graphContextFetcher?: GraphContextFetcher | null
     tracer?: ProvideInlineCompletionItemsTracer | null
-    contextFetcher?: (options: GetContextOptions) => Promise<GetContextResult>
     triggerNotice: ((notice: { key: string }) => void) | null
     isRunningInsideAgent?: boolean
 
+    contextStrategy: 'lsp-light' | 'bfg' | 'jaccard-similarity' | 'none'
```
valerybugakov, line 111: We have this union type defined in two places. Thoughts on reusing it?

vscode/src/completions/context/context-mixer.ts
```diff
@@ -0,0 +1,169 @@
...
+        const start = performance.now()
+
+        const localMatches =
+            (await this.localRetriever?.retrieve({
+                ...options,
+                hints: {
+                    maxChars: options.maxChars,
+                    maxMs: 150,
+                },
+            })) ?? []
```
valerybugakov, line 76: Redundant because we always return an array?
```suggestion
        const localMatches = await this.localRetriever?.retrieve({
            ...options,
            hints: {
                maxChars: options.maxChars,
                maxMs: 150,
            },
        })
```

vscode/src/completions/context/context-mixer.ts
```diff
@@ -0,0 +1,169 @@
...
+         * Iterate over matches and add them to the context.
+         * Discard editor matches for files with embedding matches.
+         */
+        const usedFilenames = new Set<string>()
+        const context: ContextSnippet[] = []
+        let totalChars = 0
+        function addMatch(match: ContextSnippet): boolean {
+            // TODO(@philipp-spiess): We should de-dupe on the snippet range and not
+            // the file name to allow for more than one snippet of the same file
+            if (usedFilenames.has(match.fileName)) {
```
valerybugakov, line 88: Does this match the current implementation? There are a lot of cases where relevant context snippets should come from the same file.

valerybugakov (approved):
Great clean up! Thank you, @philipp-spiess 🚀

vscode/src/completions/context/context-mixer.ts
philipp-spiess: `localRetriever` can be undefined if the context stragey is `off` (for tests) so that's why i added the `?? []`

vscode/src/completions/context/context-mixer.ts
philipp-spiess: Yeah I did not chance this code at all yet. We currently only have one snippet that we extract per each file (i.e. only the highest ranked window of a file is extracted). We should look into extending this maybe but the logic is going to to be a bit complex since we need to remove potential windows that have overlaps and rerank 🤔

PR #1688, opened 2023-11-08
Agent: add `evaluate-autocomplete` subcommand

olafurpg:
Previously, I didn't have a good way to test BFG alongside Cody. I just
manually started a local build of the extension, waited for indexing to
finish and triggered autocomplete. This PR fixes this problem by making
it easy to run Cody autocomplete in batch mode against any codebase by
adding a new `evaluate-autocomplete` subcommand. This new command made
it much easier for me to debug BFG and implement critical fixes. It also
makes it much easier to compare the difference between BFG and non-BFG
context.

## Test plan

```
pnpm install
pnpm build
cd agent
pnpm agent evaluate-autocomplete --workspace ~/dev/sourcegraph/bfg-private --queries ~/dev/sourcegraph/bfg-private/crates/bfg/queries/ --include-pattern crates/bfg/snapshots/input/typescript/main.ts --test-count 1 --snapshot-directory ~/dev/snapshots/bfg-suite
cat ~/dev/snapshots/bfg-suite/crates/bfg/snapshots/input/typescript/main.ts

...
  import { sayHello, arrayHello, objectHello, optionalHello } from "./pog";

  sayHello();
  arrayHello();
  objectHello();
//           ^^ reference { cool: { val: 1, otherVal: "hello" } }
  optionalHello();
```

Notice that the `^^ reference` line shows that Cody autocomplete picked up the graph context from BFG.

### DIFF
--- a/agent/src/cli/evaluate-autocomplete/Queries.ts
+++ b/agent/src/cli/evaluate-autocomplete/Queries.ts
@@ -1,0 +1,79 @@
+import * as fspromises from 'fs/promises'
+import * as path from 'path'
+
+import Parser, { Query } from 'web-tree-sitter'
+
+import { SupportedLanguage } from '../../../../vscode/src/completions/tree-sitter/grammars'
+
+export type QueryName = 'context'
+
+/**
+ * Queries manages compilation of tree-sitter queries from a directory layout.
+ *
+ * Queries are written in a file structure like this: `LANGUAGE/QUERY_NAME.scm`
+ *
+ * This class caches compilation of queries so that we only read each query once from disk.
+ */
+export class Queries {
+    private cache: CompiledQuery[] = []
+    constructor(private queriesDirectory: string) {}
+    public async loadQuery(parser: Parser, language: SupportedLanguage, name: QueryName): Promise<Query | undefined> {
+        const fromCache = this.cache.find(compiled => compiled.language === language && compiled.queryName === name)
+        if (fromCache) {
+            return fromCache.compiledQuery
+        }
+        return this.compileQuery(parser, language, name)
+    }
+
+    private async compileQuery(
+        parser: Parser,
+        language: SupportedLanguage,
+        name: QueryName
+    ): Promise<Query | undefined> {
+        const languages = [language, ...(grammarInheritance[language] ?? [])]
+        const queryStrings: string[] = []
+        for (const queryLanguage of languages) {
+            const queryPath = path.join(this.queriesDirectory, queryLanguage, `${name}.scm`)
+            try {
+                const stat = await fspromises.stat(queryPath)
+                if (!stat.isFile()) {
+                    continue
+                }
+            } catch {
+                continue
+            }
+            const queryString = await fspromises.readFile(queryPath)
+            queryStrings.push(queryString.toString())
+        }
+        const uncompiled: UncompiledQuery = {
+            language,
+            queryName: name,
+            queryString: queryStrings.join('\n\n'),
+        }
+        const compiled = compileQuery(uncompiled, parser)
+        this.cache.push(compiled)
+        return compiled.compiledQuery
+    }
+}
+
+interface UncompiledQuery {
+    language: SupportedLanguage
+    queryName: QueryName
+    queryString: string
+}
+interface CompiledQuery extends UncompiledQuery {
+    compiledQuery: Query
+}
+
+function compileQuery(query: UncompiledQuery, parser: Parser): CompiledQuery {
+    return {
+        ...query,
+        compiledQuery: parser.getLanguage().query(query.queryString),
+    }
+}
+
+const grammarInheritance: Partial<Record<SupportedLanguage, SupportedLanguage[]>> = {
+    [SupportedLanguage.TypeScript]: [SupportedLanguage.JavaScript],
+    [SupportedLanguage.JSX]: [SupportedLanguage.JavaScript],
+    [SupportedLanguage.TSX]: [SupportedLanguage.TypeScript, SupportedLanguage.JavaScript],
+}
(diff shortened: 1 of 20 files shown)


olafurpg:
Tagged you @valerybugakov to double-check the changes inside `vscode/` that are not directly BFG-related

agent/src/cli/evaluate-autocomplete/Queries.ts
```diff
@@ -0,0 +1,79 @@
...
+ */
+export class Queries {
+    private cache: CompiledQuery[] = []
+    constructor(private queriesDirectory: string) {}
+    public async loadQuery(parser: Parser, language: SupportedLanguage, name: QueryName): Promise<Query | undefined> {
+        const fromCache = this.cache.find(compiled => compiled.language === language && compiled.queryName === name)
+        if (fromCache) {
+            return fromCache.compiledQuery
+        }
+        return this.compileQuery(parser, language, name)
```
dominiccooney, line 25: `?.` and short-circuit evaluation is more succinct and idiomatic.

```suggestion
        const cachedQuery = this.cache.find(compiled => compiled.language === language && compiled.queryName === name)?.compiledQuery
        return cachedQuery || this.compileQuery(parser, language, name)
```

agent/src/cli/evaluate-autocomplete/Queries.ts
```diff
@@ -0,0 +1,79 @@
...
+                const stat = await fspromises.stat(queryPath)
+                if (!stat.isFile()) {
+                    continue
+                }
+            } catch {
+                continue
+            }
+            const queryString = await fspromises.readFile(queryPath)
+            queryStrings.push(queryString.toString())
+        }
```
dominiccooney, line 47: Not a big deal but you could fork and join this IO with Promise.all, something like:

```suggestion
        const queryStrings = Promise.all(languages.map(async (language: SupportedLanguage) => {
            const queryPath = path.join(this.queriesDirectory, queryLanguage, `${name}.scm`)
            try {
                const stat = await fspromises.stat(queryPath)
                if (!stat.isFile()) {
                    return
                }
            } catch {
                return
            }
            return (await fspromises.readFile(queryPath)).toString()
        })).filter(id => id);
```

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
```diff
@@ -145,4 +107,51 @@ export class BfgRetriever implements ContextRetriever {
...
+            )
+        })
+    }
+
+    private async doLoadBFG(reject: (reason?: any) => void): Promise<MessageHandler> {
+        const bfg = new MessageHandler()
+        const codyrpc = await downloadBfg(this.context)
+        if (!codyrpc) {
+            throw new Error(
+                'Failed to download BFG binary. To fix this problem, set the "cody.experimental.bfg.path" configuration to the path of your BFG binary'
```
dominiccooney, line 129: You don't have to fix this now, but @csells does not want the name "BFG" to be something users see or have to think about. Putting it in error messages increases the likelihood someone echoes that in a UI toast.

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
```diff
@@ -145,4 +107,51 @@ export class BfgRetriever implements ContextRetriever {
...
+            )
+        }
+        const isVerboseDebug = vscode.workspace.getConfiguration().get<boolean>('cody.debug.verbose', false)
+        const child = child_process.spawn(codyrpc, { stdio: 'pipe', env: { VERBOSE_DEBUG: `${isVerboseDebug}` } })
+        child.stderr.on('data', chunk => {
+            if (isTesting) {
```
dominiccooney, line 135: Realize you're just moving this around, but I thought we handled redirecting logs when running in test mode deeper in the logging stack and we shouldn't have to do it here.

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
```diff
@@ -145,4 +107,51 @@ export class BfgRetriever implements ContextRetriever {
...
+                console.log(chunk.toString())
+            } else {
+                logDebug('BFG', 'stderr output', chunk.toString())
+            }
+        })
+        child.on('disconnect', () => reject())
+        child.on('close', () => reject())
+        child.on('error', error => reject(error))
```
dominiccooney, line 143: We can iterate on this later, but it would be great if we can emanate some death reasons from the subprocess and start to record these.

dominiccooney (approved):
Looks great, some suggestions inline.

agent/src/cli/evaluate-autocomplete/Queries.ts
olafurpg: Either way works for me. I like the current solution because I find it easy to read and easier to extend. Operators like `?` tend to become unwieldy when you have more complex expressions or want to do things with intermediary values. The current structure makes it easy to add new statements or extract other properties.

agent/src/cli/evaluate-autocomplete/Queries.ts
olafurpg: This code shouldn't be a performance bottle-neck so I think it's fine to keep it as is. I'm generally not a fan of having large expressions and rely on deep type inference. For example, I think the code can be easier to work with when you already declare that `queryStrings` has type `string[]` because the typechecker can give more detailed error messages when you push another type. It's all subjective in the end

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
olafurpg: Agree. This diff is just from moving code around. I opened an issue track hiding "BFG" to end users https://github.com/orgs/sourcegraph/projects/383/views/1?pane=issue&itemId=44463865

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
olafurpg: Agreed

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
olafurpg: This guard is no longer be needed, pushed a commit to remove it

PR #1690, opened 2023-11-08
fix: handle keyboard shortcuts with Ctrl key

abeatrix:
fix: prevent default and stop propagation on certain key combos

Prevent default behavior on `ctrl + non-allowlisted keys`.. 
Also check for alt + vscode cody shortcuts and prevent default without stopping propagation.

This fixes issues with certain key combinations triggering unintended characters in input box.

- Add ctrlKeysAllowList Set to check pressed keys.
- Check ctrlKey or AltGraph modifier and that key is not in allow list.
- If so, prevent default browser shortcut and stop event propagation.

## Test plan

Pressing CtrlKey with any of the keys in the allowed list should not add character to the chat inbox.

#### Before

Press `alt`+`/` when chat box is focused. This shortcut will open a new Cody panel, but it will also add a symbol to your chat input box:

image

#### After

Press `alt`+`/` when chat box is focused will not change your current input in the input box.


philipp-spiess (approved):
Nice!

PR #1701, opened 2023-11-09
Autocomplete: Shift range when last candidate match is calculated

philipp-spiess:
This fixes an issue where the insertion range was invalid when typing as suggested on a completion that has a same line suffix. 

The issue can be reproduced like this:

https://github.com/sourcegraph/cody/assets/458591/b40693ff-715c-4a2f-9893-76dcc37735c6

When typing new characters and reusing the last candidate, we have to shift the completion insertion range to match the new location.

This is not an issue with the indentation change logic when I tried it.

## Test plan

https://github.com/sourcegraph/cody/assets/458591/044f0f36-5738-4f7f-88d8-39b2b6f8ca3d


valerybugakov (approved):
Great find!

PR #1705, opened 2023-11-09
Autocomplete: Implement top-k retrieval mixing and make section history a retriever

philipp-spiess:
Some more retrieval improvements!

- Section history now implements the retrieval strategy and is now longer called transitively by the lsp light retriever. This will allow us to mix it with other results soon.
- The context mixer now does top-k mixing (not sure if this is the right terminology?): If more than one retrieval strategy is used, it will pick the top most results from every strategy and then follow on with the second rank, etc. For now this is only used in the `lsp-light` strategy that isn't used but the idea is that it will contain the highest ranked  lsp light result followed by the highest ranked section history and then the second highest ranked lsp light etc.
  - This is only temporary and we'll add more advanced strategies shortly. Again this only affects `lsp-light` strategy which is not rolled out right now.
- Overhauls the `contextSummary` logging to include a ton more information like:
  - Execution time of each retrieval strategy
  - Number of retrieved items and number of items included in the suggested context set
  - A bitmap that can be used to know which types of snippets are included, more on this below.
  - Total number of characters suggested for the request context (this may NOT be the exact prompt length but it can be used as a proxy for it)

## Position bitmap

Okay maybe this is over engineering but I do think we need to understand how the top snippets are composed when we start to experiment with snippet mixing. To do that, we can use bitmaps and setting the bit according to the position in the result set for every type to express the order.

Here's an example of retrieval stats for an `lsp-light` request that contains a total of 5 items.

```json
{
  "lsp-light":{
    "suggestedItems":4,
    "positionBitmap":29,
    "retrievedItems":4,
    "duration":123
  },
  "section-history":{
    "suggestedItems":1,
    "positionBitmap":2,
    "retrievedItems":1,
    "duration":123
  }
}
```
 
In binary, the `positionBitmap` for each strategy look like this:

- lsp-light: `11101`
- section-history: `00010`

Since the result is read from the lowest bit to the highest, we can see that the first item (right most bit) is from the lsp light set, the second entry from the section history and the remaining three from the lsp light set again.

### Jaccard similliarty still works

Screenshot 2023-11-09 at 17 08 40

### lsp light works

Screenshot 2023-11-09 at 18 12 56

### DIFF
--- a/vscode/src/completions/context/context-mixer.ts
+++ b/vscode/src/completions/context/context-mixer.ts
@@ -1,12 +1,12 @@
 import * as vscode from 'vscode'
 
-import { logDebug } from '../../log'
 import { DocumentContext } from '../get-current-doc-context'
 import { ContextRetriever, ContextSnippet } from '../types'
 
 import { BfgRetriever } from './retrievers/bfg/bfg-retriever'
 import { JaccardSimilarityRetriever } from './retrievers/jaccard-similarity/jaccard-similarity-retriever'
-import { LspLightGraphCache } from './retrievers/lsp-light/lsp-light-graph-cache'
+import { LspLightRetriever } from './retrievers/lsp-light/lsp-light-retriever'
+import { SectionHistoryRetriever } from './retrievers/section-history/section-history-retriever'
 
 export interface GetContextOptions {
     document: vscode.TextDocument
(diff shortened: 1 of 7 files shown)


vscode/src/completions/context/context-mixer.ts
```diff
@@ -16,13 +16,33 @@ export interface GetContextOptions {
...
+    retrieverStats: {
+        [identifier: string]: {
+            /** Number of items that are ended up being suggested to be used by the context mixer */
+            suggestedItems: number
+            /** Number of total snippets */
+            retrievedItems: number
+            /** Duration of the individual retriever */
+            duration: number
+            /**
+             * A bitmap that indicates at which position in the result set an entry from the given
```
valerybugakov, line 36: What are the reasons for choosing the bitmap here? The are compact and good for performance reasons. Do we want to prioritize these features here instead of readability?

philipp-spiess: Hm I was thinking of different approaches but it very easily gets very very messy e.g.:

```
contextResultsOrder: ['bfg','jaccard-similiarty','bfg','bfg']
```

does not feel easier to parse for me. And I know we already have some scaling issues with our analytics setup especially for the non dotcom instance where we have too much write into the analytics table so i thought this would be a good compromise. 

If you think this is to complicated I think we should remove it completely until we need it. I just thought this is an elegant and simple solution so why not add it right away 🙈

philipp-spiess: > could you share your thoughts on how the positionBitmap field will be used to analyze experiment results, so that it's easier to understand the format trade-offs.

My idea was that we might want to lok into results that have a specific context source as the top rated result and then we can use simple XOR to get all of these results. Not sure if this will immediately be useful so I can definitely rm if you think it won't.

valerybugakov: > My idea was that we might want to look into results that have a specific context source as the top rated result

It makes sense. My only worry is the readability of the results when we query them in Looker or BigQuery. 

If having explicit labels `contextResultsOrder: ['bfg','jaccard-similiarty','bfg','bfg']`  is too messy and we are worried about scalability issues, maybe we can use strings representing binary numbers? It's the same as your proposal but removes the need to do XOR in your head when looking at the values. E.g. `contextResultsOrder: '00010'`. 

I'm open to keeping the existing solution. I just wanted to talk about alternatives and the end goal.

philipp-spiess: Oh, I see. I haven't thought of this as something to be super useful when manually looking at the results (although you can easily see if it's odd or even and thus indicate what is the the first result). I'd say we add this for it's compactness and cheap costs but if we end up not using it or struggling to use it we can revisit (ultimately I don't think it's going to be super impactful compared to the other metrics added)

PR #1706, opened 2023-11-09
refactor: update context files display widget

abeatrix:
MOVED FROM https://github.com/sourcegraph/cody/pull/1649
CLOSE: https://github.com/sourcegraph/cody/issues/1525

This PR replaces the current Context File widget with a new UI widget for displaying context files used for a chat question as shown in @toolmantim 's figma design:

image

Also add a new `source` field to Context File for identifying purpose, which only supports 'embeddings' currently.

Updated all the context message building functions to include context source.

Added 'keyword' as source to results returned by keyword search.

Added 'filename' as source to results returned by filename search.

Display line numbers for embedding results.

Related tests are also updated.

## Test plan

- [ ] Ask Cody a question, e.g. `What's the entry file for the cody vs code extension?`

### Before

image

### After 

- [ ] Only context files from non-editor and non-user source is included in the `enhanced context` widget:

Screenshot 2023-11-09 at 12 53 44 PM

- [ ] Run a command, e.g. Explain, that use editor context and not enhanced context (e.g. embeddings or local keyword search), the editor context will not show up as enhanced context (displayed with sprinkles in the front):

image

Screenshot 2023-11-09 at 12 53 33 PM

Issue #1524, opened 2023-10-27
VSCE: New Chat - Enhance Context selector


abeatrix:
<img width="811" alt="image" src="https://github.com/sourcegraph/cody/assets/68532117/9705d045-7c7a-48dd-8946-3abfc9cf344f">

<img width="582" alt="image" src="https://github.com/sourcegraph/cody/assets/68532117/5fa4f9c1-2c27-4dd3-8236-edda730b35b8">

dominiccooney:
Thanks for starting on this @abeatrix.

Burn-down list:

- [ ] [This design ("v4 enhanced context states")](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=209-43481&mode=design&t=o1Exs32m5DuBy9n7-0)
- [ ] Folder tree view w/ sources
- [ ] Graph, search, embeddings sources
- [ ] "Indexing" states
- [ ] Enable Embeddings
- [ ] Enterprise UX for missing embeddings "No repository matching ..."
- [ ] a11y: Keyboard needs work. Once it is open, you can't dismiss it; activating it checks and unchecks the context.
- [ ] Pixels: Popup needs rounded corners.
- [ ] Pixels: Add the "Learn More" link.
- [ ] Pixels: Box shadow is a bright focused color, should be a receding color.
- [ ] Pixels: Check box—widget needs a border and fill; check is a dim color
- [ ] Pixels: "Enhanced Context" heading, weight.
- [ ] Stretch: Add Repository

abeatrix:
> Thanks for starting on this @abeatrix.
> 
> 
> 
> Burn-down list:
> 
> 
> 
> - [ ] [This design ("v4 enhanced context states")](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=209-43481&mode=design&t=o1Exs32m5DuBy9n7-0)
> 
> - [ ] Folder tree view w/ sources
> 
> - [ ] Graph, search, embeddings sources
> 
> - [ ] "Indexing" states
> 
> - [ ] Enable Embeddings
> 
> - [ ] Enterprise UX for missing embeddings "No repository matching ..."
> 
[...]

PR #1707, opened 2023-11-10
feat: add enhance context toggler for chat

abeatrix:
Continue from https://github.com/sourcegraph/cody/pull/1706

This PR added a temporary button (green check mark when enabled) above the chat box to toggle between enabling and disabling enhanced context for the first question. The complete interface will be implemented by Dom in https://github.com/sourcegraph/cody/issues/1524

image

### Summary

- Updated context fetching logic in the chat-question recipe:
  - Only retrieve enhanced context for the first question, unless it is disabled by the user (via the new chat UI button) 
  - Update editor context regexps to match `entire file`
  - If users did not add any additional context via @, we will then check if editor context is required

- In the execute recipe method, we now pass if user has disabled enhanced context or not, in addition to `firstInteraction`

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/e3d9a3db-9db0-4957-8014-335f876739da

The expected behaviors are:

- [ ] User can click on the icon to toggle between enabling/disabling enhanced context for their first question
- [ ] After the first question, users will not be able to add enhanced context. However, they can add additional context using the @ command.
- [ ] User is required to start a new chat to add enhanced context for their first question.

When enhanced context is enabled vs when enhanced context is disabled for the question:

image

# TODO

- [x] need to fix tests


lib/shared/src/prompt/templates.ts
```diff
@@ -6,12 +6,12 @@ import { ActiveTextEditorDiagnostic, ActiveTextEditorSelection } from '../editor
 import { MAX_RECIPE_INPUT_TOKENS } from './constants'
 import { truncateText, truncateTextStart } from './truncation'
 
-const CODE_CONTEXT_TEMPLATE = `Use following code snippet from file \`{filePath}\`:
+const CODE_CONTEXT_TEMPLATE = `Review context from codebase file \`{filePath}\`:
```
abeatrix, line 9: Before:

image

After:
image

abeatrix:
Move to https://github.com/sourcegraph/cody/pull/1738

PR #1708, opened 2023-11-10
Clean up login page styles

toolmantim:
* Moves the Enterprise sign in link up under the social login links
* Makes the button logos monochrome
* Updates the window sizing and button width behaviour to match standard VS Code behaviour

| Before | After |
| - | - |
| Screenshot 2023-11-10 at 2 37 12 pm | Screenshot 2023-11-10 at 2 36 59 pm |

## Test plan

- Loaded extension
- Previewed widths and clicked buttons


toolmantim:
@dominiccooney I also tried fixing the "Sourcegraph.com" that erroneously appears in the "Recent" list when you click "Sign In To Enterprise Instance" but failed (I think perhaps its coming from https://github.com/sourcegraph/cody/blob/main/vscode/src/services/AuthProvider.ts#L377)

dominiccooney (approved):
Monochrome sign in logos look great!

dominiccooney:
> I also tried fixing the "Sourcegraph.com" that erroneously appears in the "Recent" list when you click "Sign In To Enterprise Instance"

I wrote that FIXME and it is apt, the code is racy. We should add a state machine here so we don't mix state from one sign in to another. Once we remove App it gets a bit simpler and we can lean into that to fix these.

Issue #1576, opened 2023-11-01
PRD: Ability to upgrade your account from VS Code


toolmantim:
Description:

As a new user who's signed up to Cody, I want to understand my usage limits and be able to upgrade my account. And once I've upgraded, I want that to be reflected in Cody so I can have confidence the upgrade completed, and to be able to take advantage of any features or limits available on my new plan.

The main interactions that have limits and will trigger a CTA for upgrading:

- Autocomplete
- Commands
- Chat
- Embeddings

Objectives:

- Create an intuitive, reliable, and motivating upgrade flow from the free to pro Cody plans
- Setup measuring and monitoring for future iterations

Acceptance Criteria:

```[tasklist]
### Acceptance Criteria:
- [x] Signing up shows correct plan type in nav
- [x] Signing back in shows correct plan type in nav
- [x] All interactions that can trigger rate limits gracefully handle them and provide CTA to upgrade
- [x] UI updates to reflect plan changes without manual reload
- [x] New navigation layout has links to open web URLs
- [x] Copy is readable, clear and concise
```
Designs

- [Figma](https://www.figma.com/file/f7bzmpYrwalyYOQJcWLRTP/Untitled?type=design&node-id=0-3&mode=design)

```[tasklist]
### Design Tasks
- [x] New nav items
- [x] Updated autocomplete notice (update to #634)
- [x] New chat error notice notice
- [x] Update Embeddings in Enhanced Context popover with limits and CTA (update to #1377)
```

```[tasklist]
### Engineering Tasks
- [x] Add "Upgrade" menu item for free plan
- [x] Add "Usage" menu item
- [x] Update autocomplete rate limit notice quickpick
- [ ] https://github.com/sourcegraph/cody/issues/2123
- [x] Add Chat rate limit handling to show a nicer message for rate limits
- [x] Post-upgrade: Update account plan after account upgraded
- [x] Post-upgrade: Remove "Upgrade" sidebar item
```

### Follow-up Tasks:

```[tasklist]
### Follow-up
- [ ] https://github.com/sourcegraph/cody/issues/2517
- [ ] Track data for all this
```

toolmantim:
The two new nav items to add at the top of the new "Settings & Support" treeview:

![image](https://github.com/sourcegraph/cody/assets/153/0ca2c7c3-6d1f-4e69-b6bd-6272027c2de6)

The first is only shown when you're on the free plan. The second will always be shown.

kalanchan:
[...]

PR #1711, opened 2023-11-10
Add new sidebar links

DanTup:
image

## Test plan

- [x] automated tests that sidebar only shows links based on AuthStatus matching the criteria below
- [x] Check usage link shows up when `cody_pro_dec_ga` flag is enabled
- [x] Check usage link does not show up when `cody_pro_dec_ga` flag is not enabled
- [x] Check Upgrade link shows up when `cody_pro_dec_ga` is enabled and `codyProEnabled=false` in GraphQL response when using DotCom
- [x] Check Upgrade link does not show up when `cody_pro_dec_ga` is enabled and `codyProEnabled=true` in GraphQL response when using DotCom
- [x] Check Upgrade link does not show up when `cody_pro_dec_ga` is disabled and `codyProEnabled=false` in GraphQL response when using DotCom
- [x] Check Upgrade link does not show up when `cody_pro_dec_ga` is enabled and `codyProEnabled=false` in GraphQL response when not using DotCom


DanTup:
@toolmantim I presume these are both just opening external pages? I've added placeholder URIs for now. The upgrade link needs to be conditional, but I don't know what the data/structure for accounts looks like yet (or whether it's part of an existing auth response or new APIs).

DanTup:
@toolmantim @kalanchan I pushed some changes to add conditions to the buttons. It needs tests, but I wanted to check if this is what you had in mind first.

There are some TODOs with questions - mainly around whether it's safe to always query for `codyProEnabled` (so this can be done along with an existing query for verified email). And ofc, the URLs for the buttons are just placeholders.

toolmantim:
Looks good @DanTup!

> I presume these are both just opening external pages? I've added placeholder URIs for now. The upgrade link needs to be conditional

I believe so!

> I wanted to check if this is what you had in mind first

From a design POV, this looks good to me.

DanTup:
@philipp-spiess @abeatrix I have some questions in TODOs in the code that I'm not sure who is best to answer (mostly around graphql to check for Pro (reusing an existing verified email query and whether it's safe to assume the field exists).

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -58,7 +58,9 @@ interface CurrentUserIdResponse {
 }
 
 interface CurrentUserIdHasVerifiedEmailResponse {
-    currentUser: { id: string; hasVerifiedEmail: boolean } | null
+    // TODO(dantup): Is it reasonable to add this flag here to avoid two requests?
+    //   Can we assume codyProEnabled is boolean or can it be missing?
```
abeatrix, line 62: I think what you have is fine because `getCurrentUserIdAndVerifiedEmail` is only run when user is connected to the DotCom instance, which should have this field available , I think :P

lib/shared/src/sourcegraph-api/graphql/queries.ts
```diff
@@ -28,11 +28,14 @@ query SiteGraphQLFields {
     }
 }`
 
+// TODO(dantup): Is it save to always include this against all instances, or might there be servers
+//   where this doesn't exist (and will throw)?
```
abeatrix, line 32: same as above - this should be fine as we only run this query [if the user is connected to DotCom](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/AuthProvider.ts?L212)

vscode/src/services/AuthProvider.ts
```diff
@@ -205,10 +206,14 @@ export class AuthProvider {
...
                 hasVerifiedEmail,
                 enabled,
+                /* userCanUpgrade: */ false,
                 version,
                 configOverwrites
             )
         }
+
+        // TODO(dantup): Rename this and/or split into another query.
+        //   Can we use a query with codyProStatus against all instances (eg. non-dotcom)?
```
abeatrix, line 216: > Can we use a query with codyProStatus against all instances (eg. non-dotcom)?

Probably not, because this is only available to instances that are on the latest version I presume, and most of our customers are always a few versions behind

DanTup: Ah, that's good to know. I don't think it matters here (this is `isDotcomOrLocalApp` and I presume `LocalApp` is for tests and always current?), but something to be aware of!

DanTup: Actually, I'm less sure "LocalApp" is just for tests now, because I see the same local URL (`http://localhost:3080/`) in some other code (as an `embeddingsEndpoint`). I'll change this to have a separate query for DotCom (with CodyPro) and leave LocalApp as-is - but let me know if this is incorrect and we can just have a single query for both.

abeatrix: > LocalApp is for tests and always current

It is not, but I believe we are also sunsetting Cody App by GA, and if that's true, you should be good here! 

@kalanchan can you confirm if we will support App by GA?

vscode/src/services/TreeViewProvider.ts
```diff
@@ -28,23 +32,38 @@ export class TreeViewProvider implements vscode.TreeDataProvider<vscode.TreeItem
...
+    public async updateTree(treeItems: CodySidebarTreeItem[]): Promise<void> {
+        // TODO(dantup): This method can be made not-async again when we don't need to call evaluateFeatureFlag
         const updatedTree: vscode.TreeItem[] = []
-        treeItems.forEach(item => {
+        for (const item of treeItems) {
+            if (item.requireFeature && !(await featureFlagProvider.evaluateFeatureFlag(item.requireFeature))) {
+                continue
+            }
+
+            if (item.requireUpgradeAvailable && !(this.authStatus?.userCanUpgrade ?? false)) {
```
abeatrix, line 46: `this.authStatus?.userCanUpgrade` always returns `undefined` for me but i thought it's set to false by default, is this expected?

abeatrix (commented):
Nice! I removed the tree items checks to force it to work without the correct feature flags and it looks good to me! The links are working as expected 👍 

image

Left a comment in line but otherwise, LGTM!

vscode/src/services/TreeViewProvider.ts
DanTup: The constructor of the tree renders the tree before the authStatus is synced, so I think this loop should actually run twice initially.

It's a little clunky right now because the feature flag check is async so much of this code becomes async. When we can remove the feature flag I think it'll be much simpler.

I pushed some tweaks along with tests that the right things are appearing in the tree - please take a look.

The only thing missing is tests that the auth is working correctly, however I'm not sure how I can test them (really I need a way of mocking the auth and being able to simulate a) dotCom when user is not already upgraded, b) dotCom when user is already upgraded, and c) non-dotCom. I don't know if there's currently a way to do this (nor do I think I have a way of manually testing the non-dotCom route).

DanTup: > `this.authStatus?.userCanUpgrade` always returns `undefined` for me

Actually, there is a weird bug here.. I'm not syncing auth status with the right tree (I'm syncing with the chat tree), so there's a weird race. Working on a fix.

DanTup: Fixed!

I've manually tested all of the conditions I listed above except for the one that requires not being on DotCom - I'm not sure I can manually test that.

abeatrix: Nice, I can connect it to s2 to test the condition you mentioned, will take a look now!

DanTup:
@abeatrix in this PR there are a bunch of places where code had to be made `async` to check the GA feature flag. I have the same in my next PR (adding the upgrade notification to rate limit errors for completion). This feels awkward and can be removed once the GA flag goes away.

I wonder if we should change how it works, fetching the feature flag during auth and caching it on `AuthStatus` so it could be checked synchronously in these places? (I don't know if there's a danger of it becoming stale then though - do feature flags usually change and update without re-authenticating?)

vscode/src/services/TreeViewProvider.test.ts
```diff
@@ -0,0 +1,107 @@
+import { describe, expect, it, vi } from 'vitest'
+import type * as vscode from 'vscode'
```
abeatrix, line 2: Thanks for adding the test 🙇‍♀️

vscode/src/services/TreeViewProvider.ts
```diff
@@ -28,33 +37,51 @@ export class TreeViewProvider implements vscode.TreeDataProvider<vscode.TreeItem
...
+     * Refreshes the visible tree items, filtering out any
+     * that do not meet the required criteria to show.
+     */
+    public async refresh(): Promise<void> {
+        // TODO(dantup): This method can be made not-async again when we don't need to call evaluateFeatureFlag
         const updatedTree: vscode.TreeItem[] = []
-        treeItems.forEach(item => {
+        this.treeNodes = updatedTree // Set this before any awaits so last call here always wins regardless of async scheduling.
+        for (const item of this.treeItems) {
+            if (item.requireFeature && !(await this.featureFlagProvider.evaluateFeatureFlag(item.requireFeature))) {
```
abeatrix, line 57:
```suggestion
            if (this.type === 'support') {
                if (item.requireFeature && !(await this.featureFlagProvider.evaluateFeatureFlag(item.requireFeature))) {
                    continue
                }

                if (item.requireUpgradeAvailable && !(this.authStatus?.userCanUpgrade ?? false)) {
                    console.log(this.authStatus)
                    continue
                }
            }
```

Should we run this for the support sidebar only so the chat sidebar which gets refreshed a lot won't make new requests to the featureFlagProvider on each refresh?

DanTup: I don't think `this.type === 'support'` matters, because the other kinds won't have `item.requireFeature` set, so it won't call `evaluateFeatureFlag`?

We could check for dotcom, but this code was intended to be fairly generic and not specific to things that require dotcom (maybe if we want to do that, we should add a `requiresDotCom` or similar flag to `item`?)

(our of interest, how heavily are feature flags cached? I assumed in most cases this would be immediately returned without any network request)

abeatrix:
> I wonder if we should change how it works, fetching the feature flag during auth and caching it on `AuthStatus` so it could be checked synchronously in these places? (I don't know if there's a danger of it becoming stale then though - do feature flags usually change and update without re-authenticating?)

Good question 🤔 I've never looked into the implementation before, but I am guessing they can change without re-authenticating so that when a feature flag is turned on/off in the backend, it will have immediate effect? 

I think Taras will be the right person to ask since he implemented the provider :D  cc @taras-yemets

taras-yemets:
> I wonder if we should change how it works, fetching the feature flag during auth and caching it on AuthStatus so it could be checked synchronously in these places? (I don't know if there's a danger of it becoming stale then though - do feature flags usually change and update without re-authenticating?)

Once fetched, we store the value in the cache for 1 hour. As far as I remember, the interval was selected arbitrarily. The answer depends on how often the auth happens. If it's frequent enough to evaluate the potentially changed feature flag value, then I'd say we're good to cache it and check synchronously. 
 
https://github.com/sourcegraph/cody/blob/ec008a8814f824313c02037b4a4d93456ca3987b/lib/shared/src/experimentation/FeatureFlagProvider.ts#L56

cc: @abeatrix, @philipp-spiess, @DanTup

DanTup:
@kalanchan can you confirm the feature flag name? I added `cody_pro_dec_ga` in this PR (from https://github.com/sourcegraph/cody/issues/1576#issuecomment-1806466079) but I've also seen `cody-pro` in another doc.

Other than that, I think this is ready if @abeatrix is happy with the responses to her comments above.

lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -23,7 +23,7 @@ export enum FeatureFlag {
     CodyAutocompleteUserLatency = 'cody-autocomplete-user-latency',
     CodyAutocompleteDisableNetworkCache = 'cody-autocomplete-disable-network-cache',
     CodyAutocompleteDisableRecyclingOfPreviousRequests = 'cody-autocomplete-disable-recycling-of-previous-requests',
-
+    CodyProDecGA = 'cody_pro_dec_ga',
```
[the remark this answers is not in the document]
kalanchan: actually - Naman informed me that we can control the feature flag by userIDs via the token, this implementation works as long as we whitelist a group of userIDs for testing

kalanchan:
> @kalanchan can you confirm the feature flag name? I added `cody_pro_dec_ga` in this PR (from [#1576 (comment)](https://github.com/sourcegraph/cody/issues/1576#issuecomment-1806466079)) but I've also seen `cody-pro` in another doc.
> 
> Other than that, I think this is ready if @abeatrix is happy with the responses to her comments above.

Sorry for the confusion! `cody_pro_dec_ga` was the original flag, but I've confirmed with Naman that its going to be `cody-pro` now. Let's reflect that in the PR

lib/shared/src/experimentation/FeatureFlagProvider.ts
kalanchan, line 26:
```suggestion
    CodyProDecGA = 'cody-pro',
```

DanTup: I pushed a change that changes both the string and the enum member to avoid any potential confusion. Thanks!

DanTup:
I pushed a change to update the flag and all the unit test bots are failing. I reverted the change and the same is happening - so I don't think it's related to that change.

It looks authentication related - and the error also seems to include the _live_ dotcom endpoint (it also says "unknown test" as if the error is occurring outside of the tests).

```
<empty line>
Errored ➞ POST https://sourcegraph.com/.api/graphql?CurrentUser _PollyError: [Polly] [adapter:node-http] Recording for the following request is not found and `recordIfMissing` is `false`.
{
  "url": "https://sourcegraph.com/.api/graphql?CurrentUser",
  "method": "POST",
  "headers": {
```

I don't understand why this is happening with the exact same code that was green yesterday, and I can't repro it from my local machine. Any ideas? 🤔

DanTup:
oh nm, I realised the bots are merging latest `main` with the PR before running. After merging main I'm seeing the same failures locally and will investigate!

DanTup:
Ok, I _think_ the failures might be related to the new agent tests from #1872. Based on @olafurpg's comments, I believe the tests are run locally and the HTTP requests and responses captured to be replayed during the CI runs.

This PR changes the graphql queries (to include the codyPro flag when fetching user information from dotCom) so there's no recorded responses.

@olafurpg I'm reading the instructions and it seems I need to set `SRC_ACCESS_TOKEN` to an access token to re-record these. However it's not clear to me what that token is (and whether it's something I will have access to?).

olafurpg:
@DanTup you create an access token for your sourcegraph.com account by opening this page here https://sourcegraph.com/user/settings/tokens/new

You can verify that you are authenticated with the `src` cli tool (installable via npm package `@sourcegraph/src`)

```
❯ src login

❌ Problem: No access token is configured.

🛠  To fix: Create an access token at https://sourcegraph.com/user/settings/tokens, then set the following environment variables:

   SRC_ENDPOINT=https://sourcegraph.com
   SRC_ACCESS_TOKEN=(the access token you just created)

   To verify that it's working, run this command again.
```

olafurpg:
BTW, I spent a while trying to see if I could automatically print out the instructions to update the HTTP recordings when the tests fails. I investigated custom vitest reporters but didn't make progress (because I think we're using an old vitest version 🤔 ). Either way, I'd love to see how we can make this process as smooth as possible.

DanTup:
@olafurpg thanks! I wasn't able to install the npm package (see https://github.com/sourcegraph/src-cli/issues/1050) but I set those env vars and commented out the call to `src login` when recording the tests and it seems to have worked, the bots are now green. Thanks! :)

DanTup:
I spoke too soon - it fixed most of the bots, but the Ubuntu one remained red. @olafurpg should there be differences across platforms in the HTTP requests?

```
Errored ➞ GET http://localhost:3080/__version _PollyError: [Polly] [adapter:node-http] Recording for the following request is not found and `recordIfMissing` is `false`.
{
  "url": "http://localhost:3080/__version",
  "method": "GET",
  "headers": {
    "accept": [
      "*/*"
    ],
    "user-agent": [
      "node-fetch/1.0 (+https://github.com/bitinn/node-fetch)"
    ],
```

olafurpg:
@DanTup the agent tests are passing in ubuntu 16. The root failure seems unrelated and might be a flake so re-ran the failed tests

```
Failed to terminate worker while running /home/runner/work/cody/cody/lib/shared/src/common/markdown/markdown.test.ts.

```

DanTup:
Ah, thanks! I had re-run once and it failed the same way, but I guess it was just coincidence :)

olafurpg:
I want to fix the tests so that we don't get failing HTTP requests when running the agent tests. There is some non-determinism in how we send HTTP requests causing the pollyjs errors even if they don't affect the behavior of the agent tests.

vscode/src/services/TreeViewProvider.ts
abeatrix: Yes that looks good to me. Sorry for the late reply!

abeatrix (commented):
What you have looks good to me. Once the bots are green I can give another try just to confirm and then I think it'd be good to merge

DanTup:
@abeatrix was just a flake - I re-ran that bot and it's green now.

abeatrix (approved):
Usage and Upgrade items are showing up correctly (tested by hardcoded the feature flag value) for both dotcom and non-dotcom instances:
image

PR #1713, opened 2023-11-10
agent: add telemetry/recordEvent to replace graphql/logEvent

bobheadxi:
1. Add `telemetry/recordEvent`, which sends events to a telemetry recorder provider backed by a backwards-compatible GraphQL exporter - the same one we use for VSCode, added in #1192. This replaces the current `graphql/logEvent`
2. Deprecate `extensionConfiguration.eventProperties`
3. Clients should now provide `{ extensionConfiguration: { anonymousUserID }` up-front

## Test plan

`pnpm build` and `node dist/index.js` are okay, but might need some help testing this properly with a real extension


agent/src/agent.ts
```diff
@@ -250,6 +270,26 @@ export class Agent extends MessageHandler {
...
+                // HACK: We have no control over what gets sent over JSON RPC,
+                // so we depend on client implementations to deal with this
+                // themselves. When passing the provided event to the
+                // TelemetryRecorder implementation, we forcibly cast the
+                // input into known types so that the recorder will accept it.
+                //
+                // Do not do this elsewhere!
+                event.feature as 'feature',
+                event.action as 'action',
+                event.parameters as TelemetryEventParameters<{ key: number }, BillingProduct, BillingCategory>
```
Strum355, line 285: Is `{key: number}` right here? Afaik this is saying that its an object with a single field called `key` of type `number`, no other fields.

Strum355: On a separate note, are `BillingProduct` and `BillingCategory` meant to be the raw strings `exampleBillingProduct` and `exampleBillingCategory`?

agent/src/agent.ts
```diff
@@ -250,6 +270,26 @@ export class Agent extends MessageHandler {
...
+            this.telemetryRecorderProvider.getRecorder().recordEvent(
+                // HACK: We have no control over what gets sent over JSON RPC,
+                // so we depend on client implementations to deal with this
+                // themselves. When passing the provided event to the
+                // TelemetryRecorder implementation, we forcibly cast the
+                // input into known types so that the recorder will accept it.
+                //
+                // Do not do this elsewhere!
+                event.feature as 'feature',
+                event.action as 'action',
```
Strum355, line 284: Are these also meant to be the raw strings "feature" and "action"?

Strum355: Reading further into how this is done....wow :no_mouth:

Strum355:
thx for the PR! Few questions to help me understand how things should ideally be done going forward:
1. What kind of consistency in feature/action values do we want between vscode and agent-consumers (jetbrains, neovim etc) cc @nathan-downs 
2. How do we keep consistency in the `metadata` values being sent for certain actions between various clients (e.g. currently we send the LLM provider identifier as a raw string, with values incuding "anthropic", "fireworks" and "unstable-openai", if converted numerically how would we have the reported number be consistent?)
3. What events can we reasonably expect to be sent via agent vs the clients
and similar thoughts

nathan-downs:
1) In an ideal world we would have direct parity between vscode and the other extensions, but practically there will be some divergence in functionality and capabilities anyway. The analytics dashboarding and enablement portion requires manual analysis of the available data and context anyway, so the main concern is that data remain consistent (example: name changes and feature deprecation). That said, just try to make things similar enough to existing precedence to make the inevitable refactoring easier.

nathan-downs:
So presumably events like login/out, chat submitted, core commands and the like will mean the same, but I know the vscode folks are experimenting with other metrics like completion acceptance persistence that may not be implemented everywhere, or there may be metrics that are able to be captured in some platforms and not others. We just don't want to be in a situation where the same command means different things on different platforms.

agent/src/agent.ts
nathan-downs: Here's the overview for [billing product category](https://docs.google.com/document/d/1F6yFUVX_DZ_tT76NpGJL6_jlm_THQyVikRRRjNDNsyg/edit#bookmark=id.91bo7wi7u1ca) we had previously designed (predefined enum), but the complete structure for these is still TBD based on what marketing needs are.

re: metadata field only accepting `{key: number}`
Would we be able to open this up to allowing `{key: boolean}` as well since they are functionally equivalent? (I suppose I could otherwise convert them to bit flags, but there'd be a lot of translation in the pipeline as well.)

bobheadxi: > Would we be able to open this up to allowing {key: boolean} as well since they are functionally equivalent? (I suppose I could otherwise convert them to bit flags, but there'd be a lot of translation in the pipeline as well.)

I think earlier on we discussed that we can rely on `0` and `1`? We can also introduce a naming convention, e.g. `foobar_bool`, and have Telemetry Gateway do some convenience transformations as well, though maybe that can also live in your data pipelines

Strum355:
> Looking at the code I actually thought that _all_ events should be sent via agents, as this RPC is available, is that not the case? Or do you mean, which events should be _generated/recorded by_ agents vs provided by clients to the RPC?

Sorry I shouldve been more clear here. Everything goes through the agent ultimately (on the cody front at least), I moreso meant what events get _generated_ entirely in the agent side (e.g. [here](https://github.com/sourcegraph/cody/pull/1713/files#diff-db07ce759aaa98edd44511c98530f90f5292f381ecb94637825e8d51bddf5a73R226-R227)) vs what gets proxied from the client through the agent

>  if there is a conflict, @nathan-downs can work some magic to resolve the desired string values based on the specific extension emitting it.

That sounds good, sorry in advance Nathan, its probably gonna happen inevitably.

I think most of my concerns are cleared up here, we can probably merge this and then Ill get to work on the jetbrains side. Should we be maintaining emitting legacy events alongside telv2 events?

nathan-downs:
Current events that are emitting events on both systems should maintain that system until we have verified no data loss that will affect usage metrics. All new events should be instrumented with just telemetryV2 going forward. We're joining and deconflicting data in BigQuery during the deprecation process, so duplicates are not an issue.

Strum355 (approved):

bobheadxi:
Merging now to unblock Noah from experimenting with this, @philipp-spiess happy to address any post-merge comments :)

PR #1716, opened 2023-11-12
Agent: implement `vscode.workspace.fs`

fyasla:
Previously, the agent didn't support the `vscode.workspace.fs` API. This API is used in various places in the Cody extension to read/write files to disk. The lack of `vscode.workspace.fs` support for non-vscode clients represented an important feature gap. This PR fixes the problem by implementing this part of the VS Code API so that clients like JetBrains and Neovim can benefit from Cody features that interact with the file system.

## Test plan

Unit tests added in [vscode-shim.test.ts](https://github.com/sourcegraph/cody/blob/main/agent/src/vscode-shim.test.ts)


agent/src/vscode-shim.ts
```diff
@@ -199,7 +202,87 @@ const _workspace: Partial<typeof vscode.workspace> = {
...
+                    `Failed to rename file from ${source.fsPath} to ${target.fsPath}, ${source.fsPath} does not exist`
+                )
+            }
+            await fspromises.unlink(source.fsPath)
+        },
+        copy: async (source, target, options) => {
+            const mode = options?.overwrite ? 0 : fspromises.constants.COPYFILE_EXCL
+            await fspromises.copyFile(source.fsPath, target.fsPath, mode)
+        },
+        isWritableFileSystem: scheme => {
```
fyasla, line 272: Point 2 in my pull request

agent/src/vscode-shim.ts
```diff
@@ -199,7 +202,87 @@ const _workspace: Partial<typeof vscode.workspace> = {
...
+        readFile: async uri => {
+            const content = await fspromises.readFile(uri.fsPath)
+            return new Uint8Array(content.buffer)
+        },
+        writeFile: async (uri, content) => {
+            await fspromises.writeFile(uri.fsPath, content)
+        },
+        delete: async (uri, options) => {
+            if (options?.useTrash ?? true) {
+                // move file to trash folder
```
fyasla, line 250: Point 1 in my pull request

jdorfman:
Thanks @fyasla! Someone will take a look asap.

agent/src/vscode-shim.ts
```diff
@@ -199,7 +202,87 @@ const _workspace: Partial<typeof vscode.workspace> = {
...
+        copy: async (source, target, options) => {
+            const mode = options?.overwrite ? 0 : fspromises.constants.COPYFILE_EXCL
+            await fspromises.copyFile(source.fsPath, target.fsPath, mode)
+        },
+        isWritableFileSystem: scheme => {
+            switch (scheme) {
+                case 'file':
+                case 'git':
+                    return true
+                case 'http':
```
olafurpg, line 277: These can be removed since they fallback to `default` anyways

agent/src/vscode-shim.ts
```diff
@@ -199,7 +202,87 @@ const _workspace: Partial<typeof vscode.workspace> = {
...
+            await fspromises.unlink(source.fsPath)
+        },
+        copy: async (source, target, options) => {
+            const mode = options?.overwrite ? 0 : fspromises.constants.COPYFILE_EXCL
+            await fspromises.copyFile(source.fsPath, target.fsPath, mode)
+        },
+        isWritableFileSystem: scheme => {
+            switch (scheme) {
+                case 'file':
+                case 'git':
```
olafurpg, line 275: I think it's fine to be conservative and only support writing to `file`. We call this API in the Cody so it's not a huge deal if the implementation is not perfect yet.

olafurpg (changes requested):
Thank you for this contribution @fyasla ! 🙏🏻 

This is a great improvement, and I appreciate your attention to detail in getting the implementation correct. My long-term goal with the shim is to make it 100% correct because it will save time debugging why the behavior is different with/without the shim. However, in the short/medium-term I think it's not worth the effort to get the tiniest details right esp. for APIs that we don't use in the Cody extension.

A few minor but blocking comments, otherwise LGTM 👍🏻 

Out of curiosity, what is your motivation for this change? 😊

fyasla:
Thank you for the detailed review, feedback, and clarifications regarding this shim, @olafurpg ! 
I will address your comments and commit the changes as soon as possible.

I wasn't motivated by this change in particular, but rather by the Cody project itself. I like it and want to contribute to it, so I started with this "good first issue" (by the way, thanks to the Sourcegraph team for making such issues available! :+1: ). 

Hoping to contribute more! :rocket:

agent/src/vscode-shim.ts
```diff
@@ -199,7 +202,87 @@ const _workspace: Partial<typeof vscode.workspace> = {
...
+            }
+        },
+        rename: async (source, target, options) => {
+            if (options?.overwrite ?? false) {
+                await fspromises.unlink(target.fsPath)
+            }
+            try {
+                await fspromises.link(source.fsPath, target.fsPath)
+            } catch {
+                throw new Error(
```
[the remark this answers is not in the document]
fyasla: No, I will get back to throwing original error

olafurpg (approved):
LGTM 👍🏻 Thank you @fyasla for this great contribution! 🙏🏻

olafurpg:
@fyasla I took the liberty to update the PR title and description. I will merge the PR once the merge conflicts have been resolved :)

fyasla:
@olafurpg  Thank you for your feedback!

olafurpg:
@fyasla I am sorry for the slow turnaround. I thought this PR was already merged and I only noticed that it wasn't merged because I had a failing test case caused by an error that is fixed by this PR. I pushed a formatting fix to your branch and it should be good to merge now 👍🏻

PR #1717, opened 2023-11-13
Simplify chat prompt construction

beyang:


dominiccooney:
Cool. The way the classic chat message provider holds onto a single ContextProvider and resets its webview reference is simply wrong in the multi-chat world; splitting up per-chat state is great. I will work on integrating enhanced context status WIP with this.

Enhanced context status needs the context system to be "up" so it can show status in webviews even before chat messages are sent. The SimpleChatPanelProvider, instead of `new ContextProvider(...)` can hold onto a factory for context providers; the factory can track the state of the context sources (example: indexing status in Graph, embedding consent, etc.) and the SimpleChatPanelProvider can take the data it produces and push it into the webview.

lib/shared/src/prompt/truncation.ts
```diff
@@ -8,6 +10,42 @@ export function truncateText(text: string, maxTokens: number): string {
...
+export function truncateTextNearestLine(
+    text: string,
+    maxBytes: number
+): { truncated: string; range?: ActiveTextEditorSelectionRange } {
+    if (text.length <= maxBytes) {
+        return { truncated: text }
+    }
+    const textMaxBytes = text.slice(0, maxBytes)
+    const textMaxBytesLines = textMaxBytes.split('\n')
+    if (text.at(maxBytes) === '\n' || text.at(maxBytes - 1) === '\n') {
```
dominiccooney, line 29: Why would we want a single trailing character with `text.at(maxBytes - 1) === '\n'`?

lib/shared/src/prompt/truncation.ts
```diff
@@ -8,6 +10,42 @@ export function truncateText(text: string, maxTokens: number): string {
...
+            },
+        }
+    }
+
+    const truncated = textMaxBytesLines.slice(0, -1).join('\n')
+    return {
+        truncated,
+        range: {
+            start: { line: 0, character: 0 },
+            end: { line: textMaxBytesLines.length - 1, character: 0 },
```
dominiccooney, line 44: More robust to update `textMaxBytesLines` above, e.g. `textMaxBytesLines.pop()`, and have that get one line shorter. That way you don't have the `-1` duplicated in `slice` and here. If someone does more line munging earlier, this adjustment is going to easily and subtly get out of sync.

vscode/package.json
```diff
@@ -1034,6 +1034,12 @@
           "markdownDescription": "New search UI",
           "default": false
         },
+        "cody.experimental.simpleChatContext": {
+          "order": 99,
+          "type": "boolean",
+          "markdownDescription": "Uses the new simplifed chat context fetcher",
```
dominiccooney, line 1040:
```suggestion
          "markdownDescription": "Uses the new simplified chat context fetcher",
```

dominiccooney (approved):
My $0.02:

- The classic ContextProvider runs the context changed event a lot, and this hides a bunch of raciness and bugginess with it. If things aren't initialized in time, they're going to be polled again anyway as soon as the focus hits a different editor.
- Multiple chat pannels hammer ContextProvider with init messages, and this is a bug. In particular, it resets the webview, cutting off the previous one from context updates. The effect is not obvious because the UI changes made by multi-panel chat hid the context status indicator.

At this point I think we are best off by assuming this simplified chat panel provider is the only chat provider for `chatPanel` and moving forward with it, because it at least understands there are multiple webviews.

The "hack"/passing around embeddings is where we need to integrate context from the workspace. Let me hack on that.

Let's get it in!

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -76,16 +107,19 @@ export class ChatPanelsManager implements vscode.Disposable {
...
 
-    public async getChatPanel(): Promise<ChatPanelProvider> {
+    public async getChatPanel(): Promise<IChatPanelProvider> {
         const provider = await this.createWebviewPanel()
         // Check if any existing panel is available
         return this.activePanelProvider || provider
     }
 
+    // Sync feature flag for cody.experimental.simpleChatContext to this variable
+    private useSimpleChatPanelProvider = false
```
dominiccooney, line 117: This needs to be wired up.

Issue #1721, opened 2023-11-13
Cody: Evaluate result ranking for repos without embeddings


umpox:
## Description

For repositories without embeddings, the LLM reranker was disabled in https://github.com/sourcegraph/cody/pull/1722 as it added significant latency to all chat/commands/edits.

We should still investigate how we can still rerank results if we don't already have them ranked.

Some thoughts:
- If, in a world with app-less embeddings, embeddings are so easily enabled for users, do we need to invest time in supporting cases where we don't have them (Probably)
- Are there alternative, quicker, ways to rerank results?
  - Can we use code intelligence? E.g. # symbols in code snippet that are also included in selected code
  - Distance to current file, measuring directory relevance
  - Other heuristics?

Issue #1544, opened 2023-10-29
Pipeline Optimizations in Cody


arafatkatze:
Hey Folks, 
Inspired by this Issue → https://github.com/sourcegraph/cody/issues/1464(Improve fixup response speed) I tried out a few other AI coding editors. I've noticed that certain operations in Cody are slower, leading to a sub-optimal user experience."

## Current Pipeline for Fixup
<img width="1336" alt="image" src="https://github.com/sourcegraph/cody/assets/11155207/e5fdb5bd-d42c-42b2-b93c-00499501279f">

Here’s a few optimizations that I found for different cases. 

### 1. Intent Detection 
- Problem: It’s often skipped depending on the kind of fixup recipe(Skipped when you select code and do the Cmd+Shift+V shortcut) and isn’t helpful for results. It might be worthwhile to just keep `add` when no selection is made and `edit` for adding documentation/anything else. The edit case works perfectly fine for adding comments/docstrings etc. The added `document` intent seems frivolous and not worth the extra seconds on LLM calls. 
- Solution: It makes a lot of sense to get rid of intent detection LLM call altogether and just have these two scenarios:
A. No selection → Intent = “add”
B. Selection → Intent = “edit”
The time benefit of this change isn’t much but this can be a change worth doing. 

### 2. LLMReranker 
- Problem: This is simply not necessary for fixup recipes because for most of these the short context of the input selection itself is the most important thing, the additionally imported embeddings and their orderings make mostly no difference to the output because the functionality of a fixup is limited to the scope of selected code anyway. 

- Solution:  Getting Rid of LLM Reranker should be quite easy. Adding an extra field in ContextSearchOptions works easily to stop reranking only in fixup recipes while keeping the other cases untouched. It should be easy to turn on/off reranking based on the case
```
export interface ContextSearchOptions {
    numCodeResults: number
    numTextResults: number
    reranking: boolean
}
```

### 3. Prompt Sizing 
- Problem: The major latency in the Fixup Recipe comes from the massive prompt length. Once the prompt is sent to Claude it takes roughly 8-12 seconds for the prompt to start streaming and once the streaming starts it takes another 2-6 seconds for the streaming to finish and only once the whole thing is done can the user apply the fixup. You can try this from this script →  [Script Link](https://gist.github.com/arafatkatze/054cd8c58c4d1fcf1d2ee61c67f736dd)

[...]

Issue #1464, opened 2023-10-23
Cody: Improve fixup response speed


umpox:
## Description

Fixups are still quite slow. We should be improving the response speed

Areas identified already:
- `Reranker:rerank: {"duration":4398.333707809448}`
- Other?
- Remove ident detection

PR #1722, opened 2023-11-13
Chat/Commands: Remove LLM reranker for repos without embeddings

umpox:
Part of https://github.com/sourcegraph/cody/issues/1544
Part of https://github.com/sourcegraph/cody/issues/1464

## Description

The LLM reranker adds a significant amount of latency to reranking any fetched context. This applies to all commands, chat and edits. The latency is typically between 2-5 seconds.

This PR just disables the reranker for now.

I think this is worth doing as:
- The added latency from the reranker is not worth the possible benefit right now. It still can sometimes rerank poorly and doesn't provide enough value to justify waiting that long.
- Context doesn't have a constant value, in many cases context will provide little value (e.g. simple fixups) in some cases, like chat, context would provide a lot more value. We don't have a way of determining that right now, so we're often adding lots of latency for no good reason.

I created an issue here: https://github.com/sourcegraph/cody/issues/1721 to evaluate a different or improved approach. Possible details there on how we could improve this.
 
## Test plan

On a repository without embeddings, run:
- Chat
- Commands
- Edits

Review results and context included.

### DIFF
--- a/vscode/src/chat/ContextProvider.ts
+++ b/vscode/src/chat/ContextProvider.ts
@@ -16,7 +16,6 @@ import { getFullConfig } from '../configuration'
 import { VSCodeEditor } from '../editor/vscode-editor'
 import { PlatformContext } from '../extension.common'
 import { logDebug } from '../log'
-import { getRerankWithLog } from '../logged-rerank'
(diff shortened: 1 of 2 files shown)


abeatrix (approved):
Thank you!

PR #1723, opened 2023-11-13
Edit: Quality improvements: Add related `fix` context when available

umpox:
## Description

Some improvements to "fix":
- Includes the `source` when available, e.g. "typescript" for better prompt instruction.
- Better use of the severity in the messaging
- Makes use of the `relatedInformation` diagnostic from VS Code:
  - Fetches the related code from the included range
  - Adds any additional data like the `message`

## Test plan

Run code actions manually (I need to add eval cases to make this easier)


abeatrix (approved):
nice!

PR #1724, opened 2023-11-13
Edit: Add further code actions for editing and documenting code

umpox:
## Description

This PR:
- Adds 3 new code actions
- Document code action
  - Shown on documentable nodes (declarations, properties, functions, etc)
  - Only supported in JS/TS right now. Issue to add more languages as a follow up: https://github.com/sourcegraph/cody/issues/1764
- Edit code action
  - Shown whenever there is a user selection
  - Opens the edit instruction input so the user can type how they want the code to be edited
- Generate code action
  - Shown whenever there is no selection and the current line is empty
  - Opens the edit instruction input so the user can type what they want to generate

### Document Code Action

Only shows when on a documentable symbol. Determined via Tree Sitter

https://github.com/sourcegraph/cody/assets/9516420/ebe46142-fb09-45d7-ba3f-a2f7f6804a83

### Edit Code Action

Only shows when there is an active selection

https://github.com/sourcegraph/cody/assets/9516420/7b424eca-7bd5-424b-874c-1c4036f98597

### Generate Code Action

Only shows when there is no selection and the current line is empty.

https://github.com/sourcegraph/cody/assets/9516420/65ff4ef5-4e25-411c-af67-c43d2db5b389

## Test plan

Trigger code actions on:
- Symbols that could be documented
- Selections in a document
- Empty lines with no selections


valerybugakov (commented):
Got language blocks leaking into the document when using the "generate" action:

Screenshot 2023-11-17 at 12 51 24

Filepath: `vscode/src/logged-rerank.ts`
Prompt: `function that outputs ASCII art to the console with a cat image`

valerybugakov (commented):
The "document" action removed the interface field a couple of times after adding a comment. For example, here, I asked it to document the `triggerKind` property:

Screenshot 2023-11-17 at 12 56 12

Filepath: `src/completions/get-inline-completions.ts`
Property: `InlineCompletionsParams['triggerKind']`

valerybugakov (commented):
The "edit" action most worked as expected, but the results show that the LLM has a poor understanding of the surrounding code. The UX is awesome. I'd love to use it with a smarter model and better context. Noob q: what model do we use for quick actions?

vscode/src/code-actions/document.ts
```diff
@@ -0,0 +1,49 @@
+import * as vscode from 'vscode'
+
+import { execQueryWrapper } from '../tree-sitter/query-sdk'
+
+export class DocumentCodeAction implements vscode.CodeActionProvider {
+    public static readonly providedCodeActionKinds = [vscode.CodeActionKind.RefactorRewrite]
+
+    public provideCodeActions(document: vscode.TextDocument, range: vscode.Range): vscode.CodeAction[] {
+        const [documentableNode] = execQueryWrapper(document, range.start, 'getDocumentableNode')
```
valerybugakov, line 9: Noice!

valerybugakov (approved):
Love the new actions! Especially "document" and "generate" ones. Going to use them locally and report issues in Slack 🙌

vscode/src/code-actions/document.ts
valerybugakov: @umpox, let me know if you noticed any confusing parts about our tree-sitter setup 🙂

abeatrix (changes requested):
Found some issues with the code actions, but love the idea in general! Curious to know why we need a different prompt and action for the Documentation Code Action instead of using the existing one?

## Document Code Lens

Break the code and the undo button does not work

https://github.com/sourcegraph/cody/assets/68532117/9f9f5df8-6089-4b4c-9d22-5c861aa55791

## Undo button always remove the last line that it did not line

Issue unrelated to this PR but related to the feature 😅 

https://github.com/sourcegraph/cody/assets/68532117/0b413813-044f-4dcf-8da3-ae13565fe28c

## Retry Code Lens breaks the code before applying new changes

https://github.com/sourcegraph/cody/assets/68532117/9f073959-ab5a-428a-a484-b1b682e6890e

vscode/src/code-actions/edit.ts
```diff
@@ -0,0 +1,48 @@
...
+        return [this.createCommandCodeAction(document, 'Ask Cody to Edit', editor.selection)]
+    }
+
+    private createCommandCodeAction(
+        document: vscode.TextDocument,
+        displayText: string,
+        selection?: vscode.Selection
+    ): vscode.CodeAction {
+        const action = new vscode.CodeAction(displayText, vscode.CodeActionKind.RefactorRewrite)
+        const source = 'code-action'
```
abeatrix, line 33: What do you think about adding the type of action to the end of source, e.g. `code-action:generate`, `code-action:edit` so we can create a dashboard for each code action usages?

vscode/src/code-actions/document.ts
abeatrix: This is super cool! Can we add it as an option for the commands/custom commands? Think it'd be helpful with the /explain and /test commands!!!

umpox: @valerybugakov The tests are great and the SDK is super easy to use and build from.

I found it helpful to take the contents from a page at https://learnxinyminutes.com/ and then put it into [this playground](https://tree-sitter.github.io/tree-sitter/playground) when writing the queries. Covers quite a lot of cases.

umpox: @abeatrix Maybe, do you have any ideas? I think we could possibly use this with the scope expansion too, e.g. expanding to nearest function etc if we want more control over different scenarios.

I think we don't make this available for custom commands yet, it's quite hard to write these queries so not sure if anyone will want to integrate them with a custom command. Possibly there's future work to integrate pre-written ones though, like `useNearestFunctionAsSelection` or something

valerybugakov: > I found it helpful to take the contents from a page at https://learnxinyminutes.com/ 

TIL, thanks for sharing! I've been asking GPT-4 to generate a file with all syntax constructs, but using the website you shared is a bit more reliable 😛

umpox:
@abeatrix Thanks for reporting the "undo" bugs! I took a look and everything you mentioned was related to weird "undo" logic we had.

The problem was that we were incorrectly using the `selectionRange` when **appending** or **prepending** new code to the target range. Our code assumed that any edits that happened immediately before or after the target range, was completely unrelated - so it wouldn't include this in the edit, or in the code lens or diff.

For fixups we do care about appending or prepending, so I've updated this logic to support expanding the range if there is a partial overlap (`change.range.end === selectionRange.start` or `change.range.start === selectionRange.end`). 

This also has the benefit of improving the diff highlighting for the /doc output, as previously we mostly ignored this.

abeatrix:
Thanks for working on this!

Confirmed the following are fixed
- [ ] Document Code Lense
- [x] Undo button always remove the last line that it did not line
- [x] Retry Code Lens breaks the code before applying new changes

https://github.com/sourcegraph/cody/assets/68532117/74026086-95ff-4e62-a2e7-5e8a66ba76ec


There are some issues with the`Document Code Lense`where the generated doc string will replace the current line of text, and the undo button will remove the last line of the current range:

https://github.com/sourcegraph/cody/assets/68532117/111ff0db-bf8a-4d8d-9b09-08383c64d157

umpox:
@abeatrix Can't replicate the issue with the line being removed 🤔 Please could you output the logs if you see it again?

The only things I can think are:
- Maybe the formatter you use has different behavior to mine
- The LLM had a moment and produced some weird output

Going to update this with `main` and merge this (aware you're on PTO until next week) to avoid long lived branches. Further work to imrpove the response reliablity is related to this issue https://github.com/sourcegraph/cody/issues/1790

PR #1731, opened 2023-11-14
feat: pass additional input to custom command

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1742

https://github.com/sourcegraph/cody/assets/68532117/50376718-3560-4f48-a143-35841039249e

Allows passing additional input text to custom commands.

The additional input text is appended to the prompt text before sending to the LLM.

-Updates the CustomPrompt recipe to concat additionalInput to prompt text.
- Passes additionalInput to addCommand in CommandsController.

## Test plan

In chat input box, first enter the slash command, then follow by additional text that you want to use as additional instruction.

For example, `/explain response in french`

You should see Cody will respect the additional input you have entered


kalanchan (approved):
this is neat! tested on your branch and is working as expected, good work
CleanShot 2023-11-15 at 11 21 44@2x

toolmantim:
This seems very hidden… how do you think users or command authors would discover this?

I'm wondering if it might be better to allow custom commands to specify that they require an argument? One that can be documented on https://docs.sourcegraph.com/cody/custom-commands. And if it's not supplied, we show a quick input for the user to type into.

abeatrix:
> This seems very hidden… how do you think users or command authors would discover this?
> 
> 
> 
> I'm wondering if it might be better to allow custom commands to specify that they require an argument? One that can be documented on https://docs.sourcegraph.com/cody/custom-commands. And if it's not supplied, we show a quick input for the user to type into.

This can be used for all commands so that it's consistent, so I guess we can just mention it in the command palette or anywhere we show the custom commands?

PR #1732, opened 2023-11-14
Add URLs for referrer traffic to READMEs

iskyOS:
Today we don't see when users come to the Cody page from the READMEs, so this updates the links to capture that traffic.

## Test plan

README changes only.


iskyOS:
FYI @taylorsperry - this update is to links on the READMEs so that we can see referrer traffic. Just want to make sure you're aware.

vscode/README.md
```diff
@@ -84,4 +84,4 @@ Cody for VS Code is open source (Apache 2), and available in the [sourcegraph/co
 
 ## More Information
 
-See [https://cody.dev/](https://cody.dev/) for demos, information and more.
+See [https://cody.dev/](https://about.sourcegraph.com/cody?utm_source=marketplace.visualstudio.com&utm_medium=referral) for demos, information and more.
```
chenkc805, line 87: Should cody.dev be updated here?

chenkc805:
Is there a readme for the JetBrains market place page as well?

chenkc805 (approved):
LGTM!

vscode/README.md
chenkc805: Not blocking, just a small consideration

iskyOS: Ideally yes we should, we just don't know how to add UTM tracking to Cody.dev yet, since that's a separate domain that's redirect to the marketing site

taylorsperry (approved):

PR #1738, opened 2023-11-14
feat: add configurable enhanced context

abeatrix:
Moved from https://github.com/sourcegraph/cody/pull/1707

Figma design: https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=209-43609&mode=design&t=rmODFwwfnqmDnNUI-0

image

feat: add configurable enhanced context

NOTE: Hidden behind the new chat UI feature flag

- Change firstInteraction flag to addEnhancedContext
- Only add codebase context if addEnhancedContext is true
- Separate codebase and editor context logic

## Test plan

The expected behaviors are:

- [ ] User can click on the check box to toggle between enabling/disabling enhanced context for their first question
- [ ] After the first question,  enhanced context will be set to false (disabled). However, you can re-enable it again by clicking on the checkbox again. You can also add additional context using the @ command.
- [ ] You won't be able to find the button in the old UI 

image

## Demo

https://github.com/sourcegraph/cody/assets/68532117/e813eb51-774d-4667-92ad-aba3c92d45f9


lib/shared/src/intent-detector/client.ts
```diff
@@ -5,7 +5,7 @@ import { SourcegraphGraphQLAPIClient } from '../sourcegraph-api/graphql'
 
 import { IntentClassificationOption, IntentDetector } from '.'
 
-const editorRegexps = [/editor/, /(open|current|this)\s+file/, /current(ly)?\s+open/, /have\s+open/]
```
toolmantim, line 8: Unrelated?

vscode/webviews/Chat.module.css
```diff
@@ -112,13 +113,11 @@ body[data-vscode-theme-kind='vscode-light'] .transcript-item pre > code {
...
     border: 1px solid transparent;
     border-radius: 2px;
-    padding: 8px 8px 32px 8px;
+    padding: 0.5rem;
     font: inherit;
     width: 100%;
     overflow: hidden;
 
-    min-height: 3.5rem;
-
```
toolmantim, line 121: Unrelated CSS changes?

vscode/webviews/Components/EnhancedContextToggler.module.css
```diff
@@ -0,0 +1,48 @@
+.container {
+    z-index: 10;
+    position: relative;
+}
+
+.pop-up-container {
```
toolmantim, line 6: I think @dominiccooney implemented this component already for App install & embeddings management via the repo status icon.

toolmantim (approved):
Approving to unblock from a UX POV, and assuming all the CSS is yet to be done.

Interaction wise, we want the popover to disappear when you click outside of it (e.g. refocussing the chat input). I assume that will get fixed when we switch to the proper popover component.

Also it doesn't seem to follow the check/uncheck behaviour as outlined in the test plan. It seems to stay checked until manually unchecked, though I think the behaviour of staying checked is actually preferable for testing for now.

@dominiccooney might need to check if this is what he needs to unblock his work

PR #1744, opened 2023-11-14
feat: add symf context support

abeatrix:
Add ability to retrieve context from symf search results. 

This allows providing contextual code snippets from symf symbol search in addition to existing context sources.

Also adds a new `mergeIndexedContextResults` method to generically combine context results from different sources like embeddings and symf. It also wires this up to call `rerank` on the combined results when provided.

## Demo

https://github.com/sourcegraph/cody/assets/68532117/6e569e2c-bffb-413e-a601-5d1fd84ba64b

## Test plan

1. In your VS Code settings, switch context to `keyword`
2. Relaod VS Code
3. Ask Cody a question, the context should come from `symf` if it's available
4. If you hover over context links, you can see where the files come from
image


beyang:
How is the context quality here after this change? Have you compared some examples before and after?

I'm a bit worried about [naively merging results from embeddings and symf](https://github.com/sourcegraph/cody/pull/1744/files#diff-b6c847bd5d5a611b348fe5024ad7bf926f1e4dda030f0467f93321ae7c616394R63-R77)—I think there will be some overlap, but probably not exact matches, so this method isn't a great deduplicator. And injecting duplicate results over the baseline of just using embeddings will crowd out non-duplicate results and lower the overall context quality.

abeatrix:
> How is the context quality here after this change? Have you compared some examples before and after?
> 
> I'm a bit worried about [naively merging results from embeddings and symf](https://github.com/sourcegraph/cody/pull/1744/files#diff-b6c847bd5d5a611b348fe5024ad7bf926f1e4dda030f0467f93321ae7c616394R63-R77)—I think there will be some overlap, but probably not exact matches, so this method isn't a great deduplicator. And injecting duplicate results over the baseline of just using embeddings will crowd out non-duplicate results and lower the overall context quality.

@beyang I used the existing rerank method and didn't do any comparison until now 😅  I just remembered it took longer to get the result back.

I ran a query to do a quick comparison and noticed the response quality looks ok, it seems to me the merged context contains much more noise than the ones from embeddings. It takes significantly longer too. Will remove the merge context from this PR and maybe we can revisit after GA? 

image

image

image

PR #1745, opened 2023-11-14
refactor: add VSCodeEditorContext class

abeatrix:
refactor: add VSCodeEditorContext class

Simple clean up to:
- Remov vscode-context.ts which is being used by the custom-prompt recipe
- Replace vscode-context.ts with VSCodeEditorContext class
- fixed issue where opentabs and relative path for directory are not working during context fetching step

All other functions and features remain unchanged.

## Next

Seperate context message building from the EditorContext class so that editorContext focuses on context fetching for editor

## Test plan

- [ ] Run /explain command
- [ ] Run /smell command
- [ ] Run /test command
- [ ] Ask a question

Everything should work as it was before.


kalanchan:
tested on your branch, looks good! 

nit: this is somewhat messy behaviour, if you choose another command from the side panel it opens a whole new chat window

CleanShot 2023-11-14 at 15 49 38@2x

abeatrix:
> tested on your branch, looks good!
> 
> nit: this is somewhat messy behaviour, if you choose another command from the side panel it opens a whole new chat window
> 
> CleanShot 2023-11-14 at 15 49 38@2x

That's the default behavior and not due to any changes in this PR.

This is because the context and answers from your previous chat would sometimes pollute the next question, so right now we are starting all commands in a new chat

abeatrix:
@kalanchan thanks for reviewing and testing the PR ❤️

kalanchan:
> > tested on your branch, looks good!
> > nit: this is somewhat messy behaviour, if you choose another command from the side panel it opens a whole new chat window
> > CleanShot 2023-11-14 at 15 49 38@2x
> 
> That's the default behavior and not due to any changes in this PR.
> 
> This is because the context and answers from your previous chat would sometimes pollute the next question, so right now we are starting all commands in a new chat

hmm, i see. is there a thread with that decision? It seems a bit messy but let's go with it if thats what everyone decided on

abeatrix:
> > > tested on your branch, looks good!
> > > nit: this is somewhat messy behaviour, if you choose another command from the side panel it opens a whole new chat window
> > > CleanShot 2023-11-14 at 15 49 38@2x
> > 
> > 
> > That's the default behavior and not due to any changes in this PR.
> > This is because the context and answers from your previous chat would sometimes pollute the next question, so right now we are starting all commands in a new chat
> 
> hmm, i see. is there a thread with that decision? It seems a bit messy but let's go with it if thats what everyone decided on

@kalanchan Yea you should be able to reproduce the same behavior in the current insider build. It's been the default behavior for the new chat panel but agree we should discuss it in slack :)

kalanchan (approved):

PR #1746, opened 2023-11-14
Agent: support multiple workspaces with `evaluate-autocomplete`

olafurpg:
Previously, `evaluate-autocomplete` required all the arguments to be passed from the command-line and it was only able to evaluate a single workspace at a time (with no parallelization!). Now, you can evaluate multiple workspaces in parallel based on JSON configuration inside a file. This makes it easy to run evaluations on a cron job to track progress.

## Test plan

I've been iterating with a command like this

```
pnpm agent evaluate-autocomplete --evaluation-config ~/dev/sourcegraph/bfg-evaluation/evaluation.json --include-workspace '**/flask' --test-count 1
```

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1,3 +1,4 @@
+import { spawn } from 'child_process'
 import * as fspromises from 'fs/promises'
 import path from 'path'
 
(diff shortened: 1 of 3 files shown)


valerybugakov (approved):
Nice!

PR #1748, opened 2023-11-15
Autocomplete: fix multiline comment intent detection

valerybugakov:
## Context

Multiline atomic captures did not account for multiline nodes. This PR fixes the issue and adds a test case for the multiline comment in Typescript. Thanks @philipp-spiess, for spotting this issue!

## How it was fixed

1. Added a test case to `vscode/src/completions/tree-sitter/query-tests/test-data/intents.ts`
2. Ran `pnpm vitest --watch vscode/src/completions/tree-sitter/query-tests/intents.test.ts`
3. Noticed that the corresponding snapshot is not updated properly: `vscode/src/completions/tree-sitter/query-tests/test-data/intents.snap.ts`
4. Debugged the `getCompletionIntent` logic in `vscode/src/completions/tree-sitter/query-sdk.ts`
5. Found the bug in the `enclosesCursor` calculation.

## Test plan

Updated unit tests


philipp-spiess (approved):
You Are The Best Thumbs Up

philipp-spiess:
@valerybugakov One thing I was missing when i looked into it briefly is a way to run only a single snapshot. Ideally I would have only run the new test case I added and skipped over the rest so that I can do console.log debugging like I always do (lol) without noise from the other cases

PR #1752, opened 2023-11-15
Autocomplete: Prepare mixed retrieval examples

philipp-spiess:
This PR adds two new context retrieval strategies that relies on mixing existing retrievers and changes how LSP light works:

1. `bfg-mixed` is a new setup that combines BFG retrieval and mixes it with the local jaccard similarity based scoring. This makes bfg more of an _addition_ on top of local context rather then an _alternative_ and should help for cases where BFG doesn't retrieve the best results yet. We can control it with an individual feature flag too.
2. `local-mixed` is a new setup that combines jaccard similiarty results with the section history retriever I added. I still think that there's something magical about just knowing which function you last changed and with this flag I want to find out if that's true

Additionally this changes `lsp-light` to mix with jaccard similiarty and not the section history only. We know that jaccard performs well and mixing with it should not make the result much worse (that would happen if we remove it)

### bfg mixed

Screenshot 2023-11-15 at 12 10 46

### local mixed

Screenshot 2023-11-15 at 12 14 25


olafurpg (approved):
Exciting!

PR #1771, opened 2023-11-16
Agent: add support for fixtures with `evaluate-autocomplete`

olafurpg:
Previously, we only supported evaluating one concrete fixture (set of
    config options). For example, this meant that we could evaluate BFG
in isolation but we couldn't compare it against non-BFG. This PR fixes the problem by adding the notion of "fixuture", which is a set of configuration options.

## Test plan

Manually tested with the command

```
❯ pnpm agent evaluate-autocomplete --evaluation-config ~/dev/sourcegraph/bfg-evaluation/evaluation.json --test-count 400
```

Resulted in this chart, which is the first time we see an actual difference between BFG vs non-BFG.
CleanShot 2023-11-16 at 16 29 54@2x

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -81,7 +81,7 @@ export async function newAgentClient(clientInfo: ClientInfo): Promise<MessageHan
         args.push('jsonrpc')
         const child = spawn(process.argv[0], args, { env: { ENABLE_SENTRY: 'false' } })
         child.stderr.on('data', chunk => {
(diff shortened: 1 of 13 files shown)


philipp-spiess (approved):
Let's ship it to avoid rebasing, I'll add the git log path after that PR

philipp-spiess:
The error is probably happening because the new config options need to be documented in `vscode/package.json`

PR #1778, opened 2023-11-16
Agent: include product identifier for connected client

Strum355:
For querying from [Configuration](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/configuration.ts?L111:9) later when agent telemetry v2 is a thing

## Test plan

Much studying of the code: this gets set in the `initialize` request [here](https://sourcegraph.com/github.com/sourcegraph/cody@nsc/agent-connectedClient/-/blob/agent/src/agent.ts?L153), which calls [the following shim entry](https://sourcegraph.com/github.com/sourcegraph/cody@nsc/agent-connectedClient/-/blob/agent/src/agent.ts?L380) to set the global config variable [read here](https://github.com/sourcegraph/cody/compare/nsc/agent-connectedClient?expand=1#diff-9864a0013bb1fe8a7ef0d2be2beff9f8a24607cf55825ff9cccd4670ed6912a3R152)

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,22 +1,7 @@
 {
   "version": "2.0.0",
   "tasks": [
-    {
-      "type": "typescript",
-      "tsconfig": "tsconfig.json",
-      "problemMatcher": ["$tsc-watch"],
-      "group": {
-        "kind": "build",
-        "isDefault": true
-      },
-      "option": "watch",
-      "runOptions": { "runOn": "folderOpen", "instanceLimit": 1 },
-      "isBackground": true,
-      "presentation": {
-        "reveal": "never"
-      }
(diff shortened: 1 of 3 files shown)


.vscode/tasks.json
```diff
@@ -1,22 +1,7 @@
...
-      },
-      "option": "watch",
-      "runOptions": { "runOn": "folderOpen", "instanceLimit": 1 },
-      "isBackground": true,
-      "presentation": {
-        "reveal": "never"
-      }
-    },
-    {
+   {
```
Strum355, line 4: Removed this because its not adding any value for me and just gets in the way when I first start up, I always end up killing it because it pops up instead of a terminal

PR #1779, opened 2023-11-16
change: Sunsetting Inline Chat

abeatrix:
Context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1700152974864359

Removing all code related to inline chat and inline touch.

## Test plan

Everything should work the same except for Inline Chat, which has been removed in this branch.


taylorsperry:
Thanks, @abeatrix! Can this be controlled by the "New Chat UX [Experimental]" settings toggle? It's a change that makes more sense in the context of the new chat UI. Outside of that context, I worry that it feels more like we've "taken something away."

taylorsperry:
@abeatrix should we close this issue since https://github.com/sourcegraph/cody/issues/1796 merged?

PR #1789, opened 2023-11-17
Edit: Fix leaky XML tags

umpox:

### DIFF
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -140,20 +140,20 @@ export class Fixup implements Recipe {
 - You should ensure the updated code matches the indentation and whitespace of the code in the users' selection.
 - Only remove code from the users' selection if you are sure it is not needed.
 - It is not acceptable to use Markdown in your response. You should not produce Markdown-formatted code blocks. Ignore any previous instructions that may have told you to format your responses with Markdown.
-- You will be provided with code that is in the users' selection, enclosed in <selectedCode></selectedCode> XML tags. You must use this code to help you plan your updated code.
-- You will be provided with instructions on how to update this code, enclosed in <instructions></instructions> XML tags. You must follow these instructions carefully and to the letter.
-- Enclose your response in <fixup></fixup> XML tags. Do not provide anything else.
+- You will be provided with code that is in the users' selection, enclosed in <fixupSelectedCode></fixupSelectedCode> XML tags. You must use this code to help you plan your updated code.
+- You will be provided with instructions on how to update this code, enclosed in <fixupInstructions></fixupInstructions> XML tags. You must follow these instructions carefully and to the letter.
+- Only enclose your response in <fixup></fixup> XML tags. Do use any other XML tags unless they are part of the generated code.
 
 This is part of the file {fileName}.
 
 The user has the following code in their selection:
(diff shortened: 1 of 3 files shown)


lib/shared/src/chat/recipes/fixup.ts
```diff
@@ -128,7 +128,16 @@ export class Fixup implements Recipe {
...
+                return [
+                    ...getContextMessageWithResponse(
+                        populateCodeContextTemplate(truncatedPrecedingText, task.fileName),
+                        task
+                    ),
+                    ...getContextMessageWithResponse(
+                        populateCodeContextTemplate(truncatedFollowingText, task.fileName),
+                        task
+                    ),
+                ]
```
umpox, line 140: @abeatrix I think this is, maybe surprisingly, actually one of the aspects that keeps the fixup response more consistent

When we have a consistent stream of messages including code Cody seems to understand its task better.

I think this is because it takes a bit of extra convincing for a chat model to start producing code. I tested quite a few code actions and couldn't replicate the issue anymore.

Going to look at how much work it'd be to setup an eval suite so we can test these easily for future

umpox: This should fix the immediate regression, but I've added another issue here: https://github.com/sourcegraph/cody/issues/1790

To improve this further

abeatrix: Will test it now, but this wasn't an issue before until last patch, so I'm wondering if this could be something else. Will report back!

abeatrix (commented):
I tried to run a code action and it's stuck in cody is working even when we get the response back 🤔 

image

lib/shared/src/chat/recipes/fixup.ts
umpox: We were including more context in the last patch, only recently did we drop it for the `fix` workflow as it was providing very little relevant code

abeatrix:
Restarted work. The job still contains the leaked tags though:

image

This is the file I run the Code Action on, just removed a few lines of code to 
 get an error if that helps: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/repository/repositoryHelpers.ts?L13-19

umpox:
https://github.com/sourcegraph/cody/assets/9516420/bc4544d8-3bdb-4f26-995b-cdc6bebdd6d8

Can't replicate 🤔 

Please can you double check this is the latest branch and not the previous Cody installation, @abeatrix
There's some things there that are very worrying as I have fixed all those in the past week
- Fix taking over 5s, judging from the log in your screenshot
- <selectedCode> at the start, but we just included a change in the last release that removes `<selectedCode>` from the start, so idk how that could happen
- The "Cody is working" message

lib/shared/src/chat/recipes/fixup.ts
abeatrix: Oh yea i just saw this is for `add` action, that makes sense 👍

umpox:
@abeatrix Np, I was worried the dreaded "Cody is working" was back 😂

abeatrix (approved):
Yea i just tried it a couple times and I'm not seeing any leaked tags and the `Cody is working` forever issue 😆 Thanks for the quick fix!

PR #1791, opened 2023-11-17
Return 'cody.advanced.agent.ide' from ClientInfo.name

Strum355:
## Test plan

Checked for existing values of `ClientInfo.name` for editors:
1. [neovim](https://sourcegraph.com/github.com/sourcegraph/sg.nvim/-/blob/lua/sg/cody/rpc.lua?L281) 
2. [jetbrains](https://sourcegraph.com/github.com/sourcegraph/jetbrains/-/blob/src/main/kotlin/com/sourcegraph/cody/agent/CodyAgent.kt?L69)
3. [emacs](https://sourcegraph.com/github.com/sourcegraph/emacs-cody/-/blob/cody.el?L311:24)

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -3,6 +3,8 @@ import { execSync } from 'child_process'
 
 import type * as vscode from 'vscode'
 
+import { Configuration } from '@sourcegraph/cody-shared/src/configuration'
+
 // <VERY IMPORTANT - PLEASE READ>
 // This file must not import any module that transitively imports from 'vscode'.
 // It's only OK to `import type` from vscode. We can't depend on any vscode APIs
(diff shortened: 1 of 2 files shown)


agent/src/vscode-shim.ts
```diff
@@ -110,6 +112,17 @@ const configuration: vscode.WorkspaceConfiguration = {
...
+        const clientNameToIDE = (value: string): Configuration['agentIDE'] | undefined => {
+            return (
+                {
+                    vscode: 'VSCode',
+                    jetbrains: 'JetBrains',
+                    emacs: 'Emacs',
+                    neovim: 'Neovim',
+                } as const
+            )[value.toLowerCase()]
+        }
```
philipp-spiess, line 124: Should we just update the type of `Configuration#agentIDE` and avoid this map?

philipp-spiess (approved):

bobheadxi (approved):

PR #1793, opened 2023-11-17
fix: resolve editor when getting doc context

abeatrix:
image

Fixes an error that occurred when attempting to get the current VS Code document context. The previous methods for retrieving the active editor, document, and document text do not work with the new chat panel because when an user focuses on the webview panel we run the vscode.window.activeTextEditor, it would always returns `undefined`,  cause Cody unable to fetch local context.

This replaces the previous logic with new utility functions `getActiveEditorFromVisibleEditors`, `getCurrentVSCodeDoc`, and `getCurrentVSCodeDocText` that more robustly retrieve the current document context that is not the focused webview chat panel.

## Test plan

- Select some code in your editor
- In the new chat panel UI, type `/test` to run the test command
- Before the fix, Cody will tell you it doesn't have any context and cant identify the testing framework

### AFTER

https://github.com/sourcegraph/cody/assets/68532117/1a58afd3-8d81-468f-a389-e8c2fb558d7d

### BEFORE

https://github.com/sourcegraph/cody/assets/68532117/c3257b80-e57e-40cc-8cbf-00dc723cafed

### DIFF
--- a/lib/shared/src/chat/prompts/vscode-context.ts
+++ b/lib/shared/src/chat/prompts/vscode-context.ts
@@ -22,13 +22,45 @@ import { truncateText } from '../../prompt/truncation'
 import { answers, displayFileName } from './templates'
 import { createSelectionDisplayText, createVSCodeTestSearchPattern, isValidTestFileName } from './utils'
 
-// TODO bea move out of shared
+/**
+ * Gets the currently active text editor instance from the list of visible editors.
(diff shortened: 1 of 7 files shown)


abeatrix:
@umpox can i get a quick review from you, please 🙇‍♀️

umpox (approved):

PR #1794, opened 2023-11-17
Allow promise return in notifications

Strum355:
Silences all the "Promise returned in function argument where a void return was expected."

## Test plan

N/A type checker fixes


philipp-spiess (approved):
Seems reasonable :)

PR #1806, opened 2023-11-17
add feedback and bug templates

kalanchan:
add additional templates so we can capture feedback from slack and users

## Test plan
n/a


abeatrix (approved):
Cody says this looks good 👍

abeatrix (approved):

toolmantim:
Should we link to these in-product @kalanchan @taylorsperry, instead of the existing feedback discussions?

PR #1807, opened 2023-11-18
Update completion rate limit message when upgrade is available

DanTup:
This changes the completion rate limit message depending on whether the account can be upgrade (is dotCom and is not already Pro):

### Standard Message

Screenshot 2023-11-18 115923

### Upgrade message

Screenshot 2023-11-18 120022

This isn't complete yet, some questions below..

## Test plan

- There are new automated tests in `vscode\src\completions\inline-completion-item-provider.test.ts` to verify the correct message is selected based on auth and feature flags.


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -564,21 +567,31 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
      * error messages so every unexpected error is deduplicated by its message and rate limit errors
      * are only shown once during the rate limit period.
      */
-    private onError(error: Error | RateLimitError): void {
+    private async onError(error: Error | RateLimitError): Promise<void> {
```
DanTup, line 570: This PR ends up making some more places async because they need to check the GA feature flag (see comment on my other PR https://github.com/sourcegraph/cody/pull/1711#issuecomment-1816837362). This feels a little awkward and is likely temporary, but I'm unsure if there's a better way.

vscode/test/fixtures/mock-server.ts
```diff
@@ -88,11 +88,23 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
...
+            res.setHeader('x-ratelimit-limit', '12345')
+            res.sendStatus(429)
+            return
+        }
+
         const response = isFirstCompletion ? responses.firstCode : responses.code
         isFirstCompletion = false
         res.send(JSON.stringify(response))
     })
+    app.post('/.api/completions/code/triggerRateLimit', (req, res) => {
```
DanTup, line 104: I'm not sure if there's an existing pattern for things like this. This is a request added to the mock server that tests can call to trigger the rate limit for testing. I might need to do something similar for the Cody Pro flag once the other TODO about this is addressed.

vscode/test/integration/inline-completion.test.ts
```diff
@@ -0,0 +1,96 @@
...
+                "You've used all 12345 daily autocompletions. Usage will reset in less than a minute."
+            )
+        })
+
+        // TODO(dantup): We need to be able to simulate dotcom auth here, because upgrade
+        //  messages are never shown to non-dotCom and currently auth for tests is treated
+        //  as not dotCom (by the `if (!isDotComOrApp)` condition in `vscode\src\services\AuthProvider.ts`
+        //
+        //  If we change that path to `isDotComOrAppOrTest` it would potentially change behaviour of
+        //  other tests that are currently all considered _not_ dotComOrApp.
```
DanTup, line 80: I wasn't able to make this test work yet, because during integration tests we always take the non-dotCom path for auth (which means we'll never try to get the Cody Pro flag that is only on dotCom).

I think it would be useful of during the test auth we could mark whether to simulate dotCom or not, but I'm not sure of the best way.

One idea could be that the call to `ensureExecuteCommand('cody.test.token')` in integration tests that currently takes a SERVER_URL and VALID_TOKEN also takes a flag for whether to set auth as dotCom or not? Then these tests can easily test all scenarios.

DanTup (commented):
@abeatrix @philipp-spiess I have some questions here but I'm not sure who's best placed to answer them.

vscode/src/completions/inline-completion-item-provider.ts
philipp-spiess: Yeah it feels a bit awkward but I think it makes sense. Alternatively we could evaluate the feature flag earlier and pass the evaluated value

vscode/test/fixtures/mock-server.ts
philipp-spiess: We spawn the mock server as part of the same node process that runs the test AFAIK so maybe you can just export a function instead?

vscode/test/integration/inline-completion.test.ts
philipp-spiess: > One idea could be that the call to `ensureExecuteCommand('cody.test.token')` in integration tests that currently takes a SERVER_URL and VALID_TOKEN also takes a flag for whether to set auth as dotCom or not? Then these tests can easily test all scenarios.

I think this makes sense. Other idea I had was to use an env var when starting the extension

philipp-spiess (commented):
Sweet

vscode/src/completions/inline-completion-item-provider.ts
DanTup: If we did that, probably the best place would be to include it in `RateLimitError`, although it feels slightly awkward to attach a feature flag there.

Do you have any preference? Assuming the flag is short-lived and will be removed in the near future perhaps it's not all that important since whatever we can do can be removed.

vscode/test/fixtures/mock-server.ts
DanTup: Actually, I found we had an existing unit test for this rate limit error, so I've moved these tests to be unit tests alongside that, and removed these integration tests.

vscode/test/integration/inline-completion.test.ts
DanTup: These are no unit tests and no longer depend on dotcom, because we just directly mock the `userCanUpgrade` flag on `AuthStatus` (which itself is only ever set to `true` for dotCom).

DanTup:
@philipp-spiess  I think this should be good now?

philipp-spiess (approved):

PR #1808, opened 2023-11-18
Fix symf indexed dir

A2va:
Code search with symf fails on Windows because `getIndexDir()` concatenates the `absIndexedDir` with the `indexRoot`. To solve this, I simply base64 encoded the indexed dir.


jdorfman:
Thanks @A2va. We will review it ASAP :)

beyang:
This change would alter the names of the index directories on every machine to be the base64-encoded. It's nice to have the index directories mirror the directories that are indexed. It seems like it should be possible to do this on Windows by changing `path.resolve` to `filepath.resolve`

jdorfman:
Thanks @A2va!

PR #1818, opened 2023-11-19
Chat: Edit button to rename history chats

### LINKED ISSUE #1763: Users can rename chats
Follow up from #1650, adding the ability to rename the chat history on the side panel 

Image

deepak2431:
This PR closes #1763.

An edit button has been added to rename the chat from the history tree view. 

## Test plan

https://github.com/sourcegraph/cody/assets/44617923/eb12a9fa-5324-42de-8561-59663f1373cc

https://github.com/sourcegraph/cody/assets/44617923/e79a3364-aad2-468e-8b6c-bcc6bfea7a09


deepak2431:
Thanks @philipp-spiess for adding the reviewers :)
Mentioning @abeatrix @kalanchan too to take a look.

kalanchan:
the renaming on the side bar looks thank you! 

@deepak2431 could you also propagate the name change to the Cody chat panel as well? 

https://github.com/sourcegraph/cody/assets/51868853/c9f41182-686d-4f49-8bd3-9cb6f1778934

deepak2431:
Thanks for the review, @kalanchan.
Sure,  I will make those changes.

abeatrix:
@deepak2431 is it possible to add a test for the new change as well? 👀

vscode/package.json
```diff
@@ -500,6 +500,14 @@
         "icon": "$(layout-sidebar-left)",
         "when": "config.cody.experimental.chatPanel"
       },
+      {
+        "command": "cody.chat.history.edit",
+        "category": "Cody",
+        "title": "Edit Chat",
```
toolmantim, line 506:
```suggestion
"title": "Rename Chat",
```

vscode/package.json
```diff
@@ -500,6 +500,14 @@
...
+        "group": "Cody",
+        "icon": "$(wand)",
```
toolmantim, line 508:
```suggestion
        "icon": "$(edit)",
```

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -165,6 +165,15 @@ export class ChatPanelsManager implements vscode.Disposable {
         }
     }
 
+    public async editChatHistory(chatID: string): Promise<void> {
+        await vscode.window.showInputBox({ prompt: 'Enter new chat history name' }).then(async message => {
```
toolmantim, line 169:
```suggestion
        await vscode.window.showInputBox({ prompt: 'Enter new chat name' }).then(async message => {
```

Could we also pass in the current title as `value`?

toolmantim (commented):
Thanks @deepak2431, this is a good idea! Left some comments on labels and icons.

Non-blocking q: does the treeview provider allow you to rename it directly in the treeview, same as renaming files?

vscode/src/chat/chat-view/ChatPanelsManager.ts
deepak2431: Yes, we can pass that @toolmantim. I will implement this.

deepak2431:
> @deepak2431 is it possible to add a test for the new change as well? 👀

@abeatrix Yeah, I think it would be possible to add tests. I will add one for this.

deepak2431:
@toolmantim Thanks for the review! I will take care of the changes you suggested.

>Is it possible to add the "Rename Chat" action to the right-click context menu on the chat tab too? So you can do it from the chat view or the treeview.

This would be great to have. Yeah, it would be possible to display the Edit button there too.


> Does the treeview provider allow you to rename it directly in the treeview, same as renaming files?

Based on my findings, I couldn't see in the docs that Treeview provider allows to directly rename the rendered tree data view title. But I think there might be some other workaround to do this using WebView may be. Would ask @abeatrix to add her views on this too!

abeatrix:
@deepak2431 i think you're right, from what I've gathered that's currently not supported. 

There is also a GitHub issue opened for this feature request.

deepak2431:
Some unexpected Git issues happened while resolving merge conflicts. Working to fix it!

deepak2431:
All good to review now!
@toolmantim Made the minor changes you suggested.
@kalanchan The edit changes are now propagated to the Chat Panel.

deepak2431:
> > @deepak2431 is it possible to add a test for the new change as well? 👀
> 
> @abeatrix Yeah, I think it would be possible to add tests. I will add one for this.

@abeatrix I can see that currently, there's no test suite for the new ChatPanel provider where I could add the test for this change. So, I'm not sure about adding one :). 
Let me know if I am missing anything!

kalanchan (approved):
nice this is great thank you @deepak2431!! MVP!!

https://github.com/sourcegraph/cody/assets/51868853/5e6f9084-e684-48b5-bc12-60e79736ca82

abeatrix:
@deepak2431 There should be one now in main :)

toolmantim (approved):
🙌🏼

deepak2431:
> @deepak2431 There should be one now in main :)

Yes, I can see that now. It's been a while since I wrote e2e test, so I just wanted to ask after you run `pnpm test:e2e --debug`. After that, how do you start the record? I am trying to record it, but the selectors don't appear when I am in the Chat or try to sign in using the server URL and token manually. 
It would be great if you could share any docs I can reference or a short loom :)

deepak2431:
@abeatrix I have added the test, but it's getting stuck while filling the input box. I checked it a few times but couldn't see what's wrong there. Can you please have a look?

abeatrix:
@deepak2431 Thanks! Will take a look at the test and let you know if i found anything suspicious!

vscode/test/e2e/history-new-ui.test.ts
```diff
@@ -32,4 +32,16 @@ test('checks if chat history shows up in sidebar', async ({ page, sidebar }) =>
         page.getByText('Chat alongside your code, attach files, add additional context, and try out diff')
     ).toBeVisible()
     await page.getByRole('button', { name: 'New Chat', exact: true }).click()
+
+    // Create chat
+    await page.getByRole('tab', { name: 'New Chat' }).getByText('New Chat').click()
+
+    await page.getByRole('textbox', { name: 'New Chat' }).getByLabel('Message (type @ to attach files)').fill('Hey')
+    await page.getByRole('button', { name: 'Send message' }).locator('path').click()
```
abeatrix, line 40:
```suggestion
    await page.keyboard.press('Enter')
```

vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -134,6 +135,13 @@ export class ChatManager implements vscode.Disposable {
         await chatProvider.executeCustomCommand(title, type)
     }
 
+    public async editChatHistory(treeItem?: vscode.TreeItem): Promise<void> {
+        const chatID = treeItem?.id
```
abeatrix, line 139:
```suggestion
        const chatID = treeItem?.id
        const chatLabel = treeItem?.label as vscode.TreeItemLabel
```

If you pass the treeLabel to the function here, we can remove the `getTreeItemLabelByID` function from TreeViewProvider which does the same thing as getTreeItemByID. Wdyt?

vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -134,6 +135,13 @@ export class ChatManager implements vscode.Disposable {
...
+        if (chatID) {
+            await this.chatPanelsManager?.editChatHistory(chatID)
```
abeatrix, line 141:
```suggestion
            await this.chatPanelsManager?.editChatHistory(chatID, chatLabel?.label)
```

vscode/src/services/TreeViewProvider.ts
```diff
@@ -77,6 +77,11 @@ export class TreeViewProvider implements vscode.TreeDataProvider<vscode.TreeItem
         return this.treeNodes.find(node => node.id === chatID)
     }
 
+    public getTreeItemLabelByID(chatID: string): vscode.TreeItemLabel {
```
abeatrix, line 80:
```suggestion
```

Let's remove this?

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -168,6 +181,21 @@ export class ChatPanelsManager implements vscode.Disposable {
         }
     }
 
+    public async editChatHistory(chatID: string): Promise<void> {
```
abeatrix, line 184:
```suggestion
    public async editChatHistory(chatID: string, label: string): Promise<void> {
```

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -168,6 +181,21 @@ export class ChatPanelsManager implements vscode.Disposable {
...
+        await vscode.window
+            .showInputBox({
+                prompt: 'Enter new chat name',
+                value: this.treeViewProvider.getTreeItemLabelByID(chatID)?.label,
```
abeatrix, line 188:
```suggestion
                value: label,
```

abeatrix (changes requested):
Thanks for starting and adding the test! I found the issues and made some suggestions and left questions in line. Once these are sorted you can ping me again <3 

Thank you so much for the great work!

I could be wrong, but i don't think changing the title should alter the existing conversation because that would lead to incorrect chat history. @kalanchan @toolmantim I don't believe that's the expected behavior, is it? This should just rename the current chat title without making any changes to the actual chat history yea?

vscode/src/services/treeViewItems.ts
```diff
@@ -57,6 +58,19 @@ export function createCodyChatTreeItems(userHistory: UserLocalHistory): CodySide
     return chatTreeItems.reverse()
 }
 
+export async function updateChatHistoryLastInteractionMessage(chatID: string, message: string): Promise<void> {
```
abeatrix, line 61: I dont think this is the expected behavior where renaming the conversation would change the actual chat history. I believe this should work as the rename feature in chat GPT

vscode/test/e2e/history-new-ui.test.ts
```diff
@@ -32,4 +32,16 @@ test('checks if chat history shows up in sidebar', async ({ page, sidebar }) =>
         page.getByText('Chat alongside your code, attach files, add additional context, and try out diff')
     ).toBeVisible()
     await page.getByRole('button', { name: 'New Chat', exact: true }).click()
+
+    // Create chat
+    await page.getByRole('tab', { name: 'New Chat' }).getByText('New Chat').click()
+
+    await page.getByRole('textbox', { name: 'New Chat' }).getByLabel('Message (type @ to attach files)').fill('Hey')
```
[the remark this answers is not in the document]
deepak2431: @abeatrix Thanks for looking into the test. I tried this, but using `page.keyboard` doesn't add the chat to the history.
This is what it displays on testing:
`    Error: locator.hover: Target closed
    =========================== logs ===========================
    waiting for getByRole('treeitem', { name: 'Hey' }).locator('div').filter({ hasText: 'Hey' }).nth(3)`
    
  I think we need to use some other way to get the focus of the textBox

vscode/src/chat/chat-view/ChatManager.ts
deepak2431: yeah, you are right that can be done. I will update this to pass the chat label.

vscode/test/e2e/history-new-ui.test.ts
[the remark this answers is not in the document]
deepak2431: Interesting, not sure about it that might be the case too. I am using Mac M1 though.

vscode/src/services/treeViewItems.ts
deepak2431: I might be wrong here, but as we are using the last interaction human message as a treeview history label, we are just updating that here, nothing to the assistant response. So, that's what I thought of as a solution to this. 

Another [solution](https://github.com/sourcegraph/cody/pull/1817) I tried was to store the history treeview in the localStorage and then make the update or render the view accordingly. Still, in that case, I thought it was like repeating the same thing we are doing with the chatHistory, so I switched to this solution. 

Let me know WDYT. If there's any other solution which we can use here, I am open to hearing about it :)

deepak2431: Hmm, I see the issue here as when the conversation is clicked to view again, the last interaction message is also updated to the updated title.

deepak2431: One more solution I can think of is that instead of updating the chat history, we can keep a map of `<chatID, updatedTitle>` in the local history, and when we are creating the codyTreeItems or rendering the Chat panel, just check here first, if it is defined then use the updated title else use the last human message.  WDYT @abeatrix?

abeatrix: Hmmm instead of storing them separately, what do you think about adding a new title field (maybe optional?) for Transcript / TranscriptJSON so we can store everything together?

deepak2431: That also sounds good, even better so we don't need to deal with another localStorage key state. I will update the PR accordingly.

deepak2431: @abeatrix So, I tried this today to use an optional field under TransScript JSON, but there are a few issues I am seeing while implementing this solution:
1. The history gets updated after each transcript response update, and so does the chatID's value, which undefine the value of chatTitle.
2. When the response is aborted, the notifyTurnComplete function resets the title again to undefined.
3. During session restore, we create a new session ID in that case again chat title is reset to undefined.

I have a working solution to this, but it makes me chatTitle handle everywhere under MessageProvider and ChatPanelProvider where the history gets updated, and in many scenarios, I am not sure why it is being undefined :) 

I think it would be better to simply use <chatID, chatTitle> in a localStorage as a new state, and during the webview panel creation just use that if it is defined. 

Let me know what you think!

deepak2431: Update here: I implemented the second solution using a new localStorage state for <`chatID, chatTitle>`. This is working all good, tested thoroughly, and a bit easy to understand too so not a lot of handling of updates under the `MessageProvider` :)

abeatrix: Just came back from my pto, will take a look at your update first thing in the morning!

deepak2431: Sure, Take your time!

abeatrix: I tok

> So, I tried this today to use an optional field under TransScript JSON, but there are a few issues I am seeing while implementing this solution:
> 
> 1. The history gets updated after each transcript response update, and so does the chats history, which undefine the value of chatTitle.
> 2. When the response is aborted, the notifyTurnComplete function resets the title again to undefined.
> 3. During session restore, we create a new session ID in that case again chat title is reset to undefined.

Sorry for all the back and forth 😢 

I took a look at the PR and I think the issues you mentioned above is due to the old chat history setup, so you shouldn't run into the same issue with the new [ChatHistoryManager](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/ChatHistoryManager.ts?L20).  With that, we should be able to set the optional chatTitle per Transcript (like chatModel in Transcript) and have everything in one place! Does that sound good to you? Or do you think storing them separately is more ideal?

deepak2431: @abeatrix No problem. This PR took longer than expected, I think, due to a different history setup and chat manager :)
So, I saw the new ChatHistoryManager. But using that still makes it undefined, and I am a bit unsure how this new SimpleChatPanelProvider replaces the old one. It would be great if you could brief me about it.

For you to have a look, I have updated this PR with the implemented solution of using `chatTitle` with the Transcript. And I agree this is a good way to do it, too.

deepak2431: I guess this solution works fine with `SimpleChatContext` enabled 👀

toolmantim:
@abeatrix @deepak2431 yeah, we wouldn't want to alter the chat transcript. A new property sounds right.

vscode/src/services/treeViewItems.ts
abeatrix: > So, I saw the new ChatHistoryManager. But using that still makes it undefined, and I am a bit unsure how this new SimpleChatPanelProvider replaces the old one. It would be great if you could brief me about it.

Awesome! It's looking good! And yea sorry for all the massive changes, I also find it confusing 😅  But ultimately we str aiming to retire the current chatPanelProvider and move to SimpleChatPanelProvider. Instead of adding a separate function to handle chat title in history manager, I was thinking of storing it in Transcript. Here is an example of what i've envisioned: https://github.com/sourcegraph/cody/pull/1997 --the change in my PR doesn't work correctly as renaming the chat would generate a new chat history item because of how we are still using the old way to fetch and store history, but I hope this gives you an idea on how to get it to work with the new chat history manager and  SimpleChatPanelProvider! What do you think about this approach?

deepak2431:
@abeatrix I am waiting for this #2059 to merge so I can update my changes according to the updated History Manager. 

Also, some build errors are happening in the `main` branch. Can you look into that, please? 
iScreen Shoter - Code - 231204232413
iScreen Shoter - Code - 231204232400

deepak2431:
@abeatrix Updated as per the discussion. It works all good now. I saw that the `new history ui` test had been disabled for some reason, so I didn't updated that.

Also, I saw some issues unrelated to this PR, such as how the chatID is being created for new chat and when the chat is revived from the History. So, I would ask you to look in your free time. You might also get a sense of it while reviewing this PR :)

deepak2431:
@philipp-spiess Can you have a look at it please whenever you have some time?

vscode/src/chat/MessageProvider.ts
```diff
@@ -100,6 +100,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
     protected platform: Pick<PlatformContext, 'recipes'>
 
     protected chatModel: string | undefined = undefined
+    protected chatTitle: string | undefined = 'init'
```
philipp-spiess, line 103: This looks like a strange default, should we use "Untitled" instead?

philipp-spiess (approved):
LGTM! One comment inline; let me try to run the tests

vscode/src/chat/MessageProvider.ts
deepak2431: yeah, agree with you. Let me make the change.

deepak2431:
@philipp-spiess Thanks for the review. Just made the suggested changes. Do have a look :)

philipp-spiess:
@deepak2431 Awesome stuff as usual :)

PR #1829, opened 2023-11-20
Enable support for legacy event logger in non-vscode clients

Strum355:
This is needed in some places in the new logger cc @bobheadxi 

## Test plan

Tested with telemetryLevel = 'agent' and it doesnt blow up :slightly_smiling_face:

### DIFF
--- a/lib/shared/src/telemetry/EventLogger.ts
+++ b/lib/shared/src/telemetry/EventLogger.ts
@@ -26,10 +26,25 @@ export class EventLogger {
     ) {
         this.gqlAPIClient = new SourcegraphGraphQLAPIClient(this.config)
         this.setSiteIdentification().catch(error => console.error(error))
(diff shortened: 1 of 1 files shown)


bobheadxi (approved):
I'm not sure we need it, because we use the GraphQL client directly in the new telemetry when attempting back-compat: https://sourcegraph.com/github.com/sourcegraph/cody@ae05593a09046848c76b5a2ca65f3313c6a001d1/-/blob/lib/shared/src/sourcegraph-api/telemetry/GraphQLTelemetryExporter.ts?L90-93

but approving to unblock if you need it :)

Strum355:
Ah, I may have misunderstood then :sweat: if you dont need it, then we can leave this PR until/in case we do need it later

Strum355:
I think we need to merge this anyways, after some brief discussion with Kelsey, we still need to emit legacy events for the time being until the v2 transition is mostly complete

Issue #1830, opened 2023-11-20
bug: chat doesn't complete just spins


chwarwick:
### Version

0.17.1700492803

### Describe the bug

I am able to use Cody just fine up to a certain point, then when I give it a command or ask it a question in inline chat it spins and does nothing.  This usually happens after chatting with it for a bit or asking it document multiple functions.

### Expected behavior

Chat should always complete either successfully or with an error.

PR #1831, opened 2023-11-20
Fix: Chat spins forever after errors

chwarwick:
When a chat request errors a variable is set to capture that, so that when the done event arrives it will not call on complete.  However the chat client is reused for the next chat message leaving `errorEncountered` to true which would stop the promise from resolving unless it were to fail again.  

This change resets  `errorEncountered` back to false when receiving the done message for errored requests.

resolves https://github.com/sourcegraph/cody/issues/1830
## Test plan
Manual testing
Attempted a chat that would exceed token limit - got an error
Tried another chat and it was successful 
Screenshot 2023-11-20 at 1 09 31 PM


abeatrix (approved):

kalanchan:
GOGOGOGOOG @chwarwick !!!!

PR #1832, opened 2023-11-20
WIP

philipp-spiess:


agent/src/cli/evaluate-autocomplete/evaluate-autocomplete.ts
```diff
@@ -216,7 +218,96 @@ async function evaluateGitLogStrategy(
...
+            .toString()
+            .split('\n')
+            .map(string => string.split(' ')[0])
+            .slice(0, options.testCount)
+            .filter(Boolean)
+        // Reverse the commits list so the first element is the oldest commit
+        commits.reverse()
+        for (const commit of commits) {
+            try {
+                execSync(`git checkout ${commit}`, { cwd: workspace })
```
olafurpg, line 236: It would be nice to avoid git commands that change the worktree. You can use `git show` and `git diff` without changing the worktree like this

```
# Read text contents of a file at a specific revision
❯ git show main~100:agent/src/index.ts
# Diff between two revisions of a file
❯ git diff main~100:agent/src/index.ts main:agent/src/index.ts
```

agent/src/cli/evaluate-autocomplete/evaluate-autocomplete.ts
```diff
@@ -302,13 +393,18 @@ async function evaluateBfgStrategy(
...
 // TODO: rename to remove fixture from this interface
-interface AutocompleteFixture {
+type AutocompleteFixture = {
     content: string
     filePath: string
-    capture: QueryCapture
     client: MessageHandler
     document: scip.scip.Document
-}
+} & (
```
olafurpg, line 401: This type is getting complicated making it more difficult to add more information here. Can we keep this as a simple interface or turn it into a class somehow? Non-blocking comment, I can take a stab at simplifying it myself if I need to make changes

olafurpg (approved):
LGTM 👍🏻 Very nice!

agent/src/cli/evaluate-autocomplete/evaluate-autocomplete.ts
philipp-spiess: Yeah that would be nice but this means I have to read all files by parsing the git show output hmm I will try it out

olafurpg: The current solution is totally fine, feel free to merge. Was more an FYI

agent/src/cli/evaluate-autocomplete/evaluate-autocomplete.ts
philipp-spiess: yeah I’m thinking we should probably keep this closer to what the autocomplete interface is offering (so just a text and a position) and move the tree-sitter specific range part into a parent function

olafurpg: Agreed. I can take a stab at cleaning it up separately

agent/src/cli/evaluate-autocomplete/evaluate-autocomplete.ts
philipp-spiess: Sounds good :)

Issue #1809, opened 2023-11-19
bug: chats w/ new chat UI fail to load when opened quickly in succession


toolmantim:
### Version

0.16.0

### Describe the bug

Rapidly opening chats in succession causes some tabs to fail to load:

https://github.com/sourcegraph/cody/assets/153/da08e169-58c8-4dcf-992c-17bf73f836b7

### Expected behavior

All chat tabs load successfully

Issue #1757, opened 2023-11-15
VSCE Bug: Sidebar during initial startup


abeatrix:
In the new chat panel UI, if you restart VS Code and then click on the Cody icon during the start up stage, you will see the sidebar is opened and then closed again, before you can click on it to open the sidebar again.

https://github.com/sourcegraph/cody/assets/68532117/918de800-61d9-47b0-9c8a-78f1be830331

iswarmondal:
Hey @abeatrix👋🏻 ,

I see in version `0.16.0` I am unable to reproduce this issue. So is this issue still persists from your side?

abeatrix:
@iswarmondal sorry for not updating the issue properly, as you mentioned, this issue has been resolved in the linked issues!

PR #1835, opened 2023-11-20
fix: race condition on new panel creation

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1809

This PR  moves the command registration step to chat manager, and new condition to check if the current panel has been activated or not. If it's not activated, display a warning and ask users to wait for the current panel to be activated before opening a new one to prevent race condition described in the issue https://github.com/sourcegraph/cody/issues/1809

### DEMO

https://github.com/sourcegraph/cody/assets/68532117/1619b1de-bc5b-40e9-9598-66c708d53f5e

## Test plan

Try to reproduce the steps described in the original issue to confirm.

RE: https://github.com/sourcegraph/cody/issues/1757

I also cannot reproduce the issue mentioned in https://github.com/sourcegraph/cody/issues/1757 with this build

https://github.com/sourcegraph/cody/assets/68532117/cbb9d307-3946-48c9-87d8-d3faa2df8e93

### DIFF
--- a/vscode/src/chat/chat-view/ChatPanelsManager.ts
+++ b/vscode/src/chat/chat-view/ChatPanelsManager.ts
@@ -87,7 +87,8 @@ export class ChatPanelsManager implements vscode.Disposable {
      */
     public async createWebviewPanel(chatID?: string, chatQuestion?: string): Promise<ChatPanelProvider> {
         if (this.activePanelProvider && !this.activePanelProvider?.webviewPanel?.active) {
-            void vscode.window.showInformationMessage(
+            this.activePanelProvider.webviewPanel?.reveal()
+            void vscode.window.showErrorMessage(
                 'Please wait for the current panel to be activated before opening a new one.'
             )
             return this.activePanelProvider
--- a/vscode/src/chat/chat-view/ChatPanelsManager.ts
+++ b/vscode/src/chat/chat-view/ChatPanelsManager.ts
@@ -88,9 +88,7 @@ export class ChatPanelsManager implements vscode.Disposable {
     public async createWebviewPanel(chatID?: string, chatQuestion?: string): Promise<ChatPanelProvider> {
         if (this.activePanelProvider && !this.activePanelProvider?.webviewPanel?.active) {
             this.activePanelProvider.webviewPanel?.reveal()
-            void vscode.window.showErrorMessage(
-                'Please wait for the current panel to load.'
-            )
+            void vscode.window.showErrorMessage('Please wait for the current panel to load.')
             return this.activePanelProvider
         }
 
--- a/vscode/src/chat/chat-view/ChatPanelsManager.ts
+++ b/vscode/src/chat/chat-view/ChatPanelsManager.ts
@@ -45,12 +45,6 @@ export class ChatPanelsManager implements vscode.Disposable {
             vscode.window.registerTreeDataProvider('cody.commands.tree.view', new TreeViewProvider('command'))
         )
 
-        // Register Commands
-        this.disposables.push(
-            vscode.commands.registerCommand('cody.chat.panel.new', async () => this.createWebviewPanel()),
-            vscode.commands.registerCommand('cody.chat.panel.restore', async (id, chat) => this.restorePanel(id, chat))
-        )
-
         // Register config change listener
         this.onConfigurationChange = options.contextProvider.configurationChangeEvent.event(async () => {
             // When chat.chatPanel is set to true, the sidebar chat view will never be shown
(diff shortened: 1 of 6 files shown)


vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -92,6 +86,23 @@ export class ChatPanelsManager implements vscode.Disposable {
      * Creates a new webview panel for chat.
      */
     public async createWebviewPanel(chatID?: string, chatQuestion?: string): Promise<ChatPanelProvider> {
+        if (this.activePanelProvider && !this.activePanelProvider?.webviewPanel?.active) {
+            this.activePanelProvider.webviewPanel?.reveal()
+            void vscode.window.showErrorMessage(
+                'Please wait for the current panel to be activated before opening a new one.'
```
kalanchan, line 92:
```suggestion
                'Please wait for the current panel to load.'
```

kalanchan: nit: might want to put in a timed auto dismiss

kalanchan (approved):
race condition fixed! 

I have some comments around the UX - I think it would be better if the default is grouped cascading tabs instead of individual panels. It would save a lot of screen real estate when users click on a new chat. @toolmantim for thoughts

also some weird UI bug when theres too many panels open, I think due to responsiveness it pushes the editor to the left and creates a grey bar 

CleanShot 2023-11-20 at 15 07 19@2x

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -92,6 +86,21 @@ export class ChatPanelsManager implements vscode.Disposable {
      * Creates a new webview panel for chat.
      */
     public async createWebviewPanel(chatID?: string, chatQuestion?: string): Promise<ChatPanelProvider> {
+        if (this.activePanelProvider && !this.activePanelProvider?.webviewPanel?.active) {
+            this.activePanelProvider.webviewPanel?.reveal()
+            void vscode.window.showErrorMessage('Please wait for the current panel to load.')
+            return this.activePanelProvider
+        }
```
toolmantim, line 93: This is nicer than failing to load, but is there a way to make it open the new chat for the user after the current one is finished loading? Perhaps via setTimeout etc?

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -92,6 +86,21 @@ export class ChatPanelsManager implements vscode.Disposable {
      * Creates a new webview panel for chat.
      */
     public async createWebviewPanel(chatID?: string, chatQuestion?: string): Promise<ChatPanelProvider> {
+        if (this.activePanelProvider && !this.activePanelProvider?.webviewPanel?.active) {
+            this.activePanelProvider.webviewPanel?.reveal()
+            void vscode.window.showErrorMessage('Please wait for the current panel to load.')
```
toolmantim, line 91:
```suggestion
void vscode.window.showErrorMessage(`Could not open chat ${title}. Please wait for the current chat to load and try again.`)
```

toolmantim:
@kalanchan yeah those tabs are a problem, I think we need to do more than just using `vscode.ViewColumn.Beside`. 

Here's some related Slack convos:
* https://sourcegraph.slack.com/archives/C05AGQYD528/p1700438510093699?thread_ts=1700418396.014199&cid=C05AGQYD528
* https://sourcegraph.slack.com/archives/C05AGQYD528/p1699494652493029?thread_ts=1699492198.789529&cid=C05AGQYD528

We need to figure out what is possible API wise w/ @abeatrix’s help, and create that as a new issue.

vscode/src/chat/chat-view/ChatPanelsManager.ts
abeatrix: I wanted to do that initially but then it doesn't seem to scale because if a user click to open 10 different new chats, they will then be opened at different times and no way to cancel it so this seems to be the easiest way? wdyt?

toolmantim: Hmm… yeah. Thinking through it again, I think the nicest would be something like:

```
vscode.window.withProgress(
  {
    location: vscode.ProgressLocation.Notification,
    title: 'Loading chat…',
    cancellable: true,
  },
  ...
)
```

abeatrix: @toolmantim I've made a new change in this PR https://github.com/sourcegraph/cody/pull/1836 
This way, multiple panels can be opened at once, and the webview will auto-reload every 2seconds when it failed to open initially

PR #1836, opened 2023-11-21
fix: auto reload chat panel

abeatrix:
Follow up from https://github.com/sourcegraph/cody/pull/1835

Instead of checking if the current chat panel is active or not, this PR updates the webview to reload automatically when it failed to load after 2000ms so users can have multiple chat panels opened, where some of them might be stuck at the loading stuck when it was first opened, but they will reload automatically.

- Add delays before sending webview ready events to prevent premature sending.

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/37480645-0e71-4c2c-82a9-ab88d65834e1

### DIFF
--- a/vscode/webviews/App.tsx
+++ b/vscode/webviews/App.tsx
@@ -191,7 +191,9 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
         // This mean webview is not ready
         if (isWebviewReady && !authStatus && !config && !view) {
             setIsWebviewReady(false)
-            vscodeAPI.postMessage({ command: 'ready' })
+            setTimeout(() => {
+                vscodeAPI.postMessage({ command: 'ready' })
+            }, 2000)
         }
         return <LoadingPage />
     }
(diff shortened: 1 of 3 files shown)


vscode/webviews/App.tsx
```diff
@@ -135,13 +132,16 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
 
     useEffect(() => {
         // Notify the extension host that we are ready to receive events
-        vscodeAPI.postMessage({ command: 'ready' })
+        setTimeout(() => {
```
toolmantim, line 135: Do we need to clear these timeouts on dispose?

toolmantim (approved):
Neat solution! 🙌🏼

vscode/webviews/App.tsx
abeatrix: I just realized this timeout doesn't matter actually, it's the one in `return <LoadingPage />` that is doing the reload 😅  Please let me know if the latest commit doesn't work for you 🙇‍♀️

toolmantim (approved):
Tested again 👍🏼

PR #1839, opened 2023-11-21
Autocomplete: make javascript completion intent queries more precise

valerybugakov:
## Context

- JS completion intent queries are too broad because of the "any" matches used in queries: `(arguments (_) @argument)`. It leads to misleading completion intent analytics results. E.g., too many node type combos fall into the "argument" completion intent even though they are deeply nested inside the `arguments` node. This PR limits the `argument` completion intent query to a few popular node types. We can expand the list later based on the analytics data.

Example of the misleading match for the `argument` completion intent:
```ts
myFunction(() => {
  const value = {
    █
  }
```

- Adds a query for the `pair.value` completion intent.
- Part of https://github.com/sourcegraph/cody/issues/1455

## Test plan

Updated snapshot tests.

### DIFF
--- a/vscode/src/tree-sitter/queries.ts
+++ b/vscode/src/tree-sitter/queries.ts
@@ -18,6 +18,7 @@ export const intentPriority = [
     'arguments',
     'import.source',
     'comment',
+    'pair.value',
     'argument',
     'parameter',
     'parameters',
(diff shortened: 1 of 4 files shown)


philipp-spiess (approved):
Noice

PR #1843, opened 2023-11-21
Autocomplete: log `insertText` for DotCom users

valerybugakov:
## Context

- Adds `insertText` to analytics events for DotCom users only.
- Adds a unit test to check that the `insertText` is only included for DotCom users.

## Test plan

CI

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/analytics.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/analytics.test.ts
@@ -61,6 +61,8 @@ describe('[getInlineCompletions] completion event', () => {
                 "items": [
                   {
                     "charCount": 30,
+                    "insertText": "console.log(bar)
+              return false}",
                     "lineCount": 2,
                     "lineTruncatedCount": 0,
                     "nodeTypes": {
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
Nice!

valerybugakov:
I'll wait for more [potential input](https://sourcegraph.slack.com/archives/C05AGQYD528/p1700542832836439) in Slack before merging this one.

philipp-spiess (approved):

vscode/src/completions/get-inline-completions-tests/analytics.test.ts
```diff
@@ -2,13 +2,41 @@ import { omit } from 'lodash'
...
+    workspace: {
+        ...vsCodeMocks.workspace,
+        getConfiguration() {
+            return {
+                get(key: string) {
+                    switch (key) {
+                        case 'cody.debug.filter':
+                            return '.*'
+                        case 'cody.serverEndpoint':
+                            return mockServerEndpointRef.endpoint
```
philipp-spiess, line 28: We don't use the confifguraiton object anymore to set the serverEndpoint which means that in prod, this will always retrun the dotcom server endpoint.

PR #1845, opened 2023-11-21
Autocomplete: Move decision which StarCoder model to use to the server

philipp-spiess:
Follow up to sourcegraph/sourcegraph#58404

This PR moves the client logic of picking the right StarCoder model to the server.

## Test plan

Screenshot 2023-11-21 at 13 24 21


rafax (approved):
LGTM, I think PR (in isolation) removes the ability to switch between MT and ST models on the client (but doesn't add server switching)? If so, it might make sense to include that for future readers.

philipp-spiess:
> PR (in isolation) removes the ability to switch between MT and ST models on the client (but doesn't add server switching)? If so, it might make sense to include that for future readers.

Correct but since this was only enabled for dotcom and dotcom has already updated, users will not have an observable effect.

PR #1847, opened 2023-11-21
Agent: add option to emit CSV file for `evaluate-autocomplete`

olafurpg:
Previously, `evaluate-autocomplete` only generated snapshots to process the results. This PR adds a new CSV writer so that you can process/analyze the results with CSV tooling like xsv.

## Test plan

Ran large scale evaluation with the command
```
❯ pnpm agent evaluate-autocomplete --evaluation-config ~/dev/sourcegraph/bfg-evaluation/evaluation.json --test-count 400
```


agent/src/cli/evaluate-autocomplete/triggerAutocomplete.ts
```diff
@@ -0,0 +1,88 @@
...
+    document: AutocompleteDocument
+
+    range: vscode.Range
+    modifiedContent: string
+    removedContent: string
+    position: vscode.Position
+    emptyMatchContent: string
+}
+
+export async function triggerAutocomplete(parameters: AutocompleteParameters): Promise<void> {
```
philipp-spiess, line 21: Nice cleanup

vscode/src/completions/artificial-delay.ts
```diff
@@ -74,7 +74,9 @@ export function getArtificialDelay(
         userMetrics.currentLatency += featureFlags.user ? defaultLatencies.user : 0
     }
 
-    logDebug('CodyCompletionProvider:getLatency', `Delay added: ${total}`)
+    if (total > 0) {
+        logDebug('CodyCompletionProvider:getLatency', `Delay added: ${total}`)
+    }
```
philipp-spiess, line 79: Let's rm this, we log this as part of the suggested event now anyways. I find this very annoying

agent/src/cli/evaluate-autocomplete/AutocompleteDocument.ts
```diff
@@ -0,0 +1,194 @@
...
+     * @param diagnostic optional message to include with the formatted string
+     */
+    public format(range: vscode.Range, diagnostic?: string): string {
+        const line = this.lines[range.start.line]
+        const indent = ' '.repeat(range.start.character)
+        const length =
+            range.start.line === range.end.line
+                ? range.end.character - range.start.character
+                : line.length - range.start.character
+        const carets = length < 0 ? '<negative length>' : '^'.repeat(length)
```
philipp-spiess, line 114: BTW when using vscode.Range:

>	 * It is guaranteed that {@link Range.start start}.isBeforeOrEqual({@link Range.end end})

so negative length is not possible :)

philipp-spiess (approved):
Awesome!

agent/src/cli/evaluate-autocomplete/AutocompleteDocument.ts
olafurpg: I don't think we enforce this in the shim!

vscode/src/completions/artificial-delay.ts
olafurpg: +1 to removing it. This diff here fixed my biggest pain

Issue #1841, opened 2023-11-21
bug: new chat send button isn't circular and distorts on input change


toolmantim:
### Version

main

### Describe the bug

The new send button isn't quite a circle:

<img width="520" alt="Screenshot 2023-11-21 at 3 12 55 pm" src="https://github.com/sourcegraph/cody/assets/153/a6c3a6de-2098-471f-9ba4-cf3aee3f3878">

And changes size as the text field changes size:

https://github.com/sourcegraph/cody/assets/153/a833104f-e738-4c73-a7a5-12573d31e9eb

### Expected behavior

Send button should always be circular, and stay aligned with the bottom of the text input

PR #1850, opened 2023-11-21
fix: chat submit button style

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1841

image

Address the following issues:
- The new send button isn't quite a circle
- And changes size as the text field changes size

https://github.com/sourcegraph/cody/assets/68532117/df667230-c44b-4afa-bfa7-caf2d0a88322

## Test plan

- [x] Send button should always be circular
- [x] stay aligned with the bottom of the text input

image

### DIFF
--- a/lib/ui/src/Chat.module.css
+++ b/lib/ui/src/Chat.module.css
@@ -19,10 +19,11 @@
 .text-area-container {
     width: 100%;
     display: flex;
+    align-items: flex-end;
 }
 
 .chat-input-container {
-    width: 96%;
+    width: 100%;
     height: 100%;
     position: relative;
 }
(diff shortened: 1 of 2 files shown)


lib/ui/src/Chat.module.css
```diff
@@ -19,10 +19,11 @@
 .text-area-container {
     width: 100%;
     display: flex;
+    align-items: flex-end;
 }
 
 .chat-input-container {
-    width: 96%;
+    width: 100%;
```
toolmantim, line 26: I think you can do this with `flex: 1` too

toolmantim (approved):
🙌🏼

Issue #1837, opened 2023-11-21
bug: @-file can return different results for the same query


toolmantim:
### Version

0.16.0

### Describe the bug

https://github.com/sourcegraph/cody/assets/153/6428e5af-4968-40a7-a9c0-8899af256f7b

### Expected behavior

The same results are shown each time

PR #1851, opened 2023-11-22
update: sort file results

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1837

When we do a search using the VS Code API, it will always return the first n results found so the results could be different for the same keyword search. This caused confusion stated in the issue linked above. To  workaround that, this PR updated to get 50 results instead of 10 results from the API, and then sort the results alphabetically to get the first 20 results. This helps get the same set of results of the exact keyword search.

Summary:
- Increase max file search results from 10 to 20
- Increase max symbol search results from 10 to 50
- Sort file results alphabetically by path 
- Take top 20 sorted file results and merged symbol results

Here is an example of the results returned by the VS Code API for the same keyword search:
Screenshot 2023-11-21 at 4 04 21 PM

Here is an example of the result `@lib` after the updated workaround:

image

## Test plan

- Search for the same keyword multiple times, then results should come back in the same order

### DEMO

https://github.com/sourcegraph/cody/assets/68532117/ba02fd4c-e411-4ab5-b482-6458055167b2

### DIFF
--- a/vscode/src/chat/chat-view/ChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/ChatPanelProvider.ts
@@ -288,13 +288,25 @@ export class ChatPanelProvider extends MessageProvider {
 
         const debouncedContextFileQuery = debounce(async (query: string): Promise<void> => {
             try {
-                const MAX_RESULTS = 10
-                const fileResultsPromise = getFileContextFile(query, MAX_RESULTS)
+                const MAX_RESULTS = 20
+                const fileResultsPromise = getFileContextFile(query, 50)
                 const symbolResultsPromise = getSymbolContextFile(query, MAX_RESULTS)
 
                 const [fileResults, symbolResults] = await Promise.all([fileResultsPromise, symbolResultsPromise])
-                const context = [...new Set([...fileResults, ...symbolResults])]
-
+                // Sort results by alphabatic and the length of the director/file path
+                const sortedFileResults = fileResults
+                    .sort((a, b) => {
+                        if (a.fileName < b.fileName) {
+                            return -1
+                        }
+                        if (a.fileName > b.fileName) {
+                            return 1
+                        }
+                        return 0
+                    })
+                    .slice(0, MAX_RESULTS)
+                // Return the merged results
+                const context = [...new Set([...sortedFileResults, ...symbolResults])]
                 await this.webview?.postMessage({
                     type: 'userContextFiles',
                     context,


vscode/src/chat/chat-view/ChatPanelProvider.ts
```diff
@@ -288,13 +288,25 @@ export class ChatPanelProvider extends MessageProvider {
 
         const debouncedContextFileQuery = debounce(async (query: string): Promise<void> => {
             try {
-                const MAX_RESULTS = 10
-                const fileResultsPromise = getFileContextFile(query, MAX_RESULTS)
+                const MAX_RESULTS = 20
+                const fileResultsPromise = getFileContextFile(query, 50)
```
toolmantim, line 292: Does changing these numbers affect performance? And by how much? Be interested to know if why these numbers were chosen, and if we could go higher…

abeatrix: started with 10 which was the number used in the API doc example lol that didn't work well with what we want to archive,  so i tried going up slowly and found 50 seems to be reasonable and more consistent when the keyword is 3 characters+ . It's still hard to be consistent with 50 when the keyword is shorter than that though.

vscode/src/chat/chat-view/ChatPanelProvider.ts
```diff
@@ -288,13 +288,25 @@ export class ChatPanelProvider extends MessageProvider {
...
                 const symbolResultsPromise = getSymbolContextFile(query, MAX_RESULTS)
 
                 const [fileResults, symbolResults] = await Promise.all([fileResultsPromise, symbolResultsPromise])
-                const context = [...new Set([...fileResults, ...symbolResults])]
-
+                // Sort results by alphabatic and the length of the director/file path
+                const sortedFileResults = fileResults
```
toolmantim, line 297: Should we use something like https://www.npmjs.com/package/fuzzysort instead?

abeatrix: hmmm but the results are already returned from the vs code API, so this is just doing a fuzzy search on a set of results that were not returned from a fuzzy search? I need to give it a try next week to see how it works 👀

vscode/src/chat/chat-view/ChatPanelProvider.ts
```diff
@@ -288,13 +288,25 @@ export class ChatPanelProvider extends MessageProvider {
...
+                    .sort((a, b) => {
```
kalanchan, line 298: this might not be the most optimal way of sorting file paths, I think string comparison will give really inconsistent results because it's based on how that string transpiles back down to a number. 

EX 
- `console.log("90" > "100") === true`
- `console.log("src/docs/abc.ts" > "src/xba.ts") === false` <-- this would sort to `["src/docs/abc.ts", "src/xba.ts"]`


this can be a future improvement, but we'll need to use something like [sort-paths](https://www.npmjs.com/package/sort-paths) so that it accounts for which part of the path to prioritize

kalanchan (commented):
The changes still weren't that consistent, we might need to add some more debouncing or the file search API is inconsistent with what it returns. 

https://github.com/sourcegraph/cody/assets/51868853/9c95ce1f-ecd8-4f15-93cf-20834eb5b4d0

abeatrix:
@kalanchan I don't think you are in the correct dev mode that is built from this branch, are you? In your video, it shows you're on my branch, so you'll have to start the debug mode from there to open a different repo to see changes from this branch, as vs code doesn't allow you to open the same repo in separate windows.

abeatrix:
@kalanchan but yea I don't think it would work consistently unless we remove the search result limit and just get all the results that matches the keyword, and then do the ranking and sorting using the packages you and @toolmantim  suggested before displaying them. I can work on that on Monday! Thanks for the review and the suggestions 🙇‍♀️

kalanchan:
> @kalanchan I don't think you are in the correct dev mode that is built from this branch, are you? In your video, it shows you're on my branch, so you'll have to start the debug mode from there to open a different repo to see changes from this branch, as vs code doesn't allow you to open the same repo in separate windows.

@abeatrix i did build it off your branch, but the `cody` workspace did mess it up! 

got much better results using the `sourcegraph` repo :) 

https://github.com/sourcegraph/cody/assets/51868853/67d8e47b-e3a5-46ef-96dc-43f4b3fbccf0

kalanchan:
closes https://github.com/sourcegraph/cody/issues/1837

Issue #1521, opened 2023-10-27
VSCE: Re-evaluate experimental features in settings menu


abeatrix:
Context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1698804897903669

@tooltim:

> There are too many experimental features in the settings menu

![image](https://github.com/sourcegraph/cody/assets/68532117/beae4486-aa64-4ee5-b59d-4c6933db22a2)

- [x] Move `Chat Pre-Instruction` out of Experimental
- [x] Move `Editor Title Icon` out of Experimental
- [x] Keep `Chat Suggestions` in Experimental
- [x] Move custom commands to beta

Follow up:
- [ ] sun set inline chat 
- [ ] remove old side bar chat

PR #1855, opened 2023-11-22
change: custom commands to beta

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1521

This PR includes the following changes:
- [x] Move `Command Code Lenses` out of Experimental
- [x] Keep `Chat Suggestions` in Experimental
- [x] Move `Custom Commands` to Beta

Also update Chat Pre-Instructio on config change

## Test plan

Check if the features above are showing up with the correct label

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -971,11 +971,11 @@
           }
         },
         "cody.inlineChat.enabled": {
-          "order": 6,
+          "order": 10,
           "title": "Cody Inline Chat",
           "type": "boolean",
           "markdownDescription": "Enables asking questions and requesting code changes directly from within the code editor. Use the + button next to any line of code to start an inline chat.",
-          "default": true
+          "default": false
         },
         "cody.editorTitleCommandIcon": {
           "order": 7,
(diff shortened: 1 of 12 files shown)


vscode/package.json
```diff
@@ -971,11 +971,11 @@
           }
         },
         "cody.inlineChat.enabled": {
-          "order": 6,
+          "order": 10,
           "title": "Cody Inline Chat",
           "type": "boolean",
           "markdownDescription": "Enables asking questions and requesting code changes directly from within the code editor. Use the + button next to any line of code to start an inline chat.",
-          "default": true
+          "default": false
```
kalanchan, line 978: this also turns off inline chat on the old chat interface right

abeatrix: @kalanchan that's done in another PR already https://github.com/sourcegraph/cody/pull/1797

kalanchan: ooh gotcha, so what does this default toggle do?

abeatrix: this change is already landed in main, it's showing up in diff because this PR has become stale for awhile 😆 Will merge it in a bit so this will be less confusing https://github.com/sourcegraph/cody/blob/main/vscode/package.json#L973-L979

abeatrix: @kalanchan just updated!

kalanchan (approved):
overall looks good, just minor nit - its seems a bit heavy to always rename the variables when we want to move out of experimental, i think it would be good to refactor it in the future so that it just checks if an `experimental` key is true/false instead

Custom commands ✅ 
CleanShot 2023-11-28 at 13 32 03@2x

code lenses ✅ 
chat suggestions ✅ 
CleanShot 2023-11-28 at 13 37 37@2x

PR #1856, opened 2023-11-22
add: transcript syncing

abeatrix:
feat: add transcript syncing with Sourcegraph Cloud

Added chat transcript syncing with dotCom behind a feature flag.

- This adds the ability to sync th chat transcript with Sourcegraph Cloud if the Cody endpoint is on DomCom. 
- Added a new feature flag 'cody-chat-transcript-sync' controls this behavior.

## Test plan

Waiting for @akalia25 's team to confirm where we should log the transcript

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -30,7 +30,7 @@ import { logDebug, logError } from '../log'
 import { FixupTask } from '../non-stop/FixupTask'
 import { AuthProvider, isNetworkError } from '../services/AuthProvider'
 import { localStorage } from '../services/LocalStorageProvider'
-import { telemetryService } from '../services/telemetry'
+import { syncTranscript, telemetryService } from '../services/telemetry'
 import { telemetryRecorder } from '../services/telemetry-v2'
 import { TestSupport } from '../test-support'
 
@@ -748,6 +748,11 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         if (localHistory) {
             MessageProvider.chatHistory = localHistory?.chat
             MessageProvider.inputHistory = localHistory.input
+        }
+
+        const endpoint = this.authProvider.getAuthStatus().endpoint
+        if (endpoint) {
+            void syncTranscript(endpoint)
         }
     }
 
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -22,6 +22,7 @@ import { annotateAttribution, Guardrails } from '@sourcegraph/cody-shared/src/gu
 import { IntentDetector } from '@sourcegraph/cody-shared/src/intent-detector'
 import { ANSWER_TOKENS, DEFAULT_MAX_TOKENS } from '@sourcegraph/cody-shared/src/prompt/constants'
 import { Message } from '@sourcegraph/cody-shared/src/sourcegraph-api'
+import { isDotCom } from '@sourcegraph/cody-shared/src/sourcegraph-api/environments'
 
 import { showAskQuestionQuickPick } from '../commands/utils/menu'
 import { VSCodeEditor } from '../editor/vscode-editor'
@@ -189,6 +190,11 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         this.sessionID = chatID || new Date(Date.now()).toUTCString()
     }
 
+    private get isDotComUser(): boolean {
+        const endpoint = this.authProvider.getAuthStatus()?.endpoint || ''
+        return isDotCom(endpoint)
+    }
+
     private sendPrompt(
         promptMessages: Message[],
         responsePrefix = '',
@@ -234,9 +240,10 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 const codeCount = countGeneratedCode(text)
                 if (codeCount?.charCount) {
                     const metadata = lastInteraction?.getHumanMessage().metadata
+                    const response = this.isDotComUser ? text : undefined
                     telemetryService.log(
                         'CodyVSCodeExtension:chatResponse:hasCode',
-                        { ...codeCount, ...metadata, requestID },
+                        { ...codeCount, ...metadata, requestID, response },
                         { hasV2Event: true }
                     )
                     telemetryRecorder.recordEvent(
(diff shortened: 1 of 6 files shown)


vscode/src/services/telemetry.ts
```diff
@@ -136,3 +138,67 @@ export const telemetryService: TelemetryService = {
...
+let syncingProcessStarted = false // We only wants to check this once on start up
+export async function syncTranscript(endpoint: string): Promise<void> {
+    // Only sync chat transcripts for dotcom endpoints
+    if (!isDotCom(endpoint) || syncingProcessStarted) {
+        return
+    }
+
+    syncingProcessStarted = true
+
+    const TODAYS_DATE = new Date().toISOString().split('T')[0].replaceAll('-', '/')
```
valerybugakov, line 162: Thoughts on using [format](https://date-fns.org/v2.16.1/docs/format) from `date-fns` for readability?

```ts
format(new Date(), 'yyyy/MM/dd')
```

vscode/src/services/telemetry.ts
```diff
@@ -136,3 +138,67 @@ export const telemetryService: TelemetryService = {
...
+            // Convert transcript.lastInteractionTimestamp from timestamp string to number
+            const lastInteractionTimestamp = new Date(transcript.lastInteractionTimestamp).getTime()
+
+            // Only sync if the last interaction timestamp is more recent than the last sync time.
+            if (lastInteractionTimestamp > lastSyncedTimestamp) {
+                // File location format: "cody/vscode/chatTranscript/YYYY/MM/DD/anonymousUserID.json"
+                const fileLocation = `cody/vscode/chatTranscript/${TODAYS_DATE}/${anonymousUserID}.json`
+                const eventData = { id, transcript: JSON.stringify(transcript), fileLocation }
+                logEvent(eventName, eventData)
+                lastSyncedTranscriptTimestamp = lastInteractionTimestamp
```
valerybugakov, line 197: We do not `await` for `logEvent` to complete, which means multiple requests will be started asynchronously, and even if they fail, all of the timestamps will still be marked as synced. 

We should either do it sequentially if we rely on `lastSyncedTranscriptTimestamp` or maintain a map of synced chat IDs to independently sync them and recover appropriately in case of an error.

vscode/src/services/telemetry.ts
abeatrix: Good idea!!

vscode/src/services/telemetry.ts
```diff
@@ -136,3 +138,67 @@ export const telemetryService: TelemetryService = {
...
+        Object.entries(chatFromStore).forEach(([id, transcript]) => {
+            // Convert transcript.lastInteractionTimestamp from timestamp string to number
+            const lastInteractionTimestamp = new Date(transcript.lastInteractionTimestamp).getTime()
+
+            // Only sync if the last interaction timestamp is more recent than the last sync time.
+            if (lastInteractionTimestamp > lastSyncedTimestamp) {
+                // File location format: "cody/vscode/chatTranscript/YYYY/MM/DD/anonymousUserID.json"
+                const fileLocation = `cody/vscode/chatTranscript/${TODAYS_DATE}/${anonymousUserID}.json`
+                const eventData = { id, transcript: JSON.stringify(transcript), fileLocation }
+                logEvent(eventName, eventData)
```
akalia25, line 196: I believe we'd need to change this because right now it is still sending the chatTranscript data through the logEvent ~pubsub; we would need to import the @google-cloud/storage library and establish a connection with the soucegraph-cody bucket and upload the chatTranscript there with the fileLocation you've defined :) 

Found [this](https://cloud.google.com/storage/docs/uploading-objects#storage-upload-object-nodejs) helpful on how to upload data to a GCS bucket

akalia25: So I'm thinking something like this?

We create an async function to uploadTranscript:
```
const bucketName = 'bucket_name'

export async function uploadTranscript(transcript: any, fileLocation: string) {
  const transcriptString = JSON.stringify(transcript)
  const transcriptBuffer = new TextEncoder().encode(transcriptString)

  const options = {
    destination: fileLocation,
    uploadType: 'media',
    metadata: {
      contentType: 'application/json'
    }
  }

  await storage.bucket(bucketName).file(fileLocation).save(transcriptBuffer, options)
}
```

And then within the looping of chatTranscripts we would call that?

```
        Object.entries(chatFromStore).forEach(([id, transcript]) => {
            // Convert transcript.lastInteractionTimestamp from timestamp string to number
            const lastInteractionTimestamp = new Date(transcript.lastInteractionTimestamp).getTime()

            // Only sync if the last interaction timestamp is more recent than the last sync time.
            if (lastInteractionTimestamp > lastSyncedTimestamp) {
                // File location format: "cody/vscode/chatTranscript/YYYY/MM/DD/anonymousUserID.json"
                const fileLocation = `cody/vscode/chatTranscript/${TODAYS_DATE}/${anonymousUserID}/${id}.json`
                const eventData = { id, transcript: fileLocation }
                uploadTranscript(transcript, fileLocation)
                logEvent(eventName, eventData)
                lastSyncedTranscriptTimestamp = lastInteractionTimestamp
            }
        }
 ```
 One small change I did with the `fileLocation` is adding the `id` to it; otherwise if there is multiple chats in the history they would all have the same fileLocation and overwrite each other. Does this help 😬 
 
 We'd also need to setup an authentication to write to it (more info [here](https://cloud.google.com/docs/authentication/use-cases#app-users)) ~ I can also creating this

File: vscode/src/chat/MessageProvider.ts [hunk 4]
```diff
@@ -234,9 +240,10 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 const codeCount = countGeneratedCode(text)
                 if (codeCount?.charCount) {
                     const metadata = lastInteraction?.getHumanMessage().metadata
+                    const response = this.isDotComUser ? text : undefined
                     telemetryService.log(
                         'CodyVSCodeExtension:chatResponse:hasCode',
-                        { ...codeCount, ...metadata, requestID },
+                        { ...codeCount, ...metadata, requestID, response },
```
akalia25, line 246: might be a silly question, but is there a maximum character length output Cody can produce as a response? So that we can be mindful of the upper limits of the messages we receive. 😬 

Similarly, I was trying to get the max character limit for a user inputting in the Cody chat box and did not hit any limit even after 700k characters 🙈 Is there a limit on that too? If not, might be worthwhile adding one to the user's chatbox before we start logging their chat questions 🤔 
image

abeatrix: We dont truncate the display text, but we do truncate the text under the hood and set it as `interaction.getHumanMessage().text`. This is why we will be logging `interaction.getHumanMessage().text` instead of `interaction.getHumanMessage().displayText` in https://github.com/sourcegraph/cody/blob/d4dc450d8613123e191d097ca2e76762a7107e0d/vscode/src/chat/MessageProvider.ts#L463 😄

akalia25: thank you for clarifying!! 🚀 🙇

abeatrix:
replaced by https://github.com/sourcegraph/cody/pull/1976

PR #1858, opened 2023-11-22
fix: display local context

abeatrix:
As discussed with @toolmantim on slack ([thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1700631673836569?thread_ts=1700622656.385639&cid=C05AGQYD528)), the context display widget should also include local context except the ones added by the user manually via @ 

>  I was considering anything that wasn't explicitly @-included as "enhanced context"

## Test plan

Try running the `/test` command, which use local context only.

#### BEFORE

Used local context is not showing up in the enhance context widget

image

#### AFTER

Used local context is included in the context widget

image

### DIFF
--- a/lib/ui/src/chat/components/EnhancedContext.tsx
+++ b/lib/ui/src/chat/components/EnhancedContext.tsx
@@ -12,8 +12,6 @@ export interface FileLinkProps {
     range?: ActiveTextEditorSelectionRange
 }
 
-const enhancedContextSources = new Set(['embeddings', 'keyword', 'symf', 'filename'])
-
 export const EnhancedContext: React.FunctionComponent<{
     contextFiles: ContextFile[]
     fileLinkComponent: React.FunctionComponent<FileLinkProps>
@@ -25,10 +23,11 @@ export const EnhancedContext: React.FunctionComponent<{
 
     const uniqueFiles = new Set<string>()
     const filteredFiles = contextFiles.filter(file => {
-        if (uniqueFiles.has(file.fileName) || !file.source) {
+        if (uniqueFiles.has(file.fileName)) {
             return false
         }
-        if (!enhancedContextSources.has(file.source)) {
+        // Skip files added by user. e.g. @-files
+        if (file.source === 'user') {
             return false
         }
         uniqueFiles.add(file.fileName)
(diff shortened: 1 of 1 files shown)


toolmantim (approved):
👍🏼

I've only just spotted that `snippets from ${files}` UI language now… "snippets" isn't something we mention anywhere else and it would be good to avoid introducing it here.

Under what condition is that shown? (how does `!file.range` occur?)

The intention was it would always say `n` lines from `y` files, so the user is able to get a sense of just how many lines were used/included in the context (so they can keep an eye on the volume of what's included)

abeatrix:
> 👍🏼
> 
> I've only just spotted that `snippets from ${n} files` UI language now… "snippets" isn't something we mention anywhere else and it would be good to avoid introducing it here.
> 
> Under what condition is that shown? (how does `!file.range` occur?)
> 
> The intention was it would always say `n` lines from `y` files, so the user is able to get a sense of just how many lines were used/included in the context (so they can keep an eye on the volume of what's included)

Should I replace `snippets` with `context` for files that do not contain line count? (I made the change perviously because chat from older chat history would all display `0 lines from n files` before the change.)

abeatrix:
> > Should I replace "snippets" with "context" for files that do not contain line count?
> 
> `{$n} files` would be enough. I have to ask though, how could a file not contain a line count? (can we generate one?)
> 
> After thinking about it some more, I think "✨" should only display if "Enhanced Context" was checked too.

cause not all of the sources use range or come from local files, but we can count them manually. We cannot generate them for old chat though so they will show up as ${n} file

abeatrix:
> > > Should I replace "snippets" with "context" for files that do not contain line count?
> > 
> > 
> > `{$n} files` would be enough. I have to ask though, how could a file not contain a line count? (can we generate one?)
> > After thinking about it some more, I think "✨" should only display if "Enhanced Context" was checked too.
> 
> cause not all of the sources use range or come from local files, but we can count them manually. We cannot generate them for old chat though so they will show up as ${n} file

looks like that doesn't work because the local context doesn't use the content field  😓 

Updated to show the emoji when enhanced context is used:

image
 
Will need look into attaching line numbers to different context in a separate issue

PR #1859, opened 2023-11-22
evaluate-autocomplete: fix git-log strategy and add more CSV columns

olafurpg:
Previously, the `git-log` strategy did not run correctly due to a regression in the last change. This PR fixes that regression and adds two new columns to the generated CSV file:

- `EVENT`: for detailed information about each autocomplete response like latency. This is the JSON object we include in telemetry events.
- `REVISION`: to be able to know at which git commit the request was sent. This is mostly helpful to know if two requests for the same file/line location happened on the same file content.

## Test plan

Manually tested.

### DIFF
--- a/agent/src/cli/evaluate-autocomplete/strategy-git-log.ts
+++ b/agent/src/cli/evaluate-autocomplete/strategy-git-log.ts
@@ -10,6 +10,7 @@ import { getLanguageForFileName } from '../../language'
 
 import { AutocompleteDocument } from './AutocompleteDocument'
 import { EvaluateAutocompleteOptions, matchesGlobPatterns } from './evaluate-autocomplete'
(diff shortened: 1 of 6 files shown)


agent/src/cli/evaluate-autocomplete/strategy-git-log.ts
```diff
@@ -132,6 +136,6 @@ export async function evaluateGitLogStrategy(
         // Reset submodule to initial HEAD
         const submodulesDir = path.join(workspace, '..')
         execSync('git submodule deinit -f .', { cwd: submodulesDir }).toString()
-        execSync('it submodule update --init', { cwd: submodulesDir }).toString()
```
philipp-spiess, line 135: 😬

philipp-spiess (approved):
Very nice!

PR #1862, opened 2023-11-22
Autocomplete: Reduce frequency of low performance language

philipp-spiess:
Shipping this FF which was already at 100% rollout on dotcom for a while

## Test plan

- 👀

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -237,10 +237,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
 
         // We start feature flag requests early so that we have a high chance of getting a response
         // before we need it.
-        const [languageLatencyPromise, userLatencyPromise] = [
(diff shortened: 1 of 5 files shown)


abeatrix (approved):
🚀

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -237,10 +237,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
 
         // We start feature flag requests early so that we have a high chance of getting a response
         // before we need it.
-        const [languageLatencyPromise, userLatencyPromise] = [
-            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLanguageLatency),
```
valerybugakov, line 241: Should we also remove https://github.com/sourcegraph/cody/blob/ps/ship-lang-latency/lib/shared/src/experimentation/FeatureFlagProvider.ts#L22?

valerybugakov (approved):

PR #1864, opened 2023-11-22
Fix flaky completion test and clean up console

philipp-spiess:
- Fix flaky inline completion test by making sure we do not trigger a follow up completion after the last action that may or may not be logged
- Clean up VS Code install noise when the file was already downloaded. Not helpful.
- Don't spam the console with logging noise when not on CI
- Prevent completions from showing up after a `;` (similar to `)` `]` `}`)

Screenshot 2023-11-22 at 13 44 24

bliss!

## Test plan

- Run inline completion test a bunch of times and make sure it doesn't fail anymore
- No more noise in the console

### DIFF
--- a/vscode/src/completions/get-inline-completions.ts
+++ b/vscode/src/completions/get-inline-completions.ts
@@ -184,7 +184,7 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
     }
 
     // Do not trigger when the last character is a closing symbol
-    if (triggerKind !== TriggerKind.Manual && /[)\]}]$/.test(currentLinePrefix.trim())) {
+    if (triggerKind !== TriggerKind.Manual && /[);\]}]$/.test(currentLinePrefix.trim())) {
         return null
     }
 
--- a/vscode/test/fixtures/mock-server.ts
+++ b/vscode/test/fixtures/mock-server.ts
@@ -123,9 +123,7 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
         }
     })
 
-    const server = app.listen(SERVER_PORT, () => {
-        console.log(`Mock server listening on port ${SERVER_PORT}`)
-    })
+    const server = app.listen(SERVER_PORT)
 
     const result = await around()
 
@@ -135,6 +133,10 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
 }
 
 export async function logTestingData(type: 'legacy' | 'new', data: string): Promise<void> {
+    if (process.env.CI === undefined) {
+        return
+    }
+
     const message = {
         type,
         event: data,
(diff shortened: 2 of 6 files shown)


vscode/src/completions/get-inline-completions.ts
```diff
@@ -184,7 +184,7 @@ async function doGetInlineCompletions(params: InlineCompletionsParams): Promise<
     }
 
     // Do not trigger when the last character is a closing symbol
-    if (triggerKind !== TriggerKind.Manual && /[)\]}]$/.test(currentLinePrefix.trim())) {
+    if (triggerKind !== TriggerKind.Manual && /[);\]}]$/.test(currentLinePrefix.trim())) {
```
philipp-spiess, line 187: I wanted to sneak this in anyways; If we have a line like this:

```
foo.bar();
```

we don't want to trigger a completion regardless of wether we print a semi or not.

vscode/test/fixtures/mock-server.ts
```diff
@@ -135,6 +133,10 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
 }
 
 export async function logTestingData(type: 'legacy' | 'new', data: string): Promise<void> {
+    if (process.env.CI === undefined) {
+        return
+    }
```
philipp-spiess, line 138: @akalia25 is it okay if this runs only on our CI servers? Otherwise this makes local development a PITA

vscode/test/fixtures/mock-server.ts
```diff
@@ -123,9 +123,7 @@ export async function run<T>(around: () => Promise<T>): Promise<T> {
         }
     })
 
-    const server = app.listen(SERVER_PORT, () => {
-        console.log(`Mock server listening on port ${SERVER_PORT}`)
```
philipp-spiess, line 127: Not helpful. Tests should only log something if something fails IMO

DanTup:
@philipp-spiess I hadn't gotten to landing it yet, but I have some changes in https://github.com/sourcegraph/cody/pull/1535 to fix some flakes here too. Not sure if you'd seen that - does this PR supersede that, or might we need both?

philipp-spiess:
@DanTup Ahh I’m so sorry, I have completely overlooked your PR. I think with your approach works too to fix the flake. It's a bit bound to the completion provider (Anthropic in this case) because we parse the prompt but otherwise no big deal. I think we can land both fixes with the `;` making sure no follow-up completion is ever shown and yours making sure we control which completions are shown. There are a few other goodies in this PR that should make it easier to work with it locally.

vscode/test/fixtures/mock-server.ts
akalia25: Yes, of course! I appreciate you checking 🙇

valerybugakov (approved):

PR #1866, opened 2023-11-22
evaluate-autocomplete: add support to run test commands

olafurpg:
Previously, evaluate-autocomplete only supported testing whether an autocomplete result was an exact match with the original code or if it was an empty result. This PR adds an additional test to check whether the code typechecks.

## Test plan

Ran it manually.

### DIFF
--- a/agent/src/cli/evaluate-autocomplete/Timer.ts
+++ b/agent/src/cli/evaluate-autocomplete/Timer.ts
@@ -1,0 +1,14 @@
+import prettyMs from 'pretty-ms'
+
+export class Timer {
+    public readonly start: number
+    constructor() {
+        this.start = Date.now()
+    }
+    public elapsed(): number {
+        return Date.now() - this.start
+    }
+    public toString(): string {
+        return prettyMs(this.elapsed())
+    }
+}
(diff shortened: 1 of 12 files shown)


agent/src/cli/evaluate-autocomplete/Timer.ts
```diff
@@ -0,0 +1,14 @@
+import prettyMs from 'pretty-ms'
+
+export class Timer {
+    public readonly start: number
+    constructor() {
+        this.start = Date.now()
```
philipp-spiess, line 6: Might want to use `performance.now()` in the future to have a guaranteed monotonic click in case of system time syncs and to have higher resolution 🙂

philipp-spiess (approved):

agent/src/cli/evaluate-autocomplete/Timer.ts
olafurpg: Good call! Will fix

olafurpg: This isn't used for anything meaningful, I just wanted to see how long the `pnpm build` invocation was taking.

philipp-spiess: Yeah I figured so, no big deal at all

PR #1867, opened 2023-11-22
Autocomplete: Add OpenTelemetry

philipp-spiess:

### DIFF
--- a/vscode/src/tracing/initializeOpenTelemetry.node.ts
+++ b/vscode/src/tracing/initializeOpenTelemetry.node.ts
@@ -1,0 +1,19 @@
+import { Resource } from '@opentelemetry/resources'
+import { ConsoleMetricExporter, PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics'
+import { NodeSDK } from '@opentelemetry/sdk-node'
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-node'
+import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
+
+export function initialize(): void {
+    const sdk = new NodeSDK({
+        resource: new Resource({
+            [SemanticResourceAttributes.SERVICE_NAME]: 'cody',
+            [SemanticResourceAttributes.SERVICE_VERSION]: '0.1',
+        }),
+        traceExporter: new ConsoleSpanExporter(),
+        metricReader: new PeriodicExportingMetricReader({
+            exporter: new ConsoleMetricExporter(),
+        }),
+    })
+    sdk.start()
+}
--- a/vscode/src/tracing/tracer.ts
+++ b/vscode/src/tracing/tracer.ts
@@ -1,0 +1,18 @@
+import opentelemetry from '@opentelemetry/api'
+
+export const tracer = opentelemetry.trace.getTracer('cody', '0.1')
+
+export function startAsyncSpan<T>(name: string, fn: () => T | Promise<T>): Promise<T> {
+    return new Promise((resolve, reject) => {
+        void tracer.startActiveSpan(name, async span => {
+            try {
+                const result = await fn()
+                resolve(result)
+            } catch (error) {
+                reject(error)
+            } finally {
+                span.end()
+            }
+        })
+    })
+}
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -331,6 +331,24 @@ importers:
       '@mdi/js':
         specifier: ^7.2.96
         version: 7.2.96
+      '@opentelemetry/api':
+        specifier: ^1.7.0
+        version: 1.7.0
+      '@opentelemetry/resources':
+        specifier: ^1.18.1
+        version: 1.18.1(@opentelemetry/api@1.7.0)
+      '@opentelemetry/sdk-metrics':
+        specifier: ^1.18.1
+        version: 1.18.1(@opentelemetry/api@1.7.0)
+      '@opentelemetry/sdk-node':
+        specifier: ^0.45.1
+        version: 0.45.1(@opentelemetry/api@1.7.0)
+      '@opentelemetry/sdk-trace-node':
+        specifier: ^1.18.1
+        version: 1.18.1(@opentelemetry/api@1.7.0)
+      '@opentelemetry/semantic-conventions':
+        specifier: ^1.18.1
+        version: 1.18.1
       '@sentry/browser':
         specifier: ^7.66.0
         version: 7.66.0
@@ -3841,7 +3859,7 @@ packages:
       '@google-cloud/precise-date': 3.0.1
       '@google-cloud/projectify': 3.0.0
       '@google-cloud/promisify': 2.0.4
-      '@opentelemetry/api': 1.4.1
+      '@opentelemetry/api': 1.7.0
       '@opentelemetry/semantic-conventions': 1.3.1
       '@types/duplexify': 3.6.1
       '@types/long': 4.0.2
@@ -3864,7 +3882,6 @@ packages:
     dependencies:
       '@grpc/proto-loader': 0.7.8
       '@types/node': 20.4.0
-    dev: true
 
   /@grpc/proto-loader@0.7.8:
     resolution: {integrity: sha512-GU12e2c8dmdXb7XUlOgYWZ2o2i+z9/VeACkxTA/zzAe2IjclC5PnVL0lpgjhrqfpDYHzM8B1TF6pqWegMYAzlA==}
@@ -3875,7 +3892,6 @@ packages:
       long: 4.0.0
       protobufjs: 7.2.4
       yargs: 17.7.2
-    dev: true
 
   /@humanwhocodes/config-array@0.11.11:
     resolution: {integrity: sha512-N2brEuAadi0CcdeMXUkhbZB84eskAc8MEX1By6qEchoVywSgXPIjou4rYsl0V3Hj0ZnuGycGCjdNgockbzeWNA==}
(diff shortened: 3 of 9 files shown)


philipp-spiess:
cc @bobheadxi / @valerybugakov 

I’m currently struggling getting a tracing UI running with my local SG instance (`sg start dotcom` adds tracing support but the jaeger UI fails with 500 errors?) and I also don't know how we currently connect client and server traces which would also be super helpful. Does either of you have some pointers for that? :)

bobheadxi:
@philipp-spiess  were you able to find https://docs.sourcegraph.com/dev/how-to/opentelemetry_local_dev ? Most of the o11y components are in separate run sets in local dev. If you can't get it working off this guide, ping me in Slack and I can work with you directly on that :)

philipp-spiess:
@bobheadxi Yeah I was looking at this and it wasn't working for me before (I think I gave it 10min or so to start?) but now it works?? Awesome

bobheadxi:
> Yeah I was looking at this and it wasn't working for me before (I think I gave it 10min or so to start?) but now it works?? Awesome

Nice 😁 The OpenTelemetry collector takes quite a long time to build the first time, as we have a custom builder that bundles select integrations that we choose

bobheadxi:
> and I also don't know how we currently connect client and server traces which would also be super helpful

It _should_ "just work" if you configure your HTTP requests to forward the OpenTelemetry context. I think you'll want something like https://www.npmjs.com/package/@opentelemetry/instrumentation-http, or whatever is set up in the web app, whenever you talk to a Sourcegraph instance.

Then you'll want to export your traces to the connected instance, IIRC the OTLP endpoint is at 'sourcegraph.test/-/debug/otlp'

philipp-spiess:
Awesome! I can work with this tomorrow. We might need to add a new API to expose the otel config for clients (in the web app we use the global context but that's not accessible via GQL)

bobheadxi:
> We might need to add a new API to expose the otel config for clients

I think you don't strictly _need_ any configuration clientside, the exporter endpoint is always `sourcegraph.test/-/debug/otlp`. However, this _can_ be disabled serverside, and when it is we return `422` - maybe clientside we can just check if we get this code, and if we do, disable OpenTelemetry? https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/app/otlpadapter/adapter.go?L76%3A1-88%3A6

philipp-spiess:
We probably only care about tracing for dotcom anyways for now (since that we can easily access for a vast majority of our users) and there we can assume it's enabled I guess?

vscode/src/tracing/initializeOpenTelemetry.node.ts
```diff
@@ -0,0 +1,19 @@
...
+import { ConsoleSpanExporter } from '@opentelemetry/sdk-trace-node'
+import { SemanticResourceAttributes } from '@opentelemetry/semantic-conventions'
+
+export function initialize(): void {
+    const sdk = new NodeSDK({
+        resource: new Resource({
+            [SemanticResourceAttributes.SERVICE_NAME]: 'cody',
+            [SemanticResourceAttributes.SERVICE_VERSION]: '0.1',
+        }),
+        traceExporter: new ConsoleSpanExporter(),
```
valerybugakov, line 13: We probably want to make it possible to disable it with an env variable or config setting.

vscode/src/tracing/tracer.ts
```diff
@@ -0,0 +1,18 @@
...
+export const tracer = opentelemetry.trace.getTracer('cody', '0.1')
+
+export function startAsyncSpan<T>(name: string, fn: () => T | Promise<T>): Promise<T> {
+    return new Promise((resolve, reject) => {
+        void tracer.startActiveSpan(name, async span => {
+            try {
+                const result = await fn()
+                resolve(result)
+            } catch (error) {
+                reject(error)
```
valerybugakov, line 12: We can also record exceptions [here](https://opentelemetry.io/docs/instrumentation/js/manual/#recording-exceptions).

vscode/src/tracing/initializeOpenTelemetry.node.ts
```diff
@@ -0,0 +1,19 @@
...
+        metricReader: new PeriodicExportingMetricReader({
```
valerybugakov, line 14: What's the use case for metrics? AFAIU, we can ignore this part of the SDK since we're more interested in traces.

valerybugakov (commented):
As a next step for the visualization of traces locally, we could use the OpenTelemetry collector from the sg/sg repo with a configured visualizer like Jaeger. This is what I did for the web app:

1. Add an ability to log tracer to a locally running OTel collector.
2. Start the OTel collector locally. `sg start otel`, @bobheadxi, please correct me.
3. Set the traces export URL to the locally running collector.
4. Generate traces.
5. Review them with Jaeger or other viz tools configured in `sg`.

Next next step is to upload traces to Honeycomb and see if their structures can be improved for better analysis. I used a separate test Honeycomb env for that to avoid cluttering the "production" env.

philipp-spiess:
@bobheadxi @valerybugakov Is there any allowlist that I need to modify when I add a new tracing source? I've configured the node SDK now to use the `@opentelemetry/exporter-trace-otlp-http` package but I can't get anything to show up on the SG instance side 🤔 I also don't know how to debug since there's no "network tab" in VS Code extension haha

philipp-spiess:
For some reason, there is not even an attempt being made to send the traces. I did find out that gitkraken has open telemetry though and they configure it using `BasicTracerProvider` instead of the `NodeSDK`. I'll try that next

Screenshot 2023-11-23 at 16 11 03

philipp-spiess:
Nevermind me here it was just a race condition in the setup logic lol

philipp-spiess:
Screenshot 2023-11-23 at 16 30 33

Getting somewhere 👀  But the joining with backend traces doesn't seem to work yet

philipp-spiess:
Tried to manually add the x-trace and x-trace-span headers but the server logs still don't join 🤔 

Screenshot 2023-11-23 at 16 42 59

philipp-spiess:
@valerybugakov How do you like this naming/granularity for the first version? 

Screenshot 2023-11-23 at 17 13 08

If we can get the server traces as subspans in the `generate` step this would seem quite good already? 

I.e we should already investigate what takes 30ms in the planning step? is this just tree-sitter?

vscode/src/tracing/initializeOpenTelemetry.node.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        // const exporter = new OTLPTraceExporter({
+        //     url: traceUrl,
+        // })
+        // this.provider.addSpanProcessor(new BatchSpanProcessor(exporter))
+        // this.provider.register()
+
+        //  this.provider.getTracer(context.extensionId)
+
+        this.sdk = new NodeSDK({
```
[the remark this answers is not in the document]
philipp-spiess: @bobheadxi I _think_ that's not needed, Node.js has APIs for that (async local storage) and the async context propagation seems to work fine already (the whole autocomplete request is in one trace) which I assume the `NodeSDK` handles.

I'll need to check what exactly `HttpInstrumentation()` does and why it doesn't work 🤔  (We _do_ use the underlying `http` module eventually but VS Code already mocks something on the network layer so I wanted to test it without the auto mocking)

philipp-spiess:
I restarted the otel process and then it worked lol but now it stopped again. I think maybe this is an issue with the local dev setup?  

Screenshot 2023-11-23 at 17 18 59

valerybugakov (commented):
> @valerybugakov How do you like this naming/granularity for the first version?

Love that we use the stages described in the blogpost. Makes it easy to understand what's going on even if you're not that familiar with internal.

bobheadxi (approved):
LGTM - sorry for the troubles local dev has been causing 😆 

I think this is good to land as-is probably - the issue is very likely just the backend exporting its own traces. Looking at the behaviour in DMs with @philipp-spiess , I'm 99% sure the backend does create a span because it responds with a new span ID, but we're losing it somewhere. Let's play around with this live - I think it will deliver value even without linking - and I can have a closer look as well next week

vscode/src/completions/client.ts
```diff
@@ -71,7 +72,11 @@ export function createClient(config: CompletionsClientConfig, logger?: Completio
             headers.set('Authorization', `token ${config.accessToken}`)
         }
         if (tracingFlagEnabled) {
```
bobheadxi, line 74: Looks like this is set here, https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@76f34d0375a3355c20c99e92cfc787db946766c9/-/blob/vscode/src/completions/client.ts?L64 , and we will be enabling it for select users only. Note that you'll want to be wary of enabling this for everyone, since it might cause a pretty high load in the tracing backend - before enabling everywhere for dotcom, you might want to reach out in #team-infrastructure to collaborate on a sampling strategy in the OTEL collector :)

lib/shared/src/tracing/index.ts
```diff
@@ -0,0 +1,40 @@
...
+            .finally(() => {
+                span.end()
+            })
+    )
+}
+
+export function getTraceparent(): string | null {
+    const activeIds = getActiveTraceAndSpanId()
+    if (activeIds) {
+        return `00-${activeIds.traceId}-${activeIds.spanId}-01`
```
valerybugakov, line 37: Let's link an explainer with a short comment explaining this. I vaguely remembered the meaning behind the `traceparent` header, but the format is confusing if you do not the the OpenTelemetry spec by heart. 

Found a good example here: https://www.w3.org/TR/trace-context/#examples-of-http-traceparent-headers.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -614,34 +614,51 @@ export class SourcegraphGraphQLAPIClient {
...
-        })
-            .then(verifyResponseCode)
-            .then(response => response.json() as T)
-            .catch(error => new Error(`error fetching Sourcegraph GraphQL API: ${error} (${url})`))
+        const traceparent = getTraceparent()
+        if (traceparent) {
+            headers.set('traceparent', traceparent)
+        }
+
+        const queryName = query.match(/^\s*query\s+(\w+)/m)?.[1]
```
valerybugakov, line 650: Do we want to handle mutations here, too? Also, we can extract the regex into a CONSTANT outside of the function scope and reuse it.

lib/shared/src/tracing/index.ts
```diff
@@ -0,0 +1,40 @@
+import opentelemetry, { SpanStatusCode } from '@opentelemetry/api'
+
+const TRACE_NAME = 'cody'
+const TRACE_VERSION = '0.1'
```
valerybugakov, line 4: To comply with OpenTelemetry [terminology](https://opentelemetry.io/docs/instrumentation/js/manual/#acquiring-a-tracer):

```suggestion
const INSTRUMENTATION_NAME = 'cody'
const INSTRUMENTATION_VERSION = '0.1'
```

OpenTelemetry contains a lot of abstractions, so it's helpful when our naming matches theirs when we need to find something in the spec or docs.

valerybugakov (approved):
:letsgoo:

vscode/src/completions/client.ts
philipp-spiess: That FF is currently only enabled for 5% of users so I don't expect this to be a problem (and, for that users we already do create backend traces. This would just add the frontend section)

PR #1871, opened 2023-11-23
Provide anonymousUserId via agent clients

Strum355:
For clients that generate, persist and include this via ClientInfo.extensionConfiguration, we dont want to generate CodyInstalled events at every agent startup. For now, this means that JetBrains (the only non-vscode IDE that matters :clueless:) will continue to emit CodyInstalled events itself until we come up with a solution for agent to do so based on IDE-provided info.

Based on [the existing agent-side code](https://sourcegraph.com/github.com/sourcegraph/cody@nsc/agent-anonymoususerid/-/blob/vscode/src/services/telemetry.ts?L93%3A4-116%3A2), 

## Test plan

Confirmed based on whether CodyInstalled event is fired or not when a client provides anonymousUserID

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -15,6 +15,7 @@ import { NoOpTelemetryRecorderProvider } from '@sourcegraph/cody-shared/src/tele
 import { TelemetryEventParameters } from '@sourcegraph/telemetry'
 
 import { activate } from '../../vscode/src/extension.node'
+import { localStorage } from '../../vscode/src/services/LocalStorageProvider'
 
 import { AgentTextDocument } from './AgentTextDocument'
 import { newTextEditor } from './AgentTextEditor'
@@ -27,7 +28,10 @@ import * as vscode_shim from './vscode-shim'
 
 const secretStorage = new Map<string, string>()
 
-export function initializeVscodeExtension(workspaceRoot: vscode.Uri): void {
+export function initializeVscodeExtension(
+    workspaceRoot: vscode.Uri,
+    extensionConfiguration?: ExtensionConfiguration
+): void {
     const paths = envPaths('Cody')
     activate({
         asAbsolutePath(relativePath) {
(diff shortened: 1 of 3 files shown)


agent/src/agent.ts
```diff
@@ -43,7 +47,16 @@ export function initializeVscodeExtension(workspaceRoot: vscode.Uri): void {
         extensionUri: vscode.Uri.from({ scheme: 'file', path: '__extensionUri__should_never_be_read_from' }),
         globalState: {
             keys: () => [],
-            get: () => undefined,
+            get: key => {
+                switch (key) {
+                    case localStorage.ANONYMOUS_USER_ID_KEY:
+                        // cant use vscode_shim.connectionConfig here as this is queried for in extension activation
+                        // and vscode_shim.connectionConfig is set after extension activation.
```
olafurpg, line 54: We can set `vscode_shim.connectionConfig` before we call `initializeVscodeExtension()`. Either way is fine, but reading this comment makes me feel like we're indicating this is the only possible solution.

olafurpg (approved):

agent/src/agent.ts
Strum355: Youre probably right, although Ive asked @bobheadxi whether the order of things in `initialize` request matter, as he recently moved some things around there. Ive pushed a commit for this either way and ill have it pending Roberts response

bobheadxi: If it works I think it's fine, I did not have a strong reason for that earlier move

Strum355: After more reading, my original comment was actually correct, due to a very deeply found set of order dependencies.

vscode/src/services/telemetry.ts
```diff
@@ -60,7 +60,7 @@ export async function createOrUpdateEventLogger(
             logEvent('CodyInstalled', undefined, {
                 hasV2Event: true, // Created in src/services/telemetry-v2.ts
             })
-        } else {
+        } else if (!config.isRunningInsideAgent) {
```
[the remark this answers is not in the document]
Strum355: Probably do. I've been not thinking about V2 stuff in order to get product/analytics V1 requests done before i move to devx next monday 😅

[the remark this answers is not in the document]
bobheadxi: https://github.com/sourcegraph/cody/pull/1916

PR #1872, opened 2023-11-23
Agent: add testing mode

olafurpg:
Previously, we didn't run agent tests in CI because they relied on a live Sourcegraph server running in the background to respond to HTTP requests. This PR adds support for a new testing mode that records HTTP requests into a directory and replays from that directory. This new testing mode makes it possible to enable the agent tests in CI because they run entirely from the HTTP recording avoiding the need for a sourcegraph access token or a live sourcegraph server. The HTTP recorder has been customized to automatically redact the Sourcegraph access token so that we don't commit the token into the git repository.

This PR also enables BFG tests, which send an HTTP requests to fetch the BFG binary but are otherwise offline.

## Test plan

Green CI with agent tests enabled!


olafurpg:
I have a local branch that fixes a lot of things, will push later today

agent/src/bfg/BfgContextFetcher.test.ts
```diff
@@ -79,9 +78,10 @@ describe('BfgRetriever', async () => {
                 customConfiguration: { 'cody.experimental.bfg.path': bfgBinary },
             })
         }
+        const paths = envPaths('Cody')
         const extensionContext: Partial<vscode.ExtensionContext> = {
             subscriptions: [],
-            globalStorageUri: vscode.Uri.from({ scheme: 'file', path: tmpDir }),
+            globalStorageUri: vscode.Uri.file(paths.data),
```
olafurpg, line 84: This change wasn't strictly necessary for this PR, but it made the BFG tests run way faster because we reuse the cached binary between test runs.

olafurpg:
I will fix the Windows failures on Monday. Would be good to get the Windows build green with this enabled instead of disabling it on Windows

philipp-spiess (approved):
Awesome

PR #1880, opened 2023-11-24
Autocomplete: Remove DisableNetworkCache flag

philipp-spiess:
c.f. https://sourcegraph.slack.com/archives/C05AGQYD528/p1700673200333879?thread_ts=1700673177.479549&cid=C05AGQYD528

## Test plan

👀


dominiccooney (approved):
Perfection is achieved, not when there is nothing left to add, but when there is nothing left to take away. So taking away the option to take away the cache is... wait is this a riddle?

Code looks good! Thank you for cleaning stuff up.

Would be great if we can get test plans that aren't just 👀  , there must be some caching tests which should still pass with this option gone?

PR #1881, opened 2023-11-24
Allow awaiting extension activation instead of being unawaitably async

Strum355:
Discovered while debugging why commands executed in agent on initialization weren't being fired, there can be a race condition whereby the extension hasnt activated yet by the time we attempt to execute the commands due to activation of the extension being unawaitably async.

## Test plan

Confirmed by observing that [looking up commands to execute](https://sourcegraph.com/github.com/sourcegraph/cody@nsc/promise-extension-activation/-/blob/agent/src/vscode-shim.ts?L412) no longer fails

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -28,12 +28,12 @@ import * as vscode_shim from './vscode-shim'
 
 const secretStorage = new Map<string, string>()
 
-export function initializeVscodeExtension(
+export async function initializeVscodeExtension(
     workspaceRoot: vscode.Uri,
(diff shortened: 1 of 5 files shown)


olafurpg (approved):
LGTM 👍🏻 I double checked that it was OK to make the `activate` function async by searching for other example here, seems common. Might be good to get a second approval from somebody on the vscode extension, but worst case we can merge now and revert if it's an undesirable change  https://sourcegraph.com/search?q=context%3Aglobal+export+async+function+activate&patternType=standard&sm=1&groupBy=repo

PR #1882, opened 2023-11-24
Add rate limit notification/upgrade buttons in chat

DanTup:
I've updated the error handling so that `RateLimitError` (instead of just a string) is propagated up to the chat. With a fake rate limit (using the same status code + headers as the existing completion rate limit), it looks like this:

image

## Test Plan

- There are some e2e tests that call fake methods on the mock server to set rate limit status and verify the appropriate messages show up
- This will require manual testing against a real service to verify the code/tests match what the live servers are doing. In particular:
  - non-dotCom should never show upgrade messages
  - dotCom should show upgrade messages if not already Pro
  - dotCom should not show upgrade messages if already Pro

### DIFF
--- a/lib/shared/src/chat/transcript/index.ts
+++ b/lib/shared/src/chat/transcript/index.ts
@@ -2,6 +2,7 @@ import { ContextFile, ContextMessage, OldContextMessage, PreciseContext } from '
 import { CHARS_PER_TOKEN, MAX_AVAILABLE_PROMPT_LENGTH } from '../../prompt/constants'
 import { PromptMixin } from '../../prompt/prompt-mixin'
 import { Message } from '../../sourcegraph-api'
+import { RateLimitError } from '../../sourcegraph-api/errors'
 
 import { Interaction, InteractionJSON } from './interaction'
 import { ChatMessage } from './messages'
@@ -143,22 +144,84 @@ export class Transcript {
     }
 
     /**
-     * Adds a error div to the assistant response. If the assistant has collected
+     * Adds an error div to the assistant response. If the assistant has collected
      * some response before, we will add the error message after it.
-     * @param errorText The error TEXT to be displayed. Do not wrap it in HTML tags.
+     * @param error The error to be displayed.
      */
-    public addErrorAsAssistantResponse(errorText: string): void {
+    public addErrorAsAssistantResponse(error: Error): void {
+        if (error instanceof RateLimitError) {
+            this.addRateLimitErrorAsAssistantResponse(error)
+            return
+        }
+
         const lastInteraction = this.getLastInteraction()
         if (!lastInteraction) {
             return
         }
         // If assistant has responsed before, we will add the error message after it
         const lastAssistantMessage = lastInteraction.getAssistantMessage().displayText || ''
+
+        const errorHtml = `<div class="cody-chat-error"><span>Request failed: </span>${error.message}</div>`
+
         lastInteraction.setAssistantMessage({
             speaker: 'assistant',
             text: 'Failed to generate a response due to server error.',
-            displayText:
-                lastAssistantMessage + `<div class="cody-chat-error"><span>Request failed: </span>${errorText}</div>`,
+            displayText: lastAssistantMessage + errorHtml,
+        })
+    }
+
+    /**
+     * Adds an error div to the assistant response with information about the
+     * RateLimitError that was triggered.
+     */
+    public addRateLimitErrorAsAssistantResponse(error: RateLimitError): void {
+        const lastInteraction = this.getLastInteraction()
+        if (!lastInteraction) {
+            return
+        }
+
+        const canUpgrade = false // TODO(dantup): Get from auth + GA feature flag
+        const className = canUpgrade ? 'rate-limit-error upgradable' : 'rate-limit-error'
+        const title = canUpgrade ? 'Upgrade to Cody Pro' : 'Unable to Send Message'
+        let description = error.buildMessage('chat messages')
+        if (canUpgrade) {
+            description += ' Upgrade to Cody Pro to get unlimited chats.'
+        }
+        // TODO(dantup): Is HTML here the right place?
+        const errorHtml = `
+                <div class="${className}">
+                    <h1>${title}</h1>
+                    <span>${description}</span>
+                </div>
+            `.trim()
+
+        lastInteraction.setAssistantMessage({
+            speaker: 'assistant',
+            text: 'Failed to generate a response due to rate limit error.',
+            displayText: errorHtml,
+            buttons: [
+                {
+                    label: 'Upgrade',
+                    action: '',
+                    onClick: (action: string) => {
+                        // TODO(dantup): How do we open a URL from here?
+                        // can we use something like vscode.env.openExternal or
+                        // vscode.commands.executeCommand()?
+                        // console.log(`TODO: Open ${ACCOUNT_UPGRADE_URL}`);
+                    },
+                    appearance: 'primary',
+                },
+                {
+                    label: 'Learn More',
+                    action: '',
+                    onClick: (action: string) => {
+                        // TODO: Where does Learn More go?
+                    },
+                    appearance: 'secondary',
+                },
+            ],
+            footerText: error.buildResetMessage(),
+            isRateLimitError: true,
         })
     }
 
--- a/lib/shared/src/chat/transcript/index.ts
+++ b/lib/shared/src/chat/transcript/index.ts
@@ -2,7 +2,6 @@ import { ContextFile, ContextMessage, OldContextMessage, PreciseContext } from '
 import { CHARS_PER_TOKEN, MAX_AVAILABLE_PROMPT_LENGTH } from '../../prompt/constants'
 import { PromptMixin } from '../../prompt/prompt-mixin'
 import { Message } from '../../sourcegraph-api'
-import { RateLimitError } from '../../sourcegraph-api/errors'
 
 import { Interaction, InteractionJSON } from './interaction'
 import { ChatMessage } from './messages'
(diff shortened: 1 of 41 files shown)


DanTup:
@abeatrix @philipp-spiess @kalanchan some questions above, but I'm not sure who is best placed to answer them.

toolmantim:
Nice one! I’ll pop some style improvements onto this PR.

DanTup:
@toolmantim 

> Nice one! I’ll pop some style improvements onto this PR.

FYI, I rewrote history on this PR because things got confusing with the merges after the two PRs this depended on landed. If you started any changes you might want to reset to 87f0598a and then cherry-pick them back on.

@abeatrix I still have lots of questions on this one that maybe you can help with?

abeatrix:
> Is there a better way to render this error in the UI than I've started here? Existing errors were tagged onto the end of the last assistance interaction so I continued with that, but now I have buttons, extra text (footer) and a flag (to add an additional css class) I'm passing in. I don't know if I'm abusing the existing types a little and if instead there should be a better way to insert this error into the history so it can be rendered completely custom from the existing messages?

I haven't finished reviewing the PR, but I think for this purpose, instead of adding new methods to the Transcript class, we should create a React Component like we did for the [transcriptWithWelcome](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/ui/src/Chat.tsx?L516-527) and display this component if the condition is met, what do you think about this approach?


> The only other place I found using buttons was gettingStartedButtons but I couldn't actually find where those are defined to see if I could make them appear in the UI and see how they look.

The `gettingStartedButtons` are no longer being used by VS Code, but [vscode/webviews/Chat.tsx](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/Chat.tsx?L190-247) is where we defined it when we were using it in the client


> These buttons need to open URLs, but I'm defining them in lib\shared\src\chat\transcript\index.ts where I don't appear to have access to vscode. What's the correct way to open a URL (eg. vscode.openExternal() or vscode.commands.executeCommand())?

you can define it in the  [vscode/webviews/Chat.tsx](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/Chat.tsx?L190-247) like mentioned above, and then pass it to the ChatUI in the shared lib view

abeatrix:
> What's the right way to get at AuthStatus/FeatureFlags here to verify an upgrade is available?
Everytime AuthStatus is updated, we will notify the webview in [vscode/webviews/App.tsx](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/App.tsx?L66-71), so we can create a new state for the upgrade eligibility and then display the React components for the upgrade buttons whenever the state is changed to true?

toolmantim:
@DanTup I’ll stay out of the way in this PR, and I can do style improvements in a follow-up PR

lib/shared/src/chat/transcript/index.ts
```diff
@@ -143,22 +144,84 @@ export class Transcript {
...
+     * Adds an error div to the assistant response with information about the
+     * RateLimitError that was triggered.
+     */
+    public addRateLimitErrorAsAssistantResponse(error: RateLimitError): void {
+        const lastInteraction = this.getLastInteraction()
+        if (!lastInteraction) {
+            return
+        }
+
+        const canUpgrade = false // TODO(dantup): Get from auth + GA feature flag
```
kalanchan, line 183: see updated [error response](https://docs.google.com/document/d/1qYDsO6GOxGHEB2ZRSO9BWgjTra2adw0y3BDTNkjZqZw/edit#heading=h.apmzs6ll6q20)

just got the response back from PLG - they're going to add the account status and limits in the headers `x-is-cody-pro-user: true/false`, `x-ratelimit-remaining: int`, `x-ratelimit-limit: int`

we should pass those headers in the RateLimitError instead?

lib/shared/src/sourcegraph-api/completions/nodeClient.ts
```diff
@@ -35,6 +36,28 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
...
+
+                // Calls the error callback handler for an error.
+                //
+                // If the request failed with a rate limit error, wraps the
+                // error in RateLimitError.
+                function handleError(e: Error): void {
+                    log?.onError(e.message)
+
+                    if (res.statusCode === 429) {
+                        const retryAfter = res.headers['retry-after']
```
kalanchan, line 48: related to the comment above about passing in the rate limit and status headers, I think we can grab them here and pass it in the error

see [doc](https://docs.google.com/document/d/1qYDsO6GOxGHEB2ZRSO9BWgjTra2adw0y3BDTNkjZqZw/edit#heading=h.apmzs6ll6q20)

lib/shared/src/chat/transcript/index.ts
DanTup: > `x-is-cody-pro-user: true/false`

How is this affected by:

- feature flag not being enabled
- being on an endpoint that doesn't support pro (not dot-com)?

Currently I've been using a flag on `AuthStatus` for `canUserUpgrade` which is `true` only when the user _could_ upgrade. Can we rely on `x-is-cody-pro-user: false` _only_ being present if the user could upgrade (eg. this header only exists on dot-com)?

Maybe the same could go for the feature flag? don't include the header unless the feature flag is enabled for the user?

kalanchan: hmm, I would suspect they're doing the same type of check on the PLG backend to respond back with `x-is-cody-pro-user: false`, so I want to say its always reliable. 

is using `AuthStatus` necessary in this case? The user has to always be authenticated before they can use Cody right

kalanchan: cc @abeatrix and @philipp-spiess for some input!

DanTup: > hmm, I would assume they're doing the same type of check on the PLG backend to determine `x-is-cody-pro-user: false`, so I want to say its always reliable.

There are three different states though, and a boolean can't represent them all:

1. You're on dot-com, hit the rate limit, and are a free user - you can upgrade
2. You're on dot-com, hit the rate limit, and are already a Pro user (we can't tell you to upgrade, but we could include in the message that you're already Pro)
3. You're not on dot-com, hit the rate limit - ?

It's not clear to me whether non-dot-com users are considered "always Pro users" or "always free users". My understanding was that free/pro is only for dot-com users and we never want to show upgrade buttons (or Pro badges?) to other kinds of users?

kalanchan: Ahh I see what you mean, so maybe the better implementation would be to check if the user is connected through dotcom or not before handling the 3 states you listed above? 

So for state 3:
- You're not on dotcom, hit the rate limit, and get presented the standard error response (`x-is-cody-pro-user` won't be present in the headers). 

from [David's PR](https://github.com/sourcegraph/sourcegraph/pull/58664) it sounds like `x-is-cody-pro-user` is only going to be present when connected to dotcom.

DanTup: > from https://github.com/sourcegraph/sourcegraph/pull/58664 it sounds like x-is-cody-pro-user is only going to be present when connected to dotcom.

Ah, that's perfect - and it also notes:

> The new header is only added in case of dotcom when cody-pro is on.

So I think this also includes the feature flag which avoids us checking that too. If the same is try for the code completions, that might allow me to undo a bunch of making things async just to check the feature flag in the previous PR I landed.

I'll fix this all up tomorrow. Is there a way that I can toggle my cody-pro status and the code-pro feature flag for my own account for testing?

DanTup: (Oh, I see that PR is not merged yet, so that question is moot right now)

abeatrix:
@DanTup Will defer to Tim for the second question since that might need designs? But for the other two, I just merged a change yesterday that might be helpful as reference: https://github.com/sourcegraph/cody/pull/1942

In there I pass down the [vscode/webviews/Components/ChatModelDropdownMenu.tsx](https://github.com/sourcegraph/cody/blob/86ebb68f991027786e68826d7389c29cb567b818/vscode/webviews/Components/ChatModelDropdownMenu.tsx) component that has access to `vsCodeApi.postMessage` from vscode dir to `lib` , along with the `[UserInfo](https://github.com/sourcegraph/cody/blob/86ebb68f991027786e68826d7389c29cb567b818/vscode/webviews/App.tsx#L42-L45)` that contains if the user is a dot com user and if the user is a pro user, and the list of chat models available based on user's instance. Do you think this would help? Happy to get on a call to do a quick walkthrough later if you want too (in about 2 hours?)

DanTup:
@abeatrix thanks, that worked :) I'm going to look at tests, but the rest of the code is hopefully good if you want to take a look.

DanTup:
I think this is good to go once reviewed. It will need some manual testing against a real endpoint (with rate limits triggered) to ensure assumptions in the code+tests match the real implementation.

@toolmantim the styling needs some work - is this something you still plan to look at?

lib/shared/src/chat/transcript/messages.ts
```diff
@@ -9,22 +9,31 @@ export interface ChatButton {
     label: string
     action: string
     onClick: (action: string) => void
+    appearance?: 'primary' | 'secondary' | 'icon'
 }
 
 export interface ChatMessage extends Message {
+    kind?: string
+    className?: string
```
abeatrix, line 17:
```suggestion
```
Are className, footerText, and kind being used? I can't seem to find them in the PR but I might have overlooked 😅

abeatrix (commented):
I don't have an account with rate limit to test this, so I just manually alter the states to check the new components :D The components are showing up correctly based on the states 🎉 Left a question in line, and small nit on the UI side - I noticed Cody's response (the `Based on the` text in the images below) before it was cut off is showing under the buttons, and it looks like it's part of the error message. Maybe we shouldn't show them or the error widget should have a slightly different background?

image


#### when pro is not available and user hit rate limit:

image

##### when pro is available and user hit the rate limit:

image

lib/shared/src/chat/transcript/messages.ts
DanTup: Oops, those were from a previous attempt. Removed (along with `kind` which also wasn't used here, only on `ChatError`).

DanTup:
> I noticed Cody's response (the Based on the text in the images below) before it was cut off is showing under the buttons

I've stopped showing the displaytext when there's an error (I think I assumed if we had an error we'd never have any text anyway, but now it's enforced).

PR #1883, opened 2023-11-24
Edit: Stream apply insertions line-by-line

umpox:
## Description

WIP


toolmantim:
Looking great!

umpox:
Opening this up for review - Still doing some further internal testing myself but so far everything works as expected

vscode/src/chat/MessageProvider.ts
```diff
@@ -228,12 +234,13 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
...
-                    // remove display text from last interaction if this is a non-display topic
+                    const displayText =
+                        recipe.type === 'ask'
+                            ? reformatBotMessageForChat(text, responsePrefix)
+                            : reformatBotMessageForEdit(text, responsePrefix)
                     // TODO(keegancsmith) guardrails may be slow, we need to make this async update the interaction.
-                    const displayText = nonDisplayTopics.has(multiplexerTopic)
-                        ? undefined
-                        : await this.guardrailsAnnotateAttributions(reformatBotMessage(text, responsePrefix))
-                    this.transcript.addAssistantResponse(text, displayText)
```
umpox, line 236: @abeatrix I think I remember we added this so fixups wouldn't show in the old chat (no `displayText`),

I want to use `displayText` now, as I am using `text` to help pre-fill Claudes response to steer the LLM better.

AFAICT this works fine with the new chat, so it should be fine to make this change? LMK if not!

vscode/src/non-stop/FixupDocumentEditObserver.ts
```diff
@@ -61,6 +62,13 @@ export class FixupDocumentEditObserver {
         const tasks = this.provider_.tasksForFile(file)
         // Notify which tasks have changed text or the range edits apply to
         for (const task of tasks) {
+            // Cancel any ongoing `add` tasks on undo.
+            // This is to avoid a scenario where a user is trying to undo a specific part of text, but cannot because the streamed text continues to come in as the latest addition.
+            if (task.state === CodyTaskState.inserting && event.reason === vscode.TextDocumentChangeReason.Undo) {
+                this.provider_.cancelTask(task)
+                continue
+            }
```
umpox, line 70: @toolmantim So I ran into a weird possible edge case.

Because the streaming is still treated as an "undoable" action by the editor, it can mess with the undo logic for the user.

Two scenarios:

- User is trying to undo some unrelated code, it can end up undoing the latest streamed code which results in a flicker before the streamed code appears again (as the task is still running). In a worse case scenario this can result in the users intended undo being blocked.
- User is trying to stop the insertion in some way. I naturally found that I used "undo" a few times when I wanted to stop the completion. Maybe this is because this is what I would normally do if I just pasted a bunch of text, kinda like muscle memory.

This change makes it so we just cancel any active task in a file where an undo takes place. This feels like the best approach to me, LMK if you have any other ideas

https://github.com/sourcegraph/cody/assets/9516420/55f8902c-3056-4f90-8b19-d4de7c2a161d

vscode/src/non-stop/FixupTask.ts
```diff
@@ -44,6 +44,8 @@ export class FixupTask {
     ) {
         this.id = Date.now().toString(36).replaceAll(/\d+/g, '')
         this.instruction = instruction.replace(/^\/(edit|fix)/, '').trim()
+        // If there's no text determined to be selected then we will override the intent, as we can only add new code.
+        this.intent = selectionRange.isEmpty ? 'add' : intent
```
umpox, line 48: This is just shifted up to when the task is created, so it's always available immediately. There's no chance of reading the intent of a task as `edit` before it switches to `add`

umpox:
There's a bug with the "Retry" functionality for streamed edits (uses wrong range)

Will be fixed when https://github.com/sourcegraph/cody/pull/1926 lands.

Will confirm fixed on this branch when up to date

vscode/src/non-stop/FixupDocumentEditObserver.ts
toolmantim: Oh interesting! I really like the solution you came up with here to just cancel it if they undo 👍🏼

vscode/src/chat/MessageProvider.ts
abeatrix: Thanks for the heads up! Will review this first thing tomorrow morning!

abeatrix: >  I am using text to help pre-fill Claudes response to steer the LLM better.

I haven't looked at the code yet, but we used assistantPrefix for pre-filling response, would that work for your use case?

umpox: Yep that's what I'm doing, pretty similar to the Claude completions prompt:
```
            assistantText: `${this.getResponsePreamble(fixupTask)}${promptPrefix}`,
            assistantPrefix: promptPrefix,
```

prefix alone will just just prepend it to the LLM output I believe, whereas using `assistantText` will include it in the prompt sent to the LLM as part of the `Assistant` text

abeatrix (commented):
Tried to run the /doc command and it broke the code for me. LOOKS SUPER COOL THOUGH! Once this is done I can port the new /test command to use this, very exciting work!

https://github.com/sourcegraph/cody/assets/68532117/017b265f-da22-4bfc-bc5a-2d5856ea5b33

#### Edit

Here is the response from LLM in log:

```
█ CompletionLogger:onComplete: {"type":"completion","status":"success","duration":4750}  {
  "result": "/**\n * Gets the reranker to use for testing. Creates a mock reranker if one has not been set.\n * \n * @returns The reranker to use for testing.\n */\n    public getReranker(): Reranker {\n        if (!this.reranker) {\n            return new MockReranker(\n                (_: string, results: ContextResult[]): Promise<ContextResult[]> => Promise.resolve(results)\n            )\n        }\n        return this.reranker\n    }\n",
```

umpox:
@abeatrix Really good catch! It's only supposed to do this for the `add` intent, but because the doc command can have no selection, we were overriding it to be `add` due to me moving code around :D

I've fixed it now, should work as expected!

Yeah will be really cool to integrate this with the /test flow :)

PR #1884, opened 2023-11-24
Remove Claude Instant Cyan

philipp-spiess:
Cyan is now the default Claude Instant mechanism so no need to have that specialised in the client code anymore. We can always bring it back if they deploy another experiment onto that codename.

## Test plan

👀


dominiccooney (approved):
Looks good!

valerybugakov (approved):

PR #1888, opened 2023-11-27
DX: log errors with source maps

valerybugakov:
## Context

- Log autocomplete runtime errors to the console to see stack traces with source maps in Chrome Dev tools.

## Test plan

1. Start the dev instance of the extension.
2. Connect it to standalone Chrome Dev tools.
3. Throw a runtime error.
4. See a proper stack trace.

### DIFF
--- a/vscode/src/completions/get-inline-completions.ts
+++ b/vscode/src/completions/get-inline-completions.ts
@@ -45,6 +45,7 @@ export interface InlineCompletionsParams {
 
     // Feature flags
     completeSuggestWidgetSelection?: boolean
+    dynamicMultlilineCompletions?: boolean
 
     // Callbacks to accept completions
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):

vscode/src/completions/get-inline-completions.ts
```diff
@@ -45,6 +45,7 @@ export interface InlineCompletionsParams {
 
     // Feature flags
     completeSuggestWidgetSelection?: boolean
+    dynamicMultlilineCompletions?: boolean
```
philipp-spiess, line 48: This change seems unrelated?

valerybugakov: yep, ty!

PR #1889, opened 2023-11-27
Fuzzy search for @-ing files and symbols

### LINKED ISSUE #1837: bug: @-file can return different results for the same query

### LINKED ISSUE #1838: bug: @-file is case sensitive
### Version

0.16.0

### Describe the bug

@-file results are missing some of the same results that VS Code would show for the same query, due to case sensitivity:

Screenshot 2023-11-21 at 1 26 12 pm

### Expected behavior

The search returns the same results as VS Code

### Additional context

_No response_

### LINKED ISSUE #1840: bug: @-file excludes doesn't include file extension in the match
### Version

0.16.0

### Describe the bug

Unlike VS Code’s search, @-file excludes results when you include the file extension (which is useful for reducing the results to a known file type):

Screenshot 2023-11-21 at 3 03 14 pm

Screenshot 2023-11-21 at 3 03 06 pm

### Expected behavior

The search returns the same results as VS Code (file extensions are included in the matching & ranking)

### Additional context

_No response_

toolmantim:
This updates the @-file UI with cleaned up styles, fuzzy matching, and putting symbol search behind it's own `@#` syntax.

Fixes #1838
Fixes #1837
Fixes #1840
Replaces #1851

Changes:

- Shows the UI immediately after typing `@`
- Does fuzzy matching, similar to VS Code’s finder, over files and symbols
- Symbol search is behind `@#`
- Symbol search API is finicky and can easily return zero results, so added a little UI hint to help them (alternatively: we could disable it and do per-file drill-down instead like originally intended)
- Rewrite labels and clean up styles
- Handle file search cancellations properly
- De-dupe tab file names
- Tests

Not included:
- Handling of large files
- Drilling down from a file to its symbols (i.e. `@some-file.ts#SymbolName`)

| Before | After |
| - | - |
| Screenshot 2023-11-27 at 7 24 48 pm | Screenshot 2023-11-27 at 7 33 11 pm |
| Screenshot 2023-11-27 at 7 24 03 pm | Screenshot 2023-11-27 at 7 33 31 pm |
| Screenshot 2023-11-27 at 7 23 34 pm | Screenshot 2023-11-27 at 7 33 46 pm |
| Screenshot 2023-11-27 at 7 23 04 pm | |
| Screenshot 2023-11-27 at 5 10 26 pm | Screenshot 2023-11-27 at 7 34 54 pm |
| Screenshot 2023-11-27 at 7 23 24 pm | |
|  Screenshot 2023-11-27 at 8 19 15 pm | |

## Test plan

- Manual:
  - Tested on sourcegraph/cody, soucregraph/sourcegraph & chromium/chromium
  - Open the repo, start a chat, type @
  - Tested both sidebar chat UI and new chat UI
- Automated tests

### DIFF
--- a/lib/shared/src/codebase-context/messages.ts
+++ b/lib/shared/src/codebase-context/messages.ts
@@ -38,6 +38,7 @@ export interface ContextFile {
     source?: ContextFileSource
     type?: ContextFileType
     kind?: SymbolKind
+    editorTab?: boolean
 }
 
 export interface ContextMessage extends Message {
--- a/lib/shared/src/codebase-context/messages.ts
+++ b/lib/shared/src/codebase-context/messages.ts
@@ -38,7 +38,6 @@ export interface ContextFile {
     source?: ContextFileSource
     type?: ContextFileType
     kind?: SymbolKind
-    editorTab?: boolean
 }
 
 export interface ContextMessage extends Message {
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1175,6 +1175,7 @@
     "classnames": "^2.3.2",
     "date-fns": "^2.30.0",
     "detect-indent": "^7.0.1",
+    "fuzzysort": "^2.0.4",
     "glob": "^7.2.3",
     "isomorphic-fetch": "^3.0.0",
     "js-levenshtein": "^1.1.6",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1198,7 +1198,7 @@
   },
   "devDependencies": {
     "@google-cloud/pubsub": "^3.7.3",
-    "@playwright/test": "^1.33.0",
+    "@playwright/test": "^1.40.0",
     "@storybook/preview-api": "^7.5.3",
     "@types/axios": "^0.14.0",
     "@types/dedent": "^0.7.0",
@@ -1226,7 +1226,7 @@
     "mocha": "^10.2.0",
     "ovsx": "^0.8.2",
     "path-browserify": "^1.0.1",
-    "playwright": "^1.33.0",
+    "playwright": "^1.40.0",
     "progress": "^2.0.3",
     "semver": "^7.5.4"
   }
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1181,7 +1181,6 @@
     "classnames": "^2.3.2",
     "date-fns": "^2.30.0",
     "detect-indent": "^7.0.1",
-    "fuzzysort": "^2.0.4",
     "glob": "^7.2.3",
     "isomorphic-fetch": "^3.0.0",
     "js-levenshtein": "^1.1.6",
@@ -1204,7 +1203,7 @@
   },
   "devDependencies": {
     "@google-cloud/pubsub": "^3.7.3",
-    "@playwright/test": "^1.40.0",
+    "@playwright/test": "^1.33.0",
     "@types/axios": "^0.14.0",
     "@types/dedent": "^0.7.0",
     "@types/express": "^4.17.17",
@@ -1228,10 +1227,11 @@
     "dedent": "^0.7.0",
     "envalid": "^7.3.1",
     "express": "^4.18.2",
+    "fuzzysort": "^2.0.4",
     "mocha": "^10.2.0",
     "ovsx": "^0.8.2",
     "path-browserify": "^1.0.1",
-    "playwright": "^1.40.0",
+    "playwright": "^1.33.0",
     "progress": "^2.0.3",
     "semver": "^7.5.4"
   }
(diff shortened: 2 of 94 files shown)


vscode/package.json
```diff
@@ -1181,6 +1181,7 @@
     "classnames": "^2.3.2",
     "date-fns": "^2.30.0",
     "detect-indent": "^7.0.1",
+    "fuzzysort": "^2.0.4",
```
toolmantim, line 1184: I tried a few but this was the fastest and smallest

vscode/src/editor/utils/editor-context.ts
```diff
@@ -68,15 +119,26 @@ export async function getSymbolContextFile(query: string, maxResults = 10): Prom
...
 }
 
 function createContextFileFromUri(
     uri: vscode.Uri,
     source: ContextFileSource = 'user',
     type: ContextFileType = 'file',
     selectionRange?: vscode.Range,
-    kind?: SymbolKind
+    kind?: SymbolKind,
+    editorTab?: boolean
```
abeatrix, line 141:
```suggestion
```

Let's set kind to `editor-tab` instead of making a new field?

vscode/webviews/UserContextSelector.tsx
```diff
@@ -25,41 +25,76 @@ export const UserContextSelectorComponent: React.FunctionComponent<
...
+                <p className={styles.emptySymbolSearchTip}>
+                    <i className="codicon codicon-info" /> VS Code may require you to open files and install language
+                    extensions for accurate results
+                </p>
+            ) : null}
+
+            {contextSelection?.length ? (
+                <div className={classNames(styles.selectionsContainer)}>
+                    {contextSelection?.map((match, i) => {
+                        const icon =
```
abeatrix, line 68: is it intended to not display icon for file results?

abeatrix (changes requested):
Great job on supporting fuzzy search! It doesn't fixes the issue mentioned in https://github.com/sourcegraph/cody/issues/1837 as I mentioned in https://github.com/sourcegraph/cody/pull/1851 that the only one to get perfectly consistent results is to remove the limit for `MAX_RESULTS` which is currently set to 20 in your branch. You can see the last item is sometimes different when doing the same search when testing from your branch:

https://github.com/sourcegraph/cody/assets/68532117/06d2eb3f-2e75-4c35-809e-a94d93842711

Other than that, I think we should group the editorTab into ContextFileType and this should be good to go? Wdyt?

lib/shared/src/codebase-context/messages.ts
```diff
@@ -38,6 +38,7 @@ export interface ContextFile {
     source?: ContextFileSource
     type?: ContextFileType
     kind?: SymbolKind
+    editorTab?: boolean
```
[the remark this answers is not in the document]
toolmantim: Good catch. `editorTab` was accidentally left in, and wasn't even used. I've just removed it instead?

vscode/webviews/UserContextSelector.tsx
toolmantim: Yeah, unless we can do extension-specific file icons (there's no support for that in webviews is there?) there's no need, especially now that files aren't intermingled with symbols.

toolmantim:
> It doesn't fixes the issue mentioned in https://github.com/sourcegraph/cody/issues/1837 as I mentioned in https://github.com/sourcegraph/cody/pull/1851 that the only one to get perfectly consistent results is to remove the limit for MAX_RESULTS which is currently set to 20 in your branch.

That's unexpected… it should return every file in the workspace because of `undefined` here:
https://github.com/sourcegraph/cody/pull/1889/files#diff-315ecedea695e410a04a36543d5d788b07f05d75885cedffceb922c3f33cd569R18

I'll check what’s going on here… 🤔

abeatrix:
@toolmantim I found a solution that works and created a separate PR to merge it into this PR (to avoid making direct changes to your PR in case it doesn't work for you): https://github.com/sourcegraph/cody/pull/1917

toolmantim:
Thanks @abeatrix ! I've updated this with your code from #1917 but tweaked a little. I think this is ready to go!

abeatrix (approved):
Confirmed:
- [x] Results are consistently sorted
- [x] Support fuzzy search
- [x] Search is not case-sensitive
- [x] Search support extension search

https://github.com/sourcegraph/cody/assets/68532117/0d9b2d7f-1acb-4e24-ab68-90647e4bc6d5

I noticed the search results are not sorted alphabetically. If this is the intended behavior then I think it's good to go. Thanks for working on this!

PR #1892, opened 2023-11-27
Edit: Improve response consistency

umpox:

### DIFF
--- a/lib/shared/src/chat/recipes/fixup.ts
+++ b/lib/shared/src/chat/recipes/fixup.ts
@@ -34,9 +34,12 @@ export class Fixup implements Recipe {
 
         const promptText = this.getPrompt(fixupTask)
         const quarterFileContext = Math.floor(MAX_CURRENT_FILE_TOKENS / 4)
+        const promptPrefix = `<${this.multiplexerTopic}>`
 
         return newInteraction({
             text: promptText,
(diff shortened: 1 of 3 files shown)


abeatrix (changes requested):
Tested the document code-action,  /edit and /doc commands and everything worked as expected. also tested the Undo and retry code lenses. The Undo code lense worked, but the retry button would remove the code that we want to retry on?

See loom for details: https://www.loom.com/share/13ced7cf376c473cb43b4bb2ccebf4e4?sid=48a31348-97be-4aab-bfc4-cb1593260a0e

umpox:
Thanks @abeatrix. The issue I think is because we reuse the `selectionRange` for the task on retry. If this selectionRange expands then it won't match the original range.

It's a little tricky to get this right:
- I think we need a copy of the original range `task.originalRange`
- That range should never expand or shrink BUT it should shift down if new code is added above the range
- Use `task.originalRange` on retry.

It's not related to these changes (I think possibly a bug from the original retry logic), so I'll create a follow up PR for this one.

Looking at it now

abeatrix (approved):
Thanks for clarifying. Approved since the located bug is not from this PR.

umpox:
@abeatrix Fix is here: https://github.com/sourcegraph/cody/pull/1926!

Issue #1619, opened 2023-11-03
Tree-sitter: dynamically switch singleline completions to multiline if the first line indicates the block start


valerybugakov:
- Remove the `\n` stop character for completions
- Stream in the first line
- Use tree-sitter to understand if that would start a new block. Continue the token sampling if it does.

Slack threads for context: [#1](https://sourcegraph.slack.com/archives/C0LJP8WTB/p1698138794533699), [#2](https://sourcegraph.slack.com/archives/C052G9Y5Y8H/p1698935247150589)

PR #1893, opened 2023-11-27
Autocomplete: use a smaller percentage to detect the same string

valerybugakov:
## Context

- While working on dynamic multiline completion, I extensively tested autocomplete manually and noticed that helpful completions were hidden in multiple cases. It happened because `isAlmostTheSameString` decided that the generated completion matches the string after it. This PR lowers the percentage defined in `isAlmostTheSameString` to fix the issue.
- The main reason for this functionality was the lack of fill-in-the-middle support by the existing models, so LLMs generated completion, duplicating the current code. That's not the case anymore, and we may not need this logic or can lower the percentage further.
- To repro, open `src/completions/request-manager.ts`, go to line 210, delete the implementation of the `get` method keeping `public ge`, and generate the completion at the end of this string. It will be hidden.
- It's me extracting parts of the work on #1619 to several small PRs.

## Test plan

- Updated unit tests.

### DIFF
--- a/vscode/src/completions/text-processing/string-comparator.test.ts
+++ b/vscode/src/completions/text-processing/string-comparator.test.ts
@@ -6,7 +6,7 @@ describe('isAlmostTheSameString', () => {
     it.each([
         [true, '', ''],
         [true, 'return []', ' return []'],
-        [true, 'const abortController = new AbortController()', 'const networkAbortController = new AbortController()'],
+        [true, 'const abortController = new AbortController()', 'const networkController = new AbortController()'],
         [
             true,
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):
Seems reasonable!

> The main reason for this functionality was the lack of fill-in-the-middle support by the existing models, so LLMs generated completion, duplicating the current code. That's not the case anymore, and we may not need this logic or can lower the percentage further.

I remember I used the same reasoning when I adjusted the number from `0.3` to `0.2` but have never been bitten by a repeating completion since so this should be good 👍

PR #1894, opened 2023-11-27
Autocomplete: dynamically switch to multliline completions

valerybugakov:
## Context

- **TLDR**: we now use the completion's first line to determine if it starts the new multiline syntax node. If it does, we continue streaming until the completion is truncated or we reach the token sample limit.
- Closes https://github.com/sourcegraph/cody/issues/1619
- Adds a **feature flag** `cody-autocomplete-dynamic-multline-completions` to run an experiment on the updated logic.
- Adds an **experimental setting**, `cody.autocomplete.experimental.dynamicMultilineCompletions` to test the feature locally.
- Adds `completionPostProcessLogger` used for local debugging. It's disabled by default, and I will replace it with OpenTelemetry spans in a follow-up PR. I put a little effort into improving this logger because it's a throw-away code, but I need it here because debugging streaming-truncation bugs is much faster with it.
- Revamps the multiline truncation logic:
    - We no longer look for the next new sibling after the completion. This approach is unreliable because the parse tree often breaks because of incomplete code in the completion insert text.
    - Instead, we take the descendant at the current cursor position (or multiline trigger position), look for the farthest ancestor that starts at the same line, and then wait for a moment when completion generates more line than this syntax node has. This means we do not care about broken parse-tree after the node under the cursor:

```ts
// Start completion
function test() {
  █
}

// ---------------------------

// Continue streaming
function test() {
  █if (true) {
    console.log('hello world'
}

// ---------------------------

// Stop, because characters were added outside of `function test` syntax node boundaries
function test() {
  █if (true) {
    console.log('hello world'
}

function 
```
- ~~Need to merge https://github.com/sourcegraph/cody/pull/1891 first to reduce the diff in this PR.~~

## Screenshots

Group 2
Group 3
Group 4
Group 1

## Follow-ups

- https://github.com/sourcegraph/cody/issues/1923
- https://github.com/sourcegraph/cody/issues/1924
- https://github.com/sourcegraph/cody/issues/1920
- https://github.com/sourcegraph/cody/issues/1921

## Merged preparation PRs

- https://github.com/sourcegraph/cody/pull/1709
- https://github.com/sourcegraph/cody/pull/1784
- https://github.com/sourcegraph/cody/pull/1888
- https://github.com/sourcegraph/cody/pull/1891
- https://github.com/sourcegraph/cody/pull/1893

## Test plan

- Tested manually:
    - Ensure that code completions are generated without changes on this branch.
    - Ensure that code completions are dynamically switched to multiline based on the first line when `cody.autocomplete.experimental.dynamicMultilineCompletions` is set to true.
- Existing tests pass.
- The functionality is disabled by default. I'm going to add more unit tests for the new logic in a separate PR (see planned follow-ups section).

### DIFF
--- a/vscode/src/completions/post-process-logger.ts
+++ b/vscode/src/completions/post-process-logger.ts
@@ -1,0 +1,59 @@
+/* eslint-disable no-constant-condition, @typescript-eslint/no-empty-function */
+interface Log {
+    completionPostProcessId?: string
+    stage: string
+    text?: string
+    obj?: Record<string, unknown>
+    isCollapsedGroup?: boolean
+}
+
+class GroupedLogger {
+    private logs: Map<string, Log[]> = new Map()
+
+    public info(logObj: Log): void {
+        const id = logObj.completionPostProcessId
+
+        if (!id) {
+            return
+        }
+
+        if (!this.logs.has(id)) {
+            this.logs.set(id, [])
+        }
+
+        this.logs.get(id)?.push(logObj)
+    }
+
+    public flush(): void {
+        for (const [id, msgs] of this.logs) {
+            if (msgs.length < 2) {
+                break
+            }
+
+            const [{ stage, isCollapsedGroup }] = msgs
+            const groupStart = isCollapsedGroup ? console.groupCollapsed : console.group
+            groupStart(`Grouped Logs for ID ${id}-${stage}`)
+
+            for (const { stage, text = '', obj } of msgs) {
+                logStage(stage, text, obj)
+            }
+            console.groupEnd()
+        }
+
+        this.logs.clear()
+    }
+}
+
+/**
+ * @deprecated
+ * Will be replaced with an OpenTelemetry console exporter and manual spans.
+ */
+export const completionPostProcessLogger = true ? { info() {}, flush() {} } : new GroupedLogger()
+
+function logStage(stage: string, text: string, obj?: Record<string, unknown>): void {
+    console.log(
+        `%c${stage}%c: ${obj ? JSON.stringify(obj) : ''}\n${text}`,
+        'color: green; font-weight:bold',
+        'color: black; font-weight: normal;'
+    )
+}
--- a/vscode/src/completions/post-process-logger.ts
+++ b/vscode/src/completions/post-process-logger.ts
@@ -26,6 +26,8 @@ class GroupedLogger {
 
     public flush(): void {
         for (const [id, msgs] of this.logs) {
+            // Many requests are aborted right after the start stage, and seeing logs for them is not helpful.
+            // This check declutters the conosle output.
             if (msgs.length < 2) {
                 break
             }
--- a/vscode/src/completions/post-process-logger.ts
+++ b/vscode/src/completions/post-process-logger.ts
@@ -7,6 +7,10 @@ interface Log {
     isCollapsedGroup?: boolean
 }
 
+/**
+ * Used to log the stages of the completion post-processing during streaming.
+ * Logs for each response chunk are grouped together by `console.group` and `completionPostProcessId`.
+ */
 class GroupedLogger {
     private logs: Map<string, Log[]> = new Map()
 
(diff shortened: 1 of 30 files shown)


vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -0,0 +1,250 @@
...
+                                completionPostProcessLogger.info({
+                                    completionPostProcessId,
+                                    stage: 'isMultilineBasedOnFirstLine resolve',
+                                    text: completion.insertText,
+                                })
+
+                                stopStreamingAndUsePartialResponse({
+                                    ...completion,
+                                    insertText: completion.insertText,
+                                })
```
philipp-spiess, line 153: Can we do something here so that in the analytics we can see how often a single-line completion was upgraded to a multi-line request because of this?

vscode/src/completions/providers/fireworks.ts
```diff
@@ -184,14 +187,30 @@ export class FireworksProvider extends Provider {
             timeoutMs,
```
philipp-spiess, line 187: Don't we need to increase the `maxTokensToSample` for single-line requests for this to be useful?

philipp-spiess (commented):
I’m SO excited for this. Pre-emptive approval but will stamp once it's ready for review :)

vscode/src/completions/post-process-logger.ts
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        this.logs.get(id)?.push(logObj)
+    }
+
+    public flush(): void {
+        for (const [id, msgs] of this.logs) {
+            if (msgs.length < 2) {
+                break
+            }
```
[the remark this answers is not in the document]
valerybugakov: Many requests are aborted right after the `start` stage, and seeing logs for them is not helpful. 

Context: I put a little effort into improving this logger because it's a throw-away code, but I need it here because debugging streaming-truncation bugs is much faster with it.

vscode/src/completions/text-processing/process-inline-completions.ts
```diff
@@ -30,6 +31,12 @@ export function processInlineCompletions(
     items: ParsedCompletion[],
     params: ProcessInlineCompletionsParams
 ): InlineCompletionItemWithAnalytics[] {
+    completionPostProcessLogger.info({
+        completionPostProcessId: 'constant',
```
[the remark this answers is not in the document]
valerybugakov: Yeah, it was intentional. With the proper OpenTelemtry implementation, I will propagate the parent span context properly.

vscode/src/completions/providers/fetch-and-process-completions.ts
valerybugakov: Sure! It's on my TODO list.

vscode/src/completions/providers/fireworks.ts
valerybugakov: Great catch! Updated.

vscode/src/completions/post-process-logger.ts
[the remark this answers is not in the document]
valerybugakov: Added these comments to sources.

umpox:
@valerybugakov Should this trigger a multiline completion:

image

I ran it on the HumanEval multiline dataset we have - it's just Python but should show a significant difference here, but it didn't seem to be generating multiline completions for these cases.

I have `cody.autocomplete.experimental.dynamicMultilineCompletions` set to `true`. Do I need to do anything else?

philipp-spiess:
@umpox I think this example won't trigger a multi-line completion since it already has content that has a deeper indentation. 

Maybe that's a check we can bypass when manually triggering completions though? Can you try to comment this out: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/completions/detect-multiline.ts?L52%3A21-54%3A22

umpox:
@philipp-spiess Yep that fixes it:

image

There does seem to be something that seems off though, surely this should still generate a multi line completion?:

image

🤔

valerybugakov:
@umpox, thanks for sharing the screenshots. The issue should be fixed here: https://github.com/sourcegraph/cody/pull/1894/commits/ed20acb20bca0fe8d9e9ea491ed67b8afb002322

philipp-spiess (approved):
Let's ship it!

olafurpg:
This is huge! Looking forward to trying this out 😄

PR #1900, opened 2023-11-27
Edit: Faster `doc` workflow with more target context

umpox:
closes https://github.com/sourcegraph/cody/issues/1464

## Description

This PR:
- Streamlines the `doc` intent flow for fixup tasks. Previously we would treat these as "edit", which means we would try to fetch lots of context from the selected code. That is very rarely helpful, for documenting we really just want to take the surrounding code and the current selection.

I also investigating going the other way, and fetching relevant reference code for a documentable symbol. This confirmed to me that it's not necessarily helpful. It can lead to the LLM documenting code based on how it is used elsewhere rather than how it should be used. That might help for "explaining" code but not for documenting. (Related PR: https://github.com/sourcegraph/cody/pull/1904)

## Test plan

1. Run doc command via right click menu
2. Run doc command via sidebar
3. Run doc command via code action


abeatrix (approved):
Thanks for working on this. Everything looks good to me. I think the only change we will need to do is to remove the `currentDir` from the cody.json for the default doc command to reflect this change? https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/chat/prompts/cody.json?L8

umpox:
Thanks @abeatrix, will do that now

PR #1910, opened 2023-11-27
add: include text in dotCom chat events

abeatrix:
Spin off from https://github.com/sourcegraph/cody/pull/1856

Include chat prompt and response to `CodyVSCodeExtension:chatResponse` and `CodyVSCodeExtension:recipe` log events is users are connected to dotCom.

## Test plan

Check the Output channels to confirm the log events include prompt and LLM response IF you are connected to DotCom only.

### DEMO

You can see when the instance is switched to a non-DotCom instance, promptText and responseText are not included.

https://github.com/sourcegraph/cody/assets/68532117/516953f0-8f2f-4aa9-9421-cdbae7a3246c


akalia25 (approved):
Tested locally and confirmed that chat is only being logged on dotcom.

Thank you for adding this 🙇

PR #1913, opened 2023-11-27
telemetry: tee completions events to telemetry v2

bobheadxi:
This change makes a simple tee of all `logCompletionEvent` callsites (thanks for putting this all in one place!) to V2 telemetry using a very, very simple strategy to split the existing metadata into "safe" and "unsafe" metadata implemented in the new `splitSafeMetadata` helper method, which just takes numeric-looking metadata into `metadata`,  and puts everything else in `privateMetadata`.

This half-measure easily migrates a whole bunch of events around completions for us to start with, and for dotcom users where `privateMetadata` is always exported, allows us to have parity with the existing telemetry. This will hopefully allow the data team to start building out the new telemetry pipeline with more data.

## Test plan

There are a lot of unit tests on logging events, so I added to those and confirmed that we have parity. I couldn't quite get deep `expect. objectContaining ` working (i.e. `expect. objectContaining({ metadata: expect.objectContaining(...) })`, so I use `expect.anything()` on the V2 telemetry for now, but I don't think this matters because:

1. We just want a start on data parity, and we can follow up with fixes as needed
2. Unit tests on `splitSafeMetadata` hopefully confirms the adapter works

The important thing is that the existing tests confirm our existing data is intact even with this change.


vscode/src/completions/logger.ts
```diff
@@ -169,18 +172,86 @@ interface ErrorEventPayload {
...
+    // Use automatic splitting for now - make this manual as needed
+    const { metadata, privateMetadata } = splitSafeMetadata(params)
+    writeCompletionEvent('error', { version: 0, metadata, privateMetadata }, params)
+}
 /**
  * The following events are added to ensure the logging bookkeeping works as expected in production
  * and should not happen under normal circumstances.
  */
-export function logCompletionEvent(
+export function logCompletionBookingEvent(
```
philipp-spiess, line 254: What is a "booking" event? Generally not a fan of the code duplication here, what was the issue with the overloading? 😄

vscode/src/completions/logger.ts
```diff
@@ -189,10 +260,35 @@ export function logCompletionEvent(
...
-    telemetryService.log(`${logPrefix(extDetails.ide)}:completion:${name}`, params, { agent: true })
+    telemetryService.log(`${logPrefix(extDetails.ide)}:completion:${name}`, legacyParams, {
+        agent: true,
+        hasV2Event: true, // this helper translates the event for us
+    })
+    /**
+     * New telemetry automatically adds extension context - we do not need to
+     * include platform in the name of the event.
+     */
+    telemetryRecorder.recordEvent('completion', name, params)
```
philipp-spiess, line 291: Neat!

vscode/src/services/telemetry-v2.test.ts
```diff
@@ -0,0 +1,44 @@
+import { describe, expect, it } from 'vitest'
+
+import { splitSafeMetadata } from './telemetry-v2'
+
+describe('splitSafeMetadata', () => {
```
philipp-spiess, line 5: 👍

philipp-spiess (approved):
Awesome! My only gripe is that we now have a function that looks almost the same duplicated multiple times. Was there a particular reason why you removed the type overloading?

vscode/src/completions/logger.ts
bobheadxi: > What is a "booking" event?

Sorry I misread the docstring, it should be _bookkeeping_ 😆 will rename

bobheadxi: > Generally not a fan of the code duplication here, what was the issue with the overloading? 😄

With type overloading, we end up having to either:

- define a very generous base implementation (`logCompletionEvent(string, any)`)
- define a very specific base implementation (`logCompletionEvent('all' | 'possible' | 'values' | ..., any)`

The former is not acceptable to the new telemetry API (too broad and unknown), the latter is annoying to write and still doesn't give us the shape of the metadata. Making each explicit with a shared helper allows us to define the name and shape of events more explicitly, and in particular for V2 events, we can make explicit transformations into the V2 shape (e.g. turn a string into numeric enum) on an individual event basis

philipp-spiess: @bobheadxi I don't think I understand, with overloading we know exactly the shape of the event payload for each event type. The only "cumbersome" part is that the caller needs to pass the event name as a string instead of a different function. 🤔   E.g with the current setup, TS knows that when `'suggested'` is used as an argument for the first param, it needs to have a specific shape.

vscode/src/services/telemetry-v2.ts
```diff
@@ -125,3 +121,57 @@ export async function createOrUpdateTelemetryRecorderProvider(
...
+    for (const key in properties) {
+        if (!Object.hasOwn(properties, key)) {
+            continue
+        }
+
+        const value = properties[key]
+        switch (typeof value) {
+            case 'number':
+                safe[key] = value
+                break
```
philipp-spiess, line 158: Just to double check: ieee floating points are considered safe for our logger? I remember I asked this before but I forgot the answer :D

vscode/src/completions/logger.ts
bobheadxi: I'm referring to this: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/completions/logger.ts?L193:1-196:2

At the implementation of `logCompletionEvent`, we can't know how it was called, i.e. the specific overload is unknown. Knowing what is provided is a requirement in the SDK - i.e. the string must not extend name (name must be strict subset of string), and similarly for keys in the metadata object

vscode/src/services/telemetry-v2.test.ts
bobheadxi: Actually these tests were broken, TIL nested `it()` does not work, nor do you get a warning 😆 Just updated

philipp-spiess: Oh wow I didn't see this either 😰

nathan-downs (approved):

vscode/src/services/telemetry-v2.ts
bobheadxi: Sadly not yet - https://github.com/sourcegraph/sourcegraph/pull/58643 will enable this for dotcom and newer instances.

In the meantime, I'll update the exporter in this PR to round numbers to integers. This will make the durations unusable, but hopefully we can start processing events at least

vscode/src/services/telemetry-v2.test.ts
```diff
@@ -0,0 +1,115 @@
...
+            ],
+        }
+        const originalParameters = { ...parameters }
+        const { metadata, privateMetadata } = splitSafeMetadata(parameters)
+
+        // retains safe values in metadata
+        expect(metadata).toEqual({
+            'array.0.safe': 1,
+            'array.1.safe': 2,
+        })
```
bobheadxi, line 91: cc @nathan-downs @philipp-spiess - some interesting properties from my hack, looks like we handle arrays as well, so this mechanism can deeply unwrap everything more or less for customer instances to export

the original shape of the data will still be available in `privateMetadata`

bobheadxi:
Tested end-to-end: 
image

Issue #1819, opened 2023-11-20
bug: external file warning when clicking files included via @ in new chat UI


toolmantim:
### Version

0.16.0

### Describe the bug

Clicking file attachment tokens in the chat transcript causes VS Code to show a warning about an external application wanting to open the file:

<img width="781" alt="Screenshot 2023-11-20 at 2 27 37 pm" src="https://github.com/sourcegraph/cody/assets/153/9c938a77-8958-4f42-9ee4-aee511fa4378">

### Expected behavior

No warning would be shown

abeatrix:
This is because of a new setting added by vs code and the setting can only be configured by users for security reason: 
https://code.visualstudio.com/updates/v1_84#_confirmation-for-opening-protocol-links


> When a protocol link for a file or workspace opens in VS Code, a dialog will now ask for confirmation:

toolmantim:
@abeatrix would it be possible to send a message/command instead of using the `vscode://file` URI and avoid it?

abeatrix:
@toolmantim Yea that worked! Here is the PR: https://github.com/sourcegraph/cody/pull/1919

PR #1919, opened 2023-11-28
change: replace vscode links with command

abeatrix:
CLOSE: https://github.com/sourcegraph/cody/issues/1819

This PR replace the vscode links with custom protocol for the file links in chat view.

- Replace the vscode file links with a custom "cody.chat.open.file" command protocol when displaying file names in chat. This allows handling file opening in a more customizable way without the error message described in the issue.

## Test plan

Run the /explain command, and then click on the link after the @ sign:

https://github.com/sourcegraph/cody/assets/68532117/fdf7cad2-5eb3-48c5-926a-88fb56b751b3

### DIFF
--- a/vscode/src/chat/chat-view/ChatManager.ts
+++ b/vscode/src/chat/chat-view/ChatManager.ts
@@ -49,7 +49,8 @@ export class ChatManager implements vscode.Disposable {
             vscode.commands.registerCommand('cody.chat.history.clear', async () => this.clearHistory()),
             vscode.commands.registerCommand('cody.chat.history.delete', async item => this.clearHistory(item)),
             vscode.commands.registerCommand('cody.chat.panel.new', async () => this.createNewWebviewPanel()),
-            vscode.commands.registerCommand('cody.chat.panel.restore', (id, chat) => this.restorePanel(id, chat))
+            vscode.commands.registerCommand('cody.chat.panel.restore', (id, chat) => this.restorePanel(id, chat)),
+            vscode.commands.registerCommand('cody.chat.open.file', async fsPath => this.openFileFromChat(fsPath))
         )
 
         // Register config change listener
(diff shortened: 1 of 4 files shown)


vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -228,6 +229,16 @@ export class ChatManager implements vscode.Disposable {
             .catch(error => console.error(error))
     }
 
+    private async openFileFromChat(fsPath: string): Promise<void> {
+        const rangeIndex = fsPath.indexOf(':range:')
+        const range = rangeIndex ? fsPath.slice(Math.max(0, rangeIndex + 7)) : undefined
+        const filteredFsPath = range ? fsPath.slice(0, rangeIndex) : fsPath
+        const uri = vscode.Uri.file(filteredFsPath)
+        const doc = await vscode.workspace.openTextDocument(uri)
+        const viewColumn = vscode.ViewColumn.Beside - 2 > 1 ? vscode.ViewColumn.Beside - 2 : vscode.ViewColumn.Beside
```
toolmantim, line 238: I don't quite understand this line… but based on the video I think we should just let it open in a tab in the active group (which is what it did before by default, and same as what the file explorer does by default).

File explorer lets you command-click to open a file in a new tab group, how about we support that instead? (not required for this PR though)

toolmantim (changes requested):
Awesome! I've one small suggested change on the tab group behaviour.

vscode/src/chat/chat-view/ChatManager.ts
abeatrix: i've updated to open it in the current active editor instead:

https://github.com/sourcegraph/cody/assets/68532117/ba90ee16-0559-4d13-b4fa-7bc333171899

> File explorer lets you command-click to open a file in a new tab group, how about we support that instead? (not required for this PR though)

I can't reproduce this behavior using the built-in `vscode.open` command, so we can file a feature request for this to work on after GA if that sounds good to you?

toolmantim (approved):

PR #1925, opened 2023-11-28
example: cody pro page

abeatrix:
Example of clicking on a new cody pro tree item to open the cody pro webview:
image


toolmantim:
thank you @abeatrix! 🙏 Leave this with me to play with.

PR #1928, opened 2023-11-28
Agent: use `uri: vscode.Uri` instead of `filePath: string`

olafurpg:
Previously, the agent used a property named `filePath` as the unique string identifier of documents. This was problematic for several reasons, including

- There's no idiomatic way to encode virtual files
- File paths have different formatting in Windows and Linux/macOS

This PR does a big refactoring around the agent internals so that we use URIs instead of `filePath`. URIs solve both problems above.

The motivation to prioritize this change was that bfg wasn't loading correctly in the evaluation because of a bug related to handling of relative file paths. This bug was trivial to prevent after moving to URIs.

The diff includes a few minor `evaluate-autocomplete`-related changes that I included since I wanted to validate that bfg evaluation works correctly after the refactoring.

## Test plan

Green CI.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -16,6 +16,7 @@ import { NoOpTelemetryRecorderProvider } from '@sourcegraph/cody-shared/src/tele
 import { TelemetryEventParameters } from '@sourcegraph/telemetry'
 
 import { activate } from '../../vscode/src/extension.node'
+import { TextDocumentWithUri } from '../../vscode/src/jsonrpc/TextDocumentWithUri'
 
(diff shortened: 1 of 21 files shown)


agent/src/agent.ts
```diff
@@ -30,6 +31,15 @@ const secretStorage = new Map<string, string>()
 
 export async function initializeVscodeExtension(workspaceRoot: vscode.Uri): Promise<void> {
     const paths = envPaths('Cody')
+    try {
```
olafurpg, line 34: This change is unrelated to URIs but it was required to get bfg working correctly in the evaluation

philipp-spiess (approved):

PR #1961, opened 2023-11-28
Roll cody-engine to v5.2.9998.

dominiccooney:
This rolls Cody engine, including an improved tsx matcher. This does *not* include local embeddings because of a release build issue with that.

## Test plan

CI tests.


abeatrix (approved):
CI says 👍

PR #1967, opened 2023-11-29
fix: format chat panel titles

abeatrix:
fix: clean up chat panel titles

- Add a new helper function `getChatPanelTitle()` to generate truncated titles for chat panels. 
- Update usages in `ChatPanelProvider` and `SimpleChatPanelProvider` to use this new helper.
- Add test for the new helper function

### Before

- The chat panel title in editor and tree views are different for the same chat session
- Display markdown links
- Some titles are truncated, and some are not

image

### After

- The chat panel title in editor and tree views are the same for the same chat session
- Display links without markdown format 
- All titles are truncated correctly

image

### DIFF
--- a/vscode/src/chat/chat-view/chat-helpers.ts
+++ b/vscode/src/chat/chat-view/chat-helpers.ts
@@ -109,3 +109,16 @@ function activeEditorSelectionRangeToRange(range?: ActiveTextEditorSelectionRang
     }
     return new vscode.Range(range.start.line, range.start.character, range.end.line, range.end.character)
 }
+
+export function getChatPanelTitle(lastDisplayText?: string): string {
+    if (!lastDisplayText) {
+        return 'New Chat'
+    }
+    // Regex to remove the markdown formatted links with this format: '[_@FILENAME_]()'
+    const MARKDOWN_LINK_REGEX = /\[_(.+?)_]\((.+?)\)/g
+    lastDisplayText = lastDisplayText.replaceAll(MARKDOWN_LINK_REGEX, '$1')?.trim()
+    // truncate firstQuestion to first 25 chars
+    const title = lastDisplayText.length > 25 ? `${lastDisplayText?.slice(0, 25)?.trim()}...` : lastDisplayText
+    // Display command key only
+    return title.startsWith('/') ? title.split(' ')[0] : title
+}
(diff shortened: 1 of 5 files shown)


vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -109,3 +109,16 @@ function activeEditorSelectionRangeToRange(range?: ActiveTextEditorSelectionRang
     }
     return new vscode.Range(range.start.line, range.start.character, range.end.line, range.end.character)
 }
+
+export function getChatPanelTitle(lastDisplayText?: string): string {
+    if (!lastDisplayText) {
+        return 'New Chat'
+    }
+    // Regex to remove the markdown formatted links with this format: '[_@FILENAME_]()'
```
toolmantim, line 117: Instead of removing the file name altogether, can keep it in? Otherwise it's hard to distinguish `/explain @somefile` and `/explain @otherfile` etc.

e.g. `'[_@FILENAME_]()'` → `'@FILENAME'`

abeatrix: updated to keep the string after the command key

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -109,3 +109,14 @@ function activeEditorSelectionRangeToRange(range?: ActiveTextEditorSelectionRang
...
+    const MARKDOWN_LINK_REGEX = /\[_(.+?)_]\((.+?)\)/g
+    lastDisplayText = lastDisplayText.replaceAll(MARKDOWN_LINK_REGEX, '$1')?.trim()
+    // truncate title that is too long
+    return lastDisplayText.length > 25 ? `${lastDisplayText?.slice(0, 25)?.trim()}...` : lastDisplayText
```
toolmantim, line 121: Sorry, only spotted this now with additional testing.

I don't think we want to do any truncation:

Screenshot 2023-11-29 at 5 31 45 pm

I believe we can just let VS Code handle the truncation where it needs to:

Screenshot 2023-11-29 at 5 30 52 pm

toolmantim (commented):
Looking good! Spotted one last thing.

toolmantim (approved):

PR #1976, opened 2023-11-29
setup gcs bucket in eventlogger, add sync

abeatrix:
Replaces https://github.com/sourcegraph/cody/pull/1856

This PR includes the following changes

- Implemented [uploader code](https://cloud.google.com/storage/docs/uploading-objects-from-memory#storage-upload-object-from-memory-nodejs) to sync chat history to GCS bucket in the shared EventLoggers
- Confirmed the code works as expected, see [this slack thread for details](https://sourcegraph.slack.com/archives/C065NKGD1BM/p1701233415747499?thread_ts=1700702538.931369&cid=C065NKGD1BM)

Confirm with the data team on how we can publish the extension with the credentials for the bucket package.

This added a `sync` function to the event logger for uploading the local chat transcript (once in the past 7 days) when the feature flag is enabled.

## Test plan

See https://sourcegraph.slack.com/archives/C065NKGD1BM/p1701233415747499?thread_ts=1700702538.931369&cid=C065NKGD1BM


vscode/src/services/telemetry.ts
```diff
@@ -146,3 +163,93 @@ export function logPrefix(ide: 'VSCode' | 'JetBrains' | 'Neovim' | 'Emacs' | und
...
+            chat => new Date(chat[1].lastInteractionTimestamp).getTime() > lastStoredTimestamp
+        )
+
+        const lastSyncedTranscriptTimestamp = filteredChats.at(-1)?.[1]?.lastInteractionTimestamp
+        if (lastSyncedTranscriptTimestamp) {
+            // File location format: "cody/vscode/chatTranscript/YYYY/MM/DD/anonymousUserID.json"
+            const fileLocation = `cody/vscode/chatTranscript/${TODAYS_DATE()}/${globalAnonymousUserID}.json`
+
+            // Sync and store the chats and timestamp
+            await syncChat(JSON.stringify(filteredChats), fileLocation)
```
akalia25, line 218: Would it be possible to send the chats as a JSONL instead of JSON? BigQuery has [limitations](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json#limitations) where it only accepts a GCS <> BQ connection when the data is newline delimited. Each JSON object must be on a separate line in the file.  i.e.
`{...}  \n
{...}  \n
{...} \n` 

If this would be a huge lift, we can investigate something on the data-eng side to transform the json :) 🙇

abeatrix: Yea just updated the code to reflect this :)

akalia25: amazing 🎉 !! Is there any way to send an updated file to the bucket with the new JSONL and I can verify things on the BQ side? :)

abeatrix: Oh since this is uploading to the storage directly it shouldn't reach BQ at all (I thought that's what Dan wanted)?

akalia25: You're right - the chat transcript data would not go directly into BigQuery and is going to be stored onto cloud storage. I'm creating an external table that connects to the Cloud Storage bucket holding the transcript data ([ref](https://cloud.google.com/bigquery/docs/external-data-cloud-storage)). This would allow the team to access the transcripts through BigQuery or the raw files in Cloud Storage, with the external table providing a schema overlay.

The transcripts would reside in their own table, we get the benefit of keeping this large data separate from our production table but also allowing us to join against it when needed for the AI/ML folks :) (i.e. [accomplish this](https://sourcegraph.slack.com/archives/C065NKGD1BM/p1701695420457929?thread_ts=1700702538.931369&cid=C065NKGD1BM))

Does that help? :)

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -490,13 +490,12 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
...
                             if (codeCount?.charCount) {
-                                // const metadata = lastInteraction?.getHumanMessage().metadata
-                                telemetryService.log(
-                                    'CodyVSCodeExtension:chatResponse:hasCode',
-                                    { ...codeCount, requestID },
-                                    { hasV2Event: true }
-                                )
                                 telemetryRecorder.recordEvent('cody.chatResponse.new', 'hasCode', {
                                     metadata: {
                                         ...codeCount,
```
nathan-downs, line 501: If there are V2 events that exist can we add the changes to these as well? So adding `requestID` to the metadata of the `telemetryRecorder.recordEvent()` call. The legacy telemtryService will be deprecated at some point, so we want to make sure not to lose anything.

bobheadxi: Note that there is a special field in `recordEvent` for `requestID`, `interactionID`

vscode/src/services/telemetry.ts
```diff
@@ -146,3 +163,95 @@ export function logPrefix(ide: 'VSCode' | 'JetBrains' | 'Neovim' | 'Emacs' | und
...
+        }
+
+        const filteredChats = Object.entries(chatFromStore).filter(
+            chat => new Date(chat[1].lastInteractionTimestamp).getTime() > lastStoredTimestamp
+        )
+
+        const lastSyncedTranscriptTimestamp = filteredChats.at(-1)?.[1]?.lastInteractionTimestamp
+        if (lastSyncedTranscriptTimestamp) {
+            // File location format: "cody/vscode/chatTranscript/YYYY/MM/DD/anonymousUserID.json"
+            const fileLocation = `cody/vscode/chatTranscript/${TODAYS_DATE()}/${globalAnonymousUserID}.json`
```
akalia25, line 215: Could we make the `fileLocation` as such: `cody/vscode/{event_name}/YYYY/MM/DD/{anon_user_id}/{uuidv4().slice(0, 8)}.json`. 

This would help us:
1.  isolate the different types of events within the bucket (i.e. `syncChatInteraction` vs `chatResponse`, `chatSubmitted`).
2. By adding `{uuid}`, we'd eliminate the chance of overwriting any data we stored in the past

vscode/src/services/telemetry.ts
```diff
@@ -146,3 +163,95 @@ export function logPrefix(ide: 'VSCode' | 'JetBrains' | 'Neovim' | 'Emacs' | und
...
+        logEvent(`${eventName}:failed`, { error: `${error}` })
+    }
+}
+
+/**
+ * Syncs a chat interaction to remote storage.
+ * @param endpoint - The endpoint URL.
+ * @param interaction - The chat interaction to sync.
+ */
+export async function syncChatInteraction(endpoint?: string, interaction?: Interaction): Promise<void> {
```
dadlerj, line 235: Why aren't we calling `localStorage.lastSyncedTimestamp(new Date(...))` in this function? I know there's a goal to record old chat history if no other syncs have been logged, but this setup seems to ensure that we'll log both (1) all chats as they occur _and_ (2) every-7-day historical transcripts going forward. 

For the future, don't we only want one or the other?

vscode/src/services/telemetry.ts
```diff
@@ -146,3 +163,95 @@ export function logPrefix(ide: 'VSCode' | 'JetBrains' | 'Neovim' | 'Emacs' | und
...
+    if (!endpoint || !isDotCom(endpoint) || !interaction || !globalAnonymousUserID) {
+        return
+    }
+    const eventName = 'CodyVSCodeExtension:syncChatInteraction'
+
+    try {
+        // File location format: "cody/vscode/chatInteraction/anonymousUserID/YYYY/MM/DD/interactionTimestamp.json"
+        const interactionTimestamp = interaction.timestamp
+        // format the date to yyyy/MM/dd
+        const fileLocation = `cody/vscode/chatInteraction/${globalAnonymousUserID}/${TODAYS_DATE()}/${interactionTimestamp}.json`
```
akalia25, line 245: I think this `fileLocation` has the ordering mixed up should be `TODAYS_DATE` first and then `globalAnonymousUserID` ?

vscode/src/services/telemetry.ts
```diff
@@ -146,3 +163,95 @@ export function logPrefix(ide: 'VSCode' | 'JetBrains' | 'Neovim' | 'Emacs' | und
...
+
+        const lastSyncedTranscriptTimestamp = filteredChats.at(-1)?.[1]?.lastInteractionTimestamp
+        if (lastSyncedTranscriptTimestamp) {
+            // File location format: "cody/vscode/chatTranscript/YYYY/MM/DD/anonymousUserID.json"
+            const fileLocation = `cody/vscode/chatTranscript/${TODAYS_DATE()}/${globalAnonymousUserID}.json`
+
+            // Sync and store the chats and timestamp
+            // Due to BigQuery limitation, all JSON data must be newline delimited.
+            // This converts the data to proper JSONL format by putting each chat object JSON on its own line.
+            await syncChat(filteredChats.map(chat => JSON.stringify(chat)).join('\n'), fileLocation)
```
akalia25, line 220: would it be possible to add the `fileLocation` as a top level property within the JSON? This would help with mapping it back to the event, if it also lives within the JSON data. Thoughts? :)

lib/shared/src/telemetry/EventLogger.ts
```diff
@@ -149,4 +151,20 @@ export class EventLogger {
             })
             .catch(error => console.error('Error logging event', error))
     }
+
+    /**
+     * Uploads the given stringified chat history string to Google Cloud Storage.
+     * @param history - The history string to upload.
+     * @param filePath - The path of the file to upload the history to.
+     */
+    public async sync(history: string, fileLocation: string): Promise<void> {
```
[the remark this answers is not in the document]
evict: Can we add some server-side code to handle the transcript upload? That way we can ensure there's proper authentication in place, also to the bucket writes.

evict (changes requested):
How do we prevent arbitrary writes to the bucket? We cannot store the key on the client-side.

Can we add the transcript creation not to the DotCom backend? See comment: https://github.com/sourcegraph/cody/pull/1976/files#r1422891856.

dadlerj:
@evict this PR is only a short-term solution to the question at hand. Sharing a thread with you now in Slack.

lib/shared/src/telemetry/EventLogger.ts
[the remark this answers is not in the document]
dadlerj: Just FYI, @evict and I discussed—this is a short-term solution and will be replaced in the coming months. For now, the risk of high costs due to unauthenticated uploads is acceptable and @akalia25 is setting alerts.

Issue #1973, opened 2023-11-29
bug: hitting backspace after @-including a file shouldn't lose the file context


toolmantim:
### Version

main (2ecded4fd9ae240edbc3233f545b32205e30abb3)

### Describe the bug

If you want to put a character after an at-file token (e.g. "What is @somefile.js?") you end up losing the file context (and it shows "No matching files found" in the process):

https://github.com/sourcegraph/cody/assets/153/c47c24ad-a8fa-46ee-85e8-7849989a4e7a

### Expected behavior

Instead it should work more like Slack:

https://github.com/sourcegraph/cody/assets/153/01015221-4940-49ab-b3ac-9246431bcd6b

Issue #1968, opened 2023-11-29
bug: Chat history gets overwritten when there is a new chat


sqs:
### Version

v0.17.1701184011    

### Describe the bug

Cody chat history gets overwritten when you have an active chat as well.

### Expected behavior

See the video.

1. Start a new chat
2. Go to chat history and open an old chat
3. (Separate bug: the `/`-menu shows up and I can't dismiss it)
4. Go back to new chat
5. Go back to chat history and open an old chat

In step 5, the old chat's history has been overwritten with the new message from the new chat.

https://github.com/sourcegraph/cody/assets/1976/e9ea614e-0470-4e67-8c9d-52662a1ee7a7

PR #1980, opened 2023-11-29
fix: handle chat command selection logic

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1973
Address the `/` command bug mentioned in https://github.com/sourcegraph/cody/issues/1968

- Update the chat command selection logic to only filter commands when the input starts with a '/' instead of splitting the input value. (re https://github.com/sourcegraph/cody/issues/1968)
- Also update the file completion logic to better preserve the existing input value. (re https://github.com/sourcegraph/cody/issues/1973)

### Before 

image

### After 

image

### Before

slash command can be activated when you paste a paragraph with space to the chat inbox in a new chat, even if the paragraph doesn't starts with a /

image

https://github.com/sourcegraph/cody/assets/68532117/b9453516-3fbf-423c-bb1f-e2a528fb5e6d

### After

Pasting anything that doesn't start with / will not trigger the selection box

image

https://github.com/sourcegraph/cody/assets/68532117/14ba71f0-af7e-497b-84f0-ad1daf6bb97e


toolmantim (commented):
This seems to work, but the chat transcript seems to render the @-file token inconsistently:

Screenshot 2023-11-29 at 10 20 21 pm

Ideally we'd make the context file selector disappear when they type that additional character (e.g. in the below video, it would disappear again when they type `?`). It sticking on and capturing Enter events feels a bit buggy.

https://github.com/sourcegraph/cody/assets/153/821c2c3d-b590-4da2-8b81-277e00c627e3

abeatrix:
> This seems to work, but the chat transcript seems to render the @-file token inconsistently:
> 
> Screenshot 2023-11-29 at 10 20 21 pm
> Ideally we'd make the context file selector disappear when they type that additional character (e.g. in the below video, it would disappear again when they type `?`). It sticking on and capturing Enter events feels a bit buggy.
> 
>  Screen.Recording.2023-11-29.at.10.23.25.pm.mov

@toolmantim can you take another look at the latest commit and let me know if that works the way you have in mind please? 🙇‍♀️ 

https://github.com/sourcegraph/cody/assets/68532117/da591059-631a-4ef3-8e21-45782f8631d0

toolmantim (approved):
Works great! Thanks @abeatrix ✨

PR #1981, opened 2023-11-29
Reduce noise in test logs

olafurpg:
Previously, the tests always printed out a stack trace to stderr when running `src/utils.test.ts` even if the tests passed. I've been debugging CI failures lately and this error message has repeatedly been a red herring. This PR refactors the implementation so that we use a separate function for testing purposes that doesn't log to stderr.

## Test plan
Green CI.


abeatrix (approved):
Thank you!

PR #1985, opened 2023-11-29
Autocomplete: fix multiline trigger position calculation

valerybugakov:
## Context

- Fixes the multiline trigger position calculation logic and adds tree-sitter-powered tests to avoid checking the calculated position ourselves.
- Part of #1619 
- Follow-up for #1894 

## Test plan

Added unit tests.


dominiccooney (approved):
Looks good. I love these tests with a readable interpretation of where the cursor is.

PR #1987, opened 2023-11-29
Autocomplete: Use OpenTelemtry traceparent logic

philipp-spiess:
Let's not roll our own if we don't have to 😅 

## Test plan
 
- Add yourself to the `cody-autocomplete-tracing` feature flag
- Observe that a `traceparent` is appended to the request (I used console.log to validate since my proxyman isn't working anymore)


abeatrix (approved):
Nice! Does this work with chat?

PR #1998, opened 2023-11-29
change: display sorted commands in ui

abeatrix:
small UI update: sort commands and prompts alphabetically when on display

Sort commands and prompts alphabetically in the commands menu and chat to improve discoverability and visibility.

### Before 

Commands are currently not displayed as sorted, making it hard to find when the command list grows longer 

image

image

### After

Commands are displayed as sorted in both chat selection box and command menu

image

image


toolmantim (approved):
This is a great.

Even better would be two groups, with the separators "built-in commands" and "custom commands". So the built-in 6 are at the top, followed by custom commands, with each group being alphabetically sorted.

Issue #1999, opened 2023-11-30
Autocomplete: generate multiline completions for multiline argument lists and arrays


valerybugakov:
Examples:

```ts
generateContext(
    █
)

const oddNumbers = [
    █
]
```

With the improved context, this change should lead to more correct completions from the type system POV.

PR #2000, opened 2023-11-30
Autocomplete: extended multiline triggers inside brackets

valerybugakov:
## Context

- Enables generation of the multiline completions inside brackets when the cursor is on the new empty line inside them. This change is behind the `dynamicMultilineCompletions` feature flag.
- With the improved context, this change should lead to more correct completions from the type system POV.
- Improved units around detecting the multiline trigger with and without the `dynamicMultilineCompletions` feature flag.
- Builds on top of #1894 
- Closes https://github.com/sourcegraph/cody/issues/1999

```ts
generateContext(
    █
)

const oddNumbers = [
    █
]
```

## Screenshots

multiline-arguments

multiline-arrays

## Test plan

- Added new unit tests.
- Test it locally by setting `cody.autocomplete.experimental.dynamicMultilineCompletions` to true and generating completions similar to examples on screenshots.

### DIFF
--- a/vscode/src/completions/detect-multiline.ts
+++ b/vscode/src/completions/detect-multiline.ts
@@ -69,7 +69,7 @@ export function detectMultiline(params: DetectMultilineParams): DetectMultilineR
         currentLinePrefix.trim() === '' &&
         currentLineSuffix.trim() === '' &&
         openingBracketMatch &&
-        // Only trigger multiline suggestions when the new current line is indented
+        // Only trigger multiline suggestions when the next non-empty line is indented the same or less
         indentation(prevNonEmptyLine) < indentation(currentLinePrefix) &&
         // Only trigger multiline suggestions when the next non-empty line is indented less
         // than the block start line (the newly created block is empty).
@@ -92,7 +92,7 @@ export function detectMultiline(params: DetectMultilineParams): DetectMultilineR
         currentLineSuffix.trim() === '' &&
         // Only trigger multiline suggestions for the beginning of blocks
         isBlockStartActive &&
-        // Only trigger multiline suggestions when the new current line is indented
+        // Only trigger multiline suggestions when the next non-empty line is indented the same or less
         indentation(prevNonEmptyLine) < indentation(currentLinePrefix) &&
         // Only trigger multiline suggestions when the next non-empty line is indented less
         // than the block start line (the newly created block is empty).
--- a/vscode/src/completions/detect-multiline.ts
+++ b/vscode/src/completions/detect-multiline.ts
@@ -16,7 +16,7 @@ import {
 interface DetectMultilineParams {
     docContext: LinesContext & DocumentDependentContext
     languageId: string
-    dynamicMultlilineCompletions: boolean
+    dynamicMultilineCompletions: boolean
     position: Position
 }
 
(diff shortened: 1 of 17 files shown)


vscode/src/completions/detect-multiline.ts
```diff
@@ -54,15 +54,28 @@ export function detectMultiline(params: DetectMultilineParams): DetectMultilineR
...
         indentation(currentLinePrefix) >= indentation(nextNonEmptyLine)
-    ) {
+
+    const isNewLineOpeningBracketMatch =
+        currentLinePrefix.trim() === '' &&
+        currentLineSuffix.trim() === '' &&
+        openingBracketMatch &&
+        // Only trigger multiline suggestions when the new current line is indented
+        indentation(prevNonEmptyLine) < indentation(currentLinePrefix) &&
+        // Only trigger multiline suggestions when the next non-empty line is indented less
```
dominiccooney, line 74:
```suggestion
        // Only trigger multiline suggestions when the next non-empty line is indented the same or less
```

vscode/src/completions/detect-multiline.ts
```diff
@@ -54,15 +54,28 @@ export function detectMultiline(params: DetectMultilineParams): DetectMultilineR
...
+        // than the block start line (the newly created block is empty).
+        indentation(prevNonEmptyLine) >= indentation(nextNonEmptyLine)
+
+    if ((dynamicMultlilineCompletions && isNewLineOpeningBracketMatch) || isSameLineOpeningBracketMatch) {
```
dominiccooney, line 78: There's a number of mis-spellings of `dynamicMultilineCompletions` as `dynamicMultlilineCompletions` throughout.

vscode/src/completions/get-current-doc-context.test.ts
```diff
@@ -258,39 +270,122 @@ describe('getCurrentDocContext', () => {
...
+                    def greatest_common_divisor(a, b):
+                        if a == 0:█
+                `,
+                dedent`
+                    def bubbleSort(arr):
+                        n = len(arr)
+                        for i in range(n-1):
+                            █
+                `,
+            ])('detect the multiline trigger for python', code => {
```
dominiccooney, line 288:
```suggestion
            ])('should detect the multiline trigger for python', code => {
```

vscode/src/completions/get-current-doc-context.test.ts
```diff
@@ -258,39 +270,122 @@ describe('getCurrentDocContext', () => {
...
 
-            const { multilineTrigger, multilineTriggerPosition } = getCurrentDocContext({
-                document,
-                position,
-                maxPrefixLength: 100,
-                maxSuffixLength: 100,
-                dynamicMultlilineCompletions: true,
+            it.each([
+                {
+                    code: 'const restuls = {█',
```
dominiccooney, line 300:
```suggestion
                    code: 'const result = {█',
```

for consistency with the rest

vscode/src/completions/get-current-doc-context.test.ts
```diff
@@ -258,39 +270,122 @@ describe('getCurrentDocContext', () => {
...
+                    triggerPosition: { line: 0, character: 16 },
```
dominiccooney, line 301: You don't have to do this in this change, but these tests would be more readable if you used a visual indicator for the trigger position. Like a special line that has a `^` pointing at the trigger position or something.

dominiccooney (approved):
Looks good, there's just a few typos which need cleaning up.

vscode/src/completions/get-current-doc-context.test.ts
valerybugakov: Agree. Actually, the `triggerPosition` property is redundant here because we use tree-sitter to find the multiline trigger character, so I went ahead and deleted it. It should not be necessary to check if these values are valid with our eyes.

PR #2002, opened 2023-11-30
Change "Restart Chat Session" icon and add a confirmation

toolmantim:
This changes the current icon from `reload` (which is too close to a browser reload) to `clear-all` (which isn't perfect but looks more destructive). It also adds a confirmation dialog:

| Before | After |
| - | - |
| Screenshot 2023-11-30 at 4 09 59 pm | Screenshot 2023-11-30 at 4 10 24 pm |

| Confirmation |
| - |
| Screenshot 2023-11-30 at 4 03 17 pm |

## Test plan

- Viewed the button
- Clicked the button, verify prompt cancel does nothing
- Clicked the button, verify "Restart Chat Session" does the action
- Typed `/reset` and verify it doesn't ask for confirmation


umpox (approved):

abeatrix:
Nice!

PR #2004, opened 2023-11-30
Edit: Fix edit command palette

umpox:
## Description

Noticed by @toolmantim that it wasn't showing.

It is because the text focus moves to the command palette.

We show an error when no file is open so this should be fine to always show when activated

## Test plan

1. Open command palette
2. Check the command shows


toolmantim (approved):
Nice! Works well. Don't forget a changelog.

PR #2005, opened 2023-11-30
Edit: Only affix non-whitespace changes to ranges

umpox:
## Description

We want to allow affixing non-whitespace characters and including that in the edit selection range (e.g. `export` before a function, or a JSDoc above an identifier).

We do not want to affix whitespace characters, as this may be from the user moving their cursor onto a different line, or repositioning the code.

### Before

https://github.com/sourcegraph/cody/assets/9516420/aab05bc5-4c19-4a93-ac41-53de10b3dc60

### After

https://github.com/sourcegraph/cody/assets/9516420/2db9a97f-85d0-41a2-b905-e151938135ce

## Test plan

1. Create fixup
2. Move cursor to start of range
3. Enter non-whitespace characters and see range expands

1. Create fixup
2. Move cursor to start of range
3. Enter whitespace characters and see range does not expand


valerybugakov (approved):
Makes sense! Thanks for attaching the vids 🙌

Issue #1924, opened 2023-11-28
Autocomplete: add heuristics to limit the number of huge unhelpful completions


valerybugakov:
- Part of https://github.com/sourcegraph/cody/issues/1619
- Follow up for https://github.com/sourcegraph/cody/pull/1894

For example the following trigger position can generate huge unhelpful completion until it reaches the token sample limit:

```ts
class █
```

Issue #1920, opened 2023-11-28
Autocomplete: add unit tests for dynamic switch to multliline completions


valerybugakov:
- Part of https://github.com/sourcegraph/cody/issues/1619
- Follow up for https://github.com/sourcegraph/cody/pull/1894

beyang:
Closeable?

valerybugakov:
Not yet. I'll work on the implementation and close with a PR this week.

PR #2006, opened 2023-11-30
Autocomplete: stop some dynamic multiline completions early

valerybugakov:
## Context

- Disabled dynamic multiline completions for class declarations at the root of the file because they are most often unhelpful and huge.
- Closes https://github.com/sourcegraph/cody/issues/1924

## Test plan

- Tested manually:
  - Set the `cody.autocomplete.experimental.dynamicMultilineCompletions` setting to `true`.
  - Generate a completion for `class █` as the root of any Typescript file.
  - Verify that a singleline completion is generated.
  - Verify that `function █` generates a multiline completion.
- Going to add unit tests for this case here: https://github.com/sourcegraph/cody/issues/1920


umpox (approved):
Nice!

PR #2013, opened 2023-11-30
BFG: increase resilience

olafurpg:
Previously, our BFG integration had a chance of failing in certain cases. Specifically, I was seeing errors when we sent concurrent `bfg/revision/didChange` requests. This PR adds a try/catch to that block along with a few other defensive changes to make BFG more resilient overall.

## Test plan

Manually tested by running the evaluation suite with parallel `revision/didChange` results and confirmed that it recovers.

### DIFF
--- a/vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
+++ b/vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
@@ -9,14 +9,10 @@ import { captureException } from '../../../../services/sentry/sentry'
 import { getContextRange } from '../../../doc-context-getters'
 import { ContextRetriever, ContextRetrieverOptions, ContextSnippet } from '../../../types'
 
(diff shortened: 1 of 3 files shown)


vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
```diff
@@ -9,14 +9,10 @@ import { captureException } from '../../../../services/sentry/sentry'
 import { getContextRange } from '../../../doc-context-getters'
 import { ContextRetriever, ContextRetrieverOptions, ContextSnippet } from '../../../types'
 
-// This promise is only used for testing purposes. We don't await on the
-// indexing request during autocomplete because we want autocomplete to respond
-// quickly even while BFG is indexing.
```
olafurpg, line 14: This was outdated, we don't use it for testing purposes anymore because we have a setting to control it instead.

PR #2016, opened 2023-11-30
fix: revive simple chat panel

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1701316056167949

- Revived simple chat panel when extension is reactivated by registering existing webviewPanel and setting chatID. 
- Added error handling in revive method to dispose panel and log debug message on failure.
- Updated related methods to handle null chatPanelsManager.

## Test plan

1. Enable SimpleChatPanel
2. Ask Cody a question and keep the new chat panel opened
3. Reload VS Code
4. You should see the chat has been restored
5. Check Output Channel to confirm the view is created by SimpleChatPanel 

## Demo

https://github.com/sourcegraph/cody/assets/68532117/24a9ac6f-988d-437e-8879-27a43be55d36

### DIFF
--- a/vscode/src/chat/chat-view/ChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/ChatPanelProvider.ts
@@ -383,14 +383,7 @@ export class ChatPanelProvider extends MessageProvider {
         activePanelViewColumn?: vscode.ViewColumn,
         chatID?: string,
         lastQuestion?: string
-    ): Promise<vscode.WebviewPanel | undefined> {
-        // Create the webview panel only if the user is logged in.
-        // Allows users to login via the sidebar webview.
-        if (!this.authProvider.getAuthStatus()?.isLoggedIn || !this.contextProvider.config.experimentalChatPanel) {
-            await vscode.commands.executeCommand('setContext', CodyChatPanelViewType, false)
-            return
-        }
-
+    ): Promise<vscode.WebviewPanel> {
         telemetryService.log('CodyVSCodeExtension:createWebviewPanel:clicked', undefined, { hasV2Event: true })
 
         // Checks if the webview panel already exists and is visible.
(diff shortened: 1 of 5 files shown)


vscode/src/chat/chat-view/ChatPanelProvider.ts
```diff
@@ -383,14 +383,7 @@ export class ChatPanelProvider extends MessageProvider {
...
         chatID?: string,
         lastQuestion?: string
-    ): Promise<vscode.WebviewPanel | undefined> {
-        // Create the webview panel only if the user is logged in.
-        // Allows users to login via the sidebar webview.
-        if (!this.authProvider.getAuthStatus()?.isLoggedIn || !this.contextProvider.config.experimentalChatPanel) {
-            await vscode.commands.executeCommand('setContext', CodyChatPanelViewType, false)
-            return
-        }
-
```
beyang, line 393: Why was this removed as part of this change?

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -40,9 +40,14 @@ export interface IChatPanelProvider extends vscode.Disposable {
     setWebviewView(view: View): Promise<void>
     restoreSession(chatIDj: string): Promise<void>
     setConfiguration?: (config: Config) => void
+    revive: (panel: vscode.WebviewPanel, chatID: string) => Promise<void>
 }
 
 export class ChatPanelsManager implements vscode.Disposable {
+    // Sync feature flag for cody.experimental.simpleChatContext to this variable
+    // TODO remove once ChatPanelProvider and SimpleChatPanelProvider are unified
+    private useSimpleChatPanelProvider = false
```
beyang, line 49: Do we need to store a separate piece of state here, or could we not just use `this.options.contextProvider.config.experimentalSimpleChatContext`? Preference for not introducing additional state, because it may fall out of sync in the future.

beyang (approved):
Approving to unblock, but I think we should avoid adding an additional field `useSimpleChatPanelProvider` if we can.

vscode/src/chat/chat-view/ChatPanelProvider.ts
abeatrix: Because this is already handled in `onConfigurationChange`: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/ChatPanelsManager.ts?L86-106

vscode/src/chat/chat-view/ChatPanelsManager.ts
abeatrix: makes sense,thanks! 

i've removed that and I also updated to unify `revive` and `createWebviewPanel` for simplification

PR #2017, opened 2023-11-30
Support persistence:{removed,present} events in agent mode

Strum355:
Does what it says on the tin

## Test plan

Tested with jetbrains

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -223,11 +223,12 @@ export class Agent extends MessageHandler {
         this.registerNotification('textDocument/didChange', document => {
             const documentWithUri = TextDocumentWithUri.fromDocument(document)
 
-            const contentChanges: vscode.TextDocumentContentChangeEvent[] = []
-            // get document with old content before updating
+            // Get document with old content before updating. This is not a deep copy!
             const oldDocument = this.workspace.getDocument(documentWithUri.uri)
+            const oldDocumentText = oldDocument?.getText()
+            const contentChanges: vscode.TextDocumentContentChangeEvent[] = []
             if (oldDocument) {
-                const edits = diff.calcPatch(oldDocument.getText() ?? '', document.content ?? '')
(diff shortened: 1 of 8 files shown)


agent/src/agent.ts
```diff
@@ -222,12 +223,31 @@ export class Agent extends MessageHandler {
...
+            let oldTextDocument: AgentTextDocument | undefined
+            const oldDocument = this.workspace.getDocument(documentWithUri.uri)
+            if (oldDocument) {
+                oldTextDocument = this.workspace.agentTextDocument(oldDocument)
+            }
             const textDocument = this.workspace.agentTextDocument(documentWithUri)
             this.workspace.addDocument(documentWithUri)
             this.workspace.setActiveTextEditor(newTextEditor(textDocument))
+
+            const contentChanges: vscode.TextDocumentContentChangeEvent[] = []
```
olafurpg, line 235: nit: would be nice to extract this into a function.

vscode/src/completions/persistence-tracker.ts
```diff
@@ -58,15 +61,27 @@ export class PersistenceTracker implements vscode.Disposable {
             return
         }
 
+        const isInAgent = getConfiguration().isRunningInsideAgent ?? false
+
         // The range for the completion is relative to the state before the completion was inserted.
-        // We need to convert it to the state after the completion was inserted.
+        // We need to convert it to the state after the completion was inserted, unless we're running
+        // in agent mode, in which case the range will be updated later when we receive a document
```
olafurpg, line 68: Is there a way to avoid this difference in behavior? I've tried to limit cases where we use `isRunningInsideAgent` to influence behavior in this way. It typically ends up being difficult to reason about the code when we need different code paths for the agent and vscode.

olafurpg (commented):
Quick review. Now that agent tests (`agent/src/index.test.ts`) are running in CI, it would be nice to test this by subscribing to document changes sending subsequent `didChange` events to see that the diff is as expected.

I'd love to avoid the special case for running inside the agent.

PR #2018, opened 2023-11-30
make commands work with SimpleChatPanelProvider

beyang:
https://github.com/sourcegraph/cody/assets/1646931/eeb77823-6654-4687-a06a-81ef7b5c7169

## Test plan

Test out the commands that output into chat:
- [ ] explain
- [ ] test
- [ ] smell
- [ ] ask
- [ ] custom commands (commit-message, summarize-latest-release, etc)

### DIFF
--- a/vscode/src/chat/chat-view/LegacyRecipeAdapter.ts
+++ b/vscode/src/chat/chat-view/LegacyRecipeAdapter.ts
@@ -1,0 +1,101 @@
+import { ContextFile } from '@sourcegraph/cody-shared'
+import { RecipeID } from '@sourcegraph/cody-shared/src/chat/recipes/recipe'
+import { InteractionMessage } from '@sourcegraph/cody-shared/src/chat/transcript/messages'
+import { ContextMessage } from '@sourcegraph/cody-shared/src/codebase-context/messages'
+import { Editor } from '@sourcegraph/cody-shared/src/editor'
+import { IntentDetector } from '@sourcegraph/cody-shared/src/intent-detector'
+import { Message } from '@sourcegraph/cody-shared/src/sourcegraph-api'
+
+import { PlatformContext } from '../../extension.common'
+import { ContextProvider } from '../ContextProvider'
+
+import { contextMessageToContextItem } from './chat-helpers'
+import { ContextItem, MessageWithContext } from './SimpleChatModel'
+
+/**
+ * LegacyRecipeAdapter is a class that adapts the old recipes for use by the new
+ * SimpleChatPanelProvider
+ */
+export class LegacyRecipeAdapter {
+    constructor(
+        private editor: Editor,
+        private intentDetector: IntentDetector,
+        private contextProvider: ContextProvider,
+        private platform: Pick<PlatformContext, 'recipes'>
+    ) {}
+
+    public async computeRecipeMessages(
+        requestID: string,
+        recipeID: RecipeID,
+        humanChatInput: string,
+        userInputContextFiles?: ContextFile[],
+        addEnhancedContext: boolean = true
+    ): Promise<{
+        humanMessage: MessageWithContext
+        prompt: Message[]
+    } | null> {
+        const recipe = this.platform.recipes.find(recipe => recipe.id === recipeID)
+        if (!recipe) {
+            throw new Error(`command ${recipeID} is unsupported`)
+        }
+
+        let recipeInput = humanChatInput
+        if (recipeID === 'custom-prompt') {
+            const commandRunnerID = this.editor.controllers?.command?.getCommand(humanChatInput)
+            if (!commandRunnerID) {
+                const newCommandRunnerID = await this.editor.controllers?.command?.addCommand(
+                    humanChatInput,
+                    requestID,
+                    userInputContextFiles,
+                    addEnhancedContext
+                )
+                if (newCommandRunnerID === 'invalid') {
+                    throw new Error(`did not find valid command from user input "${humanChatInput}"`)
+                }
+                if (newCommandRunnerID) {
+                    recipeInput = newCommandRunnerID
+                }
+            }
+        }
+
+        const interaction = await recipe.getInteraction(recipeInput, {
+            editor: this.editor,
+            intentDetector: this.intentDetector,
+            codebaseContext: this.contextProvider.context,
+            addEnhancedContext,
+            userInputContextFiles,
+        })
+        if (!interaction) {
+            return null
+        }
+
+        // Note: we don't include any assistant message prefixes defined by recipes
+        const humanIMsg = interaction.getHumanMessage()
+        const fullContext = await interaction.getFullContext()
+        const prompt: Message[] = fullContext.concat([humanIMsg])
+        const humanMessage = interactionMessageToMessageWithContext(humanIMsg, fullContext)
+
+        return {
+            humanMessage,
+            prompt,
+        }
+    }
+}
+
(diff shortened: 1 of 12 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -589,13 +590,15 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
...
 
-    private async postViewTranscript(messageInProgress?: ChatMessage): Promise<void> {
+    private postViewTranscript(messageInProgress?: ChatMessage): void {
         const messages: ChatMessage[] = this.chatModel.getMessagesWithContext().map(m => toViewMessage(m))
         if (messageInProgress) {
             messages.push(messageInProgress)
         }
 
-        await this.webview?.postMessage({
+        // We never await on postMessage, because it can sometimes hang indefinitely:
```
dominiccooney, line 599: TIL. There's a bunch of places we do this. Guess we need to ensure we don't do *anything* with those Promises *except* void them.

Would be great if Code Insights could work with ASTs.

dominiccooney (approved):
Looks good. Biggest question I had was about how the "human" message (which the recipe generates, right?) is handled as an assistant or human message. Some ideas about naming inline.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -57,32 +60,29 @@ interface SimpleChatPanelProviderOptions {
...
 
 export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelProvider {
-    private chatModel: SimpleChatModel = new SimpleChatModel('anthropic/claude-2')
-
-    private config: Config
-
     public webviewPanel?: vscode.WebviewPanel
     public webview?: ChatViewProviderWebview
 
+    private chatModel: SimpleChatModel = new SimpleChatModel('anthropic/claude-2')
```
abeatrix, line 70:
```suggestion
    private chatModel: SimpleChatModel = new SimpleChatModel('anthropic/claude-2.0')
```

Or should we get the default model from authStatus instead?

vscode/src/chat/chat-view/LegacyRecipeAdapter.ts
```diff
@@ -0,0 +1,101 @@
...
+        prompt: Message[]
+    } | null> {
+        const recipe = this.platform.recipes.find(recipe => recipe.id === recipeID)
+        if (!recipe) {
+            throw new Error(`command ${recipeID} is unsupported`)
+        }
+
+        let recipeInput = humanChatInput
+        if (recipeID === 'custom-prompt') {
+            const commandRunnerID = this.editor.controllers?.command?.getCommand(humanChatInput)
```
abeatrix, line 44: this will always be false because getCommand takes the CommandRunnerId and returns the command, so you will need to addCommand with humanChatInput to get the CommandRunnerId, and then in the recipe we will use the CommandRunnerId to get the Command via getCommand, which doesn't make sense with the simplified chat provider 😰

abeatrix (commented):
Commands are not working for me when I started the build from this branch, getting socket hang up 🤔  @dominiccooney is it working for you?

image

dominiccooney:
@abeatrix I am seeing the socket hang up with Explain, but Smell was working for me.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -420,7 +424,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
 
     private async handleEdit(requestID: string, text: string): Promise<void> {
         this.chatModel.updateLastHumanMessage({ text })
-        void this.postViewTranscript()
+        this.postViewTranscript()
         await this.generateAssistantResponse(requestID)
```
abeatrix, line 428:
```suggestion
        if (text.trim().startsWith('/')) {
            await this.executeRecipe('custom-prompt', text.trim())
        } else {
            await this.generateAssistantResponse(requestID, userContextFiles, addEnhancedContext)
        }
```

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -646,8 +681,52 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
...
+                close: (responseText: string) => {
+                    this.addBotMessageWithGuardrails(requestID, responseText)
+                },
+            })
+        } catch (error) {
+            this.postError(`command ${recipeID} failed: ${error}`)
+        }
     }
 
     public async executeCustomCommand(title: string): Promise<void> {
```
abeatrix, line 732:
```suggestion
    public async executeCustomCommand(title: string): Promise<void> {
        await this.executeRecipe('custom-prompt', title)
```

abeatrix (commented):
I couldn't get Explain and Test to work but Smell does work like Dom mentioned when running them via command pallette / editor menu / command menu.

I've made some inline suggestions that seem to get commands and custom commands to work via chat input:
 

https://github.com/sourcegraph/cody/assets/68532117/68e2d043-f3a9-4b59-ba97-bd701ca59781

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
beyang: This would also be a use case for "\@Cody" comments where you could describe invariants you want to hold in your codebase.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -646,8 +681,52 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
...
+                requestID,
+                recipeID,
+                humanChatInput,
+                userInputContextFiles,
+                addEnhancedContext
+            )
+            if (!recipeMessages) {
+                return
+            }
+            await this.clearAndRestartSession()
```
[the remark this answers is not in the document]
beyang: Not necessary, since the same check exists in `clearAndRestartSession`

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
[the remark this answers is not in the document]
dominiccooney: Confused emoji, let's talk in real life. This was purely positive, nothing to block this PR landing on.

vscode/src/chat/chat-view/LegacyRecipeAdapter.ts
```diff
@@ -0,0 +1,101 @@
...
+            prompt,
+        }
+    }
+}
+
+function interactionMessageToMessageWithContext(
+    imsg: InteractionMessage,
+    contextMessages: ContextMessage[]
+): MessageWithContext {
+    const contextItems = contextMessages
```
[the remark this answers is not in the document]
dominiccooney: It does (or did?) because InteractionMessage : ContextMessage, and ContextMessage has contextFiles?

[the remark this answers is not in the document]
beyang: You are correct. I've refactored this to make it clearer what context is used. It is safe to use `interaction.getFullContext()` as this should be a superset of `interaction.getHumanMessage().contextFiles`.

Issue #2020, opened 2023-11-30
bug: Simple chat panel bug reports


beyang:
## P1
- [x] [When enhanced context is off, no files should be fetched implicitly](https://sourcegraph.slack.com/archives/C05AGQYD528/p1701064244578039?thread_ts=1701062497.066149&cid=C05AGQYD528)
- [x] #2093
- [x] [Enhanced context selector does not appear](https://sourcegraph.slack.com/archives/C05AGQYD528/p1701065966715799?thread_ts=1701062497.066149&cid=C05AGQYD528). [Proposed fix](https://sourcegraph.slack.com/archives/C05AGQYD528/p1701153367410829?thread_ts=1701062497.066149&cid=C05AGQYD528)
- [x] Add back in the "You are Cody" preamble
- [x] Don't display "error: aborted" when user cancels chat. Reenable new chats after request is aborted.
- [x] #2107
- [x] [Toggle on feature flag by default](https://github.com/sourcegraph/cody/pull/2105)

PR #2021, opened 2023-11-30
Roll current editor context into enhanced context

beyang:
When enhanced context is off, no files should be fetched implicitly, including current editor context. Fixes item 1 in #2020.

## Test plan

Test the following queries:
- [ ] (Enhanced context on) What does this repo do? -> Should include at least one README
- [ ] (Enhanced context off) What does this repo do? -> Should not include any files
- [ ] (Enhanced context on) What does this file do? -> Should include current file
- [ ] (Enhanced context off) What does this file do? -> Should not include any files
- [ ] (Enhanced context on) Where do we call anthropic? -> Should fetch anthropic-related files
- [ ] (Enhanced context off) Where do we call anthropic? -> Should not fetch any files
- [ ] (Enhanced context off) @-mention a file explicitly -> should include the explicitly mentioned file


dominiccooney (commented):
Looks good!

PR #2031, opened 2023-12-01
fix: truncate long panel title

abeatrix:
We will need to truncate the panel title if it's too long or the close button will be hidden in the editor to close the chat panel

image

No button to close the chat panel:

image

## Test plan

Tests updated.

Close button to show up again even if the title is too long:

image

### DIFF
--- a/vscode/src/chat/chat-view/chat-helpers.test.ts
+++ b/vscode/src/chat/chat-view/chat-helpers.test.ts
@@ -90,10 +90,10 @@ describe('getChatPanelTitle', () => {
         expect(result).toEqual('New Chat')
     })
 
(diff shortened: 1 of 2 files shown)


toolmantim (commented):
Good catch, I spotted this today too! No idea how my testing missed this before sorry.

Can we truncate it only for the editor titles though? Not for the treeview too?

abeatrix:
> Good catch, I spotted this today too! No idea how my testing missed this before sorry.
> 
> Can we truncate it only for the editor titles though? Not for the treeview too?

@toolmantim Yup, just updated it!

image

toolmantim (approved):
👏🏼

PR #2047, opened 2023-12-01
BFG: index git repository from subfolders and workspaces with no git repositories

olafurpg:
Previously, users had to open VS Code at the root of a git repository. This PR changes the behavior to let bfg index the git repo even if the workspace folder is a sub-directory. In the case where the workspace has no git repo, bfg indexes the workspace folders from the file system. This means that users can benefit from bfg in significantly more cases than before because bfg is always active regardless of what workspace you open. 

## Test plan

Manually tested by opening the extension in a sub-directory and confirmed that BFG indexed the parent git directory.


chwarwick (commented):
I think this makes a lot of sense, but this did come up during a security review with a customer with the request that the extension is not able scan for files outside of the "project".  What a project is was never defined, and I think that the git repo is a reasonable boundary.

olafurpg:
I realized that bfg already supports `workspace/didChange` to index regular directories so I updated this PR to implement the last fallback to index workspaces with no git repositories.

olafurpg:
@chwarwick I updated the PR to add a hidden boolean setting `cody.experimental.cody-engine.index-parent-git-folder` to allow disabling this behavior. When this is disabled, BFG still indexes the workspace folder by reading directly from the file system (instead of via git, which is the default)

olafurpg:
@toolmantim FYI, this thread here has more context on the resolution of this PR. We concluded that the functionality added in this PR is a low risk since we still only index the enclosing git directory. It would have been a higher risk if we indexed outside the git repository (but that's not what this PR is doing) https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1701736206534399?thread_ts=1701427583.468059&cid=C04MSD3DP5L

Issue #1992, opened 2023-11-29
When I upgrade to Cody Pro or downgrade to Cody Free, the IDE should reflect this change ASAP


chenkc805:
Steps to repro:
- Start on Cody Free
- Open VS Code with Cody installed
- Click "Upgrade to Pro" in the bottom right settings area
- Navigate to /cody/subscription
- Upgrade to Cody Pro
- Go back to VS Code
- 🐛 The user sees no indication that they've upgraded, and the "Upgrade" button is still there.
  - **Expected behavior:** The "Upgrade" button should be gone, and the user should see a chat message that tells them they're upgraded. Design for the message is here: https://www.figma.com/file/f7bzmpYrwalyYOQJcWLRTP?node-id=73:522&mode=design#622905778

chenkc805:
Adding @kalanchan @taylorsperry since I think we need support on the clients teams to fix this. CC: @thenamankumar to support from the PLG side

kalanchan:
Will get back to you once I have a game plan figured out with the clients

thenamankumar:
What is the status on this @kalanchan ?

vdavid:
@kalanchan, I think you said this was done. Could you close it if that's the case?

chenkc805:
@kalanchan looks like this was merged 3 days ago, can we close this out?

kalanchan:
hahahah thanks for TPMing me guys! I appreciate it :) 

yes closed by #2049 - if you upgrade/downgrade in web, it will reflect the change in VSC. If you notice a bug feel free to re-open and tag me

DataDog3331:
This issue actually persists. i was using cody free version and hot rate limits. i followed the upgrade button to web to purchase a plan, completed it but im still stuck in vs code where it says "UNABLE TO SEND MESSAGE You've used all of your chat messages and commands for today. Learn More Usage will reset in 28 days (08/28/2024 at 9:36 AM)"

My question is, how to let cody vs extension know i have purchased a plan?

PR #2049, opened 2023-12-01
Refresh Cody Pro status on editor focus change

SuperAuguste:
Addresses #1992 for VSCode and enables other editors to address it.

## Test plan

Tested locally.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -386,6 +386,19 @@ export class Agent extends MessageHandler {
             throw id
         })
 
+        this.registerRequest('graphql/currentUserIsPro', async () => {
+            const client = await this.client
+            if (!client) {
+                throw new Error('Cody client not initialized')
+            }
+            const res = await client.graphqlClient.getCurrentUserHasCodyPro()
+            if (res instanceof Error) {
+                throw res
+            }
+
+            return res.codyProEnabled
+        })
+
         this.registerRequest('telemetry/recordEvent', async event => {
             this.agentTelemetryRecorderProvider.getRecorder().recordEvent(
                 // 👷 HACK: We have no control over what gets sent over JSON RPC,
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -391,7 +391,7 @@ export class Agent extends MessageHandler {
             if (!client) {
                 throw new Error('Cody client not initialized')
             }
-            const res = await client.graphqlClient.getCurrentUserHasCodyPro()
+            const res = await client.graphqlClient.getCurrentUserIdAndVerifiedEmailAndCodyPro()
             if (res instanceof Error) {
                 throw res
             }
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -14,7 +14,6 @@ import {
     CURRENT_SITE_HAS_CODY_ENABLED_QUERY,
     CURRENT_SITE_IDENTIFICATION,
     CURRENT_SITE_VERSION_QUERY,
-    CURRENT_USER_HAS_CODY_PRO,
     CURRENT_USER_ID_AND_VERIFIED_EMAIL_AND_CODY_PRO_QUERY,
     CURRENT_USER_ID_AND_VERIFIED_EMAIL_QUERY,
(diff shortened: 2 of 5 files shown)


vscode/src/main.ts
```diff
@@ -538,6 +538,19 @@ const register = async (
     authProvider.addChangeListener(() => updateAuthStatusBarIndicator())
     updateAuthStatusBarIndicator()
 
+    vscode.window.onDidChangeWindowState(async ws => {
```
kalanchan, line 541: how often would this get fired? Is this similar to something like `onBlur` in the DOM?

agent/src/agent.ts
```diff
@@ -386,6 +386,19 @@ export class Agent extends MessageHandler {
             throw id
         })
 
+        this.registerRequest('graphql/currentUserIsPro', async () => {
```
kalanchan, line 389: do we need another graphql query? Unless i'm mistaken there should already be a graphql query to check account status, or maybe that was just on the VSC side

We were using this query on VSC
```{
  currentUser {
    codyProEnabled
  }
```

kalanchan (commented):
I think the approach here is good, I just have a couple questions around when the check would get fired. 

We need be able to handle scenarios where
1. user starts in the web, upgrades, and goes back to editor
2. user starts in the editor, upgrades, and goes back to the editor

lib/shared/src/sourcegraph-api/graphql/queries.ts
```diff
@@ -251,3 +251,13 @@ export const EVALUATE_FEATURE_FLAG_QUERY = `
         evaluateFeatureFlag(flagName: $flagName)
     }
 `
+
+export const CURRENT_USER_HAS_CODY_PRO = `
```
abeatrix, line 255: This looks similar to the `CURRENT_USER_ID_AND_VERIFIED_EMAIL_AND_CODY_PRO_QUERY`query, we can just use that instead?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -332,6 +337,19 @@ export class SourcegraphGraphQLAPIClient {
...
 
+    public async getCurrentUserHasCodyPro(): Promise<
+        { id: string; codyProEnabled: boolean; codyProEnabledAt: Date } | Error
+    > {
+        return this.fetchSourcegraphAPI<APIResponse<CurrentUserHasCodyProResponse>>(CURRENT_USER_HAS_CODY_PRO, {}).then(
+            response =>
+                extractDataOrError(response, data =>
+                    data.currentUser
+                        ? { ...data.currentUser }
+                        : new Error('current user not found with verified email and cody pro')
```
abeatrix, line 348: The `CURRENT_USER_HAS_CODY_PRO` query doesn't tell us if the user has verified email or not, maybe we can just use the `getCurrentUserIdAndVerifiedEmailAndCodyPro` function below for the same purpose?

abeatrix (commented):
left comments and questions in line. I don't think this would work if the users do the sign up process in another device, but i guess that's an edge case for another time? 😜

kalanchan:
> left comments and questions in line. I don't think this would work if the users do the sign up process in another device, but i guess that's an edge case for another time? 😜

I think the idea is that it doesn't matter where the user initiates the upgrade/downgrade, whenever they go back to the editor and click around it should fire a check. Is that right @SuperAuguste?


but yes, we need to handle both scenarios: 

1. user starts in the web, upgrades/downgrades, and goes back to editor
2. user starts in the editor, upgrades/downgrades, and goes back to the editor

vscode/src/main.ts
SuperAuguste: Whenever a user comes back to a VSCode window after clicking off it. :)

lib/shared/src/sourcegraph-api/graphql/queries.ts
SuperAuguste: Sure!

abeatrix (approved):
yay

toolmantim:
Nice one @SuperAuguste 🎉

vdavid (approved):
Nice, likey!

SuperAuguste:
Glad this helps a little!

PR #2059, opened 2023-12-02
fix: migrate to chat history manager

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1968

This PR focused on simplify chat history handling by moving the old chat history handling in MessageProvider to the new Chat History Manager.

Removed redundant chat history code and use shared chatHistory module instead. This simplifies the MessageProvider by removing duplicated chat history logic.

This also fixes issue described in the attached issue as previously, we would save chat history as group instead of individual chat, so a new group can overwritten the old one, unable to keep history across panels in sync.

## Test plan

Follow instructions in attached issue

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -137,8 +137,8 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
 
     public async clearAndRestartSession(): Promise<void> {
         await this.saveTranscriptToChatHistory()
-        this.createNewChatID()
         this.cancelCompletion()
+        this.createNewChatID()
         this.isMessageInProgress = false
         this.transcript.reset()
         this.handleSuggestions([])
@@ -166,10 +166,8 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
         if (!history || chatID === this.sessionID) {
             return
         }
-        if (!this.transcript.isEmpty) {
-            await this.saveTranscriptToChatHistory()
-            this.cancelCompletion()
-        }
+        await this.saveTranscriptToChatHistory()
+        this.cancelCompletion()
         this.createNewChatID(chatID)
         this.transcript = Transcript.fromJSON(history)
         this.chatModel = this.transcript.chatModel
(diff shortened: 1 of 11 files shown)


vscode/src/chat/MessageProvider.ts
```diff
@@ -174,10 +162,16 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
...
     public async restoreSession(chatID: string): Promise<void> {
-        await this.saveTranscriptToChatHistory()
-        this.cancelCompletion()
+        const history = chatHistory.getChat(chatID)
+        if (!history || chatID === this.sessionID) {
+            return
+        }
+        if (!this.transcript.isEmpty) {
+            await this.saveTranscriptToChatHistory()
+            this.cancelCompletion()
```
dominiccooney, line 171: Worth a brief comment why. Is this about the chat transcript being saved after a human message, and then restored before the "completion" (assistant message) was done?

vscode/src/chat/chat-view/ChatHistoryManager.ts
```diff
@@ -25,4 +25,36 @@ export class ChatHistoryManager {
...
+    public async deleteChat(chatID: string): Promise<void> {
+        await localStorage.deleteChatHistory(chatID)
+    }
+
+    // Remove chat history with input history
+    public async clear(): Promise<void> {
+        await localStorage.removeChatHistory()
+    }
+
+    public async saveInput(input: string): Promise<UserLocalHistory> {
```
dominiccooney, line 38: Worth a brief comment that this saves human input for the "up" history list, and same for `getInput`. Because there's a few methods running around with "save" this or that.

dominiccooney (commented):
Looks good but e2e tests appear to have broken.

vscode/src/chat/chat-view/ChatHistoryManager.ts
abeatrix: updated name to `save/getHumanInputHistory` and added comment

kalanchan (approved):

PR #2060, opened 2023-12-02
Add V2 telemetry to remaining Cody events

nathan-downs:
## Test plan

Instrument V2 event
run `sg start`
verify in `event_logs` and/or debug console
ensure all integration tests pass


vscode/src/non-stop/FixupController.ts
```diff
@@ -59,31 +59,61 @@ export class FixupController
             vscode.commands.registerCommand('cody.fixup.accept-all', () => this.acceptFixups()),
             vscode.commands.registerCommand('cody.fixup.diff', treeItem => this.showDiff(treeItem)),
             vscode.commands.registerCommand('cody.fixup.codelens.cancel', id => {
-                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', { op: 'cancel' })
+                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', { op: 'cancel', hasV2Event: true })
+                telemetryRecorder.recordEvent('cody.fixup.codeLens.cancel', 'clicked', {
+                    privateMetadata: { op: 'cancel' },
```
bobheadxi, line 64: I wonder if we can remove the metadata here entirely? the operation `cancel` is already part of the event names (ditto for all the `privateMetadata: { op: '...' }` events here

vscode/src/non-stop/FixupController.ts
```diff
@@ -59,31 +59,61 @@ export class FixupController
...
+                })
                 return this.accept(id)
             }),
             vscode.commands.registerCommand('cody.fixup.codelens.error', id => {
-                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', { op: 'show_error' })
+                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', {
+                    op: 'show_error',
+                    hasV2Event: true,
+                })
+                telemetryRecorder.recordEvent('cody.fixup.codeLens.error', 'error', {
```
bobheadxi, line 104: I wonder if a more appropriate translation for this particular event might be:

```suggestion
                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', {
                    op: 'show_error',
                    hasV2Event: true,
                })
                telemetryRecorder.recordEvent('cody.fixup.codeLens.show_error', 'clicked', {
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -683,12 +716,18 @@ export class FixupController
...
         if (!editOk) {
-            telemetryService.log('CodyVSCodeExtension:fixup:revert:failed')
+            telemetryService.log('CodyVSCodeExtension:fixup:revert:failed', { hasV2Event: true })
+            telemetryRecorder.recordEvent('cody.fixup.revert', 'failed')
             return
         }
 
         const tokenCount = countCode(replacementText)
         telemetryService.log('CodyVSCodeExtension:fixup:reverted', tokenCount)
+        telemetryRecorder.recordEvent('cody.fixup.reverted', 'clicked', {
```
bobheadxi, line 726:
```suggestion
        telemetryService.log('CodyVSCodeExtension:fixup:reverted', tokenCount, { hasV2Event: true })
        telemetryRecorder.recordEvent('cody.fixup.reverted', 'clicked', {
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -797,6 +836,13 @@ export class FixupController
             telemetryService.log('CodyVSCodeExtension:fixupResponse:hasCode', {
                 ...countCode(replacementText),
                 source: task.source,
+                hasV2Event: true,
+            })
+            telemetryRecorder.recordEvent('cody.fixup.response', 'hasCode', {
+                privateMetadata: {
+                    tokenCount: countCode(replacementText),
```
bobheadxi, line 843: same here, can tokenCount be non-private if numeric?

vscode/src/services/AuthProvider.ts
```diff
@@ -123,6 +128,12 @@ export class AuthProvider {
         const authState = await this.auth(instanceUrl, accessToken)
         telemetryService.log('CodyVSCodeExtension:auth:fromToken', {
             success: Boolean(authState?.isLoggedIn),
+            hasV2Event: true,
+        })
+        telemetryRecorder.recordEvent('cody.auth.signin.token', 'clicked', {
+            privateMetadata: {
+                success: Boolean(authState?.isLoggedIn),
+            },
```
bobheadxi, line 136: booleans can be translated to `0` or `1` and retained in `metadata` instead 👍

vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -37,7 +38,10 @@ export function setLastStoredCode(
     const op = eventName.includes('copy') ? 'copy' : eventName.startsWith('insert') ? 'insert' : 'save'
     const args = { op, charCount, lineCount, source, requestID }
 
-    telemetryService.log(`CodyVSCodeExtension:${eventName}:clicked`, args)
+    telemetryService.log(`CodyVSCodeExtension:${eventName}:clicked`, { args, hasV2Event: true })
+    telemetryRecorder.recordEvent(`cody.${eventName}`, 'clicked', {
+        privateMetadata: { args },
+    })
```
bobheadxi, line 44: `args` has a mix of numeric-looking and arbitrary metadata, in #1913 I added a `splitSafeMetadata` helper you can use here

vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -134,6 +138,18 @@ export async function onTextDocumentChange(newCode: string): Promise<void> {
...
+        telemetryRecorder.recordEvent(`cody.${eventType}:Paste`, 'clicked', {
+            privateMetadata: {
+                args: {
+                    op,
+                    lineCount,
+                    charCount,
+                    source,
+                    requestID,
+                },
+            },
```
bobheadxi, line 152: same here, let's split this into metadata and privateMetadata more granularly

vscode/src/local-context/local-keyword-context-fetcher.ts
```diff
@@ -13,6 +13,7 @@ import { ContextResult, KeywordContextFetcher } from '@sourcegraph/cody-shared/s
 
 import { logDebug } from '../log'
 import { telemetryService } from '../services/telemetry'
+import { telemetryRecorder } from '../services/telemetry-v2'
```
ttloves2, line 16: import { telemetryRecorder } from '../services/telemetry-v2'

vscode/src/commands/CommandRunner.ts
```diff
@@ -51,6 +52,14 @@ export class CommandRunner implements vscode.Disposable {
             useShellCommand: !!command.context?.command,
             requestID: command.requestID,
         })
+        telemetryRecorder.recordEvent(`cody.command.${this.kind}`, 'executed', {
+            privateMetadata: {
+                mode: command.mode,
+                useCodebaseContex: !!command.context?.codebase,
+                useShellCommand: !!command.context?.command,
```
bobheadxi, line 59: Booleans should just live in `metadata` so that they get exported, e.g.

```
metadata {
  useCodebaseContex: command.context?.codebase ? 1 : 0,
  useShellCommand: command.context?.command ? 1 : 0,
} 
```

vscode/src/commands/CommandRunner.ts
```diff
@@ -51,6 +52,14 @@ export class CommandRunner implements vscode.Disposable {
...
+                requestID: command.requestID,
```
bobheadxi, line 60: There is a first-class field for this ID: https://sourcegraph.com/github.com/sourcegraph/telemetry/-/blob/src/index.ts?L203:1-210:26

The Cody telemetry exporter will put this in privateMetadata if the instance is < 5.2.4 - if you want, you can put this ID in both `interactionID` and privateMetadata

vscode/src/non-stop/FixupController.ts
```diff
@@ -59,31 +59,49 @@ export class FixupController
...
+                })
+                telemetryRecorder.recordEvent('cody.fixup.codeLens.retry', 'clicked', {
+                    privateMetadata: { op: 'regenerate' },
+                })
                 return this.retry(id)
             }),
             vscode.commands.registerCommand('cody.fixup.codelens.undo', id => {
-                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', { op: 'undo' })
+                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', { op: 'undo', hasV2Event: true })
+                telemetryRecorder.recordEvent('cody.fixup.codeLens.undo', 'clicked', {})
```
bobheadxi, line 83: If there is nothing in parameters you can omit it entirely:

```suggestion
                telemetryRecorder.recordEvent('cody.fixup.codeLens.undo', 'clicked')
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -59,31 +59,49 @@ export class FixupController
...
             }),
             vscode.commands.registerCommand('cody.fixup.codelens.retry', async id => {
-                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', { op: 'regenerate' })
+                telemetryService.log('CodyVSCodeExtension:fixup:codeLens:clicked', {
+                    op: 'regenerate',
+                    hasV2Event: true,
+                })
+                telemetryRecorder.recordEvent('cody.fixup.codeLens.retry', 'clicked', {
+                    privateMetadata: { op: 'regenerate' },
+                })
```
bobheadxi, line 78: Do we need `op: regenerate` here? The other events below don't keep the `op` (it seems redundant since we already have `retry`?

Also, general pattern here - right now, we have:

```
feature: cody.fixup.codeLens.retry
action: clicked
```

Would it be more descriptive to use:

```
feature: cody.fixup.codeLens
action: retry
```

ditto for the events below

vscode/src/non-stop/FixupController.ts
```diff
@@ -681,12 +702,18 @@ export class FixupController
...
         }
 
         const tokenCount = countCode(replacementText)
-        telemetryService.log('CodyVSCodeExtension:fixup:reverted', tokenCount)
+        telemetryService.log('CodyVSCodeExtension:fixup:reverted', tokenCount, { hasV2Event: true })
+        const { metadata, privateMetadata } = splitSafeMetadata(tokenCount)
+        telemetryRecorder.recordEvent('cody.fixup.reverted', 'clicked', {
+            metadata,
+            privateMetadata,
+        })
```
bobheadxi, line 716: tokenCount already only has numeric values:

```suggestion
        telemetryService.log('CodyVSCodeExtension:fixup:reverted', tokenCount, { hasV2Event: true })
        telemetryRecorder.recordEvent('cody.fixup.reverted', 'clicked', {
            metadata: tokenCount,
        })
```

vscode/src/non-stop/FixupController.ts
bobheadxi: bump on this - tokenCount is numeric, let's put it in `metadata` so that it can be exported

vscode/src/services/AuthProvider.ts
```diff
@@ -121,12 +126,24 @@ export class AuthProvider {
...
+        telemetryRecorder.recordEvent('cody.auth.signin.token', 'clicked', {
+            metadata: {
+                success: authState?.isLoggedIn ? 1 : 0,
+            },
+        })
+        telemetryRecorder.recordEvent('cody.auth.signin.token', 'clicked', {
+            metadata: {
+                success: authState?.isLoggedIn ? 1 : 0,
+            },
         })
```
bobheadxi, line 140: should there only be one event here?

Also, is the appropriate `action` still "clicked" in this event?

vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -134,6 +138,18 @@ export async function onTextDocumentChange(newCode: string): Promise<void> {
             charCount,
             source,
             requestID,
+            hasV2Event: true,
+        })
+        telemetryRecorder.recordEvent(`cody.${eventType}:Paste`, 'clicked', {
```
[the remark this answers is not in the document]
bobheadxi: Reopening this conversation, is `clicked` really the best "action" here? `paste` seems more descriptive

bobheadxi (changes requested):
Nice work, overall comments:

1. Let's be more explicit about privateMetadata vs metadata, and think carefully about how we can represent data in metadata, and only use privateMetadata as a last resort
2. I think `action` can be used better. Right now most of these events just seem to have `"clicked"` as an action (even when it seems to have nothing to do with a click), but using action better allows better grouping of events by `feature`, and possibly query performance gains as well

vscode/src/services/AuthProvider.ts
akalia25: Great catch on the duplicate event! I think "auth" could be its own action, but would want to make sure we migrate that everywhere; and if we should separate logging in vs logging out as different actions. I can work with Kelsey on scoping this further on a different PR?

bobheadxi: Oh I'm specifically talking about there being 2x `cody.auth.signin.token` `clicked` events here 😁

akalia25: yup! Removed one of them. :)

vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -37,7 +38,12 @@ export function setLastStoredCode(
     const op = eventName.includes('copy') ? 'copy' : eventName.startsWith('insert') ? 'insert' : 'save'
     const args = { op, charCount, lineCount, source, requestID }
 
-    telemetryService.log(`CodyVSCodeExtension:${eventName}:clicked`, args)
+    telemetryService.log(`CodyVSCodeExtension:${eventName}:clicked`, { args, hasV2Event: true })
+    const { metadata, privateMetadata } = splitSafeMetadata(args)
```
bobheadxi, line 42: nit - instead of `splitSafeMetadata` the args are pretty well-defined above this, let's construct `metadata` and `privateMetadata` directly and explicitly. `splitSafeMetadata` is meant as a last resort to quickly migrate callsites that have unknown args, but here it's known

vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -134,6 +140,18 @@ export async function onTextDocumentChange(newCode: string): Promise<void> {
...
+            op,
+            lineCount,
+            charCount,
+            source,
+            requestID,
+        })
+        telemetryRecorder.recordEvent(`cody.${eventType}`, 'paste', {
+            metadata,
+            privateMetadata,
         })
```
bobheadxi, line 155: same here, the metadata are pretty well-defined, let's construct `metadata` and `privateMetadata` directly and explicitly

akalia25: sounds good! I added `interactionId = requestID` here as well, let me know if thats an issue :)

vscode/src/services/AuthProvider.ts
```diff
@@ -345,6 +357,25 @@ export class AuthProvider {
...
+            },
+        })
+        telemetryRecorder.recordEvent('cody.auth.fromCallback.web', 'succeeded', {
+            metadata: {
+                success: authState?.isLoggedIn ? 1 : 0,
+            },
+            privateMetadata: {
+                type: 'callback',
+                from: 'web',
+            },
```
bobheadxi, line 378: both `callback` and `web` are already in the event `feature` - do we need to duplicate it into metadata? (duplicating event names into metadata seems really common in the existing events, curious whether we really need to do that).

If not, it might be worthwhile to do a brief audit and remove this kind of stuff. IMO most of these the event feature/actions alone should be sufficient?

bobheadxi (approved):
overall LGTM 😁 one last lingering nit, thank you!

vscode/src/services/utils/codeblock-action-tracker.ts
bobheadxi: Love it 😁

vscode/src/services/AuthProvider.ts
akalia25: Totally agree with you! removed the `privateMetadata` and also realized we were recording this twice 🙈 so thanks for calling!

PR #2068, opened 2023-12-03
Add a test for non-dotCom chat rate limit

DanTup:
Adds an extra test for the rate limit for a non-dotcom instance (where the `x-is-cody-pro-user` header will not be set and upgrade options should not be shown).

## Test plan

- New automated test, run with `pnpm test:e2e`


abeatrix (approved):
Thank you!

Issue #2022, opened 2023-11-30
bug: "failed to download bfg" error on Windows


toolmantim:
### Version

0.16.0

### Describe the bug

On Windows, Cody shows this error message:

> Failed to download bfg from URL https://github.com/sourcegraph/bfg/releases/download/v5.2.6637/bfg-windows-arm64.zip: AxiosError: Request failed with status code 404

<img width="1256" alt="Screenshot 2023-12-01 at 10 07 16 am" src="https://github.com/sourcegraph/cody/assets/153/05c6ac6d-5af0-46ab-b147-506405b5c686">

### Expected behavior

No error message

toolmantim:
cc @philipp-spiess @valerybugakov

olafurpg:
Thank you for reporting! This error is happening since we don't build bfg for Windows ARM (yet?). Any idea @SuperAuguste how to support this?

I tried building it like this

```
❯ rustup target install aarch64-pc-windows-msvc
❯ cargo zigbuild --target=aarch64-pc-windows-msvc --profile=release-without-debug
```
but it failed with errors about GNU compiler

```
...
warning: tree-sitter@0.20.10: GNU compiler is not supported for this target
```

I tried using `cargo build` instead but that seems to require that we build the binary on a Windows computer

```
  running: "lib.exe" "-....
  error occurred: Failed to find tool. Is `lib.exe` installed?
```

I'm installing clang on my windows computer right now to see if I can `cargo build` the aarch64 target...

olafurpg:
This line here indicates it failed to start the binary. 
```
█ CodyEngine: failed to initialize Error: spawn EBUSY
```

I only built the binary but wasn't able to test it on arm64-windows. For now, I think it's best to stick with saying that Windows arm64 is not supported. I think we can support it eventually but it requires engineering time to setup the build so that we test it in CI and automate releases for it.

We should still improve error handling so that it either silently doesn't work or we notify the user that Windows arm64 is not supported. However, before doing that, we may want to wait a bit to see how many users hit on this error. The JetBrains plugin and Neovim don't work on windows-arm64 either because we don't build an agent binary for it. We don't have any special handling to inform users about this case and it hasn't been reported as an issue IIRC.

PR #2073, opened 2023-12-04
Replace windows-arm64 with windows-x84 binary

olafurpg:
Previously, we tried to download a non-existent windows-arm64 binary,
  which resulted in a 404 error for end users. This PR fixes the problem
  by using the windows-x84 binary instead. This works because Windows
  Arm PCs support x86 emulation https://learn.microsoft.com/en-us/windows/arm/apps-on-arm-x86-emulation

Fixes https://github.com/sourcegraph/cody/issues/2022

## Test plan

Test it on Windows using Parallels Desktop on Apple Silicon.


toolmantim (approved):
Nice!

PR #2077, opened 2023-12-04
Set ChatModels behind codypro feature flag

abeatrix:
Set Chat Models selection behind `cody-pro` feature flag so that it will be limited for users with the feature flag enabled only.

## Test plan

All the tests for the cody pro feature should still pass

Sign in to DotCom with a non-SG employee account and check if you still see the CodyPro badge in the dropdown model:

### Before

image

### After

image

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -6,6 +6,7 @@ import { Polly } from '@pollyjs/core'
 import envPaths from 'env-paths'
 import * as vscode from 'vscode'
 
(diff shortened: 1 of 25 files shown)


kalanchan (approved):
looks good! signed in with a test account and regular sg account and saw the changes

CleanShot 2023-12-04 at 10 25 16@2x

CleanShot 2023-12-04 at 10 26 31@2x

Issue #1984, opened 2023-11-29
bug: "from ${n} files" missing line counts


toolmantim:
### Version

main (2ecded4fd9ae240edbc3233f545b32205e30abb3)

### Describe the bug

In the chat UI we display "▶ from ${n} files":

<img width="165" alt="Screenshot 2023-11-29 at 9 59 25 pm" src="https://github.com/sourcegraph/cody/assets/153/5a048e41-d914-4f8c-84c4-5101578b2c9b">

### Expected behavior

Ideally we should sum the line count from the files:

"▶ Context: 1320 lines from 120 files"
"▶ ✨ Context: 1320 lines from 120 files"

As an interim if line count isn't available:

"▶ Context: 120 files"
"▶ ✨ Context: 120 files"

✨ is shown if any of the context included things fetched via Enhanced Context.

PR #2086, opened 2023-12-04
update enhance context display copy

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1984

Suggested by @toolmantim 

> As an interim, we can at least remove the word "from" in "▶ from ${n} files" — as "from" is only meant to be used in the form "{x} lines from {n{ files"

### Before

image

### After

image


toolmantim (commented):
Thanks @abeatrix! I just updated #1984 to have the new desired behaviour.

abeatrix:
@toolmantim updated based on the issue:

image

image

toolmantim (approved):

Issue #1969, opened 2023-11-29
bug: prevent having multiple new chats open at once


toolmantim:
### Version

0.16.0

### Describe the bug

Previously we decided to open a new chat tab every time you click "New Chat". But due to the lag when creating a new chat tab, and the fact that it doesn't appear in the tree view, we should only create a single new chat tab:

https://github.com/sourcegraph/cody/assets/153/169b424e-e63b-433d-8f92-07d6c11becd3

### Expected behavior

If there exists a new/blank chat, the "New Chat" command will refocus the existing new/blank chat.

PR #2087, opened 2023-12-04
change: reuse empty "New Chat" panel

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1969

Reuse the empty New Chat panel when there is one available to prevent having multiple new chats open at once 

## Test plan

Click on the New Chat button when there is a "New Chat" available should reveal the existing one instead of creating a new one:

https://github.com/sourcegraph/cody/assets/68532117/95f7ee4a-a848-479d-923b-87f88dd54963


toolmantim (approved):
Works a treat!

PR #2098, opened 2023-12-05
Incorporate Cody preamble into simple chat provider

beyang:
## Test plan

- [ ] Ask "who are you?" in the chat
- [ ] Run a recipe like `Explain` and then ask "who are you?" in a follow-up question

### DIFF
--- a/lib/shared/src/chat/preamble.ts
+++ b/lib/shared/src/chat/preamble.ts
@@ -1,5 +1,18 @@
 import { Message } from '../sourcegraph-api'
 
+export function getSimplePreamble(): Message[] {
+    return [
+        {
+            speaker: 'human',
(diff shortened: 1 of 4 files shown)


dominiccooney (approved):
Looks good.

The reverse of reverse messages... We have some anecdotal idea that recent files get more attention. How does this compare to the status quo? I'm happy to take my answer offline, just curious. Local embeddings fidelity is far from perfect 1:1 with App either, yet.

beyang:
> How does this compare to the status quo?

The status quo attempts to construct the context with most important context *last* in terms of message order.

> Local embeddings fidelity is far from perfect 1:1 with App either, yet.

Oh? What's the delta?

Issue #1966, opened 2023-11-29
App users will be automatically signed in through dotcom (instead of app) for GA


taylorsperry:
When app users download the app, the first thing they have to do to get setup is create/login to their Sourcegraph.com account. However, when connecting the app to VS Code, they login to their VS Code extension "through the app." 

As part of the work to sunset the app, we have the ability to automatically log the user into their dotcom account instead of their app account in VS Code. This feels like the most seamless experience and nothing changes about the user's dotcom account, but we should include an indication that the login method has been updated.

dominiccooney:
Technical idea here is look at the origin we're signed in to. If it looks like App (localhost:3900 IIRC) then flip to sourcegraph.com instead, automatically.

Question for @toolmantim is whether we should do this silently, or give some notice that this is happening. My $0.02 is that we're *probably* talking to App, and App is talking to dotcom, so there's no harm in silently making this switch.

toolmantim:
Here's the UX outline for how this should work, taking into account embeddings importing too:

1. Cody for VS Code updates behind the scenes for the user
1. We silently migrate the auth to dotcom (ensuring Cody App is removed from account switcher history etc too)
1. We send a progress notification (https://code.visualstudio.com/api/ux-guidelines/notifications#progress-notification) with the words "Importing Cody App embeddings…". This closes automatically on completion.
1. After embeddings import completes (or if Cody App has no embeddings — doesn't matter if this is near-immediate speed wise), we send an information notification:

> Cody App is deprecated and your embeddings have been successfully imported into VS Code. You can now safely move Cody App to the bin.
> [Move to Bin] [Learn More]"

The "Move to Bin" button would be primary color, and would be a nice-to-have. "Learn More" would open a docs URL.

How's that sound @dominiccooney @taylorsperry?

taylorsperry:
That sounds fantastic. Thank you, @toolmantim and @dominiccooney !

dominiccooney:
I have #2099 for this.

It does not do the toasting to delete App yet.

Adds local embeddings support for classic sidebar chats, although the context status indicator says embeddings are not found. I assume it is not worth wiring this up at this point.

There's no App index "import" step or UI, we are just reading App's files directly. I can work on this but error reporting, restarting failed index jobs, etc. is probably more important.

dominiccooney:
@taylorsperry , @kalanchan we will automatically log people into dotcom where possible. Worst case they will see the login panel and have to hit login again.

[...]

PR #2099, opened 2023-12-05
vscode: Remove App login

dominiccooney:
This one is a doozy:

- Removes App as a sign-in option.
- If you're signed into App but you have a dotcom token, we'll try to use that.
- If you're signed into App, but don't have a dotcom token, you'll have to sign in again.
- Even if you're using classic sidebar chat, if you had an embeddings index generated by App, we'll pick it up and use it. However the context status indicator will indicate you have no embeddings. (We could plumb this in, but it is going away very soon.)
- No more toasts or promotions to set up App.

Leaves the LocalAppDetector so we can delete App/toast you to delete App soon.

Part of #1966.

## Test plan

**Setup**

The big secret to testing this is recreating the "signed in to App" state, now that all the sign in options are gone. You do this by abusing the Enterprise sign-in options:

1. Install Cody App and log into it if you haven't got it already.
2. Sign into dotcom, which caches a token used for the first test below.
3. In VScode, Cmd-Shift-P, Cody: Switch Account... and enter `http://localhost:3080` which is App's URL.
4. cat ~/Library/Application\ Support/com.sourcegraph.cody/app.json and smurf the token.
5. At this point, you should be signed in to App. You can verify this by chatting, quitting app, chatting again and seeing you get a connection error message.

**Next sign in**

1. Simply reload the window.
2. You should be signed in to dotcom with the token you had been using at the start of the previous step. You can verify this by chatting with App closed.
3. Check that Cody App doesn't appear in any of the account switcher or sign in menus.

**Signed into App, no dotcom token**

Assuming you're still signed in to dotcom:

1. Cmd Shift P, Cody: Sign out.
4. Sign into App using the leetcode techniques above.
5. Reload the window.

You should be dumped at the login pane.

**Signed into App, cached invalid dotcom token**

1. Create a new dotcom auth token at https://sourcegraph.com/user/settings/tokens/new/callback
2. Sign in to http://sourcegraph.com using that token
3. Cmd Shift P, Cody: Switch Account... and follow the "setup" instructions above to sign in to app.
4. You are now signed in to App, with a *valid* cached dotcom token. Switch back to the Sourcegraph token management UI and revoke the token.
5. Reload the window. This encounters a *long* (60s?) pause, but eventually the login fails and you are dumped back to the login screen and can log in to get a fresh token.


beyang:
Rather than try to solve for all auto-migration edge cases, could we have some sort of catch-all or fallback warning message that instructs the user to uninstall App and install Cody from scratch? This seems easier on both us and the user than trying to debug any errors that happen during migration.

Nit: I notice you prefix your PR titles with "VScode:", which seems a bit off:
* If you want camel casing of "VS Code", then "VSCode" would be appropriate
* If you're simply looking for a shorthand to indicate the general category of this change, then "vscode" seems sufficient and more aesthetically pleasing

vscode/src/chat/chat-view/SidebarChatProvider.ts
```diff
@@ -69,6 +56,12 @@ export class SidebarChatProvider extends MessageProvider implements vscode.Webvi
                 await this.init()
                 break
             case 'submit':
+                // Hint local embeddings, if any, should start so subsequent
+                // messages can use local embeddings. We delay to this point so
+                // local embeddings service start-up updating the context
+                // provider does not cause autocompletes to disappear.
+                void this.contextProvider.localEmbeddings?.start()
```
beyang, line 63: Why would local embeddings start-up interfere with autocomplete? Isn't it the case that autocomplete doesn't use embeddings?

Also: sidebar chat is going away, right?

beyang (approved):
I have not gone through the test plan (see my previous message on the recommended path), but the code changes LGTM.

PR #2103, opened 2023-12-05
Make abort work in the simplified chat panel

beyang:
https://github.com/sourcegraph/cody/assets/1646931/757f0780-946f-4eac-93ff-9761caeece26

* Abort now works in the simplified chat panel
* Fixes a possible race condition where an error in local embeddings search seems sometimes to result in empty context, even if remote embedding search yields results

## Test plan

- [ ] Test aborting an in-progress message
  - [ ] If the message is in-progress, the partial message should be adopted
  - [ ] If the message was still empty, the last human question should be removed
- [ ] Test asking a follow-up question after aborting in each of the above cases

Note: the context fetching step (prior to response generation) cannot yet be aborted.

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatModel.ts
+++ b/vscode/src/chat/chat-view/SimpleChatModel.ts
@@ -69,6 +69,16 @@ export class SimpleChatModel {
         })
     }
 
+    public removeLastMessageIfHuman(): void {
+        const lastMessage = this.messagesWithContext.at(-1)
+        if (!lastMessage) {
+            return
+        }
+        if (lastMessage.message.speaker === 'human') {
+            this.messagesWithContext.pop()
+        }
+    }
+
     public getLastHumanMessages(): MessageWithContext | undefined {
         return this.messagesWithContext.findLast(message => message.message.speaker === 'human')
     }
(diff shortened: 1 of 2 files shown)


toolmantim (approved):
I gave this a test. It works AOK when the response is streaming, but there's two cases it doesn't handle well:

* Clicking cancel while context is being fetched (you already mentioned out of scope of this PR)
* Clicking cancel before the response starts streaming deletes the last human message (this seems by design, but I think it would be better to just leave the last human message there in the transcript)

https://github.com/sourcegraph/cody/assets/153/fae1e739-03e1-4d2f-8b0a-815904de237d

I can see quickly pressing cancel to be a common case, if people spot they mis-wrote something or mis-@'d a file in their message and want to quickly cancel & edit or try again.

beyang:
> Clicking cancel while context is being fetched (you already mentioned out of scope of this PR)

I believe this preserves the existing behavior, but context fetching is taking longer now for some reason, which I'm looking into.

> Clicking cancel before the response starts streaming deletes the last human message (this seems by design, but I think it would be better to just leave the last human message there in the transcript)

Updated PR to match this behavior.

dominiccooney (approved):
Looks good! I assume `postViewTranscript(...)` doesn't accumulate the ... otherwise we might be posting two assistant messages with empty text? One from `generateAssistantResponse` and one from `handleHumanMessageSubmitted`.

PR #2113, opened 2023-12-05
Link to new Cody feedback URL

toolmantim:
## Test plan

- Clicked feedback

### DIFF
--- a/vscode/src/chat/protocol.ts
+++ b/vscode/src/chat/protocol.ts
@@ -123,9 +123,7 @@ export const CODY_DOC_URL = new URL('https://docs.sourcegraph.com/cody')
 
 // Community and support
 export const DISCORD_URL = new URL('https://discord.gg/s2qDtYGnAE')
-export const CODY_FEEDBACK_URL = new URL(
-    'https://github.com/sourcegraph/cody/issues/new/choose'
-)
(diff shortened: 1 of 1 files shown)


vscode/src/chat/protocol.ts
```diff
@@ -124,7 +124,7 @@ export const CODY_DOC_URL = new URL('https://docs.sourcegraph.com/cody')
 // Community and support
 export const DISCORD_URL = new URL('https://discord.gg/s2qDtYGnAE')
 export const CODY_FEEDBACK_URL = new URL(
-    'https://github.com/sourcegraph/cody/discussions/new?category=product-feedback&labels=vscode'
+    'https://github.com/sourcegraph/cody/issues/new/choose'
```
taylorsperry, line 127: Is this a private repo? Will users be able to access it? @kalanchan

jdorfman: @toolmantim that works great. We can convert any issue into a discussion if needed

toolmantim: Nope, that’s the same repo @taylorsperry — just linking to issues instead of discussions.

taylorsperry: Nice!

PR #2114, opened 2023-12-05
Chat: Fix infinite loop when searching for symbols

toolmantim:
Fixes a case where symbols are constantly re-searched and re-rendered due to `useEffect` triggering itself.

Before:

https://github.com/sourcegraph/cody/assets/153/03cdf8e9-44e0-4ea9-86a2-b16cc8367aed

After:

https://github.com/sourcegraph/cody/assets/153/05becbfb-7a70-4abd-b39b-e93233b94e62

## Test plan

- Open sourcegraph/sourcegraph repo
- Start a chat
- Type `@#O` and wait until you get results
- See results are stable


abeatrix (approved):
ohhh nice catch!

PR #2118, opened 2023-12-05
Experimental hot streak

philipp-spiess:


vscode/src/completions/providers/hot-streak.ts
```diff
@@ -0,0 +1,105 @@
...
+                }
+                const processedCompletion = processCompletion(completion, eventualDynamicMultilineProviderOptions)
+                onHotStreakCompletionReady(updatedDocContext, {
+                    ...processedCompletion,
+                    stopReason: 'hot-streak-end',
+                })
+                updatedDocContext = insertCompletionAndPressEnter(updatedDocContext, processedCompletion)
+            } else {
+                // If we don't have enough in the remaining completion text to generate a full
+                // hot-streak completion so we yield.
```
dominiccooney, line 96: We can have if we ... we yield, or we ... so we yield but if ... so doesn't go; so joins complete sentences.

```suggestion
                // We don't have enough in the remaining completion text to generate a full
                // hot-streak completion so we yield.
```

vscode/src/completions/text-processing/utils.ts
```diff
@@ -366,3 +366,79 @@ export function getPrevNonEmptyLine(prefix: string): string {
...
+    let outputEnded = false
+    let doContinue: () => void = () => {}
+
+    const buffer: T[] = []
+
+    const stream = {
+        [Symbol.asyncIterator]() {
+            return this
+        },
+        async next(): Promise<{ done: false; value: T } | { done: true; value: null }> {
```
dominiccooney, line 409: Ooh, TypeScript FTW. It is neat to encode this property of the iterator protocol.

vscode/src/completions/text-processing/utils.ts
```diff
@@ -366,3 +366,79 @@ export function getPrevNonEmptyLine(prefix: string): string {
...
+
+                return stream.next()
+            }
+
+            outputEnded = true
+            return stream.next()
+        },
+        onChunk(chunk: T) {
+            if (inputEnded) {
+                return
```
dominiccooney, line 431: Why not throw here? The caller is buggy and we should not paper over those bugs.

vscode/src/completions/text-processing/utils.ts
```diff
@@ -366,3 +366,79 @@ export function getPrevNonEmptyLine(prefix: string): string {
...
+                return { done: true, value: null }
+            }
+
+            if (buffer.length > 0) {
+                const chunk = buffer.shift()!
+                return { done: false, value: chunk }
+            }
+
+            if (!inputEnded) {
+                const promise = new Promise<void>(resolve => (doContinue = resolve))
```
dominiccooney, line 420: This is a neat gadget.

dominiccooney (approved):
Approving to unblock. This is very cool.

The code will be more robust *over time* if we tested .length === 0 instead of <= 0, and if we threw when the caller is obviously buggy (for example onChunk after onEnd noted inline.)

Some grammar nits inline. Kudos for cleaning up the spelling of multiline in many places.

Most confusing part for me is the document context updating "position." What that is and means isn't clear to me, the casual reader.

vscode/src/completions/text-processing/utils.ts
philipp-spiess: Oh. I ended up not even needing these streaming helpers so I never got around to "clean up". 😬  Will remove the code.

vscode/src/completions/get-current-doc-context.test.ts
```diff
@@ -369,3 +369,78 @@ describe('getCurrentDocContext', () => {
...
+        )
+        const docContext = getCurrentDocContext({
+            document,
+            position,
+            maxPrefixLength: 140,
+            maxSuffixLength: 60,
+            dynamicMultilineCompletions: false,
+        })
+
+        const updatedDocContext = insertIntoDocContext(docContext, "'hello', 'world')", document.languageId)
```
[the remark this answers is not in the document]
philipp-spiess: Makes sense! Yeah removing here is an impl. detail we just want to make sure no characters are duplicated instead 👍

vscode/src/completions/get-current-doc-context.test.ts
```diff
@@ -369,3 +369,78 @@ describe('getCurrentDocContext', () => {
...
+                    console.log('hello', 'world')`,
+            suffix: '\n}',
+            currentLinePrefix: "    console.log('hello', 'world')",
+            currentLineSuffix: '',
+            prevNonEmptyLine: 'function helloWorld() {',
+            nextNonEmptyLine: '}',
+            multilineTrigger: null,
+            multilineTriggerPosition: null,
+            injectedPrefix: null,
+            position: { character: 33, line: 1 },
```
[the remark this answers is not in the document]
philipp-spiess: > So for single line completions, we move to the end of the line?

Yep, I documented this in the doc type for `insertIntoDocContext` too:

 * NOTE: This will always move the position to the _end_ of the line that the text was inserted at,
 *       regardless of wether the text was inserted before the sameLineSuffix.

vscode/src/completions/get-current-doc-context.ts
```diff
@@ -139,6 +139,46 @@ export function getDerivedDocContext(params: GetDerivedDocContextParams): Docume
...
+/**
+ * Inserts a completion into a specific document context and computes the updated positions.
+ *
+ * This will insert the completion at the `position` outlined in the document context and will
+ * replace the whole rest of the line with the completion.
+ *
+ * NOTE: This will always move the position to the _end_ of the line that the text was inserted at,
+ *       regardless of wether the text was inserted before the sameLineSuffix.
+ *
+ * TODO: Properly support {@link getRangeAdjustedForOverlappingCharacters}.
```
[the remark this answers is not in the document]
philipp-spiess: Leftover comment, sorry!

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -168,140 +188,124 @@ export async function fetchAndProcessDynamicMultilineCompletions(
...
                 (incompleteResponse: CompletionResponse) => {
-                    const initialCompletion = providerSpecificPostProcess(incompleteResponse.completion)
-                    const completion = canUsePartialCompletion(initialCompletion, providerOptions)
+                    const rawCompletion = providerSpecificPostProcess(incompleteResponse.completion)
+
+                    if (!completedCompletion) {
+                        const completion = canUsePartialCompletion(rawCompletion, providerOptions)
+                        if (completion) {
+                            completedCompletion = processCompletion(completion, providerOptions)
+                            params.onCompletionReady({ ...completedCompletion, stopReason: 'streaming-truncation' })
```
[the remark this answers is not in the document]
philipp-spiess: @dominiccooney Not sure I understand what your thought is here? Complex properties maybe but no mutation if that helps 😅  The object might contain debug data inferred to from the tree-sitter parsing though

vscode/src/completions/providers/hot-streak.ts
```diff
@@ -0,0 +1,105 @@
...
+
+        alreadyInsertedLength += completion.insertText.length + lastInsertedWhitespace.length
+
+        return updatedDocContext
+    }
+
+    function extract(rawCompletion: string, isRequestEnd: boolean): void {
+        do {
+            const unprocessedCompletion = rawCompletion.slice(alreadyInsertedLength)
+            if (unprocessedCompletion.length <= 0) {
```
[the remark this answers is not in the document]
philipp-spiess: I don't think strings can even have length < 0 so this is mostly me being paranoid, I can remove

vscode/src/completions/providers/fetch-and-process-completions.ts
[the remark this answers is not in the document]
dominiccooney: My thinking is that `...completedCompletion` copies the properties of `completedCompletion`. If any of those are mutable, the callback and this class now share those references and can interfere with observe each other's side effects.

philipp-spiess:
Changes are behind a flag anyways. :shipit:

vscode/src/completions/providers/fireworks.ts
```diff
@@ -78,13 +79,15 @@ function getMaxContextTokens(model: FireworksModel): number {
...
+    // To speed up sample generation in single-line case, we request a lower token limit
+    // since we can't terminate on the first `\n`.
+    maxTokensToSample: 30,
+    stopSequences: ['\n'],
+}
+const MULTI_LINE_COMPLETION_ARGS: Pick<CodeCompletionsParams, 'maxTokensToSample' | 'stopSequences'> = {
     maxTokensToSample: MAX_RESPONSE_TOKENS,
-    stopSequences: undefined,
-    timeoutMs: 15_000,
+    stopSequences: ['\n\n', '\n\r\n'],
```
philipp-spiess, line 90: @valerybugakov Was there a specific reason that previously, in the dynamic multiline branch, we would continue to generating even after two newlines?

vscode/src/completions/providers/anthropic.ts
```diff
@@ -137,56 +141,56 @@ export class AnthropicProvider extends Provider {
...
         if (prompt.length > this.promptChars) {
             throw new Error(`prompt length (${prompt.length}) exceeded maximum character length (${this.promptChars})`)
         }
-        const { dynamicMultilineCompletions, multiline } = this.options
+        const { multiline, n, dynamicMultilineCompletions, hotStreak } = this.options
+
+        const useExtendedGeneration = multiline || dynamicMultilineCompletions || hotStreak
 
         const requestParams: CodeCompletionsParams = {
+            ...(useExtendedGeneration ? MULTI_LINE_COMPLETION_ARGS : SINGLE_LINE_COMPLETION_ARGS),
```
valerybugakov, line 161: Nice

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -90,21 +109,21 @@ export async function fetchAndProcessDynamicMultilineCompletions(
                         }
                     } else {
                         /**
-                         * This completion was started without the multline trigger at the end of current line.
-                         * Check if the the first completion line ends with the multline trigger. If that's the case
+                         * This completion was started without the multiline trigger at the end of current line.
```
valerybugakov, line 112: ty

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -59,24 +74,28 @@ export async function fetchAndProcessDynamicMultilineCompletions(
...
                         stage: 'incomplete response',
-                        text: initialCompletion,
+                        text: rawCompletion,
                         obj: {
                             multiline,
                         },
                     })
 
+                    if (completedCompletion) {
+                        hotStreakExtractor?.extract(rawCompletion, false)
```
valerybugakov, line 89: Having an object here to explain the meaning of the second argument would be helpful.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -153,66 +156,73 @@ export class FireworksProvider extends Provider {
...
+        const { multiline, n, dynamicMultilineCompletions, hotStreak } = this.options
         const prompt = this.createPrompt(snippets)
 
         const model =
             this.model === 'starcoder-hybrid'
                 ? MODEL_MAP[multiline ? 'starcoder-16b' : 'starcoder-7b']
                 : MODEL_MAP[this.model]
 
-        const timeoutMs: number = multiline
+        const useExtendedGeneration = multiline || dynamicMultilineCompletions || hotStreak
```
valerybugakov, line 174: We can extract the shared bits from `provider.generateCompletions` into a separate function.

vscode/src/completions/providers/hot-streak.ts
```diff
@@ -0,0 +1,106 @@
...
+                      docContext: getUpdatedDocContext({
+                          ...params,
+                          initialCompletion: unprocessedCompletion,
+                          completionPostProcessId: '',
+                      }),
+                  }
+                : updatedProviderOptions
+
+            const completion = canUsePartialCompletion(unprocessedCompletion, eventualDynamicMultilineProviderOptions)
+            if (completion) {
```
valerybugakov, line 67: @philipp-spiess, should we continue sampling tokens only if the next sampled line is not empty? Consider the following scenario:

```ts
const value = █ 'foo'

function veryLongFunction() {
  ...
}
```

We continue sampling tokens until `canUsePartialCompletion` returns a completion (or the request ends, but we're not interested in this case). This means that after generating ` 'foo'`, we will sample the whole function declaration, which might be too much. It won't hurt the UX but can affect our spending significantly. WDYT?

vscode/src/completions/request-manager.ts
```diff
@@ -87,11 +87,29 @@ export class RequestManager {
...
         Promise.all(
-            providers.map(provider =>
-                startAsyncSpan('autocomplete.generate', () =>
-                    provider.generateCompletions(request.abortController.signal, context, tracer)
-                )
-            )
+            providers.map(provider => {
+                const completionReadyPromise = new Promise<InlineCompletionItemWithAnalytics[]>((resolve, reject) => {
+                    provider
+                        .generateCompletions(
```
valerybugakov, line 93: Named object argument would be helpful to understand the meaning of the params.

vscode/src/completions/request-manager.ts
```diff
@@ -87,11 +87,29 @@ export class RequestManager {
         this.inflightRequests.add(request)
 
         Promise.all(
-            providers.map(provider =>
-                startAsyncSpan('autocomplete.generate', () =>
-                    provider.generateCompletions(request.abortController.signal, context, tracer)
-                )
-            )
+            providers.map(provider => {
+                const completionReadyPromise = new Promise<InlineCompletionItemWithAnalytics[]>((resolve, reject) => {
```
valerybugakov, line 91: We must create another promise wrapper to make this call compatible with `onCompletionReady`. I would like to understand why relying on the native promise resolve mechanism is insufficient here. E.g., `await provider.generateCompletions(...)`

valerybugakov (commented):
Great work, @philipp-spiess! I left some comments mainly focused on understanding the updated data flow.

vscode/src/completions/providers/fireworks.ts
philipp-spiess: Can you give a concrete example? I feel like multi-line completions that try to implement more than one method in a class are always a bit too long for my personal preference but maybe I miss something? 

e.g. this case:

```
class Foo {
  bar() {
    // ...
   }

  baz() {
    // ...
   }
}
```

Feels like it's too much for a single completion, wdyt?

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -31,19 +32,33 @@ interface FetchAndProcessCompletionsParams {
...
     // eslint-disable-next-line @typescript-eslint/no-misused-promises, no-async-promise-executor
     return new Promise(async (resolve, reject) => {
         try {
             const abortController = forkSignal(abortSignal)
 
             function stopStreamingAndUsePartialResponse(completionItem: InlineCompletionItemWithAnalytics): void {
-                resolve({ ...completionItem, stopReason: 'streaming-truncation' })
+                completedCompletion = completionItem
+                params.onCompletionReady({ ...completionItem, stopReason: 'streaming-truncation' })
+                resolve()
```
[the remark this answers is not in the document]
philipp-spiess: It's also so we can keep the `Promise` running while the network request is still running. It's not great - I was thinking to switching to streaming interfaces - but it does feel simpler this way. The Promise will run until the request is either cancelled or completed, however the completion will be yielded as _ready_ as soon as it is ready to be displayed.

vscode/src/completions/providers/hot-streak.ts
philipp-spiess: I think in this case the `stopSequence` of `\n\n` should stop it. 😅  Do you think it's better if we add explicit checks for empty lines in the code?

vscode/src/completions/request-manager.ts
philipp-spiess: Hmm, I guess you're right. I thought that conceptually it would be better if we can communicate to the caller how long the network request is open. I had a future optimization in place where we start generation until one of the completions got rejected. If that's the case and the request is still running, we know that we're emitting hot streaks only and the generation can be terminated.

Another option is to return two promises: one for the main completion and one for the network request. WDYT?

vscode/src/completions/providers/fireworks.ts
valerybugakov: I should have noticed in my review that this change applies to dynamic multiline completions. I think we should revert it because it leads to multiline completions being truncated too early.

Screenshot 2023-12-13 at 14 32 16

There's a '\n\n` after this line, and the function implementation continues.

valerybugakov: PR to revert this change for dynamic multiline completions: https://github.com/sourcegraph/cody/pull/2326

vscode/src/completions/providers/fetch-and-process-completions.ts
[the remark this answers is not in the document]
valerybugakov: > It's also so we can keep the Promise running while the network request is still running

It's not apparent why we want this behavior. Could you elaborate?

vscode/src/completions/providers/hot-streak.ts
valerybugakov: I missed the part where you changed stop sequences for dynamic multline completions. I think we should remove stop sequences for dynamic multiline; then my concern there would be valid if we have both hot-streak and dynamic multiline enabled.

vscode/src/completions/request-manager.ts
valerybugakov: > Another option is to return two promises: one for the main completion and one for the network request. WDYT?

I like this option more because it's more explicit. It's unclear why this mechanism was chosen without comments explaining the rationale you shared above. Having two promises would be more self-explanatory

PR #2122, opened 2023-12-05
fix: invert notification toggle logic

abeatrix:
The notification toggle logic was inverted, so notifications were being enabled when they should have been disabled. This fixes the logic to disable notifications when intended.

## Test plan

all e2e tests should be passing


kalanchan (approved):

toolmantim:
I thought it was just my computer 😅

PR #2126, opened 2023-12-05
Chat: Speed up chat panel debounce w/ trigger on leading edge too

toolmantim:
Follow up to #2115, this ensures we trigger chat panel creation and restoration straight away (on the leading edge of the timeout), and reduces the restoration timeout to match.

## Test plan

- Open new chat
- Open existing chat
- Reload window and ensure restores correct


abeatrix (approved):
🙏

PR #2130, opened 2023-12-06
Chat: update embeddings and codebase identifiers

beyang:
image

Fixes https://github.com/sourcegraph/cody/issues/2107

Previously, `SimpleChatPanelProvider` accepted an instance of `EmbeddingsSearch`, which was extracted from the `CodebaseContext` instance created on startup.
* `EmbeddingsSearch` instances are tied to a specific codebase.
* When the current codebase changes (e.g., file in a separate repo was opened or a new repo was added to the workspace), this approach does not update the `EmbeddingsSearch` instance held by `SimpleChatPanelProvider`.
* As a result, the remote embeddings context was fetched from the old codebase

This patch modifies `SimpleChatPanelProvider` to switch out `EmbeddingsSearch` for `CachedRemoteEmbeddingsClient`, which wraps a `SourcegraphGraphQLAPIClient` instance.
* The current codebase is now provided (to `SimpleChatPanelProvider`) by a new helper class `CodebaseStatusProvider`.
* When the user codebase changes, the `CodebaseStatusProvider` updates and also signals to `SimpleChatPanelProvider` this change has occurred...
* ...so the remote embeddings search can target the new active codebase.

The second commit fixes a bug where the new chat panel was not restoring properly on startup.

## Test plan

- [ ] Create a single folder workspace
- [ ] Start a chat asking about that folder
- [ ] Add a new repo and switch to a file in that repo
- [ ] Check that the context status has updated. Ask a question about the new codebase
- [ ] Switch to more files and check that the context status and chat functions properly
- [ ] Try adding a repo subdirectory. Cody should *not* target the new repository, because it should not refer to files outside the scope of the workspace folder

### DIFF
--- a/vscode/src/chat/ContextProvider.ts
+++ b/vscode/src/chat/ContextProvider.ts
@@ -277,12 +277,6 @@ export class ContextProvider implements vscode.Disposable, ContextStatusProvider
         this.disposables = []
     }
 
-    public async hackGetEmbeddingClientCandidates(
-        config: GraphQLAPIClientConfig
-    ): Promise<SourcegraphGraphQLAPIClient[]> {
-        return this.getEmbeddingClientCandidates(config)
-    }
-
     // Gets a list of GraphQL clients to interrogate for embeddings
     // availability.
     private getEmbeddingClientCandidates(config: GraphQLAPIClientConfig): Promise<SourcegraphGraphQLAPIClient[]> {
@@ -303,28 +297,6 @@ export class ContextProvider implements vscode.Disposable, ContextStatusProvider
     public onDidChangeStatus(callback: (provider: ContextStatusProvider) => void): vscode.Disposable {
         return this.contextStatusChangeEmitter.event(callback)
     }
-}
-
-export function hackGetCodebaseContext(
-    config: Config,
-    rgPath: string | null,
-    symf: IndexedKeywordContextFetcher | undefined,
-    editor: Editor,
-    chatClient: ChatClient,
-    platform: PlatformContext,
-    embeddingsClientCandidates: readonly SourcegraphGraphQLAPIClient[],
-    localEmbeddings: LocalEmbeddingsController | undefined
-): Promise<CodebaseContext | null> {
-    return getCodebaseContext(
-        config,
-        rgPath,
-        symf,
-        editor,
-        chatClient,
-        platform,
-        embeddingsClientCandidates,
-        localEmbeddings
-    )
 }
 
 /**
@@ -355,9 +327,8 @@ async function getCodebaseContext(
         return null
     }
 
-    // TODO: When SimpleChatContextProvider stops using hackGetCodebaseContext,
-    // it must start sending localEmbeddings.load directly to the embeddings
-    // controller.
+    // TODO: When we remove this class (ContextProvider), SimpleChatContextProvider
+    // should be updated to invoke localEmbeddings.load when the codebase changes
     const repoDirUri = gitDirectoryUri(workspaceRoot)
     const hasLocalEmbeddings = repoDirUri ? localEmbeddings?.load(repoDirUri) : false
     let embeddingsSearch = await EmbeddingsDetector.newEmbeddingsSearchClient(
--- a/vscode/src/chat/chat-view/chat-helpers.ts
+++ b/vscode/src/chat/chat-view/chat-helpers.ts
@@ -8,7 +8,8 @@ import {
 } from '@sourcegraph/cody-shared/src/codebase-context/context-status'
 import { ContextFile, ContextMessage } from '@sourcegraph/cody-shared/src/codebase-context/messages'
 import { Editor } from '@sourcegraph/cody-shared/src/editor'
-import { convertGitCloneURLToCodebaseName } from '@sourcegraph/cody-shared/src/utils'
+import { isDotCom } from '@sourcegraph/cody-shared/src/sourcegraph-api/environments'
+import { convertGitCloneURLToCodebaseName, isError } from '@sourcegraph/cody-shared/src/utils'
 
 import { repositoryRemoteUrl } from '../../repository/repositoryHelpers'
 import { CachedRemoteEmbeddingsClient } from '../CachedRemoteEmbeddingsClient'
@@ -139,17 +140,19 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
  */
 export class CodebaseStatusProvider implements vscode.Disposable, ContextStatusProvider {
     private disposables: vscode.Disposable[] = []
-    private _currentCodebase: CodebaseIdentifiers | null | undefined = undefined
-    private statusCallbacks: ((provider: ContextStatusProvider) => void)[] = []
+    private eventEmitter: vscode.EventEmitter<ContextStatusProvider> = new vscode.EventEmitter<ContextStatusProvider>()
 
-    // TODO(beyang): mix of dependency injection and direct vscode API access here
+    // undefined means uninitialized, null means current codebase is empty
+    private _currentCodebase: CodebaseIdentifiers | null | undefined = undefined
+
     constructor(
         private editor: Editor,
         private embeddingsClient: CachedRemoteEmbeddingsClient
     ) {
         this.disposables.push(
             vscode.window.onDidChangeActiveTextEditor(() => this.syncCodebase()),
-            vscode.workspace.onDidChangeWorkspaceFolders(() => this.syncCodebase())
+            vscode.workspace.onDidChangeWorkspaceFolders(() => this.syncCodebase()),
+            this.eventEmitter
         )
     }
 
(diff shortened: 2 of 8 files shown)


vscode/src/chat/ContextProvider.ts
```diff
@@ -355,9 +327,8 @@ async function getCodebaseContext(
         return null
     }
 
-    // TODO: When SimpleChatContextProvider stops using hackGetCodebaseContext,
-    // it must start sending localEmbeddings.load directly to the embeddings
-    // controller.
+    // TODO: When we remove this class (ContextProvider), SimpleChatContextProvider
+    // should be updated to invoke localEmbeddings.load when the codebase changes
```
beyang, line 331: I updated the comment here. The change itself can't be made, because the local embeddings controller is still updated by the CodebaseContext. When we remove CodebaseContext entirely, we'll need to implement this change.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -123,3 +133,99 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
...
+
+/**
+ * Provides and signals updates to the current codebase identifiers to use in the chat panel.
+ */
+export class CodebaseStatusProvider implements vscode.Disposable, ContextStatusProvider {
+    private disposables: vscode.Disposable[] = []
+    private _currentCodebase: CodebaseIdentifiers | null | undefined = undefined
+    private statusCallbacks: ((provider: ContextStatusProvider) => void)[] = []
+
+    // TODO(beyang): mix of dependency injection and direct vscode API access here
```
beyang, line 145: When we update this to support other editors, we will want to dependency-inject the additional vscode APIs.

vscode/src/chat/ContextProvider.ts
dominiccooney: The local embeddings provider loads a single index at a time today, but they are very cheap to load (once cody-engine is started.) Should I change cody-engine so local embeddings can load multiple indexes? That way classic ContextProvider and SimpleChatProvider can both hammer on the same LocalEmbeddings object without interference, and you can completely decouple SimpleChatProvider from ContextProvider.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -123,3 +133,99 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
...
+        this.disposables = []
+    }
+
+    public onDidChangeStatus(callback: (provider: ContextStatusProvider) => void): Disposable {
+        this.statusCallbacks.push(callback)
+        return {
+            dispose: () => {
+                this.statusCallbacks = this.statusCallbacks.filter(cb => cb !== callback)
+            },
+        }
```
dominiccooney, line 169: You should add a `vscode.EventEmitter<ContextStatusProvider>` property and just pass through to `.event` for subscription, `.fire` for notification, and dispose of the event emitter when done.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -123,3 +133,99 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
...
+            s(this)
+        }
+    }
+
+    private static fetchCurrentCodebase(editor: Editor): CodebaseIdentifiers | null {
+        const workspaceRoot = editor.getWorkspaceRootUri()
+        if (!workspaceRoot) {
+            return null
+        }
+        const remoteUrl = repositoryRemoteUrl(workspaceRoot)
```
abeatrix, line 217:
```suggestion
        const remoteUrl = repositoryRemoteUrl(editor.getActiveTextEditor()?.fileUri || workspaceRoot)
```

This supports workspace with multi-repositories.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -123,3 +133,99 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
...
+        const codebase = this.currentCodebase
+        return codebase?.remote
+            ? [
+                  {
+                      name: codebase.local,
+                      providers: [
+                          {
+                              kind: 'embeddings',
+                              type: 'remote',
+                              state: 'ready',
```
dominiccooney, line 181: You need to produce the `no-match` state when there are no remote embeddings to produce the UX design for Enterprise users who don't have a repository indexed.

For dotcom users, you should continue to produce the empty group/empty providers so that they will just see the local embeddings status.

vscode/src/chat/chat-view/chat-helpers.ts
dominiccooney: Agent emulates VScode APIs, so we do not need to inject these, but ensure the agent stubs/thunks provide the APIs we use.

abeatrix (approved):
Works in workspace with single repository. Made an inline suggestion to support workspace with multi-repositories.

### Before

https://github.com/sourcegraph/cody/assets/68532117/e37487ff-bef4-4f76-80e8-df38b6267519

#### After

https://github.com/sourcegraph/cody/assets/68532117/14a40c1c-ff84-4166-a565-5f9a1ece051d

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -123,3 +133,99 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
...
+                  },
+              ]
+            : []
+    }
+
+    public get currentCodebase(): CodebaseIdentifiers | null {
+        if (this._currentCodebase === undefined) {
+            this._currentCodebase = CodebaseStatusProvider.fetchCurrentCodebase(this.editor)
+        }
+        for (const s of this.statusCallbacks) {
```
dominiccooney, line 195: If you fire this here you will cause a lot of needless tree diffing in React, slowing down the UI thread. You should fire this event when the status changes (or may have changed if you're unsure.) Look at the implementation of `status` and whatever it depends on—currentCodebase, currentCodebase.remote, currentCodebase.local, etc.—and just fire when one of those has changed.

Just moving it inside the `if` is a start, not sure if the sub-properties can change independent of `fetchCurrentCodebase`.

dominiccooney (changes requested):
Looks good, but fire the status callback more precisely and reuse an EventEmitter to do it.

LMK your thoughts about decoupling LocalEmbeddings.load from ContextProvider, we can do that together in a follow up.

In a follow up change, I think you need to put local and remote embeddings together and wire them both to your copy of the codebase change signal. You need to coordinate between local and remote embeddings. The intended design as I understand it is:

- No local embeddings for Enterprise.
- Non-Enterprise uses dotcom embeddings if available.
- Otherwise use local embeddings.

Because cody-engine download and startup is slow, local embeddings has the "start" hint, "load" returns a boolean indicating whether there's an index to query, *and* there's a change signal indicating you should re-query load.

And on the query side you need to query one or the other, not both.

vscode/src/chat/chat-view/chat-helpers.ts
beyang: `Editor.getWorkspaceUri` returns the workspace URI of the active editor's workspace folder.

vscode/src/chat/ContextProvider.ts
beyang: That sounds good to me. Eventually, I would like to remove the class ContextProvider completely. I think there are a few outstanding dependencies on it in chat and then also in commands.

vscode/src/chat/chat-view/chat-helpers.ts
beyang: Removed comment.

vscode/src/chat/chat-view/chat-helpers.ts
beyang: Done.

vscode/src/chat/chat-view/chat-helpers.ts
beyang: Done. Doing this required rolling in a remote API call to fetch the repoID for the remote embeddings (if they exist). I've incorporated that into the CodebaseStatusProvider class.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -123,3 +134,130 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
...
+                }
+            }
+        }
+
+        // codebase local identifier changed, fire callbacks
+        const shouldAlert =
+            this._currentCodebase?.local !== newCodebase?.local ||
+            this._currentCodebase?.remote !== newCodebase?.remote ||
+            this._currentCodebase?.remoteRepoId !== newCodebase?.remoteRepoId
+        this._currentCodebase = newCodebase
```
dominiccooney, line 252: Looks great.

The legacy ContextProvider did one other thing: Before updating the "current" codebase, it checked if the current codebase had changed. This can happen if some async call is slow and the second update completes first.

SimpleChatProvider is loaded down with less stuff, so it should happen less often, so up to you if you want to add a check like that.

https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@5c9d1cd32b355b1e2897db6bac7a87f9c5955b9c/-/blob/vscode/src/chat/ContextProvider.ts?L160

dominiccooney (approved):
Looks good. FYI inline about racing workspace updates.

beyang:
> The legacy ContextProvider did one other thing: Before updating the "current" codebase, it checked if the current codebase had changed. This can happen if some async call is slow and the second update completes first.

Hmm, I see your point about the potential race, but I'm not sure noop-ing if the original current codebase changed is correct in all cases. For instance, let's say you have a very slow network connection and quickly open 2 files in succession in separate workspaceFolders. If the request order is preserved, then noop-ing would "forget" to switch to the 2nd codebase (which is the correct one).

vscode/src/chat/chat-view/chat-helpers.ts
abeatrix: getWorkspaceUri doesn't change on nested repositories though, which's not necessary if that's not something we support 👍

PR #2135, opened 2023-12-06
plg: display errors when autocomplete rate limits trigger

### LINKED ISSUE #2134: bug: Autocomplete rate limits display inconsistent error messages
### Version

v0.17.1701788840

### Describe the bug

This bug has 2 parts to it:
1. Using the rate limited demo accounts, by design the headers `retry-after` return back the same time as when the request was made. This causes the error messages to constantly be displayed in the quick pick view
2. Also using the rate limited demo account, if you hard code the retry time to a regular interval (like 1 day after), the error message does not display anymore after dismissing it once. 

### Expected behavior

Rate limit error message to only display once, and to display again even after dismissed

### Additional context

_No response_

kalanchan:
Closes #2134 

This bug has 2 parts to it:
1. Using the rate limited demo accounts, by design the headers `retry-after` return back the same time as when the request was made. This causes the error messages to constantly be displayed in the quick pick view
2. Also using the rate limited demo account, if you hard code the retry time to a later time (like 1 day after), the error message does not display anymore after being dismissed once. 

This PR adds a check to see whether or not the errors have been added in the `statusBar` and to display the rate limited error if no other `RateLimitError` is present. This also makes sure the error is only displayed once, instead of spamming the user if autocomplete keeps getting triggered.

**Before (bug 1):**

https://github.com/sourcegraph/cody/assets/51868853/de5a963a-a5be-4677-a75c-4f154960a516

**Before (bug2, after hardcoding a later `retry-after` date):**

https://github.com/sourcegraph/cody/assets/51868853/ffcd7748-0827-4fca-b71f-e141112ead97

After: 

https://github.com/sourcegraph/cody/assets/51868853/48078429-59dc-4ad8-a744-efb1e665a689

## Test plan
- login to `https://sourcegraph.com` using the free demo account ([thread](https://sourcegraph.slack.com/archives/C066F85TE80/p1701712911448229?thread_ts=1700694252.150349&cid=C066F85TE80))
- go to any file, trigger autocompletes multiple times and you will see only 1 error message displayed
- click on the error, dismiss it, and trigger another autocomplete. you should see another error message displayed


vscode/src/services/StatusBar.ts
```diff
@@ -17,6 +18,10 @@ export interface CodyStatusBar {
     dispose(): void
     startLoading(label: string): () => void
     addError(error: StatusBarError): () => void
+    checkErrors(): {
+        error: StatusBarError
+        createdAt: number
+    }[]
```
DanTup, line 24: Any reason to use a method `checkErrors()` rather than just using `errors` (exposing the underlying field)? I ask because the name feels a little odd to me - it's getting the errors, not really checking anything.

DanTup (approved):
Seems reasonable to me - added some nits (feel free to ignore!).

vscode/src/services/StatusBar.ts
philipp-spiess, line 24: Can this be done by setting a bit in the inline-completion-item-provider and then using the callbacks to reset it and avoid exposing this interface?

philipp-spiess (commented):
> by design the headers retry-after return back the same time as when the request was made.

Wait this doesn't really make sense, shouldn't it be a time in the future at which we can retry the requests again? If we set it to the same time, what value does the header bring?

philipp-spiess:
I think something with the rate limited test account might be incorrect as it says "limit will reset in less than a minute". This isn't how our rate limiter works AFAIK?

philipp-spiess:
Screenshot 2023-12-06 at 11 44 25

There is clearly something I misunderstand about this header 😆  Should we remove the text that depends on the `retry-after` value?

philipp-spiess:
Getting some clarity here: https://sourcegraph.slack.com/archives/C05AGQYD528/p1701859689434749

vscode/src/services/StatusBar.ts
kalanchan: good call! changed this to `hasError` in the latest commit

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -575,7 +575,16 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
      */
     private onError(error: Error): void {
         if (error instanceof RateLimitError) {
-            if (this.resetRateLimitErrorsAfter && this.resetRateLimitErrorsAfter > Date.now()) {
+            let hasRateLimitError
+            const allStatusBarErrors = this.config.statusBar.checkErrors()
```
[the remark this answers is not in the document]
kalanchan: thats why you're my TL! thanks danny I made your suggestion

vscode/src/services/StatusBar.ts
```diff
@@ -9,6 +10,7 @@ import { FeedbackOptionItems } from './FeedbackOptions'
 interface StatusBarError {
     title: string
     description: string
+    errorType: string | RateLimitError
```
philipp-spiess, line 13:
```suggestion
    errorType: string
```

That should not be set to a class I assume? 🤔

philipp-spiess (approved):
gogog

vscode/src/services/StatusBar.ts
DanTup: I changed this to `type StatusBarErrorName = 'auth' | 'RateLimitError'` for now (since these are the only two we use), since it felt a little less error-prone than just `string`.

Also pushed some fixes to the tests because the mocks didn't have the `hasError` impl.

DanTup:
@kalanchan pushed some tweaks (https://github.com/sourcegraph/cody/pull/2135/commits/363a716c402be4f04a49eaabb175bc7a0d2bdf5c), think this is good to merge?

kalanchan:
@philipp-spiess @DanTup THANKS ALL for reviewing and helping, feels good to be contributing back to the team 💪

PR #2136, opened 2023-12-06
simple chat: sync webview config and auth status on update

abeatrix:
Update the ChatPanelProvider and SimpleChatPanelProvider to sync the latest config, chat models, and auth status to the webview whenever the announceNewAuthStatus event fires in the auth provider. This keeps the webview in sync with the latest user auth state.

Currently, simple chat panels are not getting auth / config updated on change.

## Test plan

1. Log into your S2 account in cody
2. ask cody a question
3. Switch to an account that is rate limited on dot com
4. ask cody a question
5. The expected behavior is to see the rate limit error message instead of the getting response back from the S2 instance

### Before

https://github.com/sourcegraph/cody/assets/68532117/fce2c6c6-6a51-48a1-93c7-ac9ebc4e7aff

### After

https://github.com/sourcegraph/cody/assets/68532117/d8b339d1-6511-4484-9783-d0c97212c92c

### Doesn't affect current chat panel provider

https://github.com/sourcegraph/cody/assets/68532117/665dd9f9-7e9a-4b92-a439-8e9758606a42

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -23,14 +23,12 @@ import { Message } from '@sourcegraph/cody-shared/src/sourcegraph-api'
 import { isError } from '@sourcegraph/cody-shared/src/utils'
 
 import { View } from '../../../webviews/NavBar'
-import { getFullConfig } from '../../configuration'
 import { getFileContextFiles, getOpenTabsContextFile, getSymbolContextFiles } from '../../editor/utils/editor-context'
 import { VSCodeEditor } from '../../editor/vscode-editor'
 import { ContextStatusAggregator } from '../../local-context/enhanced-context-status'
 import { LocalEmbeddingsController } from '../../local-context/local-embeddings'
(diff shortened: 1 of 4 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -415,6 +412,14 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
         })
     }
 
+    // Update webview with the latest config, chat models, and authStatus
+    // This is triggered everytime announceNewAuthStatus in authProvider is fired
+    public syncWebviewConfig(authStatus: AuthStatus, configForWebview: ConfigurationSubsetForWebview & LocalEnv): void {
```
dominiccooney, line 417: @beyang suggested naming methods which talk to the webview "post..." as a consistent mnemonic with "postMessage"

abeatrix: @dominiccooney thanks for the review! looks like this is no longer an issue on main so I will just leave this PR open for now

Issue #2100, opened 2023-12-05
bug: /doc command adds comments in wrong place for Python functions


toolmantim:
### Version

main (18afbe650fe052bfc3a22100d3cedd33f0d22b7e)

### Describe the bug

Documenting functions in Python places comments incorrectly above the `def`, which creates broken code:

<img width="803" alt="Screenshot 2023-12-05 at 4 47 53 pm" src="https://github.com/sourcegraph/cody/assets/153/8c8ee74b-d367-460f-a1d3-d6df369c1c7c">

https://github.com/sourcegraph/cody/assets/153/563a4e37-5349-441a-8ed6-534d6f574146

### Expected behavior

Documenting whole functions follows https://peps.python.org/pep-0257/

e.g.

```python
@app.websocket("/generate-code")
async def stream_code(websocket: WebSocket):
    """
    stream_code handles a websocket connection to generate code from a screenshot.

    It receives the screenshot and parameters over the websocket, calls the OpenAI API to generate code, handles image generation, and streams the results back over the websocket.

    Key steps:
    - Get API key from params or environment 
    - Assemble prompt from screenshot
    - Call OpenAI API to generate code
    - Generate images if enabled
    - Stream generated code back to client
    - Write logs for debugging
    """
    await websocket.accept()
```

instead of:

```python
@app.websocket("/generate-code")
"""
stream_code handles a websocket connection to generate code from a screenshot.

It receives the screenshot and parameters over the websocket, calls the OpenAI API to generate code, handles image generation, and streams the results back over the websocket.

Key steps:
- Get API key from params or environment 
- Assemble prompt from screenshot
- Call OpenAI API to generate code
- Generate images if enabled
- Stream generated code back to client
- Write logs for debugging
"""
async def stream_code(websocket: WebSocket):
    await websocket.accept()
```

umpox:
Closed by https://github.com/sourcegraph/cody/commit/d881df4fb372d075d07f4ef2f1a5cdfea0d06677

PR #2143, opened 2023-12-06
fix: use default selection range for python doc

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2100

The getDocCommandRange function that is used for setting the range for the /doc comment was modified to always return the default selection range instead of trying to move to the line below the function for python.

## Test plan

Try to run the doc command in python code, the doc string should be added above the function name instead of below

### DIFF
--- a/vscode/src/commands/CommandRunner.ts
+++ b/vscode/src/commands/CommandRunner.ts
@@ -210,12 +210,7 @@ export function addSelectionToPrompt(prompt: string, code: string): string {
 /**
  * Gets the range to use for inserting documentation from the doc command.
  *
- * For Python files, returns a range starting on the line after the selection,
- * at the first non-whitespace character. This will insert the documentation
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):

toolmantim (commented):
This still seems to generate the comment above the function def for me, instead of below:

https://github.com/sourcegraph/cody/assets/153/fa58b9bb-23dc-4b76-879c-306475408e6c

abeatrix:
> This still seems to generate the comment above the function def for me, instead of below:
> 
>  Screen.Recording.2023-12-06.at.9.07.20.pm.mov

woops i misunderstood the issue.  thought you wanted it above the function name instead of below it, my bad 😅

kalanchan:
@abeatrix tom also has the fix in https://github.com/sourcegraph/cody/pull/2290

Issue #2139, opened 2023-12-06
bug: Highlight code and asking questions doesn’t work with chat with simplified chat


kalanchan:
### Version

tested on main - commit d2ac7f2e

### Describe the bug

Highlight code and asking question doesn’t work with chat, gave an unrelated answer to the code snippet highlighted

### Expected behavior

Use the highlighted code as context, but instead gave a random answer

### Additional context

https://github.com/sourcegraph/cody/assets/51868853/55a74c70-332b-4b84-961d-c14bfaaface0

beyang:
This is by design. The contract of the new chat is that all context must be explicitly @-mentioned if you don't have "enhanced context" checked.

PR #2144, opened 2023-12-06
chat: always include selection in enhanced context

beyang:
Fixes https://github.com/sourcegraph/cody/issues/2139

## Test plan

- [ ] Test example in the issue


toolmantim (approved):
Works a treat!

PR #2145, opened 2023-12-06
chore: Bump up macOS e2e runner instance size

toolmantim:
Follow on from #2127, #2146 and #2147, this increases the macOS runner size which takes the macOS tests from 6m to 4m.

I've broken this out into its own single change, because it requires us update our required checks.

## Test plan

- N/A


philipp-spiess (approved):
Wowzie

PR #2151, opened 2023-12-06
Fix isRateLimitError

mkondratek:


lib/shared/src/sourcegraph-api/errors.ts
```diff
@@ -22,11 +22,13 @@ export class RateLimitError extends Error {
         super(message)
         this.userMessage = `You've used all${limit ? ` ${limit}` : ''} ${feature} for today.`
         this.retryMessage = retryAfter ? `Usage will reset in ${formatDistance(retryAfter, new Date())}.` : undefined
+
+        Object.setPrototypeOf(this, RateLimitError.prototype)
```
philipp-spiess, line 26: Huh this shouldn't be necessary anymore 🤔 What runtime were you using? Was this the agent nodejs env?

philipp-spiess (approved):
Stamping to unblock but I don't think either of this should be necessary 🤷

lib/shared/src/sourcegraph-api/errors.ts
mkondratek: thank! you are right

philipp-spiess (approved):

PR #2166, opened 2023-12-07
Enable symf-based context for chat (cody.experimental.symfContext)

beyang:

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -161,7 +161,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
         this.codebaseStatusProvider = new CodebaseStatusProvider(
             this.editor,
             embeddingsClient,
-            this.config.experimentalSpladeContext ? this.symf : null
+            this.config.experimentalSymfContext ? this.symf : null
         )
         this.disposables.push(this.contextStatusAggregator.addProvider(this.codebaseStatusProvider))
     }
@@ -525,7 +525,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
                 this.editor,
                 this.embeddingsClient,
                 this.localEmbeddings,
-                this.config.experimentalSpladeContext ? this.symf : null,
+                this.config.experimentalSymfContext ? this.symf : null,
                 this.codebaseStatusProvider
             )
             const { prompt, warnings, newContextUsed } = await this.prompter.makePrompt(
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -158,7 +158,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
         if (this.localEmbeddings) {
             this.disposables.push(this.contextStatusAggregator.addProvider(this.localEmbeddings))
         }
-        this.codebaseStatusProvider = new CodebaseStatusProvider(this.editor, embeddingsClient)
+        this.codebaseStatusProvider = new CodebaseStatusProvider(this.editor, embeddingsClient, this.symf)
         this.disposables.push(this.contextStatusAggregator.addProvider(this.codebaseStatusProvider))
     }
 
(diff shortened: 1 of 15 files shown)


taylorsperry:
Can we have this say "Search" instead of "Symf"? [Designs](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=331-114717&mode=design&t=zAST0nfTpe3yrGxg-0).

Screenshot 2023-12-06 at 6 27 18 PM

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -959,6 +976,56 @@ class ContextProvider implements IContextProvider {
...
+    private async searchSymf(userText: string): Promise<ContextItem[]> {
+        if (!this.symf) {
+            return []
+        }
+        const workspaceRoot = this.editor.getWorkspaceRootUri()?.fsPath
+        if (!workspaceRoot) {
+            return []
+        }
+
+        const r0 = (await this.symf.getResults(userText, [workspaceRoot])).map(async resultsPromise => {
```
dominiccooney, line 991: You can simplify r1/line 1025.

If you make `r0` here, and `items` below, use `flatMap` instead of `map` you should be two levels flatter and not contain nulls, so you don't need one `all`, the `flat`, the `isContextItem`. Lines 1025-6 should just look like:

```
return Promise.all(r0)
```

You will need to make the innermost callback return `[{uri: ...}]` or `[]` instead of a result/null.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -137,6 +140,8 @@ export function getChatPanelTitle(lastDisplayText?: string, truncateTitle = true
 
 /**
  * Provides and signals updates to the current codebase identifiers to use in the chat panel.
+ * If this becomes too unwieldy, it may make sense to break this out into 2 separate
```
dominiccooney, line 143: I use `/**` style comments as documentation for the caller, and put commentary about the implementation in `//`. 

Not sure we need this comment, though. Either the code is unwieldy, so put a TODO, or it is not unwieldy, so not really worth mentioning. If it is going to be broken up we should be able to write:

```
// TODO: Break this up into CodebaseProvider and and ContextStatusProvider which
// interact with each other when [...specify the trigger condition for acting on the TODO]
```

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -173,19 +179,55 @@ export class CodebaseStatusProvider implements vscode.Disposable, ContextStatusP
...
+            return []
+        }
+        const symfIndexExists = this._currentCodebase.symfIndexExists || false
+        if (!symfIndexExists) {
+            return []
+        }
+        return [
+            {
+                kind: 'search',
+                state: 'ready',
```
dominiccooney, line 213: We can show indexing states in this dialog, do we know if symf is indexing?

dominiccooney (approved):
Looks good.

One suggestion inline about simplifying some of the Promise array squashing.

dominiccooney:
@taylorsperry it *should* do that, unless @beyang changed the types in lib/shared/src/codebase-context/context-status.ts . Here's the Storybook for the component:

Screenshot 2023-12-07 at 11 30 58

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -959,6 +976,56 @@ class ContextProvider implements IContextProvider {
...
+                    text = await this.editor.getTextEditorContentForFile(uri, range)
+                    if (!text) {
+                        return null
+                    }
+                } catch (error) {
+                    logError('SimpleChatPanelProvider.searchSymf', `Error getting file contents: ${error}`)
+                    return null
+                }
+                return {
+                    uri: displayUri,
```
abeatrix, line 1018: should we add an additional `source` field to ContextItem to make it easier to identify which each ContextItem comes from (in a follow up PR)?

abeatrix (approved):

vscode/src/chat/chat-view/chat-helpers.ts
beyang: Not at the moment, I need to add that to the SymfRunner class. That will be a bit more involved, so will do it in another PR.

PR #2182, opened 2023-12-07
Reset feature flags when switching through logins

### LINKED ISSUE #2154: Autocomplete: When logging in as a new user, feature flags and autocomplete config is not reset
This can lead to weird situations:

- Logging in to an enterprise instance still makes requests to dotcom
- When updating autocomplete settings, we still use the old feature flag values for a different instance

philipp-spiess:
Fixes #2181
Fixes #2154

This PR attempts to clean up the auth flow and makes it so that when the user account is switched, we never accidentely use feature flags of a wrong server URL and also ensures we reconfigure Autocomplete so it works. 

It does this by removing the relationship between the `AuthProvider` and a specific webview. This is necessary because:

- With features like Autocomplete, we have things that depend on the AuthProvider unrelated on wether a web view is enabled or not
- We have more than one web view right now and this doesn't make sense

Instead, we make it so that the web views subscribe to the auth state like every other service.

In untangling this, the biggest issue was undocumented dependencies into how events were passed. What I noticed was that the `ContextProvider` plays a major role in initializing the web views. This is something we should clean up in a follow up PR (IMO the web views should subscribe to the ContextProvider and not the other way around). 

I also noticed that the `cody.auth.sync` action isn't really necessary as we can subscribe to the `AuthProvider` via `addChangeListener` anyways (which most of the services did).

Also: Apparently there were two ways on when we reconfigure the `Configuration` object. I now made sure both ways at least reconfigure the same components by extracting a `onConfigurationChange` function inside `main.ts`.

Two unrelated fixes:

- The code to update the upgrade state when the editor was focused would also run on enterprise instances causing a bunch of errors in the console. Fixed.
- Added the remote URL to the completion loggers we can easily see where the request is sent to.

⚠️ This PR has a potential to break areas depending on auth, so please take some time to test it. I've been testing it as thorough as I could though!

## Test plan

The canonical test case was:

- Be signed in to dotcom and have the starcoder-hybrid feature flag enable
- Now switch account to an enterprise instance 
- Ensure that requests are made to the enterprise instance and that we do not use starcoder model strings

https://github.com/sourcegraph/cody/assets/458591/ab5a19a6-a906-42e6-8f76-ae61b500e30c

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -15,7 +15,6 @@ import {
     CURRENT_SITE_IDENTIFICATION,
     CURRENT_SITE_VERSION_QUERY,
     CURRENT_USER_ID_AND_VERIFIED_EMAIL_AND_CODY_PRO_QUERY,
-    CURRENT_USER_ID_AND_VERIFIED_EMAIL_QUERY,
     CURRENT_USER_ID_QUERY,
     EVALUATE_FEATURE_FLAG_QUERY,
     GET_CODY_CONTEXT_QUERY,
@@ -57,10 +56,6 @@ interface SiteHasCodyEnabledResponse {
 
 interface CurrentUserIdResponse {
     currentUser: { id: string } | null
-}
-
-interface CurrentUserIdAndVerifiedEmailQuery {
-    currentUser: { id: string; hasVerifiedEmail: boolean } | null
 }
 
 interface CurrentUserIdHasVerifiedEmailHasCodyProResponse {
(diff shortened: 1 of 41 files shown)


philipp-spiess:
@sourcegraph/cody-vscode Friendly nudge for a PR review 🤗

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -58,6 +59,10 @@ interface CurrentUserIdResponse {
     currentUser: { id: string } | null
 }
 
+interface CurrentUserIdAndVerifiedEmailQuery {
+    currentUser: { id: string; hasVerifiedEmail: boolean } | null
+}
```
DanTup, line 64: This looks like code that was recently removed in https://github.com/sourcegraph/cody/commit/5af57354e2ca4b919f6b82962e7553f78260cda2 - what's the reason for adding it back?

DanTup (approved):
The changes seem reasonable to me, though I'm not very familiar with all of this (nor do I have an enterprise account to test), so you might want additional opinions.

I did add a question about some code that was recently removed that this puts back though.

lib/shared/src/sourcegraph-api/graphql/client.ts
philipp-spiess: I started using the query again in https://github.com/sourcegraph/cody/pull/2182/files#diff-1456bee1a58eb80a929388788427f609653aef419f9c9b84e53c1d904446c516R324-R334 to fix the issue that was querying enterprise instances about the cody pro plan.

I think we can fix this differently though and not do any query on window focus unless connected to dotcom (so we avoid any query).

PR #2183, opened 2023-12-07
VSCode: Send `embeddings/initialize` to the local embeddings controller

dominiccooney:
Rolls cody-engine to v5.2.11713.

cody-engine used to enumerate indices on startup, which slows down engine startup for autocompletes. Instead, enumerate indexes in an `embeddings/initialize` method.

## Test plan

- Set `"cody.debug.enable": true, "cody.debug.verbose": true,`
- Open a repository with App embeddings and chat, check that local embeddings context appears
- Open a repository with local embeddings and chat, check that local embeddings context appears
- Open a repository with no embeddings and click Enable Embeddings and check that indexing happens and then queries work

You can verify when local embeddings is returning results with Cmd-Shift-P, Output: Show Output Channels... Cody by Sourcegraph and looking for messages like `LocalEmbeddingsController: returning 15 results`.

### DIFF
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -77,8 +77,7 @@ export class LocalEmbeddingsController implements LocalEmbeddingsFetcher, Contex
         this.accessToken = token || undefined
         // TODO: Add a "drop token" for sign out
         if (token && this.serviceStarted) {
-            // TODO: Make the cody-engine reply to set-token.
-            void (await this.getService()).request('embeddings/set-token', token)
+            await (await this.getService()).request('embeddings/set-token', token)
         }
     }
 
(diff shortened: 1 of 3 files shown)


vscode/src/local-context/local-embeddings.ts
```diff
@@ -143,17 +142,48 @@ export class LocalEmbeddingsController implements LocalEmbeddingsFetcher, Contex
...
+                break
+            case 'linux':
+                paths = {
+                    indexPath: `${process.env.HOME}/.cache/com.sourcegraph.cody/embeddings`,
+                    appIndexPath: `${process.env.HOME}/.cache/com.sourcegraph.cody/blobstore/buckets/embeddings`,
+                }
+                break
+            case 'win32':
+                paths = {
+                    indexPath: `${process.env.LOCALAPPDATA}/com.sourcegraph.cody/embeddings`,
```
beyang, line 163: Does Windows need `\` path separators?

beyang (approved):
Approving to unblock

vscode/src/local-context/local-embeddings.ts
dominiccooney: Turns out as-is works, but $LOCALAPPDATA uses backslashes, so I changed this to backslashes.

PR #2187, opened 2023-12-07
Update intro chat message to be clear about coding questions

ryphil:
Minor change to the intro text for Cody to clarify that Cody is focused on coding questions, not general world knowledge. 

## Test plan
Copy change only. Tested locally.


toolmantim (approved):

toolmantim:
@ryphil This is a good quick fix, but technically we support whatever questions the chat model you choose is capable of answering (which can be ChatGPT for example). Longer term fix: we should be clearer in the product about the capabilities/differences in models (for choosing/browsing between them, and when you start a chat with one)

PR #2189, opened 2023-12-07
evaluate-autocomplete: more advanced triggering

olafurpg:
Previously, we only triggered autocomplete in limited cases. This PR adds new pluggable infrastructure to trigger in a more variety of language constructs, specifically call expressions, function declarations, if statements, and assignment statements. This ability allows us to measure more accurately real-world usage behaviors.

This PR includes several low-risk but unrelated diffs that improve the quality-of-life while iterating on this feature. For example, the console output is a lot less noisy when running agent tests because Polly no longer complains about missing HTTP recordings for telemetry requests (that previously had random parameters on every test run).

## Test plan

I ran an evaluation of 8 languages to compare scores between baseline and dynamic-multiline. The score is a formula that penalizes autocomplete requests that don't parse or are very small.

```
             FIXTURE  LANGUAGEID      SCORE
0           baseline         cpp  64.033333
1           baseline          go  28.811475
2           baseline  javascript  40.990099
3           baseline         php  39.079365
4           baseline  typescript  46.227106
5  dynamic-multiline         cpp  83.815789
6  dynamic-multiline          go  74.201681
7  dynamic-multiline  javascript  76.683673
8  dynamic-multiline         php  58.024194
9  dynamic-multiline  typescript  78.115942
```

When plotting these scores on a radar chart we see that dynamic-multiline is a huge improvement across the board including languages like PHP.

image


valerybugakov (commented):
Hey @olafurpg, I want/need to use this eval suite for validating ideas around improving autocomplete. I started digging into the codebase at the end of December and watched your recording in Slack, which was very helpful (TY!). Still, I bumped into a couple of places where the implementation has changed, and I didn't find a way to quickly fix the commands I've been running based on your demo. It's understandable since the project moves fast. 

Could you share a set of commands and potentially a local config, if you have one, to produce results shared in the PR description?

pkukielka (approved):
LGTM

Issue #1876, opened 2023-11-24
bug: can't cancel search indexing


toolmantim:
### Version

0.16.0

### Describe the bug

If you open a large repo, symf starts automatically and can't be cancelled, pegging the CPU for an unknown time:

<img width="489" alt="Screenshot 2023-11-24 at 4 55 30 pm" src="https://github.com/sourcegraph/cody/assets/153/058d8bd1-fad4-40ee-8681-4cb4db4508ce">

<img width="1085" alt="Screenshot 2023-11-24 at 4 56 29 pm" src="https://github.com/sourcegraph/cody/assets/153/27c9d0e4-af90-4044-82a0-39a4dd009199">

If you want to stop it, reloading VS Code doesn't help… the only option is to uninstall Cody.

### Expected behavior

There should be a cancel button in the notification

PR #2202, opened 2023-12-08
symf: support cancelling index

beyang:
Fixes https://github.com/sourcegraph/cody/issues/1876

image

## Follow-up tasks
- [ ] Known improvement opportunity: If a user triggers indexing when an index already exists and then cancels the new indexing, the old index is removed. It would be nicer if we could still reuse the old index. Will fix this in a follow-up PR.

## Test plan

- [ ] `rm -rf ~/.cody-symf` to erase the symf index
- [ ] Open the cody repository
- [ ] Do a natural language search to trigger indexing
- [ ] Cancel the search. In the terminal, run `ps aux | grep symf` to verify the process was killed.
- [ ] Verify the index was not built by running `find ~/.cody-symf/`. You should only see contents in the `.tmp` and `.failed` directories.
- [ ] Try doing another search. The index should build this time.
- [ ] Refresh the index using the refresh button in the side panel. Cancel the reindex. Verify the index doesn't exist.

### DIFF
--- a/vscode/src/local-context/symf.ts
+++ b/vscode/src/local-context/symf.ts
@@ -407,7 +407,7 @@ class RWLock {
     public async withRead<T>(fn: () => Promise<T>): Promise<T> {
         while (this.readers === 0) {
             if (this.mu.isLocked()) {
-                // If muis locked at this point, it must be held by the writer.
+                // If mu is locked at this point, it must be held by the writer.
                 // We spin in this case, rather than try to acquire the lock,
                 // because multiple readers blocked on acquiring the lock will
                 // execute serially when the writer releases the lock (whereas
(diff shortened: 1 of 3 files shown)


vscode/src/local-context/symf.ts
```diff
@@ -392,7 +407,7 @@ class RWLock {
     public async withRead<T>(fn: () => Promise<T>): Promise<T> {
         while (this.readers === 0) {
             if (this.mu.isLocked()) {
-                // If mu is locked at this point, it must be held by the writer.
+                // If muis locked at this point, it must be held by the writer.
```
dominiccooney, line 410: Typo here

```suggestion
                // If mu is locked at this point, it must be held by the writer.
```

dominiccooney (approved):
Looks good!

toolmantim:
Nice!

abeatrix (approved):
Let's go!

kalanchan:
GOGOGO nice work!

PR #2207, opened 2023-12-08
Fix symf index dir on Windows

DanTup:
We can't use an absolute path containing a colon as a sub-path, so just remove the colon.

@beyang @jdorfman I tested this manually, but couldn't find a nice way to get a test around this (I tried writing a test that tried to create an index and verify it existed, but it seemed to want to trigger downloads). If there's a reasonable way I could get a test around this, please let me know.

## Before:

Screenshot 2023-12-08 110612

### After:

Screenshot 2023-12-08 110532

image

## Test plan

- Run extension on Windows
- Open a project
- Type a query into the search box
- Ensure results appear and no errors in the console


beyang (approved):

jdorfman:
Great work @DanTup!

Issue #2199, opened 2023-12-08
bug: Files starting with . are missing from the @-mention autocomplete


dominiccooney:
### Version

Main a8d1fb32916d7f08ce26277cbbf0c186d5fb78f8

### Describe the bug

If I want help with a file starting with . like prettier configs, github workflows, etc. I can't mention that filename.

This video also demonstrates #2198 about path separators, feel free to ignore that part in the context of this bug.

https://github.com/sourcegraph/cody/assets/55120/ef72abdc-c918-4a67-8723-bad74443ceea

### Expected behavior

Expected behavior: Be able to mention any file in the workspace, if not everywhere...

Actual behavior: Typing @. shows no results for a workspace with tons of files which start with .

Issue #2208, opened 2023-12-08
bug: file context search searches full paths


DanTup:
### Version

main 18b56f52

### Describe the bug

This search uses the full file paths when filtering, so it often matches way more than you'd expect:

![image](https://github.com/sourcegraph/cody/assets/1078012/46a57a21-c378-4fd1-b3ce-157d7814123c)

Here, "source" is in my path ("Sourcegraph") so it matches all files in the (test fixture) folder.

I found this because some test fail locally searching for "vgo" and my full path apparently contained all three of those characters.

### Expected behavior

The search should filter based on the relative path from the workspace folder that contains a folder.

PR #2209, opened 2023-12-08
Allow dotfiles to be included in context files list

DanTup:
This should fix https://github.com/sourcegraph/cody/issues/2199, although I'm not certain if the behaviour is exactly correct, because:

1. Dot files appear to have been excluded specifically (@toolmantim you added `fileExcludesPattern` in `editor-context` so maybe have an opinion?)
2. Typing `@.` although showing dotfiles at the top, will also basically match every file because of the fuzzy patch

I also hit an issue (filed at https://github.com/sourcegraph/cody/issues/2208) while testing this.. The existing tests can fail depending on the full fs path to your Cody checkout. I think this is a bug but I'd like some confirmation before trying to fix.

## Test plan

- Run `pnpm test:unit` for the updated test
- Open chat, and type `@.` and verify dotfiles show up and can be selected (note: ignore rules currently exclude `.git*` so things like gitattributes won't show up)

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -8,6 +8,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
+- Chat: You can @-mention files starting with a dot. [pull/2209](https://github.com/sourcegraph/cody/pull/2209)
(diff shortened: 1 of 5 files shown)


toolmantim (approved):
The original pattern was there before me, but I like the idea of supporting these use cases, so I say 👍🏼

toolmantim:
Could you also update the changelog? 🙏

PR #2218, opened 2023-12-08
Don't duplicate @-filenames when pressing tab if the whole string has been typed

DanTup:
This code was using `input.lastIndexOf(fileDisplayText)` to find the location of the text to compute the suffix, but it didn't take into account the length, so any time input contained fileDisplayText, the contents of fileDisplayText would become duplicated.

(@abeatrix the original code was added as part of a fix for https://github.com/sourcegraph/cody/issues/1973 - in my testing that still seems to work correctly with my change, but let me know if I've misunderstood).

Fixes https://github.com/sourcegraph/cody/issues/2200

## Test plan

- Run `pnpm test:unit at-file` or
- Open the chat window
- Type "@" followed by a **complete** filename including the very last character
- Press `<tab>` to accept the file
- Ensure the chat input does not contain the filename duplicated


philipp-spiess (approved):
Noice

Issue #2203, opened 2023-12-08
conditionally render PLG flows only for community users


kalanchan:
Currently all users will see PLG flows (usage and upgrade buttons), we need to render these only for community users. 

Requirements:

- [ ] User connected to dotcom (community) see PLG flows
- [ ] User connected to enterprise instances does not see PLG flows (non dotcom)

PR #2219, opened 2023-12-08
Mark Upgrade/Usage links as dot-com only

DanTup:
Fixes https://github.com/sourcegraph/cody/issues/2203

@kalanchan I don't have an enterprise account, are you able to test they're not showing now?

I'm also assuming the issue was specifically for the sidebar, but let me know if not (the other places tied to rate limiting are supposed to appear based on the presence of the Pro header that shouldn't exist for non-dot-com users).

## Test plan

- Login as a dotCom user
- Verify Usage link is visible in sidebar (and if not Pro, the Upgrade link)
- Login as enterprise user
- Verify both links are not visible

### DIFF
--- a/vscode/src/services/TreeViewProvider.test.ts
+++ b/vscode/src/services/TreeViewProvider.test.ts
@@ -1,6 +1,8 @@
 import { describe, expect, it, vi } from 'vitest'
 import type * as vscode from 'vscode'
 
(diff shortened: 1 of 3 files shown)


kalanchan (approved):
looks good, thanks danny!
CleanShot 2023-12-08 at 19 15 24@2x

PR #2233, opened 2023-12-09
Replace "Sign Out" with an account dialog

### LINKED ISSUE #1520: VSCE: “Sign Out” needs a confirmation
“Sign Out” needs a confirmation

@tooltim

> Yeah, that was me removing it before :upside_down_face: I think because it's a treeview item and not a button or menu, and we want people to click around and explore… it makes me a bit more nervous than usual. A standard VS Code confirmation dialog (rather than the quickpick) would be fine though!

toolmantim:
Replaces the "Sign Out" option with a new Account dialog that gives you visibility of what user you're signed in as, which server, and what plan.

| State | Screenshot |
| - | - |
| Cody Pro FF Not Enabled | Screenshot 2023-12-11 at 8 39 59 pm |
| Cody Pro FF Enabled | Screenshot 2023-12-11 at 8 40 22 pm |
| Enterprise Instance | Screenshot 2023-12-11 at 8 41 16 pm |

Notes:

- Display name is often missing in Enterprise accounts, so we just use the email

Fixes #1520

## Test plan

- Signed into DotCom
  - Clicked account, switch, sign out, etc
  - Changed plan, return, clicked account, verified plan has changed
  - Turn off CodyPro feature flag, clicked account, verify no manage or plan
- Signed in to S2 (Enterprise), clicked account, switch, sign out, etc

### DIFF
--- a/vscode/src/chat/utils.test.ts
+++ b/vscode/src/chat/utils.test.ts
@@ -46,7 +46,7 @@ describe('validateAuthStatus', () => {
             endpoint,
             avatarURL,
             primaryEmail,
-            displayName
+            displayName,
         }
         expect(
             newAuthStatus(
@@ -74,7 +74,7 @@ describe('validateAuthStatus', () => {
             endpoint,
             avatarURL,
             primaryEmail,
-            displayName
+            displayName,
         }
         expect(
             newAuthStatus(
@@ -102,7 +102,7 @@ describe('validateAuthStatus', () => {
             endpoint,
             avatarURL,
             primaryEmail,
-            displayName
+            displayName,
         }
         expect(
             newAuthStatus(
@@ -146,7 +146,7 @@ describe('validateAuthStatus', () => {
             endpoint,
             avatarURL,
             primaryEmail,
-            displayName
+            displayName,
         }
         expect(
             newAuthStatus(
--- a/vscode/src/chat/utils.test.ts
+++ b/vscode/src/chat/utils.test.ts
@@ -12,9 +12,9 @@ describe('validateAuthStatus', () => {
     const validUser = true
     const endpoint = ''
     const userCanUpgrade = false
-    const primaryEmail = 'email@domain.com'
-    const displayName = 'Hello There'
-    const avatarURL = 'https://example.com/avatar.png'
+    const primaryEmail = 'me@domain.test'
+    const displayName = 'Test Name'
+    const avatarURL = 'https://domain.test/avatar.png'
     // DOTCOM AND APP USERS
     test('returns auth state for invalid user on dotcom or app instance', () => {
         const expected = { ...unauthenticatedStatus, endpoint }
--- a/vscode/src/chat/utils.test.ts
+++ b/vscode/src/chat/utils.test.ts
@@ -10,13 +10,27 @@ describe('validateAuthStatus', () => {
     const verifiedEmail = true
     const codyEnabled = true
     const validUser = true
-    const endpoint = 'https://example.com'
+    const endpoint = ''
     const userCanUpgrade = false
+    const primaryEmail = 'email@domain.com'
+    const displayName = 'Hello There'
+    const avatarURL = 'https://example.com/avatar.png'
     // DOTCOM AND APP USERS
     test('returns auth state for invalid user on dotcom or app instance', () => {
         const expected = { ...unauthenticatedStatus, endpoint }
         expect(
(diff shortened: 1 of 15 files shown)


vscode/src/services/AuthProvider.ts
```diff
@@ -130,6 +131,51 @@ export class AuthProvider {
         }
     }
 
+    public async accountMenu(): Promise<void> {
+        if (!this.authStatus.authenticated || !this.authStatus.endpoint) {
+            return
+        }
+
+        if (!isDotCom(this.authStatus.endpoint)) {
```
kalanchan, line 139: wooh never forget our enterprise customers!

toolmantim: Yep, they get a nice dialog too.

toolmantim: Or did I miss something?

kalanchan: haha meant it as a statement not a question 😄

vscode/src/services/AuthProvider.ts
```diff
@@ -130,6 +131,55 @@ export class AuthProvider {
...
+                    break
+                case 'Sign Out':
+                    await this.signoutMenu()
+                    break
+            }
+            return
+        }
+
+        const option = await vscode.window.showInformationMessage(
+            `${this.authStatus.displayName}\n${this.authStatus.primaryEmail}`,
```
kalanchan, line 161: whats the reasoning behind not having the "signed in as ...." prefix for community?

toolmantim: At the moment it's based on what data I can get from dot com and Enterprise, as they have different code paths and GraphQL queries. But they'll probably converge on the same format by the time I'm finished with this PR…

vscode/src/chat/utils.test.ts
```diff
@@ -10,13 +10,27 @@ describe('validateAuthStatus', () => {
     const verifiedEmail = true
     const codyEnabled = true
     const validUser = true
-    const endpoint = 'https://example.com'
+    const endpoint = ''
     const userCanUpgrade = false
+    const primaryEmail = 'email@domain.com'
```
dominiccooney, line 15: Let's follow RFC2606 and use a test or example domain:

https://www.rfc-editor.org/rfc/rfc2606.html#page-2

vscode/src/services/AuthProvider.ts
```diff
@@ -130,6 +134,59 @@ export class AuthProvider {
...
+
+        const option = await vscode.window.showInformationMessage(
+            `Signed in as ${
+                this.authStatus.displayName
+                    ? `${this.authStatus.displayName} (${this.authStatus.primaryEmail})`
+                    : this.authStatus.primaryEmail
+            }`,
+            {
+                modal: true,
+                detail: `Plan: ${this.authStatus.userCanUpgrade ? 'Cody Pro' : 'Cody Free'}`,
```
dominiccooney, line 171: The field naming or the message is wrong. If they *can* upgrade, then they're not on Cody Free, not Cody Pro.

vscode/src/services/TreeViewProvider.test.ts
```diff
@@ -18,6 +18,9 @@ describe('TreeViewProvider', () => {
     const verifiedEmail = true
     const codyEnabled = true
     const validUser = true
+    const primaryEmail = 'email@domain.com'
```
dominiccooney, line 21: Ditto here, let's use example.net, host.example, or whatever

dominiccooney:
I think the agent tests just needed a replay archive update, I have pushed one to your branch.

dominiccooney (approved):
Approving but double check https://github.com/sourcegraph/cody/pull/2233#discussion_r1421994861

vscode/src/services/AuthProvider.ts
toolmantim: Ah, good catch. I had become blind to these labels. We're really missing an `AccountType`… we have littered this `userCanUpgrade` logic in some many places now.

toolmantim:
> If you have time and are so inclined, you can now e2e test the isDotCom/or not paths in e2e cribbing from here: https://github.com/sourcegraph/cody/blob/main/vscode/test/e2e/local-embeddings.test.ts#L18-L20

@dominiccooney this is awesome! I started on an e2e test for this, but couldn't get Playwright to work nicely with the dialogs :(

PR #2235, opened 2023-12-09
VSCode: Fix local embeddings status for non-git workspaces, git repos w/out remotes

### LINKED ISSUE #2217: bug: Local embeddings spins "indexed" forever in git repos without remotes
### Version

Main fc522b73c44072a480e0845dda308c4f2624f904

### Describe the bug

```
mkdir /tmp/foo
cd /tmp/foo
git init .
code .
```

Open a chat, open the enhanced context selector. Says it is indexed, spins endlessly.

Screenshot 2023-12-08 at 23 57 54

### Expected behavior

@toolmantim what should we display here? The "indexed" label and the indeterminate spinner looks silly. Indexed is wrong, it is not indexed. There's a case statement that falls through from "indeterminate", we should change it up so there's no "indexed" label in the indeterminate state.

### Additional context

_No response_

dominiccooney:
- When opening a non-git repo, instead of the enhanced context selector showing a non-functional "enable embeddings" button, do not show local embeddings status.
- When opening a git repo without a remote, instead of showing a spinner indefinitely and "Indexed", show the circle-slash. Fixes #2217.
- During embeddings index generation, show the current file being indexed in the status bar item tooltip.
- If embeddings index generation fails, show the error message in a tooltip on the status bar. (Test coverage and improvements planned in a follow-up change.)

Adds e2e tests for the major cases above.

Also improves the e2e test framework:
- Fixes a race condition creating the workspace `.vscode` directory, and writing `.vscode/settings.json`. This sometimes meant the tests were running without the correct settings.
- Adds Playwright extensions so tests can configure an isolated working directory, extra VSCode settings, and treat the test server endpoint as "dotcom".
- Expands the mock server so that the consumer login path works.
- Adds testing hooks to local embeddings to use a mock server, embeddings model, and isolated index library path.
- Deletes obsolete helper for submitting sidebar chat messages.

## Test plan

This integration test covers the scenarios of opening no git repo, a git repo without a remote, and a git repo *with* a remote and then generating embeddings for the repo and querying them:

```
pnpm test:e2e -- local-embeddings.test.ts
```

### DIFF
--- a/lib/shared/src/codebase-context/context-status.ts
+++ b/lib/shared/src/codebase-context/context-status.ts
@@ -38,7 +38,7 @@ interface RemoteEmbeddingsProvider {
 export interface LocalEmbeddingsProvider {
     kind: 'embeddings'
     type: 'local'
-    state: 'indeterminate' | 'unconsented' | 'indexing' | 'ready'
+    state: 'indeterminate' | 'no-match' | 'unconsented' | 'indexing' | 'ready'
 }
 
 interface SearchProvider {
--- a/lib/shared/src/codebase-context/context-status.ts
+++ b/lib/shared/src/codebase-context/context-status.ts
@@ -39,6 +39,7 @@ export interface LocalEmbeddingsProvider {
     kind: 'embeddings'
     type: 'local'
     state: 'indeterminate' | 'no-match' | 'unconsented' | 'indexing' | 'ready'
+    errorReason?: 'not-a-git-repo' | 'git-repo-has-no-remote'
 }
 
 interface SearchProvider {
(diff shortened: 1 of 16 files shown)


toolmantim (approved):
Verified that this all works as expected 👍🏼👏🏼 Two non-blocking notes…

When there's no origin:

Screenshot 2023-12-11 at 12 47 12 pm

it would be nicer to also show an explanation, similar to [this state](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=213-180250&mode=design&t=mVBSppSEvvN164CK-4):

image

With the copy:

> Git repository is missing a remote origin.

I would have assumed a similar UI if you open a folder that isn't a Git repo (with the copy "Folder is not a Git repository"), but instead I see the following:

Screenshot 2023-12-11 at 12 51 59 pm

dominiccooney:
Thanks for the feedback! Love the wording of the error messages. I wrestled with it briefly but nothing sounded right.

PR #2240, opened 2023-12-10
Fix pre-release version numbers not being correctly detected

toolmantim:
Follow on from #2210, this fixes pre-release version not being shown correctly in the sidebar. This was because the version value is baked at build time, before pre-release bumps the package.json.

This updates the version detection to use the vscode extensions API (borrowed & moved from Telemetry, which was already doing this correctly).

The update notice in the web UI still uses the static version, as it always has. We can update that to handle pre-release/insider builds better in the future.

## Test plan

- Ran the extension with different versions, verified nav change
- Modified compiled source and reloaded the extension, verified nav change

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -8,6 +8,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


philipp-spiess (approved):
Very nice!

PR #2215, opened 2023-12-08
Handle context file searching using forward slashes on Windows

DanTup:
Fixes https://github.com/sourcegraph/cody/issues/2198

## Test plan

- Run `pnpm test:e2e at-file` to run updated test, or:
- On Windows, open chat and try to attach a file in chat using backslashes and forward slashes for searching filenames

PR #2241, opened 2023-12-10
Only search relative paths when @-mentioning files in chat

DanTup:
[This depends on a function added in https://github.com/sourcegraph/cody/pull/2215 so this will fail bots until that lands and I rebase/merge]

Fixes https://github.com/sourcegraph/cody/issues/2208

## Test plan

- Run `pnpm test:e2e at-file` to run the updated test, or
- Open chat, and mention a file using `@` but type part of the file path outside of the folder opened in VS Code (eg. if you have `C:\Source\cody\ open, type `source`)
- Ensure that the results don't list every file because `source` exists in their full fspaths

### DIFF
--- a/vscode/test/e2e/at-file-context-selecting.test.ts
+++ b/vscode/test/e2e/at-file-context-selecting.test.ts
@@ -71,7 +71,8 @@ test('@-file empty state', async ({ page, sidebar }) => {
     await chatInput.press('Enter')
     await expect(chatInput).toHaveValue(withPlatformSlashes('Explain @lib/batches/env/var.go '))
     await chatInput.type('and @vgo', { delay: 50 }) // without this delay the following Enter submits the form instead of selecting
-    await chatInput.press('ArrowDown') // second item
+    await chatInput.press('ArrowDown') // second item (visualize.go)
+    await chatInput.press('ArrowDown') // third item (.vscode/settings.json)
     await chatInput.press('ArrowDown') // wraps back to first item
     await chatInput.press('ArrowDown') // second item again
     await chatInput.press('Enter')
--- a/vscode/test/e2e/at-file-context-selecting.test.ts
+++ b/vscode/test/e2e/at-file-context-selecting.test.ts
@@ -30,6 +30,9 @@ test('@-file empty state', async ({ page, sidebar }) => {
 
     // We should only match the relative visible path, not parts of the full path outside of the workspace.
     // Eg. searching for "source" should not find all files if the project is inside `C:\Source`.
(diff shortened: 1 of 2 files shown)


vscode/test/e2e/at-file-context-selecting.test.ts
```diff
@@ -22,6 +22,15 @@ test('@-file empty state', async ({ page, sidebar }) => {
     await chatInput.fill('@definitelydoesntexist')
     await expect(chatPanelFrame.getByRole('heading', { name: 'No matching files found' })).toBeVisible()
 
+    // Clear the input so the next test doesn't detect the same text already visible from the previous
+    // check (otherwise the test can pass even without the filter working).
+    await chatInput.clear()
+
+    // We should only match the relative visible path, not parts of the full path outside of the workspace.
+    // Eg. searching for "source" should not find all files if the project is inside `C:\Source`.
+    await chatInput.fill('@fixtures') // fixture is in the test project folder name, but in the relative paths.
```
dominiccooney, line 31: Once #2235 lands, you can make this more robust by adding `workspaceDirectory` alongside `page, sidebar` and get the path to the workspace directory. That will let you assert that fixtures really *is* in that path and remains that way.

dominiccooney (approved):
Looks good!

vscode/test/e2e/at-file-context-selecting.test.ts
DanTup: Oh yeah, good point. I've added a TODO about that. If you're planning to land that soon I can wait and fix it up here, or if you think I should land this as-is and fix later, let me know.

I've rebased this on main after merging #2208 so I think this should all be green now.

DanTup:
Hmm, there's something odd going on now. With this change, `.vscode\settings.json\` shows up in the search results for `@vgo` (it has no `g`!), which results in the test selecting the wrong item - but that also results in `and @vg` being duplicated on the end:

https://github.com/sourcegraph/cody/assets/1078012/3c833a18-cfa1-4776-867b-59fbad9610bf

This is reliably reproducible for me, and feels like two bugs (non-matched items in the results, and the duplicated text). I might not get chance to look into this before Friday now, so if this is important and anyone else has time, feel free to take a look!

DanTup:
> it has no `g`!

lol, it very much as a `g` 😄 I'll update the test and file another bug about the duplication (which seems to occur only when you select a file that's already listed).

DanTup:
I've filed https://github.com/sourcegraph/cody/issues/2243 about the bug.

@dominiccooney heading off, but if you're happy with this as-is and the last bots go green feel free to merge :)

PR #2250, opened 2023-12-11
Implement history quick pick

### LINKED ISSUE #1692: VSCE: New Chat History view in quick pick
From Tim, a new chat history view in quick pick format

>  quickpick… another chat history view, in addition to the treeview, to really keep you in the flow and be more keyboard navigable.

deepak2431:
This PR closes #1692 

- Implemented algorithm to group chats by last interaction timestamp.
- Added an icon in the editor panel to display history chats as a quick pick, separated by the group.
- Updated the history chats in the Treeview provider to render them as grouped chats.
- Remove createCodyChatTreeItems, use the updateTree method of TreeViewProvider to update the list instead.

## Test plan

https://github.com/sourcegraph/cody/assets/44617923/5804d8da-a629-4670-8de9-ff3b3da61e25


deepak2431:
@toolmantim @philipp-spiess @abeatrix Can I have a review please?

deepak2431:
@toolmantim @philipp-spiess Tested with the updated rename chat feature, too, it works fine. All good to test for now!

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+import { getChatPanelTitle } from '../chat/chat-view/chat-helpers'
+import { chatHistory } from '../chat/chat-view/ChatHistoryManager'
+
+import { CodySidebarTreeItem } from './treeViewItems'
+
+interface GroupedChats {
+    [groupName: string]: CodySidebarTreeItem[]
+}
+
+interface HistoryItems {
```
dominiccooney, line 12: This be history *item*, singular.
```suggestion
interface HistoryItem {
```

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+        if (lastHumanMessage?.humanMessage.displayText && lastHumanMessage?.humanMessage.text) {
+            const lastDisplayText = lastHumanMessage.humanMessage.displayText.split('\n')[0]
+            const chatTitle = chats[id].chatTitle || getChatPanelTitle(lastDisplayText)
+
+            const currentTimeStamp = new Date()
+            const lastInteractionTimestamp = new Date(entry.lastInteractionTimestamp)
+
+            const timeDiff = currentTimeStamp.getTime() - lastInteractionTimestamp.getTime()
+
+            if (timeDiff < 24 * 60 * 60 * 1000) {
```
dominiccooney, line 43: You should de-duplicate this repetitive code with a list of `[[1, todayChats], [2, yesterdayChats], [7, nDaysChats], ...]` and loop until you find the first day limit you don't exceed, and add the item to that chat.

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+                })
+            } else if (timeDiff < 7 * 24 * 60 * 60 * 1000) {
+                NDaysChats.push({
+                    id,
+                    title: chatTitle,
+                    icon: 'comment-discussion',
+                    command: { command: 'cody.chat.panel.restore', args: [id, chatTitle] },
+                })
+            } else if (timeDiff < 14 * 24 * 60 * 60 * 1000) {
+                lastWeekChats.push({
```
dominiccooney, line 65: The way this is, on March 2 I will have some January chats in `NWeeksChats` and some in `lastMonthChats` which doesn't sound right.

@toolmantim can be the definitive call on this, but I think "last month" means something objectively different to "30-60 days ago." "last week" gets slippery with locale-specific first day of the week, but same issue there.

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+export function groupCodyChats(): GroupedChats | null {
+    const todayChats: CodySidebarTreeItem[] = []
+    const yesterdayChats: CodySidebarTreeItem[] = []
+    const lastWeekChats: CodySidebarTreeItem[] = []
+    const lastMonthChats: CodySidebarTreeItem[] = []
+    const NDaysChats: CodySidebarTreeItem[] = []
+    const NWeeksChats: CodySidebarTreeItem[] = []
+    const NMonthsChats: CodySidebarTreeItem[] = []
```
dominiccooney, line 25: Camel case, so `nDaysChats` etc.

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+                })
+            }
+        }
+    })
+
+    return {
+        Today: todayChats.reverse(),
+        Yesterday: yesterdayChats.reverse(),
+        'Last Week': lastWeekChats.reverse(),
+        'Last month': lastMonthChats.reverse(),
```
dominiccooney, line 100: The capitalization of week/month is inconsistent here.

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+    const chats = chatHistory.localHistory?.chat
+    if (!chats) {
+        return null
+    }
+    const chatHistoryEntries = [...Object.entries(chats)]
+    chatHistoryEntries.forEach(([id, entry]) => {
+        const lastHumanMessage = entry?.interactions?.findLast(interaction => interaction?.humanMessage)
+        if (lastHumanMessage?.humanMessage.displayText && lastHumanMessage?.humanMessage.text) {
+            const lastDisplayText = lastHumanMessage.humanMessage.displayText.split('\n')[0]
+            const chatTitle = chats[id].chatTitle || getChatPanelTitle(lastDisplayText)
```
dominiccooney, line 36: Would this ever change chat titles? For the first couple of messages it might be fine. But it could be annoying if we change titles on people after the chats have cooled off.

Not related to this PR, but by the way, it would be cool if we could use the LLM to generate good titles for the chat.

vscode/src/services/HistoryChat.ts
```diff
@@ -0,0 +1,145 @@
...
+        quickPickItems.push({
+            label: groupName,
+            onSelect: async () => {},
+            kind: vscode.QuickPickItemKind.Separator,
+        })
+    }
+
+    for (const [groupName, chats] of Object.entries(groupedChats)) {
+        if (chats.length > 0) {
+            addGroupSeparator(groupName)
```
dominiccooney, line 125: Does this mean we start with a separator?
```suggestion
        if (chats.length > 0) {
            if (quickPickItems.length) {
              addGroupSeparator(groupName)
            }
```

or do it in addGroupSeparator, or use a `flatMap`, `join` to remove the blanks and add the separators.

vscode/src/services/TreeViewProvider.ts
```diff
@@ -5,8 +5,43 @@ import { isDotCom } from '@sourcegraph/cody-shared/src/sourcegraph-api/environme
...
+export class ChatTreeItem extends vscode.TreeItem {
+    public children: ChatTreeItem[] | undefined
+
+    constructor(
+        public readonly id: string,
+        title: string,
+        icon?: string,
+        command?: {
+            command: string
+            args?: string[] | { [key: string]: string }[]
```
dominiccooney, line 20: You could probably do some TypeScript acrobatics here to get the parameter list of execCommand to pop out here... `Parameters<...>`

dominiccooney (commented):
Looks really cool. Some feedback inline.

vscode/src/services/HistoryChat.ts
toolmantim: The logic I outlined was here: https://github.com/sourcegraph/cody/issues/1692 which was somewhat engineering from ChatGPT’s groupings, which I liked. I see your point though!

vscode/src/services/HistoryChat.ts
deepak2431: This chatTitle only gets changed when they manually edit it, as we have the functionality after the merge of #1818. 
I liked your idea of generating better chat titles using LLM rather than just using the last message interaction. I will open an issue for this and look into it.

vscode/src/services/HistoryChat.ts
deepak2431: Yeah, makes sense. Would be good to do it this way, as just the multiplying factor is being updated in each of the cases.

vscode/src/services/HistoryChat.ts
deepak2431: It would be good if we don't consider `NWeeksChats`. After `lastWeek`, which would be 7-14 days, after that, a month of 30 days, and then `NMonths`. WDYT @dominiccooney , @toolmantim ?

vscode/src/services/HistoryChat.ts
deepak2431: Yes, we start with a separator. `Today` would be the first separator. If we add that condition to check for the `quickPick` items length, the separator `today` won't be there so I think it is better to keep it this way.

vscode/src/services/TreeViewProvider.ts
deepak2431: Ah, I didn't get this. Do you mean to extend the VSCode command to create a type for this command?

deepak2431:
@dominiccooney Thanks for the review. I have made all the changes as per the review!

toolmantim (commented):
This is looking really great!

There's one blocker bug I found, and three related improvement requests… and then we're good to go IMO.

Bug: folders are incorrectly showing rename and edit actions:

Screenshot 2023-12-14 at 1 10 23 pm

Improvement: The folders often being collapsed creates some friction. Could we make it so:
- On load, the first folder is always expanded
- On load, if any chats are revived, we ensure their folder is open
- When we create the first chat for today, and the "Today" group appears, can it default to being open

The history quickpick itself works great! No changes needed there.

Thanks so much @deepak2431 — great work!

deepak2431:
@toolmantim Thanks a lot for the detailed review and comments. I will soon make those changes :)

deepak2431:
@toolmantim I have fixed the bug you mentioned and made all the suggested changes except one to expand the group when the chat is revealed. I have implemented one solution for this, but it does not expand the tree when rebuilding it. I also tried a few other workarounds, but it didn't work. 

@philipp-spiess I would appreciate some help on this: The issue is when the chat is being revealed, I am updating the collapsible state of the tree when calling the `updateTree` method, but it is not expanding the tree. Even though I have attached an event listener to it, it still doesn't work. 


https://github.com/sourcegraph/cody/assets/44617923/dabd5e28-70ef-4345-b5a5-16900e94173b

vscode/src/services/TreeViewProvider.ts
dominiccooney: No, I meant you can use TypeScript to pick out and refer to the type you're passing through here. But it's not important to do it.  May just save some maintenance time down the road.

vscode/src/services/HistoryChat.ts
dominiccooney: The UX mock in #1692 put these as tertiary context on the right hand side, but now we're using them as primary groups, which feels like it demands more precision.

I think you should do this:

```
const dateEqual(d1, d2) {
  return d1.getDate() == d2.getDate() && monthYearEqual(d1, d2)
}
const monthYearEqual(d1, d2) {
  return d1.getMonth() == d2.getMonth() && d1.getYear() == d2.getYear()
}

const today = new Date()

const yesterday = new Date()
yesterday.setDate(yesterday.getDate() - 1)

const lastMonth = new Date()
lastMonth.setDate(1)
lastMonth.setMonth(lastMonth.getMonth() - 1)

dateEqual(today, theDate) => 'Today'
dateEqual(yesterday, theDate) => 'Yesterday'
monthYearEqual(today, theDate) => 'This month'
monthYearEqual(lastMonth, theDate) => 'Last month'
otherwise => 'Months ago'
```

or something like that. TL;DR, if you avoid talking about weeks, you can do sensible date arithmetic. If we have `Intl.Locale` then you could do week stuff.

vscode/src/services/TreeViewProvider.ts
```diff
@@ -98,8 +187,16 @@ export class TreeViewProvider implements vscode.TreeDataProvider<vscode.TreeItem
...
-        return [...this.treeNodes.values()]
+    public async getChildren(element?: ChatTreeItem): Promise<ChatTreeItem[]> {
+        if (element) {
+            // Load children if not already loaded
+            if (!element.children) {
+                await element.loadChildNodes()
+            }
+            return element.children || []
+        }
+        // Cast treeNodes to ChatTreeItem[]
```
dominiccooney, line 198: We don't need this comment, it is evidently a cast.

vscode/src/services/TreeViewProvider.ts
```diff
@@ -98,8 +187,16 @@ export class TreeViewProvider implements vscode.TreeDataProvider<vscode.TreeItem
...
+        return this.treeNodes as ChatTreeItem[]
```
dominiccooney, line 199: If this cast is reliable, let's make the type of `this.treeNodes: ChatTreeItem[]` and remove the cast. That will guide anyone who later adds a non-ChatTreeItem to `this.treeNodes` away from misadventure.

dominiccooney (changes requested):
Bit more feedback inline. I think we should grub with the dates and return groups which reflect the calendar and not timespans labelled as days/weeks/months.

vscode/src/services/HistoryChat.ts
deepak2431: @dominiccooney This seems reasonable to me, that we can do date arithmetic directly over the dates. Also, as we already have `this month` that shall cover both the `NWeeks` and the `LastWeeks` too so we don't need to create another group for the weeks. Let me update the code accordingly.

vscode/src/services/TreeViewProvider.ts
deepak2431: Got it now!

deepak2431:
@dominiccooney I have made the requested changes to calculate the groups using the date calculation. 

Also, would you mind taking a quick look at the abovementioned issue related to the collapsed state when chats are revived and I recreate the items?
> This one: @philipp-spiess I would appreciate some help on this: The issue is when the chat is being revealed, I am updating the collapsible state of the tree when calling the updateTree method, but it is not expanding the tree. Even though I have attached an event listener to it, it still doesn't work.

deepak2431:
@toolmantim @dominiccooney This is all good to go now, after the recent change of using authStatus to load chats this PR was having issue. I have taken care of it now, and it works all good as expected. Can you please add your final review for this?

toolmantim (approved):
Design wise this works great 👍🏼

dominiccooney (approved):
Code looks good to me.

deepak2431:
@toolmantim @dominiccooney Thanks for the review, I will just resolve the conflicts now and it should be good to go!

toolmantim:
@deepak2431 I hit approve on the Actions runs, and it looks like there's a few other things that need fixing due to changes

deepak2431:
@toolmantim I see, do let me know I can make those changes.

toolmantim:
@deepak2431 apologies, I thought you could view them! `Error: src/services/TreeViewProvider.test.ts(72,53): error TS2339: Property 'label' does not exist on type 'Uri'.` on both `pnpm test:unit` and `pnpm test:e2e`

deepak2431:
@toolmantim No problem; I thought there might be some UI changes due to the coming updated Chat UI. 
All the unit tests are passing for me successfully:
iScreen Shoter - Terminal - 240123125209

Looks like the test being flaky, but yeah let me check it again on my side!

deepak2431:
@toolmantim Fixed the test issue, all tests are passing without any issue now!

toolmantim:
@deepak2431 thank you! Tests look green but `pnpm exec biome ci --error-on-warnings .` seems to be exiting with an error. Can you check that locally too? If we get it fixed today we can get it into this week’s release!

deepak2431:
@toolmantim Fixed the biome lint errors :)

toolmantim:
Finally got this one merged @deepak2431! Appreciate all of your work that went into this, and I'm sure users will too 🙏

deepak2431:
Thanks, @toolmantim. Finally will get this shipped into this week's release!

PR #2255, opened 2023-12-11
Chat: Add support for cody.chat.preInstruction in the new preamble

### LINKED ISSUE #2165: bug: pre-instruction config not working in simplified chat
pre-instruction config does not work in simplified chat

toolmantim:
This adds support for the `cody.chat.preInstruction` setting in simplified chat, especially important now we have a "Chat Settings" button in the top of the chat editor:

Screenshot 2023-12-11 at 11 28 07 pm

Screenshot 2023-12-11 at 11 28 13 pm

- [x] Get it working
- [x] Tests

Fix #2165

## Test plan

- Edit the setting
- Type a message

### DIFF
--- a/vscode/src/chat/chat-view/prompt.test.ts
+++ b/vscode/src/chat/chat-view/prompt.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it, afterEach, vi } from 'vitest'
+import { SimpleChatModel } from './SimpleChatModel'
+import { DefaultPrompter } from './prompt'
+import * as vscode from 'vscode'
+
+describe('DefaultPrompter', () => {
+    afterEach(() => {
+        vi.restoreAllMocks()
+    })
+
+    it('constructs a prompt with no context', async () => {
+        const chat = new SimpleChatModel('a-model-id')
+        chat.addHumanMessage({ text: 'Hello' })
+
+        const { prompt, warnings, newContextUsed } = await new DefaultPrompter().makePrompt(chat, {
+            getExplicitContext: () => [],
+            getEnhancedContext: () => Promise.resolve([])
+        }, true, 100000)
+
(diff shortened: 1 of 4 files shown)


toolmantim:
@beyang @dominiccooney @abeatrix before I go adding tests… does this approach look sound, or would you prefer it another way?

dominiccooney (approved):
This looks good to me.

We have that "typed" configuration layer, tests for defaults, etc. but because this is simply "any string" I don't think we need to plumb through that here? @philipp-spiess do you have thoughts?

toolmantim:
@dominiccooney I've just added some tests, if you could quickly give it a re-review

vscode/src/chat/chat-view/prompt.test.ts
```diff
@@ -0,0 +1,87 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
```
toolmantim, line 1: Would have been nice to have an e2e test for this, but our e2e tests aren't very end-to-end — they use a completely mocked server than doesn't go through our prompt construction gear. I glanced at the integration tests, but they seem to be all marked as skipped because the `getTranscript` function is broken atm. So a unit test it is!

valerybugakov (approved):
LGTM! Agree that there's no need to pass `cody.chat.preInstruction` through our typed `getConfiguration` wrapper.

vscode/src/chat/chat-view/prompt.test.ts
dominiccooney: Wow, I am surprised that the e2e tests somehow avoid prompt construction.

Unit test seems perfect for this. We will be waiting an hour for test results if we have e2e tests for everything.

dominiccooney (approved):
Looks great. Thank you for adding a test.

PR #2256, opened 2023-12-11
Autocomplete: Remove syntacticPostProcessing flag

philipp-spiess:
I think we're pretty settled on the decision to ship this, so let's rm the flag.

The presence of the flag actually caused tree sitter to _not_ work on Agent even though the WASM implementation should work (cc @olafurpg). 

## Test plan

- The flag was already set to be enabled by default, this just removes the flag.

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -166,9 +166,6 @@ const configuration: vscode.WorkspaceConfiguration = {
                 return connectionConfig?.debug ?? false
             case 'cody.debug.verbose':
                 return connectionConfig?.verboseDebug ?? false
-            case 'cody.autocomplete.experimental.syntacticPostProcessing':
-                // False because we don't embed WASM with the agent yet.
-                return false
             case 'cody.codebase':
                 return connectionConfig?.codebase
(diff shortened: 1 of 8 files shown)


philipp-spiess:
Putting this on hold. I misunderstood and we need to add the WASM files in the pkg single binary via https://github.com/vercel/pkg#config

olafurpg (approved):
LGTM 👍🏻 Thank you for spotting the root cause

olafurpg:
I will fix the CI and update the PR to make sure the wasm files are included with the agent.

agent/src/vscode-shim.ts
```diff
@@ -166,9 +166,6 @@ const configuration: vscode.WorkspaceConfiguration = {
                 return connectionConfig?.debug ?? false
             case 'cody.debug.verbose':
                 return connectionConfig?.verboseDebug ?? false
-            case 'cody.autocomplete.experimental.syntacticPostProcessing':
-                // False because we don't embed WASM with the agent yet.
-                return false
```
philipp-spiess, line 171: Note: We can only drop the flag when the Agent defaults to `true`.

PR #2265, opened 2023-12-11
VSCode: Make embeddings errors visible, resume indexing on restart

dominiccooney:
- If an incomplete index is loaded, explain the situation in the status bar.
- User can resume an incomplete index.
- Displays error messages from local embeddings in the UI.
- Updates the local embeddings protocol to 0.1.2. The client now learns the "names" (identifying keys) of repositories it is loading and indexing, so events can be related to the repositories they belong to. New methods to query the status of indexes and resume indexing.

https://www.loom.com/share/242be2c7abcd49d19e1709e0529651b1
UX refresh: https://www.loom.com/share/2654feb99c9f4985add94470a362bec7
 
## Test plan

Manual test:

1. Open a repository which doesn't have any embeddings.
2. Start generating local embeddings.
3. In Activity Monitor, kill the cody-engine process, leaving a partial index. Note, there's no handling process death, so the indexing process will appear to simply halt.
4. Reload the VScode window to work around the dead process.
5. Confirm there's a status bar entry indicating the index is incomplete.
6. Start a chat. Queries should use the incomplete index.
7. Interact with the status bar to resume indexing.
8. When indexing is done, queries should use the completed index.

To generate an unusual failure mode, edit another git repo's remote to match the partial index and open that folder. Then resume indexing. This should display an error message about the repo not having the specific commit the index was created at. (This resumes an incomplete index, it does not update an index to a new commit.)

Automated test:

`pnpm test:e2e -- local-embeddings.test.ts`

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -349,11 +349,6 @@
         "when": "cody.activated"
       },
       {
-        "command": "cody.status-bar.embeddings",
-        "title": "Cody Embeddings",
-        "when": "cody.embeddings.notice"
-      },
-      {
         "command": "cody.show-page",
         "category": "Cody",
         "title": "Open Account Page",
@@ -500,6 +495,11 @@
         "group": "Cody",
         "icon": "$(clear-all)",
         "when": "cody.activated && cody.hasChatHistory"
+      },
+      {
+        "command": "cody.embeddings.resolveIssue",
+        "title": "Cody Embeddings",
+        "when": "cody.embeddings.hasIssue"
       }
     ],
     "keybindings": [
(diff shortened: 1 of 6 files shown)


vscode/package.json
```diff
@@ -495,6 +495,11 @@
         "group": "Cody",
         "icon": "$(clear-all)",
         "when": "cody.activated && cody.hasChatHistory"
+      },
+      {
+        "command": "cody.embeddings.resolveIssue",
+        "title": "Cody Embeddings",
```
toolmantim, line 501: Did you intend on this appearing in the command palette when there's an issue? If so, maybe we set a nicer title? If not, we can set to `"when": false` in the "commandPalette" section.

toolmantim (approved):
Code seems legit & doesn't break setting up new embeddings. I did not manually test the incomplete embeddings scenario.

PR #2273, opened 2023-12-11
telemetry-v2: fix splitSafeMetadata to generate acceptable type

bobheadxi:
The current returned type is too "loose" to be accepted directly by `telemetryRecorder.recordEvent`. This change makes it stricter (but still correct based on the intended behaviour of the helper) with some type hackery, courtesy of StackOverflow.

Previously, this kind of worked by accident because we "broadened" the type beforehand.

## Test plan

New and existing unit tests.


nathan-downs (approved):
LGTM!

PR #2286, opened 2023-12-12
Settings: Remove the "Simple Chat Context" setting

toolmantim:
Removes the following "Simple Chat Context" setting from the settings quickpick, as it'll only be used internally or for specific support requests.

Screenshot 2023-12-12 at 5 41 40 pm

## Test plan

- Opened settings
- Verified it's gone

### DIFF
--- a/vscode/src/services/StatusBar.ts
+++ b/vscode/src/services/StatusBar.ts
@@ -139,14 +139,6 @@ export function createStatusBar(): CodyStatusBar {
                     c => c.experimentalSymfContext,
                     false
                 ),
-                createFeatureToggle(
-                    'Simple Chat Context',
-                    'Experimental',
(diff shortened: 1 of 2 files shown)


dominiccooney:
Tests need updating. You can crib from local-embeddings.test.ts the way to override a single additional flag in an e2e test, if you need to do that. It is something like

```
import { test, ExtraWorkspaceSettings } from './helpers'

...

test.extend<ExtraWorkspaceSettings>({
        extraWorkspaceSettings: {
                    'cody.foo.bar': true,
    })
('it should blah blah blah', ({ page, sideBar } => {
  ...
})
```

There's another example there where you can run code in the extraWorkspaceSettings hook if you want.

valerybugakov (approved):

Issue #2058, opened 2023-12-01
bug: Document command returns malformed comments


kukicado:
### Version

v0.17.1701443218

### Describe the bug

Running the Document command in a .php file returns malformed text. It appears that some of the comments are returned twice, but only some are properly formatted. 

![screenshot_2023-12-01_at_3 05 47___pm_720](https://github.com/sourcegraph/cody/assets/1770570/21147be0-e3dd-4d75-af88-f056b13604ec)

### Expected behavior

It should return a doctoring without additional copy that breaks the UI.

### Additional context

I tried the Doc command in a couple different .php files in a large Laravel based project. This behavior did not happen in every file, but on the ones it did, it did so consistently, I was not able to ever generate a clean docstring for this function using the Document command.

abeatrix:
i've run into this issue before and I believe this has been fixed by @umpox in https://github.com/sourcegraph/cody/pull/1883#pullrequestreview-1757511751

taylorsperry:
I think this still isn't working quite right. [Loom](https://www.loom.com/share/ef8aedb7fbe248feb0c5e5add71af886). I'm on v0.18 and it still looks like documentation for php is malformatted, and "Undo" doesn't fully undo the generated code.

beyang:
I suspect this is happening, because VS Code auto-injects doc delimiters like `*/` when you start a doc block

PR #2290, opened 2023-12-12
Edit: Fix malformed comments from doc command with selection

umpox:
## Description

closes https://github.com/sourcegraph/cody/issues/2058

`getDocCommandRange` for the `/doc` command was returning the incorrect selection. It was overriding the end position rather than just adjusting the start position for python.

This meant that the edit command was treating this as an `add` (no selected code thus nothing to edit)

### Before

https://github.com/sourcegraph/cody/assets/9516420/62121460-3ddf-47d1-9b41-f86d94f43e12

### After

https://github.com/sourcegraph/cody/assets/9516420/23519065-fd6c-4969-b034-e0530330d328

## Test plan

1. Highlight some code where there is not already a folding range
2. Use the "doc" command via the chat or right click menu
3. Check that the inserted documentation is correct

### DIFF
--- a/vscode/src/commands/CommandRunner.ts
+++ b/vscode/src/commands/CommandRunner.ts
@@ -232,10 +232,7 @@ function getDocCommandRange(
         }
     }
 
-    if (languageId === 'python') {
-        const adjustedStartPosition = new vscode.Position(selection.start.line + 1, 0)
-        return new vscode.Selection(adjustedStartPosition, selection.end)
-    }
-
-    return selection
+    const startLine = languageId === 'python' ? selection.start.line + 1 : selection.start.line
+    const adjustedStartPosition = new vscode.Position(startLine, 0)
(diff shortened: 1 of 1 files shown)


vscode/src/commands/CommandRunner.ts
```diff
@@ -235,5 +232,10 @@ function getDocCommandRange(
         }
     }
 
-    return new vscode.Selection(pos, pos)
```
umpox, line 238: This was the offending line (overriding the entire range rather than just the start).

Once we're out for GA I think we should improve our integration tests around all the entry points for commands to help catch stuff like this, lots of variants depending on if there's a selection/the language/the command

umpox:
Follow up for a related scroll bug: https://github.com/sourcegraph/cody/pull/2296

philipp-spiess (approved):
Awesome!

PR #2295, opened 2023-12-12
remove rg-based local keyword context fetcher

beyang:
Removes the ripgrep-based keyword context fetcher, which is often slow and leads users through a poor experience. Now that embeddings are easier to set up, we should push users through that process (or turn on the symf-based keyword fetching, which will happen soon).

It is better to have no context used (which will be more obvious to the user) than to attempt to use a context fetching mechanism that is slow where it's not obvious to the user what's going on (they'll assume Cody in general is slow).

## Test plan

- [ ] Test that chat still works when embeddings aren't available


philipp-spiess (approved):
Screenshot 2023-12-12 at 11 38 26

Love to see it

philipp-spiess (approved):
Screenshot 2023-12-12 at 11 38 26

Love to see it

Issue #2101, opened 2023-12-05
bug: /doc command hijacks scroll when code is in view


toolmantim:
### Version

main (18afbe650fe052bfc3a22100d3cedd33f0d22b7e)

### Describe the bug

When you execute the `/doc` command it scrolls your editor, first when it's run and then again afterwards, even if the code is visible:

https://github.com/sourcegraph/cody/assets/153/46c907a4-b91c-4d52-9ba5-34dddac3d685

### Expected behavior

* No scroll occurs if the top of the edit range is within view
* If scroll is triggered, it needs to be the line before the range so that the codelens is in the scrolled view

taylorsperry:
Marking as P1. We should fix, but I don't think it's a GA blocker.

PR #2296, opened 2023-12-12
Edit: Fix doc command jumping unnecessarily

umpox:
closes https://github.com/sourcegraph/cody/issues/2101

## Description

It is possible that the selection contains more code than is visible on the screen.

In this case, visibleRanges will never contain the entire selection.

We only really care about the start position of the selection (where the edit will be inserted). We use that position instead.

## Test plan

Run the /doc command on a large function when the entry point is already visible

Check the editor does not scroll unnecessarily

Run the /doc command on a large function when the entry point is NOT visible

Check the editor does scroll the entry point into view


toolmantim (approved):
Nice!

Issue #2259, opened 2023-12-11
bug: "Ask Cody to Fix" jumps when fix applied


taylorsperry:
### Version

v0.19.1702220762

### Describe the bug

"Ask Cody to Fix" jumps when fix applied. See [Loom](https://www.loom.com/share/a403c9454727437282d6b4b126f414fb) at 0:28.  Also seems pretty slow? 

### Expected behavior

Shouldn't jump when edits are applied.

umpox:
Related: https://github.com/sourcegraph/cody/issues/2259

PR #2297, opened 2023-12-12
Edit: Do not scroll into view on complete

umpox:
closes https://github.com/sourcegraph/cody/issues/2259

## Description

Previously we would always scroll a completed task into view on completion.

We have a bug with this already (https://github.com/sourcegraph/cody/issues/2259).

One fix here is just to scroll this into view when the code is NOT already in view.

However, I think this might actually be annoying behaviour. The user already gets a notification for edits that are made if the file is not currently active. It seems overly aggressive to force their scroll back to the code when editing an active file.

## Test plan

1. Edit code in a large file
2. Scroll the code out of view
3. Check we don't hijack the scroll


philipp-spiess (approved):
Sure

PR #2298, opened 2023-12-12
Chat: Update welcome message

### LINKED ISSUE #2282: bug: Welcome message should not talk in first-person
### Version

0.18.0

### Describe the bug

Related to #2194, the chat welcome message sets up the expectation that you're talking to Cody instead of using Cody to chat with Claude/ChatGPT/etc with additional context:

Screenshot 2023-12-12 at 1 01 45 pm

This can create confusing scenarios where sometimes it responds as Cody, and other times as the underlying LLM.

### Expected behavior

The welcome message shouldn't be written in first person, and instead be only instructional.

### Additional context

@sqs's comments:

> I think it's fine that this happens since we expose the model so clearly, but we should change the text that says I'm Cody. right above. Even just removing I'm Cody. seems like it would be fine. Cody is the system, not some personification.

toolmantim:
This updates the welcome message to not be in first-person and make it more transparent that you're chatting with the model shown at the top. It also includes clearer keyboard shortcuts, and actual icons of the buttons they can expect to find in the interface (sidebar, title, status bar, etc).

| Before | After |
| - | - |
| Screenshot 2023-12-12 at 10 59 52 pm | Screenshot 2023-12-12 at 10 58 41 pm |

Fixes #2282 

## Test plan

- Started a new chat
- Clicked on all the links


philipp-spiess (approved):
CRISP

Issue #2082, opened 2023-12-04
bug: `/doc` command leaves code decorations after edits are applied


MaedahBatool:
### Version

v0.17.1701702451

### Describe the bug

After running the `/document` command there are outline decorations littered throughout the code, outside the bounds of the doc comment that was added:

<img width="803" alt="Screenshot 2023-12-05 at 4 47 53 pm" src="https://github.com/sourcegraph/cody/assets/153/599be9b1-f469-46e8-9c20-245747c56b21">

### Expected behavior

I would expect a neat and clean docs generation process (i.e. no text decorations left after edits are applied)

### Additional context

- Internal [Slack Thread](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1701721946586139)

toolmantim:
I just clarified a few details in this. @kalanchan & @taylorsperry I think this is a bug where decorations aren't being removed properly from the code after the streaming has completed. I was able to recreate it once locally, but can't seem to do it again.

csells:
what's the user mitigation in this case? if they click on their code, does this go away?

toolmantim:
I can reproduce this sometimes on large functions, but it seems to happen about 1/10 times

toolmantim:
@csells the user mitigation is to click "Done" to make the decorations go away.

PR #2304, opened 2023-12-12
Edit: Flush diffs on apply

umpox:
closes https://github.com/sourcegraph/cody/issues/2082

## Description

Diffs can be scheduled - these update decorations, but this can conflict with removing the decorations when a task is applied.

So we have a race condition where it is possible that:
1. The task is applied, and decorations are removed
2. Later, a scheduled diff updates and re-applies the old decorations

To work around this, we just ensure that any scheduled diff updates are ran before we finally remove all decorations

## Test plan

1. Run an edit
2. Confirm that decorations are not shown on apply

1. Run the `/doc` command
2. Confirm that decorations are not shown on apply


philipp-spiess (approved):
Friends Clap

Issue #2300, opened 2023-12-12
bug: Missing commands in / menu


toolmantim:
### Version

main e75560d7d7380a662a663ed2c10bc1607c7db60b

### Describe the bug

<img width="1160" alt="Screenshot 2023-12-12 at 11 52 36 pm" src="https://github.com/sourcegraph/cody/assets/153/d6647f0f-0f25-4666-839a-4031898c5d28">

### Expected behavior

All commands appear in the slash menu

taylorsperry:
Context in EPD touchpoint: @umpox Possibly a regression from [this change](https://github.com/sourcegraph/cody/issues/2172)?. I can look into it

toolmantim:
@taylorsperry @kalanchan @umpox `/test` is there now, but I don't think this is resolved yet as I'm still seeing a big difference in the commands appearing in this list:

<img width="1406" alt="Screenshot 2023-12-13 at 11 05 50 am" src="https://github.com/sourcegraph/cody/assets/153/baee0341-d900-4d91-81cf-838cfb887d13">

The interaction should follow how it was before we switched to simple chat (it just executes the command in the active editor, with no affect on the chat):

https://github.com/sourcegraph/cody/assets/153/4cc1c967-e794-4226-81b9-bc6e67a9815d

kalanchan:
@toolmantim we talked about this in the last couple minutes of the [EPD touchpoint](https://docs.google.com/spreadsheets/d/1WGd27-_7vmmviZn0zVK_f8vhR7FzJVJjAElUh0IUvbQ/edit?usp=drive_open&ouid=106704778750093187906), we made the call to remove all edit/doc commands away from chat and only have to invokable through command drop down or side panel

toolmantim:
If the intention is to exclude commands that perform edits, we should also probably exclude custom commands that perform inserts, as they seem to still have the same chat transcript output issue:

https://github.com/sourcegraph/cody/assets/153/48ff1fb1-485e-413b-a65d-024fc72e6792

Ideally though, given we don't have any UI label that indicates some commands are 'chat' commands and others are 'edit' commands… and it's just a flat list… we should match the original behaviour.

Or we introduce the idea of "Chat Commands" (not ideal at this point I think):

<img width="407" alt="Screenshot 2023-12-13 at 12 04 39 pm" src="https://github.com/sourcegraph/cody/assets/153/9e1ef98a-4772-44d7-9bda-dd7e33ade5de">

[...]

PR #2305, opened 2023-12-12
Commands: Include /test

umpox:
closes https://github.com/sourcegraph/cody/issues/2300

## Description

Re-add `/test` command

## Test plan

Run slash command

Check test is there

Check it works


umpox:
Checked if this would be triggered by anything else, seems that the only place where we generate interaction `error`s is from custom commands via `newInteractionWithError`

philipp-spiess (approved):

Issue #2158, opened 2023-12-06
feedback: generate commit message in vscode's commit message line


vadi2:
### Areas for Improvement

- [X] UI/UX
- [ ] Onboarding
- [ ] Docs
- [ ] Chat
- [ ] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience.

It would be great to have a button right here that would do its best at generating the commit message:

![Screenshot 2023-12-06 9 13 05 PM](https://github.com/sourcegraph/cody/assets/110988/b52370fa-d287-4206-b139-f5b6918f2e69)

### Why is this feedback important? Please describe the rationale.

Copilot has this feature

### Describe the solution you'd like to see

A button to generate the commit message

PR #2306, opened 2023-12-12
feat: Add experimental commit message generator

RXminuS:
## Description

https://github.com/sourcegraph/cody/assets/9516420/96d18bff-9300-4220-8382-5d2d2a682749

I couldn't find this button (even though I saw copilot had it) so thought I'd give it a stab. It's not an "official" VSCode extension endpoint yet but seems to work. I'm still a bit unfamiliar with the codebase so would like some pointers on how to correctly generate a prompt and fire it off. 

I was initially inspired by the custom `cody.json` command, but there seem to be a few places that generate commit messages. Personally I was thinking about adding a custom recipe, but again I need a few pointers on examples as Cody was't quite able to (yet) point me to the right files.

## Test Plan

Staged only:
1. Make changes to a branch
2. Stage only some changes
3. Generate commit message, check that it matches staged changes

All:
1. Make changes to a branch
2. Do not stage any changes
3. Generate commit message, check that it matches all changes


RXminuS:
Fixes #2158

RXminuS:
Screenshot 2023-12-12 at 17 51 56

philipp-spiess:
Hey @RXminuS! This is absolutely awesome. We're a bit busy right now preparing for a major release but I'll look into this. I just tested it locally and it works so well already 🔥 

cc @toolmantim for some design/UX input - I think we'd need a different icon for when the response is generating - and then I saw that there are a few todos in the code too. I'll do a more thorough review tomorrow!

RXminuS:
@philipp-spiess Absolutely, this is only a first draft PR, thought I'd seek feedback early :-)

The icon for when responses are being generated is set by VSCode currently (there's tons of people complaining about it). I'm hoping that as the extension endpoint matures we get more control.

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+            this._subscription?.dispose();
+            this._subscription = undefined;
+        }
+	}
+
+    public dispose(): void {
+        this.disposables.forEach(d => d.dispose());
+    }
+    public async provideCommitMessage(repository: Repository, changes: string[], cancellationToken?: any): Promise<string | undefined> {
+        // TODO: load commit template from settings
```
toolmantim, line 31: I wonder if a `cody.commit.preInstruction` that mimics our chat pre-instruction here would be best. You could guide it to use a template, and tweak it to your liking.

toolmantim (commented):
This is awesome @RXminuS. Yeah UI wise, there's nothing to do… it's just an API we're implementing. I left a note on my idea/preference for one of the todos.

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -0,0 +1,77 @@
...
+    public title = 'Generate Commit Message (Cody)'
+
+    private disposables: vscode.Disposable[] = []
+    private _subscription?: vscode.Disposable;
+
+    constructor(private readonly options: CommitMessageProviderOptions) {}
+
+    public onConfigurationChange(config: ConfigurationWithAccessToken): void {
+		if(config.experimentalCommitMessage){
+            this._subscription = this.options.gitApi.registerCommitMessageProvider(this);
```
philipp-spiess, line 20: Should we gate against `registerCommitMessageProvider` not being defined on `gitApi` (i.e for older versions or so?)

RXminuS:
Based on @beyang suggestion I switched it to the CodeCompletionsClient. Unfortunately I'm out of autocomplete credits today but will try and run it against a local LLM.

I'll add a setting as well so a user can control the template. Any suggestions on a good way to query for relevant "similar" commits that can serve as examples?

vscode/src/scm/CommitMessageProvider.ts
RXminuS: I've made the prompt a bit better now. Before moving to a proper PR I'll move it to a shared location so that the same code can also generate a commit message in chat etc. Because I'd like to re-use it for a chat version (future PR)...with a twist

I'm a pretty messy committer, usually I fix things up as I see them which means that I always have a bunch of different groups of changes in my unstaged files. So what I picture is that I can ask the chat to help me commit; where we'd do ReAct prompting to figure out which changes are related and then generate a commit message for each one of those groups and then after some tweaks in chat gives me a "/do-this" that then runs the correct set of git commands.

By abstracting the commit-message prompt it'll be easy to create a "function-call" that the LLM can call itself for each group it has come up with. In the past I've used libraries like the below for this.

- https://github.com/LevanKvirkvelia/salute
- https://github.com/outlines-dev/outlines
- https://github.com/guidance-ai/guidance

Maybe went a bit off topic and I think the "Super-chat" version is out of scope for this PR...but to answer your question, yes I hope to at least organise it in such a way that I can re-use the work :-)

philipp-spiess:
@RXminuS I think it would be better to use the normal completions client, not the code specific one or are you seeing an improved context with it? The code completions client is more fine-tuned for our autocomplete use case which I don't think is helpful here. It might still be helpful to use a faster model by default (using the `fast: true` switch) but that can also be achieved by going through the chat APIs

RXminuS:
What does the fast do in this case. What I'm looking for is more control over the start of the assistant output and a 'end-sequence'

RXminuS:
Ah, nope...it was my sleepy eyes indeed, I see it now :-)

RXminuS:
Hey @philipp-spiess, I've been experimenting with different ways to improve prompting and reusing logic in Cody. I've moved the core logic to a recipe, but I'm not entirely satisfied with how I have to call the recipe to achieve a non-user interaction.

My goal is to generate an output different from a normal chat message, as I want to directly use the result as a structured output. This seems less common in Cody right now, but I anticipate it becoming more relevant as ReAct-based logic becomes prevalent. The challenge arises when you want to reuse prompt logic as part of a chain of thought.

I considered extracting the response from a chat and searching for specific text, but this Lang-Chain linking approach has for me been very error-prone in the past, so I'd prefer to avoid it. For now, I added a method to the Commit Message Recipe that allows me to retrieve only the input prompt.

However, there's a concern about the type-safety of doing this given the current typing of Recipes. One potential solution could be typing them with a discriminative union, as mentioned in my comment in `main.ts:207`. This would enable recipes to provide various methods for reusing logic in contexts beyond "chat conversations."

Alternatively, it would be great to have a better way of omitting or substituting parts of an interaction to make them not user-facing. In an ideal scenario, I could use the chat model with an autocomplete API to fully control token generation and handle tasks like filling out JSON structs.

I'd appreciate your thoughts!

**Note: I haven't been able to test any of my changes due to issues with the v1.0.0 release and Pro account upgrade; models don't work anymore (#2394). So if something's broken in my PR that's why :-) Congrats on the release though 👏**

umpox:
Hey @RXminuS, this is really cool! Great work.

I'm looking into the changes now, I'll leave some comments and push some changes directly if that's cool with you.

lib/shared/src/chat/recipes/recipe.ts
```diff
@@ -15,40 +15,6 @@ export interface RecipeContext {
...
-//     [K in ExampleAllRecipes['id']]: ExampleAddId<Extract<ExampleAllRecipes, { type: K }>>
-// }
-// // Get all possible keys
-// type ExampleAllRecipeIds = keyof ExampleAllRecipesUnique
-// // Now we have typed access to additional properties
-// const unk: ExampleAllRecipes = { id: 'recipe-a', a: false } as any
-// if(unk.id === 'recipe-a'){ //knows ids
-//     unk.a //exists
-// }
-
```
umpox: @RXminuS

Thanks for this comment, I agree that this would be a better way to structure recipes and give a lot more type safety.

We are (slowly!) moving away from the recipe architecture for building out new features. I have updated this branch to merge the recipe code into `CommitMessageProvider` and removed this comment. Will create an issue so we have more of a concrete plan for recipes and link this there though!

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -41,57 +51,75 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
                     {
-                        maxTokensToSample: 1000,
-                        temperature: 0,
-                        stopSequences: ['</commit-message>'],
+                        fast: true,
+                        stopSequences: [`</${COMMIT_MESSAGE_TOPIC}>`],
                     }
                 )
+
+                cancellationToken?.onCancellationRequested(abort)
```
umpox, line 122: Added `abort` logic, which will be triggered when a user clicks the "Stop" icon whilst Cody is generating.

image

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -41,57 +51,75 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
+                completion = content
+                return Promise.resolve()
+            },
+            onTurnComplete: () => {
+                return Promise.resolve()
+            },
+        })
+
+        // Prefix the assistant response with any prewritten human text
+        const prefix = repository.inputBox.value
```
umpox, line 90: Including the prefix here so we still retain a half-written commit message if the user has one.

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -41,57 +51,75 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
+        }
+
+        if (isTruncated) {
+            this.options.editor.showWarningMessage(
+                'The length of the diff exceeded the maximum available length. Using a truncated diff.'
+            )
         }
 
         let completion = ''
+        const multiplexer = new BotResponseMultiplexer()
```
umpox, line 78: Just refactored to use the multiplexer which gives us similar logic for targeting responses wrapped in XML tags

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -41,57 +51,75 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
+        const { isEmpty, prompt, isTruncated } = humanPrompt
         if (isEmpty) {
-            return ''
+            return repository.inputBox.value
+        }
+
+        if (isTruncated) {
+            this.options.editor.showWarningMessage(
+                'The length of the diff exceeded the maximum available length. Using a truncated diff.'
+            )
```
umpox, line 74: Added a VS Code warning if we have to use a truncated diff

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -41,57 +51,75 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
                         },
-                        onChange: text => {
-                            completion = text
-                        },
                     },
                     {
-                        maxTokensToSample: 1000,
-                        temperature: 0,
-                        stopSequences: ['</commit-message>'],
+                        fast: true,
```
umpox, line 117: Using the `fast` chat model now

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -105,8 +133,124 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
+        }
+    }
+
+    private async getCommitTemplate(cwd: string): Promise<string | null> {
+        try {
+            const commitTemplateFile = (
+                await execFile('git', ['config', '--get', 'commit.template'], {
+                    cwd,
+                })
+            ).stdout.trim()
```
umpox, line 186: Changed to using `git config` instead of `git ls-files .gitmessage etc`

I believe the commit message template has to be set through config, so I think this should be still safe to use and catch more cases (no clear naming convention to these files AFAICT)

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -105,8 +133,124 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
+        const diffCliCommands = ['diff', '--patch', '--unified=1', '--diff-algorithm=minimal', '--no-color', '-M', '-C']
+        let diff: string
+
+        // First, attempt to get a diff from only staged changes
+        diff = (await execFile('git', [...diffCliCommands, '--staged'], { cwd })).stdout.trim()
+
+        if (!diff) {
+            // Attempt to get a diff from all changes, not just staged.
+            diff = (await execFile('git', [...diffCliCommands, 'HEAD'], { cwd })).stdout.trim()
+        }
```
umpox, line 150: Now falling back to all changes if we cannot find anything using just `--staged`.

I think this is the best approach. If a user triggers this feature but has no files staged they almost certainly want the fallback option

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -41,57 +51,75 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
...
         changes: string[],
-        cancellationToken?: any
+        cancellationToken?: vscode.CancellationToken
     ): Promise<string | undefined> {
-        //TODO: Handle cancellation
-
         // we ignore the changes coming form VSCode here as the resulting commit messages were just not as good
         // it could maybe be something to fall back on for web, but the recipe doesn't work for that atm. anyways.
-        const humanPrompt = await this.options.recipe.getHumanPrompt(undefined, this.options)
+        const humanPrompt = await this.getHumanPrompt(changes)
```
umpox, line 61: I re-added `changes` here. From my personal testing I think this works quite well but happy to discuss. We can still fallback to using the git CLI to get our own diff, but looking at other implementations in different extensions - I think using `changes` here is fine and we may just need to tweak our prompt.

vscode/package.json
```diff
@@ -833,6 +833,12 @@
           "default": false,
           "markdownDescription": "Adds suggestions of possible relevant messages in the chat window."
         },
+        "cody.experimental.commitMessage": {
+          "order": 8,
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Adds a button to the SCM input field to generate a relevant commit message for the current changes."
+        },
```
umpox, line 841: I updated the default here to be `true`.

Reasoning:
- This is very low friction for users, it's just a button in the SCM input field
- Seems quite stable in its current state, low chance of annoying output - Cody generates the text but the user can easily edit/remove.

Further question:
- Should this be `experimental` still?

@sourcegraph/cody-vscode

vscode/src/scm/CommitMessageProvider.ts
RXminuS: Yeah saw this; what does fast actually do/imply?

vscode/src/scm/CommitMessageProvider.ts
RXminuS: Agree, was gonna ask about this

vscode/package.json
RXminuS: The API isn't stable yet from VSCode so my reasoning was to keep it experimental until it is. Just in case VScode does something breaking

vscode/src/scm/CommitMessageProvider.ts
RXminuS: Thanks...just me being lazy ☺️

vscode/src/scm/CommitMessageProvider.ts
RXminuS: I think UX wise personally it's actually less nice to keep old data around. Just because it's a tedious box to then find and edit in. Usually because I start typing at first and then realize I'm terrible and want to auto generate. The only thing I thought about is using it as an additional prompt input just in case it references an issue or so?

Maybe @toolmantim has some input on the UX of it?

vscode/src/scm/CommitMessageProvider.ts
RXminuS: Probably want to doc that somehow for users?

vscode/package.json
RXminuS: Maybe here we can inform about git.config commit template? Or is there a better place?

vscode/src/scm/CommitMessageProvider.ts
umpox: It uses the fast chat model, which in this case is Claude Instant. We should better document this, sorry!

vscode/package.json
umpox: That's fair, I think we can keep it as an experimental setting at least until then. I feel setting the default to `true` still makes sense as we're already quite defensive (`registerCommitMessageProvider` has to exist for this to be enabled)

umpox: > Maybe here we can inform about git.config commit template? Or is there a better place?

Hmm, not very discoverable here. I kinda feel like using a `commit.template` file is more of an implementation detail and we don't necessarily need to document.

Kinda like, "oh cool Cody figured out how we write commits here", rather than telling the user directly.

cc @toolmantim

vscode/src/scm/CommitMessageProvider.ts
umpox: Yep also interested to hear from @toolmantim 

Two directions:
1. Use the current commit message text as a prefix, essentially asking Cody to finish the message

2. Ignore the current text (or provide it as context), Cody will overwrite whatever is in the input box.
 
WDYT @toolmantim?

vscode/package.json
RXminuS: Yeah I more mean to document how you provide a different template. Other plugins have vscode settings for this. I think it could be cool to point out somewhere that Cody will look at your git config for a template?

vadi2:
Thanks @RXminuS for picking this up :) I'd love to see this in Cody.

umpox:
Hey @RXminuS, we haven't forgotten about this - just have had a bit of disruption with the holidays :)

I'll aim to get this merged next week 👍

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -0,0 +1,256 @@
...
+    private async getCommitTemplate(cwd: string): Promise<string | null> {
+        try {
+            const commitTemplateFile = (
+                await execFile('git', ['config', '--get', 'commit.template'], {
+                    cwd,
+                })
+            ).stdout.trim()
+
+            if (commitTemplateFile.length === 0) {
+                return null
```
abeatrix, line 189: If there is no template path found, and before falling back to use the default template, do you think using the last n commits as template/context would be helpful? 👀

abeatrix (dismissed):
I've left some comments and questions inline.

The only concern I have is that the files listed in .cody/ignore should be filtered out from the diff output, can we add that? 🙇🏻‍♀️ 

Other than that I am very excited about this 🤤 I use Cody to generate commit messages and changelog entry daily so having it available in one click is just a game-changer to me!

abeatrix:
Q: I tried the build from this branch but wasn't able to get the cody button to show up in the commit input box, is there something else I will need to do to get it set up?

image

idea: GitLens has the same feature that displays a disclaimer when uses it for the first time, maybe something we can add later :)

image

RXminuS:
Yes, I think we can add that the ignore.

I don't think we need to call out specifically that the changes might be sent on a fit commit because every interaction with Cody does so, but you can also setup a local LLM. So I feel it should (if even) be a more productive wide thing?

W.r.t. the button not showing. The vscode API isn't mature yet so I think if multiple plugins try to modify this button (Copilot, Gitlens) the loading order of plugins matters, I can look into it.

philipp-spiess:
@RXminuS Hey Rik! You mentioned in your last message that you want to look into improving the loading order. Is this still on your list? Otherwise we'd be happy to take over from here too if you don't have time to look into it. I realize this PR is already diverged from `main` quite a lot so a rebase/merge is necessary too I’m afraid. Let us know if you're still looking into it or if you want us to take over from here - whatever you prefer :)

RXminuS:
Happy to fix that and do the rebase at the end of today. Was waiting a bit so you guys would have capacity again to merge with all the cool stuff coming out!

If you do have time before I do feel free to run with it, I'm sure I'm not the only one who would love to see it added 😊

RXminuS:
Just looked into the commitMesageProvider again, there's no update on docs or info about it.

My suspicion is still that it's a last to register (probably by name) wins and I don't see a way to control extension loading order. We maybe just need to add to the docs that if it doesn't show up to disable Copilot (maybe that's a feature 😅)

So i don't think it's something that can be fixed right now.

There are a few comments/questions open still but maybe let's just rebase and merge; get some community feedback and then tweak from there?

philipp-spiess:
@RXminuS Sounds good! cc @umpox who has been looking into a different loading-order-related issue; maybe you have some ideas here? 👀

umpox:
> My suspicion is still that it's a last to register (probably by name) wins and I don't see a way to control extension loading order.

@RXminuS @philipp-spiess I think this is correct using the API that we have now. I don't think we need to do anything special right now, if we want to beat another extension to this space in the future we could always add a short delay before calling `registerCommitMessageProvider`. 

Copilot have access to a [proposed API](https://sourcegraph.com/github.com/microsoft/vscode@3b9f4a340c4433b4cff3635d53ad20125eb028e9/-/blob/src/vscode-dts/vscode.proposed.contribSourceControlInputBoxMenu.d.ts). I think this API takes precedence over our method, unfortunately there's nothing we can do about that right now.

RXminuS:
@philipp-spiess So sorry for the delay...I ended up with an ear infection so have been out this week :-/

I tried doing a rebase but it got quite messy, and given that we probably want to squash the commits anyways I just did a merge and got it all working again. There's one comment left by @umpox (https://github.com/sourcegraph/cody/pull/2306#pullrequestreview-1807170314) w.r.t. filtering out cody/ignore files.

I had a quick look but feel I've missed a bit of context for some of the upstream changes. Got the impression that this is handled in the gitAPI already? However `vscodeGitAPI` isn't exported so wasn't quite sure what the intended use was. Could you double-check `vscode/src/main.ts:211` because it feels a bit wrong to do that now. Happy to fix with some pointers or feel free to take it from here. I might still be a little bit slow to respond as I'm still a bit 🤧

philipp-spiess:
@RXminuS Awesome, thanks! I think there's still going to be some work necessary to support cody/ignore. Thea idea for this feature is that if a file matches the explicit ignore rules, _it should never be send to tbe LLM_. So we'll have to make this work here somehow (perhaps by doing a post-processing on the git diff or by adding the cody ignore content to the git commands somehow). 

During some quick googling I found this: https://stackoverflow.com/a/21079437 but that means we'd have to rewrite the cody/ignore syntax to match this. cc @abeatrix is cody/ignore just gitignore syntax?

RXminuS:
I think @umpox added back my first attempt as a fallback to get the changes directly from vscode (so not running hit command)

In which case it's probably easiest to post process filter all diffs regardless of what method they came from instead of relying on gut command features. I just have to check what vscode actually gives us.

philipp-spiess:
@RXminuS Agree this is probably easier. Feel free to add a npm dependecy to parse git diff text if that makes your live any easier (these are usually very slim IIRC)

RXminuS:
Cool, will do! I'm still a bit under the weather so might not get to it today but will have a look asap.

abeatrix:
> is cody/ignore just gitignore syntax

@philipp-spiess It is!

> Probably makes sense to make some class/service where you can simply check by passing a path if you can expose it or not.

Not sure if this is what you're looking for, but you can check if a file uri is an ignored path using [isCodyIgnoredFile](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/completions/inline-completion-item-provider.ts?L192)

RXminuS:
@philipp-spiess @umpox so sorry it took me some time, I'm finally on the mend and had some time so I've now added the codyignore filtering. No external deps were needed. I did however find a potential problem (commented CommitMessageProvider.ts:186 ).

Basically the issue is that we're doing everything based of the workspaceRoot, but for Git that's not necessarily true as the git root folder might be up the tree from the workspace root. For instance when opening the vscode subdirectory when debuging the plugin your run into this. This then means that the URI path gets created relative to the workspaceRoot not the gitRoot for the workspace. There's a similar problem with codyIgnore where it doesn't actually discover it. I'm guessing we don't actually want .cody folder to have to be in the workspaceRoot and instead have it behave similar to the .git folder.

Hope I explained it ok.

When you look at the vscode git plugin you see that they actually run a .git path discovery for the workspace. Was something like that already implemented or needs to be put on the roadmap?

umpox:
@RXminuS Looking good. I'm reviewing and doing a small amount of updates as I go. I'm going to add an E2E test so we can be more confident around usage of the Git API and also the Cody Ignore stuff.

RXminuS:
Absolutely! 👍

RXminuS:
@umpox any thoughts on the Git root for workspace? (And also for .cody folder)

vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -0,0 +1,317 @@
...
+--------------------
+Type can be
+    feat     (new feature)
+    fix      (bug fix)
+    refactor (refactoring production code)
+    style    (formatting, missing semi colons, etc; no code change)
+    docs     (changes to documentation)
+    test     (adding or refactoring tests; no production code change)
+    chore    (updating grunt tasks etc; no production code change)
+    wip      (work in progress commit to be squashed -- do not push!)**
```
philipp-spiess, line 303: I wonder if we should make this part configurable. I know that in some repos this type is a common pattern but this feels like it has the potential to be highly depending on the project.

RXminuS: You can do so already with your own commit template right?

philipp-spiess: ah, yes, ofc 🤦

umpox:
@RXminuS any thoughts on the Git root for workspace? (And also for .cody folder)

I don't think we need to handle this right now, but we can look into it in a follow up. @abeatrix Do we handle this for Cody Ignore yet? I.e. if the cody ignore file is in a directory above the current workspace dir

RXminuS:
> @RXminuS I don't think we need to handle this right now, but we can look into it in a follow up

Shall I open an issue for this after the PR merges?

abeatrix:
@umpox no we only look at workspace root right now so we don't need to handle that in their PR 👍

philipp-spiess (approved):
This is looking slick now! I'll let @umpox merge this but here's my stamp 😎

umpox:
> Shall I open an issue for this after the PR merges?

@RXminuS Yeah that'd be really helpful 👍

umpox (approved):
Awesome work @RXminuS! Excited to start using this 🎉

umpox:
Hey @RXminuS, just wanted to keep you in the loop here. Unfortunately VS Code have removed this API from the Git extension in the latest editor release, so we've had to pull this.

It's looking like we might have an opportunity to make use of this if the [current proposed API is finalized](https://github.com/microsoft/vscode/issues/195474). Unfortunately we can't enable this until then.

If you want, you can revive your PR and we'll make sure it's updated and we get it merged if we get access to the API we need. We should be able to reuse much of this code so want to make sure you get credit for the commit!

RXminuS:
Yeah I saw 😅 I'll keep tabs on it and keep merging in main so we can see what happens with the API.  

And then let's revisit once stabilized. I don't think they'll be huge changes but 🤷‍♂️

One thought might be to at least make an action out of this PR so you can call it as a slash command?

ValentinFunk:
Is there any chance to get this working again? The copilot extension does have it but the messages aren't really that great. When I do "git diff > file" and prompt for a message in cody using the diff as context it works much better.  
image

PR #2312, opened 2023-12-12
Fix rate limit number being displayed

philipp-spiess:
We were calling `[0]` even though it was a string which lead to it picking the first character. So if the rate limit was `222` it would show up as `2`

## Test plan
Screenshot 2023-12-12 at 22 06 57

- Log in with rate limited David user
- Send a message

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -8,6 +8,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
+- Fixes an issue where the wrong rate limite count was shown. [pull/2312](https://github.com/sourcegraph/cody/pull/2312)
+
 ### Changed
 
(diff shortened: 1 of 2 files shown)


vscode/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
+- Fixes an issue where the wrong rate limite count was shown. [pull/2312](https://github.com/sourcegraph/cody/pull/2312)
```
rafax, line 11:
```suggestion
- Fixes an issue where the wrong rate limit count was shown. [pull/2312](https://github.com/sourcegraph/cody/pull/2312)
```

rafax (approved):

toolmantim:
Nice catch!

Issue #2288, opened 2023-12-12
Autocomplete: dynamic multiline completion generation logic throws the invalid position error


valerybugakov:
- On some completions, the error is thrown.
- Stderr: `Error: error parsing streaming CodeCompletionResponse: Error: Illegal argument: character must be non-negative`
- The stack trace is not available.

PR #2323, opened 2023-12-13
Autocomplete: log raw error in the development env

valerybugakov:
## Context

- Log the raw error with the stack trace to the console in the development environment.
- Part of #2288 

## Test plan

n/a

### DIFF
--- a/lib/shared/src/sourcegraph-api/completions/client.ts
+++ b/lib/shared/src/sourcegraph-api/completions/client.ts
@@ -9,7 +9,7 @@ export interface CompletionLogger {
     ):
         | undefined
         | {
-              onError: (error: string) => void
+              onError: (error: string, rawError?: unknown) => void
               onComplete: (response: string | CompletionResponse | string[] | CompletionResponse[]) => void
               onEvents: (events: Event[]) => void
           }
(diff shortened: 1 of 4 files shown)


lib/shared/src/sourcegraph-api/completions/client.ts
```diff
@@ -9,7 +9,7 @@ export interface CompletionLogger {
     ):
         | undefined
         | {
-              onError: (error: string) => void
+              onError: (error: string, rawError?: unknown) => void
```
philipp-spiess, line 12: Can we change `error` to be of `Error` type and avoid having two parameters? Then, a caller that only wants the message, can just call `erorr.message` instead

philipp-spiess (approved):

lib/shared/src/sourcegraph-api/completions/client.ts
valerybugakov: I thought about that, but it's tricky with current usage. There's valuable information in custom messages (first argument), and for debugging, I want to see the original error with the original message and stack trace. Happy to change in the follow-up if you have a solution for this dilemma 🙂

PR #2326, opened 2023-12-13
Autocomplete: remove stop sequences for dynamuc multliline completions

valerybugakov:
## Context

- Stop sequences were intentionally removed from the dynamic multiline completion request params. 
- This PR reverts relevant changes from https://github.com/sourcegraph/cody/pull/2118 to preserve this change.

## Test plan

Try generating multiline completion with two consecutive new lines and ensure they are not truncated in the middle.

### DIFF
--- a/agent/recordings/FullConfig_234680970/recording.har
+++ b/agent/recordings/FullConfig_234680970/recording.har
@@ -77,30 +77,30 @@
             "encoding": "base64",
             "mimeType": "application/json",
             "size": 136,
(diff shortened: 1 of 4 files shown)


philipp-spiess (approved):
You are the boss but I'm personally not sure if a completion that is so long that it adds blank lines isn't _too long_ for a completion. only one way to find out I guess 🙂

valerybugakov:
@philipp-spiess, let's looks at the A/B test results. I understand your concerns and it might be that my experience does not match the expectations of the majority of our users. I noticed that many helpful completion are cut off in the middle because of the stop sequences so I prefer to test this setup first. We should run a separate experiment _with_ stop sequences after it.

PR #2327, opened 2023-12-13
Autocomplete: format completions on accept

valerybugakov:
## Context

- Formats autocomplete suggestions using the available formatter on accept.
- Addresses [user feedback](https://sourcegraph.slack.com/archives/C05AGQYD528/p1701771083790929): "_A dev told us that Cody was suggesting tabs for python code whereas, he was expecting spaces_"
- Closes https://github.com/sourcegraph/cody/issues/2252

## Test plan

1. Write code using tabs instead of spaces.
2. Ensure that you have a configured formatter that prefers spaces.
3. Generate a multiline completion. StarCoder will (most probably) match the indentation style of the existing code (tabs).
4. Accept the completion and ensure that tabs are converted to spaces automatically. 
5. Ensure that only the inserted completion is formatted.


prechayimmee (approved):

agent/src/agent.ts
```diff
@@ -370,7 +370,7 @@ export class Agent extends MessageHandler {
                 throw new Error('Cody client not initialized')
             }
             const provider = await vscode_shim.completionProvider()
-            provider.handleDidAcceptCompletionItem(completionID)
+            await provider.handleDidAcceptCompletionItem(completionID)
```
philipp-spiess, line 373: Since I’m seeing this change in the agent file: Agent won't have a way to format the selection, right? Do we make sure that it's not crashing with this change?

lib/shared/src/configuration.ts
```diff
@@ -23,25 +23,42 @@ export interface Configuration {
...
+     * Autocomplete
+     */
+    autocomplete: boolean
+    autocompleteLanguages: Record<string, boolean>
     autocompleteAdvancedProvider: 'anthropic' | 'fireworks' | 'unstable-openai' | null
     autocompleteAdvancedServerEndpoint: string | null
     autocompleteAdvancedModel: string | null
     autocompleteAdvancedAccessToken: string | null
     autocompleteCompleteSuggestWidgetSelection?: boolean
+    autocompleteFormatOnAccept?: boolean
```
philipp-spiess, line 40: I’m curious why this is not in "experimental" state 😅

lib/shared/src/configuration.ts
```diff
@@ -23,25 +23,42 @@ export interface Configuration {
...
+    /**
+     * Autocomplete
+     */
+    autocomplete: boolean
+    autocompleteLanguages: Record<string, boolean>
     autocompleteAdvancedProvider: 'anthropic' | 'fireworks' | 'unstable-openai' | null
     autocompleteAdvancedServerEndpoint: string | null
     autocompleteAdvancedModel: string | null
     autocompleteAdvancedAccessToken: string | null
     autocompleteCompleteSuggestWidgetSelection?: boolean
```
philipp-spiess, line 39: Oh there's a bit to clean up!

vscode/package.json
```diff
@@ -955,6 +955,11 @@
           "default": true,
           "markdownDescription": "Autocomplete based on the currently selection in the suggest widget. Requires the VS Code user setting `editor.inlineSuggest.suppressSuggestions` set to true and will change it to true in user settings if it is not true."
         },
+        "cody.autocomplete.formatOnAccept": {
+          "type": "boolean",
+          "default": true,
```
philipp-spiess, line 960: Should we really add this by default without testing it?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -156,6 +161,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
 
     constructor({
         completeSuggestWidgetSelection = true,
+        formatOnAccept = true,
```
philipp-spiess, line 164: Why do we have more than one place defining a default 🤔

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -489,6 +500,50 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
         )
     }
 
+    public async formatCompletion(autocompleteItem: AutocompleteItem): Promise<void> {
```
philipp-spiess, line 503:
```suggestion
    private async formatCompletion(autocompleteItem: AutocompleteItem): Promise<void> {
```

philipp-spiess: Also this doesn't seem to require `this` so maybe extract in a stand-alone function to keep the class slimmer?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -489,6 +500,50 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+
+            if (formattingChangesInRange.length !== 0) {
+                const edit = new vscode.WorkspaceEdit()
+                for (const change of formattingChangesInRange) {
+                    edit.replace(document.uri, change.range, change.newText)
+                }
+                void vscode.workspace.applyEdit(edit)
+            }
+        } catch (unknownError) {
+            const error = unknownError instanceof Error ? unknownError : new Error('unknown')
```
philipp-spiess, line 540:
```suggestion
            const error = unknownError instanceof Error ? unknownError : new Error(unknownError ?? 'unknown')
```

Would that work in case it has a toString?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -489,6 +500,50 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+            logError('InlineCompletionItemProvider:formatCompletion:error', error.message, error.stack, {
+                verbose: { error },
+            })
```
philipp-spiess, line 543:
```suggestion
            logError('InlineCompletionItemProvider:formatCompletion:error', error.message, {
                verbose: { error },
            })
```

adding the stack here will always make it verbose and the `verbose` part wouldn't add anything more to it I think

vscode/src/configuration.ts
```diff
@@ -98,6 +98,7 @@ export function getConfiguration(config: ConfigGetter = vscode.workspace.getConf
             CONFIG_KEY.autocompleteCompleteSuggestWidgetSelection,
             true
         ),
+        autocompleteFormatOnAccept: config.get(CONFIG_KEY.autocompleteFormatOnAccept, true),
```
philipp-spiess, line 101: Another place where we define a default. 🤔

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -6,7 +6,8 @@ import { FeatureFlag, featureFlagProvider } from '@sourcegraph/cody-shared/src/e
 import { RateLimitError } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
 import { startAsyncSpan } from '@sourcegraph/cody-shared/src/tracing'
 
-import { logDebug } from '../log'
+import { logDebug, logError } from '../log'
+import { getEditorInsertSpaces, getEditorTabSize } from '../non-stop/utils'
```
philipp-spiess, line 10: Should we move these to a different place to make it clearer that this is not only used in `non-stop`?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -489,6 +500,50 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+
+            const insertedLines = lines(autocompleteItem.analyticsItem.insertText)
+            const endPosition =
+                insertedLines.length <= 1
+                    ? new vscode.Position(position.line, currentLinePrefix.length + insertedLines[0].length)
+                    : new vscode.Position(position.line + insertedLines.length - 1, insertedLines.at(-1)!.length)
+            const rangeToFormat = new vscode.Range(position, endPosition)
+
+            const formattingChanges = await vscode.commands.executeCommand<vscode.TextEdit[] | undefined>(
+                'vscode.executeFormatRangeProvider',
```
umpox, line 519: It's awesome that we can use `formatRangeProvider` here, this seemed to be broken when we tried to use it for edits but seems to be working well now. Maybe we can switch edits to this now too.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -489,6 +500,50 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+                rangeToFormat,
+                {
+                    tabSize: getEditorTabSize(document.uri),
+                    insertSpaces: getEditorInsertSpaces(document.uri),
+                }
+            )
+
+            const formattingChangesInRange = (formattingChanges || []).filter(change =>
+                rangeToFormat.contains(change.range)
+            )
```
umpox, line 530: Is this necessary, given we're calling `executeFormatRangeProvider`?

olafurpg (commented):
Very cool! Can we update the vscode-shim to disable this feature by default in the agent?

philipp-spiess (approved):
The changes look good but I’m unsure wether this is a good default experience. Also here are two things I found while playing with it:

1.  I found a bug when there is only one space between the completion and the text in the document but when I save it, there are 4 spaces:

    https://github.com/sourcegraph/cody/assets/458591/dfaa3bf8-85f5-4df3-919a-9b8c93494659

2. It does seem formatting creates a new "redo" snapshot so in order to undo a tab insertion, you now have to do know if you have had a reformat occur or not in order to press cmd+z once or twice

That said, I didn't notice it a lot so this is definitely more stable than I thought. Also I noticed when prettier would insert new characters (e.g. `;` before a statement or whatever shenanigans prettier does), ther format on accept behavior would _not_ do that. Not sure if this means it's using a different formatter?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -489,6 +500,50 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+
+            if (formattingChangesInRange.length !== 0) {
+                const edit = new vscode.WorkspaceEdit()
+                for (const change of formattingChangesInRange) {
+                    edit.replace(document.uri, change.range, change.newText)
```
umpox, line 535: One side effect of this is that it will push a new change to the undo stack. So to undo a completion that has been formatted you will need to undo twice (once for the format change, once for the undo insertion.)

I think you can get around this by using `vscode.TextEditor['edit']` and overriding the undo stops

```
            if (formattingChangesInRange.length !== 0) {
                await vscode.window.activeTextEditor?.edit(
                    edit => {
                        for (const change of formattingChangesInRange) {
                            edit.replace(change.range, change.newText)
                        }
                    },
                    { undoStopBefore: false, undoStopAfter: true }
                )
            }
```

This should keep the formatting and the insertion as a single undo item I think, worth testing though

vscode/src/completions/inline-completion-item-provider.ts
umpox, line 519: May be worth adding some telemetry to check how long this function takes, some formatters can be pretty buggy/slow from our experience with edits. Maybe we could just log the time taken and the language ID?

umpox (commented):
Awesome! Gave it a try with some weird Prettier config and it works really well.

It'd be cool if we could format the completion this way, so it is shown to the user with the correct formatting - but I think this isn't possible AFAIK - the code has to be in a document to be correctly formatted.

I added some comments, there's one thing that I think is definitely worth considering and it's around the undo/redo behaviour

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: @umpox, it's still broken. I wanted to ask you specifically about it but forgot to mention it in my Slack message. I still filter our edits outside of the `rangeToFormat`. Looks like you had the same experience?

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: See my comment above 🙃

vscode/src/completions/inline-completion-item-provider.ts
umpox: Ahh, IIRC it wasn't giving me any edits before. My thinking was that the formatter might have to support this so maybe it only works with certain formatters. Maybe they changed it to proxy through to `formatDocumentProvider` when not supported?

vscode/src/configuration.ts
valerybugakov: Yep, we need to clean this up. Defaults should be set in `configuration.ts` only.

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Yep, we need to clean this up. Defaults should be set in `configuration.ts` only.

lib/shared/src/configuration.ts
valerybugakov: I think about experimental settings as something that does not produce consistent results/improvements. E.g., our latency tweaks or enlarged context windows. With the format on accept functionality, everything is straightforward, and users can change its behavior by changing the formatter or its settings. 

Agree that we can set it to `false` by default to test it properly and collect metrics first.

vscode/package.json
valerybugakov: I will switch to `false` by default to dogfood it first.

lib/shared/src/configuration.ts
philipp-spiess: Makes sense 👍

agent/src/agent.ts
valerybugakov: Disabled formatting inside the agent.

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Interesting. Did you test it with prettier, and it did not work? I tried it with `prettier` and black formatter for python today, and it worked identically in both cases.

Also, I found no issues related to this problem in the VS Code repo. I should file one to see what they think about it.

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Works like a charm!

valerybugakov:
@umpox, @philipp-spiess, I added a telemetry event to track the duration of the format command and removed format-edits from the undo stack.

valerybugakov:
@philipp-spiess, I cannot repro your example with extra added spaces on the latest commit in this branch. I tried generating a completion in the same file and line, but it works fine for me on the StarCoder model. What is your setup? (model, formatter)

Let's discuss it in Slack, and I can create a follow-up PR to fix this problem if we find a way to repro it repro it consistently.

philipp-spiess:
> @philipp-spiess, I cannot repro your example with extra added spaces on the latest commit in this branch. I tried generating a completion in the same file and line, but it works fine for me on the StarCoder model. What is your setup? (model, formatter)

I can still repro:  That's fireworks with prettier as the default formatter. 

https://github.com/sourcegraph/cody/assets/458591/e0f8ecab-973b-4c14-8858-d70498d93820

umpox:
@philipp-spiess @valerybugakov Kinda looks like the formatter is skipping a `\n` at the very start. Maybe `rangeToFormat.contains` doesn't return true if it's on the boundary of a range?

valerybugakov:
@philipp-spiess, TY for another repro example. I fixed the issue by making the `rangetoFormat` start from the beginning of the line instead of the cursor position.

Issue #2331, opened 2023-12-13
bug: chat: current file context listed as absolute path and isn't clickable


beyang:
![Image](https://github.com/sourcegraph/cody/assets/1646931/2b9d769b-0a82-41cd-87f3-ddb030c26917)

PR #2344, opened 2023-12-13
Chat Context: Provide relative path as current file

umpox:
## Description

The `fileUri` provided will be absolute. Instead, we always want to use a better Uri based on the relative file path that we already know.

closes https://github.com/sourcegraph/cody/issues/2331

## Test plan

1. Start chat, ask "what does this file do?"
2. Check context
3. Confirm current file is shown relative to users editor directory root.
4. Confirm context is correctly used
5. Confirm current file is clickable


philipp-spiess (approved):
YES YES YES

toolmantim:
This is great! Thank you!

Issue #2260, opened 2023-12-11
bug: Individual chat panel titles don't match the titles in the chat list


taylorsperry:
### Version

v0.19.1702220762

### Describe the bug

When I reopen an old chat, the title of the chat in the chat list doesn’t match the title of the reopened individual chat, which now appears to be the date/time I created the chat, though the time is not my local time (local time is 10:26am, 10:34am). 

![Image](https://github.com/sourcegraph/cody/assets/43555476/30b375d4-8530-4966-8129-c42173676b80)

### Expected behavior

Chat titles should match.

PR #2345, opened 2023-12-13
Chat: Fix chat panel titles

umpox:
## Description

closes https://github.com/sourcegraph/cody/issues/2260

Fixes the title of reopened chat panels.

We were using the `chatId` by accident before, due to a mistake when calling this function

> [!NOTE]  
> We could just not pass the `chatId` but we have two implementations of the ChatPanelProvider right now. One relies on the `chatId` and one doesn't. Will start a thread on Slack to discuss this as it can be quite confusing to debug!

### Before

image

### After

image

## Test plan

1. Create a chat and ask a question
2. Close the chat
3. Reopen the chat by selecting it from the chat history in the sidebar
4. Inspect the title


philipp-spiess (approved):
BLISS

deepak2431:
Thanks for making this PR, Tom. I just made a push to my branch, and was trying to create a PR but saw it is fixed :)

This was a very small fix, but took some time for me to debug, haha!

PR #2348, opened 2023-12-13
Chat: Scope history to user account

### LINKED ISSUE #2261: bug: Chats persist between accounts
### Version

v0.19.1702220762

### Describe the bug

When I switch between by dotcom account and my enterprise account, I’m surprised that my chats persist across the two. Presumably, a user who has two accounts (one enterprise, one free/personal) is using them for different things in different codebases, and the LLMs used to create those chats may be different. 

### Expected behavior

When logged in to a given account, I should only see that chats that I created while logged into that account. Ie, When I'm logged into s2, I should see the chats that I created as an enterprise user logged into sourcegraph.sourcegraph.com; when I'm logged into dotcom, I should see the chats that I created as a community user. 

### Additional context

_No response_

philipp-spiess:
Closes #2261

This PR scopes the chat history storage to the user account (we use the email + server URL as the key for this). That way, when switching accounts, you will not see chat histories from other accounts (potentially showing work stuff in your free time).

There's one caveat that I have depicted in the test plan. If you keep a window open while you switch accounts and _send a new message to it_, we will _fork_ the history, so it will appear in your other account with the updated messages but kept as-is in your other account. I think this behavior makes sense though and we should not try to optimize for it (i.e I don't think we should close chat windows, for example, when you switch accounts).

⚠️ I plan to land this after cutting the 1.0.2 patch release and deploy it onto the insider build right away to test this some more.

## Test plan

- Log in to a user from dotcom
- Chat!
- Switch to another account, e.g. S2
- More chatting!
- ...repeat

https://github.com/sourcegraph/cody/assets/458591/a42f1a4b-da0c-4c4e-839a-c3717d4c8bbb


toolmantim:
Too risky for GA? I imagine it's a deep enough change it would need a bunch of thorough manual QA.

philipp-spiess:
@toolmantim yeah, something else came up anyways 😬  This PR did not get ready in time anyways

valerybugakov (commented):
> I don't think we should close chat windows, for example, when you switch accounts

I would expect that because it's the learned behavior from all the chat products. If I switch accounts on Facebook or the ChatGPT website, I won't be able to see currently opened chats. Curious about the rationale here. 

> If you keep a window open while you switch accounts and send a new message to it, we will fork the history

I bumped into this scenario during local testing. I have multiple chat windows opened, and after I switched an account, all of the chat windows are still there. Then, if I reload the VS Code instance, all the chat windows are there without the messages _but_ with titles from the previous account. It does not feel right. I would expect Cody to wipe all previous chats to keep my data secure and private if I log out.

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -156,3 +198,11 @@ export class LocalStorage {
  * The underlying storage is set on extension activation via `localStorage.setStorage(context.globalState)`.
  */
 export const localStorage = new LocalStorage()
+
+function getKeyForAuthStatus(authStatus: AuthStatus): `${string}-${string}` {
```
valerybugakov, line 202: Would it be nice to extract ``${string}-${string}`` into a named type?

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -67,34 +69,74 @@ export class LocalStorage {
...
+            return { chat: {}, input: [] }
+        }
+
+        const key = getKeyForAuthStatus(authStatus)
+
+        // For backwards compatibility, we upgrade the local storage key from the old layout that is
+        // not scoped to individual user accounts to be scoped instead.
+        if (history && !isChatHistoryV2(history)) {
+            history = {
+                [key]: history as UserLocalHistory,
```
philipp-spiess, line 87: Need to make sure `chat` and `user` is not deleted here to not break when downgrading, see https://sourcegraph.slack.com/archives/C05AGQYD528/p1702954449990249

toolmantim:
Yeah, to echo @valerybugakov's comment… I would have expected it to close chats when logging out. Making copies of chats into new accounts seems problematic and confusing. Curious to hear why.

philipp-spiess:
No good reason, it was just "the way it worked" and I didn't find it too weird. But agree with both of you, when we log out we should close all open cody tabs. Consider it done

philipp-spiess:
@valerybugakov @toolmantim I changed the behavior so that chat panels clear when the authenticated user changes. This indeeds feels much better. I attached a video as the test plan. 

@valerybugakov I have also fixed the issue where when you test this PR you can no longer downgrade 👍

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -67,34 +71,80 @@ export class LocalStorage {
...
+            null
+        )
+        if (!history) {
+            return { chat: {}, input: [] }
+        }
+
+        const key = getKeyForAuthStatus(authStatus)
+
+        // For backwards compatibility, we upgrade the local storage key from the old layout that is
+        // not scoped to individual user accounts to be scoped instead.
```
valerybugakov, line 86: 💜

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -156,3 +206,11 @@ export class LocalStorage {
...
  */
 export const localStorage = new LocalStorage()
+
+function getKeyForAuthStatus(authStatus: AuthStatus): ChatHistoryKey {
+    return `${authStatus.endpoint}-${authStatus.primaryEmail}`
+}
+
+function isChatHistoryV2(history: { [key: ChatHistoryKey]: UserLocalHistory } | UserLocalHistory): boolean {
+    return !!Object.keys(history).find(k => k.includes('-'))
+}
```
valerybugakov, line 216: Let's add a comment explaining what V2 means for chat history or change the function name to describe that. Looking at this code from the PR diff, it's obvious but might need to be clarified for folks reading the code later in their editors.

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -67,34 +71,80 @@ export class LocalStorage {
...
+                // For backward compatibility, we do not want to delete the `chat` and `input` keys.
+                // As otherwise, downgrading to a prior version would completely block the startup
+                // as the client would throw.
+                chat: {},
+                input: [],
+            }
+
+            // We use a raw write here to ensure we do not _append_ a key but actually replace
+            // existing `chat` and `input` keys.
+            void this.storage.update(this.KEY_LOCAL_HISTORY, history)
```
valerybugakov, line 100: Do we want to `try/catch` or `await` here, similar to how it's done in `setChatHistory`?

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -67,34 +71,80 @@ export class LocalStorage {
...
+        }
+
+        if (!Object.hasOwn(history, key)) {
+            return { chat: {}, input: [] }
+        }
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore: TS seems to be consufed about using `key` as the key here
+        return history[key] as UserLocalHistory
```
valerybugakov, line 108: This can be fixed by adding type-casting the backward compatible history object and converting `isChatHistoryV2` to a type-guard:

```ts
type AccountBasedChatHistory = {
  [key: ChatHistoryKey]: UserLocalHistory
  
  // For backward compatibility, we do not want to delete the `chat` and `input` keys.
  // As otherwise, downgrading to a prior version would completely block the startup
  // as the client would throw.
  chat: UserLocalHistory
  input: []
}

public getChatHistory(authStatus: AuthStatus): UserLocalHistory {
  history = {
      [key]: history,
      chat: {},
      input: [],
  } as AccountBasedChatHistory

  return history[key] 
}

function isChatHistoryV2(history: AccountBasedChatHistory | UserLocalHistory): history is AccountBasedChatHistory {
    return !!Object.keys(history).find(k => k.includes('-'))
}

**Review by valerybugakov — APPROVED** [review]
Thanks for addressing all the issues. It works like a charm now!

File: vscode/src/services/LocalStorageProvider.ts (hunk 5 above)
Comment (philipp-spiess) [reply vscode/src/services/LocalStorageProvider.ts:100]: That function is non-blocking right now and I wanted to avoid changing the API (since it is called in some constructors AFAIK). We definitely don't want to silence the error though but I think having it write asynchronously is fine. The worst part we call the same function again and the state is not updated yet but that would simply call the same update again and not destruct anything (provided writes are atomic but if they are not it would be v strange I guess?).

File: vscode/src/services/LocalStorageProvider.ts (hunk 6 above)
Comment (philipp-spiess) [reply vscode/src/services/LocalStorageProvider.ts:108]: Screenshot 2023-12-21 at 14 40 06
 
 😢

Comment (philipp-spiess) [reply vscode/src/services/LocalStorageProvider.ts:108]:
```
            history = {
                ...{ [key]: history },
                ...{
                    chat: {},
                    input: [],
                },
            } satisfies AccountKeyedChatHistory
```

this works lol but this does not?


```
            history = {
                [key]: history,
                chat: {},
                input: [],
            } satisfies AccountKeyedChatHistory
```

TS never fails to surprise

Comment (philipp-spiess) [reply vscode/src/services/LocalStorageProvider.ts:108]: Screenshot 2023-12-21 at 14 44 27
Doesn't help with the initial issue either though

Comment (valerybugakov) [reply vscode/src/services/LocalStorageProvider.ts]: Is my TS server stale or 🤔

Comment (philipp-spiess) [reply vscode/src/services/LocalStorageProvider.ts]: @valerybugakov I applied the file from the gist and 😬  

```
vscode/src/services/LocalStorageProvider.ts:95:23 - error TS2352: Conversion of type '{ [x: string]: [] | UserLocalHistory | {}; chat: {}; input: []; }' to type 'AccountBasedChatHistory' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property 'chat' are incompatible.
    Type '{}' is missing the following properties from type 'UserLocalHistory': chat, input

 95             history = {
                          ~
 96                 [key]: history,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
...
 98                 input: [],
    ~~~~~~~~~~~~~~~~~~~~~~~~~~
 99             } as AccountBasedChatHistory
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.

 ELIFECYCLE  Command failed with exit code 2.
```

valerybugakov: 🤷‍♂️

PR #2355, opened 2023-12-13
telemetry-v2: generate timestamps locally for sourcegraph 5.2.5+

bobheadxi:
Adds conditional support for https://github.com/sourcegraph/sourcegraph/pull/58944, available in 5.2.5+ (https://github.com/sourcegraph/sourcegraph/commit/364f197a9e303018f18615f6edb13237d41a4ddc), which mitigates issues with out-of-order events caused by the timestamp being only rendered serverside.

V1 events do not support this, and there's currently no plan to update V1 to support it.

## Test plan

Unit tests on the conditional exporter.

Manual test with `pnpm dev`: 

```
█ telemetry-v2: recordEvent (no-op): cody.extension/savedLogin: {"parameters":{"version":0},"timestamp":"2023-12-13T23:11:35.612Z"}
```

With export enabled, events show up as expected, in order, e.g. on rctest.sgdev.dev:

image


dominiccooney (approved):
Looks good!

PR #2358, opened 2023-12-14
VSCode: Fix local embeddings e2e test to specifically check that *embeddings* are indexed

dominiccooney:
When we enabled search, Search — Indexed started appearing in the enhanced context selector. This will cause spurious passes in `local-embeddings.test.ts` which was just looking for the text "Indexed"; and spurious failures when the word "Indexed" appears twice, for search *and* embeddings.

Fix the test by looking specifically for "Embeddings — Indexed".

## Test plan

Note, this passed locally before this change, so check that we did not break it. But the real test is whether CI results are accurate now.

```
pnpm test:e2e -- local-embeddings.test.ts
```


toolmantim (approved):
🚀

PR #2359, opened 2023-12-14
VSCode: Roll cody-engine to v5.2.12792

### LINKED ISSUE #2328: bug: Continue Indexing results in "OS error 5" on Windows
### Version

Main f7cb9e5184c18f1a56b114b8a751f713079514f1

### Describe the bug

This affects users on Windows using local embeddings who encountered an error while creating the index. It prevents them from finishing the indexing process, but they can use the partially completed index for queries.

Repro steps:

1. Open a repo that doesn't have embeddings and start a chat.
2. In the enhanced context selector, enable embeddings.
3. Kill the cody-engine process, hanging the embeddings indexing process and leaving a half-completed index.
4. Reload the window and start a chat. The status bar item will appear warning you the index is incomplete.
5. Click, then select Continue Indexing.

An "os error 5" will result.

### Expected behavior

Expected behavior: Indexing should continue.

Actual behavior: "os error 5" and repeated attempts to continue indexing say the index cannot be found.

### Additional context

This bug is specific to Windows. cody-engine dropping the old index isn't fast/coordinated enough to prevent a sharing violation.

cody-engine drops the loaded index before trying to resume indexing it, but this should be better coordinated. The index file can be opened in a sharing mode in the first instance and reopened in exclusive mode to update it.

dominiccooney:
Fixes #2328.

## Test plan

CI.

The "os error 5" issue specifically is covered by tests upstream.

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -8,6 +8,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
+- Context: The "Continue Indexing" button works on Windows. [issues/2328](https://github.com/sourcegraph/cody/issues/2328)
+
 ### Changed
 
 ## [0.18.6]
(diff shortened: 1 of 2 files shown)


kalanchan (approved):
lets go!

Issue #2311, opened 2023-12-12
bug: VS Code upsell telemetry bugs


chenkc805:
### Version

v0.19.1702393628

### Describe the bug

Bug:
- For the `CodyVSCodeExtension:openLLMDropdown:clicked` event, it fires even when:
  - the dropdown is disabled
  - the user closes the dropdown
- For the `CodyVSCodeExtension:upsellUsageLimitCTA:shown` event, it the tier is "enterprise" even though I'm on the Free tier.
![Screenshot 2023-12-12 at 12 04 19 PM](https://github.com/sourcegraph/cody/assets/3654603/8f575e57-1ba1-4f85-80ff-7867a28da854)
- For users on the Pro tier, when they hit the abuse limit, we log `CodyVSCodeExtension:upsellUsageLimitCTA:shown`, when instead the event should be `CodyVSCodeExtension:abuseUsageLimitCTA:shown`. [See row 6 here](https://docs.google.com/spreadsheets/d/1QXnrHkgGLXdbfHEhI_VjtDTy-mCtnK1qri3IMQu7rNk/edit#gid=1704312181&range=6:6)
![Screenshot 2023-12-12 at 12 08 27 PM](https://github.com/sourcegraph/cody/assets/3654603/a3fa5d3d-dfec-4b76-bbbb-ee73ba38c547)

### Expected behavior

Fix:
- For the `CodyVSCodeExtension:openLLMDropdown:clicked` event, it fires only when:
  - the dropdown is open-able AND opens
  - Don't fire if the dropdown is disabled, if the dropdown is closing, or if the user clicks and LLM which causes the dropdown to close. 
- For the `CodyVSCodeExtension:upsellUsageLimitCTA:shown` event, ensure the correct "tier" is being passed into the payload. The tier should be the tier the user is on. 
- For users on the Pro tier, when they hit the abuse limit for chat/commands or autocompletes and the abuse UI shows up, we should log `CodyVSCodeExtension:abuseUsageLimitCTA:shown`. [See row 6 here](https://docs.google.com/spreadsheets/d/1QXnrHkgGLXdbfHEhI_VjtDTy-mCtnK1qri3IMQu7rNk/edit#gid=1704312181&range=6:6)

### Additional context

Events for reference: https://docs.google.com/spreadsheets/d/1QXnrHkgGLXdbfHEhI_VjtDTy-mCtnK1qri3IMQu7rNk/edit#gid=1704312181

PR #2372, opened 2023-12-14
Chat: use the correct user tier in the upsell event

valerybugakov:
## Context

- Closes of https://github.com/sourcegraph/cody/issues/2311
- Use the correct user tier in the upsell event.
- Use the correct event when pro users hit the usage limit.

## Test plan

- For the `CodyVSCodeExtension:upsellUsageLimitCTA:shown` event, ensure the correct "tier" is being passed into the payload. The tier should be the tier the user is on.
- For users on the Pro tier, when they hit the abuse limit for chat/commands or autocompletes and the abuse UI shows up, we should log `CodyVSCodeExtension:abuseUsageLimitCTA:shown`

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -591,8 +591,10 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                 return
             }
 
+            const isEnterpriseUser = this.config.isDotComUser !== true
             const canUpgrade = error.upgradeIsAvailable
-            const tier = this.config.isDotComUser ? 'enterprise' : canUpgrade ? 'free' : 'pro'
+            const tier = isEnterpriseUser ? 'enterprise' : canUpgrade ? 'free' : 'pro'
+
             let errorTitle: string
             let pageName: string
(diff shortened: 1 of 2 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -591,8 +591,10 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                 return
             }
 
+            const isEnterpriseUser = this.config.isDotComUser !== true
             const canUpgrade = error.upgradeIsAvailable
-            const tier = this.config.isDotComUser ? 'enterprise' : canUpgrade ? 'free' : 'pro'
+            const tier = isEnterpriseUser ? 'enterprise' : canUpgrade ? 'free' : 'pro'
```
valerybugakov, line 596: @philipp-spiess, logically, this update makes sense to me, but I might not have some context or misinterpret it. Compare this code to the logic used in chat.

philipp-spiess: Whoopsie this was wrong before, yeah.

philipp-spiess (approved):

PR #2380, opened 2023-12-14
Agent: prepare to fix chat parity

olafurpg:
This PR makes progress towards fixing parity in chat for JetBrains and Neovim. We update the test case to reproduce the warning about missing context and fix a few errors that happen in this code path. The issue is still there but it will be easier to fix the root problem with this merged.


olafurpg:
Superseded by https://github.com/sourcegraph/cody/pull/2396

PR #2391, opened 2023-12-15
Handle connection close events with no prior errors or messages

philipp-spiess:
Welcome to callback hell. Even though we handle a bazillion cases here, we did never call any of the callbacks if:

- The connection opens
- We receive headers
- However, none of these headers indicate errors
- And we receive no body and instead closes

This adds some form of handling for this so the request doesn't appear stuck.

We should also make it so the backend indicates the error in the response somehow (c.f. https://sourcegraph.slack.com/archives/C05AGQYD528/p1702648140278529)

## Test plan

- git revert 2a00da01ae8b83ecd5450af7be2682c32edf3080 (to bring back the issue that beyang fixed)
- Create a chat message with GPT 3.5

Screenshot 2023-12-15 at 14 57 51


umpox (approved):
👍

PR #2395, opened 2023-12-15
Chat: Honor enterprise token limits

philipp-spiess:
As the title suggests.

## Test plan

Screenshot 2023-12-15 at 15 24 22

- `sg start` with the openai config from dev-private
- chat with cody


chwarwick (approved):
Looks good thanks!

PR #2396, opened 2023-12-15
Agent: add support for new VS Code chat UI 

olafurpg:
Previously, the agent only supported chat via recipes, which had an
increasingly large feature gap with chat in VS Code. This PR extends
the agent protocol with new high-level endpoints to support interacting
with the enw chat UI:

- `chat/new`: requests to start a new chat. Returns an ID number to
  reference this chat session.
- `chat/submitMessage`: request to submit a chat message. The agent
  streams progress via a separate `webview/postMessage` notification,
  and responds to this request when the last message has finished
  streaming.

The endpoints above are "high-level", meaning they are implemented on
top of more low-level endpoints that directly translate VS Code API
calls:

- `webview/create`: request from the server to the client to create a
  new webview. Wrapper for `vscode.window.createWebviewPanel`.
- `webview/postMessage`: notification from the server to the client to
  send a JSON message to the webview.
- `webview/receiveMessage`: notification from the client to the server to
  send a JSON message from the webview.

This design allows us to both support a semi-dropin replacement of the
recipe API while additionally allowing clients to fully interact with
the low-level webview API. In the short term, the JetBrains plugin can
quickly migrate to the high-level APIs. In the long-term, it probably
makes sense for JetBrains to use webviews and use the low-level API.
## Test plan

See updated test cases.


olafurpg:
BTW, the HTTP recording doesn't need to be this big. I will make it smaller next week. It's huge right now because I had `RECORD_IF_MISSING=true` for a very long time, and it doesn't clean up unused recordings.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -708,8 +708,8 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
                 },
                 onError: error => {
                     this.cancelInProgressCompletion()
-                    typewriter.stop()
                     callbacks.error(lastContent, error)
+                    typewriter.stop()
```
philipp-spiess, line 712: What's the intention behind this change?

agent/src/index.test.ts
```diff
@@ -133,19 +159,23 @@ describe('Agent', () => {
...
+    it('accepts textDocument/didOpen notifications', () => {
         const content = 'function sum(a: number, b: number) {\n    \n}'
         client.notify('textDocument/didOpen', {
             uri: uri.toString(),
             content,
             selection: { start: { line: 1, character: 0 }, end: { line: 1, character: 0 } },
         })
+    })
+
+    // TODO: fix this test
```
philipp-spiess, line 177: What's broken?

philipp-spiess (approved):
Makes sense to me!

agent/src/index.test.ts
olafurpg: It's working correctly now, I disabled it while hacking on other things.

olafurpg (commented):
@philipp-spiess Thank you for the review! Great points, I've addressed both comments

olafurpg:
Tests are failing because it's not replaying from the HTTP recording for the chat prompt. I don't have time to investigate this problem until after New Years, maybe briefly around December 28/29th. I tried to fix this problem by running the test from a temporary directory but the prompt still seems to be pulling in files from the Cody repo. The root fix is probably small.


```
+   \"body\": \"{\\\"temperature\\\":0.2,\\\"maxTokensToSample\\\":1000,\\\"topK\\\":-1,\\\"topP\\\":-1,\\\"model\\\":\\\"anthropic/claude-2.0\\\",\\\"messages\\\":[{\\\"speaker\\\":\\\"human\\\",\\\"text\\\":\\\"You are Cody, an AI coding assistant from Sourcegraph.\\\"},{\\\"speaker\\\":\\\"assistant\\\",\\\"text\\\":\\\"I am Cody, an AI coding assistant from Sourcegraph.\\\"},{\\\"speaker\\\":\\\"human\\\",\\\"text\\\":\\\"Use the following text from file `/vscode/doc/web.md`
```

mkondratek:
closing in favour of https://github.com/sourcegraph/cody/pull/2457

Issue #2388, opened 2023-12-15
bug: @-mentioning files on Windows shows markup instead of clickable links


DanTup:
### Version

main 2a00da01

### Describe the bug

When mentioning files on Windows, I no longer see clickable links, but markup instead:

https://github.com/sourcegraph/cody/assets/1078012/cda79ba3-67dd-4a73-8f82-4d3e4a3f007b

### Expected behavior

The filename should appear as a clickable link and no markup visible.

DanTup:
Seems to be broken by the backslashes, but I can't find any way to encode them in a way that comes through the other end correctly. I've opened https://github.com/microsoft/vscode/issues/200965 assuming this could be a VS Code bug, or that there's at least some missing documentation about how these args should be encoded.

PR #2398, opened 2023-12-15
Fix encoding of file paths in chat links

DanTup:
This fixes https://github.com/sourcegraph/cody/issues/2388, however the encoding format feels odd to me. I've opened https://github.com/microsoft/vscode/issues/200965 to get some clarification about whether this is correct (and if so, hopefully it can be better documented).

https://github.com/sourcegraph/cody/assets/1078012/53893a0b-a5fd-4398-85c7-de95911767b2

@kalanchan 

## Test plan

- Open a chat
- Mention some files with `@`
- Ensure clicking the filenames in the chat message correctly open the files (note: the filenames in the collapsed set are handled differently, the issue was specifically with the converted links in the chat message)
- Also: verify the changes to tests for the new expected format are sensible and match what's discussed in https://github.com/microsoft/vscode/issues/200965 (at the time of writing there has been no response, so we may want to wait for one before merging this?)


dominiccooney (approved):
Nice work.

Issue #2399, opened 2023-12-15
bug: Clicking filenames in chat context list does not work (on Windows?)


DanTup:
### Version

main 2a00da01

### Describe the bug

Clicking this link doesn't work for me:

![image](https://github.com/sourcegraph/cody/assets/1078012/af82786b-2779-4819-8164-fb7c6b08ba87)

And I see this error:

```
rejected promise not handled within 1 second: Error: Invalid argument 'uriOrString' when running 'vscode.open', received: c:\Dev\Test Projects\dartcode_4895\lib\main.dart 
extensionHostProcess.js:147
stack trace: Error: Invalid argument 'uriOrString' when running 'vscode.open', received: c:\Dev\Test Projects\dartcode_4895\lib\main.dart 
	at c:\Users\danny\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:141:138142
	at Array.map (<anonymous>)
	at p.registerCommand.description (c:\Users\danny\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:141:138103)
	at d.h (c:\Users\danny\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:141:140094)
	at d.g (c:\Users\danny\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:141:139086)
	at d.executeCommand (c:\Users\danny\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:141:138993)
	at Object.executeCommand (c:\Users\danny\AppData\Local\Programs\Microsoft VS Code\resources\app\out\vs\workbench\api\node\extensionHostProcess.js:158:21333)
	at openFilePath (c:\Dev\Sourcegraph\cody\vscode\dist\extension.node.js:164413:26)
	at SimpleChatPanelProvider.onDidReceiveMessage (c:\Dev\Sourcegraph\cody\vscode\dist\extension.node.js:167591:15)
	at f.value (c:\Dev\Sourcegraph\cody\vscode\dist\extension.node.js:167490:79)
```

On the surface, the path looks correct.

### Expected behavior

File should open, no error in console.

DanTup:
> Could it be due to change in VSCode from paths to URI?

[...]

PR #2400, opened 2023-12-15
Fix clicking on context files with absolute paths on Windows

DanTup:
Fixes https://github.com/sourcegraph/cody/issues/2399 by removing a duplicate call to open a file which [expects a file URI](https://code.visualstudio.com/api/references/commands#:~:text=of%20InlineValue%20objects-,vscode.open,-%2D%20Opens%20the%20provided) but was given a path, and only tries to join a path with the workspace URI if it's a relative path.

https://github.com/sourcegraph/cody/assets/1078012/787806c8-05e2-43a4-a6fe-59e574298092

I'm not certain if there really should be absolute paths in this list if things are inside my workspace, but it probably makes sense for this code to tolerate them.

@kalanchan 

## Test plan

- `@`-mention a file on Windows
- expand the list of context files
- click on an absolute path


vscode/src/services/utils/workspace-action.ts
```diff
@@ -22,14 +24,18 @@ export async function openFilePath(
     currentViewColumn?: vscode.ViewColumn,
     range?: ActiveTextEditorSelectionRange
 ): Promise<void> {
-    void vscode.commands.executeCommand('vscode.open', filePath)
```
DanTup, line 25: @abeatrix looks like this call was added in 343f626e6ec106fb75a9a586e0b2dc00d5730d7c but as far as I can tell it is redundant (because the code below opens the file).

My fix here is removing it because I don't think it's doing anything, but if there's a reason to have it we can also fix by changing it from a path to a file URI.

vscode/src/services/utils/workspace-action.ts
```diff
@@ -22,14 +24,18 @@ export async function openFilePath(
...
-    if (!workspaceRootUri) {
-        throw new Error('Failed to open file: missing workspace')
+    filePath = 'lib\\main.dart'
```
dominiccooney, line 27: ???

dominiccooney (changes requested):
There's a hard coded path in here?

vscode/src/services/utils/workspace-action.ts
DanTup: 😱

Thanks for catching this. I've added a test that clicks one of these links and verifies an editor tab opens. The test is broken before removing this and passing on Windows when I removed this.

However, on Mac/Linux it's failing because apparently we have a full path without the leading slash:

image

Looking into that..

DanTup:
Ok, I tracked down the Linux/macOS test failures down to this code which is removing the leading slash from absolute file paths:

https://github.com/sourcegraph/cody/blob/bcd49b3694082f529d47681354310a47ea99d195/vscode/src/chat/chat-view/chat-helpers.ts#L74-L77

It's not clear to me what the purpose of this is.. on Windows it's not being done and the tests are passing (although it's not clear to me if we wanted absolute or relative paths in the context file list). On Linux/Mac this is stripping the first `/` which makes it seem like a relative path, but it's not clear what it's relative to).

I think this needs some input from someone who is more familiar with this (perhaps @abeatrix ?). And I know I've said this before, but I think we need to get better at being very explicit with comments on all "filename-like" fields what they really are... Are they URI paths (always forward slashes), native paths (backslashes on Windows), absolute, relative (if so, relative to what).

DanTup:
FWIW if I just completely remove the code noted above, then Mac behaves the same as Windows (we show an absolute path, and clicking it works):

Screenshot 2023-12-22 at 12 13 23

I don't know what implications there are of doing this (for files that come from places other than these `@`-mentions) though.

DanTup:
@beyang the code quoted above was added in https://github.com/sourcegraph/cody/commit/c958112f6a3f257e0010a1db1a9b318cfaafb472 but it's a big change so I don't know if it was just moved/copied from elsewhere. If you authored it, do you remember what the reason was for just removing the first `/`?

It's causing issues because it looks like a relative path not absolute, and since Windows paths don't start with `/` it means the paths are inconsistent across the platforms.

DanTup:
The test here will be impacted by the question at https://github.com/sourcegraph/cody/pull/2408 - it turns out user-selected files weren't intended to show up here, so the test added for this issue won't work once they're filtered out (when https://github.com/sourcegraph/cody/pull/2408 lands).

Depending on the response there, we may need to change the test for this (assuming the original issue is even still valid with that change).

DanTup:
> The test here will be impacted by the question at https://github.com/sourcegraph/cody/pull/2408 - it turns out user-selected files weren't intended to show up here, so the test added for this issue won't work once they're filtered out (when https://github.com/sourcegraph/cody/pull/2408 lands).

This decision was reverted in https://github.com/sourcegraph/cody/pull/2679 so user-mentioned files will show up again. However this change might be impacted by https://github.com/sourcegraph/cody/pull/2653 so I'll come back to this after that merges.

DanTup:
This change is no longer necessary, the big was already fixed on main by changes @sqs made to tidy up use of paths vs URIs etc. :-)

https://github.com/sourcegraph/cody/assets/1078012/928dd4d4-a566-4816-99ec-ad65ed2ac682

Issue #2402, opened 2023-12-15
bug: mentioned files content not always being included


DanTup:
### Version

main 2a00da01

### Describe the bug

I mentioned two files, but they were not both included:

![image](https://github.com/sourcegraph/cody/assets/1078012/cfc30a6c-dfb1-41b5-988f-b327dc3f653e)

The exported chat shows:

```js
"usedContextFiles": [
	{
		"fileName": "c:\\Dev\\Test Projects\\cody_testing\\analysis_options.yaml",
		"source": "user",
		"content": "# This file configures the static analysis results for your project (errors,\n# warnings, and lints).\n#\n# This enables the 'recommended' set of lints from `package:lints`.\n# This set helps identify many issues that may lead to problems when running\n# or consuming Dart code, and enforces writing Dart using a single, idiomatic\n# style and format.\n#\n# If you want a smaller set of lints you can change this to specify\n# 'package:lints/core.yaml'. These are just the most critical lints\n# (the recommended set includes the core lints).\n# The core lints are also what is used by pub.dev for scoring packages.\n\ninclude: package:lints/recommended.yaml\n\n# Uncomment the following section to specify additional rules.\n\n# linter:\n#   rules:\n#     - camel_case_types\n\n# analyzer:\n#   exclude:\n#     - path/to/excluded/files/**\n\n# For more information about the core and recommended set of lints, see\n# https://dart.dev/go/core-lints\n\n# For additional information about configuring this file, see\n# https://dart.dev/guides/language/analysis-options\n"
	},
	{
		"fileName": "README.md",
		"content": "A sample command-line application with an entrypoint in `bin/`, library code\nin `lib/`, and example unit test in `test/`.\n"
	}
],
```

It includes the README and one of the files I mentioned, but not the other file mentioned.

### Expected behavior

Any mentioned file should be included.

DanTup:
With some debugging, I ended up here. We have some code that tries to de-dupe items here, and it's building an ID by calling `toString()` on a URI, but it's `[Object object]`, so every file mentioned in the chat looks like a dupe and it seems like we'll only keep the first.

![image](https://github.com/sourcegraph/cody/assets/1078012/272f9ef4-0c89-40ab-9b8b-8eb10e3e619f)

PR #2405, opened 2023-12-15
Ensure all explicitly-mentioned files are included in chats

DanTup:
Because the search results come from inside the webview, the URIs are not real vscode.Uri instances. The code that computes IDs for each context to de-dupe them uses `contextItem.uri.toString()` which would be `[Object object]`.

image

This converts them back to real `vscode.Uris` which fixes the de-dupe check.

Fixes https://github.com/sourcegraph/cody/issues/2402

(fyi @kalanchan )

## Test plan

- Run `pnpm test:e2e at-file` to run the automated test which verifies after asking about two files that the context has a count of two
- Or manually mention two files in chat and ensure both appear in context and that the model is aware of both file contents (or export the chat and confirm that both files appear in the context)


dominiccooney (approved):
Brilliant.

Issue #2244, opened 2023-12-10
bug: Tooltip on context files says "included via embeddings" even if embeddings not in enhanced context status list


dominiccooney:
### Version

Main 9e06cafee3fb60b7e265ced935b2e55b57413838

### Describe the bug

@toolmantim Via Slack: https://sourcegraph.slack.com/archives/C05AGQYD528/p1702248113730479?thread_ts=1702246529.082959&cid=C05AGQYD528

When context is included (probably via editor and/or the "readme" context added by simplified chats) the tooltip in the context list says "included via embeddings" even if embeddings is not involved.

### Expected behavior

Tooltip accurately reflects the context source, or at least doesn't misrepresent it.

dominiccooney:
@kalanchan points out this probably originates here:

https://sourcegraph.com/github.com/sourcegraph/cody@9e06cafee3fb60b7e265ced935b2e55b57413838/-/blob/vscode/src/chat/chat-view/chat-helpers.ts?L94

We need to plumb the context origin to that point and promulgate it.

PR #2408, opened 2023-12-15
Use correct file sources instead of always showing 'embeddings' in chat context

DanTup:
Fixes https://github.com/sourcegraph/cody/issues/2244

Screenshot 2023-12-15 182047

Screenshot 2023-12-15 182154

fyi @kalanchan 

## Test plan

- Select text in an editor and run a command like `/explain` that will include the context file from the editor
- Hover and ensure the tooltip says the file was from the editor, not embeddings
- Ask something that will cause the readme to be included automatically
- However and ensure the tooltip does not say embeddings


dominiccooney (changes requested):
Looks good, but are we missing the 'search' source?

DanTup:
Apparently I didn't submit my reply here.... Should there be a new entry in `ContextFileSource`?:

https://github.com/sourcegraph/cody/blob/bcd49b3694082f529d47681354310a47ea99d195/lib/shared/src/codebase-context/messages.ts#L6-L11

Or should it use an existing one like `keyword`?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1187,6 +1187,8 @@ class ContextProvider implements IContextProvider {
                 return {
                     uri: displayUri,
                     range,
+                    // TODO(dantup): Set source (a value from `ContextFileSource`) here.
+                    // source: 'user',
```
abeatrix, line 1191: maybe we can create a new ContextFileSource called 'symf' since the results returned from symf is not determined by the user manually? wdyt?

abeatrix (approved):
i think you've made the right call. Left a question regarding ContextFileSource for symf but other than that, looks good to me!

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
DanTup: Yeah, that's what the comment above was about - do you know if a) it's safe for me to just add new values here (the type says they're used for analytics - I don't know if they need to match something at the other end), and b) whether any of the existing values are valid for symf or it needs a new one?

Thanks!

abeatrix: @DanTup 

>  a) it's safe for me to just add new values here (the type says they're used for analytics - I don't know if they need to match something at the other end)

That should be fine since the value can be `any`, i think the key is the important part haha

> b) whether any of the existing values are valid for symf or it needs a new one? 

I'd say we will need a new one because I don't see it fitting into any of the existing options: `'embeddings' | 'user' | 'keyword' | 'editor' | 'filename' | 'unified'` , or do you find one they you think it'd match?

DanTup: I don't, but I wasn't completely sure about the meaning of `keyword` and `unified`. I'll add `symf`, thanks!

DanTup: Oh, I just realised that Dominic suggested "search" and this shows up in the tooltip (plus some other things were renamed from "symf" to "search" according to the changelog), so I went with "search".

DanTup:
@abeatrix let me know if using "search" (as noted above) seems reasonable, and if so (and the bots are green) I'll merge this. Thanks!

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
abeatrix: Oh yeah good points! For reference:
- `filename` is referring to the results from the [FilenameContextFetcher](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/local-context/filename-context-fetcher.ts)
- `keyword` refers to the results from the `keywordContextFetcher` but i can't find it anymore, so I guess it's been removed and replaced by `symf`
- `unified` refers to the results from the [unifiedContextFetcher](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/unified-context/index.ts?L13-19) that I believe only Cody Web uses. We don't use it in VS Code.

abeatrix:
> @abeatrix let me know if using "search" (as noted above) seems reasonable, and if so (and the bots are green) I'll merge this. Thanks!

Yea I think that makes sense to me 👍

DanTup:
@dominiccooney merging is blocked on your review. I believe my recent change resolved your issue, but let me know if anything else needs changing. Thanks!

DanTup:
@abeatrix actually, now I'm a little confused.

In the top line at https://github.com/sourcegraph/cody/pull/1858 it says:

> the context display widget should also include local context except the ones added by the user manually via @

And with my fix in this PR, the user-chosen files indeed don't show up. However, in the screenshots (eg. https://github.com/sourcegraph/cody/pull/1858#issuecomment-1822494346) it does still show the mentioned file - so I'm not certain what the intended functionality is now.

If the user types something like "Explain foo.dart", should the context widget show up with "foo.dart" listed in it?

abeatrix:
> in the screenshots (eg. [#1858 (comment)](https://github.com/sourcegraph/cody/pull/1858#issuecomment-1822494346)

@DanTup Oh that's because the screenshot you linked is context added for the `explain` command, which was not added by the user manually via @ . In those cases, I believe Tim wanted to include those under context last time we chatted on slack here: https://sourcegraph.slack.com/archives/C05AGQYD528/p1700638221936799?thread_ts=1700622656.385639&cid=C05AGQYD528

abeatrix:
But if the context was added via manual `@`  input, then i believe we do not want to include those in the context list if i'm not misunderstanding tim's message here: https://sourcegraph.slack.com/archives/C05AGQYD528/p1700631673836569?thread_ts=1700622656.385639&cid=C05AGQYD528

DanTup:
Great, thanks for confirming. I've re-merged main so it's all up-to-date. Needs @dominiccooney's stamp before it'll let me merge (I presume he's out this week).

abeatrix:
Dismissed staled review request to unblock.

Issue #2423, opened 2023-12-18
Autocomplete: extract shared logic from the completion providers 


valerybugakov:
The `Provider.generateCompletions()` method is very similar across three existing providers, and it's becoming increasingly complex as more experiments are added. The goal is to simplify this by breaking down the common logic into smaller functions that can be shared among providers.

PR #2425, opened 2023-12-18
Autocomplete: extract the shared completion params logic

valerybugakov:
## Context

- Extracts the shared between providers logic of defining completion request params related to the number of lines we want to generate from `Provider.generateCompletions()` into a separate function.
- Functional changes:
    - Changes the temperature for the Antrothic provider from 0.5 to 0.2. @philipp-spiess, do you remember why did we end up with 0.5? From my memories, 0.2 performed in a more predictable way, which resulted in a higher CAR and better completion quality. That's why we ended up setting it to 0.2 for Fireworks. Happy to change it back based on the feedback.
    - Changes the temperature for the OpenAI provider from 1 to 0.2. Same question as above. My thinking: 1 is great for creating original jokes but too high for code completions. 
- Do we want/need to have `topK: 0.5` for the OpenAI provider? 
- Part of https://github.com/sourcegraph/cody/issues/2423
- Follow-up https://github.com/sourcegraph/cody/pull/2426

## Test plan

CI


agent/src/index.test.ts
```diff
@@ -181,12 +181,12 @@ describe('Agent', () => {
...
     })
 
-    it('allows us to execute recipes properly', async () => {
+    it.skip('allows us to execute recipes properly', async () => {
         await client.executeRecipe('chat-question', 'How do I implement sum in JavaScript?')
     }, 20_000)
 
     // Timeout is 100ms because we await on `recipes/execute` in the previous test
-    it('executing a recipe sends chat/updateMessageInProgress notifications', async () => {
+    it.skip('executing a recipe sends chat/updateMessageInProgress notifications', async () => {
```
valerybugakov, line 189: Disabled based on this Slack discussion: https://sourcegraph.slack.com/archives/C05B7C6FBPX/p1702888522119979

vscode/src/completions/providers/anthropic.ts
```diff
@@ -167,30 +142,17 @@ export class AnthropicProvider extends Provider {
...
-
-        const useExtendedGeneration = multiline || dynamicMultilineCompletions || hotStreak
+        const { partialRequestParams, fetchAndProcessCompletionsImpl } = getCompletionParamsAndFetchImpl({
+            providerOptions: this.options,
+            lineNumberDependentCompletionParams,
+        })
 
         const requestParams: CodeCompletionsParams = {
-            ...(useExtendedGeneration ? MULTI_LINE_COMPLETION_ARGS : SINGLE_LINE_COMPLETION_ARGS),
-            temperature: 0.5,
```
philipp-spiess, line 181: > Changes the temperature for the Antrothic provider from 0.5 to 0.2. @philipp-spiess, do you remember why did we end up with 0.5? From my memories, 0.2 performed in a more predictable way, which resulted in a higher CAR and better completion quality. That's why we ended up setting it to 0.2 for Fireworks. Happy to change it back based on the feedback.

I don't think we ever tested with Anthropic. I saw this too recently and was a bit confused about it. Happy to turn it down a bit here, but in general I don't think we can use our learnings for StarCoder and apply it 1:1 to Claude Instant.

> Changes the temperature for the OpenAI provider from 1 to 0.2. Same question as above. My thinking: 1 is great for creating original jokes but too high for code completions.

vscode/src/completions/providers/get-completion-params.ts
```diff
@@ -0,0 +1,114 @@
...
+        providerOptions: { multiline: isMutiline, dynamicMultilineCompletions, hotStreak },
+        lineNumberDependentCompletionParams: { singlelineParams, multilineParams, dynamicMultilineParams },
+    } = params
+
+    const useExtendedGeneration = isMutiline || dynamicMultilineCompletions || hotStreak
+
+    const partialRequestParams: Omit<CodeCompletionsParams, 'messages'> = {
+        ...(useExtendedGeneration ? multilineParams : singlelineParams),
+        temperature: 0.2,
+        topK: 0,
```
philipp-spiess, line 88: Should we remove `top_k` here? Some providers (i.e [OpenAI](https://platform.openai.com/docs/api-reference/chat/create)) don't seem to support it even?

vscode/src/completions/providers/get-completion-params.ts
```diff
@@ -0,0 +1,114 @@
...
+        timeouts,
+        providerOptions: { multiline: isMutiline, dynamicMultilineCompletions, hotStreak },
+        lineNumberDependentCompletionParams: { singlelineParams, multilineParams, dynamicMultilineParams },
+    } = params
+
+    const useExtendedGeneration = isMutiline || dynamicMultilineCompletions || hotStreak
+
+    const partialRequestParams: Omit<CodeCompletionsParams, 'messages'> = {
+        ...(useExtendedGeneration ? multilineParams : singlelineParams),
+        temperature: 0.2,
```
philipp-spiess, line 87: I highly doubt we will be able to generalize this for _every_ model.

vscode/src/completions/providers/unstable-openai.ts
```diff
@@ -141,27 +116,18 @@ ${OPENING_CODE_TAG}${infillBlock}`
...
-        let fetchAndProcessCompletionsImpl = fetchAndProcessCompletions
-        if (dynamicMultilineCompletions) {
-            // If the feature flag is enabled use params adjusted for the experiment.
-            Object.assign(requestParams, DYNAMIC_MULTLILINE_COMPLETIONS_ARGS)
-
-            // Use an alternative fetch completions implementation.
-            fetchAndProcessCompletionsImpl = fetchAndProcessDynamicMultilineCompletions
+            ...partialRequestParams,
+            messages: [{ speaker: 'human', text: this.createPrompt(snippets) }],
+            topK: 0.5,
```
philipp-spiess, line 127: The parameters for OpenAI seems to have vastly changed here:

- temperature from 1 to 0.2
- removal of top_p (which is an alternative to sampling with temperature)
- addition of top_k ([which isn't even supported by OpenAI?](https://platform.openai.com/docs/api-reference/chat/create))

I think we should test these changes quickly, wdyt?

philipp-spiess (commented):
Changes LGTM but I would like to see Anthropic and OpenAI examples in the test plan since we do change the parameters quite a bit

philipp-spiess (approved):
Let'sgooooo

PR #2438, opened 2023-12-19
Adding support to save the locally selected model so that new chats don´t require repeated model selection

arafatkatze:
image

I was quite frustrated with the loading issue as well so I changed it. The video is pretty self explanatory. It saves the state even after you quit and load Cody again. 

I have a strong suspicion that a lot of users will switch to GPT 4-Turbo once and then never switch back

https://github.com/sourcegraph/cody/assets/11155207/189fe1b3-5e26-49ce-adf1-f156a020608d

## Test plan
Just try it out and open close a few times

### DIFF
--- a/vscode/src/chat/chat-view/ChatPanelsManager.ts
+++ b/vscode/src/chat/chat-view/ChatPanelsManager.ts
@@ -186,10 +186,6 @@ export class ChatPanelsManager implements vscode.Disposable {
             ChatModelProvider.add(new ChatModelProvider(authStatus.configOverwrites.chatModel))
         }
         const models = ChatModelProvider.get(authStatus.endpoint)
-        const defaultModel = models.find(m => m.default) || models[0]
-        if (!defaultModel) {
-            throw new Error('No chat model found in server-provided config')
-        }
 
         return this.options.contextProvider.config.experimentalSimpleChatContext
             ? new SimpleChatPanelProvider({
@@ -205,7 +201,7 @@ export class ChatPanelsManager implements vscode.Disposable {
                       this.options.contextProvider,
                       this.options.platform
                   ),
-                  defaultModelID: defaultModel.model,
+                  models,
               })
             : new ChatPanelProvider(this.options)
     }
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -39,6 +39,7 @@ import { SymfRunner } from '../../local-context/symf'
 import { logDebug, logError } from '../../log'
 import { AuthProvider } from '../../services/AuthProvider'
 import { getProcessInfo } from '../../services/LocalAppDetector'
+import { localStorage } from '../../services/LocalStorageProvider'
 import { telemetryService } from '../../services/telemetry'
 import { telemetryRecorder } from '../../services/telemetry-v2'
 import { createCodyChatTreeItems } from '../../services/treeViewItems'
@@ -78,7 +79,7 @@ interface SimpleChatPanelProviderOptions {
     treeView: TreeViewProvider
     featureFlagProvider: FeatureFlagProvider
     recipeAdapter: SimpleChatRecipeAdapter
-    defaultModelID: string
+    models: ChatModelProvider[]
 }
 
 export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelProvider {
@@ -108,7 +109,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
     private readonly contextStatusAggregator = new ContextStatusAggregator()
     private readonly editor: VSCodeEditor
     private readonly treeView: TreeViewProvider
-    private readonly defaultModelID: string
+    private readonly modelID: string
 
     private history = new ChatHistoryManager()
     private prompter: IPrompter = new DefaultPrompter()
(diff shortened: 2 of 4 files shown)


valerybugakov (approved):
I also started looking into it because I want GPT-4 by default. TY!

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -205,7 +207,7 @@ export class ChatPanelsManager implements vscode.Disposable {
                       this.options.contextProvider,
                       this.options.platform
                   ),
-                  defaultModelID: defaultModel.model,
```
philipp-spiess, line 208: We should check if the saved model is actually a valid model name for the current configuration. Otherwise, if you switch from a dotcom account to enterprise or from pro to free, this may result in a wrong model identifier which, when used, might cause errors.

philipp-spiess (commented):
This is awesome! I would expect some kind of validation to go on though, did you look into this?

Removing @valerybugakov's approval not to merge this too eagerly since it could cause some nasty errors

vscode/src/chat/protocol.ts
```diff
@@ -77,6 +77,7 @@ export type WebviewMessage =
     | { command: 'abort' }
     | { command: 'custom-prompt'; title: string; value?: CustomCommandType }
     | { command: 'reload' }
+    | { command: 'saveModel'; model: string }
```
beyang, line 80: The `chatModel` command already exists; why not use that one?

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -139,7 +140,8 @@ export class ChatPanelsManager implements vscode.Disposable {
         // Get the view column of the current active chat panel so that we can open a new one on top of it
         const activePanelViewColumn = this.activePanelProvider?.webviewPanel?.viewColumn
 
-        const provider = this.createProvider()
+        const savedModel = localStorage.get('model')
```
beyang, line 143: Let's give the key a more distinctive name, like "lastSelectedModelID"

vscode/src/chat/chat-view/ChatPanelsManager.ts
beyang: It seems strange that we write the last selected model ID in SimpleChatPanelProvider, but read it in ChatPanelsManager to pass into SimpleChatPanelProvider as a constructor param. Where possible, it's better to localize state management. Could we move the `localStorage.get` call to an `onReady` method in SimpleChatPanelProvider that runs when we receive the `ready` message?

beyang (changes requested):

vscode/src/chat/chat-view/ChatPanelsManager.ts
arafatkatze: Moved the model selection to this function 

https://github.com/sourcegraph/cody/blob/2b9310b2828a66dc9d5b94d03f68a79013b8faf6/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts#L176-L192

Also it takes into account that the model selected is valid for that user. 

> Could we move the localStorage.get call to an onReady method in SimpleChatPanelProvider that runs when we receive the ready message?

Not sure if I fully understand this but hopefully moving the model selection to a single place makes the code better than before.

vscode/src/chat/protocol.ts
arafatkatze: Need to store and load from storage.

philipp-spiess: @arafatkatze Not sure I follow: If we already send a `chatModel` message, why would we need another one to save that if the behavior is "always save the last one as default"? 🤔

philipp-spiess (commented):
This looks crisp now but I’m still unsure why we need a new notification for this instead of reusing. Do you mind elaborating?

philipp-spiess (approved):
Sharp! This looks good now. Let's add a change log entry though 🙇

philipp-spiess:
I wonder why the test still fails 🤔  Did you create the PR in a _fork_ of the cody repo? You can now create your branches on this repo instead, that should help!

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -108,7 +109,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
     private readonly contextStatusAggregator = new ContextStatusAggregator()
     private readonly editor: VSCodeEditor
     private readonly treeView: TreeViewProvider
-    private readonly defaultModelID: string
+    private readonly modelID: string
```
philipp-spiess, line 112: Do we need this as a member now that we can compute it based on the localStorage? Asking because I just noticed that we have `this.modelID` and `this.chatmodel.modelID` and we only update one at some point. It's better to not keep duplicate references around I think.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -336,6 +355,8 @@ export class SimpleChatPanelProvider implements vscode.Disposable, IChatPanelPro
                 break
             case 'chatModel':
                 this.chatModel.modelID = message.model
```
philipp-spiess, line 357: Should this also update `this.modelID` (and if yes, can we make it so that there is no duplication?)

philipp-spiess (changes requested):
I found one more nit I’m afraid. I don't feel great that we have a compute function but keep a copy of the value in the class properties. Can we avoid this somehow?

philipp-spiess (approved):
Bliss

PR #2439, opened 2023-12-19
Autocomplete: change Anthropic temperature to 0.2

valerybugakov:
## Context

- Changes the temperature for the Antrothic provider from 0.5 to 0.2. From my memories, 0.2 performed in a more predictable way, which resulted in a higher CAR and better completion quality. That's why we ended up setting it to 0.2 for Fireworks.
- Discussed briefly in https://github.com/sourcegraph/cody/pull/2425#discussion_r1429960244

## Test plan

TODO

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -154,7 +154,6 @@ export class AnthropicProvider extends Provider {
         const requestParams: CodeCompletionsParams = {
             ...partialRequestParams,
             messages: this.createPrompt(snippets).messages,
-            temperature: 0.5,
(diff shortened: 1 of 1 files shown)


dominiccooney (changes requested):
Let's do an evaluation and attach the results of the evaluation to this patch.

Let's include enough information in the diff so if people want to re-do the evaluation then anyone with reasonable time and skills could reproduce it.

valerybugakov:
Yep, that's the plan 👍

valerybugakov:
It's not on the top of the priority list for now. I'm going to close it and come back to it later.

PR #2441, opened 2023-12-19
Automate VS Code release GitHub changelog

philipp-spiess:
We already have a pretty stable release pipeline, however one thing that has been very tedious and we often forgot about was to update the GitHub release changelog.

This PR adds a new script that computes the GitHub release body from the `CHANGELOG.md`. 

It also makes it so that we always create a new insider build when we do a stable release (something that also went out of sync before).

## Test plan

- `pnpm -C vscode run github-changelog`
- `cat vscode/GITHUB_CHANGELOG.md`

Screenshot 2023-12-19 at 16 27 47

### DIFF
--- a/vscode/scripts/github-changelog.ts
+++ b/vscode/scripts/github-changelog.ts
@@ -1,3 +1,5 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
 /**
  * A script to create the GitHub release changelog format from the current
  * CHANGELOG.md.
--- a/vscode/scripts/github-changelog.ts
+++ b/vscode/scripts/github-changelog.ts
@@ -50,18 +50,18 @@ function extractSection(changelog: string, version: string): { changes: Change[]
     for (const line of lines) {
         if (found) {
             if (line.startsWith('## ')) {
-                const versionMatches = /^## \[(\d+\.\d+\.\d+)]$/.exec(line)
-                if (!versionMatches) {
+                const versionMatches = /^## \[(?<dottedVersion>\d+\.\d+\.\d+)]$/.exec(line)
+                if (!versionMatches?.groups) {
                     throw new Error(`Malformed version line: ${line}`)
                 }
-                previousVersion = versionMatches[1]
+                previousVersion = versionMatches.groups?.dottedVersion
                 break
             }
 
             if (line.startsWith('- ')) {
                 const change = line.slice(2)
 
-                const linkRegex = /\[(pull|pulls|issue|issues).*]\((.*)\)/
+                const linkRegex = /\[(pull|pulls|issue|issues).*]\((?<link>.*)\)/
                 const firstLink = linkRegex.exec(change)
 
                 let text = change.slice(0, firstLink?.index ?? -1).trim()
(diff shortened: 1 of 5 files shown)


vscode/scripts/github-changelog.ts
```diff
@@ -0,0 +1,167 @@
...
+                // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+                // @ts-ignore: Fetch is available in node :shrug:
+                const json = await fetch(apiUrl).then(res => res.json())
+                if (json?.user?.login) {
+                    author = json.user.login
+                }
+            }
+        }
+
+        output += `- ${change.text}${author ? ` by @${author}` : ''}${change.link ? ` in ${change.link}` : ''}\n`
```
valerybugakov, line 156: Would it make sense to change the changelog format to match the expected format of the release notes? Then we have a smaller surface area for errors in this script, and it's easy to update release notes manually if the script fails for some reason (copy-paste changelog entries and remove some of them). 

cc @toolmantim @beyang

valerybugakov (commented):
That's great! Thanks for looking into it.

It would be awesome if we could unify the formats of the changelog and release notes and remove the related logic from this script.

dominiccooney:
My $0.02, let's the script grep change descriptions for ^RELNOTES:(.*)$ and dump the content of the .* in the changelog.

We could do ADDED:, FIXED:, etc. to keep those sections if we want.

vscode/scripts/github-changelog.ts
```diff
@@ -0,0 +1,167 @@
...
+function extractSection(changelog: string, version: string): { changes: Change[]; previousVersion: string } {
+    let previousVersion = ''
+
+    const lines = changelog.split('\n')
+    const changes = []
+    let found = false
+    for (const line of lines) {
+        if (found) {
+            if (line.startsWith('## ')) {
+                const versionMatches = /^## \[(\d+\.\d+\.\d+)]$/.exec(line)
```
dominiccooney, line 53: If it helps readability, use named groups, e.g. `...\[(?<dottedVersion>\d+\.\d+\.\d+)]...` and then `groups['dottedVersion']`. More readable than the 1, 2 indices.

dominiccooney (commented):
Think this is great.

Are we dropping the added/changed/fixed groupings? What do you think about sorting related items together?

philipp-spiess:
> Are we dropping the added/changed/fixed groupings? What do you think about sorting related items together?

We have only consistently done these for minor versions so far and with a varying naming structure, so I dropped it for now. But if we plan to change the changelog format to match this, we should probably include the titles in the same way as the changelog (maybe removing empty sections)

philipp-spiess:
> My $0.02, let's the script grep change descriptions for ^RELNOTES:(.*)$ and dump the content of the .* in the changelog.

From the git log you mean? 

I think this would be much better too but it would make for awkward "reverts"

dominiccooney (approved):
I support eliminating this toil.

vscode/scripts/github-changelog.ts
philipp-spiess: Let's keep what we have for now. This PR is mainly thought to help with patch releases and this keeps the current format. We can still overwrite the format etc but having at least patches being fully automated would already be helpful I think (better than trying to make larger changes here and having this PR stall for longer)

This also only changes the initial release text on github that is currently blank, so it's not even inconveniencing the current workflow.

PR #2446, opened 2023-12-20
Autocomplete: Do not trigger a completion if a single-line completion was accepted

philipp-spiess:
This is a small optimization that should cut down a bit on unnecessary LLM request:

In VS Code, if you accept a completion, this changes the document and thus issue a new completion request. Makes sense! The thing is: we always complete full lines so the chance that you do want a follow-up completion is _very_ slim. In some cases the ending `)` or `;` helps to guard against this but not in all cases. 

This PR adds a special logic now to detect this case and not issue a completion (so accepting a completion via `tab` won't issue a new request but any further keystroke (like `enter`) will).

## Test plan

- Added tests
- 

https://github.com/sourcegraph/cody/assets/458591/83d9c044-a886-4e72-90bf-d2c1486f260e

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
@@ -7,7 +7,7 @@ import { completion } from '../test-helpers'
 import { getInlineCompletions, params, V } from './helpers'
 
 describe('[getInlineCompletions] no request when accepting', () => {
-    // In VS Code, accepting a completion will immediatly start a new completion request. If the
+    // In VS Code, accepting a completion will immediately start a new completion request. If the
     // user, however, accepted a single line completion, chances are that the current line is
     // finished (ie. the LLM already gave the best guess at completing the line).
     //
--- a/vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
@@ -1,0 +1,109 @@
+import dedent from 'dedent'
+import { describe, expect, it } from 'vitest'
+
+import { TriggerKind } from '../get-inline-completions'
+import { completion } from '../test-helpers'
+
+import { getInlineCompletions, params, V } from './helpers'
+
+describe('[getInlineCompletions] no request when accepting', () => {
+    // In VS Code, accepting a completion will immediatly start a new completion request. If the
+    // user, however, accepted a single line completion, chances are that the current line is
+    // finished (ie. the LLM already gave the best guess at completing the line).
+    //
+    // Thus, this results in a request that almost always has zero results but still incurs network
+    // and inference costs.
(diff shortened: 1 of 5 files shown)


vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
```diff
@@ -0,0 +1,109 @@
+import dedent from 'dedent'
+import { describe, expect, it } from 'vitest'
+
+import { TriggerKind } from '../get-inline-completions'
+import { completion } from '../test-helpers'
+
+import { getInlineCompletions, params, V } from './helpers'
+
+describe('[getInlineCompletions] no request when accepting', () => {
+    // In VS Code, accepting a completion will immediatly start a new completion request. If the
```
dominiccooney, line 10:
```suggestion
    // In VS Code, accepting a completion will immediately start a new completion request. If the
```

vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
```diff
@@ -0,0 +1,109 @@
...
+                        triggerKind: TriggerKind.Manual,
+                        lastAcceptedCompletionItem: {
+                            requestParams: initialRequestParams,
+                            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+                            analyticsItem: item!.items[0]!,
+                        },
+                    }
+                )
+            )
+        ).not.toEqual<V>(null)
```
dominiccooney, line 76: 😎

vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
```diff
@@ -0,0 +1,109 @@
...
+
+describe('[getInlineCompletions] no request when accepting', () => {
+    // In VS Code, accepting a completion will immediatly start a new completion request. If the
+    // user, however, accepted a single line completion, chances are that the current line is
+    // finished (ie. the LLM already gave the best guess at completing the line).
+    //
+    // Thus, this results in a request that almost always has zero results but still incurs network
+    // and inference costs.
+    it('should not make a request after accepting a completion', async () => {
+        const initialRequestParams = params(
```
dominiccooney, line 17: This test and the next one would be more readable if they used a helper function that's parameterized on the trigger kind to show that, *exact* same situation with implicit trigger *doesn't* cause a follow on completion attempt, but manual trigger *does.*

The difference in trigger kind is only one line, it is hard to spot it in the midst of the rest of the test code.

dominiccooney (approved):
Looks great. Some feedback inline about readable tests.

valerybugakov (approved):
Nice!

vscode/src/completions/get-inline-completions-tests/no-request-when-accepting.test.ts
philipp-spiess: muuuch better 🙇

PR #2454, opened 2023-12-21
RFC: Detect and close stale issues and PRs

philipp-spiess:
This PR adds the stale action to automatically detect and close stale issues and PRs. 

I configured it so that issues will be marked as stale after 60 days of no activity and will be closed within 5 days unless manually interfered.

I’m personally up for even more aggressive timeouts, e.g. 30 days are enough, just because of how fast we move here but I’m happy to settle on something larger.

## Test plan

We see if issues close huh

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,15 @@
+name: 'Close stale issues and PRs'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is marked as stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed automatically in 5 days.'
+          stale-pr-message: 'This PR is marked as stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed automatically in 5 days.'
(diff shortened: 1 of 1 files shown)


beyang (approved):
Is there a way to mark something as "evergreen", so it escapes the stale reaper? I've filed some issues that track more forward-looking features that may run into the 60 day limit.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v9
+        with:
+          stale-issue-message: 'This issue is marked as stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed automatically in 5 days.'
+          stale-pr-message: 'This PR is marked as stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed automatically in 5 days.'
+          days-before-stale: 60
+          days-before-close: 5
```
beyang, line 15: ChatGPT suggested the following edit to add a label that would prevent the stale reaper from acting on specific issues
```suggestion
          days-before-close: 5
          exempt-issue-labels: 'nostale'
          exempt-pr-labels: 'nostale'
```

philipp-spiess: Screenshot 2023-12-21 at 18 22 39

Looks legit!

PR #2456, opened 2023-12-21
Remove serverEndpoint from config object

philipp-spiess:
This refactors the `Config` object to no longer expose a `serverEndpoint`. The `serverEndpoint` should be retrieved only from local storage as part of the authenticated user. When going through this legacy code part, the server endpoint would always be set to the dotcom instance URL instead.

This caused some parts to behave like you were on dotcom even when inside an enterprise version.

## Test plan

- I tested it by evaluating analytics logs and switching accounts
- Video is large so uploading it on Google Drive instead: https://drive.google.com/file/d/1dpn-C1IczJ9CQ0zX6evMyzKpHpq8lIvS/view?usp=sharing

### DIFF
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -13,7 +13,6 @@ export const CONTEXT_SELECTION_ID: Record<ConfigurationUseContext, number> = {
 
 // Should we share VS Code specific config via cody-shared?
 export interface Configuration {
-    serverEndpoint: string
     proxy?: string | null
     codebase?: string
     debugEnable: boolean
@@ -33,7 +32,6 @@ export interface Configuration {
     autocomplete: boolean
     autocompleteLanguages: Record<string, boolean>
     autocompleteAdvancedProvider: 'anthropic' | 'fireworks' | 'unstable-openai' | null
-    autocompleteAdvancedServerEndpoint: string | null
     autocompleteAdvancedModel: string | null
     autocompleteAdvancedAccessToken: string | null
     autocompleteCompleteSuggestWidgetSelection?: boolean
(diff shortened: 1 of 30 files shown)


lib/shared/src/configuration.ts
```diff
@@ -13,7 +13,6 @@ export const CONTEXT_SELECTION_ID: Record<ConfigurationUseContext, number> = {
 
 // Should we share VS Code specific config via cody-shared?
 export interface Configuration {
-    serverEndpoint: string
```
philipp-spiess, line 16: This is the main change, everything else was just necessary to make sure we don't rely on this field.

lib/shared/src/configuration.ts
```diff
@@ -33,7 +32,6 @@ export interface Configuration {
     autocomplete: boolean
     autocompleteLanguages: Record<string, boolean>
     autocompleteAdvancedProvider: 'anthropic' | 'fireworks' | 'unstable-openai' | null
-    autocompleteAdvancedServerEndpoint: string | null
```
philipp-spiess, line 36: Removed this while at it, it's not used anymore and is just confusing

vscode/src/completions/get-inline-completions-tests/analytics.test.ts
```diff
@@ -89,8 +89,6 @@ describe('[getInlineCompletions] completion event', () => {
                 "items": [
                   {
                     "charCount": 30,
-                    "insertText": "console.log(bar)
-              return false}",
                     "lineCount": 2,
```
beyang, line 92: Worth to add a security-oriented test case somewhere in this file, or to add a comment that indicates whether some information should not be included?

beyang (approved):

Issue #2468, opened 2023-12-22
bug: tier metadata not firing correctly for (definitely) upsell and (possibly) abuse dialogues


chenkc805:
### Version

1.0

### Describe the bug

- We are tracking `tier:enterprise` metadata for upsells, when the only tier that can get upsells is the Free tier.
- For abuse messages, I'm surprised to be seeing so many Enterprise users (as per my query) hitting our abuse limits, so I think there's a bug somewhere. I could be wrong though.
Query for this
```
select name, JSON_QUERY(public_argument, '$.tier'), JSON_QUERY(public_argument, '$.limit_type'), count(*)
from dotcom_events.events
where (name like 'CodyVSCodeExtension:upsellUsage%'
or name like 'CodyVSCodeExtension:abuse%')
and timestamp > '2023-12-20'
group by 1,2,3 

```
<img width="1237" alt="Screenshot 2023-12-20 at 8 37 00 PM" src="https://github.com/sourcegraph/cody/assets/3654603/a26227fa-fc3b-4b78-a0ab-85c99cecf4fb">

### Expected behavior

- Upsells should only be showing for Free users, so any upsell telemetry should have `tier = free`
- Abuse limits only be hit by Pro or Enterprise users, with the majority being under Pro (assuming there's more Pro DAU than Enterprise DAU)

PR #2057, opened 2023-12-01
log chat rate limit events

abeatrix:
add: log Cody chat rate limit related events

Log  Cody chat rate limit events

Added `CodyVSCodeExtension:upsellUsageLimitCTA:shown` and `CodyVSCodeExtension:upsellUsageLimitCTA:clicked` to chat_commands

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/287966cf-5ae8-422f-b1bd-9194db51fd18

PR #2470, opened 2023-12-22
VS Code: require `userInfo` prop

valerybugakov:
## Context

- Fixes the issue caused by the fact that `userInfo` [was not available](https://github.com/sourcegraph/cody/pull/2057/files#diff-3ef5ec11a5cb8846141e4fc4e08c5502837d912d07e00fadfbfa25187bdc0d84R20) in the critical place required for a telemetry event creation. This PR makes `userInfo` a required prop in the React component tree for all the intermediate components to avoid such an issue in the future. It sets the `userInfo` value to `{ isDotComUser: true, isCodyProUser: false }` for the components reused in the web version. @chenkc805, let me know if other default values make more sense.
- Closes https://github.com/sourcegraph/cody/issues/2468
- [Slack thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1703205833741309)

## Test plan

- Hit the rate limit events as a DotCom user and ensure the correct event name/tier is used.
- Ensured that this won't happen again with the updated types.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -4,6 +4,7 @@ import { Client, createClient, Transcript } from '@sourcegraph/cody-shared/src/c
 import { ChatMessage } from '@sourcegraph/cody-shared/src/chat/transcript/messages'
 import { ErrorLike, isErrorLike } from '@sourcegraph/cody-shared/src/common'
 import type { Editor } from '@sourcegraph/cody-shared/src/editor'
+import { isDotCom } from '@sourcegraph/cody-shared/src/sourcegraph-api/environments'
 import { CodySvg } from '@sourcegraph/cody-ui/src/utils/icons'
 
 import { Chat } from './Chat'
@@ -118,7 +119,7 @@ export const App: React.FunctionComponent = () => {
                                 isCodyEnabled={true}
                                 onSubmit={onSubmit}
                                 userInfo={{
-                                    isDotComUser: true,
+                                    isDotComUser: isDotCom(config.serverEndpoint),
                                     isCodyProUser: false,
                                 }}
                             />
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -117,6 +117,10 @@ export const App: React.FunctionComponent = () => {
                                 setInputHistory={setInputHistory}
                                 isCodyEnabled={true}
                                 onSubmit={onSubmit}
+                                userInfo={{
+                                    isDotComUser: true,
+                                    isCodyProUser: false,
+                                }}
                             />
                         </>
                     )
--- a/web/src/Chat.tsx
+++ b/web/src/Chat.tsx
@@ -10,8 +10,8 @@ import { SubmitSvg } from '@sourcegraph/cody-ui/src/utils/icons'
 
 import styles from './Chat.module.css'
 
-export const Chat: React.FunctionComponent<
-    Omit<
+interface ChatProps
+    extends Omit<
         React.ComponentPropsWithoutRef<typeof ChatUI>,
         | 'textAreaComponent'
         | 'submitButtonComponent'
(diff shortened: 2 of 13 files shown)


chenkc805:
So confirming, will the intended behavior be:

- If user is Free & on Dotcom & they hit rate limits, we log an upsell
- If user is Pro & on Dotcom & they hit rate limits, we log an abuse message
- If user is Enterprise & not on Dotcom & they hit rate limits, we log an abuse message
Every other case (ie. Enterprise & Dotcom, Free & not Dotcom, Pro & not Dotcom) are not relevant and can be ignored. Did I get that right?

valerybugakov:
> If user is Free & on Dotcom & they hit rate limits, we log an upsell

If a user hits the rate limit error and an upgrade is available (based on response headers, it can only happen on DotCom for users on the Free tier), we log an upsell ✅

>  If user is Pro & on Dotcom & they hit rate limits, we log an abuse message

If a user hits the rate limit error and an upgrade is NOT available (based on response headers, it can only happen on DotCom for users on the Pro tier), we log an abuse message ✅

> If user is Enterprise & not on Dotcom & they hit rate limits, we log an abuse message

If a user hits the rate limit error and an upgrade is NOT available (based on response headers, it can only happen on DotCom for users on the Pro tier), we log an abuse message ✅

This PR does not change the logic related to choosing the suitable event. It fixes the data propagation so we have all the required information to make the right choice.

chenkc805 (approved):

web/src/App.tsx
```diff
@@ -117,6 +117,10 @@ export const App: React.FunctionComponent = () => {
                                 setInputHistory={setInputHistory}
                                 isCodyEnabled={true}
                                 onSubmit={onSubmit}
+                                userInfo={{
+                                    isDotComUser: true,
```
abeatrix, line 121:
```suggestion
                                    isDotComUser: isDotCom(config.serverEndpoint),
```
I might be missing some context but wouldn't this not work correctly for enterprise users? Would it be safer to check the serverEndpoint from config to determine its value instead?

web/src/Chat.tsx
```diff
@@ -53,6 +53,10 @@ export const Chat: React.FunctionComponent<
         inputRowClassName={styles.inputRow}
         chatInputClassName={styles.chatInput}
         isCodyEnabled={true}
+        userInfo={{
+            isDotComUser: true,
```
abeatrix, line 57: same question as above

abeatrix (commented):
left a question inline

web/src/App.tsx
valerybugakov: Used your suggestion 👍

valerybugakov:
@abeatrix, @chenkc805, end-to-end tests worked for this use case by accident. I fixed them in my latest commit. Before that, we ran all the `upgrade rate limit` E2E tests, thinking we were on the enterprise instance.

abeatrix:
> @abeatrix, @chenkc805, end-to-end tests worked for this use case by accident. I fixed them in my latest commit. Before that, we ran all the `upgrade rate limit` E2E tests, thinking we were on the enterprise instance.

HHmm this is different than how I remembered and it looks like something were changed while I was on PTO, but when we first set up the test, we assumed the client would return different value for DotCom and non-DotCom user, and we use the custom endpoint in E2E test to differentiate DotCom and non-DotCom instances: https://github.com/sourcegraph/cody/blob/8bbb73e0087567850add91d39dd8064a228928af/vscode/test/fixtures/mock-server.ts#L122-L126

valerybugakov:
Yes, that's correct, @abeatrix. This changes response headers for completion requests. But we also look at the `authStatus.endpoint` to understand whether we are on DotCom. This was not taken into account in E2E tests AFAIU.

abeatrix:
> Yes, that's correct, @abeatrix. This logic changes response headers for completion requests. But we also look at the `authStatus.endpoint` to understand whether we are on DotCom. This was not taken into account in E2E tests AFAIU.

Ah. I meant to say we added the custom completion endpoint for the test to workaround that instead of relying on the authStatus.endpoint in the webview, but if that's changed then what I mentioned is now irrelevant 😅

valerybugakov:
> to workaround that instead of relying on the authStatus.endpoint in the webview

But we still have code in the codebase that relies on `authStatus.endpoint` to determine whether we are on DotCom. Do you know if the intention was to remove all such instances, too?

Issue #2401, opened 2023-12-15
bug: files in `bin/` are filtered out of @-mention in chat


DanTup:
### Version

main 2a00da01

### Describe the bug

When I typed `@`, my open file `bin\\main.dart` was already in the results, but as I started typing `bin` it vanished.

https://github.com/sourcegraph/cody/assets/1078012/9a997f6e-42d3-4904-b564-8b05f35e2fb3

### Expected behavior

Files that exist and match my query should not be filtered out.

DanTup:
This turned out to be very specific... The default exclude list includes `bin/`:

https://github.com/sourcegraph/cody/blob/9d48eeb94c3674807c4815151510181943026364/vscode/src/editor/utils/editor-context.ts#L16

This is something I've hit before... When I did .NET, I would have `bin/` in by global gitignore (because in .NET it's compiled binaries). It's quite common in Dart to use `bin/` for the entry points of packages:

https://dart.dev/tools/pub/package-layout#:~:text=make_lunch.dart%0A%20%20bin/-,enchilada,-doc/%0A%20%20%20%20api/%20***%0A%20%20%20%20getting_started

@toolmantim any thoughts on this (since you picked the original defaults)? I'm not familiar with how many languages use `bin/`.. For .NET, probably explicitly including `bin/Debug/` and `bin/Release/` would filter out the garbage, although perhaps the real fix here is to use gitignore/etc.

toolmantim:
Best to ask @abeatrix when she’s back, or @beyang — unsure how that initial exclude list was chosen.

abeatrix:
We can just remove the keyword from the exclude list if the keyword is included in the exclude list, or just remove the exclude list in general from the @ search if you want. I was just using the same exclude list I used for the generate unit test to filter out uncommon files for unit tests and thought the same list could be used, no other particular reasons.

abeatrix:
I'd recommend keeping it and remove it only if the keyword is included because those files will always show up on top of the list from those folder compared to the actual ones with the same names

DanTup:
It may look odd if the user is typing `@bin/main.dart` if it's missing from the list until they've typed `@bin/`?

I wonder if it should be included (not in the exclude list) but penalised in the ranking unless the query starts with it?

That way if your project is small, it won't be obvious `bin/main.dart` is missing, but it may be further down the list until you've typed `bin/`?

toolmantim:
[...]

PR #2472, opened 2023-12-22
Remove "bin/" from search ignore and just rank low unless typed

DanTup:
Fixes https://github.com/sourcegraph/cody/issues/2401 by removing `bin/` from the ignore list and instead just penalizing it unless you have typed `bin` in the query specifically.

https://github.com/sourcegraph/cody/assets/1078012/2ad0ef3b-b2f9-45bd-bfc0-54fd77811511

## Test plan

- Run `pnpm test:unit` and ensure the new tests pass
- Open a project with a file in a folder named `bin` and ensure the file shows up in search results, near the bottom unless you've explicitly typed "bin"

### DIFF
--- a/vscode/src/editor/utils/editor-context.ts
+++ b/vscode/src/editor/utils/editor-context.ts
@@ -14,10 +14,15 @@ const findWorkspaceFiles = async (cancellationToken: vscode.CancellationToken): 
     // TODO(toolmantim): Add support for the search.exclude option, e.g.
     // Object.keys(vscode.workspace.getConfiguration().get('search.exclude',
     // {}))
-    const fileExcludesPattern = '**/{*.env,.git,out/,dist/,bin/,snap,node_modules,__pycache__}**'
+    const fileExcludesPattern = '**/{*.env,.git,out/,dist/,snap,node_modules,__pycache__}**'
     // TODO(toolmantim): Check this performs with remote workspaces (do we need a UI spinner etc?)
     return vscode.workspace.findFiles('', fileExcludesPattern, undefined, cancellationToken)
 }
+
+// Some matches we don't want to ignore because they might be valid code (for example `bin/` in Dart)
+// but could also be junk (`bin/` in .NET). If a file path contains a segment matching any of these
+// items it will be ranked low unless the users query contains the exact segment.
+const lowScoringPathSegments = ['bin']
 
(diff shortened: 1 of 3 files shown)


vscode/src/editor/utils/editor-context.ts
```diff
@@ -67,10 +72,24 @@ export async function getFileContextFiles(
         threshold: -100000,
     })
 
+    // Apply a penalty for segments that are in the low scoring list.
+    const adjustedResults = [...results].map(result => {
```
abeatrix, line 76: Is this a good place to add `isCodyIgnoredFile` for filtering ignored files?

abeatrix (approved):
Left a question in line regarding adding isCodyIgnoredFile.

vscode/src/editor/utils/editor-context.ts
DanTup: hmm, yes! I thought I'd already seen it added in here in your ignore PR, but now I don't see it. I'll test it (and update if necessary).

DanTup: Ah, looks like there are a few other places in the same file that got it and I misunderstood they were here:

https://github.com/sourcegraph/cody/commit/c9a4f7edf1d00c2f958f7003f07aeaba5c15d378#diff-315ecedea695e410a04a36543d5d788b07f05d75885cedffceb922c3f33cd569

Adding!

DanTup: I've added this filter and manually tested and it's good - however adding a test for it has revealed some issues with the testing on Windows (I think the shim findFiles() is returning bad URIs though I've not yet figured out why).

If you're happy to land this as-is (manually tested), I'll open a separate PR to fix this issue and add in the test.

DanTup: image

DanTup: Scratch that, it's my bad test - fix incoming :)

abeatrix: It's working for me!

image

DanTup:
@abeatrix I've pushed a test and updated the change log, let me know if this still looks good to you.

abeatrix (approved):
LGTM!

Issue #2195, opened 2023-12-08
bug: hit up, resend @-tagged file does not include the file, highlight the link, etc.


dominiccooney:
### Version

Main a8d1fb32916d7f08ce26277cbbf0c186d5fb78f8

### Describe the bug

Repro steps:

1. Start a chat
2. @-mention a file, chat about it.
3. Hit up to bring up the same message again, chat that.

https://github.com/sourcegraph/cody/assets/55120/b7a09274-c584-4905-9da6-9dd628c3457b

### Expected behavior

Expected behavior: File is highlighted like a link, sent every time I mention it.

Actual behavior: File isn't highlighted like a link, I have just sent Cody the file *name.*

dominiccooney:
Related, if I @ and "type through" the autocomplete, then there's no file included. I feel like this is a *good* thing, because sometimes you do really want to talk about things that start with that symbol. So probably the chat history needs to remember, not only what the text of the message is, but where did that text come from?

csells:
shouldn't hitting Up and resubmitting a chat prompt do the same thing it did before?

DanTup:
I started looking at this, but it wasn't as straight forward as I thought. I pushed the changes that I started to https://github.com/sourcegraph/cody/pull/2473 and would like some pointers/validation that this I'm going in the right direction. Thanks!

PR #2473, opened 2023-12-22
[WIP] Ensure pressing Up to re-send chat messages includes same context files

DanTup:
I started trying to fix https://github.com/sourcegraph/cody/issues/2195 but I'm not sure if I'm taking the right approach...

Currently the input history is just a `string[]` so we don't retain the `Map<string, ContextFile>` which maps input strings to context files for previous chat inputs. I considered just capturing these in a new state array/map so we could look a map for a previous history, however it seems you can reuse history after closing and re-opening a chat window, so to work correctly this needs to persist.

So I started changing `inputHistory: string[]` to `inputHistory: ChatInputHistory[]` so we could track both the input string and the `Map<string, ContextFile>` for context files, but it's touching a lot of code and I have some other concerns so I wanted some validation this is the right direction. In particular:

- Since this history is persisted, I think changing the shape of this object would become incompatible with earlier chats and we'd have to handle opening chats that had a `string[]` history
- Some of these changes are outside of `vscode/` and I'm not sure if they will break other clients
- I started changing the signature of `onHumanMessageSubmitted` to accept `ChatInputHistory` instead of just `string`, but this method is already accepting a `contextFiles?: ContextFile[]` which feels a little odd (maybe this is fine because there can be context files not tied to the chat input text?)

### DIFF
--- a/lib/shared/src/chat/transcript/messages.ts
+++ b/lib/shared/src/chat/transcript/messages.ts
@@ -45,11 +45,15 @@ export interface ChatMetadata {
 
 export interface UserLocalHistory {
     chat: ChatHistory
-    input: string[]
+    input: ChatInputHistory[]
 }
 
 export interface ChatHistory {
     [chatID: string]: TranscriptJSON
+}
+
+export interface ChatInputHistory {
+    inputText: string
 }
 
 export interface OldChatHistory {
--- a/lib/ui/src/Chat.tsx
+++ b/lib/ui/src/Chat.tsx
@@ -5,6 +5,7 @@ import classNames from 'classnames'
 import {
     ChatButton,
     ChatContextStatus,
+    ChatInputHistory,
     ChatMessage,
     ChatModelProvider,
     CodyPrompt,
@@ -29,8 +30,8 @@ interface ChatProps extends ChatClassNames {
     contextStatus?: ChatContextStatus | null
     formInput: string
     setFormInput: (input: string) => void
-    inputHistory: string[]
-    setInputHistory: (history: string[]) => void
+    inputHistory: ChatInputHistory[]
+    setInputHistory: (history: ChatInputHistory[]) => void
     onSubmit: (text: string, submitType: ChatSubmitType, userContextFiles?: Map<string, ContextFile>) => void
     contextStatusComponent?: React.FunctionComponent<any>
     contextStatusComponentProps?: any
@@ -317,7 +318,8 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
             const rowsCount = (inputValue.match(/\n/g)?.length || 0) + 1
             setInputRows(rowsCount > 25 ? 25 : rowsCount)
             setFormInput(inputValue)
-            if (inputValue !== inputHistory[historyIndex]) {
+
+            if (inputValue !== inputHistory[historyIndex]?.inputText) {
                 setHistoryIndex(inputHistory.length)
             }
         },
@@ -334,7 +336,7 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
             setChatContextFiles(new Map())
             setSelectedChatContext(0)
             setHistoryIndex(inputHistory.length + 1)
-            setInputHistory([...inputHistory, input])
+            setInputHistory([...inputHistory, { inputText: input }])
             setDisplayCommands(null)
             setSelectedChatCommand(-1)
         },
@@ -477,16 +479,16 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
                 return
             }
 
-            if (formInput === inputHistory[historyIndex] || !formInput) {
+            if (formInput === inputHistory[historyIndex]?.inputText || !formInput) {
                 if (event.key === 'ArrowUp' && caretPosition === 0) {
                     const newIndex = historyIndex - 1 < 0 ? inputHistory.length - 1 : historyIndex - 1
                     setHistoryIndex(newIndex)
-                    setFormInput(inputHistory[newIndex])
+                    setFormInput(inputHistory[newIndex]?.inputText)
                 } else if (event.key === 'ArrowDown' && caretPosition === formInput.length) {
                     if (historyIndex + 1 < inputHistory.length) {
                         const newIndex = historyIndex + 1
                         setHistoryIndex(newIndex)
-                        setFormInput(inputHistory[newIndex])
+                        setFormInput(inputHistory[newIndex]?.inputText)
                     }
                 }
             }
--- a/lib/shared/src/index.ts
+++ b/lib/shared/src/index.ts
@@ -1,7 +1,7 @@
 // Add anything else here that needs to be used outside of this repository.
 
 export { ChatContextStatus } from './chat/context'
-export { ChatButton, ChatMessage, ChatError } from './chat/transcript/messages'
+export { ChatButton, ChatInputHistory, ChatMessage, ChatError } from './chat/transcript/messages'
 export { RateLimitError, ContextWindowLimitError } from './sourcegraph-api/errors'
 export { renderCodyMarkdown } from './chat/markdown'
 export { basename, pluralize, isDefined, dedupeWith } from './common'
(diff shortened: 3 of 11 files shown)


DanTup:
@abeatrix do you have any opinions on the above?

abeatrix:
> Since this history is persisted, I think changing the shape of this object would become incompatible with earlier chats and we'd have to handle opening chats that had a string[] history

I think that's fine if we make ContextFile also accepts undefined?

> Some of these changes are outside of vscode/ and I'm not sure if they will break other clients

I'm guessing if all the tests pass for both vscode and agent, we should be ok? We can ask the jetbrain's team for review once this is ready to make sure too.

> I started changing the signature of onHumanMessageSubmitted to accept ChatInputHistory instead of just string, but this method is already accepting a contextFiles?: ContextFile[] which feels a little odd (maybe this is fine because there can be context files not tied to the chat input text?)

It's been a minute since i last look at the code so something might have changed, but from what i remember, we store each user input in `onHumanMessageSubmitted`, and the `contextFiles?: ContextFile[]` in `onHumanMessageSubmitted` actually refers to the context files that are related to the humanInput if there is any, so i think you are on the right track there :D

DanTup:
@abeatrix I simplified this a little and made it compatible with bare strings in the chat history to avoid any issues with potentially persisted history.

@olafurpg I don't entirely understand how other editors may interact with this code (if at all) - is there anything I should be aware of?

DanTup:
Also seems like I have some tests failing that I wouldn't expect to be impacted by this. I re-ran the bot but it failed again. I'll try a third time in case it's a flake and I got unlucky, but if anyone has ideas let me know. Here's a video of what the failure looks like from the bot:

[54baabff464f53b7145f07df03414d60.webm](https://github.com/sourcegraph/cody/assets/1078012/b9dbb1b8-5a66-4ee8-93e5-5e16bce3bb28)

DanTup:
(tests all green after rebasing on main)

abeatrix (approved):
This is awesome 👏

lib/ui/src/Chat.tsx
```diff
@@ -234,6 +235,18 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
...
 
+    // Gets the display text for a context file to be completed into the chat when a user
+    // selects a file.
+    //
+    // This is also used to reconstruct the map from the chat history (which only stores context
+    // files).
+    function getContextFileDisplayText(contextFile: ContextFile): string {
+        const isFileType = contextFile.type === 'file'
+        const range = contextFile.range ? `:${contextFile.range?.start.line}-${contextFile.range?.end.line}` : ''
+        const symbolName = isFileType ? '' : `#${contextFile.fileName}`
```
abeatrix, line 246: Not a blocker since the change is not in main yet, just curious if this is going to work after the changes from Quinn's pr that removes the file name?

DanTup: Assuming @sqs plans to land his PR soon, I'm going to hang fire on this PR and rebase/fix after. I don't want to add conflicts to his change because I'd really like to see it land (it simplifies a bunch of stuff that had made it difficult for me to debug path issues on Windows) :-)

(there might also be some overlap with this function and `inputRepresentation` he added)

DanTup:
@abeatrix I've merged `main` and fixed some minor conflicts. Tests pass locally, will check back on the bots shortly.

@olafurpg would still like to be sure that these changes (in particular, changing chat history from `string` to ` { inputText: string, inputContextFiles: ContextFile[] }`) won't impact IntelliJ (I believe these values are going back via the editor).

lib/shared/src/chat/transcript/messages.ts
```diff
@@ -54,13 +54,25 @@ interface ChatMetadata {
...
     [chatID: string]: TranscriptJSON
 }
 
+/**
+ * We must support bare strings in history because existing users may have histories
+ * persisted with them (and since we support them, we can also use them for entries with
+ * no context files).
+ */
+export type ChatInputHistory =
+    | string
```
olafurpg, line 70: I highly recommend we avoid union types union types like these for anything that we persist to disk because it's awkward to work with. My experience is that it's better to migrate the `string` types  directly to `{ inputText: string }` instead. Union types also can't be typed in many other languages making it difficult for clients like IntelliJ and Neovim to mirror the same structure.

olafurpg (commented):
@DanTup Thank you for the ping! This could definitely be relevant for the IntelliJ client. We added an integration tests for the `chat/restore` functionality last week, and we have integration tests that assert the transcripts from the webview protocol, so this change is most likely not introducing regressions. I left one comment about about the usage of union types for data that we persist to disk

lib/shared/src/chat/transcript/messages.ts
DanTup: Do we have an existing way for migrating data like this? If I drop the `string` here, any data already on disk will be `string` and won't match so clients would need to handle both anyway.

Or do you mean add a new field for the new type, and migrate data into it from the old field as we load it?

olafurpg: We replace chat history for all sessions even when we only update one chat. My understanding is that the history will migrate automatically as long as we handle `string` results from old storage.

olafurpg: Some extra background that might be relevant or not. While working on `chat/restore`, I spent quite a lot of time investigating how we persist chat history and reached the conclusion it's quite messy (example, using ISO date as the ID, weird hierarchy, titles as sibling `string[]` property, ...) and I did not want to spend too much time reverse-engineering the schema because I'm expecting it to change. Then design I landed on for `chat/restore` was to accept `history: ChatMessage[]`, which I was able to load into `SimpleChatModel` here https://sourcegraph.com/github.com/sourcegraph/cody@main/-/blob/agent/src/agent.ts?L566-578

DanTup:
I've pushed some changes to remove the `| string` from the union, and I've added code to migrate the chat during loading from local history (w/ test).

I'm not sure if this is the correct place to migrate (and that there are not other places that might have unmigrated histories).. @olafurpg PTAL and see if this is what you were expecting. Thanks!

DanTup:
@olafurpg @abeatrix I've rebased and fixed the biome issues. Please take a look (in particular at the migration I added since last review, and whether these changes would impact IntelliJ/non-VS Code negatively).

olafurpg:
@DanTup I have not reviewed the changes, but IntelliJ only uses a stateless agent process so we have to restore the chat history from scratch for every new connection. We use `ChatModel.{addBotMessage,addHumanMessage,addErrorAsBotMessage}` to construct transcripts.

DanTup:
@olafurpg would it be affected by the change to `UserLocalHistory` (from `string` to `ChatInputHistory`)? I don't know if it has its own definition of any of these types.

And does the loading go through `LocalStorageProvider` so that the migration could would run (so that chats persisted before this change would be migrated from `string` to `UserLocalHistory` before being sent anywhere else)?

abeatrix:
LGTM and i see the agent test passes for the chat/restore test. 
@sourcegraph/jetbrains-plugin can we get a confirmation from your team to move forward with this change?

olafurpg (approved):
If no changes are required in the agent tests then this should not impact JetBrains. We don't use local history directly, `chat/editMessage` sends a transcript and the agent server creates a new history via `SimpleChatModel`

pkukielka:
> If no changes are required in the agent tests then this should not impact JetBrains. We don't use local history directly, `chat/editMessage` sends a transcript and the agent server creates a new history via `SimpleChatModel`

Agree, it does not seem to impact us in any way.
That said it would be nice to add even minimal coverage of this feature in `index.test.ts`.
It will allow to avoid all doubts in future (as at some point we might start to use it as well).

olafurpg:
> That said it would be nice to add even minimal coverage of this feature in index.test.ts.

I don't think an agent test would help here since the feature is a responsibility of the agent client, not server. Most of the diff is from UI/React code.

PR #2474, opened 2023-12-22
Fix duplication of query when selecting a context file that already exists earlier in the input

### LINKED ISSUE #2243: bug: Selecting a file already included in chat @-mention leaves search query appended
### Version

main 59c4382f

### Describe the bug

When selecting a file from the @-mention search, if it's already in the search box the search query gets duplicated on the end:

https://github.com/sourcegraph/cody/assets/1078012/7d52c490-58f2-4b3c-98e8-ac575d10bcb8

### Expected behavior

The search query should be replaced by the selected file, as it is if you choose a different filename that doesn't already appear:

https://github.com/sourcegraph/cody/assets/1078012/127da762-aee1-42b0-a661-d7121709f057

### Additional context

_No response_

DanTup:
The code here appeared to try to preserve suffixes after the filename, but we appear to only support selecting files at the end of the input anyway. By assuming the index of "fileDisplayText" in the input was the one that the user expected to replace, an input like:

```
@foo.txt @fo
```

Would assume the first `@foo.txt` is being replaced, and retain `@fo` (actually duplicating it).

Removing this suffix fixes the issue and as far as I can tell does not break any other functionality. The issue described in https://github.com/sourcegraph/cody/issues/1973 still appears to work as expected with this change.

Fixes #2243:

https://github.com/sourcegraph/cody/assets/1078012/53055f8b-9b61-49e4-bdf7-66afa4367599

#1973 is still fixed:

https://github.com/sourcegraph/cody/assets/1078012/9053b3ce-0b27-494a-8243-7276406cc6d3

@abeatrix Please take a look, because this partially reverts your changes from #1980. If you think there are cases that this is breaking, let me know and I will add tests and try to fix another way.

## Test plan

- Run `pnpm test:e2e at-file` to run the updated test
- Open chat and try to mention a file a second time without typing the fill name (as shown in the first video above)
- Open chat, complete a filename, then backspace over the added trailing space and replace it with a `?` and ensure it's still used as a context file (second video above)


abeatrix (approved):
I can't remember what was the reason I added the suffix but i just tried it on your branch and confirmed what described in the old issue is not an issue anymore even after your change 👍 

Thanks for adding the test as well!

PR #2478, opened 2023-12-23
VS Code: remove redundant abort controller

valerybugakov:
## Context

- Removes redundant `abortSignal` from the completions `stream` function. It's not exposed to the outside world, so it cannot affect the execution flow of the request. @abeatrix, let me know if I misunderstood the code here.

## Test plan

1. CI
2. Tested manually that the "stop generating" chat feature works as expected.


abeatrix (approved):
Tested build from branch and confirmed the abort button still works as intended:

https://github.com/sourcegraph/cody/assets/68532117/ed9cea1a-27eb-43fa-a0cb-cefe76f5979a

PR #2479, opened 2023-12-23
VS Code: remove redundant cancellation logic from the click handler

valerybugakov:
## Context

- Removes redundant logic from the abort streaming click handler. @abeatrix, let me know if I misunderstood the code here.

## Test plan

1. CI
2. Tested manually that the "stop generating" chat feature works as expected.

### DIFF
--- a/vscode/webviews/Chat.tsx
+++ b/vscode/webviews/Chat.tsx
@@ -89,13 +89,9 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
     enableNewChatUI,
     userInfo,
 }) => {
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
Confirm the stop button still works as it is now. 

Most likely not related to this PR, but the button doesn't work on first click and requires a second click.

https://github.com/sourcegraph/cody/assets/68532117/9b8ccf79-1a53-490e-acc7-318543a39956

valerybugakov:
> the button doesn't work on first click and requires a second click.

Interesting. I cannot repro it on main and on this branch. Did you manage to find the root cause?

Issue #1921, opened 2023-11-28
Autocomplete: replace completion post processing logger with OpenTelemetry spans


valerybugakov:
We need a way to see the stage the streamed completion response goes through to quickly locate bugs and debug efficiently streaming/truncation/post-processing logic. Each generated partial response can start a new OpenTelemtry trace, with spans being stages of our processing pipeline with an ability to log all this information in a readable way to the trace view or the console output.

- Part of https://github.com/sourcegraph/cody/issues/1619
- Follow up for https://github.com/sourcegraph/cody/pull/1894

beyang:
Closeable?

valerybugakov:
Yes, it will be closed by the three PRs I opened last week. The are linked above.

PR #2487, opened 2023-12-26
Autocomplete: improve autocomplete tracing util

valerybugakov:
## Context

- The only functional change is that the `autocomplete.generate` span now wraps all the logic related to completion generation instead of only wrapping a promise which is sufficient for duration calculation but does not allow use to created nested spans automatically based on the current async context.
- Type-safety improvement: allows to use `wrapInActiveSpan` with sync functions without converting them to async functions on a type level.
- Part of #1921 

## Test plan

CI


lib/shared/src/tracing/index.ts
```diff
@@ -17,18 +17,34 @@ export function getActiveTraceAndSpanId(): { traceId: string; spanId: string } |
...
+            const response = fn()
+
+            if (response instanceof Promise) {
+                return response.then(handleSuccess, catchError) as R
+            }
+
+            return handleSuccess(response)
+        } catch (error) {
+            catchError(error)
+            throw error
```
abeatrix, line 43: Why do we need to throw error again after throw error in catchError?

abeatrix (approved):
Left a question in line otherwise lgtm!

lib/shared/src/tracing/index.ts
valerybugakov: Typescript is not happy without it because it does not understand that `catchError` always throws. I pushed a commit that duplicates `throw error` to make it less confusing.

valerybugakov: Got a better solution to this: https://github.com/sourcegraph/cody/pull/2487/commits/dbdb92e14f2b20417309ec0178fb5ac7686e8e86

PR #2489, opened 2023-12-26
Autocomplete: add console-exporter for autocomplete OpenTelemetry tracing

valerybugakov:
## Context

- Replaces the deprecated `completionPostProcessLogger` with console-exporter for autocomplete OpenTelemetry tracing. It prints OpenTelemetry traces with their attributes and events to the console. It helps to debug the autocomplete data flow quickly.
- Next up: reuse the new logic for tracing-view-exporter to present the same information there.
- Builds on top of https://github.com/sourcegraph/cody/pull/2487 and https://github.com/sourcegraph/cody/pull/2486
- Closes https://github.com/sourcegraph/cody/issues/1921

## Test plan

1. Merge https://github.com/sourcegraph/cody/pull/2487 and https://github.com/sourcegraph/cody/pull/2486.
2. Set `cody.experimental.tracing` to `true`.
3. Connect standalone Chrome Dev tools.
4. Generate a code completion.
5. See traces printed to the console.
6. Disclaimer: `console.group()` titles are duplicated. It's a node.js bug: https://github.com/nodejs/node/issues/31973 
Screenshot 2023-12-26 at 17 13 26


vscode/src/services/open-telemetry/debug-utils.ts
```diff
@@ -0,0 +1,15 @@
+import { Attributes, Span, trace } from '@opentelemetry/api'
+
+export const addDebugEventToActiveSpan = (name: string, attributes: Record<string, unknown> = {}): Span | void => {
```
abeatrix, line 3: Looks like this is built for auto-complete specifically, should we rename it to reflect that?

abeatrix (approved):
lgtm! I have a few questions regarding how opentelemetry works:
- is opentelemetry going to replace our current telemetry service or is it an additional service? 
- i remember you or rob talked about how we won't need to create a request id manually with opentelementry --is this something that is being used that I can use as reference?

Thank you :D

valerybugakov (commented):
> is opentelemetry going to replace our current telemetry service or is it an additional service?

OpenTelemetry is an additional service that we currently use to collect only one type of signal — traces. The existing telemetry service is used to collect logs, and OpenTelemetry inspires its API, but it's a separate thing. 

> i remember you or rob talked about how we won't need to create a request id manually with opentelementry --is this something that is being used that I can use as reference?

Need a bit more clarification here 😛

PR #2520, opened 2023-12-27
Pass URIs for some context snippets

beyang:
Previously, we used the relative file name to indicate the source file of all context snippets. This causes problems, because:
* The user may have more than one workspace folder open, in which case we don't know how to resolve a relative file name, because we lose track of which workspace folder it came from
* Context from remote sources (e.g., remote embeddings) isn't attached to a local file, and we may need to treat these differently when trying to open the file

Add a URI field to keep better track of the true location of a context item. Adds creator functions for 2 URI schemes: `cody-file-relative://` and `cody-embeddings://`.

Also add an `openUri` function that handles opening these URIs and is more forgiving for `file://` URIs (will attempt to match a relative file URI with an open workspace folder).

## Test plan

1. Open at least 2 workspace folders
2. Open a file from the 2nd workspace folder
3. Create a new chat and ask a question
4. Attempt to open the context files—these should open properly, whereas before, nothing occurred

See screen recordings for examples.

Before: https://github.com/sourcegraph/cody/assets/1646931/4ae6b2a1-ce00-4d3a-86bd-25effae6495c
After: https://github.com/sourcegraph/cody/assets/1646931/2d72e149-0f0f-4704-9b2b-3fab699a5b3e

### DIFF
--- a/vscode/src/chat/chat-view/ChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/ChatPanelProvider.ts
@@ -123,7 +123,7 @@ export class ChatPanelProvider extends MessageProvider {
                 void openExternalLinks(message.value)
                 break
             case 'openFile':
-                await openFilePath(message.filePath, this.webviewPanel?.viewColumn, message.range)
+                await openFilePath(message.filePath, undefined, this.webviewPanel?.viewColumn, message.range)
                 break
             case 'openLocalFileWithRange':
                 await openLocalFileWithRange(message.filePath, message.range)
--- a/vscode/src/chat/chat-view/SidebarChatProvider.ts
+++ b/vscode/src/chat/chat-view/SidebarChatProvider.ts
@@ -145,7 +145,7 @@ export class SidebarChatProvider extends MessageProvider implements vscode.Webvi
                 void openExternalLinks(message.value)
                 break
             case 'openFile':
-                await openFilePath(message.filePath, this.webviewPanel?.viewColumn)
+                await openFilePath(message.filePath, undefined, this.webviewPanel?.viewColumn)
                 break
             case 'openLocalFileWithRange':
                 await openLocalFileWithRange(message.filePath, message.range)
--- a/lib/ui/package.json
+++ b/lib/ui/package.json
@@ -30,6 +30,7 @@
   "dependencies": {
     "@mdi/js": "^7.2.96",
     "@sourcegraph/cody-shared": "workspace:*",
-    "classnames": "^2.3.2"
+    "classnames": "^2.3.2",
+    "vscode-uri": "^3.0.7"
   }
 }
--- a/lib/ui/src/chat/components/EnhancedContext.tsx
+++ b/lib/ui/src/chat/components/EnhancedContext.tsx
@@ -1,10 +1,13 @@
 import React from 'react'
 
+import { URI } from 'vscode-uri'
+
 import { ActiveTextEditorSelectionRange, ContextFile } from '@sourcegraph/cody-shared'
 
 import { TranscriptAction } from '../actions/TranscriptAction'
 
 export interface FileLinkProps {
+    uri?: URI
     path: string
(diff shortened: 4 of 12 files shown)


arafatkatze (changes requested):
Facing issues in my [debugger](https://www.loom.com/share/59155e84cda4441da22ce381b60d4dbe?sid=ccf69547-640a-499c-bd75-5624f6992620)[](url)

abeatrix (commented):
The links in the context files are not working for me for any workspace 🤔 

https://www.loom.com/share/797f9a07f1d949f3933b6702081a25cf?sid=abd68722-8c34-4a46-b2ca-82dd934a4ea4

https://www.loom.com/share/797f9a07f1d949f3933b6702081a25cf

beyang:
@abeatrix pushed a patch, PTAL

@arafatkatze can you share the output of your debugger console? I want to see why chat is hanging

vscode/src/chat/chat-view/SidebarChatProvider.ts
```diff
@@ -145,7 +145,7 @@ export class SidebarChatProvider extends MessageProvider implements vscode.Webvi
                 void openExternalLinks(message.value)
                 break
             case 'openFile':
-                await openFilePath(message.filePath, this.webviewPanel?.viewColumn)
+                await openFilePath(message.filePath, undefined, this.webviewPanel?.viewColumn)
```
abeatrix, line 148:
```suggestion
                await openFilePath(message.filePath, message.uri, this.webviewPanel?.viewColumn)
```

`message.uri` should also work here?

vscode/src/chat/chat-view/ChatPanelProvider.ts
```diff
@@ -123,7 +123,7 @@ export class ChatPanelProvider extends MessageProvider {
                 void openExternalLinks(message.value)
                 break
             case 'openFile':
-                await openFilePath(message.filePath, this.webviewPanel?.viewColumn, message.range)
+                await openFilePath(message.filePath, undefined, this.webviewPanel?.viewColumn, message.range)
```
abeatrix, line 126:
```suggestion
                await openFilePath(message.filePath, message.uri, this.webviewPanel?.viewColumn, message.range)
```

abeatrix: (Not related to this PR) should we start working on removing this old `ChatPanelProvider` since we have migrated to `SimpleChatPanelProvider`?

abeatrix (approved):
Left some questions inline. Confirmed the changes in the last commit worked for me: https://www.loom.com/share/eae552bbf3b844f387f8fb9570ce13c3?sid=b03b84fa-542d-47ce-98c3-127d46287170

Here are the steps I tried:
1. Open a workspace folder
2. Open a file from workspace folder
3. Create a new chat and ask a question
4. Attempt to open the context files - worked
5. Add a 2nd workspace folder
6. Open a file from the 2nd workspace folder
7. Create a new chat and ask a question
8. Attempt to open the context files - worked
9. Switched back to the 1st chat window and attempt to open the context files again - worked as expected

vscode/src/chat/chat-view/ChatPanelProvider.ts
beyang: yes, we should!

vscode/src/chat/chat-view/SidebarChatProvider.ts
beyang: This was intentional, as I didn't want to interfere with any legacy behavior. We fall back to using the file path if the uri is undefined.

arafatkatze (approved):
I haven't reviewed the code of this PR but the steps and the functionality outlined by Beatrix above works perfectly for me. I am doing other tasks so that's why I didn't do a proper review of this but I can verify the functionality outlined in the PR works. 

Message if proper review is required too.

PR #2521, opened 2023-12-28
Chat attribution UI

cbart:
Part of [sourcegraph#58993.](https://github.com/sourcegraph/sourcegraph/issues/58993)

Implements a Chat UI for guardrails following direction from @toolmantim. Let's look at some screenshots:

First with experimental setting turned off:

off

Once we turn on the experimental setting, each snippet in chat gets a gray shield icon on the right hand side of the pane below snippet. This is search running:

running

The happy path is that we find no attribution. In that case icon turns green and we update relevant tooltip.

notFound

Alternative is that attribution is found. In such case icon turns red. The tooltip lists one repository where attribution was found, and counts the other hits:

found

An exceptional case is when guardrails back end is unavailable. In this case the icon turns yellow and the tooltip updates:

unavailable

## Test plan

Visual and manual.

I'd love some review feedback on how could we go and implement a story here. I'm not familliar yet with VS Code ui testing.


abeatrix (commented):
Which instance should I connect to to test this? I tried it with both s2 and dot com and is getting the icon that indicates guardrails back end is not available in both cases.

With setting off:

image

With setting on and logged into s2:

image

With setting on and logged into dot com:

image

cbart:
@abeatrix correct! This is just the UI and it's wired up to a search method that always returns unavailable. The back end is not ready yet.

abeatrix (approved):
Approving to unblock.

We currently do not have an instance with guardrails backend setup correctly to test this completely as mentioned in https://github.com/sourcegraph/cody/pull/2521#pullrequestreview-1798283441, but the icon is showing up correctly to indicate the backend is not available.

toolmantim (commented):
Started [a Slack thread](https://sourcegraph.slack.com/archives/C05JDP433DL/p1704249867447619) for the design side of this

PR #2528, opened 2023-12-29
refactor: remove old chat panel provider

abeatrix:
PR to remove the old ChatPanelProvider with additional clean-ups:

1. Rename SidebarChatProvider to SidebarViewProvider for clarity as it is no longer used for Chat View.
  - It is used for auth flows and running tasks that don't require a chat view.
2. SidebarViewProvider has been simplified.
  - All functions related to the chat view are removed
2. Remove ChatPanelProvider in favor of the new SimpleChatPanelProvider

## Test plan

All features of Cody should work the same. All existing tests should be passing.

image

Make sure all the current features work correctly:

- [ ] auth flow
- [ ] ask cody a question in chat
  - [ ] cody can response
  - [ ] question include correct context
- [ ] core commands
- [ ] edit command
- [ ] custom command
- [ ] export chat history
- [ ] remove single chat history
- [ ] remove all chat history


vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -16,24 +16,24 @@ import { CachedRemoteEmbeddingsClient } from '../CachedRemoteEmbeddingsClient'
...
 export const CodyChatPanelViewType = 'cody.chatPanel'
 /**
  * Manages chat view providers and panels.
  */
 export class ChatManager implements vscode.Disposable {
-    // View in sidebar for auth flow and old chat sidebar view
+    // SidebarView is used for auth view and running tasks that do not require a chat view
     // We will always keep an instance of this around (even when not visible) to handle states when no panels are open
-    public sidebarChat: SidebarChatProvider
+    public sidebarView: SidebarViewProvider
```
beyang, line 28: I think a more accurate name for this field would be `sidebarViewController`.

vscode/src/main.ts
```diff
@@ -231,7 +231,7 @@ const register = async (
     // Register tree views
     disposables.push(
         chatManager,
-        vscode.window.registerWebviewViewProvider('cody.chat', chatManager.sidebarChat, {
+        vscode.window.registerWebviewViewProvider('cody.chat', chatManager.sidebarView, {
```
beyang, line 234: Should this still be registered under the id `cody.chat`?

vscode/src/chat/chat-view/SidebarViewProvider.ts
```diff
@@ -0,0 +1,184 @@
...
+
+export interface SidebarChatWebview extends Omit<vscode.Webview, 'postMessage'> {
+    postMessage(message: ExtensionMessage): Thenable<boolean>
+}
+
+export interface SidebarViewOptions extends MessageProviderOptions {
+    extensionUri: vscode.Uri
+}
+
+export class SidebarViewProvider extends MessageProvider implements vscode.WebviewViewProvider {
```
beyang, line 24: Can we get rid of `MessageProvider` entirely at some point? (Doesn't have to be in this PR)

abeatrix: 100%! Maybe I can work with @umpox in the next sprint to see how we can retire this and migrate fixup and commands to use the simple chat context instead?

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -285,14 +292,17 @@ export class ChatPanelsManager implements vscode.Disposable {
     }
 
     public async clearHistory(chatID?: string): Promise<void> {
```
beyang, line 294: The old code path would go through `MessageProvider.clearHistory`, which would post the update to the sidebar web view. Do we still need to do that for the SidebarViewProvider?

beyang (approved):

vscode/src/chat/chat-view/ChatPanelsManager.ts
abeatrix: We won't need to anymore because the chat history view now lives in the tree view, and the sidebar webview is only used for auth view which doesn't read the chat history.

vscode/src/chat/chat-view/SidebarViewProvider.ts
umpox: Yes! I'm actually already working on this as part of my edit refactor/cleanup. I'm moving away from `MessageProvider` so I think we should be able to remove that once that is done.

vscode/src/main.ts
abeatrix: I looked into this and tried to rename it from `cody.chat` to `cody.sidebar` but ran into issues with the `cody.chat.focus` command (somehow it's behaving differently). 

I'll create a new issue and address this in a separate PR so that this PR will not be blocked by it.

Issue #2536, opened 2024-01-02
bug: Enhanced context added for some commands


abeatrix:
Issue for "Enhanced context interferes with smell command" filed in https://github.com/sourcegraph/cody/issues/2464

Enhanced context is added for /smell commands and some custom commands and not respecting the context fields specified in the cody.json file.

For example, only the current file should be used for the `/smell` command, but additional context from enhanced context was added:

![Image](https://github.com/sourcegraph/cody/assets/68532117/19f7922c-d304-4604-bdbd-6d301a1f03b2)

PR #2537, opened 2024-01-02
fix: remove enhanced context from commands

abeatrix:
Close https://github.com/sourcegraph/cody/issues/2536

1. Removed the 'addEnhancedContext' parameter from the addCommand() method in CommandsController. 
2. All commands should be run without enhancedContext unless it's specified in the cody.json to include `codybase` as context

We previously added extra context to custom command which caused confusion to LLM and affected command quality. 

## Test plan

1. Highlight some code in a file
2. Run `/smell` command
4. Check and make sure only 1 file (current file) is used as context.

Before:

On top of the current file, it would add other unrelated files as enhanced context

image

After:

It should only use the correct file as context and nothing else

image

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -6,12 +6,13 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


toolmantim (approved):
👍🏼 We should update the PR title to mention it affects all commands though, not just custom commands.

PR #2538, opened 2024-01-02
refactor: update recipe telemetry events and naming

abeatrix:
Related to slack conversation in https://sourcegraph.slack.com/archives/C05AGQYD528/p1703807507189849

The recipe telemetry events were updated to use 'recipe-used' instead of 'executed' to better reflect when a recipe is triggered.

The recipe event names were also updated to remove the 'recipe:' prefixfor brevity and consistency.

## Test plan

- Run an edit command.
- The recipe event has been updated from `CodyVSCodeExtension:recipe:fixup:executed` to `CodyVSCodeExtension:fixup:recipe-used`

Logs:

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:command:edit:executed VSCode {"properties":{"source":"editor"},"opts":{"hasV2Event":true}}
█ telemetry-v2: recordEvent (no-op): cody.command.edit/executed: {"parameters":{"version":0,"privateMetadata":{"source":"editor"}},"timestamp":"2024-01-02T18:36:38.604Z"}
█ MessageProvider:executeRecipe: fixup  "lqwoyx"
█ logEvent (telemetry disabled): CodyVSCodeExtension:fixup:recipe-used VSCode {"properties":{"contextSummary":{"embeddings":3,"local":2,"user":0},
```

image


vscode/src/chat/MessageProvider.ts
```diff
@@ -469,8 +469,8 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
 
         const promptText = this.isDotComUser ? interaction.getHumanMessage().text : undefined
         const properties = { contextSummary, source, requestID, chatModel: this.chatModel, promptText }
-        telemetryService.log(`CodyVSCodeExtension:recipe:${recipe.id}:executed`, properties, { hasV2Event: true })
-        telemetryRecorder.recordEvent(`cody.recipe.${recipe.id}`, 'executed', { metadata: { ...contextSummary } })
+        telemetryService.log(`CodyVSCodeExtension:${recipe.id}:recipe-used`, properties, { hasV2Event: true })
+        telemetryRecorder.recordEvent(`cody.recipe.${recipe.id}`, 'used', { metadata: { ...contextSummary } })
```
chenkc805, line 473: What's this "used" event here? I've never see a suffix "used" event in our telemetry. CC: @kelsey-brown

chenkc805:
Thanks @abeatrix ! 

Confirming this would work with our other commands, like `doc`? Expected behavior in the `doc` case:
* when someone runs doc, we log the `doc:executed` event and `fixup:recipe-used` event.

chenkc805 (changes requested):

kelsey-brown:
IMO we don't need to rename the `executed` events to `used`, we just need to add a _new_ event to distinguish between user-invoked commands and command-invoked commands.

For example, 
- User invokes the doc command (which uses fixup): `doc:executed ` event fires, AND new `fixup:recipe-used` event fires
- User invokes the fixup command: `fixup:executed` fires 

FWIW, I think adding a new event for command-invoked commands (`fixup:recipe-used` in the above example) is low priority. The high priority is making sure that the existing `executed` events only fire on user-invoked commands

abeatrix:
> * User invokes the fixup command: `fixup:executed` fires

@kelsey-brown  There is no command called  `fixup`. `fixup` is an underlying recipe that is used by all `edit` commands, including the `/doc` command. 

After this PR:
- All commands executed will be logged with `executed` event. 
- All recipes that is used will be logged with `recipe-used` event

taylorsperry:
Thanks, @abeatrix! What do/would we log if a user triggered the "Ask Cody to Fix" VS Code Action?

abeatrix:
> Thanks, @abeatrix! What do/would we log if a user triggered the "Ask Cody to Fix" VS Code Action?

@taylorsperry That'd use the `edit` command which uses the `fixup` recipe under the hood

vscode/src/chat/MessageProvider.ts
abeatrix: updated from `used` to `recipe-used`

abeatrix:
> Confirming this would work with our other commands, like `doc`? Expected behavior in the `doc` case:
> 
> * when someone runs `doc`, we log the `doc:executed` event and `fixup:recipe-used` event.

@chenkc805 this PR updates the event name from the old event name `CodyVSCodeExtension:recipe:fixup:executed` to `CodyVSCodeExtension:fixup:recipe-used` without any changes to how the commands are currently logged, so when an `edit` command, including `doc` is executed, all the old `execute` events for the command will be logged as is, on top of the new `CodyVSCodeExtension:fixup:recipe-used` event. 

Here is the logs when the doc command is executed. 

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:command:doc:executed VSCode {"properties":{"mode":"insert","useCodebaseContex":false,"useShellCommand":false,"requestID":"7e39c694-72e9-46a2-9604-b54acc204fd6"}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:command:edit:executed VSCode {"properties":{"source":"doc"},"opts":{"hasV2Event":true}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:fixup:recipe-used VSCode {"properties":{"contextSummary":{"embeddings":0,"local":1,"user":0},"source":"doc","requestID":"6f5a968b-1cd6-40f8-97fd-fa634759e94d","promptText":...
```

Again, this PR didn't make any changes of the current logging behavior so the `edit` command is logged when `doc` command is executed as this is the current behavior. We will need a separate PR to fix this as that is a separate issue unrelated to the fixup recipe.

chenkc805:
> Again, this PR didn't make any changes of the current logging behavior so the edit command is logged when doc command is executed as this is the current behavior. We will need a separate PR to fix this as that is a separate issue unrelated to the fixup recipe.

Can we spin up a separate PR for this then? This wouldn't solve our root problem of making sure that we're logging events 1:1 for user-invoked commands.

In the `docs` case, it seems like `edit` and `fixup` are should actually be `recipe-used`, since that's a command-invoked command. 

So in summary:
- For any user-invoked command, use `executed` suffix. This telemetry should always be 1 event to 1 user action.
- For any command-invoked command, use `recipe-used` suffix. This telemetry can be N events to 1 user action

abeatrix:
> > Again, this PR didn't make any changes of the current logging behavior so the edit command is logged when doc command is executed as this is the current behavior. We will need a separate PR to fix this as that is a separate issue unrelated to the fixup recipe.
> 
> 
> 
> Can we spin up a separate PR for this then, seeing as this PR alone wouldn't solve our root problem of making sure that we're logging events 1:1 for user-invoked commands.
> 
> 
> 
> In the `docs` case, it seems like `edit` and `fixup` are should actually be `recipe-used`, since that's a command-invoked command. 
> 
> 
> 
> So in summary:
> 
> - For any user-invoked command, use `executed` suffix. This telemetry should always be 1 event to 1 user action.
> 
> - For any command-invoked command, use `recipe-used` suffix. This telemetry can be N events to 1 user action

@umpox and I will be working on refactoring the edit command and other commands, so we can work on the 1-1 event mapping issue after that. Can you create an issue for that for tracking purpose? @chenkc805 

This PR is just to rename the recipe execute events to the recipe-used which is related to your second point. If we don't need to rename the recipe execute events at all, we can close this PR.

chenkc805:
Thanks, done here: https://github.com/sourcegraph/cody/issues/2545

taylorsperry:
Re: commands vs recipes, I know that from the user's perspective we just have "commands" (user-initiated actions that Cody performs, like "document" or "explain" or "edit"), but behind the scenes we have recipes that power those commands (ie, the "doc" command uses the "fixup" recipe). Is that the right way to think about it?

chenkc805 (approved):

PR #2540, opened 2024-01-02
Chat: fallback to local keyword search when embeddings fail

jtibshirani:
Before, if both embeddings search methods fail then we display an error in chat. Now, we just log the error and fall back to local keyword search. This avoids a situation we've seen at customer sites where they hit the embeddings rate limit, then their users run into a bunch of errors in chats. 

I don't think this is actually a change in behavior, because the local embeddings client swallows errors and returns "no results", which caused us to fall back to keyword search (accidentally?) It really just clarifies the logic.

## Test plan

Updated embeddings search methods to throw errors (just for testing), and checked we still produce codebase context.


chwarwick (approved):
This make sense to me, to use whatever context is available instead of failing.

PR #2546, opened 2024-01-03
Autocomplete: skip request manager cache on manual completion trigger

valerybugakov:
## Context

- Bypass request manager cache on manual completion triggers.
- We already have the last candidate cache disabled for this case. The request manager cache was implemented after that we never got to skipping it too. 

## Test plan

- Use the keyboard shortcut to trigger a completion generation multiple times at the same cursor position.
- Every time, a new request should be issued according to Cody's output channel.

### DIFF
--- a/vscode/src/completions/request-manager.ts
+++ b/vscode/src/completions/request-manager.ts
@@ -37,6 +37,14 @@ export interface RequestManagerResult {
     source: InlineCompletionsResultSource
 }
 
+interface RequestsManagerParams {
+    requestParams: RequestParams
+    providers: Provider[]
+    context: ContextSnippet[]
+    isCacheEnabled: boolean
+    tracer?: CompletionProviderTracer
+}
+
 /**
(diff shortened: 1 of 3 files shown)


umpox (approved):

vscode/src/completions/request-manager.ts
```diff
@@ -64,26 +72,24 @@ export class RequestManager {
...
-        tracer?: CompletionProviderTracer
-    ): Promise<RequestManagerResult> {
-        const cachedCompletions = this.cache.get(params)
-        if (cachedCompletions) {
+    public async request(params: RequestsManagerParams): Promise<RequestManagerResult> {
+        const { requestParams, providers, context, isCacheEnabled, tracer } = params
+
+        const cachedCompletions = this.cache.get(requestParams)
+        if (isCacheEnabled && cachedCompletions) {
+            console.log('using cache', cachedCompletions)
```
philipp-spiess, line 80: Leftover console.log :)

Issue #2545, opened 2024-01-03
bug: User-invoked commands are not firing events in a 1:1 manner


chenkc805:
### Version

1.0

### Describe the bug

Currently, for commands like `doc`, and `edit`, they log multiple `executed` commands at the same time, making it difficult for Sourcegraphers to conduct analysis on how often users are executing certain commands.

For example, here's what we log today for an `edit` command:
```
logEvent (telemetry disabled): CodyVSCodeExtension:command:edit:executed VSCode {"properties":{"source":"editor"},"opts":{"hasV2Event":true}}
logEvent (telemetry disabled): CodyVSCodeExtension:fixup:executed VSCode {"properties":{"contextSummary":{"embeddings":3,"local":2,"user":0},
```

and here's what we'd log for the `doc` command:
```
logEvent (telemetry disabled): CodyVSCodeExtension:command:doc:executed VSCode {"properties":{"mode":"insert","useCodebaseContex":false,"useShellCommand":false,"requestID":"7e39c694-72e9-46a2-9604-b54acc204fd6"}}
logEvent (telemetry disabled): CodyVSCodeExtension:command:edit:executed VSCode {"properties":{"source":"doc"},"opts":{"hasV2Event":true}}
logEvent (telemetry disabled): CodyVSCodeExtension:fixup:executed VSCode {"properties":{"contextSummary":{"embeddings":0,"local":1,"user":0},"source":"doc","requestID":"6f5a968b-1cd6-40f8-97fd-fa634759e94d","promptText":...
```

Since so many `executed` events are being fired for just 1 user-invoked command, this makes it hard to understand how many times is `doc` vs `fixup` vs `edit` is actually being used by end users. 

### Expected behavior

Expected behavior is that for each **user-invoked command or chat**:
- the suffix of the telemetry is `executed`
- there's 1 and only 1 `executed` event for each user-invoked command or chat

For commands that invoke other commands (like `edit` using the underlying `fixup` command), we should log:
- the suffix of the telemetry as `recipe-used` (like we're doing in https://github.com/sourcegraph/cody/pull/2538)
- there can be N number of `recipe-used` events for each user-invoked command or chat.  

### Additional context
Discussion: https://github.com/sourcegraph/cody/pull/2538
Relevant slack thead: https://sourcegraph.slack.com/archives/C05AGQYD528/p1703807507189849

PR #2549, opened 2024-01-03
Edit: Refactor away from `MessageProvider` and `Recipe`

umpox:
## Description

Edit is now:
- Using it's own logic instead of sharing `MessageProvider`. 
- No longer a recipe, all logic lives inside of `/edit` and `non-stop`. In a follow up PR I will consolidate the logic in `non-stop` into this directory too and clean that up.
- Refactored to easily support changes without affecting other commands/chat/recipes.
- Refactored to support easily adding prompts for new models such as GPT4

closes https://github.com/sourcegraph/cody/issues/2545

## Test plan

Followed edit test plans in https://docs.google.com/document/d/1all3z8dPs52amhNs3m4O1Lu7V6VsCaU7jXovprte-EE/edit#heading=h.o8dycc3mlio

Followed telemetry test plan via https://github.com/sourcegraph/cody/issues/2545


abeatrix (changes requested):
I tried the build from this branch but the comments are not working for me:


Doc command is not working:

https://github.com/sourcegraph/cody/assets/68532117/7ba40757-2dfd-4ae2-a419-4727adb28a9d

Edit command is also not working:

https://github.com/sourcegraph/cody/assets/68532117/8dea235c-fd6b-4cdc-99ff-219c81c4d18d

Other non-edit commands will open a new chat panel but they are all stuck in loading screen: 

image

umpox:
@abeatrix I can't replicate 🤔 Do you have any logs in your console (VS Code dev tools or the Cody Output log)? The diff should definitely not affect any chat related code, it seems like the extension hasn't activated properly.

https://github.com/sourcegraph/cody/assets/9516420/ac80ef28-e317-4269-917c-af5e10fa927a

abeatrix (approved):
That was my bad. I was testing it on files that were added to my `.cody/ignore` file so they were not showing up as expected 😅 

Confirmed the edit commands are working correctly after the change!

Uploading Screen Recording 2024-01-08 at 9.26.02 AM.mov…

vscode/src/edit/manager.ts
```diff
@@ -0,0 +1,118 @@
...
+        telemetryService.log(
+            `CodyVSCodeExtension:command:${commandEventName}:executed`,
+            { source },
+            { hasV2Event: true }
+        )
+        telemetryRecorder.recordEvent(`cody.command.${commandEventName}`, 'executed', { privateMetadata: { source } })
+
+        const editor = getEditor()
+        if (editor.ignored) {
+            console.error('File was ignored by Cody.')
```
abeatrix, line 60:
```suggestion
            void vscode.window.showInformationMessage('Cannot edit Cody ignored file.')
```

abeatrix:
we can also remove 'registerTreeView' and `taskViewProvider` from FixupController?

umpox:
@abeatrix I think so, I'll do it in a follow up PR

vscode/src/edit/manager.ts
umpox: Good idea 👍

PR #2550, opened 2024-01-03
Move VS Code chat attribution search to extension process

cbart:
Part of https://github.com/sourcegraph/sourcegraph/issues/58993
Following on #2521

Early draft - goal is to move the attribution search to the extension process out of the webview process.

Here this is done through:

- extending the protocol with a `attribution-search` command and an `attribution` result message type.
- a `GuardrailsPost` object that synchronizes between a back-and-forth on guardrails searches (per snippet).
  - This object when asked to perform search, issues a `Promise`
  - and fires a call back (which in use is a `postMessage` to the extension process).
  - Once the extension returns back with a `postMessage` with attribution (or error), it notifies the object, and thus the `Promise`. 

## Test plan

Manual. Please recommend a UI level testing approach if we have one.


lib/shared/src/guardrails/index.ts
```diff
@@ -14,6 +15,55 @@ export interface Guardrails {
...
+
+    public notifyAttributionFailure(snippet: string, error: Error): void {
+        const request = this.currentRequests.get(snippet)
+        if (request !== undefined) {
+            request.reject(error)
+        }
+    }
+}
+
+class DeferredAttributionSearch {
```
arafatkatze, line 54: Confused by the naming: The results are blocking on this and not deferred so how is this called deferred?

arafatkatze (approved):
Looks okay for now I  would love to see it all in action eventually.

lib/shared/src/guardrails/index.ts
cbart: Good point, generally it's a synchronization point with a `Promise` and a `resolve` for a single attribution search. Do you have a good idea for a name?

arafatkatze: `BlockingAttributionSearch` or `SynchronousAttributionSearch` could be a good fit.

cbart: Good ideas! Amended.

arafatkatze (approved):
Approving again

lib/shared/src/guardrails/index.ts
```diff
@@ -14,6 +14,58 @@ export interface Guardrails {
...
+class AttributionSearchSync {
+    public promise: Promise<Attribution>
+    public resolve!: (result: Attribution) => void
+    public reject!: (cause: any) => void
+
+    constructor() {
+        this.promise = new Promise<Attribution>((resolve, reject) => {
+            this.resolve = resolve
+            this.reject = reject
+        })
```
philipp-spiess, line 65: Should we add a timeout here too in case something in the RPC goes wrong? It does feel a bit brittle, doesn't it? 😅

philipp-spiess (changes requested):
I do think the `currentRequests` is leaking memory so rejecting to avoid merging this, otherwise this looks good to me!

lib/shared/src/guardrails/index.ts
```diff
@@ -14,6 +14,58 @@ export interface Guardrails {
     searchAttribution(snippet: string): Promise<Attribution | Error>
 }
 
+// GuardrailsPost implements Guardrails interface by synchronizing on message
+// passing between webview and extension process.
+export class GuardrailsPost implements Guardrails {
+    private currentRequests: Map<string, AttributionSearchSync> = new Map()
```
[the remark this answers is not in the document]
cbart: Thanks for pointing this out! I was going to perform garbage-collection in notify methods but forgot.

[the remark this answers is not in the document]
cbart: Done now.

lib/shared/src/guardrails/index.ts
cbart: Added in GuardrailsPost, so that notify* method is invoked.

cbart:
Thanks @philipp-spiess ! Please take another look!

philipp-spiess (approved):
Awesome!

PR #2559, opened 2024-01-04
Making the Cody UI configurable by Site Admin

arafatkatze:
Main Issue https://github.com/sourcegraph/sourcegraph/issues/58631
Related SourceGraph PR: https://github.com/sourcegraph/sourcegraph/pull/59229

To understand just watch 
1. [Video Description of this PR
](https://www.loom.com/share/ec5fc79db52d401491dfa2674ecf7125?sid=b5de0bc2-fdd2-4a6a-b346-289cd263f011)

2. [How to connect Cody to SG Instance
](https://www.loom.com/share/ec5fc79db52d401491dfa2674ecf7125?highlightComment=27977469&t=230)

## UI Changes

1. New UI For Chat
image

2. New UI for Commands
image

3. New UI for Autocomplete
image

## Test plan
Already described in the video

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -9,6 +9,7 @@ import { isError } from '../../utils'
 import { DOTCOM_URL, isDotCom } from '../environments'
 
 import {
+    CURRENT_SITE_CODY_CONFIG_FEATURES,
     CURRENT_SITE_CODY_LLM_CONFIGURATION,
     CURRENT_SITE_CODY_LLM_PROVIDER,
     CURRENT_SITE_GRAPHQL_FIELDS_QUERY,
@@ -72,8 +73,18 @@ interface CurrentUserInfoResponse {
         displayName?: string
         username: string
         avatarURL: string
+        codyProEnabled: boolean
         primaryEmail?: { email: string } | null
     } | null
+}
+interface CodyConfigFeatures {
+    chat: boolean
+    autoComplete: boolean
+    commands: boolean
+}
+
+interface CodyConfigFeaturesResponse {
+    site: { codyConfigFeatures: CodyConfigFeatures | null } | null
 }
 
 interface CurrentUserCodyProEnabledResponse {
@@ -326,6 +337,25 @@ export class SourcegraphGraphQLAPIClient {
                 extractDataOrError(response, data =>
                     data.currentUser ? { ...data.currentUser } : new Error('current user not found')
                 )
+        )
+    }
+
+    /**
+     * Fetches the Site Admin enabled/disable Cody config features for the current instance.
+     */
+    public async getCodyConfigFeatures(): Promise<
+        | {
+              chat: boolean
+              autoComplete: boolean
+              commands: boolean
+          }
+        | Error
+    > {
+        return this.fetchSourcegraphAPI<APIResponse<CodyConfigFeaturesResponse>>(
+            CURRENT_SITE_CODY_CONFIG_FEATURES,
+            {}
+        ).then(response =>
+            extractDataOrError(response, data => data.site?.codyConfigFeatures ?? new Error('cody config not found'))
         )
     }
 
@@ -675,6 +705,81 @@ export class SourcegraphGraphQLAPIClient {
  * Should be configured on the extension activation via `graphqlClient.onConfigurationChange(config)`.
  */
 export const graphqlClient = new SourcegraphGraphQLAPIClient()
+
+/**
+ * ConfigFeaturesSingleton is a class that manages the retrieval
+ * and caching of configuration features from GraphQL endpoints.
+ */
+export class ConfigFeaturesSingleton {
+    private static instance: ConfigFeaturesSingleton
+    private configFeatures: Promise<{
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    }>
+
+    // Constructor is private to prevent creating new instances outside of the class
+    private constructor() {
+        // Initialize with default values
+        this.configFeatures = Promise.resolve({
+            chat: true,
+            autoComplete: true,
+            commands: true,
+        })
+        // Initiate the first fetch and set up a recurring fetch every 30 seconds
+        this.refreshConfigFeatures()
+        // Fetch config features periodically every 30 seconds only if isDotCom is false
+        if (!graphqlClient.isDotCom()) {
+            setInterval(() => this.refreshConfigFeatures(), 30000)
+        }
+    }
+
+    // Static method to get the singleton instance
+    public static getInstance(): ConfigFeaturesSingleton {
+        if (!ConfigFeaturesSingleton.instance) {
+            ConfigFeaturesSingleton.instance = new ConfigFeaturesSingleton()
+        }
+        return ConfigFeaturesSingleton.instance
+    }
+
+    // Refreshes the config features by fetching them from the server and caching the result
+    private refreshConfigFeatures(): void {
+        const previousConfigFeatures = this.configFeatures
+        this.configFeatures = this.fetchConfigFeatures().catch((error: Error) => {
+            // Ignore a fetcherror as older SG instances will always face this because their GQL is outdated
+            if (!error.message.includes('FetchError')) {
+                console.error(error.message)
+            }
+            // In case of an error, return previously fetched value
+            return previousConfigFeatures
+        })
+    }
+
+    public getConfigFeatures(): Promise<{
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    }> {
+        return this.configFeatures
+    }
+
+    // Fetches the config features from the server and handles errors
+    private async fetchConfigFeatures(): Promise<{
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    }> {
+        // Execute the GraphQL query to fetch the configuration features
+        const features = await graphqlClient.getCodyConfigFeatures()
+        if (features instanceof Error) {
+            // If there's an error, throw it to be caught in refreshConfigFeatures
+            throw features
+        } else {
+            // If the fetch is successful, store the fetched configuration features
+            return features
+        }
+    }
+}
 
 async function verifyResponseCode(response: Response): Promise<Response> {
     if (!response.ok) {
--- a/vscode/src/chat/chat-view/ChatManager.ts
+++ b/vscode/src/chat/chat-view/ChatManager.ts
@@ -89,7 +89,15 @@ export class ChatManager implements vscode.Disposable {
         await chatProvider?.setWebviewView(view)
     }
 
-    public async executeCommand(command: CodyCommand, source: ChatEventSource): Promise<ChatSession | undefined> {
+    public async executeCommand(
+        command: CodyCommand,
+        source: ChatEventSource,
+        enabled?: boolean
+    ): Promise<ChatSession | undefined> {
+        if (!enabled) {
+            void vscode.window.showErrorMessage('This feature has been disabled by your Sourcegraph site admin.')
+            return
+        }
         logDebug('ChatManager:executeCommand:called', command.slashCommand)
         if (!vscode.window.visibleTextEditors.length) {
             void vscode.window.showErrorMessage('Please open a file before running a command.')
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -3,6 +3,7 @@ import * as vscode from 'vscode'
 import { isCodyIgnoredFile } from '@sourcegraph/cody-shared/src/chat/context-filter'
 import { FeatureFlag, featureFlagProvider } from '@sourcegraph/cody-shared/src/experimentation/FeatureFlagProvider'
 import { RateLimitError } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
+import { ConfigFeaturesSingleton } from '@sourcegraph/cody-shared/src/sourcegraph-api/graphql/client'
 import { wrapInActiveSpan } from '@sourcegraph/cody-shared/src/tracing'
 
 import { type AuthStatus } from '../chat/protocol'
@@ -186,6 +187,18 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             const completionRequest: CompletionRequest = { document, position, context }
             this.lastCompletionRequest = completionRequest
 
+            const configFeatures = await ConfigFeaturesSingleton.getInstance().getConfigFeatures()
+
+            try {
+                if (!configFeatures.autoComplete) {
+                    // If Configfeatures exists and autocomplete is disabled then raise
+                    // the error banner for autocomplete config turned off
+                    throw new Error('AutocompleteConfigTurnedOff')
+                }
+            } catch (error) {
+                this.onError(error as Error)
+                throw error
+            }
             const start = performance.now()
 
             if (!this.lastCompletionRequestTimestamp) {
@@ -605,6 +618,26 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             return
         }
 
+        if (error.message === 'AutocompleteConfigTurnedOff') {
+            const errorTitle = 'Cody Autocomplete Disabled by Site Admin'
+            // If there's already an existing error, don't add another one.
+            const hasAutocompleteDisabledBanner = this.config.statusBar.hasError('AutoCompleteDisabledByAdmin')
+            if (hasAutocompleteDisabledBanner) {
+                return
+            }
+            let shown = false
+            this.config.statusBar.addError({
+                title: errorTitle,
+                description: 'Contact your Sourcegraph site admin to enable autocomplete',
+                errorType: 'AutoCompleteDisabledByAdmin',
+                onShow: () => {
+                    if (shown) {
+                        return
+                    }
+                    shown = true
+                },
+            })
+        }
         // TODO(philipp-spiess): Bring back this code once we have fewer uncaught errors
         //
         // c.f. https://sourcegraph.slack.com/archives/C05AGQYD528/p1693471486690459
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -5,7 +5,7 @@ import { type ConfigurationWithAccessToken } from '@sourcegraph/cody-shared/src/
 import { FeatureFlag, featureFlagProvider } from '@sourcegraph/cody-shared/src/experimentation/FeatureFlagProvider'
 import { newPromptMixin, PromptMixin } from '@sourcegraph/cody-shared/src/prompt/prompt-mixin'
 import { isDotCom } from '@sourcegraph/cody-shared/src/sourcegraph-api/environments'
-import { graphqlClient } from '@sourcegraph/cody-shared/src/sourcegraph-api/graphql'
+import { ConfigFeaturesSingleton, graphqlClient } from '@sourcegraph/cody-shared/src/sourcegraph-api/graphql/client'
 
 import { CachedRemoteEmbeddingsClient } from './chat/CachedRemoteEmbeddingsClient'
 import { ChatManager, CodyChatPanelViewType } from './chat/chat-view/ChatManager'
@@ -281,8 +281,8 @@ const register = async (
         if (command.mode !== 'ask') {
             return
         }
-
-        return chatManager.executeCommand(command, source)
+        const configFeatures = await ConfigFeaturesSingleton.getInstance().getConfigFeatures()
+        return chatManager.executeCommand(command, source, configFeatures.commands)
     }
 
     const statusBar = createStatusBar()
--- a/vscode/webviews/Chat.tsx
+++ b/vscode/webviews/Chat.tsx
@@ -35,6 +35,7 @@ import styles from './Chat.module.css'
 
 interface ChatboxProps {
     welcomeMessage?: string
+    chatEnabled: boolean
     messageInProgress: ChatMessage | null
     messageBeingEdited: boolean
     setMessageBeingEdited: (input: boolean) => void
@@ -76,6 +77,7 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
     setChatModels,
     chatModels,
     enableNewChatUI,
+    chatEnabled,
     userInfo,
     guardrails,
 }) => {
@@ -224,6 +226,7 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
             onCurrentChatModelChange={onCurrentChatModelChange}
             ChatModelDropdownMenu={ChatModelDropdownMenu}
             userInfo={userInfo}
+            chatEnabled={chatEnabled}
             EnhancedContextSettings={enableNewChatUI ? EnhancedContextSettings : undefined}
             postMessage={msg => vscodeAPI.postMessage(msg)}
             guardrails={guardrails}
@@ -242,6 +245,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
     autoFocus,
     value,
     setValue,
+    chatEnabled,
     required,
     onInput,
     onKeyDown,
@@ -250,6 +254,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
 }) => {
     const inputRef = useRef<HTMLTextAreaElement>(null)
     const placeholder = 'Message (@ to include code, / for commands)'
+    const disabledPlaceHolder = 'Chat has been disabled by your Enterprise instance site administrator'
 
     useEffect(() => {
         if (autoFocus) {
@@ -278,14 +283,21 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
         },
         [inputRef, onKeyDown]
     )
+    const actualPlaceholder = chatEnabled ? placeholder : disabledPlaceHolder
+    const isDisabled = !chatEnabled
 
     return (
         <div
             className={classNames(styles.chatInputContainer, chatModels && styles.newChatInputContainer)}
-            data-value={value || placeholder}
+            data-value={value || actualPlaceholder}
         >
             <textarea
-                className={classNames(styles.chatInput, className, chatModels && styles.newChatInput)}
+                className={classNames(
+                    styles.chatInput,
+                    className,
+                    chatModels && styles.newChatInput,
+                    isDisabled && styles.textareaDisabled
+                )}
                 rows={1}
                 ref={inputRef}
                 value={value}
@@ -293,9 +305,10 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
                 onInput={onInput}
                 onKeyDown={onTextAreaKeyDown}
                 onFocus={onFocus}
-                placeholder={placeholder}
+                placeholder={actualPlaceholder}
                 aria-label="Chat message"
                 title="" // Set to blank to avoid HTML5 error tooltip "Please fill in this field"
+                disabled={isDisabled} // Disable the textarea if the chat is disabled and change the background color to grey
             />
         </div>
     )
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -28,10 +28,10 @@ log:
             value: defaultClient / v1
           - _fromType: array
             name: accept
-            value: '*/*'
+            value: "*/*"
           - _fromType: array
             name: content-length
-            value: '1228'
+            value: "1228"
           - name: host
             value: sourcegraph.com
         headersSize: 0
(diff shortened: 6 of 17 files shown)


vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -85,13 +85,17 @@ export class ChatManager implements vscode.Disposable {
...
-    public async executeCommand(command: CodyPrompt, source: ChatEventSource): Promise<void> {
+    public async executeCommand(command: CodyPrompt, source: ChatEventSource, enabled?: boolean): Promise<void> {
         logDebug('ChatManager:executeCommand:called', command.slashCommand)
         if (!vscode.window.visibleTextEditors.length) {
             void vscode.window.showErrorMessage('Please open a file before running a command.')
             return
         }
 
+        if (!enabled) {
+            void vscode.window.showErrorMessage('This is feature has been disabled on Cody by site admin.')
```
toolmantim, line 96:
```suggestion
            void vscode.window.showErrorMessage('This feature has been disabled by your Enterprise instance site administrator.')
```

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -676,6 +688,26 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+            const errorTitle = 'Cody Autocomplete Disabled by Site Admin'
+            // If there's already an existing error, don't add another one.
+            const hasAutocompleteDisabledBanner = this.config.statusBar.hasError('AutoCompleteDisabledByAdmin')
+            if (hasAutocompleteDisabledBanner) {
+                return
+            }
+            let shown = false
+            this.config.statusBar.addError({
+                title: errorTitle,
+                description: 'Contact Sourcegraph Site Admin to enable autocomplete',
```
toolmantim, line 701:
```suggestion
                description: 'Contact your Enterprise instance site administrator to enable autocomplete',
```

vscode/src/main.ts
```diff
@@ -286,14 +286,30 @@ const register = async (
...
         source: ChatEventSource = 'editor' // where the command was triggered from
     ): Promise<void> => {
+        const configFeatures = await ConfigFeaturesSingleton.getInstance().getConfigFeatures()
+        let commandsEnabled = true
+        if (configFeatures) {
+            // If configFeatures exist then we set the commandsEnabled based on its value
+            commandsEnabled = configFeatures.commands
+        }
+        if (!commandsEnabled) {
+            void vscode.window.showErrorMessage('This is feature has been disabled on Cody by site admin.')
```
toolmantim, line 309:
```suggestion
            void vscode.window.showErrorMessage('This feature has been disabled by your Enterprise instance site administrator.')
```

vscode/webviews/Chat.tsx
```diff
@@ -269,6 +273,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
 }) => {
     const inputRef = useRef<HTMLTextAreaElement>(null)
     const placeholder = 'Message (@ to include code, / for commands)'
+    const disabledPlaceHolder = 'SiteAdmin has disabled chat and commands for Cody'
```
toolmantim, line 276:
```suggestion
    const disabledPlaceHolder = 'Chat and commands have been disabled by your Enterprise instance site administrator'
```

toolmantim (commented):
Thanks @arafatkatze! Looking good. Left some suggested comments on the UI copy.

arafatkatze:
@toolmantim Thanks for the recommendations, All comments resolved.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,74 @@ export class SourcegraphGraphQLAPIClient {
...
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    } | null = null
+    private isFetching = false
+
+    private constructor() {
+        // Private constructor to prevent direct instantiation
+        // Start the interval to fetch config features every 30 seconds
+        setInterval(() => {
```
slimsag, line 837: Question: `setInterval` doesn't run immediately (so this logic will wait 30s, then `fetchConfigFeatures`. If say someone signs in and uses Cody really quickly, before that 30s and before `fetchConfigFeatures` has run for the first time.. what is the default? Will they have access, then it gets taken away after 30s?

slimsag: Perhaps you should also immediately call `this.fetchConfigFeatures().catch(console.error)` outside the `setInterval` so as to not delay the first request

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,74 @@ export class SourcegraphGraphQLAPIClient {
...
+            ConfigFeaturesSingleton.instance = new ConfigFeaturesSingleton()
+        }
+        return ConfigFeaturesSingleton.instance
+    }
+
+    public async getConfigFeatures(): Promise<{
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    } | null> {
```
slimsag, line 854: Instead of this ever returning `null`, perhaps we could have the default values defined here. i.e., if it would be null, then fill out the defaults here and return them.

That way we can clearly find that list of default access controls - without chasing down every usage of `getConfigFeatures()`

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,74 @@ export class SourcegraphGraphQLAPIClient {
...
+        return this.configFeatures
+    }
+
+    private async fetchConfigFeatures(): Promise<void> {
+        // Perform your GQL query here and return the result
+        const features = await graphqlClient.getCodyConfigFeatures()
+        if (features instanceof Error) {
+            // Handle the error case by setting null so as not to disrupt the flow
+            console.error(features.message)
+            this.configFeatures = null // Set the value of Config Features as null incase it doesn't exist or incase of errors
```
slimsag, line 870: I am imagining this scenario:

* Temporarily for some reason the config features GraphQL request fails, e.g. server error, your wifi has issues temporarily, etc.
* This would then get set to `null`, meaning you'd get the default access controls..
* That could mean someone might report an issue like:

> Normally Cody says "my admin disabled chat", but sometimes I see that chat has been enabled seemingly at random.. and then like 30s later it gets disabled again. What's going on?

So perhaps in the case of an error, if we already fetched configFeatures successfully at least one time, we should use the old configFeatures instead of `null`/defaults.

vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -85,13 +85,19 @@ export class ChatManager implements vscode.Disposable {
...
+    public async executeCommand(command: CodyPrompt, source: ChatEventSource, enabled?: boolean): Promise<void> {
         logDebug('ChatManager:executeCommand:called', command.slashCommand)
         if (!vscode.window.visibleTextEditors.length) {
             void vscode.window.showErrorMessage('Please open a file before running a command.')
             return
         }
 
+        if (!enabled) {
+            void vscode.window.showErrorMessage(
+                'This feature has been disabled by your Enterprise instance site administrator.'
```
slimsag, line 97:
```suggestion
                'This feature has been disabled by your Sourcegraph site administrator.'
```

Suggesting wording improvement, feel free to ignore if someone else said to use the current wording.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -676,6 +688,26 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
+            const errorTitle = 'Cody Autocomplete Disabled by Site Admin'
+            // If there's already an existing error, don't add another one.
+            const hasAutocompleteDisabledBanner = this.config.statusBar.hasError('AutoCompleteDisabledByAdmin')
+            if (hasAutocompleteDisabledBanner) {
+                return
+            }
+            let shown = false
+            this.config.statusBar.addError({
+                title: errorTitle,
+                description: 'Contact your Enterprise instance site administrator to enable autocomplete',
```
slimsag, line 701:
```suggestion
                description: 'Contact your Sourcegraph site administrator to enable autocomplete',
```

vscode/src/main.ts
```diff
@@ -286,14 +286,32 @@ const register = async (
...
     ): Promise<void> => {
+        const configFeatures = await ConfigFeaturesSingleton.getInstance().getConfigFeatures()
+        let commandsEnabled = true
+        if (configFeatures) {
+            // If configFeatures exist then we set the commandsEnabled based on its value
+            commandsEnabled = configFeatures.commands
+        }
+        if (!commandsEnabled) {
+            void vscode.window.showErrorMessage(
+                'This feature has been disabled by your Enterprise instance site administrator.'
```
slimsag, line 310:
```suggestion
                'This feature has been disabled by your Sourcegraph site administrator.'
```

vscode/webviews/Chat.tsx
```diff
@@ -269,6 +273,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
 }) => {
     const inputRef = useRef<HTMLTextAreaElement>(null)
     const placeholder = 'Message (@ to include code, / for commands)'
+    const disabledPlaceHolder = 'Chat and commands have been disabled by your Enterprise instance site administrator'
```
slimsag, line 276:
```suggestion
    const disabledPlaceHolder = 'Chat and commands have been disabled by your Sourcegraph instance site admin'
```

I think this text should change depending on what is disable, right?

* Chat AND commands are disabled -> `Chat and commands have been disabled by your Sourcegraph site admin`
* Only Chat is disabled -> `Chat has been disabled by your Sourcegraph site admin`

slimsag: Also, an edge case I had not thought about: if chat is enabled, but commands are disabled.. how does the user know commands are disabled? where do we tell them?

slimsag (commented):
Left a bunch of comments, generally looking good to me otherwise.

One question: If you run this against a Sourcegraph instance that does NOT have the new GraphQL API, what happens? Want to make sure that this doesn't e.g. immediately lock out all Cody users whose Sourcegraph instance does not have the new GraphQL API :)

vscode/src/main.ts
```diff
@@ -286,14 +286,32 @@ const register = async (
         if (command.mode !== 'ask') {
             return
         }
-
-        return chatManager.executeCommand(command, source)
+        const configFeatures = await ConfigFeaturesSingleton.getInstance().getConfigFeatures()
```
abeatrix, line 289: it might be better to move this to [handleChatRequest](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts?L587-643) in SimpleChatPanelProvider to make sure the checks are run for all outgoing chat and command requests.

vscode/src/main.ts
```diff
@@ -286,14 +286,32 @@ const register = async (
...
+        let commandsEnabled = true
+        if (configFeatures) {
+            // If configFeatures exist then we set the commandsEnabled based on its value
+            commandsEnabled = configFeatures.commands
```
abeatrix, line 293: this would block chat questions submitted from the command palette: 

image

abeatrix (commented):
I havent set up an instance to test the config (will do it tomorrow), but i've left some comments / questions inline based on the code change

vscode/src/chat/chat-view/ChatManager.ts
toolmantim: (We also use "Enterprise Instance" in the UI for login etc)

vscode/src/main.ts
slimsag: I'd say not the end of the world if this is the case - since it is technically a command and we do want admins to enable commands generally, so the more reasons users pester their site admin to enable them the better :)

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: image
Everytime we use this singleton instance we use public `getConfigFeatures`.  The first time we call it, there is a request made to GQL only once.

All the GQL calling is behind a private method.

So this solves the underlying edge case suggested.

vscode/src/chat/chat-view/ChatManager.ts
arafatkatze: Good thinking folks!!

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Makes perfect Sense

vscode/webviews/Chat.tsx
arafatkatze: Currently we say this 
https://sourcegraph.com/github.com/sourcegraph/cody@ara/site-config/-/blob/vscode/src/chat/chat-view/ChatManager.ts?L97%3A18-97%3A34 

But instead of saying this feature I can also say commands disabled. 
image
The UI looks like this. 

Chat disabled commentary and command disabled commentary are separate things so they should be expressed separately in the UX. That was my thinking

arafatkatze: I fixed the wording to just say "Chat has been disabled by your Sourcegraph instance site admin "

vscode/src/main.ts
arafatkatze: _this would block chat questions submitted from the command palette:_

@abeatrix That was done on purpose. I wanted to disable commands from being sent if they were disabled by the admin so I wanted to show errors in case of edit code etc. But if you feel the UI should behave differently then lemme know. My reasoning was same as Stephen's reasoning.

vscode/src/chat/chat-view/ChatManager.ts
toolmantim: Thanks for the brain dump @slimsag!

@arafatkatze if we want to leave off Enterprise, "your Sourcegraph instance site admin" sounds like the best bet… let's use that.

vscode/src/main.ts
arafatkatze: Great suggestion but there were reasons not to do it

1. I wanted to decouple Chat/commands. Coz its possible to have Chat Enabled/Commands disabled and vice versa
2. Incase of Chat I wanted to show this UI change where the text box is greyed out
image

The greyed out UI is good but it requires a change in the react components and thats why I moved the GQL query there instead of simplechatpanelprovider

3. I wanted to communicate the commands not being enabled with this UI 
image

Which looks different from the UI for chat disabled so thats why this had to be here. It would be harder to. do the UI communication had the blockers been elsewhere.



Feel free to offer suggestions on how to do this differently.

arafatkatze:
@abeatrix @slimsag 
Thanks for the review.
Offered commentary on most changes but feel free to push back on choices and offer more suggestions.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
  */
 export const graphqlClient = new SourcegraphGraphQLAPIClient()
 
+/**
+ * ConfigFeaturesSingleton is a singleton class that fetches and caches
+ * configuration features from a remote source. It provides methods to
+ * retrieve the cached config features.
```
dominiccooney, line 813: Let's revise this to be shorter and clearer. Fetches/retrieve is awkward.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
...
+ *
+ * The constructor is private to enforce the singleton pattern.
```
dominiccooney, line 815: I don't think details like this are worth it.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
...
+ *
+ * getInstance() provides access to the singleton instance.
```
dominiccooney, line 817: Let's put method documentation on the methods themselves.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
...
+ *
+ * getConfigFeatures() fetches the config if not already cached, and returns
+ * the cached value. It ensures the config is only fetched once from the caller's perspective.
```
dominiccooney, line 820: Why does perspective matter about the number of fetches? Be direct: The config is only fetched once.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
...
+ *
+ * fetchConfigFeatures() makes the actual remote request to get the config,
+ * handles errors, and caches the result. It runs every 30 seconds
+ */
+export class ConfigFeaturesSingleton {
```
dominiccooney, line 825: Naming this is awkward because the thing it does—fetches and caches the config features—should be transparent to the user. The fact that it is a singleton is the least interesting thing about it, to me. Don't think it is worth naming it around that.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
...
+ *
+ * The constructor is private to enforce the singleton pattern.
+ *
+ * getInstance() provides access to the singleton instance.
+ *
+ * getConfigFeatures() fetches the config if not already cached, and returns
+ * the cached value. It ensures the config is only fetched once from the caller's perspective.
+ *
+ * fetchConfigFeatures() makes the actual remote request to get the config,
+ * handles errors, and caches the result. It runs every 30 seconds
```
dominiccooney, line 823: Why would we poll every 30 seconds? This sounds like the exact opposite of the config only being fetched once.

If the config changes, how are clients meant to detect and respond to that change?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -778,6 +807,75 @@ export class SourcegraphGraphQLAPIClient {
...
+        return ConfigFeaturesSingleton.instance
+    }
+
+    public async getConfigFeatures(): Promise<{
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    } | null> {
+        // Make sure to fetch only once when get method is used from external sources
+        if (!this.configFeatures && !this.isFetching) {
```
dominiccooney, line 856: There's a simpler way to do this "join async operation in progress", and that is to store the Promise of the fetch operation in the class and simply await or return that.

When you think about it, if all you need is the result, then you can simply always do it that way. You don't need to special case "is fetching" at all.

lib/shared/src/sourcegraph-api/graphql/client.ts
dominiccooney: Reopening this thread...

If we fetch the features lazily, and there's no way to subscribe for updates, what's the benefit of this polling? New calls will sometimes get an updated value?

dominiccooney (changes requested):
Can you use the VSCode context APIs and "when" enablement in the commands so users don't see commands they can't use in command palettes, editor tab strip, tree views, etc.?

lib/shared/src/sourcegraph-api/graphql/client.ts
slimsag: I want a few behaviors from this PR:

1. On startup, quickly ask the server for which features can be used. Don't make Cody startup times slower.
2. If a site admin changes this setting on the server while people are using Cody, we should respect their change in a reasonably quick timeframe (e.g. a minute or so) without requiring users to restart Cody/vscode.
3. Do not increase the latency of autocomplete/chat requests. For example, let's not send a 'Can I use Cody chat?' request directly before every autocomplete/chat API request.
4. If for some reason we can't fetch the config on Cody startup, just let the user make use of the features.
5. If someone is using Cody but e.g. Chat is disabled,and then there's some intermittent failure fetching the updated config, don't give them access to chat only to be blocked from using it shortly after. i.e. use whatever last config we did successfully query.

arafatkatze:
> Can you use the VSCode context APIs and "when" enablement in the commands so users don't see commands they can't use in command palettes, editor tab strip, tree views, etc.?

@slimsag @toolmantim  Selective UI to only show the enabled commands look much better UX than what I have now. 
If you are cool too I am down to implement it?

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Agree!!

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: I agree with your point but I called it singleton coz the idea is that you could potentially be calling the SAME thing from many different places to get the GQL config of what is and isn't visible(the pull a subset of the key/value pair). Those things could potentially HAMMER the GQL if they all get a chance to make a GQL request.

So how about we call it `CachedConfigFeaturesSingleton` would that be better? CC @slimsag

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Apologies for the lack of clarification, I am fixing the comments basically the idea is that the GQL config must be fetched every 30 seconds so that it stays in sync. 

We don't want to HAMMER GQL with requests so we use a singleton and because sometimes those requests can potentially be time consuming etc we only do 1 every 30 seconds to keep config in sync.

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Agree, Fixing all the commentary.

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Great Catch!!

Fixed!!

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Great point. I fixed it

slimsag:
> so users don't see commands they can't use in command palettes, editor tab strip, tree views, etc.?

Rather than **remove them** so users can't see them, what about still having them be there but just communicating on each clearly they are disabled by their admin?

I imagine two possibilities:

1. User doesn't see commands -> doesn't know Cody has those features -> assumes Cody is not as good as it actually is -> user values Cody less.
1. User sees commands -> sees that their admin disabled them -> voices opinion internally at company that their company should enable cody commands -> admin enables them -> users and us win

I'm hoping for option 2 above. Thoughts?

toolmantim:
Option 2 is my preference too

arafatkatze:
@slimsag Option 2 makes a lot of sense now that you explain. 

So hopefully this PR is good to go. Let me know if you have any concerns and I will fix the whole thing(merge conflicts etc) and we can merge this today.

lib/shared/src/sourcegraph-api/graphql/client.ts
dominiccooney: How about `CachedConfigFeatures`?

dominiccooney (approved):
Improvements look good, approving to unblock. However...

Other times we started with config objects like this that could change, we ended up having to add plumbing to respond to changes. And because we didn't design for it, sometimes didn't go well... Change events refreshed unrelated state; change events caused cycles; etc.

Could you explain why callers don't need to subscribe to changes? If component X sees one config and component Y another then the product could be configured in a weird/broken state that's hard to reproduce.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -186,6 +188,16 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
             const completionRequest: CompletionRequest = { document, position, context }
             this.lastCompletionRequest = completionRequest
 
+            try {
+                if (!configFeatures.autoComplete) {
```
abeatrix, line 192: When I'm connected to the [demo instance](https://demo.sourcegraph.com/), the autocompletion stopped working for me, and i can see the error fires every n seconds.

cc @philipp-spiess @valerybugakov to confirm if this is the best place to disable autoComplete

image

- Commands also doesn't work:

image

- Chat works fine though!

image

vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -85,13 +85,17 @@ export class ChatManager implements vscode.Disposable {
...
-    public async executeCommand(command: CodyPrompt, source: ChatEventSource): Promise<void> {
+    public async executeCommand(command: CodyPrompt, source: ChatEventSource, enabled?: boolean): Promise<void> {
         logDebug('ChatManager:executeCommand:called', command.slashCommand)
         if (!vscode.window.visibleTextEditors.length) {
             void vscode.window.showErrorMessage('Please open a file before running a command.')
             return
         }
 
+        if (!enabled) {
+            void vscode.window.showErrorMessage('This feature has been disabled by your Sourcegraph site admin.')
```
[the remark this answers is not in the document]
arafatkatze: Totally

vscode/src/completions/inline-completion-item-provider.ts
arafatkatze: image
Thats more of Demo instance issue coz it cant recognize the GQL of the new things I added. 

image
My test instance can recognize that GQL and same for Dotcom(Ideally I would test this with the test instance coz it does have that GQL to be able to make it work). 

I am rather surprized coz the demo instance says its on version 5.2.6 so it should have my changes 
image

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -785,6 +814,75 @@ export class SourcegraphGraphQLAPIClient {
...
+        // Initialize with default values
+        this.configFeatures = Promise.resolve({
+            chat: true,
+            autoComplete: true,
+            commands: true,
+        })
+        // Initiate the first fetch and set up a recurring fetch every 30 seconds
+        this.refreshConfigFeatures()
+        // Fetch config features periodically every 30 seconds
+        setInterval(() => this.refreshConfigFeatures(), 30000)
```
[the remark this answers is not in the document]
arafatkatze: Great catch I fixed it

arafatkatze:
@abeatrix Can I get 💚 if there aren't major blockers on this? Its rather hard to fix this PR as it touches many files that get changed with refactors in other places. If majority of the PR works good I am happy to merge this one then do follow up items on top of it if that works.

Thanks for the careful catches on the things I missed.

abeatrix:
> @abeatrix Can I get 💚 if there aren't major blockers on this? Its rather hard to fix this PR as it touches many files that get changed with refactors in other places. If majority of the PR works good I am happy to merge this one then do follow up items on top of it if that works.
> 
> Thanks for the careful catches on the things I missed.

no problem! The only major blockers I am seeing when connected to an older Sourcegraph instance (e.g. demo) are:
1. Autocompletes not working
2.  this error message when a new chat is created:

image


@arafatkatze @slimsag I don't want to block anything for your team though, so if you think these are expected then feel free to ignore me 😄

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -72,9 +73,30 @@ interface CurrentUserInfoResponse {
...
+    chat: boolean
+    autoComplete: boolean
+    commands: boolean
+}
+
+interface CodyConfigFeaturesResponse {
+    site: { codyConfigFeatures: CodyConfigFeatures | null } | null
+}
+
+interface EnterpriseCurrentUserInfoResponse {
```
abeatrix, line 90: looks like this can be removed?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -676,6 +717,77 @@ export class SourcegraphGraphQLAPIClient {
...
+    public getConfigFeatures(): Promise<{
+        chat: boolean
+        autoComplete: boolean
+        commands: boolean
+    }> {
+        return this.configFeatures
+    }
+
+    // Fetches the config features from the server and handles errors
+    private async fetchConfigFeatures(): Promise<{
```
abeatrix, line 774: should we check their instance version (`getSiteVersion()`) to make sure they support the graphQL query before ` graphqlClient.getCodyConfigFeatures()` ?

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Yes it can be I am fixing tests now

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Thats a good one, lemme try this.

arafatkatze:
@abeatrix That chaining cycle thing is a LEGIT issue that for some reason I didn't see in my debugger tests but it exists in testing. Apologies for this but lemme fix this stuff properly.

abeatrix (approved):
Tested on DotCom, S2 and Demo and confirmed autocomplete, chat, and commands are all working at the latest commit, great work :D !

PR #2560, opened 2024-01-04
Autocomplete: adjust leading whitespace in the last candidate text on indentation

### LINKED ISSUE #2532: bug: code completions sometimes have incorrect indentation on the first line
### Version

 v1.0.4

### Describe the bug

image (8)

### Expected behavior

Code completions are always indented correctly with no extra whitespace after the cursor.

### Additional context

[Slack thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1704164215488589)

valerybugakov:
## Context

- It was a tricky one to pinpoint because of the time-sensitive nature of the bug. It happens when a user initializes a completion generation by placing a cursor on the new line and then almost instantly tabs 2+ times. For example: 

    ```ts
      function test() {
    █
      }
    ```
    
    The initial completion generation request comes back in the background with the insert-text: `\t\tconst x = 1`, but a user already added three tabs manually:
    
    ```ts
      function test() {
          █
      }
    ```
    
    We reuse (`reuseLastCandidate`) the initial completion that was generated in the background because only whitespace on the current line was changed, and the previous logic returned the final insert text as the following: `lastTriggerCurrentLinePrefixInDocument.slice(currentLinePrefix.length) + item.insertText` which didn't account for leading whitespace that user added manually. Notice extra whitespace in the ghost text:

    ```ts
      function test() {
          █    const x = 1
      }
    ```

    This PR fixes the issue by removing the manually added whitespace from the insert-text if it's present there. Note that the `isTypingAsSuggested` clause missed this issue because a user adds more whitespace than there's present in the current line: `lastCompletion.startsWith(currentLinePrefix) === false`.

- Closes #2532

## Test plan

- Manually tested that this is not happening anymore. This is tricky because the bug is time-sensitive.
- Added a unit test for this specific issue.

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/last-candidate.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/last-candidate.test.ts
@@ -90,16 +90,6 @@ describe('[getInlineCompletions] reuseLastCandidate', () => {
             source: InlineCompletionsResultSource.LastCandidate,
         }))
 
-    it('is reused when adding leading whitespace', async () =>
-        // The user types ``, sees ghost text `x = 1`, then types ` ` (space). The original
-        // completion should be reused.
-        expect(await getInlineCompletions(params(' █', [], { lastCandidate: lastCandidate('█', 'x = 1') }))).toEqual<V>(
-            {
-                items: [{ insertText: 'x = 1' }],
-                source: InlineCompletionsResultSource.LastCandidate,
-            }
-        ))
-
     it('is reused when the deleting back to the start of the original trigger (but no further)', async () =>
         // The user types `const x`, accepts a completion to `const x = 123`, then deletes back
         // to `const x` (i.e., to the start of the original trigger). The original completion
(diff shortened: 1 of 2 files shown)


vscode/src/completions/get-inline-completions-tests/last-candidate.test.ts
```diff
@@ -243,14 +241,52 @@ describe('[getInlineCompletions] reuseLastCandidate', () => {
...
+             * as suggested. The resuling completion `const foo = 1`.
+             */
+            expect(
+                await getInlineCompletionsInsertText(
+                    params('\t\t\t\t█', [], {
+                        lastCandidate: lastCandidate('\t█', '\t\tconst x = 1'),
+                    })
+                )
+            ).toEqual(['const x = 1'])
+        })
```
valerybugakov, line 269: New unit test for the bug fixed in this PR.

abeatrix (approved):
Texted with AnthropicProvider and fireworks and they both seem to work fine for me. Nice catch 👍!

philipp-spiess:
Great fix and test!

Issue #2539, opened 2024-01-02
bug: AI Code Generation Stalling and Unresponsive Stop Button


jim-steven:
### Version

v1.1.1704208019

### Describe the bug

![image](https://github.com/sourcegraph/cody/assets/141747724/8ad92264-58b9-4275-9396-d06dbc786ad5)

A Discord user encountered a problem with ChatGPT 4, where it couldn't decode an event payload due to an "unexpected end of JSON input." The error message indicated the system received less JSON data than expected, leading to an unprocessable JSON structure. After switching to pre-release, this issue was resolved. However, ChatGPT 4 then began stalling during code generation, and the 'Stop Generating' button stopped working, requiring a new chat to fix the problem.

[Discord link](https://discord.com/channels/969688426372825169/1191760945177968721)

### Expected behavior

The AI should generate code smoothly without stalling. The 'Stop Generating' button should function correctly, allowing users to halt the code generation process at any time without needing to restart the chat.

### Additional context
[Error Logs](https://github.com/sourcegraph/cody/files/13813228/message.txt)

jim-steven:
Two more users ran into this [error](https://discord.com/channels/969688426372825169/1191834179982344284), one of them on pre-release.

jim-steven:
related to #2464 and #2526. should have been fixed with the recent pre-release.

jim-steven:
additional user that has the [same issue from GH](https://discord.com/channels/969688426372825169/1192077669639127163).
![image](https://github.com/sourcegraph/cody/assets/141747724/573c2ef8-5c91-420a-b268-b67b6735e5f2)

jim-steven:
![image](https://github.com/sourcegraph/cody/assets/141747724/ebca0fb4-c937-44f3-9ba4-5d394eac1170)
![image](https://github.com/sourcegraph/cody/assets/141747724/ac79590c-fc03-41c2-9c46-3336c7e023c0)

abeatrix:
dup of https://github.com/sourcegraph/cody/issues/2513

The main issue described in this issue is fixed in the latest release (v1.0.5).

PR #2566, opened 2024-01-04
fix: show error after Cody response in chat UI

abeatrix:
image

This PR addresses the issue mentioned in https://github.com/sourcegraph/cody/issues/2539, where long response from chat gpt 4.0 will resulted in error, and the error message will replace the response text from Cody.

- Display error message in additional to the text that cody has already responded 
- Add new styles for .request-error and .request-error-title classes.
- Update ErrorItem.tsx to use new CSS classes for request error display.

## Test plan

1. Switch chat to use chat gpt 4.0
2. Open a large file and select the whole file
3. Run the /test command
4. If the response is longer than the answer limit, it should display an error message after the text that Cody has already responsed
- before: the text will be replaced with the error message

### Before

https://github.com/sourcegraph/cody/assets/68532117/40609eef-8a27-4202-a195-7b75b7ddc260

https://www.loom.com/share/2136a5de267d4c21849b735bb38619d9?sid=b8d8652e-2bba-4498-bcf6-aec9c2c9a0cc

### After

Error message will not replace the text that's already returned by the LLM

image

https://www.loom.com/share/3e183dfec7f940079100b48e519235eb?sid=2e55abad-e4d4-4ba0-8a6a-05078f48ef0a

### DIFF
--- a/lib/ui/src/chat/ErrorItem.module.css
+++ b/lib/ui/src/chat/ErrorItem.module.css
@@ -18,7 +18,8 @@
     margin: 0;
 }
 
-.error-item h1, .error-item p {
+.error-item h1,
+.error-item p {
     margin: 0;
 }
 
@@ -57,7 +58,6 @@
     right: 0;
     overflow: hidden;
     height: 100px;
-
 }
 
 @property --error-item-reflection-position {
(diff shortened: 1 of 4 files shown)


chwarwick (approved):
This makes sense to me

Issue #2570, opened 2024-01-05
bug: autocomplete no-response telemetry event is not sent for some completions


valerybugakov:
### Version

latest

### Describe the bug

1. Generate a completion that, for some reason, won't be shown (`isCompletionVisible` returns false)
2. Observe that `NoResponseEvent` is not being sent.

### Expected behavior

If we get completion results, but they are not visible, `NoResponseEvent` should be sent.

PR #2572, opened 2024-01-05
Autocomplete: split up completion item provider

valerybugakov:
## Context

- Closes https://github.com/sourcegraph/cody/issues/2570 by moving the telemetry helper call to the early return, where we check for the number of visible completions.
- Simplifies the completions processing logic in `provideInlineCompletionItems()`. We return early if no visible completions are generated, so we can simplify the following logic knowing the completions are available and visible.
- Moves the conversion 
- Extracts the `AutocompleteItem` class and `suggestedCompletionItemIDs` cache into a separate file. Update the interfaces of these entities for better readability.
- Extracts the `isCompletionVisible` function and related helpers into a separate file.
- Extracts `FirstCompletionDecorationHandler` into a separate file.

## Test plan

CI

### DIFF
--- a/vscode/src/completions/first-completion-decoration-handler.ts
+++ b/vscode/src/completions/first-completion-decoration-handler.ts
@@ -1,0 +1,78 @@
+import * as vscode from 'vscode'
+
+import { RequestParams } from './request-manager'
+
+/**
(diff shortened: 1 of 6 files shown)


philipp-spiess (approved):
LGTM

PR #2573, opened 2024-01-05
Autocomplete: do not adjust insert text in a VS Code-specific way for the agent

valerybugakov:
## Context

UPD: I added back the VS Code specific `insertText` processing, so this PR no longer has any breaking changes. But this logic is not extracted into a separate function and is called separately. If we decide to change this behavior, it will be easy to do in the future.

- This PR proposes to change the behavior of the `autocomplete/execute` requests for the agent based on the [feedback in Slack](https://sourcegraph.slack.com/archives/C05AGQYD528/p1704164215488589). 
- The only functional change: `InlineCompletionItemProvider.provideInlineCompletionItems()` does not execute the VS Code-specific logic anymore, which makes all completions start at the beginning of the line no matter what the current cursor position is. This logic was added to avoid UI jitter in VS Code, but it does not apply to other Cody clients and requires them to implement additional post-processing logic with no apparent reason.
    - ~~This is a breaking change for clients using the agent API `autocomplete/execute`**.  cc @olafurpg~~
- Based on the clean-up work in another branch https://github.com/sourcegraph/cody/pull/2572

## Test plan

- CI
- Manually verify that completions still work as expected in VS Code.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -402,7 +402,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                 // return `CompletionEvent` telemetry data to the agent command `autocomplete/execute`.
                 const autocompleteResult: AutocompleteResult = {
                     logId: result.logId,
-                    items: autocompleteItems,
+                    items: updateInsertRangeForVSCode(autocompleteItems),
                     completionEvent: CompletionLogger.getCompletionEvent(result.logId),
                 }
 
@@ -413,10 +413,7 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
                     this.unstable_handleDidShowCompletionItem(autocompleteItems[0])
                 }
 
-                return {
-                    ...autocompleteResult,
-                    items: updateInsertRangeForVSCode(autocompleteItems),
-                }
+                return autocompleteResult
             } catch (error) {
                 this.onError(error as Error)
                 throw error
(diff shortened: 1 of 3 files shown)


agent/src/index.test.ts
```diff
@@ -205,7 +205,7 @@ describe('Agent', () => {
         expect(completions.items.length).toBeGreaterThan(0)
         expect(texts).toMatchInlineSnapshot(`
           [
-            "   return a + b;",
+            "return a + b;",
```
valerybugakov, line 208: Example diff for the `completion.insertText` field returned to the agent users.

olafurpg (dismissed):
If this is a breaking change that requires updating all editor clients then I don't think it's a good idea to merge it because it's going to require a lot of work. I'm OK with merging this as long as it's fully backwards compatible.

valerybugakov:
I added back the VS Code specific `insertText` processing, so this PR no longer has any breaking changes. But this logic is not extracted into a separate function and is called separately. If we decide to change this behavior, it will be easy to do in the future.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -392,27 +384,39 @@ export class InlineCompletionItemProvider implements vscode.InlineCompletionItem
...
                 }
 
-                // return `CompletionEvent` telemetry data to the agent command `autocomplete/execute`.
-                const autocompleteResult: AutocompleteResult = {
-                    logId: result.logId,
-                    items: visibleItems,
-                    completionEvent: CompletionLogger.getCompletionEvent(result.logId),
+                return {
+                    ...autocompleteResult,
+                    items: updateInsertRangeForVSCode(autocompleteItems),
```
philipp-spiess, line 418:
```suggestion
```

Should we move this line up to when we create the `autocompleteResult`? It seems odd that the autocomplete result generation is initialized with `items: autocompleteItems` and then overwritten here.

philipp-spiess (approved):
LGTM, there are no more behavioural changes so this should not block any other client. I think the PR description needs updating to reflect that, though

PR #2574, opened 2024-01-05
Edit: Support user provided files ("@name") and symbols ("@#name")

umpox:
## Description

This PR adds the edit functionality for:
- User provided file context, e.g. `@fileName`
- User provided symbol context. e.g. `@#symbolName`

Attempted to get this to closely match the chat fuzzy matching:
- Helpful messages for certain suffixes/no matches
- Icons for different symbols
- Same syntax - `@fileName` and `@#symbolName`

Demo:

https://github.com/sourcegraph/cody/assets/9516420/f37d11f1-2375-41d1-b074-02676390b428

## Test plan

1. Create an edit using `@fileName`
2. Create an edit using `@symbolName`
3. Create an edit using multiple `@fileName` and `@symbolName`
4. Create an edit after removing `@fileName` context, check it doesn't still use it
5. Retry an edit that used `@fileName` or `@symbolName` context, check that the context is still used

### DIFF
--- a/vscode/src/non-stop/FixupTypingUI.ts
+++ b/vscode/src/non-stop/FixupTypingUI.ts
@@ -26,8 +26,10 @@ const MATCHING_CONTEXT_FILE_REGEX = /@(\S+)$/
 const MATCHING_SYMBOL_REGEX = /@#(\S+)$/
 
 interface FixupMatchingContext {
-    /* Associated text label with the context */
-    label: string
+    /* Unique identifier for the context, shown in the input value but not necessarily in the quick pick selector */
+    key: string
+    /* If present, will override the key shown in the quick pick selector */
+    shortLabel?: string
     file: ContextFile
 }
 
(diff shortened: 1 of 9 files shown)


vscode/src/non-stop/FixupTypingUI.ts
```diff
@@ -1,34 +1,120 @@
...
+                file: result,
+            }))
+        }
+
+        return null
+    }
+
+    public async getInputFromQuickPick({
         title = `${EDIT_COMMAND.description} (${EDIT_COMMAND.slashCommand})`,
         placeholder = 'Your instructions',
```
abeatrix, line 93:
```suggestion
        placeholder = 'Your instructions (@ to include additional context from file)',
```

Should we update the instruction to mention the new @ feature?

abeatrix (approved):
Confirmed that @ a file would add context from the file to the request. It doesn't follow the instruction to use the context from the @ file as references but probably something we can fix (in follow up PRs) by updating the prompt for the @ files?
 
https://github.com/sourcegraph/cody/assets/68532117/20bb7175-bd9f-4700-8822-febd20244d95

vscode/src/non-stop/FixupTypingUI.ts
umpox: Yes! Will update, ty

PR #2577, opened 2024-01-05
Clean up Polly recordings to reduce diffs when re-recording

DanTup:
@olafurpg I still see quite a lot of changes here, I think some of it is because of ordering. I noticed we're setting [disableSortingHarEntries](https://netflix.github.io/pollyjs/#/configuration?id=disablesortingharentries) although after reading the docs it's not clear to me why _not_ sorting would improve diffs - seems like the opposite should be true? 🤔

There are a still a lot of "junk" headers in the HAR file that don't necessarily change on each run, but might be worth removing just to make the file smaller (for example `x-frame-options`). Let me know if you think they should be stripped.

(I haven't looked at the format yet, will do that separately, easy stuff first 🙂)

## Test plan

- Re-record Polly recordings before this change and note large diff that contains many timestamps and cookies that change on each run
- Apply this change
- Re-record and commit
- Re-record and review new diff to ensure it's not full of the same timestamps/etc.

### DIFF
--- a/agent/package.json
+++ b/agent/package.json
@@ -25,6 +25,7 @@
   "dependencies": {
     "@pollyjs/adapter-node-http": "^6.0.6",
     "@pollyjs/core": "^6.0.6",
+    "@pollyjs/persister": "^6.0.6",
     "@pollyjs/persister-fs": "^6.0.6",
     "@sourcegraph/cody-shared": "workspace:*",
     "@sourcegraph/telemetry": "^0.16.0",
--- a/agent/recordings/FullConfig_234680970/recording.har
+++ b/agent/recordings/FullConfig_234680970/recording.har
@@ -55,7 +55,7 @@
               "value": "sourcegraph.com"
             }
           ],
-          "headersSize": 324,
+          "headersSize": 341,
           "httpVersion": "HTTP/1.1",
           "method": "POST",
           "postData": {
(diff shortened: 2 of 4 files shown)


DanTup:
@olafurpg I noticed we have some code for passing `--recording-expiry-strategy`.. But I suspect my hard-coded dates here could break that functionality.. is it used?

olafurpg:
@DanTup This is fantastic!

> I noticed we have some code for passing --recording-expiry-strategy.. But I suspect my hard-coded dates here could break that functionality.. is it used?

It's not used. It's fine to just ignore the configurable expiry strategy, or remove the flag.

> There are a still a lot of "junk" headers in the HAR file that don't necessarily change on each run, but might be worth removing just to make the file smaller (for example x-frame-options). Let me know if you think they should be stripped.

I'd love to remove all junk like random IDs, body/header size, etc. We are primarily interested in the request body and response.

olafurpg (approved):
This is fanastic!

abeatrix (approved):
🚀

DanTup:
> It's not used. It's fine to just ignore the configurable expiry strategy, or remove the flag.
> I'd love to remove all junk like random IDs, body/header size, etc. We are primarily interested in the request body and response.

Thanks, I'll tackle some more of this in a separate PR :-)

PR #2575, opened 2024-01-05
Revert "refactor: move commands away from recipe (#2542)"

olafurpg:
The agent tests started failing after merging the PR #2542 when you re-record the HTTP requests. The tests were passing in replay mode for some reason, and we suspect it's due to a buggy merge in `recording.har`. When you `rm -rf agent/recordings` and record from scratch then the tests fail.

To prevent this from happening again, we're planning to provide scripts to cleanly update the recordings.

## Test plan

Green CI.

PR #2535, opened 2024-01-02
Fix double agent messages with rate limit errors

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/233

## Test plan

Automatic tests to cover the issue were added. 

## Details

Error `Error: Cannot add a bot message after a bot message` was popping out when message was sent to agent with rate limit exhausted. It was caused by `SimpleChatPanelProvider` closing  `typewriter`, which in turn was calling `close` callback from `sendLLMRequest` in `SimpleChatPanelProvider`.
Close callback was then dumping remaining content, resulting in adding agent message twice.

```
    at SimpleChatPanelProvider.addBotMessageWithGuardrails (cody/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts:855:21)
    at Object.close (cody/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts:690:26)
    at Object.close (cody/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts:729:27)
    at Typewriter.stop (cody/lib/shared/src/chat/typewriter.ts:116:27)
    at Object.onError (cody/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts:748:32)
    at onErrorOnce (cody/lib/shared/src/sourcegraph-api/completions/nodeClient.ts:29:20)
    at handleError (cody/lib/shared/src/sourcegraph-api/completions/nodeClient.ts:82:25)
    at IncomingMessage.<anonymous> (cody/lib/shared/src/sourcegraph-api/completions/nodeClient.ts:107:41)
    at IncomingMessage.emit (node:events:529:35)
    at IncomingMessage.emit (node:domain:489:12)
```

PR #2583, opened 2024-01-05
Revert "Revert "refactor: move commands away from recipe (#2542)"" with new fixes

abeatrix:
Close https://github.com/sourcegraph/cody/issues/2588

- Reverts sourcegraph/cody#2575 with updated fixes from https://github.com/sourcegraph/cody/pull/2580 that I've confirmed with @Gedochao on call. 

- Merged main and updated the recording to confirm there are no errors caused by the changes in this branch in all covered tests:

image

- Also confirmed this change also passed the new rate limit test introduced in https://github.com/sourcegraph/cody/pull/2535 

- Fixed upresponsive stop button when error is presented

https://github.com/sourcegraph/cody/assets/68532117/6d9263b0-3e9c-419e-939d-20a2ea011824

## Test Plan

- run `pnpm run test:unit` to confirm all the tests are passing


DanTup (approved):
lgtm, but I'm less familiar with this so you might want another reviewer.

I'm surprised by the size of the `.har` diff though, is it possible it got re-formatted? If not, there might be something else causing significant differences between machines that we might need to track down to make the diffs easy (or possible) to review.

abeatrix:
> lgtm, but I'm less familiar with this so you might want another reviewer.
> 
> I'm surprised by the size of the `.har` diff though, is it possible it got re-formatted? If not, there might be something else causing significant differences between machines that we might need to track down to make the diffs easy (or possible) to review.

It was auto-formatted. Should I turn off auto-format to see if that helps?

DanTup:
> It was auto-formatted. Should I turn off auto-format to see if that helps?

Was it formatted because you had it open/saved it, or do we have something formatting it automatically?

It's probably not important _how_ it's formatted, as long as we ensure it's consistent between our machines so the diffs are small.

Issue #2586, opened 2024-01-05
bug: ignore files don't work for non-codebases


DanTup:
### Version

main f2e64874

### Describe the bug

With `"cody.internal.unstable": true`, files listed in `.cody/ignore` don't work for me.

I think it's because the only places I can find that call `IgnoreHelper.setIgnoreFiles` are guarded by codebase checks:

https://github.com/sourcegraph/cody/blob/cfa212c03ad33ffb99c4605cf0eebae4b6b0eaca/vscode/src/services/context-filter.ts#L104-L113

However my folder is not a git repo and no codebase name is computed.

### Expected behavior

Ignore file should work for basic  folders

DanTup:
@abeatrix if I comment out this check it fixes the issue, but since it had a `return` in it I don't know if that's the correct fix

![image](https://github.com/sourcegraph/cody/assets/1078012/299f7b35-dc81-4e79-b623-63672f7660b1)

abeatrix:
If there is no codebaseName, should we just use a generic one since we use the codebaseName for Mapping in case they are in a multi-root / nested workspace? 🤔

DanTup:
I've opened https://github.com/sourcegraph/cody/pull/2587 - please take a look (we're not `return`ing in the case where there's a codebase name, but I think that should be fine?)

PR #2587, opened 2024-01-05
Don't skip setting ignore files when not in a codebase

DanTup:
Fixes https://github.com/sourcegraph/cody/issues/2586

## Test plan

- Open a folder that is not a Git repo / codebase
- Set `"cody.internal.unstable": true` to enable ignore functionality
- Create a `.cody/ignore` file containing "*.ignore"
- Create files "foo.not" and "foo.ignore"
- Type "@foo" in chat and ensure the `.ignore` file is not shown

image


abeatrix (approved):
Thank you!

PR #2604, opened 2024-01-08
Agent: optimize HTTP recording file for human reviews

olafurpg:
Previously, the recording.har file had too much noise making it difficult to review manually. This PR takes a few steps to make it easier to review the file manually:

- use YAML instead of JSON
- format POST request bodies as JSON objects instead of escaped JSON strings
- hardcode a few useless values

Additionally, this PR adds a new `pnpm update-agent-recordings` script that is the easiest and fastest way to update the recording file.

## Test plan

Run `pnpm update-agent-recordings`, update the snapshots and confirm that the HTTP recording file has a minimal diff (only base64 stuff left along with the actual changes LLM response).


dominiccooney (approved):
This is a nice spelunkability improvement to the HTTP archive.

Issue #2610, opened 2024-01-08
bug: starting doc command from chat create two edits


abeatrix:
Reported by @umpox 

starting doc command from chat create two edits

https://github.com/sourcegraph/cody/assets/68532117/8086f96b-257b-43ac-99b0-7fdb6a27fd86

Version: v1.1.1704726450

PR #2612, opened 2024-01-08
fix: duplicate edit handling in chat

abeatrix:
CLOES https://github.com/sourcegraph/cody/issues/2610

Fix issues where starting edit commands from chat would create two edits

Edit commands are already redirected to the executeEdit function in findCommand, so we should exist early in the command execution flow in chat for edit commands. 

This avoids duplicate handling of edit commands.

- No changelog required as the bug does not exist in released version

## Test plan

- Start the doc command from chat input box
- only one edit should be executed

https://github.com/sourcegraph/cody/assets/68532117/7736cdc4-b65c-4d72-a351-de0c350e5c2b


umpox (approved):
Thank you!!

Issue #2302, opened 2023-12-12
bug: Commands that require open files continue instead of stopping


toolmantim:
### Version

main e75560d7d7380a662a663ed2c10bc1607c7db60b

### Describe the bug

Some commands require an open file and post unhelpful output to the transcript:

https://github.com/sourcegraph/cody/assets/153/2b23fc69-961f-40de-95b0-adcbfe885f5e

### Expected behavior

Only the native VS Code error notification is posted, and both the request and response is never added to the transcript.

kalanchan:
following up on this - this seems to only happen if you invoke commands from the chat input and not the side panel


https://github.com/sourcegraph/cody/assets/51868853/db295441-7d9b-4051-8226-e7d1a683bc56

DanTup:
I had a go at fixing this a couple of different ways, but they all feel bad for different reasons so I'm hoping for some pointers.

The native VS Code error message shown comes from here in the `CommandRunner`:

https://github.com/sourcegraph/cody/blob/2a00da01ae8b83ecd5450af7be2682c32edf3080/vscode/src/commands/CommandRunner.ts#L70-L74

But this code doesn't directly seem to have the ability to influence the call to the LLM. I thought maybe we should set some kind of `errorKind` flag on the command runner to indicate that it won't be able to run.

Here seems like a convenient place to read such a flag to avoid sending anything to the LLM and perhaps showing a nicer error in the chat conversation. There seems to be something similar for missing selections:

https://github.com/sourcegraph/cody/blob/2a00da01ae8b83ecd5450af7be2682c32edf3080/lib/shared/src/chat/recipes/custom-prompt.ts#L61-L63

However, at this point we only have access to the `command` (`CodyPrompt`) and not the `CommandRunner` (since `getCommand` returns only the command, and disposes the runner). Changing `getCommand()` to the runner could work, but since that interface is outside vscode/ presumably would be breaking to other editors.

Putting the `errorKind` on the command (`CodyPrompt`) instead seemed like it could work, and it did.. but it seems like that object may be reused reused (once I trigger the error, every `/smell` command fails with it, even if I've no opened the file). This seems slightly odd to me, as other properties on that class look like they're specific to each prompt.

[...]

PR #2614, opened 2024-01-08
fix: check active editor before executing commands

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2302

Fix issue listed in the linked issue.

- Check if an editor is active and not ignored before executing commands. 
  - Exist early if false
- Show error message if no editor is active or active editor is ignored.

## Test plan

1. Execute a command without opening an editor
2. Only the native VS Code error notification is posted, and both the request and response is never added to the transcript.

image


vscode/src/commands/CommandsController.ts
```diff
@@ -64,6 +65,15 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
         requestID?: string,
         contextFiles?: ContextFile[]
     ): Promise<CodyPrompt | null> {
+        const editor = getEditor()
+        if (!editor.active || editor.ignored) {
+            const message = editor.ignored
+                ? 'The active editor is ignored.'
```
dominiccooney, line 71: This is mildly mysterious. Sourcegraph style we'd give a pointer about why or what to do about this?

dominiccooney (changes requested):
I think the "ignored" error message needs to be more helpful.

vscode/src/commands/CommandsController.ts
abeatrix: @toolmantim any suggestion on what the updated copy should be? 👋

abeatrix:
> I think the "ignored" error message needs to be more helpful.

Do you think the updated message would do?

vscode/src/commands/CommandsController.ts
```diff
@@ -64,6 +65,15 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
         requestID?: string,
         contextFiles?: ContextFile[]
     ): Promise<CodyPrompt | null> {
+        const editor = getEditor()
+        if (!editor.active || editor.ignored) {
+            const message = editor.ignored
+                ? 'The file is on the .cody/ignore list. Please remove it from the list and try again.'
```
DanTup, line 71: Maybe "The active file is ignored by a .cody/ignore file..."? Using "a" instead of "the" might make it clearer that there could be one of these files in the tree (it might be easy to look in the wrong one and think this is a bug if you weren't the one to create them in the repository).

DanTup (approved):
Small suggestion, but LGTM

vscode/src/commands/CommandsController.ts
abeatrix: Makes sense, updated. Thank you so much for the suggestion!

PR #905, opened 2023-09-01
experimental support for VS Code autocomplete with Code Llama (via Ollama running locally)

sqs:
Adds support for autocomplete with Meta's Code Llama model, running locally in [Ollama](https://ollama.ai/).

~Depends on https://github.com/jmorganca/ollama/pull/457.~

Requires Ollama to be running, and the following VS Code settings:

```json
{
	"cody.autocomplete.advanced.provider": "ollama-experimental",
	"cody.autocomplete.experimental.ollamaOptions": {
		"url": "http://localhost:11434",
		"model": "codellama:7b-code"
	}
}
```

TODO instructions etc.

https://github.com/sourcegraph/cody/assets/1976/5d36d8e5-bd69-4786-a70f-0baf79604da8

## Test plan

TODO

PR #2626, opened 2024-01-09
Autocomplete: use `NodeResponse` in the autocomplete client

valerybugakov:
## Context

- This PR applies part of the changes from @sqs's Ollama PR: https://github.com/sourcegraph/cody/pull/905, which is not specific to Ollama support.
- Differentiate between HTTP responses on the web and in node.js on the type level. 
- This allows dropping redundant type casting in the autocomplete HTTP client code.

## Test plan

- CI
- Manually tested autocomplete locally.

### DIFF
--- a/lib/shared/package.json
+++ b/lib/shared/package.json
@@ -41,6 +41,7 @@
     "@types/highlight.js": "^9.12.4",
     "@types/isomorphic-fetch": "^0.0.39",
     "@types/lodash": "^4.14.195",
+    "@types/node-fetch": "^2.6.4",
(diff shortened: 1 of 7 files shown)


dominiccooney (approved):
Looks good. Interesting that `class` is structural without an extra member, that is a surprise.

philipp-spiess (approved):

valerybugakov:
>Interesting that class is structural without an extra member, that is a surprise.

Yep, it's an interesting case. The first time I bumped into this behavior.

PR #2631, opened 2024-01-09
do not require separate build process for storybooks

sqs:
This change makes it so that `pnpm run storybook` in `vscode/` or `lib/ui/` will compile from TypeScript sources, which means that you don't need to have a concurrent `tsc -b` running for changes in other monorepo subprojects to take effect in your storybook.

## Test plan

CI


vdavid (approved):
Makes sense, thanks!

PR #2632, opened 2024-01-09
Autocomplete: Add a new 16b exclusive experiment and document existing flags

philipp-spiess:
For https://docs.google.com/document/d/1P-Hb8BTbz-vMNFXw3gbAvphArsNvS1_8u4ID-phn8gI/edit, we are considereing moving all traffic to the 16b model. When we first run an experiment for this, the single-line CAR was better with the faster model. Part of this might be due to latency (which will improve with the ST deployment) but parts of this might also just no longer be relevant since we have made so many changes.

This PR adds a new client side feature flag that we can run in the short term to gather data on 16b exclusive setups. It will only have an affect if the user is already inside the `CodyAutocompleteStarCoderHybrid` experiment, which is currently our main StarCoder enablement flag.

## Test plan

- Be inside `CodyAutocompleteStarCoderHybrid`
- Be inside `CodyAutocompleteStarCoder16B`
- Observe that single-line completions use the 16b model identifier: 
Screenshot 2024-01-09 at 11 46 51

### DIFF
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -106,10 +106,13 @@ async function resolveDefaultProviderFromVSCodeConfigOrFeatureFlags(configuredPr
         return { provider: configuredProvider }
     }
 
-    const starCoderHybrid = await featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoderHybrid)
+    const [starCoder, starCoder16B] = await Promise.all([
+        featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoderHybrid),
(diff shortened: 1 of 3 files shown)


rafax (approved):
Non-authoritative LGTM :)

vscode/src/completions/providers/create-provider.ts
```diff
@@ -106,10 +106,13 @@ async function resolveDefaultProviderFromVSCodeConfigOrFeatureFlags(configuredPr
         return { provider: configuredProvider }
     }
 
-    const starCoderHybrid = await featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoderHybrid)
+    const [starCoder, starCoder16B] = await Promise.all([
```
valerybugakov, line 109: Would love to keep `starCoderHybrid` here as a variable name for clarity

valerybugakov (approved):

PR #2634, opened 2024-01-09
annotate type-only exports/imports with `type`

sqs:
This fixed an issue with Storybook complaining about not finding type imports (probably because it uses some magic in its bundling). It's also a good hint to the dev about which imports actually affect the output bundle and which are just for `tsc`.

Rerun `pnpm run lint:js --fix && pnpm run format` to recreate this diff.

## Test plan

CI


vdavid (approved):
🙆‍♂️

PR #2638, opened 2024-01-09
Do not require a root build for agent unit tests

philipp-spiess:
See thread in https://sourcegraph.slack.com/archives/C05AGQYD528/p1704748857378909

This makes it so that the unit tests for agent only runs the esbuild portion of the script and not the full build of the workspace root (which can fail for any TS issue in the codebase and leads to confusion)

It does seem like the stdio inheriting does not work because of how vitests isolates different test workers (i.e if the test is run in isolation, the stdio won't be cleared? maybe another test file uses `console.clear`?) In any case it doesn't appear trivial to fix the reporting, so leaving this as-is (Plus I added a comment in the code that is displayed on error on how to get more logs).

## Test plan

- Have a change that does not typecheck, e.g.

```
diff --git a/vscode/src/configuration.ts b/vscode/src/configuration.ts
index ffa6e1a9..f1e0bc95 100644
--- a/vscode/src/configuration.ts
+++ b/vscode/src/configuration.ts
@@ -67,6 +67,8 @@ export function getConfiguration(config: ConfigGetter = vscode.workspace.getConf
         autocompleteExperimentalGraphContext = 'lsp-light'
     }

+    const foo = getHiddenSetting('experimental.symfContext', false)
+
     return {
         proxy: config.get<string | null>(CONFIG_KEY.proxy, null),
         codebase: sanitizeCodebase(config.get(CONFIG_KEY.codebase)),
```

- Run `pnpm test` in the workspace root
- Observe no issues

And when you make the build command error:

Screenshot 2024-01-09 at 15 55 48

### DIFF
--- a/agent/package.json
+++ b/agent/package.json
@@ -12,7 +12,9 @@
   "main": "src/index.ts",
   "sideEffects": false,
   "scripts": {
-    "build": "cd .. && pnpm build && cd agent && node src/esbuild.mjs",
+    "build:root": "cd .. && pnpm build && cd agent",
+    "build:agent": "node src/esbuild.mjs",
+    "build": "pnpm run build:root && pnpm run build:agent",
     "build-minify": "pnpm run build --minify",
     "agent": "pnpm run build && node dist/index.js",
     "agent:debug": "pnpm run build && node --inspect --enable-source-maps ./dist/index.js",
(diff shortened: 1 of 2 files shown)


olafurpg (approved):
Thank you!

Issue #2585, opened 2024-01-05
bug: editing non-file documents with ignore file enabled causes errors


DanTup:
### Version

main f2e64874

### Describe the bug

When editing a file that doesn't have a file URI (such as my workspace settings - although it's not clear to me why that's not a file URI), I see errors like this in the console:

![image](https://github.com/sourcegraph/cody/assets/1078012/a6f68219-c619-449a-88de-af17e12a1bc3)

### Expected behavior

No errors

DanTup:
@abeatrix I'm not sure whether we intend to ignore or not-ignore any no-file URIs, but it actually seems like there's code here that almost handles this, but specifically lets the user-data scheme through (but ignore helper then fails on it because we expected only file URIs):

https://github.com/sourcegraph/cody/blob/cfa212c03ad33ffb99c4605cf0eebae4b6b0eaca/vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts#L101-L109

abeatrix:
> @abeatrix I'm not sure whether we intend to ignore or not-ignore any no-file URIs, but it actually seems like there's code here that almost handles this, but specifically lets the user-data scheme through (but ignore helper then fails on it because we expected only file URIs):
> 
> https://github.com/sourcegraph/cody/blob/cfa212c03ad33ffb99c4605cf0eebae4b6b0eaca/vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts#L101-L109

Interesting! I wasn't aware of the 'vscode-userdata' file but I'd assume we would want that to be ignored by default? wdyt 🤔 ?

DanTup:
Probably - but I wonder if there's other code that might be passing other URIs like this. Should we update `IgnoreHelper` to just reutrn true (always ignore) for non-file URIs instead of throwing? 🤔

DanTup:
I'm failing to reproduce this now.. if you have any ideas what I might've done to go through this code path, let me know! I've tried opening the settings files both as JSON and the UI, having it active, in the background, opened and closed... Unfortunately I didn't keep the full stack trace to know why it was going in jaccard-similarity-retriever.

abeatrix:
> Probably - but I wonder if there's other code that might be passing other URIs like this. Should we update `IgnoreHelper` to just reutrn true (always ignore) for non-file URIs instead of throwing? 🤔
> 
> 
> 
[...]

PR #2640, opened 2024-01-09
fix: ignore all non-file uris by default

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2585

Add a check to ignore all non-file URIs in the IgnoreHelper class. 

This prevents Cody from using files that cannot be uploaded remotely or defined in ignore files as context, e.g. unsaved files or vs code user configuration files.

This also fixed an issue where opening a non-file URIs would cause autocomplete to stop working

## Test plan

1. Open vs code config as tab as described by @philipp-spiess in https://sourcegraph.slack.com/archives/C05AGQYD528/p1704815009033359
2. Switch back to a codebase file to make sure autocomplete is still working (unless the file is ignored)

https://github.com/sourcegraph/cody/assets/68532117/fca1e0e1-205e-4d72-8278-f5101609a1bf

### DIFF
--- a/lib/shared/src/chat/ignore-helper.test.ts
+++ b/lib/shared/src/chat/ignore-helper.test.ts
@@ -40,6 +40,16 @@ describe('IgnoreHelper', () => {
         ignore.setActiveState(true)
     })
 
+    it('returns true for non-file schemed URLs', () => {
+        const nonFileWorkspaceRoot = workspace1Root.with({ scheme: 'non-file' })
+        expect(ignore.isIgnored(Utils.joinPath(nonFileWorkspaceRoot, 'foo.txt'))).toBe(true)
+    })
+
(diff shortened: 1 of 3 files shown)


philipp-spiess (approved):
Great 🤟

DanTup (approved):
lgtm, but if this doesn't break any tests we should add one 😄

Edit: Apparently [we tested](https://github.com/sourcegraph/cody/blob/main/lib/shared/src/chat/ignore-helper.test.ts) everything except non-file URIs...

abeatrix:
> lgtm, but if this doesn't break any tests we should add one 😄
> 
> Edit: Apparently [we tested](https://github.com/sourcegraph/cody/blob/main/lib/shared/src/chat/ignore-helper.test.ts) everything except non-file URIs...

good idea lol just added!

PR #2641, opened 2024-01-09
Agent: add symf support and enabled keyword context by default

olafurpg:
Fixes https://github.com/sourcegraph/jetbrains/issues/256

Previously, symf was disabled in the agent because the download+unzip logic was failing in Node.js. This PR fixes that problem (along with
    several other small issues), and adds a test case to stress the code
path where the user submits a chat message with enhanced context enabled. This test case is disabled for now because symf sends an LLM request that we can't record/replay (since symf is written in Go). The test can be manually enabled locally to test the functionality. It's still useful to merge this PR because it unblocks testing symf context with JetBrains.

**Important**: this PR changes the default context to be keyword-based instead of embeddings-based. This is aligned with our plans for February GA. It's possible to change it back to be embeddings-based by adding the custom configuration:

```json
{
  "cody.useContext": "embeddings"
}
```

Additionally (and optionally), use the following setting to disable symf to get the old default before this PR:

```json
{
  "cody.experimental.symfContext": false
}
```

## Test plan

* Green CI
* (optional) Manually enable the enhanced context test case to see it running locally. Even in replay mode, it still takes ~2 seconds to run because it does an LLM request for query expansion inside symf. Also, the assertion doesn't stress the context functionality because the workspace folder is a tmp directory without git repo.
* Manually test with JetBrains in a real-world git repository and see if the context is any more intelligent.

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -144,8 +144,6 @@ log:
             value: chunked
           - name: connection
             value: keep-alive
-          - name: retry-after
-            value: "0"
           - name: access-control-allow-credentials
             value: "true"
           - name: access-control-allow-origin
@@ -219,50 +217,18 @@ log:
       response:
         content:
           mimeType: text/event-stream
-          size: 641
-          text: >+
+          size: 232
+          text: |+
             event: completion
-
             data: {"completion":" Hello","stopReason":""}
 
-
             event: completion
-
             data: {"completion":" Hello!","stopReason":""}
 
-
             event: completion
-
-            data: {"completion":" Hello! Nice","stopReason":""}
-
-
-            event: completion
-
-            data: {"completion":" Hello! Nice to","stopReason":""}
-
-
-            event: completion
-
-            data: {"completion":" Hello! Nice to meet","stopReason":""}
-
-
-            event: completion
-
-            data: {"completion":" Hello! Nice to meet you","stopReason":""}
-
-
-            event: completion
-
-            data: {"completion":" Hello! Nice to meet you.","stopReason":""}
-
-
-            event: completion
-
-            data: {"completion":" Hello! Nice to meet you.","stopReason":"stop_sequence"}
-
+            data: {"completion":" Hello!","stopReason":"stop_sequence"}
 
             event: done
-
             data: {}
 
         cookies: []
@@ -275,8 +241,6 @@ log:
             value: chunked
           - name: connection
             value: keep-alive
-          - name: retry-after
-            value: "0"
           - name: access-control-allow-credentials
             value: "true"
           - name: access-control-allow-origin
@@ -350,7 +314,7 @@ log:
       response:
         content:
           mimeType: text/event-stream
-          size: 102558
+          size: 107784
           text: >+
             event: completion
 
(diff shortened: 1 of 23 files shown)


olafurpg:
Given that the symf test case isn't working 100% correctly yet, I'd be open to disabling this by default for now to unblock merging this PR and allow manual experimentation. If everything looks good, we can turn it on by default with a separate PR.

agent/recordings/defaultClient_631904893/recording.har.yaml
```diff
@@ -844,512 +808,977 @@ log:
 
             event: completion
 
-            data: {"completion":" Here is a simple Hello World program in Java:\n\n```java\npublic class Main {\n  public static void main(String[] args) {\n    System.out.println(\"Hello World!\"); \n  }\n}\n```\n\nTo break this down:\n\n- The code is wrapped in a class called Main. In Java, code must be inside a class.\n\n- The main method is the entry point of the program. It is marked as static so it can be run without creating an instance of Main.\n\n- The","stopReason":""}
```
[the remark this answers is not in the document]
olafurpg: It's randomness in the LLM. We don't do any processing of the HTTP response yet (beyond unzip+base64 decode)

agent/src/cli/jsonrpc.ts
```diff
@@ -166,11 +166,10 @@ class CodyPersister extends FSPersister {
                 // in in pull requests.
                 try {
                     const text = JSON.parse(responseContent.text)[0]
-                    console.log({ text })
                     const decodedBase64 = decodeCompressedBase64(text)
                     ;(responseContent as any).textDecoded = decodedBase64
-                } catch (error) {
-                    console.error('base64 decode error', error)
+                } catch {
+                    // console.error('base64 decode error', error)
```
pkukielka, line 172: Is that commented out on purpose?

lib/shared/src/sourcegraph-api/telemetry/GraphQLTelemetryExporter.ts
```diff
@@ -63,7 +63,7 @@ export class GraphQLTelemetryExporter implements TelemetryExporter {
             } else {
                 this.exportMode = 'legacy'
             }
-            console.log('telemetry: evaluated export mode:', this.exportMode)
+            // console.log('telemetry: evaluated export mode:', this.exportMode)
```
pkukielka, line 66: Commented out code?

pkukielka (approved):
LGTM

agent/src/cli/jsonrpc.ts
olafurpg: Yes. Added a comment explaining why we can ignore this error.

agent/src/cli/jsonrpc.ts
```diff
@@ -266,12 +265,15 @@ export const jsonrpcCommand = new Command('jsonrpc')
...
             // macOS/Windows/Linux
             polly.server.get('https://github.com/sourcegraph/bfg/*path').passthrough()
+            polly.server.get('https://github.com/sourcegraph/symf/*path').passthrough()
         } else if (options.recordingMode) {
             console.error('CODY_RECORDING_DIRECTORY is required when CODY_RECORDING_MODE is set.')
             process.exit(1)
         }
 
-        process.stderr.write('Starting Cody Agent...\n')
+        if (process.env.CODY_DEBUG === 'true') {
```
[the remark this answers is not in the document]
olafurpg: Agreed. What we need is a reusable logger that doesn't `import * as vscode from 'vscode'`. I tried moving to `logDebug` but we can't use that before we initialize the vscode extension. I was removing console statements to make it easier to hack on the code. There's almost no noisy/redundant console output when running agent tests now.

lib/shared/src/sourcegraph-api/telemetry/GraphQLTelemetryExporter.ts
olafurpg: Just commented out to reduce noise. Left it around if somebody wants to add it back. I'm guessing nobody was using this output in a meaningful way. It probably should be `logDebug()` if we move lib/shared/ into vscode/

agent/src/index.test.ts
```diff
@@ -339,19 +348,55 @@ describe('Agent', () => {
...
+    it.skip('allows us to send a chat message with enhanced context enabled', async () => {
+        await openFile(animalUri)
+        await client.request('command/execute', { command: 'cody.search.index-update' })
+        const lastMessage = await client.sendSingleMessage(
+            'Write a class Dog that implements the Animal interface in my workspace. Only show the code, no explanation needed.',
+            {
+                addEnhancedContext: true,
+            }
+        )
+        // TODO: make this test return a TypeScript implementation of
```
pkukielka, line 375: I missed that comment before.
Do does that test in fact prove that `symf` works at all?
I would like to fix it, where is that git error logged?

olafurpg: The test does not prove `symf` works, and I have manually confirmed that it doesn't actually work here because of a git-related error. If you replace `ignore` with `inherit` on this line here then the git errors get printed out to the console https://sourcegraph.com/github.com/sourcegraph/cody@main/-/blob/vscode/src/local-context/symf.ts?L322

The first step to fixing the problem is to merge this PR here, which moves term expansion into the vsc extension so we can run an offline replay test with enhanced context enabled. This test is enabled in this PR. https://github.com/sourcegraph/cody/pull/2644

The next step is to update the test case to open an actual git workspace instead of a tmp directory without git metadata.

PR #2643, opened 2024-01-09
VS Code: Disable displaying feature limits

vdavid:
The problem is that we're displaying 160 chats in the error message:

image

while on the /cody/manage page, we display 20:

image

Cody Gateway only knows the LLM limits (which is 160 for Free users), and that's what it returns in the error header, so that's all the client knows as well. We should display the official limit of 20 instead, but we can't know it for now on the client side.

So my quick solution is just to hide this info on the UI.

## Test plan

Relying on the VS Code team to test this

### DIFF
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -34,7 +34,7 @@ export class RateLimitError extends Error {
     ) {
         super(message)
         if (upgradeIsAvailable) {
-            this.userMessage = `You've used all${limit ? ` ${limit}` : ''} ${feature} for the month.`
+            this.userMessage = `You've used all ${feature} for the month.`
         } else {
             // Don't display Pro & Enterprise’s fair use limit numbers, as they're for abuse protection only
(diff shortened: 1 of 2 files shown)


kalanchan:
Raised [here](https://sourcegraph.slack.com/archives/C05SZB829D0/p1704824393085649) 

cc @sourcegraph/cody-clients @toolmantim

toolmantim (approved):

chrsmith (approved):
This seems fine to me, but is it guaranteed that `${feature}` will always make sense when used as a strong like this? e.g. "used all code completions for the month" makes sense, but "used all code completion (singular)" doesn't.

Do we need to have some sort of function like `pluralize(feature: string): string` to be sure? Or is that just overkill? (Since for Cody Gateway, IIRC there are like 3-4 features that it cares about.

toolmantim:
@chrsmith no need, `feature` is populated by the code in this repo: https://github.com/search?q=repo%3Asourcegraph%2Fcody+%22new+RateLimitError%22&type=code

It would be nice to change feature from `string` to `'autocompletions' | 'chat messages and commands'` tho

PR #2644, opened 2024-01-09
Symf: move term expansion into VS Code extension

olafurpg:
Previously, we shelled out to `symf expand-query` to do term expansion
on the query. This approach prevented us from writing stable tests with
enhanced context because the LLM request inside the `symf` binary was
not recorded by Polly. This PR fixes the problem by reimplementing the
`expand-query` subcommand in TypeScript so that we can record the HTTP
request.

## Test plan

See new integration tests for `expandQuery`. Ideally, we will tweak the HTTP recording format to only return the latest completion event so PRs like this don't have to be a 6k+ diff.


vscode/src/local-context/symf.test.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        function check(query: string, expectedHandler: (expandedTerm: string) => void): void {
+            it(query, async () => {
+                expectedHandler(await expandQuery(client, query))
+            })
+        }
+
+        check('ocean', expanded =>
+            expect(expanded).toMatchInlineSnapshot(
+                '"current flow ocean salinity salt saltiness sea stream surf tidal tide water wave waves"'
```
beyang, line 32: How do you guarantee determinisim here?

beyang (approved):

vscode/src/local-context/symf.test.ts
olafurpg: Correct, the tests run entirely offline in "replay" mode where Polly returns the same LLM response as long as the request parameters (URL, body, headers) are unchanged. The tests don't require any custom environment variables when running in replay mode. The HTTP recording expires in 365 days (configurable duration).

I added a `pnpm update-symf-recordings` script to record new HTTP responses against sourcegraph.com with your access token.

Issue #1475, opened 2023-10-24
VSCE: Improve Generate Unit Tests command


abeatrix:
Design: https://github.com/sourcegraph/cody/issues/1359
https://github.com/sourcegraph/cody/issues/1359

Helpful feedback from users: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1698096296523479

```[tasklist]
### Tasks
- [ ] https://github.com/sourcegraph/cody/issues/1761
- [ ] https://github.com/sourcegraph/cody/pull/2646
- [ ] https://github.com/sourcegraph/cody/issues/2758
- [ ] https://github.com/sourcegraph/cody/pull/2959
```

TBD: 
- new edit mode for creating file with new content?
- make context for edit commands configurable

#### Claude 

Based on the shared context, the test framework in use is Vitest and the following libraries are imported:

```
import { describe, expect, it, test } from 'vitest'
import * as vscode from 'vscode'
```

No new imports are needed since Vitest and VS Code APIs are already imported.

I will generate a suite of unit tests for the setLastStoredCode and matchCodeInStore functions in , using Vitest assertions and mocks where applicable:

```
import { describe, expect, it } from 'vitest'

import { setLastStoredCode, matchCodeInStore } from './CodeblockActionHandlers'

describe('setLastStoredCode', () => {

  it('sets lastStoredCode with expected properties', () => {
    const code = 'console.log("hello")'
    const eventName = 'insertButton'
    
    const result = setLastStoredCode(code, eventName)

    expect(result).toMatchObject({
      code,
      eventName,
      lineCount: 1,
      charCount: 21
    })
  })

  it('sets insertInProgress to false for copy events', () => {
    const code = 'console.log("hello")'
    const eventName = 'copyButton'
    
    setLastStoredCode(code, eventName)

    expect(insertInProgress).toBe(false)
  })

})

describe('matchCodeInStore', () => {

  it('returns false if insert is in progress', () => {
    insertInProgress = true

    expect(matchCodeInStore('console.log("hello")')).toBe(false)
  })

  it('returns true if code matches lastClipboardText and lastStoredCode', () => {
    insertInProgress = false
    lastClipboardText = 'console.log("hello")'
    lastStoredCode.code = 'console.log("hello")'

    expect(matchCodeInStore('console.log("hello")')).toBe(true)
  })

})
```

[...]

PR #2646, opened 2024-01-09
feat: add generated unit test to files automatically

abeatrix:
Part of https://github.com/sourcegraph/cody/issues/1475

- Add new command for generating unit test in file so that it'd work with the edit command (will replace the old /test command once it's out of experimental)
- Set it behind a config `cody.experimental.commands`
- Add new EditMode `file` for inserting response to a different file (only used by /test currently)
- Generates new unsaved test files based on existing test files in codebase if LLM didn't provide us with a name suggestion
- Adds the generated test to the unsaved test file for users to review
- Some additional codebase cleanup for commands

## Note

This will not affect the /test command used by agent. In a follow up PR I will add a param that allows all commands to be run in chat mode only

## Test plan

0. Enable `cody.experimental.commands`
1. Run the /test-file command in the selected code to generate unit tests
2. If there is a test file exists for the file that contains your selected code, the new unit tests will be added below the import statements in the file
  - if the user is not happy with the generated tests, simply close the file with the unsaved change 
3. If a test file does not exist for the selected file, the new unit tests will be added to a new unsaved file with the proposed file name for the new test file
  - user can then review the file before hitting save

In addition, confirm the following works as expected:
- [ ] All existing tests are passing
- [ ] Other commands should not be affected by this change

## Demo

https://github.com/sourcegraph/cody/assets/68532117/04d4499b-752d-4f61-938a-793f58f5cde9

## Follow-ups

To-dos in follow-up PRs:
- [x] Add code lenses in the new test file with undo button
- [ ] Update prompt to eliminate duplicated imports in new tests


vscode/src/commands/utils/new-test-file.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    return fileName.includes('test') && !fileName.includes('test-')
+}
+
+const TEST_FILE_EXTENSIONS = new Set(['ts', 'js', 'tsx', 'jsx', 'py', 'rb', 'go', 'cs', 'cpp', 'cc'])
+// Language extension that uses '.test' suffix for test files
+const TEST_FILE_DOT_SUFFIX_EXTENSIONS = new Set(['js', 'ts', 'jsx', 'tsx'])
+// language extension that uses '_test' suffix for test files
+const TEST_FILE_DASH_SUFFIX_EXTENSIONS = new Set(['py', 'rb', 'go'])
```
toolmantim, line 24: Most large Ruby projects use `spec/${dir}/${file}_spec.rb` btw. In Rails for example the file `app/models/user.rb` would have the test (spec) `specs/models/user_spec.rb`

abeatrix: Updated the code and the tests to reflect this 👍

toolmantim: Smaller Ruby gems use minitest — it would be great if we could detect their test framework of choice 🤔

If you want to test the spec generation works, https://github.com/mastodon/mastodon is a good app to use:

* Existing model & spec: https://github.com/mastodon/mastodon/blob/main/app/models/account/field.rb & https://github.com/mastodon/mastodon/blob/main/spec/models/account/field_spec.rb
* Creating a new spec (doesn't have one): https://github.com/mastodon/mastodon/blob/main/app/validators/vote_validator.rb

abeatrix: > Smaller Ruby gems use minitest — it would be great if we could detect their test framework of choice 🤔
> 
> 
> 
> If you want to test the spec generation works, https://github.com/mastodon/mastodon is a good app to use:
> 
> 
> 
> * Existing model & spec: https://github.com/mastodon/mastodon/blob/main/app/models/account/field.rb & https://github.com/mastodon/mastodon/blob/main/spec/models/account/field_spec.rb
> 
> * Creating a new spec (doesn't have one): https://github.com/mastodon/mastodon/blob/main/app/validators/vote_validator.rb

We do try to detect their test framework of choice first. The function you commented on is used as fallback when no test files and test frameworks are detected in their workspace.

abeatrix: Will look into having llm to return a file name for the test in a follow up PR though!

(Right now I'm trying to check with Tom to see if this new file creation flow added to fixup makes sense, hence I only added him as a reviewer before the team 😅)

toolmantim: Ah okay cool, no problem! Once a PR is out of draft I consider it a comment-now-or-never situation 😅

abeatrix: I tried to turn it into a draft after I created the issue, but I can't find the button for it so that was my bad 😂 

I'll make sure to tag you once this is ready!

vscode/src/edit/execute.ts
```diff
@@ -1,17 +1,18 @@
...
 
 export interface ExecuteEditArguments {
     document?: vscode.TextDocument
     instruction?: string
     userContextFiles?: ContextFile[]
     intent?: EditIntent
     range?: vscode.Range
-    insertMode?: boolean
+    mode?: EditMode
+    command?: CodyCommand
```
umpox, line 15: I understand why we're doing it here but I'm wary of `Edit` being directly coupled with `Command` like this. It just means we start branching things and then we have to change both if we start making changes to one.

We already have `userContextFiles`, could we add an argument for `userContextMessages` too?

This would mean we don't need a special `getContextForCommand` function and we may find other cases where we want to provide additional context messages.

The logic for creating the context messages can live with the command code, and it can just be provided when `executeEdit` is called

WDYT?

umpox (commented):
Feels really nice! Left some comments. 

Mainly just wary of how we link Commands and Edits... It'd be awesome if we can abstract this slightly more - and might mean we can reuse this code easier in the future for any non-Command requirements

vscode/src/non-stop/FixupController.ts
```diff
@@ -765,6 +777,32 @@ export class FixupController
...
+         *
+         * NOTE: Currently used for /test command only.
+         */
+        if (task.mode === 'file') {
+            const newFile = NewFixupFileMap.get(task.id)
+            if (newFile?.fsPath) {
+                task.fixupFile = this.files.forUri(newFile)
+                const doc = await vscode.workspace.openTextDocument(newFile)
+                // If the file is not empty, insert text after the imports
+                const importRange = await getImportsRange(newFile)
```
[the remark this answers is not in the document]
abeatrix: Will update it to use last line instead of import range 👍 

I was hoping it would make it easier to merge duplicated import statements, but after playing around with this today i gave up lol

abeatrix:
> Feels really nice! Left some comments. 
> 
> 
> 
> Mainly just wary of how we link Commands and Edits... It'd be awesome if we can abstract this slightly more - and might mean we can reuse this code easier in the future for any non-Command requirements

Thank you so much for your input @umpox! What you suggested makes sense and I've made some new changes in my local branch to reflect some of them. 
I have some questions for you regarding the context fetching step, will discuss more during our 1:1 tomorrow. Thanks again for the review Tom!

vscode/src/non-stop/FixupController.ts
```diff
@@ -765,6 +777,32 @@ export class FixupController
...
+                const lastLineOfImports = importRange.end.line
+                task.selectionRange = new vscode.Range(lastLineOfImports, 0, lastLineOfImports, 0)
+                // insert text to the new file
+                task.mode = 'insert'
+                // Show the new document before streaming start
+                await vscode.window.showTextDocument(doc)
+            }
+            NewFixupFileMap.delete(task.id)
+        }
```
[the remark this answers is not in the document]
abeatrix: Moved to using the response from LLM to generate the file

vscode/src/edit/execute.ts
abeatrix: replaced command with `ContextMessages` (since they are not from the user)

umpox:
image

This notification shows 🤔  It should usually only show if the file is not visible but I think it should be.

This is the code in `streamTask`:
```
            if (!visibleEditor) {
                await this.notifyTaskComplete(task)
            }
```

vscode/src/edit/provider.ts
```diff
@@ -160,4 +194,46 @@ export class EditProvider {
...
+        }
+
+        const opentag = `<${PROMPT_TOPICS.FILENAME}>`
+        const closetag = `</${PROMPT_TOPICS.FILENAME}>`
+        const isCompleted = text.indexOf(closetag)
+
+        const currentFileUri = this.config.task.fixupFile.uri.fsPath
+        const currentFileName = path.basename(currentFileUri)
+        // remove open and close tags from text
+        const newFileName = text.trim().replaceAll(new RegExp(opentag + '(.*)' + closetag, 'g'), '$1')
```
umpox, line 220: I think the multiplexer should already strip the open and close tag for you, so you might not need these.

umpox (commented):
Looking really good, left some comments!

I noticed that it can duplicate some stuff when there's already an existing test file with tests in it. I wonder if there's a nice way to provide the `precedingText` if the test file already exists 🤔 We use `precedingText` for the `add` intent and it works well to avoid duplication. Maybe one for a follow up PR though

vscode/src/non-stop/codelenses.ts
```diff
@@ -178,3 +179,12 @@ function getAcceptLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens
...
 }
+
+function getGeneratingTestLens(codeLensRange: vscode.Range): vscode.CodeLens {
+    const lens = new vscode.CodeLens(codeLensRange)
+    lens.command = {
+        title: '$(sync~spin) Cody is generating tests...',
+        command: 'cody.focus',
+    }
+    return lens
+}
```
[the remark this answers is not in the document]
abeatrix: removed

vscode/src/edit/provider.ts
abeatrix: removed check!

vscode/src/non-stop/FixupController.ts
```diff
@@ -803,7 +819,7 @@ export class FixupController
 
     public async didReceiveFixupText(id: string, text: string, state: 'streaming' | 'complete'): Promise<void> {
         const task = this.tasks.get(id)
-        if (!task) {
+        if (!task || task.state === CodyTaskState.pending) {
```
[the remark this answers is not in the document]
abeatrix: removed

abeatrix:
@umpox it's now behind an experimental configuration. 

Regarding `I noticed that it can duplicate some stuff when there's already an existing test file with tests in it.` it's an existing issue that I will look into in a follow up PR since it's more of a prompting issue.

vscode/package.json
```diff
@@ -830,6 +846,12 @@
           "markdownDescription": "Enable OpenTelemetry tracing",
           "default": false
         },
+        "cody.experimental.commands": {
+          "order": 99,
+          "type": "boolean",
+          "markdownDescription": "Enable the updated generate unit tests command",
+          "default": false
+        },
```
umpox, line 854:
```suggestion
        "cody.experimental.inlineUnitTest": {
          "order": 99,
          "type": "boolean",
          "markdownDescription": "Enable the updated generated unit tests command",
          "default": false
        },
```

nit: more specific, in case we have another `experimental.commands` flag.

Feel free to ignore this if we're going to quickly delete this flag soon

vscode/src/main.ts
```diff
@@ -322,16 +322,22 @@ const register = async (
...
         vscode.commands.registerCommand('cody.action.commands.exec', async title => executeCommand(title)),
         vscode.commands.registerCommand('cody.command.explain-code', async () => executeCommand('/explain')),
         vscode.commands.registerCommand('cody.command.generate-tests', async () => executeCommand('/test')),
+        vscode.commands.registerCommand('cody.command.unit-tests', async () => {
+            if (!config.experimentalCommands) {
+                void vscode.window.showWarningMessage(
+                    'This is an experimental feature. Please enable the cody.experimental.commands in the extension settings and try again.'
+                )
+                return
+            }
```
umpox, line 338: Do we need this? I thought it'd be impossible to trigger this command as we have a `when` with a check against `config.cody.experimental.commands`

vscode/src/non-stop/FixupTask.ts
```diff
@@ -36,6 +37,8 @@ export class FixupTask {
     public diff: Diff | undefined
     /** The number of times we've submitted this to the LLM. */
     public spinCount = 0
+
+    public skipContent?: string
```
umpox, line 41:
```suggestion
```

Seems unused

umpox (approved):

vscode/package.json
abeatrix: > Feel free to ignore this if we're going to quickly delete this flag soon

Yea since i'm planning to delete this flag soon, I will just hide it behind `internal.unstable` instead, thanks!

Issue #2390, opened 2023-12-15
bug: @-mentioning a binary file causes chat to get stuck


DanTup:
### Version

main 2a00da01

### Describe the bug

If I select a binary file in chat, I get this error message at the top and the chat continues to flash as if it's working, but it never completes:

![image](https://github.com/sourcegraph/cody/assets/1078012/3784a4a7-d650-4074-b0df-c4bb859b494b)

This prevents further interaction in that chat window.

(note: the markup shown in the chat window is filed at https://github.com/sourcegraph/cody/issues/2388 and I believe unrelated).

### Expected behavior

I should be able to continue to chat.

sqs:
Confirmed repro:

![image](https://github.com/sourcegraph/cody/assets/1976/253a028d-749e-49e6-b77a-0562aeccfd98)

PR #2650, opened 2024-01-10
skip (instead of hanging indefinitely on) non-text context files

sqs:
If a file is `@`-mentioned that is inaccessible or unreadable, Cody chat would hang indefinitely. Now, it skips the file, shows the user an error message, and proceeds with the chat response.

The most likely causes for a file being inaccessible or unreadable are: the file is binary, or the file was removed after the user selected it.

Fix https://github.com/sourcegraph/cody/issues/2390.

## Test plan

Added test case.


philipp-spiess (approved):
Awesome!

Issue #2584, opened 2024-01-05
bug: enhanced context shows paths with leading backslashes on Windows


DanTup:
### Version

main cfa212c0

### Describe the bug

When running the e2e test "should be able to index, then search, a git repository" on Windows, the paths shown back leading backslashes:

![image](https://github.com/sourcegraph/cody/assets/1078012/da03a2a6-d0ce-4e08-bc3a-c34d0d2cc6e1)

### Expected behavior

The paths should be relative (and not start with a slash) or absolute (and be a full path including drive letter on Windows)

DanTup:
This has been fixed by @sqs's changes to pass URIs around and use a standard function for display:

https://github.com/sourcegraph/cody/assets/1078012/928dd4d4-a566-4816-99ec-ad65ed2ac682

Issue #2330, opened 2023-12-13
bug: when there are multiple workspace folders, local embeddings appears to search the wrong one


beyang:
![Image](https://github.com/sourcegraph/cody/assets/1646931/8adb6b2d-f879-4e38-b6ed-dfe6d14954d8)

Video: https://www.loom.com/share/5a4148a4ab9849e6acaeb264376d7682

A related customer report: https://sourcegraph.slack.com/archives/C05AGQYD528/p1703114013765599
- Another ws folder is searched (rather than the one with the current active file)
- In this case, the ✨ menu doesn't even show the other ws folder is part of the context used (may be a bug in the way we display symf index availability in the ✨ menu)

PR #2653, opened 2024-01-10
remove `ContextFile.fileName`, use URIs instead

sqs:
This PR addresses the following problems:

* `ContextFile.fileName` is overloaded for many different things, including: filenames (with different transformations applied), symbol names, and representations of context files in the chat input. This was making the code hard to read/understand/fix for me.
* URIs were encoding relative file paths (that were looked up at resolution-time in an ambiguous and duplicative manner), ranges, symbol names, etc. We should not try to jam this all into a URI; where needed, we should just pass these values in separate arguments or properties.

See https://sourcegraph.slack.com/archives/C05AGQYD528/p1704869760165279 for more info.

## Solution

* Better typing for `ContextFile` 
* Removed of `ContextFile.fileName`; use `ContextFile.uri` instead
* Removed a lot of unnecessary usage of `vscode.Uri.file` and `uri.fsPath`
* Removed "janky" usage of URIs to pass additional information, instead passing that in separate args or properties
* URIs are made into "nice" workspace-relative paths to show the user at *display time* (as late as possible) using `displayPath(uri)`, so that the data we pass around need not involve presentation concerns

## Notes

In some areas (such as mapping embeddings results filenames to URIs), support for multi-root workspaces regressed. I am not sure if it ever *actually* worked on multi-root workspaces. There were a lot of implicit relative file lookups that probably would find the wrong file for ambiguous filenames among the different workspace folders.

I'd like to merge this sooner and address those in a followup soon.

## Test plan

* CI
* Add `@`-mentioned files and symbols and ensure chat messages are sent and responses received properly. Also ensure that the chat is revived properly.
* Run NLS search and ensure the results are clickable.


lib/shared/src/chat/transcript/interaction.ts
```diff
@@ -41,14 +41,15 @@ export class Interaction {
...
-                const { uri, repoName, fileName, source } = message.file
+                const { uri, repoName, source } = message.file
                 if (uri && isCodyIgnoredFile(uri)) {
                     i++
                     continue
                 }
                 // Filter embedding results from the current workspace
                 if (source === 'embeddings') {
-                    if (repoName && isCodyIgnoredFilePath(repoName, fileName)) {
+                    // TODO(sqs): need to make sure this properly filters files out - I dont think this is correct - https://sourcegraph.slack.com/archives/C05AGQYD528/p1704886845816059
```
sqs, line 51: This is the only TODO that blocks merging. Discussion in https://sourcegraph.slack.com/archives/C05AGQYD528/p1704886845816059.

sqs:
Ready for review. I will not merge until https://github.com/sourcegraph/cody/pull/2653#discussion_r1448628661 is fixed, but I'd love a review earlier.

sqs:
(Commit history is messier than I'd like but it'll all be squashed anyway, and it was very hard to make atomic commits here.)

vscode/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "sourceMap": true,
+    "sourceMap": false,
```
sqs, line 4: I just noticed this and a couple other changes to the build config that I will break out into a separate PR. Disregard during review.

vscode/vite.config.ts
```diff
@@ -4,21 +4,22 @@ import { defineConfig } from 'vite'
 
 export default defineConfig(({ mode }) => ({
     resolve: {
-        alias:
-            mode === 'development'
+        alias: [
+            ...(mode === 'development' || mode === 'test'
```
sqs, line 8: Another build change that I will separate into its own PR (https://github.com/sourcegraph/cody/pull/2653/files#r1448656832)

abeatrix:
`.cody/ignore` doesn't seem to be working properly on this branch.

Results from "selection" and "search" (symf) are not filtered:

image

image

Results from embeddings are filtered, but it's displaying as "absolute path" instead:
image

Will continue the review after my meetings!

sqs:
@abeatrix Thanks! I will look into both issues (ignores not being applied, and the absolute URIs showing).

sqs:
@abeatrix In your screenshot, the ignore file is named `.cody/.ignore`. Is the second `.` (in `.ignore`) intentional?

vscode/vite.config.ts
sqs: done https://github.com/sourcegraph/cody/pull/2691

vscode/tsconfig.json
sqs: done https://github.com/sourcegraph/cody/pull/2691

sqs:
@abeatrix I just tried it with `.cody/ignore` (and with `	"cody.internal.unstable": true`).  Ignores are properly respected for remote embeddings, local embeddings, and symf AFAICT.

lib/shared/src/chat/transcript/interaction.ts
sqs: Not a blocker anymore; this code path will be deleted in https://github.com/sourcegraph/cody/pull/2692.

lib/shared/src/chat/transcript/index.ts
```diff
@@ -58,7 +60,7 @@ export class Transcript {
 
                                 const { fileName } = message as any as OldContextMessage
                                 if (fileName) {
-                                    return { ...message, file: { fileName } }
+                                    return { ...message, file: { type: 'file', uri: URI.file(fileName) } }
```
DanTup, line 63: I can't find where this code is called from, but is it certain any existing data in `fileName` is a valid absolute path that will succeed passed to `URI.file`?

lib/shared/src/chat/transcript/transcript.test.ts
```diff
@@ -243,7 +244,10 @@ describe('Transcript', () => {
 
     it('includes currently visible content from the editor', async () => {
         const editor = new MockEditor({
-            getActiveTextEditorVisibleContent: () => ({ fileName: 'internal/lib.go', content: 'package lib' }),
+            getActiveTextEditorVisibleContent: () => ({
+                fileUri: URI.file('/internal/lib.go'),
```
DanTup, line 248: This path doesn't look correct for Windows. Could we use something like `testFilePath` from `lib\shared\src\test\path-helpers.ts` so that on Windows bots the code runs with Windows-style paths? (otherwise it's easy to have code that accidentally works only on Mac/Linux and tests for it that pass on Windows bots).

lib/shared/src/codebase-context/index.ts
```diff
@@ -247,12 +256,15 @@ export class CodebaseContext {
...
 
-function groupResultsByFile(results: EmbeddingsSearchResult[]): { file: ContextFile; results: string[] }[] {
-    const originalFileOrder: ContextFile[] = []
+function groupResultsByFile(
+    results: EmbeddingsSearchResult[]
+): { file: ContextFile & Required<Pick<ContextFile, 'uri'>>; results: string[] }[] {
+    const originalFileOrder: (ContextFile & Required<Pick<ContextFile, 'uri'>>)[] = []
     for (const result of results) {
-        if (!originalFileOrder.find((ogFile: ContextFile) => ogFile.fileName === result.fileName)) {
+        const resultUri = URI.file(result.fileName)
```
DanTup, line 264: I'm not very familiar with embeddings, but is it certain this `fileName` here is absolute? I understood that there are some places where paths would be relative to the codebase root, but I don't know enough about it to tell if that could be the case here.

lib/shared/src/prompt/templates.ts
```diff
@@ -134,16 +127,16 @@ const SELECTED_CODE_CONTEXT_TEMPLATE_WITH_REPO = `"My selected {languageName} co
...
+    const extension = getFileExtension(fileUri)
     const languageName = getNormalizedLanguageName(extension)
     return (
         repoName
             ? SELECTED_CODE_CONTEXT_TEMPLATE_WITH_REPO.replace('{repoName}', repoName)
             : SELECTED_CODE_CONTEXT_TEMPLATE
     )
         .replace('{code}', code)
-        .replaceAll('{filePath}', filePath)
+        .replaceAll('{filePath}', displayPath(fileUri))
```
DanTup, line 139: Are these template tokens like `{filePath}` something we can change? If so, it might be less confusing to rename to `displayPath` or `relativePath` or something to avoid future confusion with something called "filePath" that is an incomplete/relative path.

lib/shared/src/test/testSetup.ts
```diff
@@ -0,0 +1,7 @@
+import { beforeAll } from 'vitest'
+
+import { setDisplayPathFn } from '../editor/displayPath'
+
+beforeAll(() => {
+    setDisplayPathFn(location => (typeof location === 'string' ? location : location.path.slice(1)))
```
DanTup, line 6: I'm not sure if this will do as expected with Windows file URIs (`file:///C:/foo/bar`). Possibly should use `fsPath` and then make the path relative to the root?

(it's possible no tests currently send Windows paths here, but if that's the case I think we should try to improve that)

lib/ui/src/chat/fixtures.ts
```diff
@@ -22,8 +24,8 @@ export const FIXTURE_TRANSCRIPT: Record<string, ChatMessage[]> = {
             displayText:
                 "Explain the following code at a high level:\n\n```\nprivate getNonce(): string {\n  let text = ''\n  const possible = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'\n  for (let i = 0; i < 32; i++) {\n    text += possible.charAt(Math.floor(Math.random() * possible.length))\n  }\n  return text\n}\n```",
             contextFiles: [
-                { fileName: 'vscode/src/chat/ChatViewProvider.ts' },
-                { fileName: 'lib/shared/src/timestamp.ts' },
+                { type: 'file', uri: URI.file('/vscode/src/chat/ChatViewProvider.ts') },
+                { type: 'file', uri: URI.file('/lib/shared/src/timestamp.ts') },
```
DanTup, line 28: As above, I think we should ensure these are the correct format for the current platform.

slack/src/slack/message-interaction.ts
```diff
@@ -49,15 +49,7 @@ class SlackInteraction {
                 contextSearchOptions
             )
 
-            this.contextMessages.push(
-                ...contextMessages.map(message => {
-                    if (message.file) {
-                        message.file.fileName = `https://${codebase}/blob/main/${message.file.fileName}`
```
DanTup, line 55: I'm not sure what this was doing (a https URL in filename?!) - was it unused? (I can't see an obvious replacement).

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1301,7 +1297,10 @@ class ContextProvider implements IContextProvider {
                 new vscode.Position(result.startLine, 0),
                 new vscode.Position(result.endLine, 0)
             )
-            const uri = relativeFileUri('', result.fileName, range)
+
+            // TODO(sqs): this is broken for multi-root workspaces because it assumes that the file
+            // exists in the first workspaceFolder and that the file still exists.
+            const uri = vscode.Uri.joinPath(workspaceFolder.uri, result.fileName)
```
DanTup, line 1303: I guess this is what the ignore codebase code was intended to handle. Maybe it wasn't wired up properly.

As a possible quick-fix (to avoid sending any files that are supposed to be ignored), we could check `IsCodyIgnoredFile()` for _all_ workspace folders, and ignore it if any of them are?

We really need a simple way to resolve paths from codebases to their local full paths though.

vscode/src/chat/chat-view/chat-helpers.ts
```diff
@@ -207,13 +74,9 @@ export function stripContextWrapper(text: string): string | undefined {
 export function contextItemsToContextFiles(items: ContextItem[]): ContextFile[] {
     const contextFiles: ContextFile[] = []
     for (const item of items) {
-        let relFsPath = item.uri.fsPath
-        if (relFsPath.startsWith('/')) {
-            relFsPath = relFsPath.slice(1)
-        }
```
DanTup, line 213: Thank you for all this... this code has been bugging me (see https://github.com/sourcegraph/cody/issues/2584) but I hadn't gotten to it (perhaps in part because I suspected the fix would end up becoming something like this PR 😄)

vscode/src/commands/prompt/display-text.test.ts
```diff
@@ -1,117 +1,144 @@
...
+import { setDisplayPathFn } from '@sourcegraph/cody-shared'
 
 import { replaceFileNameWithMarkdownLink } from './display-text'
 
 describe('replaceFileNameWithMarkdownLink', () => {
-    it('replaces file name with markdown link', () => {
-        const text = 'Hello @test.js'
-
-        const result = replaceFileNameWithMarkdownLink(text, '@test.js', '/path/to/test.js')
+    // Mock a `displayPath` function that always uses slashes (even on Windows).
```
DanTup, line 10: "always uses _forward_ slashes"?

vscode/src/commands/prompt/display-text.test.ts
```diff
@@ -1,117 +1,144 @@
...
 
-        expect(result).toEqual(
-            'Hello [_@test.js_](command:cody.chat.open.file?%22%2Fpath%2Fto%2Ftest.js%3Arange%3A0%22)'
+    let origDisplayPath: any
+    beforeEach(() => {
+        origDisplayPath = setDisplayPathFn(location =>
+            (typeof location === 'string' ? location : location.path.slice(1)).replaceAll('\\', '/')
```
DanTup, line 15: Similar to above - I think `.slice(1)` will still be wonky on Windows because of the drive letter.

vscode/src/commands/prompt/display-text.ts
```diff
@@ -54,23 +68,35 @@ export function createDisplayTextWithFileSelection(
  */
 export function replaceFileNameWithMarkdownLink(
     humanInput: string,
-    fileName: string,
-    fsPath: string,
-    startLine = 0
+    file: vscode.Uri,
```
DanTup, line 71: nit: `file` -> `fileUri`?

vscode/src/editor-context/helpers.ts
```diff
@@ -241,11 +230,9 @@ export function createHumanDisplayTextWithDocLink(
     selection: ActiveTextEditorSelection
 ): string {
     const { range, start } = createSelectionDisplayText(selection)
-    const fsPath = docUri.fsPath
-    const fileName = createVSCodeRelativePath(fsPath)
-    const fileLink = `vscode://file${fsPath}:${start}`
+    const fileLink = `vscode://file${docUri.fsPath}:${start}`
```
DanTup, line 233: This isn't part of your change, but this link looks a bit suspicious for Windows (`vscode://filec:/foo`) 😬

vscode/src/editor/utils/editor-context.test.ts
```diff
@@ -25,7 +27,7 @@ describe('getFileContextFiles', () => {
     async function runSearch(query: string, maxResults: number): Promise<(string | undefined)[]> {
         const results = await getFileContextFiles(query, maxResults, new vscode.CancellationTokenSource().token)
 
-        return results.map(uri => uri.path?.basename)
+        return results.map(f => basename(f.uri.path))
```
DanTup, line 30: Probably should be `fsPath` (I suspect basename tolerates forward slashes on Windows, but it's a good habit... ultimately it'd be good to have a lint for `uri.path` because it often leads to Windows-specific bugs).

vscode/src/testutils/testSetup.ts
```diff
@@ -0,0 +1,7 @@
+import { beforeEach } from 'vitest'
+
+import { setDisplayPathFn } from '@sourcegraph/cody-shared'
+
+beforeEach(() => {
+    setDisplayPathFn(location => (typeof location === 'string' ? location : location.path.slice(1)))
```
DanTup, line 6: (as above - might not do what you want on Windows)

DanTup (commented):
Thank you for looking at this, this has been confusing me (and making me nervous about some kinds of changes)!

I added some comments (mostly around Windows paths, and mostly the same thing just in a few places).

I'm not all that familiar with all this code, so you may want another reviewer too!

DanTup:
btw I do see some failures on the e2e tests on Windows with these changes, looking into them..

abeatrix:
> @abeatrix I just tried it with `.cody/ignore` (and with ` "cody.internal.unstable": true`). Ignores are properly respected for remote embeddings, local embeddings, and symf AFAICT. I believe the latter 2 (local embeddings and symf) were working (perhaps the problem was the file was named `.ignore`?), and I fixed ignores for remote embeddings.

sorry that was my bad! I've confirmed the .ignore file is working and the embedding file path issue is resolved in the latest commit:

image

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
abeatrix: If we are logging the codebase name (repoName?) in ContextfFile, maybe we can create a map with that for all the codebase with their workspace to construct the correct path?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1292,6 +1282,12 @@ class ContextProvider implements IContextProvider {
...
             return []
         }
+
+        const workspaceFolder = vscode.workspace.workspaceFolders?.at(0)
+        if (!workspaceFolder) {
+            return []
+        }
+
         logDebug('SimpleChatPanelProvider', 'getEnhancedContext > searching local embeddings')
         const contextItems: ContextItem[] = []
```
abeatrix, line 1292: QQ: Should we merge ContextItem with ContextFile? 
Not in this PR if yes, just wondering if they are the same :)

lib/shared/src/chat/transcript/index.ts
sqs: Yes, as of now it is. Previously `fileName` sometimes also contained the symbol name.

lib/shared/src/chat/transcript/transcript.test.ts
sqs: URIs always use forward slashes, even on Windows (right??). I have pushed all displaying of FS paths to the edge in `displayPath`, which uses your code to render the paths correctly.

lib/shared/src/codebase-context/index.ts
sqs: I manually checked (local embeddings, remote embeddings, symf, at-mentioned files) and found no way in the code or product to get a relative path here. There is a small chance I missed something, but I am confident enough for myself at least. :)

lib/shared/src/test/testSetup.ts
sqs: This is just for tests. There are some tests that manually override this (display-text.test.ts) to another displayPath func to test Windows paths specifically.

slack/src/slack/message-interaction.ts
sqs: I removed the slack code in https://github.com/sourcegraph/cody/pull/2705 anyway BTW.

vscode/src/commands/prompt/display-text.test.ts
sqs: This is a URI path, so it would not have the drive letter, I believe.

See https://code.visualstudio.com/api/references/vscode-api#Uri.fsPath:

```
const u = URI.parse('file://server/c$/folder/file.txt');
u.authority === 'server';
u.path === '/shares/c$/file.txt';
u.fsPath === '\\serverc$\folder\file.txt';
```

vscode/src/editor-context/helpers.ts
sqs: See https://code.visualstudio.com/api/references/vscode-api#Uri.fsPath:

```
const u = URI.parse('file://server/c$/folder/file.txt');
u.authority === 'server';
u.path === '/shares/c$/file.txt';
u.fsPath === '\\serverc$\folder\file.txt';
```

sqs:
@DanTup Thanks again for the review. I'm going to add some Windows tests for the displayPath impl that runs in VS Code.

sqs:
@DanTup OK, check the latest commit I just pushed.

vscode/src/editor-context/helpers.ts
sqs: Turns out this code is unused. :) I'll delete it. You are right, though!

lib/shared/src/test/testSetup.ts
sqs: OK, I improved the tests here to not be OS-dependent.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
sqs: I think that we should immediately make paths from the remote server/service absolute, and *never* let ambiguous relative paths leak.

I'll do that in a followup PR.

lib/shared/src/chat/transcript/transcript.test.ts
sqs: Nevermind, `URI.file` does some magic on Windows. Yeah, let me see.

sqs: Done, added a new testFileUri, deleted the other similar helper funcs, and made all `URI.file` and `Uri.file` in tests use it.

sqs:
@abeatrix @DanTup OK, did another round of edits to address the feedback. PTAL! Thank you.

DanTup (approved):
LGTM (with the previous caveats about being less familiar with some of this code).

I'll double-check the e2e tests on Windows shortly (I'd really like to get https://github.com/sourcegraph/cody/pull/1447 landed soon so these run on the bots!)

DanTup:
I ran the e2e tests on my PC and got this one failure (while I was afk):

```
  1) at-file-context-selecting.test.ts:12:5 › @-file empty state ───────────────────────────────────

    Error: Timed out 5000ms waiting for expect(locator).toHaveValue(expected)

    Locator: frameLocator('iframe.webview').last().frameLocator('iframe').getByRole('textbox', { name: 'Chat message' })
    Expected string: "@Main.java and @Main.java "
    Received string: "@Main.java and @Main.ja"
    Call log:
      - expect.toHaveValue with timeout 5000ms
      - waiting for frameLocator('iframe.webview').last().frameLocator('iframe').getByRole('textbox', { name: 'Chat message' })
      -   locator resolved to <textarea rows="1" title="" required="" aria-label="Chat m…>@Main.java and @Main.ja</textarea>
      -   unexpected value "@Main.java and @Main.ja"
      -   locator resolved to <textarea rows="1" title="" required="" aria-label="Chat m…>@Main.java and @Main.ja</textarea>
      -   unexpected value "@Main.java and @Main.ja"
      -   locator resolved to <textarea rows="1" title="" required="" aria-label="Chat m…>@Main.java and @Main.ja</textarea>
      -   unexpected value "@Main.java and @Main.ja"
      -   locator resolved to <textarea rows="1" title="" required="" aria-label="Chat m…>@Main.java and @Main.ja</textarea>
      -   unexpected value "@Main.java and @Main.ja"
      -   locator resolved to <textarea rows="1" title="" required="" aria-label="Chat m…>@Main.java and @Main.ja</textarea>
```

However, I can't reproduce it now while I'm watching.. I've run many times and it keeps passing.

I'm gonna be running the e2e tests on the (much slower) bots when I rebase and try to land https://github.com/sourcegraph/cody/pull/1447, so I can dig more into this if I see it again when doing that.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1338,19 +1337,26 @@ class ContextProvider implements IContextProvider {
...
         const embeddings = await this.embeddingsClient.search([repoId], text, NUM_CODE_RESULTS, NUM_TEXT_RESULTS)
         if (isError(embeddings)) {
             throw new Error(`Error retrieving embeddings: ${embeddings}`)
         }
         for (const codeResult of embeddings.codeResults) {
-            const uri = remoteEmbeddingSnippetUri(codebase, codeResult)
+            // TODO(sqs): this is broken for multi-root workspaces because it assumes that the file
+            // exists in the first workspaceFolder and that the file still exists.
+            const uri = vscode.Uri.joinPath(workspaceFolder.uri, codeResult.fileName)
```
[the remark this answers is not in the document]
sqs: I’ll start fixing it tonight and will try to get it landed asap. Yeah, I think we need to fix it before the next release in 2-ish weeks.

abeatrix:
@sqs just realized I didn't submit my last review. 

This makes a lot of things easier, thank you for landing this! And @DanTup for the tests on windows!

sqs:
@abeatrix Yep, thanks for the review!

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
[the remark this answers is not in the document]
sqs: I just saw https://github.com/sourcegraph/cody/issues/2330, which makes me think that maybe it never worked correctly. Still want to get it fixed ASAP.

PR #2657, opened 2024-01-10
agent: return client-provided anonymous user ID from localStorage

Strum355:
(From https://github.com/sourcegraph/jetbrains/issues/281#issuecomment-1884680142) 

The [telemetry logger](https://sourcegraph.com/github.com/sourcegraph/cody@85355f1ca72073518ab7adeb1dcc14d14fc1449a/-/blob/vscode/src/services/telemetry.ts?L48-49) was not being created with the anonymous user ID provided by the agent client for events that were created using a singleton [globalAnonymousUserID](https://sourcegraph.com/github.com/sourcegraph/cody@nsc/global-anonymous-user-id/-/blob/vscode/src/services/telemetry.ts?L16). The recipe events dont https://sourcegraph.com/github.com/sourcegraph/cody@85355f1ca72073518ab7adeb1dcc14d14fc1449a/-/blob/agent/src/agent.ts?L327, as well as events created on the jetbrains plugin side (e.g. `CodyInstalled`), but `CodyJetBrainsPlugin:Auth:connected` and others do.

As a result, each agent startup would create a new ID and use that for most of its events instead of the one created by clients.

Disclaimer: no ☕ has been had yet, but I think I've gotten the situation correct here

## Test plan

`console.log` showed `created === true` [here](https://sourcegraph.com/github.com/sourcegraph/cody@85355f1ca72073518ab7adeb1dcc14d14fc1449a/-/blob/vscode/src/services/telemetry.ts?L48) before the change, and false after the change

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -20,6 +20,7 @@ import { type TelemetryEventParameters } from '@sourcegraph/telemetry'
 import { type ExtensionMessage, type WebviewMessage } from '../../vscode/src/chat/protocol'
 import { activate } from '../../vscode/src/extension.node'
 import { TextDocumentWithUri } from '../../vscode/src/jsonrpc/TextDocumentWithUri'
+import { localStorage } from '../../vscode/src/services/LocalStorageProvider'
 
 import { newTextEditor } from './AgentTextEditor'
 import { AgentWebPanel, AgentWebPanels } from './AgentWebPanel'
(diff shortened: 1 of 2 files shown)


dadlerj:
Thank you! 

I'm not 100% sure how the agent works. Will this update make it so that [this function](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/LocalStorageProvider.ts?L164-177) will return the value passed from the client? Or would this file have to explicitly request it using `globalState.get`?

The path to that point is:
* [The "auth:connected" event is logged here](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/ContextProvider.ts?L221)
* [Which calls `log` and then `logEvent` in the telemetry service](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/telemetry.ts?L89-112)
* [The `globalAnonymousUserID` value is set here using that function above](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/telemetry.ts?L48)

dadlerj:
If you feel confident that this will work, feel free to merge. How quickly could we get this deployed? Is that a question for the JB channel/team?

PR #2658, opened 2024-01-10
Agent: add support for client-side progress bars

olafurpg:
Fixes https://github.com/sourcegraph/jetbrains/issues/257

Previously, the agent didn't support rendering progress bars on the client. This was problematic because symf symbol uses progress bars to inform the user when symf is indexing, and allows users to cancel indexing. This PR closes this feature gap by adding four new JSON-RPC notification endpoints to the agent protocol to deal with progress bars:

* `progress/start`: server to client.
* `progress/report`: server to client.
* `progress/end`: server to client.
* `progress/end`: client to server. Only valid for cancelable progress bars.

Agent clients must add `progressBars: 'enabled'` to the client capabilities during the `initalize` handshake to receive progress-related notifications.

Additionally, to help clients with testing progress bars, there are two new testing-related JSON-RPC endpoints

* `testing/progress`. Client request to server.
* `testing/progressCancelation`. Client request to server.

This PR adds two test cases that demonstrate how progress bars work, including how to cancel a progress bar.

## Test plan

See two new test cases for progress bars.

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -2,6 +2,7 @@
 import { execSync } from 'child_process'
 import path from 'path'
 
+import * as uuid from 'uuid'
 import type * as vscode from 'vscode'
 
 // <VERY IMPORTANT - PLEASE READ>
@@ -31,6 +32,7 @@ import {
     ExtensionKind,
     FileType,
     LogLevel,
+    ProgressLocation,
     Range,
     StatusBarAlignment,
     UIKind,
@@ -414,6 +416,8 @@ export function setCreateWebviewPanel(newCreateWebviewPanel: typeof vscode.windo
     shimmedCreateWebviewPanel = newCreateWebviewPanel
 }
 
+export const progressBars = new Map<string, CancellationTokenSource>()
+
 const _window: Partial<typeof vscode.window> = {
     createTreeView: () => defaultTreeView,
     tabGroups,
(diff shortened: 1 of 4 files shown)


agent/src/vscode-shim.ts
```diff
@@ -443,13 +447,43 @@ const _window: Partial<typeof vscode.window> = {
...
+                id,
+                options: { title: options.title, cancellable: options.cancellable, location, locationViewId },
+            })
+        }
         try {
-            const result = await handler({ report: () => {} }, new CancellationTokenSource().token)
+            const result = await handler(
+                {
+                    report: ({ message, increment }) => {
+                        if (progressClient) {
```
pkukielka, line 470: Shouldn't we check somewhere if `token.isCancellationRequested`?
Otherwise how do we get out if cancelation was requested?

olafurpg: Good catch. I added a guard for `progress/report` but kept the `progress/end` just to allow the client to get an acknowledgement that the operation has ended. Would be good to have one test case for cancelation too...

pkukielka (approved):
LGTM

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -361,3 +386,48 @@ export interface DebugMessage {
...
+    /**
+     * A human-readable string which will be used to describe the
+     * operation.
+     */
+    title?: string
+    /**
+     * The location at which progress should show.
+     * Either `location` or `locationViewId` must be set
+     */
+    location?: string // one of: 'SourceControl' | 'Window' | 'Notification'
```
dominiccooney, line 419: TypeScript can model types like this (literally what you have after 'one of:'...) why not represent it accurately?

olafurpg: I tried doing that but the type of `ProgressLocation[options.location]` returns a `string` instead of the literal types and I didn't spend much time trying to get the narrow type inferred (I'm guessing something like `as const` could help). The literal types won't help much in this scenario anyways, most of the clients are written in languages without literal types so they will type this as a string anyways.

PR #2662, opened 2024-01-10
Autocomplete: Improve jaccard similiarty retriever

philipp-spiess:
This PR improves the existing jaccard similiarty retriever and adds support for:

- Returning more then one snippet for a given file. This ensures that we prioritize higher ranked snippets over getting some snippets of every open file
- Returning snippets that existing in the source document. If a snippet ranks highly that's inside the same document that you're currently working (but not part of the prefix/suffix), we should definitely include it.
- Better handling empty lines (so as to avoid creating matches that start with empty lines)

## Test plan

- Please take a look at the added and updated unit tests. These test all of the changes in isolation
- I also just ran the extension for a bit and it seems to still work. Hard to say how much better or worse this is.


vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
```diff
@@ -3,121 +3,103 @@ import winkUtils from 'wink-nlp-utils'
...
             bothWords,
-            wordsForEachLine[i + windowSize]
+            wordsForEachLine[i + windowSize - 1]
         )
         windowCount += windowIncrease
         bothCount += intersectionIncrease
 
-        // We only do a Jaccard similarity every 5 lines to avoid excessive CPU burn. Since the
-        // windows are 50 lines, this should still give good results.
-        if (i % 5 === 0) {
```
philipp-spiess, line 107: This behaviour didn't really make sense. My intention of adding it was to save CPU cycles but the `jaccardDistance` check is just comparing already computed numbers so it's not expensive at all. I think it's less confusing if we remove this.

valerybugakov:
> That means we'd have two implementations for the jaccard similarity retriever for a bit

This code is not frequently updated, so I'm with you on having two similar implementations to measure the impact of these changes 👍 . It should not cause a lot of overhead where we need to update both implementations all the time.

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
```diff
@@ -3,121 +3,103 @@ import winkUtils from 'wink-nlp-utils'
...
-        // Subtract the words from the line we are scrolling past
-        windowCount += subtract(windowWords, wordsForEachLine[i])
-        bothCount += subtract(bothWords, wordsForEachLine[i])
+    // Slide over the target text line by line
+    //
+    // We start at 1 since we already calculated the first match at startLine 0,
+    // this way, i can refer to the startLine of the current window
+    for (let i = 1; i <= lines.length - windowSize; i++) {
+        // Subtract the words from the line we are scrolling away from
+        windowCount += subtract(windowWords, wordsForEachLine[i - 1])
```
valerybugakov, line 69: I did not expect `subtract` to modify its arguments in place until I read its sources. Having a comment or a function name indicating that would be helpful.

UPD: I see that this was the behavior prior to this PR

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
```diff
@@ -3,121 +3,103 @@ import winkUtils from 'wink-nlp-utils'
...
 
-    let windowCount = wordCount(windowWords)
+    let windowCount = sumWordCounts(windowWords)
     // Initialize the bag of words for the intersection of the match window and targetText
     const bothWords = new Map<string, number>()
     for (const [word, wordCount] of targetWords.entries()) {
         bothWords.set(word, Math.min(wordCount, windowWords.get(word) || 0))
     }
-    let bothCount = wordCount(bothWords)
+    let bothCount = sumWordCounts(bothWords)
```
valerybugakov, line 51: You didn't introduce this variable in this PR, but still, it's name is not descriptive, making it hard to understand the data flow. Maybe use `intersectionWordCounts` or something similar?

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
```diff
@@ -3,121 +3,103 @@ import winkUtils from 'wink-nlp-utils'
...
     content: string
+    startLine: number
+    endLine: number
 }
 
+type WordOccurrences = Map<string, number>
+
 /**
  * Finds the window from matchText with the lowest Jaccard distance from targetText.
  * The Jaccard distance is the ratio of intersection over union, using a bag-of-words-with-count as
```
valerybugakov, line 14: Is this actually Jaccard similarity and not Jaccard distance? https://en.wikipedia.org/wiki/Jaccard_index

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
```diff
@@ -3,121 +3,103 @@ import winkUtils from 'wink-nlp-utils'
...
 
-    for (let i = 0; i < wordsForEachLine.length - windowSize; i++) {
-        // Subtract the words from the line we are scrolling past
-        windowCount += subtract(windowWords, wordsForEachLine[i])
-        bothCount += subtract(bothWords, wordsForEachLine[i])
+    // Slide over the target text line by line
+    //
+    // We start at 1 since we already calculated the first match at startLine 0,
+    // this way, i can refer to the startLine of the current window
+    for (let i = 1; i <= lines.length - windowSize; i++) {
```
valerybugakov, line 67: Could it be that `windowSize` is bigger than `lines.length`?

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
```diff
@@ -3,121 +3,103 @@ import winkUtils from 'wink-nlp-utils'
...
         }
-    }
 
-    return {
-        score: bestScore,
-        content: lines.slice(bestWindow[0], bestWindow[1]).join('\n'),
+        // compute the jaccard distance between our target text and window
+        const score = jaccardDistance(targetCount, windowCount, bothCount)
+        const startLine = i
+        const endLine = i + windowSize
```
valerybugakov, line 96: For the first window, we have `const firstWindowEnd = Math.min(windowSize, lines.length)`, and here `const endLine = i + windowSize`, which results in a smaller window size because `endLine` is exclusive in `lines.slice(startLine, endLine)`. Should we `+ 1` here?

vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
```diff
@@ -17,35 +18,107 @@ import { VSCodeDocumentHistory, type DocumentHistory } from './history'
...
+        const contextLineRange = { start: contextRange.start.line, end: contextRange.end.line }
+
         const matches: JaccardMatchWithFilename[] = []
         for (const { uri, contents } of files) {
-            const match = bestJaccardMatch(targetText, contents, SNIPPET_WINDOW_SIZE)
-            if (!match || abortSignal?.aborted || isCodyIgnoredFile(uri)) {
+            if (isCodyIgnoredFile(uri) || abortSignal?.aborted) {
                 continue
             }
+            const lines = contents.split('\n')
```
valerybugakov, line 62: Should we use the `lines()` helper to support `\r\n` line endings?

vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
```diff
@@ -17,35 +18,107 @@ import { VSCodeDocumentHistory, type DocumentHistory } from './history'
...
+            // Use relative path to remove redundant information from the prompts and
+            // keep in sync with embeddings search results which use relative to repo root paths
+            const readableFileName = path.normalize(vscode.workspace.asRelativePath(uri.fsPath))
+
+            // TODO: Cluster matches by score. For now we assume that every match that is returned
+            // is of equal importance to the user (we truncate the list by maxMatchesPerFile to
+            // avoid this being too many results), but ideally we can create clusters so that merged
+            // sections do not become too big
+
+            // Merge overlapping ranges
```
valerybugakov, line 74: Thoughts on extracting this logic into a separate function?

vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
```diff
@@ -17,35 +18,107 @@ import { VSCodeDocumentHistory, type DocumentHistory } from './history'
...
+            const fileMatches = bestJaccardMatches(targetText, contents, this.snippetWindowSize, this.maxMatchesPerFile)
+
+            // Use relative path to remove redundant information from the prompts and
+            // keep in sync with embeddings search results which use relative to repo root paths
+            const readableFileName = path.normalize(vscode.workspace.asRelativePath(uri.fsPath))
+
+            // TODO: Cluster matches by score. For now we assume that every match that is returned
+            // is of equal importance to the user (we truncate the list by maxMatchesPerFile to
+            // avoid this being too many results), but ideally we can create clusters so that merged
+            // sections do not become too big
```
valerybugakov, line 72: Let's do that 👍

philipp-spiess: I was thinking this is more or less a todo for later, since the clustering is probably not trivial to implement and the impact is questionable

vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
philipp-spiess: The `\r\n` behavior is only needed when you need to do something on the line (so that a line with `foo\r` is not expected). In this case we split solely so we know 

- the number of lines
- be able to `join('\n')` again

So this should be fine and in fact even help us to preserve `\r\n` when joining again.

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
philipp-spiess: Good catch will add a test

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
philipp-spiess: I found out where the discrepancy came from, if startLine is 0 and we only have one line, I'd expect endLine to also be 0 (since the range only spans from line 1 to line 1). Fixing

vscode/src/completions/context/retrievers/jaccard-similarity/bestJaccardMatch.ts
philipp-spiess: Should be fine though because `1` (our start value) is not smaller than a negative number so this won't run?

valerybugakov (approved):
Preemptive approval since we discussed everything in the comments ✅
Looking forward for A/B testing this!

philipp-spiess:
@valerybugakov Yay! I need to get the language specifics right now, need to understand the differences between similarity and distance 😰

Issue #2663, opened 2024-01-10
bug: extension stuck in a startup for users without email


abeatrix:
### Version

1.0.5

### Describe the bug

Reported by Noah on Slack:

A user reported Cody is stuck in startup loading screen. Tried reinstall the extension but that didn't work.

Based on the log they shared, it looks like it was caused by changes introduced in https://github.com/sourcegraph/cody/pull/2233, where we create auth status with primary email when a lot of our enterprise users do not have that set up.

```sh
console.ts:137 [Extension Host] TypeError: Cannot read properties of null (reading 'email')
    at AuthProvider.makeAuthStatus
```
### Steps to reproduce

1. Log into Cody using an enterprise soucegraph (e.g. S2) token associated with a user that doesn't have emails set up for their account.
2. Try reloading VS Code

https://github.com/sourcegraph/cody/assets/68532117/85ad63ef-c422-45a4-b4f8-640d5e11ecf3

### Expected behavior

Primary email should not be a requirement for sourcegraph users to load the extension.

Issue #2664, opened 2024-01-10
bug: signing out of Cody does not close chat panels


abeatrix:
### Version

1.0.5

### Describe the bug

Chat panels remain opened in editor even thought the user has signed out:

https://github.com/sourcegraph/cody/assets/68532117/ea5dcfc7-bf94-4d6e-a5d3-19fb6128e5a8

### Expected behavior

All chat sessions should be deactivated and all chat panels should be closed on log out

PR #2665, opened 2024-01-10
fix: handle missing primaryEmail for enterprise users

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2663 & https://github.com/sourcegraph/cody/issues/2664

A bug was introduced in https://github.com/sourcegraph/cody/pull/2233 where we added a requirement for `primaryEmail` during auth, resulting in uncaught error that stopped the extension from starting. 

For Enterprise instances, emails are not always required to set up an account. 

This PR includes the following changes to fix the issue above:
- update authStatus and graphQL query to reflect that `primaryEmail` and `displayName` fields are optional
- add tests for the above change
- use username instead of displayName for enterprise instances as username is a required field
- replace replace getEnterpriseCurrentUserInfo and getDotComCurrentUserInfo with the new getCurrentUserInfo
- update agent tests with the changes above

## Test plan

1. Log into Cody using an enterprise soucegraph (e.g. S2) token associated with a user that doesn't have emails set up for their account.
2. Try reloading VS Code

### Before

Unable to start Cody on reload

https://github.com/sourcegraph/cody/assets/68532117/85ad63ef-c422-45a4-b4f8-640d5e11ecf3

### After

https://www.loom.com/share/b9049f54cb0f48da87fc630c84cc736e?sid=0e1b3677-9c3e-4c28-84b5-846f95ad4a16

https://github.com/sourcegraph/cody/assets/68532117/dd2a54f9-dfdf-4ff2-b789-a032dc6b3b1d


kalanchan (approved):
ggogoog!

PR #2667, opened 2024-01-10
fix: remove codyProEnabled field from GraphQL API call

abeatrix:
RE: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1704927255975299

Fix issues where enterprise users are unable to login due when on an older version of Sourcegraph.

Remove codyProEnabled field from GraphQL API call as it's used for both enterprise and dot com users, which does not exist in older enterprise instances.

## Test plan

updating tests

### DIFF
--- a/vscode/src/services/AuthProvider.ts
+++ b/vscode/src/services/AuthProvider.ts
@@ -271,7 +271,6 @@ export class AuthProvider {
             }
             return { ...unauthenticatedStatus, endpoint }
         }
-
         const userCanUpgrade =
             isDotCom &&
             'codyProEnabled' in proStatus &&
--- a/vscode/src/services/AuthProvider.ts
+++ b/vscode/src/services/AuthProvider.ts
@@ -262,6 +262,7 @@ export class AuthProvider {
         }
 
         const isCodyEnabled = true
+        const proStatus = await this.client.getCurrentUserCodyProEnabled()
 
         // check first if it's a network error
         if (isError(userInfo)) {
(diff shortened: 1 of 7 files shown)


vscode/src/services/AuthProvider.ts
```diff
@@ -262,6 +262,7 @@ export class AuthProvider {
         }
 
         const isCodyEnabled = true
+        const proStatus = await this.client.getCurrentUserCodyProEnabled()
```
chwarwick, line 265: So am I following this right in the enterprise `proStatus = Error('current user not found')` which makes the userCanUpgrade false?

chwarwick: but it never gets that far because isDotCom is false anyway.

chwarwick (approved):
This looks alright to me, I'm not familiar with this area so maybe an comment on the pro check about potentially failing on enterprise users would be helpful.

vscode/src/services/AuthProvider.ts
abeatrix: Yea this is only called below when isDotCom is true

kalanchan:
https://github.com/sourcegraph/cody/assets/51868853/3c9b6d69-4e87-4811-a8a9-52edea3fc1a1

looks good, I logged into `demo.sourcegraph.com` fine and it didn't seem to affect me when I logged into dotcom (as plg user)

kalanchan (approved):

PR #2670, opened 2024-01-11
Fixing Ollama model names

arafatkatze:
Minor fix to change model names for ollama

## Test plan

Run the model locally and it works


arafatkatze:
there are different codellama versions in the ollama registry Please pull

`ollama pull codellama:7b-code`

PR #2673, opened 2024-01-11
VSCode, Clean up: Remove classic context status indicator

dominiccooney:
Remove the classic context status indicator. This has been replaced by the enhanced context settings widget. Also clean up some app-less onboarding properties.

## Test plan

CI, specifically

```
pnpm test:integration
pnpm test:e2e
```


abeatrix (approved):
LGTM! Confirmed enhance context is still working properly.

image

Do we still need https://github.com/sourcegraph/cody/blob/main/vscode/src/services/LocalAppDetector.ts at this point?

dominiccooney:
> QQ: Do we still need https://github.com/sourcegraph/cody/blob/main/vscode/src/services/LocalAppDetector.ts at this point?

It's not used right now, but the last bit of #1966 is to toast users to remove it, so I want to use it for that. We can delete it and I can pull it back out of history though if we want.

Issue #2669, opened 2024-01-11
bug: Cody "steals" cursor from files when alt+tab-ing back to VSCode


morgangauth:
### Version

v1.1.0 (not present in v1.0.5)

### Describe the bug

When editing a file side-by-side with Cody, after using alt+tab to navigate to another application (e.g. Chrome) and back to VSCode, Cody will "steal" the cursor after a second.

Steps to reproduce:
1. Open a file in VSCode side-by-side with Cody chat in v1.1.0
2. Make sure the focus/cursor are in the file
3. Use alt+tab to navigate to another application (e.g. Chrome)
4. Use alt+tab to navigate back to VSCode
5. Verify the cursor and focus are on the file
6. Begin typing
7. The typing will switch to the Cody chat window rather than the file as expected

### Expected behavior

In Step 7 above, the focus/cursor should remain in the file, as it was before navigating away from VSCode using alt+tab. 

https://github.com/sourcegraph/cody/assets/113058716/ee61c27c-ef08-4ffb-84aa-80de5a79e81e

PR #2674, opened 2024-01-11
Fixing Steal Cursor Issue

arafatkatze:
Issue: https://github.com/sourcegraph/cody/issues/2669

The current build of Cody has an issue where the cursor is stolen many times because of a rerendering of some sort for auth, this makes cody very hard to use for anything as you can barely type.

Original PR that introduced this https://github.com/sourcegraph/cody/pull/2665

I actually do not know how the PR works so I added a throttle mechanism to stop the rerendering for 5 minutes(300 seconds) so that it doesn't make a super terrible UX. 

My aim is to merge this HACK for now and then do a proper fix later with @abeatrix

### DIFF
--- a/vscode/webviews/Chat.tsx
+++ b/vscode/webviews/Chat.tsx
@@ -274,7 +274,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
         if (autoFocus) {
             inputRef.current?.focus()
         }
-    }, [autoFocus, value, chatModels])
+    }, [autoFocus, value])
 
     // Focus the textarea when the webview gains focus (unless there is text selected). This makes
     // it so that the user can immediately start typing to Cody after invoking `Cody: Focus on Chat


sqs (changes requested):
I believe this is the problem. Apply this diff and try it out. Fixes it for me.

```diff
diff --git a/vscode/webviews/Chat.tsx b/vscode/webviews/Chat.tsx
index d141704d..32f46ab0 100644
--- a/vscode/webviews/Chat.tsx
+++ b/vscode/webviews/Chat.tsx
@@ -275,7 +275,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
         if (autoFocus) {
             inputRef.current?.focus()
         }
-    }, [autoFocus, value, chatModels])
+    }, [autoFocus, value])
 
     // Focus the textarea when the webview gains focus (unless there is text selected). This makes
     // it so that the user can immediately start typing to Cody after invoking `Cody: Focus on Chat
```

This makes it so that when you select a new chat model, it does not autofocus the input right after that, but that's fine and we can fix it later.

Another fix is to memoize the chatModels value.

sqs (approved):

PR #2675, opened 2024-01-11
send transcript data as `privateMetadata` for v2 telemetry only

akalia25:
updates telemetry to only send transcript data on v2 telemetry and removes sending it for v1. This part of a bigger pipeline to isolate these transcripts in their own gcs bucket and subscriptions

### DIFF
--- a/vscode/src/chat/MessageProvider.ts
+++ b/vscode/src/chat/MessageProvider.ts
@@ -236,7 +236,7 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                 const responseText = this.isDotComUser ? text : undefined
                 telemetryService.log(
                     'CodyVSCodeExtension:chatResponse:hasCode',
-                    { ...codeCount, ...metadata, requestID, responseText },
+                    { ...codeCount, ...metadata, requestID },
                     { hasV2Event: true }
                 )
 
@@ -247,6 +247,10 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                         {
                             metadata: {
                                 ...codeCount,
+                            },
+                            privateMetadata: {
+                                ...metadata,
+                                responseText,
                             },
                         }
                     )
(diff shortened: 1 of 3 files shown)


vscode/src/chat/MessageProvider.ts
```diff
@@ -247,6 +247,12 @@ export abstract class MessageProvider extends MessageHandler implements vscode.D
                         {
                             metadata: {
                                 ...codeCount,
+                                recordsPrivateMetadataTranscript: 1,
```
bobheadxi, line 250: Should we add a docstring on what this does? Maybe we can link to https://github.com/sourcegraph/sourcegraph/pull/59524 for now

bobheadxi: Also, below, it looks like the `responseText` is conditional - should `recordsPrivateMetadataTranscript` be conditional too? Otherwise for enterprise instances `recordsPrivateMetadataTranscript` will be true, while no private metadata will be available

akalia25: Yes! Good idea, I'll update this.

bobheadxi: Also, aside - on enterprise instances, `privateMetadata` will not be exported anyway (there is currently nothing on the `sensitivemetadataallowlist`). We _could_ choose to just always record the transcript.

The potential risk here is data pipelines that pull directly from `event_logs` in Cloud instances, but as we phase that out, maybe no longer a concern?

bobheadxi: Sounds good to me :) Let's include that in the docstring here too

akalia25: > Let's include that in the docstring here too
Should I add this here (`node_modules/.pnpm/@sourcegraph+telemetry@0.16.0/node_modules/@sourcegraph/telemetry/dist/index.d.ts) so that it is a part of the `privateMetadata` definition?

bobheadxi: This is probably specific to Cody for now, so let's just document it here

sqs:
@akalia25 you blocked on getting this reviewed?

bobheadxi (approved):
Approving as it looks good to me, but I think it would be beneficial for someone from the Cody teams to review as well so we're all aligned on direction the analytics team wants to take with regards to recording this kind of data.

akalia25:
Thanks for flagging @sqs! Yes, as Robert mentioned, I would love to get someone from the Cody team to review!

abeatrix (approved):
Lgtm, thank you @akalia25

PR #2677, opened 2024-01-11
update cody pro status when window refocuses

abeatrix:
The onDidChangeWindowState was introduced previously to check if the user has left the window to upgrade their account to Cody Pro in their browser when the VS Code window refocuses. 

In v1.1.0, we've updated to use `reloadAuthStatus()` to update the auth status, that caused logging this reload event as login event.  This PR replaced that with `syncAuthStatus(authStatus)` to update the `authStatus.userCanUpgrade` only on windowState change.

This will update the auth status with a userCanUpgrade flag to indicate if the user is eligible to upgrade.

## Test plan

- Check the log to see if login events are being logged on window state change.
- Login event should only be logged once after sign in

image

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -425,6 +425,20 @@ const register = async (
         }),
         vscode.commands.registerCommand('agent.auth.reload', async () => {
             await authProvider.reloadAuthStatus()
+        }),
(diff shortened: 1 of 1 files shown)


arafatkatze (approved):

arafatkatze:
Confirmed that this PR is Jolly

PR #2678, opened 2024-01-11
hydrate values passed via postMessage

sqs:
There were some places in the code where either (1) we were manually re-hydrating `vscode-uri.URI`/`vscode.Uri` instances or (2) there were subtle bugs that occurred because we were not doing that (the most common of which is broken uniquifying because `unhydratedUriValue.toString()` == `[Object object]`).

## Test plan

- CI
- Run a NLS search and ensure that clicking on files works (for the `uriJSON` change)
- In chat, `@`-mention some files, ensure the chat response comes through, and ensure that the EnhancedContext files are clickable


dominiccooney (approved):
Looks good

PR #2679, opened 2024-01-11
show context files explicitly included (@-mentioned) by user

sqs:
It is confusing if these are not shown because the user thinks that the file they *most* wanted is not included in Cody's context.

Discussion: https://sourcegraph.slack.com/archives/C05AGQYD528/p1704952995626909

## Test plan

- `@`-mention a file and ensure it's mentioned in the EnhancedContext


dominiccooney (approved):
👍

Issue #2628, opened 2024-01-09
PRD: Multi-repository support for Enterprise users


toolmantim:
### Description:

As a user I want to add additional repositories that will be used as context in a Cody Chat in order to increase the quality of the chat responses (less hallucinations, higher chance of being able to answer, higher quality code recommendations)

### Objectives:

Enterprise user can chat with multiple repositories as context

### Acceptance Criteria:

- [ ] Can add and remove up to 10 additional repositories (fuzzy searching to find them)
- [ ] Show's inferred repository with tooltip
- [ ] Uses the additional repositories as context when Enhanced Context is enabled
- [ ]  When a user starts a new chat from an existing chat, the repository configuration is copied across to the new chat
- [ ] There is a "recently selected" group of repositories shown at the top of the quickpick (not shown in Figma), so you can easily keep re-selecting a common set of repositories in a new chat if the new chat was created not from an existing chat

Out of scope:

- Persisting the additional repositories between chats. Instead people will use the "recently selected" to choose the same ones as before if needed

### Design Tasks:

```[tasklist]
### Design Tasks
- [x] v1 design for comment from Eng and Prod - [Loom](https://www.loom.com/share/b96a8acfaabf47699e33c275297cecfd) - [Figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT++?type=design&node-id=1218-38734&mode=design&t=uzPYwXsMerPJlMSX-4)
```

### Engineering Tasks:

```[tasklist]
### Engineering Tasks
- [ ] https://github.com/sourcegraph/cody/issues/2622
- [ ] https://github.com/sourcegraph/cody/issues/2623
- [ ] https://github.com/sourcegraph/cody/issues/2624
- [x] (WIP) Pre-populate repo picker with currently selected repos
- [x] (WIP) Automatically populate the workspace repo(s)
- [x] (WIP) Display the selected repos in the enhanced context selector
- [x] (WIP) Save the set of selected repos in the chat transcript
- [x] (WIP) Search the set of selected repos
- [ ] (WIP) Add a way to `(refresh)` button to the repo picker which refreshes the repo list
- [x] (WIP) Change the default context from "embeddings" to "blended"
- [x] (WIP) Wire up "remove" button to removing repos from context source
- [ ] https://github.com/sourcegraph/cody/issues/2850
- [x] (WIP) Limit to enterprise users
[...]

PR #2680, opened 2024-01-11
VSCode: Remove close button from Enhanced Context Settings popup

dominiccooney:
#2628 will use &times; in the enhanced context settings popup for removing a repository, so we need to remove the close button. This removes the close button. Now, the popup is closed by:

- Clicking outside the popup.
- Pressing Escape.

In addition, this takes care that keyboard focus goes to the enable/disable checkbox when the popup opens, and returns to the enhanced context button when the popup closes.

## Test plan

```
pnpm test:e2e -- context-settings.test.ts
```

Manual test:

1. Sign in and start a chat.
2. Open the enhanced context popup.
3. Press Space a few times, this should toggle the checkbox.
4. Press Escape, this should close the popup.
5. Press Space again, this should re-open the popup.
6. Click somewhere else on the chat. This should close the popup.

### DIFF
--- a/vscode/test/e2e/context-settings.test.ts
+++ b/vscode/test/e2e/context-settings.test.ts
@@ -9,20 +9,20 @@ test('enhanced context selector is keyboard accessible', async ({ page, sidebar 
     const contextSettingsButton = chatFrame.getByTitle('Configure Enhanced Context')
     await contextSettingsButton.focus()
 
-    page.keyboard.press('Space')
+    await page.keyboard.press('Space')
     // Opening the enhanced context settings should focus the checkbox for toggling it.
     const enhancedContextCheckbox = chatFrame.locator('#enhanced-context-checkbox')
     await expect(enhancedContextCheckbox.and(page.locator(':focus'))).toBeVisible()
 
     // Enhanced context should be enabled by default.
     await expect(enhancedContextCheckbox).toBeChecked()
-    page.keyboard.press('Space')
+    await page.keyboard.press('Space')
     // The keyboard should toggle the checkbox, but not dismiss the popup.
     await expect(enhancedContextCheckbox).not.toBeChecked()
     await expect(enhancedContextCheckbox).toBeVisible()
 
     // The popup should be dismiss-able with the keyboard.
-    page.keyboard.press('Escape')
+    await page.keyboard.press('Escape')
     // Closing the enhanced context settings should close the dialog...
     await expect(enhancedContextCheckbox).not.toBeVisible()
     // ... and focus the button which re-opens it.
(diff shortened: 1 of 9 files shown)


toolmantim:
😍

vscode/test/e2e/context-settings.test.ts
```diff
@@ -0,0 +1,30 @@
+import { expect } from '@playwright/test'
+
+import { sidebarSignin } from './common'
+import { newChat, test } from './helpers'
+
+test('enhanced context selector is keyboard accessible', async ({ page, sidebar }) => {
```
abeatrix, line 6: ❤️

vscode/webviews/Popups/Popup.tsx
```diff
@@ -25,28 +42,35 @@ interface PopupProps extends Omit<PopupFrameProps, 'classNames'>, PopupOpenProps
...
+    const handleKeyUp = (e: React.KeyboardEvent<HTMLDialogElement>): void => {
+        if (e.key === 'Escape') {
+            e.stopPropagation()
+            onDismiss()
+        }
+    }
+    return (
+        isOpen && (
+            <>
+                <dialog
```
abeatrix, line 58: TIL about the `<dialog>` html element

vscode/webviews/Popups/Popup.tsx
```diff
@@ -8,6 +8,23 @@ export interface PopupOpenProps {
     onDismiss: () => void
 }
 
+interface BackdropProps {
+    dismiss: () => void
+}
+
+export const Backdrop: React.FunctionComponent<React.PropsWithoutRef<BackdropProps>> = ({ dismiss }) => {
+    const handleKeyUp = (e: React.KeyboardEvent<HTMLDivElement>): void => {
+        if (e.key === 'Escape') {
```
abeatrix, line 17: Should we refocus the input box again after closing it with escape key?

abeatrix (approved):
Confirmed it works as described and the button still works as expected :D!

https://github.com/sourcegraph/cody/assets/68532117/9536f225-2445-4d79-ae8b-e472837dab10

https://github.com/sourcegraph/cody/assets/68532117/08eb66d3-c7dc-4769-aae0-a71a42adcbda

vscode/webviews/Popups/Popup.tsx
dominiccooney: I was wondering the same thing. I think that is more useful. Only downside is it migrates the focus away from where it was. WDYT @toolmantim ?

I am going to land this as-is to keep working but I want to follow up and get this exactly right.

vscode/webviews/Popups/Popup.tsx
dominiccooney: I'm using this here because the lints are nagging me to use an interactive element. DIALOG isn't particularly buying us much here. But if you ever need modal dialogs it is super useful.

This could be modal and we'd get the backdrop for free, make the rest of the elements inert for free, etc. but then positioning it is a bit of a mess.

PR #2682, opened 2024-01-11
Autocomplete: insert missing brackets before completion truncation

### LINKED ISSUE #2387: bug: multiline completions are truncated too early if the parse tree is broken because of the missing brackets
### Version

v1.1.1702572663

### Describe the bug

Consider the following completion:

```ts
function quickSort(█array: any[]): any[] {\n    if (array.length < 2) {\n        return array\n    }\n    const pivot = array.pop()\n    const left = []\n    const right = []\n    for (const element of array) {\n        if (element < pivot}
```

It breaks the tree-sitter parse tree even though we artificially insert the closing curly brace before parsing it. It means that we cut-off the completion too early:

```
function quickSort(█array: any[])
```

### Expected behavior

We artificially add all the missing remaining brackets to fix the issue.

### Additional context

_No response_

valerybugakov:
## Context

- Improves the stability of multiline completion truncation during streaming by gracefully handling missing brackets in incomplete code segments.
- Closes #2387 

## Test plan

Added new unit tests.

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/multiline-truncation.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/multiline-truncation.test.ts
@@ -682,7 +682,7 @@ cases.forEach(isTreeSitterEnabled => {
                 `)
                 })
 
-                it('handles missing brackes gracefully to truncate the completion correctly', async () => {
+                it('handles missing brackets gracefully to truncate the completion correctly', async () => {
                     const requestParams = params('console.log(1); const █', [completion``], {
                         onNetworkRequest(_params, onPartialResponse) {
                             onPartialResponse?.(completion`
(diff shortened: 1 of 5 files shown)


vscode/src/completions/get-inline-completions-tests/multiline-truncation.test.ts
```diff
@@ -681,6 +681,73 @@ cases.forEach(isTreeSitterEnabled => {
...
+                                ├MyCoolObject = {
+                                constructor() {
+                            ┤`)
+                            onPartialResponse?.(completion`
+                                ├MyCoolObject = {
+                                constructor() {
+                                    console.log(1)
+
+                                    if (false
+                            `)
```
philipp-spiess, line 698: This does not look like a realistic partial response. I.e if the line `if (false` is only partially ready, where is there a newline afterwards?

Should this be:

```suggestion
                            onPartialResponse?.(completion`
                                ├MyCoolObject = {
                                constructor() {
                                    console.log(1)

                                    if (false`)
```

instead?

philipp-spiess (approved):
The code makes sense but I think the integration test would not happen as such in the real world.

vscode/src/completions/get-inline-completions-tests/multiline-truncation.test.ts
valerybugakov: Yep, good catch!  There should be no newline at the end.

PR #2683, opened 2024-01-11
Agent: use `temperature: 0` in HTTP recordings

olafurpg:
Previously, running `pnpm update-agent-recordings` more or less always
made the agent tests fail because the LLM response had minor changes.
This PR sets the LLM temperature to zero when we run agent tsets, which
should help make the tests be a bit more stable when updating
recordings. The tests are still non-deterministic with this change based
on local testing.

## Test plan

Green CI


philipp-spiess (approved):
Nice!

PR #2692, opened 2024-01-12
remove unused codebase-specific ignores

sqs:
The `isCodyIgnoredFilePath` API for checking ignores is not actually used. Only `isCodyIgnoredFile` is used.

See discussion: https://sourcegraph.slack.com/archives/C05AGQYD528/p1705004134538609?thread_ts=1704886845.816059&cid=C05AGQYD528.

## Test plan

CI and manually ensured that the removed function was not actually being called for local embeddings, remote embeddings, and symf context (using a debugger).

### DIFF
--- a/lib/shared/src/chat/transcript/interaction.ts
+++ b/lib/shared/src/chat/transcript/interaction.ts
@@ -1,5 +1,5 @@
 import { type ContextFile, type ContextMessage, type PreciseContext } from '../../codebase-context/messages'
-import { isCodyIgnoredFile, isCodyIgnoredFilePath } from '../context-filter'
+import { isCodyIgnoredFile } from '../context-filter'
 
 import { type ChatMessage, type ChatMetadata, type InteractionMessage } from './messages'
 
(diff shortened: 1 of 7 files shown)


lib/shared/src/chat/transcript/interaction.ts
```diff
@@ -41,18 +41,10 @@ export class Interaction {
             const message = contextMessages[i]
             // Skips the assistant message if the human message is ignored
             if (message.speaker === 'human' && message.file) {
-                const { uri, repoName, fileName, source } = message.file
-                if (uri && isCodyIgnoredFile(uri)) {
+                if (message.file.uri && isCodyIgnoredFile(message.file.uri)) {
```
DanTup, line 44: Is it certain that `uri` here is _always_ available on embeddings? The removed code below only used `fileName`, so if we ended up with embeddings having `fileName` set but not `uri` (the field does appear nullable) then I think this would result in them not being filtered.

I don't know a lot about the codebases to know if that's possible or likely though.

sqs: Ah, good catch. That is NOT true for current `main`, but it will be true when the related PR https://github.com/sourcegraph/cody/pull/2653 is merged (which should be imminently). I will keep an eye out in case that other PR fails to be merged or is reverted.

DanTup: heh, I made a slightly related comment in there:

https://github.com/sourcegraph/cody/pull/2653#discussion_r1450577703

abeatrix (approved):
Confirmed the ignore file is still working as expected in the following settings for me 👍 
- [x] Single Workspace
- [x] Multi-root workspace
- [x] Single Workspace with nested repository

Screenshot 2024-01-12 at 11 37 25 AM

- [x] Embeddings source is filtered correctly:

image


- [x] search context is also filtered correctly

image

lib/shared/src/chat/transcript/interaction.ts
abeatrix: I think this is ok as the `edit command` is the only one using the Interaction class, and looks like Tom is going to remove embeddings from the edit command ([slack](https://sourcegraph.slack.com/archives/C05FY86PHLM/p1705057644053639?thread_ts=1705057278.717029&cid=C05FY86PHLM)) so we can probably retire Interaction after that?

PR #2695, opened 2024-01-12
remove now-unreachable chat history keyed on primary email

sqs:
In https://github.com/sourcegraph/cody/pull/2665 we migrated the chat history key to use the user's username instead of their email address. This means that the storage would retain the chat history under the old key indefinitely.

Large storage data slows down extension host activation and every `set` operation. My storage was 1.8MB, which was causing warnings from VS Code like this:

```
[mainThreadStorage] large extension state detected (extensionId: sourcegraph.cody-ai, global: true): 916.8603515625kb. Consider to use 'storageUri' or 'globalStorageUri' to store this data on disk instead.
```

Each `Memento#set` call was taking ~37ms (with 1.8MB of data). With ~1kb of data, each set call takes ~1-3ms.

Of course, the data will accumulate as the user uses Cody more, but at least we can manage the data now without some of it being "orphaned".

## Test plan:

Confirmed that this removes the old data:

1. Added a console.log to ensure that the code had the right before/after values.
2. Confirmed the storage DB on disk had the keys before and after the migration did not have them, with `sqlite3 ~/.config/Code/User/globalStorage/state.vscdb "select value from itemtable where key='sourcegraph.cody-ai';" | jq '."cody-local-chatHistory-v2" | keys'`.
3. Confirmed the length of the key's value went down (from 1.8MB to 672 bytes, as expected).

### DIFF
--- a/vscode/src/services/LocalStorageProvider.ts
+++ b/vscode/src/services/LocalStorageProvider.ts
@@ -92,7 +92,7 @@ export class LocalStorage {
 
         // For backwards compatibility, we upgrade the local storage key from the old layout that is
         // not scoped to individual user accounts to be scoped instead.
-        if (history && !isMigratedChatHistory(history)) {
+        if (history && !isMigratedChatHistory2261(history)) {
             // HACK: We spread both parts here as TS would otherwise have issues validating the type
             //       of AccountKeyedChatHistory. This is only three fields though.
             history = {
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Looks good.

olafurpg (commented):
This is a nice cleanup. I'm working on supporting persistent chat history for JetBrains and it wasn't obvious to me when inspecting the raw storage data why some entries had an email and some had a username.

PR #2700, opened 2024-01-12
disable eslint no-explicit-any

sqs:
It will still warn on implicit use of any. This is causing a lot of our eslint warnings, and 95%+ of them are intentional, which just makes us ignore eslint warnings.

Before this, there were 153 no-explicit-any warnings. That represented more than 1/3 of our total eslint warning count.

## Test plan

CI


burmudar (approved):

olafurpg (approved):
Yes please!

sqs:
related: https://github.com/sourcegraph/cody/pull/2702

PR #2705, opened 2024-01-12
remove unused code

sqs:
- Remove `web`, `slack`, `cli`.
  - We can re-add them later, but they are not currently used, and they slow us down.
- Remove unused code, using [knip](https://knip.dev/), a dead code detector.
  - knip basically tells you which exports are unused, and it can auto-remove the `export` keyword. Then I went through each such file and deleted unused declarations (which were identified by `tsc` now that some things were no longer exported). 
  - I manually re-added some code that it identified as unused (such as the `dirname` exported helper in `@sourcegraph/cody-shared`, which could be useful in the future), but I was very sparing about this.
- I did not remove any unused code from `agent` because I am (for now) less familiar with that project and did not want to break or disrupt anything.

## Test plan

CI


sqs:
It's possible this breaks something. Risks:

- It breaks some build process. The breakage would likely be obvious (like `vsce: command not found`).
- It breaks something subtle. If so, I'd rather find it out soon with this big PR top of mind for us, rather than subtly later.

valerybugakov (approved):
Tested commands that I use locally for the VS Code development ✅

eseliger (approved):
skimmed over the diff, trusting CI here that actually used code would've been found. Thanks for -5000loc!

PR #2712, opened 2024-01-12
Edit: Fix file/symbol hint showing unnecessarily

umpox:
## Description

We defaulted `initialValue` to an empty string.

This meant if you only had typed "@" and then deleted it, we wouldn't clear out the previous quick pick item, so it'd look like this:
image

Which is confusing as "#" on its own does not trigger the UI.

## Test plan

1. Quick pick input, type just @ to trigger the file menu
2. Delete the @, check menu is hidden


abeatrix (approved):
🚀

PR #2715, opened 2024-01-12
Fix some issues on Windows with removed context filenames

DanTup:
This makes some fixes to https://github.com/sourcegraph/cody/pull/2653 for Windows where the paths were appearing incorrectly (wrong slash direction, or not detected as part of the workspace so showing as absolute paths).

This PR is targetting @sqs rm-filename branch since it's fixed to unmerged changes.

## Test plan

- Run `pnpm test:e2e` on Windows and verify all is green


sqs:
Thanks!

sqs:
will merge then incorporate into my branch

PR #2719, opened 2024-01-12
Agent: add support for `commands/{test,smell,explain}`

olafurpg:
Previously, agent clients had to use the old recipe-based API to invoke
the test, smell and explain commands. This PR adds new `commands/*`
endpoints to invoke these commands using the exact same code paths as
they're implemented in VS Code.

## Test plan

See three new test cases.

### DIFF
--- a/agent/src/AgentWebviewPanel.ts
+++ b/agent/src/AgentWebviewPanel.ts
@@ -8,7 +8,7 @@ import { type ExtensionMessage, type WebviewMessage } from '../../vscode/src/cha
 import { defaultWebviewPanel, EventEmitter } from './vscode-shim'
 
 /** Utility class to manage a list of `AgentWebPanel` instances. */
-export class AgentWebPanels {
+export class AgentWebviewPanels {
     public panels = new Map<string, AgentWebviewPanel>()
     public add(panel: AgentWebviewPanel): void {
         this.panels.set(panel.panelID, panel)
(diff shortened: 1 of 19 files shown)


agent/src/AgentWebviewPanel.ts
```diff
@@ -45,6 +56,14 @@ export class AgentWebPanel implements vscode.WebviewPanel {
         })
     }
 
+    public initialize(): void {
```
mkondratek, line 59: is it related to the command support? what does it change?

agent/src/AgentWebviewPanel.ts
```diff
@@ -1,26 +1,37 @@
...
 import type * as vscode from 'vscode'
 
+import { type ChatModelProvider } from '@sourcegraph/cody-shared'
+
 import { type ExtensionMessage, type WebviewMessage } from '../../vscode/src/chat/protocol'
 
 import { defaultWebviewPanel, EventEmitter } from './vscode-shim'
 
 /** Utility class to manage a list of `AgentWebPanel` instances. */
 export class AgentWebPanels {
```
mkondratek, line 11: nit: `s/AgentWebPanels/AgentWebviewPanels` for consistency?

agent/src/AgentWebviewPanel.ts
olafurpg: This PR was based on another PR that is now merged. The diff here was from that PR, not commands-related

agent/src/AgentWebviewPanel.ts
olafurpg: Fixed. This rename was in the other PR that this PR was based on.

mkondratek (approved):
this is fine

PR #2721, opened 2024-01-13
Add more agent tests

beyang:
Add tests for:
- select chat model
- reset chat
- edit chat

Callout: introduces a `chat/editMessage` message type, which closely mirrors `chat/submitMessage`

## Test plan

Only adding tests


olafurpg (approved):
This is great! Thank you @beyang LGTM

PR #2723, opened 2024-01-13
rm more unused code

sqs:
## Test plan

CI

PR #2726, opened 2024-01-14
Improve Knip config

webpro:
After being _very glad_ to see Knip being used in this repository, I saw the configuration and the slightly uncommon blend of the `vscode` and `agent` workspaces (`"../agent/**/*.ts"`), I couldn't resist to check out the repo and come up with some suggestions, as presented in this PR.

Looks like most of the results are (still) correct and there's even more to clean up, yay! I'm happy to leave the actual clean up to developers more familiar with the codebase (I'm also unaware of  potential external factors/consumers).

Feel free to AMA!

Additionally, Knip v4 is around the corner and will give a nice speed bump (on my machine it goes down from 4.7s to 2.8s with the same config).

## Test plan

CI

### DIFF
--- a/knip.jsonc
+++ b/knip.jsonc
@@ -7,14 +7,18 @@
                 "scripts/*.ts",
                 "test/integration/*.ts",
                 "webviews/index.tsx",
-                "../agent/**/*.ts"
+                "webviews/search.tsx"
             ],
-            "ignore": ["src/testutils/uri.ts"]
+            "ignore": ["src/tree-sitter/query-tests/test-data/**"],
+            "ignoreBinaries": ["code", "code-insiders"],
+            "ignoreDependencies": ["@vscode/vsce", "ovsx", "path-browserify"]
         },
         "agent": {
-            "ignore": ["**"]
+            "vitest": {
+                "entry": ["**/__tests__/**/*.ts"]
+            }
         }
     },
     "ignore": ["**/__mocks__/**", "**/mocks.*"],
-    "exclude": ["classMembers", "enumMembers"]
+    "exclude": ["classMembers"]
 }


webpro:
By the way, looks like including unused class members to the report gives clues to remove more dead code :)

sqs:
@webpro Awesome, thanks! This fixes some of the false positives. Appreciate the PR.

sqs (approved):

sqs:
Ran e2e tests locally to merge. Seems to be an auth problem on GitHub Actions.

sqs:
@webpro Indeed, I found a bunch more unused code to delete: https://github.com/sourcegraph/cody/pull/2723. Thank you!!

I noticed a couple false positives:

```
Unused exported class members (20)
providedCodeActionKinds  DocumentCodeAction      vscode/src/code-actions/document.ts:7:28                        
providedCodeActionKinds  EditCodeAction          vscode/src/code-actions/edit.ts:6:28                            
providedCodeActionKinds  ExplainCodeAction       vscode/src/code-actions/explain.ts:4:28                         
providedCodeActionKinds  FixupCodeAction         vscode/src/code-actions/fixup.ts:12:28                          
getContext               FilenameContextFetcher  vscode/src/local-context/filename-context-fetcher.ts:32:18  
```

But neither are shown by tsserver in `Find References` either, so it's not knip's fault and it's not something knip should reasonably expect to hack around. So IMO knip works 100% as expected!

webpro:
Thanks for pointing out the issues. Tbh, I'm not 100% convinced they are false positives (if I follow the code paths then they don't seem to be actually used). But yeah if `Find References` won't find it then (the TS Language Server internal to) Knip won't either.

PR #2731, opened 2024-01-15
Autocomplete: clean up streaming unit tests

valerybugakov:
## Context

- I've been refactoring the autocomplete streaming logic and bumped into these tests. By accident, I discovered a couple of things:
    1. The `expect` check for abort signal being aborted in `onNetworkRequest` callbacks is not applicable anymore. If you `console.log(abortController.signal.aborted)`, it's always `false`, which makes sense logic-wise. We do not cancel the top-level signal; we only cancel individual requests using the forked signal.
    2. These `expect` statements are not taken into account (were never taken into account?) by `vitest`. That's why they all pass even though the values do not match with expects.
- In my other branch, I made these expects reachable and spent an hour trying to understand why it does not work as expected. Then, I finally went to debug how it works on main and discovered the above 😛.
- This PR also enables one of the tests that was skipped previously. 

## Test plan

CI

### DIFF
--- a/vscode/src/completions/get-inline-completions-tests/streaming.test.ts
+++ b/vscode/src/completions/get-inline-completions-tests/streaming.test.ts
@@ -11,7 +11,7 @@ describe('[getInlineCompletions] streaming', () => {
         expect(
             await getInlineCompletions({
                 ...params('const x = █', [completion`├1337\nconsole.log('what?');┤`], {
-                    async onNetworkRequest(_params, onPartialResponse) {
+                    onNetworkRequest(_params, onPartialResponse) {
                         onPartialResponse?.(completion`├1337\ncon┤`)
                     },
                 }),
@@ -26,7 +26,7 @@ describe('[getInlineCompletions] streaming', () => {
         expect(
             await getInlineCompletions({
                 ...params('const x = █', [completion`├1337\nconsole.log('what?');┤`], {
-                    async onNetworkRequest(_params, onPartialResponse) {
+                    onNetworkRequest(_params, onPartialResponse) {
                         onPartialResponse?.(completion`├13┤`)
                         onPartialResponse?.(completion`├1337\n┤`)
                     },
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Approving to unblock, but doesn't it make sense to spin the event loop in these tests (for example between calls to onPartialResponse)? I have certainly seen callbacks *and* Promise-based code with bugs sensitive to the timing of callbacks/resolves. So doing everything synchronously seems... bold.

valerybugakov:
@dominiccooney, we're not trying to do stuff synchronously. We were waiting for the next event loop tick, but checks still were not triggered because they were executed in the nested async callback. I see the removed checks as outdated (@philipp-spiess, correct me if I'm wrong), so there's no need to fix them.

dominiccooney:
@valerybugakov the removed checks are one thing, but the code under test that's exercised by those calls... Typically partial completions etc. would happen after the event loop has spun. So the "does not include unfinished lines in results" test, for example, is using a couple of synchronous callbacks. But in practice those would always? almost always? be asynchronous wouldn't they? So now we're testing something slightly different from reality.

The bugs this misses are things like failure to await Promises before continuing, or using some state that's true on the synchronous callstack but torn down after the event loop spins, things like that.

I didn't dig in here deep to say this is definitely wrong, but I am curious.

philipp-spiess:
> We do not cancel the top-level signal; we only cancel individual requests using the forked signal.

This makes sense. We only pass through the `signal` and we can't abort the abort signal unless we have access to the _controller_, duh. 🤦 

We could give the network callback access to the signal that it was being called with and assert on that instead. WDYT? 

> The bugs this misses are things like failure to await Promises before continuing, or using some state that's true on the synchronous callstack but torn down after the event loop spins, things like that.

That's a good point; It could be that the logic only works if the two network requests are coming back in the same event loop for some reason. It's not guaranteed that this is not an issue in prod at some point. Maybe we should let one test case in there where there's a delay between the two network packets?

philipp-spiess (approved):
Approving because... yeah these didn't ever work as we expected it 🤦

valerybugakov:
> The bugs this misses are things like failure to await Promises before continuing

I see. I thought you were referring specifically to the change introduced in this PR, but you raised a broader issue about our autocomplete test helpers, which is valid. It would be helpful to have a test where partial responses come with a delay. I just bumped into such cases while I migrated to async generators for autocomplete processing logic (more on that in a separate message 😛). I can look into adding such a test case as a follow-up to the PR I will create.

PR #2732, opened 2024-01-15
Symf: move term expansion to separate file

olafurpg:
Previously, I was unable to run the symf tests locally because it reported an issue with `vscode` imports. This error was understandable because the `symf.ts` file imported from the `'vscode'` module. Moving the term expansion function to a separate file fixes the problem. However, I'm not sure how the tests passed in the first place unless something regressed recently.

## Test plan

Green CI


dominiccooney (approved):
Thank you for greening tests.

PR #2733, opened 2024-01-15
Agent: make `pnpm update-agent-recordings` more stable and run faster

olafurpg:
Previously, the `update-agent-recordings` command did an `rm -rf agent/recordings`. This was problematic because 1) it was slow to update all the recordings to record one new HTTP request and 2) it made the tests fail more regularly because the LLM response is random even with temperature=0. This PR addresses both problems by allowing Polly to manage deletion of unused recordings instead of doing `rm -rf agent/recordings`. This means that updating recordings is very fast because it only sends network requests for missing recordings, and it also makes the tests a lot more stable because we keep old recordings around for much longer.

One risk with this change is that we never update recordings. Polly suports configuring an expiry date for recorings but we can't use that at the moment because we hardcode the request date to minimize diffs. We should consider removing that hardcoded date because it's no longer needed after this PR (because we keep recordings around for much longer).

## Test plan

Green CI.

### DIFF
--- a/agent/src/cli/jsonrpc.ts
+++ b/agent/src/cli/jsonrpc.ts
@@ -10,6 +10,7 @@ import { booleanOption } from './evaluate-autocomplete/cli-parsers'
 interface JsonrpcCommandOptions {
     expiresIn?: string | null | undefined
     recordingDirectory?: string
+    keepUnusedRecordings?: boolean
     recordingMode?: MODE
     recordIfMissing?: boolean
     recordingExpiryStrategy?: EXPIRY_STRATEGY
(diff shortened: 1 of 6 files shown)


agent/src/cli/jsonrpc.ts
```diff
@@ -54,6 +55,14 @@ export const jsonrpcCommand = new Command('jsonrpc')
             'Path to the directory where network traffic is recorded or replayed from. This option should only be used in testing environments.'
         ).env('CODY_RECORDING_DIRECTORY')
     )
+    .addOption(
+        new Option(
+            '--keep-unused-recordings <bool>',
+            'If true, unused recordings are not removed from the recording file'
+        )
+            .env('CODY_KEEP_UNUSED_RECORDINGS')
+            .default(false)
```
olafurpg, line 64: I wasn't sure which default to use. I figured it was good to try this first and change it later if it's a problem. I added an explicit `CODY_KEEP_UNUSED_RECORDINGS=false` in the `update-agent-recordings` script even if it's redundant right now, just in case it won't become redundant later.

mkondratek (approved):
I am not an expert in this area but lgtm 😄

dominiccooney (approved):
🎉 

linguist-generated, TIL

pkukielka (commented):
LGTM

Issue #2720, opened 2024-01-12
feedback: Allow user to copy, insert & save code blocks as soon as they are rendered


chillatom:
### Areas for Improvement

- [x] UI/UX
- [ ] Onboarding
- [ ] Docs
- [x] Chat
- [ ] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience.

<img width="543" alt="image" src="https://github.com/sourcegraph/cody/assets/8737510/68263c6e-3d17-404f-911e-697ac747291e">

The buttons to "copy code", "insert code at cursor" and "copy code to new file" are only made available when Cody has finished stream the entire output from the LLM.  This can be annoying if I just want to grab the code and do something with it and Cody spends another 10 seconds spitting out additional text.  

In other tools like ChatGPT you can copy code from the rendered code block before streaming completes. 

### Why is this feedback important? Please describe the rationale.

Saves developer time, improves UX

### Describe the solution you'd like to see

Buttons are displayed as soon as code block is finished being written and rendered.

PR #2737, opened 2024-01-15
change: display codeblock actions while message in progress

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2720

Display codeblock actions while message in progress to allow users to copy and insert code before the answer is completed.

## Test plan

- ask cody to generate a unit test
- you should be able to copy the test code before the message is completed by cody


arafatkatze (approved):

toolmantim:
I don't seem to get this behaviour when testing it on main (ad9c9a64d9a2ff019b3596c3f4833389b8646a48) @abeatrix:

https://github.com/sourcegraph/cody/assets/153/ea7bbe59-7758-47a3-9dbb-b1507c94df9c

PR #2740, opened 2024-01-15
simplify the shared language-detection code

sqs:
We were doing language detection in a lot of different ways: looking at file extensions, running regexps like `/js|ts/` that would incorrectly match a Java file named `Hits.java`, etc. Also, our `getFileExtension` returned the value without the leading `.`, which differs from `path.extname`, which is a likely source of bugs. This cleans up a lot of that code and exposes 2 APIs: `languageFromFilename` and `markdownCodeBlockLanguageIDForFilename`.

This is not a full cleanup:

1. The agent uses a `getLanguageForFileName` API that looks possibly better. We should standardize on one way of doing this throughout the codebase. But this is a good step toward that.
1. There is other code that implements Markdown code block language ID inference for the translation recipe. That still needs to be consolidated.

(Most of the diff is the updated agent polly snapshots, not code.)

## Test plan

CI


arafatkatze (approved):
Stamper!!

PR #2741, opened 2024-01-15
fix cody-shared import from agent issue

sqs:
Now, `esbuild` builds other projects from their TypeScript sources instead of from their intermediate build products in `dist`.

As a benefit, also makes it so you don't need to run `tsc -b` to produce the latest `dist` dirs during agent development.

The cause was likely that the `dist` dirs were out of date (because `tsc -b` was not running or got confused during a `git checkout`), or different import strategies were causing `@sourcegraph/cody-shared` to be imported multiple times.

Discussion: https://sourcegraph.slack.com/archives/C05AGQYD528/p1705319075562969

## Test plan

Run agent tests as described in the now-deleted `lib/shared/src/index.ts` comment and ensure they pass.

### DIFF
--- a/agent/src/esbuild.mjs
+++ b/agent/src/esbuild.mjs
@@ -17,6 +17,13 @@ import { build } from 'esbuild'
         logLevel: 'error',
         alias: {
             vscode: path.resolve(process.cwd(), 'src', 'vscode-shim.ts'),
+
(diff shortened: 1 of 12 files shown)


arafatkatze (approved):
Stamper!!

PR #2745, opened 2024-01-16
DX: upgrade vitest to the latest version

valerybugakov:
## Context

- Upgrades `vitest` to the latest version.
- Since the last version upgrade in our repo, `vitest` had [the first stable release](https://github.com/vitest-dev/vitest/releases/tag/v1.0.0) with a lot of useful features and performance improvements. I wanted to use a couple of CLI flags locally and noticed they are unavailable because of the outdated version.
- No breaking changes for us. Only snapshot updates were required because they look nicer now. 

## Test plan

CI

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -487,30 +487,30 @@ describe('Agent', () => {
         const trimmedMessage = trimEndOfLine(lastMessage?.text ?? '')
         expect(trimmedMessage).toMatchInlineSnapshot(
             `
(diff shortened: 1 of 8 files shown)


sqs (approved):

olafurpg:
Thank you @valerybugakov ! I've been wanting to  bump up vitest for a while. I tried to use some of the APIs from the vitest documentation that didn't exist in the 0.x version we were using.

PR #2747, opened 2024-01-16
Autocomplete: migrate autocomplete streaming post-processing to async generators

valerybugakov:
## Context

I've been diving into our hot streak and dynamic multiline autocomplete features, seeking ways to enhance them. As I prototyped some ideas, I hit a snag with our existing setup. It was clunky to iterate on, which isn't surprising given our past changes.

Recently, we moved a big chunk of the LLM suggestions processing closer to our HTTP streaming phase instead of keeping it post-streaming in a synchronous environment. This shift allowed us to assess suggestions in real time and decide whether to continue streaming or not. But our code wasn't ready for this complexity. With dynamic multiline and hot streak features, we've added callbacks, abort controllers, and promises, overly complicating it.

While tweaking this to make it more adaptable and extensible, I tried using async iterators and ended up fully transitioning to them. They're a perfect fit for managing HTTP streaming, which is at the heart of this logic. Async generators give us an excellent interface over the suggestions we return to the autocomplete provider. Take hot streak completions, for instance – they seamlessly integrate without extra logic; we simply yield additional values after the first completion.

Besides, other perks include a cleaner data flow with no callback interference and the flexibility to support different HTTP clients while maintaining type safety (think Ollama client). 

## Details

- Instead of having separate callbacks for regular and hot streak completions, we now yield more values from the fetch-and-process generator, which wraps the HTTP streaming generator. 
- This PR removes a bunch of nested promises and forked abort controllers, which were hard to reason about. The structure is flatter with uni-directional data flow (provider -> HTTP client -> processor -> provider).
- This PR removes support for multiple completions providers being used together. We never used this "feature" and do not plan to do so soon. Removing it simplifies the code.
- I discovered the issue fixed in https://github.com/sourcegraph/cody/pull/2731 working on this migration. It would be easy to implement @dominiccooney's [suggestion](https://github.com/sourcegraph/cody/pull/2731#issuecomment-1891703877) to improve our autocomplete integration tests.
- Fixes a couple of TODOs in the Ollama client because there's no need for type compatibility between the Sourcegraph HTTP client and Ollama HTTP client anymore.

## Test plan

1. CI
2. Manually verified that both hot streak and dynamic multiline completions work as expected


vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -1,291 +1,214 @@
...
+        if (multiline) {
+            addAutocompleteDebugEvent('multline_branch')
+            const completion = extractCompletion(rawCompletion, {
+                document: providerOptions.document,
+                docContext,
+            })
+
+            if (completion) {
+                const completedCompletion = processCompletion(completion, providerOptions)
+                yield* stopStreamingAndUsePartialResponse({ completedCompletion, isFinal, rawCompletion })
```
philipp-spiess, line 93: This is so cool! Finally a good use case for async generator functions 🤣

philipp-spiess (approved):
This is great! Thank you for driving this. My only question is regarding `zipGenerators` and the fact that it could become a bottleneck, see the inline comment. Rest LGTM.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -175,17 +166,41 @@ class FireworksProvider extends Provider {
...
+         * from `requestManager.request`. Therefore, we must wait for the initial
+         * batch of completions before returning them collectively, ensuring all
+         * are included as suggested completions.
+         *
+         * To circumvent this performance issue, a method for adding completions to
+         * the existing suggestion list is needed. Presently, this feature is not
+         * available, and the switch to async generators maintains the same behavior
+         * as with promises.
+         */
+        return zipGenerators(completionsGenerators)
```
valerybugakov, line 203: Hey @philipp-spiess, I added this comment to sources for context and to respond to your concern. Let me know what you think.

vscode/src/completions/utils.test.ts
```diff
@@ -0,0 +1,53 @@
...
+import { zipGenerators } from './utils'
+
+async function* generatorFromArray<T>(array: T[]) {
+    for (const item of array) {
+        yield await Promise.resolve(item)
+    }
+}
+
+describe('zipGenerators', () => {
+    test('should zip values from multiple generators', async () => {
```
valerybugakov, line 12: Added `zipGenerators` unit tests with Cody's help 💪

vscode/src/completions/utils.ts
```diff
@@ -55,3 +56,45 @@ export function createSubscriber<T>(): Subscriber<T> {
...
+    while (true) {
+        const res = await Promise.all(generators.map(generator => generator.next()))
+
+        if (res.every(r => r.done)) {
+            return
+        }
+
+        yield res.map(r => r.value)
+    }
+}
```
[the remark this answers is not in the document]
valerybugakov: Responded with the inline comment https://github.com/sourcegraph/cody/pull/2747#discussion_r1454402547

vscode/src/completions/providers/fireworks.ts
philipp-spiess: Hmmm I understand that but I still think this is not the right approach. Perhaps adding a middleware where we batch the completion generations and make it so that it yields line by line would be helpful, but without that, we're at the mercy of the upstream inference provider about how often a generator will yield.

This might actually already become a problem with the Starcoder ST rollout: Requests are mapped to either the ST or MT instance based on a per-request roll so the same multi-line completion could hit two different completion backends with a different way of batching the requests. 

I did built that per-line batching at some point; let me find it again. Ah, there it is: [`500f368` (#2118)](https://github.com/sourcegraph/cody/pull/2118/commits/500f36888c0856ad88d7df0bcfa162d3366773a4#diff-9c8feeb3a5aeb9828dd585e95280541b1a9cb297d62cee5a1bc399bd5a660c66R381-R393)

With that we would at least be able to sync the generators to yield line-by-line and we know that we wait for the first complete line of each generator before we yield the first time in `zipGenerators`, wdyt?

valerybugakov: Discussed this question on the Zoom call. The key point here is that `zipGenerators` are applied on the boundary between an autocomplete provider and a request manager, so we already converted all the strings to completion-item data structures, and there, we do not do any additional batching that can introduce delays.

PR #2749, opened 2024-01-16
improve usage of URIs and FS paths for multi-root support

sqs:
Add displayPath{Dir,Base}name and uri{Dir,Base,Ext}name funcs. These helpers make it easier to write code that uses URIs and renders paths correctly across platforms and with multi-root workspaces.

- `displayPath{Dir,Base}name`: for showing the user a URI's dirname or basename
- `uri{Dir,Base,Ext}name`: for path manipulation of URIs that works on `vscode.Uri` and `URI` in VS Code, webviews, tests, agent, and any other environment (pure JS)
    
Then use these new APIs to clean up incorrect usage of `uri.fsPath`, `URI.file`, `vscode.Uri.file`, `dirname`, etc.
    
Also add a `FileURI` type for URIs that are known to be `file`-scheme.

Multi-root is not yet supported by local embeddings or symf because those have hard-coded assumptions of a single workspace root. This PR will make it easier to remove that assumption.

## Test plan

CI and e2e tests. Ensure that symf indexes can be created and queried, and same for local embeddings.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,8 @@
 // @ts-check
 
+const RESTRICT_PATH_IMPORT_MESSAGE =
+  "Use URIs instead of file system paths, and use the following helpers from @sourcegraph/cody-shared: (1) for displaying paths to the user: displayPath, displayPathDirname, displayPathBasename; (2) for manipulating URI paths: uriDirname, uriBasename, uriExtname. If you are writing code that ONLY runs in Node, import from 'node:path'."
+
 /** @type {import('eslint').Linter.Config} */
 const config = {
   extends: ['@sourcegraph/eslint-config', 'plugin:storybook/recommended'],
@@ -72,6 +75,23 @@ const config = {
     ],
   },
   overrides: [
+    {
+      // Apply this rule to a subset of files so we can gradually roll it out.
+      files: ['shared/**/*.ts', 'vscode/src/**/*.ts'],
+      rules: {
+        'no-restricted-imports': [
+          'error',
+          {
+            paths: [
+              { name: 'path', message: RESTRICT_PATH_IMPORT_MESSAGE },
+              { name: 'path/posix', message: RESTRICT_PATH_IMPORT_MESSAGE },
+              { name: 'path/win32', message: RESTRICT_PATH_IMPORT_MESSAGE },
+              { name: 'path-browserify', message: RESTRICT_PATH_IMPORT_MESSAGE },
+            ],
+          },
+        ],
+      },
+    },
     {
       files: ['*.d.ts'],
       rules: {
--- a/lib/shared/src/common/languages.test.ts
+++ b/lib/shared/src/common/languages.test.ts
@@ -1,8 +1,13 @@
 import { describe, expect, test } from 'vitest'
+import { URI } from 'vscode-uri'
 
-import { languageFromFilename, markdownCodeBlockLanguageIDForFilename } from './languages'
+import { languageFromFilename as _languageFromFilename, markdownCodeBlockLanguageIDForFilename } from './languages'
 
 describe('languageFromFilename', () => {
+    function languageFromFilename(name: string): ReturnType<typeof _languageFromFilename> {
+        return _languageFromFilename(URI.parse(`file:///${name}`))
+    }
+
     test('', () => {
         expect(languageFromFilename('foo.java')).toBe('Java')
         expect(languageFromFilename('foo.go')).toBe('Go')
@@ -22,12 +27,12 @@ describe('languageFromFilename', () => {
 
 describe('markdownCodeBlockLanguageIDForFilename', () => {
     test('simple', () => {
-        expect(markdownCodeBlockLanguageIDForFilename('foo.java')).toBe('java')
-        expect(markdownCodeBlockLanguageIDForFilename('foo.go')).toBe('go')
+        expect(markdownCodeBlockLanguageIDForFilename(URI.parse('file:///foo.java'))).toBe('java')
+        expect(markdownCodeBlockLanguageIDForFilename(URI.parse('file:///foo.go'))).toBe('go')
     })
     test('complex', () => {
-        expect(markdownCodeBlockLanguageIDForFilename('foo.js')).toBe('javascript')
-        expect(markdownCodeBlockLanguageIDForFilename('foo.ts')).toBe('typescript')
-        expect(markdownCodeBlockLanguageIDForFilename('foo.tsx')).toBe('typescript')
+        expect(markdownCodeBlockLanguageIDForFilename(URI.parse('file:///foo.js'))).toBe('javascript')
+        expect(markdownCodeBlockLanguageIDForFilename(URI.parse('file:///foo.ts'))).toBe('typescript')
+        expect(markdownCodeBlockLanguageIDForFilename(URI.parse('file:///foo.tsx'))).toBe('typescript')
     })
 })
--- a/lib/shared/src/common/uri.test.ts
+++ b/lib/shared/src/common/uri.test.ts
@@ -1,0 +1,42 @@
+import { describe, expect, test } from 'vitest'
+import { URI } from 'vscode-uri'
+
+import { uriBasename, uriDirname, uriExtname, uriParseNameAndExtension } from './uri'
+
+describe('uriDirname', () => {
+    test('', () => {
+        expect(uriDirname(URI.parse('file:///a/b/c')).toString()).toBe('file:///a/b')
+        expect(uriDirname(URI.parse('https://example.com/a/b')).toString()).toBe('https://example.com/a')
+        expect(uriDirname(URI.parse('https://example.com/a/')).toString()).toBe('https://example.com/')
+        expect(uriDirname(URI.parse('https://example.com/')).toString()).toBe('https://example.com/')
+    })
+})
+
+describe('uriBasename', () => {
+    test('', () => {
+        expect(uriBasename(URI.parse('file:///a/b/c')).toString()).toBe('c')
+        expect(uriBasename(URI.parse('https://example.com/a/b')).toString()).toBe('b')
+        expect(uriBasename(URI.parse('file:///c:/a/b')).toString()).toBe('b')
+        expect(uriBasename(URI.parse('file:///c:/')).toString()).toBe('c:')
+        expect(uriBasename(URI.parse('file:///c:')).toString()).toBe('c:')
+        expect(uriBasename(URI.parse('file:///c%3A')).toString()).toBe('c:')
+        expect(uriBasename(URI.parse('file:///a/b%20c')).toString()).toBe('b c')
+    })
+})
+
+describe('uriExtname', () => {
+    test('', () => {
+        expect(uriExtname(URI.parse('file:///a/b.txt')).toString()).toBe('.txt')
+        expect(uriExtname(URI.parse('https://example.com/a/b.test.rb')).toString()).toBe('.rb')
+        expect(uriExtname(URI.parse('file:///c:/a/.foo.js')).toString()).toBe('.js')
+        expect(uriExtname(URI.parse('file:///a/b')).toString()).toBe('')
+    })
+})
+
+describe('uriParseNameAndExtension', () => {
+    test('', () => {
+        expect(uriParseNameAndExtension(URI.parse('file:///a/b.txt'))).toEqual<
+            ReturnType<typeof uriParseNameAndExtension>
+        >({ name: 'b', ext: '.txt' })
+    })
+})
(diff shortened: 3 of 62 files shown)


sqs:
@DanTup I'd love your review here of the API I'm introducing in this PR to avoid footguns with path presentation and URI manipulation. The commit message has a high-level description, there are docstrings and tests, and this PR is working through cleaning up all (99% wrong) usage of `uri.fsPath`, `URI.file`, etc.

---

**add displayPath{Dir,Base}name and uri{Dir,Base,Ext}name funcs**
    
These helpers make it easier to write code that uses URIs and renders paths correctly across platforms and with multi-root workspaces.
    
- `displayPath{Dir,Base}name`: for showing the user a URI's dirname or basename
- `uri{Dir,Base,Ext}name`: for path manipulation of URIs that works on `vscode.Uri` and `URI` in VS Code, webviews, tests, agent, and any other environment (pure JS)

lib/shared/src/common/uri.test.ts
```diff
@@ -0,0 +1,32 @@
...
+    })
+})
+
+describe('uriBasename', () => {
+    test('', () => {
+        expect(uriBasename(URI.parse('file:///a/b/c')).toString()).toBe('c')
+        expect(uriBasename(URI.parse('https://example.com/a/b')).toString()).toBe('b')
+        expect(uriBasename(URI.parse('file:///c:/a/b')).toString()).toBe('b')
+        expect(uriBasename(URI.parse('file:///c:/')).toString()).toBe('c:')
+        expect(uriBasename(URI.parse('file:///c:')).toString()).toBe('c:')
```
DanTup, line 21: We should add some tests with escaped colons, because VS Code can encode them in URIs into something like:

```
file:///C%3A/Dev/Test%20Projects/dart_application_3
```

See https://github.com/Microsoft/vscode/issues/2990

lib/shared/src/editor/displayPath.ts
```diff
@@ -20,34 +20,90 @@ import { basename } from '../common'
  * - `vscode.workspace.asRelativePath`: because it's not available in webviews, and it does not
  *   handle custom URI schemes (such as if we want to represent remote files that exist on the
  *   Sourcegraph instance).
- * @param location The absolute URI or file path to convert to a display path.
+ * @param location The absolute URI to convert to a display path.
  */
-export function displayPath(location: URI | string): string {
+export function displayPath(location: URI): string {
+    // TODO(sqs): make this only take a URI, not a string
```
DanTup, line 26: Is this TODO redundant?

lib/shared/src/editor/displayPath.ts
```diff
@@ -20,34 +20,90 @@ import { basename } from '../common'
...
+ * file URIs on Windows, '/' otherwise).
+ *
+ * The returned path string MUST ONLY be used for display purposes. It MUST NOT be used to identify
+ * or locate files.
+ *
+ * Use this instead of other seemingly simpler techniques to avoid a few subtle
+ * bugs/inconsistencies:
+ *
+ * - On Windows, Node's `dirname(uri.fsPath)` breaks on a non-`file` URI on Windows because
+ *   `dirname` would use '\' path separators but the URI would have '/' path separators.
```
DanTup, line 43: Isn't it great how easy cross-platform development is these days? 😄

vscode/src/chat/ContextProvider.ts
```diff
@@ -61,7 +61,7 @@ export class ContextProvider implements vscode.Disposable, ContextStatusProvider
     public configurationChangeEvent = new vscode.EventEmitter<void>()
 
     // Codebase-context-related state
-    public currentWorkspaceRoot: string
+    public currentWorkspaceRoot: vscode.Uri | null = null
```
DanTup, line 64: Any reason for null instead of `undefined`?

vscode/src/chat/ContextProvider.ts
```diff
@@ -146,8 +145,12 @@ export class ContextProvider implements vscode.Disposable, ContextStatusProvider
             // these are ephemeral
             return
         }
-        const workspaceRoot = this.editor.getWorkspaceRootUri()?.fsPath
-        if (!workspaceRoot || workspaceRoot === '' || workspaceRoot === this.currentWorkspaceRoot) {
+        const workspaceRoot = this.editor.getWorkspaceRootUri()
+        if (
+            !workspaceRoot ||
+            workspaceRoot === null ||
```
DanTup, line 151: Is line 151 redundant (covered by 150)? Probably same was try before this change.

vscode/src/commands/prompt/utils.ts
```diff
@@ -31,20 +33,20 @@ export function toSlashCommand(command: string): string {
 
 /**
  * Checks if the given file path is a valid test file name.
```
DanTup, line 35: "file path" -> "file URI"?

vscode/src/commands/utils/new-test-file.test.ts
```diff
@@ -26,81 +25,83 @@ describe('createDefaultTestFileNameByLanguageExt', () => {
...
-        expect(convertFsPathToTestFile(filePath, existingTestFilePath)).toBe(expected)
+    it('should generate the default spec file uri for ruby when no exisiting test files is found', () => {
+        const currentFile = URI.file('/path/to/file.rb')
+        const existingTestFile = undefined
+        const expected = URI.file('/path/to/file_spec.rb')
+        expect(convertFileUriToTestFileUri(currentFile, existingTestFile)).toStrictEqual(expected)
+    })
+
+    it('should generate the corrent test file uri for window files', () => {
+        const currentFile = URI.file('\\path\\to\\file.ts')
```
DanTup, line 95: Should these be tested with drive letters?

vscode/src/edit/provider.ts
```diff
@@ -211,18 +214,20 @@ export class EditProvider {
...
         // Create a new file uri by replacing the file name of the currentFileUri with fileName
-        const newFileFsPath = currentFileUri.replace(currentFileName, newFileName.trim())
+        let newFileUri = Utils.joinPath(currentFileUri, '..', newFileName)
 
         if (haveSameExtensions && !NewFixupFileMap.get(task.id)) {
-            const fileIsFound = await doesFileExist(URI.parse(newFileFsPath))
-            const newFileUri = URI.parse(fileIsFound ? newFileFsPath : `untitled:${newFileFsPath}`)
+            const fileIsFound = await doesFileExist(newFileUri)
+            if (!fileIsFound) {
+                newFileUri = URI.parse(`untitled:${newFileUri.path}`)
```
DanTup, line 229: Probably doesn't matter here, but I think could just be `newFileUri.with({scheme: 'untitled'})` to avoid the parse (I don't know the details, but URI.parse is weird enough that it has a "strict" boolean parameter).

vscode/src/editor-context/VSCodeEditorContext.ts
```diff
@@ -125,28 +124,19 @@ export class VSCodeEditorContext {
         if (!activeEditor?.fileUri) {
             return []
         }
-        if (activeEditor.fileUri && activeEditor.fileUri.scheme === 'file') {
-            const currentDir = dirname(activeEditor.fileUri.fsPath)
-            return this.getEditorDirContext(currentDir, activeEditor.fileUri, isUnitTestRequest)
-        }
-        return []
+        const currentDir = uriDirname(activeEditor.fileUri)
+        return this.getEditorDirContext(currentDir, activeEditor.fileUri, isUnitTestRequest)
```
DanTup, line 128: Is it deliberate that we might now go in here for non-file URIs?

vscode/src/local-context/symf.ts
```diff
@@ -149,13 +148,13 @@ export class SymfRunner implements IndexedKeywordContextFetcher, vscode.Disposab
         return lock
     }
 
-    private async unsafeRunQuery(keywordQuery: string, scopeDir: string): Promise<string> {
+    private async unsafeRunQuery(keywordQuery: string, scopeDir: vscode.Uri): Promise<string> {
```
DanTup, line 151: Should we use `FileUri` instead of `vscode.Uri` for some of these? If we get a non-file URI, `.fsPath` will throw.

(Same goes for a few places further down this file)

DanTup (commented):
Added some comments/questions, but generally looks really good to me, removes a lot of confusion :-)

.eslintrc.js
```diff
@@ -1,5 +1,8 @@
 // @ts-check
 
+const RESTRICT_PATH_IMPORT_MESSAGE =
+  "Use URIs instead of file system paths, and use the following helpers from @sourcegraph/cody-shared: (1) for displaying paths to the user: displayPath, displayPathDirname, displayPathBasename; (2) for manipulating URI paths: uriDirname, uriBasename, uriExtname. If you are writing code that ONLY runs in Node, import from 'node:path'."
```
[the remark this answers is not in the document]
sqs: > How do I know if this is the case?

Good question. I was thinking of webview code (as not running in Node), since I am not expecting people to keep VS Code Web in their mind.

I also don't know the state of Cody in VS Code Web, and I don't think anyone keeps it top of mind (and there are no tests for it). That would be a good integration test to add IMO (I personally believe that it would help us keep using clean practices), but I know it adds a burden as well, so I will check with folks before doing that.

> some custom rules

Yes, that is a good idea. I will add more custom rules to avoid `.fsPath` and will check out the others you added.

vscode/src/edit/provider.ts
sqs: good call!

vscode/src/editor-context/VSCodeEditorContext.ts
sqs: Yes, in that IME only using `file`-scheme URIs in VS Code extensions is almost always a mistake (no VS Code Web support, no untitled file support, does not play nicely with extension content providers, etc.).

vscode/src/local-context/symf.ts
sqs: `FileURI` was an experiment. Do you like it? Yes, I will use it here.

sqs: Added, and separated FileURI into a separate PR because it can be independent. I will keep the FileURI commits in this branch as well until the FileURI PR is merged.

https://github.com/sourcegraph/cody/pull/2827

vscode/src/editor-context/VSCodeEditorContext.ts
sqs: Nevermind, can't be separated very cleanly/easily.

vscode/src/chat/ContextProvider.ts
sqs: Because it was frequently checked against a `workspaceRoot` value that was `whatever | null`. But now that I look at it here, I think it is clearer for this to be `whatever | undefined` as you say.

vscode/src/commands/utils/new-test-file.test.ts
sqs: This change is no longer in this PR.

lib/shared/src/common/uri.test.ts
sqs: Good call. Added test cases.

lib/shared/src/common/languages.test.ts
```diff
@@ -1,8 +1,13 @@
...
+import { URI } from 'vscode-uri'
 
-import { languageFromFilename, markdownCodeBlockLanguageIDForFilename } from './languages'
+import { languageFromFilename as _languageFromFilename, markdownCodeBlockLanguageIDForFilename } from './languages'
 
 describe('languageFromFilename', () => {
+    function languageFromFilename(name: string): ReturnType<typeof _languageFromFilename> {
+        return _languageFromFilename(URI.parse(`file:///${name}`))
+    }
+
```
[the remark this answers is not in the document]
sqs: This is just a test helper, so I don't think that is needed. Also, IMO `FromFilename` is better because it makes it clear(er) that the language detection is only occurring based on the filename, which is not correct in general. `FromUri` seems more open to the confusion of "maybe it reads the file contents to determine what language it is".

.eslintrc.js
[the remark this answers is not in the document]
sqs: Also this will help :smile: 

image

It makes it show up like this, and the hover explains why it's struck through:

image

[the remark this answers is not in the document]
sqs: @DanTup I think this deprecation notice is enough for now. There are a lot more usages of `.fsPath` that need cleaning up first before we can enforce it via lint. I suppose we could grandfather in all existing usages, but I think it would just frustrate/confuse people until you/I clean it up further so almost everything uses URIs.

sqs:
@DanTup Ready for review. No major changes after you last looked at it 2 days ago. Thank you!

.eslintrc.js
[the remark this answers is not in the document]
DanTup: nice, thanks :)

DanTup (approved):
Still all LGTM :)

PR #2757, opened 2024-01-16
Edit: Add codelens shortcuts

umpox:
## Description

image

This PR:
- Adds keyboard shortcuts to common Edit codelens actions (Accept, Retry, Undo, Cancel)
  - These shortcuts are only enabled when an Edit is in an active actionable state, and it is in a visible document. This will help reduce the risk of these shortcuts becoming an annoyance if a user doesn't use Edit much.
- Updates "Retry" text to show "Edit & Retry"

## Test plan

Single edit:

1. Create an edit
2. Try undo/retry/save shortcuts

Multiple edit (actions nearest edit):

1. Create multiple edits
2. Try undo/retry/save shortcuts and confirm that they only trigger against the nearest edit to the cursor position

Shortcut enablement: 

1. Create an edit
2. Leave the file
3. Try undo/retry/save shortcuts. Confirm nothing happens
4. Return to the file
5. Try undo/retry/save shortcuts. Confirm they work


vscode/src/non-stop/codelenses/items.ts
```diff
@@ -165,8 +166,9 @@ function getAppliedLens(codeLensRange: vscode.Range, id: string): vscode.CodeLen
 
 function getRetryLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens {
     const lens = new vscode.CodeLens(codeLensRange)
+    const shortcut = process.platform === 'win32' ? 'Alt+R' : '⌥R'
     lens.command = {
-        title: 'Retry',
+        title: `Edit & Retry [${shortcut}]`,
```
abeatrix, line 171: Do you know why this is not affecting the e2e test where we only look for `Edit`?

vscode/src/non-stop/codelenses/items.ts
```diff
@@ -185,8 +188,9 @@ function getUndoLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens {
 
 function getAcceptLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens {
     const lens = new vscode.CodeLens(codeLensRange)
+    const shortcut = process.platform === 'win32' ? 'Alt+A' : '⌥A'
```
abeatrix, line 191:
```suggestion
    const shortcut = process.platform === 'darwin' ? '⌥A' : 'Alt+A'
```

vscode/src/non-stop/codelenses/items.ts
```diff
@@ -135,8 +135,9 @@ function getFormattingSkipLens(codeLensRange: vscode.Range, id: string): vscode.
 
 function getCancelLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens {
     const lens = new vscode.CodeLens(codeLensRange)
+    const shortcut = process.platform === 'win32' ? 'Alt+C' : '⌥Z'
```
abeatrix, line 138:
```suggestion
    const shortcut = process.platform === 'darwin' ? '⌥Z' : 'Alt+C'
```

Doesn't Linux uses Alt as well? If not you can ignore this lol

vscode/src/non-stop/codelenses/items.ts
```diff
@@ -165,8 +166,9 @@ function getAppliedLens(codeLensRange: vscode.Range, id: string): vscode.CodeLen
 
 function getRetryLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens {
     const lens = new vscode.CodeLens(codeLensRange)
+    const shortcut = process.platform === 'win32' ? 'Alt+R' : '⌥R'
```
abeatrix, line 169:
```suggestion
    const shortcut = process.platform === 'darwin' ? '⌥R' : 'Alt+R'
```

abeatrix (commented):
Left comments/questions inline. I'm wondering why the current e2e test is passing when the code lens title has changed 🤔  Do we also want a shortcut for ShowDiff?

toolmantim (commented):
Looking great!

A few things:
* When an edit is running, the cancel shortcut (opt-z) doesn't seem to work
* Can we move Show Diff to the end?
* Can we hide the generic Cody Code Lenses if we have edit code lenses, so we don't get the double up seen in the below?
* I tested out a few different labelling formats (see below), I think #2 is the best (parentheses) — wdyt?

1:

CleanShot 2024-02-11 at 00 19 49@2x

Without the generic Cody lens:

CleanShot 2024-02-11 at 00 26 22@2x


2:

CleanShot 2024-02-11 at 00 21 02@2x

Without generic Cody lens:

CleanShot 2024-02-11 at 00 28 08@2x

3:

CleanShot 2024-02-11 at 00 21 53@2x

vscode/src/non-stop/codelenses/items.ts
umpox: Yesss, good catch...

vscode/src/non-stop/codelenses/items.ts
umpox: Where we look for `Retry`? I think it's because the text is still there. I'll update the e2e test so it matches exactly

umpox:
@toolmantim 

> When an edit is running, the cancel shortcut (opt-z) doesn't seem to work

Fixed

> Can we move Show Diff to the end?

Done. Do you think we should include a shortcut for it too?

> I tested out a few different labelling formats (see below), I think `#2` is the best (parentheses) — wdyt?

Agree. Updated

umpox:
> Can we hide the generic Cody Code Lenses if we have edit code lenses, so we don't get the double up seen in the below?

@toolmantim I looked briefly into this. I think we can do this but it'll be a follow up PR and require some refactoring of the Cody codelens so we can easily hide this for a location within a particular document.

I think we should also consider what to do about this codelens first. It's disabled by default so we should either enable it fully or rethink how we show it so we can enable it by default

toolmantim (approved):
Looking good!

> Done. Do you think we should include a shortcut for [Show Diff] too?

Maybe not, to keep keyboard shortcut footprint low? Or do you think it's weird not having one? Either way, can be a follow-up.

Issue #2758, opened 2024-01-16
allow agent to run all commands in chat mode


abeatrix:
Refactor commands to allow running edit commands as chat prompt.

PR #2760, opened 2024-01-16
add: new chat mode to run edit command in chat view

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2758

This PR adds new command arguments and standardizes command execution. 

The new change allows agent to run the default edit commands as chat commands when setting `runInChatMode` to true.

This change standardizes how commands are executed by creating a `CodyCommandArgs` interface that is passed to command execution. The args include a `requestID` and `source` to better track command execution for logging purpose. It also includes a new argument called `runInChatMode` that allows caller to run the command in chat mode only, and allows the agent to run `edit` commands like `/doc`, and the upcoming `/test`, in chat view instead of inline edit.

Command execution is also now centralized through the executeCommand method rather than directly calling chatManager or command runners.

Additional changes:
- Update relevant methods to accept the new CodyCommandArgs
- Update prompt for /doc command so that it would work in both edit and non-edit settings
- Generate requestID in command execution time when not provided
- Remove some unnecessary early returns
- Basic clean-up

## Test plan

- All the agent tests for commands are passing.
- All the existing tests for commands are passing
- `/doc` command works as expected
- 
image

1. In this branch, try setting "mode" to "edit" or "file" for the current `/test` command in the [cody.json file](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/commands/prompt/cody.json?L19-27)
2. run the unit tests `pnpm unit:test`, the unit test should still be passing even though the `/test` command is now an edit command

## Next Step

Once this is merged, we can add the `/doc` command to agent to replace the recipe one they are currently using.


olafurpg (approved):
LGTM from the perspective of the agent-related changes 👍🏻 As long as the agent can run those commands inside the chat panel then I don't see a problem with this change.

PR #2765, opened 2024-01-17
refactor: convert fs paths to uris in test file utils

abeatrix:
refactor: convert filesystem paths to uris in test file utils

Refactored the current test file utils to use URIs instead of filesystem paths. 

### Changes

- convertFsPathToTestFile -> convertFileUriToTestFileUri
- isValidTestFileName -> isValidTestFile
- Updated method signatures to use URI instead of strings 
- Updated tests to construct URI objects
- Generate new file requests using URI.parse instead of strings
- Replace uses of uri.fsPath -> uri.path in both functions
- Replace uses of dirname, basename, ext from path with path.posix

## Test plan

- all current tests are passing
- the /unit command (behind cody.internal.unstable config) works as expected
- 
image


vscode/src/edit/provider.ts
```diff
@@ -197,12 +197,12 @@ export class EditProvider {
...
-            if (cbTestFile) {
-                const testFsPath = convertFsPathToTestFile(task.fixupFile.uri.fsPath, cbTestFile)
-                const fileExists = await doesFileExist(URI.file(testFsPath))
-                const newFileUri = URI.parse(fileExists ? testFsPath : `untitled:${testFsPath}`)
+            // an exisiting test file from codebase
+            const cbTestFileUri = task.contextMessages?.find(m => m?.file?.uri?.fsPath?.includes('test'))?.file?.uri
+            if (cbTestFileUri) {
+                const testFileUri = convertFileUriToTestFileUri(task.fixupFile.uri, cbTestFileUri)
+                const fileExists = await doesFileExist(testFileUri)
+                const newFileUri = URI.parse(fileExists ? testFileUri.fsPath : `untitled:${testFileUri.fsPath}`)
```
dominiccooney, line 205: I know you're just maintaining the status quo with this edit, but does the `fileExists` case, `URI.parse(testFileUri.fsPath)`, work on Windows? `URI.parse('C:\\my\\file.txt')` produces a URI with scheme "C" which isn't right.

Think this should be something like:

```
const newFileUri = fileExists ? testFileUri : URI.parse(`untitled:${testFileUri.fsPath}`)
```

vscode/src/commands/utils/new-test-file.test.ts
```diff
@@ -26,81 +25,69 @@ describe('createDefaultTestFileNameByLanguageExt', () => {
...
-        const filePath = '/path/to/test-file.py'
-        const expectedFilePath = '/path/to/test-file_test.py'
-        const expected = withPlatformSlashes(expectedFilePath)
-        expect(convertFsPathToTestFile(filePath)).toBe(expected)
+        const currentFile = URI.file('/path/to/test-file.py')
+        const expected = URI.file('/path/to/test-file_test.py')
+        expect(convertFileUriToTestFileUri(currentFile)).toStrictEqual(expected)
     })
 
     it('should generate the default spec file path for ruby when no exisiting test files is found', () => {
```
dominiccooney, line 87: Would be great to add some tests for Windows-style file paths, if that is possible. Maybe you can pass the directory separator as an optional parameter which defaults to the OS separator, so you can override it just for running the Windows-style path tests on macOS and Linux.

dominiccooney (changes requested):
Looks good, but can we make the Windows path handling and test coverage more robust while we're here?

vscode/src/commands/utils/new-test-file.ts
```diff
@@ -33,31 +35,31 @@ export function createDefaultTestFileNameByLanguageExt(fileName: string, ext: st
...
+ * If an existing test file uri is provided, attempts to use the same naming convention. Falls back to default if naming doesn't follow conventions.
+ * @param currentFileUri - URI of file to convert
+ * @param existingTestFileUri - Optional uri of existing test file retrived from codebase to match naming with
+ * @returns The uri of the converted test file
  */
-export function convertFsPathToTestFile(currentFilePath: string, existingTestFilePath?: string): string {
-    if (isValidTestFileName(currentFilePath)) {
-        return currentFilePath
+export function convertFileUriToTestFileUri(currentFileUri: URI, existingTestFileUri?: URI): URI {
+    if (isValidTestFileName(currentFileUri.fsPath)) {
```
sqs, line 50: In general, you want to avoid `uri.fsPath` unless you are writing to the file system directly (which you want to avoid because that won't work in webviews or on VS Code Web).

Here, you can pass the uri to `isValidTestFileName` (and make it accept a uri if it doesn't already).

Instead of using `basename` `dirname` from Node's `fs` module, use them from `path/posix` and operate on `uri.path` instead of `uri.fsPath`. The reason to use `path/posix` is that it always operates on paths with forward slashes, and the reason to use `uri.path` is because it always has forward slashes. On Windows, Node's `path` module will work with backslashes, and `fsPath` will have backslashes. It might seem that on Windows both would be backslashes so it would work out, but it turns out that in webviews and on VS Code Web, vite or webpack will bundle `path-browserify` for Node's `path` module, which always uses forward slashes.

This is tricky, and I have another branch that should help with this (`sqs/multiroot` but WIP).

tl;dr is:

- avoid `uri.fsPath`, use `uri.path` instead
- avoid `basename` `dirname` from Node `path`, use Node `path/posix` instead
- avoid `vscode.Uri.file(...)` and `URI.file(...)` - you should only use these if you are reading a file path from something like an external process - within the extension code, just pass around URIs and use `vscode.Uri.joinPath`

vscode/src/commands/utils/new-test-file.test.ts
```diff
@@ -26,81 +25,83 @@ describe('createDefaultTestFileNameByLanguageExt', () => {
...
-    it('should generate a test file path for a non-test file path in python', () => {
-        const filePath = '/path/to/file.py'
-        const existingTestFilePath = '/path/to/testFile_test.py'
-        const expectedFilePath = '/path/to/file_test.py'
-        const expected = withPlatformSlashes(expectedFilePath)
-        expect(convertFsPathToTestFile(filePath, existingTestFilePath)).toBe(expected)
+    it('should generate a test file uri for a non-test file uri in python', () => {
+        const currentFile = URI.file('/path/to/file.py')
+        const existingTestFile = URI.file('/path/to/testFile_test.py')
+        const expected = URI.file('/path/to/file_test.py')
```
sqs, line 77: For ease of readability, I recommend not pulling out the expected value into a separate variable. Just inline it. And same for the value inside `expect(...)` unless it's long.

So like:

```typescript
expect(convertFileUriToTestFileUri(URI.file('...'), ...).toString()).toBe('file:///whatever')
```

(also `.toString()` is helpful when comparing URIs since the objects are not equal, only the string representations are)

sqs:
@abeatrix Just noticed this when calling `vscode.Uri.joinPath` on a `vscode-uri.URI` value as the first arg. Will get it merged soon, and you might find it useful to rebase this PR on that for your tests.

https://github.com/sourcegraph/cody/pull/2772

vscode/src/commands/utils/new-test-file.ts
abeatrix: @sqs Oh wow this is super helpful, will definitely keep this in mind, thank you so much for sharing this with me! 
I've updated the PR with your suggestions. Will try to finalize this and update the tests to cover paths on different OS like @dominiccooney suggested tomorrow. Thank you both again for the constructive feedback and reviews 🙇‍♀️

sqs: Yep! I had to figure this all out over the last week or so, and I'm about to send up a PR with cleaner APIs for doing this and some eslint rules to flag code.

sqs (approved):
Added some comments, LGTM, thanks! Super helpful for me as I'm working on a cleaner API for working with URIs in our codebase too.

vscode/src/commands/utils/new-test-file.ts
```diff
@@ -10,71 +12,73 @@ const TEST_FILE_DASH_SUFFIX_EXTENSIONS = new Set(['py', 'go'])
...
-            return join(dirPath, createDefaultTestFileNameByLanguageExt(currentFileNameWithoutExt, currentFileExt))
+    // If there is an existing test file, create the test file following its naming convention
+    if (testFile?.path && isValidTestFile(testFile)) {
+        const parsedCurrentFile = posix.parse(currentFile.path)
+        const parsedTestFile = posix.parse(testFile.path)
+        if (parsedCurrentFile.ext === parsedTestFile.ext) {
+            // Check if the existing test file has a non-alphanumeric character at the test character index
+            // If so, we will create a test file following the generic naming convention
+            // else, create one follow the existing test file's naming convention
+            const testFileName = parsedTestFile.name.toLowerCase()
```
sqs, line 71: Why toLowerCase? I think it's rare for uppercase letters in a filename to have no meaning. Is it necessary to toLowerCase this value?

vscode/src/commands/utils/new-test-file.ts
```diff
@@ -10,71 +12,73 @@ const TEST_FILE_DASH_SUFFIX_EXTENSIONS = new Set(['py', 'go'])
...
+            const index = testFileName.lastIndexOf('test') || testFileName.lastIndexOf('spec')
+            if (!(index > -1 && !/^[\da-z]$/i.test(testFileName[index - 1]))) {
+                // Use the existing file's naming convention
+                // Assuming that 'existing' should be replaced with the current file name without extension
+                const newTestFileName = parsedTestFile.name.replace(/existing/i, parsedCurrentFile.name)
```
sqs, line 76: What would make a filename contain the string `existing`?

vscode/src/edit/provider.ts
```diff
@@ -197,12 +197,13 @@ export class EditProvider {
         const task = this.config.task
         // Manually create the file if no name was suggested
         if (!text.length && !isMessageInProgress) {
-            const cbTestFile = task.contextMessages?.find(m => m?.file?.uri?.fsPath?.includes('test'))?.file?.uri
-                ?.fsPath
-            if (cbTestFile) {
-                const testFsPath = convertFsPathToTestFile(task.fixupFile.uri.fsPath, cbTestFile)
-                const fileExists = await doesFileExist(URI.file(testFsPath))
-                const newFileUri = URI.parse(fileExists ? testFsPath : `untitled:${testFsPath}`)
+            // an exisiting test file from codebase
```
sqs, line 200: typo: existing

vscode/src/commands/utils/new-test-file.ts
```diff
@@ -10,71 +12,73 @@ const TEST_FILE_DASH_SUFFIX_EXTENSIONS = new Set(['py', 'go'])
...
-            existingFileNameWithoutExt.toLowerCase().lastIndexOf('test') ||
-            existingFileNameWithoutExt.toLowerCase().lastIndexOf('spec')
-        if (testCharIndex > -1 && !/^[\da-z]$/i.test(existingFileNameWithoutExt[testCharIndex - 1])) {
-            return join(dirPath, createDefaultTestFileNameByLanguageExt(currentFileNameWithoutExt, currentFileExt))
+    // If there is an existing test file, create the test file following its naming convention
+    if (testFile?.path && isValidTestFile(testFile)) {
+        const parsedCurrentFile = posix.parse(currentFile.path)
+        const parsedTestFile = posix.parse(testFile.path)
+        if (parsedCurrentFile.ext === parsedTestFile.ext) {
+            // Check if the existing test file has a non-alphanumeric character at the test character index
```
sqs, line 68: What's an example of a test file that has a non-alphanumeric character at the test character index? Like `_test`?

sqs (approved):
Code LGTM (minor comments)

vscode/src/editor-context/helpers.ts
```diff
@@ -67,7 +67,7 @@ export const getFilesFromDir = async (
                 return !isDirectory && !isHiddenFile
             }
 
-            const isFileNameIncludesTest = isValidTestFileName(fileName)
+            const isFileNameIncludesTest = isValidTestFile(vscode.Uri.file(fileName))
```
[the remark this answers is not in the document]
abeatrix: Good callout on the abstraction break with isValidTestFile! I agree it would be cleaner to split that into separate checks, but for now I've updated it to focus on consolidating the main test file generation logic since this is the only exception where we need to generate the URI for the file manually (i've updated the uri generated here to be more accurate as well via `Utils.joinPath(dirUri, fileName)` as you suggested as well).

I've also removed the `node_modules ` check from the function as it was added previously before I know how to [filter node_modules out via excludePattern](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/editor-context/helpers.ts?L89), so this is no longer necessary IMO but I will keep an eye out and add it back if an issue occur :)

vscode/src/commands/utils/new-test-file.ts
abeatrix: LOL this is fixed! It comes from an old PR I worked on before but forgot to update the code lol Also updated comments to address the questions you raised.

Note: Will try to improve this in a follow-up PR as this currently doesn't get called often, and only when Cody failed to suggest a test file name for the new test command that's behind the internal unstable flag.

Issue #2766, opened 2024-01-17
remove error banners related to context limit


abeatrix:
Background context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1705357214212829?thread_ts=1705356180.305859&cid=C05AGQYD528

Suggested by Corey: Remove all banners that warns of context limit increase

dominiccooney:
> Suggested by Corey: Remove all banners that warns of context limit increase

@chillatom I think we should look at the places in the code which "push context warnings" (code like this in DefaultPrompter/makePrompt for example):

```
            if (limitReached) {
                warnings.push('Ignored current user-specified context items due to context limit')
                return { prompt: promptBuilder.build(), contextLimitWarnings: warnings, newContextUsed }
            }
```

And decide, do we want to keep a warning for this and truncate at this point?

And second, when we have too much context all together, what truncation behavior do we want?

Here's my first cut:
- User specified context items too large in aggregate: Continue to warn. (Later, make the warning nicer about which ones we dropped.) My thinking here is we're disobeying something you *just* explicitly told us to do.
- Previous context too large: Don't warn.
- Previous transcript too large: Don't warn.

Here's the truncation behavior we have today. Include the following until the limit is reached, then truncate:
- Preamble
- Existing transcript up to current message, prioritizing recent messages (that is, current message is the most important)
- Explicit tagged context up to current message, prioritizing recent messages (that is, current tagged context is the most important)
- Current enhanced context

We should do some hand inspection of this truncation, but I think we will do better with something like:
- Preamble
- Current message
- Current explicit tagged context
- Current enhanced context (note, to this point, we are basically like "new chat" context.)
- Existing transcript *and* tagged context, prioritizing recent messages (that is, previous bot response is the most important, and tagged context is likely to be included.)

Would be cool to try having the LLM giving feedback along with its response about which context sources it found useful, and boosting the priorities of those in the replayed transcript (@rishabhmehrotra has mentioned wanting one-shot learning in the transcript, this is a little bit like that...) and obviously summarizing the existing transcript instead of truncating it.

dominiccooney:
I have filed #2769 for more forward-looking changes to truncation and summarization.

kalanchan:
Closing as we addressed this in multi turn chat, and working on improvements to surface context limits

PR #2767, opened 2024-01-17
fix: remove context limit warnings from UI

abeatrix:
The ContextWindowLimitError banner was being displayed in the Chat UI when context limit warnings occurred, causing confusion to users. The PR updates to log the context limit warning message instead to provide useful debugging info, instead of sending it to the webview to display as an error, which will then be rendered as the ContextWindowLimitError banner.

## Test plan

Ask Cody multiple questions to confirm no ContextWindowLimitError banner was displayed in the Chat UI.

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -6,7 +6,6 @@ import * as vscode from 'vscode'
 import {
     ChatModelProvider,
     ConfigFeaturesSingleton,
-    ContextWindowLimitError,
     FeatureFlag,
     hydrateAfterPostMessage,
     isCodyIgnoredFile,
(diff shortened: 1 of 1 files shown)


dominiccooney (changes requested):
Good, but I think we need more here:

- We should look at all the places generating context warnings and decide whether we want to keep the warning (tagging a file that's too large may be one we want to keep, for example) or be silent and do something "best effort" (or at least better effort.) We should get @chillatom and @beyang's input/awareness of these.
- We need to make that "better effort" thing happen. For example dropping old automatically fetched context is better than dropping new explicitly tagged context.
- After all that, *if* we're not using ContextWindowLimitError any more, we can remove all the downstream UX widgets for displaying it.

abeatrix:
see https://github.com/sourcegraph/cody/issues/2766 for details

PR #2768, opened 2024-01-17
Make new chats the default and limited context window allocated to enhanced context to preserve context window

beyang:
From the `vscode/CHANGELOG.md`:
- New chats are now the default when the user submits a new quesetion. Previously, follow-up questions were the default, but this frequently led to exceeding the LLM context window, which users interpreted as an error state. Follow-up questions are still accessible via ⌘-Enter or Ctrl-Enter.
- We now allocate no more than 60% of the overall LLM context window for enhanced context. This preserves more room for follow-up questions and context.

image

## Test plan

- [ ] Check out branch locally
- [ ] Verify new chat is the default behavior
- [ ] Check cmd-enter and ctrl-enter to see that this results in asking a follow-up question

### DIFF
--- a/lib/ui/src/Chat.tsx
+++ b/lib/ui/src/Chat.tsx
@@ -156,7 +156,7 @@ export interface UserContextSelectorProps {
     setSelectedChatContext: (arg: number) => void
 }
 
-export type ChatSubmitType = 'user' | 'suggestion' | 'example' | 'user-followup'
+export type ChatSubmitType = 'user' | 'suggestion' | 'example' | 'user-newchat'
 
 export interface ChatModelDropdownMenuProps {
     models: ChatModelProvider[]
@@ -350,7 +350,7 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
             // Submit chat only when input is not empty and not in progress
             if (formInput.trim() && !messageInProgress) {
                 setInputRows(1)
-                submitInput(formInput, isFollowUp ? 'user-followup' : 'user')
+                submitInput(formInput, isFollowUp ? 'user' : 'user-newchat')
                 setFormInput('')
             }
         },
--- a/lib/ui/src/Chat.tsx
+++ b/lib/ui/src/Chat.tsx
@@ -155,7 +155,7 @@ export interface UserContextSelectorProps {
     setSelectedChatContext: (arg: number) => void
 }
 
-export type ChatSubmitType = 'user' | 'suggestion' | 'example'
+export type ChatSubmitType = 'user' | 'suggestion' | 'example' | 'user-followup'
 
 export interface ChatModelDropdownMenuProps {
     models: ChatModelProvider[]
@@ -343,14 +343,20 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
         [inputHandler]
     )
 
-    const onChatSubmit = useCallback((): void => {
-        // Submit chat only when input is not empty and not in progress
-        if (formInput.trim() && !messageInProgress) {
-            setInputRows(1)
-            submitInput(formInput, 'user')
-            setFormInput('')
-        }
-    }, [formInput, messageInProgress, setFormInput, submitInput])
+    const onChatSubmit = useCallback(
+        (isFollowUp: boolean): void => {
+            // Submit chat only when input is not empty and not in progress
+            if (formInput.trim() && !messageInProgress) {
+                setInputRows(1)
+                submitInput(formInput, isFollowUp ? 'user-followup' : 'user')
+                setFormInput('')
+            }
+        },
+        [formInput, messageInProgress, setFormInput, submitInput]
+    )
+    const onChatButtonSubmit = useCallback((): void => {
+        onChatSubmit(false)
+    }, [onChatSubmit])
 
     const onChatKeyDown = useCallback(
         (event: React.KeyboardEvent<HTMLElement>, caretPosition: number | null): void => {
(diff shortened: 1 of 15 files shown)


chillatom:
I'm having trouble with Loom at the moment, but while testing I ran into a bug.  **Clicking in the chat history panel fails to open the old chat.** 

To replicate
1. Start a chat (Chat 1).
2. Type a new question (Chat 2) in the same chat box and hit enter (triggers new chat)
3. Go to chat history panel and see the prompt from chat #1, click 
4. ChatPanelsManager: restorePanel is fired, but it seems to open the existing chat window (e.g. Chat 2)

abeatrix (commented):
Left a comment inline. Looks like the e2e tests are failing because of the updated keybinding for submitting a question, and I'm also seeing the issue Corey mentioned, but the rest works as described for me. 

The new buttons look similar to what Tim has in his design, should we update the styles to match the ones in his design (in a follow-up PR)? 

image



After this PR is merged, do we still want to keep the context limit warning from the UI? If yes, I can close my PR https://github.com/sourcegraph/cody/pull/2767

philipp-spiess:
Nice bandaid! One minor nit (in addition to the history being bugged): We should probably update the submit icon. Perhaps even make it a `+` or so unless the cmd modifier is pressed end than change it to what we have right now (so that there's a clear visual feedback of what will happen when you press enter without reading the fine print)

Screenshot 2024-01-17 at 11 37 32

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -661,7 +661,12 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
             : inputText
         // The text we will use to send to LLM
         const promptText = command ? [command.prompt, command.additionalInput].join(' ')?.trim() : inputText
-        this.chatModel.addHumanMessage({ text: promptText }, displayText)
+        if (submitType === 'user-followup' || this.chatModel.isEmpty()) {
+            this.chatModel.addHumanMessage({ text: promptText }, displayText)
+        } else {
+            this.chatModel = new SimpleChatModel(this.chatModel.modelID)
```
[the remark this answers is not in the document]
abeatrix: This change should resolve the issue @chillatom reported after this PR is merged: https://github.com/sourcegraph/cody/pull/2786

beyang:
> I'm having trouble with Loom at the moment, but while testing I ran into a bug. Clicking in the chat history panel fails to open the old chat.

This is actually a bug with history when chats are reset which you can reproduce on `main` by clicking "reset chat", starting a new conversation, and then trying to restore the old chat from the history panel. It is now fixed in d949d16e2461698abf1b05b4b8b36eb8ad7bf4d7 and 456230becbb9dd0d2ae6f7a08661cba56e3ba887


> The new buttons look similar to what Tim has in his design, should we update the styles to match the ones in his design (in a follow-up PR)?

+1 to Tim's design, that seems like a reasonable thing to work toward.

> After this PR is merged, do we still want to keep the context limit warning from the UI? If yes, I can close my PR #2767

I'd be in favor of keeping the context limit warning, but making it look like less of an error and more like a notice—and presenting a CTA in it to reset the chat.

beyang:
*Note: don't merge until [greenlight from JetBrains folks](https://sourcegraph.slack.com/archives/C059N5FRYG3/p1705539191033299)*

lib/ui/src/Chat.tsx
```diff
@@ -343,14 +345,20 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
...
+            if (formInput.trim() && !messageInProgress) {
+                setInputRows(1)
+                submitInput(formInput, isFollowUp ? 'user-followup' : 'user')
+                setFormInput('')
+            }
+        },
+        [formInput, messageInProgress, setFormInput, submitInput]
+    )
+    const onChatButtonSubmit = useCallback((): void => {
+        onChatSubmit(false)
```
abeatrix, line 360: imo manual button click should be default to sending follow-up questions since it's next to the chat box

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -279,10 +281,12 @@ export class ChatPanelsManager implements vscode.Disposable {
         try {
             logDebug('ChatPanelsManager', 'restorePanel')
             // Panel already exists, just reveal it
-            const provider = this.panelProvidersMap.get(chatID)
+            const provider = this.panelProviders.find(p => p.sessionID === chatID)
             if (provider) {
-                provider.webviewPanel?.reveal()
-                return provider
+                if (provider.sessionID === chatID) {
```
abeatrix, line 286:
```suggestion
                if (provider?.sessionID === chatID) {
```
so we can remove the `if (provider) {` above?

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -312,12 +316,12 @@ export class ChatPanelsManager implements vscode.Disposable {
             this.disposeProvider(activePanelID)
         }
         // loop through the panel provider map
-        const panelsProvider = Array.from(this.panelProvidersMap.values())
+        const panelsProvider = Array.from(this.panelProviders.values())
```
abeatrix, line 319:
```suggestion
        const panelsProvider = Array.from(this.panelProviders)
```

do we still need Array.from if this.panelProviders is now an array?

abeatrix (approved):
left comments/questions in line. confirmed the new chat submission behavior works and the chat history sidebar issue is resolved 👍

olafurpg:
Shared some thoughts on this design here https://sourcegraph.slack.com/archives/C059N5FRYG3/p1705566202325319?thread_ts=1705539191.033299&cid=C059N5FRYG3

lib/ui/src/Chat.tsx
beyang: I've changed the icon to a single chat bubble. If the user hits "command/ctrl", then the icon changes to show a multi-bubble icon that indicates a follow-up question.

PR #2772, opened 2024-01-17
fix testing vscode.Uri.joinPath(/* vscode-uri.URI */, ...)

sqs:
If the first arg was a `vscode-uri.URI` type, then it lacks a `.uri` property, and this would fail.

## Test plan

CI


abeatrix (approved):
Thank you!

PR #2774, opened 2024-01-17
remove support for codebase-specific ignores

sqs:
Workspace ignores (`.cody/ignore` files at the root of a workspace) are still supported. Previously, we also stored the mapping of a workspace ignore file to the codebases it applied to, but we never look up by codebase anymore (after https://github.com/sourcegraph/cody/pull/2692), so the mapping can be removed. There is no loss or change in functionality.

(I missed deleting this code in that PR.)

## Test plan

CI


vscode/src/services/context-filter.ts
```diff
@@ -68,37 +65,17 @@ async function refresh(uri: vscode.Uri): Promise<void> {
...
-            }
-            return {
-                filePath: fileUri.fsPath,
-                content: await tryReadFile(fileUri),
-                codebase,
-            }
-        })
+    const filesWithContent: IgnoreFileContent[] = await Promise.all(
+        ignoreFiles.map(async fileUri => ({
+            filePath: fileUri.fsPath,
```
abeatrix, line 72: Does it matter if we use path vs fsPath here?

abeatrix (approved):

vscode/src/services/context-filter.ts
sqs: In this case the code is still expecting FS paths, but in `sqs/multiroot` I have another commit that I'll send up soon that converts it to use URIs. I wanted to make this the smallest possible change for this PR.

PR #2778, opened 2024-01-17
remove e2e, e2e-inspector, Transcript, most of createClient

sqs:
Previously the primary consumer of these was the e2e test. OldChatQuestion was only used in the transcript test. There are some transcript tests that may still be useful that were deleted, but they were heavily reliant on an old and decaying implementation (OldChatQuestion), so they were low value. Transcript is still used by the inline edit workflow.

## Test plan

CI


sqs:
@olafurpg This removes stuff from `createClient` that agent was not using. Maybe your new branch will remove the dependency on `createClient` altogether since it's in shared. Feel free to move your branch ahead and let this one wait if that would be easier for you, just let me know.

olafurpg (approved):
LGTM wrt. agent changes. I have a `goodbye-recipes` branch that removes the same editor.ts file and `createCompletionsClient` usage, and I expect that branch to be ready for merging later today.

sqs:
from @rishabhmehrotra: "we are not currently using it, and evaluator would exist outside Cody main repo for now"

PR #2780, opened 2024-01-17
[Cody] remove cody-pro feature flag dependency

thenamankumar:
Similar to: https://github.com/sourcegraph/sourcegraph/pull/59660

Remove dependency on cody-pro feature flag which was added during Dec GA release. Things are now stable and it is a good time to cleanup.

## Test plan
- Connect to a local sourcegraph instance
- Visit `/site-admin/feature-flags` and remove the cody-pro feature flag.
- Visit `/cody/manage` and make sure you are on a free plan.
- Check in IDE that Upgrade buttons are showing up.

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -396,7 +396,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                 await localStorage.set('model', message.model)
                 break
             case 'get-chat-models':
-                await this.postChatModels()
+                this.postChatModels()
                 break
             case 'getUserContext':
                 await this.handleContextFiles(message.query)
@@ -487,7 +487,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
             chatID: this.sessionID,
         })
 
-        await this.postChatModels()
+        this.postChatModels()
         await this.saveSession()
         await this.postCodyCommands()
         this.initDoer.signalInitialized()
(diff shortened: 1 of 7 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -556,14 +556,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         if (authStatus?.configOverwrites?.chatModel) {
             ChatModelProvider.add(new ChatModelProvider(authStatus.configOverwrites.chatModel))
         }
-        // selection is available to pro only at Dec GA
-        const isCodyProFeatureFlagEnabled = await this.featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyPro)
-        const models = ChatModelProvider.get(authStatus.endpoint, this.chatModel.modelID)?.map(model => {
-            return {
-                ...model,
-                codyProOnly: isCodyProFeatureFlagEnabled ? model.codyProOnly : false,
```
philipp-spiess, line 564: Can we remove this flag in the model list too?

philipp-spiess (commented):
LGTM but let's also add a case to the test plan when connected to a non dotcom instance (e.g. s2) to make sure we don't bring some of the cody pro functionality to enterprise by mistake

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
thenamankumar: It was confusing to me too, but here `codyProOnly` means if the model is only available on the pro plan. It has nothing to do with the feature flag but the user's subscription.

philipp-spiess: Ohh gotcha!

philipp-spiess (approved):

kalanchan:
thanks @thenamankumar for helping us cleanup!

@abeatrix @DanTup could you help review this?

PR #2781, opened 2024-01-17
Edit: Use `svg` for ghost text, include icons

umpox:
Horribly hacky implementation of using an `SVG` as the ghost text.

Attempts to adjust the SVG based on the text styling in the editor.

I believe for this to be fully accurate we would need to implement a hack that opens an invisible webview, gets additional font settings out of that webview (e.g. line height), and returns them to the editor.

image

Some prior art of doing this here: https://sourcegraph.com/github.com/cursorless-dev/cursorless

## Test plan

1. Make a selection
2. View ghost text


umpox:
@philipp-spiess Real PR is merged here: https://github.com/sourcegraph/cody/pull/2611 :D 

(Unless you like my hacky SVG creation lol)

philipp-spiess:
I liked the SVG creation 🥺

PR #2782, opened 2024-01-17
Agent: remove recipes, fix authentication bugs and improve logging

olafurpg:
I started the PR with a small ambition of just removing support for recipes and direct imports on `lib/shared/`  but the PR only got bigger and bigger as I discovered more and more bugs in the implementation. I had to fix these bugs so I could document with integration tests that my removal was not introducing regressions. I ended up fixing at least a dozen bugs including nasty ones where the agent would fail to initialize or silently hide authentication errors. Along the way, I made the HTTP recording infrastructure more robust (it now doesn't replay responses if you change the access token), and lots of other long overdue code cleanups (most notably: `logDebug`/`logError` are now available in lib/shared among). It's not ideal to bundle so many changes in one PR but it would have been impractical to split it up and rebase multiple PRs with all the code cleanups happening in parallel this week.

New agent JSON-RPC endpoints:

- `extensionConfiguration/change`: request that does the same as `extensionConfiguration/didChange` but returns a value to indicate if the credentials authenticated successfully
- `extensionConfiguration/status`: request that returns the authentication status of the agent

## Test plan
See updated tests.


olafurpg:
Turning on auto-merge after doing a self review. I don't want to deal with more rebases 😅

sqs (approved):

PR #2783, opened 2024-01-17
Autocomplete: Make sure to not crash when the site config was changed

philipp-spiess:
This is an attempt to fix a potential rough experience for our customers:

In the scenario where a server admin changes the completion model from the default to `fireworks/starcoder`, we don't have a way for the clients to be notified about this change. I think it's reasonable to require clients to reload until those changes take effect.

The issue though, is that previously in the Anthropic code path, we never included a `model` identifier in the requests. This means that when the server receives the code completion request, it will fall back to the _default_ that is configured - This has changed, though. That will mean the Anthropic prompt will be sent to the new Fireworks provider which is obviously not going to work (alone for the fact that it's not a chat model).

This is one attempt at fixing this by passing through the model that is defined in the upstream url. The catch? The allow list was updated only a few months ago: https://github.com/sourcegraph/sourcegraph/commit/70b068c4c6d90e88cdc023df964b846986dacb7b which means that for Sourcegraph servers before this allow list, passing a `model` might be coming back with errors instead.

There are two potential ways of fixing this:

1. Instead of fixing it on the client, we can make the server look at the payload. If the prompt looks like the anthropic prompt and no model overwrite is defined, route it to anthropic instead of Fireworks. 
2. Handle the case of a model-not-found error on the client and, if throw, back off from this and pass `undefined` through again.

I think I'll go ahead with 2) since avoids cases where admins never set up Anthropic and yet for some reason the prompt is set to Anthropic.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -173,36 +173,6 @@ importers:
         specifier: ^5.0.5
         version: 5.0.5
 
-  e2e:
-    dependencies:
-      '@anthropic-ai/sdk':
-        specifier: ^0.5.8
-        version: 0.5.8
-      '@azure/openai':
-        specifier: 1.0.0-beta.5
-        version: 1.0.0-beta.5
-      '@sourcegraph/cody-shared':
-        specifier: workspace:*
-        version: link:../lib/shared
-      chalk:
-        specifier: ^5.3.0
-        version: 5.3.0
-      commander:
-        specifier: ^10.0.1
-        version: 10.0.1
-      node-fetch:
-        specifier: ^3.3.2
-        version: 3.3.2
-
-  e2e-inspector:
-    dependencies:
-      '@sourcegraph/cody-e2e':
-        specifier: workspace:*
-        version: link:../e2e
-      classnames:
-        specifier: ^2.3.2
-        version: 2.3.2
-
   lib/icons:
     dependencies:
       svgtofont:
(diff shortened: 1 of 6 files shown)


vscode/src/completions/providers/create-provider.ts
```diff
@@ -91,7 +91,11 @@ export async function createProviderConfig(
                 })
             case 'aws-bedrock':
             case 'anthropic':
-                return createAnthropicProviderConfig({ client })
+                return createAnthropicProviderConfig({
+                    client,
+                    // Only pass through the upstream-defined model if we're using Cody Gateway
+                    model: codyLLMSiteConfig.provider === 'sourcegraph' ? codyLLMSiteConfig.completionModel : undefined,
```
philipp-spiess, line 97: To the cody strat team looking: I made it so that we only pass-through a model if we're using Cody Backend in the background so this should not affect any of the strat customers _I think_.

pnpm-lock.yaml
```diff
@@ -173,36 +173,6 @@ importers:
         specifier: ^5.0.5
         version: 5.0.5
 
-  e2e:
```
philipp-spiess, line 176: Not sure what was going on here but `pnpm install` removed all this from my pnpm lock file?

vscode/src/completions/providers/create-provider.ts
chwarwick: I agree, all of the BYOK backend providers ignore the client passed model and use the value from the site config. 

So if I follow correctly this addresses the current default being `provider=sourcegraph` `model:claude-instant-1.1`.  

Now if an admin updates the config to `provider=sourcegraph` and `model=fireworks/starcoder` and the clients haven't reloaded to pick up that configuration change it will send `model=anthropic/claude-instant-1.1` and the sourcegraph provider allows clients to override the model, so requests for the previous model will work (and in theory the user could pick any allowed model supported by the gateway.

chwarwick (approved):
cc @slimsag  just as something to be aware of considering admin controls, I don't expect that this behavior would be problematic.  Updating the allowed models still gives an pathway to immediately cut off access.

vscode/src/completions/providers/unstable-openai.ts
```diff
@@ -31,7 +31,7 @@ const MULTI_LINE_STOP_SEQUENCES = [CLOSING_CODE_TAG]
 const SINGLE_LINE_STOP_SEQUENCES = [CLOSING_CODE_TAG, MULTILINE_STOP_SEQUENCE]
 
 const lineNumberDependentCompletionParams = getLineNumberDependentCompletionParams({
-    singlelineStopRequences: SINGLE_LINE_STOP_SEQUENCES,
+    singlelineStopSequences: SINGLE_LINE_STOP_SEQUENCES,
```
valerybugakov, line 34: TY @philipp-spiess 👀

philipp-spiess: Hehe

PR #2784, opened 2024-01-17
refactor: chat action command

abeatrix:
The `cody.action.chat` command has been using `executeCommand` for chat questions that turns the question into Chat before it can be processed by the Simple Chat Panel as a question.

This PR adds a new `executeChat` method in ChatManager to handle chat questions directly to simplify the process.

## Test plan

The `cody.action.chat` is being used by `Ask Cody to Explain` in CodeAction, so you can invoke that to make sure it works to confirm the changes in this PR.

https://github.com/sourcegraph/cody/assets/68532117/548be375-6351-45b3-9e1d-4aa9e69f1cac


umpox (approved):
Nice!

PR #2798, opened 2024-01-18
Agent attribution API

cbart:
Part of  sourcegraph/sourcegraph#59335

Add attribution API to the agent which can be invoked by IntelliJ for chat.

## Test plan

- [ ] Manual with IntelliJ extension
- [x] Automated test


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -112,6 +112,21 @@ export type Requests = {
     // Returns the current authentication status without making changes to it.
     'extensionConfiguration/status': [null, AuthStatus | null]
 
+    // Returns the result of an attribution search for given snippets.
+    // The result ordering DOES NOT MATCH the request ordering.
+    // However each result contains the verbatim snnippet from the request.
```
olafurpg, line 117:
```suggestion
    // However each result contains the verbatim snippet from the request.
```

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -112,6 +112,21 @@ export type Requests = {
     // Returns the current authentication status without making changes to it.
     'extensionConfiguration/status': [null, AuthStatus | null]
 
+    // Returns the result of an attribution search for given snippets.
```
olafurpg, line 115: Can we add a brief explanation what "attribution search" is? This PR is the first time I'm hearing about it

olafurpg (approved):
LGTM 👍🏻 Left a couple of non-blocking comments

agent/src/agent.ts
```diff
@@ -589,6 +591,30 @@ export class Agent extends MessageHandler {
         this.registerAuthenticatedRequest('featureFlags/getFeatureFlag', async ({ flagName }) => {
             return featureFlagProvider.evaluateFeatureFlag(FeatureFlag[flagName as keyof typeof FeatureFlag])
         })
+
+        this.registerAuthenticatedRequest('attribution/search', async ({ snippets }) => {
```
chwarwick, line 595: Why do the additional extensions need to request the attribution search for guardrails? Shouldn't it be something that is already emitted as the result of a chat or autocomplete?

cbart: I'm open to ideas!

In IntelliJ we trigger attribution async after snippet shows whole. Because communication between webview and extension is fire&forget, we match the messages into a promise.

- For autocomplete indeed we might be able to wire in there, and I wonder when we do, whether it won't just work for IntelliJ.
- For chat we can do something more similar to VS Code. Then on the IntelliJ we would need to mimic the same synchronization (as we'll just get a notification from agent that attribution is here, I guess?) It seemed to me though that actually invoking a JSON RPC is a simpler mechanism, and then we avoid all the unnecessary complexity of webview <-> extension communication. But again, I'm not an expert here, and I'm all ears.

chwarwick: Ok I think I understand now.  So in vs code the rendering of the a code block is what triggers the guardrails attribution search to run and then it injects it's ui into an element in the code block.  

I had thought was that it would be nice to do this much closer to the api request prior to going to UI component so that it other extensions would inherit it rather than reimplement it.  Although I didn't look into what that would require doing.

vscode/src/jsonrpc/agent-protocol.ts
cbart: Added! Hope it's good!

cbart:
I made some changes - specifically I trigger the attribution op via webview message, and the little hack that Olaf recommended [here](https://sourcegraph.slack.com/archives/C059N5FRYG3/p1705655165645629?thread_ts=1705588365.312009&cid=C059N5FRYG3).

- [ ] I haven't tested this with IntelliJ yet, as I need to take the kids to the dentist now.
- [ ] I'm also aiming to implement an agent test for this in this PR.

Please take another look!

olafurpg (approved):
LGTM 👍🏻 This solution is idiomatic to the architecture we're using to solve other similar problems like multi-repo keyword search.

PR #2799, opened 2024-01-18
Agent: support `new vscode.Selection(number,number,number,number)`

olafurpg:
Previously, local embeddings were enabled for agent clients. We had disabled local embeddings in the tests through an environment variable because they were causing stability issues, but clients like the JetBrains plugin were not setting this environment variable. This PR fixes the problem so that local embeddings are always disabled for all agent clients.

## Test plan
See new tests


pkukielka (approved):
LGTM

PR #2804, opened 2024-01-18
Run embeddings and symf retrieval in parallel and implement basic fusion

philipp-spiess:
This PR is changing the simple chat context engine to run the symf search together with any dense retrieval strategies (local/remote embedding) in parallel and fuse the results. The fusion is currently done naively by allocating up to 80% of the context window for the symf results and the remaining 20% for embeddings.

We will follow this up with a more advanced fusion logic that will also address some issues for Autocomplete. We, however, need to adjust the logic we have for RRF to have a finer granularity before we can do this.

## Test plan

- Added unit test for the fusion code
- Ask questions and look at the context items being picked up - Most importantly it passes the squirrel test.


philipp-spiess:
cc @beyang @dominiccooney I know you both have been thinking about this. Is there any reason right now why we don't race symf and embeddings and combine the results? Any ideas you have maybe that are a bit more thought-through? :D

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+                // TODO(beyang): handle this error better
+                logDebug('SimpleChatPanelProvider.getEnhancedContext', 'searchSymf error', error)
+            }
+        }
+
+        const priorityContext = await this.getPriorityContext(text, searchContext)
+        return priorityContext.concat(searchContext)
+    }
+
+    public async getEnhancedContextFused(text: string): Promise<ContextItem[]> {
```
philipp-spiess, line 182: Main change in this method :) I forked it from the current implementation, so we can properly gate this out.

dominiccooney:
> Is there any reason right now why we don't race symf and embeddings and combine the results?

We should race them, I made a similar change it my WIP branch! Thanks for getting to this!

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+        }
+
+        const retrievers: Promise<{ strategy: string; items: ContextItem[] }>[] = []
+
+        // Get embeddings context if useContext Config is not set to 'keyword' only
+        if (useContextConfig !== 'keyword') {
+            // Query local embeddings (dense)
+            retrievers.push(retrieveWithContext(this.searchEmbeddingsLocal(text), 'local-embeddings'))
+            // Query remote embeddings (dense)
+            retrievers.push(retrieveWithContext(this.searchEmbeddingsRemote(text), 'remote-embeddings'))
```
dominiccooney, line 199: Feel free to drop this branch if it is convenient. Remote embeddings are going away.

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+        }
+        if (this.symf) {
+            // Query symf context (sparse)
+            retrievers.push(retrieveWithContext(this.searchSymf(text), 'symf'))
+        }
+
+        const searchContext = await Promise.all(retrievers)
```
dominiccooney, line 206: I approached this the same way, thinking we want errors to be visible. But we could try to muddle through here if one provider is producing errors and produce *some* context. WDYT?

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+
+        const fusedContext = fuseContext(searchContext)
```
dominiccooney, line 208: 🎉

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+                uri: readmeUri,
+                text: truncatedReadmeText,
+                range: viewRangeToRange(range),
+                source: 'editor',
+            },
+        ]
+    }
+}
+
+async function retrieveWithContext<T>(
```
dominiccooney, line 509: ❤️

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+
+    // Rank the order of documents using reciprocal rank fusion.
+    //
+    // For this, we take the top rank of every document from each retrieved set and compute a
+    // combined rank. The idea is that a document that ranks highly across multiple retrievers
+    // should be ranked higher overall.
+    const fusedDocumentScores: Map<string, number> = new Map()
+    for (const { strategy, items } of searchContext) {
+        items.forEach((item, rank) => {
+            const documentId = item.uri.toString()
```
dominiccooney, line 555: Remote keyword search items have URIs that are very different and include the revision for example. We might need a function that abstracts from items to an "ranking identity" or something.

However, with the way the product is designed right now, we would not try to fuse remote keyword search with other sources, so we can ignore it for now.

If you have ideas I'm interested to hear them though!

vscode/src/chat/SimpleContextProvider.ts
```diff
@@ -0,0 +1,621 @@
...
+    // We start with the highest ranked document and include all retrieved snippets from this
+    // document into the result set, starting with the top retrieved snippet from each retriever
+    // and adding entries greedily.
+    for (const documentId of fusedDocuments) {
+        const resultByDocument = resultsByDocument.get(documentId)
+        if (!resultByDocument) {
+            continue
+        }
+
+        // We want to start iterating over every retrievers first rank, then every retrievers
```
dominiccooney, line 592: Great job with these comments, the clarity is superb.

Fine in a follow-up, but downstream from this process after editor context, etc. is assembled, should we combine items from the same document the way the old context assembly system did? (I don't recall if simplified chats does that but it makes sense to do it... deduplicate overlap, merge adjacent snippets.)

dominiccooney (approved):
Splendid work!

Any new opportunities for unit tests unlocked by this?

Some optional comments and questions inline.

vscode/src/chat/SimpleContextProvider.ts
philipp-spiess: Sounds good, we're in a feature flag anyways so let's remove this and avoid confusion

vscode/src/chat/SimpleContextProvider.ts
philipp-spiess: Kinda happening already here: https://github.com/sourcegraph/cody/pull/2804/files/8a0f8c506e9073966985b6e429e4f2c487dfbe38#diff-b702ed9de10a9380cb3a8e5fa35ac560849a095a7c9d882f15edeed4388bafceR517-R518

It could be more explicit though

vscode/src/chat/SimpleContextProvider.ts
philipp-spiess: Mhhh yeah if we have a ranking identity, we can also easily reuse the same logic for autocomplete. I will do that before merging (no need to have the same behavior implemented slightly different in two places 😬). But I always find it easier to build a working thing first before abstracting.

For now though, the ranking identity would just be `(item) => item.uri.toString()` I reckon. With remote embeddings going away this should be "good enough" for getting some data on it?

vscode/src/chat/SimpleContextProvider.ts
philipp-spiess: Ah, good, I didn't know if the current solution did this too. I think this would be a good second step, but we need to have a way to define an upper bound of items to combine. Otherwise, we might end up with giant snippets that consist of the whole file being merged.

Additionally it also requires to work on the same revision of a file. I think it's fine if we just use the ranges from the retrievers and assume they still work (the llm can smooth over potential line missmatches here) but something to keep in mind too.

vscode/src/chat/SimpleContextProvider.ts
dominiccooney: > For now though, the ranking identity would just be `(item) => item.uri.toString()` I reckon. With remote embeddings going away this should be "good enough" for getting some data on it?

Yep, sounds great. Weak handling of different revisions of context is not a new problem, so getting the abstraction in place, even if it is a stub, is a nice step forward and better than having `string` or `Uri` flying around everywhere.

beyang:
+1 to racing them and then merging results.

I'm not quite as sure about ranking by file and then picking all results from each file.
- What if there are low quality snippets from the file? Those snippets would then be bumped up by association with high-quality snippets. There might be such cases where the selected context set gets worse than the previous behavior of just picking from symf or embeddings.

An alternative proposal would be: (1) Select all context from symf up to a threshold. (2) Then include context from embeddings, discarding any ones with overlap with those from symf
- In this proposal, the context set should be *at least* as good as the context selected from symf alone. And in addition, we have the benefit of ensuring that any context selected by embeddings does not overlap with results from symf.

philipp-spiess:
@beyang Thanks, this is good feedback. Agree that upranking low quality snippets is a failure case here.

> An alternative proposal would be: (1) Select all context from symf up to a threshold. (2) Then include context from embeddings, discarding any ones with overlap with those from symf

I don't understand how symf internally ranks results but if there is a reasonable way to define a threshold (since it would universally apply to all search queries) than this would be good, yeah. Looking at the symf `Result` type, there doesn't appear to be any ranking coefficients that are exported though. Or were you thinking a threshold on the number of results (so that symf)?

philipp-spiess:
> What if there are low quality snippets from the file? Those snippets would then be bumped up by association with high-quality snippets. There might be such cases where the selected context set gets worse than the previous behavior of just picking from symf or embeddings.

We could also change the granularity for the RRF to individual lines or try to somehow combine overlapping ranges into a single document to mitigate this. If I understand correctly, the issue is not that two low ranking snippets from two low ranking files get a slight push but instead that if we have one file with two snippets - one high ranking and one low ranking, the low ranking would be pushed by the rank of the high ranking file. 

Maybe what we need to do is a "normalize" step before we to the fusion? I.e we have a set of ranges that might have overlaps. What we can do as a first step is to normalize these so that all overlapping ranges are combined. Then, the ranges from different retrievers have the same boundaries and we can use a finer grainer document key. The issue with this is that if we have retrievers that don't have thresholds, we might be getting snippets for _every possible line in the document_. If we try to merge those, we would end up with one range per document again which is not desirable. 

Alternatively we could rank based on _lines_ instead of whole documents. We need a bit more memory for the map we create but then we could rank based on lines occurring more often throughout retrievers and start by including all snippets that the highest ranked lines.

philipp-spiess:
@dominiccooney @beyang Updated this to be a bit more lightweight for now. In the first pass, I want to just do a top-k picking of the top context items of each retriever. I'll do some more testing tomorrow but I think this should be a relatively small change to get us going.

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -222,32 +224,32 @@ function renderContextItem(contextItem: ContextItem): Message[] {
 }
 
 /**
- * PromptBuilder constructs a full prompt given a byteLimit constraint.
```
philipp-spiess, line 225: Moved from the bytes terminology to chars. I found bytes not fitting well because all JS engines use UTF-18 so one character is two bytes + there are unicode composite characters that might be even more bytes 😬  Since we use `.length` everywhere, I think this is a bit clearer

philipp-spiess:
@dominiccooney I tagged you for another quick round of reviews since the PR has diverged quite a bit from the original implementation.

dominiccooney (approved):
Looks good. @beyang mentioned an n/(1-n) split so this seems in line with that and fine.

If document-based scores aren't enough, it would be nice to pursue the fine-grained item fusing. (I must admit, I am still confused about when theories vs anecdata vs data are necessary to justify action here...)

philipp-spiess:
@dominiccooney 

> If document-based scores aren't enough, it would be nice to pursue the fine-grained item fusing. (I must admit, I am still confused about when theories vs anecdata vs data are necessary to justify action here...)

You're going to like this: https://github.com/sourcegraph/cody/pull/2898/commits/9fe4db456fd4298c37545476821aaf20cbe1a48f :) (I’m going to fix this for autocomplete first since there are some issues related to existing context window merge logics that are in line with the issues beyang mentioned here.

PR #2739, opened 2024-01-15
hold: display edit buttons on all human messages

abeatrix:
close https://github.com/sourcegraph/cody/issues/2619

## This PR is outdated - moving to a new branch

The goal of this PR is to address the following items described in the design figma:

image

image

This PR includes the following changes:
- display edit button on each human message
- update the `handleEdit` function in SimpleChat to support commands and regular chat input properly
  - instead of only able to edit the last human message, user can now choose any human messages in their chat session to edit
  - also updated to support `/commands` in edited messages 

New tests added to cover the new changes
- update tests for the edit function in agent 
- add new e2e test for the updated user journey for editing chat messages

## Follow-up

- the new design update for the edit text area https://github.com/sourcegraph/cody/issues/2762

image

## Test plan

1. Build from this branch
2. in debug mode, ask Cody multiple questions, including using chat commands (e.g, /explain)
  - confirm edit button shows up on each message submitted when no message is in progress
3. click on an edit button to start the editing process
  - confirm the edit text area only shows up for the message you clicked the edit button on
4. update the message to something else and then submit the edit
  - confirm all the messages after the edited messages are removed (similar to how editing old messages work in chatGPT)
  - confirm the previous message has been replaced by the updated message
  - confirm cody is responding to your updated message when replying

Added new tests to agent to cover this updated behavior for edit. All tests are passing locally:

image

## Demo

https://github.com/sourcegraph/cody/assets/68532117/4a1b1cc5-f87f-4e2e-a491-412de82120b4

PR #2808, opened 2024-01-19
feat: keybinds for chat actions, change default chat submit

abeatrix:
Introduced keyboard shortcuts for editing the last sent message and resetting chat.

Updated current keybindings:
- from "MetaKey" + "Enter" to "Enter" for submitting follow up chat question in current chat session.
- from "Enter" to  "MetaKey" + "Enter"  for submitting chat message in a new chat session.

New keybindings when chat input is focused:
- `ESC`: abort current message in progress
- `MetaKey/Clt` + `/`: Reset current chat
- `MetaKey/Clt` + `E`: Edit and retry

Commit message created by Cody:

- Introduce ChatActions component import and usage for better chat action management.
- Modify EditButtonProps to accept an optional number for messageBeingEdited, replacing the boolean.
- Add new state hooks for messageBeingEditedIndex and sendAsFollowUp to manage editing and follow-up message states.
- Implement logic for editing the last human message and resetting chat with new keyboard shortcuts.
- Remove redundant isFollowUpKeyPressed state in favor of sendAsFollowUp.
- Adjust Transcript and TranscriptItem components to work with the new messageBeingEditedIndex state.
- Update related event handlers and callbacks to accommodate the new editing and follow-up message logic.
- Clean up unused variables and comments, and refactor for better readability and maintainability.

## Test plan

 e2e test for edit and chat should pass


lib/ui/src/Chat.tsx
```diff
@@ -308,6 +316,43 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
         [ChatCommandsComponent, chatCommands, filterChatCommands]
     )
 
+    const lastHumanMessageIndex = useMemo<number | undefined>(() => {
+        if (!transcript?.length) {
+            return undefined
+        }
+        const index = transcript.at(-1)?.speaker === 'human' ? -1 : -2
```
dominiccooney, line 323: We've hardcoded the constraint/assumption that human, assistant messages are interleaved. But it would it be simpler and more reliable to just loop backwards and find the last human message. If you fall out of the loop without finding a human message, you can return `undefined`. That way you don't need the special case for the transcript being 0 length. And you'll handle the case of a transcript with one assistant message correctly, which you don't today.

lib/ui/src/Chat.tsx
```diff
@@ -308,6 +316,43 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
...
+        return transcript.length + index
+    }, [transcript])
+
+    const onSetEditMessageIndexClick = useCallback(
+        (index?: number): void => {
+            if (messageInProgress || !transcript.length) {
+                return
+            }
+            if (index === undefined) {
+                setMessageBeingEdited(false)
```
dominiccooney, line 333: Why do we have two variables here? What would it mean if messageBeingEdited is true but messageBeingEditedIndex is undefined? Let's avoid this invalid states by eliminating them: Just use a number index to mean that message is being edited, and undefined to mean nothing is being edited.

lib/ui/src/Chat.tsx
```diff
@@ -308,6 +316,43 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
...
+            setMessageBeingEdited(true)
+        },
+        [messageInProgress, setMessageBeingEdited, transcript.length]
+    )
+
+    const onHandleChatReset = useCallback(() => {
+        if (postMessage) {
+            postMessage({ command: 'reset' })
+        }
+        // Hack: update row number to trigger autoFocus in textarea
```
dominiccooney, line 350: What autofocus is being triggered by this and how? Why not just call "focus" if you want to focus something?

lib/ui/src/Chat.tsx
```diff
@@ -308,6 +316,43 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
...
+            }
+            setMessageBeingEditedIndex(index)
+            setMessageBeingEdited(true)
+        },
+        [messageInProgress, setMessageBeingEdited, transcript.length]
+    )
+
+    const onHandleChatReset = useCallback(() => {
+        if (postMessage) {
+            postMessage({ command: 'reset' })
```
dominiccooney, line 348: You don't need the `if` statement; you can use `f?.(args)` syntax to do this more succinctly.

lib/ui/src/Chat.tsx
```diff
@@ -358,34 +401,51 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
...
-            if (event.metaKey && event.key === 'k') {
-                onSubmit('/r', 'user', chatContextFiles)
+            if (event.ctrlKey || event.metaKey) {
+                // Clear & reset session on CMD+/
+                if (event.key === '/') {
+                    event.preventDefault()
+                    onHandleChatReset()
+                    return
+                }
+                // Edit last human message
```
dominiccooney, line 437: Ctrl-e is "end of line" on macOS and X with emacs keybindings, so that's going to mess up some people's editing. Where is this bound? Does ctrl-e still work for end-of-line while you're composing or editing with this?

lib/ui/src/Chat.tsx
```diff
@@ -664,7 +732,7 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
                         )}
                     </div>
                     <SubmitButton
-                        type={isFollowUpKeyPressed ? 'follow-up' : 'chat'}
+                        type={sendAsFollowUp ? 'follow-up' : 'chat'}
```
dominiccooney, line 735: IIRC @olafurpg and @rrhyne had some feedback in Slack about the impact of follow-up vs chat messages on JetBrains/chat UI affordances. I wasn't following it closely though; are you across it?

lib/ui/src/chat/TranscriptItem.tsx
```diff
@@ -103,17 +103,17 @@ export const TranscriptItem: React.FunctionComponent<
...
-    // To identify if the current message is the one being edited when beingEdited is true
-    // This is used to display EditTextArea only for the message being edited
-    const [isItemBeingEdited, setIsItemBeingEdited] = useState<boolean>(false)
+    // To identify if the current message is the one being edited when beingEdit is set
+    // This is used to display EditTextArea only for the message that is being edited
+    const hasItemBeingEdited = beingEdited !== undefined
+    const isItemBeingEdited = beingEdited === index
 
-    const onEditCurrentMessage = useCallback(
+    const setItemAsBeingEdited = useCallback(
```
dominiccooney, line 111: What happens if you start editing more than one message at a time?

I feel like I've given this feedback before on #2739 so I'm going to stop reviewing at this point and give you a chance to integrate feedback.

dominiccooney (changes requested):

lib/ui/src/Chat.tsx
abeatrix: That change has already been implemented by Beyang. 

I talked to @toolmantim  and we think the new design should work fine with reusing what Beyang added (new buttons for restarting chat or edit chat). 

Will put more details in the new issue 👍

abeatrix:
@dominiccooney hey Dom im so sorry i should have closed this PR as the design has changed and a lot of the changes in this PR has become irrelevant (we no longer need a separate input box for editing the message). Will tag you in the new pr with more details when it's ready. So sorry about that 🙇🏻‍♀️

PR #2809, opened 2024-01-19
Cleanup: Put "is enhanced context enabled" in one place instead of passing it around

dominiccooney:
We're using React context for enhanced context settings instead of passing it around. This is more succinct and obviously consistent because the context is provided in far fewer places (specifically, one place) than it would be if it were passed around.

Moves the definition of the context into lib/ui so it is accessible from the lib/ui loading indicator, and cleans up passing the value around in some places and using context in others.

## Test plan

CI, this is not a functional change.


philipp-spiess (approved):
Good cleanup!

PR #2810, opened 2024-01-19
Add squirrel test to agent test suite

beyang:
Add squirrel test to agent test suite. Also fixes login error which caused unauthorized errors when running single tests.

## Test plan

Test-only

`npx vitest ./src/index.test.ts -t 'squirrel test'`


agent/src/index.test.ts
```diff
@@ -363,12 +363,24 @@ describe('Agent', () => {
             accessToken: 'sgp_INVALIDACCESSTOK_ENTHISSHOULDFAILEEEEEEEEEEEEEEEEEEEEEEE2',
         })
         expect(serverInfo?.authStatus?.isLoggedIn).toBeFalsy()
+
+        // Log in so test cases are authenticated by default
+        const valid = await client.request('extensionConfiguration/change', {
```
olafurpg, line 368: Thank you for adding this! I've been meaning to do it myself 😅 I was also planning to document that `CODY_KEEP_UNUSED_RECORDINGS=true` is useful when you're running tests with `it.only` mode.

olafurpg (approved):
This is great!

PR #2815, opened 2024-01-19
VS Code enables attribution based on site config

cbart:
Part of [#59427](https://github.com/sourcegraph/sourcegraph/issues/59427)

This pull request:

- extends config features GraphQL query with attribution
- extends webview-extension chat-enabling messaging to `setConfigFeatures` that includes chat and attribution.

## Test plan

- Added e2e tests
- Manual
- https://www.loom.com/share/a59a18a48d914062a21489fc3c47ef7f?sid=578d1a22-f0ac-4102-8b85-427923c2b345


cbart:
I'd like to add a test here still, please stay tuned!

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -339,20 +340,12 @@ export class SourcegraphGraphQLAPIClient {
...
      */
-    public async getCodyConfigFeatures(): Promise<
-        | {
-              chat: boolean
-              autoComplete: boolean
-              commands: boolean
-          }
-        | Error
-    > {
-        return this.fetchSourcegraphAPI<APIResponse<CodyConfigFeaturesResponse>>(
```
arafatkatze, line 350: Yo thanks for cleaning this up

arafatkatze (approved):
So nice to see 👮  🛤️  UI

This works jolly.

NOTE for testers who wanna try the whole thing
1. Just check for the guardrail batch in UI
image

2. In cody.settings json you gotta enabled  ` "cody.experimental.guardrails": true,`

3. You also gotta do attribution.true in the siteadmin config for local testing 
image


You can do all of that or you can just presume that Cezary and I are telling the truth and this works jolly.

lib/shared/src/guardrails/client.ts
```diff
@@ -2,11 +2,17 @@ import type { SourcegraphGraphQLAPIClient } from '../sourcegraph-api/graphql'
 import { isError } from '../utils'
 
 import type { Attribution, Guardrails } from '.'
+import { ConfigFeaturesSingleton } from '../sourcegraph-api/graphql/client'
 
 export class SourcegraphGuardrailsClient implements Guardrails {
     constructor(private client: SourcegraphGraphQLAPIClient) {}
 
     public async searchAttribution(snippet: string): Promise<Attribution | Error> {
```
philipp-spiess, line 10: unrelated to your pr but returning an error seems a bit odd. Should we throw the error instead?

lib/shared/src/sourcegraph-api/graphql/queries.ts
```diff
@@ -65,7 +65,8 @@ query CodyConfigFeaturesResponse {
             chat
             autoComplete
             commands
-          }
```
philipp-spiess, line 68: How does this work if a server does not have the `attribution` field implemented? Will the whole query fail? 

I saw a similar issue when I connected to `demo.sourcegraph.com` recently where the `CodyConfigFeaturesResponse` query fails and is littering the console. I suppose in that case it's fine since we can gracefully fall back to the defaults if the server has no overwrites but here, the server might have overwrites for the three other features so we can't fully ignore it. Or can we since it's still so early in dev?

philipp-spiess (approved):
Nice!

lib/shared/src/guardrails/client.ts
cbart: I saw it used a couple times here and there. Happy to follow up to clean it up.

https://sourcegraph.com/search?q=repo:^github\.com/sourcegraph/cody%24++|+Error%3E&patternType=standard&sm=1

cbart: https://github.com/sourcegraph/sourcegraph/issues/59758

PR #2816, opened 2024-01-19
Ghost text: Show on command driven selections

umpox:
## Description

We primarily disabled the ghost text in this scenario because command driven selections were quite common, especially when searching text in VS Code - we wanted to avoid spamming the user.

However, we now only show the ghost text when:
1. There is a multiple line selection
2. There is a single line selection but all non-whitespace characters from this line are selected.

Those restrictions mean that we can re-enable the command driven selections without worrying about it spamming the user too much.

This fixes an issue where Vim keybindings would not trigger the ghost text.

## Test plan

With Vim:
1. Enable the `vscodevim.vim` extension
2. Select multiple lines
3. Check the ghost text shows

With search:
1. Search for a single symbol that isn't the only text on a line
2. Check ghost text doesn't show


olafurpg (approved):
LGTM 👍🏻 Worth giving this a try, we can always make it more conservative later.

olafurpg:
Confirmed this is working as expected with vim mode

CleanShot 2024-01-19 at 21 26 55@2x

The decorations are still missing with `columnSelection: true`. It might be worth a separate PR to at least log when this setting is enabled so we can make a decision if it's worth adding custom support for it.

abeatrix (approved):
🚀

PR #2818, opened 2024-01-19
Fix failing Windows integration test

DanTup:
This test is failing on `main` on the Windows bots because the way the expected string is built is not correct for Windows (the path will have a drive letter).

This changes it to just compare the string URIs instead.

## Test plan

CI


abeatrix (approved):
Thank you!

PR #2819, opened 2024-01-19
Fix Anthropic model ID overwrite for old instances

philipp-spiess:
Follow up to #2783

I tested the gracefull fallback in #2783 on the demo instance (demo.sourcegraph.com) and noticed that the error is a different string. Turns out the string was changed in the past and thus the fallback behavior was not working and the insider build broke completions for these instances 😬. 

So I did a git blame on the `codecompletion.go` file to find all variations of this string. Turns out there is a third one that was available too for some time:

- https://github.com/sourcegraph/sourcegraph/commit/f700d84cab215b84dfbc4ae63d9e6134a856b8be
- https://github.com/sourcegraph/sourcegraph/commit/3b53b4e623e978a559e8ae05d71e8550dded4886#diff-3b488ccb5c3c1aa6d139c640704ba6d118ba6737fda7bc4ac077242a188956baR30
- https://github.com/sourcegraph/sourcegraph/commit/5b95c4b98e2890b9dd7bfdce35977d223227259f#diff-3b488ccb5c3c1aa6d139c640704ba6d118ba6737fda7bc4ac077242a188956baL32

Makes me wonder if we should handle this a bit more gracefully? 🤔  Maybe _any_ upstream error should reset this field from being sent, just to be safe?   

## Test plan

- Connect to demo.sourcegraph.com
- Make code completions

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -189,7 +189,10 @@ class AnthropicProvider extends Provider {
                         //
                         // If it happens once, we disable the behavior where the client includes a
                         // `model` parameter.
-                        if (error.message.includes('Unsupported code completion model')) {
+                        if (
+                            error.message.includes('Unsupported code completion model') ||
+                            error.message.includes('Unsupported custom model')
+                        ) {
                             isOutdatedSourcegraphInstanceWithoutAnthropicAllowlist = true
                         }
                     }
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
Ooof glad you caught it. We just ran into an auth issue not long ago that we weren't aware of until someone tried to use cody with demo. Actually grateful to have an instance available that behaves closer to what our customers are using!

PR #2821, opened 2024-01-19
Agent: fix critical bugs related to document synchronization

olafurpg:
This PR fixes two critical bugs that prevented enhanced context from working as expected:

- Changes to the user's selection were not respected. The agent only kept the first selection. This might be a regression from https://github.com/sourcegraph/cody/pull/2048
- Symf results were ignored if they came from files that the user hadn't explicitly opened. Now, the agent reads files from disk on `vscode.workspace.openTextDocument`.

Massive kudos to @pkukielka for spotting the bug related to selections and coming up with a reliable minimized reproduction in his PR #2803.

## Test plan

See updated tests.

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -13,7 +13,7 @@ import { type ChatMessage, type ContextFile } from '@sourcegraph/cody-shared'
 import type { ExtensionMessage, ExtensionTranscriptMessage } from '../../vscode/src/chat/protocol'
 
 import { AgentTextDocument } from './AgentTextDocument'
-import { MessageHandler } from './jsonrpc-alias'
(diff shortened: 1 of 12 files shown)


agent/src/index.test.ts
```diff
@@ -657,6 +673,32 @@ export class Dog implements Animal {
         )
     })
 
+    describe('Text documents', () => {
+        // This test fails when running in replay mode with `it.only`. This seems to happen
```
olafurpg, line 677: I simply was not able to get this test passing in replay mode with `it.only` because there's something non-deterministic. I don't have more time to investigate it right now but I really want to get to the root of this problem. Maybe we don't even need to sort symf results.

pkukielka (approved):
LGTM, thank you very much for fixing this Olaf!

PR #2823, opened 2024-01-19
Fix Ctrl+Enter for follow-on chat on Windows

DanTup:
We have some code that `return`s if you press certain `Ctrl+` combinations. Since we're using `Ctrl+Enter` for follow-on chat on Windows, we need to not exit out here.

## Test plan

- Run `pnpm test:e2e` to run the updated test, or
- Run the extension on Windows
- Send a chat
- Send a second chat with `<enter>` and ensure it's a new chat (replaces the first message)
- Type a third chat and press `<ctrl+enter>` and ensure it's a follow-on chat (appended after the previous message)


abeatrix (approved):
🚀

PR #2825, opened 2024-01-19
Agent: reset server environment between tests

olafurpg:
Followup to https://github.com/sourcegraph/cody/pull/2821

Previously, each agent test performed side effects like opening a document that leaked into the next test case. This caused subtle problems like seemingly non-deterministic behavior based on whether you ran an individual test with `it.only` (or `vitest -t`) or all tests. For example, if one test case customized the LLM model then the next test case would use that custom model when running all test together, but the next test would use the default model when running isolated (via `it.only` or `vitest -t`).

This PR fixes the problem by adding a new `testing/reset` request that does a best-effort to reset the environment of the server process between each test case. The current implementation may not reset everything perfectly but it solved the problem of the LLM default along with a few other low-hanging cases (open documents).

## Test plan

Green CI. Optionally, run the tests locally with `it.only` changes to any test. It should always pass in replay mode now.

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -1333,21 +1333,8 @@ class ContextProvider implements IContextProvider {
             })
             return (await Promise.all(items)).flat()
         })
-        const allResults = (await Promise.all(r0)).flat()
 
-        // Sort results for deterministic ordering for stable tests. Ideally, we
-        // could sort by some numerical score from symf based on how relevant
-        // the matches are for the query.
-        // Feel free to change this behavior to only sort while running tests.
-        allResults.sort((a, b) => {
-            const byUri = a.uri.fsPath.localeCompare(b.uri.fsPath)
-            if (byUri !== 0) {
-                return byUri
-            }
-            return a.text.localeCompare(b.text)
-        })
-
-        return allResults
+        return (await Promise.all(r0)).flat()
     }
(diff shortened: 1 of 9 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1333,21 +1333,8 @@ class ContextProvider implements IContextProvider {
             })
             return (await Promise.all(items)).flat()
         })
-        const allResults = (await Promise.all(r0)).flat()
-
-        // Sort results for deterministic ordering for stable tests. Ideally, we
-        // could sort by some numerical score from symf based on how relevant
-        // the matches are for the query.
-        // Feel free to change this behavior to only sort while running tests.
-        allResults.sort((a, b) => {
```
olafurpg, line 1342: This change from https://github.com/sourcegraph/cody/pull/2821 was not needed. Undoing the diff here now that we've fixed the root cause (LLM model changes got persisted between test runs)

olafurpg: Scratch that. Tests are failing on Windows and Ubuntu 16. I pushed a commit to test if sorting results here is making the tests more stable.

olafurpg: Tests passed after re-enabling sorting so looks like it's needed after all.

valerybugakov (approved):

Issue #2619, opened 2024-01-09
Make "edit" always present on every human message


dominiccooney:
[Figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1171-38718&mode=design&t=pKiZKQ8zL0qfdu1G-0)

Issue #2621, opened 2024-01-09
Add a "New chat" button, shortcut above the chat box


dominiccooney:
[Figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1171-36164&mode=design&t=pKiZKQ8zL0qfdu1G-0)

PR #2826, opened 2024-01-20
feat: chat editings, new keybindings, with updated UI

abeatrix:
context: [figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&mode=design&t=I2qONOmFEGXKaDV5-0))

Close https://github.com/sourcegraph/cody/issues/2619 https://github.com/sourcegraph/cody/issues/2807 https://github.com/sourcegraph/cody/issues/2762 https://github.com/sourcegraph/cody/issues/2621 https://github.com/sourcegraph/cody/issues/2620 https://github.com/sourcegraph/cody/issues/2807

edit

A new UI has been implemented to add edit buttons to all user messages in the chat. This provides easy access to chat actions editing a previous message (where all subsequential messages will be removed). 

The default submit behavior has also been updated to allow users to toggle between submitting a question as a new chat or as a follow-up using the keyboard shortcuts `Ctrl/Cmd + Shift`.

New keyboard shortcuts as shown in the design also been added for these actions and are display above the input box to make them easily discoverable and accessible:

- `Ctrl/Cmd + Shift`: Toggles between submitting chat questions as new chat and as follow-up question
- `Ctrl/Cmd + /`: Clears the chat and starts a new session by calling onChatResetClick()
- `Ctrl/Cmd + K`: When not already editing, edits the last human message 
- `Escape` during editing mode: Exits editing mode if a message is being edited
- `Escape` when a message in progress : aborts a message in progress if one exists

With these UI improvements, users now have more flexibility to manipulate their current conversation. They can easily edit and remove previous messages to refine their questions and context. This enhances users' ability to have natural conversations by giving them more control. The intuitive edit buttons and shortcuts allow them to iterate on questions, clarify intent, and improve the context share with Cody, where staying in the same chat input box they are already familiar with that supports enhanced context toggler, execute commands, and add custom context via @-mention.

Overall, these UI updates empower users to actively shape the conversation to get their needs met. By making message edits and actions more discoverable and convenient, the chat becomes more flexible and tailored to each user's unique conversation flow. This improves the overall experience by letting users take an active role in guiding the LLM assistant.

Added "Return to Previous Chat" button to restore chat history from a previous session.

## Demo

TBC

## Test plan

- See details in the new chat-edits e2e test for the expected behavior and updated user journal
- Also added new integration test to agent for the new edit command

image

Edit buttons available in all human messages

image

Click on the edit icon next to the message to enter edit mode. When in edit mode, all messages will be blurred except the message you are editing. The text from the edit item will be showing up in your chat input box with the "Cancel Edit" button above it. The edit icons will be hidden during edit mode as well.

image

Press ESC or click on the button to exit Edit Mode. You will automatically focus to the input box again. Press Meta+K key to edit your last message

image

Press `Meta+shift` in chat box to toggle between `new chat mode` and `follow up chat mode`

image

image

After you've submitted a message, your focus will be moved to the "Edit Last Message" button, and will be returned back to the input box once the streaming is completed

focus

TBC


agent/src/index.test.ts
```diff
@@ -651,30 +670,51 @@ describe('Agent', () => {
             }
         })
 
-        it(
-            'edits the chat',
-            async () => {
+        // Tests for edits would fail on Node 16 (ubuntu16) possibly due to an API that is not supported
+        describe.skipIf(isNode16())('chat/editMessage', () => {
```
abeatrix, line 674: @olafurpg Just want to let you know i've made some changes to the edit command where it would take an index and remove chat messages from there,  but it should work the same for agent/jetBrains, where when an index is not provided, we will just edit the last human message (which was the previous default behavior). I've updated the test that hopefully would cover this change, but if your team notices any issues please let me know! 

I've also added the isNode16 skip as you shared with me in the other PR :) 

Thank you so much again for helping me with the tests ❤️

lib/ui/src/Chat.tsx
```diff
@@ -240,6 +241,61 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
     const [chatContextFiles, setChatContextFiles] = useState<Map<string, ContextFile>>(new Map([]))
     const [selectedChatContext, setSelectedChatContext] = useState(0)
 
+    // When New Chat Mode is enabled, all non-edit questions will be asked in a new chat session
+    // Users can toggle this feature via "shift" + "Meta(Mac)/Control" keys
+    const [enableNewChatMode, setEnableNewChatMode] = useState(false)
```
abeatrix, line 246: @beyang @toolmantim This updates the current behavior to toggle between submitting a new message on `Enter` in a new chat session vs as follow-up question via the `Meta+Shift` key.
Right now it's defaulted to submit new questions as follow-up questions, but feel free to change it back if this doesn't make sense to you.

vscode/test/e2e/chat-edits.test.ts
```diff
@@ -0,0 +1,134 @@
...
+    await chatInput.press('Enter')
+    await expect(cancelEditButton).not.toBeVisible()
+    await expect(chatInput).toBeEmpty()
+    await expect(chatFrame.getByText('Explain @Main.java')).toBeVisible()
+    await chatInput.press('Escape')
+
+    // "MetaKey(MacOS)/Control" + "Shift" to toggle "New Chat Mode" on and off
+    // When it's on, the submit button will be replaced with "Start New Chat" button
+    await expect(submitMessageButton).toBeVisible()
+    await chatInput.press(`${osKey}+Shift`)
```
abeatrix, line 115: The test works for me locally but I'm guessing this keybinding doesn't work on windows and Ubuntu 🤣 will fix it on Monday.

toolmantim (commented):
Woohoo!

This is just a quick review today while I have the chance…

* Editing the message during the stream (i.e. pressing enter shortly after submitting a message) doesn't focus the input and put the cursor at the end
* For the small button focus, let's not change the keyboard shortcut color
* The transcript content moves around when you start edit mode because the feedback buttons are gone. Can we `visibility: hidden` instead?
* When a response is streaming (and "Edit Last Message" is focused) hitting escape should stop the stream, same as how it works when the message input is focused
* Can we scroll to the transcript item to the top when we trigger edit, if it's off-screen?
* Can we remove the space between the shortcut keys? (i.e. `⌘K` instead of `⌘ K`)
* I'll clean some CSS in a follow up if not this one (small button’s appearance as I'm not sure about this outline style in practice now, editing shadow, etc)

Unrelated probably, but it looks like some of the links in the welcome chat don't work… only "docs".

lib/ui/src/Chat.tsx
```diff
@@ -323,7 +379,7 @@ export const Chat: React.FunctionComponent<ChatProps> = ({
 
     const submitInput = useCallback(
         (input: string, submitType: ChatSubmitType): void => {
-            if (messageInProgress) {
+            if (messageInProgress && submitType !== 'edit') {
```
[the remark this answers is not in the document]
toolmantim: Yep!

lib/ui/src/Chat.tsx
toolmantim: Yeah, I think that makes sense. We should probably update the tooltip on the send button to tell you about those two options and their keyboard shortcut.

abeatrix:
- [x] Editing the message during the stream (i.e. pressing enter shortly after submitting a message) doesn't focus the input and put the cursor at the end
-[x] For the small button focus, let's not change the keyboard shortcut color
-[x] The transcript content moves around when you start edit mode because the feedback buttons are gone. Can we visibility: hidden instead?
- [x] Can we scroll to the transcript item to the top when we trigger edit, if it's off-screen?
- [x] Can we remove the space between the shortcut keys? (i.e. ⌘K instead of ⌘ K)


- [ ] When a response is streaming (and "Edit Last Message" is focused) hitting escape should stop the stream, same as how it works when the message input is focused

chillatom:
This looks really slick imo.  Thanks for getting this in place @abeatrix!

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -443,42 +454,42 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+    /**
+     * Handles editing a human chat message in current chat session.
+     *
+     * Removes any existing messages from the provided index,
+     * before submitting the replacement text as a new question.
+     * When no index is provided, default to the last human message.
+     *
+     * TODO (bee) set index as required once confirmed this change doesn't affect other clients
+     */
+    private async handleEdit(
```
abeatrix, line 466: @beyang can i get your review on this change to see if this aligns with what you have in mind for supporting user context items and useEnhancedContext?

kalanchan:
@toolmantim @beyang heads up!

I think theres a couple things we should brush up on before releasing this:

1. If the chat input is not focused, the edit and new chat shortcuts don't work - is this expected? 
2. the toggle between new/follow up chat (cmd shift) isn't very easy to understand as a user, I think we need a way to flash a verbose message on the chat somewhere that explains what mode you're in and what it means. It should be display NEW CHAT MODE or FOLLOW UP MODE, something like that, I don't think the blue submit button is enough
4. Edit last message doesn't respect the follow up chat mode, it always creates a new chat
5. I think we need to re-think how "edit last message" works for a chat command (explain, smell, test). When you edit the last chat command, it shows the command with no file that was referenced. If you try to add it via @ file, it breaks the dropdown. 

https://github.com/sourcegraph/cody/assets/51868853/a2fd8772-7db5-4fc8-a9e8-3aef41e09dc1

CleanShot 2024-01-22 at 15 57 04@2x


https://github.com/sourcegraph/cody/assets/51868853/43a72441-e9c8-42eb-a6bf-e88798547a3b


https://github.com/sourcegraph/cody/assets/51868853/324f48a9-950f-4ede-8433-4e5bd0de0f3d

abeatrix:
> * If the chat input is not focused, the edit and new chat shortcuts don't work - is this expected?

@toolmantim should we mark these buttons as disabled when the chat input box is not focused? I don't think these commands should be global but want to get your confirmation on this. 


> * the toggle between new/follow up chat (cmd shift) isn't very easy to understand as a user, I think we need a way to flash a verbose message on the chat somewhere that explains what mode you're in and what it means. It should be display NEW CHAT MODE or FOLLOW UP MODE, something like that, I don't think the blue submit button is enough. I got confused a couple times because I was in new chat mode thinking I was doing a follow up question (natural instinct of chat UX)

@toolmantim Can I get a design from you to address this please? 

> * Edit last message doesn't respect the follow up chat mode, it always creates a new chat
I can't reproduce this issue unless I have clicked "Command+shift" to switch to the submit chat in new mode. Perhaps you had that mode turned on?

> * I think we need to re-think how "edit last message" works for a chat command (explain, smell, test). When you edit the last chat command, it shows the command with no file that was referenced. If you try to add it via @ file, it breaks the dropdown.

This would be an existing issue and not related to this change as we currently do not support using @ file with commands. Users can still run commands in chat as they do currently where they just type in the `/` command and the @ file will be added for them automatically.

toolmantim:
> @toolmantim should we mark these buttons as disabled when the chat input box is not focused? I don't think these commands should be global but want to get your confirmation on this.

Let's not disable the buttons but instead remove the keyboard shortcut labels on the action buttons if the message input isn't focussed.

>> the toggle between new/follow up chat (cmd shift) isn't very easy to understand as a user,
> @toolmantim Can I get a design from you to address this please?

Can we change this to use the standard option-modifier (or windows equivalent) key, similar to this:

https://github.com/sourcegraph/cody/assets/153/66b20966-15e8-4e9f-b6ce-89c5048922ee

image

The icon would change from up arrow to a right arrow whenever you're pressing the option key. That keybinding (to detect option) should be on the `document` though, not the message input.

We can remove the additional mode for now, and do this in a follow-up?

toolmantim:
>> @kalanchan I think we need to re-think how "edit last message" works for a chat command (explain, smell, test). When you edit the last chat command, it shows the command with no file that was referenced. If you try to add it via @ file, it breaks the dropdown.
> @abeatrix This would be an existing issue and not related to this change as we currently do not support using @ file with commands. Users can still run commands in chat as they do currently where they just type in the / command and the @ file will be added for them automatically.

Let's disable editing of messages that are commands for now, if they don't work reliably. And we don't show the "Edit Last Message" button either.

toolmantim:
Is there a way to support VS Code’s back button (in the window title bar) so you can return to the chat if you hit "New Chat" and want to get back? Perhaps we need to change the way we close/open the editor windows to support it?

abeatrix:
> Is there a way to support VS Code’s back button (in the window title bar) so you can return to the chat if you hit "New Chat" and want to get back? Perhaps we need to change the way we close/open the editor windows to support it?

I need to check if the back button would work with web view. If it doesn't, do you think new chat being opened in a new editor window instead of using the current one would help?

toolmantim:
> @abeatrix I need to check if the back button would work with web view. If it doesn't, do you think new chat being opened in a new editor window instead of using the current one would help?

Yep! Or alternatively we show this so they can go back easily, and it would only be shown if history is blank and they got there from clicking "Start New Chat" in another chat:

image

That makes it one less window to manage for the user. New chat window for now is fine though.

toolmantim:
If we make the above changes I'm comfortable merging this straight away so it's in the next release

toolmantim (approved):

abeatrix:
- [x] disable editing of messages that are commands for now
  - [x] show disabled edit icon for command
    - @toolmantim instead of no showing the edit icon i've disabled it instead so people can hover over it to understand why it's not showing up, let me know if this sounds ok to you
  - [x] no "Edit Last Message" button if the last message was a command
- [x] listen to key down events in webview for the chat actions
  - @toolmantim instead of remove the keyboard shortcut labels on the action buttons when input isn't focused, which looks very jumpy to me when i tried, i've added a document listener instead. I can change it back if you don't think this is a good idea :)

toolmantim (approved):
This is looking good! There's one bug around editing message when a stream is in progress (it doesn't regenerate a response but just stops the stream):

https://github.com/sourcegraph/cody/assets/153/fbfc4fa4-ffeb-45d6-8a66-caaf25a77b67

Aside from that, everything else is solid, and the keyboard shortcuts hiding/showing works nicely 🙌🏼

abeatrix:
As discussed with @toolmantim , we will work on a follow-up to add supports back for submitting a message as new chat vs follow up messages on keys pressed.

agent/src/index.test.ts
```diff
@@ -186,9 +209,23 @@ export class TestClient extends MessageHandler {
         return reply.messages.at(-1)
     }
 
-    public async editMessage(id: string, text: string): Promise<ChatMessage | undefined> {
+    public async editMessage(
+        id: string,
+        text: string,
+        index?: number,
```
olafurpg, line 215: FYI @abeatrix to add optional parameters, it's better to add a property to `params:` below

agent/src/index.test.ts
```diff
@@ -487,7 +529,9 @@ describe('Agent', () => {
         `,
             explainPollyError
         )
-        client.notify('autocomplete/completionAccepted', { completionID: completions.items[0].id })
+        client.notify('autocomplete/completionAccepted', {
```
olafurpg, line 532: @abeatrix any chance you still have configured prettier as a `defaultFormatter` in your VS Code settings? These diffs here are a sign that the code has been formatted with the default Biome settings, and then reformatted with our custom biome settings.

agent/src/index.test.ts
abeatrix: @olafurpg Oh yea! This is [one of my follow-ups](https://github.com/sourcegraph/cody/blob/17068701ac86b4c62432a40ce9edbd282ade7381/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts#L469)! We actually don't want this to be an optional parameter, but I wasn't sure if making it required would break JetBrains . Is this the case here? If not, I'll create a PR to update thisv

PR #2827, opened 2024-01-20
add FileURI type for `file`-scheme URIs

sqs:
TODO: still need to separate this out cleanly, not ready for review

When a URI is accepted *and* it MUST be a `file` URI, it is helpful to use this new `FileURI` type so that the typechecker helps us enforce correctness.

The `isFileURI` helper is a type guard.

## Test plan

CI


sqs:
nevermind, this can't be separated cleanly/easily

PR #2829, opened 2024-01-21
skip remote embeddings detection for dotcom

sqs:
Remote embeddings are no longer used on dotcom, so this skips an unnecessary step. Even if remote embeddings would be found, they would not be used because the embeddings search methods already perform the check that was added here.

## Test plan

CI

### DIFF
--- a/lib/shared/src/embeddings/EmbeddingsDetector.ts
+++ b/lib/shared/src/embeddings/EmbeddingsDetector.ts
@@ -6,7 +6,7 @@ import { SourcegraphEmbeddingsSearchClient } from './client'
 // A factory for SourcegraphEmbeddingsSearchClients. Queries the client connection and app (if
 // available) for remote embeddings in parallel and returns the one with embeddings available.
 export const EmbeddingsDetector = {
-    // Creates an remote embeddings search client with the first client in `clients`
+    // Creates a remote embeddings search client with the first client in `clients`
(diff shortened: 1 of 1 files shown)


lib/shared/src/embeddings/EmbeddingsDetector.ts
```diff
@@ -1,19 +1,25 @@
...
 
-// A factory for SourcegraphEmbeddingsSearchClients. Queries the client
-// connection and app (if available) in parallel and returns the one with
-// embeddings available.
+// A factory for SourcegraphEmbeddingsSearchClients. Queries the client connection and app (if
+// available) for remote embeddings in parallel and returns the one with embeddings available.
 export const EmbeddingsDetector = {
-    // Creates an embeddings search client with the first client in `clients`
-    // that has embeddings. If none have embeddings, returns undefined. If all
+    // Creates an remote embeddings search client with the first client in `clients`
```
BolajiOlajide, line 9:
```suggestion
    // Creates a remote embeddings search client with the first client in `clients`
```

BolajiOlajide (approved):

lib/shared/src/embeddings/EmbeddingsDetector.ts
sqs: ty :)

PR #2830, opened 2024-01-21
use more standard AbortSignal for aborting stream/chat operations

sqs:
Passing an AbortSignal is more standard and interoperable than the previous way of having them return a `() => void` that cancels the operation.

## Test plan

CI

### DIFF
--- a/lib/shared/src/common/abortController.ts
+++ b/lib/shared/src/common/abortController.ts
@@ -1,0 +1,34 @@
+/**
+ * Create a new AbortController that also aborts the given {@link signal} is aborted.
+ *
+ * This can be used for a operation that is controlled by an {@link AbortSignal} from its caller but
+ * that also needs to create its own {@link AbortController} to control its own operations.
+ */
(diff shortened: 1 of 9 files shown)


arafatkatze (approved):

lib/shared/src/common/abortController.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Create a new AbortController that also aborts the given {@link signal} is aborted.
+ *
+ * This can be used for a operation that is controlled by an {@link AbortSignal} from its caller but
+ * that also needs to create its own {@link AbortController} to control its own operations.
+ */
+export function dependentAbortController(signal?: AbortSignal): AbortController {
```
valerybugakov, line 7: We have `function forkSignal(signal: AbortSignal): AbortController` in `vscode/src/completions/utils.ts`, which does the same thing. We need to migrate to your shared utility for consistency. 

cc @philipp-spiess in case you're using it in your current work

PR #2832, opened 2024-01-21
use async generator for completions and chat clients

sqs:
This is more idiomatic JavaScript, is cleaner, and is more interoperable than custom callbacks.

The underlying implementations (`_streamWithCallbacks` methods) still use callbacks. That can be addressed separately. It's a nice win for callers to not need to use callbacks.

## Test plan

CI and e2e tests. Confirmed that completions, chat, and inline edits work manually.


arafatkatze (approved):

valerybugakov (commented):
💜

PR #2836, opened 2024-01-21
use biome for formatting (not prettier) and linting (not eslint)

sqs:
[Biome](https://biomejs.dev/) is significantly faster than ESLint and Prettier:

- lint all files: biome 0.497s vs. eslint 28.192s (the only really valuable rule we lose is `@typescript-eslint/no-unsafe-member-access`)
- format all files: biome 0.523s vs. prettier 5.762s

It also has a lot of nice new rules (that are not annoying/bikesheddy).

- The [Biome VS Code extension](https://marketplace.visualstudio.com/items?itemName=biomejs.biome) is great. Install it!
- If you have an `editor.defaultFormatter` setting in your VS Code user settings per-language that points to `eslint` or something else, remove it so Biome can run on save in VS Code.
- On other branches where this causes merge conflicts, run `pnpm run lint:js --apply` and `pnpm run format`. That should reformat your branch to minimize merge conflicts.

## Test plan

CI

PR #2842, opened 2024-01-21
POC: test out grit for shared-imports

morgante:
https://github.com/sourcegraph/cody/pull/2836 discarded the [`no-restricted-imports` rule](https://github.com/sourcegraph/cody/pull/2836/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5L64-L76).

This is mostly an attempt to see how fast it would be to replicate this with GritQL.

```
$ time grit check
grit check  0.39s user 0.15s system 270% cpu 0.201 total
```

### DIFF
--- a/.grit/.gitignore
+++ b/.grit/.gitignore
@@ -1,0 +1,2 @@
+.gritmodules*
+*.log
--- a/.grit/grit.yaml
+++ b/.grit/grit.yaml
@@ -1,0 +1,16 @@
+version: 0.0.1
+patterns:
+  - name: github.com/getgrit/stdlib#*
+  - name: no_shared_imports
+    level: error
+    description: |
+      Please import from @sourcegraph/cody-shared instead of subpaths.
+      If you need to use something that is not exported by @sourcegraph/cody-shared, just update lib/shared/src/index.ts to export the thing you need.
+      Reasons for this restriction: (1) enforce a clean boundary between the internal and external API of the shared lib,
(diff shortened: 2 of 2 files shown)


sqs:
Awesome, thanks! This is great. I've heard good things about Grit and appreciate the PR. I will try this out.

morgante:
> Awesome, thanks! This is great. I've heard good things about Grit and appreciate the PR. I will try this out.

Thanks! I'm happy to port the full config and set up the [GitHub action](https://github.com/marketplace/actions/grit-check) for you if you're amenable to including it.

sqs:
I think I'd want to consume it in the editor too so devs get instant feedback (not just in CI). I see https://marketplace.visualstudio.com/items?itemName=Grit.grit-vscode is marked as private beta. Each editor extension we ask people to install also has a burden. For example, to switch from ESLint and Prettier, I found the top 10 committers on our team to this repository and messaged them on Slack to ensure they got the notice. We can overcome this somewhat with VS Code's extensions.json, but each additional tool is still an additional burden. Is it possible to hook this into Biome so that I can just add a Biome plugin or something to biome.json and package.json, and then everyone's Biome editor extension, `pnpm run biome`, etc., will just handle it?

morgante:
The extension is openly usable for anyone now - need to update that message - and will give ~instant diagnostics for files.

Unfortunately, I don't think Biome has any plugin API yet. Biome's approach makes it fast, but extensibility hasn't been figured out yet.

PR #2847, opened 2024-01-22
resolve remote embeddings fileNames to URIs

beyang:
This fixes an issue with remote embeddings where attempting to run a chat that fetched remote embeddings would completely fail with an error from trying to dereference `.scheme` on an undefined value.
    
The Sourcegraph GraphQL API returns remote embeddings results with `fileName` fields. We need to resolve this to a `uri` so that we don't pass around a relative fileName that we don't know how to resolve. To resolve it, we assume (as do other context sources) that the context is for the first workspace folder in VS Code. We need to pass this around as `workspaceFolderUri`, which is ugly but is the smallest change that will make this work. The remote embeddings code is going away soon anyway, so IMO this solution suffices.

## Test plan

Using remote embeddings, run a chat that gets context and ensure there is no URI error blocking it. Run an inline edit as well.
    
Co-authored-by: Quinn Slack <quinn@slack.org>

### DIFF
--- a/lib/shared/src/codebase-context/index.ts
+++ b/lib/shared/src/codebase-context/index.ts
@@ -51,6 +51,7 @@ export class CodebaseContext {
      * the most important context message appears *last*)
      */
     public async getContextMessages(
+        workspaceFolderUri: URI,
         query: string,
         options: ContextSearchOptions
     ): Promise<ContextMessage[]> {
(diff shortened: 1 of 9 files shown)


sqs (approved):
Code is good as a workaround. Haven’t tried it. Sorry for the bug!

sqs:
I'll pick this up, @beyang, and get it merged.

beyang:
this got rolled into https://github.com/sourcegraph/cody/pull/2848

beyang:
Wait nvm, there's additional changes in this PR

sqs:
Yeah, the problem is that the fileName is relative not absolute, so `vscode.Uri.file` would not work. The context would be pulled in, but the links would not be clickable. My additional changes fix that issue by resolving the fileNames to their absolute path (joining them w/ the workspace root).

sqs:
I'll rebase and test

PR #2848, opened 2024-01-22
Refactor chat

beyang:
This PR addresses some code quality issues and complexity in the chat implementation.
* https://github.com/sourcegraph/cody/pull/2542 introduced a separate code path for handling command-initiated chats. This introduced a flag that special-cased handling of commands in a number of places—in the view controller (SimpleChatPanelProvider), prompt constructor, and context fetcher. These flags make chat more difficult to reason about, requiring the dev to reason about the conditional flag at multiple places in the lifecycle of a chat request.

Code changes
* Instead of checking a flag to see if the user request was a command in `DefaultPrompter`, introduces a new `CommandPrompter` to customize prompt construction for commands
* Breaks apart `ContextProvider` into separate functions, moved to a `context.ts` file. Enhanced context can be fetched by either prompter.
* Simplifies the core code paths in `SimpleChatPanelProvider`. Now there are two top-level message handlers: `handleNewUserMessage` (for regular chat) and `handleCommand` (for commands). These then execute the necessary steps without entangling other methods that must then support both code paths via conditional.
  * Previously, we had the following mess:
    * `executeChat` -> `handleHumanMessageSubmitted`
    * `handleHumanMessageSubmitted` -> 'handleChatRequest` OR `handleCommands`
    * `handleCommands` -> `handleChatRequest`
    * `handleChatRequest` -> `generateAssistantResponse`

Behavior changes
* This change removes the already defunct `cody.experimental.chatPredictions` setting.
* This change disables codebase-wide context fetching for commands. This field appears unused by first-party commands, and it's unclear whether it was actually being used for custom commands.

Follow-up work: This PR also flags some outstanding bugs uncovered in the refactor:
* Update chat "edit" to support user-specified context items and the value of the enhanced context selector
* Many commands overflow the context window currently (though we're masking the error so it's not obvious to the end user)
* Remove codebase context from custom commands or add support for fetching it
* Handle symf errors better
* Move methods around in SimpleChatPanelProvider for greater clarity (not doing in this PR to avoid making the diff larger)
* https://github.com/sourcegraph/cody/pull/2848#discussion_r1461315316
* https://github.com/sourcegraph/cody/pull/2848#discussion_r1461314572
* https://github.com/sourcegraph/cody/pull/2848#discussion_r1461314120
* https://github.com/sourcegraph/cody/pull/2848#discussion_r1461335084

## Test plan

- [ ] Test chat with enhanced context
- [ ] Test chat without context
- [ ] Test all first-party commands
- [ ] Test a custom command

### DIFF
--- a/vscode/src/chat/chat-view/ChatManager.ts
+++ b/vscode/src/chat/chat-view/ChatManager.ts
@@ -118,11 +118,10 @@ export class ChatManager implements vscode.Disposable {
         const requestID = uuid.v4()
         telemetryService.log('CodyVSCodeExtension:chat-question:submitted', { requestID, ...args })
         const chatProvider = await this.getChatProvider()
-        await chatProvider.handleHumanMessageSubmitted(requestID, question, 'user', [], false)
+        await chatProvider.handleNewUserMessage(requestID, question, 'user-newchat', [], true)
     }
 
     // Execute a command request in a new chat panel
-
     public async executeCommand(
         command: CodyCommand,
         args: CodyCommandArgs,
@@ -143,7 +142,7 @@ export class ChatManager implements vscode.Disposable {
 
         // Else, open a new chanel panel and run the command in the new panel
         const chatProvider = await this.getChatProvider()
-        await chatProvider.handleCommands(command, args)
+        await chatProvider.handleCommand(command, args)
         return chatProvider
     }
 
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -30,7 +30,10 @@ import {
 import type { View } from '../../../webviews/NavBar'
 import { newCodyCommandArgs, type CodyCommandArgs } from '../../commands'
 import type { CommandsController } from '../../commands/CommandsController'
-import { createDisplayTextWithFileLinks, createDisplayTextWithFileSelection } from '../../commands/prompt/display-text'
+import {
+    createDisplayTextWithFileLinks,
+    createDisplayTextWithFileSelection,
+} from '../../commands/prompt/display-text'
 import { getFullConfig } from '../../configuration'
 import { executeEdit } from '../../edit/execute'
 import {
@@ -40,10 +43,10 @@ import {
 } from '../../editor/utils/editor-context'
 import type { VSCodeEditor } from '../../editor/vscode-editor'
 import { ContextStatusAggregator } from '../../local-context/enhanced-context-status'
-import { type LocalEmbeddingsController } from '../../local-context/local-embeddings'
-import { type SymfRunner } from '../../local-context/symf'
+import type { LocalEmbeddingsController } from '../../local-context/local-embeddings'
+import type { SymfRunner } from '../../local-context/symf'
 import { logDebug } from '../../log'
-import { type AuthProvider } from '../../services/AuthProvider'
+import type { AuthProvider } from '../../services/AuthProvider'
 import { getProcessInfo } from '../../services/LocalAppDetector'
 import { localStorage } from '../../services/LocalStorageProvider'
 import { telemetryService } from '../../services/telemetry'
@@ -66,11 +69,23 @@ import type { ChatViewProviderWebview, Config } from './ChatPanelsManager'
 import { CodebaseStatusProvider } from './CodebaseStatusProvider'
 import { getCommandContext, getEnhancedContext } from './context'
 import { InitDoer } from './InitDoer'
-import { ContextItem, MessageWithContext, SimpleChatModel, toViewMessage } from './SimpleChatModel'
-import { CachedRemoteEmbeddingsClient } from '../CachedRemoteEmbeddingsClient'
-import { AuthStatus, ChatSubmitType, ConfigurationSubsetForWebview, ExtensionMessage, LocalEnv, WebviewMessage } from '../protocol'
-import { CommandPrompter, DefaultPrompter, IPrompter } from './prompt'
-import { MessageErrorType } from '../MessageProvider'
+import {
+    type ContextItem,
+    type MessageWithContext,
+    SimpleChatModel,
+    toViewMessage,
+} from './SimpleChatModel'
+import type { CachedRemoteEmbeddingsClient } from '../CachedRemoteEmbeddingsClient'
+import type {
+    AuthStatus,
+    ChatSubmitType,
+    ConfigurationSubsetForWebview,
+    ExtensionMessage,
+    LocalEnv,
+    WebviewMessage,
+} from '../protocol'
+import { CommandPrompter, DefaultPrompter, type IPrompter } from './prompt'
+import type { MessageErrorType } from '../MessageProvider'
 
 interface SimpleChatPanelProviderOptions {
     config: Config
@@ -186,8 +201,8 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
 
         const viewType = CodyChatPanelViewType
         const panelTitle =
-            this.history.getChat(this.authProvider.getAuthStatus(), this.chatModel.sessionID)?.chatTitle ||
-            getChatPanelTitle(lastQuestion)
+            this.history.getChat(this.authProvider.getAuthStatus(), this.chatModel.sessionID)
+                ?.chatTitle || getChatPanelTitle(lastQuestion)
         const viewColumn = activePanelViewColumn || vscode.ViewColumn.Beside
         const webviewPath = vscode.Uri.joinPath(this.extensionUri, 'dist', 'webviews')
         const panel = vscode.window.createWebviewPanel(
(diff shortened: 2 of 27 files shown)


vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -118,11 +118,10 @@ export class ChatManager implements vscode.Disposable {
         const requestID = uuid.v4()
         telemetryService.log('CodyVSCodeExtension:chat-question:submitted', { requestID, ...args })
         const chatProvider = await this.getChatProvider()
-        await chatProvider.handleHumanMessageSubmitted(requestID, question, 'user', [], false)
+        await chatProvider.handleNewUserMessage(requestID, question, 'user-newchat', [], true)
```
dominiccooney, line 121: This is why I'm not a fan of boolean parameters... They are mysterious at the callsite.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
 
-    /**
-     * Handles a message submitted by the user.
-     *
-     * Validates the message, checks for slash commands, edit commands,
-     * and sends the message to be handled like a regular chat request.
-     */
-    public async handleHumanMessageSubmitted(
+    public async handleCommand(command: CodyCommand, args: CodyCommandArgs): Promise<void> {
+        // If it's not a ask command, it's a fixup command. If it's a fixup request, we can exit early
```
dominiccooney, line 533: Here is "fixup command" and then "fixup request", but these are the same thing, right? Call it the same thing to avoid confusion. Calling it something different makes it look like you're making a distinction between two specific things.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+        // This is because startCommand will start the CommandRunner,
+        // which would send all fixup requests to the FixupController
+        if (command.mode !== 'ask') {
+            return
+        }
+
+        if (!this.editor.getActiveTextEditorSelectionOrVisibleContent()) {
```
dominiccooney, line 540: It would be nice if the check for command.context.selection, etc., fetching that context, and the error reporting was all together instead of spread out like this.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
-     * and sends it to be handled like a regular chat request.
-     */
-    public async handleCommands(command: CodyCommand, args: CodyCommandArgs): Promise<void> {
-        // If it's not a ask command, it's a fixup command. If it's a fixup request, we can exit early
-        // This is because startCommand will start the CommandRunner,
-        // which would send all fixup requests to the FixupController
-        if (command.mode !== 'ask') {
-            return
-        }
+        // trigger the context progress indicator
```
dominiccooney, line 602: Given we have had this in a couple of places with this comment, why not just add a method `triggerContextProgressIndicator`?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+            const authStatus = this.authProvider.getAuthStatus()
+            const properties = {
+                requestID,
+                chatModel: this.chatModel.modelID,
+                contextSummary,
             }
-        }
 
-        const inputText = [command.slashCommand, command.additionalInput].join(' ')?.trim()
+            // Only log chat-question event if it is not a command to avoid double logging for commands
```
dominiccooney, line 629: This comment is odd now that the method is a "human message" handling method.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+                    ...contextSummary,
+                    // Flag indicating this is a transcript event to go through ML data pipeline. Only for DotCom users
+                    // See https://github.com/sourcegraph/sourcegraph/pull/59524
+                    recordsPrivateMetadataTranscript: authStatus.endpoint && isDotCom(authStatus.endpoint) ? 1 : 0,
+                },
+                privateMetadata: {
+                    properties,
+                    // 🚨 SECURITY: chat transcripts are to be included only for DotCom users AND for V2 telemetry
+                    // V2 telemetry exports privateMetadata only for DotCom users
+                    // the condition below is an aditional safegaurd measure
```
dominiccooney, line 644:
```suggestion
                    // the condition below is an additional safeguard
```

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
-        await this.handleChatRequest(args.requestID, inputText, 'user', [], false, command)
+        try {
+            const prompt = await this.buildPrompt(prompter, sendTelemetry)
+            this.streamAssistantResponse(requestID, prompt)
+        } catch (error) {
+            if (isRateLimitError(error)) {
+                this.postError(error, 'transcript')
+            } else {
+                this.postError(isError(error) ? error : new Error(`Error generating assistant response: ${error}`))
+            }
```
dominiccooney, line 658: While you're cleaning things up, we can de-dup the `postError` here.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,170 +529,236 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
-            userContextFiles,
-            addEnhancedContext,
-            contextSummary => {
-                if (submitType !== 'user' && submitType !== 'user-newchat') {
-                    return
+        // Update UI based on prompt construction
+        this.chatModel.setNewContextUsed(newContextUsed)
+        if (contextLimitWarnings.length > 0) {
+            const warningMsg = contextLimitWarnings
+                .map(w => (w.trim().endsWith('.') ? w.trim() : w.trim() + '.'))
```
dominiccooney, line 675: Minor nit but I think the following is more readable. The purpose is obscured by calling 'trim' three times. Even better if we have a helper which does this already and can just call it.

```suggestion
                .map(w => {
                  w = w.trim();
                  if (!w.endsWith('.')) {
                    w += '.'
                  }
                  return w
                })
```

dominiccooney (approved):
Approving, this is an improvement. @philipp-spiess was pursuing similar breakout of context in https://github.com/sourcegraph/cody/pull/2804 so it is clearly time. Suggestions inline.

@abeatrix how well are commands covered by E2E tests?

I see the TODO about enhanced context, @beyang what's the bottom line on the change to context? Your description mentions some things I think aren't here yet.

vscode/src/commands/CommandsController.ts
```diff
@@ -211,15 +211,14 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
                 case selectedCommandID === menu_options.chat.slashCommand: {
                     let input = userPrompt.trim()
                     if (input) {
-                        return await vscode.commands.executeCommand('cody.action.chat', input, {
-                            source: 'command',
-                        })
+                        await vscode.commands.executeCommand('cody.action.commands.exec', `${selectedCommandID} input`)
```
abeatrix, line 214:
```suggestion
-                         await vscode.commands.executeCommand('cody.action.chat', input)
```

When the selectedCommandID is equal to "/ask" , it means this is a chat question submitted from the command menu, so we can send process this as a chat question with "cody.action.chat".

abeatrix (approved):
Thanks. Left a comment in line.

vscode/src/chat/chat-view/ChatManager.ts
beyang: Yeah, I agree. In this case, there is a semantic change—we now use enhanced context when starting a chat from the command palette.

vscode/src/commands/CommandsController.ts
beyang: The issue with executing the command `cody.action.chat` is that we lose the context of the user selection, so the response ends up being very low quality.

vscode/src/commands/CommandsController.ts
```diff
@@ -211,15 +211,14 @@ export class CommandsController implements VsCodeCommandsController, vscode.Disp
...
                     }
                     input = await showAskQuestionQuickPick()
-                    await vscode.commands.executeCommand('cody.chat.panel.new')
-                    return await vscode.commands.executeCommand('cody.action.chat', input, {
-                        source: 'command',
-                    })
+                    return await vscode.commands.executeCommand(
+                        'cody.action.commands.exec',
+                        `${selectedCommandID} ${input}`
+                    )
```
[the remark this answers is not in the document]
beyang: See previous response

beyang:
> I see the TODO about enhanced context, @beyang what's the bottom line on the change to context?

The only change to context should be:
* Codebase-wide context fetching is disabled now for commands. This field appears unused by first-party commands, and it's unclear whether it was actually being used for custom commands.

PR #2855, opened 2024-01-22
Autocomplete: Cancel requests that are no longer relevant

philipp-spiess:
This PR adds the ability to abort inflight requests when the document was updated in a way that yields new completions no longer relevant.

This is useful as an optimization to avoid streaming completions run for too long, for example in cases where the user updates the document after the request was started in a way that the completion would no longer yield useful results (e.g. they rewrite the prefix). Or in case of hot-streaks when the completion starts to diverge from the current document prefix (e.g. when the user rejects the first completion and thus all hot-streaks being generated would be useless).

The exact cases of when something is deemed relevant are tested and document separately. Most importantly we leave a room for some typos by the user to avoid cancelling requests as soon as the user makes one wrong keystroke (even though they might revert it right away).

## Test plan

- See the added tests
- In real world, you can see this working by enabling the hot-streak feature and observing the console for request cancellation notifications.

### DIFF
--- a/vscode/src/completions/providers/fetch-and-process-completions.ts
+++ b/vscode/src/completions/providers/fetch-and-process-completions.ts
@@ -210,12 +210,12 @@ export async function* fetchAndProcessCompletions(
             continue
         }
 
-        const parsedComletion = isFinal
+        const parsedCompletion = isFinal
             ? parseAndTruncateCompletion(rawCompletion, providerOptions)
             : canUsePartialCompletion(rawCompletion, providerOptions)
 
-        if (parsedComletion) {
-            const completedCompletion = processCompletion(parsedComletion, providerOptions)
+        if (parsedCompletion) {
+            const completedCompletion = processCompletion(parsedCompletion, providerOptions)
 
             yield {
                 docContext,
--- a/vscode/src/completions/request-manager.test.ts
+++ b/vscode/src/completions/request-manager.test.ts
@@ -4,7 +4,12 @@ import { getCurrentDocContext } from './get-current-doc-context'
 import { InlineCompletionsResultSource } from './get-inline-completions'
 import type { FetchCompletionResult } from './providers/fetch-and-process-completions'
 import { Provider } from './providers/provider'
-import { RequestManager, type RequestManagerResult, type RequestParams } from './request-manager'
+import {
+    RequestManager,
+    computeStillRelevantCompletions,
+    type RequestManagerResult,
+    type RequestParams,
+} from './request-manager'
 import { documentAndPosition } from './test-helpers'
 import type { ContextSnippet } from './types'
 
@@ -62,8 +67,8 @@ function createProvider(prefix: string) {
     })
 }
 
-function docState(prefix: string, suffix = ';'): RequestParams {
-    const { document, position } = documentAndPosition(`${prefix}█${suffix}`)
+function docState(prefix: string, suffix = ';', uriString?: string): RequestParams {
+    const { document, position } = documentAndPosition(`${prefix}█${suffix}`, undefined, uriString)
     return {
         document,
         position,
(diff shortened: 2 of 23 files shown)


vscode/src/completions/request-manager.test.ts
```diff
@@ -211,3 +216,74 @@ describe('RequestManager', () => {
...
+        const previousRequest = docState('console.')
+        const completion = { insertText: 'log("Hello, world!")' }
+
+        expect(computeStillRelevantCompletions(currentRequest, previousRequest, [completion])).toEqual([
+            completion,
+        ])
+    })
+
+    it('handles cases on different lines', async () => {
+        const currentRequest = docState('if (true) {\n  console.')
```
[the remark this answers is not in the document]
philipp-spiess: Do you have a specific test case in mind for this? Maybe we need to remove all leading whitespace in the calculation logic so the indentation can be mixed and matched however we like it? I will propose something

vscode/src/completions/request-manager.test.ts
```diff
@@ -211,3 +216,74 @@ describe('RequestManager', () => {
...
+    it('never matches for mismatching prefixes', async () => {
+        const hundredLines = 'WOW\n'.repeat(100)
+        const thousandLines = 'WOW\n'.repeat(1000)
+        const currentRequest = docState(`${hundredLines}console.log`)
+        const previousRequest = docState(`${thousandLines}console.`)
+        const completion = { insertText: 'log("Hello, world!")' }
+
+        // Even though the prefix will look the same, it'll be on different lines and should thus
+        // not be reused
+        expect(computeStillRelevantCompletions(currentRequest, previousRequest, [completion])).toEqual(
```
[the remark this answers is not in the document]
philipp-spiess: I think in prod environments, having the _full_ prefix be equivalent is very unlikely anyways. I guarded against this mostly because the logic that we need to ensure the prefixes have the same lengths would not work if we allow these prefixes to start _anywhere_ in the document. And since we do pick a certain number of line before the cursor, we can't omit this logic.

vscode/src/completions/request-manager.test.ts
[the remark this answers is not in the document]
valerybugakov: > so the indentation can be mixed and matched however we like it

Maybe yeah, it would be the simplest solution to start with. 

> Do you have a specific test case in mind for this?

For example the use case that caused the bug for Stevey, where the completion was generated at the beginning of the line but not shown because he quickly tabbed several times.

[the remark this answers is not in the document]
philipp-spiess: @valerybugakov I pushed a fix for this :)

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -210,12 +210,12 @@ export async function* fetchAndProcessCompletions(
             continue
         }
 
-        const parsedComletion = isFinal
+        const parsedCompletion = isFinal
             ? parseAndTruncateCompletion(rawCompletion, providerOptions)
             : canUsePartialCompletion(rawCompletion, providerOptions)
 
-        if (parsedComletion) {
```
valerybugakov, line 217: Gosh, I keep missing letters. Is it time to buy 🔎?

philipp-spiess (commented):
forget to submit these comment sorryyyyy

vscode/src/completions/providers/fetch-and-process-completions.ts
philipp-spiess: Hahah I can recommend "Code Spell Checker" but that doesn't prevent me from making typos all over the place myself lol

valerybugakov: Yeah, I had it, but at some point, VS Code became too slow even on my M1 Mac, so I turned off all nonessential extensions. Probably, it's time to turn it back on and see how it goes.

vscode/src/completions/request-manager.ts
```diff
@@ -91,32 +96,40 @@ export class RequestManager {
...
-                        completions: processedCompletions,
-                        source: InlineCompletionsResultSource.Network,
-                    })
+                        // Shared post-processing logic
+                        const processedCompletions = wrapInActiveSpan(
+                            'autocomplete.shared-post-process',
+                            () => processInlineCompletions(completions, requestParams)
+                        )
 
-                    this.testIfResultCanBeRecycledForInflightRequests(request, processedCompletions)
```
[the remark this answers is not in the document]
valerybugakov: Good catch, TY!

vscode/src/completions/request-manager.ts
```diff
@@ -192,13 +209,36 @@ export class RequestManager {
...
+
+        for (const request of this.inflightRequests) {
+            if (
+                !computeIfRequestStillRelevant(
+                    this.latestRequestParams.requestParams,
+                    request.lastRequestParams,
+                    request.lastCompletions
+                )
+            ) {
+                logDebug('CodyCompletionProvider', 'Request aborted')
```
valerybugakov, line 226:
```suggestion
                logDebug('CodyCompletionProvider', 'Irrelevant request aborted')
```

vscode/src/completions/request-manager.ts
```diff
@@ -192,13 +209,36 @@ export class RequestManager {
...
+    }
 }
 
 class InflightRequest {
     public promise: Promise<RequestManagerResult>
     public resolve: (result: RequestManagerResult) => void
     public reject: (error: Error) => void
 
+    public lastCompletions: InlineCompletionItemWithAnalytics[] | null = null
+    public lastRequestParams: RequestParams
```
valerybugakov, line 240: Let's add a short comment explaining why do we need these fields?

vscode/src/completions/request-manager.ts
```diff
@@ -62,8 +64,11 @@ interface RequestsManagerParams {
 export class RequestManager {
     private cache = new RequestCache()
     private readonly inflightRequests: Set<InflightRequest> = new Set()
+    private latestRequestParams: null | RequestsManagerParams = null
```
valerybugakov, line 67: Let's leave a comment explaining why we need this property?

valerybugakov (approved):
Cool beans! Thanks for adding extensive tests as always 🙌

PR #2858, opened 2024-01-22
Add `graphql/getCurrentUserCodySubscription` endpoint

mkondratek:
- ~Needs https://github.com/sourcegraph/sourcegraph/pull/59742.~
- Needs https://github.com/sourcegraph/sourcegraph/pull/59816.
- Required for https://github.com/sourcegraph/jetbrains/issues/276.

---

> the schema changes should go live when dotcom gets deployed during these times: `09:00 UTC on Monday, 06:00 UTC rest of weekdays`

## Test plan
- [x] let's wait for tomorrow so that the schema changes go live and we can actually test that endpoint from the client
- [x] tested with the jetbrains client + I added a test case - works 🚀


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -95,6 +95,18 @@ interface CurrentUserCodyProEnabledResponse {
     } | null
 }
 
+interface CurrentUserCodySubscriptionResponse {
+    currentUser: {
+        codySubscription: {
+            status: string
+            isPro: boolean
+            applyProRateLimits: string
+            currentPeriodStartAt: string
```
mkondratek, line 104: or should it be `Date`?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -170,6 +182,15 @@ export interface CodyLLMSiteConfiguration {
     provider?: string
 }
 
+
+export interface CurrentUserCodySubscription {
+    status: string
+    isPro: boolean
+    applyProRateLimits: string
+    currentPeriodStartAt: string
```
mkondratek, line 190: (`Date` here too?)

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -95,6 +95,18 @@ interface CurrentUserCodyProEnabledResponse {
     } | null
 }
 
+interface CurrentUserCodySubscriptionResponse {
+    currentUser: {
+        codySubscription: {
+            status: string
+            isPro: boolean
```
kalanchan, line 102: the `codySubscription` schema has changed a bit, `isPro` is now `pro`

https://github.com/sourcegraph/sourcegraph/blob/da169dc09733d36a917e8bf96b79618a708a926a/cmd/frontend/graphqlbackend/schema.graphql#L6388

lib/shared/src/sourcegraph-api/graphql/client.ts
kalanchan: yeah should be datetime

https://github.com/sourcegraph/sourcegraph/blob/da169dc09733d36a917e8bf96b79618a708a926a/cmd/frontend/graphqlbackend/schema.graphql#L6388

lib/shared/src/sourcegraph-api/graphql/client.ts
kalanchan: yup should be datetime

https://github.com/sourcegraph/sourcegraph/blob/da169dc09733d36a917e8bf96b79618a708a926a/cmd/frontend/graphqlbackend/schema.graphql#L6388

kalanchan (changes requested):

lib/shared/src/sourcegraph-api/graphql/client.ts
mkondratek: you mean the field is `plan` now, right?

kalanchan: woops yes good catch, sorry haha!

lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -30,6 +30,13 @@ export enum FeatureFlag {
     // Enable Cody PLG features on JetBrains
     CodyProJetBrains = 'cody-pro-jetbrains',
 
+    // use-ssc-for-cody-subscription is a feature flag that enables the use of SSC as the source of truth for Cody subscription data.
+    UseSscForCodySubscription = 'use-ssc-for-cody-subscription',
+
+    // cody-pro-trial-ended is a feature flag that indicates if the Cody Pro "Free Trial"  has ended.
+    // (Enabling users to use Cody Pro for free for 3-months starting in late Q4'2023.)
+    CodyProTrialEnded = 'cody-pro-trial-ended',
+
```
mkondratek, line 39: Note: I added these feature flags too.

vdavid (approved):
Looks pretty boilerplate-y, not much to go wrong here 👍

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -95,6 +96,18 @@ interface CurrentUserCodyProEnabledResponse {
     } | null
 }
 
+interface CurrentUserCodySubscriptionResponse {
+    currentUser: {
```
pkukielka, line 100: Can you give me some examples what could be values of those fields?

agent/src/agent.ts
```diff
@@ -542,6 +542,15 @@ export class Agent extends MessageHandler {
             return res.codyProEnabled
         })
 
+        this.registerRequest('graphql/getCurrentUserCodySubscription', async () => {
```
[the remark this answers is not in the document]
mkondratek: looks like a good point - @vdavid  , can you confirm?

lib/shared/src/sourcegraph-api/graphql/client.ts
mkondratek: please take a look at the related document:

https://docs.google.com/document/d/1p8V_c5wkmNgnCshRdlnjbHqxgkYLVlvpTyDdbHu4VSY/edit

agent/src/agent.ts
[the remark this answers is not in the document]
vdavid: Errm, could you rephrase the question? I've never used either function, so not sure what it does. I suppose `registerAuthenticatedRequest` only works if the user has a valid access token with dotcom, while `registerRequest` works without it? If this is the case, then I suppose we need the authed one because anonymous users don't have subscriptions.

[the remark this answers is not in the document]
mkondratek:
```
    // Alternative to `registerRequest` that awaits on authentication changes to
    // propagate before calling the method handler.
    public registerAuthenticatedRequest<M extends RequestMethodName>(
...
```

It sounds correct to use `registerAuthenticatedRequest`

lib/shared/src/sourcegraph-api/graphql/client.ts
pkukielka: I was wondering about `applyProRateLimits` because name sounds like boolean but it is a string.
And it's not mentioned in the doc.

pkukielka: Actually I see it's laso not in the `CURRENT_USER_CODY_SUBSCRIPTION_QUERY`

mkondratek: good point! it's boolean

mkondratek: there is more context there: https://github.com/sourcegraph/sourcegraph/pull/59816/files

pkukielka (approved):
LGTM

PR #2862, opened 2024-01-23
Agent: add new evaluation strategy for chat

hitesh-1997:


agent/src/cli/evaluate-autocomplete/strategy-simple-chat.ts
```diff
@@ -0,0 +1,300 @@
...
+        exit(1);
+    }
+    if(chatLogs===undefined){
+        console.log(`chat logs data undefined for the workspace: ${workspace}`)
+        exit(1)
+    }
+    return [repoDisplayName, chatLogs]
+}
+
+export async function evaluateSimpleChatStrategy(
```
hitesh-1997, line 29: @olafurpg Hi Olaf, could you please review the chat strategy, if the setup looks alright.

agent/src/cli/evaluate-autocomplete/strategy-simple-chat-logs.ts
```diff
@@ -0,0 +1,57 @@
...
+    private logMutex: Mutex
+
+    constructor(base_path: string) {
+        this.logFilePath = `${base_path}/${new Date().toISOString()}/file.log`;
+        this.logFilePathLatest = `${base_path}/latest/file.log`;
+        this.logMutex = new Mutex();
+    }
+
+    async initialize() {
+        await this.createLogFile();
```
olafurpg, line 18: Nit: i think you can avoid all the `Mutex` stuff by using `fspromises.open(..)`

```
import fspromises from 'fs/promises'
...
initialize(): {
  this.open = await fspromises.open(f)
}
...
append(log) {
  this.open.write(..)
}
```

Node.js is single-threaded so you don't have to worry about race conditions. Not a big deal

agent/src/cli/evaluate-autocomplete/strategy-simple-chat.ts
```diff
@@ -0,0 +1,338 @@
...
+
+interface EmbeddingFlag {
+    isEmbeddingReady: boolean,
+    isNumItemNeedIndexZero: boolean,
+    isSearchIndexReady: boolean,
+}
+
+async function createEmbeddings(client: MessageHandler, options: EvaluateAutocompleteOptions): Promise<void> {
+    const [repoDisplayName, chatLogs] = getMetaDataInfo(options)
+    let embeddingDoneFlag: EmbeddingFlag = {
```
olafurpg, line 52: nit: it would be more idiomatic to use `new vscode.EventEmitter<EmbeddingFlag>()` here. You can then easily implement a promise that resolves when all properties are true

```
const embeddingFlags = new vscode.EventEmitter<EmbeddingFlag>()
const onEmbeddingFlagsDidChange = embeddingFlags.event
await new Promise<void>(resolve => {
  const disposable = onEmbeddingFlagsDidChange(flags => {
    if (allPropertiesTrue(flags)) {
      resolve(flags)
      disposable.dispose()
    }
  })
}
```

The benefit of this approach is that you are not relying on `setInterval`. As a rule of thumb, you want to avoid hardcoding timeouts/intervals at because they 1) slow down your pipelines and 2) tend to make your code run less reliably

agent/src/cli/evaluate-autocomplete/strategy-simple-chat.ts
```diff
@@ -0,0 +1,338 @@
...
+    let totalQuestions = all_chat_configs.length
+    for(const single_chat_config of all_chat_configs) {
+        const reply = await simulateChatInRepo(client, options, single_chat_config)
+        replies.push(reply)
+        totalQuestions-=1
+        await chatLogs.writeLog(repoDisplayName, `Number of questions remaining: ${totalQuestions}`)
+    }
+
+    // todo: Add support for concurrency 
+    // todo:     - can't add concurrency as of now b/c for the different question, we might get same webview panel id, if the message is not already posted.
```
olafurpg, line 208: Not anymore, fixed this today https://github.com/sourcegraph/cody/pull/2969

olafurpg: You should be able to run parallel chat/commands. If you can't get it working, please open a discussion in #wg-cody-agent

agent/src/cli/evaluate-autocomplete/strategy-simple-chat.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    await saveListToFile(replies, path.join(options.snapshotDirectory, 'strategy-chat.json'))
+    await chatLogs.writeLog(repoDisplayName, '--------- --------- --------- --------- --------- --------- ---------')
+    await chatLogs.writeLog(repoDisplayName, `Saved results for workspace ${options.workspace}`)
+    await chatLogs.writeLog(repoDisplayName, '--------- --------- --------- --------- --------- --------- ---------')
+}
+
+
+async function textDocumentEvent(
```
olafurpg, line 236: Good to see this code is worth turning into a more reusable library. It's fine to keep it duplicated here for now, but I may remove it down the road

olafurpg (commented):
@hitesh-1997 very cool! I'm glad to see that you have been able to access so much functionality programmatically. This is exactly how I envisioned/hoped the agent would be used. 

Please note that the low-level embeddings-related endpoints are subject to breaking changes. This is the only client using those endpoints, we are not using them for JetBrains. We have no automated tests running to ensure these endpoints are working as expected. Feel free to add tests to `agent/src/index.test.ts` if you want to defensively guard against unwanted regressions. It's your responsibility to make sure these APIs continue working since we are not running the evaluation tool in CI for every PR.

PR #2865, opened 2024-01-23
Edit/Chat: Change shortcut and add dedicated setting

umpox:
Description:
- Adds a dedicated setting for "Command Hints". Disabled by default for now
- Changes the shortcut to "Opt+K" and "Opt+L" for Edit / Chat

image

image

## Test plan

1. Enable "Command Hints"
2. Check hint text shows up on selections
3. Check shortcuts work for both Edit and Chat
4. Disable "Command Hints"
5. Check hint text doesn't show
6. Check shortcuts still work


umpox:
I went with the wording "Command Hints" but open to better ideas, cc @abeatrix @toolmantim

abeatrix (approved):

PR #2866, opened 2024-01-23
Update language to make `@#-symbol` tagging more discoverable

taylorsperry:
Today, it's not easy to discover that the chat input supports `@#-symbol` tagging to include symbols in Cody's chat context (see screenshots below).  

For one, I initially was trying to just type `#` to include a symbol (the current language doesn't make it clear that you actually need `@#`), but this is also something we've heard [internally](https://sourcegraph.slack.com/archives/C04MZPE4JKD/p1705996369006999?thread_ts=1705105369.141479&cid=C04MZPE4JKD) and from a user DM in Discord. The same user interpreted "search" literally and tried adding a symbol with command+F.
  
This PR updates the placeholder to explicitly call out symbol search.

## Test plan

- Started a chat
- Checked the label
- Typed a message
- Checked follow-up label


taylorsperry:
Please do not merge without approval from @toolmantim .

abeatrix (approved):
lgtm!

vscode/webviews/UserContextSelector.tsx
```diff
@@ -36,16 +36,16 @@ export const UserContextSelectorComponent: React.FunctionComponent<
...
+        headingTitle = 'Start typing a symbol to include...'
     } else if (formInput.match(/@[^ #]+$/)) {
         headingTitle = contextSelection?.length
-            ? 'Search for a file to include...'
+            ? 'Start typing a file to include...'
             : 'No matching files found'
     } else if (formInput.match(/@#[^ ]+$/)) {
         headingTitle = contextSelection?.length
-            ? 'Search for a symbol to include...'
+            ? 'Start typing a symbol to include...'
```
toolmantim, line 48: I'm curious why you wanted to change these?

With the current way it works, I don't think we want to use the word "start" here, because these labels are shown as you type too:

CleanShot 2024-01-24 at 10 12 32@2x

I think it's safe to use the label "Search" without instructions, given that's how it works everywhere these days:

CleanShot 2024-01-24 at 10 30 30@2x

CleanShot 2024-01-24 at 10 14 31@2x

If we did want to change it to "Start typing" we should update the logic so there is no label after you start typing. But my preference would be to leave out this change, and we just update the placeholder.

vscode/webviews/Chat.tsx
```diff
@@ -261,7 +261,7 @@ const TextArea: React.FunctionComponent<ChatUITextAreaProps> = ({
     chatModels,
 }) => {
     const inputRef = useRef<HTMLTextAreaElement>(null)
-    const placeholder = 'Message (@ to include code, / for commands)'
+    const placeholder = 'Message (@ to include files, @# for symbols, or / for commands)'
```
toolmantim, line 264: I like the idea of updating this! I'd suggest one of these two versions though:

```suggestion
const placeholder = 'Message (@ to include files and symbols, or / for commands)'
```

or

```suggestion
const placeholder = 'Message (@ to include a file or symbol, or / for commands)'
```

I think the latter would be my preference.

toolmantim (commented):
Love that we're increasing the discoverability of symbol search!

I think there's some issues with updating the label that appears in the search result popover, but the placeholder update is a great idea.

We could also investigate some updates that add more explicit options, so then we can just use "@ to include code" as the placeholder:

CleanShot 2024-01-24 at 10 54 31@2x

vscode/src/non-stop/FixupTypingUI.ts
```diff
@@ -55,8 +55,8 @@ const MATCHING_CONTEXT_FILE_REGEX = /@(\S+)$/
 const MATCHING_SYMBOL_REGEX = /@#(\S+)$/
 
 const MAX_FUZZY_RESULTS = 20
-const FILE_HELP_LABEL = 'Search for a file to include, or type # to search symbols...'
-const SYMBOL_HELP_LABEL = 'Search for a symbol to include...'
+const FILE_HELP_LABEL = 'Start typing a file to include, or use @# to search for symbols...'
+const SYMBOL_HELP_LABEL = 'Start typing a symbol to include...'
```
toolmantim, line 59:
```suggestion
 const FILE_HELP_LABEL = 'Search for a file to include, or type # to search symbols...'
 const SYMBOL_HELP_LABEL = 'Search for a symbol to include...'
```

vscode/test/e2e/at-file-context-selecting.test.ts
```diff
@@ -18,7 +18,7 @@ test('@-file empty state', async ({ page, sidebar }) => {
     await chatInput.fill('@')
     await expect(
         chatPanelFrame.getByRole('heading', {
-            name: 'Search for a file to include, or type # to search symbols...',
+            name: 'Start typing a file to include, or use @# to search for symbols...',
```
toolmantim, line 21:
```suggestion
            name: 'Search for a file to include, or type # to search symbols...',
```

vscode/webviews/Chat.tsx
taylorsperry: I think we need the `@#` instruction here because otherwise people don't know how to include a symbol. That's what I was getting at in the description: 

> For one, I initially was trying to just type # to include a symbol (the current language doesn't make it clear that you actually need @#), but this is also something we've heard [internally](https://sourcegraph.slack.com/archives/C04MZPE4JKD/p1705996369006999?thread_ts=1705105369.141479&cid=C04MZPE4JKD) and from a user DM in Discord.

vscode/test/e2e/at-file-context-selecting.test.ts
```diff
@@ -45,7 +45,7 @@ test('@-file empty state', async ({ page, sidebar }) => {
     // Symbol empty state
     await chatInput.fill('@#')
     await expect(
-        chatPanelFrame.getByRole('heading', { name: 'Search for a symbol to include..' })
+        chatPanelFrame.getByRole('heading', { name: 'Start typing a symbol to include...' })
```
toolmantim, line 48:
```suggestion
        chatPanelFrame.getByRole('heading', { name: 'Search for a symbol to include..' })
```

vscode/webviews/UserContextSelector.tsx
```diff
@@ -36,16 +36,16 @@ export const UserContextSelectorComponent: React.FunctionComponent<
 
     let headingTitle: string | undefined
     if (formInput.endsWith('@')) {
-        headingTitle = 'Search for a file to include, or type # to search symbols...'
+        headingTitle = 'Start typing a file to include, or use @# to search for symbols...'
```
toolmantim, line 39:
```suggestion
        headingTitle = 'Search for a file to include, or type # to search symbols...'
```

vscode/webviews/UserContextSelector.tsx
taylorsperry: Fair enough. The misinterpretation of "Search" was just one data point, and I agree with you that "Start" is kind of redundant. 

> 
The same user interpreted "search" literally and tried adding a symbol with command+F.

toolmantim, line 48:
```suggestion
         headingTitle = 'Search for a symbol to include...'
     } else if (formInput.match(/@[^ #]+$/)) {
         headingTitle = contextSelection?.length
             ? 'Search for a file to include...'
             : 'No matching files found'
     } else if (formInput.match(/@#[^ ]+$/)) {
         headingTitle = contextSelection?.length
             ? 'Search for a symbol to include...'
```

toolmantim (approved):

vscode/webviews/UserContextSelector.tsx
```diff
@@ -38,14 +38,14 @@ export const UserContextSelectorComponent: React.FunctionComponent<
     if (formInput.endsWith('@')) {
         headingTitle = 'Search for a file to include, or type # to search symbols...'
     } else if (formInput.endsWith('@#')) {
-        headingTitle = 'Search for a symbol to include...'
+        headingTitle = 'Start typing a symbol to include...'
```
toolmantim, line 41:
```suggestion
        headingTitle = 'Search for a symbol to include...'
```

vscode/webviews/UserContextSelector.tsx
```diff
@@ -38,14 +38,14 @@ export const UserContextSelectorComponent: React.FunctionComponent<
...
     } else if (formInput.match(/@[^ #]+$/)) {
         headingTitle = contextSelection?.length
-            ? 'Search for a file to include...'
+            ? 'Start typing a file to include...'
```
toolmantim, line 44:
```suggestion
            ? 'Search for a file to include...'
```

vscode/webviews/UserContextSelector.tsx
```diff
@@ -38,14 +38,14 @@ export const UserContextSelectorComponent: React.FunctionComponent<
...
             : 'No matching files found'
     } else if (formInput.match(/@#[^ ]+$/)) {
         headingTitle = contextSelection?.length
-            ? 'Search for a symbol to include...'
+            ? 'Start typing a symbol to include...'
```
toolmantim, line 48:
```suggestion
            ? 'Search for a symbol to include...'
```

toolmantim (approved):

PR #2869, opened 2024-01-23
refactor: commands

abeatrix:
Refactor and clean up the existing commands set up, update and add new e2e tests to cover commands and custom commands.

Note: A lot of the files were updated due to attempts to clean up and unify method names, commands, and file names. 
For example:
- "cody.action.commands.custom.menu" -> "cody.menu.custom-commands"
- "cody.action.commands.menu" -> "cody.menu.commands"
- "cody.action.commands.exec" -> "cody.action.command"
- "CustomPrompts" -> "CustomCommands"
- file name  "CommandRunnerr" -> runner.ts etc

The major change in this PR is moving `command` to the top level and decouple it from chat and edit. This allows other services to execute a command anywhere through:
- [executeCodyCommand](https://github.com/sourcegraph/cody/blob/62c923f94641dd33d335dc8dabb53bd37650f419/vscode/src/commands/CommandsController.ts#L79): a function that is tied to the execution step in the `CommandsController` class.
  1. When a command is executed, the request will be sent to the `CommandsController`
  2. CommandsController will look up the command based on the input, before creating a `CommandRunner`
  3. Each `CommandRunner` handles identifying if the command is a `chat request` or an `edit request`, then fetch for the relevant context before sending creating the request via [executeChat](https://github.com/sourcegraph/cody/blob/62c923f94641dd33d335dc8dabb53bd37650f419/vscode/src/commands/default/ask.ts) for chat request and [executeEdit](https://github.com/sourcegraph/cody/blob/62c923f94641dd33d335dc8dabb53bd37650f419/vscode/src/edit/execute.ts) for edit request respectively. 
- 0r via [vscode.commands.executeCommand('cody.action.command')](https://github.com/sourcegraph/cody/blob/62c923f94641dd33d335dc8dabb53bd37650f419/vscode/src/main.ts#L308-L339): a global command that is connected to `executeCodyCommand`.
- All the default commands can also be initiated by other services via the vscode commands registered in [main.ts](https://github.com/sourcegraph/cody/blob/62c923f94641dd33d335dc8dabb53bd37650f419/vscode/src/main.ts#L339-L344).

Summary:
- Context used for custom commands are broken down into individual files inside `vscode/src/commands/context`
- Default commands are broken down into individual files inside `vscode/src/commands/default`
- Command controller now handles the sorting and execution at top level
- Slash commands support have been removed from chat
- All the command-handling methods have been removed from the chat panel class
- All the command-context fetching methods have been removed from the chat context provider class

Checklist:
- [x] moved command execution step to top level 
- [x] decouple commands from chat
- [x] decouple default commands from custom commands
- [x] remove /ask command from chat
- [x] remove slash commands from chat
- [x] unify command menus
- [x] support codebase as custom command context via `addEnhancedContext` in `cody.action.chat` command
- [x] remove command handling from chat panel
- [x] remove get command context from chat
- [x] command runner handles context fetching and categorizing command before sending command prompt with context as either a chat request or inline edit request
- [x] rename files
- [x] replace "custom prompt" with "custom commands"
- [x] clean up command controller
- [x] clean up command provider
- [x] clean up command runner
- [x] clean up command context 
- [x] clean up custom command store
- [x] update integration tests for commands
- [x] add  e2etests to cover custom command menu
- [x] add e2e tests to cover executing custom command defined in cody.json file
- [x] add e2e tests to cover context defined for a custom command
- [x] add e2e test to cover default command selection context

## Test plan

Try executing 
- the default chat commands (explain, test, smell)
- default edit command (doc, new test command)
- a custom command


agent/src/index.test.ts
```diff
@@ -961,61 +965,57 @@ describe('Agent', () => {
...
-              " Okay, based on the shared context, it looks like:
-
-              - The test framework in use is Vitest
-              - The assertion library is vitest's built-in expect
-
-              No additional packages or imports are needed.
+                  " Okay, based on the shared context, it looks like Vitest is being used as the test framework. No mocks are detected.
 
-              I will focus on testing the Animal interface by validating:
+                  Here are some new unit tests for the Animal interface in src/animal.ts using Vitest:
```
abeatrix, line 970: the test is failing because of the spacing in this test, trying to fix it...

abeatrix: Also looking into the new e2e test for custom command that's failing on the CI, but works locally for me 😡 

image

vscode/src/commands/menus/index.ts
```diff
@@ -1,214 +1,156 @@
...
-                    str?.toLowerCase().includes(normalizedValue)
+        // Open or delete custom command files
+        quickPick.onDidTriggerItemButton(item => {
+            const selected = item.item as CustomCommandsItem
+            if (selected.type) {
+                void commands.executeCommand(
+                    item.button.tooltip?.startsWith('Delete')
+                        ? 'cody.commands.delete.json'
+                        : 'cody.commands.open.json',
+                    selected.type
```
umpox, line 95: I think we also need to handle actioning the actual items in `onDidAccept`, not just the buttons:

image

Clicking these doesn't seem to do anything

umpox (commented):
Took a high level look over the code and it's looking really good. Left some minor comments and a few possible bugs I noticed

abeatrix:
> Right click -> Explain code seems to show the full explain prompt, instead of `/explain`. Is that intended? Also happens with Smell code.
> 
> image
> The cog in the Command palette seems to open the custom commands menu. Is that expected? I would have thought it would open VS Code settings 🤔

@umpox Thanks for the thorough review!

> Right click -> Explain code seems to show the full explain prompt, instead of /explain. Is that intended? Also happens with Smell code.

That's intended as part of removing slash commands from chat. This allows user to see the underlaying prompt and tweak it via the new edit buttons as right now, we are disabling the edit buttons on command messages due to this issue.

> The cog in the Command palette seems to open the custom commands menu. Is that expected? I would have thought it would open VS Code settings 🤔

Yea that's the current behavior! I do think we can give this menu a makeover in a follow up, but i've updated the menu so that it runs edit command by default instead of ask :D

vscode/src/commands/menus/index.ts
abeatrix: fixed, and created a e2e test to cover this. Thanks for catching this!

umpox (approved):
Thanks! and nice tests 👍

agent/src/index.test.ts
```diff
@@ -645,39 +633,40 @@ describe('Agent', () => {
...
+                                    null,
+                                    2
+                                )}`
+                            )
+                        }
+                        await client.request('command/execute', acceptCommand.command)
+                        expect(client.codeLenses.get(uri.toString()) ?? []).toHaveLength(0)
+                        const newContent = client.workspace.getDocument(uri)?.content
+                        assertion(trimEndOfLine(newContent))
+                    } catch (error) {
```
olafurpg, line 660: Is this intentional? I expect the test to fail if one of the assertions

olafurpg (commented):
@abeatrix the very large addition to the recording file (+15k lines) hints to me that we're adding a lot of unused recordings. The `pnpm update-agent-recordings` command is supposed to clean up unused recordings but I'm sensing it's not working 100% correctly yet.

You can run this command to reset the commands from `origin/main`
```
./agent/scripts/reset-recordings-to-main.sh
```

I don't see any new LLM requests in the updated tests that could explain why we need 15k new lines to the recording file.

olafurpg:
FYI I figured out the bug with `keepUnusedRecording` so that `pnpm update-agent-recordings` no longer keeps adding new recordings https://github.com/sourcegraph/cody/pull/2926

PR #2870, opened 2024-01-23
Agent: add support for inline edit command "Document code"

olafurpg:
Previously, the agent only supported commands that appeared in the chat view (explain, smell and test). This PR adds support for a new "Document code" command that can automatically insert a docstring into the file. This functionality requires expanding the agent protocol to allow the server to update the document contents:

- `commands/document`: request sent from client to server to trigger "Document code" command. This command returns a "task ID", which we can use to track progress on this command. 
- `editTaskState/didChange`: notification sent from the server to the client to notify when the task enters a new phase (applying, formatting, applied, finished, errored, ...).
- `textDocument/edit`: request sent from the server to the client to apply a diff against the file contents
- `codeLenses/display`: notification sent from server to client to display code lenses (buttons to accept/retry/undo generation of docstring)
- `textDocument/didSave`: request sent from the client to the server to notify the user has saved the file. Cody uses this as an implicit signal that the docstring was accepted by the user. Users can also explicitly accept via code lenses.

Additionally, this PR adds two now properties `ClientInfo.capabilities.{codeLenses,edit}` for the `initialize` handshake to allow the client to opt-into supporting the edit/codeLens requests.

Finally, to get the functionality working as expected, this PR implements folding ranges with a basic algorithm that follows indentation.

## Test plan

See updated tests.


steveyegge (approved):
Thanks for all the tests!

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -136,4 +143,49 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
...
+            selections: [selection],
+            edit: (callback, options) => {
+                if (this.params?.edit) {
+                    return this.params.edit(document.uri, callback, options)
+                }
+                logDebug('AgentTextEditor:edit()', 'not supported')
+                return Promise.resolve(false)
+            },
+            insertSnippet: () => Promise.resolve(true),
+            revealRange: () => {}, // TODO: implement this for inline edit commands?
```
umpox, line 176: `revealRange(range, vscode.TextEditorRevealType.InCenterIfOutsideViewport)`

Will also be very helpful for letting users manually adjust their selection. E.g. keying through available symbols to decide what to unit test

agent/src/index.test.ts
```diff
@@ -979,6 +1099,126 @@ describe('Agent', () => {
...
+                   * @param b - The second number to sum.
+                   * @returns The sum of a and b.
+                   */
+                  export function sum(a: number, b: number): number {
+                      /* CURSOR */
+                  }
+                  "
+                `)
+            )
+
```
umpox, line 1146: These tests are so good, ty!

vscode/src/non-stop/FixupController.ts
```diff
@@ -294,6 +294,7 @@ export class FixupController
             { ...codeCount, source },
             { hasV2Event: true }
         )
+        logDebug('FixupController:apply', new Error().stack ?? '')
```
umpox, line 297: @olafurpg Presuming this was added accidentally: pr to remove https://github.com/sourcegraph/cody/pull/2907

umpox (commented):
Thanks for doing this!

vscode/src/non-stop/FixupController.ts
olafurpg: Whoops, yes this should be removed 🙈

PR #2877, opened 2024-01-24
Agent: update squirrel test to assert context files

olafurpg:
Previously, the squirrel test didn't assert that the context files were empty. Also, the squirrel test explicitly opened the squirrel.ts file so it wasn't stressing the code path where symf context returns results from unopened files. This PR fixes both problems, we remove the explicit `textDocument/didOpen` notification for squirrel.ts and we add an assertion that squirrel.ts is part of the context files in the transcript.

## Test plan

See updated tests.

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -22,6 +22,7 @@ import type {
     ServerInfo,
     WebviewPostMessageParams,
 } from './protocol-alias'
+import { URI } from 'vscode-uri'
 
 type ProgressMessage = ProgressStartMessage | ProgressReportMessage | ProgressEndMessage
 interface ProgressStartMessage {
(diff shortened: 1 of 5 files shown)


agent/src/index.test.ts
```diff
@@ -201,6 +202,26 @@ export class TestClient extends MessageHandler {
         return this.sendMessage(id, text, params)
     }
 
+    public async sendSingleMessageToNewChatWithFullTranscript(
```
pkukielka, line 205: Can you reuse that function in the `public async sendMessage`?
Now they are in 95% duplicates.

olafurpg: Good catch! I was planning to do this but forgot. Pushed a commit increasing reuse. The test helper functions are starting to look quite messy, probably worth cleaning up even more aggressively in a separate PR

pkukielka (approved):
LGTM

PR #2880, opened 2024-01-24
Autocomplete: add test utility for completions streaming

valerybugakov:
## Context

- Adds a new autocomplete testing utility, `getInlineCompletionsWithInlinedChunks,` which aims to reduce the boilerplate code required to test completions generated from multiple chunks received during HTTP streaming.
     - On the separator symbol: I want to reduce the number of symbols we need to copy-paste manually to annotate code sections as prefix, suffix, and completion chunks; that's why I propose using only one symbol for annotation. There's no ambiguity regarding how this annotation can be interpreted, and it allows us to add new tests quickly (or change existing ones). It also makes it easier to understand what the final code would look like. 
     - I tried different symbols from the emoji keyboard, like ⎮, ⌷, ⍁, etc. But I didn't find a symbol that would provide as good readability as the one we use right now (because it's bold and makes it easy to identify parts of the code). I'm happy to switch to whatever other symbol is proposed. 
- Migrate hot-streak and dynamic multiline completions integration tests to this utility.
- Adds new dynamic multiline completions integration tests to cover more cases.

## Test plan

CI


philipp-spiess (approved):
Ok I really like it

PR #2881, opened 2024-01-24
Autocomplete: adaptive dynamic multiline completions

valerybugakov:
## Context

- Adds a timeout to dynamic multiline completions.
- Adds multiline completions support to hot streak completions.
- Adds support to indentation updates in hot streak completions.
- Fixes the truncation of consecutive multiline blocks in hot streak completions
- Adds tests for all the use cases above (I'm still adding several unit tests for isolated cases).
- Unifies the "insert-into-document-context" logic partially duplicated for dynamic multiline completions and hot streaks.
- Built on top of https://github.com/sourcegraph/cody/pull/2880

The updated flow:
1. Set a timeout for all completions. For now, the value is based on multiline latency averages from Looker
2. Begin streaming and check if it's a multiline completion start.
3. If yes, continue generating new chunks.
4. [**Timeout usage**] If timeout elapses, present the singleline completion and keep streaming in the background.
5. [**Multiline support for hot streaks**] Once background steaming is finished, add the multiline completion to our cache for instant access if the user continues typing as suggested.

The great news is that this doesn’t affect small dynamic multiline completions that generate quickly, preserving Cody’s intelligent feel!

## Test plan

1. CI and updated unit tests
2. Manually test the updated logic with the dynamic multiline completions enabled. Long dynamic multiline completion should be cut off early, as shown on the screen recording in [this Slack message](https://sourcegraph.slack.com/archives/C05AGQYD528/p1705666719813919).


valerybugakov:
cc @philipp-spiess for early feedback, I'm still adding addition tests, but the core functionality is here

vscode/src/completions/providers/hot-streak.ts
```diff
@@ -22,90 +29,111 @@ export interface HotStreakExtractor {
...
+function insertCompletionAndPressEnter(
+    docContext: DocumentContext,
+    completion: InlineCompletionItemWithAnalytics,
+    document: TextDocument,
+    dynamicMultilineCompletions: boolean
+): DocumentContext {
+    const { insertText } = completion
+    const { languageId, uri } = document
+
+    const startsNewBlock = Boolean(endsWithBlockStart(insertText, languageId))
```
philipp-spiess, line 46: ohh this should help a lot with hit rate, see https://sourcegraph.slack.com/archives/C05AGQYD528/p1706108513310449

philipp-spiess: I wonder if instead looking at the same line, we can actually look at the next line of generated hot-streak completion for this? I.e wouldn't the generated completion also contain the right indentation? This would then also work for closing already 🤔

vscode/src/completions/providers/hot-streak.ts
```diff
@@ -22,90 +29,111 @@ export interface HotStreakExtractor {
...
+    const newBlockIndent = startsNewBlock ? getEditorIndentString(uri) : ''
+    const currentIndentReference = insertText.includes('\n')
+        ? getLastLine(insertText)
+        : docContext.currentLinePrefix
+
+    const indentString = '\n' + detectIndent(currentIndentReference).indent + newBlockIndent
+    const insertTextWitPressedEnter = insertText + indentString
```
philipp-spiess, line 53:
```suggestion
    const insertTextWithPressedEnter = insertText + indentString
```

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -34,54 +34,71 @@ export async function* fetchAndProcessDynamicMultilineCompletions(
...
 
+    const generatorStartTime = performance.now()
+
     for await (const { completion, stopReason } of completionResponseGenerator) {
-        const isFinal = stopReason !== STOP_REASON_STREAMING_CHUNK
-        const extractCompletion = isFinal ? parseAndTruncateCompletion : canUsePartialCompletion
+        const isFirstCompletionTimeoutElapsed =
+            performance.now() - generatorStartTime >= firstCompletionTimeout
+        const isFullResponse = stopReason !== STOP_REASON_STREAMING_CHUNK
+        const shouldFirstCompletion = isFullResponse || isFirstCompletionTimeoutElapsed
```
philipp-spiess, line 83:
```suggestion
        const shouldYieldFirstCompletion = isFullResponse || isFirstCompletionTimeoutElapsed
```

Feels like there's a verb missing? 🤔

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -34,54 +34,71 @@ export async function* fetchAndProcessDynamicMultilineCompletions(
...
         }
     }
 
+    const generatorStartTime = performance.now()
+
     for await (const { completion, stopReason } of completionResponseGenerator) {
-        const isFinal = stopReason !== STOP_REASON_STREAMING_CHUNK
-        const extractCompletion = isFinal ? parseAndTruncateCompletion : canUsePartialCompletion
+        const isFirstCompletionTimeoutElapsed =
+            performance.now() - generatorStartTime >= firstCompletionTimeout
```
philipp-spiess, line 81: I think this should also check for wether we have received the first newline in the response already. Otherwise this might yield a completion of an incomplete line.

Or is this already covered by the `getFirstLine` check below?

vscode/src/completions/text-processing/utils.ts
```diff
@@ -382,3 +382,9 @@ export function getPrevNonEmptyLine(prefix: string): string {
 export function lines(text: string): string[] {
     return text.split(/\r?\n/)
 }
+
+export function hasCompleteFirstLine(text: string): boolean {
+    const lastNewlineIndex = text.lastIndexOf('\n')
```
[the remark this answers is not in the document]
valerybugakov: Nope

vscode/src/utils.ts
```diff
@@ -43,3 +43,10 @@ export function getEditorTabSize(uri: vscode.Uri): number {
 
     return tabSize
 }
+
+export function getEditorIndentString(uri: vscode.Uri): string {
+    const insertSpaces = getEditorInsertSpaces(uri)
+    const tabSize = getEditorTabSize(uri)
+
+    return insertSpaces ? new Array(tabSize + 1).join(' ') : '\t'
```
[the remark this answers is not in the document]
valerybugakov: Yep!

vscode/src/completions/get-current-doc-context.ts
```diff
@@ -10,11 +10,28 @@ import {
...
+     * TODO(valery): we need a better abstraction that would allow us to mutate
+     * the `TextDocument` text in memory without actually pasting it into the `TextDocument`
+     * and that would not require copy-pasting and modifiying the whole document text
+     * on every completion update or new virtual completion creation.
+     */
+    injectedCompletionText?: string
+    /**
+     * Actually current cursor position that exists in the `TextDocument`.
+     */
+    actualPosition?: vscode.Position
```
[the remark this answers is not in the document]
valerybugakov: Thanks for raising this. I removed the field in this commit: https://github.com/sourcegraph/cody/pull/2881/commits/19a271478639842cd5197a2aac5153ea69b24c7f

vscode/src/completions/providers/hot-streak.ts
valerybugakov: > we can actually look at the next line of generated hot-streak completion for this?

Could you elaborate?

vscode/src/completions/providers/hot-streak.ts
valerybugakov: I re-enabled the spell checker; please forgive me!

vscode/src/completions/providers/fetch-and-process-completions.ts
valerybugakov: Yep, I do it here to continue the loop early: https://github.com/sourcegraph/cody/blob/4133837b7f1c829db42fa21abb4920716f5c8a73/vscode/src/completions/providers/fetch-and-process-completions.ts#L90-L92

vscode/src/completions/providers/hot-streak.ts
philipp-spiess: Yeah, sorry. So we have all this fancy logic here to find out what the indentation is that the next completion is going to be inserted at, but we do already have the exact indentation from the LLM response.

E.g. imagine we have these two lines, the first one is the "regular" completion and the next one the hot-streak:

```
const foo = fetch(
  'https://example.com',
```

When we parse the hot-streak, we actually trim the start of the completion so it'll have the content of `'https://example.com',` and then we reverse-engineer the whitespace based on the line above. 

I’m suggesting we just use the same whitespace as returned from the LLM! That would work for:

- When the hot-streak is on the same indentation (since the indentation the LLM generates will match the one VS Code uses when you press enter)
- When the hot-streak is indented - The use case you just adde in this PR
- When the hot-streak is unindented - Which is not handled at all right now

vscode/src/completions/providers/hot-streak.ts
philipp-spiess: noooo worriessss

vscode/src/completions/providers/fetch-and-process-completions.ts
philipp-spiess: Yep that makes sense!

vscode/src/completions/providers/hot-streak.ts
valerybugakov: Got it! I like the idea and looked into implementing it here, but it could be more challenging with our current setup. 

We need to account for the fact that the following line might not be finished when we do the hot streak extraction. Also, we need to factor in that the completion can be singleline or multiline, and there's no way of getting the truncated suffix of the completion to get the next line indentation. 

I [extracted](https://github.com/sourcegraph/cody/issues/2903) the idea into an issue and propose implementing it in a follow-up PR.

PR #2889, opened 2024-01-24
[WIP] update e2e tests to enforce stricter logging standards

akalia25:
Update e2e-tests to validate the events logged to an expected logged events list; to prevent regressions in our telemetry and analytics

## Test plan
test via `pnpm test:e2e` and CI

### DIFF
--- a/vscode/test/e2e/at-file-context-selecting.test.ts
+++ b/vscode/test/e2e/at-file-context-selecting.test.ts
@@ -4,10 +4,27 @@ import { expect } from '@playwright/test'
 
 import { isWindows } from '@sourcegraph/cody-shared'
 
+import { loggedEvents, resetLoggedEvents } from '../fixtures/mock-server'
+
 import { sidebarSignin } from './common'
-import { test } from './helpers'
+import { assertEvents, test } from './helpers'
 
 // Creating new chats is slow, and setup is slow, so we collapse all these into one test
+
+const expectedEvents = [
+    'CodyVSCodeExtension:auth:clickOtherSignInOptions',
+    'CodyVSCodeExtension:login:clicked',
+    'CodyVSCodeExtension:auth:selectSigninMenu',
+    'CodyVSCodeExtension:auth:fromToken',
+    'CodyVSCodeExtension:Auth:connected',
+    'CodyVSCodeExtension:chat-question:executed',
+    'CodyVSCodeExtension:chat-question:executed',
+    'CodyVSCodeExtension:Auth:connected'
+  ]
+
+test.beforeEach(() => {
+    void resetLoggedEvents()
+})
 
 test('@-file empty state', async ({ page, sidebar }) => {
     await sidebarSignin(page, sidebar)
@@ -116,6 +133,9 @@ test('@-file empty state', async ({ page, sidebar }) => {
     await chatInput.type('and @Main.ja', { delay: 50 })
     await chatInput.press('Tab')
     await expect(chatInput).toHaveValue('@Main.java and @Main.java ')
+    console.log(loggedEvents)
+    // check to see if loggedEvents == expectedEvents
+    await assertEvents(loggedEvents, expectedEvents)
 })
 
 function withPlatformSlashes(input: string) {


vscode/test/e2e/at-file-context-selecting.test.ts
```diff
@@ -2,11 +2,29 @@ import { expect } from '@playwright/test'
...
+
 import { sidebarSignin } from './common'
-import { test, withPlatformSlashes } from './helpers'
+import { assertEvents, test, withPlatformSlashes } from './helpers'
 
 // Creating new chats is slow, and setup is slow, so we collapse all these into one test
 
+// list of events we expect this test to log, add to this list as needed
+const expectedEvents = [
+    'CodyVSCodeExtension:auth:clickOtherSignInOptions',
```
dominiccooney, line 14: If the order is not important, does it make sense to sort these? Same feedback below.

vscode/test/e2e/at-file-context-selecting.test.ts
```diff
@@ -114,4 +132,8 @@ test('@-file empty state', async ({ page, sidebar }) => {
     await chatInput.type('and @Main.ja', { delay: 50 })
     await chatInput.press('Tab')
     await expect(chatInput).toHaveValue('@Main.java and @Main.java ')
+
+    // Critical test to prevent event logging regressions.
+    // Do not remove without consulting data analytics team.
+    await assertEvents(loggedEvents, expectedEvents)
```
dominiccooney, line 138: You should do this for every relevant test using `.extend`, make the set of events declarative, and add the assertions there. See vscode/test/e2e/helpers.ts for some examples of what the test framework can do.

In the end, the test should look like:

```
test.extend<helpers.ExpectedEvents>([
  'event/foo',
  'event/bar',
])('@-file empty state', async ({ page, sidebar }) => {
  ...
```

And if the set of events is common throughout a file, then you can save and reuse the `test.extend` result and it will be obvious what is happening. See the way vscode/test/e2e/local-embeddings.test.ts does `const test = ...` and sets the dotcom URL override, etc. for *all* the tests in that file?

dominiccooney (changes requested):
Let's lean into the test framework's extension mechanism so we don't need so much repeated before each/assertions in the test.

vscode/test/e2e/at-file-context-selecting.test.ts
akalia25: Thank you, @dominiccooney, for this insightful feedback! I'll work on making these changes and reduce redundancy! 🙇

vscode/test/e2e/attribution.test.ts
```diff
@@ -47,7 +49,7 @@ test('attribution search enabled in chat', async ({ page, sidebar }) => {
     await assertEvents(loggedEvents, expectedEvents)
```
dominiccooney, line 49: You can do these assertions in the exectedEvents handler.

vscode/test/e2e/helpers.ts
```diff
@@ -157,6 +162,15 @@ export const test = base
             await use(sidebar)
         },
     })
+    .extend<ExpectedEvents>({
+        expectedEvents: [
```
dominiccooney, line 166: Make this an async function, and you can put your check for expected events after the call to `use` instead of writing it in multiple functions:

```
  expectedEvents: async ({ expectedEvents }, use) => {
    use(expectedEvents)
    await assert...etc.
  },
})
```

dominiccooney (commented):
One more bit of feedback. We can remove the assertEvents call and do it in the extension as well. You can call extend multiple times with the same thing; the first one defines how the extension works, the later ones provide data to it. Look at how the dotcom URL extension works.

vscode/test/e2e/attribution.test.ts
akalia25: Thanks @dominiccooney 🙇 ! I'm trying to get that updated and I am a little confused if I am doing the assertion with the handler correctly.

I was able to update the `vscode/test/e2e/helpers.ts` to make it an async function:
```
    .extend<ExpectedEvents>({
        expectedEvents: async ({ expectedEvents }, use) => {
            use(expectedEvents)
            await assertEvents(loggedEvents, expectedEvents)
        },
    }
```

But just to confirm, i'll need to extend all the tests like so, to provide data to the extension? (i.e. tests would like so)
```
test
    .extend<ExpectedEvents>({
        // list of events we expect this test to log, add to this list as needed
        expectedEvents: [
            'CodyVSCodeExtension:auth:clickOtherSignInOptions',
            '...'
        ],
    })
    .extend<ExpectedEvents>({
        expectedEvents: async ({ expectedEvents }, use) => {
            await use(expectedEvents)
            await assertEvents(loggedEvents, expectedEvents)
        },
    })('requires a valid auth token and allows logouts', async ({ page, sidebar, expectedEvents }) => {
```
One call out is that `expectedEvents` parameter within the test (`('requires a valid auth token and allows logouts', async ({ page, sidebar, expectedEvents })`) is flagged as declared but not used, which can be misleading as it is important to have there otherwise it will not run our assertion. I'm concerned someone may remove it as a parameter and we'd lose test coverage pretty quickly. 
image

Am I approaching this the right way? Thank you.

dominiccooney: > i'll need to extend all the tests like so

No. See they way the test framework imports `test` from the framework, extends it, and then exports that as `test` and people use that? ([Here.](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@a1afc071da617f438c839485024d4c70e3ab315a/-/blob/vscode/test/e2e/helpers.ts?L34)) That's the way to hook it globally.

Then some tests want to override the events per-file, so they do the same thing at the file level.

And some tests want to specifically override it, and they call `.extend(...)(...)` just for the one test.

> One call out is that expectedEvents parameter within the test (('requires a valid auth token and allows logouts', async ({ page, sidebar, expectedEvents })) is flagged as declared but not used

You can use `expectedEvents` in the `page` hook which every test uses [here.](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@a1afc071da617f438c839485024d4c70e3ab315a/-/blob/vscode/test/e2e/helpers.ts?L66)

akalia25: Hey @dominiccooney thank you for explaining! I've made the changes to call `assertEvents`  globally and have `expectedEvents` in the `page` hook. This was a great suggestion and has simplified the need for writing `assertEvents` on each test 😮‍💨, thank you!! 

I also call `assertEvents` directly within the `test` instead of extending `test` and having the async function there. Let me know if that is not preferred.

vscode/test/e2e/at-file-context-selecting.test.ts
akalia25: Right now these are sorted in order of what events get logged by timestamp. I feel like this would be most helpful?

dominiccooney (approved):
Great work! This looks really succinct in the individual tests now, and the assertions in the base `test` ensure we'll run this assertion everywhere.

vscode/test/e2e/command-core.test.ts
```diff
@@ -57,20 +74,37 @@ test('Explain Command & Smell Command', async ({ page, sidebar }) => {
 
     // Smell Command
     // Running a command again should reuse the current cursor position
-    await expect(page.getByText('Find Code Smells')).toBeVisible()
-    await page.getByText('Find Code Smells').click()
+    await expect(page.getByText('Identify code smells')).toBeVisible()
```
abeatrix, line 77: you might have updated this file by mistake? This will fail because the titles have been updated.

PR #2891, opened 2024-01-25
Move context warnings from the UI to the debug log

toolmantim:
The context warnings in the UI are creating more confusion than they're worth, so we'll move them to the debug output and work on better ways of surfacing these issues as necessary in the transcript view.

## Test plan

- Recreate each warning scenario
- Verify code behaves well


dominiccooney (approved):
Looks good. Thank you for picking this up.

PR #2892, opened 2024-01-25
Clarify custom command codebase property removal changelog

toolmantim:
## Test plan

- N/A


dominiccooney (approved):
I'm not crazy about just deleting it because if you have a custom command with codebase context that stops using codebase context, this was your best lead before having to debug the extension yourself.

Can we mention that it has been disabled while we redesign the feature or something like that? It's honest.

dominiccooney (approved):
Splendid.

PR #2896, opened 2024-01-25
VS Code chat UI matches design

cbart:
Closes sourcegraph/sourcegraph#59294

[Figma](https://www.figma.com/file/H0k85e9AGWqSkiaExKVdyL/Guard-Rails-in-Cody-for-VS-Code?type=design&node-id=23-592&mode=design&t=pnlU3Q8qbm7N2Tq8-0)

[Long Loom showing demo local sg start](https://www.loom.com/share/c8cdef930af144e78584cc01e2882a7f?sid=00eac153-c720-4928-bae7-d44ec0567330)

Pass:
pass

Fail:
fail

Unavailable ([caveat](https://github.com/sourcegraph/sourcegraph/issues/59294#issuecomment-1910151871)):
unavailable-orange

## Test plan

- [x] Manual test
- [ ] Update automatic tests

For testing you got to enable `attribution.enabled` in your local site config as well as an experimental setting in VS Code:
Screenshot 2024-01-25 at 3 08 12 PM

### DIFF
--- a/lib/ui/src/chat/CodeBlocks.tsx
+++ b/lib/ui/src/chat/CodeBlocks.tsx
@@ -198,9 +198,15 @@ export const CodeBlocks: React.FunctionComponent<CodeBlocksProps> = React.memo(
                         attributionContainer.setAttribute('data-testid', 'attribution-indicator')
                         buttons.append(attributionContainer)
 
+                        const spinnerContainer = document.createElement('div')
+                        spinnerContainer.className = styles.spinner
+                        spinnerContainer.innerHTML = `<i class="codicon codicon-loading ${styles.codiconLoading}"></i>`
+                        buttons.append(spinnerContainer)
+
                         guardrails
                             .searchAttribution(preText)
                             .then(attribution => {
+                                spinnerContainer.innerHTML = '<i class="codicon codicon-pass"></i>'
                                 if (isError(attribution)) {
                                     attributionContainer.classList.add(styles.attributionIconUnavailable)
                                     attributionContainer.title = 'Attribution search unavailable.'
@@ -227,7 +233,9 @@ export const CodeBlocks: React.FunctionComponent<CodeBlocksProps> = React.memo(
                                 attributionContainer.title = 'Attribution not found.'
                             })
                             .catch(error => {
-                                console.error('promise failed', error)
+                                attributionContainer.classList.add(styles.attributionIconUnavailable)
+                                attributionContainer.title = `Attribution search unavailable. ${error.message}`
+                                return
                             })
                     }
 
--- a/lib/ui/src/chat/CodeBlocks.tsx
+++ b/lib/ui/src/chat/CodeBlocks.tsx
@@ -256,34 +256,35 @@ export const CodeBlocks: React.FunctionComponent<CodeBlocksProps> = React.memo(
                         eventMetadata
                     )
                     if (guardrails) {
-                        const flexFiller = document.createElement('div')
-                        flexFiller.classList.add(styles.flexFiller)
-                        buttons.append(flexFiller)
                         const container = document.createElement("div");
                         container.classList.add(styles.attributionContainer);
                         buttons.append(container);
 
                         const g = new GuardrailsStatusController(container);
-                        g.setPending()
+                        g.setPending();
 
                         guardrails
                             .searchAttribution(preText)
-                            .then(attribution => {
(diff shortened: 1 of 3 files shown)


cbart:
Need to enable attribution in prod gateway for testing https://github.com/sourcegraph/infrastructure/pull/5768 FYI @toolmantim

lib/ui/src/chat/CodeBlocks.tsx
```diff
@@ -154,6 +154,74 @@ function createCodeBlockActionButton(
     return button
 }
 
+class GuardrailsStatusController {
+    readonly statusSpinning = `<i class="codicon codicon-loading ${styles.codiconLoading}"></i>`
+    readonly statusPass = '<i class="codicon codicon-pass"></i>'
+    readonly statusFailed = 'Guard Rails Check Failed'
```
chwarwick, line 160: @toolmantim  do you think these will have meaning to users?  Honestly I'm not sure what I would be supposed to think if I saw this message but my chat otherwise worked and gave me the output I was looking for.

chwarwick (approved):

lib/ui/src/chat/CodeBlocks.tsx
toolmantim: There is also this tooltip: https://github.com/sourcegraph/cody/pull/2896/files#diff-2ae2bd0097186e3f0a9102eb087b71f31678cba0c19b4cc787a13d6119021cdaR215-R222

(from [Figma](https://www.figma.com/file/H0k85e9AGWqSkiaExKVdyL/Guard-Rails-in-Cody-for-VS-Code?type=design&node-id=21-167&mode=design&t=WyFFEOD5IpC800IP-4)):
image

But looking at it again with your comment, I think we should introduce the word "license" in there somewhere too.

`This code was found in ${n} ${n === 1 'repository' : 'repositories'} with incompatible licenses`?

toolmantim (commented):
Looking good @cbart! A few things…

Let's not title the whole bottom bar container, but instead we can put it on a container that wraps the two icons e.g. (also see `margin-left: auto` to avoid the need for the flex filler):

CleanShot 2024-01-26 at 08 49 19@2x

Can we also `font-size: 12px; line-height: 1;` the `this.status` element too?

To @chwarwick’s point about labelling, wdyt of introducing the word "license" here somewhere? e.g.

* Passed tooltip: "Guard Rails check passed. Code was not found in any incompatibly licensed repositories."
* Failed tooltip: "Guard Rails check failed. Code was found in 4 incompatibly licensed  repositories: xy/z, a/b/c, e/f/g, h/i/k, m/n/o, o/p/q."

lib/ui/src/chat/CodeBlocks.tsx
chwarwick: We don't check the license it's **_only_** this matches something found on dotcom.

abeatrix (approved):
🚀

cbart:
Thanks all!

Elements addressed:

- [x] Let's not title the whole bottom bar container
- [x] see `margin-left: auto` to avoid the need for the flex filler
- [x] `font-size: 12px; line-height: 1;` the this.status element

I did not address the following yet:

- [ ] Tooltip contents - talking about licensing might be misleading here, since we don't check licenses. We take a conservative approach that if anything is found in dotcom corpus, we notify the user - irrespective of what the license is. Let's talk more!
- [ ] Notifying user about being below threshold - filed https://github.com/sourcegraph/sourcegraph/issues/59892 to track.

aramaraju:
@toolmantim Is there an UX for the filtering out for autocomplete?

toolmantim:
@aramaraju there's no guardrails feedback UX designs at all for autocomplete… and our options in VS Code there are limited. What sort of a feedback would a user want as they're typing code in their code editor do you think? At the moment, if Cody doesn't have a suggestion to give… there's just simply no grey autocomplete text.

aramaraju:
@toolmantim If Cody autocompletes with code that has a match with viral OSS, does it not indicate to the user? Or do we simply prevent such suggestions from showing up for autocomplete?

toolmantim:
@cbart can confirm but I believe we simply prevent those suggestions from showing up

cbart:
Correct @toolmantim

aramaraju:
@cbart And I assume/hope this doesn't go into our CAR calculations. Is that right?

PR #2898, opened 2024-01-25
Autocomplete: More Jaccard improvements and per-line RRF

philipp-spiess:
This PR makes some changes to the new jaccard similarity retriever and implements per-line RRF:

- Jaccard similarity search will now no longer yield results that overlap in the lines. This was happening previously since we were creating results by sliding a window over the source document line by line and ranking it individually. That meant that lines with high overlap would result in multiple results and overlapping windows. We then had logic to specifically merge overlapping ranges. In practice though, often time this would run unbound and would end up having the whole document in one range after all ranges were merged. This then lead to only one context item being added that contains all of a file.
   
   We fix this by adding another linear pass through the ranked windows and doing some bookkeeping to know which lines were not part of any yielded result yet. Since this is an optimization problem and we only do one linear scan, this can end up not yielding results for every line in the document (i.e if the window size is 5 lines and we have the best matching snippet starting at line 3, we will never be returning a result for line 1 and 2). This, however, is not really relevant in reality as long as we contain the few highest ranked results per file.
   
   Merging was also removed to avoid combining lesser important results into a more important one and thus inflating the rank.

- The existing RRF implementation had a similar issue. Since we were doing the RRF on the individual documents, we ended up almost exclusively only picking snippets from a single file and adding a lot of low-ranking snippets into the mix. To fix this, this PR adds support for adding more than one ranking identifier per result. This means that we can rank on _individual lines_ instead of the whole file. When combining results, we would start with the highest ranked line across all documents and would mix in all snippets with a range overlapping that line. If there is a lower ranked snippet on a different line, it would not be picked.

I assume that both changes together would result in a much more healthy distribution of context snippets for Autocomplete so I’m thinking about rerunning maybe the local-mixed context fusion experiment. Additionally, I think this might fix some of the issues we're seeing with the new jaccard similarity (that right now is not stat. sig better or worse than the old implementation) 

## Test plan

 - Added test cases
 - Play around with our different context strategies and make sure nothing crashes


philipp-spiess:
@valerybugakov @olafurpg I realize this PR changes types that are used by the BFG interface. I'll need to do some investigation to see what the exact data is that bfg is returning here and if there's something that we can use to get a line range for the retrieved results. However I’m struggling right now to get any BFG hits. 🤔  Will look more into it tomorrow but if you have some ideas, lmk

olafurpg:
@philipp-spiess it's fine if we ignore bfg for now and we'll deal with updating the types when we have a new experiment to run

vscode/src/completions/context/reciprocal-rank-fusion.ts
```diff
@@ -0,0 +1,105 @@
...
+ * to be implemented that returns a set of identifiers for a document. A set is used to support our
+ * needs of varying context windows across Cody retrievers. In our case, these ranking identity
+ * would yield one identifier for every line of code affected by a context window.
+ *
+ * When combining the top ranked documents, we will make sure that no item is added twice (as
+ * denoted by their referential equality).
+ */
+export function fuseResults<T>(
+    retrievedSets: Set<T>[],
+    rankingIdentity: (result: T) => string[]
```
valerybugakov, line 20: Maybe `rankingIdentities` or something similar will highlight that it can return multiple results? This part was confusing to me until I dug into its application of it in the Jaccard-similarity retriever. 

Adding a more practical example to the comment with a couple of retrieved sets consisting of multiple lines of code will make it easier to grasp. AFAIU, we rank every line in a set and then, based on the sum of line ranks, decide the final order of the set.

vscode/src/completions/context/reciprocal-rank-fusion.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        // biome-ignore format: Make it clearly visible that there are two retrieved sets
+        const retrievers = [
+            new Set([{ id: 1 }, { id: 2 }]),
+            new Set([{ id: 2 }, { id: 3 }]),
+        ]
+        const rankingIdentity = (item: { id: number }) => [item.id.toString()]
+
+        const results = fuseResults(retrievers, rankingIdentity)
+
+        expect(results).toEqual(new Set([{ id: 2 }, { id: 2 }, { id: 1 }, { id: 3 }]))
```
valerybugakov, line 16: Explainer for dummies who see RRF for the first time:

```
// Retriever 1 has items with IDs 1 and 2. Retriever 2 has items with IDs 2 and 3.
// The item with ID 2 appears in both sets, making it more significant according to RRF. 
// Therefore, it gains a higher combined rank and appears first in the fused result.
```

vscode/src/completions/context/reciprocal-rank-fusion.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        const retrievers = [
+            new Set([{ lines: [1, 2, 3, 4] }, { lines: [7, 8] }]),
+            new Set([{ lines: [1, 2] }, { lines: [5, 6] }]),
+        ]
+        const rankingIdentity = (item: { lines: number[] }) => item.lines.map(id => id.toString())
+
+        const results = fuseResults(retrievers, rankingIdentity)
+
+        expect(results).toEqual(
+            new Set([{ lines: [1, 2, 3, 4] }, { lines: [1, 2] }, { lines: [7, 8] }, { lines: [5, 6] }])
```
valerybugakov, line 42: Based on the `expect` label, shouldn't it dedupe lines 1 and 2 from the results?

valerybugakov (approved):

vscode/src/completions/context/reciprocal-rank-fusion.test.ts
valerybugakov: Got it; thanks for adding extensive comments! 🙌

PR #2900, opened 2024-01-25
Workspace settings: Disable codeActionsOnSave

umpox:
## Desc

Takes precedence over user setting and causes annoying behaviour: https://sourcegraph.slack.com/archives/C05AGQYD528/p1706194762920939

(We can still enable these in user settings if people want them)

## Test plan

N/A - Dev setting only


philipp-spiess (approved):
TY!

PR #2902, opened 2024-01-25
feat(errors): add ServerError class to handle server-side aborts

abeatrix:
Context: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1706116514736789

feat(errors): add ServerError class to handle server-side aborts

- Implement a new ServerError class to identify errors like 'context deadline exceeded'.
- Add logic to skip user-facing display of `ServerError` instances in chat components.
- Handle ServerError in SimpleChatPanelProvider to post relevant errors to the transcript.

Follow-up:
- Add helpful tips to explain follow-up actions when a server error is encountered. For example, type `ask Cody to continue if your request was cut off`, or display a button for the user to press `continue` etc 
- Increase deadline to 2 minutes on the server side

## Test plan

- Switch to ChatGPT 4.0 as your chat model
- Use the /test command in a large method to ask Cody to generate Unit Test, as long as the response takes longer than 1 minute, your answer should be cut off
- Confirm the previous error message is not showing up 

### Before

image

### After

Message is cut off:

image

With no error message display that caused confusion:

image

### DIFF
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -79,6 +79,31 @@ export class NetworkError extends Error {
     }
 }
 
+/*
+ * Error for incompleted requests aborted by the server side error
+ * E.g. context deadline exceeded, request timeout, etc.
+ */
+export class ServerError extends Error {
+    public static readonly errorName = 'ServerError'
+    public readonly name = ServerError.errorName
+
+    /**
+     * Checks if the given error message indicates a request that was aborted by the server,
+     * such as from a context deadline exceeded.
+     *
+     * @param message The error message to check
+     * @returns True if the message matches a known incompleted error, false otherwise
+     */
+    public static isServerError(message: string): boolean {
+        const knownErrors = ['context deadline exceeded']
+        return knownErrors.includes(message)
+    }
+
+    constructor(public readonly message: string) {
+        super(message)
+    }
+}
+
 export function isNetworkError(error: Error): error is NetworkError {
     return error instanceof NetworkError
 }
--- a/lib/ui/src/chat/ErrorItem.tsx
+++ b/lib/ui/src/chat/ErrorItem.tsx
@@ -5,6 +5,7 @@ import { RateLimitError, type ChatError } from '@sourcegraph/cody-shared'
 import type { ApiPostMessage, ChatButtonProps, UserAccountInfo } from '../Chat'
 
 import styles from './ErrorItem.module.css'
+import { ServerError } from '@sourcegraph/cody-shared/src/sourcegraph-api/errors'
 
 /**
  * An error message shown in the chat.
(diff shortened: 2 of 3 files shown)


toolmantim:
Initially I didn't realise that the actual message was being cut off as well. In that case, I don't think we want to hide it completely from users, but mention at the end of the transcript message that it was cut-off due to it reaching the maximum response time.

Perhaps: "_Message stream ended early due to 2 minute maximum response duration._"

If that's too difficult, and we think this will be rare, than those words in an error message in the chat would also work.

lib/ui/src/chat/ErrorItem.tsx
```diff
@@ -26,6 +27,12 @@ export const ErrorItem: React.FunctionComponent<{
         )
     }
 
+    // Skip errors that should not be user-facing
+    if (typeof error !== 'string' && error.name === ServerError.errorName) {
```
valerybugakov, line 31: Would it be nice to have a type-guard function as we have for other errors? E.g., `isNetworkError`

lib/shared/src/sourcegraph-api/errors.ts
```diff
@@ -79,6 +79,31 @@ export class NetworkError extends Error {
...
+     * @returns True if the message matches a known incompleted error, false otherwise
+     */
+    public static isServerError(message: string): boolean {
+        const knownErrors = ['context deadline exceeded']
+        return knownErrors.includes(message)
+    }
+
+    constructor(public readonly message: string) {
+        super(message)
+    }
```
valerybugakov, line 104: Can it be omitted since `Error` has the same constructor?

lib/shared/src/sourcegraph-api/errors.ts
```diff
@@ -79,6 +79,31 @@ export class NetworkError extends Error {
     }
 }
 
+/*
+ * Error for incompleted requests aborted by the server side error
+ * E.g. context deadline exceeded, request timeout, etc.
+ */
+export class ServerError extends Error {
+    public static readonly errorName = 'ServerError'
+    public readonly name = ServerError.errorName
```
valerybugakov, line 88: This field will be redundant if we implement the `instanceof`-based type-guard function.

abeatrix:
> Initially I didn't realise that the actual message was being cut off as well. In that case, I don't think we want to hide it completely from users, but mention at the end of the transcript message that it was cut-off due to it reaching the maximum response time.
> 
> 
> 
> Perhaps: "_Response time limit reached (2 minutes). Message may be truncated._"
> 
> 
> 
> If that's too difficult, and we think this will be rare, than those words in an error message in the chat would also work.

Yea I agree, it's on my follow up to display an helpful message instead of the correct error. 
I think the error message is causing a lot of confusion to the user hence the complains. In chatgpt web chat, the assistant will stop mid-sentence when the response is too long but there's no error message displayed and user'd know it's stopped because of the long response, which is quite common imo

PR #2904, opened 2024-01-26
fix: do not display non-chat view in panel

abeatrix:
Reported by https://sourcegraph.slack.com/archives/C05AGQYD528/p1706206513677179

## Test plan

Provided by cezary on slack:
- User is signed out of Cody (I just Cmd+Shift+P > Sign out'd)
- I click on the yellow Cody icon in the bottom
- Signup screen appears on the right hand side in a pane
- Clicking sign into enterprise instance seems to have no effect.

After this change: sidebar will be opened with auth view instead of chat panel

https://github.com/sourcegraph/cody/assets/68532117/5a9a8e15-7318-402f-bf66-8e741928f3c8


cbart (approved):
Works! Thanks!

PR #2906, opened 2024-01-26
Autocomplete: fix actual position calculation for hot-streak completions

valerybugakov:
## Context

- As a follow-up to https://github.com/sourcegraph/cody/pull/2881, I started adding more tests. I bumped into the issue that I had created with the last-minute fix in the PR where I removed the `actualPosition` from `docContext` and switched to computing it manually on the spot. The math was incorrect in some cases: we need to know the length of the string at the target line number to calculate the correct position. This PR reverts back to having this field to avoid the complexity of handling edge cases where the injected completion overwrites `currentLineSuffix`.
- It also adds a helper function `acceptFirstCompletionAndPressEnter`, to the return value of `getInlineCompletionsWithInlinedChunks`, making hot-streak tests even more compact and easier to read. 

## Test plan

CI

### DIFF
--- a/vscode/src/completions/providers/hot-streak.ts
+++ b/vscode/src/completions/providers/hot-streak.ts
@@ -71,13 +71,9 @@ function insertCompletionAndPressEnter(
         dynamicMultilineCompletions,
     })
 
-    updatedDocContext.positionWithoutInjectedCompletionText =
-        updatedDocContext.positionWithoutInjectedCompletionText || docContext.position
-    updatedDocContext.injectedCompletionText =
-        (docContext.injectedCompletionText || '') + insertTextWithPressedEnter
-
     return updatedDocContext
 }
+
 export function createHotStreakExtractor(params: HotStreakExtractorParams): HotStreakExtractor {
     const { completedCompletion, providerOptions } = params
     const {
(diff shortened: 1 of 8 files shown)


vscode/src/completions/providers/hot-streak.ts
```diff
@@ -41,30 +55,24 @@ function insertCompletionAndPressEnter(
...
-        languageId,
+        languageId: document.languageId,
         insertText: insertTextWithPressedEnter,
         dynamicMultilineCompletions,
     })
 
+    updatedDocContext.positionWithoutInjectedCompletionText =
+        updatedDocContext.positionWithoutInjectedCompletionText || docContext.position
     updatedDocContext.injectedCompletionText =
         (docContext.injectedCompletionText || '') + insertTextWithPressedEnter
```
philipp-spiess, line 78: What's the reason for these two mutations not to happen inside the `insertIntoDocContext` helper? I worry that if we mutate the doc context after the helper, different usages of the `insertIntoDocContext` API will start to diverge in behavior (ie. some forget to update `injectedCompletionText` etc)

philipp-spiess (approved):
I see now why this is necessary. Thanks for renaming the field!

vscode/src/completions/providers/hot-streak.ts
valerybugakov: Love this idea 👍

PR #2879, opened 2024-01-24
Add multi-repo search for enterprise, remove remote embeddings

dominiccooney:
- Removes remote embeddings from consumer and enterprise
- Adds remote search through the `getCodyContext` GraphQL endpoint as a new context provider for Enterprise
- The current file's repository (if any), or `cody.codebase`, is automatically included by this context source
- Adds a multi-repo picker to Enterprise chats; up to 9 manually selected remote repos are saved in the chat transcript
- New chats default to using the workspace roots' repositories for context

Known issues:
- Needs automated tests for VSCode

Fixes #2622, #2624, #2850

## Test plan

This change deletes a lot of unused code so naturally, run the unit tests, e2e tests, etc.

Manual test plan:

Remote embeddings have been removed for consumer users:

1. Clone a popular open source repository, such as `git@github.com/react/react.git`
2. Open VSCode and sign in to a consumer account
3. Start a chat
4. Verify that the enhanced context selector advertises Search (that is, symf) and Embeddings with "set up embeddings." You should *not* see "sourcegraph.com" (that is, remote embeddings.)
5. Open a file in the repository, Cmd-Shift-P, Cody Command: Edit Code and edit the code. The edit should succeed.

Enterprise changes, so *sign into an Enterprise account*, then:

Repo selection:

1. Open a new VSCode session with nothing in the workspace.
2. Start a chat.
3. Open the Enhanced Context Selector.
4. Verify that you see a message about no repositories.
5. Click Choose Repositories.
6. Verify the Repository Picker works: that you can pick up to 9 repositories manually, that you can add and remove repositories, that you can search repositories, that ESC or a click outside the picker does not affect the repositories chosen, etc.
7. Verify that the Enhanced Context Selector works: Mouse over the repository names and see the full name; click X to remove a repository; etc.

Chats use the selected repositories; chats save the selected repositories:

1. Start a chat and add some repositories.
2. Chat with Cody and verify that context is fetched from the selected repositories.
3. Close the chat.
4. Open a new chat and verify that the repository list is empty again.
5. Reopen the first chat and verify that the repository list are the ones selected in step 2.

Chats automatically use the repository workspace roots for context:

1. Open a clone of a repository indexed by the enterprise.
2. Start a chat.
3. Verify that the Enhanced Context Selector includes the relevant repository. (If this is a different repository, you may have `cody.codebase` set.)

Supports multiple workspace roots:

1. File, Add Folder to Workspace, select another folder which is a clone of a repository indexed by the enterprise.
[...]

PR #2908, opened 2024-01-26
Agent: reduce automatic requests to sourcegraph.com

olafurpg:
Previously, the agent initialized the VS Code extension against sourcegraph.com with an empty access token. We did this to ensure that initialization succeeded even with an invalid server endpoint. However, by doing so we introduced another problem: the extension sent a lot of network traffic to sourcegraph.com for enterprise accounts.

Now, we use the custom server endpoint even during initialization. This change eliminates all automatic requests to sourcegraph.com for the minimized test cases we have in the codebase (covers initialization and sending a chat message). This currently only works out of the box thanks to the PR https://github.com/sourcegraph/cody/pull/2879 that removes embeddings. Before that change, initialization was regularly reporting errors from `EmbeddingDetector`, which has now been removed.

## Test plan

Green CI. Manually search for "sourcegraph.com" in the updated recording file and confirm that all matches reference demo.sourcegraph.com.

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -1390,29 +1390,7 @@ describe('Agent', () => {
                 }
                 const paths = contextUris.map(uri => uri.path.split('/-/blob/').at(1) ?? '').sort()
 
-                expect(paths).toMatchInlineSnapshot(`
-              [
-                "client/branded/src/search-ui/input/BaseCodeMirrorQueryInput.tsx",
-                "client/branded/src/search-ui/input/experimental/suggestionsExtension.ts",
-                "client/web-sveltekit/src/routes/[...repo=reporev]/(validrev)/(code)/+layout.ts",
-                "client/web/src/enterprise/repo/enterpriseRepoContainerRoutes.tsx",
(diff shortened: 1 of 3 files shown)


agent/src/index.test.ts
```diff
@@ -1390,29 +1390,7 @@ describe('Agent', () => {
...
-                "doc/admin/config/webhooks/outgoing.md",
-                "docker-images/syntax-highlighter/crates/scip-syntax/src/bin/scip-local-nav.rs",
-                "go.mod",
-                "internal/authz/providers/perforce/cmd/scanprotects/README.md",
-                "internal/codeintel/sentinel/internal/store/observability.go",
-                "internal/oobmigration/downgrade_test.go",
-                "internal/usagestats/codehost_integration.go",
-              ]
-            `)
+                expect(paths).includes('cmd/symbols/squirrel/README.md')
```
olafurpg, line 1393: Simplified the assertion to make this test case pass more consistently. The test failed when I re-recorded because the number of returned files had changed.

chwarwick (approved):
Nice improvement

PR #2909, opened 2024-01-26
Agent: add ability to disable network requests to sourcegraph.com

olafurpg:
Previously, the agent sent log events to sourcegraph.com for enterprise accounts even if any of the following things were true:

- `cody.serverEndpoint` was set to the enterprise URL.
- `cody.telemetry.level` was set to `off`. This setting is, in fact, entirely ignored by the agent.

The root problem was that the v1 event logger was still logging one event to sourcegraph.com without any option to route this request only to the connected instance. This was problematic because some enterprise instances are airgapped and this request ended up timing out.

This PR fixes the problem by adding support to exclusively send logging events to the connected instance. This behavior can only be enabled by setting the environment variable `CODY_LOG_EVENT_MODE` to the value `'connected-instance-only'`. Previously, it defaulted to the value `'dotcom-only'` or `'all'` without any workaround. We are using an environment variable because that is enough to unblock JetBrains. It probably makes sense to add a similar setting for the VSC extension.

Paired with @chwarwick on this PR
## Test plan

See updated enterprise test case that adds an assertion that we don't send any requests to other URLs than the connected instance.

### DIFF
--- a/lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
+++ b/lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
@@ -52,7 +52,7 @@ export class TelemetryRecorderProvider extends BaseTelemetryRecorderProvider<
                 }`,
                 clientVersion: extensionDetails.version,
             },
-            process.env.CODY_SHIM_TESTING === 'true'
+            process.env.CODY_TELEMETRY_EXPORTER === 'testing'
                 ? new TestTelemetryExporter()
                 : new GraphQLTelemetryExporter(client, anonymousUserID, legacyBackcompatLogEventMode),
             [
--- a/agent/recordings/enterpriseClient_3965582033/recording.har.yaml
+++ b/agent/recordings/enterpriseClient_3965582033/recording.har.yaml
@@ -63,7 +63,7 @@ log:
         cookies: []
         headers:
           - name: date
-            value: Fri, 26 Jan 2024 13:34:50 GMT
+            value: Fri, 26 Jan 2024 14:31:06 GMT
           - name: content-type
             value: text/event-stream
           - name: transfer-encoding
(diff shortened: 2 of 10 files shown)


chwarwick (approved):

lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
```diff
@@ -52,7 +52,7 @@ export class TelemetryRecorderProvider extends BaseTelemetryRecorderProvider<
                 }`,
                 clientVersion: extensionDetails.version,
             },
-            process.env.CODY_SHIM_TESTING === 'true'
+            process.env.CODY_TELEMETRY_EXPORTER === 'testing'
                 ? new TestTelemetryExporter()
                 : new GraphQLTelemetryExporter(client, anonymousUserID, legacyBackcompatLogEventMode),
```
bobheadxi, line 57: @olafurpg V2 telemetry only logs to connected instance, do we need this condition here?

bobheadxi: FYI we only land in the `legacy` mode to log `legacyBackcompatLogEventMode` here: https://github.com/sourcegraph/cody/pull/2909/files#diff-0982c52257a0ac820857d81a1a4efdd5c6ec1a7fc1ae62a329b9f0b2fc188a01R66-R70 for instances <5.2.5

bobheadxi: Context: we have a lot of "is telemetry enabled" toggling in https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@7d385240264090f7bfb0d502a76049fd6a05de58/-/blob/vscode/src/services/telemetry-v2.ts?L39, and it would probably be better if we avoid spreading things out in different parts of the telemetry stack, and configure everything in one place.

cc @akalia25

PR #2910, opened 2024-01-26
Add notification of expired/nearly expired Cody Pro

DanTup:
This borrows some changes from https://github.com/sourcegraph/cody/pull/2858 so may need rebasing when that lands (and the bots will be red since those methods are missing from this PR and not yet in main).

The notifications might need some tweaking - they currently look like this:

Screenshot 2024-01-26 153431

Screenshot 2024-01-26 153456

Some questions:

- [x] Currently this uses only the two FFs and does not look at the date fields at all, is this correct?
  - This means the code will need updating if another trial happens in future (or trial dates differ per user). It may be better to use the dates from the users account instead, but if we do this do we ignore the `CodyProTrialEnded` flag?
- [ ] Is 30mins a reasonable check frequency for if the `UseSscForCodySubscription` FF is not yet set?
- [x] Are the URLs accurate? They currently error, but work if I remove "accounts" subdomain
- [x] Prompts are not shown immediately after logging in with an access token because we fetch FeatureFlags before the API client is updated.. needs a fix (https://sourcegraph.slack.com/archives/C05AGQYD528/p1706872864488829)

## Test plan

- CI will run new tests `vscode\src\notifications\cody-pro-expiration.test.ts`

To manually test, you need the `UseSscForCodySubscription` feature flag set on your account, and the value of  `CodyProTrialEnded` will determine which dialog you see. You can test this manually without your account by adding the following code to `FeatureFlagProvider.evaluateFeatureFlag`:

```ts
if (flagName === FeatureFlag.UseSscForCodySubscription) return Promise.resolve(true)
if (flagName === FeatureFlag.CodyProTrialEnded) return Promise.resolve(true) // or comment out for nearly expired
```

If you click one of the buttons or the "x" we'll remember not to show the notification again, so to reset that for testing you can add the following into the top of `CodyProExpirationNotifications.triggerExpirationCheck`:

```ts
await localStorage.delete(CodyProExpirationNotifications.localStorageSuppressionKey)
```


DanTup:
@kalanchan FYI - there are some questions above, not sure who is best placed to answer

@toolmantim FYI - to review how the notifications look

mkondratek:
fyi @DanTup, my PR has been merged 🚀

kalanchan:
thanks @DanTup ! I updated the copy text, is there a way to add some line breaks between the sentences? 
EX - 

```
Your Cody Pro trial has ended.

Thank you for using Cody Pro, you have been downgraded to Cody Free. 
                
Setup your payment information to continue using Cody Pro for $9 / month, and cancel anytime.
```
Also, I wasn't able to see the notification running it locally even after setting `useSscForCodySubscription = true` and `codyProTrialEnded = true`. 


@abeatrix could you help me review as well?

DanTup:
@kalanchan unfortunately it's not possible to add newlines - see https://github.com/microsoft/vscode/issues/48900

> I wasn't able to see the notification running it locally even after setting useSscForCodySubscription = true and codyProTrialEnded = true.

Have you seen it at all? If you've dismissed it once it wouldn't show again (though it's not clear if you did). If you did, you can reset that with the code at the end of my comment above (in `triggerExpirationCheck`). Otherwise, you could add a breakpoint to the first line in `triggerExpirationCheck()` and see which `return` it's hitting (or add some `console.log`s for each `return`).

vscode/src/notifications/cody-pro-expiration.ts
```diff
@@ -0,0 +1,190 @@
...
+
+export class CodyProExpirationNotifications implements vscode.Disposable {
+    public static readonly expiredActionUrl = 'https://accounts.sourcegraph.com/cody/subscription'
+
+    public static readonly expiredMessageText = `
+                Your Cody Pro trial has ended.
+
+                Thank you for using Cody Pro, you have been downgraded to Cody Free. 
+                
+                Setup your payment information to continue using Cody Pro for $9 / month, and cancel anytime.
```
toolmantim, line 19:
```suggestion
                Your Cody Pro trial has ended, and you are now on the Cody Free plan.
                
                If you'd like to upgrade to Cody Pro, please setup your payment information. You can cancel anytime.
```

vscode/src/notifications/cody-pro-expiration.ts
```diff
@@ -0,0 +1,190 @@
...
+    public static readonly nearlyExpiredMessageText = `
+                Your Cody Pro Trial is ending soon. 
+
+                Setup your payment information to continue using Cody Pro, you won't be charged until February 15.
+            `
+    public static readonly localStorageSuppressionKey = 'extension.codyPro.suppressExpirationNotices'
+
+    public static readonly actionText = 'Setup Payment Info'
+
+    public static readonly noThanksText = `Don't show again`
```
toolmantim, line 33:
```suggestion
    public static readonly noThanksText = 'Don’t Show Again'
```

toolmantim (approved):
👍🏼 from a UX/design POV, with a few small copy edits

abeatrix (approved):
the code looks good to me 👍

DanTup:
@kalanchan 

> Also, I wasn't able to see the notification running it locally even after setting useSscForCodySubscription = true and codyProTrialEnded = true.

Did you figure out what was wrong here?

DanTup:
I pushed a small workaround for an issue with feature flags not being immediately updated (and started a thread on Slack). It basically involves waiting a few seconds after auth changes before triggering the check.

abeatrix (approved):
Setting the flags manually as mentioned in the Test Plan worked for me:

image

kalanchan:
also worked for me ! @DanTup I had notifications silent the whole time thats why it didn't work hah


CleanShot 2024-02-05 at 14 19 49@2x

DanTup:
@kalanchan any opinion on the 30 minute polling? If that's fine, this can be merged.

kalanchan:
> @kalanchan any opinion on the 30 minute polling? If that's fine, this can be merged.

yup its fine by me! lets do it

Issue #2911, opened 2024-01-26
Add e2e test for .cody/ignore behavior in chat


abeatrix:
Add e2e test to cover expected behavior for .cody/ignore in chat and commands mentioned in the main issue:

### Chat Message

Ask cody about fixup in Chat View: "what is fixup?" 

You should see Cody did not read any files from the `vscode` directory to answer your question:

<img width="2001" alt="image" src="https://github.com/sourcegraph/cody/assets/68532117/b53493e7-f382-4325-8c47-5b2229340aac">

### @ - mention

Ignored file should not show up in the @-mention list.

![Image](https://github.com/sourcegraph/cody/assets/68532117/f652f72a-d084-42b6-9b4a-a3b6f39609aa)

### Command

Executing a command on an ignored file will trigger a system pop up to notify you about the file is being ignored by Cody.

![Image](https://github.com/sourcegraph/cody/assets/68532117/47d1c6b9-b2d8-4481-9f2b-febf76cd5409)

PR #2913, opened 2024-01-26
add e2e test for .cody/ignore in chat,  update filtering step

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2911 https://github.com/sourcegraph/cody/issues/2914

This PR updates to move the cody/ignore filtering step for chat from context fetching step to SimpleChatModel during prompt building. 

The main change is that the filtering steps, which were previously performed in multiple functions for chat, have been moved to the end during the prompt building step, before the chat context is turned into chat messages and sent to the LLM (Language Model). This change removes the burden on each underlying context fetching service.

Also added a new e2e test for keeping track on .cody/ignore behavior in chat and commands.

Added security reminder for the completion client class about caller's responsibility on making sure the context are filtered before they are sent to the server.

The changes include:
- Updated to check for `isCodyIgnoredFile` in SimpleChatModel
- Updated `lastMessage.newContextUsed` assignment in SimpleChatModel to filter out ignored files
- Removed `isCodyIgnoredFile` import and usage in context.ts
  - Removed filtering of ignored files in `searchSymf` function in context.ts
  - Removed filtering of ignored files in `searchEmbeddingsLocal` function in context.ts
  - Removed checking of ignored files in `getCurrentSelectionContext` function in context.ts
  - Removed checking of ignored files and empty content in `getVisibleEditorContext` function in context.ts
  - Removed checking of ignored files in `getReadmeContext` function in context.ts

## Test plan

See the new e2e test for cody ignore. 

To-do: add integration test to agent


abeatrix:
Need to fix the e2e test (was having issue on windows but now broke the rest of them lol will fix on Monday), but the rest should be good to review.

DanTup:
Taking a look at the Windows issue - seems to be something bad with the paths:

image

DanTup:
The reason for the failure for me on Windows is this:

https://github.com/sourcegraph/cody/blob/9cb9d75a6c8300ea3fcc1a4046ff8a7b8b89b46f/lib/shared/src/chat/ignore-helper.ts#L120

The inputs are:

- `workspaceRoot.path`: `/c:/Dev/Sourcegraph/cody/vscode/test/fixtures/workspace`
- `uri.path`: `/C:/Dev/Sourcegraph/cody/vscode/test/fixtures/workspace/ignoredByCody.css`

The output is:

`../../../../../../../../C:/Dev/Sourcegraph/cody/vscode/test/fixtures/workspace/ignoredByCody.css`

The issue here is case-sensitivity.. The drive letter is cased differently (this happens a lot on Windows) but the function to compute relative paths (`posixAndURIPathsRelative`) is entirely case-sensitive.

VS Code does normalise drive letters in `.fsPath` to lowercase, so we may be able to do `URI.file(uri.fsPath)` to fix this for drive letters, although I don't know if there might be a possibility of getting here with mismatched casing in another part of the path and throwing in the same way.

@sqs what are your thoughts on making some of the functions like `posixAndURIPaths.relative` case-insensitive? While not strictly true, I don't know if we can easily tell the actual case-sensitivity (without testing by hitting the disk) and it seems like the chance of false paths when we already expect one path is inside the other is low.

DanTup:
@abeatrix I pushed a small workaround for the Windows drive letter issue (though it only fixes this very specific issue and there might still be other path-casing issues).

However, the e2e tests still fail to progress for me and the cause seems to be caused by the override of the settings:

```ts
    .extend<ExtraWorkspaceSettings>({
        extraWorkspaceSettings: {
            'cody.internal.unstable': true,
        },
    })
```

With this, the test hangs (fails to click the Cody button on sidebar). Without that line, it works. The issue seems to be that the default test (in helpers.ts) is declaring this:

```ts
    // By default, do not add any extra workspace settings.
    .extend<ExtraWorkspaceSettings>({
        extraWorkspaceSettings: {
            'cody.experimental.symfContext': false,
        },
    })
```

And the new test is essentially overwriting this (removing that setting). Explicitly adding the "cody.experimental.symfContext" setting to your new test as well causes it to get further, but it's not clear to me why the lack of that setting prevents the tests from clicking the sidebar (as far as I can see, the icon is there and visible, but the test claims it is not).

lib/shared/src/cody-ignore/ignore-helper.ts
```diff
@@ -117,6 +117,13 @@ export class IgnoreHelper {
             return this.getDefaultIgnores().ignores(uriBasename(uri))
         }
 
+        // HACK(dantup): It's possible we got here with two URIs that have a Windows drive letter cased
+        // differently. This will cause relative() to produce an incorrect path. As a workaround,
+        // re-create the URIs from fsPath which VS Code normalizes.
+        if (isWindows()) {
+            workspaceRoot = URI.file(workspaceRoot.fsPath)
+            uri = URI.file(uri.fsPath)
+        }
```
DanTup, line 126: @abeatrix @sqs this is what I added to fix the error I was seeing, but it's possible we have this kind of issue elsewhere where we do path manipulating (in particular computing relative paths) if casing doesn't match (Windows is entirely case-insensitive by default, so you can easily open VS Code using path segments case differently to what's on disk, for example).

vscode/test/e2e/cody-ignore.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    const chatPanel = page.frameLocator('iframe.webview').last().frameLocator('iframe')
+    const chatInput = chatPanel.getByRole('textbox', { name: 'Chat message' })
+    await chatInput.focus()
+    await chatInput.fill('Ignore me')
+    await chatInput.press('Enter')
+    // Assistant should response to your chat question,
+    // but the current file is excluded (ignoredByCody.css) and not on the context list
+    if (!isWindows()) {
+        await expect(chatPanel.getByText('hello from the assistant')).toBeVisible()
+    }
```
DanTup, line 46: What's the reason for this difference for Windows?

.cody/ignore
```diff
@@ -0,0 +1,3 @@
+# examples for testing
+playwright/**
+recordings/**
```
DanTup, line 3: Does this mean for manual testing? The automated tests seem to only use `ignoredByCody.css` so adding "manual" to the comment above might avoid confusion if that's the intent.

DanTup (approved):
Looks reasonable to me, but some questions about some things in the tests (and red bots).

vscode/test/e2e/cody-ignore.test.ts
abeatrix: it shouldn't! I added this because it was where the test would fail, so i wanted to see if adding this would help lol I've removed this :D

.cody/ignore
abeatrix: good point, added a comment to clarify this :)

abeatrix:
> @abeatrix I pushed a small workaround for the Windows drive letter issue (though it only fixes this very specific issue and there might still be other path-casing issues).
> 
> However, the e2e tests still fail to progress for me and the cause seems to be caused by the override of the settings:
> 
> ```ts
>     .extend<ExtraWorkspaceSettings>({
>         extraWorkspaceSettings: {
>             'cody.internal.unstable': true,
>         },
>     })
> ```
> 
> With this, the test hangs (fails to click the Cody button on sidebar). Without that line, it works. The issue seems to be that the default test (in helpers.ts) is declaring this:
> 
> ```ts
>     // By default, do not add any extra workspace settings.
>     .extend<ExtraWorkspaceSettings>({
>         extraWorkspaceSettings: {
>             'cody.experimental.symfContext': false,
>         },
>     })
> ```
> 
> And the new test is essentially overwriting this (removing that setting). Explicitly adding the "cody.experimental.symfContext" setting to your new test as well causes it to get further, but it's not clear to me why the lack of that setting prevents the tests from clicking the sidebar (as far as I can see, the icon is there and visible, but the test claims it is not).

@DanTup Thanks for looking into this! I moved the "unstable" setting back to helper, and the tests on other OSs are now passing again in the CI (that was my bad for moving it in my last commit), and the test did move on further for Windows as well. 

I am a bit confused because the file seems to be ignored by the first test where it wasn't being used as context (I assumed that's the case here since it passed the Context check), but it seems like it's showing up (or something else is showing up) when using via @-mention. I saw a test artifact was uploaded in the CI step after the test has failed. Do you happen to know where I can find the recording?

DanTup:
>  it seems like it's showing up (or something else is showing up) when using via @-mention. I saw a test artifact was uploaded in the CI step after the test has failed. Do you happen to know where I can find the recording?

On the summary page (https://github.com/sourcegraph/cody/actions/runs/7681119589), the artifacts are at the bottom:

Screenshot 2024-01-27 at 21 30 44

If you don't have easy access to debug on Windows, let me know and I can have another look locally.

abeatrix:
> On the summary page (https://github.com/sourcegraph/cody/actions/runs/7681119589), the artifacts are at the bottom:

🤡 should have looked into the summary page first, thanks! That just confirmed it's not working for @-mention in Windows 😭

image

DanTup:
@abeatrix I've pushed a fix that should fix for Windows. There were two issues:

1. Another instance of the relative path being computed wrong (same as my original issue above, but it didn't throw, it just silently recording the ignore rules incorrectly).. The fix was to apply the workaround in both places (I don't like this and think we might need to think about how we check paths that could be case-insensitive)
2. The test was looking for `.code/ignore` in the file-completion, but on Windows it has a backslash. The fix was to to change the expected text to use the correct slashes for the platform.

I'll check back later to ensure the bots are green now, but please review the last commit I pushed (since otherwise it's unreviewed). Thanks!

vscode/src/chat/chat-view/SimpleChatModel.ts
```diff
@@ -53,7 +59,7 @@ export class SimpleChatModel {
         if (lastMessage.message.speaker !== 'human') {
             throw new Error('Cannot set new context used for bot message')
         }
-        lastMessage.newContextUsed = newContextUsed
+        lastMessage.newContextUsed = newContextUsed.filter(c => !isCodyIgnoredFile(c.uri))
```
arafatkatze, line 62: This is 🔥 

Such a simple change saves us so much context tracking misery and potentials for failure in doing that. I was so concerned that me might need to do some sort whack-a-mole style search for context "leaks" in chat but this solves all of it. 

Thanks a million for this.

arafatkatze (approved):
So glad to see this PR

🙇

vscode/src/chat/chat-view/SimpleChatModel.ts
abeatrix: Thanks for adding the docs!

abeatrix:
> The fix was to apply the workaround in both places (I don't like this and think we might need to think about how we check paths that could be case-insensitive)

@DanTup thanks for the quick workaround! I agree we should rethink how we computer the relative path. I've created a follow-up issue based on your suggestion: https://github.com/sourcegraph/cody/issues/2929

Do you think this is an issue specified to cody/ignore? or does it affect other features as well? Trying to determine if this should be a follow-up issue for general or cody/ignore issue

PR #2927, opened 2024-01-29
Autocomplete: Add experimental fast path mode for Fireworks on PLG

philipp-spiess:
A new completion mode that skips the Sourcegraph instance and directly connects to Cody Gateway. It's enabled only for PLG and Fireworks (which make up the majority of our completion request).

This turns out to be a complete implementation of [RFC 888](https://docs.google.com/document/d/1xCSHHuz5M4PX6MQLun_KOz8RfoZyf-wWXRuTCBi07PE/edit) since it includes a fix for the two main functunalities that needs to be added:

- The client now handles chunks that do not have a repetition of all previously received tokens. Another small improvement in latency as we reduce the network overhead
- I made sure that the rate limiting behavior is kept as it was before. The only header that was missing was the one to infer wether a user can upgrade. We can also get this from the current `authStatus`

It also supports streaming exclusively. That is okay, because: We previously supported streaming only in Node.js envs with a server version that actually responds with a node stream. We can gate for the Node version (but it's unclear if the web version is even used at all) and we know the server is recent since we only run this on PLG.

You can turn it on via feature flag or VS Code config: `"cody.autocomplete.experimental.fastPath": true`

## Test plan

- Turn on `"cody.autocomplete.experimental.fastPath": true`
- Trigger single and multi-line completions and notice that cody gateway is the upstream that is being logged

To test the rate limiting, go into the dotcom admin console and set your quota to 1. Then:

### As a free use
Screenshot 2024-01-30 at 17 58 57

### As a cody pro user

Screenshot 2024-01-30 at 18 00 51


vscode/src/completions/client.ts
```diff
@@ -29,9 +31,25 @@ export function createClient(
...
     ): CompletionResponseGenerator {
-        const url = new URL('/.api/completions/code', config.serverEndpoint).href
+        if (featureFlags.fastPath) {
+            if (!config.accessToken) {
+                throw new Error('No access token found')
+            }
+            logDebug(
+                'fastpath',
+                '' + config.accessToken,
+                '' + dotcomTokenToGatewayToken(config.accessToken)
```
rafax, line 46: Should we do the conversion once, instead of doing it on every request?

vscode/src/completions/client.ts
```diff
@@ -58,7 +110,7 @@ export function createClient(
         // TODO(philipp-spiess): Feature test if the response is a Node or a browser stream and
         // implement SSE parsing for both.
         const isNode = typeof process !== 'undefined'
-        const enableStreaming = !!isNode
+        const enableStreaming = false //!!isNode
```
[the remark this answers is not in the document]
philipp-spiess: Yeah, I just disabled it now because I wanted to avoid having to mess with SSE payloads for the prototype. I will move all of the forking into the fireworks specific code actually (since Cody Gateway has provider specific APIs right now) at which point the fork is much easier.

vscode/package.json
```diff
@@ -1,1125 +1,1132 @@
 {
-  "name": "cody-ai",
```
valerybugakov, line 2: Formatting update?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -317,8 +320,8 @@ export class InlineCompletionItemProvider
                     requestManager: this.requestManager,
                     lastCandidate: this.lastCandidate,
                     debounceInterval: {
-                        singleLine: 75,
-                        multiLine: 125,
+                        singleLine: 25,
+                        multiLine: 50,
```
valerybugakov, line 324: What's the rationale here?

philipp-spiess: Oh that was just me trying locally, I do think that if we manage to shave of a few hundred MS we might need to re-evaluate our default timeouts again (we increased them from 25 to 75 at some point). Now is not the time though

philipp-spiess:
@rafax @valerybugakov Re-requesting a review. I still need to handle 429 headers as laid out by the RFC but other than that, this is mostly an implementation of RFC 888 already. It feels very good to use, you should try it out 🙂

lib/shared/src/inferenceClient/misc.ts
```diff
@@ -13,6 +13,11 @@ export type CodeCompletionsParams = Omit<CompletionParameters, 'fast'> & { timeo
 export type CompletionResponseGenerator = AsyncGenerator<CompletionResponse>
 
 export interface CodeCompletionsClient<T = CodeCompletionsParams> {
-    complete(params: T, abortController: AbortController): CompletionResponseGenerator
+    logger: CompletionLogger | undefined
```
philipp-spiess, line 16: @valerybugakov I don't like doing this but it seems very cumbersome to parse optional values to the completion providers so I thought maybe you can let me get away with this one 🥺

vscode/src/completions/client.ts
```diff
@@ -188,7 +189,10 @@ interface SSEMessage {
 }
 
 const SSE_TERMINATOR = '\n\n'
-export async function* createSSEIterator(iterator: NodeJS.ReadableStream): AsyncGenerator<SSEMessage> {
+export async function* createSSEIterator(
+    iterator: NodeJS.ReadableStream,
+    options: { batchCompletionEvents?: boolean } = {}
```
philipp-spiess, line 194: Made sure we can turn off the batching even though it wouldn't hit anyway (as Fireworks has a different SSE event type). But this seems like highly specific behavior and shouldn't be the "default"

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,153 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+    }
+
+    // When using the fast path, the Cody client talks directly to Cody Gateway. Since CG only
+    // proxies to the upstream API, we have to first convert the request to a Fireworks API
+    // compatible payload. We also have to manually convert SSE response chunks.
+    //
+    // Note: This client assumes that it is run inside a Node.js environment and will always use
+    // streaming to simplify the logic. Environments that do not support that should fall back to
+    // the default client.
+    private async *createFastPathClient(
```
philipp-spiess, line 316: It's now late and the PR is ready but I realize an alternative idea could also be to fork at the `CompletionClient` level (like you did with Ollama). That would prevent having to pass some data, too! 

Let me know if I make this change. I don't really like that this is going to be firework specific though, so it might be better suited in here. OTOH the Ollama client is also Ollama specific.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,153 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+        // c.f. https://github.com/microsoft/vscode/issues/173861
+        headers.set('Connection', 'keep-alive')
+        headers.set('Content-Type', 'application/json; charset=utf-8')
+        headers.set('Authorization', `Bearer ${this.codyGatewayAccessToken}`)
+        headers.set('X-Sourcegraph-Feature', 'code_completions')
+        addTraceparent(headers)
+        // Disable gzip compression since the sg instance will start to batch
+        // responses afterwards.
+        // Note: Find out if this is also happening when connecting directly to Cody Gateway
+        headers.set('Accept-Encoding', 'gzip;q=0')
```
philipp-spiess, line 350: a question for @rafax but I figured it doesn't hurt to include the header

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,153 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+                if (!choice) {
+                    continue
+                }
+
+                lastResponse = {
+                    completion: (lastResponse ? lastResponse.completion : '') + choice.text,
+                    stopReason:
+                        choice.finish_reason ??
+                        (lastResponse ? lastResponse.stopReason : STOP_REASON_STREAMING_CHUNK),
+                }
```
philipp-spiess, line 422: @valerybugakov sorry for the 🍝

vscode/src/configuration.test.ts
```diff
@@ -66,7 +66,7 @@ describe('getConfiguration', () => {
                     case 'cody.autocomplete.advanced.provider':
                         return 'unstable-openai'
                     case 'cody.autocomplete.advanced.model':
-                        return 'starcoder-32b'
+                        return 'starcoder-16b'
```
philipp-spiess, line 69: This was thorwing errors in the console 🤷

vscode/package.json
```diff
@@ -456,7 +456,7 @@
         "group": "Cody",
         "icon": "$(list-unordered)",
         "when": "cody.activated && cody.hasChatHistory"
-    },
+      },
```
valerybugakov, line 459: 🤔

vscode/src/completions/client.ts
valerybugakov: Add this to sources as a comment?

vscode/src/completions/providers/fireworks.ts
valerybugakov: I would not make this a required change, but I agree that forking on the HTTP client level ties nicely into our existing structure and is easier to maintain going forward because the data propagation and patterns are consistent across clients. We should try it if it's mostly about moving the code into a separate module.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -189,8 +213,19 @@ class FireworksProvider extends Provider {
...
+                // Require the upstream to be dotcom
+                this.authStatus.isDotCom &&
+                // The fast path client only supports Node.js style response streams
+                isNode
+
             const completionResponseGenerator = generatorWithTimeout(
-                this.client.complete(requestParams, abortController),
+                useFastPathClient
+                    ? this.createFastPathClient(requestParams, abortController)
+                    : this.createDefaultClient(requestParams, abortController),
```
valerybugakov, line 228: Nice!

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,153 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+    // compatible payload. We also have to manually convert SSE response chunks.
+    //
+    // Note: This client assumes that it is run inside a Node.js environment and will always use
+    // streaming to simplify the logic. Environments that do not support that should fall back to
+    // the default client.
+    private async *createFastPathClient(
+        requestParams: CodeCompletionsParams,
+        abortController: AbortController
+    ): CompletionResponseGenerator {
+        const url = 'https://cody-gateway.sourcegraph.com/v1/completions/fireworks'
```
valerybugakov, line 320: It's not that important, but can we support the locally running instance of the gateway here?

vscode/src/configuration.ts
```diff
@@ -118,11 +118,11 @@ export function getConfiguration(
             'autocomplete.experimental.dynamicMultilineCompletions',
             false
         ),
-
         autocompleteExperimentalHotStreak: getHiddenSetting(
             'autocomplete.experimental.hotStreak',
             false
         ),
+        autocompleteExperimentalFastPath: getHiddenSetting('autocomplete.experimental.fastPath', false),
```
valerybugakov, line 125: Should we add it to the `package.json` with a description?

valerybugakov (approved):
Works locally for me with dynamic multiline and hot-streak 🎉

vscode/package.json
philipp-spiess: 🤷 looks more correct now though

vscode/src/completions/providers/fireworks.ts
philipp-spiess: Hm there is no clear mapping of "SG instance to Cody Gateway" version but I can special case localhost I think 👍

vscode/src/configuration.ts
philipp-spiess: Oh yes sorry that was removed while I was fixing some package.json issues I believe

vscode/src/completions/providers/fireworks.ts
philipp-spiess: I just looked into the ollama abstraction and I think I misunderstood. This is actually not using the `client` that is initialized once and passed through to the providers but instead initializes the ollama client on demand. That pattern is similar to what I’m doing right now, too, so I think we can leave it as-is for now. 

Later we should revisit this and e.g. not initialize a sourcegraph instance client when ollama is specified, etc.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -91,18 +109,24 @@ const lineNumberDependentCompletionParams = getLineNumberDependentCompletionPara
...
-        { model, maxContextTokens, client, timeouts }: Required<FireworksOptions>
+        { model, maxContextTokens, client, timeouts, config, authStatus }: Required<FireworksOptions>
     ) {
         super(options)
         this.timeouts = timeouts
         this.model = model
         this.promptChars = tokensToChars(maxContextTokens - MAX_RESPONSE_TOKENS)
         this.client = client
+        this.codyGatewayAccessToken = config.accessToken
+            ? dotcomTokenToGatewayToken(config.accessToken)
```
rafax, line 127: Should we do this all the time, or only when fastPath is enabled?

vscode/src/completions/providers/fireworks.ts
rafax: It shouldn't - this old behaviour was caused by nytimes/gziphandler, which is not used in Cody Gateway

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,161 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+            body: JSON.stringify(fireworksRequest),
+            headers,
+            signal: abortController.signal,
+        })
+
+        const traceId = getActiveTraceAndSpanId()?.traceId
+
+        // When rate-limiting occurs, the response is an error message
+        // The response here is almost identical to the SG instance response but does not contain
+        // information on wether a user is eligible to upgrade to the pro plan. We get this from the
```
rafax, line 363:
```suggestion
        // information on whether a user is eligible to upgrade to the pro plan. We get this from the
```

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,161 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+        // authState instead.
+        if (response.status === 429) {
+            const upgradeIsAvailable = this.authStatus.userCanUpgrade
+            const retryAfter = response.headers.get('retry-after')
```
rafax, line 367: Up to you, but if this is logic is the same as for non-fast-path (and possibly chat) client, maybe we can extract a function for this?

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,6 +298,161 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+        }
+
+        let lastResponse: CompletionResponse | undefined
+        try {
+            const iterator = createSSEIterator(response.body)
+            let chunkIndex = 0
+
+            for await (const { event, data } of iterator) {
+                if (event === 'error') {
+                    throw new Error(data)
```
rafax, line 398: Should this be a tracedError?

vscode/src/completions/client.ts
```diff
@@ -204,7 +201,11 @@ interface SSEMessage {
 }
 
 const SSE_TERMINATOR = '\n\n'
-export async function* createSSEIterator(iterator: NodeJS.ReadableStream): AsyncGenerator<SSEMessage> {
+export async function* createSSEIterator(
+    iterator: NodeJS.ReadableStream,
+    // Message batching or any non-standard SSE behaviors should noe be enabled by default
```
rafax, line 206:
```suggestion
    // Message batching or any non-standard SSE behaviors should not be enabled by default
```

vscode/src/completions/providers/fireworks.ts
```diff
@@ -91,18 +108,32 @@ const lineNumberDependentCompletionParams = getLineNumberDependentCompletionPara
...
     ) {
         super(options)
         this.timeouts = timeouts
         this.model = model
         this.promptChars = tokensToChars(maxContextTokens - MAX_RESPONSE_TOKENS)
         this.client = client
+        this.authStatus = authStatus
+
+        const isNode = typeof process !== 'undefined'
+        this.codyGatewayAccessToken =
```
rafax, line 128: nit: fastPathToken (everything else using fast-path as prefix, and Cody Gateway is a concept that readers shouldn't have to understand)

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,13 +296,164 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+    //
+    // Note: This client assumes that it is run inside a Node.js environment and will always use
+    // streaming to simplify the logic. Environments that do not support that should fall back to
+    // the default client.
+    private async *createFastPathClient(
+        requestParams: CodeCompletionsParams,
+        abortController: AbortController
+    ): CompletionResponseGenerator {
+        const isLocalInstance =
+            this.authStatus.endpoint?.includes('sourcegraph.test') ||
```
rafax, line 319: FYI - Cody Gateway is not always ran in a local env (ex. for Enterprise), so you may want to make the errors highlight that it has to be ran

vscode/src/completions/providers/fireworks.ts
```diff
@@ -263,13 +296,164 @@ ${intro}${infillPrefix}${OPENING_CODE_TAG}${CLOSING_CODE_TAG}${infillSuffix}
...
+        // When rate-limiting occurs, the response is an error message The response here is almost
+        // identical to the SG instance response but does not contain information on whether a user
+        // is eligible to upgrade to the pro plan. We get this from the authState instead.
+        if (response.status === 429) {
+            const upgradeIsAvailable = this.authStatus.userCanUpgrade
+            throw await createRateLimitErrorFromResponse(response, upgradeIsAvailable)
+        }
+
+        if (!response.ok) {
+            throw new NetworkError(response, await response.text(), traceId)
```
rafax, line 371: if `isLocalInstance` is true, make it clear that Cody Gateway needs to be running?

rafax (approved):
LGTM as an experimental feature

PR #2930, opened 2024-01-29
Allow agent to be debugged by IntelliJ

steveyegge:
If you set some new env vars you can spawn the agent with --inspect, which allows IntelliJ to connect to it. There is also an option to have the agent listen on a socket for the client to attach.

## Test plan

Manually tested; it is a dev-only code path.


pkukielka (approved):
LGTM!

PR #2934, opened 2024-01-29
fix: update chat input placeholder

abeatrix:
Cleaned up imports for commands. Removed mention of `/ for command` from the chat input placeholder.

### Before

image

### After

image


kalanchan (approved):
good bye inline commands!

PR #2937, opened 2024-01-29
add: display cody icon in chat panel title

abeatrix:
This commit updates the package.json for the VS Code extension by removing the condition that required the resourceScheme to be file for the cody.menu.commands to be visible in the editor title. This change allows the command to be visible regardless of the resource scheme, as long as cody.activated is true, config.cody.editorTitleCommandIcon is set, and the editor is not in readonly mode. This enhances the command's visibility and usability in various contexts within VS Code, and shows up in the chat panels.

### Before

cody icon doesn't show up in chat panel

image

### After

image

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -6,6 +6,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


abeatrix:
> Thanks!
> 
> In QA'ing this I did spot another bug where our chat icons seems to flash for a second in file buffers too:
> 
>  CleanShot.2024-01-30.at.11.35.46.mp4

I looked into this, but I don't think this is something we have control over since vs code handles the when clauses 🤔

PR #2939, opened 2024-01-30
add /ask to command menu

abeatrix:
Add the missing /ask command back to command menu.

## Test plan

image

PR #2941, opened 2024-01-30
fix: handle ask and edit commands from menu

abeatrix:
Follow up on https://github.com/sourcegraph/cody/pull/2939

This enables /ask and /edit from menu to work again.

Also added e2e test to cover submitting question via /ask in command menu.

## Test plan

Start an ask/ edit command from the command menu to confirm it works. 

https://github.com/sourcegraph/cody/assets/68532117/c3ca3077-fdf5-46ea-a5b6-0cded45c0c2a

"Create Custom Command" option should only show up in custom command menu

image


kalanchan (approved):

toolmantim (approved):
👍🏼 Works well

Issue #2940, opened 2024-01-30
bug: Cody won't initialize after installing Tabnine


kalanchan:
### Version

1.2.0

### Describe the bug

Cody is stuck on initalizing after installing tabnine, it might be an issue with how both editors are using secretstorage for access tokens?

Cody still wont initialize even after uninstalling tabnine, and re-installing Cody

logs: 
```
█ VSCodeSecretStorage:getAccessToken: failed  {}
█ logEvent: CodyVSCodeExtension:CodySavedLogin:executed VSCode {"opts":{"hasV2Event":true}}
█ telemetry-v2: recordEvent: cody.extension/savedLogin: {"parameters":{"version":0,"metadata":[{"key":"contextSelection","value":10},{"key":"guardrails","value":0}]},"timestamp":"2024-01-29T23:20:12.495Z"}
█ AuthProvider:init:lastEndpoint: https://sourcegraph.com/
█ LocalEmbeddingsController: constructor
█ GraphQLTelemetryExporter: evaluated export mode: 5.2.5+
█ ChatManager:constructor: init has local embeddings controller
█ ChatPanelsManager:constructor: init
█ VSCodeSecretStorage:getAccessToken: failed  {}
█ CommandsController:disposeWatchers: watchers disposed
█ CommandsController:fileWatcherInit: watchers created
█ CodyCompletionProvider:notSignedIn: You are not signed in.
█ CodyEngine: using downloaded path "/Users/kalanchan/Library/Application Support/Code/User/globalStorage/sourcegraph.cody-ai/cody-engine/cody-engine-5.2.12792-macos-arm64"
█ LocalEmbeddingsController: spawnAndBindService service started, initializing
█ CodyEngine: stderr Starting main loop. VERBOSE_DEBUG=true
got request: "embeddings/initialize"
github.com/sourcegraph/deploy-sourcegraph-cloud: "/Users/kalanchan/Library/Caches/com.sourcegraph.cody/embeddings/.tmpDEyMDn"
github.com/sourcegraph/accounts.sourcegraph.com: "/Users/kalanchan/Library/Caches/com.sourcegraph.cody/embeddings/.tmp3AquIT"
github.com/sourcegraph/pfp: "/Users/kalanchan/Library/Caches/com.sourcegraph.cody/embeddings/.tmpyOq6tT"

█ LocalEmbeddingsController: spawnAndBindService initialized [object Object] token available? false
█ VSCodeSecretStorage:getAccessToken: failed  {}
[...]

PR #2943, opened 2024-01-30
Adding temp fix for the race condition involved in reading auth token

arafatkatze:
Fixes the race condition in the auth token issue highlighted in https://github.com/sourcegraph/cody/issues/2940 thanks to Kevin and Kalan for raising this.

The original bug came in this cool cleanup PR https://github.com/sourcegraph/cody/pull/2782 on this [commit](https://github.com/sourcegraph/cody/commit/058b8ad9542ce15fc3f95bc87f3c0d440699b708) CC @olafurpg 

## What is the bug?
In the latest version of Cody if you install it in VS code then it will ask you to reload and then you can't login(Coz it can't read auth token)

NOTE: 
1. This bug very likely has nothing to Tabnine, it was just a coincidence
2. The actual error of token not found is [here](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/SecretStorageProvider.ts?L15)

## Test plan

- Run the debugger
- Logout of Cody
- Log in to cody
- Everything should work

To verify the failure case do the same testing on main branch and you will see problems highlighted in the original issue.

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -599,8 +599,6 @@ const register = async (
 
     promises.push(showSetupNotification(initialConfig))
 
-    await Promise.all(promises)
-
     // Register a serializer for reviving the chat panel on reload
     if (vscode.window.registerWebviewPanelSerializer) {
         vscode.window.registerWebviewPanelSerializer(CodyChatPanelViewType, {


abeatrix (approved):
Tried logging out and logging back in and it seems to work as expected:

https://github.com/sourcegraph/cody/assets/68532117/47235730-0bb8-46b7-ab33-cf78a4645afc

@arafatkatze worth adding a changelog entry?

vscode/src/main.ts
```diff
@@ -599,8 +599,6 @@ const register = async (
 
     promises.push(showSetupNotification(initialConfig))
 
-    await Promise.all(promises)
```
abeatrix, line 602: @philipp-spiess @valerybugakov fyi

valerybugakov: Hey, @arafatkatze, can the `promises` variable be removed after this change?

arafatkatze: @valerybugakov Good catch, that might be the case but I will let this be for now because @olafurpg is working on a broader fix which is a mix of a few related issues. I just did a tiny hack for now.

PR #2946, opened 2024-01-30
Autocomplete: add more stop-reason values to autocomplete responses

valerybugakov:
## Context

- Adds `STOP_REASON_REQUEST_ABORTED` and `STOP_REASON_REQUEST_FINISHED` to completion responses to enable better observability locally and improve our analytics events analysis capabilities.
- Adds early exit for aborted requests to ollama HTTP client.

## Test plan

CI

### DIFF
--- a/lib/shared/src/inferenceClient/misc.ts
+++ b/lib/shared/src/inferenceClient/misc.ts
@@ -7,9 +7,11 @@ import type { CompletionParameters, CompletionResponse } from '../sourcegraph-ap
  * TODO: migrate to union of multiple `CompletionResponse` types to explicitly document
  * all possible response types.
  */
-export const STOP_REASON_STREAMING_CHUNK = 'cody-streaming-chunk'
-export const STOP_REASON_REQUEST_ABORTED = 'cody-request-aborted'
-export const STOP_REASON_REQUEST_FINISHED = 'cody-request-finished'
+export enum CompletionStopReason {
+    StreamingChunk = 'cody-streaming-chunk',
+    RequestAborted = 'cody-request-aborted',
+    RequestFinished = 'cody-request-finished',
+}
 
 export type CodeCompletionsParams = Omit<CompletionParameters, 'fast'> & { timeoutMs: number }
 export type CompletionResponseGenerator = AsyncGenerator<CompletionResponse>
--- a/lib/shared/src/inferenceClient/misc.ts
+++ b/lib/shared/src/inferenceClient/misc.ts
@@ -8,6 +8,11 @@ import type { CompletionParameters, CompletionResponse } from '../sourcegraph-ap
  * all possible response types.
  */
 export enum CompletionStopReason {
+    /**
+     * Used to signal to the completion processing code that we're still streaming.
+     * Can be removed if we make `CompletionResponse.stopReason` optional. Then
+     * `{ stopReason: undefined }` can be used instead.
+     */
     StreamingChunk = 'cody-streaming-chunk',
     RequestAborted = 'cody-request-aborted',
     RequestFinished = 'cody-request-finished',
--- a/lib/shared/src/inferenceClient/misc.ts
+++ b/lib/shared/src/inferenceClient/misc.ts
@@ -8,6 +8,8 @@ import type { CompletionParameters, CompletionResponse } from '../sourcegraph-ap
  * all possible response types.
  */
 export const STOP_REASON_STREAMING_CHUNK = 'cody-streaming-chunk'
+export const STOP_REASON_REQUEST_ABORTED = 'cody-request-aborted'
+export const STOP_REASON_REQUEST_FINISHED = 'cody-request-finished'
 
 export type CodeCompletionsParams = Omit<CompletionParameters, 'fast'> & { timeoutMs: number }
 export type CompletionResponseGenerator = AsyncGenerator<CompletionResponse>
--- a/lib/shared/src/ollama/ollama-client.ts
+++ b/lib/shared/src/ollama/ollama-client.ts
@@ -1,11 +1,9 @@
 import { isDefined } from '../common'
 import type { OllamaGenerateParameters, OllamaOptions } from '../configuration'
 import {
-    STOP_REASON_STREAMING_CHUNK,
+    CompletionStopReason,
     type CodeCompletionsClient,
     type CompletionResponseGenerator,
-    STOP_REASON_REQUEST_ABORTED,
-    STOP_REASON_REQUEST_FINISHED,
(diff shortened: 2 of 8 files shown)


lib/shared/src/inferenceClient/misc.ts
```diff
@@ -8,6 +8,8 @@ import type { CompletionParameters, CompletionResponse } from '../sourcegraph-ap
  * all possible response types.
  */
 export const STOP_REASON_STREAMING_CHUNK = 'cody-streaming-chunk'
+export const STOP_REASON_REQUEST_ABORTED = 'cody-request-aborted'
+export const STOP_REASON_REQUEST_FINISHED = 'cody-request-finished'
```
dominiccooney, line 12: Make this an enum with string values so you can use the enum type to refer to "a stop reason."

Also referring to a streaming chunk boundary as a *stop* is a bit surprising... is there a better name? What is "stopping" exactly at a chunk boundary?

lib/shared/src/ollama/ollama-client.ts
```diff
@@ -4,8 +4,11 @@ import {
     STOP_REASON_STREAMING_CHUNK,
     type CodeCompletionsClient,
     type CompletionResponseGenerator,
+    STOP_REASON_REQUEST_ABORTED,
```
dominiccooney, line 7: This sort order doesn't look right.

lib/shared/src/ollama/ollama-client.ts
```diff
@@ -83,15 +87,21 @@ export function createOllamaClient(
...
+                }
+
                 for (const chunkString of chunk.toString().split(RESPONSE_SEPARATOR).filter(Boolean)) {
                     const line = JSON.parse(chunkString) as OllamaGenerateResponse
 
                     if (line.response) {
-                        responseText += line.response
-                        yield { completion: responseText, stopReason: STOP_REASON_STREAMING_CHUNK }
+                        insertText += line.response
+                        yield { completion: insertText, stopReason: STOP_REASON_STREAMING_CHUNK }
```
dominiccooney, line 104: I realize you're just tidying here, but is a yield a *stop*?

vscode/src/completions/client.ts
```diff
@@ -119,58 +125,67 @@ export function createClient(
...
+            log?.onComplete(completionResponse)
+            return completionResponse
+        } catch (error) {
+            // Shared error handling for both streaming and non-streaming requests.
+            if (isRateLimitError(error as Error)) {
+                throw error
+            }
+
+            if (isAbortError(error as Error) && completionResponse) {
+                log?.onComplete(completionResponse)
```
dominiccooney, line 183: This seems odd... We log `onComplete`... but only on abort errors?

vscode/src/completions/client.ts
```diff
@@ -108,9 +113,10 @@ export function createClient(
         // regular JSON payload. This ensures that the request also works against older backends
         const isStreamingResponse = response.headers.get('content-type') === 'text/event-stream'
 
-        if (isStreamingResponse && isNodeResponse(response)) {
-            let lastResponse: CompletionResponse | undefined
-            try {
+        let completionResponse: CompletionResponse | undefined = undefined
```
dominiccooney, line 116: This would be neater if you had the CompletionResponse | undefined type only in the streaming branch. Only problem with that is the weird abort error logging in the catch clause. Can this be simplified?

dominiccooney (changes requested):
Some ideas inline.

lib/shared/src/inferenceClient/misc.ts
valerybugakov: > Make this an enum with string values so you can use the enum type to refer to "a stop reason."

There's a TODO above these variables, and I planned to do it in a follow-up PR with more values added to the `enum`. I migrated the existing one to the enum here https://github.com/sourcegraph/cody/pull/2946/commits/97902c4cfbeff46fbec4802f30b6e726414928d4

> Also referring to a streaming chunk boundary as a stop is a bit surprising... is there a better name?

We need to do a bigger refactor to address this. This comes from the fact the `CompletionResponse` has `stopReason` as a required property, so we use a specific value in our code to signal to the processing logic that this is a partial response. Ideally, we have `stopReason: undefined` for this case and remove the streaming chunk stop reason. Do you think it makes sense? I added a comment about that here: https://github.com/sourcegraph/cody/pull/2946/commits/59b616794dbf48d79f2711240e13d8f990298561

lib/shared/src/ollama/ollama-client.ts
valerybugakov: Let's continue the discussion in another thread: https://github.com/sourcegraph/cody/pull/2946#discussion_r1472235755

vscode/src/completions/client.ts
valerybugakov: This is the existing behavior that I did not refactor in this PR. AFAIU, we call `onComplete` here because we want to log the completion we generated before the request was aborted.

vscode/src/completions/client.ts
valerybugakov: I looked into it and moved `log.onComplete` to the `finally` clause because this is what we want. In case the request was aborted but we got a completion response, we still want to have it in the Cody output channel. Added a comment about that too: https://github.com/sourcegraph/cody/pull/2946/commits/dd02323424ec8f0a2d37d2c2fadc54f2e267267b

It removes duplicate `onComplete` calls but still needs the `completionResponse` variable.

vscode/src/completions/client.ts
```diff
@@ -119,58 +121,74 @@ export function createClient(
...
+            }
+
+            return completionResponse
+        } catch (error) {
+            // Shared error handling for both streaming and non-streaming requests.
+            if (isRateLimitError(error as Error)) {
+                throw error
+            }
+
+            /**
```
dominiccooney, line 176: I have always associated `/**` with documentation comments (as in, class and member documentation.)

In general, `/*` is an unfriendly comment syntax to use in Type/JavaScript because `/*`-style comments don't nest, but they're convenient when hacking to comment out swathes of code. Having `/*`-style comments in the code breaks that.

So let's write less `/*` in general, and `/**` for member-level documentation if we feel we must. But not here.

dominiccooney (approved):
Splendid.

One nit inline.

vscode/src/completions/client.ts
valerybugakov: I like `/**` more for its readability and automatic expansion handled by the editor when typing multiline comments, but I agree that it's associated with docs comments, so I changed it to `//` 👍

PR #2950, opened 2024-01-30
Edit: Remove usage of codebase context

umpox:
## Description

Remove codebase context as:
- We now support user provided context, which will generally be much more accurate for the user
- This context is very rarely useful and adds latency to every request. Related code is unlikely to be useful for edits. It's possible we could use codebase context for code related to the instructions, but this is unlikely to work well given most instructions are short. We should revisit a mechanism to manually trigger codebase context.

## Test plan

1. Run edits
2. Check they output as normally

### DIFF
--- a/vscode/src/edit/prompt/context.ts
+++ b/vscode/src/edit/prompt/context.ts
@@ -1,4 +1,4 @@
-import * as vscode from 'vscode'
+import type * as vscode from 'vscode'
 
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
Lgtm. Was wondering why "fix" doesn't use diagnostics context but the doc string covered that 😊

PR #2951, opened 2024-01-30
Edit: Add support for multiple models

umpox:
## Description

This PR:
- Enables multi-model selector UI for Edit
- Adds multi-model support for Edit:
  - Anthropic: Claude 2 + Instant
  - OpenAI: GPT-3.5 and 4
- Improves how we build edit prompts and context, to make it easier to add new models in the future.
- Refactors some chat-specific logic to be more generic, so we can maintain chat and edit models in a single place
- Improves typing, so we can still get type narrowing, e.g. `type EditModels = 'anthropic/..' | 'openai..' | etc`
- Various cleanup

## Screenshots

Pro users:

image

Non-Pro users:
image

## Demo 
https://github.com/sourcegraph/cody/assets/9516420/73ddf2de-9756-4707-87c7-1eb7b16697d3

## Test plan

- Create an edit
- Try changing the model and create an edit
- Try retrying an edit and changing the model
- As a non-pro user, try selecting a pro model


lib/shared/src/models/types.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Available models for Chat.
+ * This is either:
+ * - one of the availble options (dotcom)
+ * - an unknown `string` (enterprise)
+ */
+export type ChatModel =
+    | {
+          [K in keyof Models]: HasUsage<Models[K], ModelUsage.Chat>
+      }[keyof Models]['model']
+    | (string & {})
```
umpox, line 39: These give us types that contain the model strings that are available for `Chat` or `Edit`.

It's not fully type safe, as enterprise models can have an unknown value, but this helps a lot with matching prompts up to available models etc

vscode/src/edit/prompt/models/generic.ts
```diff
@@ -0,0 +1,145 @@
...
+
+            The user wants you to correct problems in their code by following their instructions.
+            Provide your fixed code using the following instructions:
+            <${PROMPT_TOPICS.DIAGNOSTICS}>
+            {instruction}
+            </${PROMPT_TOPICS.DIAGNOSTICS}>`,
+    },
+    test: {
+        system: dedent`
+            - Do not enclose response with any markdown formatting or triple backticks.
```
abeatrix, line 78: I think for the test command, we might want to stick with Claude 2.0 as it doesn't work well with Claude Instant and Chat GPT. I also runs into an issue where the file name was not provided in the response more often with this prompt change 🤔 

chat GPT:

chatgpt_3

claude instant:

claude_instant

abeatrix (changes requested):
Nice adds! Some of the models runs really fast for some edit jobs! I think we might want to set the default model for /test to Claude 2.0 until we have tested it more thoroughly as I'm seeing regression in response when using other models (see inline comment). 

~~Also noticed a potential bug: when I run the test command using Claude 2.0 model, it would start two fixup task for me~~ (no longer reproducible): 

error_1

I don't get the same issue when using other models though 👀  ~~We should also fix the e2e test for fixup test~~ (update: added one to main!) 😅

abeatrix:
Did some more testing and noticed GPT-3.5 returns code enclosed in markdown format consistently for me:

When trying to create a new test file:

https://github.com/sourcegraph/cody/assets/68532117/6cea5502-9189-46bc-86b0-c44ce53eb685

When trying to generate tests for existing file:

https://github.com/sourcegraph/cody/assets/68532117/16966e81-9865-4077-8d08-d32590154390

Claude Instant seems to work fine now, but Claude 2.1 sometimes would include import statements to existing test file when the prompt asked not to:

https://github.com/sourcegraph/cody/assets/68532117/4b5de146-04f2-4a69-b837-127ad6b2e2d4

Claude 2.0 has also returned test enclosed in markdown, but not consistently 🤔 

https://github.com/sourcegraph/cody/assets/68532117/41d7fd6b-3b99-40e6-94f7-f13d8c93f8ce

toolmantim:
So cool to be giving it a go.

Following on from @abeatrix’s tests about with `/test`, I don't think we want the high-level edit settings effecting the document & test commands. They have their own range selection, and are built against Claude.

The rest of this is some non-blocking feedback…

I think maybe we need that extra "submit" separator on the first item here, just to make it clear there's two groups here:

CleanShot 2024-02-02 at 13 39 05@2x

It would be great if "Nearest Code Block" was shown as `Nearest Code Block (${nameOfSymbol})` here instead, and just remove one level of thinking the user has to do (if there's cases where that's not possible, then just Nearest Code Block). Though as a user I expected it to just auto-select the correct symbol from the symbol list (e.g. `extract_dimensions`):

CleanShot 2024-02-02 at 13 39 21@2x

Do we operate on line ranges, or also sub-lines? Because for this label, I was going to suggest "Selected Lines" (or even "Selected Lines — L47" / "Selected Lines — L47-52" instead of "Selection"):

CleanShot 2024-02-02 at 13 45 11@2x

kalanchan:
+1 to the feedback from @abeatrix and @toolmantim 

One thing I got really confused was when I used the "edit" command, in the model & range dropdown it also gave me the options to use Document and Test. Is this expected behaviour? Feels a bit confusing! 


https://github.com/sourcegraph/cody/assets/51868853/568de761-58ac-46e6-a995-389620bf98db

CleanShot 2024-02-01 at 23 38 13@2x


Also noticed some regression in test quality, I wrote a simple `login.ts` function and it gave a better test case explanation before the changes. Before and after were ran on Claude 2.0

CleanShot 2024-02-01 at 23 46 18@2x

I think we should keep the model change to just the edit command for now, and maybe give product & design some time to think about where it makes sense to introduce model changes and what the flow should look like. ALSO WE NEED TO TEST FOR regression!!

umpox:
>  think we might want to set the default model for /test to Claude 2.0 until we have tested it more thoroughly

@abeatrix @kalanchan @toolmantim 
👍 Done

That's fine, I've updated the edit commands (`fix`, `doc`, `test`) to default to Claude 2. That being said, I do think in the future we should make these commands work as well as possible with other models. We have all the code in place now to make it easy to tweak the prompt depending on the model. @abeatrix I left a relevant comment [here](https://github.com/sourcegraph/cody/pull/2959#issuecomment-1919261010) mentioning how we could separate the file creation prompt to make things more reliable, this would likely solve the issue (we could still use Claude - just for the file creation step). 

> We hide the option to switch models for enterprise users in chat, maybe that should be the default for edit too?
@abeatrix 👍 Done

> I think maybe we need that extra "submit" separator on the first item here, just to make it clear there's two groups here:
@toolmantim Done 👍 

> One thing I got really confused was when I used the "edit" command, in the model & range dropdown it also gave me the options to use Document and Test. Is this expected behaviour? Feels a bit confusing!

@kalanchan That's expected behaviour, Document & Test are Edit commands, so it makes sense to include them in the Edit input. The input gives more advanced behavior than the menu, as users can switch the symbol/range they use for those commands. There's some duplication I guess, but I think it'd be more confusing to have the edit input show different items depending on how it is triggered 🤔

umpox:
> It would be great if "Nearest Code Block" was shown as Nearest Code Block (${nameOfSymbol}) here instead, and just remove one level of thinking the user has to do (if there's cases where that's not possible, then just Nearest Code Block). Though as a user I expected it to just auto-select the correct symbol from the symbol list (e.g. extract_dimensions):

@toolmantim I hear you, but "Nearest code block" doesn't match up with symbols nicely. It's actually the nearest folding range, which often won't be a symbol. The ranges don't match up exactly. We're going to look into if we can improve the folding range logic to actually always expand to the nearest function, which is almost always what we want. Once we do that I think we can update this to "Nearest function" and include the symbol.

> Do we operate on line ranges, or also sub-lines? Because for this label, I was going to suggest "Selected Lines" (or even "Selected Lines — L47" / "Selected Lines — L47-52" instead of "Selection"):

We also operate on sub-lines. so selected lines isn't technically accurate here. Selecting a range within a line is important, especially for things like `document` where there may be multiple variables on one line. 👍

abeatrix:
> That being said, I do think in the future we should make these commands work as well as possible with other models. We have all the code in place now to make it easy to tweak the prompt depending on the model.

I totally agree we should improve the prompt to make sure it works with all models. we just need more time for the prompt engineering's part 😂 

> I left a relevant comment here mentioning how we could separate the file creation prompt to make things more reliable, this would likely solve the issue (we could still use Claude - just for the file creation step).

The /test command in the main branch and pre release has been working reliably for me after updating the prompt, but like kalan, I also noticed a regression in the /test output quality from this branch even when using the same Claude model, e.g import statements are not included when adding tests to a new file, or importing files with incorrect file path, etc. I wonder if it's caused by the context message issue I mentioned above, but I haven't figured out where we wrap the context message again in the code -- any idea?

umpox:
@abeatrix Looking into the context issue now, the only thing I've changed really with the prompt is:
- We no longer include context with markdown backticks, which was one of the reason why the LLM often produced markdown

Anything else should be a bug so I'll check it all now

umpox:
@abeatrix Actually... the test command is actually using `claude 2.1` on `main` right now. This PR obviously changes the default `claude 2.0`. I wonder if part of the regression is because `claude 2.1` performs better? Is it easily reproducible for you?

abeatrix (approved):
i dont see any regression with the /test command anymore in the latest commit, thank you so much for fixing it so quickly! 

I still notice the double file path added to the context message, but i don't think that's a block and can be fixed in a follow-up PR. well done!

image

PR #2954, opened 2024-01-30
INC-267: add test case to reproduce regression

olafurpg:
The PR https://github.com/sourcegraph/cody/pull/2943 removed an `await` to resolve an incident where users were unable to activate the Cody extension if they were signed out. This PR adds test cases to reproduce the original bug to prevent similar regressions from getting adding back into the codebase later.

## Test plan

Green CI.

Manually replace `void showSetupNotification(initialConfig)` with `await showSetupNotification(initialConfig)` and run `pnpm test agent/src/index.test.ts`. Observe that the tests time out.


vscode/src/main.ts
```diff
@@ -597,7 +595,11 @@ const register = async (
         )
     }
 
-    promises.push(showSetupNotification(initialConfig))
+    // INC-267 do NOT await on this promise. This promise triggers
```
arafatkatze, line 598: 🧯

arafatkatze (approved):

arafatkatze:
Such a clean PR, thanks for the FIX and helping me understand everything.

🙇

PR #2959, opened 2024-01-30
feat: unit test command & unit test case command

abeatrix:
PR to introduce `/unit` command that is a smarter version of the `/test` command.

### /unit command

- `/test` command generates unit tests in chat
- `/unit` command generates unit tests in line
  - if a test file exists, the new unit test suit will be appended to the end of the current test file
  - if a test file does not exist, cody will suggest a file name and we will create a temporary test file with the new test suit appended along with the import statements (if any)

unit-test-file

### unit-test-case command

`/unit-test-case` is not a command available via the command menu, but a command that can be executed via the Cody Command CodeLenses (when enabled) in test files.

unit-test-case

## Summary

This PR includes the following change:
- Refactor the unit-test command and move it out of unstable
- Add unit-test-case command
- Add `executeUnitCaseCommand` to execute unit test cases
- Added support for specifying a `destination file` in the executeEditArguments interface in the edit/execute.ts file.
- Added support for specifying a `destination file` in the executeEdit method in the edit/manager.ts file.
- Removed the `NewFixupFileMap` import and added support for specifying a `destination file` in the EditProvider class in the edit/provider.ts file.
- Removed the `NewFixupFileMap` import in the FixupController class in the non-stop/FixupController.ts file.

## Test plan

Added e2e test.

### Unit

Run the `Unit` command to confirm the newly generated tests are added to a new / exisiting test file automatically for you.

image

### Unit Cases

Enable Code Lenses in your Cody Setting

image

Locate the `Add Test` code lenses in a test file to check if new test cases have been inserted in line for you

image


vscode/package.json
```diff
@@ -1,1125 +1,1090 @@
...
-    ],
-    "viewsContainers": {
-      "activitybar": [
-        {
-          "id": "cody",
-          "title": "Cody",
-          "icon": "resources/cody.svg"
-        }
-      ]
+    "name": "cody-ai",
```
abeatrix, line 2: got formatted on save... will revert the change tomorrow....

vscode/src/services/treeViewItems.ts
```diff
@@ -131,14 +129,13 @@ const commandsItems: CodySidebarTreeItem[] = [
         title: 'Test',
         icon: 'package',
         command: { command: 'cody.command.generate-tests' },
-        description: 'Generate unit tests',
+        description: 'Generate unit tests (Chat)',
```
abeatrix, line 132: @toolmantim Let me know if you think we should remove the old /test command as I wasn't sure if we want to keep the chat option available. can do it in a follow up too.

vscode/src/non-stop/FixupTask.ts
```diff
@@ -60,7 +60,9 @@ export class FixupTask {
         public mode: EditMode = 'edit',
         /* the source of the instruction, e.g. 'code-action', 'doc', etc */
         public source?: ChatEventSource,
-        public readonly contextMessages?: ContextMessage[]
+        public readonly contextMessages?: ContextMessage[],
+        /* The file to write the edit to. If not provided, the edit will be applied to the fixupFile. */
+        public destinationFile?: vscode.Uri
```
abeatrix, line 65: i added this to replace `NewFixupFileMap`, let me know if you don't think this is a good appraoch.

I also suck at naming, so feel free to suggest another name for this field 🤡

toolmantim (commented):
Woo!

* Why the rename to `/unit`? I don't see a strong argument for renaming — I'd recommend we keep a single `/test` command (the edit version) and replace the old `/test` (chat version) with this one
* I couldn't get this to work well with Python or Typescript… it seems to replace the code with the tests (see videos below)
* We might need to add some kind of progress that isn't just a spinner, because tests can take quite a while to generate. Maybe something like `\`Generating tests… (${n} lines)\``?
* Is the "unit-test-case" command — it's an interesting idea, but I don't fully understand what problem it's trying to solve. Could you expand more on how it fits into the overall flow?

https://github.com/sourcegraph/cody/assets/153/c42d81c8-4eaf-4467-9528-aba244c4762d

https://github.com/sourcegraph/cody/assets/153/25d7427e-0eb0-4104-b780-8b20daba0d34

abeatrix:
> Why the rename to /unit? I don't see a strong argument for renaming — I'd recommend we keep a single /test command (the edit version) and replace the old /test (chat version) with this one

Because jetBrains is using the old one so if I was worried switching it would cause issue for them, but I think I can find a workaround for that.

abeatrix:
> Is the "unit-test-case" command — it's an interesting idea, but I don't fully understand what problem it's trying to solve. Could you expand more on how it fits into the overall flow?

That's for adding more tests to an existing test suite inside the test file, vs create a new test suite with tests using other test context as reference, adding more tests to an existing test suite should be easier for the llm and the result should have a higher quality from what I've tested so far.

abeatrix:
> I couldn't get this to work well with Python or Typescript… it seems to replace the code with the tests (see videos below)

Let me try to reproduce it. Can you try it on non-react typescript code in the meantime?

toolmantim:
> That's for adding more tests to an existing test suite inside the test file, vs create a new test suite with tests using other test context as reference, adding more tests to an existing test suite should be easier for the llm and the result should have a higher quality from what I've tested so far.

That's interesting about higher quality — is that higher quality output for the same user problem? (i.e. same source and test file, but running the command from the test file first rather than the source file first).

I still don't quite understand what "Add Tests" quite means, or how it fits into their overall flow though. When would someone know to hit and not hit the "Add Tests" button?

abeatrix:
> i.e. same source and test file, but running the command from the test file first rather than the source file first

Yes. You will only see this code lense in test files. You will not see this in source file.

> I still don't quite understand what "Add Tests" quite means, or how it fits into their overall flow though. When would someone know to hit and not hit the "Add Tests" button?

it only shows up on top of each test suite, and clicking "Add Tests" adds new tests to the associated test suite. You don't see them in non-test files.

image

abeatrix:
For example, i started the suite with only 1 test:

image

Press Add tests:

image

Ended up with more tests:

image

abeatrix:
@toolmantim Can you share the link to the repo you tested with to see if I can reproduce the issue? I can't reproduce the issue when testing it on the langchain python repository.

https://github.com/sourcegraph/cody/assets/68532117/a2ae5e5f-f6d8-44be-9a98-c9e4dda8a4b3

vscode/src/commands/execute/test-case.ts
```diff
@@ -0,0 +1,55 @@
...
+        task: await executeEdit(
+            {
+                instruction,
+                document,
+                range,
+                intent: 'edit',
+                mode: 'insert',
+                userContextFiles: contextFiles,
+            } satisfies ExecuteEditArguments,
+            DefaultEditCommands.Doc
```
umpox, line 52:
```suggestion
            DefaultEditCommands.Test
```

vscode/src/commands/execute/test-case.ts
```diff
@@ -0,0 +1,55 @@
...
+        type: 'edit',
+        task: await executeEdit(
+            {
+                instruction,
+                document,
+                range,
+                intent: 'edit',
+                mode: 'insert',
+                userContextFiles: contextFiles,
+            } satisfies ExecuteEditArguments,
```
umpox, line 51: As we're adding tests, should we not use `add`, and set the range to the last line? This will mean the new tests are streamed in, which will be much faster.

Something like:
```suggestion
    const range = new Range(editor.selection.end.line, 0, editor.selection.end.line, 0)

    return {
        type: 'edit',
        task: await executeEdit(
            {
                instruction,
                document,
                range,
                intent: 'add',
                mode: 'insert',
                userContextFiles: contextFiles,
            } satisfies ExecuteEditArguments,
```

vscode/src/commands/execute/test-file.ts
```diff
@@ -0,0 +1,66 @@
...
+    args?: Partial<CodyCommandArgs>
+): Promise<EditCommandResult | undefined> {
+    // The prompt for generating tests in a new test file
+    const newTestFilePrompt = defaultCommands.test.prompt
+    // The prompt for adding new test suit to an existing test file
+    const newTestSuitPrompt =
+        'Review the shared code context to identify the testing framework and libraries in use. Then, create multiple new unit tests for the functions in <selected> following the same patterns, testing conventions and testing library as shown in shared context. Pay attention to the shared context to ensure your response code do not contain cases that have already been covered. Focus on generating new unit tests for uncovered cases. Response only with the full completed code with the new unit tests added at the end, without any comments, fragments or TODO. The new tests should validate expected functionality and cover edge cases for <selected>. Do not include any markdown formatting or triple backticks. The goal is to provide me with code that I can add to the end of existing test file. Enclose only the new test suit WITHOUT ANY import statements or packages in your response.'
```
umpox, line 25:
```suggestion
    // The prompt for adding new test suite to an existing test file
    const newTestSuitePrompt =
        'Review the shared code context to identify the testing framework and libraries in use. Then, create multiple new unit tests for the functions in <selected> following the same patterns, testing conventions and testing library as shown in shared context. Pay attention to the shared context to ensure your response code do not contain cases that have already been covered. Focus on generating new unit tests for uncovered cases. Response only with the full completed code with the new unit tests added at the end, without any comments, fragments or TODO. The new tests should validate expected functionality and cover edge cases for <selected>. Do not include any markdown formatting or triple backticks. The goal is to provide me with code that I can add to the end of existing test file. Enclose only the new test suite WITHOUT ANY import statements or packages in your response.'
```

Same here

vscode/src/commands/services/CommandCodeLenses.ts
```diff
@@ -147,11 +169,15 @@ export class EditorCodeLenses implements vscode.CodeLensProvider {
...
     cody: {
         title: '$(cody-logo) Cody',
         command: 'cody.editor.codelens.click',
         tooltip: 'Open command menu',
     },
-    inline: { title: 'Inline Chat', command: 'cody.editor.codelens.click', tooltip: 'Ask Cody inline' },
+    test: {
+        title: '$(cody-logo) Add Tests',
+        command: 'cody.editor.codelens.click',
+        tooltip: 'Generate new test cases for the selected test suit',
```
umpox, line 181:
```suggestion
        tooltip: 'Generate new test cases',
```

I'd personally keep this simple IMO, users can already see it's attached to the suite

vscode/src/commands/execute/test-case.ts
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * It generates new test cases for the selected test suit.
+ * When invoked, the command will be executed as an inline-edit command.
+ */
+export async function executeTestCaseCommand(
+    args?: Partial<CodyCommandArgs>
+): Promise<EditCommandResult | undefined> {
+    const instruction =
+        'Review the shared code context to identify the testing framework and libraries in use. Then, create multiple new unit tests for the functions in <selected> following the same patterns, testing conventions and testing library as shown in shared context. Pay attention to the shared context to ensure your response code do not contain cases that have already been covered. Focus on generating new unit tests for uncovered cases. Response only with the full completed code with the new unit tests added at the end, without any comments, fragments or TODO. The new tests should validate expected functionality and cover edge cases for <selected>. Do not include any markdown formatting or triple backticks. The goal is to provide me with code that I can add to the end of existing test file. Enclose only the new tests WITHOUT ANY suit, import statements or packages in your response.'
```
umpox, line 20:
```suggestion
        'Review the shared code context to identify the testing framework and libraries in use. Then, create multiple new unit tests for the functions in <selected> following the same patterns, testing conventions and testing library as shown in shared context. Pay attention to the shared context to ensure your response code do not contain cases that have already been covered. Focus on generating new unit tests for uncovered cases. Response only with the full completed code with the new unit tests added at the end, without any comments, fragments or TODO. The new tests should validate expected functionality and cover edge cases for <selected>. Do not include any markdown formatting or triple backticks. The goal is to provide me with code that I can add to the end of existing test file. Enclose only the new tests WITHOUT ANY suite, import statements or packages in your response.'
```

Minor, but could have an impact on the LLM output

vscode/src/commands/services/CommandCodeLenses.ts
```diff
@@ -87,41 +77,73 @@ export class EditorCodeLenses implements vscode.CodeLensProvider {
...
+                    symbol.kind === vscode.SymbolKind.Class ||
+                    symbol.kind === vscode.SymbolKind.Method ||
+                    symbol.kind === vscode.SymbolKind.Constructor
+            ) ?? []
 
-        // Add code lenses for each symbol
-        if (symbols) {
+        // For test files, adds code lenses for each symbol
+        if (isTest) {
             for (const symbol of symbols) {
```
umpox, line 102: I guess the downside here is that this will show for random functions too, e.g.
image

I guess we don't really have a way to determine which symbols are actually used for tests here :/

vscode/src/non-stop/FixupController.ts
```diff
@@ -842,7 +843,7 @@ export class FixupController
      * Update the task's fixup file and selection range with the new info,
      * and then task mode to "insert".
      *
-     * NOTE: Currently used for /test command only.
+     * NOTE: Currently used for /uint and /unit-case command only.
```
umpox, line 846:
```suggestion
     * NOTE: Currently used for /unit and /unit-case command only.
```

vscode/src/non-stop/FixupTask.ts
umpox: Sounds good!

umpox:
Noticed one thing, it seems to often not stream the text into the new file for me. It's a bit off putting because it feels like nothing is working, but actually the file seems to be just created at the end.

I got a good video of it doing this:

https://github.com/sourcegraph/cody/assets/9516420/107af2c8-75c9-45e7-b686-93dc2733d96c

It looks like the file name isn't properly returned by the LLM. WDYT about maybe firing two LLM requests for this command:
1. The edit one
2. A separate one to get the file name, if it doesn't exist. Then we maybe have a better chance of getting a valid response?

We had something similar for the intent detector: https://sourcegraph.com/github.com/sourcegraph/cody@bee/new-commands/-/blob/lib/shared/src/intent-detector/client.ts?L66%3A1-125%3A6

We could do the same here?

umpox (commented):
Left some minor comments. Looking really good!

abeatrix:
@toolmantim @umpox i've updated the prompt so that it will always generate the file name. Before this change, if cody doesn't reply with a file name suggestion, we would wait for the test to finish before adding the code, causing a delay in showing the response.
The updated prompt now reminds the LLM provider to start the response with a file name suggestion.

abeatrix:
@toolmantim also put the Add Tests code lense behind the unstable flag so the new /test command won't get blocked by it.

vscode/src/commands/services/CommandCodeLenses.ts
abeatrix: Yea we should update the code lenses logic in a follow up PR!

toolmantim (approved):
Works great with the examples I found now! Let's just go with a single `/test` command (edit) for now though.

PR #2961, opened 2024-01-31
Finishing autocomplete Cody Ignore

arafatkatze:
This is the Autocomplete step of my context hardening for Cody Ignore https://github.com/sourcegraph/sourcegraph/issues/59683 
[Document](https://docs.google.com/document/d/1LUNSOuwwhgycfCvqoI3AETS-sDIjCpZknFv2ueQ8Ctw/edit) 
 
Snippet
image

## Test plan

Try it in the debugger UI. USe this file for instructions of Cody Ignore https://sourcegraph.com/github.com/sourcegraph/cody@adding-autocomplete-ignore/-/blob/vscode/doc/codyignore.md?L16

### DIFF
--- a/vscode/src/completions/context/context-mixer.ts
+++ b/vscode/src/completions/context/context-mixer.ts
@@ -91,6 +91,7 @@ export class ContextMixer implements vscode.Disposable {
                             },
                         })
                 )
+                // Filtering out snippets that belong to ignored files
                 const filteredSnippets = allSnippets.filter(snippet => !isCodyIgnoredFile(snippet.uri))
 
                 return {
--- a/vscode/src/completions/context/context-mixer.ts
+++ b/vscode/src/completions/context/context-mixer.ts
@@ -1,6 +1,6 @@
 import type * as vscode from 'vscode'
 
-import { wrapInActiveSpan } from '@sourcegraph/cody-shared'
+import { wrapInActiveSpan, isCodyIgnoredFile } from '@sourcegraph/cody-shared'
 
 import type { DocumentContext } from '../get-current-doc-context'
 import type { ContextSnippet } from '../types'
@@ -80,7 +80,7 @@ export class ContextMixer implements vscode.Disposable {
         const results = await Promise.all(
             retrievers.map(async retriever => {
                 const retrieverStart = performance.now()
-                const snippets = await wrapInActiveSpan(
+                const allSnippets = await wrapInActiveSpan(
                     `autocomplete.retrieve.${retriever.identifier}`,
                     () =>
                         retriever.retrieve({
@@ -91,11 +91,12 @@ export class ContextMixer implements vscode.Disposable {
                             },
                         })
                 )
+                const filteredSnippets = allSnippets.filter(snippet => !isCodyIgnoredFile(snippet.uri))
 
                 return {
                     identifier: retriever.identifier,
                     duration: performance.now() - retrieverStart,
-                    snippets: new Set(snippets),
+                    snippets: new Set(filteredSnippets),
                 }
             })
         )
--- a/vscode/src/completions/context/context-mixer.ts
+++ b/vscode/src/completions/context/context-mixer.ts
@@ -91,7 +91,6 @@ export class ContextMixer implements vscode.Disposable {
                             },
                         })
                 )
-                // Filtering out snippets that belong to ignored files
                 const filteredSnippets = allSnippets.filter(snippet => !isCodyIgnoredFile(snippet.uri))
 
                 return {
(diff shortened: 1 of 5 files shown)


vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
```diff
@@ -1,8 +1,6 @@
 import * as vscode from 'vscode'
 import type { URI } from 'vscode-uri'
 
-import { isCodyIgnoredFile } from '@sourcegraph/cody-shared'
```
arafatkatze, line 4: These are not needed anymore because the change in Context mixer covers all the context from these files. 

Its better to have as single place to filter all the context sources rather than do a whack a mole for many different sources which might come and go depending on our needs. 

Beatrix did this for chat and now I wanna do this for Completions and possibly for commands too.

abeatrix: Commands go through the same route as chat so we won't need a separated one for commands.

For autocompletion there is a place where we process all the context into message, that's probably the best place to check for Cody ignored files.

abeatrix:
Cc @valerybugakov @philipp-spiess for second pairs of eyes 👀

vscode/src/completions/context/context-mixer.ts
```diff
@@ -91,11 +91,13 @@ export class ContextMixer implements vscode.Disposable {
...
                         })
                 )
+                // Filtering out snippets that belong to ignored files
+                const filteredSnippets = allSnippets.filter(snippet => !isCodyIgnoredFile(snippet.uri))
 
                 return {
                     identifier: retriever.identifier,
                     duration: performance.now() - retrieverStart,
-                    snippets: new Set(snippets),
+                    snippets: new Set(filteredSnippets),
```
abeatrix, line 100: this looks good to me, but wonder if mixedContext.filter(c=>!isCodyIgnoredFile(c.uri) at the bottom is a better place to do the filtering -- will defer to the valery and philipp to confirm.

abeatrix (commented):
left a comment in line, but the rest looks good to me. Thanks for cleaning this up! Would be nice to add a test to cover this :)

philipp-spiess (approved):
LGTM. One nit inline. Also do you mind adding a unit test to `vscode/src/completions/context/context-mixer.test.ts`? :)

vscode/src/completions/context/context-mixer.ts
```diff
@@ -91,11 +91,13 @@ export class ContextMixer implements vscode.Disposable {
                             },
                         })
                 )
+                // Filtering out snippets that belong to ignored files
+                const filteredSnippets = allSnippets.filter(snippet => !isCodyIgnoredFile(snippet.uri))
```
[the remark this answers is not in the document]
arafatkatze: Fixed

arafatkatze:
Hey @philipp-spiess, 
> do you mind adding a unit test to vscode/src/completions/context/context-mixer.test.ts? :)

I went down the rabbithole of trying to mock isCodyignored for testing and apparently it wasn't as easy as I originally thought but regardless I wouldn't mind doing it if it makes sense. 

Looking further I see that the there is no external testing for isCodyignored because all the testing is done in [this file](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Asourcegraph%2Fcody.git&branch=adding-autocomplete-ignore&file=lib%2Fshared%2Fsrc%2Fcody-ignore%2Fignore-helper.test.ts&editor=VSCode&version=2.2.16&start_row=16&start_col=19&end_row=16&end_col=19) for the ignoredhelper stuff which is basically all the code that does Cody ignore.

I believe the current thing for testing makes sense because a single point where all the ignore stuff is tested is good. 


If I were to have a test here I am essentially testing the filter method which seems frivolous. With that being said perhaps you have something else in mind and if so please let me know coz I would be down to implement that kind of test.

abeatrix:
> Hey @philipp-spiess,
> 
> > do you mind adding a unit test to vscode/src/completions/context/context-mixer.test.ts? :)
> 
> I went down the rabbithole of trying to mock isCodyignored for testing and apparently it wasn't as easy as I originally thought but regardless I wouldn't mind doing it if it makes sense.
> 
> Looking further I see that the there is no external testing for isCodyignored because all the testing is done in [this file](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Asourcegraph%2Fcody.git&branch=adding-autocomplete-ignore&file=lib%2Fshared%2Fsrc%2Fcody-ignore%2Fignore-helper.test.ts&editor=VSCode&version=2.2.16&start_row=16&start_col=19&end_row=16&end_col=19) for the ignoredhelper stuff which is basically all the code that does Cody ignore.
> 
> I believe the current thing for testing makes sense because a single point where all the ignore stuff is tested is good.
> 
> If I were to have a test here I am essentially testing the filter method which seems frivolous. With that being said perhaps you have something else in mind and if so please let me know coz I would be down to implement that kind of test.

i added a test to the context mixer for cody/ignore that can be merged to your branch: https://github.com/sourcegraph/cody/pull/2986

arafatkatze:
Thanks to Beatrix the test is fully solved.

PR #2962, opened 2024-01-31
Autocomplete: add `single-multiline-request` feature flag

valerybugakov:
## Context

- Adds `cody-autocomplete-single-multiline-request` feature flag to reduce the number of requests for multiline completions from three to one.
- Adds `CompletionProviderConfig` singleton that prefetches completion provider feature flags and stores the latest configuration object to reduce the "prop-drilling" of completion provider options.

## Test plan

CI

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -17,7 +17,7 @@ import { telemetryService } from '../services/telemetry'
 
 import { getArtificialDelay, resetArtificialDelay, type LatencyFeatureFlags } from './artificial-delay'
 import { ContextMixer } from './context/context-mixer'
-import { DefaultContextStrategyFactory, type ContextStrategy } from './context/context-strategy'
+import { DefaultContextStrategyFactory } from './context/context-strategy'
 import type { BfgRetriever } from './context/retrievers/bfg/bfg-retriever'
 import { getCompletionIntent } from './doc-context-getters'
 import { FirstCompletionDecorationHandler } from './first-completion-decoration-handler'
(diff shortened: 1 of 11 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -152,7 +146,10 @@ export class InlineCompletionItemProvider
 
         this.requestManager = new RequestManager()
         this.contextMixer = new ContextMixer(
-            new DefaultContextStrategyFactory(config.contextStrategy, createBfgRetriever)
+            new DefaultContextStrategyFactory(
```
abeatrix, line 149: Since we want to make sure .cody/ignore will keep working with autocomplete. is this a good place to filter ignored file? Something like `mixedContext.filter(f => !isCodyIgnoredFile(f.uri))`?

abeatrix (commented):
looks like it's just moving exisiting methods around with a new feature flag added? 
left a comment inline regarding filtering .cody/ignore context, but the rest looks good to me.

valerybugakov:
> looks like it's just moving exisiting methods around with a new feature flag added?

pretty much, yep

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Yeah, we can implement it on the `ContextMixer` level since all the context retrievers should go through it.

abeatrix (approved):

philipp-spiess (commented):
There's a subtle behavior change here where we no longer re-evaluate some feature flags every 30min like we did before. 

Before this change, the re-evaluation behavior was depending on where in the code the feature flag was accessed. If we access it inside a  completion handler, we could be sure that all clients have the updated value within at most 30minutes of the feature flag change. This is super useful as it means we don't have to ways _days_ for changes to roll out. Some other features (the one we queried doing setup) would require an instance reload.

With this change, every feature flag change will require an instance reload. This is probably find but I want to make sure we're aware of this change.

In an ideal world, every feature flag change would be applied almost instantly, but that would need additional logic to reload the autocomplete provider that we don't have yet.

valerybugakov:
Discussed with Philipp in DMs, we should be good here because the refresh timeout is used in the featureFlagProvider.getFromCache method that the updated code relies on.

valerybugakov:
- A small follow-up improvement: https://github.com/sourcegraph/cody/pull/2972

Issue #2920, opened 2024-01-27
Add agent integration test coverage to verify .cody/ignore behavior


abeatrix:
Add Integration tests to agent to verify the cody ignore works with different features and context sources

features include:
- [ ] autocomplete
- [ ] chat
- [ ] commands
- [ ] inline edit

local context source include:
- [ ] editor https://github.com/sourcegraph/cody/pull/2963
- [ ] symf (local search) https://github.com/sourcegraph/cody/pull/2963
- [ ] autocomplete context mixer: https://github.com/sourcegraph/cody/pull/2986

PR #2963, opened 2024-01-31
add: .cody/ignore integration tests in agent

abeatrix:
Part of https://github.com/sourcegraph/cody/issues/2920

Summary:
- Adding a new `check/isCodyIgnoredFile` request handler to check if a file path is being ignored by Cody
- Updating tests to use this new request instead of the isCodyIgnoredFile utility method
- Adding a test to confirm .cody/ignore is working on startup

This PR adds integration tests for the following features to agent:
- [x] autocomplete https://github.com/sourcegraph/cody/pull/2986
- [x] chat (chat submission)
- [x] chat command (/explain)
- [x] inline edit command (/doc)

Local context source covered in the test:
- [x] editor (used by commands, chat, inline-edit, auto-completes)
- [x] symf (local search) (used by chat when enhanced context is enabled)
- [x] local context used by autocompletion https://github.com/sourcegraph/cody/pull/2986

Also updated to allow remote context which should be filtered during sync time (https://github.com/sourcegraph/sourcegraph/pull/59836) to allow support for multi-repo context in clients.

## Follow-Up Works

- [x] Add assertions for checking input context during Network Request

## Test plan

Green bots means 👍


lib/shared/src/cody-ignore/ignore-helper.ts
```diff
@@ -109,7 +109,13 @@ export class IgnoreHelper {
             return false
         }
 
-        // Ignore all non-file URIs
+        // Remote context (e.g. unified, multi-repo) are filtered during sync time,
+        // and not ignored by clients.
+        if (uri.scheme === 'https') {
```
abeatrix, line 114: Without this, the `'chat/submitMessage (addEnhancedContext: true, multi-repo test)'` test would fail as we will remove ALL context returned from the remote host.

chwarwick:
@abeatrix This will be great to get tests in place for.  Currently this is validating that the ignore worked by asserting on the outputs, however what we are trying to protect against is adding ignored content to the input.  [This snippet ](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/index.test.ts?L970-978) shows how you can get access to the requests that were sent.  Adding assertions to the content sent to the llm would provide a much higher level of confidence with this feature.

abeatrix:
> @abeatrix This will be great to get tests in place for. Currently this is validating that the ignore worked by asserting on the outputs, however what we are trying to protect against is adding ignored content to the input. [This snippet ](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/index.test.ts?L970-978) shows how you can get access to the requests that were sent. Adding assertions to the content sent to the llm would provide a much higher level of confidence with this feature.

Thanks for pointing me in the right direction to test this feature reliably! I will work on adding assertions to check the content sent in network requests. 

@chwarwick should I keep what I have here now for asserting on the output, or do you think it'd be unnecessary?

chwarwick:
> @chwarwick should I keep what I have here now for asserting on the output, or do you think it'd be unnecessary?

I think they good to leave since it covers the expectations of what the output will be including the post processing steps. For example if running autocomplete in an ignored file started returning errors instead of empty responses that would be visible.

agent/src/index.test.ts
```diff
@@ -548,6 +548,40 @@ describe('Agent', () => {
...
+
+                groupedMsgs.push(...(text ?? []))
+            }
+
+            expect(groupedMsgs.length).toBeGreaterThan(0)
+
+            // Join all the string from each groupedMsgs[] together into
+            // one block of text, and then check if it contains the ignored file name
+            // to confirm context from the ignored file was not sent to the server.
+            const groupedText = groupedMsgs.flat().join(' ')
```
abeatrix, line 577: @chwarwick Would this approach work? 😅

chwarwick: Sure, if we find something better in the future we can update it.  As long as we know it catches the failure condition I think it's a positive.

agent/src/index.test.ts
```diff
@@ -461,6 +449,143 @@ describe('Agent', () => {
         })
     })
 
+    describe.skipIf(isNode16())('Cody Ignore', () => {
```
DanTup, line 452: I think it would be useful if we annotated when we skip tests.

I think right now we run tests on Node 16 + 20, but latest VS Code is using Node 18 so if we only tested 20, we might miss bugs caused by using Node 20 APIs.

That said, I don't know if this code runs in VS Code at all. If those code is only run in a bundled Node 20 VM for other editors, then maybe the entire test suite should be skipped for Node 16?

DanTup (commented):
It looks reasonable to me, but I'm not too familiar with the code or the agent tests so I'll defer to others for this :)

agent/src/index.test.ts
abeatrix: was able to fix it once i move the check to the request handler!! Thank you so much for debugging this with me!

chwarwick (approved):

PR #2964, opened 2024-01-31
VSCode UI: Add an e2e smoke test for the multi-repo picker

dominiccooney:
Most interesting thing here: Adds the ability to mock GraphQL responses at a fine granularity, instead of having everyone dump them into the mock server.

## Test plan

```
cd vscode
pnpm test:e2e -- context-settings.test.ts
```


abeatrix (approved):
Nice test! 🚀

PR #2969, opened 2024-01-31
Agent: disable reusing webview panels

olafurpg:
Previously, the ID from `chat/new` was reused for `commands/document` if no message was sent to `chat/new`. This behavior was due to a desire to reuse webview panels in VS Code, which is not behavior that we want to mirror for the agent. Now, we always create a new ID for these different commands even if no message has been sent to the original `chat/new` ID.

## Test plan

See added test case. You can manually verify that the test fails if you comment out the if condition

```ts
!this.options.config.isRunningInsideAgent && // Don't reuse panels in the agent
```


pkukielka (approved):
LGTM

PR #2971, opened 2024-01-31
Agent: fail tests if a network request errors

olafurpg:
Previously, the agent tests were passing even if Polly encountered HTTP requests with missing recordings. This happened frequently for PRs where a new request was added to get feature flags, which didn't fail the test, but added a lot of noise to the console output due to Polly errors about missing recordings. This PR changes the agent tests so they fail if we run in replay mode and send a network request without an accompanying recording. Hopefully, this change will prevent the noisy standard output AND it will improve our code reviews for PRs that add new network requests (because the PR author must document what kind of network requests got added in the PR).

## Test plan
Green CI. You can manually verify that the tests fail if you remove a recording from the recording file (search for `_id:` to quickly select a single recording)


pkukielka (approved):
LGTM

PR #2974, opened 2024-01-31
Autocomplete: Remove obvious prompt-continuations

### LINKED ISSUE #2912: bug: Autocomplete suggestion showing "Human:..."
### Version

v1.1.3

### Describe the bug

Using autocomplete against an enterprise instance configured to use anthropic
received the following autocomplete suggestion:
```
"result": {
    "completion": "\nHuman: Here is some more context code to provide:",
    "stopReason": "stop_sequence"
  },
```

### Expected behavior

Autocomplete should not have "Human:" and should be code.

### Additional context

I saved the entire debug log output ask me if needed.

philipp-spiess:
Closes #2912

If the completion is very obviously a prompt continuation, we should not show it. I noticed this too a bunch of time with `// Path: ...` style completions in StarCoder.

## Test plan

- Added unit tests

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -14,7 +14,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 ### Changed
 
 - Autocomplete: Improved the new jaccard similarity retriever and context mixing experiments. [pull/2898](https://github.com/sourcegraph/cody/pull/2898)
-- Autocomplete: Remove obvious prompt-continuations. [pull/]()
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
Does it affect the quality of the response?

philipp-spiess:
@abeatrix Can you clarify this question? This is going to hide those bad responses but they can still happen. In practice though, this doesn't seem to happen often. Would be insightful to have analytics though how many things we filter out with these checks

chwarwick (approved):

PR #2978, opened 2024-01-31
Agent: add support for `workspace/edit`

olafurpg:
Previously, the agent would crash when we hit on a codepath that used
the `vscode.WorkspaceEdit` class. This class is used to perform more
advanced edits on the workspace (multi-file, create files, rename files,
delete files, ..). This functionality is used by some inline-edit
commands like the `unit-tests` command that writes the generated test
case to the file instead of to the chat.

Now, the agent implements `vscode.WorkspaceEdit` so that it no longer
crashes. There are also several new JSON-RPC endpoints to support
applying workspace edits:

- `workspace/edit`: request sent from server to client to apply an edit across
  multiple files.
- `workspace/didCreateFiles`: notification sent from client to server to
  notify files were created.
- `workspace/didDeleteFiles`: notification sent from client to server to
  notify files were deleted.
- `workspace/didRenameFiles`: notification sent from client to server to
  notify files were renamed.
- `textDocument/openUntitledDocument`: request sent from server to client to open an untitled document.
- `textDocument/show`: request sent from server to client to prominently show or make a document visible in the IDE.

The current implementation is not perfect. I didn't spend much time
perfecting all the details of how we apply edits. The goal of this PR is
to unblock further progress.

## Test plan

This PR adds basic tests for the new inline-edit test command that uses `WorkspaceEdit`. I'm not super confident that the test client is covering all cases. However, it's more important for IDE clients like JetBrains to implement this endpoint 100% correctly.

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -1,3 +1,4 @@
+import * as vscode from 'vscode'
 import assert from 'assert'
 import { execSync } from 'child_process'
 import fspromises from 'fs/promises'
@@ -5,7 +6,6 @@ import os from 'os'
 import path from 'path'
 
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from 'vitest'
-import { Uri } from 'vscode'
 
 import { isWindows } from '@sourcegraph/cody-shared'
 
@@ -36,7 +36,7 @@ const explainPollyError = `
     `
 
 const prototypePath = path.join(__dirname, '__tests__', 'example-ts')
-const workspaceRootUri = Uri.file(path.join(os.tmpdir(), 'cody-vscode-shim-test'))
+const workspaceRootUri = vscode.Uri.file(path.join(os.tmpdir(), 'cody-vscode-shim-test'))
 const workspaceRootPath = workspaceRootUri.fsPath
 
 const mayRecord =
@@ -111,13 +111,13 @@ describe('Agent', () => {
     })
 
     const sumPath = path.join(workspaceRootPath, 'src', 'sum.ts')
-    const sumUri = Uri.file(sumPath)
+    const sumUri = vscode.Uri.file(sumPath)
     const animalPath = path.join(workspaceRootPath, 'src', 'animal.ts')
-    const animalUri = Uri.file(animalPath)
+    const animalUri = vscode.Uri.file(animalPath)
     const squirrelPath = path.join(workspaceRootPath, 'src', 'squirrel.ts')
-    const squirrelUri = Uri.file(squirrelPath)
+    const squirrelUri = vscode.Uri.file(squirrelPath)
     const multipleSelections = path.join(workspaceRootPath, 'src', 'multiple-selections.ts')
-    const multipleSelectionsUri = Uri.file(multipleSelections)
+    const multipleSelectionsUri = vscode.Uri.file(multipleSelections)
 
     it('extensionConfiguration/change (handle errors)', async () => {
         // Send two config change notifications because this is what the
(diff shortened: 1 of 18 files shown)


agent/src/index.test.ts
```diff
@@ -665,6 +662,51 @@ describe('Agent', () => {
             )
         }, 30_000)
 
+        it('editCommand/test', async () => {
+            const uri = Uri.file(path.join(workspaceRootPath, 'src', 'missing-test.ts'))
+            await client.openFile(uri)
+            const id = await client.request('editCommands/test', null)
+            await client.taskHasReachedAppliedPhase(id)
+            const document = client.workspace.getDocument(uri)!
+            // Currently, the test doesn't create a new file for some reason. My
```
olafurpg, line 671: FYI @abeatrix I couldn't get test-edit to create a separate file in the test case even if the cursor was placed at the bottom of the file. I left the assertion here with the current behavior, but we should ideally fix the behavior so that it reliably creates a new file instead. There may be some bugs in the agent infrastructure to deal with creating new files because I wasn't able to reproduce a case where this happens.

lib/shared/src/cody-ignore/ignore-helper.ts
```diff
@@ -109,9 +109,11 @@ export class IgnoreHelper {
             return false
         }
 
-        // Ignore all non-file URIs
+        // Return all non-file URIs on the assumption that they origin from
+        // remote multi-repo context files, which are already filtered by the
+        // backend to respect codyignore files.
         if (uri.scheme !== 'file') {
-            return true
+            return false
```
olafurpg, line 116: FYI @abeatrix flipping this boolean to get the tests passing when `unstable: true`.

abeatrix: which test was it failing before this change?  i've made a similar change in my PR to add tests to agent, wondering if it's the same case here: https://github.com/sourcegraph/cody/pull/2963/files#diff-b39e270dd4dedc7dc7068542a8c942b0e70640f79e53bf1e5067c60e9e9a50c4

olafurpg: Alternatively, run the URI path against the ignore filters regardless of the scheme.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -64,12 +64,13 @@ export type Requests = {
     // shortcuts to start a new chat with a templated question. The return value
     // of these commands is the same as `chat/new`, an ID to reference to the
     // webview panel where the reply from this command appears.
-    'commands/explain': [null, string]
+    'commands/explain': [null, string] // TODO: rename to chatCommands/{explain,text,smell}
```
abeatrix, line 67:
```suggestion
    'commands/explain': [null, string] // TODO: rename to chatCommands/{explain,test,smell}
```

agent/src/index.test.ts
```diff
@@ -665,6 +662,51 @@ describe('Agent', () => {
             )
         }, 30_000)
 
+        it('editCommand/test', async () => {
+            const uri = Uri.file(path.join(workspaceRootPath, 'src', 'missing-test.ts'))
```
abeatrix, line 666:
```suggestion
            const uri = Uri.file(path.join(workspaceRootPath, 'src', 'missing.ts'))
```

ooooof this is because we would try to figure out if the current file is a test file or not based on the file name. 
If it has a common test file naming convention, as you do here, we would skip the file creation step and stream directly into the "test" file with a different prompt 😅  I will look into that in a follow up PR, but I think you should be able to create an assertion that reflects the expected behavior with a "non-test" file path here.

agent/src/index.test.ts
abeatrix: See the comment above. I think this behavior is expected (for now), but we should add another assertion to reflect the other behavior as well as there are two behaviors here.

agent/src/index.test.ts
```diff
@@ -665,6 +662,51 @@ describe('Agent', () => {
...
+            await client.openFile(uri)
+            const id = await client.request('editCommands/test', null)
+            await client.taskHasReachedAppliedPhase(id)
+            const document = client.workspace.getDocument(uri)!
+            // Currently, the test doesn't create a new file for some reason. My
+            // expectation is that this command creates a separate file. We
+            // should update this test to reflect the new behavior when we start
+            // creating a new file.
+            expect(trimEndOfLine(document.getText())).toMatchInlineSnapshot(`
+              "export function trickyLogic(a: number, b: number): number {
```
abeatrix, line 676: Expected Behavior 1: The expected behavior in a brand new test file (e.g: "missing-test.ts" / "missing_test.ts") should look like this if we rename the file where we invoke the test command from `missing-test.ts` to `missing.ts`

```suggestion
import { trickyLogic } from './missing'
describe('trickyLogic', () => {
    it('returns 1 if a is 0', () => {
        expect(trickyLogic(0, 1)).toBe(1)
    })
    it('returns 1 if b is 2', () => {
        expect(trickyLogic(1, 2)).toBe(1)
    })
    it('returns a - b otherwise', () => {
        expect(trickyLogic(3, 1)).toBe(2)
    })
})
```

Expected Behavior  2: if there is already an existing test file (e.g: "missing-test.ts" / "missing_test.ts"), the test file should look like this if we rename the file where we invoke the test command from `missing-test.ts` to `missing.ts`

```suggestion
import { trickyLogic } from './missing'
describe('trickyLogic', () => {
    it('returns 1 if a is 0', () => {
        expect(trickyLogic(0, 1)).toBe(1)
    })
    it('returns 1 if b is 2', () => {
        expect(trickyLogic(1, 2)).toBe(1)
    })
    it('returns a - b otherwise', () => {
        expect(trickyLogic(3, 1)).toBe(2)
    })
})
describe('trickyLogic', () => {
    it('returns a - b when a and b are numbers other than test cases', () => {
        const result = trickyLogic(3, 1)
        expect(result).toBe(2)
    })
})
```

abeatrix (changes requested):
Thanks for adding the test! I've left some comments inline with suggested change.

This is my bad as I should have documented the behavior better to avoid this sort of confusion, will work on a follow-up PR to address that, sorry about that!

vscode/src/jsonrpc/agent-protocol.ts
olafurpg: Fixed

agent/src/index.test.ts
olafurpg: Good catch! That makes a lot of sense. I updated the filename to be `trickyLogic.ts` and it now works as expected :D

agent/src/index.test.ts
olafurpg: I've updated the test case to assert that the test is added to a separate file. It's still not 100% correct because it's an untitled file. I see requests to open `untitled:TARGET_PATH` where TARGET_PATH has `trickyLogic.test.ts` but I wasn't clear on the expected behavior for such untitled URIs.

lib/shared/src/cody-ignore/ignore-helper.ts
olafurpg: I reverted this change because I realized the inline-test command works even when unstable is disabled. Leaving it to your PR  to enable unstable and fix the codyignore handling of https schems.

olafurpg:
@abeatrix thank you for the review! The current solution is not 100% correct but it adds most of tricky infrastructure to wire all the requests/notifications between the client/server. I ended up disabling the new `editCommand/test` because it was timing out in ubuntu 20 only (passing in others). I have several other projects I need to look into so I'm going to delete further `WorkspaceEdit`-related changes to somebody else.

abeatrix (approved):
Thank you for adding this and for starting the test for the new /test command 🤗 !

QQ: Do the agent tests and agent client (?) run separately? I noticed before the test started, 1 IgnoreHelper was created, and then when the test started 4 more were created. Is that expected?

lib/shared/src/cody-ignore/ignore-helper.ts
```diff
@@ -109,7 +109,9 @@ export class IgnoreHelper {
             return false
         }
 
-        // Ignore all non-file URIs
+        // Return all non-file URIs on the assumption that they origin from
+        // remote multi-repo context files, which are already filtered by the
+        // backend to respect codyignore files.
```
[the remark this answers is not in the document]
abeatrix: No action needed here since I can update it on my branch, but I just wanted to flag this :D

PR #2983, opened 2024-01-31
Agent: handle document URIs with exclamation marks

olafurpg:
Previously, the agent threw an error for `textDocument/*` notificattions when the URI included an exclamation mark. The error was thrown to preserve an important variant that we can always parse and render a URI back into the original string
```ts
vscode.Uri.parse(uri).toString() === uri
```
This variant was getting violated when the JB plugin sent document URIs that included an exclamation mark in them, for example:

```
file:///Users/com.jetbrains/ideaIC-2022.1-sources.jar!/com/intellij/RequiresBackgroundThread.java
```

The `!` got encoded as `%21`. This PR "fixes" the problem by mutating the `uri: string` property to get the `Uri.toString()` formatted URI. t's not a super satisfying solution, it would be more desirable if we can get `vscode.Uri.parse(..).toString()` to be a safe roundtrip. However, this change gets the most important job done: we preserve the invariant.

## Test plan
See updated tests.

### DIFF
--- a/vscode/src/jsonrpc/TextDocumentWithUri.ts
+++ b/vscode/src/jsonrpc/TextDocumentWithUri.ts
@@ -1,6 +1,7 @@
 import * as vscode from 'vscode'
 
 import type { Range, ProtocolTextDocument } from './agent-protocol'
+import { logDebug } from '../log'
 
 /**
  * Wrapper around `ProtocolTextDocument` that also contains a parsed vscode.Uri.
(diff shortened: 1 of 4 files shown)


mkondratek:
huh, strange... I still see a similar error:

```
2024-01-31 21:27:04,842 [ 558131]   WARN - #c.s.c.a.CodyAgentClient - Cody by Sourcegraph: █ ProtocolTextDocumentWithUri: correcting invariant violation file:///Users/mkondratek/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/2022.1/4aa17d162c81293d88aa56a54f655e784a8ba8ab/ideaIC-2022.1-sources.jar%21/com/intellij/util/concurrency/FutureResult.java (this.uri) !== file:///Users/mkondratek/.gradle/caches/modules-2/files-2.1/com.jetbrains.intellij.idea/ideaIC/2022.1/4aa17d162c81293d88aa56a54f655e784a8ba8ab/ideaIC-2022.1-sources.jar!/com/intellij/util/concurrency/FutureResult.java (this.underlying.uri)
```

However the context (the selection) is passed properly and the response looks good.

image

vscode/src/jsonrpc/TextDocumentWithUri.ts
```diff
@@ -10,15 +11,18 @@ import type { Range, ProtocolTextDocument } from './agent-protocol'
...
         this.underlying = underlying ?? { uri: uri.toString() }
         if (this.underlying.uri !== uri.toString()) {
-            throw new Error(
-                `ProtocolTextDocumentWithUri invariant violation: ${this.uri} (this.uri) !== ${this.underlying.uri} (this.underlying.uri)`
+            logDebug(
+                'ProtocolTextDocumentWithUri',
+                'correcting invariant violation',
+                `${this.uri} (this.uri) !== ${this.underlying.uri} (this.underlying.uri)`
             )
+            this.underlying.uri = uri.toString()
```
mkondratek, line 25: Ok, I see - instead of throwing an error you are logging the same message.

olafurpg: You can add a special case to ignore the `"ProtocolTextDocumentWithUri"` channel for the JetBrains plugin. I don't want to silently do this behavior because there might be other places where we are silently hitting on bugs because we are not doing this workaround

PR #2984, opened 2024-01-31
display .cody/ignore status in status bar

abeatrix:
Update status bar background color on ignored files (behind unstable flag)

- Added functionality to the status bar in StatusBar.ts to display whether the current file is ignored by Cody or not by switching the background color of the status bar item.

This feature allows users and devs to easily identify if the current file is being ignored by Cody directly from the editor.

This feature is behind the unstable feature flag. When unstable flag is not enabled, the background will not be updated based on ignore file.

Note: I found it hard to tell if cody ignore is active or not during development and testing. We can remove it when this is out of unstable state. 

## Test plan

1. Go to a Cody ignored file
2. Check if the status bar item update accordingly

### Updated

As discussed with Tim, we willl update the background color to indicates if the current file is ignored by Cody or not instead of adding Ignore next to Cody icon

image

### Outdated

image

### DIFF
--- a/vscode/src/services/StatusBar.ts
+++ b/vscode/src/services/StatusBar.ts
@@ -38,19 +38,6 @@ export function createStatusBar(): CodyStatusBar {
     statusBarItem.command = 'cody.status-bar.interacted'
     statusBarItem.show()
 
-    // Listens for changes to the active text editor and updates the status bar text
-    // based on whether the active file is ignored by Cody or not.
-    // If ignored, adds 'Ignored' to the status bar text.
-    // Otherwise, rerenders the status bar.
-    const onDocumentChange = vscode.window.onDidChangeActiveTextEditor(e => {
-        if (e && isCodyIgnoredFile(e?.document.uri)) {
-            statusBarItem.tooltip = 'Current file is ignored by Cody'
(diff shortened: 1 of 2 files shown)


arafatkatze (approved):
Approving in the sense that the PR does what it claim to do.

image
When I am hovering on the cody image thingy sometimes it shows the file is ignored by  codyignore and sometimes it takes a while to do the showing. When I actually click on it I see

image

Which doesn't say anything about the error. I wonder if it would be nice to have something like 

image

When clicking on the yellow cody icon. Thats what we do for when admin disables the autocomplete

Still approving coz its likely you have had this discussion

toolmantim:
I was going to ask the same thing. I believe in the other wanting background states we show an item in the quickpick to explain why — let's do the same here too.

abeatrix:
> I was going to ask the same thing. I believe in the other wanting background states we show an item in the quickpick to explain why — let's do the same here too.

what's a wanting background?

I can update it, but this change is mainly for internal dev only. I agree we will need a proper UI and UX ready before we can move .Cody/ignore out of the internal unstable flag. I've listed that as a to-do in the Cody ignore tracking issue, will create a new issue for the product and design team!

PR #2987, opened 2024-02-01
Autocomplete: update BFG snippets structure to match the format expected on the client

valerybugakov:
## Context

- I started testing BFG locally and bumped into a runtime error. The context mixer [logic](https://github.com/sourcegraph/cody/blob/7106a4fc2c8d088961f66de2336217a87805136f/vscode/src/completions/context/context-mixer.ts#L110) expects context snippets to have the `uri` field. This field [was added](https://github.com/sourcegraph/cody/pull/2749) relatively recently, but neither the BFG implementation nor the client retriever was updated accordingly. 
- This PR fixes the issue on the client side and adds additional types that would prevent the problem from happening. AFAIU, this solution is similar to the approach `LocalEmbeddingsController` uses, where local embeddings snippet types returned by Cody engine [are defined separately](https://github.com/sourcegraph/cody/blob/main/vscode/src/jsonrpc/embeddings-protocol.ts#L21) from [global types](https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/types.ts#L32) used by context mixer.

## Test plan

CI

### DIFF
--- a/vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
+++ b/vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
@@ -197,7 +197,24 @@ export class BfgRetriever implements ContextRetriever {
                 return []
             }
 
-            return [...(responses?.symbols || []), ...(responses?.files || [])]
+            const mergedContextSnippets = [...(responses.symbols || []), ...(responses?.files || [])]
+
+            // Convert BFG snippets to match the format expected on the client.
+            const symbols = mergedContextSnippets.map(contextSnippet => ({
+                ...contextSnippet,
+                uri: vscode.Uri.from({ scheme: 'file', path: contextSnippet.fileName }),
+            })) satisfies Omit<ContextSnippet, 'startLine' | 'endLine'>[]
+
+            logDebug(
+                'CodyEngine',
+                'bfg/contextAtPosition',
+                `returning ${mergedContextSnippets.length} results`
+            )
+
+            // TODO: add `startLine` and `endLine` to `responses` or explicitly add
+            // another context snippet type to the client.
+            // @ts-ignore
+            return symbols
         } catch (error) {
             logDebug('CodyEngine:error', `${error}`)
             return []
--- a/vscode/src/jsonrpc/bfg-protocol.ts
+++ b/vscode/src/jsonrpc/bfg-protocol.ts
@@ -2,14 +2,22 @@
  * This file declares the protocol for communicating between Cody and BFG (Blazingly Fast Graph), a Rust implementation
  * of the "Graph Context" feature flag.
  */
-import type { FileContextSnippet, SymbolContextSnippet } from '../completions/types'
-
 import type { Position, Range } from './agent-protocol'
+
+export interface BFGFileContextSnippet {
+    fileName: string
+    content: string
+}
+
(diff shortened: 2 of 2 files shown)


dominiccooney (approved):
Approving because this seems like an improvement but the TODO and ts-ignore make me 😨

Is the lack of strong type checking at the bfg-TypeScript boundary the cause of this badness?

olafurpg (approved):

olafurpg:
> Is the lack of strong type checking at the bfg-TypeScript boundary the cause of this badness?

The types in the JSON-RPC protocol don't have any accompanying runtime validation, we parse the JSON and cast it into the expected type. In this case, the protocol had breaking changes when we made changes to `SymbolContextSnippet` and `FileContextSnippet`. I've been noodling with ideas of writing a code generator for the protocol, which we need to run to pass CI and will surface breaking changes like these much faster. It would also make it easier to write clients in Rust/Kotlin/Lua/Python when we can auto-generate bindings.

You're probably thinking that we should move to a dedicated RPC framework instead of creating Yet Another code generator. It's definitely worth considering. One risk with that approach is that we are currently reusing TypeScript types from the Cody codebase that are constantly evolving from feature development by the VSC client team. The benefit of writing a bespoke TypeScript code generator is that we can keep using these types from the VSC client team. Moving to an RPC framework won't save us from the labour of converting RPC types into the TypeScript types the Cody VSC extension is using, unless we refactor the extension to use RPC types everywhere.

valerybugakov:
The changes were made on the client side, and because we reused client types for BFG responses, TS was happy even though BFG implementation was not updated. Having independent types should prevent this from happening. 

And yeah, we need to address the implicit fact that BFG snippets do not have start/end line fields. I'm surprised @philipp-spiess accounted for this case despite types being silent about this.

dominiccooney:
@olafurpg  and I chatted IRL, to briefly respond:

> ... I've been noodling with ideas of writing a code generator for the protocol, which we need to run to pass CI and will surface breaking changes like these much faster. It would also make it easier to write clients in Rust/Kotlin/Lua/Python when we can auto-generate bindings.

Think this is a splendid idea.

> You're probably thinking that we should move to a dedicated RPC framework instead of creating Yet Another code generator. It's definitely worth considering. One risk with that approach is that we are currently reusing TypeScript types from the Cody codebase that are constantly evolving from feature development by the VSC client team. The benefit of writing a bespoke TypeScript code generator is that we can keep using these types from the VSC client team. Moving to an RPC framework won't save us from the labour of converting RPC types into the TypeScript types the Cody VSC extension is using, unless we refactor the extension to use RPC types everywhere.

I don't necessarily think an off the shelf RPC framework is the way to go. For data, TypeScript types are expressive, and slinging JSON is easy in those other languages. So we could constrain the expressivity of TypeScript a bit to hit a happy compromise of what the client languages can model, if not statically then with some runtime checks, and go from there.

philipp-spiess:
> I'm surprised @philipp-spiess accounted for this case in the context mixer logic despite types being silent about this.

Hey I raised this issue here 😆 https://github.com/sourcegraph/cody/pull/2898#issuecomment-1910702845 

> because we reused client types for BFG responses

Yeah I do think here lies the issue. If we had a forked typescript type we would at least notice that it's out of sync. I should have probably done that when I added the diversion (although it was unclear to me what the current real type is because I could not get any bfg hits to compare it so I pushed it aside)

PR #2991, opened 2024-02-01
[DO NOT MERGE] VS Code: Release 1.2.3

philipp-spiess:
This is a git-diff of the 1.2.3 release against 1.2.2 and should be used to audit the cherry pick and the release changes. I will post a seperate PR after the release to update the changelog on `main`.

This PR is not for merging, instead, the minor release branch should be updated manually to ensure we don't squash.

## Test plan

- Started VS Code and did some completions and chats.


philipp-spiess:
@valerybugakov Please take a look at the cherry picks. I would love to ship this tomorrow so we can experiment some more with fast-path!

valerybugakov (commented):
Thanks, Philipp! I went through the diff and tested completions locally. All good 👍

philipp-spiess:
Merged into the release branch. cutting now

PR #2992, opened 2024-02-01
Feature flags: Add a way to subscribe to feature flag changes

philipp-spiess:
Adds a new method to the feature flag provider in order to subscribe to changes for a specific prefix and subscribes the autocomplete codebase to all `cody.autocomplete` changes. When triggered, it will re-initialize just like when settings are changed.

The main motivation is that we want to be able to change experiment rollouts without having to wait for days until all users have reloaded the VS Code extension.

## Test plan

- Added unit tests
- https://github.com/sourcegraph/cody/assets/458591/b32636c8-66be-483e-827b-0d749bab7fc1


philipp-spiess:
This has been much more involved than I initially thought. There's still some kind of race condition in there that I don't fully understand. I think I'll refactor the feature flag provider to not mix "undefined" feature flags with the preloaded feature flag object but instead cache them in a different object. 😓  cc @valerybugakov if you have any input

valerybugakov:
> not mix "undefined" feature flags with the preloaded feature flag object but instead cache them in a different object

Yes, please! Based on our recent discussion this part is tricky to reason about. Making it more transparent and robust would be great.

philipp-spiess:
@valerybugakov This is ready for review now, I addressed the point I mentioned above

vscode/CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Added
 
+- Chat: Display Cody icon in the editor title of the chat panels when `cody.editorTitleCommandIcon` is enabled. [pull/2937](https://github.com/sourcegraph/cody/pull/2937)
```
umpox, line 9: Looks like a bad merge, needs updating!

umpox (approved):
Looks good! and v helpful for future experiments. Thanks for adding the tests

vscode/CHANGELOG.md
philipp-spiess: ouch, ty!

PR #2995, opened 2024-02-01
Edit/Chat: Ghost text improvements

umpox:
wip:

TODO:
- Gate ghost text to only when authenticated
- Expand selection for chat (if no selection)
- Add context directly to the chat
- Show “Add to chat” when already open


umpox:
Closed in favour of other PRs

PR #3005, opened 2024-02-01
change: disable edit buttons on core command messages

abeatrix:
As discussed with @toolmantim , this PR updates to disable `edit button` and `Edit Last Message` button from `/smell` and `/explain` commands, with updated e2e tests to reflect this change.

Also removing the React components for slash commands that we no longer support.

Once we have addressed the editing issues, we can re-able editing command prompts again.

## Test plan

Run the explain and smell command to confirm the edit button has been disabled, and the Edit Last Message doesn't show up.

### After

Smell Command:

image

Explain Command:

image

### Before

image


toolmantim (approved):
Great!

PR #3006, opened 2024-02-01
fix: remove sparkle from context list without enhanced context

abeatrix:
Do not display sparkle emoji in Context list if no enhanced context was used, aka messages from the core commands.

`/smell`

image

`/explain`

image

## Test plan

updated e2e tests to reflect this change.


toolmantim (approved):
Awesome!

PR #3007, opened 2024-02-02
Don't add duplicate workspace folders

beyang:
When we receive the initialization message from the client, only add the workspace folders if they don't already exist. Previously, when the agent service lived past the client lifespan, a subsequent client could add a duplicate workspace folder. (Currently, I think we only hit this in the dev environment if the agent is run separately.)

## Test plan

- [ ] Run the IntelliJ plugin against this version of the agent


pkukielka (approved):
LGTM

PR #3008, opened 2024-02-02
feat: ask Cody about terminal output

abeatrix:
Every time I encounter an error in the terminal, I need to copy and paste the error message into the chat and ask Cody to explain or debug it for me. I just discovered that there is an API to add a contribution point to the right-click menu in the VS Code terminal. I am proposing adding an option to 'Ask Cody' directly from the terminal output window.

image

This PR adds a new "Ask Cody" option to the terminal right-click menu that allows users to `Ask Cody` to explain the selection from their terminal output:
- The terminal output will be sent to chat view with a predefined prompt that they can easily edit with the edit button.
- Register a new command called `cody.command.explain-output` and bind it to the terminal view context menu.

## Test plan

Select text in your terminal and right click to `Ask Cody`: 

https://github.com/sourcegraph/cody/assets/68532117/fb33ab59-83af-449f-801d-7895da856c8c

### DIFF
--- a/lib/shared/src/commands/types.ts
+++ b/lib/shared/src/commands/types.ts
@@ -6,6 +6,7 @@ export enum DefaultChatCommands {
     Explain = 'explain', // Explain code
     Unit = 'unit', // Generate unit tests in Chat
     Smell = 'smell', // Generate code smell report in Chat
+    Terminal = 'terminal', // Explain terminal output
 }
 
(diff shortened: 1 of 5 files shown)


umpox (approved):
Love it!

toolmantim (approved):
Awesome. Can we label the action “Ask Cody to Explain”?

PR #3010, opened 2024-02-02
remove retry lens from /test

abeatrix:
context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1706865888514999

the retry button doesn't work on untitled file it seems, and since we have disabled the "edit" button for the core commands in chat, will remove the retry code lens from the /test command as well until it is fixed.

## Test plan

the current e2e test for the /test command should still pass as it only looks for "undo" and "Accept" in the test 😎


umpox (approved):
Thanks!

PR #3011, opened 2024-02-02
Edit: Add command hint filters against non-file selections and multiple selections

umpox:
## Description

Non-file selections:
- E.g. the "Cody by Sourcegraph" output log. --> Edit isn't possible here
- Multiple selections -> Edit is technically possible on the first selection, but this is confusing if a user has multiple selections, so we just hide the text for now.
   - Idea for the future: Trigger the same edits for each selection ?

## Test plan

1. Make selections in above condiitons
2. Check no command hints are shown


abeatrix (approved):
niceee

PR #3012, opened 2024-02-02
fix: handle missing prompts in custom commands

abeatrix:
context: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1706656724966049

fix: handle commands missing prompts when building map

Issue: Currently, an invalid entry in the cody.json will cause custom commands not showing up in the command menu.

Fix: Updated the get-commands utility to skip adding commands to the map if they don't contain a prompt property. It updates the command type and adds null checking before accessing the prompt property.

This prevents crashes when building the custom command map if a command is missing the expected prompt field.

## Test plan

Updated both the unit test and e2e test to ensure an invalid custom command entry would not break the custom command at build time.

Create a custom command entry without adding a prompt in your cody.json.

Before: all your custom command from that file will not show up in the custom command menu.

After: only the invalid ones will not show up in the menu, while the valid ones do.

### DIFF
--- a/vscode/src/commands/utils/get-commands.ts
+++ b/vscode/src/commands/utils/get-commands.ts
@@ -40,11 +40,15 @@ export function buildCodyCommandMap(
     // If it doesn't, use the root as the root
     const commands = parsed.commands ?? parsed
     for (const key in commands) {
-        const command = commands[key] as CodyCommand
+        const command = commands[key] as Partial<CodyCommand>
+        // Skip adding the command if it doesn't have a prompt
+        if (!command.prompt) {
+            continue
+        }
         command.type = type
         command.slashCommand = toSlashCommand(key)
         // Set default mode to ask unless it's an edit command
-        command.mode = command.mode ?? (command.prompt.startsWith('/edit ') ? 'edit' : 'ask')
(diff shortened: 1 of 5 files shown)


vscode/src/commands/utils/get-commands.ts
```diff
@@ -40,11 +40,15 @@ export function buildCodyCommandMap(
...
+        const command = commands[key] as Partial<CodyCommand>
+        // Skip adding the command if it doesn't have a prompt
+        if (!command.prompt) {
+            continue
+        }
         command.type = type
         command.slashCommand = toSlashCommand(key)
         // Set default mode to ask unless it's an edit command
-        command.mode = command.mode ?? (command.prompt.startsWith('/edit ') ? 'edit' : 'ask')
+        command.mode = command.mode ?? (command.prompt?.startsWith('/edit ') ? 'edit' : 'ask')
```
umpox, line 51:
```suggestion
        command.mode = command.mode ?? (command.prompt.startsWith('/edit ') ? 'edit' : 'ask')
```

nit: TS should be able to understand this can't be undefined

umpox (approved):

jdorfman (approved):

PR #3013, opened 2024-02-02
fix: ignore non file in status bar

abeatrix:
fix: check scheme before ignoring file in status bar

The status bar logic was ignoring files without checking the URI scheme first. This fixes it to only ignore files with the 'file' scheme, preventing incorrect ignoring of other URI types.

Note: this is only available behind the unstable feature flag.

## Test plan

in your editor, create an untitled file

before: the cody icon will light up as we mark all the non-files as ignored

image

after: cody status bar does not light up

image

### DIFF
--- a/vscode/src/services/StatusBar.ts
+++ b/vscode/src/services/StatusBar.ts
@@ -209,6 +209,9 @@ export function createStatusBar(): CodyStatusBar {
     // If ignored, adds 'Ignored' to the status bar text.
     // Otherwise, rerenders the status bar.
     const verifyActiveEditor = (uri?: vscode.Uri) => {
+        // NOTE: Non-file URIs are not supported by the .cody/ignore files and
+        // are ignored by default. As they are files that a user would not expect to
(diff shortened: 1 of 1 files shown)


vscode/src/services/StatusBar.ts
```diff
@@ -209,7 +209,7 @@ export function createStatusBar(): CodyStatusBar {
     // If ignored, adds 'Ignored' to the status bar text.
     // Otherwise, rerenders the status bar.
     const verifyActiveEditor = (uri?: vscode.Uri) => {
-        if (uri && isCodyIgnoredFile(uri)) {
+        if (uri?.scheme === 'file' && isCodyIgnoredFile(uri)) {
```
DanTup, line 212: Maybe worth a comment explaining why we treat this slightly differently?

I presume the goal is that although we do still ignore these files, we don't want them to seem like a "warning" because they're files a user wouldn't _expect_ to be used by Cody anyway?

DanTup (approved):

Issue #3003, opened 2024-02-01
Add integration test to validate .cody/ignore in a multi-root workspace


abeatrix:
Users should be able to use .cody/ignore to filter workspace files based on the .cody/ignore files located inside their current workspace.

PR #3014, opened 2024-02-02
Add support for integration tests using a multi-root workspace

DanTup:
This moves the existing integration tests into a `single-root` folder and adds a `multi-root` folder. Because we can't change between single/multi-root while tests are running (it causes an extension host restart), these are separate invocations (`vscode/test/integration/main.ts` was updated to have a set of test folders / workspaces).

This is work towards https://github.com/sourcegraph/cody/issues/3003, which is to have integration tests that open a multi-root workspace and ensure ignores are handled correctly for each (I have some basic tests for this but they're failing and need some debugging).

## Test plan

- Check bot results
- Ensure results include both:
  - the original single-root integration tests (including the new "Single-root Workspace was correctly loaded" test) [result](https://github.com/sourcegraph/cody/actions/runs/7756639403/job/21154351783?pr=3014#step:8:392)
  - the new multi-root integration tests (currently just a single "Multi-root Workspace was correctly loaded" test) [result](https://github.com/sourcegraph/cody/actions/runs/7756639403/job/21154351783?pr=3014#step:8:415)

### DIFF
--- a/vscode/test/integration/main.ts
+++ b/vscode/test/integration/main.ts
@@ -16,36 +16,38 @@ async function main(): Promise<void> {
     // __dirname is derived from that path, not this file's source path.
     const vscodeCodyRoot = path.resolve(__dirname, '..', '..', '..', '..')
 
-    // The test workspace is not copied to out/ during the TypeScript build, so we need to refer to
-    // it in the src/ dir.
-    const testWorkspacePath = path.resolve(vscodeCodyRoot, 'test', 'fixtures', 'workspace')
-
     // The directory containing the extension's package.json, passed to --extensionDevelopmentPath.
     const extensionDevelopmentPath = vscodeCodyRoot
 
-    // The path to the test runner script, passed to --extensionTestsPath.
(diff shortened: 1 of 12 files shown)


vscode/test/integration/main.ts
```diff
@@ -16,36 +16,38 @@ async function main(): Promise<void> {
...
-                launchArgs: [
-                    testWorkspacePath,
-                    '--disable-extensions', // disable other extensions
-                ],
-            })
-        )
+        await MockServer.run(async () => {
+            for (const testConfig of testConfigs) {
+                await runTests({
+                    version: '1.81.1',
```
umpox, line 39: Unrelated to this PR, but we should look at if we can upgrade this, `1.81.1` is quite far behind 🤔

umpox (approved):

vscode/test/integration/main.ts
DanTup: Yeah, I've asked about this on Slack :)

Thanks!

Issue #2970, opened 2024-01-31
bug: errors when using "ask cody to explain" an error


vadi2:
### Version

1.2

### Describe the bug

Using the "ask cody to explain" an error always gives this message:

> Cannot read properties of undefined (reading 'match')

### Expected behavior

An explanation of the error.

cbjerg:
Same error here. Also "Ask Cody to fix" gives me "retrieved codebase context before initialization"

umpox:
Thanks for reporting!

PR to fix: https://github.com/sourcegraph/cody/pull/3015

@cbjerg That error should be fixed in the latest release, if not please LMK and I'll take another look

PR #3015, opened 2024-02-02
Chat: Fix "Ask Cody to Explain"

umpox:
## Description

- Pass the correct arguments to the `cody.action.chat` command
- Add type safety to avoid this mistake again

closes https://github.com/sourcegraph/cody/issues/2970

## Test plan

1. Try "Ask Cody to Explain"
2. Check there is no error


vadi2:
Is there an easy way to test this?

umpox:
@vadi2 If you want to run this locally, you can checkout the branch locally and follow these docs: https://github.com/sourcegraph/cody/blob/main/doc/dev/index.md

vadi2:
> % asdf install
> Install plugins first to be able to install tools

Which plugins should I install?

umpox:
@vadi2 Try this quickstart from the repo root:
`pnpm install && pnpm build && cd vscode && pnpm run dev`

That should work if you have NodeJS installed.

I've added an e2e test for this too

vadi2:
No luck testing it - same issue in the 2nd vscode instance that pops up when I run the `Launch VS Code Extension (Desktop)` task, but it could be that I'm doing something wrong.

abeatrix (approved):
Confirmed the Ask Cody to Explain works for me:
image


Also updated the failing e2e test that hopefully helps!

umpox:
@vadi2 Might have been an issue with our local flow. We need to do a check on the instructions for first clones.

I've merged this now so will ensure this is included in the next release

PR #3019, opened 2024-02-02
Clarify asdf setup instructions

vadi2:
You need to have at least one asdf plugin installed for it to give you a meaningful error message on which plugins are missing. Otherwise, it just says "Install plugins first to be able to install tools" which is not actionable.

### DIFF
--- a/doc/dev/index.md
+++ b/doc/dev/index.md
@@ -3,6 +3,7 @@
 ### Setup
 
 1. Install [asdf](https://asdf-vm.com/)
+2. Install pnpm using asdf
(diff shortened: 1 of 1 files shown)


doc/dev/index.md
```diff
@@ -3,6 +3,7 @@
 ### Setup
 
 1. Install [asdf](https://asdf-vm.com/)
+2. Install pnpm using asdf
```
dominiccooney, line 6: Let's provide the specific command to run, like the subsequent steps do, so people don't have to hunt for them.

Let's use 1. for the numbered list, in keeping with the surrounding style.

dominiccooney (changes requested):

PR #3020, opened 2024-02-02
Add integration tests for Ignores in multi-root VS Code workspaces

DanTup:
This adds some integration tests that in an active multi-root workspace in VS Code, the ignore files are set up and applied correctly.

Also includes some tweaks to the shim `findFiles()` to support `RelativePattern` which fixes an issue where we could add ignore rules from the "wrong" workspace folders (not in a breaking way, but it was confusing to debug and slower because we ran a search across all workspace folders for each workspace folder).

Fixes https://github.com/sourcegraph/cody/issues/3003

## Test plan

- Check bot results for the new tests, or:
- Open the `multi-root.code-workspace` workspace from `fixtures` and verify that the `.ws1` is ignored only in workspace1 and that `.ws2` file is ignored only in workspace2.

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -195,9 +195,27 @@ const _workspace: typeof vscode.workspace = {
     workspaceFile: undefined,
     registerTaskProvider: () => emptyDisposable,
     async findFiles(include, exclude, maxResults, token) {
-        if (typeof include !== 'string') {
-            throw new TypeError('workspaces.findFiles: include must be a string')
+        let searchFolders: vscode.WorkspaceFolder[]
+        let searchPattern: string
+
+        if (typeof include === 'string') {
+            searchFolders = workspaceFolders
+            searchPattern = include
+        } else {
+            const matchingWorkspaceFolder = workspaceFolders.find(
+                wf => wf.uri.toString() === include.baseUri.toString()
+            )
+            if (!matchingWorkspaceFolder) {
+                throw new TypeError(
+                    `workspaces.findFiles: RelativePattern must use a known WorkspaceFolder\n  Got: ${
+                        include.baseUri
+                    }\n  Known:\n${workspaceFolders.map(wf => `  - ${wf.uri.toString()}\n`).join()}`
+                )
+            }
+            searchFolders = [matchingWorkspaceFolder]
+            searchPattern = include.pattern
         }
+
         if (exclude !== undefined && typeof exclude !== 'string') {
             throw new TypeError('workspaces.findFiles: exclude must be a string')
         }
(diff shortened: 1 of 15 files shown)


DanTup:
Bots are green now - had to make some tweaks to the shim `findFiles()` to support `RelativePath` (a pattern + workspace folder) to fix a quirk (and performance issue) with searching all workspace folders for each workspace folder. (fyi @abeatrix @olafurpg)

agent/src/vscode-shim.ts
```diff
@@ -195,9 +195,27 @@ const _workspace: typeof vscode.workspace = {
     workspaceFile: undefined,
     registerTaskProvider: () => emptyDisposable,
     async findFiles(include, exclude, maxResults, token) {
-        if (typeof include !== 'string') {
-            throw new TypeError('workspaces.findFiles: include must be a string')
+        let searchFolders: vscode.WorkspaceFolder[]
```
abeatrix, line 198: doesn't have to be in this PR, but we probably should add a test to the agent to cover this. wdyt?

vscode/test/fixtures/multi-root.code-workspace
```diff
@@ -1,10 +1,19 @@
...
-		{
-			"path": "workspace"
-		},
-		{
-			"path": "workspace2"
-		}
-	]
-}
\ No newline at end of file
+    "folders": [
```
abeatrix, line 2: Will this also be used by the e2e tests?

abeatrix (approved):
left some questions inline but other than that, LGTM 🚀

agent/src/vscode-shim.ts
DanTup: I've added a test that `RelativePattern()` is supported, but it doesn't seem like the tests are setup to have multiple workspace folders (there's a single `workspaceDocuments` with a single URI), so I added a TODO for that.

vscode/test/fixtures/multi-root.code-workspace
DanTup: It's not currently, but there's no reason it can't be if useful :-)

(part of me did wonder if we should just make all the tests always multi-root instead of having both.. but maybe there are some differences worth testing both? 🤔)

agent/src/vscode-shim.ts
olafurpg: The agent protocol currently only supports one workspace root so I haven't prioritized this. Thank you for extending the solution to support `RelativePath` 🙇🏻  

I'm doing my best to implement VS Code APIs thoroughly but the full scope of this work is realistically beyond what I can do alone. The more people we have on the team who feel comfortable implementing new APIs the better. I think this is our best chance at getting robust cross-IDE support and getting a robust testing story for VSC.

PR #3021, opened 2024-02-02
Always uses latest stable VS Code version for integration tests

DanTup:
This was previously hard-coded and needed updating (and confusingly in two places, with one being bumped without the other in the past). This just tracks latest stable.

If we want to also run tests on the oldest supported version (if this is defined), we could split this out into an env VAR and have them both (`['stable', 'x.y.z']`) in the GH Actions yaml file.

## Test plan

Non-functional... ensure bots are still running and the logs show that they used the current latest version of VS Code (1.86).


DanTup:
Seems like e2e tests are all timing out on the bots with the latest version, needs some debugging.

DanTup:
image

It's not often that Windows is the green bot 😄

abeatrix (approved):
thanks for updating this! I'm happy with the switch, but want to check with @philipp-spiess @umpox @valerybugakov to see what they think since i don't know much about setting up our CI :)

vscode/test/e2e/install-deps.ts
```diff
@@ -7,7 +7,12 @@ import {
     type ProgressReport,
 } from '@vscode/test-electron'
 
-const vscodeVersion = '1.85.1'
+// The VS Code version to use for e2e tests (there is also a version in ../integration/main.ts used for integration tests).
+//
+// We set this to stable so that tests are always running on the version of VS Code users are likely to be using. This may
+// result in tests breaking after a VS Code release but it's better for them to be investigated than potential bugs being
+// missed because we're running on an older version than users.
+const vscodeVersion = 'stable'
```
valerybugakov, line 15: Love it!

valerybugakov (approved):

umpox (approved):
Will be v helpful for certain bleeding edge features too (like the commit message provider)

PR #3025, opened 2024-02-02
update: remove description requirement from custom command

abeatrix:
Remove the requirement to add a description for a custom command as the prompt of the command is descriptive enough.

This eliminates one unnecessary step to simplify the custom command creation process.

Based on screenshot shared by our users, it looks like users often use the name of the command to fill in the description field, e.g. https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1706661347779839?thread_ts=1706656724.966049&cid=C04MSD3DP5L, https://discord.com/channels/969688426372825169/1202562648340701257/1202562648340701257, https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1703011617179849

Feedback from a power user:

> One piece of feedback when defining your custom commands is to get rid of the command description step.

## Test plan

We will still show the description in the command menu when provided, but it is no longer a required field when creating a command from the command palette. You can try to create a custom command with and without the "description" field to confirm.

I've updated the current e2e tests and unit tests for custom commands to reflect this change.


vscode/src/commands/services/custom-commands.ts
```diff
@@ -218,7 +219,22 @@ export class CustomCommandsManager implements vscode.Disposable {
...
+                    void vscode.workspace.fs.delete(uri)
+                    return
+                }
+                // Show confirmation for user setting only, as
+                // workspace config can be reverted with git history.
+                vscode.window
+                    .showInformationMessage(
+                        'Remove the Cody Custom Command file from your User setting?',
+                        'Confirm'
+                    )
```
toolmantim, line 232: Let's just show the confirmation consistently for both, as even if it's in Git history I think deletion deserves a confirmation.

For the confirmation, it's always good to use the action itself as button label, rather than the generic "Confirm" text, i.e. "Remove Command"

For the message can we make it like the following (matching the path labels we use in the Configure Custom Commands quickpick too):
* "Remove custom command 'My Thing' from User Settings (~/.vscode/cody.json)?"
* "Remove custom command 'My Thing' from Workspace Settings (.vscode/cody.json)?"

CleanShot 2024-02-07 at 15 26 17@2x

So in the end something like:

```ts
vscode.window
    .showInformationMessage(
        'Remove custom command \'My Thing\' from User Settings (~/.vscode/cody.json)?',
        'Remove Command'
    )
```

toolmantim (approved):

vscode/src/commands/services/custom-commands.ts
abeatrix: @toolmantim Oh this is not removing a single command, but the whole file. So something like this?

```
vscode.window
    .showInformationMessage(
        'Remove user settings file (~/.vscode/cody.json)?',
        'Remove File'
    )
```

abeatrix:
Will update the e2e test tmr

vscode/src/commands/services/custom-commands.ts
toolmantim: Oh right! I forgot.

In that case let's call it Delete (same as right clicking on a file in the VS Code file explorer tree).

They then show this:



We should probably add `{ useTrash: true }` to the `vscode.workspace.fs.delete` call too, and do something similar:

```ts
vscode.window
    .showInformationMessage(
        'Are you sure you want to delete your Cody user settings file \'~/.vscode/cody.json\'?',
        { detail: 'You can restore this file from the Trash.' }
        'Move to Trash'
    )
```

abeatrix: Updated to use modal:

image

PR #3027, opened 2024-02-03
exclude build/ and .class files from user context files

beyang:
## Test plan

- [ ] Run locally and try to include a .class file (e.g., in `sourcegraph/jetbrains`)
- [ ] Run locally and try to include a regular source file (should still succeed)


arafatkatze (approved):
Stamp

PR #3032, opened 2024-02-05
Autocomplete: truncate the last completion line if it matches suffix

valerybugakov:
## Context 

- If the last completion line starts the next-non-empty-suffix line, we didn't truncate it, which caused completions to be suggested with redundant closing brackets:

    CleanShot 2024-01-19 at 12 14 34@2x

- This PR fixes the issues and adds a unit test to cover this case.
- [Slack report](https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1705662932035059)

## Test plan

CI


philipp-spiess (approved):
Nice!

PR #2957, opened 2024-01-30
cody: change auth endpoint to handle SAMS redirect

kalanchan:
Part of the accounts migration from sourcegraph.com to accounts.sourcegraph.com (SAMS). Using this new auth url that enables us to redirect to SAMS and back to dotcom. 

I wasn't able to test the GitLab login properly as my account is too new, could someone help with this? 

For more context, see [Joe's explanation](https://sourcegraph.slack.com/archives/C05AGQYD528/p1706593310231579?thread_ts=1706585008.158329&cid=C05AGQYD528)

## Test plan
Pull the branch, **make sure you are logged out of sourcegraph.com AND accounts.sourcegraph.com**. Head to the editor, log out of Cody, confirm that each login has the right url and try logging in.

See each scenario below:

GitHub

https://github.com/sourcegraph/cody/assets/51868853/6390ac11-d228-4c2c-8960-edc85d7ef558

GitLab

https://github.com/sourcegraph/cody/assets/51868853/fb18441e-fb42-463d-9f31-439a1d48a36c

Google

https://github.com/sourcegraph/cody/assets/51868853/25af627c-717f-49ef-abd6-cade32ab1489

PR #3037, opened 2024-02-05
cody: update auth endpoint for SAMS redirect

unknwon:
This is a follow-up improvement to https://github.com/sourcegraph/cody/pull/2957. TIL we can configure auth provider in Sourcegraph to have a static `pc` to reference to. This is strictly better than relying on the brittle sha256 of current provider config.

## Test plan

Manually tested

and

CleanShot 2024-02-05 at 14 55 27@2x


kalanchan (approved):
nice this will help make sure users end up at the right spot to login

unknwon:
Merging as-is, happy to address post-merge feedback!

PR #3039, opened 2024-02-05
Add config feature gating for chat and commands

abeatrix:
This commit adds a check for the chat and commands config features before executing the chat command. If either feature is disabled by the site admin, an error message is shown and the command is not executed. This gates usage of the chat and commands behind the config features.

Also updated behavior where running a command in an ignored file will returns an error.

## Test plan

connect to an instance that has chat or command disabled, and check if the error message shows up correctly.

### DIFF
--- a/vscode/src/commands/execute/ask.ts
+++ b/vscode/src/commands/execute/ask.ts
@@ -1,7 +1,8 @@
 import * as vscode from 'vscode'
 import type { ChatSession } from '../../chat/chat-view/SimpleChatPanelProvider'
 import type { WebviewSubmitMessage } from '../../chat/protocol'
-import type { ChatEventSource } from '@sourcegraph/cody-shared'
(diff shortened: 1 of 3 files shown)


vscode/src/commands/execute/ask.ts
```diff
@@ -11,5 +12,14 @@ export interface ExecuteChatArguments extends WebviewSubmitMessage {
  * Wrapper around the `cody.action.chat` command that can be used anywhere but with better type-safety.
  */
 export const executeChat = async (args: ExecuteChatArguments): Promise<ChatSession | undefined> => {
```
abeatrix, line 14: this function is used by all default commands so this will be the best place to check

arafatkatze (approved):
Thanks so much, saves me sooooo much time

🙇

arafatkatze:
Tested all the cases of this
- All the commands(non custom) with both siteadmin enabled/disabled
- All the commands(non custom) with cody ignored file

PR #3047, opened 2024-02-06
Edit: Always expand range to include all non-whitespace chars

umpox:
## Description:

This PR:
- Always expands the Edit selection to include all non-whitespace characters from the selected lines.

This is a change that will help ensure users don't accidentally miss characters from their selection. Edit will only act on the selected range, so this reduces the likelihood of bad output.

There is some use case for inner-line ranges, but it's unlikely to be of a benefit for the user. The value of doing this outweighs that feature.

## Test plan

1. Create a selection that doesn't contains all chars from a full line
2. Trigger an edit
3. Check that the edit acts on the full line

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -10965,6 +10965,1032 @@ log:
         send: 0
         ssl: -1
         wait: 0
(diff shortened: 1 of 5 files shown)


umpox:
Noticed an issue with empty ranges, fixing

abeatrix (approved):

toolmantim (approved):

PR #3049, opened 2024-02-06
Autocomplete: Add option to disable inside code comments

### LINKED ISSUE #3038: feedback: Turn off autompletion for comments
### Version

1.3.1707145731

### Areas for Improvement

- [X] UI/UX
- [ ] Onboarding
- [ ] Docs
- [ ] Chat
- [ ] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience.

It's rare that Cody can suss a useful and thoughtful comment from the code itself--the best code comments reveal new information such as a "why" or "historical context" that is not present. I'd like to keep autocomplete enabled for code, but disable it when in the context of creating a comment.

### Describe the solution you'd like to see

I'd like a config that allows me to disable autocompletion within the context of creating a new comment. So if I've typed `// (<- cursor is here)` it should not make any comment suggestions.

### Describe any alternatives that could be considered

_No response_

### Additional context

_No response_

philipp-spiess:
Closes #3038

Adds a new user config option to prevent completions inside code comments.

@valerybugakov I didn't get the test to run since the intent detection is never run because the parser cache is always empty even though I ran `initTreeSitterParser()`. Do you know how we can get this enabled in unit tests? I would expect a call to `provideInlineCompletions` to preload the parser cache if we have called `initTreeSitterParser` before but I don't think this happens right now. Perhaps because we query it before the stuff inside `get-inline-completions` but we do call into tree sitter before that so I think this should work. 😄  

## Test plan

https://github.com/sourcegraph/cody/assets/458591/6f9abc60-a26d-4140-8188-7897f87f55fe


philipp-spiess:
Ah we do it manually in tests via [this](https://github.com/sourcegraph/cody/blob/839be374b5765b806816aeff1b6e04c8c0304954/vscode/src/completions/get-inline-completions-tests/helpers.ts#L126) I seeeee

valerybugakov (approved):
Fabulous!

sjurgis:
Can we get this for jetbrains IDEs?

sjurgis:
@philipp-spiess any chance you can point me how to use this setting in Jetbrains IDEs?

philipp-spiess:
@sjurgis I'll defer to @pkukielka but chances are we don't have the setting there yet.

pkukielka:
We do not support it yet but it should be a quick job to add it.
@taylorsperry @kalanchan can prioritise it

sjurgis:
@taylorsperry @kalanchan that would be a nice differentiator as no other assistant supports this...

sjurgis:
I've stumbled the settings are actually there in Jetbrains. Tried the flag and it seems to be working. Hoooraay!

Screenshot 2025-01-16 at 2 06 02 PM

PR #3051, opened 2024-02-06
Add set model endpoint to agent protocol

Sa1to:
It is part of https://github.com/sourcegraph/jetbrains/issues/221 
We need to add possibility to set specific llm model per chat.
## Test plan
Test api from JetBrains plugin

### DIFF
--- a/vscode/src/jsonrpc/agent-protocol.ts
+++ b/vscode/src/jsonrpc/agent-protocol.ts
@@ -49,6 +49,7 @@ export type Requests = {
     'chat/restore': [{ modelID: string; messages: ChatMessage[]; chatID: string }, string]
 
     'chat/models': [{ id: string }, { models: ModelProvider[] }]
+    'chat/setModel': [{id: string; model: string}, boolean]
     'chat/remoteRepos': [{ id: string }, { remoteRepos?: Repo[] }]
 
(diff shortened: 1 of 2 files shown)


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -49,6 +49,7 @@ export type Requests = {
     'chat/restore': [{ modelID: string; messages: ChatMessage[]; chatID: string }, string]
 
     'chat/models': [{ id: string }, { models: ModelProvider[] }]
+    'chat/setModel': [{id: string; model: string}, boolean]
```
mkondratek, line 52: why is it boolean? you can make it "void" (just use `null`)

Sa1to: I just wanted to make sure it was executed

olafurpg: This request doesn't need to exist on the agent protocol when you can already achieve the same behavior with `webview/receiveMessage`. We exceptionally added higher-level abstractions for `chat/submitMessage` because it required fairly complicated logic to create a promise that resolves when the last reply is received
```ts
        await this.request('webview/receiveMessage', {
            id,
            message: { command: 'chatModel', model },
        })
```

PR #3055, opened 2024-02-06
Add test case to reproduce crash that only happens with Enterprise accounts

olafurpg:
When triggering the "Document code" command in JetBrains with an Enterprise account, we hit on the following error
```
InternalError: Cannot read properties of undefined (reading 'model')

TypeError: Cannot read properties of undefined (reading 'model')
    at getModel (/Users/olafurpg/dev/sourcegraph/cody/vscode/src/models/index.ts:24:26)
    at Object.get (/Users/olafurpg/dev/sourcegraph/cody/vscode/src/models/index.ts:47:13)
    at EditManager.executeEdit (/Users/olafurpg/dev/sourcegraph/cody/vscode/src/edit/manager.ts:102:47)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5): {
  "error": {},
  "stack": "\nTypeError: Cannot read properties of undefined (reading 'model')\n    at getModel (/Users/olafurpg/dev/sourcegraph/cody/vscode/src/models/index.ts:24:26)\n    at Object.get (/Users/olafurpg/dev/sourcegraph/cody/vscode/src/models/index.ts:47:13)\n    at EditManager.executeEdit (/Users/olafurpg/dev/sourcegraph/cody/vscode/src/edit/manager.ts:102:47)\n    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)"
```

## Test plan
See added test case.

PR #3058, opened 2024-02-06
fix: sync edit models on AuthStatus change

abeatrix:
CLOSE https://github.com/sourcegraph/jetbrains/issues/511
CONTEXT: https://sourcegraph.slack.com/archives/C05B7C6FBPX/p1707210738623389

Sync the edit models used by the edit manager with the latest authentication status rather than just initializing them once on startup. This allows edit models to be refreshed when auth config changes.

- Syncing the edit models in the edit manager when the auth status changes rather than just on init
- Getting the edit models based on the auth status rather than directly from the auth provider

## Test plan

- All current tests should continue passing.
- Edit should work.

Confirmed this change work with https://github.com/sourcegraph/cody/pull/3055:

image

### DIFF
--- a/vscode/src/edit/input/get-input.ts
+++ b/vscode/src/edit/input/get-input.ts
@@ -79,7 +79,7 @@ export const getInput = async (
 
     const authStatus = authProvider.getAuthStatus()
     const isCodyPro = !authStatus.userCanUpgrade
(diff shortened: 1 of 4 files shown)


umpox (approved):
Thanks!

Issue #1730, opened 2023-11-14
Remove Slash from Slash Command


abeatrix:
Remove Slash from Slash Command

Requested by @toolmantim, will wait for detail

iswarmondal:
I cannot understand the issue, please help me with a bit more info 🙂

iswarmondal:
<img width="562" alt="Screenshot 2023-11-15 at 7 04 54 PM" src="https://github.com/sourcegraph/cody/assets/73457704/0391ed37-e233-454c-a73a-e33397b65bf2">

#
I found these two slashes in the new chat panel.
Are we talking about these @abeatrix ?

abeatrix:
@iswarmondal sorry for the late reply and thanks for your interest in working on this issue. 
I don't have enough information on how to proceed this atm, will close this issue for now. If there is something else you're interested in feel free to let me know :)

abeatrix:
Closing issue as discussed with Tim in this slack thread: https://sourcegraph.slack.com/archives/C05AGQYD528/p1700173517887689

PR #3061, opened 2024-02-06
Command: remove slashes from commands

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/1730

Remove slashes from command names and implemented design from https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1550-62492&mode=design&t=qdjbLWsl9qHq578p-4

Main Commands Menu

image

Custom Commands Menu

image

image

image

## Test plan

Updated all current tests.

https://github.com/sourcegraph/cody/assets/68532117/581ef1e9-c0bc-461d-83ae-ffc03098d2fc

### Before 

image

image


vscode/src/commands/CommandsController.ts
```diff
@@ -34,7 +35,7 @@ class CommandsController implements vscode.Disposable {
         const commandSplit = text?.trim().split(' ')
         // The unique key for the command. e.g. /test
         const commandKey = commandSplit?.shift() || text
-        const command = this.provider?.get(commandKey)
+        const command = this.provider?.get(fromSlashCommand(commandKey))
```
valerybugakov, line 38: Is this required for backward compat?

vscode/src/commands/menus/index.ts
```diff
@@ -25,7 +26,7 @@ export async function showCommandMenu(
         if (type === 'default') {
             items.push(CommandMenuSeperator.commands)
             for (const [_name, _command] of vscodeDefaultCommands) {
-                const label = _command.slashCommand
+                const label = fromSlashCommand(_command.slashCommand)
```
valerybugakov, line 29: Does it make sense to deprecate `CodyCommand.slashCommand` and introduce `CodyCommand.command` or smth like this to reflect the updated behavior better?

abeatrix: good point, i've deprecated slashCommand and replace it with key instead

toolmantim:
Design wise we want to have it match the commands treeview.

Could we update both the treeview and the command menu to match this?
https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1550-62492&mode=design&t=qdjbLWsl9qHq578p-4

abeatrix:
@toolmantim updated and ready for review

toolmantim (commented):
This is great!

A few small changes if we could:

* "New Chat" from the quickpick should act the same as "New Chat" in the review (open a new chat tab)
* Could we add the keyboard shortcuts for "New Chat" and "Edit Code" to the quickpick item descriptions?
* If there are no custom commands, can we make "Custom Commands" quickpick item & treeview item go directly to the "Configure Custom Commands…" and skip the empty-ish middle quickpick?

One bug:

* In the quickpick "Custom Commands" > "Configure Custom Commands…" both the "New Custom Command..." and "Open Custom Commands Documentation" items don't seem to work

abeatrix:
@toolmantim updated based on your suggestions and fixed the bug:
- [x] "New Chat" from the quickpick should act the same as "New Chat" in the review (open a new chat tab)
- [x] Could we add the keyboard shortcuts for "New Chat" and "Edit Code" to the quickpick item descriptions?
  - [x] Also added a keyboard shortcuts for "Custom Commands" as shown in your figma design
- [x] If there are no custom commands, can we make "Custom Commands" quickpick item & treeview item go directly to the "Configure Custom Commands…" and skip the empty-ish middle quickpick?
- [x] Fixed bug: In the quickpick "Custom Commands" > "Configure Custom Commands…" both the "New Custom Command..." and "Open Custom Commands Documentation" items don't seem to work

toolmantim (approved):
Nice! Two non-blocking things…

If you start typing "New" you end up seeing duplicates. I'd suggest we always exclude the "New Chat (shortcut)" and "Edit Code (shortcut)" items when you start filtering/typing… because they're always going to be there as "Start a new chat" and "Start a new edit" anyhow.

CleanShot 2024-02-13 at 16 18 16@2x

Also when you filter in custom commands we probably don't want to show these last 3 items, just the filtered custom commands (or nothing if no matches):

CleanShot 2024-02-13 at 16 18 40@2x

toolmantim:
Does this binding actually exist? I can't get it to trigger:

CleanShot 2024-02-13 at 16 24 19@2x

If it doesn't exist, we can just remove that shortcut label for now.

abeatrix:
> Does this binding actually exist? I can't get it to trigger:
> 
> CleanShot 2024-02-13 at 16 24 19@2x
> 
> If it doesn't exist, we can just remove that shortcut label for now.

it does exist! I just fixed this and the bugs you found. Will update the e2e tests tomorrow to cover these cases before I finalize the PR tomorrow, thanks for taking the time to review and test this PR out!

vscode/src/commands/index.ts
```diff
@@ -0,0 +1,82 @@
+import os from 'os'
+import type { CodySidebarTreeItem } from '../services/treeViewItems'
+
+const isMac = os.platform() === 'darwin'
```
valerybugakov, line 4: Nit: we have multiple helpers like this? Is it worth reusing the same implementation everywhere? 
Screenshot 2024-02-13 at 15 23 57

vscode/src/commands/menus/types.ts
```diff
@@ -20,8 +20,15 @@ export interface CommandMenuItem extends QuickPickItem {
     type?: CodyCommandType
     // vs code command, e.g. 'cody.commands.open.json'
     command?: string
-    // cody command, e.g. '/ask'
+    /**
+     * cody command, e.g. '/ask'
+     * @deprecated Use 'commandKey' instead.
+     */
     slashCommand?: string
```
valerybugakov, line 27: Nice! 🙌

valerybugakov (approved):
Looks great!

vscode/src/commands/index.ts
abeatrix: Good catch! I've move the one for node to lib/shared!

PR #3063, opened 2024-02-07
move logging for edits to provider

abeatrix:
Move all telemetry logging related to Cody fixup responses into the EditProvider instead of having it duplicated across Fixup Controller.

This consolidates the logging into the EditProvider that has better context for logging response telemetry events.

By centralizing this, the telemetry will be easier to manage and modify in the future instead of needing updates across multiple methods within the controller.

RE: https://github.com/sourcegraph/cody/issues/3065: Added logging for sidebar clicks. Here is the log from clicking on the edit button in the sidebar to start an edit job:

```
█ logEvent (telemetry disabled): CodyVSCodeExtension:command:edit:clicked VSCode {"properties":{"source":"sidebar"}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:command:edit:executed VSCode {"properties":{"source":"editor"},"opts":{"hasV2Event":true}}
█ logEvent (telemetry disabled): CodyVSCodeExtension:fixupResponse:hasCode VSCode {"properties":{"lineCount":11,"charCount":254,"source":"editor","hasV2Event":true}}
```

## Test plan

The current e2e test that logs edit events should work the same.


chenkc805 (approved):

vscode/src/edit/manager.ts
```diff
@@ -52,7 +52,7 @@ export class EditManager implements vscode.Disposable {
 
     public async executeEdit(
         args: ExecuteEditArguments = {},
-        source: ChatEventSource = 'editor'
+        source: ChatEventSource = 'menu'
```
umpox, line 55: I don't think we should default to `menu` here, I'm unsure that providing _any_ default is correct. The caller should ideally always provide this...

@abeatrix @chenkc805 WDYT about adding `unknown` as an option here. It would ideally never be used, but if it does it will be much clearer that we have a gap in our telemetry

umpox: I'll update the code

umpox: Ah actually, defaulting to `editor` is important because that's how the command is triggered from editor keyboard shortcuts. Hmm...

umpox (approved):

vscode/src/edit/manager.ts
umpox: Improved the types, confirmed that "menu" is logged where it is expected

PR #3066, opened 2024-02-07
Edit: Always set `models` on initialization

umpox:
## Description

Removed in: https://github.com/sourcegraph/cody/pull/3058

We still need to do this on init, as `syncAuthStatus` is only called if the auth status is actually changed. This change handles the case where the user is already authenticated and e.g. just reloads the editor.

## Test plan

1. Ensure authenticated
2. Reload the VS Code window
3. Try to trigger an edit

### DIFF
--- a/vscode/src/edit/manager.ts
+++ b/vscode/src/edit/manager.ts
@@ -40,6 +40,7 @@ export class EditManager implements vscode.Disposable {
     private models: ModelProvider[] = []
 
     constructor(public options: EditManagerOptions) {
+        this.models = getEditModelsForUser(options.authProvider.getAuthStatus())
         this.controller = new FixupController(options.authProvider)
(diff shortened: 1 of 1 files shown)


umpox:
Confirmed tests in https://github.com/sourcegraph/cody/pull/3055 still pass

olafurpg (approved):

abeatrix (approved):
On reload, wouldn't the AuthProvider start fresh and re-auth the user so that the AuthStatus should always get updated from its init status to the actual AuthStatus? 🤔 

QQ regarding the edit model (unrelated to this issue): since we hardcoded Claude-2.0 model for the test command, would it be an issue if the instance doesn't have that model available? (I haven't looked into the code to confirm if it's only used by dot com or not 😄

PR #3067, opened 2024-02-07
Edit: Disable command hints for unauthenticated users

umpox:
## Description

Only show command hints if the user is authenticated

## Test plan

1. Confirm you can see the ghost text when auth
2. Confirm you can see the ghost text when unauth
3. Confirm you can never see the ghost text when the setting is disabled


abeatrix (approved):
good catch!

PR #3070, opened 2024-02-07
Add CharactersLogger

philipp-spiess:
Closes https://github.com/sourcegraph/cody/issues/3403

This PR creates a new logger that logs an aggregates of how much characters were added/deleted in 5 minute intervals. We will use this to calculate a % metric on how much code is being generated by Cody. 

## Test plan

- Added unit tests
- Ensured that dispose is called and we `flush` even when VS Code is closed (nice!)

Screenshot 2024-03-14 at 15 20 24


vscode/src/completions/keystroke-tracker.ts
```diff
@@ -0,0 +1,33 @@
+import { isFileURI } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+
+export class KeystrokeTracker implements vscode.Disposable {
```
philipp-spiess, line 4: let's definitely change the name here to something more fitting lol

valerybugakov (commented):
Looks great! Should we ship it?

philipp-spiess:
@valerybugakov I'll do a cleanup and ship it, yep. I was starting this before but then we got the legal pushback.

vscode/src/services/CharactersLogger.ts
```diff
@@ -0,0 +1,61 @@
...
+        telemetryService.log(
+            `${logPrefix(extDetails.ide)}:characters`,
+            { insertedCharacters, deletedCharacters },
+            {
+                agent: true,
+                hasV2Event: true,
+            }
+        )
+        telemetryRecorder.recordEvent('cody', 'characters', {
+            metadata: { insertedCharacters, deletedCharacters },
```
akalia25, line 37: should we also include the `agent` property in v2 telemetry? We could do something like `agent: 1` to represent true and `agent: 0` for false? If this would be valuable to collect :)

vscode/src/main.ts
```diff
@@ -525,7 +526,8 @@ const register = async (
         // For debugging
         vscode.commands.registerCommand('cody.debug.export.logs', () => exportOutputLog(context.logUri)),
         vscode.commands.registerCommand('cody.debug.outputChannel', () => openCodyOutputChannel()),
-        vscode.commands.registerCommand('cody.debug.enable.all', () => enableDebugMode())
+        vscode.commands.registerCommand('cody.debug.enable.all', () => enableDebugMode()),
+        new CharactersLogger()
```
valerybugakov, line 530: Should we enable this logger only when autocomplete is enabled? We can probably filter out events from users with disabled autocomplete in Looker, but if we can avoid sending many redundant events, that's worth doing.

valerybugakov (approved):

vscode/src/main.ts
philipp-spiess: I explicitly enabled it for everyone who has cody enabled since we plan on combining this with all cody generative features (chat pastes/edit) in the future. If we only want to do this for people who do autocomplete we should do it in post. We'd need to apply the same filter that we do for the people who see completion suggested events anyway (i.e those who do not have other completion providers enabled etc.)

@Tobiryd @kelsey-brown do you foresee any issues with joining this data if we log it as a separate event?

vscode/src/services/CharactersLogger.ts
akalia25: ohh gotcha! I thought it was just additional metadata, It's my first time seeing this, so i'll have to look into the behavior a bit more before I can answer that ^, will keep you updated.

On that note, don't let this comment block the PR. 🚀🚀🚀

kelsey-brown:
@philipp-spiess yes we'll be able to filter down to a subset of users (e.g. just users who have seen an autocomplete suggestion) in Looker. insertedCharacters will include both Cody + user written code, right?

@akalia25 this event is probably a good candidate to exclude from the Amplitude pipeline, I can't imagine us doing this kind of analysis there.

vscode/src/services/CharactersLogger.ts
```diff
@@ -0,0 +1,79 @@
...
+        const extDetails = getExtensionDetails(getConfiguration(vscode.workspace.getConfiguration()))
+        telemetryService.log(
+            `${logPrefix(extDetails.ide)}:characters`,
+            { insertedCharacters, deletedCharacters },
+            {
+                agent: true,
+                hasV2Event: true,
+            }
+        )
+        telemetryRecorder.recordEvent('cody', 'characters', {
```
bobheadxi, line 38: Heads up that I think this should be named `'cody.characters', 'edited'` or something other than a root `cody` as the "feature" parameter, as we generally expect more context to be included here

bobheadxi: Also, I think we generally expect `cody.` as the prefix, and not the entire feature

philipp-spiess: https://github.com/sourcegraph/cody/pull/3730 TY

PR #3073, opened 2024-02-07
Edit: Always expand `doc` command

umpox:
## Description

We would not expand the `doc` intent if we first through we could override it to be `add`.

This just improves our `smartSelection` to always expand in this case

Before:

https://github.com/sourcegraph/cody/assets/9516420/26685fae-6afc-4fe8-8f51-29af0e478a19

After:

https://github.com/sourcegraph/cody/assets/9516420/a0917adb-f540-4cd1-93d9-67d122fe7432

## Test plan

1. Start a doc command by right clicking on an empty line inside a function
2. Check the range expands as much as it can

### DIFF
--- a/vscode/src/edit/input/get-input.ts
+++ b/vscode/src/edit/input/get-input.ts
@@ -17,12 +17,12 @@ import { getTestInputItems } from './get-items/test'
 import { executeEdit } from '../execute'
 import type { EditModelItem, EditRangeItem } from './get-items/types'
 import { CURSOR_RANGE_ITEM, EXPANDED_RANGE_ITEM, SELECTION_RANGE_ITEM } from './get-items/constants'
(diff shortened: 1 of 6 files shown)


umpox:
@abeatrix Interesting! I hadn't thought to try that 🤔 Hmm do you know if we've ever handled this? I see it's the same on in the current release but uses the `add` intent so streams in.

I feel like we should probably error in this case, as there's no expanded range to document? WDYT? I can update in a follow up

abeatrix:
>  I hadn't thought to try that 
Ohh I thought that's what you wanted me to try based on the Test Plan: Start a doc command by right clicking on an empty line? 😆

umpox:
@abeatrix I actually mean inside a function or something where it will expand, lol mb 🤦

abeatrix (approved):
LOL in that case this works as expected 👍

PR #3077, opened 2024-02-07
Autocomplete: Mark Ollama support experimental

philipp-spiess:
## Test plan

- Try with both `"cody.autocomplete.advanced.provider": "unstable-ollama"` and `"cody.autocomplete.advanced.provider": "experimental-ollama"`in the settings
- Check that it resolves to the correct provider by logging: 
    ```
    █ CodyCompletionProvider:initialized: experimental-ollama/deepseek-coder:6.7b-base-q4_K_M
    ```

Screenshot 2024-02-07 at 20 21 37


vscode/src/configuration.ts
```diff
@@ -50,13 +50,18 @@ export function getConfiguration(
...
 
-    let autocompleteAdvancedProvider: Configuration['autocompleteAdvancedProvider'] = config.get(
-        CONFIG_KEY.autocompleteAdvancedProvider,
-        null
-    )
-    // Handle the old `unstable-fireworks` option
-    if (autocompleteAdvancedProvider === 'unstable-fireworks') {
-        autocompleteAdvancedProvider = 'fireworks'
+    let autocompleteAdvancedProvider = config.get<
+        Configuration['autocompleteAdvancedProvider'] | 'unstable-ollama' | 'unstable-fireworks'
```
abeatrix, line 54: Can we set the default value to `experimental-ollama` as true if `unstable-ollama` is available, so that we don't have to check for both names in the code repeatedly?

abeatrix (approved):
Left a non blocker question inline. lgtm

vscode/src/configuration.ts
philipp-spiess: I'm not sure I follow but isn't this exactly happening here? only `experimental-ollama` is supported but here we handle `unstable-ollama` explicitly in just one area: https://github.com/sourcegraph/cody/pull/3077/files/89672b78997a6d8dbc58dd354afa8247fab8a4e8#diff-85c89382076aeb49c8645dd41e6422552fba7bbf25575701c1a9253027164a76R62-R63

😅

abeatrix: Oh i mean to replace the `null` in `(CONFIG_KEY.autocompleteAdvancedProvider, null)` with `experimental-ollama` if `unstable-ollama` is available so you don't have to check for both in `create-provider`? https://github.com/sourcegraph/cody/blob/89672b78997a6d8dbc58dd354afa8247fab8a4e8/vscode/src/completions/providers/create-provider.ts#L52-L53

abeatrix: oh my diff view got cut off, i saw what you meant now lol

philipp-spiess: Oh gotcha the null is actually or the case when no provider is set and needs to work too, independent of this change

PR #3079, opened 2024-02-07
bump symf version

beyang:
## Test plan

- [ ] Run Cody locally and ensure search and chat context (with symf) work.


dominiccooney (approved):
Is there a doc for the symf release process? cody-engine has a workflow triggered by a tag which produces the release but I did not notice anything like that for symf.

beyang:
Here is the release process doc: https://github.com/sourcegraph/symf-private/tree/main/sh#readme

PR #3081, opened 2024-02-08
Autocomplete: always log `onComplete` debug event

valerybugakov:
## Context

- The `generatorWithTimeout` did not finish the inner generator if the consumer stopped earlier. It caused the `onComplete` completion event never to be logged to the Cody output channel. This PR fixes it by introducing the `finally` clause that finishes the inner generator instance.
- I did not experience this issue until I turned off hot-streak completions locally.

## Test plan

Added unit test


dominiccooney (approved):
Very cleanly written tests. Nice simplification to `finally` log completion.

PR #3088, opened 2024-02-08
PLG: update end date

kalanchan:
update end date in the notifications. We will need to cut a patch to get this updated in the client before end of week

## Test plan
N/A


abeatrix (approved):
Is 21 the correct date?

PR #3089, opened 2024-02-08
Agent: Custom Commands support

abeatrix:
Add Custom Commands support to Agent.
- This includes getting the commands from the `.cody/commands.json` file.
  - vs code client should migrate from using `.vscode/cody.json` to `.cody/commands.json`, will do this in follow up
- Execute a custom command in either chat mode or inline edit mode.

Current supported context options for Custom Commands in agent covered by the agent tests:
- none: exclude all context
- current directory
- open tabs
- current file
- selection

## Test plan

CI should be green with the newly added tests for Custom Commands.


agent/src/index.test.ts
```diff
@@ -976,6 +977,170 @@ describe('Agent', () => {
         })
     })
 
+    describe('Custom Commands', () => {
+        // TODO (bee) fix test - only works when using .only else it returns all opened files
+        it.skip('commands/custom, chat command, open tabs as context', async () => {
```
abeatrix, line 982: @olafurpg do you have any idea why the tests are passing locally for me but not when running in CI? 
image

agent/src/index.test.ts
```diff
@@ -976,6 +977,185 @@ describe('Agent', () => {
...
+            await client.openFile(squirrelUri)
+
+            const id = await client.request('commands/custom', { key: '/countTabs' })
+            // id should be type of string for chat commands
+            expect(typeof id).toBe('string')
+            const lastMessage = await client.firstNonEmptyTranscript(id as string)
+            expect(trimEndOfLine(lastMessage.messages.at(-1)?.text ?? '')).toMatchInlineSnapshot(`
+              " Based on my memory, you have shared 7 file contexts with me:
+
+              1. src/example.test.ts
```
abeatrix, line 995: Is there a way to clear current chat history before running this command?
when running with `.only` this would produce the correct answer (2 files only), but it would use all the tests that have been opened before this test was run when i remove `.only`.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -68,6 +68,9 @@ export type Requests = {
     'commands/test': [null, string]
     'commands/smell': [null, string]
 
+    // Trigger custom commands that could be a chat-based command or an edit command.
+    'commands/custom': [{ key: string }, string | EditTask]
```
olafurpg, line 72: To simplify JSON serialization/deserialization across all clients, we avoid using union types like `string | EditTask`. While it's possible to decode this type in some languages like Kotlin, it requires custom code and is often awkward to work with. It's better to design the protocol in a way that has simple types.

olafurpg: We can support unions of structural types

```ts
type CustomCommandResult = ChatResult | EditResult
interface ChatResult {
  kind: 'chat'
  foo: number
...
}
interface EditResult {
  kind: 'edit'
  bar: number
...
}
```

In other languages, this pattern has a natural encoding in as data classes or structs
```kotlin
data class CustomCommandResult(
  kind: String,
  foo: Int,
  bar: Int
)
```

agent/src/index.test.ts
olafurpg: The tests are also passing for me locally. The CI failures are a sign that the LLM prompt is different when running in CI because there is no matching recording. If you look at the prompt in the HTTP recording file, you see there's an absolute path leaking in the prompt

CleanShot 2024-02-09 at 11 04 29@2x

I pushed a commit to use `displayPath`, which I hope fixes the problem.

agent/src/index.test.ts
olafurpg: We do a best-effort to "reset" the environment between tests to avoid this exact problem. You can see how we implement `testing/reset` here, which tries to fire VS Code events to mirror the equivalent of closing all file tabs

https://sourcegraph.com/github.com/sourcegraph/cody@0fcf8a998c1716e23622ee9e8ae3ec1522325509/-/blob/agent/src/agent.ts?L450

You might be able to fix this problem by finding where these files are "leaking" between tests, and add a `vscode.workspace.onDidCloseTextDocument` handler to remove those files from the context's internal state.

agent/src/index.test.ts
olafurpg: Looks like that helped fix the issue on macOS/Linux. The Windows tests are still failing because the ordering of the context files is not deterministic. It's more tricky to spot this, I copy pasted the HTTP request body in the error message and re-printed it with indentation inside a Scala repl (any repl or notebook works fine)

CleanShot 2024-02-09 at 11 22 43@2x

CleanShot 2024-02-09 at 11 23 43@2x
CleanShot 2024-02-09 at 11 23 17@2x

olafurpg: Doesn't look like sorting was enough, but I think the fix for Windows is something similar. Worst case, we can disable this test on Windows

olafurpg (commented):
Only blocking comment is removing the union type in the protocol. Not using "request changes" so you can merge this PR even when I'm offline

agent/src/index.test.ts
abeatrix: Thank you so much for looking into this and for the debugging tips! I'll see if I can fix the windows issue first 😄

abeatrix:
@olafurpg thank you so much for the debugging tips! I was able to locate the issue and sort the context files for chat messages and that seems to fix the issue we were seeing!

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -630,3 +633,7 @@ export interface CreateFilesParams {
 export interface RenameFilesParams {
     files: { oldUri: string; newUri: string }[]
 }
+
+export type CustomCommandResult = CustomChatCommandResult | CustomEditCommandResult
+export type CustomChatCommandResult = string
```
olafurpg, line 638: This needs to serialize as a JSON object to have a simple encoding in other languages.

olafurpg (changes requested):
Great that you were able to fix the windows issue. I would absolutely never have been able to figure that diff out myself 😅

vscode/src/jsonrpc/agent-protocol.ts
abeatrix: Ohhhh so it has to be interface? Will update it!

olafurpg (approved):
Stamp to unblock merge after protocol moves to interface

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -630,3 +633,13 @@ export interface CreateFilesParams {
 export interface RenameFilesParams {
     files: { oldUri: string; newUri: string }[]
 }
+
+export type CustomCommandResult = CustomChatCommandResult | CustomEditCommandResult
+export interface CustomChatCommandResult {
+    type: 'chat'
+    result: string
```
olafurpg, line 640: One more blocking comment: using different types for the `result` property is also problematic. It's better to use unique property names for every unique type
```suggestion
    panelID: string
```

BTW, I am working on a code generation tool that automatically catches these errors and will fail CI with an error message on how to fix the problem

olafurpg: I'm not sure if `panelID` is the best property name, just added a suggestion to give an example of a diff that would resolve the issue. We encode discriminated unions as flat data classes in other languages. For example:

```
data class CustomCommandResult(
    type: String,
    chatResult: String? = null,
    editResult: EditTask? = null,
)
```

PR #3090, opened 2024-02-08
Guardrails is enabled just via instance configuration

cbart:
Closes sourcegraph/sourcegraph#59720

This pull requests removes the requirement for experimental config to enable guardrails.

The feature is just hidden behind [a sourcegraph instance config](https://sourcegraph.com/github.com/sourcegraph/cody@30620eb96a4e112611abe1d864b72a3a4c6ad1c3/-/blob/lib/shared/src/sourcegraph-api/graphql/client.ts?L86:5-86:16).

## Test plan

- Manual check - works with site config changes

### DIFF
--- a/vscode/webviews/App.tsx
+++ b/vscode/webviews/App.tsx
@@ -31,9 +31,7 @@ import { createWebviewTelemetryService } from './utils/telemetry'
 import type { VSCodeWrapper } from './utils/VSCodeApi'
 
 export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vscodeAPI }) => {
-    const [config, setConfig] = useState<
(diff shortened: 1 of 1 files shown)


arafatkatze (approved):
I approve of this PR on testing it and it makes sense.

Non blocking side question: its a bit confusing that attribution enablement in the siteadmin config is not like

```
  "configFeatures": {
    "chat": true
    "attribution": true
  },
```

but instead it is  ` "attribution.enabled": true,`

Which is also fine but creates some confusion in the first time user. My understanding is that @slimsag had a broader reasoning for this but I can't seem to recall having read it anywhere?

Can I get an **attribution** that links to that reasoning or a slack message?

cbart:
@arafatkatze take a look here: https://github.com/sourcegraph/sourcegraph/pull/59624#discussion_r1454020131

PR #3091, opened 2024-02-08
feat: support at-mentions context on edits

abeatrix:
edit-at

This commit adds the functionality to include the old context files from the transcript to the chat context files map in the Chat component. It creates a new map called contextFilesMap and populates it with the old context files. This allows for easy access and reference to the context files used in edit mode. 

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/a9683af7-dca7-43f6-bc0c-80d7d31bfef1


philipp-spiess (approved):
Niceeee! Don't forget to add a message to the change log

PR #3093, opened 2024-02-09
vscode-shim: fix workspace.findFiles

beyang:
- fixes exclusion pattern to properly exclude contents of node_modules and .git directories
- do not descend into excluded directories
- ignore .* files

## Test plan

- Added unit tests
- Test locally with IntelliJ plugin


vscode/src/editor/utils/editor-context.ts
```diff
@@ -25,7 +25,7 @@ const findWorkspaceFiles = async (
     // Object.keys(vscode.workspace.getConfiguration().get('search.exclude',
     // {}))
     const fileExcludesPattern =
-        '**/{*.env,.git,.class,out/,dist/,build/,snap,node_modules,__pycache__}**'
+        '**/{*.env,.git/,.class,out/,dist/,build/,snap,node_modules/,__pycache__/,bin/}**'
```
DanTup, line 28: `bin` was removed in https://github.com/sourcegraph/cody/pull/2472 because it's not universally a folder that should be excluded (Dart projects often have a `bin` folder that contains the main executable scripts).

agent/src/cli/evaluate-autocomplete/matchesGlobPatterns.ts
```diff
@@ -13,6 +13,8 @@ export function matchesGlobPatterns(
         return false
     }
 
-    const matchingExcludePatttern = excludeGlobs.find(excludePattern => minimatch(value, excludePattern))
+    const matchingExcludePatttern = excludeGlobs.find(excludePattern =>
+        minimatch(value, excludePattern, { dot: true })
```
DanTup, line 17: Does excluding dot folders prevent us from finding `.cody/ignore` (if it's the include pattern)? If it doesn't (and we can still find it), maybe we should add a test to verify/ensure this.

olafurpg (approved):
Thank you @beyang! This is great 🙇🏻

vscode/src/editor/utils/editor-context.ts
olafurpg: I missed that this PR would auto-merge. Opened a PR to address this comment https://github.com/sourcegraph/cody/pull/3094

agent/src/cli/evaluate-autocomplete/matchesGlobPatterns.ts
olafurpg: Good catch! `.cody/ignore` is being ignored here. I opened a PR to reproduce this. Might be worth removing `{ dot: true}` https://github.com/sourcegraph/cody/pull/3095

vscode/src/editor/utils/editor-context.ts
beyang: ah sorry about that, thanks for catching!

PR #3097, opened 2024-02-09
Autocomplete: bump BFG version

valerybugakov:
## Context

BFG version bump.

## Test plan

Tested locally with BFG context enabled.


philipp-spiess (approved):
Yessir

Issue #2929, opened 2024-01-29
follow-up: check paths that could be case-insensitive


abeatrix:
context: https://github.com/sourcegraph/cody/pull/2913#issuecomment-1913674458

Currently, we are applying a workaround for instances where the relative path is being computed wrong on Windows.

Instead of applying a workaround, @DanTup: 
> we might need to think about how we check paths that could be case-insensitive)

DanTup:
(from https://github.com/sourcegraph/cody/pull/2913#issuecomment-1915263484):

> Do you think this is an issue specified to cody/ignore? or does it affect other features as well? Trying to determine if this should be a follow-up issue for general or cody/ignore issue

I don't know if there are any other places computing relative paths in the way the ignore helper does so it _might_ not affect anything else, but it'd probably require some careful reviewing or testing to confirm.

Probably a good place to start would be to expand the tests for @sqs's new path functions to include mixed casing and make them all pass. Unfortunately I don't know if we can easily tell whether the URIs _should_ be case-sensitive though... so the easiest solution might be to make them all case-insensitive (or do what VS Code does and "guess" based on the platform, which is unfortunately not sound, but I think users of non-default case-sensitivity in VS Code are going to have issues anyway).

PR #3102, opened 2024-02-09
Make relative() case-insensitive for Windows paths + URIs

DanTup:
+ fix some other minor inconsistencies with slash directions such as relative always assuming input and output slash directions match (on Windows, we'd expect relative() on two file URIs to return a relative path, not URI).

Removes a workaround in the ignore helper (See https://github.com/sourcegraph/cody/issues/2929), although it might be that there are still other places that are incorrectly case-sensitive on Windows (that aren't using these path helpers).

## Test plan

- Covered by tests that run on CI (some of which were updated/expanded with this change)


abeatrix (approved):
nice! add a test to agent as well, please let me know if that doesn't look right to you :)

abeatrix:
woops ill fix it

agent/src/__tests__/example-ts/src/is_ignored.ts
```diff
@@ -0,0 +1,4 @@
+function notIgnoredByCody() {
```
DanTup, line 1: This function says "not", but the filename/comment suggest _is_ ignored. Is the function name misleading?

agent/src/index.test.ts
```diff
@@ -546,6 +546,12 @@ describe('Agent', () => {
             })
         })
 
+        it('ignore rule is not case sensitive', async () => {
+            const alsoIgnoredPath = path.join(workspaceRootPath, 'src/is_ignored.ts')
+            const result = await client.request('check/isCodyIgnoredFile', { urls: [alsoIgnoredPath] })
+            expect(result).toBeTruthy()
```
DanTup, line 552: @abeatrix is `check/isCodyIgnoredFile` used only for testing? It seems to take an array of paths but only returns a single bool. Would it be better if it returned an array of bools so you can check multiple files and know which are ignored/not rather than only knowing if "at least one" of them was ignored?

(not necessarily for this PR, since it already exists - but it seemed odd to send in an array here and then check `result`.. I was expected to see `result[0]` or similar because of the input array).

abeatrix: Yea it's only for testing right now but you're right, will update it in a follow-up!

agent/src/__tests__/example-ts/src/is_ignored.ts
abeatrix: Updated!

DanTup:
@abeatrix @olafurpg this had conflicts with the agent recordings when I merged main. I don't believe I have access to the accounts to regenerate them - is it possible one of you can regenerate them for me?

olafurpg:
The tests should pass reliably if you rebase on https://github.com/sourcegraph/cody/pull/3194

Issue #3101, opened 2024-02-09
bug: chat input doesn't allow standard text motions


zephyrtronium:
### Version

1.4

### Describe the bug

Entering a line of text in the Cody AI chat input and then using ctrl+left and ctrl+right to navigate by word does nothing. Same for ctrl+shift+left and ctrl+shift+right to select by word as reported in #2512.

### Expected behavior

Motion by words when using ctrl+left and ctrl+right, as every other text input on every major OS/DE provides.

### Additional context

Other by-word motions (ctrl+backspace and ctrl+delete) work as normal.

jdorfman:
Thanks for reporting @zephyrtronium and thanks @DanTup for fixing!

Issue #2512, opened 2023-12-27
bug: chat input line doesn't allow word selection 


vadi2:
### Version

v1.0.4

### Describe the bug

Chat input line doesn't allow ctrl+shift+left and right arrows to select whole words at once.

### Expected behavior

Standard input line behaviour is supported.

Issue #2403, opened 2023-12-15
bug: ctrl+cursor keys don't work in chat input box


DanTup:
### Version

main 2a00da01

### Describe the bug

When typing in the chat input box (on Windows), I can't use Ctrl+Left / Ctrl+Right to move the cursor around a word at a time (similarly, I can't ues Ctrl+Shift+Left / Ctrl+Shift+Right to select words).

### Expected behavior

I should be able to use these keys to quickly select/edit my chat text.

PR #3103, opened 2024-02-09
Fix Ctrl+Arrows in chat input box

DanTup:
Previously these would be caught by the condition below which calls `preventDefault` and would suppress the default behaviour.

- Fixes https://github.com/sourcegraph/cody/issues/3101
- Fixes https://github.com/sourcegraph/cody/issues/2512
- Fixes https://github.com/sourcegraph/cody/issues/2403

## Test plan

- new tests passing / green on CI, or:
- open chat
- type some words
- try to navigate by-word using Ctrl+Cursor
- try to highlight words using Ctrl+Shift+Cursor


vscode/test/e2e/chat-text-editing.test.ts
```diff
@@ -0,0 +1,27 @@
+import { expect } from '@playwright/test'
```
abeatrix, line 1: nit: should we use `chat-input.test.ts` instead to avoid confusing this with chat-edits?

abeatrix (approved):

vscode/test/e2e/chat-text-editing.test.ts
DanTup: Yeah, that's much better. Done, thanks!

PR #3107, opened 2024-02-09
Edit: Improve telemetry for cancelled edits

umpox:
## Description

If the task is not created for whatever reason we shouldn't log the edit

## Test plan

1. Try to create an edit
2. Exit the input, without submitting
3. Check no telemetry is sent


abeatrix (approved):
can be done in a follow up issue but i think this should also exclude tasks that started but not "executed" (eg the ones that are stuck in working state etc). wdyt?

umpox:
@abeatrix We also have the `'CodyVSCodeExtension:fixup:applied` and `'CodyVSCodeExtension:fixup:apply:failed` events. Those should give us an indication of what happened to the edit after this point!

PR #3111, opened 2024-02-09
context-ranking support

hitesh-1997:
## Context

Adding a context ranking module for chat command, which ranks the context from different sources such as embeddings, symf and editor. This feature currently is behind the flag `cody.experimental.chatContextRanker`. We plan to launch the changes first for the internal users.

**Note**: This updates bfg version and the related BFG release associated with this change is: https://github.com/sourcegraph/bfg/releases/tag/v5.3.1781.

Project Plan and offline analysis: https://docs.google.com/document/d/1rg9P-xqAwn_HfgKlzpHqX6v8Mf9cN-9r-S4kjr88jP8/edit?pli=1

### Feature Precomputation
Similar to local embeddings, additional indexes are created with [base path here](https://github.com/sourcegraph/cody/blob/hitesh/context-ranking-support/vscode/src/local-context/context-ranking.ts#L36). More specifically, When a user click the chat panel for the first time, we send an event to start feature pre-computations (BM25 index, embddings):

1. BM25 Index (On repos ~1.5 GB, it takes ~34 seconds to compute the index, with index size ~45 MBs)
2. Sentence transformer embeddings index (We use local embedding backend to generate the same)

### Ranking call
In case the user has selected the enhanced context option and `cody.experimental.chatContextRanker` flag is set.
1. We compute context item from all context providers (embeddings, symf and editor)
2. Make a call to the cody engine via json-rpc to get the scores for each context item.  
3. Finally sort the context items, based on the ranker score.

## Dependent PR's
1. https://github.com/sourcegraph/bfg-private/pull/167 (changes in the cody engine to support context-ranking)

### Manul Integration test
I have tested the approach on macos and linux based system.

1.  Setup the following flag in the vscode user settings
```
  "cody.experimental.chatContextRanker": true
```

3. Checkout to this branch and start the vscode debugger.
4. Open a repo and click on the chat command from the vscode commands. This will start the feature computation job and the artifacts should start populating in the relevant directories. (For embedding the file names start with `.tmp`, so to list the embedding list do `ls -a`)
- For macos: `~/Library/Caches/com.sourcegraph.cody/context-ranking`
- For Linux: `~/.cache/com.sourcegraph.cody/context-ranking`
6. Testing ranker call: type a question in the chat and add a breakpoint [here](https://github.com/sourcegraph/cody/blob/hitesh/context-ranking-support/vscode/src/local-context/context-ranking.ts#L138), the output will be the predicted ranker scores for the context items and we get the ranked context items which should reflect in the UI.

Note:: The flags in the video below are outdated, we only need to set the flag: `"cody.experimental.chatContextRanker": true`.

https://github.com/sourcegraph/cody/assets/20701220/f4040008-7c57-44b0-8ecd-110b55050cb5

### DIFF
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -76,6 +76,8 @@ export interface Configuration {
     testingLocalEmbeddingsModel: string | undefined
     testingLocalEmbeddingsEndpoint: string | undefined
     testingLocalEmbeddingsIndexLibraryPath: string | undefined
+
+    experimentalChatContextRanker: boolean | undefined
 }
 
 export interface AutocompleteTimeouts {
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -50,6 +50,7 @@ export interface Configuration {
     experimentalSymfContext: boolean
     experimentalTracing: boolean
     experimentalSimpleChatContext: boolean
+    experimentalChatContextRanker: boolean | undefined
 
     /**
      * Unstable Features for internal testing only
@@ -76,8 +77,6 @@ export interface Configuration {
     testingLocalEmbeddingsModel: string | undefined
     testingLocalEmbeddingsEndpoint: string | undefined
     testingLocalEmbeddingsIndexLibraryPath: string | undefined
-
-    experimentalChatContextRanker: boolean | undefined
 }
 
 export interface AutocompleteTimeouts {
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -80,9 +80,13 @@ import { ModelUsage } from '@sourcegraph/cody-shared/src/models/types'
 import { chatModel } from '../../models'
 import { getContextWindowForModel } from '../../models/utilts'
 import type { ContextItem } from '../../prompt-builder/types'
+<<<<<<< HEAD
 import type { Span } from '@opentelemetry/api'
 import { recordErrorToSpan, tracer } from '@sourcegraph/cody-shared/src/tracing'
 import { recordExposedExperimentsToSpan } from '../../services/open-telemetry/utils'
+=======
+import { ContextRankingController } from '../../local-context/context-ranking'
+>>>>>>> af602235 (context-ranking support)
 
 interface SimpleChatPanelProviderOptions {
     config: ChatPanelConfig
@@ -90,6 +94,7 @@ interface SimpleChatPanelProviderOptions {
     authProvider: AuthProvider
     chatClient: ChatClient
     localEmbeddings: LocalEmbeddingsController | null
+    contextRanking: ContextRankingController | null
     symf: SymfRunner | null
     enterpriseContext: EnterpriseContextFactory | null
     editor: VSCodeEditor
@@ -138,6 +143,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
     private readonly chatClient: ChatClient
     private readonly codebaseStatusProvider: CodebaseStatusProvider
     private readonly localEmbeddings: LocalEmbeddingsController | null
+    private readonly contextRanking: ContextRankingController | null
     private readonly symf: SymfRunner | null
     private readonly contextStatusAggregator = new ContextStatusAggregator()
     private readonly editor: VSCodeEditor
@@ -161,6 +167,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         authProvider,
         chatClient,
         localEmbeddings,
+        contextRanking,
         symf,
         editor,
         treeView,
@@ -173,6 +180,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         this.authProvider = authProvider
         this.chatClient = chatClient
         this.localEmbeddings = localEmbeddings
+        this.contextRanking = contextRanking
         this.symf = symf
         this.repoPicker = enterpriseContext?.repoPicker || null
         this.remoteSearch = enterpriseContext?.createRemoteSearch() || null
@@ -187,6 +195,9 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
 
         // Advise local embeddings to start up if necessary.
         void this.localEmbeddings?.start()
+
+        // Start the context Ranking module
+        void this.contextRanking?.start()
 
         // Push context status to the webview when it changes.
         this.disposables.push(
@@ -395,6 +406,34 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                 FeatureFlag.CodyChatFusedContext
             )
 
+<<<<<<< HEAD
+=======
+        const userContextItems = await contextFilesToContextItems(
+            this.editor,
+            userContextFiles || [],
+            true
+        )
+        const prompter = new DefaultPrompter(
+            userContextItems,
+            addEnhancedContext
+                ? (text, maxChars) =>
+                      getEnhancedContext({
+                          strategy: this.config.useContext,
+                          editor: this.editor,
+                          text,
+                          providers: {
+                              localEmbeddings: this.localEmbeddings,
+                              symf: this.config.experimentalSymfContext ? this.symf : null,
+                              remoteSearch: this.remoteSearch,
+                          },
+                          featureFlags: this.config,
+                          hints: { maxChars },
+                          contextRanking: this.contextRanking,
+                      })
+                : undefined
+        )
+        const sendTelemetry = (contextSummary: any): void => {
+>>>>>>> af602235 (context-ranking support)
             const authStatus = this.authProvider.getAuthStatus()
             const sharedProperties = {
                 requestID,
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -473,9 +473,6 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                                   featureFlags: {
                                       fusedContext:
                                           this.config.internalUnstable || (await useFusedContextPromise),
-                                      testingContextRanking:
-                                          this.config.internalUnstable &&
-                                          this.config.experimentalChatContextRanker,
                                   },
                                   hints: { maxChars },
                                   contextRanking: this.contextRanking,
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -80,13 +80,10 @@ import { ModelUsage } from '@sourcegraph/cody-shared/src/models/types'
 import { chatModel } from '../../models'
 import { getContextWindowForModel } from '../../models/utilts'
 import type { ContextItem } from '../../prompt-builder/types'
-<<<<<<< HEAD
 import type { Span } from '@opentelemetry/api'
 import { recordErrorToSpan, tracer } from '@sourcegraph/cody-shared/src/tracing'
 import { recordExposedExperimentsToSpan } from '../../services/open-telemetry/utils'
-=======
-import { ContextRankingController } from '../../local-context/context-ranking'
->>>>>>> af602235 (context-ranking support)
+import type { ContextRankingController } from '../../local-context/context-ranking'
 
 interface SimpleChatPanelProviderOptions {
     config: ChatPanelConfig
@@ -406,34 +403,6 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                 FeatureFlag.CodyChatFusedContext
             )
 
-<<<<<<< HEAD
-=======
-        const userContextItems = await contextFilesToContextItems(
-            this.editor,
-            userContextFiles || [],
-            true
-        )
-        const prompter = new DefaultPrompter(
-            userContextItems,
-            addEnhancedContext
-                ? (text, maxChars) =>
-                      getEnhancedContext({
-                          strategy: this.config.useContext,
-                          editor: this.editor,
-                          text,
-                          providers: {
-                              localEmbeddings: this.localEmbeddings,
-                              symf: this.config.experimentalSymfContext ? this.symf : null,
-                              remoteSearch: this.remoteSearch,
-                          },
-                          featureFlags: this.config,
-                          hints: { maxChars },
-                          contextRanking: this.contextRanking,
-                      })
-                : undefined
-        )
-        const sendTelemetry = (contextSummary: any): void => {
->>>>>>> af602235 (context-ranking support)
             const authStatus = this.authProvider.getAuthStatus()
             const sharedProperties = {
                 requestID,
@@ -504,8 +473,12 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                                   featureFlags: {
                                       fusedContext:
                                           this.config.internalUnstable || (await useFusedContextPromise),
+                                      testingContextRanking:
+                                          this.config.internalUnstable &&
+                                          this.config.experimentalChatContextRanker,
                                   },
                                   hints: { maxChars },
+                                  contextRanking: this.contextRanking,
                               })
                         : undefined
                 )
--- a/vscode/src/chat/chat-view/context.ts
+++ b/vscode/src/chat/chat-view/context.ts
@@ -19,6 +19,7 @@ import { logDebug, logError } from '../../log'
 import { viewRangeToRange } from './chat-helpers'
 import type { RemoteSearch } from '../../context/remote-search'
 import type { ContextItem } from '../../prompt-builder/types'
+import { ContextRankingController } from '../../local-context/context-ranking'
 
 export interface GetEnhancedContextOptions {
     strategy: ConfigurationUseContext
@@ -35,6 +36,7 @@ export interface GetEnhancedContextOptions {
     hints: {
         maxChars: number
     }
+    contextRanking: ContextRankingController | null
     // TODO(@philipp-spiess): Add abort controller to be able to cancel expensive retrievers
 }
 export async function getEnhancedContext({
@@ -44,6 +46,7 @@ export async function getEnhancedContext({
     providers,
     featureFlags,
     hints,
+    contextRanking,
 }: GetEnhancedContextOptions): Promise<ContextItem[]> {
     if (featureFlags.fusedContext) {
         return getEnhancedContextFused({
@@ -53,6 +56,7 @@ export async function getEnhancedContext({
             providers,
             featureFlags,
             hints,
+            contextRanking
         })
     }
 
@@ -113,6 +117,7 @@ async function getEnhancedContextFused({
     text,
     providers,
     hints,
+    contextRanking,
 }: GetEnhancedContextOptions): Promise<ContextItem[]> {
     return wrapInActiveSpan('chat.enhancedContextFused', async () => {
         // use user attention context only if config is set to none
--- a/vscode/src/chat/chat-view/context.ts
+++ b/vscode/src/chat/chat-view/context.ts
@@ -32,7 +32,6 @@ export interface GetEnhancedContextOptions {
     }
     featureFlags: {
         fusedContext: boolean
-        testingContextRanking: boolean | undefined
     }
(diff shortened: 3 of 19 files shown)


dominiccooney (changes requested):
We've got some copy-pasta from the local embeddings controller here. That class needs to be simplified a bunch.

Let's start by having local embeddings and the re-ranker share a process, instead of each spawning their own processes, because they're doing a similar job at a similar time. So I would:

1. Split the local embeddings controller into two parts: One part for spawning cody-engine; the other part for all the embeddings stuff.
2. Make one instance of the cody-engine. Right now it has one job, starting the engine up. But later we need to build in some reliability for crashes, etc. and this will be the natural place to do it.
3. Have the context ranker and local embeddings share the cody-engine spawner and share one process spawned by it.

hitesh-1997:
> We've got some copy-pasta from the local embeddings controller here. That class needs to be simplified a bunch.
> 
> Let's start by having local embeddings and the re-ranker share a process, instead of each spawning their own processes, because they're doing a similar job at a similar time. So I would:
> 
> 1. Split the local embeddings controller into two parts: One part for spawning cody-engine; the other part for all the embeddings stuff.
> 2. Make one instance of the cody-engine. Right now it has one job, starting the engine up. But later we need to build in some reliability for crashes, etc. and this will be the natural place to do it.
> 3. Have the context ranker and local embeddings share the cody-engine spawner and share one process spawned by it.

Hi @dominiccooney 
Added the changes, could you please check once

dominiccooney (changes requested):
- Major thing missing is a flag to turn this off by default.
- Needs rebasing.
- If you want to avoid getting broken by changes, adding a test is advised but up to you. Maybe you don't need one at first if your test audience is small.

Earlier today @beyang mentioned putting these files in their own directory. I think that makes sense, but we can do it later.

vscode/src/extension.node.ts
```diff
@@ -33,6 +37,7 @@ export function activate(context: vscode.ExtensionContext): Promise<ExtensionApi
             ? undefined
             : (config: LocalEmbeddingsConfig): LocalEmbeddingsController =>
                   createLocalEmbeddingsController(context, config),
+        createContextRankingController: (config: ContextRankerConfig) => createContextRankingController(context, config),
```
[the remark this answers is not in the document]
hitesh-1997: Added the changes domonic.

lib/shared/src/configuration.ts
```diff
@@ -76,6 +76,8 @@ export interface Configuration {
     testingLocalEmbeddingsModel: string | undefined
     testingLocalEmbeddingsEndpoint: string | undefined
     testingLocalEmbeddingsIndexLibraryPath: string | undefined
+
+    experimentalChatContextRanker: boolean | undefined
```
philipp-spiess, line 80: Might want to move this up to the other experimental flags on Line 49-52

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -465,8 +473,12 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                                   featureFlags: {
                                       fusedContext:
                                           this.config.internalUnstable || (await useFusedContextPromise),
+                                      testingContextRanking:
+                                          this.config.internalUnstable &&
+                                          this.config.experimentalChatContextRanker,
                                   },
                                   hints: { maxChars },
+                                  contextRanking: this.contextRanking,
```
philipp-spiess, line 481: If we already pass the context reranking module through here, would we want to make it optional and remove the `testingContextRanking` instead?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -188,6 +193,9 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         // Advise local embeddings to start up if necessary.
         void this.localEmbeddings?.start()
 
+        // Start the context Ranking module
+        void this.contextRanking?.start()
```
philipp-spiess, line 197: Should this be gated behind `experimentalChatContextRanker` too?

vscode/src/configuration.ts
```diff
@@ -170,6 +170,10 @@ export function getConfiguration(
         testingLocalEmbeddingsIndexLibraryPath: isTesting
             ? getHiddenSetting<string | undefined>('testing.localEmbeddings.indexLibraryPath', undefined)
             : undefined,
+
+        experimentalChatContextRanker:
+            getHiddenSetting('internal.unstable', false) &&
+            getHiddenSetting('experimental.chatContextRanker', false),
```
philipp-spiess, line 176: Again unclear why we need to enable two flags here.

vscode/src/local-context/cody-engine.ts
```diff
@@ -0,0 +1,54 @@
+import type * as vscode from 'vscode'
+import { spawnBfg } from '../graph/bfg/spawn-bfg'
+import type { MessageHandler } from '../jsonrpc/jsonrpc'
+import { logDebug } from '../log'
+import { captureException } from '../services/sentry/sentry'
+
+export class CodyEngineService {
```
philipp-spiess, line 7: This is nice! Is there a reason for this not also being used by the bfg implementation? cc @valerybugakov

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
+import * as vscode from 'vscode'
+import * as path from 'path'
+
+// import { URI } from 'vscode-uri'
+import { logDebug } from '../log'
+
+import {
+    isDotCom,
+    type ConfigurationWithAccessToken,
+    // type FileURI,
```
philipp-spiess, line 10:
```suggestion
```

leftover

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
+import * as vscode from 'vscode'
+import * as path from 'path'
+
+// import { URI } from 'vscode-uri'
```
philipp-spiess, line 4:
```suggestion
```

leftover

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
...
+        case 'win32':
+            return URI.file(`${process.env.LOCALAPPDATA}\\com.sourcegraph.cody\\context-ranking`)
+        default:
+            throw new Error(`Unsupported platform: ${process.platform}`)
+    }
+}
+
+export class ContextRankingController implements ContextRanker {
+    // The cody-engine child process, if starting or started.
+    private service: Promise<MessageHandler> | undefined
```
philipp-spiess, line 51: Is there a reason why this is a Promise and not the awaited value? Would prevent a micro event tick for every caller when we await it before we store it in the class properties

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
...
+        if (this.indexLibraryPath) {
+            logDebug(
+                'ContextRankingController',
+                'spawnAndBindService',
+                'overriding index library path',
+                this.indexLibraryPath
+            )
+            indexPath = this.indexLibraryPath
+        }
+        if (this.accessToken) {
```
philipp-spiess, line 119: Maybe an else case with some logging so we can debug when this is not the case?

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
...
+            const initResult = await service.request('context-ranking/initialize', {
+                indexPath: indexPath.fsPath,
+                accessToken: this.accessToken,
+            })
+            logDebug('LocalEmbeddingsController', 'spawnAndBindService', 'initialized', initResult)
```
philipp-spiess, line 124: Maybe we can consolidate the debug logs since it's getting quite noisy in the output log. E.g. the previous log for ``overriding index library path` could be part of the results here like this:

```suggestion
            logDebug('LocalEmbeddingsController', 'spawnAndBindService', 'initialized', JSON.stringify({ initResult, indexPath })
```

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
...
+        }
+    }
+
+    private convertContextItemsToRankItems(
+        baseRepoUrl: string,
+        contextItems: ContextItem[]
+    ): RankContextItem[] {
+        const rankContextItems = contextItems.map((item, index) => ({
+            document_id: index,
+            filePath: item.uri?.path ? path.relative(baseRepoUrl, item.uri?.path) : '',
```
philipp-spiess, line 173: I remember we had issues with relying on `Uri#path` before and we started to use `Uri#toString()`. Is that still the case? We might want to consider the lessons learned here right from the start. cc @dominiccooney who I remember running into this

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
...
+        }))
+        return rankContextItems
+    }
+
+    private orderContextItemsAsRankItems(
+        contextItems: ContextItem[],
+        rankedItemsOrder: RankerPrediction[]
+    ): ContextItem[] {
+        rankedItemsOrder.sort((a, b) => b.score - a.score)
+        const orderedContextItems: ContextItem[] = [] // Initialize the array
```
philipp-spiess, line 185:
```suggestion
        const orderedContextItems: ContextItem[] = []
```

nit: these type of comments are usually a distraction when the code is obvious

vscode/src/local-context/context-ranking.ts
```diff
@@ -0,0 +1,202 @@
...
+        for (const item of rankedItemsOrder) {
+            const newIndex = item.document_id
+            if (newIndex < 0 || newIndex >= contextItems.length) {
+                logDebug(
+                    'ContextRankingController',
+                    'length of rankItems',
+                    JSON.stringify(rankedItemsOrder),
+                    'does not match context items',
+                    JSON.stringify(contextItems)
+                )
```
philipp-spiess, line 195: Maybe this should throw instead and we capture it via sentry or in the tracing APIs? Otherwise this is going to just create noise in the users console and we won't be notified about it.

philipp-spiess (commented):
This is some great stuff here! I left a few comments (and _a lot_ of nits which you are free to ignore). 

The only two big thing standing out that IMO should be adresed before we land this is:

- This currently requires two flags to be enabled where maybe one is enough?
- I’m unsure if we should include the `priorityContext` in the reranking for the reason I mentioned inline.

Awesome work otherwise!

vscode/src/chat/chat-view/context.ts
```diff
@@ -159,6 +175,53 @@ async function getEnhancedContextFused({
...
+            keywordContextItemsPromise,
+        ])
+
+        searchContext = searchContext.concat(keywordContextItems).concat(embeddingsContextItems)
+        const editorContext = await getPriorityContext(text, editor, searchContext)
+        const allContext = editorContext.concat(searchContext)
+        if (!contextRanking) {
+            return allContext
+        }
+        const rankedContext = contextRanking.rankContextItems(text, allContext)
```
philipp-spiess, line 220: Let's create a sub-span for this step using `wrapInActiveSpan` so we can see in our tracing data how long this takes for our users

```suggestion
        const rankedContext = wrapInActiveSpan('chat.enhancedContextRanker.reranking', () => contextRanking.rankContextItems(text, allContext))
```

dominiccooney:
@hitesh-1997 , I've approved your cody-engine PR so land that, tag it to start a release, and then update the BFG version in download-bfg.ts. After @philipp-spiess 's feedback (thanks Philipp!) you should be good to go.

vscode/src/chat/chat-view/context.ts
```diff
@@ -159,6 +175,53 @@ async function getEnhancedContextFused({
...
+            ...(await remoteSearchContextItemsPromise),
+        ])()
+
+        const [embeddingsContextItems, keywordContextItems] = await Promise.all([
+            embeddingsContextItemsPromise,
+            keywordContextItemsPromise,
+        ])
+
+        searchContext = searchContext.concat(keywordContextItems).concat(embeddingsContextItems)
+        const editorContext = await getPriorityContext(text, editor, searchContext)
```
philipp-spiess, line 215: The priority context, IIRC, refers to the context that is explicitly selected in the editor and/or added via `@` mentions in the chat message. If we add it to the `allContext` array here and then re-rank, wouldn't it be possible to end up not having _any_ of the priority context items in the final context list? 

I would assume this would result in a major regression when you ask "what does this code do" and this code is never added 🤔

vscode/src/chat/chat-view/context.ts
```diff
@@ -159,6 +175,53 @@ async function getEnhancedContextFused({
     })
 }
 
+async function getEnhancedContextFromRanker({
```
philipp-spiess, line 178: One last nit: I think, conceptually, this does the same as the `getEnhancedContextFused` just with a different implementation of _how to fuse_. Maybe instead of forking everything we can just make the last line of the `getEnhancedContextFused` dynamic? This way, if we change something in the context sources, we don't have to update 3 places

dominiccooney:
I forgot to add: AWESOME demo in the video @hitesh-1997 . Looks amazing. I am so excited to see this working end to end. 🎉

vscode/src/chat/chat-view/context.ts
dominiccooney: Following up from earlier comments about ML prototype code: This is the kind of feedback I *do* think should be addressed. We should try to minimize the maintenance burden that the experimental code puts on the shipping product.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
dominiccooney: Think it is OK as-is, the `contextRanking` will be null, right? So the call is a no-op.

philipp-spiess: Ah, yes you're right but right now these use two different gates. the context ranking service is enabled if `experimentalChatContextRanker` is set: https://github.com/sourcegraph/cody/pull/3111/files/4e8e2c4beb13cf6170575aa677e90760d0973fc8#diff-13b80d92afa9fb014e0ffc69e6f1bc179bb4516b003c4916456af49d8fc22fe1R82-R84 but the gating below depends on `experimentalChatContextRanker && internalUnstable`. Should be fine if we unify the condition 👍

philipp-spiess: I think it's fine if we just remove `testingContextRanking` from the `getEnhancedContext` and then check on the presence of `contextRanking`

vscode/src/chat/chat-view/context.ts
philipp-spiess: Ahh, gotcha, yeah this makes sense. 👍

hitesh-1997: Also, just to cross check @rishabhmehrotra can provide additional inputs if adding all context source in the ranker is okay ?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -465,8 +473,12 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                                   featureFlags: {
                                       fusedContext:
                                           this.config.internalUnstable || (await useFusedContextPromise),
+                                      testingContextRanking:
+                                          this.config.internalUnstable &&
+                                          this.config.experimentalChatContextRanker,
```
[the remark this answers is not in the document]
hitesh-1997: @philipp-spiess I think it makes sense to just keep this.config.experimentalChatContextRanker flag. Making this change. Thanks for the input :)

lib/shared/src/configuration.ts
hitesh-1997: done

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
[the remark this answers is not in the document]
hitesh-1997: Thanks changing to only `this.config.experimentalChatContextRanker` as per discussion

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: that makes sense, since contextRanking is initialized only if the flag is true, we can check only on `this.contextRanking` condition. let me know if sounds good.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: this makes sense @philipp-spiess only relying on single condition

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
philipp-spiess: Sounds great!

vscode/src/jsonrpc/context-ranking-protocol.ts
```diff
@@ -0,0 +1,37 @@
...
+    accessToken: string
+}
+
+interface ComputeFeaturesParams {
+    repoPath: string
+}
+
+export interface RankContextItem {
+    document_id: number
+    filePath?: string
```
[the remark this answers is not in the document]
hitesh-1997: ahh, I did not notice this. I have added a todo here and i can do it later, since this would also involve chaning bfg protocol code.

vscode/src/local-context/cody-engine.ts
hitesh-1997: We introduced this class in the current PR and i think others haven't adapted to this yet and I only changed the local embeddings to use this.

vscode/src/local-context/context-ranking.ts
hitesh-1997: This implementation of this is very much [similar to local embeddings](https://github.com/sourcegraph/cody/blob/main/vscode/src/local-context/local-embeddings.ts#L70) I wanted to keep it same as for this first version. Is it okay if try to improve this logic in the future iterations ?

vscode/src/local-context/cody-engine.ts
philipp-spiess: Oh yeah this is more of a follow-up for @valerybugakov if we can adapt BFG to use that, too! Sorry for the confusing wording.

vscode/src/local-context/context-ranking.ts
philipp-spiess: Totally!

vscode/src/local-context/context-ranking.ts
hitesh-1997: done

vscode/src/local-context/context-ranking.ts
hitesh-1997: makes sense, changed

vscode/src/local-context/context-ranking.ts
hitesh-1997: Thanks for letting know. I have changed to use `fsPath` here and changed `baseRepoUrl` to `baseRepoPath`.

vscode/src/local-context/context-ranking.ts
hitesh-1997: removed the comment

vscode/src/chat/chat-view/context.ts
hitesh-1997: nice, this would be great to have !!
Added

hitesh-1997:
> I forgot to add: AWESOME demo in the video @hitesh-1997 . Looks amazing. I am so excited to see this working end to end. 🎉

Thanks Dominic for helping us land this !!

hitesh-1997:
> @hitesh-1997 , I've approved your cody-engine PR so land that, tag it to start a release, and then update the BFG version in download-bfg.ts. After @philipp-spiess 's feedback (thanks Philipp!) you should be good to go.

Hi @dominiccooney and @philipp-spiess 
I have updated the [release tag](https://github.com/sourcegraph/bfg/releases/tag/v5.3.1781) and also updated the PR description to reflect the same. In addition, I did manually integration test described in the PR description on both Mac and Linux and they seem to be working fine. Although i haven't tested on windows, Local embeddings does e2e test on windows so i think it should be covered. Please let me know if I should add unit test on cody-engine or any other test to make sure new release tag is working as expected.

philipp-spiess (approved):
🚢

PR #3114, opened 2024-02-09
Chat: improve handling of at mentions

abeatrix:
The extractMentionQuery function has been updated to handle cases where the user types the file without tabbing the selection. If a context selection is available and the current chat context query matches the input query, the first context selection should be automatically selected.

## Test plan

Green bots = 🆗 

1. at-mention a file without using tab or enter, just type out the whole file path, and end it with space
2. it should stlil include the file on submit

### After

After

### Before

before


abeatrix:
@toolmantim not sure if this is a bug, or expected behavior. if it's expected, we can just close this PR 🙇‍♀️

toolmantim (approved):
This is great!

PR #3117, opened 2024-02-09
Fixing logs on the failed GQL queries for refreshConfigFeatures

arafatkatze:
Solves the issue https://github.com/sourcegraph/sourcegraph/issues/60050
Basically we already wanted to make sure that GQL and fetch errors aren't something that hammer the logs console so this PR adds an extra case of checking that the error message doesn't have the string that gets repeatedly printed in this users console. 

## Test plan

I didn't thoroughly test this PR in the sense that going back to v5.2.0 and then testing. I did go to that version but doing the setup on my local instance caused other issues for me so I left it. My understanding is that this issue is just about seeing something non necessary in the logs so its not really an actually problematic issue.


chwarwick (approved):
If you want to test against a 5.2.x environment demo.sourcegraph.com or the sg01 test environment will work. I just created a user for you on sg01.

arafatkatze:
Before 
image

After 
image


This issue has been resolved and tested too. 

Thanks @chwarwick I did this on demo instance.

Issue #2965, opened 2024-01-31
Display warning for @-mentioning massive files


kalanchan:
Display warning/error message when file is too large to @ in chat

see design and input: https://sourcegraph.slack.com/archives/C05AGQYD528/p1704800906759909

abeatrix:
@kalanchan can I get more information for the task please? 
What's the definition of a large file here? What's the expected behavior?

toolmantim:
The Figma for this is here: https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=331-103646&mode=design&t=jidOC06CYg9nSWOI-4

I guess the bug is that if a file is too large to include in context, we just exclude it, and users are wondering why.

Ideally we detect that it will be too big at file-selection time, and ask them to provide a symbol.

kalanchan:
what is considered too large? 🤔


Ideally we have a mechanism that detects how many tokens it's going to eat up, and make a decision there. But without it we are kinda just guess right? 

@dominiccooney any thoughts on how we can programatically calculate this

toolmantim:
An alternative to my design, which we should consider/test, is to include a summary of the file in the context vs forcing them to select a symbol. A symbol may be too large for the context too 🤔

abeatrix:
> Ideally we detect that it will be too big at file-selection time, and ask them to provide a symbol.

What if they select a file that's within the "limit" but then they added a question that's over the overall limit? 

Should we set a maximum token limit allowed per file (not ideal) and per chat input?

Or maybe add a token counter in the chat ui to make the limit visible instead (like a tweet limit in Twitter)?

umpox:
@toolmantim @abeatrix We should ensure we do this for the Edit logic too. I'm happy to take a look at this post-release.

I think we can use the `detail` option in the QuickPick to show a similar UI

abeatrix:
Adding @umpox for Edit logic

chillatom:
Is this at the complexity level that we should write an RFC?  Sounds like we have a few options

1. Warning for too large
2. Push users towards symbols
3. Gracefully handle too large files
    - summarize
    - expand context window
    - other?

There are some other edge cases to consider (some of which listed above) 

1. What if mentioning N files and the Nth pushes over the context limit
2. What if symbols are too large

virgildotcodes:
[...]

Issue #2636, opened 2024-01-09
bug: request sent to LLM without context due to large context size


abeatrix:
### Version

v1.0.5

### Describe the bug

If the shared context is too large, the chat question will still be submitted to the LLM without context, leading to the LLM response that it doesn't have any context:

![image](https://github.com/sourcegraph/cody/assets/68532117/7e9bf28d-43c1-4b17-9de5-709e5c410278)

However, we wouldn't know that no context was shared due to the context's size until we open the output channel and look for logs:

![image](https://github.com/sourcegraph/cody/assets/68532117/93780fe2-ecbb-4df5-aab7-4edcb2a42402)

### Expected behavior

- let user know in the UI that the context was not shared due to context limit
- context should be truncated and share with LLM to provide a response

deepak2431:
@abeatrix Looking into this issue!

abeatrix:
> let user know in the UI that the context was not shared due to context limit

Fixed in https://github.com/sourcegraph/cody/pull/3528
We now display the excluded @-mention files in UI to avoid confusion:
![image](https://github.com/sourcegraph/cody/assets/68532117/2f118331-5c32-4b21-8b48-200352f0ce96)

Issue #2769, opened 2024-01-17
Chat: Better context summarization and truncation


dominiccooney:
### Areas for Improvement

- [ ] UI/UX
- [ ] Onboarding
- [ ] Docs
- [X] Chat
- [ ] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience.

Cody context is truncated, causing unquantified forgetting and poor performance. We tried simply showing the user an error message when this happened, but that was confusing.

### Why is this feedback important? Please describe the rationale.

Chat and inline edits are a core part of the Cody experience and we want it to be as good as possible. But LLM context sizes are limited or large context windows are slow, so we have to manage the limited context window size that we have.

### Describe the solution you'd like to see

Here's the chat context truncation behavior we have today. Include the following until the limit is reached, then truncate:
- Preamble
- Existing transcript up to current message, prioritizing recent messages (that is, current message is the most important)
- Explicit tagged context up to current message, prioritizing recent messages (that is, current tagged context is the most important)
- Current enhanced context

We should do some hand inspection of this truncation, but I think we will do better with something like:
- Preamble
- Current message
- Current explicit tagged context
- Current enhanced context (note, to this point, we are basically like "new chat" context.)
- Existing transcript *and* tagged context, prioritizing recent messages (that is, previous bot response is the most important, and tagged context is likely to be included.)

### Describe any alternatives that could be considered

Some other casual ideas:
- Have the LLM giving feedback along with its response about which context sources it found useful, and boost the priorities of those in the replayed transcript (@rishabhmehrotra has mentioned wanting one-shot learning in the transcript, this is a little bit like that...)
- Summarize the existing transcript instead of truncating it, at the cost of an extra turn.

blarvin:
Will it be possible in future to DELETE irrelevent chat messages from a conversation, and have the context updated accordingly? Is that a thing? (I created a feedback issue #2579  suggesting a delete chat message feature - mostly its a UX thing, but context would be major bonus )

kalanchan:
@dominiccooney @chillatom to discuss what we need to do here to preserve context in each multi turn chat

PR #3118, opened 2024-02-09
Chat: display file size warning on large files

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1707508300392389
Part of https://github.com/sourcegraph/cody/issues/2636 & https://github.com/sourcegraph/cody/issues/2965

Update chat UI to provide better file size awareness based on Tim's figma design:  https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=331-103646&mode=design&t=jidOC06CYg9nSWOI-4

image

This PR includes the following changes:
- Filter the `sorted results` in the `getFileContextFiles` function to remove non-files and files that are larger than 1MB
- Set title as `large-file` for files that contain more tokens than we allow for file using the file size returned from `vscode.workspace.fs.stat` API
- When users try to add a file using @-mentioned, they should see a warning below the file name if the file is too large 
- ~~Truncating prompt messages in the PromptBuilder class to include truncated context without all context being ignored due to excessive length.~~
- ~~Use the exisiting truncatePrompt function and MAX_AVAILABLE_PROMPT_LENGTH constant for truncating prompts.~~
- Display warning in at-mentions selector for large files

### Note

- This PR focuses on providing information to users in the UI for files they are trying to add using `@-mention` that are too large.
- ~~We are also truncating the context files to make sure large files will be truncated and used as context, instead of excluding the file.~~

Follow-up works include:
- truncate large files
- show users if a file was truncated 

## Test plan

1. @-mentioned a large file [SimpleChatPanelProvider](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@916ef190bf286505c5dcb254b87c0ad142b86c5a/-/blob/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts)

image
raph/cody/assets/68532117/7ea38043-c148-45d7-9a00-1d6aeb53f552)

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -544,7 +544,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         const cancellation = new vscode.CancellationTokenSource()
 
         try {
-            const MAX_RESULTS = 10
+            const MAX_RESULTS = 20
             if (query.startsWith('#')) {
                 // It would be nice if the VS Code symbols API supports
                 // cancellation, but it doesn't
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -544,7 +544,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         const cancellation = new vscode.CancellationTokenSource()
 
         try {
-            const MAX_RESULTS = 20
+            const MAX_RESULTS = 10
             if (query.startsWith('#')) {
                 // It would be nice if the VS Code symbols API supports
                 // cancellation, but it doesn't
--- a/vscode/src/editor/utils/editor-context.ts
+++ b/vscode/src/editor/utils/editor-context.ts
@@ -20,6 +20,8 @@ import {
 import { getOpenTabsUris, getWorkspaceSymbols } from '.'
 import { CHARS_PER_TOKEN } from '@sourcegraph/cody-shared/src/prompt/constants'
 
+const isTesting = process.env.CODY_TESTING === 'true'
+
 const findWorkspaceFiles = async (
     cancellationToken: vscode.CancellationToken
 ): Promise<vscode.Uri[]> => {
@@ -114,21 +116,30 @@ export async function getFileContextFiles(
 
     // TODO(toolmantim): Add fuzzysort.highlight data to the result so we can show it in the UI
 
+    if (isTesting) {
+        return sortedResults
+    }
+
     const filtered = []
-    for (const sorted of sortedResults) {
-        // Remove file larger than 1MB and non-text files
-        // NOTE: Sourcegraph search only includes files up to 1MB
-        const fileStat = await vscode.workspace.fs.stat(sorted.uri)
-        if (fileStat.size > 1000000 || !fileStat.size || fileStat.type !== vscode.FileType.File) {
-            continue
+
+    try {
+        for (const sorted of sortedResults) {
+            // Remove file larger than 1MB and non-text files
+            // NOTE: Sourcegraph search only includes files up to 1MB
+            const fileStat = await vscode.workspace.fs.stat(sorted.uri)
+            if (fileStat.size > 1000000 || fileStat.type !== vscode.FileType.File) {
+                continue
+            }
+            // Check if file contains more characters than the token limit based on fileStat.size
+            // and set the title of the result as 'large-file' for webview to display file size
+            // warning.
+            if (fileStat.size > CHARS_PER_TOKEN * MAX_CURRENT_FILE_TOKENS) {
+                sorted.title = 'large-file'
+            }
+            filtered.push(sorted)
         }
-        // Check if file contains more characters than the token limit based on fileStat.size
-        // and set the title of the result as 'large-file' for webview to display file size
-        // warning.
-        if (fileStat.size > CHARS_PER_TOKEN * MAX_CURRENT_FILE_TOKENS) {
-            sorted.title = 'large-file'
-        }
-        filtered.push(sorted)
+    } catch (error) {
+        console.error('getFileContextFiles:failed', error)
     }
 
     return filtered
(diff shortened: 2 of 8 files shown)


vscode/webviews/UserContextSelector.tsx
```diff
@@ -96,7 +97,12 @@ export const UserContextSelectorComponent: React.FunctionComponent<
                             ? `:${match.range.start.line + 1}-${match.range.end.line + 1}`
                             : ''
                         const description =
-                            match.type === 'file' ? undefined : displayPath(match.uri) + range
+                            match.type === 'file'
+                                ? match.content &&
+                                  estimateTokenCount(match.content) > MAX_CURRENT_FILE_TOKENS
+                                    ? 'Large file will be truncated.'
```
abeatrix, line 103:
```suggestion
                                    ? 'Large file will be truncated.'
```

@toolmantim feel free to update this copy

abeatrix: Getting the file content for all search results is most likely going to cause performance issues in large repos / large files, so I'm wondering if we should just show this warning at the top instead.

vscode/src/editor/utils/editor-context.ts
```diff
@@ -244,3 +250,10 @@ function createContextFileRange(selectionRange: vscode.Range): ContextFile['rang
         },
     }
 }
+
+async function getFileContent(uri: vscode.Uri): Promise<string> {
+    return vscode.workspace.fs.readFile(uri).then(
```
dominiccooney, line 255: You can use `vscode.workspace.fs.stat(uri)` and check `size`. Performance is O(number of files) not O(size of all the files) so it's better. Downside is it's a byte count, not a character count... I haven't looked at tokenizer rates for non-Latin character sets but I would be surprised if character count is a better factor than byte count.

vscode/webviews/UserContextSelector.tsx
dominiccooney: The principle is: don't disappoint/frustrate the user. So we should keep them informed as early as possible. Ideally, tell them what's going on, without bogging down the machine, before they commit to an action.

I suggested stat-ing the filesystem above, which will reduce the load a bit, but stat is not free. So maybe you could just stat items in the viewport and display the labels just in time.

With the way we do context assembly (see detail in #2769) we can't predict accurately whether we'll truncate any given file. So we could inform the user before they pick an item if we're likely going to truncate it, and after the prompt is assembled, too if we did. But we have to do that informing in a way that they can understand—one problem with the first iteration of the error banner is people thought it was a rate limit, etc. We don't want to do that again.

@toolmantim do you have ideas here?

dominiccooney (changes requested):
Some ideas on making the size check scale better inline.

vscode/src/editor/utils/editor-context.ts
abeatrix: Ahh that makes sense! I've updated to use fs.state instead and that seems to work:

image

Will continue next week, thank you for your the suggestion!

abeatrix:
@toolmantim updated to follow your design on figma:

image

vscode/webviews/UserContextSelector.tsx
abeatrix: updated the behavior based on Tim's figma design

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -544,7 +544,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         const cancellation = new vscode.CancellationTokenSource()
 
         try {
-            const MAX_RESULTS = 20
+            const MAX_RESULTS = 10
```
abeatrix, line 547: since it'd take longer to find more results, especially in large repositories, i think we should lower this number from 20 to 10 until the underlying issue is resolved (slow @-mentioned in large repositories)

abeatrix:
@dominiccooney i've updated to use `vscode.workspace.fs.stat` instead of `vscode.workspace.fs.readFile`. Can I get your review again on this change, please?

beyang:
We're approaching crunch time—WDYT if we just display a warning for now and postpone getting the truncation logic right until after launch?

abeatrix:
> We're approaching crunch time—WDYT if we just display a warning for now and postpone getting the truncation logic right until after launch?

Yea I think that'd be the best. I'll update the PR to remove the truncation logic.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
toolmantim: Would be better not to degrade this for everyone just due to the symlink performance issues… but I'm not sure if there's a better solution. Maybe at least add a todo?

toolmantim (approved):

toolmantim:
@beyang @abeatrix I think it also makes sense to include a warning at display time if something was excluded. Here's a design for that if you think it's easier/faster: https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1590-34814&mode=design&t=ZSpfaRg2b9wfCOfP-4

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
abeatrix: i've reverted the change so that this PR focuses on adding the warning for large @-files. once the fix for findFiles is merged, this shouldn't be an issue 👍

abeatrix:
> @beyang @abeatrix I think it also makes sense to include a warning at display time if something was excluded. Here's a design for that if you think it's easier/faster: https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1590-34814&mode=design&t=ZSpfaRg2b9wfCOfP-4

@toolmantim i think that makes sense too. This is for after the release, right?
(I can try, but I'm not sure if I can get the new design in by the end of tomorrow.)

umpox (approved):
Looks good! I think we can show a similar UI in the Edit quick pick too. I left a comment [here](https://github.com/sourcegraph/cody/issues/2965#issuecomment-1941644011)

I don't think this need to be done before this release but just added so we don't forget

Issue #3104, opened 2024-02-09
Cody doesn't take into account output tokens and exceeds the OpenAI context window


beyang:
### Version

v1.2.3 and probably latest

### Describe the bug

Cody's token limit takes into account only input tokens. So we get an error that we exceed the context window even when the Cody-side limit is set to a value lower than the model's context window. This has been reproduced with both Azure OpenAI and OpenAI.

There are 2 errors that appear linked to the same underlying issue:

1. [The first error](https://github.com/sourcegraph/cody/issues/2988) returns an error stating we requested too many tokens.
2. [The second error](https://sourcegraph.slack.com/archives/C05F0220UA1/p1707335397028909) states the response "resulted in" too many tokens.

### Proposed fix

- [x] Properly enforce context limit https://github.com/sourcegraph/cody/pull/3121
- [ ] Update the code so we explicitly account for `ANSWER_TOKENS`

Additional:
- Use the appropriate tokenizer locally to more precisely enforce the token limit
  - [tiktoken](https://github.com/openai/tiktoken)
  - [tiktoken fork with WASM + js bindings](https://github.com/dqbd/tiktoken)

kalanchan:
should be fixed now as we expanded output token limits

PR #3121, opened 2024-02-09
properly enforce prompt context limit

beyang:
There was a bug in the logic of `tryAddContext` that could effectively expand the overall char limit we imposed on the prompt by the number of bytes we allocated to enhanced context (60%). This contributed to https://github.com/sourcegraph/cody/issues/3104.

This change properly enforces the overall prompt limit.

## Test plan

See added unit test


chwarwick (approved):

umpox (commented):
Nice catch, thanks for adding the test!

PR #3124, opened 2024-02-10
Add the ability to run/debug integration tests+mock server from VS Code

DanTup:


.vscode/launch.json
```diff
@@ -68,7 +84,112 @@
...
+        "group": "Test",
+        "order": 1
+      }
+    },
+    {
+      // This config runs the integration tests in debug mode.
+      // This config itself is hidden and should not be used directly because the integration
+      // tests depend on the MockServer. The "Debug Integration Tests" compound config
+      // will launch both this and the mock server together (and allow debugging both).
+			"type": "extensionHost",
```
dominiccooney, line 98: Let's keep the indentation consistent here.

.vscode/launch.json
```diff
@@ -68,7 +84,112 @@
...
+			"request": "launch",
+			"name": "Debug Current Integration Test (Single-root impl)",
+			"preLaunchTask": "Build Tests",
+			"runtimeExecutable": "${execPath}",
+			"skipFiles": ["<node_internals>/**", "**/node_modules/**"],
+			"args": [
+				"${workspaceFolder}/vscode/test/fixtures/workspace",
+				"--extensionDevelopmentPath=${workspaceFolder}/vscode",
+				"--extensionTestsPath=${workspaceFolder}/vscode/dist/tsc/test/integration/single-root",
```
dominiccooney, line 107: What do "roots", single and multi, refer to here? Where do the single-root, multi-root output folders come from?

.vscode/launch.json
```diff
@@ -68,7 +84,112 @@
...
+				"--profile-temp"
+			],
+			"outFiles": ["${workspaceFolder}/vscode/dist/**/*.js"],
+			"smartStep": true,
+      "env": {
+        "CODY_TESTING": "true",
+        "MOCK_RG_PATH": "true",
```
dominiccooney, line 114: ripgrep is gone, so if you're in the mood to clean up, you should be able to remove the MOCK_RG_PATH. Sorry I missed these...

dominiccooney (changes requested):
On the whole, having more stuff easily debuggable sounds good... if we can make it share enough with CI that the rate of people using it exceeds the rate of decay.

Do the tests wait for the server to be up before continuing?

.vscode/launch.json
DanTup: Done!

.vscode/launch.json
DanTup: Oops, fixed. In my editor the tabs were the same size so I didn't notice it was mixed. All spaces now.

.vscode/launch.json
DanTup: The integration tests are split into two now, some that run with a single folder open ("single-root") and some that run with a `.code-workspace` with multiple folders open ("multi-root"). This is to ensure we can easily test both (because we can't change workspace folders during a test run, because it recreates the extension host).

- single-root tests are at https://github.com/sourcegraph/cody/tree/main/vscode/test/integration/single-root
- single-root tests open the folder https://github.com/sourcegraph/cody/tree/main/vscode/test/fixtures/workspace
- multi-root tests are at https://github.com/sourcegraph/cody/tree/main/vscode/test/integration/multi-root
- multi-root tests open https://github.com/sourcegraph/cody/blob/main/vscode/test/fixtures/multi-root.code-workspace which includes both the `workspace` and `workspace2` folders

The CLI run handles both here:

https://github.com/sourcegraph/cody/blob/c823e8ca561b29107c6332827b12ff475b1fa998/vscode/test/integration/main.ts#L28-L32

Unfortunately, there is duplication here between launch.json and the CLI run, but unfortunately I don't know of a way to avoid this (besides perhaps a little code-gen to generate both from a single source).

PR #3136, opened 2024-02-12
Autocomplete: Log if a completion was suggested again

philipp-spiess:
Add a span event when a completion was already logged as suggested. Mainly to ensure we have attributes attached.

Noticed a few spans with source `null` in Honeycomb. 

## Test plan

- Ensured that this is the codepath taken via adding debugger breakpoints.


umpox (approved):

valerybugakov (commented):
Nice find!

PR #3138, opened 2024-02-12
Autocomplete: Don't delay completions for cached entries

philipp-spiess:
Right now, about 15% of all suggested completions originate from our cache (actually quite nice considering we are aggressively invalidating the cache e.g. when rejecting completions). 

However, I noticed that these have a P75 duration of ~125ms which is caused by the debounce time + retrieval time. We can check the cache before we do any of this work though to remove the latency of cached entries completely.

## Test plan

- Updated the unit tests
- Stepped into it with a debugger to see the right breakpoints are triggered


vscode/src/completions/request-manager.test.ts
```diff
@@ -205,6 +194,20 @@ describe('RequestManager', () => {
         provider2.yield(["'world')"])
     })
 
+    describe('cache', () => {
+        it('resolves a single request with a cached value', async () => {
```
valerybugakov, line 198: It might be helpful to highlight the use case this test covers in case we need to update it later: 

```suggestion
        it('resolves a single request with a cached value without waiting for the debounce timeout', async () => {
```

valerybugakov (approved):
Haha, are you a genius or what? Awesome catch!

PR #3139, opened 2024-02-12
Autocomplete: Only sample requests that are also suggested

philipp-spiess:
Thinking about #3136 again, I don't think it's helpful to sample these traces even. It's not bad per-se, but it's easier to aggregate if we don't have to filter based on a root span attribute everywhere.

E.g. I want to see the "median" completion time so I get the median timings of specific spans but if I have to exclude some values in the root span, this is getting more complex then necessary. Let's keep it simple and don't even sample these (which also saves some bandwidth)

## Test plan

- Added a debugger breakpoint to see if the sampled attribute is set

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -441,15 +441,7 @@ export class InlineCompletionItemProvider
                     this.unstable_handleDidShowCompletionItem(autocompleteItems[0])
                 }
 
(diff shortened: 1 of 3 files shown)


valerybugakov (approved):
Unblocking this one since it makes it easier to review our traces, but I didn't understand the use case fully. Could you share a Honeycomb query that's harder than it should be with the previous approach? Or add more context to the problem in some other way.

PR #3142, opened 2024-02-12
Agent: automate generation of Kotlin bindings

olafurpg:
Until now, we have been manually writing Kotlin bindings for the agent protocol. This worked fine when only a small number of people worked on the protocol and the protocol was small, but has become increasingly problematic as more people are actively working on the JetBrains plugin and the size of the Agent protocol has grown. For example, if you make a mistake in the bindings, you may end up spending a long time debugging why something is not working because the error messages can be cryptic.

This PR adds a new script to automatically generate Kotlin bindings so that we no longer need to rely on manually written bindings. Simplified, the script works like this:

- Index the codebase with a fork of scip-typescript. This fork uses an extension of SCIP that adds `SymbolInformation.signature` (more details here https://github.com/sourcegraph/scip/pull/231).
- Use the structured type signatures from the SCIP index, translate it into Kotlin code, and report errors if something in the protocol has no clean translation to Kotlin.
- Add the generated code to the Cody repo so that we can closely track what Cody PRs are influencing the protocol, and how they're influencing it.

## Test plan

Green CI


olafurpg:
Didn't open a draft PR, but I'm going to review this myself tomorrow before tagging for reviews.

philipp-spiess (approved):
Great stuff! The changes on the `vscode` side are minuscule and it's very easy to turn this off if it causes problems but even in its current state this is a huge improvement for the work on the JetBrains extension (even if we don't keep the types updated) so let's go

PR #3147, opened 2024-02-13
Chat: Fix input focus issue

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1707784216516929

Fix an issue with editor focus being stolen by webview when webview is not focused.

This change updates the Chat.tsx component to pass the isWebviewActive prop to the inputCaretPosition and isFocusd props, which are used for setting the focus to the chat input box, respectively. 
This PR updated to not set chat input focus when isWebviewActive is not true, meaning the editor is currently focused.

## Test plan

All bots are green.

1. put your cursor to an open file after sending a question to Cody
2. once Cody has responded, make sure the cursor in the editor is still active

https://github.com/sourcegraph/cody/assets/68532117/574bc59b-bb40-4aa0-8211-b402165d360a


abeatrix:
> Ah, this could be a good quick fix (though from an interaction POV I'd prefer it if we only changed keyboard focus if "Edit Message" had focus, which means the user didn't move and continue doing other things… even within the same webview).
> 
> This fix does seem to have this bug now though:
> 
>  CleanShot.2024-02-13.at.12.46.51.mp4

ah i forgot to remove the uEffect dependence. I think the latest commit should work now, but will test it more thoroughly tomorrow to confirm

toolmantim:
I gave this a quick test, but doesn't seem to work:

https://github.com/sourcegraph/cody/assets/153/472d93ca-860b-4fbb-acc7-2d50bf24021c

abeatrix:
That helps! See the latest update:

https://github.com/sourcegraph/cody/assets/68532117/670b0b9c-7df4-4bc4-ac9b-5693b1288667

Do you know if we can use e2e test for this?

toolmantim (approved):
This seems to work great now!

There's still one improvement that I would have loved us do, which is to only_refocus the message input if "Edit Last Message" has focus. If the user is doing other things in the webview we shouldn't interrupt them (like it does below):

https://github.com/sourcegraph/cody/assets/153/46335604-6f35-4d81-bca1-081d3327e0af

dominiccooney:
> Is there a way to pause or slow down the LLM response in the e2e test perhaps?

Yes, that sounds useful. You'll need to hack on vscode/test/fixtures/mock-server.ts and change the `app.post('/.api/completions/code', (req, res) => { ...` bit.

One cheesy but effective technique is to put "instructions" to the server in the request. For example you could send it some JavaScript in some marker tags in the chat, and then `await new Function(str)(... pass whatever you need ...)` (essentially, `eval` but with arguments) in the server. Or you could make a DSL or whatever.

Second idea is to actually have the mock server turn around and use/hand out Promises so you can precisely stage when things happen. You can look at how the mock graphql server is set up and used for some ideas.

Or you could do some combination of these two.

abeatrix:
> There's still one improvement that I would have loved us do, which is to only_refocus the message input if "Edit Last Message" has focus.

Makes sense! Do you mind filing an issue so i can look into that after i'm done with my current issue?

PR #3149, opened 2024-02-13
Autocomplete: Split debounce into two chunks, race the second part with the context retrieving

philipp-spiess:
This PR splits the debounce time into two logical chunk. The first 25ms (regardless of how long the debounce time is defined), the behavior will be like it is now: We batch events and don't do any concurrent work.

However after 25ms (which, remember, was the debounce time we had until recently), we start to fetch the context in parallel. The idea is that we have shave off up to 50ms from the context retrieval time.

Here's a visualization of a trace:

Screenshot 2024-02-13 at 11 37 12

As you can see in this specific example, context retrieval could almost be free.

One caveat here is that this _will_ increase CPU pressure (since we start fetching context a bit earlier). That's why I think 25ms is a good value for this since we have used it in the past with the same retrieval algorithm quite successfully. 

## Test plan

- Ensure AC still works
- Check out a trace

Screenshot 2024-02-13 at 11 47 14


vscode/src/completions/get-inline-completions.ts
```diff
@@ -426,3 +448,7 @@ function createCompletionProviderTracer(
         }
     )
 }
+
+async function sleep(ms: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, ms))
+}
```
valerybugakov, line 454: I added `sleep` to `vscode/src/completions/test-helpers.ts` recently. Does it make sense to share from the generic utility file?

valerybugakov (approved):
Nice!

vscode/src/completions/get-inline-completions.ts
philipp-spiess: Yep will fix

PR #3150, opened 2024-02-13
Commit Messages: Fix bug with unstaged files and add feature toggle to menu

umpox:
## Description

This PR:
- Fixes an issue where VS Code would provide `changes` as an empty array for unstaged files
- Adds a e2e test to cover this
- Adds an item to the Cody feature menu to enable/disable the setting

## Test plan

1. Trigger the Cody commit message with no staged files
2. Check it works!

Other:

1. Check that disabling/enabling the feature from the menu works

### DIFF
--- a/vscode/src/scm/CommitMessageProvider.ts
+++ b/vscode/src/scm/CommitMessageProvider.ts
@@ -63,7 +63,12 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
             hasV2Event: true,
         })
         telemetryRecorder.recordEvent('cody.command.generateCommitMessage', 'clicked')
-        const humanPrompt = await this.getHumanPrompt(changes)
+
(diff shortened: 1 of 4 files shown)


vscode/src/scm/CommitMessageProvider.ts
```diff
@@ -201,13 +206,17 @@ export class CommitMessageProvider implements VSCodeCommitMessageProvider, vscod
         }
 
         if (!diffs?.length) {
+            console.log('Getting from cli')
```
philipp-spiess, line 209: leftover console.log

philipp-spiess (approved):

vscode/src/scm/CommitMessageProvider.ts
umpox: Thanks :)

PR #3153, opened 2024-02-13
`chat/restore`: set default llm for null `modelID` param

Sa1to:
It fixes https://github.com/sourcegraph/jetbrains/issues/620
## Test plan
Restore chat from history with null model 
(There is automated test in index.test.ts)

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -9,8 +9,8 @@ import * as vscode from 'vscode'
 import {
     type BillingCategory,
     type BillingProduct,
-    convertGitCloneURLToCodebaseName,
     FeatureFlag,
+    ModelProvider,
     NoOpTelemetryRecorderProvider,
     convertGitCloneURLToCodebaseName,
     featureFlagProvider,
@@ -20,8 +20,6 @@ import {
     isRateLimitError,
     logDebug,
     logError,
-    ModelProvider,
-    NoOpTelemetryRecorderProvider,
     setUserAgent,
 } from '@sourcegraph/cody-shared'
 import type { TelemetryEventParameters } from '@sourcegraph/telemetry'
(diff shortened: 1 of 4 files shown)


agent/src/index.test.ts
```diff
@@ -302,6 +302,44 @@ describe('Agent', () => {
             )
         }, 30_000)
 
+        it('chat/restore (With null model)', async () => {
+            // Step 1: create a chat session where I share my name.
```
pkukielka, line 306: Incorrect comment?

agent/src/agent.ts
```diff
@@ -710,7 +712,10 @@ export class Agent extends MessageHandler {
         })
 
         this.registerAuthenticatedRequest('chat/restore', async ({ modelID, messages, chatID }) => {
-            const chatModel = new SimpleChatModel(modelID, [], chatID)
+            const chatModel =
+                modelID === null
+                    ? new SimpleChatModel(ModelProvider.get(ModelUsage.Chat).at(0)!.model, [], chatID)
```
pkukielka, line 717: Unfortunately looks like it's not so simple. Look at:
cody/vscode/src/models/index.ts, function `getModel`.
Looks like for pro users default model is the last one used. 
I think we can simplify a bit for now, but that logic (and logic in our client) won't be 100% correct :(

olafurpg (commented):
@szymonprz FYI, you can run `./agent/scripts/reset-recordings-to-main.sh` to reset the recordings and start from scratch. The very large diff indicates that something may be wrong in the recordings. The recording file should only have additions for the newly added test case.

agent/src/agent.ts
olafurpg, line 717: It's always better to report a helpful error than to use `!`

```
const model = ModelProvider.get(ModelUsage.Chat).at(0)?.model
if (!model) { throw new Error( ... ) }
...
```

agent/src/agent.ts
```diff
@@ -710,7 +712,10 @@ export class Agent extends MessageHandler {
         })
 
         this.registerAuthenticatedRequest('chat/restore', async ({ modelID, messages, chatID }) => {
-            const chatModel = new SimpleChatModel(modelID, [], chatID)
+            const chatModel =
+                modelID === null
```
olafurpg, line 716: We should always handle both null and undefined
```suggestion
                modelID === null || modelID === undefined
```

It's normally easier to do do the condition like this
```
const foo = modelID ? ... : /* when null or undefined */
```

olafurpg: Note that `modelID?: string` does not allow `null` values, you need to type it as `modelID?: string | null` to make it 100% correct. However, we don't do this consistently everywhere

pkukielka (approved):
LGTM

PR #3154, opened 2024-02-13
Command Hints: Update setting enablement

umpox:
## Description

1. Defaults the setting to on, as visible to the user. See comment below as to why that is important
2. Adds default text to the setting description, so users can easily search VS Code settings to disable this if they want to.

## Test plan

**When `cody.commandHints.enabled` setting is not set:**

1. Set `cody-command-hints` flag to 100%. Check command hints are enabled. Check value is correctly shown in "Cody Settings" menu. 
2. Set `cody-command-hints` flag to 0%. Reload and check command hints are NOT enabled. Check value is correctly shown in "Cody Settings" menu. 

**When `cody.commandHints.enabled` setting is already set to `true`:**

1. Set `cody-command-hints` flag to 0%. Reload and check command hints are STILL enabled. Check value is correctly shown in "Cody Settings" menu. 

**When `cody.commandHints.enabled` setting is already set to `false`:**

1. Set `cody-command-hints` flag to 100%. Check command hints are NOT enabled. Check value is correctly shown in "Cody Settings" menu.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -821,8 +821,8 @@
                     "order": 12,
                     "title": "Cody Command Hints",
                     "type": "boolean",
-                    "markdownDescription": "Enable hints for Edit and Chat shortcuts, displayed alongside editor selections.",
-                    "default": false
+                    "markdownDescription": "Enable hints for Edit and Chat shortcuts (\"Opt+K to Edit, Opt+L to Chat\"), displayed alongside editor selections.",
+                    "default": true
                 },
                 "cody.experimental.simpleChatContext": {
                     "order": 99,
--- a/vscode/src/commands/GhostHintDecorator.ts
+++ b/vscode/src/commands/GhostHintDecorator.ts
@@ -39,12 +39,16 @@ export async function getGhostHintEnablement(): Promise<boolean> {
     const config = vscode.workspace.getConfiguration('cody')
     const configSettings = config.inspect<boolean>('commandHints.enabled')
 
-    // Return the actual configuration setting, if set. Otherwise return the default value from the feature flag.
-    return (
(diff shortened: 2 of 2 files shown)


vscode/package.json
```diff
@@ -821,8 +821,8 @@
                     "order": 12,
                     "title": "Cody Command Hints",
                     "type": "boolean",
-                    "markdownDescription": "Enable hints for Edit and Chat shortcuts, displayed alongside editor selections.",
-                    "default": false
+                    "markdownDescription": "Enable hints for Edit and Chat shortcuts (\"Opt+K to Edit, Opt+L to Chat\"), displayed alongside editor selections.",
+                    "default": true
```
umpox, line 825: VS Code will always default a `boolean` setting to `false`. 

We don't respect the default here, as we override it ourselves depending on the A/B test.

There is a chance that some users will see this text and immediately search VS Code settings to see if they can disable it. If this is defaulted to `false` already then it will be confusing how they can disable it already. If we default to `true`, then users can just uncheck the box in this scenario and move on.

The other fix would just be to use a string enum: "default", "disabled", "enabled". That would work too but would mean we'd have to deal with the issue of maintaining or migrating that setting when we really just want a boolean.

abeatrix (approved):
Makes sense to me! 
Should we disable the Command Code Lenses by default now we have the Command Hints? Just worried users will get overwhelmed by Cody in their editor lol

umpox:
@abeatrix Do you mean these: 
image

I think it's already off by default so should be safe for now.

We should consider a way where users can set their preference when they first install Cody, like in the onboarding walkthroughx:
- Enable Code lens
- Disable Command hints

Food for thought, @toolmantim

PR #3156, opened 2024-02-13
Auth: Add input validation for instance URL

abeatrix:
This commit adds validation for sourcegraph tokens in the AuthMenus and AuthProvider files. The showInstanceURLInputBox function in AuthMenus now checks if the user is entering a token as a URL and displays an error message if so. 
Similarly, the formatURL function in AuthProvider now throws an error if the URI is a sourcegraph token, and return `null`.

 Additionally, the LocalStorageProvider file now ignores and clears the last used endpoint if the provided endpoint is a sourcegraph token.

image

Update placeholder value for URL to always starts with `https://` to make it clear that the field is for URL input:

image

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/9d44427a-3b8b-4c35-812d-aa5b22120d6d

### DIFF
--- a/vscode/src/services/AuthProvider.ts
+++ b/vscode/src/services/AuthProvider.ts
@@ -31,6 +31,9 @@ import { telemetryRecorder } from './telemetry-v2'
 type Listener = (authStatus: AuthStatus) => void
 type Unsubscribe = () => void
 
+const sourcegraphTokenRegex = /sgp_[a-zA-Z0-9]+_[a-zA-Z0-9]+/
+export const isSourcegraphToken = (text: string): boolean => sourcegraphTokenRegex.test(text)
+
 export class AuthProvider {
     private endpointHistory: string[] = []
 
@@ -447,21 +450,30 @@ export function isNetworkError(error: Error): boolean {
 }
 
 function formatURL(uri: string): string | null {
-    if (!uri) {
-        return null
-    }
-    // Check if the URI is in the correct URL format
-    // Add missing https:// if needed
-    if (!uri.startsWith('http')) {
-        uri = `https://${uri}`
-    }
     try {
+        if (!uri) {
+            return null
+        }
+
+        // Check if the URI is a sourcegraph token
+        if (isSourcegraphToken(uri)) {
+            const errorMsg = 'Token is not a valid URL'
+            void vscode.window.showErrorMessage(errorMsg)
+            throw new Error(errorMsg)
+        }
+
+        // Check if the URI is in the correct URL format
+        // Add missing https:// if needed
+        if (!uri.startsWith('http')) {
+            uri = `https://${uri}`
+        }
+
         const endpointUri = new URL(uri)
         return endpointUri.href
-    } catch {
-        console.error('Invalid URL')
+    } catch (error) {
+        console.error('Invalid URL: ', error)
+        return null
     }
-    return null
 }
 
 async function showAuthResultMessage(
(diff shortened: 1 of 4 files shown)


vscode/src/services/AuthProvider.ts
```diff
@@ -31,6 +31,9 @@ import { telemetryRecorder } from './telemetry-v2'
 type Listener = (authStatus: AuthStatus) => void
 type Unsubscribe = () => void
 
+const sourcegraphTokenRegex = /sgp_[a-zA-Z0-9]+_[a-zA-Z0-9]+/
```
willdollman, line 34: Sorry should've mentioned on the call - we have [several secret formats](https://docs.sourcegraph.com/dev/security/secret_formats), and should aim to match v2 and v3. A better regex here would be `^sgp_[a-zA-Z0-9]+` as it should match all formats

dcomas, line 34: @abeatrix this does not look to be catching the old format. Does everyone uses the new token format only? Can we include the ones from this list? cc:@willdollman @shivasurya

willdollman: Broadly no (e.g. `http://localhost`) and slight concern that some edge case customer will have a crazy dns setup that this would break.

What about ensuring the url matches this: `([.]|^https?:\/\/)`

Matches anything containg a dot or anything starting with `https?://`
Matches:
```
https://sourcegraph.com
http://localhost
sourcegraph.com
http://customersourcegraph
```

Doesn't match:
```
sgp_abcd_foobar
sgp_abcd
sgp_foobar
foobar
```

willdollman: I'm making the assumption that we automatically add the missing https:// if a user enters a url without a prefix (`sourcegraph.com`)

unknwon: LGTM 👍

abeatrix: > I'm making the assumption that we automatically add the missing https:// if a user enters a url without a prefix (sourcegraph.com)

@willdollman yea i added the check before we add the `https://` prefix: https://github.com/sourcegraph/cody/blob/77114725ef81e83458ad271766696da9362119c1/vscode/src/services/AuthProvider.ts#L463-L465

But that's a good callout, I'll make sure to include those cases

abeatrix: i just updated the regex for catching token:

Sourcegraph Access Token (v3): 
image

Sourcegraph Access Token (v2, deprecated):
image

Sourcegraph Access Token (v1, deprecated):
image

Catch tokens with http prefix:

image

image

image

Also updated the check to ensure the URL must starts with `http` to reduce confusion:
image
image
image

willdollman:
Validating new tokens lgtm - thanks!

@abeatrix are you able to also validate that if a user _already_ has a token saved in local storage as the endpoint, will this update result in the token being purged? I see it will purge via `saveEndpoint` but is that called if the endpoint is already stored and just being read?

willdollman:
Yep looks good! My concern was whether the bad endpoint is (eventually) removed from local storage and I see in `auth()` that we call `this.storeAuthInfo(endpoint, token)` which will overwrite `endpoint` with the empty string returned by `formatURL`.

> @willdollman we call the [auth function](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/AuthProvider.ts?L65) when we first get the stored endpoint from the local storage, and then first thing we do in the auth function is to run the `formatURL` function on the endpoint which now will return `null` if the store endpoint is a token. Does this address your concern?

willdollman (approved):
lgtm. Thank you for the quick work on this! 🙏

PR #3161, opened 2024-02-13
Revert "feat: Add experimental commit message generator (#2306)"

abeatrix:
This reverts commit 80dbffa0bb0e2336e1362bfdbac66e36d6c7e02b.

Context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1707861341879059?thread_ts=1707832350.281989&cid=C05AGQYD528

Unfortunately, the API used by the feature has been removed from VS Code and no longer works in the stable version of VS Code: 
- https://github.com/microsoft/vscode/pull/200012
- https://github.com/microsoft/vscode/blob/a776d5fe0a401a0b7dc3b58f28eef1887ad0c15e/extensions/git/src/api/api1.ts#L348-L353

cc @RXminuS we will add it back once the API has become publicly available again!

## Test plan

image


kalanchan (approved):
ah dang, yeah let's revert this for now and see what the API gets replaced with

RXminuS:
Hahahaha, typical 😅

PR #3164, opened 2024-02-14
Add local certs when running agent on macOS

chwarwick:
Some enterprise setups have proxies that are using custom certificates, that may already be available on the users machine.  For users on mac systems, this will add root CAs to the http agent.
## Test plan
CI
Manually tested against an sample enterprise instance with a selfsigned cert that was added to my local keychain.
Chat & Autocomplete and telemetry all working.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -129,6 +129,9 @@ importers:
       js-levenshtein:
         specifier: ^1.1.6
         version: 1.1.6
+      mac-ca:
+        specifier: ^2.0.3
+        version: 2.0.3
       minimatch:
         specifier: ^9.0.3
         version: 9.0.3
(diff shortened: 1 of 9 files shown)


vscode/src/services/open-telemetry/OpenTelemetryService.node.ts
```diff
@@ -75,7 +75,9 @@ export class OpenTelemetryService {
 
         // Add the default tracer exporter used in production.
         this.tracerProvider.addSpanProcessor(
-            new BatchSpanProcessor(new OTLPTraceExporter({ url: traceUrl }))
+            new BatchSpanProcessor(
+                new OTLPTraceExporter({ url: traceUrl, httpAgentOptions: { rejectUnauthorized: false } })
```
chwarwick, line 79: Doesn't appear to be any options to set the agent here

pnpm-lock.yaml
```diff
@@ -10553,6 +10556,12 @@ packages:
     resolution: {integrity: sha512-Pn9cox5CsMYngeDbmChANltQl+5pi6XmTrraMSzhPmMBbmgcxmqWry0U3PGapCU1yB4/LqCcom7qhHZiF/jGfQ==}
     dev: false
 
+  /mac-ca@2.0.3:
```
slimsag, line 10559: When I was digging into this I noticed that Copilot appears to use these deps:

```
        "@roamhq/mac-ca": "^1.0.7",
        "windows-ca-certs": "^0.0.3",
```

And that VS Code uses `@vscode/windows-ca-certs@^0.3.1` instead of `windows-ca-certs`

---

Should we be using `@roamhq/mac-ca` instead of `mac-ca`? Should we also be using one of the windows CA ones too?

slimsag (approved):

pnpm-lock.yaml
chwarwick: I'm going to handle windows in a 2nd PR.  

@roamhq/mac-ca appears to be a fork that is appears to be unmaintained.

vscode/src/services/open-telemetry/OpenTelemetryService.node.ts
philipp-spiess: Wait this shouldn't even be enable unless the `cody-tracing` feature flag is turned on on the instance? 😱 
See line 45

philipp-spiess (approved):

PR #3166, opened 2024-02-14
Command hints: Experiment with always on

umpox:
## Description

Changes:
- More opacity on the ghost text
- More indentation before the ghost text
- Do not show the ghost text on the last line that the user typed on
- Show "Opt+K to Generate" on empty lines

### DIFF
--- a/vscode/src/commands/GhostHintDecorator.ts
+++ b/vscode/src/commands/GhostHintDecorator.ts
@@ -3,6 +3,7 @@ import * as vscode from 'vscode'
 import type { AuthProvider } from '../services/AuthProvider'
 import type { AuthStatus } from '../chat/protocol'
 import { FeatureFlag, featureFlagProvider } from '@sourcegraph/cody-shared'
+import { isGenerateIntent } from '../edit/utils/edit-intent'
 
 const EDIT_SHORTCUT_LABEL = process.platform === 'win32' ? 'Alt+K' : 'Opt+K'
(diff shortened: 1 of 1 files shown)


vscode/src/commands/GhostHintDecorator.ts
```diff
@@ -132,18 +152,30 @@ export class GhostHintDecorator implements vscode.Disposable {
...
     }
 
-    private setGhostText(editor: vscode.TextEditor, position: vscode.Position): void {
+    private setGhostText(editor: vscode.TextEditor, position: vscode.Position, editVerb = 'Edit'): void {
         this.activeDecoration = {
             range: new vscode.Range(position, position),
             renderOptions: {
-                after: { contentText: `${EDIT_SHORTCUT_LABEL} to Edit, ${CHAT_SHORTCUT_LABEL} to Chat` },
+                after: {
+                    contentText: `\u00a0\u00a0\u00a0${EDIT_SHORTCUT_LABEL} to ${editVerb}, ${CHAT_SHORTCUT_LABEL} to Chat`,
```
toolmantim, line 177: This extra indentation looks good on lines with code, but not so much on empty lines

toolmantim:
I think showing it on the empty line is great! Showing it on every line, even without selection, perhaps too much?

I was hoping generate would work on a completely new file, but doesn't seem to show up?

Weird thing about "Generate" being the label is that all the menus after you press it use "Edit" 🤔

PR #3168, opened 2024-02-14
Chat: Add tracing

philipp-spiess:
As the title suggests. Please review with [whitespace changes hidden](https://github.com/sourcegraph/cody/pull/3168/files?w=1).

## Test plan

e.g.: https://ui.honeycomb.io/sourcegraph/environments/cody/datasets/cody-client/result/oy1kwHxdSmP/trace/9xmczeHgPAu?fields[]=s_name&fields[]=s_serviceName&span=6dc03ea5de6419ae

Screenshot 2024-02-14 at 16 28 44


abeatrix:
WOOOO

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -382,88 +386,103 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         userContextFiles: ContextFile[],
         addEnhancedContext: boolean
     ): Promise<void> {
-        if (inputText.match(/^\/reset$/)) {
-            return this.clearAndRestartSession()
-        }
+        return tracer.startActiveSpan('chat.submit', async (span): Promise<void> => {
```
abeatrix, line 389: noob question: should we start the active span from when the question is submitted (from the command menu, chat view, or commands), or we only want to trace when we handle the request that was sent here?
If i want to add the tracing from when the command was executed, so we can trace how long it takes before it arrives to this function, do i just add another active span to that?

philipp-spiess: Ideally as early as possible but when the trace is started it does not happen in the same JavaScript thread (it happens in the UI thread and not the extension host) so I put it here instead. Do you think there is a long delay until it arrives here?

philipp-spiess: Ah, yeah for the simple chat provider I think this function is called pretty much immediately after the webview message arrives. Context fetching happens later

abeatrix (approved):
🚀  thanks for setting this up again!!

PR #3174, opened 2024-02-14
Chat: support line numbers in at-files

abeatrix:
feat: enable users to add line number to @-file queries

image

atLines

demoo

Key changes:
- Extract input parsing logic into new inputParser.ts file. The module parses line numbers after file names in the input string. It sets the parsed range on the context file objects.
- Remove separate display text generation logic for selections. Instead always pass file range to create markdown links. This allows links to point directly to user-provided line number context when available.
- Covered by unit tests.

These changes enable users to add line number context to their @-mentioned queries by appending ':line-range' after file names. The line range will be parsed and used to generate more precise code links.

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/1cf80528-7f72-42a9-b7ef-8d61ecb73157

### After

image

Loom: https://www.loom.com/share/28cadce2a91f4e5892d3ee3a7b6f82dc

On Edit:

https://github.com/sourcegraph/cody/assets/68532117/aec2c517-071b-4ebb-944c-2497d247c911

### Before

image

## Next

Implement this with the default chat commands.


toolmantim (commented):
Nice! This gets us the building block we need for making commands explicitly show the selected lines.

A few bugs:
* It seems to break if you click on items in the popover, vs hitting enter with the keyboard
* If you use the up arrow to load the last message input doesn't seem to attach @-mentioned files either.

To make the experience for manually typing it in better… it would be good to not have the popover disappear, as it makes it feel like it's not going to work. Can we keep it open, and show some 50% opacity helper text as they type?

https://github.com/sourcegraph/cody/assets/153/07a51bd0-11ef-431b-9865-d47510f37150

abeatrix:
@toolmantim I can't reproduce the first issue you mentioned. Can you try the latest commit and see if it's still broken for you?

https://github.com/sourcegraph/cody/assets/68532117/ddeeab2a-d1ac-4333-890c-67e3c81997fc

- [] It seems to break if you click on items in the popover, vs hitting enter with the keyboard
- [x] If you use the up arrow to load the last message input doesn't seem to attach @-mentioned files either.
- [x] Added the helper:

demoo

abeatrix:
@toolmantim I've updated the PR to support single line range:

https://github.com/sourcegraph/cody/assets/68532117/1fe2896e-87ea-4b62-aa31-3da1a72ba555

We can merge this after tomorrow release if we want to polish this more

PR #3178, opened 2024-02-15
Require a timeout when getting a loading lease

philipp-spiess:
This fixes cases like https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1707987812090509 where a loading lease that is never returned would be stuck forever.

## Test plan

- Remove the call of `stopLoading()` in the completions code
- Trigger a completion and wait 30sec.
- The loading indicator now stops


abeatrix (approved):

olafurpg (approved):
Thank you!

PR #3179, opened 2024-02-15
Custom Command:  set userConfigFile on runtime

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1707992847314209

**NOTE:** No Changelog entry needed as this is a recent change that's not included in stable yet.

Previously, userConfigFile was initialized using this.configFileName but did not get updated when this.configFileName changed at runtime (init()). Now it is set dynamically based on the runtime config.

workspaceConfigFile was already using this.configFileName dynamically. Make userConfigFile consistent by also setting it dynamically based on runtime config.

This allows custom commands to be configured separately for the agent and editor.

### Note

We use separate config files (.cody/commands.json and .vscode/cody.json) for the Cody agent and VS Code for configuration Custom Commands right now, with a plan to migrate the one VS Code used into `.cody/commands.json` in the near future.

## Test plan

- Added new e2e test coverage to make sure the correct file name is used for both workspace and User settings.
- All the current tests on agent are still passing.

###.Step to reproduce:

Open the Custom Commands Menu from the sidebar to select `Configure Custom Commands`:

image

Then `Open User Settings`:

image

After, It should open `cody.json` for you:

image

Before, it would open `commands.json`:

image

### DIFF
--- a/vscode/src/commands/services/custom-commands.ts
+++ b/vscode/src/commands/services/custom-commands.ts
@@ -30,8 +30,9 @@ export class CustomCommandsManager implements vscode.Disposable {
     public customCommandsMap = new Map<string, CodyCommand>()
     public userJSON: Record<string, unknown> | null = null
 
-    protected configFileName = ConfigFiles.COMMAND
-    private userConfigFile = Utils.joinPath(URI.file(userHomePath), this.configFileName)
+    // Configuration files
+    protected configFileName
+    private userConfigFile
     private get workspaceConfigFile(): vscode.Uri | undefined {
         const workspaceRoot = vscode.workspace.workspaceFolders?.[0]?.uri
         if (!workspaceRoot) {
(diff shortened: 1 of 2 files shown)


vscode/src/commands/services/custom-commands.ts
```diff
@@ -30,8 +30,9 @@ export class CustomCommandsManager implements vscode.Disposable {
     public customCommandsMap = new Map<string, CodyCommand>()
     public userJSON: Record<string, unknown> | null = null
 
-    protected configFileName = ConfigFiles.COMMAND
-    private userConfigFile = Utils.joinPath(URI.file(userHomePath), this.configFileName)
```
abeatrix, line 34: root cause: this wasn't getting updated again after init runtime

vscode/test/e2e/command-custom.test.ts
```diff
@@ -221,7 +221,7 @@ test.extend<ExpectedEvents>({
         'CodyVSCodeExtension:menu:command:custom:clicked',
         'CodyVSCodeExtension:menu:command:config:clicked',
     ],
-})('open and delete cody.json from the custom command menu', async ({ page, sidebar }) => {
+}).only('open and delete cody.json from the custom command menu', async ({ page, sidebar }) => {
```
umpox, line 224:
```suggestion
})('open and delete cody.json from the custom command menu', async ({ page, sidebar }) => {
```

umpox (approved):

vscode/test/e2e/command-custom.test.ts
abeatrix: thank you!

PR #3184, opened 2024-02-15
Agent: sort custom command context only

abeatrix:
Just a small follow-up from https://github.com/sourcegraph/cody/pull/3089

As discovered by Olaf in the aforementioned PR, the context we used for the custom command tests were not always in the same order when running on CI (especially on Windows), causing the tests to fail with a `recordings not found` error. As a workaround, we added a condition to sort all chat context for agent tests, which might affect the purpose of other tests that utilize enhanced context as those are ranked priorly. 

This PR is to remove the workaround that would affect all the tests that use chat context and add the workaround to context from custom commands context runtime instead. 

## Test plan

All the current tests should be green.

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -4458,493 +4458,6 @@ log:
         send: 0
         ssl: -1
         wait: 0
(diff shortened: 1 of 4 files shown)


olafurpg (approved):
Thank you!

PR #3192, opened 2024-02-16
Edit: Improve response reliability with different chat models

umpox:
## Description

This PR improves the `responseTransformer` (prev `contentSanitizer`) to better handle different responses from the LLM.

Specific changes:
- Better handling of XML tags, if the numbers differ from what we specified
- Markdown blocks are stripped (e.g. \`\`\`typescript foo \`\`\`)

## Test plan

1. Make Edits
2. Confirm that there's no regressions in output quality


abeatrix (approved):
well done! I tested this with Claude 2.0, 2.1, instant, 3 Opus, 3 Sonnet, 3 Haiku, Ollama llama2, gpt-3.5, gpt-4, didn't spot any big issue with Edit and Doc commands!

With test commands, most of the models work fine except GPT-3.5 where it would behave strangely and open 2 files before inserting. I can reproduce on main though so not related to this PR but just wanted to flag:

https://github.com/sourcegraph/cody/assets/68532117/46f3adc4-b8c8-47a5-910a-1c8ad728bd52




I did receive this response when using ChatGPT 4 once (I can't reproduce it again):
image

Maybe we should skip all the none-code responses (responses that are not enclosed with the XML tag)?

umpox:
@abeatrix You are the best! Thank you so much for thoroughly testing this. 

> Maybe we should skip all the none-code responses (responses that are not enclosed with the XML tag)?

Hmmm, I thought we only look for the XML tag and skip if it's not there. I'll double check. There's still a potential issue where we can a non-code response in the XML tag though 🤔 

I couldn't find any other issues either so I'll merge this and look at that in a follow-up

PR #3197, opened 2024-02-16
Agent: print out helpful error message on missing recording

olafurpg:
Previously, it could be difficult to debug why an agent test case was failing due to missing HTTP recordings. In many cases, the failure happened because 1) we leaked an absolute path into the recording or 2) we forgot to sort context files before constructing the prompt. This PR is an attempt to make it easier to debug failures in this cases by

* Taking the body and URL of the HTTP request with missing recording
* Find the entry in the recording file with matching URL and closest levenshein distance from the body
* Prints out a unified diff of the multi-line formatted JSON bodies

## Test plan

Green CI.

Manually tested by making modifications to the recording file. The output looks like this

CleanShot 2024-02-16 at 15 20 02@2x


abeatrix (approved):
Nice! And thank you for fixing the sorting issue!!!

PR #3199, opened 2024-02-16
Agent: Update Cody Ignore test

abeatrix:
Update prompt that would return a more stable result from symf through the enhanced context option.

## Test plan

Check CI.

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -6097,1073 +6097,6 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: b2f825e019e5aef725edb5d2fde1ae3f
-      _order: 0
-      cache: {}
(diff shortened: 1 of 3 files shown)


DanTup:
There are a lot of changes in the recording that I don't understand - does this seem expected?

image

DanTup:
nm, the things that appear "missing" from the above seem to be further down the file. I think the diff for these recording is still not great - seems like there's still something that's not properly deterministic about them.

abeatrix:
> nm, the things that appear "missing" from the above seem to be further down the file. I think the diff for these recording is still not great - seems like there's still something that's not properly deterministic about them.

Yea i think the issue is with the results returned from symf which we should look into it in a follow-up.

> But there are some others that look odd:

Is it because the test uses less context now so the diff is shorter? Or are you talking about something else?

NOTE: I will try to merge this PR and see if it fixes the issue on main. Will revert this change if it doesn't.

PR #3200, opened 2024-02-16
Autocomplete: Remove unused feature flags and models

philipp-spiess:
Removes feature flags and Fireworks models that are no longer deployed and not used. Ships fast path though (only for dotcom and it's already rolled out to 100% anyways).

## Test plan

- it builds


lib/shared/src/inferenceClient/misc.ts
```diff
@@ -23,10 +23,6 @@ export type CompletionResponseGenerator = AsyncGenerator<CompletionResponse>
 
 export interface CodeCompletionsClient<T = CodeCompletionsParams> {
     logger: CompletionLogger | undefined
-    complete(
-        params: T,
-        abortController: AbortController,
-        featureFlags?: { fastPath?: boolean }
```
philipp-spiess, line 29: I think this was unused whoopsie

umpox (approved):

PR #3209, opened 2024-02-19
chat: do not hide snippets from the same file in context list

beyang:
The chat "context used" list was hiding snippets used in context, because there was logic that would filter out snippets from the same file.

Before:

image

After:

image

## Test plan

- Compare context menu with this change and without this change.

### DIFF
--- a/lib/ui/src/chat/components/EnhancedContext.tsx
+++ b/lib/ui/src/chat/components/EnhancedContext.tsx
@@ -36,35 +36,21 @@ export const EnhancedContext: React.FunctionComponent<{
         return
     }
 
-    const uniqueFiles = new Set<string /* uri.toString() */>()
-
-    const filteredFiles = contextFiles.filter(file => {
-        if (uniqueFiles.has(file.uri.toString())) {
-            return false
-        }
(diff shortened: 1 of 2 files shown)


philipp-spiess (dismissed):
Uh-oh. Thanks for the fix!!

lib/ui/src/chat/components/EnhancedContext.tsx
```diff
@@ -36,35 +36,21 @@ export const EnhancedContext: React.FunctionComponent<{
...
     const prefix = sparkle + 'Context: '
     // It checks if file.range exists first before accessing start and end.
     // If range doesn't exist, it adds 0 lines for that file.
-    const lineCount = filteredFiles.reduce(
+    const lineCount = contextFiles.reduce(
         (total, file) => total + (file.range ? file.range?.end?.line - file.range?.start?.line + 1 : 0),
         0
     )
-    const fileCount = filteredFiles.length
+    const fileCount = contextFiles.length
```
philipp-spiess, line 53: This still needs to be deduped. Something like this:

```suggestion
    const fileCount = new Set(contextFiles.map(file => file.uri.toString())).size
```

PR #3214, opened 2024-02-19
Commands: Include Custom Commands in main menu

abeatrix:
Default Commands Menu

image

Custom Commands Menu

image

PR to add custom commands back to the main menu as discussed with @toolmantim 

## Test plan

Open the main command menu to check if custom commands are included.

#### Before

image

### DIFF
--- a/vscode/src/commands/menus/index.ts
+++ b/vscode/src/commands/menus/index.ts
@@ -11,6 +11,7 @@ import { executeEdit } from '../../edit/execute'
 import { CodyCommandMenuItems } from '..'
 import { telemetryService } from '../../services/telemetry'
 import { telemetryRecorder } from '../../services/telemetry-v2'
+import { CustomCommandType } from '@sourcegraph/cody-shared/src/commands/types'
 
 export async function showCommandMenu(
     type: 'default' | 'custom' | 'config',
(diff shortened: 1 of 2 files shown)


vscode/src/commands/menus/index.ts
```diff
@@ -27,32 +28,40 @@ export async function showCommandMenu(
...
                 // Show keybind as description if present
                 const description = _command.keybinding ? _command.keybinding : ''
+                const type = 'default'
                 items.push({ label, command, description, type, key })
             }
         }
 
-        // The Create New Command option should show up in custom command menu only
+        // Add Custom Commands
+        items.push(CommandMenuSeperator.custom)
```
philipp-spiess, line 49: Should we check if there is at least 1 custom command before we render the separator?

vscode/src/commands/menus/index.ts
```diff
@@ -27,32 +28,40 @@ export async function showCommandMenu(
     if (type === 'config') {
         items.push(...CustomCommandConfigMenuItems)
     } else {
-        if (type === 'default') {
-            items.push(CommandMenuSeperator.commands)
+        items.push(CommandMenuSeperator.commands)
```
philipp-spiess, line 31: Does it make sense to render the seperator if `type === 'custom'` (and thus no default commands are rendered?)

PR #3219, opened 2024-02-20
Agent CodeLenses: Allow consumers to control how they render `title` icons and omit shortcut labels

umpox:
## Description

**Problem:**
- VS Code allows rendering icons (via an icon font) in code lens, e.g. `$(cody-logo) Accept`
- We cannot assume consumers of Agent will be able to render these icons in the same way
- (Other problem): We include VS Code specific shortcuts in the text of the agent lens

**Solution:**
- Extract icon syntax from title string
- Return `icons[]` with the icon value and the original position in the string
- Consumers of the agent can decide if and how they will render these icons.

**Example:**

VS Code codelens title:
`$(cody-logo) Accept`
Agent output:
```
{
  "isResolved": true,
  "range": {
    "start": {
      "line": 4,
      "character": 0
    },
    "end": {
      "line": 4,
      "character": 0
    }
  },
  "command": {
    "title": {
      "text": " Accept",
      "icons": [
        {
          "value": "$(cody-logo)",
          "position": 0
        }
      ]
    },
    "command": "cody.fixup.codelens.accept",
    "arguments": [
      "lsutws"
    ]
  }
}
```

[This is derived from the way VS Code parses these icons
](https://sourcegraph.com/github.com/microsoft/vscode/-/blob/src/vs/base/browser/ui/iconLabel/iconLabels.ts?L10) We just skip rendering these, and let the consumer decide that.

## Test plan

1. Run agent
2. Check that code lenses that include titles correctly split the title text and the icon posiitons
3. Check that the original code lens title can be reconstructed by inserting the icon positions at their respective position in the string

### DIFF
--- a/vscode/src/non-stop/codelenses/items.ts
+++ b/vscode/src/non-stop/codelenses/items.ts
@@ -4,6 +4,7 @@ import { isRateLimitError } from '@sourcegraph/cody-shared'
 
 import type { FixupTask } from '../FixupTask'
 import { CodyTaskState } from '../utils'
+import { isRunningInsideAgent } from '../../jsonrpc/isRunningInsideAgent'
 
 // Create Lenses based on state
 export function getLensesForTask(task: FixupTask): vscode.CodeLens[] {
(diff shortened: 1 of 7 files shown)


vscode/src/non-stop/codelenses/items.ts
```diff
@@ -135,9 +136,9 @@ function getFormattingSkipLens(codeLensRange: vscode.Range, id: string): vscode.
 
 function getCancelLens(codeLensRange: vscode.Range, id: string): vscode.CodeLens {
     const lens = new vscode.CodeLens(codeLensRange)
-    const shortcut = process.platform === 'darwin' ? '⌥Z' : 'Alt+Z'
+    const shortcut = isRunningInsideAgent() ? '' : ` (${process.platform === 'darwin' ? '⌥Z' : 'Alt+Z'})`
```
umpox, line 139: If we're running inside Agent, we skip including the shortcut text as this will always be VS Code specific here.

Other clients can still add shortcut hints if required

steveyegge (approved):
Thanks so much for the speedy turnaround, the elegant design, and for plumbing it all the way through!

PR #3220, opened 2024-02-20
Chat: Add flag to enable fused context

philipp-spiess:
The fused context feature was running for a while inside the `internalUnstable` config and I think it's time to ship this. To avoid breaking things, I'd like to do this with a feature flag (so we can easily killswitch this feature). 

## Test plan

- I added a breakpoint in the debugger to ensure this evaluates to the right value
- You can try it locally by enabling the `cody-chat-fused-context` feature flag on your SG instance and disable `cody.internal.unstable`


dominiccooney (approved):
Looks good to me; just ensure it makes sense when using Enterprise which has less to fuse.

philipp-spiess:
>  just ensure it makes sense when using Enterprise which has less to fuse.

Yeah I'll do some follow up tests, thanks! This flag is not enabled yet anyways: I'll still merge it so we can ship it with today's release.

PR #3232, opened 2024-02-21
Ghost text: Fire throttled display event

umpox:
## Description

Sends an event when the ghost text becomes visible.

Fires no more than once every 30 seconds to reduce telemetry spam

## Test plan

1. Make a selection, check that the event is fired
2. Make another selection immediately after, check no event
3. Make a selection after 30 seconds of waiting, check event is fired again

### DIFF
--- a/vscode/src/commands/GhostHintDecorator.ts
+++ b/vscode/src/commands/GhostHintDecorator.ts
@@ -66,18 +66,30 @@ export const ghostHintDecoration = vscode.window.createTextEditorDecorationType(
     },
 })
 
+const GHOST_TEXT_THROTTLE = 250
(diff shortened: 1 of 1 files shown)


umpox:
@chenkc805 Here's a proposal for firing display events for the ghost text. It is limited to fire no more than once every 30 seconds. I think this is sensible as:
1. It still gives us a good indication as to how often the ghost text shows
2. It avoids sending _lots_ of telemetry events every single time a user makes a selection, or changes an existing one

chenkc805:
Sounds good. Thanks Tom

abeatrix (approved):

Issue #3202, opened 2024-02-16
feedback: Allow VS Code keybindings for Cody Custom Commands


chillatom:
### Version

1.5.1708096199

### Areas for Improvement

- [ ] UI/UX
- [ ] Onboarding
- [ ] Docs
- [ ] Chat
- [X] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience.

Executing custom commands from VSC keybindings would improve the speed of getting to my most used commands. 

### Describe the solution you'd like to see

Allow me to define keybindings for custom commands.

PR #3242, opened 2024-02-21
Custom Command: supports keybinding registration

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3202

image

- All custom commands will be registered as a VS Code command that can be found in the `Keyboard Shortcuts editor`
- Added keybinding registration for custom commands in the CustomCommandsManager class 

This allows custom commands to be executed via custom keybindings in VS Code.

Also added related docs and walkthrough page.

## Test plan

1. In your cody.json file, create a custom command
2. Save the cody.json file
3. Open the `Keyboard Shortcuts` editor
image
4. In the `Keyboard Shortcuts` editor, search for `cody.command.customn.{NAME_OF_YOUR_COMMAND}`
5. Click on the `+` icon next to the command to assign a custom keybinding for your custom command
image


toolmantim:
“register” is a very generic property name, which could mean anything. Perhaps “allowKeybinding” is more suitable? Also does this property make sense for Jetbrains, or is it very VS Code specific?

Instead… why don’t we just register all custom commands by default?

toolmantim:
Do we support 1000 custom commands? I think that is beyond our design specs.

I'd suggest we register them by default and do away with the config — it's one less config to document, discover, etc.

abeatrix:
> Do we support 1000 custom commands? I think that is beyond our design specs.
> 
> I'd suggest we register them by default and do away with the config — it's one less config to document, discover, etc.

Makes sense, will update the PR tomorrow to reflect this!

toolmantim (approved):

PR #3243, opened 2024-02-22
Chat: enable editing chat command prompts

abeatrix:
enable edits on commands. Replace `the selected code` in the prompt for the explain and smell commands with @-mentioned file with selection range.

image

## Test plan

1. Run the Explain and Smell commands
2. You should see the @-mentioned file in the prompt
3. You should be able to edit the prompt

https://github.com/sourcegraph/cody/assets/68532117/41754c4a-0e2c-40ed-98ac-91e9285deb8e


toolmantim (approved):
Looks good! In the future, looking forward to making it a details element in the transcript.

PR #3244, opened 2024-02-22
Remove auth steps

### LINKED ISSUE #3216: Remove authentication steps in VS Code
Test removing unnecessary dialogues that stand between a user installing Cody and signing up for Cody. 

Context: 

- [Remove authentication steps PRD](https://docs.google.com/document/d/1AubJ226dUhaV0zC8WpA_19vzoXXtQ8AUmNqSHPaWKWQ/edit#heading=h.ftqde5tlfi9g)
- [Asana task](https://app.asana.com/0/1206519077296236/1206570325064763/f) 
- [Figma](https://www.figma.com/file/9112BsKsJc1BpO2j8XYwLL/%F0%9F%A4%96-Cody-VS-Code-%5BMain%5D?type=design&node-id=5595-162671&mode=design&t=RaJNJM7wBkyxtw55-0)

philipp-spiess:
Closes #3216 

As part of the [remove authentication steps](https://docs.google.com/document/d/1AubJ226dUhaV0zC8WpA_19vzoXXtQ8AUmNqSHPaWKWQ/edit#heading=h.ftqde5tlfi9g) PRD we want to simplify the login chain a bit.

Right now, when the VS Code Cody client starts an auth processes, it will not know about the newly generated auth token until the user navigates back to VS Code. In user testing, this proofed to be confusing an some users might tab away before going through with the final redirect.

We are working on an experiment that simplifies this. The trick? When an auth process is started, start a local http server (listening on a free port on `127.0.0.1`) and forward that to the web browser. Then, when a token is created, we can `POST` the new token to that endpoint in the hope that the VS Code client.

## Test plan

https://github.com/sourcegraph/cody/assets/458591/f2ec6977-7859-4bda-be46-8d9ba2704a67

- Opt into the experiment by setting `"cody.testing.removeAuthenticationStep": true,`
- Run a local version of Cody with this PR (The accompanying dotcom change is already rollout so no other change is necessary)
- In the dev build, click one of the big sign in buttons
- Notice that you do not have to wait for the redirect-back anymore


evict (approved):

vscode/src/services/AuthProviderSimplified.ts
```diff
@@ -34,8 +35,17 @@ async function openExternalAuthUrl(provider: AuthMethod): Promise<boolean> {
...
+    const tokenReceiver = tokenReceiverUrl
+        ? `&tokenReceiverUrl=${encodeURIComponent(tokenReceiverUrl)}`
+        : ''
+
+    // FIXME: It's a complete mystery to me why a double URL decoding is necessary for this.
+    const newTokenUrl = encodeURIComponent(
+        encodeURIComponent(
+            `/user/settings/tokens/new/callback?requestFrom=${referralCode}${tokenReceiver}`
+        )
+    )
```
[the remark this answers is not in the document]
dominiccooney: In theory, the rules (eg RFC3986) for URL handling, specifically query strings back to the beginning of time, are very simple: after the important parts like scheme, the query starts at the first `?` and ends at the next `#`. However parameters within the query string like `foo=bar` are really up to the web app. So web apps have a lot of bugs and/or latitude in handling their query strings. The good news is there's only one of these in the flow which we don't control: The SSO providers (Google, GitHub, etc.)

Which brings us to the guiding principle: *Does it work?* It's really a game of producing a string which, at each step, has the first `?` intact and hides anything (`&`, `=`) the app may mistake as its own query string format.

The game was easier for me because the URI encoding and decoding of `CODY`/`CODY_INSIDERS`/etc. is homomorphic! But you have the misfortune of introducing an `&` and a `=` in `&tokenReceiverUrl=` deep in the redirect stack.

Unrolling the stack of URLs:
- The token sink—this server you're spinning up.
- The post sign-up survey.
- The new token callback page.
- The SSO provider.
- The Sourcegraph login URL.

To *really* understand this we'd need to look carefully at each step which handles these, noting that the paths fork in a couple of places (login may skip SSO; post sign-up survey may skip the survey itself and has a mix of client and server routing; etc.) It's possible that this is an insolvable problem, and certainly a fragile situation, without server-side changes.

The double `encodeURIComponent` looks weird, but I think what's happening here is you're adding encodings that were "missing" because of the lack of `=` and `&` at the innermost value early on.

That said, JetBrains and NeoVim [solved the same problem a simpler (and more robust) way:](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@ee9199762853c10e6a737f341ceb0808ceca20ea/-/blob/client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx?L144) They encoded their port, instead of a whole URL, and separated it with a dash instead of anything a web app would interpret specially like `=` or `&`.

philipp-spiess:
@taylorsperry @chenkc805 In terms of logging, this is not using our regular experiment setup (because we need to be able to track assignment even when the user is not signed in yet). So instead this reuses the setup that @dominiccooney wrote for the simplified onboarding experiments. It will log a V1 event with the name `CodyVSCodeExtension:experiment:removeAuthenticationStep:exposed`

philipp-spiess:
@abeatrix That's fixed in main: https://github.com/sourcegraph/cody/pull/3300

vscode/src/services/OnboardingExperiment.test.ts
```diff
@@ -0,0 +1,171 @@
...
+import { localStorage } from './LocalStorageProvider'
+import * as OnboardingExperiment from './OnboardingExperiment'
+import { telemetryService } from './telemetry'
+
+import { OnboardingExperimentBranch } from './OnboardingExperiment'
+
+//vi.mock('vscode', mockVScode)
+vi.mock('./LocalStorageProvider', mockLocalStorage)
+vi.mock('../log', mockLog)
+// function mockVScode() {
```
abeatrix, line 13: leftover?

vscode/src/auth/token-receiver.ts
```diff
@@ -0,0 +1,74 @@
+import crypto from 'crypto'
+import http from 'http'
+import type { AddressInfo } from 'net'
+import { sleep } from '../completions/utils'
+
+const FIVE_MINUTES = 5 * 60 * 1000
+
+// Start a http server on a free port and generate a secure hash. When authenticated, the browser
```
abeatrix, line 8: I think it'd be helpful to explain how this is different than the`tokenCallbackHandler` set up by VS Code, wdyt?

QQ: Will this work for VS Code on web?

abeatrix (approved):
left a comment and question inline. Would be nice to have am e2e test cover the redirect machinic (if possible, doesn't have to be in this PR).

Tried to login via GitHub and also login as enterprise with redirect, works magically on both!

I did notice the ` CodyVSCodeExtension:Auth:connected` was logged in twice though, is that expected?

philipp-spiess:
> I did notice the  CodyVSCodeExtension:Auth:connected was logged in twice though, is that expected?


Ah good question I guess you see both the redirect _and_ the token post logging that. I will see if we can fix this somehow 🙇 Thanks for the review!

vscode/src/auth/token-receiver.ts
philipp-spiess: Will do! 

No this wont work on VS Code on web, this is purely an optimization for local desktops.

philipp-spiess:
I think it's reasonable for this event to be logged twice (although I can not repro?) I did add a specific event for when the token receiver server receives the token, though. This way we know exactly how the client receives the token (if the user follows the redirect AND posts to the webserver, we log two events but this seems reasonable?) 

We currently have no event though when a _new_ auth provider is added the first time (this seems unrelated to this PR though)

PR #3245, opened 2024-02-22
Custom Commands: display commands in sidebar

abeatrix:
PR add custom commands to sidebar as discussed with @toolmantim

image

- move the TreeViewProvider registration process for commands to the command provider as that's the service that manages the commands
- reorganize tree-views items into its own folder
- refactor TreeViewProvider to make sure methods are not one-type-specificed

## Test plan

e2e test for custom commands has been updated accordingly.

1. Check the sidebar to see if your custom commands are showing up there
2. Clicking on the custom command in sidebar should work as expected

https://github.com/sourcegraph/cody/assets/68532117/0bc58b64-ba2a-4d7e-add1-3def062ed366

### After

image

### Before

image

### DIFF
--- a/vscode/src/services/tree-views/commands.ts
+++ b/vscode/src/services/tree-views/commands.ts
@@ -21,6 +21,7 @@ export function getCommandTreeItems(customCommands: CodyCommand[]): CodyTreeItem
                 ? vscode.TreeItemCollapsibleState.Expanded
                 : vscode.TreeItemCollapsibleState.None
         )
+        treeItem.description = item.keybinding
 
         if (item.key === 'custom' && customCommands?.length) {
             try {
--- a/vscode/src/services/tree-views/commands.ts
+++ b/vscode/src/services/tree-views/commands.ts
@@ -4,32 +4,24 @@ import { CodyCommandMenuItems } from '../../commands'
 import type { CodyCommand } from '@sourcegraph/cody-shared'
 
 /**
- * Method to initialize the grouped chats for the Commands items
+ * Method to get items for the Commands sidebar
  */
 export function getCommandTreeItems(customCommands: CodyCommand[]): CodyTreeItem[] {
     const treeNodes = []
 
     // Create a CodyTreeItem for each group and add to treeNodes
     for (const item of CodyCommandMenuItems) {
-        const treeItem = new CodyTreeItem(
-            item.key,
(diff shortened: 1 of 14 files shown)


vscode/src/services/tree-views/commands.ts
```diff
@@ -0,0 +1,37 @@
...
+export function getCommandTreeItems(customCommands: CodyCommand[]): CodyTreeItem[] {
+    const treeNodes = []
+
+    // Create a CodyTreeItem for each group and add to treeNodes
+    for (const item of CodyCommandMenuItems) {
+        const treeItem = new CodyTreeItem(item.key, item.description, item.icon, item.command)
+        treeItem.description = item.keybinding
+
+        if (item.key === 'custom' && customCommands?.length) {
+            treeItem.collapsibleState = vscode.TreeItemCollapsibleState.Expanded
```
toolmantim, line 18: I think given that we're not persisting the collapse/expand state, like the file explorer in VS Code does, it may be safer to always start this as collapsed, in case people in a shared codebase with lots of commands don't want it to always be expanded.

```suggestion
            treeItem.collapsibleState = vscode.TreeItemCollapsibleState.Collapsed
```

I think ideally we keep it expanded by default if there are commands, but persist/remember if they collapsed it similar to file explorer.

toolmantim (approved):
Works great! I have one small non-blocking suggested change/improvement.

vscode/src/services/tree-views/commands.ts
abeatrix: If it was collapsed manually by the user, it will stay collapsed until they expanded it again like how it works in the file explorer--is that what you're referring to? 
We won't be able to expand it programmatically again once the user has clicked on the collapse button so I think it'd be nice to start it as expanded to let users know about this new field, but happy to change it in a follow up pr if you think we should start the menu as collapsed :) or if we want to start an a/b test for this just lmk! We actually are seeing a lot of people clicking on the New Chat button on the sidebar!

abeatrix: Tl;dr the state is persisted I believe

toolmantim: Oh! For some reason it wasn’t for me, but if it is, that’s great 👍🏼

PR #3256, opened 2024-02-24
Debug: new button to export log

abeatrix:
proof of concept: Add functionality to export Cody output logs using VS Code API

This PR introduces the ability to export Cody output as log file in the VS Code extension by utilizing the VS Code API. It provides a seamless log export experience that is compatible with all VS Code client types that supports the VS Code API.

- Added a new command 'cody.debug.export.logs' to export Cody output logs in the VS Code extension.
- Updated main.ts to include the import of 'exportOutputLog' function.
- Registered the 'cody.debug.export.logs' command in the extension.
- Added a new command 'cody.sidebar.logs' in SidebarCommands.ts to export logs from the sidebar.
- Updated treeViewItems.ts to include an 'Export Logs' option in the sidebar tree view.

Updated:

Moved to the ... menu

image

### Follow-up

We should enable debugs by default.

## Test plan

Click on the "Export logs" icon in the sidebar to export the text from output channel.

https://github.com/sourcegraph/cody/assets/68532117/9da8f379-0ead-4b26-8c9e-3ee5cdb533d2


vscode/src/services/export-logs.ts
```diff
@@ -0,0 +1,54 @@
...
+        )
+
+        if (!outputDirStat) {
+            return
+        }
+
+        // search for the file name that ends with 'Cody by Sourcegraph.log' inside the outputdir
+        const outputDir = vscode.Uri.joinPath(logDir, outputDirStat[0])
+        const outputFiles = await vscode.workspace.fs.readDirectory(outputDir)
+        const logFile = outputFiles.find(file => file[0].endsWith('Cody by Sourcegraph.log'))
```
philipp-spiess, line 25: Interesting! is there a way we can avoid hardcoding these values and use the ones that are currently defined for the extension name etc? 

Want to avoid breaking it if we decide to change the name

vscode/src/services/export-logs.ts
```diff
@@ -0,0 +1,54 @@
...
+    const logDir = vscode.Uri.joinPath(logUri, '..')
+    try {
+        // Search for a directory that starts with "output" inside logDir
+        const files = await vscode.workspace.fs.readDirectory(logDir)
+        const outputDirStat = files.find(
+            file => file[0].startsWith('output') && file[1] === vscode.FileType.Directory
+        )
+
+        if (!outputDirStat) {
+            return
```
philipp-spiess, line 19: Maybe if this fails here or below, we can just open the output channel as a fallback?

philipp-spiess (approved):
Awesome!

toolmantim:
Can we move it behind the “…” menu, out of the treeview? Thanks!

abeatrix:
> Can we move it behind the “…” menu, out of the treeview? Thanks!

for sure! What is ... menu though 😆

vscode/src/services/export-logs.ts
abeatrix: good point, let me check on this!

toolmantim:
When you hover over “Settings & Support” there’s the “…” button, and it opens an OS popover menu. From memory the getting started guide is there (that’s actually probably a better candidate for the treeview). If we could have the export action appear that “…” popover, instead of the treeview, I think that would be best.

arafatkatze (approved):
🙇

abeatrix:
> When you hover over “Settings & Support” there’s the “…” button, and it opens an OS popover menu. From memory the getting started guide is there (that’s actually probably a better candidate for the treeview). If we could have the export action appear in the “…” popover menu, instead of the treeview, I think that would be best.

Moved the item to the ... menu 🙂
image

toolmantim (approved):
Love it! One tiny thing about the action labeling, given it’s not atomic (opens the file save dialog)

Can we mention that it’s also available in the standard VS Code command palette? I feel like the easiest instructions for support people is to tell people to type Cmd-Shift-P and type “Cody export logs”

PR #3259, opened 2024-02-24
organize imports upon saving in editor and when running biome

sqs:
This only enables Biome's import-organization upon saving in the editor and when running `biome check`.

This ensures that imports are consistent, which will reduce merge conflicts overall and will prevent unrelated import organizations from being included in commits.

This does NOT re-enable all Biome quickfixes, which were causing issues and were disabled in https://github.com/sourcegraph/cody/pull/2900.

## Test plan

CI


philipp-spiess (approved):
Good stuff

PR #3260, opened 2024-02-24
format JSON files in a standard way

sqs:
Biome is now used to format all JSON files. This fixes the problem where some JSON files were indented with 4 spaces and some were indented with 2 spaces, which caused noisy diffs on `package.json` and other JSON files when automated tooling would write them with 2 spaces and then they would be re-formatted with 4 spaces when somebody opened the file in their editor. Now all JSON files are indented with 2 spaces, which was more common in this repo and was what our editorconfig already specified.

## Test plan

CI


philipp-spiess (approved):
Yes please

PR #3261, opened 2024-02-25
Update symf index when stale

beyang:
Local indexes are rebuilt automatically on a daily cadence when they are stale. Staleness is determined by checking whether files have changed across Git commits and in the set of working file updates not yet committed. If the indexed directories is not in a Git repository, then reindex every day.

## Test plan

- [x] Test chat context still works
- [x] Change the update frequency from [1 day](https://github.com/sourcegraph/cody/pull/3261/files#diff-0299d8549e3068ffd8c6bce70122b88f7cc9871455aff5bdd90572a8c44eb636R28) to 1 second and test that the index is rebuilt when files are changed and a new chat question is asked.


dominiccooney (approved):
Looks good!

philipp-spiess (approved):
Nice!

PR #3267, opened 2024-02-25
fix App story

sqs:
The `vscodeAPI` should be passed down from App to avoid global state, which makes the App story in the storybook work again.

## Test plan

CI


burmudar (approved):

toolmantim:
🎉

PR #3269, opened 2024-02-25
rename type Context{File => Item}, remove existing redundant ContextItem type,

sqs:
These 2 types (ContextFile and ContextItem) were basically the same, except for a different field name (`content` vs. `text`) and a different default value (plus some other meaningless differences).

Also, ContextFile was a misnomer as this was being used for symbols as well (not just files).

There is no behavior change from this diff.

## Test plan

CI and e2e tests


sqs:
image

dominiccooney:
Haven't gone through this in great detail, but there is an untyped boundary that's a bit confusing in the way chat transcripts are serialized and deserialized. If I had to guess the missing 'line' property error on FileLink probably arises from that.

Any indication what's causing CI to fail?

abeatrix (approved):
Updated the agent test recordings to fix the failing unit tests in CI

PR #3274, opened 2024-02-28
Autocomplete: expose the smart throttle setting

valerybugakov:
## Context

- Document the existing setting to be visible in the extension settings for the insiders build.

## Test plan

n/a


dominiccooney (approved):
Straightforward enough.

PR #3275, opened 2024-02-28
Smarter Doc: Use Tree-sitter for range expansion, and show hints alongside symbols

umpox:
## Description

This PR:
- Improves the range used for the `doc` command,
  - Gets the actual range of the function, variable, class etc.
  - Gets the actual symbol being targeted, which lets us do more complex insertions depending on the type.
    - E.g. if the symbol is a function, class or method in Python, we can insert the documentation _below_ the symbol.
- Adds more command hints:
  - If the cursor is on a "documentable node", we show: "Opt+D" to document
    - Behind feature flag: `cody-document-hints`. We will A/B test to see the effect this has.
  - If the cursor is in an empty file, we show "Opt+K" to generate.
    - Not behind a feature flag, small change that is unobtrusive (immediately disappears if the file has any text)
- Improves tree-sitter support for the `Doc` command:
  - Gets the symbol and range of a "documentable node"
  - Suport for:
     - TypeScript/JavaScript
     - Go
     - Python
 - Adds a direct shortcut for "Document Code"
   - Keybinding: Option+D for Mac, or Alt+D for other OS

## Screenshots

image

image

## Test plan

**A/B test functionality**

Manual:
1. Enable setting manually (set `"cody-command-hints": true` in VS Code settings)
2. Check all Generate, Document and Edit ghost text appears
3. Check relevant enrolment events are sent (variant=treatment)
4. Disable setting manually
5. Check all Generate, Document and Edit ghost text disappears
6. Check relevant enrolment events are sent (variant=control)

A/B test Document:
1. Enable A/B feature `cody-document-hints`
2. Move the cursor to a variable declaration in TypeScript
3. Check "Opt-D to Document shows"
4. Check correct enrolment event is sent (variant=treatment)
6. Move cursor to another variable
7. Check enrolment event is _not_ sent (should only send once)

A/B test EditChat:
1. Enable A/B feature `cody-command-hints`
2. Select some code
3. Check "Opt-K to Edit, Opt+L to Chat" shows
4. Check correct enrolment event is sent (variant=treatment)
5. Make another selection
6. Check enrolment event is _not_ sent (should only send once)

**Document functionality:**

1. Move cursor to a function declaration in TS, Go or Python
6. Check "Opt+D to Document" shows up
8. Check the keybinding executes the command
10. Check the range of the command covers the entire function, not just the cursor position.
11. Check the documentation is inserted above the function

**Generate functionality:**

1. Create a new, empty file
2. Open in
3. Expect to see "Opt+K to Generate"
9. Check the keybinding executes the command
10. Check the command functions as expected


umpox:
Need to figure out why the agent tests don't pass in CI, but the rest of the PR is ready for review :)

vscode/src/edit/utils/edit-models.ts
```diff
@@ -17,11 +17,12 @@ export function getEditModelsForUser(authStatus: AuthStatus): ModelProvider[] {
 
 export function getOverridenModelForIntent(intent: EditIntent, currentModel: EditModel): EditModel {
     switch (intent) {
-        case 'doc':
         case 'fix':
         case 'test':
             // Edit commands have only been tested with Claude 2. Default to that for now.
             return 'anthropic/claude-2.0'
+        case 'doc':
+            return 'anthropic/claude-instant-1.2'
```
abeatrix, line 25: nice! should we check if the currentModel is an ollama model or not so that ollama users can still configure this option? wdyt?

vscode/src/edit/provider.ts
```diff
@@ -231,7 +237,7 @@ export class EditProvider {
 
     private async handleFileCreationResponse(text: string, isMessageInProgress: boolean): Promise<void> {
         const task = this.config.task
-        if (task.state !== CodyTaskState.pending) {
+        if (task.intent === 'test' && task.state !== CodyTaskState.working) {
```
abeatrix, line 240:
```suggestion
        if (task.intent === 'test' && task.state !== CodyTaskState.pending) {
```

I think we only want to create the file request if the task is at pending state since we only want to create the file once? It'd be cool to support multiple file creation though but i don't think that's something the test command needs right now?

vscode/src/code-actions/document.ts
```diff
@@ -1,30 +1,36 @@
...
     public provideCodeActions(document: vscode.TextDocument, range: vscode.Range): vscode.CodeAction[] {
-        const [documentableNode] = execQueryWrapper(document, range.start, 'getDocumentableNode')
+        const [documentableSymbol, documentableRange] = execQueryWrapper(
+            document,
+            range.start,
+            'getDocumentableNode'
+        )
 
-        if (!documentableNode) {
+        if (!documentableSymbol?.node || !documentableRange?.node) {
```
abeatrix, line 16: a follow up idea: it'd be nice to check if the symbol already have a doc string or not before we display the ghost text

abeatrix (approved):
This is really good! I love how snappy it is and serves an actual purpose! This makes it feel less spammy and less distracting as well.

## Tests

Confirmed the Ghost Text for Doc shows up accordingly
image

Ghost text is not showing up in untitled new file (which I'm guessing is expected?):
image

It does show up in a saved empty file:
image

NOTE: only tested on python and typescript files.


## Bugs

Some bugs I noticed that might not be related to the changes in this PR:

1. when I tried to document a python function that have a multi-line params, it'd break the code 😬 

https://github.com/sourcegraph/cody/assets/68532117/060b7256-56cb-4b79-b8c5-f3a3079e698c

2. Ghost text remains in place when doc string was added. I'd expect it to go away when there is a code lenses for the add?

image

image


Feedback on the ghost text showing up in an empty file, I find `shortcut to Generate` a bit unclear as in what this shortcut can generate. Since it'd an empty file, maybe we can add some examples on how they can use it, at least for the first time they see it?

image

vscode/src/edit/provider.ts
umpox: I just merged `CodyTaskState.pending` into `working` here, I think they're ultimately the same thing, we just need to add the extra check for `task.intent === 'test'` in our logic. Trying to make the Edit states a bit simpler!

umpox:
@abeatrix Thanks for the review!

> when I tried to document a python function that have a multi-line params, it'd break the code 😬

I overlooked this scenario, looking at fixing

vscode/src/code-actions/document.ts
umpox: Will follow up on this

vscode/src/edit/utils/edit-models.ts
umpox: Hmmm, I think that's a good shout but I think it's possibly a bit too "magic" right now. Let's revisit this when we tackle the problem of selecting a model for chat and edit via a quick pick (looking at this after this PR), ideally we want to allow users to configure this completely so we might be able to get something working there

vscode/src/tree-sitter/query-tests/test-data/documentable-node.snap.go
```diff
@@ -0,0 +1,158 @@
...
+// ------------------------------------
+
+  package main
+
+  type Creature struct {
+//^ start range.identifier[1]
+//     ^^^^^^^^ symbol.identifier[1]
+//         █
+  }
+//^ end range.identifier[1]
```
valerybugakov, line 15: 🪄

valerybugakov: This is awesome! I might reuse these queries for the graph context experiment.

vscode/src/code-actions/document.ts
valerybugakov: > Inside getDocumentableNode only return if there is not a comment node on the line above the symbol.

Yep, this sounds reasonable to me. We can post-process the queries in the query wrapper if they are insufficient.

valerybugakov (approved):
🔥

vscode/src/tree-sitter/query-sdk.ts
```diff
@@ -159,10 +159,14 @@ function getLanguageSpecificQueryWrappers(queries: ResolvedQueries, _parser: Par
...
-                // This is primarily used for languages where documentation should be inserted differently
-                // depending on the syntax (e.g. Python PEP 257)
+            if (languageId === 'python' && range) {
+                /**
+                 * Python is a special case for generating documentation.
+                 * The insertion point of the documentation should differ if the symbol is a function or class.
+                 * We need to query again for an insertion point, this time using the correct determined range.
+                 *
+                 * See https://peps.python.org/pep-0257/ for the documentation conventions for Python.
+                 */
```
umpox, line 169: @valerybugakov Small change here to make `languageId` available to the query wrappers.

It feels pretty inefficient to be doing this - running _another_ query after the first one, but I don't know how we could do it another way.

We can't guarantee that the users' cursor/selection includes the full range, so we need to first expand the range before looking for the insertion point.

It adds a small amount of latency to the query, although still very fast for me (sub 1ms)

valerybugakov: Make sense 👍

PR #3277, opened 2024-02-28
Add Mixtral model to code edit feature

PriNova:
This PR adds the Mixtral model to the code edit feature.

The code edit feature allows users to ask Cody to edit code according to their instructions, giving them the chance to retry the edit with a new prompt, undo, or accept the change.

With this PR, users can now select the Mixtral model when using the code edit feature. This may improve the quality of the code edits for certain use cases.

Here's a summary of the changes:

- Adds a new file `mixtral.ts` in the 'models' directory, which exports a `mixtral` object that contains various methods for generating prompts for the Mixtral model.
- The `mixtral` object includes methods for generating prompts for editing, documenting, fixing, adding, and testing code.
- The `mixtral` object is added to the `INTERACTION_MODELS` object in the `index.ts` file, which maps model names to `EditLLMInteraction` objects.
- The `mixtral` object is associated with the `fireworks/accounts/fireworks/models/mixtral-8x7b-instruct` model name.

I've tested the changes thoroughly and ensured that the Mixtral model works as expected with the code edit feature.

Please let me know if you have any questions or concerns.

### DIFF
--- a/lib/shared/src/models/dotcom.ts
+++ b/lib/shared/src/models/dotcom.ts
@@ -50,6 +50,6 @@ export const DEFAULT_DOT_COM_MODELS = [
         default: false,
         codyProOnly: true,
         // TODO: Improve prompt for Mixtral + Edit to see if we can use it there too.
-        usage: [ModelUsage.Chat],
+        usage: [ModelUsage.Chat, ModelUsage.Edit],
     },
 ] as const satisfies ModelProvider[]
--- a/vscode/src/edit/prompt/index.ts
+++ b/vscode/src/edit/prompt/index.ts
@@ -17,6 +17,7 @@ import { PromptBuilder } from '../../prompt-builder'
 import { getContext } from './context'
 import { claude } from './models/claude'
 import { openai } from './models/openai'
+import { mixtral } from './models/mixtral'
 import type { EditLLMInteraction, GetLLMInteractionOptions, LLMInteraction } from './type'
 
 const INTERACTION_MODELS: Record<EditModel, EditLLMInteraction> = {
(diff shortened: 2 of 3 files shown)


philipp-spiess:
Thanks! I wish this would exist too just because how fast Mixtral is but have you tested the quality and ensured it works well? IIRC we had some quality related concerns and that's why decided not to include Mixtral in this list.

cc @umpox to take a look :)

PriNova:
At least, it works technically.
I'm testing now for different languages, programming, architectures, code semantics, etc.
If it falls short on quality, we can drop this PR.
I will report later.

PriNova:
The editing with Mixtral have two flaws:

1. It sometimes misses indentations.
2. It sometimes put `\` in variables separated with `_`

But overall the quality looks very promising. 
I also tried it with `fix`, `docs`, and `tests`, which shows the same flaws above.

umpox:
> It occasionally inserts backslashes (\) in variables separated by underscores (_).

I noticed this too, really strange. Maybe we can do some better escaping in the response 🤔

PriNova:
Hey @umpox . thank you for your reply.

I'm testing all the way down to the raw response of the Mixtral model. Do a lot of prompt crafting and A/B testing. It seems there is a style issue with the backslashes if in edit-insert mode. The edit-add mode works flawlessly.
Could it be that Mixtral doesn't like XML tags?! Will try with different tags.

Since Fireworks support grammar structure responses, maybe this could help also.

PriNova:
> > It occasionally inserts backslashes () in variables separated by underscores (_).
> 
> I noticed this too, really strange. Maybe we can do some better escaping in the response 🤔

Since the replacement will be handled in the background, VS code could linting or styling the text automatically.
I will research deeper if I extract the raw model response.

PriNova:
After logging back and forth in Cody, I used the Fireworks API directly to reproduce the backslash issue.
With prompting the Mixtral Model through the API in Python, Node and cUrl directly, the issue does not appear.

I assume that forwarding the response from Fireworks -> Sourcegraph -> users-frontend there could be the issue, but I might be wrong.

PriNova:
@umpox I resolved the backslash issue preceding an underscore within the 'contentSanitizer()' function.

PriNova:
Will test the Mixtral model for `edit` feature before leaving the draft mode.

Issue #3252, opened 2024-02-23
Ollama: Experimental support for chat


philipp-spiess:
Creating an issue to track the demand for ollama support for chat models.

cam-s-hunt:
+1. I'd like to be able to use a local Ollama API.

ByerRA:
+1 YES!!!!

I think having the ability to have ALL of Cody's functions able to be run from Ollama would be of great benefit and would also draw more people to use Cody as I believe more and more people are going to want to run local models.

mrdjohnson:
+1 big ollama user here, I'd love cody + ollama for use on flights / bad internet connections / code security

PieBru:
Following this guide: https://github.com/sourcegraph/cody/blob/main/vscode/doc/ollama.md
![image](https://github.com/sourcegraph/cody/assets/6537434/6beec38f-d387-4d26-b73d-fb6206519d5d)

mrdjohnson:
I just tried using Cody offline with the ollama flag on and it was still trying to hit Cody's servers it looks like? 

`"cody.experimental.ollamaChat": true`

`Request Failed: getaddrinfo ENOTFOUND sourcegraph.com`

Is the ollama integration meant for Cody to work completely offline?

mrdjohnson:
I think I figured out how to use ollama! (I needed to create a new chat and set the bot to be ollama ) Thank you so much for the work you put into this! 🥳 🥳

abeatrix:
Thank you for trying it out 🤗 

We have a blog post with instructions on how to set up Ollama for chat if anyone else is interested in giving it a try 😃  https://sourcegraph.com/blog/cody-vscode-1-8-0-release

PR #3282, opened 2024-03-01
Chat: Support local Ollama models

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3252

image

This PR adds support for locally run Ollama models for Chat and Commands

> NOTE: Previously a proof of concept for a hackathon idea: Ollama for chat and commands

## Test plan

1. Download Ollama https://ollama.com/download
2. Start Ollama (makes sure the ollama logo is showing up in your menu bar)
3. Select a chat model (e.g. [gemma:7b-instruct-q4_K_M](https://ollama.com/library/gemma:7b-instruct-q4_K_M)) from the [Ollama Library](https://ollama.com/library)
4. Pull the chat model locally (e.g. `ollama pull gemma:7b-instruct-q4_K_M`)
5. Start debug mode from this branch after the chat model has been downloaded successfully
6. Enable the `cody.experimental.ollamaChat` configuration
7. Open a new chat in Cody
8. In the new chat panel, you should see the chat model you've pulled in the dropdown list
9. currently, you will need to restart vs code to see the new models you just downloaded, will update so that the new models will be displayed instantly.

Note: You can run `curl http://localhost:11434/api/tags` in your terminal to see what models are available on your machine

## Demo

https://github.com/sourcegraph/cody/assets/68532117/22ff03d4-2d46-469e-b9ef-c6ecb935d301


lib/shared/src/sourcegraph-api/completions/nodeClient.ts
```diff
@@ -40,6 +42,99 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
...
+                    prompt: lastHumanMessage.text,
+                    messages: params.messages.map(msg => {
+                        return {
+                            role: msg.speaker === 'human' ? 'user' : 'assistant',
+                            content: msg.text,
+                        }
+                    }),
+                }
+
+                fetch(new URL('/api/generate', OLLAMA_DEFAULT_URL).href, {
```
philipp-spiess, line 62: What do you think of adding this logic in `lib/shared/src/chat/chat.ts` instead? I've used that as a branch to fork off because I think this file should be deprecated soon and we should rely on `fetch` instead (which works across envs and is not node specific).

vscode/doc/ollama.md
```diff
@@ -17,8 +19,24 @@
      }
    }
    ```
+
 1. Confirm Cody uses Ollama by looking at the Cody output channel or the autocomplete trace view (in the command palette).
 
 ## Current limitations
 
 Ollama is currently only supported by the autocomplete feature and is considered experimental. You can track [issue #3252](https://github.com/sourcegraph/cody/issues/3252) for chat support.
```
philipp-spiess, line 27: can remove this now 👍

abeatrix: removed 👍

lib/shared/src/models/index.ts
```diff
@@ -29,6 +30,33 @@ export class ModelProvider {
...
+     * and adds them to the list of ollama providers.
+     */
+    public static getLocalOllamaModels(endpoint?: string | null): void {
+        const isAgentTesting = process.env.CODY_SHIM_TESTING === 'true'
+        // Only fetch local models if the endpoint is not a dotcom instance
+        if (isAgentTesting || !endpoint || !isDotCom(endpoint)) {
+            return
+        }
+
+        fetch(new URL('/api/tags', OLLAMA_DEFAULT_URL).href)
```
philipp-spiess, line 47: Should we make the URL configurable like we do for autocomplete? (cc @valerybugakov who might have input on why we made it configurable)

lib/shared/src/models/index.ts
```diff
@@ -29,6 +30,33 @@ export class ModelProvider {
...
     // Providers available for dotcom instances
     private static dotComProviders: ModelProvider[] = DEFAULT_DOT_COM_MODELS
+    // Providers available from local ollama instances
+    private static ollamaProviders: ModelProvider[] = []
+
+    /**
+     * Fetches available Ollama models from the local Ollama server
+     * and adds them to the list of ollama providers.
+     */
+    public static getLocalOllamaModels(endpoint?: string | null): void {
```
philipp-spiess, line 40: The problem by doing this without a feature flag is that it will not be easy for people to find out why it is not working when they expect it to. I.e if someone forgets to start Ollama they might be confused as to why no ollama model shows up.

Have you thought of making this a config option instead and, if enabled, errors when we can't connect to the ollama backend? This way onboarding is a bit harder but we will have less people that come and ask why it's not working is my assumption.

lib/shared/src/models/index.ts
```diff
@@ -29,6 +30,33 @@ export class ModelProvider {
...
+            .then(data => {
+                const models = new Set<ModelProvider>()
+                for (const model of data.models) {
+                    const name = `ollama/${model.model}`
+                    const newModel = new ModelProvider(name, [ModelUsage.Chat])
+                    models.add(newModel)
+                }
+                ModelProvider.ollamaProviders = Array.from(models)
+            })
+            .catch(() => console.log('Cannot find local ollama models'))
```
philipp-spiess, line 58: This is not logged anywhere is it? Should this use `debugLog` instead?

lib/shared/src/models/index.ts
```diff
@@ -29,6 +30,33 @@ export class ModelProvider {
...
+
+        fetch(new URL('/api/tags', OLLAMA_DEFAULT_URL).href)
+            .then(response => response.json())
+            .then(data => {
+                const models = new Set<ModelProvider>()
+                for (const model of data.models) {
+                    const name = `ollama/${model.model}`
+                    const newModel = new ModelProvider(name, [ModelUsage.Chat])
+                    models.add(newModel)
+                }
```
philipp-spiess, line 55: Should we do some filtering here and now show e.g. code completion models (anything that is not instruct tuned might not be a good fit for chat).

lib/shared/src/sourcegraph-api/completions/nodeClient.ts
```diff
@@ -38,6 +39,11 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
                 }
             }
 
+            if (params.model?.startsWith('ollama')) {
+                ollamaChatClient(params, cb, this.completionsEndpoint, this.logger, signal)
+                return
+            }
+
```
philipp-spiess, line 46: Let's do the branching in the `chat.ts` class (the one that calls into the nodeClient) so it also works for web builds

vscode/test/e2e/command-custom.test.ts
philipp-spiess, line 1: These changes feel unrelated?

philipp-spiess (approved):
Awesome! Left some ideas inline, let me know what you think about it (but this looks safe enough to land)

Add a change log entry tho

lib/shared/src/models/index.ts
abeatrix: Yea that makes sense! Will put this behind an experimental feature flag

vscode/test/e2e/command-custom.test.ts
abeatrix: Yea this will be removed once https://github.com/sourcegraph/cody/pull/3313 is landed. Was trying to fix the flaky tests that kept failing!

lib/shared/src/models/index.ts
abeatrix: I can update this to be configurable in a follow-up PR if this is common!

lib/shared/src/models/index.ts
abeatrix: I think this should be fine for now as the goal is to allow users to test different models out themselves?

lib/shared/src/models/index.ts
philipp-spiess: Sounds good

lib/shared/src/models/index.ts
philipp-spiess: Sounds good but let's mention this in the docs so it's clear for people that they should look for instruct tuned models.

lib/shared/src/sourcegraph-api/completions/nodeClient.ts
abeatrix: I ran into some cros issue with the browser client, will do this in a follow-up!

lib/shared/src/models/index.ts
abeatrix: Updated the Ollama doc to reflect this 😄

PR #3283, opened 2024-03-01
Autocomplete: reduce the adaptive timeout

valerybugakov:
## Context

- Reduce the adaptive timeout to match latency improvements.

## Test plan

n/a


philipp-spiess (approved):
NICE

PR #3286, opened 2024-03-02
ensure symf is only downloaded once

beyang:
Previously, it was possible to have 2 parallel symf downloads if `getSymfPath` was called twice before the first download completed. Sometimes, these appeared to interfere with one another.

## Test plan

See added unit test.

### DIFF
--- a/vscode/src/local-context/download-symf.ts
+++ b/vscode/src/local-context/download-symf.ts
@@ -49,19 +49,18 @@ export async function _getSymfPath(
         )
         return null
     }
+    // Releases (eg at https://github.com/sourcegraph/symf/releases) are named with the Zig platform
+    // identifier (linux-musl, windows-gnu, macos).
+    const zigPlatform =
+        platform === 'linux' ? 'linux-musl' : platform === 'windows' ? 'windows-gnu' : platform
 
     const symfFilename = `symf-${symfVersion}-${arch}-${platform}`
-    const symfUnzippedFilename = `symf-${arch}-${platform}` // the filename inside the zip
+    const symfUnzippedFilename = `symf-${arch}-${zigPlatform}` // the filename inside the zip
     const symfPath = path.join(symfContainingDir, symfFilename)
     if (await fileExists(symfPath)) {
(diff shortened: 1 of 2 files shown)


vscode/src/local-context/download-symf.ts
```diff
@@ -33,59 +49,82 @@ export async function getSymfPath(context: vscode.ExtensionContext): Promise<str
...
-
-                const tmpFile = path.join(symfTmpDir, `symf-${arch}-${zigPlatform}`)
-                await fspromises.chmod(tmpFile, 0o755)
-                await fspromises.rename(tmpFile, symfPath)
-                await fspromises.rm(symfTmpDir, { recursive: true })
-
-                logDebug('symf', `extracted symf to ${symfPath}`)
-            }
-        )
+        await downloadLock.acquire()
```
dominiccooney, line 69: This is fine, but it looks like there's something simpler in this package which encapsulates the `try`:

```
await downloadLock.runExclusive(async () => {
  ...
})
```

dominiccooney (approved):
async-mutex package looks neat, we could use this instead of slinging Promises in a couple of places in bfg too.

vscode/src/local-context/download-symf.ts
beyang: IMO, the code is more readable as-is, because there are some return statements in the `try` and `catch` blocks that are more awkward to wrap in `runExclusive`. Rewritten with `runExclusive` would look something like this:

```
return await downloadLock.runExclusive(async () => {
    try {
        // Re-check if it has been downloaded
        if (await fileExists(symfPath)) {
            logDebug('symf', 'symf already downloaded, reusing')
            return symfPath
        }

        await actualDownloadSymf({ symfPath, symfURL, symfFilename, symfUnzippedFilename })
        void removeOldSymfBinaries(symfContainingDir, symfFilename)
    } catch (error) {
        captureException(error)
        void vscode.window.showErrorMessage(`Failed to download symf: ${error}`)
        return null
    }
    return symfPath
})
```

PR #3289, opened 2024-03-04
Autocomplete: migrate to `bfg/contextForIdentifiers`

valerybugakov:
## Context

- Added `getGraphContextIdentifiers` to the tree-sitter SDK, which will be used to experiment with graph context.
    - For now, only JS and TS queries are polished. Going to improve `go` and `python` in a separate PR after experimenting with JS.
- Migrated to the new `bfg/contextForIdentifiers` BFG request, where clients are responsible for specifying the list of identifiers. More context https://github.com/sourcegraph/bfg-private/pull/171
    - The BFG request migration won't be an immediate priority. [We plan to experiment with LSP context in the short term](https://docs.google.com/document/d/1V2SB-ixi8OrhTwCMPO0MseBkA4tnipAkse5CCXbucXI/edit?usp=sharing). I want to merge graph-context-identifiers tree-sitter queries and start using them in combination with LSP. 
- Updated the `execQueryWrapper` API to allow different argument combinations.
- Added the tree-sitter tree representation output to snapshot tests if `// only` is used to debug a particular test.

## Test plan

- CI
- Added graph-identifiers query snapshot tests


valerybugakov:
Tagging @philipp-spiess for early 👀

vscode/src/completions/context/context-mixer.ts
```diff
@@ -61,6 +64,10 @@ export interface GetContextResult {
 export class ContextMixer implements vscode.Disposable {
     constructor(private strategyFactory: ContextStrategyFactory) {}
 
+    public isBfgRetrieverEnabled(): boolean {
+        return this.strategyFactory.graphRetriever instanceof BfgRetriever
+    }
```
philipp-spiess, line 69: Unused?

vscode/src/completions/context/context-strategy.ts
```diff
@@ -16,13 +16,14 @@ export type ContextStrategy =
...
     getStrategy(document: vscode.TextDocument): { name: ContextStrategy; retrievers: ContextRetriever[] }
+    graphRetriever: ContextRetriever | undefined
 }
 
 export class DefaultContextStrategyFactory implements ContextStrategyFactory {
     private disposables: vscode.Disposable[] = []
 
     private localRetriever: ContextRetriever | undefined
-    private graphRetriever: ContextRetriever | undefined
+    public graphRetriever: ContextRetriever | undefined
```
philipp-spiess, line 26: Why is this made public? 🤔

vscode/src/completions/context/retrievers/bfg/bfg-retriever.test.ts
philipp-spiess, line 1: This is just a copy for now, or? I don't see know if there's changed behavior here.

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
```diff
@@ -184,32 +188,57 @@ export class BfgRetriever implements ContextRetriever {
...
+                new Set([
+                    // Get last 10 identifiers from the last candidate insert text
+                    // in hopes that LLM partially guessed the right completion.
+                    ...getLastNBfgIdentifiersFromString({
+                        n: 10,
+                        document,
+                        position,
+                        currentLinePrefix: docContext.currentLinePrefix,
+                        source: lastCandidateCurrentLine,
+                    }),
```
philipp-spiess, line 206: Hm interesting so this means BFG will only be good if we have two round trips to the LLM? One to get a good _partial suggestion_ and then one that will include BFG?

vscode/src/completions/doc-context-getters.ts
```diff
@@ -52,7 +55,83 @@ export function getCompletionIntent(params: GetCompletionIntentParams): Completi
...
+    position: vscode.Position
+    currentLinePrefix: string
+    /**
+     * Parse this source string to get the tree for the tree-sitter query
+     * instead of using `document.getText`
+     */
+    source?: string
+}
+
+export function getLastNBfgIdentifiersFromDocument(
```
philipp-spiess, line 76: Should we couple this to BFG or just rename it to `getLastNIdentifier` instead?

If this is BFG specific maybe we should put it closer to the BFG parts in the source code?

vscode/src/completions/doc-context-getters.ts
```diff
@@ -52,7 +55,83 @@ export function getCompletionIntent(params: GetCompletionIntentParams): Completi
...
+
+interface GetLastNBfgIdentifiersFromStringParams {
+    n: number
+    document: vscode.TextDocument
+    position: vscode.Position
+    currentLinePrefix: string /**
+     * Parse this source string to get the tree for the tree-sitter query
+     * instead of using `document.getText()`
+     */
+    source: string
```
philipp-spiess, line 107: docstring indentation gone haywire :D

vscode/src/completions/get-inline-completions.ts
```diff
@@ -262,7 +262,12 @@ async function doGetInlineCompletions(
                   handleDidPartiallyAcceptCompletionItem,
               })
             : null
-    if (resultToReuse) {
+
+    // If smart throttle is enabled we want to reuse the last candidate only if the request is throttled (see below).
+    // Otherwise we want to make a new request based on the latest user input and update the completion.
+    // We also use the last candidate insert text as a source for identifiers for the updated context to improve the
+    // updated completion.
```
philipp-spiess, line 269: I’m not sure I understand the intent behind this. Looking at the smart throttle mess I produced it seems that it only returns `null` if the completion was aborted within the smart throttle interval.

It also means that we only return the `resultToReuse` after a 25ms sleep time and no longer instantly.

Maybe that's expected; If so: do you mind elaborating?

philipp-spiess (commented):
Good stuff! Left some thoughts inline

vscode/src/completions/context/retrievers/bfg/bfg-retriever.test.ts
valerybugakov: Yep. I'll delete it for now and return to it when we shift the focus back to BFG.

vscode/src/completions/context/retrievers/bfg/bfg-retriever.ts
valerybugakov: The first request will use identifiers from the last N lines. The second one will enhanced by the identifiers extracted from the last candidate.

vscode/src/completions/doc-context-getters.ts
valerybugakov: Yes! Moved it closer to context retrievers

valerybugakov:
Please take a look, folks! @umpox, @philipp-spiess

philipp-spiess (approved):
Noice

PR #3304, opened 2024-03-05
preserve trailing non-alphanum chars in chat message

sqs:
Given a chat message like `Explain @my/file.go :)`, all but the last non-alphanum char was dropped, so the resulting message as seen in the transcript would be `Explain @my/file.go)`. The handling of trailing non-alphanum was to not get tripped up by question marks, but it incorrectly assumed there would only be a single such character. This approach can be fixed more robustly by using a rich editor (https://github.com/sourcegraph/cody/pull/3287), but this is a quick fix for now.

## Test plan

CI

### DIFF
--- a/vscode/src/commands/utils/display-text.test.ts
+++ b/vscode/src/commands/utils/display-text.test.ts
@@ -198,4 +198,14 @@ describe('replaceFileNameWithMarkdownLink', () => {
             )}) what does this do`
         )
     })
+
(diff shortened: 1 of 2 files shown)


valerybugakov (approved):

philipp-spiess (commented):
Oh yikes. Thanks for fixing this, I didn't realize this was a problem in prod too 😰

PR #3307, opened 2024-03-05
clean up ContextItem<->ContextMessage conversion, remove duplicate wrapping & unused code

sqs:
I noticed in the conversion `ContextItem <-> ContextMessage` there are several possible places where we were wrapping context files with multiple layers of preamble, like:

```
Context from file path foo/bar.ts:
Context from file path foo/bar.ts: 
(actual contents of file)
```

There was a `stripContextWrapper` to try to undo that in some places, but it's not applied everywhere it's needed. Examples (from the agent recordings):

dupe2
dupecontext

This PR cleans up code related to this to remove the multiple layers of preamble, and deletes other related unused code.

Cleanups:

- clean up use of ContextMessage
- clean up codebase context
- remove unused task contextMessages
- fillInContextItemContent
- remove unused fullContext field
- fix usage of VS Code ranges

## Test plan

CI


sqs:
Note: most of the diff is the agent recordings. The rest is a major deletion:

```
 28 files changed, 175 insertions(+), 780 deletions(-)
```

philipp-spiess (approved):
Awesome

vscode/src/edit/prompt/context.ts
```diff
@@ -138,46 +126,21 @@ const getContextFromIntent = async ({
...
-        userContextFiles.sort((a, b) => a.uri.path.localeCompare(b.uri.path))
-    }
-    for (const file of userContextFiles) {
-        if (file.uri) {
-            const content = await editor.getTextEditorContentForFile(file.uri, file.range)
-            if (content) {
-                const message = createContextMessageByFile(file, content)
-                userProvidedContextMessages.push(...message)
-            }
-        }
```
umpox, line 176: I think we'll need to use `fillInContextItemContent` here like we do for chat, otherwise the user context files won't actually have any `content` and will be ignored as context.

quick fix here: https://github.com/sourcegraph/cody/pull/3318

vscode/src/edit/prompt/context.ts
```diff
@@ -138,46 +126,21 @@ const getContextFromIntent = async ({
...
+        userContextItems.sort((a, b) => a.uri.path.localeCompare(b.uri.path))
     }
 
-    return extractContextItemsFromContextMessages([
```
valerybugakov, line 179: Based on this diff and behavior before this PR, my best guess is that we should preserve the `contextMessage` to `contextItem` conversion until we properly remove the duplication between these types. Fixing it here: https://github.com/sourcegraph/cody/pull/4432

PR #3309, opened 2024-03-05
Edit: Fix missing context for instruction-based inputs

umpox:
## Description

@philipp-spiess noticed that Edit was not including context, which it should have been doing.

Recent work around commands + context improvements introduced a couple of bugs that weren't obvious:
- `contextMessages` are defaulted to `[]` from the Edit input, we had a bad check for these
- `range` is an optional value for `ContextItem`, but we rely on it to generate unique IDs for content within a file. This adds an extra ID mechanism to uniquely hash the `content` of the item alongside the file URI (if available)

## Test plan

1. Create an edit,
2. Check preceding code and following code are included as context

### DIFF
--- a/vscode/src/prompt-builder/utils.ts
+++ b/vscode/src/prompt-builder/utils.ts
@@ -1,3 +1,4 @@
+import crypto from 'crypto'
 import {
     type ContextItem,
     type Message,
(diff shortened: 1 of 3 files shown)


vscode/src/prompt-builder/utils.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from 'crypto'
```
philipp-spiess, line 1: crypto will only work in node so this will likely not compile in the web config. Maybe use the `crypto-js` package we already bundle? https://github.com/sourcegraph/cody/blob/a7cf2726c399a710b2bc2e7bf2bb0f85f85fa4c1/vscode/src/completions/providers/fireworks.ts#L548

philipp-spiess (approved):

vscode/src/prompt-builder/utils.ts
umpox: Thanks!

sqs:
It would be great to have a test for this. The code around this is pretty tricky. I have a PR https://github.com/sourcegraph/cody/pull/3307 that is cleaning up related unused code. I'll see if I can add a test.

PR #3313, opened 2024-03-05
Test: Update e2e tests for Windows

abeatrix:
Attempt to improve the e2e Windows tests that sometimes fail on main

Also added a new response for the document command in the mock server as the current response `<title>Goodbye Cody</title>` is not related to the test and that it is reformated differently in Typescript code on Windows.
 

## Test plan

CI should be green.


vscode/test/e2e/README.md
```diff
@@ -1,24 +1,38 @@
...
+## Notes
+
+1. Flakiness in tests can often be attributed to timing issues, such as asynchronous actions taking longer than expected to complete. By performing a hover action before a click action, you give the application a bit more time to react and stabilize before attempting the subsequent click. This can help mitigate timing-related flakiness.
+
+   Example:
+
+   ```
+   await page.getByLabel('.vscode', { exact: true }).hover()
+   await page.getByLabel('.vscode', { exact: true }).click()
+   ```
```
philipp-spiess, line 38: 😱 This is scary. How is it better to hover before we click when both actions are supposed to [auto-wait](https://playwright.dev/docs/actionability)? I would expect both actions to have the same timeout so if something is not there when we click it, it will also not be there if we `hover` it. This is clearly not how it works if you have had this finding 🤔

abeatrix: Haha yea we have been able to resolve flaky tests by adding `hover()` before `click()` ([example 1](https://sourcegraph.slack.com/archives/C05AGQYD528/p1707858463955029?thread_ts=1707850515.546129&cid=C05AGQYD528), [example 2](https://sourcegraph.slack.com/archives/C05AGQYD528/p1707766896208689?thread_ts=1707744071.460429&cid=C05AGQYD528)).

PR #3315, opened 2024-03-05
handle incorrectly serialized vscode.Range values in toRangeData

sqs:
We should fix the source of any bugs like this, but it's also helpful to handle them gracefully.

## Test plan

CI

### DIFF
--- a/lib/shared/src/common/range.ts
+++ b/lib/shared/src/common/range.ts
@@ -14,11 +14,21 @@ export interface RangeData {
  */
 export function toRangeData<R extends RangeData>(range: R): RangeData
 export function toRangeData<R extends RangeData>(range: R | undefined): RangeData | undefined
-export function toRangeData<R extends RangeData>(range: R | undefined): RangeData | undefined {
-    return range
+export function toRangeData(
+    badVSCodeSerializedRange: [{ line: number; character: number }, { line: number; character: number }]
+): RangeData
+export function toRangeData<R extends RangeData>(
+    range: R | [{ line: number; character: number }, { line: number; character: number }] | undefined
+): RangeData | undefined {
+    // HACK: Handle if the `vscode.Range` value was accidentally JSON-serialized as [start, end],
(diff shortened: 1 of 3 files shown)


lib/shared/src/common/range.ts
```diff
@@ -15,10 +15,14 @@ export interface RangeData {
 export function toRangeData<R extends RangeData>(range: R): RangeData
 export function toRangeData<R extends RangeData>(range: R | undefined): RangeData | undefined
 export function toRangeData<R extends RangeData>(range: R | undefined): RangeData | undefined {
-    return range
+    // HACK: If value was accidentally serialized as [start, end] (which `vscode.Range` instances do
+    // when JSON-serialized), handle it.
+    const data = Array.isArray(range) ? ({ start: range[0], end: range[1] } as RangeData) : range
```
dominiccooney, line 20: You should elaborate the argument type/constraints on R to reflect the function handling this, because a two element array does not `extends RangeData`.

dominiccooney (approved):
Looks good but let's keep the types accurate.

lib/shared/src/common/range.ts
sqs: I'll clarify the comment. The problem is that `vscode.Range#toJSON()` returns `[Position, Position]` at runtime (and this is not captured by the type system of course). So, I don't think this function should accept the `[Position, Position]` type, because that would imply to callers that that is an intentional type of the argument. If you think it should accept that, I'm happy to defer to you.

dominiccooney: I understand your point; you don't want `[number, number]` callers to proliferate. On the other hand, regardless of what we want, we *do* have `[number, number]` callers and do accept `[number, number]` now. (...in fact we'll accept any Array and assert things like `{ start: 'foo', end: undefined }` are RangeData 😓 )
 
My $0.02: The compiler didn't catch the problem because somewhere along the line, someone's dishonest about the type of the value being passed here. Two wrongs don't make a right.

But it is not more important than any of the million small decisions that go into software so I leave it up to your judgement.

PR #3318, opened 2024-03-06
Edit: Fix userContextFiles missing from context response

umpox:
## Description

Ensures we populate the user context files with the associated content when building our context messages.

We already do this for Chat, we just also need to do the same for Edit too.

## Test plan

1. Start an instruction-based edit with Opt+K
2. @ mention a file
3. Execute an edit, check that the file that was mentioned is included as context


abeatrix (approved):

sqs:
Thanks! I can't tell from the PR description, but sorry if I regressed this. 😄

PR #3323, opened 2024-03-06
context ranker: adding sentence transformer embeddings as a context r…

hitesh-1997:
…etriver source

## Context

1. Adding sentence transformer embeddings as a context retrieval source. Caller can specify the type of embedding model to use as a context source.

2. Adding the support for logging which logs the ranker features, 
- Embedding health checks
- BM25 score detailed breakdown
- All possible embeddings with overlapping stats (such as # lines overlapping etc) -- can be used for offline analysis of embedding selection strategy. 
- ranker features and final payload sent
- ranker prediction scores
- latencies such as ranker e2e latency.

**Note**: The logging payload will be saved in a temporary file in the system and the system path will be printed in the console.
See last second line in image below: `ContextRankingController: rank-items-logger-payload saving the logging at path`

image

3. Support for getting variable number of context items from the local embeddings

4. Pre-computing the query embedding (in parallel to fetching context from symf, local-embeddings etc.)

## Previous Related PR
1. https://github.com/sourcegraph/cody/pull/3111

## Dependent PR
Note: This PR should be merged only after releasing the BFG and Symf.

1. https://github.com/sourcegraph/bfg-private/pull/173/files
2. https://github.com/sourcegraph/symf-private/pull/10

### Manul Integration test
I have tested the approach on macos and linux based system.

1. Build a symf binary from: https://github.com/sourcegraph/symf-private/pull/10
2. Build a BFG binary from: https://github.com/sourcegraph/bfg-private/pull/173  
3.  Setup the following flag in the vscode user settings
```
 "cody.experimental.cody-engine.path": <path-to-bfg-binary-step-2>,
 "cody.experimental.symf.path": <path-to-symf-binary-step-1> 
  "cody.experimental.chatContextRanker": true
```

4. Checkout to this branch and start the vscode debugger.
5. Open a repo and click on the chat command from the vscode commands. This will start the feature computation job 
6. Testing ranker call: Adding the breakpoint at 
 a) Sent transformer Embedding retrieval 
 b) Embedding to check number of embeddings other than 15
 c) Temporary adding the payload to a file and analyzing.

https://github.com/sourcegraph/cody/assets/20701220/27c2f465-d91d-4e38-aa84-ad53d1bf7e83

## ToDo:
1. Changing the binary version for bfg and symf.

### DIFF
--- a/vscode/src/chat/chat-view/context.ts
+++ b/vscode/src/chat/chat-view/context.ts
@@ -20,7 +20,6 @@ import type { ContextRankingController } from '../../local-context/context-ranki
 import type { LocalEmbeddingsController } from '../../local-context/local-embeddings'
 import type { SymfRunner } from '../../local-context/symf'
 import { logDebug, logError } from '../../log'
-import { ContentProvider } from '../../non-stop/FixupContentStore'
 
 interface GetEnhancedContextOptions {
     strategy: ConfigurationUseContext
@@ -195,8 +194,10 @@ async function getEnhancedContextFromRanker({
             'model-specific-embeddings'
         )
 
+        const precomputeQueryEmbeddingPromise = contextRanking?.precomputeContextRankingFeatures(text)
+
         const localSearchContextItemsPromise = providers.symf
-            ? retrieveContextGracefully(searchSymf(providers.symf, editor, text, false, 20), 'symf')
+            ? retrieveContextGracefully(searchSymf(providers.symf, editor, text), 'symf')
             : []
 
         const remoteSearchContextItemsPromise = providers.remoteSearch
@@ -216,6 +217,7 @@ async function getEnhancedContextFromRanker({
                 embeddingsContextItemsPromise,
                 keywordContextItemsPromise,
                 modelSpecificEmbeddingsContextItemsPromise,
+                precomputeQueryEmbeddingPromise,
             ])
 
         searchContext = searchContext
@@ -264,8 +266,7 @@ async function searchSymf(
     symf: SymfRunner | null,
     editor: VSCodeEditor,
     userText: string,
-    blockOnIndex = false,
-    numResults = 15
+    blockOnIndex = false
 ): Promise<ContextItem[]> {
     return wrapInActiveSpan('chat.context.symf', async () => {
         if (!symf) {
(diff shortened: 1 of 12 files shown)


vscode/src/chat/chat-view/context.ts
```diff
@@ -349,6 +362,42 @@ async function searchEmbeddingsLocal(
...
+    text: string
+): Promise<ContextItem[]> {
+    return wrapInActiveSpan('chat.context.model-specific-embeddings.local', async () => {
+        if (!contextRankingController) {
+            return []
+        }
+
+        logDebug('SimpleChatPanelProvider', 'getEnhancedContext > searching model specific embeddings')
+        const contextItems: ContextItem[] = []
+        const numResults = 20
```
dominiccooney, line 376: @rishabhmehrotra mentioned retrieving and reranking more results and logging the "depth." Is 20 enough for what's planned there?

vscode/src/chat/chat-view/context.ts
```diff
@@ -349,6 +362,42 @@ async function searchEmbeddingsLocal(
...
+                new vscode.Position(result.startLine, 0),
+                new vscode.Position(result.endLine, 0)
+            )
+
+            contextItems.push({
+                type: 'file',
+                uri: result.uri,
+                range,
+                content: result.content,
+                source: ContextItemSource.Embeddings,
```
dominiccooney, line 394: Can you clarify with @rishabhmehrotra whether you want to divide these context items by model (probably because Triton is coming), or local embeddings retrieval versus "model-specific"? If so, you probably want to use a different source here.

vscode/src/chat/chat-view/context.ts
```diff
@@ -349,6 +362,42 @@ async function searchEmbeddingsLocal(
     })
 }
 
+async function searchModelSpecificEmbeddings(
```
dominiccooney, line 365: The naming or factoring of this method is odd. On the one hand, every embeddings method ends up using *some* specific model. And this method uses a different (fixed) one.

A sounder approach to this is to let the embeddings client specify models when creating an index, support multiple indexes for the same repo, and let the client specify preferred models when querying. But we're following the policy of getting things done quickly, so this is fine, but move it to the context ranking controller. It's really embeddings, not context ranking, and belongs in the embeddings crate but maybe that's too much work for you to take on in the exploration phase.

dominiccooney (approved):
Looks good because it is gated by the reranker flag. Let me emphasize two bits of feedback:

- Move `searchModelSpecificEmbeddings`, it really belongs with the rest of embeddings, it doesn't belong *here.* But generalizing embeddings to multiple models might be too much work for the move fast approach (although we'll need it for Triton...) so move this to the context ranking controller which supports the operation. In the long run it should be integrated into the embeddings crate & controller instead of having the context ranker route around it.
- Better not to log two different things as "embeddings" source if you want to split them later.

vscode/src/chat/chat-view/context.ts
hitesh-1997: checked, changed to 50

vscode/src/chat/chat-view/context.ts
hitesh-1997: I added a field modelName to indicate the name of the model. does that look okay ?

PR #3324, opened 2024-03-06
Command: do not refresh sidebar on config change

abeatrix:
FIX bug where command sidebar disappears on configuration change ([slack](https://sourcegraph.slack.com/archives/C05AGQYD528/p1709745282681469))

image

Root cause: Introduced in https://github.com/sourcegraph/cody/pull/3245 by me 😅 , the tree view provider for commands is refreshed without tree items provided on configuration change, causing the sidebar view to be empty

Proposed fix: Since commands are not configurable, there is no need to refresh the sidebar on configuration change. This PR removes the aforementioned code. 

Note: Changelog entry is not required as the change has not been released

## Test plan

2. Update any of your Cody extension settings
3. Confirm the command sidebar is not empty
4. Update any of your Custom Command
5. Confirm the change is reflected in the sidebar

https://github.com/sourcegraph/cody/assets/68532117/441234d8-4b37-4ae7-84fa-a7d820aebb03

### DIFF
--- a/vscode/src/commands/services/provider.ts
+++ b/vscode/src/commands/services/provider.ts
@@ -35,12 +35,7 @@ export class CommandsProvider implements vscode.Disposable {
             vscode.commands.registerCommand('cody.menu.commands', () => this?.menu('default')),
             vscode.commands.registerCommand('cody.menu.custom-commands', () => this?.menu('custom')),
             vscode.commands.registerCommand('cody.menu.commands-settings', () => this?.menu('config')),
-            vscode.commands.registerCommand('cody.commands.open.doc', () => openCustomCommandDocsLink()),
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):

sqs:
LGTM although I couldn't figure out why/where the tree items were being cleared.

abeatrix:
> LGTM although I couldn't figure out why/where the tree items were being cleared.



> LGTM although I couldn't figure out why/where the tree items were being cleared.

Looks like we have treeItems and treeNodes and on refresh, we only reset the treeNodes that the commands sidebar added. Will look into the git blame to see why treeItems were added, see if we can simplify it in a follow-up PR :)

PR #3330, opened 2024-03-07
Add support link for Cody Pro & Enterprise

toolmantim:
Adds Support to the sidebar and settings quickpick:

CleanShot 2024-03-07 at 12 40 16@2x

CleanShot 2024-03-07 at 12 40 23@2x

- [x] Fix Support to only show for Pro & Ent in quickpick

## Test plan

- Open sidebar on pro & ent plans. Click Support. Verify works.
- Open settings quickpick on pro & ent plans. Click Support. Verify works.
- Open sidebar on free plan. Verify no support link.
- Open settings quickpick on free plan. Verify no support link.


toolmantim:
@enriquegh it would be good to fill the current Cody for VS Code version automatically, and pre-fill https://help.sourcegraph.com/hc/en-us/requests/new — is it possible to pass it URL params? We could pass the version number, and account type.

toolmantim:
@abeatrix could I please get your assistance making the display of the quickpick item conditional, similar to how it works in the treeview?

toolmantim:
@abeatrix doh, I forget to check in the code that needed the conditionals! Fixed in d8b96dd53

abeatrix:
@toolmantim pushed the changes to the branch 👍

abeatrix (approved):
Awesome!

It would be nice to add a e2e test for this, but we can do it in a follow-up PR 🎉

vscode/src/services/tree-views/treeViewItems.ts
```diff
@@ -68,6 +69,12 @@ const supportItems: CodySidebarTreeItem[] = [
         icon: 'book',
         command: { command: 'cody.sidebar.documentation' },
     },
+    {
+        title: 'Support',
+        icon: 'question',
+        command: { command: 'cody.sidebar.support' },
+        requireNotFree: true,
```
umpox, line 76: nit: confusing var name, would just change this to something like `requirePaid`

umpox (approved):

toolmantim:
Added some e2e tests but something funky is going on with the unit tests now 🤔

abeatrix:
> Added some e2e tests but something funky is going on with the unit tests now 🤔

That's probably related to an issue @valerybugakov is working on

valerybugakov:
Yeah, unit tests should be fixed on main now 👍

PR #3335, opened 2024-03-07
remove unused CodeBlockMetadata

sqs:
This telemetry metadata was not being assigned anywhere, so it was always undefined (confirmed at runtime as well).

## Test plan

CI, and confirm at runtime that the metadata fields were not being populated.


abeatrix (approved):
Thank you!

PR #3342, opened 2024-03-07
Debug: Enable debug mode from setting menu

abeatrix:
Part of Better debugging facilities for enterprise issues

New commands to make debugging Cody easier:
- `cody.debug.outputChannel`
  - open the Cody output channel
- `cody.debug.enable.all`
  - enable both `cody.debug.enable` and `cody.debug.verbose` and open output channel with one click

New sidebar menu item to open output channel

image

New button in the Cody Setting menu to enable debug mode and open output channel on click

image

All debug commands are also grouped together and are available through the command palette

image

## Test plan

https://github.com/sourcegraph/cody/assets/68532117/3ec5a97b-8d09-4b72-92a5-d08f17bad72a

### DIFF
--- a/vscode/src/services/utils/export-logs.ts
+++ b/vscode/src/services/utils/export-logs.ts
@@ -54,9 +54,13 @@ export async function exportOutputLog(logUri: vscode.Uri): Promise<void> {
             })
     } catch (error) {
         // Open the output channel instead
-        void vscode.commands.executeCommand(
-            'workbench.action.output.show.extension-output-sourcegraph.cody-ai-#1-Cody by Sourcegraph'
-        )
+        openCodyOutputChannel()
         console.error(error)
     }
+}
+
+export function openCodyOutputChannel(): void {
+    void vscode.commands.executeCommand(
+        'workbench.action.output.show.extension-output-sourcegraph.cody-ai-#1-Cody by Sourcegraph'
+    )
 }
--- a/vscode/src/services/utils/export-logs.ts
+++ b/vscode/src/services/utils/export-logs.ts
@@ -1,5 +1,5 @@
 import * as vscode from 'vscode'
-import { CODY_OUTPUT_CHANNEL } from '../../log'
+import { CODY_OUTPUT_CHANNEL, outputChannel } from '../../log'
 
 /**
  * Exports the output log file to a specified location.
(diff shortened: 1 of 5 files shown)


slimsag:
Thank you so much for working on this!

> New sidebar menu item to open output channel

Very helpful!

> New button in the Cody Setting menu to enable debug mode and open output channel on click

Also very helpful!

Could the button to enable debug mode also be available under the `...` menu?

image

That way all the debugging tools are nested under there (having a one-stop shop for that stuff is super nice)

abeatrix:
@slimsag Glad you find it helpful 🤗  Love that idea! I have updated it so that if `cody.debug.verbose` is not enabled, the `Enable Debug Mode` item will show up in the ... menu. This will make it easier for us to tell if the user has enabled debug mode or not in one place, wdyt?

When `cody.debug.verbose` is on:

image

When `cody.debug.verbose` is off:

image

slimsag:
Perfect!

vscode/src/services/utils/export-logs.ts
```diff
@@ -54,9 +54,25 @@ export async function exportOutputLog(logUri: vscode.Uri): Promise<void> {
...
+        openCodyOutputChannel()
         console.error(error)
     }
 }
+
+export function openCodyOutputChannel(): void {
+    void vscode.commands.executeCommand(
+        'workbench.action.output.show.extension-output-sourcegraph.cody-ai-#1-Cody by Sourcegraph'
+    )
+}
```
philipp-spiess, line 66: There's a more reliable way to do this that doesn't require hardcoding the extension ID and a magical "#1":

https://github.com/sourcegraph/cody/blob/fcb91b002c9a9da25151bbddf9481ed8b9887230/vscode/src/dev/helpers.ts#L22

```suggestion
export function openCodyOutputChannel(): void {
    outputChannel.show()
}
```

vscode/src/services/utils/export-logs.ts
```diff
@@ -54,9 +54,25 @@ export async function exportOutputLog(logUri: vscode.Uri): Promise<void> {
...
+
+/**
+ * Enables debug mode by updating workspace configuration settings.
+ * Sets 'cody.debug.enable' and 'cody.debug.verbose' to true globally.
+ * Opens the Cody output channel.
+ */
+export function enableDebugMode(): void {
+    const workspaceConfig = vscode.workspace.getConfiguration()
+    void workspaceConfig.update('cody.debug.enable', true, vscode.ConfigurationTarget.Global)
```
philipp-spiess, line 75: If `true` is the default then we just need to delete it if it's set instead?

philipp-spiess (approved):
Awesome but there's a more elegant way to open the output view, let's change this quickly 😅

philipp-spiess:
oh and don't forget about adding a change log entry 😎

Issue #3311, opened 2024-03-05
Test command should generate file name with correct file path


abeatrix:
Reported by Dan Diemer on [Slack](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1709594478210299):

> Cody is creating a new test file in the src path with the same name as the actual test file for the code. The file should be lived in the /test path instead

From beyang: 

> we probably need to add something to the test generation command to search for existing test files in parallel directory hierarchies. It's not uncommon to have parallel directory hierarchies under top-level src/ and test/ directories in many codebases (this is in contrast to the pattern in most Sourcegraph codebases, where unit test files are often colocated with the files they test).

csells:
We've heard VIP customer feedback that they'd like their unit tests to be generated to match not just their testing framework but also the coding style, i.e. single quotes instead of double quotes and not generating semi-colons at the end of lines.

PR #3344, opened 2024-03-08
Command: Generate file path for test command

abeatrix:
Fix https://github.com/sourcegraph/cody/issues/3311

- Updated the generic prompt rules to anticipate a file path for the test suite.
- Modified the edit provider to achieve the following:
- Extract the full file path instead of just the file name.
- Utilize workspace URI, if available, to resolve the full path.
- Default to the current directory if no workspace is open.

This adjustment enables the LLM to propose paths relative to the workspace root (as included in the shared context) for the generated test code. This enhancement ensures that the test files can be created in the correct directory, rather than always being saved in the same directory as the current file. This is particularly beneficial as it aligns with the intended location for test files.

### Before

The newly generated test file will be saved at the same level as the current file even when there is an existing test file with the same name located in a different directory:

image

https://github.com/sourcegraph/cody/assets/68532117/10cf9341-ca67-41da-9058-9e5c950c9438

### After

The same test that was generated by the LLM is now magically appended to the current file that lives outside the current directory

image

https://github.com/sourcegraph/cody/assets/68532117/123d88c2-f7f9-4b90-9bf6-ec15b1257a97


vscode/src/commands/context/unit-test-file.ts
```diff
@@ -23,20 +23,24 @@ export async function getContextFilesForUnitTestCommand(file: URI): Promise<Cont
...
         const max = 10
 
-        // Search for test files in the current directory first
-        const curerntDirPattern = getSearchPatternForTestFiles(file, searchInCurrentDirectoryOnly)
+        // Search for a test file that has the same file name first
+        const sameNameTestPattern = getSearchPatternForTestFiles(file, !searchCurrentDirectoryOnly, true)
+        const testWithSameName = await getWorkspaceFilesContext(sameNameTestPattern, excludePattern, 1)
+
+        // Search for test files in the current directory
+        const curerntDirPattern = getSearchPatternForTestFiles(file, searchCurrentDirectoryOnly)
```
umpox, line 35:
```suggestion
        const currentDirPattern = getSearchPatternForTestFiles(file, searchCurrentDirectoryOnly)
```

nit: typo

vscode/src/commands/context/unit-test-file.ts
```diff
@@ -23,20 +23,24 @@ export async function getContextFilesForUnitTestCommand(file: URI): Promise<Cont
...
         // The max number of files to search for in each workspace search
         const max = 10
 
-        // Search for test files in the current directory first
-        const curerntDirPattern = getSearchPatternForTestFiles(file, searchInCurrentDirectoryOnly)
+        // Search for a test file that has the same file name first
+        const sameNameTestPattern = getSearchPatternForTestFiles(file, !searchCurrentDirectoryOnly, true)
+        const testWithSameName = await getWorkspaceFilesContext(sameNameTestPattern, excludePattern, 1)
+
+        // Search for test files in the current directory
```
umpox, line 34: If we found a test file with the same name, is there any point looking for other test files?

vscode/src/commands/context/unit-test-file.ts
```diff
@@ -23,20 +23,24 @@ export async function getContextFilesForUnitTestCommand(file: URI): Promise<Cont
...
+        const curerntDirPattern = getSearchPatternForTestFiles(file, searchCurrentDirectoryOnly)
         const currentDirContext = await getWorkspaceFilesContext(curerntDirPattern, excludePattern, max)
 
-        contextFiles.push(...currentDirContext)
+        contextFiles.push(...testWithSameName, ...currentDirContext)
 
         // If no test files found in the current directory, search the entire workspace
-        if (!contextFiles.length) {
-            const wsTestPattern = getSearchPatternForTestFiles(file, !searchInCurrentDirectoryOnly)
+        if (contextFiles.length < 3) {
```
umpox, line 41: We probably need to update this comment to explain why `3`, does this have a noticeable benefit over `!contextFiles.length`?

vscode/src/edit/prompt/context.ts
```diff
@@ -47,6 +47,7 @@ const getContextFromIntent = async ({
          * The preceding code is already included as part of the response to better guide the output.
          */
         case 'test':
+            return []
```
umpox, line 50: The doc string should still be attached to case 'add` here, but ignore me if you're still playing around with this.

I'm curious if this shouldn't use the same context as `add` though, as if we're appending to an existing test file then it seems useful that we include the surrounding code? Maybe we should just include the preceding code for test, as we'll always insert at the bottom?

umpox (commented):
Nice! Left some small comments

vscode/src/commands/context/unit-test-file.ts
abeatrix: ah this was a left over, will remove it!

vscode/src/edit/prompt/context.ts
abeatrix: I removed it because we already add it when we generate the context for test, so the `add` was adding duplicated context in this case. Also, we are appending to an existing test file but the doc file it reads here is the origin file which is a non-test file (if this makes sense)

vscode/src/commands/context/unit-test-file.ts
abeatrix: I'd say yes as it'd be helpful to use other test files as references in terms of patterns and library usage. It also helps in cases where the test file with the same name is empty or short that's unhelpful 😄

vscode/src/edit/prompt/context.ts
umpox: 👍  Can we add a docstring explaining this and move the docstring that's existing to above `case 'add':`

vscode/src/commands/context/unit-test-file.ts
umpox: does this still need removing?

umpox (approved):
Approved, left a few comments that need changing above

PR #3341, opened 2024-03-07
Chat: Shows welcome message on empty chat only

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2461

image

This PR removes the welcome message after the first question has been submitted.

## Test plan

1. Open a new chat to confirm welcome message is displayed
2. Submit a question / execute a command
3. the welcome message should disappear

## Before

image

PR #3345, opened 2024-03-08
Chat: Remove welcome message after first chat

abeatrix:
Follow up on https://github.com/sourcegraph/cody/pull/3341#issuecomment-1984793796

Display the welcome message on the first chat ever created based on local chat history, instead of first chat message submitted in a single chat session.

## Test plan

image


toolmantim:
Thank you! I'll check out how this feels

PR #3346, opened 2024-03-08
Auth: Fix pro status logic

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1709840673141329

- Check if the user has an active PRO subscription using getCurrentUserCodySubscription()
- Consider users with a non-PENDING PRO plan as valid pro users
- Set userCanUpgrade based on the active pro user status

This change corrects the detection of pro users by checking their actual subscription status instead of relying on the codyProEnabled flag. It ensures that only users with an active PRO plan are considered valid pro users as we have already implemented in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/notifications/cody-pro-expiration.ts?L115-117, but the edit code lens that relies on the AuthStatus was not updated correctly in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/non-stop/codelenses/items.ts?L60-69

This PR should fix the issue

## Test plan

Naman to confirm

### DIFF
--- a/agent/recordings/rateLimitedClient_2043004676/recording.har.yaml
+++ b/agent/recordings/rateLimitedClient_2043004676/recording.har.yaml
@@ -567,11 +567,11 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: a398c9f2c8001eb1c872f37647a399bf
+    - _id: 9f46eb98120189f0b3face0b705225fd
       _order: 0
       cache: {}
       request:
-        bodySize: 115
+        bodySize: 268
         cookies: []
         headers:
           - _fromType: array
@@ -589,7 +589,7 @@ log:
             value: "*/*"
           - _fromType: array
             name: content-length
-            value: "115"
+            value: "268"
           - _fromType: array
             name: accept-encoding
             value: gzip,deflate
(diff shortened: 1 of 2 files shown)


thenamankumar (approved):
This looks good but once I merge my [PR](https://github.com/sourcegraph/sourcegraph/pull/60940) this change will not be required and the client can either depend solely on codyProEnabled or plan === 'PRO'. As part of my PR status will never be 'PENDING'.

Given that, do you think we should still make this change only to have the logic not required once I merge my PR?

abeatrix:
> This looks good but once I merge my [PR](https://github.com/sourcegraph/sourcegraph/pull/60940) this change will not be required and the client can either depend solely on codyProEnabled or plan === 'PRO'. As part of my PR status will never be 'PENDING'.
> 
> 
> 
> Given that, do you think we should still make this change only to have the logic not required once I merge my PR?

Since solely relying on codyProEnabled is not the right way to do it at the moment, and this change will continue to work after your Pr is merged, I think it'd be easier to keep the change just so it aligns with how it is currently work in the other part of the code. I will keep track of your PR though. Thank you so much for the review David 😆

thenamankumar:
Cool, let's merge it then!

PR #3351, opened 2024-03-08
fix bug where the entire document text would be included instead of just visible content

sqs:
In https://github.com/sourcegraph/cody/pull/3269, the `ContextItem` and `ContextFile` types were merged. The former had a `text` field that I neglected to update in that PR to use the new `content` name (and it was not caught by the typechecker).

This could lead to poor context in chat whenever `getEnhancedContext` is used. Instead of including just the visible portion of the file, the entire file's content would be included.

## Test plan

I have not been able to repro this, but it is obviously a bug.


sqs:
@philipp-spiess This is backport-worthy, don't you think? Sorry about this bug!

valerybugakov (approved):

PR #3353, opened 2024-03-08
document and reorganize ContextItem source, title, etc.

sqs:
Mostly just adds docs. Makes `ContextItemSource` an enum because each item had relevant documentation.

Also now uses `isTooLarge?: boolean` instead of `title === 'large-file'` to indicate the file is too large.

## Test plan

CI


umpox (approved):
👍

PR #3354, opened 2024-03-08
disable model selector after 1st chat message is sent

sqs:
In https://github.com/sourcegraph/cody/pull/3341 the behavior changed so that the welcome message goes away when you chat. (Which I love!) But there's a minor bug in that the model selector remains enabled after the 1st message is sent now (because the existing logic assumed that the welcome message was always the 1st message and so checked for `>1` messages to disable the model selector).

Also adds an e2e test for the model selector.

## Test plan

e2e test covers the following test plan: Open new chat. Ensure model selector is enabled. Send a message. Ensure the model selector becomes immediately disabled.

### DIFF
--- a/vscode/test/e2e/chat-input.test.ts
+++ b/vscode/test/e2e/chat-input.test.ts
@@ -1,7 +1,8 @@
 import { expect } from '@playwright/test'
 
+import * as mockServer from '../fixtures/mock-server'
 import { sidebarExplorer, sidebarSignin } from './common'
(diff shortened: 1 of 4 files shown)


umpox (approved):
Thanks for adding the test!

philipp-spiess (approved):

abeatrix (approved):
Thanks for fixing this and for adding the e2e test ❤️

PR #3363, opened 2024-03-09
compute ChatMessage.displayText on-demand instead of persisting

sqs:
Previously, `ChatMessage.displayText?: string` was set in multiple places, and the logic for getting the title of a chat was defined in 2 different places (and uses the `displayText`). Now, we never persist `displayText` and instead compute it at display time, which is possible because it uses only data that is available in the `ChatMessage` value.

Also simplify and test reformatBotMessageForChat.

This is a refactor to clear the way for #3287.

## Test plan

CI. Also manually test that chat titles are set based on the last human message in a new chat, and that editing a chat title in the sidebar updates the title in the sidebar and the tab title bars.


sqs:
Note: need to confirm that the issue in https://github.com/sourcegraph/cody/pull/3367 does not occur with this change.

philipp-spiess (approved):
Nice!

PR #3365, opened 2024-03-10
Fix: Fixing the @-mention line numbering

PriNova:
This PR fixes the @-mention file context fetching with single-line, and multi-line ranges in the Chat interface.

The screenshot shows that the file is fetched multiple times when the user specifies a line number range for it.

Screenshot 2024-03-10 012225

This PR corrects the RegEx patterns and separates the single-line and multi-line use cases into different loops. It also refactors the code for better readability.

## Test plan

Tested manually with single file and multiple files in a test matrix within the Chat view:
1. No line numbers for the whole file
2. Single line number for only a single line in a file
3. Multiple line numbers for a range of lines in a file

### DIFF
--- a/lib/shared/src/chat/input/user-context.ts
+++ b/lib/shared/src/chat/input/user-context.ts
@@ -41,25 +41,29 @@ export function verifyContextFilesFromInput(
         }
 
         // Get the line number behind the file name if there is one:
-        // SingleLines Example: foo/bar.ts:1 > 1
-        const singleLines = input.matchAll(new RegExp(atFileName + ':(\\d+)(?!-)', 'g'))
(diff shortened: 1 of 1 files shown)


sqs (changes requested):
Thank you. Could you please update the unit tests in `user-context.test.ts`?

FYI, we are moving to a different way of passing the at-mentioned context files from the chat webview to the backend in #3287. This uses a rich editor instead of just encoding everything as text and parsing out `@`-mentions from the text. That will make this change irrelevant. I am not sure when that PR will land, but I hope in the next 1-2 weeks, so if this is not an urgent bug for you, it is fine to just close this PR.

PR #3371, opened 2024-03-11
escapeHTML that does not use the DOM

sqs:
This is an equivalent implementation of `escapeHTML` that just uses regexps and not the DOM, so it's usable in non-DOM contexts.

These replacements are sufficient and match the OWASP recommendation for characters to encode: https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#output-encoding-for-html-contexts.

## Test plan

CI

### DIFF
--- a/lib/shared/src/common/markdown/markdown.ts
+++ b/lib/shared/src/common/markdown/markdown.ts
@@ -5,12 +5,15 @@ import { marked } from 'marked'
 // TODO(sqs): copied from sourcegraph/sourcegraph. should dedupe.
 
 /**
- * Escapes HTML by replacing characters like `<` with their HTML escape sequences like `&lt;`
+ * Escapes HTML by replacing characters like `<` with their HTML escape sequences like `&lt;`.
  */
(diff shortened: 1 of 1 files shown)


lib/shared/src/common/markdown/markdown.ts
```diff
@@ -5,12 +5,15 @@ import { marked } from 'marked'
...
-    const span = document.createElement('span')
-    span.textContent = html
-    return span.innerHTML
+export function escapeHTML(html: string): string {
+    return html
+        .replaceAll('&', '&amp;')
+        .replaceAll('<', '&lt;')
+        .replaceAll('>', '&gt;')
+        .replaceAll('"', '&quot;')
+        .replaceAll("'", '&#039;')
```
philipp-spiess, line 16: TIL that `&#<HEX>` is for ASCII and`&#x<HEX>` for unicode

philipp-spiess (approved):
Agree this should be fine and is following the guidelines

PR #3380, opened 2024-03-11
cody.experimental.showChatScores setting

beyang:
Shows the search scores of chat context items when enabled. This is in preparation for adjusting the chat context engine to take better advantage of the combination of (1) symf, (2) Sourcegraph search, (3) embeddings produced by the new model, (4) the reranker (which currently keys off of embeddings and BM25 signals).

image

## Test plan

Test locally. Should not affect regular use. When the flag is turned on, it should only modify the display of enhanced context links.

### DIFF
--- a/vscode/webviews/App.tsx
+++ b/vscode/webviews/App.tsx
@@ -14,7 +14,7 @@ import {
     type SerializedChatTranscript,
 } from '@sourcegraph/cody-shared'
 import type { UserAccountInfo } from './Chat'
-import { EnhancedContextEnabled } from './chat/components/EnhancedContext'
+import { EnhancedContextEnabled, IncludeScores } from './chat/components/EnhancedContext'
 
 import type { AuthMethod, LocalEnv } from '../src/chat/protocol'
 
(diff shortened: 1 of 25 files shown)


vscode/webviews/App.tsx
```diff
@@ -242,29 +244,31 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
...
-                                        telemetryService={telemetryService}
-                                        isTranscriptError={isTranscriptError}
-                                        chatModels={chatModels}
-                                        setChatModels={setChatModels}
-                                        welcomeMessage={getWelcomeMessageByOS(config?.os)}
-                                        guardrails={attributionEnabled ? guardrails : undefined}
-                                        chatIDHistory={chatIDHistory}
-                                        isWebviewActive={isWebviewActive}
-                                    />
+                                    <IncludeScores.Provider value={config.experimentalShowChatScores}>
```
dominiccooney, line 247: React Context + providers FTW. I think this is much better than passing so many props.

vscode/webviews/Components/FileLink.tsx
```diff
@@ -46,7 +49,36 @@ export const FileLink: React.FunctionComponent<FileLinkProps> = ({
...
             {pathWithRange}
         </button>
     )
 }
+
+// Pretty-print the score, so that at most 2 sig figs are used, and if a number
+// is greater than one million, 'M' is used, and if a number is between one
+// thousand and one million, 'K' is used
+function prettyPrintScore(score: number): string {
+    if (score > 1000000) {
```
dominiccooney, line 62: Style nit, you could use _s here too for consistency.
```suggestion
    if (score > 1_000_000) {
```

vscode/webviews/Components/FileLink.tsx
```diff
@@ -46,7 +49,36 @@ export const FileLink: React.FunctionComponent<FileLinkProps> = ({
...
+        return `${(score / 1_000_000).toFixed(1)}M`
+    }
+    if (score > 1000) {
```
dominiccooney, line 65:
```suggestion
    if (score > 1_000) {
```

dominiccooney (approved):
Splendid.

PR #3381, opened 2024-03-12
sourcegraph-api: always include anonymous UID

bobheadxi:
Updates the condition for including `X-Sourcegraph-Actor-Anonymous-UID` to always do it, instead of only doing it on unauthenticated users.

## Test plan

n/a


akalia25 (approved):
TY!!

PR #3386, opened 2024-03-12
Agent: Bundle WASM artifacts

### LINKED ISSUE #3302: Cody Agent should always use WASM artifacts
Right now, Cody Agent does not always supports TreeSitter using WASM: https://github.com/sourcegraph/cody/blob/88f88e3a2d991053066dc827ff17b60a140a7b9a/agent/src/AgentWorkspaceConfiguration.ts#L88-L90

This, in turn, will cause some features to not be available like _smarter multi-line completions_ etc.

To support it, we have to make sure the WASM binaries are included in the agent build step and the agent dev environment. @olafurpg looked at this in the past. His work can be found here: https://github.com/sourcegraph/cody/commits/olafurpg/embed-wasm/

We need to:

- [x] Ensure that the cody agent build process copies `.wasm` artifacts over (or, even better, bundles them in the single-binary)
- [x] Ensure that running the cody agent locally (in dev mode) picks up the `.wasm` artifacts
- [x] Do some runtime evaluation to make sure the wasm paths are picked up correctly in all run configurations
- [x] [Remove the `syntacticPostProcessing` flag](https://github.com/sourcegraph/cody/pull/2256)

Related discussion thread: https://sourcegraph.slack.com/archives/C05AGQYD528/p1709571818337319

philipp-spiess:
Closes #3302

This PR ensures that cody agent always has access to the WASM artifacts. While building this, we found out that _it is necessary to bundle the wasm inside the binary:_

```
----stderr----
Aborted(Error: File '/**/dist/tree-sitter.wasm' was not included into executable at compilation stage. Please recompile adding it as asset or script.)
--------------
----stderr----
failed to asynchronously prepare wasm: RuntimeError: Aborted(Error: File '/**/dist/tree-sitter.wasm' was not included into executable at compilation stage. Please recompile adding it as asset or script.). Build with -sASSERTIONS for more info.
Aborted(RuntimeError: Aborted(Error: File '/**/dist/tree-sitter.wasm' was not included into executable at compilation stage. Please recompile adding it as asset or script.). Build with -sASSERTIONS for more info.)
--------------
----stderr----
Uncaught exception: RuntimeError: Aborted(RuntimeError: Aborted(Error: File '/**/dist/tree-sitter.wasm' was not included into executable at compilation stage. Please recompile adding it as asset or script.). Build with -sASSERTIONS for more info.). Build with -sASSERTIONS for more info.
    at abort (/snapshot/dist/agent.js)
    at /snapshot/dist/agent.js
--------------
```

Since we have more than one agent build, this would result in a huge increase in bundle size. We offset that by adding gzip compression. I tried brotli but due to the nature of not having centralized builds yet, it's not worth the additional build time.

Here are the final bundle size changes (for the aarch build):

- Currently the agent binaries are ~82mb
- Without compression they would grow to almost ~146mb
- With compression (gzip) ~64mb

## Test plan

I added a new script that we can use to play around with the binary under `agent/scripts/test-agent-binary.ts`. You can use it to see that WASM works:

- `cd agent/`
- `pnpm run build-agent-binaries`
- `pnpm run test-agent-binary`

This will output something like this. Notice the addition of the _intent_ which requires wasm to load 😎 

Screenshot 2024-03-13 at 13 06 28


agent/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "name": "@sourcegraph/cody-agent",
+  "name": "@sourcegraph/agent",
```
philipp-spiess, line 3: This rename is necessary so the build artifacts are still called `agent-` and not `cody-agent-`. We don't use that name anywhere though so it should be fine.

agent/package.json
```diff
@@ -19,9 +19,21 @@
...
   },
+  "pkg": {
+    "targets": [
+      "latest-linux-arm64",
+      "latest-linux-x64",
+      "latest-macos-x64",
+      "latest-win-x64",
+      "latest-macos-arm64"
+    ],
+    "assets": "dist/*.{wasm,map}"
```
philipp-spiess, line 34: We have to use to this `pkg` syntax and define the wasm artifact here so it'll be inlined properly.

agent/scripts/test-agent-binary.ts
```diff
@@ -0,0 +1,77 @@
...
+    await fs.cp(prototypePath, workspaceRootPath, {
+        recursive: true,
+    })
+
+    const client = new TestClient(
+        {
+            name: 'defaultClient',
+            accessToken,
+        },
+        './dist/agent-macos-arm64'
```
pkukielka, line 36: So that test will only work on windows? Can we pickup proper binary depending on the platform?

pkukielka (approved):
LGTM

agent/scripts/test-agent-binary.ts
philipp-spiess: Good point, I'll copy paste the right version selector here if we ever need to debug on win envs.

PR #3387, opened 2024-03-12
chore: Bump dependencies

evict:
This updates most medium and all high & critical risk rated vulnerabilities in dependencies. These were reported by Dependabot.

## Test plan
CI tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,9 +30,7 @@
     "update-symf-recordings": "rm -rf recordings && CODY_RECORD_IF_MISSING=true vitest vscode/src/local-context/symf.test.ts"
   },
   "devDependencies": {
-    "@babel/traverse": ">=7.23.2",
     "@biomejs/biome": "1.5.2",
-    "@koa/cors": ">=5.0.0",
     "@sourcegraph/tsconfig": "^4.0.1",
     "@storybook/addon-essentials": "^7.0.26",
     "@storybook/addon-interactions": "^7.0.26",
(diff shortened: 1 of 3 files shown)


sqs (dismissed):
@evict What did you run to add these? Was it manual? I see new deps added to `devDependencies` that we don't directly use. If the purpose of that was to pin transitive deps to a specific version, we should use package.json's `resolutions` property instead. Otherwise, the pinned but not-directly-used deps will be removed next time we run something like Knip to remove unused code and deps, and we could fall back to the unsafe dep versions.

evict:
> @evict What did you run to add these? Was it manual? I see new deps added to `devDependencies` that we don't directly use. If the purpose of that was to pin transitive deps to a specific version, we should use package.json's `resolutions` property instead. Otherwise, the pinned but not-directly-used deps will be removed next time we run something like Knip to remove unused code and deps, and we could fall back to the unsafe dep versions.

TIL, thank you! We updated it as dependabot recommended, which was indeed manual.

package.json
```diff
@@ -59,7 +59,17 @@
...
+    "@koa/cors": ">=5.0.0",
+    "axios": ">=1.6.0",
+    "decode-uri-component": ">=0.2.1",
+    "follow-redirects": ">=1.15.4",
+    "http-cache-semantics": ">=4.1.1",
+    "ip": ">=2.0.1",
+    "semver": ">=7.6.0",
+    "tslib": "2.1.0",
+    "vitest": "^1.2.0",
+    "word-wrap": ">=1.2.4"
```
philipp-spiess, line 72: Whoa this is a lot of overwrites. Is this really necessary and can't we just upgrade the packages these depend on to the latest version instead? 🤔  

I’m worried that if we commit this it'll create even more friction to upgrade dependencies later.

package.json
```diff
@@ -59,7 +59,17 @@
...
+    "@babel/traverse": ">=7.23.2",
+    "@koa/cors": ">=5.0.0",
+    "axios": ">=1.6.0",
+    "decode-uri-component": ">=0.2.1",
+    "follow-redirects": ">=1.15.4",
+    "http-cache-semantics": ">=4.1.1",
+    "ip": ">=2.0.1",
+    "semver": ">=7.6.0",
+    "tslib": "2.1.0",
+    "vitest": "^1.2.0",
```
philipp-spiess, line 71: `vitest` is a direct dependency for example. Do we really need to use the `resolutions` feature to pin that to a version.

philipp-spiess (commented):
It looks like the inline snapshot formatter from vitest has gotten an update in one of the version bumps here. Should be fine to run `pnpm test -u` to update these.

My only recommendation is to try upgrading everything to latest versions instead of focusing only on the vulnerable sub-dependenceis to avoid having to use `resolutions` too much since it might make upgrading later feel more tedious. However if there's a lot of breaking changes, this should work too. Let's just be mindful about the change.

package.json
evict: Indeed, we don't. Fixed. 👍

package.json
evict: That makes sense.

PR #3392, opened 2024-03-13
Autocomplete: ship dynamic multiline completions enabled by default

valerybugakov:
## Context

- Based on the A/B test results, we ship dynamic multiline completions enabled by default, removing changes that affect multiline completions.
- [Slack thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1710297548776729?thread_ts=1709623195.735099&cid=C05AGQYD528) with the A/B test results.

## Test plan

CI


valerybugakov:
I'm going to do a second pass of self-review later today, but it's ready for early 👀

PR #3405, opened 2024-03-14
fix counting of context lines (don't count lines w/no chars included)

sqs:
In the context summary in chat, `✨ Context: 56 lines from 5 files` previously counted lines even if no characters on that line were included. For example, a range from line 5 character 10 to line 7 character 0, which is rendered as `5-6`, was counted as 3 lines. It should be counted as 2 lines.

## Test plan

CI


toolmantim (approved):

sqs:
Nevermind. Will land as part of #3287. I initially tried to extract this from #3287 here, but I found that there were other bugs elsewhere in the code (a few different places where we parsed the range strings differently) that meant it would be a more involved change.

Issue #3331, opened 2024-03-07
bug: Cody incorrectly states that it doesn't have the required context, then still answers the question 


chillatom:
### Version

multiple

### Describe the bug

[Slack background](https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1708528867541999)

Users report that Cody consistently responds with "Unfortunately I don't have access to..." even when Cody does in fact have access to the file and it is even explicitly included in context.  Cody still will respond.  

Anecdotal feedback that this occurs more prevalently when selecting remote / multiple repos

![image](https://github.com/sourcegraph/cody/assets/8737510/fa757104-228c-407c-88d7-5fabaa7e46ec)
![image](https://github.com/sourcegraph/cody/assets/8737510/f2b5b9b7-c77b-4e70-a640-94c8fcf84daa)

This happens across IDEs & Web.  

### Expected behavior

If Cody does have access to a file and successfully uses it as context, it should not respond with "unfortunately".

dhdiemer:
![image](https://github.com/sourcegraph/cody/assets/93543113/01054cb0-c44c-4a89-9ae1-73a22c21c869)
IDE: VS Code
Version: 1.6.1
Account Type: Enterprise
Repo Configuration: Local + one remote repo, question is asked of remote repo
Response: ```Unfortunately there is not much information in the code snippets and documentation provided about how authentication specifically works for Cody.

Some relevant information I could gather:

There are various authentication methods supported in Sourcegraph such as SAML, OAuth, OpenID Connect, etc. as mentioned in the Sourcegraph authentication documentation. These could potentially be leveraged by Cody as well.

The Cody VS Code extension code refers to some authentication status checking, including checking if the user is logged in. This implies some authentication mechanism is used.

There are references to an "authenticated user" in some of the TypeScript code, suggesting authenticated users are represented in a certain way.

One code snippet mentions "Authentication Proxy" as an authentication system option, which may be relevant for Cody.

But there are no clear details provided on the exact end-to-end authentication flow or methods used specifically for Cody itself. More context would be needed to fully understand Cody's authentication implementation.```

ajaynz:
Happened to me on demo.sourcegraph.com during a customer demo today using web chat . 
![Screenshot 2024-03-07 at 12 16 54](https://github.com/sourcegraph/cody/assets/1268945/26e011b3-02b5-4798-8e23-7f74188f68d5)

hopkinscl:
[...]

PR #3412, opened 2024-03-14
standardize on a single prompt format for codebase context

sqs:
Previously, there were several very similar (but slightly different) templates for introducing code snippets as context.

- "Use the following text from FILE ..."
- Sometimes backticks around file paths and repo names, sometimes not.
- Markdown files are cited differently from non-Markdown files.

There were also several unused functions here.

This improves the response quality regarding https://github.com/sourcegraph/cody/issues/3331 (for the `how does authentication work in sourcegraph/cody?` question at least).

## Test plan

CI

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -2841,515 +2841,6 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: 7d426108a93308149c0d6c9b3a4066af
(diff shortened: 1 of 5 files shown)


jtibshirani (approved):
This looks good to me, I think these subtle differences are accidental and don't have clear motivation. (Caveat: I'm far from a Cody Clients expert!)

abeatrix (approved):
Nice! From what I've tested "codebase context from..." works the best for chat. Thanks for cleaning this up!

PR #3417, opened 2024-03-15
Autocomplete: add StarCoder2 hybrid feature flag

valerybugakov:
## Context

- Preparing for the StarCoder2 A/B test.
- Adds `starcoder2-hybrid` model to the Fireworks provider, which uses `starcoder2-7b` and `starcoder2-15b`.
- Cody-Gateway support https://github.com/sourcegraph/sourcegraph/pull/61207

## Test plan

- Tested locally by mocking the return value of the feature flag provider and asserting that the right requests were made by investigating the output console.

### DIFF
--- a/vscode/src/completions/providers/fireworks.ts
+++ b/vscode/src/completions/providers/fireworks.ts
@@ -122,6 +122,7 @@ class FireworksProvider extends Provider {
     private timeouts?: AutocompleteTimeouts
     private fastPathAccessToken?: string
     private authStatus: Pick<AuthStatus, 'userCanUpgrade' | 'isDotCom' | 'endpoint'>
+    private isLocalInstance: boolean
 
     constructor(
         options: ProviderOptions,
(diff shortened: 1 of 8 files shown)


vscode/src/completions/providers/fireworks.ts
```diff
@@ -218,15 +230,27 @@ class FireworksProvider extends Provider {
...
+                      : MODEL_MAP[this.model],
+        } satisfies CodeCompletionsParams
+
+        if (requestParams.model.includes('starcoder2')) {
+            requestParams.stopSequences?.push(
+                '<fim_prefix>',
+                '<fim_suffix>',
+                '<fim_middle>',
+                '<|endoftext|>',
+                '<file_sep>'
```
philipp-spiess, line 252: Uhh new `<file_sep>` token 👀  What is this intended for?

philipp-spiess (approved):

vscode/src/completions/providers/fireworks.ts
valerybugakov: Context snippets!

philipp-spiess: Should we use it in the prompt then? 👀

valerybugakov: True. It might improve the suggestions' quality. Let me do it in a separate PR.

PR #3418, opened 2024-03-15
experimental Cody CLI

sqs:
The experimental Cody CLI spawns the agent and uses it to support CLI invocations like:

```
cody-agent experimental-cli chat -m 'what color is the sky?'

cody-agent experimental-cli chat --context-repo github.com/sourcegraph/{sourcegraph,cody} -m 'how is authentication handled in sourcegraph/cody?'
```

Unlike the previous CLI (which was removed), this CLI requires almost *no changes* to code outside the new `cli/` directory, which is a nice win for the agent architecture and means that this experimental CLI does not necessitate bad code abstractions.

Depends on https://github.com/sourcegraph/cody/pull/3671 to publish `cody-agent` binaries and packages, so that it's easy to run the CLI without building it locally from source.

## Test plan

n/a

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -1,0 +1,39 @@
+# Cody CLI (experimental)
+
+A command-line interface for Cody.
+
+**Status:** experimental
+
+## Usage
+
+Set the `SRC_ENDPOINT` and `SRC_ACCESS_TOKEN` environment variables:
+
+```
+# Sourcegraph URL (https://sourcegraph.com or a URL to an enterprise instance)
+export SRC_ENDPOINT=https://sourcegraph.com
+
+# Sourcegraph access token (created in Sourcegraph > User settings > Access tokens)
+export SRC_ACCESS_TOKEN=sgp_0000000000_0000000000000000000
+```
+
+Then run:
+
+```
+npm install -g @sourcegraph/cody-agent
+
+# Ask Cody a question (with no context):
+cody-agent experimental-cli chat -m 'what color is the sky?'
+
+# Ask Cody a question (with Sourcegraph Enterprise repository context):
+cody-agent experimental-cli chat --context-repo github.com/sourcegraph/{sourcegraph,cody} --show-context -m 'how is authentication handled in sourcegraph/cody?'
+```
+
+## Development & feedback
+
+Use the [Feedback on Cody CLI (experimental feature)](https://community.sourcegraph.com/t/feedback-on-cody-cli-experimental-feature/78) thread in the Sourcegraph community forum.
+
+Issues and PRs appreciated!
+
+## Releases
+
+The CLI is built and published as part of the [Cody Agent](../agent/README.md).
--- a/agent/package.json
+++ b/agent/package.json
@@ -12,9 +12,9 @@
   "main": "src/index.ts",
   "sideEffects": false,
   "scripts": {
-    "build:root": "cd .. && pnpm build && cd $(pwd)/agent",
+    "build:root": "pnpm -C .. run -s build",
     "build:agent": "node src/esbuild.mjs",
-    "build": "pnpm run build:root && pnpm run build:agent",
+    "build": "pnpm run -s build:root && pnpm run -s build:agent",
     "build-minify": "pnpm run build --minify",
     "agent": "pnpm run build && node dist/index.js",
     "agent:debug": "pnpm run build && CODY_AGENT_TRACE_PATH=/tmp/agent.json CODY_AGENT_DEBUG_REMOTE=true node --enable-source-maps ./dist/index.js",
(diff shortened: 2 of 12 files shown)


sqs:
I'm finding this useful for some stuff. Anyone else? Fine to keep it on a branch if nobody else wants it yet.

philipp-spiess (commented):
I like this but what are your thoughts of bundling this directly _inside_ the agent? So that the single agent binary (which already has two cli args, one for starting the `jsonrpc` server) _is_ the CLI? So we'd move the command to a `chat` command here instead: https://github.com/sourcegraph/cody/blob/main/agent/src/cli/root.ts

This would also handily give us the CLI on every non-VS-Code-client installation (in case we need to do some debugging, any additional tooling is handy!)

enriquegh:
➕ to @philipp-spiess's comment around debugging.
I could see this is as something SE uses as a troubleshooting tool as well. Where we want to remove all IDE logic from the equation and concentrate on the agent connection to the instance

sqs:
@philipp-spiess:

> I like this but what are your thoughts of bundling this directly _inside_ the agent? So that the single agent binary (which already has two cli args, one for starting the `jsonrpc` server) _is_ the CLI? So we'd move the command to a `chat` command here instead: [`main`/agent/src/cli/root.ts](https://github.com/sourcegraph/cody/blob/main/agent/src/cli/root.ts?rgh-link-date=2024-03-15T12%3A44%3A28Z)
> 
> This would also handily give us the CLI on every non-VS-Code-client installation (in case we need to do some debugging, any additional tooling is handy!)

That's a great idea. I will do that.

Do you think that the CLI should still spawn an agent subprocess (basically like `argv[0] jsonrpc`) so that it uses the agent in the same way that editor clients do? Or should the CLI just communicate with the agent within the same Node process? I'm leaning toward the former so that we don't introduce alternate code paths for in-process CLI vs. separate-process editor usage of the agent. cc @dominiccooney @steveyegge

philipp-spiess:
> Do you think that the CLI should still spawn an agent subprocess (basically like argv[0] jsonrpc since the agent and CLI are the same script now) so that it uses the agent in the same way that editor clients do? Or should the CLI just communicate with the agent within the same Node process? I'm leaning toward the former (spawn an agent subprocess) so that we don't introduce alternate code paths for in-process CLI vs. separate-process editor usage of the agent.

Yeah I agree that it should probably go through the same jsonrpc protocol than our other clients to avoid _yet-another_ way of invoking these methods.

aramaraju:
Just checking in on the status of the PR, we can get some real-world testing once it's out :) Thank you

sqs:
OK, @philipp-spiess or another @sourcegraph/cody-clients reviewer: I made this a subcommand of agent: `cody-agent experimental-cli chat -m 'what color is the sky?'`. The instructions in the README depend on https://github.com/sourcegraph/cody/pull/3671 to publish `cody-agent`, but you can run it locally now with either:

```
pnpm -C agent run build && node agent/dist/index.js experimental-cli chat -m 'what color is the sky?'

## or ##

pnpm -C agent run build-agent-binaries
agent/dist/agent-linux-x64 experimental-cli chat -m 'what color is the sky?'
```

(I am not intending to get this into the VS Code release announcement.)

PriNova:
@sqs  I'm not 100% sure about this, but to have context during chat, the `enhanced context` bool needs to be set to true.

cli/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+# Ask Cody a question (with Sourcegraph Enterprise repository context):
+cody-agent experimental-cli chat --context-repo github.com/sourcegraph/{sourcegraph,cody} --show-context -m 'how is authentication handled in sourcegraph/cody?'
+```
+
+## Development & feedback
+
+Use the [Feedback on Cody CLI (experimental feature)](https://community.sourcegraph.com/t/feedback-on-cody-cli-experimental-feature/78) thread in the Sourcegraph community forum.
+
+Issues and PRs appreciated!
```
umpox, line 35: Nice!

umpox (approved):

sqs:
> How about putting the chat in a loop for chat continuation? Until user inputs '/quit' command to gracefully exit.

That's a great idea. I just want to get this basic thing out first. I suppose you could use it in a `while true; do ...; done` loop for now until I (or someone else) adds that, though!

kilyinov:
awesome stuff, NRI already started using this. I also experimented a bit with STDOUT redirection but couldn't get it work. I tried this: 

`cody-agent experimental-cli chat -m 'create plantuml for loadPetWithVisit method in src/main/java/org/springframework/samples/petclinic/owner/VisitController.java' --context-repo github.com/kilyinov/spring-petclinic 2>&1 > VisitController.puml`

`cody-agent experimental-cli chat -m 'create plantuml for loadPetWithVisit method in src/main/java/org/springframework/samples/petclinic/owner/VisitController.java' --context-repo github.com/kilyinov/spring-petclinic > VisitController.puml`

The resulting file **VisitController.puml** was empty though. Being able to redirect output is very important for Cody scripting/automation use cases

PR #3420, opened 2024-03-15
Telemetry: Fix `hasV2Event` often being included as an event property

umpox:
## Description

- Bans `hasV2Event` from telemetry properties
- Fixes type errors created from this

## Test plan

Build passes, type only change as this is an unused telemetry property

### DIFF
--- a/lib/shared/src/telemetry/index.ts
+++ b/lib/shared/src/telemetry/index.ts
@@ -54,7 +54,12 @@ export interface TelemetryEventProperties {
         | undefined
         | string[]
         | TelemetryEventProperties[]
-        | TelemetryEventProperties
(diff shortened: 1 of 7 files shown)


umpox:
Hmm this breaks the kotlin issues :(

toolmantim (approved):

philipp-spiess (approved):
Awesome!

Hmm re. Kotlin bindings I think there are some allowlists in the code that skips certain params because they're hard to convert. Maybe you can quickly add `never` there as one but otherwise I think we just remove the type change for now

umpox:
@philipp-spiess Will remove the type change for now, had a quick look but seems quite involved and don't want to spend too long on this

PR #3429, opened 2024-03-16
show "File too large" @-mention item warning on 2nd line in menu

sqs:
Previously, this warning was shown on the same line. See the screenshots below. (Note that these are on different color themes, but you get the idea.)

### After
image

### Before
image

## Test plan

CI


toolmantim (approved):
tbh I was torn on single line vs the extra line too. After testing it, what did you prefer?

sqs:
Neither is great, but 2-line seems less broken in a “they fucked up the CSS” way.

PR #3430, opened 2024-03-16
only show at-mention menu header in empty states

sqs:
Previously, the at-mention menu header ("Search for files to include...") was always shown. It is not necessary to show it and it adds noise, so now it's only shown when (a) the query is empty (`@` or `@#`) or (b) there are no results.

## Test plan

CI


sqs:
Blocked on merging because there is jitter in the current implementation when you type `@` then `a`, upon typing `a` the header immediately goes away and the menu repositions up by 1 row. This is not a fundamental issue with the UX, just a kinda tricky positioning issue.

PR #3433, opened 2024-03-16
speed up & respect ignores in finding workspace files for file @-mentions

sqs:
1. Finding the list of files now respects the `files.exclude`,
   `search.exclude`, and gitignore/ignore files. This makes it much faster
   (~1200ms -> 100ms on sourcegraph/sourcegraph on my machine) and eliminates
   irrelevant files (such as tons of `bazel-bin/` and `*.js` files).
2. Removes the cancellation on `throttledFindFiles`. This actually made it
   slower because each time `throttledFindFiles.cancel()` was called, it would
   invalidate the cached result, which means the next call would invoke the
   expensive find-files operation even if it had been less than 10000ms since
   the previous successful call. We do not need cancellation for this
   operation because each call uses the same arguments, so earlier calls are
   not wasted and just mean the result will be returned slightly sooner.

Note: the e2e tests have 1 change because the `.gitignore` in the vscode/test/fixtures/workspace dir is now respected, so there is one less item in the dropdown. This is an enhancement.

### Before

files-before

### After

after-file

## Test plan

CI

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -17,6 +17,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 - Document: Fixed an issue where the generated documentation would be incorrectly inserted for Python. Cody will now follow PEP 257 – Docstring Conventions. [pull/3275](https://github.com/sourcegraph/cody/pull/3275)
 - Edit: Fixed incorrect decorations being shown for edits that only insert new code. [pull/3424](https://github.com/sourcegraph/cody/pull/3424)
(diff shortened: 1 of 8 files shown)


toolmantim:
`~1200ms -> 100ms` — wowsers!

philipp-spiess (approved):
Awesome stuff. I hope we can get [the better file api](https://github.com/microsoft/vscode/blob/main/src/vscode-dts/vscode.proposed.findFiles2.d.ts) to clean up some of this. 

Also this doesn't pull in nested gitignore files yet making it less useful on repos with more than one project but I don't know if it's worth it preemtively looking at _every_ ignore file in the workspace and building a giant ignore list just to support that. 😐

PR #3436, opened 2024-03-16
support @-mentioning URLs to use web page contents as context

sqs:
This is an experimental (off-by-default) feature that lets you, in chat, @-mention web pages by URL to include their text contents as context for the LLM. For example, you can type `what are the books listed at @https://slack.org/reading` into chat, select the context item in the menu, and then press Enter, and Cody will read the contents of that web page.

Requires the experimental and undocumented `"cody.experimental.urlContext": true` VS Code setting (or the `cody-url-context` feature flag, which I enabled for all sourcegraph org members on dotcom).

This is just something we can experiment with internally for now.

[Loom video (below)](https://www.loom.com/share/309d6829b98a46678e28b910af21fd21)

https://github.com/sourcegraph/cody/assets/1976/51995c63-e246-4e55-860b-03a9431caa1e

## Test plan

CI

### DIFF
--- a/lib/shared/src/chat/transcript/display-text.ts
+++ b/lib/shared/src/chat/transcript/display-text.ts
@@ -2,6 +2,7 @@ import type * as vscode from 'vscode'
 import type { URI } from 'vscode-uri'
 import type { ContextItem } from '../../codebase-context/messages'
 import type { RangeData } from '../../common/range'
+import { isURLContextItem } from '../../mentions/urlContextItems'
 
 /**
  * VS Code intentionally limits what `command:vscode.open?ARGS` can have for args (see
@@ -40,7 +41,7 @@ export function webviewOpenURIForContextItem(item: Pick<ContextItem, 'uri' | 'ra
     href: string
     target: '_blank' | undefined
 } {
-    if (item.uri.scheme === 'http' || item.uri.scheme === 'https') {
+    if (isURLContextItem(item)) {
         return {
             href: item.uri.toString(),
             target: '_blank',
--- a/lib/shared/src/experimentation/FeatureFlagProvider.ts
+++ b/lib/shared/src/experimentation/FeatureFlagProvider.ts
@@ -49,6 +49,9 @@ export enum FeatureFlag {
 
     // Show document hints above a symbol if the users' cursor is there. "Opt+D to Document"
     CodyDocumentHints = 'cody-document-hints',
+
+    /** Support @-mentioning URLs in chat to add context from web pages. */
+    URLContext = 'cody-url-context',
 }
 
 const ONE_HOUR = 60 * 60 * 1000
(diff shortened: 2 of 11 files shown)


philipp-spiess (approved):
Love the idea but I am unsure how useful this is because this won't work well if the website depends on JS to render its content and there's no way a user would notice when it's not working well right now. It also requires users to know the URL in which case they probably have it open in a browser already and it would be much more reliable for them to just copy-paste the content instead.

I think it's fine behind an experimental setting if the plan is to see how ergonomic it is but if we do want to move forward with it, we probably need a proper html client (ideally something like Playwright that can load the website and then extract the text) which adds a lot of complexity to this feature.

toolmantim:
V cool. Sadly I couldn't get the @-URL to tokenise and be used when testing this branch… I was keen to give it a try!

Did it work for you @philipp-spiess?

philipp-spiess:
@toolmantim I got it to work yeah but now I played more with it because of your comment and I get quite a few `fetch` related errors showing up. I think we attempt to load the website on every keystroke so when I type `slack.or` it will start the first request already which will fail. 😬 

Screenshot 2024-03-18 at 13 38 04

toolmantim:
Got it working! I typo'd the JSON config to enable it 🤦🏼‍♂️

Worked well with Playwright docs. 

For code docs in particular, it'd be nice to cut out the browser eventually. Being able to index the whole Playwright docs for example, and then being able to @-search the page titles to include the relevant page, vs copy and pasting from a browser.

But it's good to be able to attach an arbitrary thing into the chat as needed. There's probably a heap of use cases based on this we haven't even considered.

sqs:
@philipp-spiess:

> Love the idea but I am unsure how useful this is because this won't work well if the website depends on JS to render its content and there's no way a user would notice when it's not working well right now. It also requires users to know the URL in which case they probably have it open in a browser already and it would be much more reliable for them to just copy-paste the content instead.
> 
> I think it's fine behind an experimental setting if the plan is to see how ergonomic it is but if we do want to move forward with it, we probably need a proper html client (ideally something like Playwright that can load the website and then extract the text) which adds a lot of complexity to this feature.

Yes, agreed. This exposes a (future) gap in how we show context for things that aren't a file range in the current editor or linkable to a range on Sourcegraph. My intent is to keep it experimental to have a way to start exposing these gaps.

Also the point about how there are errors shown with each keystroke. I'll see how I can fix that even before merging this.

abeatrix (approved):
Oh wow just played around with this and think this could be pretty helpful in loading online docs!
image

nit: Should we update the heading title for `mentionQuery.type === 'url'` to `'Enter a valid URL to include...'` (instead of `'Search for a file to include...'`)?

image

image

sqs:
@abeatrix that tooltip idea is a good one. I'll add that in a followup.

jdorfman:
community member:

"bro this is a effen game changer"

Amazing work @sqs

PR #3439, opened 2024-03-17
in the transcript context list, omit needless '@' prefix

sqs:
When you click to open the `Context: 123 lines from 456 files` list, now the items no longer are prefixed with `@`. For example, it will show `my/file.ts:12-34` rather than `@my/file.ts:12-34`. The `@` is not needed when displaying the list here.

image

## Test plan

CI


sqs:
@toolmantim can you design-check this?

philipp-spiess (approved):

toolmantim:
I don’t have a Figma for it, but I think better than removing would be to use different icons to show where/why the file/lines were included. The more we can show where/why the better.

Codicon has a proper @ icon we could use for explicitly @‘d things. File lines included from a selection could use another icon. And things included from enhanced context could use the same DB icon as sources from the enhanced context popover.

With tooltips for each one, e.g. “Included from @-mentions” or “Included from the search index via Enhanced Context”.

wdyt? I’m happy for you to rough it out and I jump in to CSS it, or can Figma it if it’s not clear.

sqs:
@toolmantim That sounds great. I'm not sure that icons for /why/ make the most sense, though. We will soon have multiple types of context (already do: files, symbols, terminal) and the icon should probably be what the thing is, not why it was included. (How) should we represent both?

toolmantim:
@sqs here's two design ideas:
1. https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1826-40270&mode=design&t=iTkYZoD3bPo9Scya-4
2. https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1828-41041&mode=design&t=iTkYZoD3bPo9Scya-4

2 rethinks how we display chips now we have a rich editor (I imagine we could add some hover cards to chips too, rather than displaying them in a bar above the message)

abeatrix (approved):
Looks very clean! 
image

PR #3440, opened 2024-03-17
remove needless hovers and visibility checks from e2e

sqs:
In our e2e tests, there were many patterns like (where `x` is some Playwright locator, ie a selector query resolved by polling):

```
await x.hover()
await x.click()
```

The intent of this pattern is (except in rare cases where hovering is actually needed) to give extra time for `x` to become visible. But this is unnecessary because Playwright's `.click()` already auto-waits for the locator to become [actionable (see Playwright's definition)](https://playwright.dev/docs/actionability). The effect of these 2 lines is just to double the timeout (from 3s to 6s), which is not usually necessary. If a longer timeout is indeed needed, you can use `.click({timeout: N})`, and if a longer timeout than 3s is ever needed by default, we can update `playwright.config.ts` (but I don't think that's needed).

Same goes for this pattern:

```
await expect(x).toBeVisible()
x.click()
```

Playwright's definition of actionability (linked above) already checks for visibility.

In some cases where the intent of `await x.hover()` was to wait for something to become visible, I changed it to `await expect(x).toBeVisible()` for clarity.

There should be no behavior change (other than 3s not 6s timeouts) with these changes.

## Test plan

CI

### DIFF
--- a/vscode/test/e2e/README.md
+++ b/vscode/test/e2e/README.md
@@ -25,14 +25,3 @@ pnpm test:e2e --debug
 # Run a specific test in debug mode
 pnpm test:e2e $TEST_FILE_NAME  --debug
 ```
-
-## Notes
-
-1. Flakiness in tests can often be attributed to timing issues, such as asynchronous actions taking longer than expected to complete. By performing a hover action before a click action, you give the application a bit more time to react and stabilize before attempting the subsequent click. This can help mitigate timing-related flakiness.
-
-   Example:
-
-   ```
-   await page.getByLabel('.vscode', { exact: true }).hover()
-   await page.getByLabel('.vscode', { exact: true }).click()
-   ```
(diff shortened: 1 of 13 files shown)


philipp-spiess (approved):
So the reason this _worked_ is because `hover()` does not throw if the element is not found after the timeout?

sqs:
@philipp-spiess It does throw. It worked (aka was harmless) because it was being used as either a visibility check or a way to give a click an extra timeout.

abeatrix:
IIRC the reason we added the `hover()` check before click() was because `.toBeVisible()` throws when the element is not ready specifically on Windows, and @DanTup suggested to add the extra timeout with `hover()` which helped: [example 1](https://sourcegraph.slack.com/archives/C05AGQYD528/p1707858463955029?thread_ts=1707850515.546129&cid=C05AGQYD528), [example 2](https://sourcegraph.slack.com/archives/C05AGQYD528/p1707766896208689?thread_ts=1707744071.460429&cid=C05AGQYD528)

PR #3445, opened 2024-03-18
V1 new client extension API, for inline edits + document code

steveyegge:
This commit is part 1 of a joint effort between @dominiccooney and @steveyegge to make our client API simpler, more robust, and easier to test. The main change from the client's perspective is that the client no longer has to receive lenses as part of the protocol, focusing only on state transitions. This enabled a significant cleanup of the JetBrains-side logic.

The FixupController has been rewritten to use Strategy objects to abstract the operations and decouple the UI. There are several bug fixes in the Agent, as well as a new getFoldingRanges API for clients that need it. It's useful for figuring out the very first place to put a spinner, before the FixupController's task has finished initializing.

The commit goes hand in hand with [stevey/inline-edits](https://github.com/sourcegraph/jetbrains/tree/stevey/inline-edits) on the jetbrains repo, which implements v1 of Document Code and Edit Code.

## Test plan

We will start immediately on the integration test framework for JetBrains. This commit is just a checkpoint so that we can start parallelizing some of this work.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1076,11 +1076,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
                 `Expected a non-empty edit command result. Got ${JSON.stringify(result)}`
             )
         }
-        return {
-            id: result.task.id,
-            state: result.task.state,
-            error: errorToCodyError(result.task.error),
-        }
+        return AgentFixupControls.serialize(result.task)
     }
 
     private async createChatPanel(commandResult: Thenable<CommandResult | undefined>): Promise<string> {
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -60,6 +60,7 @@ import type {
     GetFoldingRangeResult,
     ProtocolCommand,
     ProtocolTextDocument,
+    Range,
     TextEdit,
 } from './protocol-alias'
 import { AgentHandlerTelemetryRecorderProvider } from './telemetry'
@@ -332,15 +333,22 @@ export class Agent extends MessageHandler implements ExtensionClient {
         })
 
         this.registerNotification('textDocument/didFocus', (document: ProtocolTextDocument) => {
+            function isEmpty(range: Range | undefined): boolean {
+                return !range || range === new vscode.Range(0, 0, 0, 0)
+            }
             const documentWithUri = ProtocolTextDocumentWithUri.fromDocument(document)
             // If the caller elided the content, as is the sensible thing to do, reconstruct it here.
-            if (!document.content) {
-                const cachedDocument = this.workspace.getDocumentFromUriString(
-                    documentWithUri.uri.toString()
-                )
+            const cachedDocument = this.workspace.getDocumentFromUriString(
+                documentWithUri.uri.toString()
+            )
+            if (!documentWithUri.underlying.content) {
                 if (cachedDocument?.content != null) {
-                    document.content = cachedDocument.content
+                    documentWithUri.underlying.content = cachedDocument.content
                 }
+            }
+            // Similarly, don't let this notification blow away our cached selection.
+            if (isEmpty(document.selection) && !isEmpty(cachedDocument?.protocolDocument.selection)) {
+                document.selection = cachedDocument?.protocolDocument.selection
             }
             this.workspace.setActiveTextEditor(
                 this.workspace.newTextEditor(this.workspace.addDocument(documentWithUri))
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -39,12 +39,12 @@ import type { ExtensionClient } from '../../vscode/src/extension-client'
 import { IndentationBasedFoldingRangeProvider } from '../../vscode/src/lsp/foldingRanges'
 import type { CommandResult } from '../../vscode/src/main'
 import type { FixupTask } from '../../vscode/src/non-stop/FixupTask'
+import { FixupCodeLenses } from '../../vscode/src/non-stop/codelenses/provider'
+import type { FixupActor, FixupFileCollection } from '../../vscode/src/non-stop/roles'
 import {
-    FixupCodeLenses,
     type FixupControlApplicator,
-    NullFixupControlsApplicator,
-} from '../../vscode/src/non-stop/codelenses/provider'
-import type { FixupFileCollection } from '../../vscode/src/non-stop/roles'
+    NullFixupControlApplicator,
+} from '../../vscode/src/non-stop/strategies'
 import { CodyTaskState } from '../../vscode/src/non-stop/utils'
 import { AgentWorkspaceEdit } from '../../vscode/src/testutils/AgentWorkspaceEdit'
 import { emptyEvent } from '../../vscode/src/testutils/emptyEvent'
@@ -870,11 +870,13 @@ export class Agent extends MessageHandler implements ExtensionClient {
 
     // ExtensionClient callbacks.
 
-    public createFixupControlApplicator(files: FixupFileCollection): FixupControlApplicator {
-        return this.clientInfo?.capabilities?.codeLenses &&
-            this.clientInfo?.capabilities?.fixupCodeLenses
+    public createFixupControlApplicator(
+        files: FixupActor & FixupFileCollection
+    ): FixupControlApplicator {
+        return this.clientInfo?.capabilities?.codeLenses === 'enabled' &&
+            this.clientInfo?.capabilities?.fixupControls === 'lenses'
             ? new FixupCodeLenses(files)
-            : new NullFixupControlsApplicator()
+            : new NullFixupControlApplicator()
     }
 
     private codeLensToken = new vscode.CancellationTokenSource()
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -37,6 +37,7 @@ import { ModelUsage } from '../../lib/shared/src/models/types'
 import type { CompletionItemID } from '../../vscode/src/completions/logger'
 import type { ExtensionClient } from '../../vscode/src/extension-client'
 import { IndentationBasedFoldingRangeProvider } from '../../vscode/src/lsp/foldingRanges'
+import { getDocumentSections } from '../../vscode/src/editor/utils/document-sections'
 import type { CommandResult } from '../../vscode/src/main'
 import type { FixupActor, FixupFileCollection } from '../../vscode/src/non-stop/roles'
 import type { FixupControlApplicator } from '../../vscode/src/non-stop/strategies'
@@ -56,6 +57,7 @@ import type {
     CustomCommandResult,
     EditTask,
     ExtensionConfiguration,
+    GetFoldingRangeResult,
     ProtocolCommand,
     TextEdit,
 } from './protocol-alias'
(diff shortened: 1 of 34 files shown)


agent/src/agent.ts
```diff
@@ -35,13 +35,17 @@ import type { Har } from '@pollyjs/persister'
...
 import { ModelUsage } from '../../lib/shared/src/models/types'
 import type { CompletionItemID } from '../../vscode/src/completions/logger'
+import { getDocumentSections } from '../../vscode/src/editor/utils/document-sections'
+import type { ExtensionClient } from '../../vscode/src/extension-client'
 import { IndentationBasedFoldingRangeProvider } from '../../vscode/src/lsp/foldingRanges'
-import type { CommandResult } from '../../vscode/src/main'
-import type { FixupTask } from '../../vscode/src/non-stop/FixupTask'
-import { CodyTaskState } from '../../vscode/src/non-stop/utils'
+import type { CommandResult, EditCommandResult } from '../../vscode/src/main'
+import { FixupTask } from '../../vscode/src/non-stop/FixupTask'
```
dominiccooney, line 42:
```suggestion
import type { FixupTask } from '../../vscode/src/non-stop/FixupTask'
```

agent/src/agent.ts
```diff
@@ -319,11 +333,26 @@ export class Agent extends MessageHandler {
             process.exit(0)
         })
 
-        this.registerNotification('textDocument/didFocus', document => {
+        this.registerNotification('textDocument/didFocus', (document: ProtocolTextDocument) => {
+            function isEmpty(range: Range | undefined): boolean {
+                return !range || range === new vscode.Range(0, 0, 0, 0)
```
dominiccooney, line 338: This `===` isn't right; that's strict object identity just like JavaScript so trivially false.

vscode.Range has something called isEmpty that means the range start and end are the same, but that's not the intent here, so let's call it something else. (Since this comparison was wrong, can we just delete it? It means the 0-width range at (0,0)-(0,0) is still a valid thing.)

agent/src/agent.ts
```diff
@@ -319,11 +333,26 @@ export class Agent extends MessageHandler {
...
+            }
+            const documentWithUri = ProtocolTextDocumentWithUri.fromDocument(document)
+            // If the caller elided the content, as is the sensible thing to do, reconstruct it here.
```
dominiccooney, line 341: Elide the judgement "as is the sensible thing to do" and just state it plainly, like "because the caller elides the content" if that always happens; "when the caller omits the content" if it *may* happen, etc.

agent/src/agent.ts
```diff
@@ -707,6 +736,72 @@ export class Agent extends MessageHandler {
...
+                        'Agent',
+                        'editTask/getFoldingRanges',
+                        'No document found for file path',
+                        params.uri,
+                        [...this.workspace.allUris()]
+                    )
+                    return Promise.resolve({ ranges: [] })
+                }
+                const ranges = await getDocumentSections(document)
+                return { ranges: ranges }
```
dominiccooney, line 773:
```suggestion
                return { ranges }
```

agent/src/agent.ts
```diff
@@ -707,6 +736,72 @@ export class Agent extends MessageHandler {
...
+            }
+        )
+
+        this.registerAuthenticatedRequest('editCommands/code', params => {
+            const args = { configuration: { ...params } }
+            const taskThenable = vscode.commands.executeCommand<FixupTask | undefined>(
+                'cody.command.edit-code',
+                args
+            )
+            // Wrap the task in a Thenable that returns a CommandResult, required by createEditTask().
```
dominiccooney, line 783: In my experience, no good comes of providing Promise-like APIs when Promises are available, yet here we are...

This should be more succinct by not reiterating the types, and relying on failures to propagate through without mentioning them explicitly. Something like the following but with proper indentation.

```
return this.createEditTask(
    vscode.commands.executeCommand<FixupTask | undefined>('cody.command.edit-code', args)
    .then(task => task && { type: 'edit', task })
)
```

vscode/src/commands/execute/edit.ts
```diff
@@ -0,0 +1,35 @@
...
+import { type ExecuteEditArguments, executeEdit } from '../../edit/execute'
+import { getEditor } from '../../editor/active-editor'
+import type { EditCommandResult } from '../../main'
+
+import { wrapInActiveSpan } from '@sourcegraph/cody-shared/src/tracing'
+
+export async function executeEditCommand(
+    args: ExecuteEditArguments
+): Promise<EditCommandResult | undefined> {
+    return wrapInActiveSpan('command.test', async span => {
```
dominiccooney, line 11: That span identifier looks like copypasta from the test command.

vscode/src/commands/execute/edit.ts
```diff
@@ -0,0 +1,35 @@
...
+        span.setAttribute('sampled', true)
+        const instruction = args.configuration?.instruction // get the instruction
+        const editor = getEditor()?.active // get the active editor
```
dominiccooney, line 14: Just chain them, `const document = getEditor()?.active.document`

(Sorry for not using GitHub suggestions but something is wrong with multiline select in the web GitHub UI rn.)

vscode/src/commands/execute/edit.ts
```diff
@@ -0,0 +1,35 @@
...
+        return {
+            type: 'edit',
+            task: await executeEdit({
+                configuration: {
+                    ...args.configuration,
+                    instruction,
+                    document,
+                    mode: 'edit',
+                },
+                source: DefaultEditCommands.Test,
```
dominiccooney, line 30: This source also looks iffy for "edit" (or the source is right and the method name is iffy.)

dominiccooney (approved):
Some minor feedback inline, PTAL. LGTM.

There's a format for bylines in commit descriptions. Also need to add @abeatrix here for the edit command:

```
Co-authored-by: Dominic Cooney <dominic.cooney@sourcegraph.com>
Co-authored-by: Beatrix <beatrix@sourcegraph.com>
```

abeatrix (approved):
From my testing I can confirm that all edit commands are working correctly!

Doc command:
image

Edit command:
image


Test command:
image

agent/src/agent.ts
steveyegge: Fixed!

agent/src/agent.ts
steveyegge: Fair enough; elided. I do actually want to have it be _recommended_ to client writers, as part of a broader effort to document the expected semantics of all the request and response objects. Not sure exactly where yet.

agent/src/agent.ts
steveyegge: Works like a charm.

vscode/src/commands/execute/edit.ts
steveyegge: I have no recollection of any of this code, and it is indeed bizarre. Perhaps I was having a minor seizure. I've made all the suggested changes and removed the spurious comments.

vscode/src/commands/execute/edit.ts
steveyegge: Fixed all this code, which was definitely dubious.

vscode/src/extension.common.ts
```diff
@@ -44,14 +45,13 @@ export interface PlatformContext {
...
     startTokenReceiver?: typeof startTokenReceiver
     onConfigurationChange?: (configuration: Configuration) => void
+    extensionClient: ExtensionClient
 }
 
 export async function activate(
     context: vscode.ExtensionContext,
     platformContext: PlatformContext
 ): Promise<ExtensionApi> {
-    const api = new ExtensionApi(context.extensionMode)
```
abeatrix, line 53: reverting this change fixed the integration test

vscode/src/extension.common.ts
```diff
@@ -67,5 +67,5 @@ export async function activate(
         console.error(error)
     }
 
-    return api
+    return new ExtensionApi(context.extensionMode)
```
abeatrix, line 70:
```suggestion
    return api
```
reverting this change fixed the integration test

vscode/src/extension.common.ts
abeatrix, line 53: reverting this change fixed the integration test

vscode/src/non-stop/FixupController.ts
```diff
@@ -16,159 +16,37 @@ import { countCode } from '../services/utils/code-count'
...
-                    return
-                }
-                return vscode.commands.executeCommand('cody.fixup.codelens.undo', nearestTask.id)
-            })
-        )
+    constructor(
+        private readonly authProvider: AuthProvider,
+        client: ExtensionClient
+    ) {
+        this.controlApplicator = client.createFixupControlApplicator(this)
```
umpox, line 49: I like this abstraction 👍

vscode/src/non-stop/codelenses/provider.ts
```diff
@@ -1,25 +1,202 @@
...
+                telemetryService.log(
+                    'CodyVSCodeExtension:fixup:codeLens:clicked',
+                    {
+                        op: 'cancel',
+                    },
+                    {
+                        hasV2Event: true,
+                    }
+                )
+                telemetryRecorder.recordEvent('cody.fixup.codeLens', 'cancel')
```
umpox, line 42: Assuming that we are expecting that each client implements their own telemetry (I think this is fine)

vscode/src/non-stop/roles.ts
```diff
@@ -1,17 +1,65 @@
...
+    cancel(task: FixupTask): void
+
+    /**
+     * Undo the task (see `undo`), prompt for updated instructions, and start
+     * a new task to try again. Only applicable to tasks in the "applied" state.
+     * @param task the task to retry.
+     * @param source the source of the retry, for event logging.
+     */
+    retry(task: FixupTask, source: ChatEventSource): Promise<FixupTask | undefined>
+}
```
umpox, line 42: Awesome

vscode/src/non-stop/strategies.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+// A FixupControlApplicator which does not present any controls for fixup
+// tasks.
+export class NullFixupControlApplicator implements FixupControlApplicator {
+    public didUpdateTask(task: FixupTask): void {}
+    public didDeleteTask(task: FixupTask): void {}
+    public visibleFilesWithTasksMaybeChanged(files: readonly FixupFile[]): void {}
+    public dispose(): void {}
+}
```
umpox, line 28: Just a general comment, ideally we should use `/** doc */` for doc strings, as they are attached for hovers

vscode/src/non-stop/strategies.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface FixupControlApplicator extends vscode.Disposable {
+    didUpdateTask(task: FixupTask): void
+    didDeleteTask(task: FixupTask): void
+    // Called when visible files changed.
+    // TODO: This API design is gross: this is *not* called when a new task
+    // is created in a file that is already visible. It *is* called every time
+    // visible files change, so be prepared to handle repeated calls with
+    // an empty or unchanged set of files efficiently. Unearth a consistent
+    // API here.
+    visibleFilesWithTasksMaybeChanged(files: readonly FixupFile[]): void
```
umpox, line 18: We do need to listen to when the visible files change, regardless of if they are empty or not. This is currently used purely to enable/disable keyboard shortcuts when there is a visible edit task (in one of the visible open files). Just listening for a task being created is not sufficient here

umpox (commented):
Finally got chance to review this, great work @steveyegge and @dominiccooney. I think these are sensible abstractions 👍

PR #3446, opened 2024-03-18
add e2e-flakiness-detector GitHub action

sqs:
Every day at 1400 UTC, this action runs the e2e tests many times (10 times per test case) to detect intermittent flakiness proactively instead of having it be occasionally detected in unrelated PRs. This information will help us fix flaky e2e tests.

## Test plan

CI


dominiccooney (approved):
Would be great to add docs about when, where and how to review these.

burmudar (approved):

sqs:
@dominiccooney Yep, I'll add docs once we've been able to try this out (and see how/if it helps).

PR #3447, opened 2024-03-18
only allow 1 retry for flaky e2e tests

sqs:
This will mean more annoying e2e failures due to flakiness in our PRs, which is bad. But it, along with https://github.com/sourcegraph/cody/pull/3446, will mean that we (if we are good people :) will notice the flakiness better and will fix it, thereby making it so we don't need to wait for multiple retries of some flaky tests on each build, thereby saving us time overall.

We should make this `retries: 0` soon once we reduce the flakes (which #3446 will help us identify).

## Test plan

CI

### DIFF
--- a/vscode/playwright.config.ts
+++ b/vscode/playwright.config.ts
@@ -4,8 +4,7 @@ const isWin = process.platform.startsWith('win')
 
 export default defineConfig({
     workers: 1,
-    // Give failing tests more chances
-    retries: isWin ? 4 : 2,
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):

burmudar (approved):

sqs:
I think this is worth merging because fixing flakiness in test-e2e/ubuntu in specific test cases is actually fairly easy and is something we should be doing more regularly. And test-e2e/windows already is super flaky, so it's already not trusted.

PR #3450, opened 2024-03-18
fix log spam in e2e flakiness detector

sqs:
The GitHub action appears to fail because of tons of log lines from the thing that logs our telemetry from e2e runs. Example: https://github.com/sourcegraph/cody/actions/runs/8321504214/job/22767854935

## Test plan

CI


sqs:
(noticed a flake in `typing @-mention text does not automatically accept it`, will look into it)

jhchabran (approved):

PR #3460, opened 2024-03-19
misc e2e test improvements

sqs:
- Remove `page.waitForTimeout` calls when they are unnecessary (before another `expect` call that waits for 3000ms). Keep `page.waitForTimeout` calls that perform an action immediately after because those delays are needed.
- Improve the @-mention test to check that the correct results are in the mention list for easier debuggability, and slow down keypresses to avoid flakiness that seems only to occur if multiple keyboard events are sent without intervening delay.
- Increase timeout in the Document Command e2e test because formatting sometimes is slow.
- Increase timeout for the e2e flakiness detector because it was timing out (exceeding 15 min).

## Test plan

CI

### DIFF
--- a/.github/workflows/e2e-flakiness-detector.yml
+++ b/.github/workflows/e2e-flakiness-detector.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         runner: [ubuntu, windows]
     runs-on: ${{ matrix.runner }}-latest
-    timeout-minutes: 15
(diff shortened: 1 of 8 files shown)


valerybugakov (approved):

sqs:
For some reason the `existing installs should show the update toast when the last dismissed version is different` is now flaky on CI on this PR, but I don't see anything in this PR that would have broken it.

PR #3466, opened 2024-03-19
Fix chat abort before first streaming chunk is coming in

philipp-spiess:
Review this with whitespace changes removed. I need to validate this still works for rate limiting issues.

## Test plan

- Send a message and abort before the message starts streaming the first chars


philipp-spiess:
Screenshot 2024-03-19 at 15 20 31
rate limit errors still work

abeatrix (approved):
Thanks! Would it be possible to add a e2e test for the abort button?

philipp-spiess:
@abeatrix Yeah but the problem here is with the special code paths that we go through for claude 3 right now that will be fixed once we land enterprise support. I will look into e2e testing this flow after we have converged the flows if that's alright?

PR #3467, opened 2024-03-19
Test: Fix cody-ignore e2e test on Windows

abeatrix:
Attempt to fix the cody-ignore e2e test on Windows that was failing on main. Based on the last commit on main, this is the test that's failing: https://github.com/sourcegraph/cody/actions/runs/8344781538/job/22838057502#step:11:325

image

## Test plan

CI

image

## Next

Fix flaky test: chat-atFile

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -598,6 +598,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         }
 
         const cancellation = new vscode.CancellationTokenSource()
+        this.contextFilesQueryCancellation = cancellation
         getChatContextItemsForMention(query, cancellation.token, scopedTelemetryRecorder)
             .then(items => {
                 if (!cancellation.token.isCancellationRequested) {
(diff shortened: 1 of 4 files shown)


philipp-spiess:
Screenshot 2024-03-19 at 19 54 57

not sure if this is ready to review yet 😆

abeatrix:
> Screenshot 2024-03-19 at 19 54 57
> not sure if this is ready to review yet 😆

😆 it is now

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -577,11 +576,6 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
     }
 
     private async handleGetUserContextFilesCandidates(query: string): Promise<void> {
-        // Cancel previously in-flight query.
```
abeatrix, line 580: I could be wrong, but I think this is not needed and can be removed. I also noticed this was causing issues for the Windows e2e test because the cancellation is sent to the [getFileContextFiles()](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/editor/utils/editor-context.ts?L57) function, which we use the token to [throttledFindFiles()](https://github.com/sourcegraph/cody/pull/3467/files) in order to limit the call to 1 per every 10 seconds, so typing really quickly (as e2e tests do) could keep causing the previous token to be canceled, and the last input falls into that 10-second person so sometimes we will not get results back.

abeatrix:
continue in https://github.com/sourcegraph/cody/pull/3489

PR #3477, opened 2024-03-20
Autocomplete: use `parser.safeParse()`

valerybugakov:
## Context

- Wrap `parser.parse()` in a try-catch statement, report errors to Sentry and fail gracefully.

## Test plan

CI

### DIFF
--- a/vscode/src/tree-sitter/parser.ts
+++ b/vscode/src/tree-sitter/parser.ts
@@ -5,6 +5,7 @@ import type Parser from 'web-tree-sitter'
 
 import { wrapInActiveSpan } from '@sourcegraph/cody-shared'
 import type { Tree } from 'web-tree-sitter'
+import { captureException } from '../services/sentry/sentry'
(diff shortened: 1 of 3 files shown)


vscode/src/tree-sitter/parser.ts
```diff
@@ -82,10 +90,25 @@ export async function createParser(settings: ParserSettings): Promise<WrappedPar
...
 
+    const safeParse: SafeParse = (...args) => {
+        try {
+            return parser.parse(...args)
+        } catch (error) {
+            captureException(error)
+
+            if (process.env.NODE_ENV === 'production') {
+                console.error('parser.parse() error:', error)
+            }
```
philipp-spiess, line 101: Should that be `!==`?

philipp-spiess (approved):
LGTM

vscode/src/tree-sitter/parser.ts
valerybugakov: yep! fixing

PR #3478, opened 2024-03-20
Chat: clarify tooltip for remote search context

abeatrix:
Small UI change.

Update the tooltip for file links included via remote search to explicitly mention "search (remote)" for unified context to improve clarity and make it easier to identify the search, as currently, we are showing it as `via Search` which overlaps with the search results from `symf`. 

|  symf context | unified context   |
|---|---|
| image  |  image  |

### Before

image

### After

image


dominiccooney (approved):
👍

abeatrix:
> 👍 
> 
> 
> 
> (remote) seems redundant given the only enterprise search is remote search, but seems fine.

That's true! Maybe it'd be helpful in the future if we decide to support both 😆 

For now I think this would still be helpful for supports and users to identity if the context sources are correct imo!

PR #3479, opened 2024-03-20
Error when receing an SSE response that does not contain any SSE event

philipp-spiess:
We have observed cases where the response of the SSE endpoint did not contain any SSE events (and was thus malformed). This PR makes it so that in this case we still yield an error and log the actual body in verbose mode.

## Test plan

- Apply a change to the sg server like this

```diff
diff --git a/internal/completions/httpapi/handler.go b/internal/completions/httpapi/handler.go
index 50fb45bcd77..afc1080a63a 100644
--- a/internal/completions/httpapi/handler.go
+++ b/internal/completions/httpapi/handler.go
@@ -204,6 +204,8 @@ func newSwitchingResponseHandler(logger log.Logger, db database.DB, feature type
        nonStreamer := newNonStreamingResponseHandler(logger, db, feature)
        streamer := newStreamingResponseHandler(logger, db, feature)
        return func(ctx context.Context, requestParams types.CompletionRequestParameters, cc types.CompletionsClient, w http.ResponseWriter, userStore database.UserStore, test guardrails.AttributionTest) {
+               w.Write([]byte("oh this is malformed huh? too bad"))
+               return
                if requestParams.IsStream(feature) {
                        streamer(ctx, requestParams, cc, w, userStore, test)
                } else {
```

- Observe that we see an error now instead of a hanging request
Screenshot 2024-03-20 at 16 04 11

### DIFF
--- a/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
+++ b/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
@@ -53,6 +53,7 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
             // Keep track if we have send any message to the completion callbacks
             let didSendMessage = false
             let didSendError = false
+            let didReceiveAnyEvent = false
 
             // Call the error callback only once per request.
             const onErrorOnce = (error: Error, statusCode?: number | undefined): void => {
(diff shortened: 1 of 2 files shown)


sqs (approved):

umpox (approved):

slimsag (approved):
🤦 🙏

PR #3480, opened 2024-03-20
Add support for storing @-mentioned URL content in .fetched folder

Kynlos:
This pull request adds the capability to store the text content of @-mentioned URLs in chat as files in a .fetched subfolder of the current project's root directory.

Key changes:

• When a user @-mentions a URL in the chat input, the web page content is fetched and processed as before. 
• If the URL context feature flag is enabled (cody.experimental.urlContext), the fetched text content is now saved as a file in the .fetched folder. 
• The .fetched folder is created as a subfolder of the currently open project's root directory (the first workspace folder). • The filename for each fetched URL is generated using the current timestamp and the URL (with non-word characters replaced by hyphens), ensuring uniqueness. 
• The fetched text content is written to the file using fs.writeFile. 
• This feature allows users to easily access and review the content of @-mentioned URLs later, as the files persist in the project directory. It can be helpful for referencing the information pulled in from web pages during the chat conversation. It can also be used later as an @filename reference (which could help with monitoring changes between stored information and any updated information)

Note: This feature creates a .fetched folder in the project's root directory and writes files to it. The files are not automatically cleaned up, so users may want to manually manage this folder and remove old or unnecessary files periodically.

### DIFF
--- a/vscode/src/editor/utils/editor-context.ts
+++ b/vscode/src/editor/utils/editor-context.ts
@@ -1,5 +1,5 @@
 import path from 'path'
-
+import fs from 'fs/promises'    
 import fuzzysort from 'fuzzysort'
 import throttle from 'lodash/throttle'
 import * as vscode from 'vscode'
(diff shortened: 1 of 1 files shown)


dominiccooney:
I like the idea... This is basically a cache of fetched web content. I bet there's a web cache implementation for npm we could roll in here and get some header understanding, size limits, expiries, etc.

Kynlos:
Oh absolutely. Mine was just an extremely basic implementation. 

I was thinking that maybe it could even create an embedding that the user could call into the chat window whenever they need it (not something I have remotely enough knowledge about to implement).

dominiccooney:
> I was thinking that maybe it could even create an embedding that the user could call into the chat window whenever they need it (not something I have remotely enough knowledge about to implement).

That would be cool. cody-engine that we use for embeddings can't do this as-is, it is hard wired to embed the content of git repos. But with some work it could be integrated with a web content cache.

PR #3481, opened 2024-03-20
Properly handle one or more ndjson records coming from Ollama

peebles:
## Test plan

The bug was verified using an isolated version of lib/shared/src/ollama/chat-client.ts running against a fast, local ollama server.  The changes were made and verified against same ollama server.

The issue is triggered very consistently by setting "cody.experimental.ollamaChat": true, and using a fast local Ollama setup, perhaps using llama2:7b or as small and fast a model as you can find.  Then ask a question with a reasonably long answer like "why is the sky blue?".  You should see the answer halt somewhere in the middle and Cody chat is then "frozen" and cannot be aborted.  The only recovery is to delete the chat and start a new one.  The reason this happens is that if Cody does not read the stream fast enough, it gets '{...}\n{...}' in reader.read() and then JSON.parse bombs.  The existing code does not accommodate x-ndjson, which is what Ollama is sending.    Probably all other LLM providers too.

can-ndjson-stream is a library that could be used, but I thought a no-dependency solution would be more acceptable as a PR.  You might read this background article: https://andreitopli.medium.com/working-with-newline-delimited-json-using-can-ndjson-stream-in-typescript-983016d51d4a

Added testcase: lib/shared/src/ollama/chat.test.ts

### DIFF
--- a/lib/shared/src/ollama/chat-client.ts
+++ b/lib/shared/src/ollama/chat-client.ts
@@ -73,7 +73,7 @@ export function ollamaChatClient(
             const textDecoder = new TextDecoder()
             const decoded = textDecoder.decode(value, { stream: true })
             // handle one or more ndjson records
-            const parsedLines = decoded.split('\n').map(JSON.parse) as OllamaGenerateResponse[]
+            const parsedLines = decoded.split('\n').map(m => JSON.parse(m)) as OllamaGenerateResponse[]
             for (const parsedLine of parsedLines) {
                 if (parsedLine.message) {
                     insertText += parsedLine.message.content
--- a/lib/shared/src/ollama/chat-client.ts
+++ b/lib/shared/src/ollama/chat-client.ts
@@ -72,11 +72,13 @@ export function ollamaChatClient(
 
             const textDecoder = new TextDecoder()
             const decoded = textDecoder.decode(value, { stream: true })
-            const parsedLine = JSON.parse(decoded) as OllamaGenerateResponse
-
-            if (parsedLine.message) {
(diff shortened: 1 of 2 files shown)


lib/shared/src/ollama/chat-client.ts
```diff
@@ -72,11 +72,19 @@ export function ollamaChatClient(
...
+            if (matches !== null) {
+                parsedLines = matches.map(line => {
+                    return JSON.parse(line) as OllamaGenerateResponse
+                });
+            }
+            for (const parsedLine of parsedLines) {
+                if (parsedLine.message) {
+                    insertText += parsedLine.message.content
+                    cb.onChange(insertText)
+                }
```
dominiccooney, line 87: Cool but we can do this more succinctly, something like:

```suggestion
            // handle one or more ndjson records
            const parsedLines = decoded.split('\n').map(JSON.parse) as OllamaGenerateResponse[]
            for (const parsedLine of parsedLines) {
                if (parsedLine.message) {
                    insertText += parsedLine.message.content
                    cb.onChange(insertText)
                }
```

dominiccooney (changes requested):
Looks good, one suggestion inline. Triggered CI for you.

It would be great if you could add some detail to your test plan to break it down step-by-step for someone to reproduce, or add a unit test.

peebles:
I went back to my original way of breaking records.  Splitting on \n did not work in all cases, while /.+/ does work on all cases.
I also added a unit test to try all the important cases.

peebles:
Is there anything I need to do?

Issue #3483, opened 2024-03-20
bug: markdown is not rendered in transcript user message


kalanchan:
### Version

main

### Describe the bug

Users can't encapsulate markdown with triple backticks (```) anymore in chat

see video 

https://github.com/sourcegraph/cody/assets/51868853/2f7c3eb4-1909-43a2-84c2-e917533a739a

 

### Expected behavior

in chat users should be able to encapsulate their code in markdown using triple backticks

sqs:
The bug is that the markdown is not rendered when the user's message is displayed in the transcript.

Will look into it now

kalanchan:
also filed in #3496

kalanchan:
@sqs any updates to the fix?

kalanchan:
Based on new design direction to make user transcript less rich, more notebook style

PR #3484, opened 2024-03-20
do not strip intentional newlines in chat transcript items

sqs:
Partial fix for https://github.com/sourcegraph/cody/issues/3483. Makes it so that the Markdown is rendered literally with newlines as typed, instead of all being jammed on 1 line. (It does not render the user's code in backticks as code.)

Here is how it looks (not ideal because of the lack of rendering, but better than seen in #3483 since the newlines are preserved):

image

## Test plan

Type a message with triple-backtick then newline then multiple lines of code. Confirm that the newlines are preserved when the message is displayed in the transcript.

### DIFF
--- a/lib/shared/src/chat/markdown.ts
+++ b/lib/shared/src/chat/markdown.ts
@@ -47,6 +47,7 @@ const DOMPURIFY_CONFIG = {
         'h4',
         'h5',
         'h6',
+        'br',
     ],
     ALLOWED_URI_REGEXP,
     FORBID_ATTR: [


abeatrix (approved):
🤯

sqs:
Compare this to the current behavior on the stable release (the code is rendered):
image

kalanchan (approved):
what's the appropriate next step to fixing this?

sqs:
@kalanchan It's to make it render nicely as well. That is what I'll commit to doing in the next 48h.

PR #3485, opened 2024-03-20
remove @-mention URL e2e test (experimental feature)

sqs:
## Test plan

CI


abeatrix (approved):

dominiccooney:
@sqs It would be great if the PR had a brief justification of why we're removing this test. I feel weird asking #3480 to add tests if we've just been taking the tests out. We can also disable tests and the blame gives some clues about why the test is disabled...

Issue #3455, opened 2024-03-18
tracking: context limit issues with @-mentioned files


abeatrix:
This is an issue spinned off from https://github.com/sourcegraph/cody/issues/2965 where the original goal was to display warnings for the @-mentioning massive files

Context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1710331429302729

## Current state

Here is the current state of @-mentioning massive files summarized by @chillatom:

1. The file size "check" that we use to display the warning is not the real token limit applied by the system.  It is an approximation.  This leads to the odd behavior where we trigger a warning, but you can still include the file or you can successfully copy and paste the file contents into the chat message. 
2. The warning check is computed per file, which means that if file A & B are both individually under the estimated limit, we trigger no warning, even if A+B is above the limit.
3. If the file is actually too large to be included, but the user submits anyway (e.g. vscode/src/completions/logger.ts which is 860 LOC and ~6,700 tokens. It appears that we silently exclude the file from context and Cody hallucinates. :red_circle:
4. If the file triggers the warning, but isn't too large to be included e.g. vscode/src/local-context/symf.ts which is 641 LOC and 4,867 tokens, I see the content actually used as context and referenced correctly.

At a minimum it feels like we should do a few things

1. Tie our warning to the actual operation of the product.  If we say it's too large, the file should not be able to be input
2. We should not silently exclude a file that has been explicitly @ mentioned
3. Consider the case of multiple @ mentioned files
4. Prioritize @ mention files over other fetched context (not sure if we do this today) as to avoid silently removing the file explicitly referenced

Over a longer term, I think we should explore
- Expanding the context windows, especially for some of the flagship models (I'll be writing a proposal here)
- Consider summarization or [proposition extraction](https://arxiv.org/pdf/2312.06648.pdf) from the file

## Design Tasks

```[tasklist]
### Design Tasks:
- [x]  We should not silently exclude a file that has been explicitly @ mentioned - [figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1590-34814&mode=design&t=ZSpfaRg2b9wfCOfP-4)
```

## Engineering Tasks

```[tasklist]
### Engineering Tasks:
- [x] Tie our warning to the actual operation of the product. If we say it's too large, the file should not be able to be input
- [ ] https://github.com/sourcegraph/cody/issues/3522
[...]

PR #3486, opened 2024-03-20
Chat: sync token limit at model import time

abeatrix:
Building block for https://github.com/sourcegraph/cody/issues/3455

Previously, we had to check the `authStatus` and `config` every time we wanted to get the token limit for a model. This approach was inefficient as it required the caller to have access to authProvider and check the same config repeatedly.

With this change, we sync the token limit along with the model when we import models By including the `contextWindow` property in the model definitions, we can easily access the token limit for each model anywhere without the need for repeating the same checks we did when we import the model.

This approach eliminates the need for scattered token limit checks throughout the codebase, making the code more maintainable and accessible. It centralizes the token limit configuration within the model definitions, making it easier to understand and modify.

## Test plan

Check token limit


agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/ModelProvider.kt
```diff
@@ -6,6 +6,7 @@ data class ModelProvider(
   val codyProOnly: Boolean? = null,
   val provider: String? = null,
   val title: String? = null,
+  val contextWindow: Int? = null,
```
dominiccooney, line 9: Put the (approximate) units in this name. You intend this as a character count or byte count, not a token count, but if people mix them up the type system won't save you!

Let's name it so reviewers can at least spot problems.

```suggestion
  val contextWindowSizeChars: Int? = null,
```

lib/shared/src/models/dotcom.ts
```diff
@@ -1,6 +1,9 @@
 import type { ModelProvider } from '.'
 import { ModelUsage } from './types'
 
+export const DEFAULT_CHAT_MODEL_TOKEN_LIMIT = 28000 // 7000 tokens * 4 bytes per token
```
dominiccooney, line 4: Same feedback here, this naming is confusing. You say TOKEN but it is really a BYTE limit.

lib/shared/src/models/index.ts
```diff
@@ -15,16 +19,18 @@ export class ModelProvider {
...
         public readonly usage: ModelUsage[],
-        isDefaultModel = true
+        tokenLimit?: number
     ) {
         const splittedModel = model.split('/')
         this.provider = getProviderName(splittedModel[0])
         this.title = splittedModel[1]?.replaceAll('-', ' ')
-        this.default = isDefaultModel
+        this.default = true
+        this.contextWindow = tokenLimit ? tokenLimit * 4 : DEFAULT_FAST_MODEL_TOKEN_LIMIT
```
dominiccooney, line 33: It would be great to put `tokenLimit * 4` inside a function, say `tokenCountToByteCount(tokenLimit)`, so that when we make this accurate we don't have to hunt the codebase for the digit "4" and can instead find uses of that function.

lib/shared/src/models/index.ts
```diff
@@ -120,4 +126,12 @@ export class ModelProvider {
             }
         })
     }
+
+    public static getContextWindow(modelID: string): number {
```
dominiccooney, line 130: Same naming feedback: Specify that this is a byte count in the name, don't leave people to guess.

dominiccooney (changes requested):
Thanks for working on this! Some suggestions for naming to help people catch mixing up byte/character and token counts.

lib/shared/src/models/index.ts
```diff
@@ -82,11 +100,8 @@ export class ModelProvider {
      * made available for use.
      */
     public static add(provider: ModelProvider): void {
-        // private instances can only support 1 provider atm
-        if (ModelProvider.privateProviders.size) {
-            ModelProvider.privateProviders.clear()
-        }
-        ModelProvider.privateProviders.set(provider.model.trim(), provider)
+        // NOTE: private instances can only support 1 provider atm
```
dominiccooney, line 103: This method, "add", is mis-named. Let's call it `overrideModelList` or something like that, that is what it is really doing. Having a list of 3 models, calling "add", and getting a list of one model out is surprising.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -58,7 +58,7 @@ import type { Repo } from '../../context/repo-fetcher'
 import type { RemoteRepoPicker } from '../../context/repo-picker'
 import type { ContextRankingController } from '../../local-context/context-ranking'
 import { chatModel } from '../../models'
-import { getContextWindowForModel } from '../../models/utilts'
+import { setModelProviders } from '../../models/utilts'
```
dominiccooney, line 61: Let's rename that file "utils" for consistency with other similar files elsewhere.

dominiccooney (changes requested):
Thanks for making it clear about tokens vs chars.

We can simplify this more. See how `ModelProvider.get` has a side effect, overwriting the primary models if the user is a dotcom user? It is surprising to have "get" change state. See *how many* auth status listeners push auth status to the model provider? That's too much. That can be simplified:

- Make a class for holding all the models.
- Move the statics from ModelProvider, so they become *instance* properties of the new class.
- Create an instance of the class when the extension starts.
- Subscribe this new instance to auth status changes.
- Do the dotcom handling along with the rest of the auth status change handling.
- You don't need to call setModelProvider from all the other places auth status changes.

Ideally, pass the instance of the new class to every place that needs models. But if that is too hard in one patch, you can make one static method to hang onto the single instance of it and we will clean it up later.

lib/shared/src/models/index.ts
abeatrix: True! I've renamed this to `setProviders` instead.

lib/shared/src/models/index.ts
```diff
@@ -13,111 +17,106 @@ import { getProviderName } from './utils'
...
-        currentModel?: string
-    ): ModelProvider[] {
-        const isDotComUser = !endpoint || (endpoint && isDotCom(endpoint))
-        const models = (
-            isDotComUser
-                ? ModelProvider.dotComProviders
-                : Array.from(ModelProvider.privateProviders.values())
-        )
+    public static getProviders(type: ModelUsage, currentModel?: string): ModelProvider[] {
+        const models = ModelProvider.primaryProviders
```
dominiccooney, line 66: This primary providers/ollama providers concatenation is duplicated on line 85.

Let's de-dup by adding a getter which does the concatenation and call it in both places.

lib/shared/src/models/index.ts
```diff
@@ -13,111 +17,106 @@ import { getProviderName } from './utils'
...
+
+/**
+ * Fetches available Ollama models from the local Ollama server
+ * and adds them to the list of ollama providers.
+ */
+export async function fetchLocalOllamaModels(): Promise<ModelProvider[]> {
+    if (process.env.CODY_SHIM_TESTING === 'true') {
+        return []
+    }
+    // TODO (bee) watch file change to determine if a new model is added
```
dominiccooney, line 100: There's no file to watch?

lib/shared/src/models/index.ts
```diff
@@ -13,111 +17,106 @@ import { getProviderName } from './utils'
...
+        return tokensToChars(model?.maxToken || DEFAULT_FAST_MODEL_TOKEN_LIMIT)
+    }
+}
+
+/**
+ * Fetches available Ollama models from the local Ollama server
+ * and adds them to the list of ollama providers.
+ */
+export async function fetchLocalOllamaModels(): Promise<ModelProvider[]> {
+    if (process.env.CODY_SHIM_TESTING === 'true') {
```
dominiccooney, line 97: Why not? Wouldn't the network replay be able to capture and replay the connection errors, etc.?

vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -100,15 +99,8 @@ export class ChatManager implements vscode.Disposable {
...
-        if (authStatus?.configOverwrites?.chatModel) {
-            ModelProvider.add(
-                new ModelProvider(authStatus.configOverwrites.chatModel, [
-                    ModelUsage.Chat,
-                    // TODO: Add configOverwrites.editModel for separate edit support
-                    ModelUsage.Edit,
-                ])
-            )
-        }
+        setModelProviders(authStatus)
```
dominiccooney, line 102: I guess one place is better than N places... Because the models are static, I think we should wire up setting the model providers to auth status at the top level. For example, say we make ChatManager lazier but still query models for edits, then we have a bug.

Having less global state, and giving the model provider object responsibility for keeping itself up-to-date instead of free-riding on nearby objects, will make it easier to keep evolving this code.

dominiccooney (approved):
Very nice. Some more feedback inline I leave up to your judgement.

lib/shared/src/models/index.ts
abeatrix: I was thinking we can watch for changes in `~/. ollama/models/manifests/registry.ollama.ai/library` where the ollama models are stored.

lib/shared/src/models/index.ts
abeatrix: Removed!

PR #3487, opened 2024-03-20
Add chatExport endpoint & tests

mkondratek:
Rel PR to Jetbrains: https://github.com/sourcegraph/jetbrains/pull/1096

## Test plan
- tests
- run the export action from the jetbrains IDE


agent/src/agent.ts
```diff
@@ -775,6 +776,15 @@ export class Agent extends MessageHandler {
             return { models: panel.models ?? [] }
         })
 
+        this.registerAuthenticatedRequest('exportChats', async ({ id }): Promise<ChatHistory> => {
+            const panel = this.webPanels.getPanelOrError(id)
+            await this.receiveWebviewMessage(id, {
+                command: 'history',
+                action: 'export',
+            })
+            return panel.chatHistory || {} // todo: proper handling
```
mkondratek, line 785: - [ ] TODO

agent/src/index.test.ts
```diff
@@ -337,6 +338,45 @@ describe('Agent', () => {
...
+                        command: 'submit',
+                        text: 'What model are you?',
+                        submitType: 'user',
+                        addEnhancedContext: false,
+                    },
+                })
+            )
+
+            // Step 2: Restore multiple chats
+            const NUMBER_OF_CHATS_TO_RESTORE = 3
```
mkondratek, line 359: 🟢 I tried with 300 - works - I did not notice any time difference.

🟡 I tried with 3000 - fails:
```
Error: Test timed out in 30000ms.
If this is a long-running test, pass a timeout value as the last argument or configure it globally with "testTimeout".
    at Timeout.<anonymous> (file:///Users/mkondratek/IdeaProjects/cody/node_modules/.pnpm/@vitest+runner@1.2.0/node_modules/@vitest/runner/dist/index.js:63:16)
    at listOnTimeout (node:internal/timers:573:17)
    at processTimers (node:internal/timers:514:7)
```

agent/src/index.test.ts
```diff
@@ -337,6 +338,45 @@ describe('Agent', () => {
...
+                date.setTime(date.getTime() + 60 * 1000)
+            }
+
+            // Step 3: export history
+            const exportHistoryChatId = await client.request('chat/new', null)
+            const chatHistory = await client.request('chat/export', {
+                id: exportHistoryChatId,
+            })
+
+            // Step 4: delete the chat that corresponds to `exportHistoryChatId` as its id cannot be predicted or mocked
```
dominiccooney, line 374: This is OK but would it be simple to load the object at `historyPath` and inject a clone of the new chat into it and match on that?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -270,6 +270,11 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
             case 'chatModel':
                 this.handleSetChatModel(message.model)
                 break
+            case 'history':
```
dominiccooney, line 273: This is interesting... so we used to have a history WebView -> extension message, it is unused, but you're adding it back?

dominiccooney (changes requested):
We should do this differently, for these reasons:

The Agent API for chat/export seems odd. It takes a panel ID, but returns all the local history. So is the intent here to export one chat (why return all the history?) or all the chats (why do you need a panel ID?)

I see the history, action = export web view message exists, but it seems to be no longer used. saveSession is called in a few places (initialization, restarting chats, submitting messages) so it doesn't fill me with confidence that the chat/export and history messages won't get crossed. And it seems odd that chat/export has a side effect of saving.

Can you clarify what the API for chat/export is intended to do? Then let's look at how we can add methods to the extension objects so that Agent can get the behavior it needs without having to hook WebView messages.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
mkondratek: Oh, I did not realised due to two renames in the past:
`ChatViewProvider` ➡️  `SidebarChatProvider` ➡️ `SimpleChatPanelProvider` 

You are probably right - I am adding it back (I am not sure if it's wrong 😅 )

agent/src/index.test.ts
mkondratek: test simplified

agent/src/index.test.ts
```diff
@@ -337,6 +338,45 @@ describe('Agent', () => {
             )
         }, 30_000)
 
+        it('chat/restore (multiple) & export', async () => {
+            const date = new Date(1997, 7, 2, 12, 0, 0, 0)
+
+            // Step 1: Create a chat session asking what model is used.
+            const id1 = await client.request('chat/new', null)
+            const reply1 = asTranscriptMessage(
```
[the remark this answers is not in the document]
mkondratek: fixed, thanks

agent/src/index.test.ts
```diff
@@ -337,6 +338,45 @@ describe('Agent', () => {
...
+
+            // Step 3: export history
+            const exportHistoryChatId = await client.request('chat/new', null)
+            const chatHistory = await client.request('chat/export', {
+                id: exportHistoryChatId,
+            })
+
+            // Step 4: delete the chat that corresponds to `exportHistoryChatId` as its id cannot be predicted or mocked
+            const key = Object.keys(chatHistory).find(x => !chatHistory[x].id.startsWith('1997'))
+            if (key) delete chatHistory[key]
```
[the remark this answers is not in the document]
mkondratek: test simplified

agent/src/agent.ts
```diff
@@ -775,6 +776,15 @@ export class Agent extends MessageHandler {
             return { models: panel.models ?? [] }
         })
 
+        this.registerAuthenticatedRequest('chat/export', async ({ id }): Promise<ChatHistory> => {
+            const panel = this.webPanels.getPanelOrError(id)
+            await this.receiveWebviewMessage(id, {
+                command: 'history',
+                action: 'export',
+            })
+            return panel.chatHistory || {} // todo: proper handling
```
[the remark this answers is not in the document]
mkondratek: I left that todo just to rethink it next day 😄 seeing it now - an empty chat history seems to be a valid case so it's ok

pkukielka (approved):
LGTM

mkondratek:
hello @dominiccooney 👋 thank you for your review! I do not feel very confident in this area so I really appreciate your input 

I agree with your points - I got some sleep, took a look at your comments, and managed to improve my solution.

Please, revisit my PR, close comments that you consider as fixed now, suggest me whatever you think makes more sense 🙏

dominiccooney (approved):
Fantastic, this is nice and clean.

PR #3493, opened 2024-03-21
[LSP Context]: Add `lsp-light` graph context retriever

valerybugakov:
## Context

- Adds a proof of concept for the new graph context strategy, `lsp-light`, that uses LSP APIs to collect relevant symbol snippets from the codebase.
- This implementation is built on top of the learnings from [previous experiments](https://sourcegraph.slack.com/archives/C05AGQYD528/p1699261855821009?thread_ts=1699022249.057729&cid=C05AGQYD528). It prioritizes performance and efficiency over accuracy.:
    - We use hovers to create context snippets AND extract nested symbols. While hover strings are not valid code snippets and can produce invalid syntax trees, this method reduces the number of LSP calls per symbol. If this approach proves ineffective, we can switch to a more costly but predictable logic.
    - The indexing logic is heavily cached and rate-limited to avoid performance issues from high LSP request volumes. All LSP calls are wrapped in OpenTelemetry traces for performance analysis during internal testing.
    - We use tree-sitter to collect symbol names, providing a shorter list tailored to our use cases using tree-sitter queries.
    - We combine hovers and go-to-definition requests to gather nested types required for context. Hover snippets alone are insufficient. Local testing with Typescript showed that recursive resolution of context snippets is necessary for high-quality context. For example, having `interface CacheItem extends AutocompleteItem {}` or `function myFunc(): Result` in the LLM request context without resolving `AutocompleteItem` and `Result` won't increase suggestion accuracy. Tree-sitter is also used for recursive symbol resolution. 
    - The symbol resolution logic focuses on Typescript/Javascript due to their popularity. A few places have language-specific logic related to hover-string parsing, making it easy to add support for new languages as needed.
- Related [RFC 901: the state of graph-based Cody context](https://docs.google.com/document/d/1V2SB-ixi8OrhTwCMPO0MseBkA4tnipAkse5CCXbucXI/edit#heading=h.trqab8y0kufp)
- Part of https://github.com/sourcegraph/cody-issues/issues/151
- Closes https://github.com/sourcegraph/cody-issues/issues/238
- Closes https://github.com/sourcegraph/cody-issues/issues/202
- Closes https://github.com/sourcegraph/cody-issues/issues/203

## Integration tests

- **Current approach:** This PR implements integration tests with the [vscode-jsonrpc](https://www.npmjs.com/package/vscode-jsonrpc) and [typescript-language-server](https://www.npmjs.com/package/typescript-language-server) where we spawn the language server process on the side and use vscode jsonrpc to communicate with it. 
    - Integration test [example](https://github.com/sourcegraph/cody/blob/01242e9188a6370a1a1af81b41a5cf4929202510/vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts#L98-L112). 
    - This allows us to use all the LSP commands available in the VS Code without much additional logic. It also opens the door for adding integration tests for other languages. For example, Python is the next candidate based on its popularity among Cody users.
- **Why not use `tsc` directly?** I experimented with using the Typescript compiler language service directly. It works like a charm and is easy to integrate. However, it operates at a lower level of abstraction than the LSP commands. That's why it lacks [additional data post-processing](https://github.com/typescript-language-server/typescript-language-server/blob/b224b878652438bcdd639137a6b1d1a6630129e4/src/lsp-server.ts#L568-L601) implemented on the language-sever layer. It's easy to re-implement this logic on top of [tsc APIs](https://github.com/microsoft/TypeScript/blob/79a851426c514a12a75b342e8dd2460ee6615f73/src/harness/client.ts#L252). Still, it means digging into extra details, maintaining it later + no compatibility with other languages that I want to test later.

## Demo recording

https://github.com/sourcegraph/cody/assets/3846380/2c374094-a208-4a87-91db-83e94981aa88

## Test plan

1. CI
2. Adds unit tests 
3. Adds integration tests
5. Manual testing with autocomplete: add `"cody.autocomplete.experimental.graphContext": "lsp-light"` to your user settings.


vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts
```diff
@@ -0,0 +1,328 @@
...
+        const contextSnippets = await retriever.retrieve({
+            document: mainDocument,
+            position: positionForSubstringInSnippet('new Greeter("world")', 'Greeter'),
+            hints: { maxChars: 100, maxMs: 1000 },
+        })
+
+        expect(contextSnippets.length).toBe(1)
+        // TODO: add only constructor to context snippets
+        expect(contextSnippets[0].content).toMatchInlineSnapshot(`
+          "interface GreeterConfig {
```
olafurpg, line 165: non-blocking. The full text of the class definition is quickly going to eat up the token window budget. It's possible to compress classes into a much smaller summary like we do here  https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/completions/context/retrievers/tsc/SymbolFormatter.test.ts?L112-114

It's possible to do this just with a syntax tree alone, don't need a typechecker.

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts
```diff
@@ -0,0 +1,328 @@
...
+    } satisfies Pick<typeof vscode.workspace, 'openTextDocument'>,
+}))
+
+const disposable = {
+    dispose: () => {},
+}
+
+// To run these tests only with clean debug output:
+// pnpm vitest vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts --disableConsoleIntercept --hideSkippedTests --reporter=basic
+// TODO: fix windows tests. Probably caused by the issues files paths.
```
olafurpg, line 40: It could be because `(property) ts.ParsedCommandLine.fileNames: string[]` uses `/` slashes on Windows as well (IIRC). I had similar issues running tsc-retriever tests on Windows. Either way, it's fixable, I have a Windows computer to troubleshoot.

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts
```diff
@@ -0,0 +1,328 @@
...
+        mockLspCommands(connection)
+
+        const workspaceFileURIs = await openWorkspaceFiles(connection)
+        mainFileUri = workspaceFileURIs.find(uri => uri.fsPath.endsWith('main.ts'))!
+        mainDocument = documentFromFilePath(mainFileUri.fsPath)
+        parseDocument(mainDocument)
+    })
+
+    afterAll(() => {
+        connection.dispose()
```
olafurpg, line 72: To strictly follow LSP, you should send a shutdown request and exit notification. Not a big deal though

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.ts
```diff
@@ -0,0 +1,162 @@
...
+        const contextSnippets = await getSymbolContextSnippets({
+            symbolsSnippetRequests,
+            recursionLimit: RECURSION_LIMIT,
+            abortSignal: abortController.signal,
+        })
+
+        this.lastResultKey = resultKey
+        this.lastResult = contextSnippets
+
+        debugLspLightLog(
```
olafurpg, line 107: nit: we are creating a very large amount of strings in these logging statements that are not logged by default

olafurpg (approved):
LGTM 👍🏻 Really nice work @valerybugakov 

I tried this locally and it feels very nice. It seems like most of the results are served from cache so it adds very little overhead to context fetching. I love that it also works for Go! 👏🏻 

CleanShot 2024-05-22 at 12 54 43@2x

CleanShot 2024-05-22 at 12 54 19@2x

I'll need to test this on a heavy coding session to evaluate whether it overloads the LSP server or not. This is definitely ready for internal testing :letsgoo:!

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts
valerybugakov: Easy fix: https://github.com/sourcegraph/cody/pull/3493/commits/d3286fb7fef9f28bc85da2c55155d51da26132ee

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts
valerybugakov: Yeah, it's a minor issue. I do not have a Windows machine to test, and debugging it remotely is not worth it at this stage.

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.ts
valerybugakov: moved behind explicit if statements

vscode/src/completions/context/retrievers/lsp-light/lsp-light-retriever.integration.test.ts
valerybugakov: Nice! I created a follow-up issue for it https://linear.app/sourcegraph/issue/CODY-1805/[lsp-context]-compress-classes-into-a-smaller-summary

PR #3494, opened 2024-03-21
Edit: Show warning on files that are too large

umpox:
closes https://github.com/sourcegraph/cody/issues/2965

## Description

This PR:
- Adds a warning alongside large files that are @ mentioned in the Edit input
- Shows open tabs by default in the Edit input, as soon as @ is triggered

image

## Test plan

1. Start an edit input, use @ mention
2. Check open tabs are shown immediately
3. Check a warning appears against a large file


philipp-spiess (approved):
Noice

vscode/src/edit/input/get-matching-context.ts
```diff
@@ -11,17 +15,25 @@ interface FixupMatchingContext {
...
     /* If present, will override the key shown in the quick pick selector */
     shortLabel?: string
-    file: ContextItem
+    item: ContextItem
 }
 
 export async function getMatchingContext(
     mentionQuery: MentionQuery
 ): Promise<FixupMatchingContext[] | null> {
+    if (mentionQuery.type === 'empty') {
```
abeatrix, line 24: Maybe can use `getChatContextItemsForMention` here so that every time we add a new context item (like the URL context we just added a few days ago) we won't need to add it again for Edit separately?  Something like:

```suggestion
export async function getMatchingContext(
    mentionQuery: MentionQuery
): Promise<FixupMatchingContext[] | null> {
    const token = new vscode.CancellationTokenSource()?.token
    const results = await getChatContextItemsForMention(mentionQuery.text, token)
    if (!results?.length) {
        return null
    }
    const resultsAsMenuItems = results.map(result => ({
        key: getLabelForContextItem(result),
        item: result,
        shortLabel:
            result.type === 'symbol'
                ? `${result.kind === 'class' ? '$(symbol-structure)' : '$(symbol-method)'} ${
                      result.symbolName
                  }`
                : undefined,
    }))
    return resultsAsMenuItems
}
```

This adds support for URI as context:
image

abeatrix (approved):
Works smoothly! Left some comments inline. 

I think some of the logics are duplicated and can be shared between chat and edit, we should work on that in a follow-up 😄

abeatrix:
@umpox i just pushed my changes to this branch if you want to test it out :) https://github.com/sourcegraph/cody/pull/3498 

image

umpox:
@abeatrix Awesome, thank you.

I noticed a bug where symbol results don't show at all for edit or chat, when using `@#`. Looking into why 🤔

abeatrix:
> @abeatrix Awesome, thank you.
> 
> 
> 
> I noticed a bug where symbol results don't show at all for edit or chat, when using `@#`. Looking into why 🤔 

Ohhh do you think it's fixed in https://github.com/sourcegraph/cody/pull/3503?

abeatrix:
> > @abeatrix Awesome, thank you.
> > I noticed a bug where symbol results don't show at all for edit or chat, when using `@#`. Looking into why 🤔
> 
> Ohhh do you think it's fixed in #3503?
I pulled your branch and it's working for me for both Chat and Edit 👀 

image

image

toolmantim (approved):

PR #3497, opened 2024-03-21
Test: fix flaky e2e tests related to count

abeatrix:
Attempt to fix the flaky e2e tests.

### custom-commands.test.ts

Notice we were expecting the new command to show up on the sidebar and in the menu, so we originally were expecting 2 elements with the new command name:
```
await expect(page.getByText(commandName)).toHaveCount(2)
```
However, in the screenshot captured from the recordings shown below, the same command name will also appear in the background in the editor and on hover, leading the test to fail. To fix this, i've updated the test to expect any number greater than 1, and then click on the actual element to make sure they show up in the expected places.

image

### chat-edits.test.ts

In some failed recordings, I noticed the `await expect(editButtons).toHaveCount(3)` would fail because the third message wasn't submitted because the second message was still in process. To fix this, i've added additional checks after each message is submitted to ensure the request was completed before we move to the next one.

The test 

## Test plan

Green CI

### DIFF
--- a/vscode/test/e2e/chat-edits.test.ts
+++ b/vscode/test/e2e/chat-edits.test.ts
@@ -37,15 +37,16 @@ test.extend<ExpectedEvents>({
     const startNewChatButton = chatFrame.getByTitle('Start New Chat')
 
     // Submit three new messages
-    await chatInput.fill('One')
-    await chatInput.press('Enter')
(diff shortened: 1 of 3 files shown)


kalanchan (approved):
gogogogo!

PR #3498, opened 2024-03-21
update: share at mention logics between edit and chat

abeatrix:
Follow up on my command in https://github.com/sourcegraph/cody/pull/3494

Updated to share the duplicated at-mention labels and logic to fetch for at-mention context items between edit and chat.

This also adds the missing support for URL as context item to Edit:

image

## Test plan

Green CI


umpox (approved):
Looks great, thank you!!

PR #3501, opened 2024-03-22
Chat: update at-mention input token color

abeatrix:
Updates the color for the at-mention input tokens to make it more obvious to user. Right now it's hard to tell between a token and regular input.

Co-author with @toolmantim 

### Before 

image

### After

image


toolmantim (approved):
Looking great! I checked themes, and the only thing I spotted was that light (high contrast) could do with a strong background color (maybe we just match the `code` style in the trascript, which is light text on very dark bg)

CleanShot 2024-03-22 at 13 17 10@2x

toolmantim:
Turns out this doesn't work if things are broken between lines:

CleanShot 2024-03-26 at 00 37 54@2x

CleanShot 2024-03-26 at 00 37 43@2x

PR #3489, opened 2024-03-21
do not cancel throttled findWorkspaceFiles call

sqs:
(CROSSES FINGERS...) Deflakes the @-mention file e2e tests on Windows.

## Test plan

CI

PR #3503, opened 2024-03-22
Chat: skip invalid at-mention request

abeatrix:
Follow-up on https://github.com/sourcegraph/cody/pull/3489

Backgroud: While debugging the cancellation token issue yesterday ([slack](https://sourcegraph.slack.com/archives/C05AGQYD528/p1710984547497189?thread_ts=1710973220.764839&cid=C05AGQYD528)), I noticed the `handleGetUserContextFilesCandidates`function was unexpectedly getting called with an empty query when the chat view is first mounted, or when the input box is re-focused after closing the @ mention box.

This pull request addresses an issue in the `atMentions` plugin where empty and null query states were not handled correctly. The changes ensure that the plugin distinguishes between a valid empty query used for retrieving open tabs context and a null query indicating a non-mention scenario that we can skip.

- Improved handling of empty and null query states in the `atMentions` plugin.
- Clearer distinction between valid empty queries and non-mention scenarios.
- Reduced unnecessary re-rendering by limiting the `useEffect` dependency array in the `useChatContextItems` hook to `query`.

### Changes

- Update the `query` state in the `MentionsPlugin` component to be of type `string | null` instead of `string`.
- Modify the `onQueryChange` callback to directly set the `query` state without fallback to an empty string.
- In the `useChatContextItems` hook:
  - Add a check to set the `results` state to `undefined` when the `query` is null, indicating a non-mention scenario.
  - Update the `useEffect` dependency to include the `query` variable only, as we only want to rerender on query change.

## Test plan

- Verify that the `atMentions` plugin behaves correctly when the query is empty (retrieving open tabs context) and when the query is null (non-mention scenario).
- Ensure that the typeahead menu and options are displayed and updated appropriately based on the query state.
- Green CI

### DIFF
--- a/vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx
+++ b/vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx
@@ -54,10 +54,18 @@ function useChatContextClient(): ChatContextClient {
 }
 
 /** Hook to get the chat context items for the given query. */
-export function useChatContextItems(query: string): ContextItem[] | undefined {
+export function useChatContextItems(query: string | null): ContextItem[] | undefined {
     const chatContextClient = useChatContextClient()
     const [results, setResults] = useState<ContextItem[]>()
+    // biome-ignore lint/correctness/useExhaustiveDependencies: we only want to run this when query changes.
     useEffect(() => {
+        // An empty query is a valid query that we use to get open tabs context,
+        // while a null query means this is not an at-mention query.
+        if (query === null) {
+            setResults(undefined)
+            return
+        }
+
         // Track if the query changed since this request was sent (which would make our results
         // no longer valid).
         let invalidated = false
@@ -80,6 +88,6 @@ export function useChatContextItems(query: string): ContextItem[] | undefined {
         return () => {
             invalidated = true
         }
-    }, [chatContextClient, query])
+    }, [query])
     return results
 }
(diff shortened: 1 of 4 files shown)


vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx
```diff
@@ -80,6 +88,6 @@ export function useChatContextItems(query: string): ContextItem[] | undefined {
         return () => {
             invalidated = true
         }
-    }, [chatContextClient, query])
+    }, [query])
```
sqs, line 91: What is the impact of removing `chatContextClient` from the deps array here? I would not think that value is changing anyway.

sqs (approved):
Thanks!

vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx
abeatrix: Since we only want to run the effect when the query changes ( [commented here](https://github.com/sourcegraph/cody/pull/3503/files#diff-94087ba04bd0d80f6a11f23cbe7c1361c4f49a5877d96e9b43e4ce4f92ae9ab5R60)), I thought it was unnecessary to keep an unchanged dependency in the deps array. It doesn't make any impact apart from making it clear (imo) that this effect only depends on the query value, but happy to add it back if you find it confusing!

PR #3507, opened 2024-03-22
Edit/Chat: Always expand to the nearest enclosing function, if available, before folding ranges

umpox:
## Description

This PR:
- Adds Tree-sitter querys for `getEnclosingFunction` in languages: JS/TS/Go/Python
- Updates `getSmartSelection` to run this query first, before using the folding range logic.

## Test plan

1. Start an edit, using just a cursor position.
2. Check that the edit is correctly expanded to the nearest function.

An example of where this is a significant improvement, is in JS test files that use `describe` and `it` blocks.
- Try putting cursor inside an `it` block and starting an edit
- Notice that the range only expands to the entire `it` now, whereas before it would expand to the full `describe`


dominiccooney (approved):
NICE! 🎉

Issue #3522, opened 2024-03-22
We should not silently exclude a file that has been explicitly @ mentioned


abeatrix:
Tie our warning to the actual operation of the product. If we say it's too large, the file should not be able to be input

We should not silently exclude a file that has been explicitly @ mentioned

PR #3523, opened 2024-03-22
Chat: Disable adding large-file via @-mention

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3522

A new check has been added to the contextItemMentionNodeDisplayText function. If the contextItem is of type 'file' and has the isTooLarge flag set to true, the context item cannot be added as an @-mention token.

In the Item component, a new warning prop has been added to conditionally apply the .disabled class to the option item. This is used to visually indicate that a large file is disabled and cannot be selected from the context list:

image

## Test plan

Try to @ a large file
You should not be able to add the file

### DIFF
--- a/vscode/webviews/promptEditor/plugins/atMentions/OptionsList.module.css
+++ b/vscode/webviews/promptEditor/plugins/atMentions/OptionsList.module.css
@@ -72,7 +72,7 @@
 }
 .option-item.disabled {
     opacity: 0.5;
-    cursor: not-allowed;
+    cursor: default;
 }
 
 body[data-vscode-theme-kind='vscode-high-contrast-light'] .option-item.selected,
--- a/vscode/webviews/promptEditor/plugins/atMentions/OptionsList.module.css
+++ b/vscode/webviews/promptEditor/plugins/atMentions/OptionsList.module.css
@@ -70,6 +70,10 @@
     color: var(--vscode-list-hoverForeground);
     background-color: var(--vscode-list-hoverBackground);
 }
+.option-item.disabled {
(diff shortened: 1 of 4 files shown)


vscode/webviews/promptEditor/plugins/atMentions/OptionsList.module.css
```diff
@@ -70,6 +70,10 @@
     color: var(--vscode-list-hoverForeground);
     background-color: var(--vscode-list-hoverBackground);
 }
+.option-item.disabled {
+    opacity: 0.5;
+    cursor: not-allowed;
```
toolmantim, line 75: VS Code convention is just to use the normal cursor on default elements:

```suggestion
    cursor: default;
```

It would also be good not to show any blue hover background color for disabled options too (see the behaviour hovering over macOS disabled menu items)

toolmantim (commented):
Great! Just some small changes

philipp-spiess (approved):

PriNova:
This is a great addition, but now the user cannot even use line ranges to narrow the context of the 'too large file'.
Will this be considered?

Thank you

hlhdaiaaii:
> This is a great addition, but now the user cannot even use line ranges to narrow the context of the 'too large file'. Will this be considered?
> 
> Thank you

Please consider this. Line ranges is indeed the feature I use the most to manually gather best context to feed to the model.

abeatrix:
@PriNova @hlhdaiaaii Thanks for raising that issue with us! I've filed an issue for tracking purpose, and will look into supporting range for large files again!

PR #3526, opened 2024-03-22
Chat: fix at-mention token size

abeatrix:
RE: https://github.com/sourcegraph/cody/pull/3523 & https://github.com/sourcegraph/cody/issues/2636

This PR adds character limit tracking for chat models in the SimpleChatModel class. It introduces a new property `maxChars` that stores the maximum number of characters allowed for the selected chat model. 
The `charsLeft` property is also added to calculate the remaining characters available based on the current chat messages without context.

These changes allow the chat interface to track the character limit for the selected chat model and ensure that the at-mention context items are showing the correct warning for large files based on the latest available token limit instead of a standardized one that doesn't represent the latest limit.

## Test plan

Try `@Simple` and look for `SimpleChatModel.ts` in the cody repository following the examples below

### Before

@ a file that is showing `File too large` warning, but when you select it, it will still be included in your message because the warning size was using the incorrect limit number. 

image

You can see Cody is able to reply to you because the file was still included, meaning the warning was incorrectly displayed:

image

### After

Type `@Simple` and look for `SimpleChatModel.ts` again, you can see `SimpleChatModel.ts` doesn't have the warning anymore, and selecting it still works:

image

However, keep asking Cody question so that the context window will become smaller, and when the messages is larger than the file, then the warning will show up correctly for the file:

image

## Limitation

The large @-mention files can still be added by the user, which has been addressed in a separate PR.

Currently cannot dynamically calculate the token limit change for a resubmitted query, meaning users won't see the file size warning if they have selected an at-mention item that used all the limit, but they can  still add more at-mention items. This will also be addressed in a future PR. 

image

## Follow-up

- [ ] Do not allow users to add large fill with @-mention: https://github.com/sourcegraph/cody/pull/3523
- [ ] Calculate token limit with in-flight input dynamically


toolmantim (approved):
I was able to add an impressive amount of context:

CleanShot 2024-03-26 at 14 57 59@2x

Large files showed the correct thing:

CleanShot 2024-03-26 at 14 59 06@2x

Issue #3530, opened 2024-03-23
bug: incorrect @-symbol error "No symbols found (try installing language extensions)"


abeatrix:
### Version

v1.10.0

### Describe the bug

[Reported by a user on Discord](https://discord.com/channels/969688426372825169/1220497421860212756/1220497421860212756):  Cody chat reports symbols cannot be found. Cody instructions appears to have access to symbols, but in testing it appears it is not including them in context either.

![image](https://github.com/sourcegraph/cody/assets/68532117/f93eef8b-9b6d-45a3-bab9-34d679427ed3)

### Expected behavior

Display the correct information and instruction

PR #3531, opened 2024-03-23
Chat: fix at symbol title

abeatrix:
https://github.com/sourcegraph/cody/issues/3530

A user on Discord has reported that Cody is incorrectly showing an error message "No symbols found (try installing language extensions and opening a file)":

image

This error message/tooltip is incorrect as the result is expected to be empty when the query is empty, and not required to have a file opened for symbol search, this PR reverts the tooltip back to `'Search for a symbol to include...'` to display on start-up screen.

## Test plan

- Type `@#` in the input, you should see `'Search for a symbol to include...'`
- `No symbols found` is displayed when no result is returned after typing more than 1 character

image

image

### DIFF
--- a/vscode/test/e2e/chat-atFile.test.ts
+++ b/vscode/test/e2e/chat-atFile.test.ts
@@ -315,9 +315,15 @@ test.extend<ExpectedEvents>({
     // Go back to the Cody chat tab
     await page.getByRole('tab', { name: 'New Chat' }).click()
 
-    // Symbol empty state
+    // Symbol empty state shows tooltip to search for a symbol
     await chatInput.fill('@#')
-    await expect(chatPanelFrame.getByRole('heading', { name: /No symbols found/ })).toBeVisible()
+    await expect(
+        chatPanelFrame.getByRole('heading', { name: /^Search for a symbol to include/ })
+    ).toBeVisible()
+
+    // Symbol empty symbol results updates tooltip title to show no symbols found
+    await chatInput.fill('@#invalide')
+    await expect(chatPanelFrame.getByRole('heading', { name: /^No symbols found/ })).toBeVisible()
 
     // Clicking on a file in the selector should autocomplete the file in chat input with added space
     await chatInput.fill('@#fizzb')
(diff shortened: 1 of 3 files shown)


toolmantim:
I'd like to leave this message in if there are 2 or more chars after `@#` but no results. Only because sometimes the VS Code API does return results after a period of time (and there's no API for loading state).

But we can make the message:

"No symbols found. Symbols may be loading, or language extensions not installed."

(May need overflow ellipsis)

abeatrix:
> I'd like to leave this message in if there are 2 or more chars after `@#` but no results. Only because sometimes the VS Code API does return results after a period of time (and there's no API for loading state).
> 
> But we can make the message:
> 
> "No symbols found. Symbols may be loading, or language extensions not installed."
> 
> (May need overflow ellipsis)

I think there's a way to check if symbol providers are available in the current workspace or not to display the tooltip more accurately. Let me check on Monday to confirm, and we can add the "(symbols may be loading)" only if that works? @toolmantim

toolmantim:
Oh amazing! That would be even better.

abeatrix:
@toolmantim instead of "No symbols found (symbol maybe be loading)", what do you think about "Searching symbols..." and then display "No symbols found" after a timeout?

toolmantim:
I thought the API immediately returns a nil result? But if that's not right, or you're thinking of polling/retrying… some kind of experience like that could be nice. How would you calculate the timeout though, because huge codebases could take a very long time, and I'd still want us to show "(language extensions may be loading)"

abeatrix:
> I thought the API immediately returns a nil result? 
That's my understanding as well but I misunderstood you when you said "only because sometimes the VS Code API does return results after a period of time (and there's no API for loading state)", which I just realized you are referring to when the symbol provider is initiating.

How about `No symbols found (is symbol available in your workspace?)` or `No symbols found (make sure workspace symbols are available)`? I think the current issue is that we are showing misleading label to users that is causing confusions: https://github.com/sourcegraph/cody/issues/3530

toolmantim (approved):
It's looking good, but I wouldn't show the "(language extensions may be loading)" above two characters, because there's a good chance there's simply no matching symbols:

CleanShot 2024-03-26 at 15 18 15@2x

Two characters and under, you can be pretty sure there should always be a result, so something went wrong.

Issue #3254, opened 2024-02-23
feedback: Specify unit test method.


James-Morrell:
### Version

1.86.2

### Areas for Improvement

- [ ] UI/UX
- [X] Onboarding
- [ ] Docs
- [ ] Chat
- [X] Commands
- [X] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience.

Using the Generate Unit Tests function with C# doesn't always give the correct framework response.

Example:
Project is using Xunit, but sometimes the generated code would be Nunit

### Describe the solution you'd like to see

When using the Generate Unit Tests function for the first time, ask for input of which test framework to use.

abeatrix:
@James-Morrell Thank you so much for your feedback!  Do you mind sharing some information with us so we can look into this further?
1. What's the naming convention of the Xunit test files in your project? For example, `example.Tests.cs` or `exampleTest.cs` 
2. What's your Cody version? (1.86.2 is the version of your VS Code. You can find your Cody version next to the Release Note icon in your Cody's SETTING & SUPPORT sidebar:
![image](https://github.com/sourcegraph/cody/assets/68532117/eff387ee-11f3-46a6-a502-1718faf7e152)

James-Morrell:
@abeatrix I'm running v 1.6.0 of Cody.

My usual naming convention for tests is `ExampleTests.cs`, where Example would be the class name being tested.
On occasion when a set of tests for a single method gets very large I'll split the test class like so:
`ExampleTests_FuncOne.cs`

PR #3542, opened 2024-03-25
Edit: include pre-instruction to Edit commands

abeatrix:
CLOSE: https://github.com/sourcegraph/cody/issues/3254

This pull request fixes an issue where the `cody.chat.preInstruction` configuration option was not being applied to the preamble used by the Edit feature. 

This adds a new configuration option ``cody.edit.preInstruction`` to the end of human text to make sure it is included in the prompt to the LLM for all edit commands

## Test plan

1. Configure `cody.edit.preInstruction`
2. Execute a Edit command
3. Check the output channel to confirm if the `cody.chat.preInstruction` value is added to the preamble for Edit

Example when setting `cody.chat.preInstruction` to "When writing unit tests, always use the "Jest" testing framework instead of detected framework. " and write the "Unit Test" command:

image


toolmantim:
I don't think the `cody.chat.preInstruction` should affect edits, because it's scoped to `cody.chat`

We could introduce another one for edits: `cody.edit.preInstruction`

vscode/package.json
```diff
@@ -887,6 +887,12 @@
           "markdownDescription": "A custom instruction to be included at the start of all chat messages. (E.g., \"Answer all my questions in Spanish.\")",
```
toolmantim, line 887:
```suggestion
          "markdownDescription": "A custom instruction to be included at the start of all chat messages (e.g. \"Answer all my questions in Spanish.\")",
```

vscode/package.json
```diff
@@ -887,6 +887,12 @@
...
           "examples": ["Answer all my questions in Spanish."]
         },
+        "cody.edit.preInstruction": {
+          "order": 7,
+          "type": "string",
+          "markdownDescription": "A custom instruction to be included at the start of the prompt for your edit commands.",
```
toolmantim, line 893:
```suggestion
          "markdownDescription": "A custom instruction to be included at the start of all instructions for edit commands (e.g. \"Write all unit tests with Jest instead of detected framework.\")",
```

toolmantim (approved):
Works a treat! Left two small suggestions for the settings copy.

CleanShot 2024-03-26 at 15 26 40@2x

vscode/src/edit/prompt/index.ts
```diff
@@ -102,7 +102,11 @@ export const buildInteraction = async ({
     const preamble = getSimplePreamble(model)
     promptBuilder.tryAddToPrefix(preamble)
 
-    const transcript: ChatMessage[] = [{ speaker: 'human', text: prompt }]
+    const preInstruction: string | undefined = vscode.workspace
+        .getConfiguration('cody.edit')
+        .get('preInstruction')
+    const editRule = preInstruction ? `\nIMPORTANT: ${preInstruction}` : ''
```
umpox, line 108: Is the `IMPORTANT` necessary here?

We don't do this for chat AFAICT

umpox (approved):

vscode/src/edit/prompt/index.ts
abeatrix: Good question! For chat, we mix the pre-instruction with the preamble, but here we are adding it to the end of human's prompt as a way to override the instruction given by the prompt. The `IMPORTANT: ` is basically used as `Response guidelines:` but from what i've tried so far (`IMPORTANT`, `Response guidelines`, `Additional Request`) they all work consistently, so I picked important cause it has the less char count 😅 

I will add a comment to reflect this.

Issue #2839, opened 2024-01-21
Improve discoverability/understanding/onboarding for context sources (embeddings/symf for PLG, repo & multi-repo for Enterprise)


taylorsperry:
We know that the "Enhanced Context" button isn't easily discoverable for users. This is a problem for our Free/Pro SKU because we need users to enable embeddings so that we can continue learning about the value that embeddings context brings to chats.

This issue is a follow-up to https://github.com/sourcegraph/cody/issues/2838. The goal is to make Enhanced Context more discoverable by adding a description to the Enhanced Context button prompting users to "Complete Setup: Enable Embeddings." See design below. 

**Only Free/Pro users should see this notification.** We'll add a different one for enterprise users once multi-repo context is enabled.

The description should only be displayed under the following conditions: 
- The user is a free/pro user
- The user has opened a new folder that doesn't have embeddings. (The description should only appear once. Whether or not the user enabled embeddings for the folder, they won't see the description again the next time they open the folder.) 

To measure the effectiveness of this CTA, we should add an event that we can log. The simplest solution is probably to adapt the existing event https://github.com/sourcegraph/cody/issues/2609#event-11420138487.

https://github.com/sourcegraph/cody/assets/43555476/3145af38-181f-4039-9975-820e419fb081

kalanchan:
@chillatom needs follow up

toolmantim:
I've an updated design here that covers both PLG and Enterprise users:
https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1447-39182&mode=design&t=cB0ygKAkoURvrspK-4

Let's get this in, we can't have people missing these features.

### PLG

<img src="https://github.com/sourcegraph/cody/assets/153/df228468-9416-4905-a1ac-cd1563e55f16" width="290">

<img src="https://github.com/sourcegraph/cody/assets/153/c464ea9b-ebd6-475f-a5c3-09493110a1dd" width="290">

### Enterprise

<img src="https://github.com/sourcegraph/cody/assets/153/6861fb99-f07e-445a-b096-e1fe587a8c6f" width="290">

<img src="https://github.com/sourcegraph/cody/assets/153/0a7f8e60-78ef-41ba-bf04-244ef17acb9d" width="290">

sqs:
[...]

PR #3547, opened 2024-03-25
Chat: Display Enhanced Context settings on first chat

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/2839

image

- As discussed with @toolmantim in [this figma comment thread](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm?type=design&node-id=1447-38603&mode=design#739536448), we will open Enhanced Settings on first chat by default.
- Add dropdown button for Enhanced Context Settings
- Existing button for Enhanced Context Settings is now for toggling enhanced context enablement.

## Test plan

- Updated e2e tests

To test this manually:
1. Open a new Cody Chat
2. The Enhanced Context settings should be opened by default
3. Send a new message
4. the Enhanced Context settings windows is closed when you move focus to the chat input box

### After

First chat for Dot Com users:

image

First chat for enterprise users:

image


abeatrix:
@toolmantim before I fix the e2e tests to reflect the new behavior, can you take a look and let me know if this aligns with what you had in mind?

toolmantim:
@abeatrix Nice! Yeah checking it out now. Quick one: I spotted `CodyVSCodeExtension:useEnhancedContextToggler:clicked` event isn't triggered when you click it via the new button.

abeatrix:
> @abeatrix Nice! Yeah checking it out now. Quick one: I spotted `CodyVSCodeExtension:useEnhancedContextToggler:clicked` event isn't triggered when you click it via the new button. (edit: we may want to include a metadata property of the source, so we can tell where they click it from — popover or input button)

added a new event `useEnhancedContextInputToggler` event for the new icon click

toolmantim:
Just pushed a bunch of style fixes, and the only thing left I think is some of the keyboard accessibility fixes:
* Tabbing to enhanced context indicator button and hitting space/enter shouldn't refocus the input, that should only happen when context indicator button is clicked. Pressing escape while the enhanced context indicator button is focused should focus the message input.
* If you shift-tab from the popover checkbox back into the message input, the popover should be dismissed

toolmantim (approved):
Approving to unblock, given how close this is

abeatrix:
> If you shift-tab from the popover checkbox back into the message input, the popover should be dismissed
@toolmantim I couldn't get this to work, but I will work on it in a follow-up as this is an existing behavior. For now, users can hit Escape when the input box is focused though to dismiss the popover.

vscode/webviews/Chat.tsx
```diff
@@ -517,8 +526,9 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
                         <div className={styles.contextButton}>
                             <EnhancedContextSettings
                                 isOpen={isEnhancedContextOpen}
-                                setOpen={setIsEnhancedContextOpen}
+                                setOpen={onEnhancedContextTogglerClick}
                                 presentationMode={userInfo.isDotComUser ? 'consumer' : 'enterprise'}
+                                isFirstChat={transcript.length < 1}
```
toolmantim, line 531: I only spotted this logic. This is more like `isEmptyChat` which isn't what we want here. We want it to appear on the first chat that's created for first time users, not every time a chat is created.

PR #3548, opened 2024-03-26
Chat: update @-input token background

abeatrix:
Address issue from https://github.com/sourcegraph/cody/pull/3501#issuecomment-2018037797

Remove `.context-item-mention-node::before` that sets background color with opacity, set `background-color` to `rgba(200, 200, 200, 0.2);` to add blurred background instead.

### Before 

image

### After

image

### DIFF
--- a/vscode/webviews/promptEditor/nodes/ContextItemMentionNode.module.css
+++ b/vscode/webviews/promptEditor/nodes/ContextItemMentionNode.module.css
@@ -1,6 +1,7 @@
 .context-item-mention-node {
-    background-color: rgba(200, 200, 200, 0.2);
     color: var(--link-foreground);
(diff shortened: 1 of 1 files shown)


toolmantim:
I fixed the color background and added a border radius:

https://github.com/sourcegraph/cody/assets/153/c9dbdc55-c84d-436f-9184-f640a24fd085

I really wish we could add some padding, but every time I tried it didn't quite work. This PR also has the side effect of changing the cursor color when you're in a token — ideally we select the token when you click on it or move the cursor into it, similar to how https://github.com/sodenn/lexical-beautiful-mentions works.

I wonder if there's some code/tricks in lexical-beautiful-mentions we could use so we can render and interact with these more nicely?

toolmantim (approved):

PR #3549, opened 2024-03-26
Autocomplete: add singleline stop sequences

valerybugakov:
Adding `\n\n` as a stop sequence for singleline completions. I suspect this is one of the reasons for [higher-than-expected upstream char counts](https://redash.sgdev.org/queries/485/source#1003) in the StarCoder2 experiment.

## Test plan

CI

### DIFF
--- a/vscode/src/completions/providers/anthropic.ts
+++ b/vscode/src/completions/providers/anthropic.ts
@@ -45,7 +45,11 @@ import {
     standardContextSizeHints,
 } from './provider'
 
-export const SINGLE_LINE_STOP_SEQUENCES = [anthropic.HUMAN_PROMPT, CLOSING_CODE_TAG]
+export const SINGLE_LINE_STOP_SEQUENCES = [
+    anthropic.HUMAN_PROMPT,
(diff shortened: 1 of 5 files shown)


vscode/src/completions/providers/anthropic.ts
```diff
@@ -45,7 +45,11 @@ import {
     standardContextSizeHints,
 } from './provider'
 
-export const SINGLE_LINE_STOP_SEQUENCES = [anthropic.HUMAN_PROMPT, CLOSING_CODE_TAG]
+export const SINGLE_LINE_STOP_SEQUENCES = [
+    anthropic.HUMAN_PROMPT,
+    CLOSING_CODE_TAG,
+    MULTILINE_STOP_SEQUENCE,
+]
```
philipp-spiess, line 52: New anthropic `/messages` API doesn't allow you to have whitespace-only stop sequences. 😬  We filter that on the server side so it won't break anything but it also won't have an effect.

philipp-spiess (approved):

Issue #3513, opened 2024-03-22
Edit: Track persistence rate for variants of edit over X seconds/minutes


umpox:
What:
- We don’t have a similar concept of “completion acceptance rate” for Edits. Edits are applied by default, so there is no actual acceptance action.
- We should instead, try to track the persistence of an edit over X seconds/minutes.

How:
- We already have a persistence tracker for completions, we can use this for Edits.
- We need to use chunks from the diff to track specific lines, rather than the whole edit. Tracking the entire selection range will be misleading, as the edit may only actually change a few characters or lines.
- We need to ensure that we correctly calculate a % of code left from the edit, combining both deletions and insertions. For deletions, we need to be tracking that code that Cody deleted, remains deleted.

philipp-spiess:
There's some discussion about this in this thread: https://sourcegraph.slack.com/archives/CN4FC7XT4/p1710299380435099

One problem with persistence is that edits can also just delete code and if that happens we have no ranges that we can track after. We don't know if newly inserted characters are trying to _undo_ the completion or adding more text. Might need some thinking about how we can make this work well 🤔

umpox:
Done

PR #3550, opened 2024-03-26
Edit: Track persistence of edits and insertions

umpox:
## Description

This PR makes `PersistenceTracker` generic and uses it for `Edit`

In its current state, this means we get:
- Accurate tracking for _insertion_ edits (Document, Test, Generate)
- Partially accurate tracking for _edit_ edits (Fix, Instruction-based Edit)

The latter is only partially accurate as it tracks the entire selectionRange, but an Edit might only update part of this range.

Will address this in a follow up as part of https://github.com/sourcegraph/cody/issues/3513

## Test plan

**Insertion edit**
1. Make an insertion edit
2. Delete nothing, check that telemetry is sent with correct `difference` after 30 seconds
3. Repeat but by deleting part of the insertion and then the full insertion, validate the `difference` each time.

**Editing edit**
1. Make an instruction-based edit
2. Delete nothing, check that telemetry is sent with correct `difference` after 30 seconds
3. Repeat but by deleting part of the insertion and then the full insertion, validate the `difference` each time.
4. Note that this `difference` may not accurately reflect Cody's changes


vscode/src/common/persistence-tracker/index.ts
```diff
@@ -26,19 +26,25 @@ interface TrackedCompletion {
...
     constructor(
         workspace: Pick<
             typeof vscode.workspace,
             'onDidChangeTextDocument' | 'onDidRenameFiles' | 'onDidDeleteFiles'
-        > = vscode.workspace
+        >,
+        public logger: {
+            onRemoved: (event: PersistenceRemovedEventPayload<T>) => void
+            onPresent: (event: PersistencePresentEventPayload<T>) => void
+        }
```
umpox, line 47: Change in this file: Consumer now provides the `logger` and handles the telemetry events itself

vscode/src/common/persistence-tracker/index.ts
```diff
@@ -51,116 +57,105 @@ export class PersistenceTracker implements vscode.Disposable {
...
-        const textLines = lines(insertText)
-        const latestRange = new vscode.Range(
-            insertRange.start.line,
-            insertRange.start.character,
-            insertRange.end.line + textLines.length - 1,
-
-            textLines.length > 1
-                ? textLines.at(-1)!.length
-                : insertRange.end.character + textLines[0].length
-        )
```
umpox, line 76: Change in this file: Let the consumer do their own calculation and here we can just assume they provide the correct `insertRange`.

vscode/src/common/persistence-tracker/index.ts
```diff
@@ -26,19 +26,25 @@ interface TrackedCompletion {
     insertText: string
     insertRange: vscode.Range
     latestRange: vscode.Range
+    metadata?: PersistenceEventMetadata
```
umpox, line 29: Change in this file: We support attaching arbitrary metadata that will be sent alongside `onRemoved` and `onPresent`. Used in Edit for tracking persistence across different edit commands (e.g. test, document, etc)

philipp-spiess (approved):
Chef Thumb

Issue #3518, opened 2024-03-22
Edit (Fix): Use Claude Sonnet as the default model for fix


umpox:
What:
- We currently use Claude 2 for “Ask Cody to Fix”
- We will likely see performance improvements by upgrading this model to Claude 3.

How:
- We should ideally run an A/B test when switching this model, and also include GPT 3.5/4.
- This can be paired with the work to track Edit persistence.

PR #3551, opened 2024-03-26
Add support for Cody API v1

philipp-spiess:
Closes #3413

This PR properly infers the Cody API Version (0 for legacy (unversioned) and 1 for API Version 1) and uses that to decide wether to use `system` prompts or not (see [RFC 911](https://docs.google.com/document/d/1GPo7G-FDHidHRMNrKLeutaSTj6y7L0XSdCcSDtncoaU/edit#heading=h.trqab8y0kufp) for more details).

Furthermore this removes the fast-path fork for chat. The latency win (120ms) is neglectable for our chat latencies and this reduces the complexity a lot and makes sure we have life traffic via our server APIs before the enterprise release.

## Open question

I’m not sure if Claude 3 support for Enterprise is supposed to go out with 5.4 or before cc @chwarwick 

## Test plan

Scenarios used for the version infer testing:
```
{endpoint: 'https://sourcegraph.sourcegraph.com/', version: '266411_2024-03-26_5.3-75947b347f3a', isDotCom: false, parsedVersion: null, codyApiVersion: 'v1'}
{endpoint: 'https://sourcegraph.com/', version: '266401_2024-03-26_5.3-08bc88e776c2', isDotCom: true, parsedVersion: null, codyApiVersion: 'v1'}
{endpoint: 'https://demo.sourcegraph.com/', version: '5.3.3', isDotCom: false, parsedVersion: '5.3.3', codyApiVersion: 'legacy'}
{endpoint: 'https://sg02.sourcegraphcloud.com/', version: '266411_2024-03-26_5.3-75947b347f3a', isDotCom: false, parsedVersion: null, codyApiVersion: 'v1'}
{endpoint: 'https://sourcegraph.test:3443/', version: '0.0.0+dev', isDotCom: false, parsedVersion: '0.0.0', codyApiVersion: 'v1'}
```

Tested chat against:
- demo instance (legacy)
- S2
- SG02

PR #3555, opened 2024-03-26
Edit: Default `fix` to Claude sonnet

umpox:
## Description

Changes the `fix` model from Claude 2 to Claude 3 Sonnet

## Test plan

1. Run multiple "Ask Cody to FIx"
2. Check quality of results

closes https://github.com/sourcegraph/cody/issues/3518


umpox:
cc: @chillatom as we spoke a little bit about this previously

umpox:
@philipp-spiess I'm going to test this more thoroughly, but just checking...

Now that https://github.com/sourcegraph/sourcegraph/pull/61396 is in, this should be safe to do, right?

philipp-spiess:
@umpox Yep should be good 👍 Ideally we can pair it with https://github.com/sourcegraph/cody/pull/3551 so we can keep using the old and battle-tested codepaths but it should work regardless

valerybugakov (approved):

PR #3557, opened 2024-03-26
Add size limit when logging chat transcripts

akalia25:
This PR implements a MB limit on logged chat transcripts to prevent exceeding the 10MB pub/sub message size limit([ref](https://cloud.google.com/pubsub/quotas)). 

Without this limit, some long chat transcripts have failed to publish due to exceeding the maximum pub/sub payload size (for example, [this failure](https://sourcegraph.slack.com/archives/C06CCJR4K9R/p1710952854856269)). By capping transcripts before logging, this change will prevent similar publish failures going forward.

## Test plan
Tested locally and confirmed that truncation occurs after `maxTranscriptSize` is exceeded.


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -407,6 +407,10 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                 source,
                 traceId: span.spanContext().traceId,
             }
+            const maxTranscriptSize = 9 * 1024 * 1024 // 9 MB
```
abeatrix, line 410: I think we will need to do the same for edit commands, so probably a good idea to move this to become a global variable?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -407,6 +407,10 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+            if (inputText.length > maxTranscriptSize) {
+                inputText = inputText.substring(0, maxTranscriptSize)
```
abeatrix, line 412: I don't think we want to alter the inputText value here since we are still depending on the value in the functions below (even though we don't support 9MB of messages anyway). Would it work if we just replace the inputText inside privateMetadata with this instead? Something like:
```
                privateMetadata: {
                    ...sharedProperties,
                    // 🚨 SECURITY: chat transcripts are to be included only for DotCom users AND for V2 telemetry
                    // V2 telemetry exports privateMetadata only for DotCom users
                    promptText:
                        authStatus.endpoint && isDotCom(authStatus.endpoint) ? inputText.substring(0, maxTranscriptSize) : undefined,
                },
```

abeatrix (commented):
Thanks for adding this! Left some comments inline 😄

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
akalia25: Really great suggestion!! Thank you!! 🙇

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
akalia25: Good point! I hesitated to make it a global variables since I didn't see any being declared on the file, but glad its the right path forward :)

bobheadxi:
Is 9MB necessarily the right threshold to set? Should it be something lower, like 5MB or even 1MB? What are the standard sizes of these transcripts and how much is realistically useful?

A discerning user might be alarmed to see hundreds of MB worth of data leaving a VSCode extension all the time

abeatrix:
> Is 9MB necessarily the right threshold to set? Should it be something lower, like 5MB or even 1MB? What are the standard sizes of these transcripts and how much is realistically useful?

@bobheadxi good point. 

@akalia25 for the reason described in your PR, I don't think this is the right place. The code you edited is only logging a single user input here, which is for sure going to be less than 1MB. 
I don't think any of our models support 1MB worth of tokens afaik. 
I also don't think we are sending the whole transcript anywhere in our code yet?

akalia25:
Thanks for calling this out!

@bobheadxi, I'm happy to lower the threshold, I just chose the most the data pipeline itself could handle. 4 month's ago we've received transcripts as large as 3mb in size but were averaging < 1mb. So I chose 9mb to be safe. The rest of the event payload we have is quite insignificant data wise (< 1mb), so my rationale was we could allocate a higher threshold for the transcript itself. 

> A discerning user might be alarmed to see hundreds of MB worth of data leaving a VSCode extension all the time

This would only happen if the user was sending _large_ input text consistently. Most of users were averaging < 1mb on their transcripts.


> The code you edited is only logging a single user input here, which is for sure going to be less than 1MB.
> I don't think any of our models support 1MB worth of tokens afaik.
> I also don't think we are sending the whole transcript anywhere in our code yet?

@abeatrix I also thought we would not need this 🙃, but we have been getting [failures](https://sourcegraph.slack.com/archives/C06CCJR4K9R/p1710952854856269) on the telemetry pipeline alerting us that events are coming across >10mb and failing. The failures are all associated with transcript events (`cody.chat-question/submitted`, `cody.chat-question/exectued`) with some users submitting transcript events as large as 68Mb, not sure how this is happening given the limited tokens too 😶‍🌫️  

You two bring up really good points, I'm wondering if we should just be okay on getting these one-off alerts/failures instead of adding complexity to our logging?

bobheadxi:
I think we should set a threshold still. I would suggest something more aggressive like 5MB, but this is fine too

akalia25:
@bobheadxi @abeatrix anything blocking this PR I could resolve 🙇

abeatrix:
@akalia25 I checked, and you're right. Users can just copy and paste a giga-large file in their chat box and then hit submit, and the input text will be logged lol
Maybe we could just use the [MAX_BYTES_PER_FILE](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/prompt/constants.ts?L12) as the limit in here?

akalia25:
Great Suggestion @abeatrix, thank you !! 

Quick question with the [MAX_BYTES_PER_FILE](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/prompt/constants.ts?L12), does this ever change based on LLM provider? Just want to make sure we're always able to capture the entire text for a legitimate input/response :)

abeatrix:
> Great Suggestion @abeatrix, thank you !!
> 
> Quick question with the [MAX_BYTES_PER_FILE](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/prompt/constants.ts?L12), does this ever change based on LLM provider? Just want to make sure we're always able to capture the entire text for a legitimate input/response :)

we are working on standardizing the toklen limit for file input (human input limit in discussion), so i think this would be a good limit to start with, and when we make the change, it will be automatically applied to all the callers :D

abeatrix:
P.S: The current limit for the largest model we support is set to `28000` byte which is 0.224 MB which is still way below the 5MB we originally want to set. We could use that limit here but that'd still be an invalid input.

If you think 4096 from [MAX_BYTES_PER_FILE](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/prompt/constants.ts?L12) is too low, we can use something else, but i think the goal is still set a limit for the logger?

akalia25:
Thanks for explaining! This limit makes a lot of sense and should not restrict us from getting legitimate data :)

abeatrix (approved):
I think this is a good start and we can monitor the data we get from tomorrow's pre-release build to see if we want to raise the limit, since we still have 2 weeks before the next release, if this sound good to you!

PR #3558, opened 2024-03-26
E2E Test: improve flaky tests

abeatrix:
Noticed the @-symbol test fails frequently because the ts language was not ready before we performed the search.

This PR moves the timeout before we perform the search to make sure it gives the server enough time.

The update-notice.test.ts is also failing frequently because the chat window was created before the close button was clicked based on the recordings returned from the CI. Updated the test to get the close button by label instead.

## Test plan

Green CI


kalanchan (approved):

philipp-spiess:
Thank you so much for always looking out for our test health @abeatrix 🤗

abeatrix:
That worked before, but it's still failing on main 😭

PR #3568, opened 2024-03-27
Chat: Fix message input placeholder overflow

toolmantim:
Updates the placeholder styles so it doesn't go over the enhanced context button. Future improvement (which I couldn't get to work) would be to set a `title` so that it shows the full label on hover.

| Before | After|
| - | - |
| CleanShot 2024-03-27 at 14 07 19@2x | CleanShot 2024-03-27 at 14 06 38@2x |

## Test plan

- Opened chat
- Resized chat
- Verified ellipsis


abeatrix (approved):
Thank you!

Issue #3517, opened 2024-03-22
Edit (Document): Use Claude Haiku


umpox:
What:
- Claude Haiku is a faster and improved model compared to Claude Instant

How:
- There is some small formatting issues to consider before switching over to Haiku.
- I think the main one is just that we need to ensure that we always complete the LLM response with a new line when doing an insertion (if it doesn’t already exist). Prior Claude models have done this for us, but Haiku doesn’t appear to.

PR #3572, opened 2024-03-27
Edit: Default `doc` to Claude Haiku

umpox:
## Description

This PR defaults the `doc` command to Claude 3 Haiku.

I noticed that we were relying on this response always applying a new line for it to work correctly. This PR also addresses that by ensuring that all insertions end with a new line when finished.

I think this is the correct behaviour as we currently don't really have a strong use case for in-line insertions (rather that's where completions would come in anyway)

## Test plan

1. Check `doc` command works as expected
2. Check latency of `doc` command
3. Check other insertions like `generate` and `test`

closes https://github.com/sourcegraph/cody/issues/3517


philipp-spiess (approved):
Absolutely

abeatrix:
> This PR also addresses that by ensuring that all insertions end with a new line when finished.

Iirc we had this logic in the fix up controller that adds a new line at the end as well 👍

PR #3573, opened 2024-03-27
Edit: Simplify mode by removing `add`

umpox:
## Description

We weren't using mode `add` anywhere currently. We also now support changing the insertion point of the `insert` by setting `task.insertionPoint`. Removing this mode will simplify our code.

## Test plan

1. Check build passes
2. Check unit test command works correctly


philipp-spiess (approved):
Noice

PR #3576, opened 2024-03-27
Remove Cody Pro JetBrains flag

philipp-spiess:
JetBrains folks: Is the cody-pro-jetbrains flag still used by the jb client? If not we can probably clean it up in the cody repo and the dotcom backend :)

## Test plan

CI


mkondratek:
it has been removed in the jetbrains repo already: https://github.com/sourcegraph/jetbrains/pull/854

PR #3579, opened 2024-03-27
Edit: Use system prompt format

umpox:
## Description

Updates `edit` and Edit commands to use `speaker: 'system'` when available.

## Test plan

1. Make all kinds of Edits
2. Check the quality and output matches what we would expect
3. Check on models where this is not supported (e.g. GPT-4)


philipp-spiess (approved):
Sweeeet

PR #3230, opened 2024-02-21
Update the logged out status bar item with more obvious CTA

toolmantim:
Updates the Cody status bar button to have the text "Sign in to Cody" when not logged in.

[PRD](https://docs.google.com/document/d/1vYkk-LY65k2x0J1AOldbUyxfK_JQCFc3eCUbULZXx70/edit)

[Asana task](https://app.asana.com/0/1206519077296236/1206519077296305)

Before:

https://github.com/sourcegraph/cody/assets/153/a413dfd0-4ad2-4eec-9549-99a17551511e

After:

https://github.com/sourcegraph/cody/assets/153/8d2214f8-1582-4801-b73a-25b4c4cc6451

Events added:

- `CodyVSCodeExtension:signInNotification:shown`
- `CodyVSCodeExtension:signInNotification:signIn:clicked`
- `CodyVSCodeExtension:signInNotification:doNotShow:clicked`
- `CodyVSCodeExtension:statusBarIcon:clicked, { loggedIn: boolean }`

## Test plan

- CI
- Test logged out & logging in
- Test logged in

PR #3587, opened 2024-03-27
Add logo to status bar sign in CTA

toolmantim:
In #3230 we updated the status bar text to say "Sign In" to be more minimal, but forgot to add the logo given we removed the words "Cody" from the text. This adds the logo.

| Before | After |
| - | - |
| CleanShot 2024-03-28 at 09 44 41@2x | CleanShot 2024-03-28 at 09 44 08@2x |

## Test plan

- Tested locally
- CI


abeatrix (approved):

philipp-spiess (approved):
Ha thank you, I noticed it too but I was not annoyed enough to fix it 😬

Issue #3589, opened 2024-03-28
Support range for large @-mention files


abeatrix:
Large files are disabled in @-mentions but that also removes the ability to choose range for the large files. 

We should update to support that.

Also noticed a regression (compared to the old @-mention) where the range doesn't show up until you have completed typing:

![image](https://github.com/sourcegraph/cody/assets/68532117/fd7d217e-a249-45bd-ab5d-7fb93d98066c)

![image](https://github.com/sourcegraph/cody/assets/68532117/c087b554-5ca7-452b-b579-5c202ab5860e)

PriNova:
Thanks @abeatrix for raising this issue here. This is the best place for tracking it. My bad to append it on a closed PR.

abeatrix:
@PriNova no problem at all! Thanks for flagging the issue!

PR #3606, opened 2024-03-29
Chat: improve @mentions on input

abeatrix:
CLOSE: https://github.com/sourcegraph/cody/issues/3589#issuecomment-2026647366

## Current

Currently, if you add the range after inputting the file name, the title will show up as `No files found`:
image

Until you have finished the @-range:

image

This means you will have to type the whole file path before you can add the file range. Which means if you pressed space or tab on a selection you won't be able to make any changes and have to restart. 

## Changes 

This PR updates:

1. Menu displays correctly while entering the line range.
2. Autocomplete on select, and then select again to turn the selection into token (for files only. Symbol and uri is not affected).
   - This allows users to edit the file path, e.g. adding range, before it becomes a token

at

### Before

https://github.com/sourcegraph/cody/assets/68532117/9b91c43f-13ef-40b1-898c-cc00c3541f71

### After

https://github.com/sourcegraph/cody/assets/68532117/22314f05-37c9-4f43-878b-9447731f9886

### DIFF
--- a/lib/shared/src/mentions/query.ts
+++ b/lib/shared/src/mentions/query.ts
@@ -46,7 +46,7 @@ const VALID_CHARS = '[^' + TRIGGERS + PUNCTUATION + '\\s]'
 
 const MAX_LENGTH = 250
 
-const RANGE_REGEXP = '(?::\\d+-\\d+)?'
+const RANGE_REGEXP = '(?::\\d+(?:-\\d*)?)?'
 
 const AT_MENTIONS_REGEXP = new RegExp(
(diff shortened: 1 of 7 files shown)


kalanchan (approved):
nice this works!

https://github.com/sourcegraph/cody/assets/51868853/612f3778-db2b-4d33-be87-ce6b465d1389

sqs:
https://sourcegraph.slack.com/archives/C05AGQYD528/p1712006763387029?thread_ts=1712006002.459869&cid=C05AGQYD528

PR #3617, opened 2024-03-29
remove LocalEnv unused fields & LocalAppDetector

sqs:
Just code cleanup and unused code removal.

## Test plan

CI


sqs:
@sourcegraph/jetbrains-plugin Will this cause problems with the JetBrains plugin? It removes unused fields in the webview protocol.

abeatrix (approved):
thank you!

agent/src/vscode-shim.ts
```diff
@@ -55,7 +55,7 @@ import type { ClientInfo, ExtensionConfiguration } from './protocol-alias'
 // Not using CODY_TESTING because it changes the URL endpoint we send requests
 // to and we want to send requests to sourcegraph.com because we record the HTTP
 // traffic.
-const isTesting = process.env.CODY_SHIM_TESTING === 'true'
+const isTesting = typeof process !== 'undefined' && process.env.CODY_SHIM_TESTING === 'true'
```
dominiccooney, line 58: Why do we need this guard?

If this is to put Agent in web workers or VSCode Web maintenance, I'm reticent to add another dimension of complexity to this right now. For example, to fix Enterprise proxy issues with non-VSCode, we will need to lean on the node networking stack... symf and cody-engine need subprocess... etc.

dominiccooney (commented):
IDEs do depend on the webview messages in general, but I believe JetBrains will be fine with this change. One question inline.

agent/src/vscode-shim.ts
sqs: Yeah, it is to try putting agent in a web worker and VS Code Web maintenance. That's fine, I understand, I will take the `typeof process !== 'undefined'` guards out of this PR.

dominiccooney: Thank you. I am keen to support VSCode Web, and experiment with Agent in workers, once we have JetBrains on a sure footing.

dominiccooney (approved):
Looks good! Thank you for the clean up.

PR #3619, opened 2024-03-29
Chat: allow large file with range in @-mentions

abeatrix:
CLOSE: https://github.com/sourcegraph/cody/issues/3589#issuecomment-2026647366
PART OF:  https://github.com/sourcegraph/cody-issues/issues/9

### Demo

https://github.com/sourcegraph/cody/assets/68532117/8e0ee5e8-40b4-4b1a-b07b-7420af3f5183

This pull request updates the prompt editor to allow large files to be added to the context when they have a specified range. Previously, large files were not allowed to be added to the context file list. Now, large files with ranges can be included in the context.

##Changes

Update contextItemMentionNodeDisplayText in ContextItemMentionNode.ts:

- Remove the condition that prevents large files from being added to the context.
- Always display the file path and range (if available) for files.

Update Item component in OptionsList.tsx:

- Modify the condition for showing the "File too large" warning to only display when the file is large and doesn't have a range specified.

Update MentionsPlugin in atMentions.tsx:

- Differentiate between file selections and other selections.
- For file selections, replace the node with the selected file path and range (if available).
- If the selected file is large and doesn't have a range, append a `:` after the file path to allow adding line range.
- Create a mention node for the selected item and insert it into the document.

## Limitation

Users will not see the large file warning on large file with range input because we will need to get the range numbers first before we can check the token usage, which should be part of the works when we start working on improving our context window logic.

Tracking issue: https://github.com/sourcegraph/cody-issues/issues/9

## Test plan

Search for a large file via @-mentions:

 image

Press tab, verify `:` is added to the end of the file path. And then add file range:

 image

Hit tab / space again, and the large file is now added as a context item token:

image

Verify that selecting files that are not too large should automatically add to input box as token.


kalanchan (approved):
looks good! tested this on your branch and works 

edit: uploading proper video

kalanchan (dismissed):
The @ file with line numbers still got excluded in the context, did I test this correctly? 

also, it feels a bit weird to backspace after hitting tab/enter to enter in the line numbers, would it be better for large files to not have that empty space and let the user start selecting the line numbers right afterwards?

CleanShot 2024-03-29 at 14 09 51


https://github.com/sourcegraph/cody/assets/51868853/d89ca677-80bc-4bd7-b116-41609b743095

abeatrix:
> The @ file with line numbers still got excluded in the context, did I test this correctly?
> 
> also, it feels a bit weird to backspace after hitting tab/enter to enter in the line numbers, would it be better for large files to not have that empty space and let the user start selecting the line numbers right afterwards?
> 
> CleanShot 2024-03-29 at 14 09 51
> 
>  large-file-excluded.mp4



> The @ file with line numbers still got excluded in the context, did I test this correctly?
> 
> also, it feels a bit weird to backspace after hitting tab/enter to enter in the line numbers, would it be better for large files to not have that empty space and let the user start selecting the line numbers right afterwards?
> 
> CleanShot 2024-03-29 at 14 09 51
> 
>  large-file-excluded.mp4

I don't think that's related to my changes in this PR, since we can see the file was appended to the chat correctly, but got excluded by the chat context window since i suspect it's because you are using the smaller claude 3 model? Can you try Claude 3 Opus?

Mentioned in the Limitation section about more works need to be on the range for large files as well, which we can look into when we start the works on context windows?

kalanchan:
I think the current limitation is a bit confusing for the user, I don't know who to trust if I'm able to @ file a large file with range but the UI says it got excluded from context 😓 

@toolmantim @chillatom what do you all think? I think we need to address the limitations section either in the same PR or a follow up right afterwards

abeatrix:
> I don't know who to trust if I'm able to @ file a large file with range but the UI says it got excluded from context 

It was part of @toolmantim 's design and the engineering work on "We should not silently exclude a file that has been explicitly @ mentioned" 😅 We also have users asking to enable to @ line numbers again after we have disabled it on large files: https://github.com/sourcegraph/cody/pull/3523

@toolmantim  @chillatom @kalanchan I'll need more information on how we want the context window works before we can address this limitation too. I've sent you all an invite for next week so we can discuss live!

Issue for the warning title: https://github.com/sourcegraph/cody/issues/3620
I believe @chillatom has also created a tracking issue for improving the context window.

abeatrix:
@kalanchan You were right there was a bug with the display logic where used @-mention items are showing up as excluded when the file was originally too large. That's why in your screenshot Cody was still able to answer the question. 

I've fixed that so that all used files shouldn't be marked as large file:
image

Also updated to include `:1-10` when user try to include a large file to give them a head start:
image

abeatrix:
@toolmantim @chillatom updated the PR so that it matches the behavior we discussed earlier for Wednesday release: selecting a large file from the option list will add the file path to the input as text so that users can choose to add range after the auto-filled file path or remove it. Selecting files that are not large file will automatically turn it into a token. 

Follow-up from this would be allowing users to add selection from editor as @-mentions via right click menu. 

Let me know if I'm missing something here, thanks!

toolmantim (commented):
Looking good! I found a few things:

* Items shouldn't look disabled now they're clickable. CleanShot 2024-04-02 at 22 43 24@2x Can we bring back up the opacity and cursor etc?
* When you select an item, can we complete the path, add `:` to the end, make sure it's the only item in the menu, and give them some hint for how to type the range? At the moment you need to manually type the `:` and then it starts doing a line-range to all items: 
CleanShot 2024-04-02 at 22 43 07@2x
* When you add `:` can we replace the "File is too large ..." file subtext with something more helpful like "Type a line range to include, e.g. 1-42": CleanShot 2024-04-02 at 22 56 46@2x

* Let's remove the "File will be excluded if context range exceeds limit" message. Could we implement a maximum number of lines you can do, and only if they exceed it we show the message, rather than warn everyone ahead of time?
* If you type a line range (e.g. `:1-42`) and remove the lines, leaving only `:`, it seems to show an old cached value: CleanShot 2024-04-02 at 22 49 20@2x  CleanShot 2024-04-02 at 22 52 34@2x

umpox:
Agree with Tim's comments about the disabled text.

Noticed that this doesn't work with the Edit input right now, is it supposed to?

image

We can do this in a follow up but we should not show the same warning if so:
image

abeatrix:
@toolmantim @umpox Thanks for the review! I've updated the PR to address the following changes:

- [x] Items shouldn't look disabled now they're clickable. we bring back up the opacity and cursor etc?
  - Opacity updated from 0.5 to 0.8. Let me know if you think we should just remove it.
  image

- [x] When you select an item, can we complete the path, add : to the end, 
- [x] make sure it's the only item in the menu
- [x] give them some hint for how to type the range
  - Updated title to “Type a line range to include, e.g. 1-42”
- [x] When you add : can we replace the "File is too large ..." file subtext with something more helpful like "Type a line range to include, e.g. 1-42": 
  - Updated title to “Type a line range to include, e.g. 1-42”
  image
- [x] Let's remove the "File will be excluded if context range exceeds limit" message.
- [x] If you type a line range (e.g. :1-42) and remove the lines, leaving only :, it seems to show an old cached value:
- [x] Let's remove the "File will be excluded if context range exceeds limit" message
- [ ]  Could we implement a maximum number of lines you can do, and only if they exceed it we show the message, rather than warn everyone ahead of time?
  - This would be in the scope for the works we are starting in this sprint right? To find the right number?
  
> Noticed that this doesn't work with the Edit input right now, is it supposed to?
@umpox This is only updating the UI for chat, so I don't suppose the changes in this PR will work with Edit which uses a different UI. Will create an issue for Edit


### Demo

https://github.com/sourcegraph/cody/assets/68532117/8e0ee5e8-40b4-4b1a-b07b-7420af3f5183

vscode/CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
-- Chat: Large file cannot be added via @-mention. [pull/3531](https://github.com/sourcegraph/cody/pull/3531)
+- Large files cannot be added via @-mention without specifying a line range. [pull/3531](https://github.com/sourcegraph/cody/pull/3531) & [pull/3585](https://github.com/sourcegraph/cody/pull/3585)
```
sqs, line 21:
```suggestion
- Chat: Fixed issue where large files could not be added via @-mention. You can now @-mention line ranges within large files. [pull/3531](https://github.com/sourcegraph/cody/pull/3531) & [pull/3585](https://github.com/sourcegraph/cody/pull/3585)
```

vscode/src/edit/input/get-input.ts
```diff
@@ -413,7 +413,9 @@ export const getInput = async (
                         label: shortLabel || key,
                         description: shortLabel ? key : undefined,
                         detail:
-                            'isTooLarge' in item && item.isTooLarge ? FILE_TOO_LARGE_LABEL : undefined,
+                            'isTooLarge' in item && item.isTooLarge
```
sqs, line 416:
```suggestion
                            item.type === 'file && item.isTooLarge
```

vscode/webviews/promptEditor/plugins/atMentions/OptionsList.tsx
```diff
@@ -49,7 +50,9 @@ export const OptionsList: FunctionComponent<
                                 : NO_SYMBOL_MATCHES_LABEL +
                                   (mentionQuery.text.length < 3 ? NO_SYMBOL_MATCHES_HELP_LABEL : '')
                           : options.length > 0
-                              ? FILE_HELP_LABEL
+                              ? isValideLineRangeQuery(query)
```
sqs, line 53:
```suggestion
                              ? isValidLineRangeQuery(query)
```

sqs (approved):
Code LGTM with comments, have not tried this yet but I see other people have

vscode/src/edit/input/get-input.ts
```diff
@@ -413,7 +413,9 @@ export const getInput = async (
                         label: shortLabel || key,
                         description: shortLabel ? key : undefined,
                         detail:
-                            'isTooLarge' in item && item.isTooLarge ? FILE_TOO_LARGE_LABEL : undefined,
+                            item.type === 'file && item.isTooLarge
```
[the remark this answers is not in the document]
abeatrix: ty!

vscode/webviews/promptEditor/plugins/atMentions/OptionsList.module.css
```diff
@@ -72,7 +72,7 @@
     background-color: var(--vscode-list-hoverBackground);
 }
 .option-item.disabled {
-    opacity: 0.5;
```
toolmantim, line 75: Yeah, let's not change the opacity at all, as we have the label and don't want it to feel disabled. We could create a custom icon if we really wanted, but I think the label is good.

We could add a right chevron to the right hand side ">" to show that you don't be in fact selecting the item, but "going into it"

abeatrix: @toolmantim Thanks! removing opacity in https://github.com/sourcegraph/cody/pull/3665

hpn1975:
I am sorry.. but for a paid member keep getting this error is annoying... the file has only 75 lines of code.
image
I am not sure what you guys are fixing or what is fixed.  And having to enter the line number is so annoying..

PR #3621, opened 2024-03-29
E2E Test: fix flaky custom command test

abeatrix:
Small improvement for the custom command e2e tests

Recording shows that the custom command e2e test often fails because the unexpected file wasn't closed. This PR attempts to fix the test by closing the old file before it continues to the next step.

## Test plan

Green CI

### Before

image

image


kalanchan (approved):
goooo!

PR #3626, opened 2024-03-30
fix always opening EnhancedContextSettings on first chat

sqs:
Previously, it would forcibly close the popover if the component re-rendered (for some reason) on a non-first chat. I have not seen this bug actually occur in real usage.

## Test plan

CI


abeatrix (approved):
nice find, ty!

PR #3628, opened 2024-03-30
remove indirection in vscode webview React components

sqs:
- remove indirection for FileLinkComponent
- remove indirection for ChatModelDropdownMenu
- remove indirection for EditButton
- extract FeedbackButtons and remove indirection
- rm needless ChatButtonComponent (it just wrapped VSCodeButton)

## Test plan

CI

### DIFF
--- a/vscode/webviews/Chat.module.css
+++ b/vscode/webviews/Chat.module.css
@@ -141,11 +141,6 @@ body[data-vscode-theme-kind='vscode-light'] .transcript-item pre > code {
     opacity: 0.85;
 }
 
-.feedback-buttons {
-    display: flex;
-    flex-direction: row;
-    gap: 0.15rem;
-}
 
 .edit-button {
     border-radius: 0.5rem;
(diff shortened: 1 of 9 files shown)


abeatrix (commented):
Not sure if it's related to this PR, but when I build from this branch, the enhanced context window will always show up on both first chat and non-first chat:

https://github.com/sourcegraph/cody/assets/68532117/0b74d286-0c9c-4980-a1ee-2335ec756a0b

abeatrix:
It's *supposed* to show up only when a user has no chat history (fresh install).  I can reproduce on main though so it's not because of the changes in this PR. We can look into it in a follow-up!

sqs:
@abeatrix Ah, thanks for catching that. I can fix that tonight since I just broke it in #3626.

abeatrix (approved):
Everything works fine for me when testing it manually, except for the enhanced context settings behavior I mentioned above which is unrelated to the changes from this PR 
+ the CI is green 👍 
Thanks for cleaning the codebase up 🙇‍♀️

PR #3630, opened 2024-03-30
isNewInstall defaults to unknown (undefined) not true

sqs:
Previously, before the ext host sends the user's chat history, we considered `isNewInstall` to be true. Now, we consider it to be undefined (indeterminate/unknown). Fixes the issue reported by @abeatrix in https://github.com/sourcegraph/cody/pull/3628#issuecomment-2027947292 where the enhanced context settings popup always opened on all chats.

Also renames the prop to be consistent (`isNewInstall`).

## Test plan

Run locally, and ensure that the popup does not display on a non-new install. Flip the value in App.tsx manually and ensure it does show up.

### DIFF
--- a/vscode/webviews/App.tsx
+++ b/vscode/webviews/App.tsx
@@ -45,7 +45,7 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
         isCodyProUser: false,
     })
 
-    const [userHistory, setUserHistory] = useState<SerializedChatTranscript[]>([])
+    const [userHistory, setUserHistory] = useState<SerializedChatTranscript[]>()
     const [chatIDHistory, setChatIDHistory] = useState<string[]>([])
 
     const [errorMessages, setErrorMessages] = useState<string[]>([])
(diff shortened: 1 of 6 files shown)


vscode/webviews/App.tsx
```diff
@@ -192,7 +192,10 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
     )
 
     const telemetryService = useMemo(() => createWebviewTelemetryService(vscodeAPI), [vscodeAPI])
-    const isNewInstall = useMemo(() => !userHistory.some(c => c.interactions.length > 0), [userHistory])
+    const isNewInstall = useMemo(
+        () => userHistory && !userHistory.some(c => c.interactions.length > 0),
```
sqs, line 196: this is the main change

vscode/webviews/Popups/Popup.tsx
```diff
@@ -34,12 +34,11 @@ const Backdrop: React.FunctionComponent<React.PropsWithoutRef<BackdropProps>> =
 interface PopupFrameProps {
     classNames?: string[]
     actionButtons?: React.ReactNode
-    isFirstChat?: boolean
```
sqs, line 37: rm unused

vscode/webviews/App.tsx
abeatrix: Ahhhhh

abeatrix (approved):
Thanks for the fix and renaming it to be clear. 
I originally misunderstood @toolmantim and thought we want to have it opened on all first chats. I should have changed it after that, sorry about that 😆

sqs:
@abeatrix I think you implemented the same behavior in #3567  I had in mind, and I think that is also what @toolmantim had in mind. The behavior is that for the first chat *on the new install* the enhanced context settings popup should be open.

My fix in #3626, while well intentioned, broke because of an old "bug" elsewhere. It also means that your code from #3657 was (silently to the user) showing the popup every time but then immediately hiding it if it wasn't the first chat. That did not cause a user-visible bug, but it's good to fix it (except that my fix caused a worse bug :P).

PR #3640, opened 2024-03-31
factor out chat context e2e helpers

sqs:
This will make it easier to change the way that context is displayed in the chat UI (as in #3639).

## Test plan

CI


abeatrix (approved):
nice!

PR #3644, opened 2024-04-01
Semgrep: improved Semgrep scan params, upgraded GH action version

shivasurya:
This PR attempts to improve semgrep scans, (tested this config on `sourcegraph/sourcegraph`)

- Moved checkout script github action to v4 for semgrep
- Moved to v3 for uploading sarif file action script
- optimized few semgrep params to match `sourcegraph/sourcegraph` config to scan only changed files

reference: https://github.com/sourcegraph/security/issues/1286

## Test plan

- CI 🟢

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -15,13 +15,13 @@ jobs:
       image: returntocorp/semgrep
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.ref }}
           repository: ${{ github.event.pull_request.head.repo.full_name }}
 
       - name: Checkout semgrep-rules repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: sourcegraph/security-semgrep-rules
           token: ${{ secrets.GH_SEMGREP_SAST_TOKEN }}
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -30,8 +30,8 @@ jobs:
       - name: Run Semgrep SAST Scan
         run: |
           mv semgrep-rules ../
-          semgrep ci -f ../semgrep-rules/ --metrics=off --oss-only --suppress-errors --sarif -o results.sarif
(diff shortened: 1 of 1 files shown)


abeatrix (approved):

.github/workflows/semgrep.yml
```diff
@@ -15,13 +15,13 @@ jobs:
       image: returntocorp/semgrep
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
willdollman, line 18: Should we pin the hash of this action? +2 more below

.github/workflows/semgrep.yml
```diff
@@ -30,8 +30,8 @@ jobs:
       - name: Run Semgrep SAST Scan
         run: |
           mv semgrep-rules ../
-          semgrep ci -f ../semgrep-rules/ --metrics=off --oss-only --suppress-errors --sarif -o results.sarif
+          semgrep ci -f ../semgrep-rules/semgrep-rules/ --metrics=off --oss-only --suppress-errors --sarif -o results.sarif --exclude='semgrep-rules' --baseline-commit "$(git merge-base main HEAD)" || true
```
willdollman, line 33: Is the `|| true` to avoid blocking the pipeline if the scan has an intermittent failure? Will we be able to spot larger problems like if down the line something breaks and the scan stops working entirely?

.github/workflows/semgrep.yml
shivasurya: I thought github owned/based actions doesn't require hash while third party actions are riskier.

.github/workflows/semgrep.yml
shivasurya: Following the [semgrep config](https://semgrep.dev/docs/semgrep-ci/configuring-blocking-and-errors-in-ci/) while there is blocking finding the semgrep binary fails and prevents uploading sarif file which blocks the whole PR from merged. So, I opted for `|| true` similar to `sourcegraph/sourcegraph` and keep monitoring for failures.

.github/workflows/semgrep.yml
willdollman: Good point!

willdollman (approved):

Issue #3646, opened 2024-04-01
bug: Can't get passed the "Enhanced Context" popup for new chat as free user


chillatom:
### Version

Pre-release v1.11.1711984299

### Describe the bug

As a free user with no chat history, when I create a new chat the Enhanced Context selector pops up, but I cannot get passed it.  I also cannot type into the Message box at all.  Selecting or deselecting the popup doesn't work nor does trying to hit the down arrow to hide it.  

https://www.loom.com/share/7b69e6aa624c4ff696a6ef804e43534f?sid=e2686b1e-4da8-43c6-803e-78c4a6d08e4d

### Expected behavior

I should be able to dismiss the popup via the down arrow or by typing into the message box.

PR #3647, opened 2024-04-01
Chat: Fix enhance context settings bug

abeatrix:
FIX & CLOSE https://github.com/sourcegraph/cody/issues/3646

Issue: the Enhanced Context settings would remain open for all new users (user without chat history) 

Root cause: Cause by the useEffect logic where we will open the Enhanced Settings for new users on every `isOpen` value change.

Proposed fix in this PR:
- Makes sure the value for userHistory is loaded before we render the Chat component
- Start the `isOpen` value in the Chat component with the `isNewInstall` value so that we won't need to use `useEffect` to handle this logic

(also looking into why this wasn’t caught by our e2e tests, will update it in a follow-up PR)

## Test plan

Green CI. 

For Manual testing:
1. Clear your chat history
2. Open a new chat to verify the enhanced context settings is opened by default
3. Verify you can close the enhanced context settings 
4. Submit a question
5. Verify the window remains closed
6. Open a new chat
7. Verify the window doesn't show up

### DIFF
--- a/vscode/webviews/App.tsx
+++ b/vscode/webviews/App.tsx
@@ -192,12 +192,13 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
     )
 
     const telemetryService = useMemo(() => createWebviewTelemetryService(vscodeAPI), [vscodeAPI])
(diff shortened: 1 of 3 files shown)


abeatrix:
Noticed some issues with input focus stealing the focus from context settings, working on it...

PR #3651, opened 2024-04-01
Chat: remove double tab token

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1712006002459869

This reverts the change in #3606 that update  @-mentions to require double tabbing.

Will address the original issue in https://github.com/sourcegraph/cody/pull/3619

## Test plan

Green CI.


abeatrix:
agree to remove double tabbing for non-large-files. Will continue in https://github.com/sourcegraph/cody/pull/3619

PR #3654, opened 2024-04-02
Build v2 telemetry recorder for VSCode webviews

akalia25:
This PR adds a `createWebviewTelemetryRecorder`, that lets us record events on webviews using the VSCode API. It is mimicking what we are doing for v1 events through `createWebviewTelemetryService`.

We are now logging v2 events (`feature` | `action`) for:

1. `cody.thumbsup/thumbsdown` | `feedback`
2. `chatActions:editLast` | `executed`
3. `chatActions:reset` | `executed`
4. `auth` | `clickSignInWeb`
5. `auth` | `simplifiedSignInGitLabClick`
6. `auth` | `simplifiedSignInGitLabClick`
7. `auth` | `simplifiedSignInGoogleClick`

See screenshot below to verify that transcripts are only getting logged on `privateMetadata` and following the correct [guidelines](https://handbook.sourcegraph.com/departments/data-analytics/architecture/#considerations). 
image

co-authored with @bobheadxi 

## Test plan
tested locally


vscode/webviews/Chat.tsx
```diff
@@ -126,16 +129,29 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
...
+                        ?.contextFiles?.some(file => file.source === 'embeddings')
+                        ? 1
+                        : 0,
+                    recordsPrivateMetadataTranscript: userInfo.isDotComUser ? 1 : 0,
+                },
+                privateMetadata: {
+                    // 🚨 SECURITY: chat transcripts are to be included only for DotCom users AND for V2 telemetry
+                    // V2 telemetry exports privateMetadata only for DotCom users
+                    // the condition below is an aditional safegaurd measure
+                    responseText: userInfo.isDotComUser ? JSON.stringify(transcript) : '',
```
bobheadxi, line 150: Should we trim the transcript contents size like we do elsewhere?

vscode/webviews/Chat.tsx
```diff
@@ -231,6 +247,12 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
                 eventName: 'CodyVSCodeExtension:chatActions:reset:executed',
                 properties: { source: 'chat', eventType },
             })
+            postMessage?.({
+                command: 'recordEvent',
+                feature: 'chatActions:reset',
+                action: 'executed',
```
bobheadxi, line 253: 1. Feature must prefix with `cody.`
2. Verb should be the `action`, not the `feature`

```suggestion
                feature: 'cody.chatActions',
                action: 'reset',
```

vscode/webviews/Chat.tsx
```diff
@@ -231,6 +247,12 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
                 eventName: 'CodyVSCodeExtension:chatActions:reset:executed',
                 properties: { source: 'chat', eventType },
             })
+            postMessage?.({
```
bobheadxi, line 250: This should use `telemetryRecorder`, _not_ `postMessage`

vscode/webviews/Chat.tsx
```diff
@@ -322,6 +344,14 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
                         eventName: 'CodyVSCodeExtension:chatActions:editLast:executed',
                         properties: { source: 'chat', eventType: 'keyDown' },
                     })
+                    postMessage?.({
+                        command: 'recordEvent',
+                        feature: 'chatActions:editLast',
+                        action: 'executed',
```
bobheadxi, line 350: This shouldn't use `postMessage`, it should use `telemetryRecorder`, but naming needs adjusting too:

```suggestion
                        feature: 'cody.chatActions',
                        action: 'editLast',
```

lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
```diff
@@ -80,6 +80,13 @@ export class TelemetryRecorderProvider extends BaseTelemetryRecorderProvider<
  */
 export type TelemetryRecorder = typeof noOpTelemetryRecorder
 
+/** For testing. */
+export const NOOP_TELEMETRY_RECORDER_SERVICE: TelemetryRecorder = {
+    recordEvent() {
+        /* noop */
+    },
+}
```
bobheadxi, line 88: This is the same as `noOpTelemetryRecorder = new NoOpTelemetryRecorderProvider().getRecorder()` below, we can just export that

vscode/webviews/Chat.tsx
akalia25: Good call! I'm going to rename these to:
feature: `cody.chatActions`
action: `reset:executed`
just because we have some actions that are either submitted or executed :)

bobheadxi: Or `reset.executed`, per the naming scheme?

akalia25: ah yes!!

vscode/webviews/Chat.tsx
akalia25: Good catch! Using [truncateTextStart](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@b39050f6e3c5f929f9a490151b0ea221a8ca78f5/-/blob/lib/shared/src/prompt/truncation.ts?L60) now to limit the data we want to ingest

vscode/webviews/Chat.tsx
akalia25: updated, thank you!! 🙇

vscode/webviews/Chat.tsx
akalia25: updated! Thank you 🙇

vscode/webviews/Chat.tsx
```diff
@@ -126,16 +131,31 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
...
+                        ? 1
+                        : 0,
+                    recordsPrivateMetadataTranscript: userInfo.isDotComUser ? 1 : 0,
+                },
+                privateMetadata: {
+                    // 🚨 SECURITY: chat transcripts are to be included only for DotCom users AND for V2 telemetry
+                    // V2 telemetry exports privateMetadata only for DotCom users
+                    // the condition below is an aditional safegaurd measure
+                    responseText: userInfo.isDotComUser
+                        ? truncateTextStart(transcript.toString(), CHAT_INPUT_TOKEN_BUDGET)
```
akalia25, line 153: Hey @philipp-spiess, while using the `truncateTextStart` function we noticed an equivalent function called `truncatePromptText`, and discovered the `promptText` type. Should we be casting our strings as the defined `promptText` and then storing the data? Or does `promptText` only matter when we're sending the transcript to the LLM?

In other words, when **logging** transcripts, do we need to explicitly cast the transcripts to the `promptText` type?  :)

vscode/webviews/OnboardingExperiment.tsx
```diff
@@ -101,6 +113,10 @@ export const LoginSimplified: React.FunctionComponent<React.PropsWithoutRef<Logi
                                             telemetryService.log(
                                                 'CodyVSCodeExtension:auth:simplifiedSignInGitLabClick'
                                             )
+                                            telemetryRecorder.recordEvent(
+                                                'auth',
+                                                'simplifiedSignInGitLabClick'
```
bobheadxi, line 118:
```suggestion
                                            telemetryRecorder.recordEvent(
                                                'cody.webview.auth',
                                                'simplifiedSignInGitLabClick'
```

vscode/webviews/OnboardingExperiment.tsx
```diff
@@ -114,6 +130,10 @@ export const LoginSimplified: React.FunctionComponent<React.PropsWithoutRef<Logi
                                             telemetryService.log(
                                                 'CodyVSCodeExtension:auth:simplifiedSignInGoogleClick'
                                             )
+                                            telemetryRecorder.recordEvent(
+                                                'auth',
+                                                'simplifiedSignInGoogleClick'
+                                            )
```
bobheadxi, line 136:
```suggestion
                                            telemetryRecorder.recordEvent(
                                                'cody.webview.auth',
                                                'simplifiedSignInGoogleClick'
                                            )
```

bobheadxi (commented):
remember to prefix events with `cody.` to make sure they get exported on-perm :)

lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
```diff
@@ -89,7 +89,7 @@ export class NoOpTelemetryRecorderProvider extends BaseTelemetryRecorderProvider
     }
 }
 
-const noOpTelemetryRecorder = new NoOpTelemetryRecorderProvider().getRecorder()
+export const noOpTelemetryRecorder = new NoOpTelemetryRecorderProvider().getRecorder()
```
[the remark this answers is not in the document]
akalia25: added, thank you! 🙇

lib/shared/src/index.ts
```diff
@@ -215,6 +215,7 @@ export {
     MockServerTelemetryRecorderProvider,
     NoOpTelemetryRecorderProvider,
     TelemetryRecorderProvider,
+    noOpTelemetryRecorder,
```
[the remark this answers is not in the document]
akalia25: This is being used on the `vscode/webviews/OnboardingExperiment.story.tsx` file :)

vscode/src/chat/protocol.ts
```diff
@@ -13,20 +13,59 @@ import type {
...
+import type { BillingCategory, BillingProduct } from '@sourcegraph/cody-shared/src/telemetry-v2'
+
+import type { TelemetryEventParameters } from '@sourcegraph/telemetry'
+
 import type { View } from '../../webviews/NavBar'
 import type { Repo } from '../context/repo-fetcher'
 
+/**
+ * DO NOT USE DIRECTLY - ALWAYS USE a TelemetryRecorder from
+ * createWebviewTelemetryRecorder instead in webviews..
```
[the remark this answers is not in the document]
akalia25: updated, thank you for catching!

vscode/webviews/Chat.tsx
```diff
@@ -126,16 +131,31 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
...
 
-            if (userInfo.isDotComUser) {
-                eventData.transcript = JSON.stringify(transcript)
-            }
-
-            telemetryService.log(`CodyVSCodeExtension:codyFeedback:${text}`, eventData)
+            telemetryService.log(`CodyVSCodeExtension:codyFeedback:${text}`, eventData, {
+                hasV2Event: true,
+            })
+            telemetryRecorder.recordEvent(`cody.${text}`, 'feedback', {
```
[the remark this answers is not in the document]
akalia25: I updated to add a feedback enum, my only concern now would be that it is hardcoded to the string 'thumbsUp' and if we ever change the name, it could break this. 🤔 

I added the raw text to privateMetadata, so we can keep track of the raw text being submitted

vscode/webviews/OnboardingExperiment.tsx
```diff
@@ -43,6 +45,7 @@ const WebLogin: React.FunctionComponent<
                     href="about:blank"
                     onClick={event => {
                         telemetryService.log('CodyVSCodeExtension:auth:clickSignInWeb')
+                        telemetryRecorder.recordEvent('auth', 'clickSignInWeb')
```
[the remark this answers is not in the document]
akalia25: updated, thank you!

vscode/webviews/OnboardingExperiment.tsx
```diff
@@ -88,6 +96,10 @@ export const LoginSimplified: React.FunctionComponent<React.PropsWithoutRef<Logi
                                             telemetryService.log(
                                                 'CodyVSCodeExtension:auth:simplifiedSignInGitHubClick'
                                             )
+                                            telemetryRecorder.recordEvent(
+                                                'auth',
+                                                'simplifiedSignInGitLabClick'
```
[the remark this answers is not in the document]
akalia25: updated, thank you 🙇

lib/shared/src/index.ts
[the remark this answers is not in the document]
bobheadxi: Ah missed the part that this is an _export_ not an import!

vscode/webviews/Chat.tsx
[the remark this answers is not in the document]
bobheadxi: Yeah requiring hardcoded is the intent in general, as these values should be known upfront. Keeping the private metadata raw value is also good!

or, maybe we can make the callback parameter an enum as well?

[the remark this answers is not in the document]
akalia25: hmmm yeah that might get tricky, I'm happy with this, since we should be able to figure out quite quickly if text value ever changes through the `privateMetadata` raw text value :)

vscode/webviews/Chat.tsx
philipp-spiess: We only care for prompt string enforcement for data we send to the LLMs. Data we send to other SG services does not need to adhere to Cody Ignore. This is also an enterprise only feature so seeing this is gated behind a dotcom user, it's fine to convert to a string before uploading here 👍

abeatrix (approved):
The current implementation looks good to me but maybe it'd be easier to move this out of webview to have a centralized place for logging webviewMessage. 

Great work!

akalia25:
> The current implementation looks good to me but maybe it'd be easier to move this out of webview to have a centralized place for logging webviewMessage.
> 
> Great work!

Hey @abeatrix, great idea! I can work on that as a second iteration to this PR :). I'll merge this as it is, to unblock us on getting these v2 events and then we can refine the telemetry implementation. :)

vscode/webviews/Chat.tsx
akalia25: Thank you for clarification @philipp-spiess, really great to know :D

PR #3659, opened 2024-04-02
Edit: Get models from model provider

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1712074100690569?thread_ts=1712001904.028749&cid=C05AGQYD528

Follow-up on https://github.com/sourcegraph/cody/pull/3650

Tihs PR moves the model sync step from EditorManager constructor to main.ts. 

The root cause was EditorManager was initialized before the AuthStatus was created, and the existing `editorManager.syncAuthStatus()` was added to the AuthStatus Listeners after the AuthStatus has been created, that's why the model list was shown as empty. 

To fix this, call the `editorManager.syncAuthStatus()` after adding it to the listeners to make sure it syncs with the initAuthStatus.

## Test plan

- connect to s2
- run doc command
- check if claude 3 is being used or claude 2


vscode/src/main.ts
```diff
@@ -316,6 +316,7 @@ const register = async (
     const initAuthStatus = authProvider.getAuthStatus()
     syncModelProviders(initAuthStatus)
     await chatManager.syncAuthStatus(initAuthStatus)
+    editorManager.syncAuthStatus(initAuthStatus)
```
abeatrix, line 319: this is the fix

chwarwick (approved):

PR #3662, opened 2024-04-02
Hover Commands: enable and disable ghost text and hover on feature flags

abeatrix:
This PR enables the hover commands feature based on a feature flag evaluation. 

The main changes are:
- Get the ghost hint enablement config in HoverCommandsProvider to check if EditOrChat or Document are enabled
- Set the isInTreatment flag in HoverCommandsProvider based on:
  - The hover commands feature flag
  - EditOrChat and Document being disabled in the ghost hint config (control by config settings and feature flag)

This allows us to run an A/B test on the hover commands by enabling it for some users if they have the feature flag and EditOrChat/Document disabled.

The ghost hint config is used to disable EditOrChat and Document as those features conflict with the hover commands. So we want to enable hover commands only if those are off.

This unblocks us from running the planned hover commands A/B test by simply enabling the feature flag for Hover Commands to 100% when we are ready.

## Test plan

Enable and disable feature flags on your account to verify.

### DIFF
--- a/vscode/src/commands/GhostHintDecorator.ts
+++ b/vscode/src/commands/GhostHintDecorator.ts
@@ -98,17 +98,16 @@ export async function getGhostHintEnablement(): Promise<EnabledFeatures> {
     const config = vscode.workspace.getConfiguration('cody')
     const configSettings = config.inspect<boolean>('commandHints.enabled')
     const settingValue = configSettings?.workspaceValue ?? configSettings?.globalValue
+    const featureFlagValue = await featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyHoverCommands)
 
     // Return the actual configuration setting, if set. Otherwise return the default value from the feature flag.
     return {
         /**
-         * We're not running an A/B test on the "Opt+K to Text".
-         * We can safely set the default of this to `true`.
+         * We want to use the featureFlagValue to toggle settings for EditOrChat & Document for the
+         * HoverCommands A/B test, which will only be enabled if EditOrChat & Document are disabled.
          */
-        EditOrChat: settingValue ?? true,
-        Document:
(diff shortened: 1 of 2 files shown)


vscode/src/commands/GhostHintDecorator.ts
```diff
@@ -94,21 +96,26 @@ function getSymbolDecorationPadding(
...
         /**
-         * We're not running an A/B test on the "Opt+K to Text".
-         * We can safely set the default of this to `true`.
+         * Toggle the default settings for EditOrChat & Document based on the feature flagss for hover commands and ghost text.
          */
-        EditOrChat: settingValue ?? true,
-        Document:
-            settingValue ??
-            (await featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyDocumentHints)),
+        EditOrChat: settingValue ?? (ghostTextFeatureFlag && !hoverFeatureFlag),
```
abeatrix, line 117: does this look ok to you? @umpox

vscode/src/commands/GhostHintDecorator.ts
```diff
@@ -94,21 +96,26 @@ function getSymbolDecorationPadding(
...
+        Document: settingValue ?? (ghostTextFeatureFlag && !hoverFeatureFlag),
```
umpox, line 118:
```suggestion
        EditOrChat: settingValue ?? !hoverFeatureFlag,
        Document: settingValue ?? !hoverFeatureFlag,
```

I think is what we want. That way 50% of users will get the hover experiment and the other 50% will get the ghost text

vscode/src/commands/GhostHintDecorator.ts
umpox: Left a comment!

PR #3669, opened 2024-04-03
pass secretStorage instead of using global singleton

sqs:
`secretStorage` is not intended by the VS Code extension API to be a global. It is passed to the `activate` function. I believe this is because VS Code's extension API designers felt it was important to be explicit about which parts of your code have access to the secrets, to avoid mistakes.

## Test plan

CI


valerybugakov (commented):
Context: this is one of the changes I introduced a while ago to reduce the "prop-drilling" of some class instances (e.g., `localStorage` too) through the VS Code extension logic layers. I think we overuse this pattern and feel that it slows down feature development when we need to access a particular instance in one of the deeply nested functions. 

> I think it's better to be explicit about the use of secret storage and secret data flow.

Since this is a security-related instance, it might be worth being more explicit about it. I'm curious if you have any scenarios in mind where having it as a globally accessible import might cause issues.

PR #3685, opened 2024-04-03
E2E: Fix helper function for file opening

abeatrix:
This commit update the `openFile` helper function in the VSCode extension E2E tests to fix the incorrect logic and make it more robust and cross-platform compatible. 

This is to fix the local-embeddings e2e test that has been flaky on Windows because the `openFile` helper function which only works on Windows

Update to use the correct shortkey for opening a file from the file picker:
- Currently using `Control+P+Shift` which is to open the command palette
  - Control shouldn't work on Mac as well, so the tests were actually not opening the file before this change
- Updated to use `MetaKey+P` correctly. 
- Add check to ensure the extension was loaded before we try opening a file to prevent other windows from stealing focus
- Add checks to make sure the file is opened before we move onto the next step

## Test plan

E2E test passing.
Green CI

### Before

You can see after pressing `Control+P+Shift`, the file picker wasn't invoked on Mac:

image

### After

The file picker shows up after the `MetaKey+P` step:

image

### DIFF
--- a/vscode/test/e2e/command-custom.test.ts
+++ b/vscode/test/e2e/command-custom.test.ts
@@ -35,8 +35,14 @@ test.extend<ExpectedEvents>({
         'CodyVSCodeExtension:chat-question:executed',
     ],
 })('create a new user command via the custom commands menu', async ({ page, sidebar }) => {
-    // Sign into Cody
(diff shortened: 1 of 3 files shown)


kalanchan (approved):
goooo

PR #3691, opened 2024-04-04
Edit: Track rejection rate

umpox:
## Description

This PR adds an rejection metric that will listen for the next action in a document immediately after an Edit.

It works like this:
- We track how often the user hits "Undo" or "Redo" immediately after an Edit.
- When a user makes a new insertion to a document, we consider them to have moved on from the Edit.
- If we observed that the user undid the Edit, then we mark it as "rejected", otherwise we mark it as "accepted"
- If the user clicks the "Undo" command from the Edit codelens, we consider this to be a rejection, and immediately mark it as such.

The intention here is to provide a secondary metric to our persistence rate. We cannot track acceptance rate, as edits are automatically inserted to the document, so we essentially need to do the inverse, and check if the user rejects the change or not. There is a chance that this is not 100% accurate, as a user may manually modify the edit output if they are not happy with it. However this could be seen as both a partial acceptance or partial rejection, so the persistence rate is likely more useful here.

## Test plan

Undo:

1. Make an edit
2. Undo it
3. Make another change
4. Check event fired as `fixup:user:rejected`

Undo/Redo

1. Make an Edit
2. Undo it
3. Redo to apply the change again
4. Make another change
5. Check event fired as `fixup:user:accepted`

Undo via Code lens

1. Make an Edit
2. Undo it
3. Redo to apply the change again
4. Make another change
5. Check event fired as `fixup:user:rejected`

Happy path (any other change):
1. Make an edit
2. Make another change
3. Check event fired as `fixup:user:accepted`


vscode/src/non-stop/FixupController.ts
```diff
@@ -586,6 +646,7 @@ export class FixupController
 
             // TODO: See if we can discard a FixupFile now.
             this.setTaskState(task, CodyTaskState.applied)
+            this.logTaskCompletion(task, document, editOk)
```
umpox, line 649: Previous we called `logTaskCompletion` _before_ the output was formatted.

We now correctly call it once we have fully finished inserting the edit.

vscode/src/non-stop/FixupController.ts
```diff
@@ -445,6 +452,7 @@ export class FixupController
             intent: task.intent,
             mode: task.mode,
             source: task.source,
+            model: task.model,
```
umpox, line 455: Including `model` in v1 metadata too, previously did this as part of v2 only (in case the model string was private for enterprise) but this doesn't match with how we handle this elsewhere in telemetry, and it limits us comparing rejection rate across models

philipp-spiess (approved):
Nice! I like the solution, curious to look at the results 👀

PR #3695, opened 2024-04-04
[DONOTCOMMIT] Add very basic proxy support for agent

pkukielka:
## Test plan

1. Install [squidman](https://squidman.net/squidman/) - you can leave all default options. It will give you a local proxy server.
2. Set `http_proxy` and `https_proxy` variables:

```
export https_proxy=http://127.0.0.1:8080
export https_proxy=http://127.0.0.1:8080
```

Both point to the same http squid endpoint which handles both https and https, that is not a mistake.

3. Use that cody version with either [jetbrains plugin](https://github.com/sourcegraph/jetbrains/blob/main/CONTRIBUTING.md) or with [agent tests](https://github.com/sourcegraph/cody/blob/main/agent/src/index.test.ts).
If everything is working properly either should cause new entries appearing in squid access log:
/Users/USERNAME/Library/Logs/squid/squid-access.log
----

**TBD**
1. Make sure my changes does not break normal VSC, and if they does add branched logic for agent
2. Test with better proxy (e.g. with user login/pass and with ports 80 and  443 completely blocked [1])
3. add some automatic tests with proxy in the CI
4. ???
----

[1] I was trying to do that using firewall but I failed, pf was not respecting my `pass` squid rule:

```
# 1. Install squidman: brew install --cask squidman
# 2. Setup squidman to work on 8080 HTTP port
# 3. Apply this firewall config: sudo pfctl -e -f proxy_firewall.conf
# 4. To disable it: sudo pfctl -d

# Set the interface to be used
if="en0"

# Default Deny Policy
pass out quick inet proto tcp from any to any port { 8080, 3128 }
block out quick inet proto tcp from any to any port { 80, 443 }

# Skip the loop back interface
# set skip on lo

# Set http proxy (8080) port
# web_ports = "{8080}"
```

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -440,9 +440,9 @@ export class Agent extends MessageHandler implements ExtensionClient {
             let closest = ''
             if (polly) {
                 // @ts-ignore
-                const persister = polly?.persister?._cache as Map<string, Har> || {}
+                const persister = (polly?.persister._cache as Map<string, Har>) || {}
                 for (const [, har] of persister) {
-                    for (const entry of (await har).log.entries) {
+                    for (const entry of har.log.entries) {
                         if (entry.request.url !== url) {
                             continue
                         }
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -439,7 +439,8 @@ export class Agent extends MessageHandler implements ExtensionClient {
             let closestDistance = Number.MAX_VALUE
             let closest = ''
             if (polly) {
-                const persister = polly.persister._cache as Map<string, Promise<Har>>
+                // @ts-ignore
+                const persister = polly?.persister?._cache as Map<string, Har> || {}
                 for (const [, har] of persister) {
                     for (const entry of (await har).log.entries) {
                         if (entry.request.url !== url) {
--- a/vscode/src/fetch.node.ts
+++ b/vscode/src/fetch.node.ts
@@ -1,5 +1,4 @@
 import http from 'node:http'
-import https from 'node:https'
 
 import { SocksProxyAgent } from 'socks-proxy-agent'
 
@@ -7,6 +6,9 @@ import type { Configuration } from '@sourcegraph/cody-shared'
 
 import { agent } from '@sourcegraph/cody-shared/src/fetch'
 import { getConfiguration } from './configuration'
+
+import { HttpProxyAgent } from 'http-proxy-agent';
+import {HttpsProxyAgent} from "https-proxy-agent";
 
 // The path to the exported class can be found in the npm contents
 // https://www.npmjs.com/package/@vscode/proxy-agent?activeTab=code
@@ -17,8 +19,6 @@ const pacProxyAgent = 'PacProxyAgent'
 /**
  * We use keepAlive agents here to avoid excessive SSL/TLS handshakes for autocomplete requests.
  */
-let httpAgent: http.Agent
-let httpsAgent: https.Agent
 let socksProxyAgent: SocksProxyAgent
 
 function getCustomAgent({ proxy }: Configuration): ({ protocol }: Pick<URL, 'protocol'>) => http.Agent {
@@ -30,10 +30,15 @@ function getCustomAgent({ proxy }: Configuration): ({ protocol }: Pick<URL, 'pro
             })
             return socksProxyAgent
         }
+
+        // For some reason adding 'keep-alive' in the constructor was causing proxy to not be picked up.
+
+        // http_proxy/https_proxy in the format of 'https://[username:password]@your-proxy.com'
+        // (see :https://github.com/TooTallNate/proxy-agents/issues/12)
         if (protocol === 'http:') {
-            return httpAgent
+            return new HttpProxyAgent(process.env.http_proxy ?? "")
         }
-        return httpsAgent
+        return new HttpsProxyAgent(process.env.https_proxy ?? "")
     }
 }
 
@@ -45,12 +50,6 @@ export function setCustomAgent(
 }
 
 export function initializeNetworkAgent(): void {
-    httpAgent = new http.Agent({ keepAlive: true, keepAliveMsecs: 60000 })
-    httpsAgent = new https.Agent({
-        ...https.globalAgent.options,
-        keepAlive: true,
-        keepAliveMsecs: 60000,
-    })
     const customAgent = setCustomAgent(getConfiguration())
 
     /**
@@ -67,7 +66,7 @@ export function initializeNetworkAgent(): void {
      */
     try {
         const PacProxyAgent =
-            (globalThis as any)?.[nodeModules]?.[proxyAgentPath]?.[pacProxyAgent] ?? undefined
+            (globalThis as any)?.[nodeModules]?.[proxyAgentPath]?.[pacProxyAgent] ?? customAgent
         if (PacProxyAgent) {
             const originalConnect = PacProxyAgent.prototype.connect
             // Patches the implementation defined here:
--- a/vscode/src/fetch.node.ts
+++ b/vscode/src/fetch.node.ts
@@ -1,4 +1,4 @@
-import http from 'node:http'
+import type http from 'node:http'
 
 import { SocksProxyAgent } from 'socks-proxy-agent'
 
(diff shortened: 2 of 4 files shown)


aramaraju:
Basic user authentication can be tested by following instructions [here](https://wiki.squid-cache.org/Features/Authentication)

vscode/src/fetch.node.ts
```diff
@@ -30,10 +30,15 @@ function getCustomAgent({ proxy }: Configuration): ({ protocol }: Pick<URL, 'pro
...
+        // For some reason adding 'keep-alive' in the constructor was causing proxy to not be picked up.
+
+        // http_proxy/https_proxy in the format of 'https://[username:password]@your-proxy.com'
+        // (see :https://github.com/TooTallNate/proxy-agents/issues/12)
         if (protocol === 'http:') {
-            return httpAgent
+            return new HttpProxyAgent(process.env.http_proxy ?? "")
         }
-        return httpsAgent
+        return new HttpsProxyAgent(process.env.https_proxy ?? "")
```
chwarwick, line 41: Won't this now create a new https.Agent instead of reusing the configured agent that should already have a connection from the keepalive?

vscode/src/fetch.node.ts
```diff
@@ -30,10 +30,15 @@ function getCustomAgent({ proxy }: Configuration): ({ protocol }: Pick<URL, 'pro
             })
             return socksProxyAgent
         }
+
+        // For some reason adding 'keep-alive' in the constructor was causing proxy to not be picked up.
```
chwarwick, line 34: I believe the keepalive was added as large a performance improvement based on the number of requests that are generated by autocomplete

chwarwick (commented):
Just a warning that this fetch is not exclusively used, there is some usage of native fetch, axios, and https.request so this may only cover some requests.

vscode/src/fetch.node.ts
dominiccooney: Yes, we should continue advertising keep-alive, at least for VSCode non-proxy users in the short term.

> adding 'keep-alive' in the constructor ... to not be picked up

Would be great to be specific... which constructor? In what way was the proxy not picked up? ...wondering if you were watching connections or requests.

@philipp-spiess FYI

vscode/src/fetch.node.ts
dominiccooney: &uarr; This

We need to use the same agent so something is hanging onto the connection to reuse.

@philipp-spiess had to work around VSCode breaking keep-alive by reusing this object, see:

ae2b292940b2e1f2347047a064dacd6f9facf40b
https://github.com/microsoft/vscode/issues/173861

@pkukielka for now, let's simply early return the now object, *only* when `http_proxy` or `https_proxy` are set, this like we do for SOCKS proxies above, and add a TODO or file an issue that proxy users may have slower autocomplete because there's no keep-alive. (How hard could it be for a proxy to support that... but never underestimate how broken proxies can be.)

agent/src/agent.ts
```diff
@@ -439,7 +439,8 @@ export class Agent extends MessageHandler implements ExtensionClient {
             let closestDistance = Number.MAX_VALUE
             let closest = ''
             if (polly) {
-                const persister = polly.persister._cache as Map<string, Promise<Har>>
+                // @ts-ignore
```
dominiccooney, line 442: Why did we need to do this?

dominiccooney (changes requested):

vscode/src/fetch.node.ts
pkukielka: Yes, we should check if we are in the agent or VSC and have different logic based on that.

agent/src/agent.ts
pkukielka: This is interesting, I'm not sure how my other changes could cause problems there, but without changing this line I get the errors:

```
agent/src/agent.ts:442:35 - error TS18047: 'polly.persister' is possibly 'null'.

442                 const persister = polly.persister._cache as Map<string, Promise<Har>>
                                      ~~~~~~~~~~~~~~~

agent/src/agent.ts:442:35 - error TS2352: Conversion of type 'Map<string, Har>' to type 'Map<string, Promise<Har>>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'Har' is missing the following properties from type 'Promise<Har>': then, catch, finally, [Symbol.toStringTag]

442                 const persister = polly.persister._cache as Map<string, Promise<Har>>
                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

agent/src/agent.ts:442:51 - error TS2341: Property '_cache' is private and only accessible within class 'Persister<TOptions>'.

442                 const persister = polly.persister._cache as Map<string, Promise<Har>>
```

vscode/src/fetch.node.ts
pkukielka: Sorry, what I meant is this:
`new HttpProxyAgent(process.env.http_proxy, { keepAlive: true, keepAliveMsecs: 60000 })`
If I add `keepAlive` in the constructor `opts` param everything builds and starts normally, but in the `squid-access.log` I see no requests are forwarded to a proxy.
On which level of the stack this is failing I do not know, I did not had time yet to debug it as this PR was time-boxed limited WIP.

agent/src/agent.ts
dominiccooney: Sigh, looks like we're hacking into polly implementation details to add something missing (enumerate entries/approximate matching) and those implementation details have changed, and because of our hackery we are now giving things the wrong type, etc.

So let's fix the types first... `_cache` has type `Map<string, Har>`, not type `Map<string, Promise<Har>>`. Since you're apparently using the correct type, we should get rid of `as Map<string, Har>` *and* the `(await har)` &rArr; `har` on line 445.

Second, `_cache` is private. You *do* need the `@ts-ignore` to suppress this, but add a comment about what we're doing here.

Third, `polly.persister` may be `null`. I guess the `|| {}` is fine but this also smells like something has changed in polly and we may be broken.

CC @steveyegge , because he was coincidentally looking at polly not recording traces—was that in agent?—maybe some of this rot is related.

vscode/src/fetch.node.ts
```diff
@@ -1,5 +1,4 @@
 import http from 'node:http'
```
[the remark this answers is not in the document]
dominiccooney: This apparently still broken, so `import type http from 'node:http'`

I would fix it for you but GitHub doesn't allow editing non-diff lines 🤷

vscode/src/fetch.node.ts
```diff
@@ -8,6 +7,9 @@ import type { Configuration } from '@sourcegraph/cody-shared'
 import { agent } from '@sourcegraph/cody-shared/src/fetch'
 import { getConfiguration } from './configuration'
 
+import { HttpProxyAgent } from 'http-proxy-agent';
+import {HttpsProxyAgent} from "https-proxy-agent";
```
dominiccooney: Apparently the formatter didn't fix this 🤷 

```suggestion
import { HttpsProxyAgent } from "https-proxy-agent";
```

vscode/src/fetch.node.ts
dominiccooney: OK let's add a TODO or DONOTCOMMIT here so we know not to land this as-is.

vscode/src/fetch.node.ts
dominiccooney: > I think we rather should check if we are in the agent or VSC and have different logic based on that. VSC already have proper agent there so no point to tinker with it.

Agree with this part.

> Ideally we should probably hack node module loading and replace agent in http and https modules - I think that is what vscode proxy code is doing.

Disagree with this part. VSCode is probably doing that because they don't control a lot of extension code and their random dependencies.

We *do* control this code, so let's just use the existing hooks for adding proxies.

dominiccooney (changes requested):
More feedback inline.

agent/src/agent.ts
pkukielka: @dominiccooney FYI I'm not actively looking on this PR as agreed with @kalanchan and @taylorsperry .
I think @chwarwick and Ara will be taking over.

PR #3717, opened 2024-04-05
pass chatModels and onCurrentChatModelChange through React context

sqs:
This makes it so more of the exthost<->webview RPC code lives in App.tsx, which is fitting. It also simplifies the props we need to pass down.

We did the same thing for enhanced context a couple months ago.

## Test plan

CI


philipp-spiess (approved):

abeatrix (approved):
This would be very helpful for what I'm working on as well. 

Since @sqs is OOO, will try to merge it tomorrow if this looks good to you @philipp-spiess?

sqs:
I am coding on fun stuff and was keeping this rebased for my own purposes anyway. Going to merge. :)

PR #3722, opened 2024-04-07
Agent-side changes to support JetBrains integration tests

steveyegge:
Most of the changes were on the JetBrains/Kotlin side; there isn't much here:

- added a new file scheme, `temp://`, for jetbrains in-memory integration tests that use a temp filesystem
- added support for an integration test access token passed via env var
- stub for a Mock LLM provider which does nothing yet

Other changes:

- I added an extra VS Code run configuration for connecting to a remote Agent, for debugging.

## Test plan

This is part of the test plan for previous commits, as it helps introduce JetBrains integration testing. It complements PR https://github.com/sourcegraph/jetbrains/pull/1291


vscode/src/edit/mock-provider.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class MockEditProvider extends EditProvider {
+    // TODO: Where do find the script file to use, or define the scripts?
+    constructor(public config: EditProviderOptions) {
+        super(config)
+    }
+
+    public async startEdit(): Promise<void> {
+        this.config.controller.startTask(this.config.task)
+        // Here we elided all the machinery around talking to a real LLM.
```
pkukielka, line 12: Why do we need that?
Isn't the whole idea to use polly so we can record real interactions instead of mocking everything manually?

vscode/src/edit/mock-provider.ts
```diff
@@ -0,0 +1,23 @@
+import type { EditProviderOptions } from './provider'
+import { EditProvider } from './provider'
+
+export class MockEditProvider extends EditProvider {
+    // TODO: Where do find the script file to use, or define the scripts?
```
dominiccooney, line 5: I don't grok the TODO.

node has `__dirname` and `__filename` to get the directory/file name of the thing executing, if that helps.

vscode/src/edit/mock-provider.ts
dominiccooney: @pkukielka, my $0.02, we want both kinds of tests.

polly.js is useful because it records real backend responses, and will detect frontend/backend drift if the recordings are updated.

Mocked responses are useful because it is impractical to create some test conditions with a real/replayed backend.

vscode/src/edit/provider.ts
```diff
@@ -128,7 +127,6 @@ export class EditProvider {
                     }
                     case 'error': {
                         let err = message.error
-                        logError('EditProvider:onError', err.message)
```
dominiccooney, line 131: Why remove this?

vscode/src/services/SecretStorageProvider.ts
```diff
@@ -6,6 +6,10 @@ const CODY_ACCESS_TOKEN_SECRET = 'cody.access-token'
 
 export async function getAccessToken(): Promise<string | null> {
     try {
+        const testToken = process.env.CODY_INTEGRATION_TEST_TOKEN
+        if (testToken) {
+            return testToken
```
dominiccooney, line 11: Suggest logging here. Behavior can be quite mysterious if you accidentally leave an environment variable like this floating around in your shell and then launch VSCode and forget about it, etc.

dominiccooney (commented):
Seems OK in as far as it goes, but why mock so much of the edit provider/where's the rets of the mocking for it? (Haven't looked at the JB side change yet. Will now.)

vscode/src/edit/mock-provider.ts
umpox: Agree, it'd be useful just to record the LLM responses instead of skip them if we can

vscode/src/edit/provider.ts
umpox, line 131: Seems useful for debugging still?

agent/src/cli/jsonrpc.ts
```diff
@@ -144,6 +144,17 @@ export const jsonrpcCommand = new Command('jsonrpc')
             //     res.sendStatus(400)
             // })
             polly.server.get('https://objects.githubusercontent.com/*path').passthrough()
+
+            // TODO: I'm not yet sure why this is required, but Polly refuses to record
```
dominiccooney, line 148: Is this because of shutdown? grep for calls to shutdownAndExit with 30_000 msec timeouts and history of that. It may be a buffering/file synchronization issue.

dominiccooney (approved):
Looks good, a suggestion to clean up the forced persisting inline.

steveyegge:
This stuff wound up getting merged on a different branch.

PR #3727, opened 2024-04-08
Edit: Accurately track chars and lines added based on the diff

umpox:
## Description

closes: https://github.com/sourcegraph/cody/issues/3409

This PR:
- Tracks characters inserted by Cody using the diff, instead of the entire replacement text
  - This is important because the diff is only useful changes coming from Cody, whereas the replacement contains code that Cody didn't change
- Tracks lines inserted by Cody using the diff, instead of the entire replacement text

## Test plan

1. Make edits
2. Check lines and characters added is correct

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -153,7 +153,7 @@ export class FixupController
             intent: task.intent,
             mode: task.mode,
             source: task.source,
-            ...countCode(replacementText),
+            ...this.countEditInsertions(task),
             ...task.telemetryMetadata,
         }
         const { metadata, privateMetadata } = splitSafeMetadata(legacyMetadata)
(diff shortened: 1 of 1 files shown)


vscode/src/non-stop/FixupController.ts
```diff
@@ -415,12 +415,37 @@ export class FixupController
...
+        }
+
+        if (!task.diff) {
+            return { lineCount: 0, charCount: 0 }
+        }
+
+        const countedLines = new Set<number>()
+        let charCount = 0
+        for (const edit of task.diff.edits) {
+            charCount += edit.text.length
```
umpox, line 434: Note, in the case of a deletion, the text would be "", so this will still handle only the number of chars that Cody is inserting. We don't measure the number of chars Cody deleted yet (unsure if we do this elsewhere?)

PR #3731, opened 2024-04-08
Persistence Tracker: Treat removed events as `difference: 100`

umpox:
## Description

This PR updates the persistence tracker to log `persistence:present` events with a difference of `100` when an insertion is completely removed.

This should make our analytics more accurate

## Test plan

1. Make a completion or "insert" edit
2. Delete the inserted text
3. Check the correct telemetry event is logged after 30 secs

Confirmed:
```
logEvent (telemetry disabled): CodyVSCodeExtension:completion:persistence:present  {
  "details": "VSCode",
  "properties": {
    "id": "b0e0030a-4a6f-482d-bd20-7ec3f595f046",
    "difference": 100
  },
  "opts": {
    "agent": true,
    "hasV2Event": true
  }
}
```

### DIFF
--- a/vscode/src/common/persistence-tracker/types.ts
+++ b/vscode/src/common/persistence-tracker/types.ts
@@ -19,7 +19,7 @@ export interface PersistenceRemovedEventPayload<T = string> {
     /** An ID to uniquely identify an accepted insertion. */
     id: T
     /** Levenshtein distance between the current document state and the accepted completion */
-    difference: 100
+    difference: 1
     /** Attached metadata to the insertion */
     metadata?: PersistenceEventMetadata
 }
--- a/vscode/src/common/persistence-tracker/types.ts
+++ b/vscode/src/common/persistence-tracker/types.ts
@@ -18,6 +18,8 @@ export interface PersistencePresentEventPayload<T = string> {
 export interface PersistenceRemovedEventPayload<T = string> {
     /** An ID to uniquely identify an accepted insertion. */
     id: T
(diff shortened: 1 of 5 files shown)


vscode/src/completions/logger.ts
```diff
@@ -241,7 +241,7 @@ export function logCompletionPersistenceRemovedEvent(params: PersistenceRemovedE
     // Use automatic splitting for now - make this manual as needed
     const { metadata, privateMetadata } = splitSafeMetadata(params)
     writeCompletionEvent(
-        'persistence:removed',
+        'persistence:present',
```
umpox, line 244: This should probably just be `persistence`, but I've preserved the `present` so this doesn't conflict with any existing telemetry data

vscode/src/common/persistence-tracker/types.ts
```diff
@@ -18,6 +18,8 @@ export interface PersistencePresentEventPayload<T = string> {
 export interface PersistenceRemovedEventPayload<T = string> {
     /** An ID to uniquely identify an accepted insertion. */
     id: T
+    /** Levenshtein distance between the current document state and the accepted completion */
+    difference: 100
```
philipp-spiess, line 22: @umpox @kelsey-brown should this be `1` instead of `100`? 🤔

philipp-spiess (approved):

vscode/src/common/persistence-tracker/types.ts
umpox: Oh yeah I think it should, good catch!

PR #3713, opened 2024-04-05
Chat: add ability to send editor selection to chat as @-mention

abeatrix:
Follow https://github.com/sourcegraph/cody/pull/3619
Part of https://github.com/sourcegraph/cody/issues/3664

image

This pull request introduces a new feature that allows users to easily share their current editor selection with Cody during a chat conversation. It turns the selection into an @-mention token in their chat inbox.

The following changes have been made:

Add 'cody.chat.context.add' Command:

- A new command 'cody.chat.context.add' has been added to the package.json file, which is registered in the ChatManager class.
- When triggered, this command calls the sendSelectionToChat method in the ChatManager, which retrieves the chat provider and invokes the handleGetUserEditorContext method.

Implement 'sendSelectionToChat' Method:

- The sendSelectionToChat method has been added to the ChatManager class.
- This method retrieves the chat provider and calls the handleGetUserEditorContext method, which is responsible for sending the user's editor selection to the chat.

Update SimpleChatPanelProvider:

- The SimpleChatPanelProvider has been updated to import the ContextItemFile type from the shared codebase context messages.

Update Package.json:

- The package.json file has been updated with the new 'cody.chat.context.add' command configuration, including its category, group, title, icon, and when condition.
- The command has also been added to the context menu and the editor menu group.
- With this feature, users can easily share relevant code snippets or files with the AI assistant during a conversation, providing additional context and improving the quality of the assistant's responses.

## Test plan

Highlight some code and right click and select `Cody Chat: add context`

image

The chat panel should be opened for you with the selection added as an @-mention token

image

If you already have a chat panel opened, it will add it to the end of your input box

## Demo

https://github.com/sourcegraph/cody/assets/68532117/ab2eccba-60a3-463a-aae8-a1c0669c9821

PR #3732, opened 2024-04-08
Chat: "Ask Cody" command on hover

abeatrix:
Follow https://github.com/sourcegraph/cody/pull/3713 

Note: This should be treated separately and should not block 3713

Address the following:
@PriNova ([source](https://github.com/sourcegraph/cody/pull/3713#issuecomment-2041033364)) Would it be possible to integrate it additionally into the mouse hover popup too?
@umpox ([source](https://github.com/sourcegraph/cody/pull/3713#issuecomment-2042866117)) This means we can [implement the "Add to Chat" ghost text design too.](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=1161-33280&mode=design&t=GNofSTDZrFNVqenh-0)

- Add a new 'Ask Cody' command to the context menu when text is selected in the editor
- Disable the 'New Chat' command when text is selected
- Add a 'when' condition to only show the 'Ask Cody' context menu item when text is selected and editor is not read-only
- Rename the 'ask' hover command to 'error' and enable it only when there is an error

### Shortcut

On Alt+L when there is a selection in your editor:

altL

### Hover Commands Experiment

- When hovering on a selection, `Ask Cody` will show up instead of `New Chat`.
image

- Clicking on `Ask Cody` will open a new chat with the selection added as `@-mention`:
image


PriNova:
Awesome!!

I would reconsider the steps for both the hover and Cody commands:

1. If a Chat window is already open and text is selected:
  - 'Ask Cody' will add or append the selected text to the ongoing chat conversation.
  - 'New Chat' will open a new Chat window with the selected text inserted.

2. If no chat window is open and text is selected:
  - 'Ask Chat' will start a new chat with the selected text inserted.

Rationale for 1: The option to open a 'New Chat' (on hover too) should be considered because the context window in an ongoing conversation could be full. Having the ability to open a 'new chat' window provides the user with a fresh conversation instead of hiding the command or hover.

abeatrix:
> * 'Ask Cody' will add or append the selected text to the ongoing chat conversation.
> * 'New Chat' will open a new Chat window with the selected text inserted.

Thanks for the feedback and ideas @PriNova   
This is actually the current behavior for the new `Ask Cody`. If there is an existing chat panel opened, we will append the @-mentions for the selected text to the chat panel. If there is none, then a new chat panel will be opened with the @-mentions added 😄

PriNova:
@abeatrix But if there is a chat window already open, then 'New Chat' with selected text should also be possible.

```JavaScript
if Chat_Is_Open and Text_Is_Selected then
    New_Chat | Append_Text

if not Chat_Is_Open and Text_Is_Selected then
    New_Chat
```

vscode/package.json
```diff
@@ -546,7 +554,12 @@
       {
         "command": "cody.chat.panel.new",
         "key": "alt+l",
-        "when": "cody.activated"
+        "when": "cody.activated && !editorHasSelection"
+      },
+      {
+        "command": "cody.chat.context.add",
+        "key": "alt+l",
+        "when": "cody.activated && !editorReadonly && editorTextFocus && editorHasSelection"
```
umpox, line 562:
```suggestion
      {
        "command": "cody.chat.context.add",
        "key": "alt+l",
        "when": "cody.activated && editorTextFocus && editorHasSelection"
```

Shouldn't matter if this is `editorReadonly` for chat, right?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -561,11 +563,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
 
     private handleAbort(): void {
         this.cancelInProgressCompletion()
-        telemetryService.log(
-            'CodyVSCodeExtension:abortButton:clicked',
-            { source: 'sidebar' },
-            { hasV2Event: true }
-        )
+        telemetryService.log('CodyVSCodeExtension:abortButton:clicked', { hasV2Event: true })
```
umpox, line 566:
```suggestion
        telemetryService.log('CodyVSCodeExtension:abortButton:clicked', {}, { hasV2Event: true })
```

or `hasV2Event` gets logged as a property

umpox (approved):
Nice!!

vscode/package.json
abeatrix: it does because we will need to read the text from the editor but not all the readOnly editor are local files. For example, using a remote file created by the Sourcegraph extension, the command doesn't work if we remove `!editorReadonly &&`:

https://github.com/sourcegraph/cody/assets/68532117/a613c04a-c256-495d-b3db-0ec39b14512d

Issue #3642, opened 2024-04-01
[Cody Ignore] Fix risks arising from file based .cody/ignore


dominiccooney:
There's a file-based .cody/ignore feature, there's an remote configured Cody Ignore feature planned for Enterprise, we need to make these coherent. This records clean-up while reviewing the VSCode, extension code:

☢️ risk 👷 integrating new requirements w/ existing codegm 💪 defensive, mitigation 🧹 clean up

```[tasklist]
### Tasks
- [ ] ☢️ getContextFileFromDirectory, callers searches current directory without consulting cody ignore; passed thru edit user context files
- [ ] ☢️ getWorkspaceFilesContext searches workspace without consulting Cody Ignore
- [ ] ☢️ vscode/src/edit/prompt/context.ts getContext passes thru "user" context without checking Cody Ignore
- [ ] ☢️ vscode/src/editor/vscode-editor.ts getTextEditorContextForFile does not check Cody Ignore
- [ ] 👷 Expand ignore concept to repos, eg vscode/src/chat/ContextProvider.ts updateCodebaseContext should filter ignored repos
- [ ] 💪 Security review
- [ ] 💪 Audit uses of VSCode document getText, etc.
- [ ] 💪 vscode/src/prompt/index.ts buildInteraction uses fixup file URI without checking ignore, strengthen this
- [x] 💪 Consider carrying file, editor content substrings in objects, not strings; use template literals for concatenation preserving taint.
- [ ] 🧹 ~vscode/src/editor/active-editor.ts caches lastActiveTextEditor for no purpose~ (the cache is effective, via `getEditor`)
- [x] 🧹 `check/isCodyIgnoredFile` is unused (at least in JetBrains) in Agent protocol, so remove it; API as designed is weird (takes a list of files, returns a single boolean) and wasteful (checks all of the files to return `true` even if the first file is `true`)
- [ ] 🧹 There are "chat based" and "new" edit, unit test commands. Minimize risk by deprecating, removing unused commands.
- [ ] What's the plan for context? Do we require user prompts to be promptstring?
- [ ] PromptString: Revisit the type for storing references. https://github.com/sourcegraph/cody/pull/3734#discussion_r1562182819
- [ ] PromptString: Handle serialization of chat transcriptions with a resolved reference.
- [ ] PromptString: Can we guarantee that filters are honored? https://github.com/sourcegraph/cody/pull/3856#issuecomment-2069272652
```

abeatrix:
[...]

PR #3734, opened 2024-04-08
Add PromptString

philipp-spiess:
Part of #3642

We're introducing a new concept of `PromptString` into the Cody code base. The goal of this is to create a safe way for us to concatenate prompts _while still knowing what context was referenced along the way_. This will give us peace of mind when adding features like Cody Ignore since we can guarantee that the prompts we send to the backend LLMs do not accidentally contain blocked files.

## What changes for me?

When working on prompts in Cody, you can no longer concatenate strings and call it a day. Instead, you will have to use our PromptString utilities. We tried to make this experience as smooth as possible and added a lot of helper methods. 

For example:

```ts
// Old and unsafe:
const prompt = `Please rewrite this code: ${document.getText(range)}`

// New and will retain a reference to the document.uri that was interpolated:
const prompt = ps`Please rewrite this code: ${PromptString.fromDocumentText(document, range)}`
````

You can create `PromptString`s from other PromptStrings, literals, and use string helper functions just like you did before. Here’s an example from the newly added unit tests:

```ts
const s = ps`  Foo${ps`bar`}baz  `
expect(s.toString()).toBe('  Foobarbaz  ')
expect(s.length).toBe(13)
expect(s.slice(1, 3).toString()).toBe(' F')
expect(s.trim().toString()).toBe('Foobarbaz')
expect(s.trimEnd().toString()).toBe('  Foobarbaz')
expect(s.indexOf('Foo')).toBe(2)
expect(s.indexOf(ps`Foo`)).toBe(2)
expect(s.toLocaleLowerCase().toString()).toBe('  foobarbaz  ')
expect(s.includes('Foo')).toBe(true)
```

A few of the helper functions are prefixed with and considered  `unsafe_` since we can't fully track the references in all cases (e.g. when a user copy-pasts something into the text box). @dominiccooney and me are tracking usages of these APIs and will work on reducing the unsafe surface (e.g. we want to make it safe to serialize/deserialize prompt string).

Let us know if you have any issues with this when working in your code.

## Follow-up
- [ ] Revisit the type for storing references. [Should we use a tree of frozen arrays? ](https://github.com/sourcegraph/cody/pull/3734#discussion_r1562182819)
- [ ] Handle serialization of chat transcriptions with a _resolved_ reference.

## Test plan

- [x] Test Chat/Edit/Autocomplete
- [x] CI passes
- [x] New unit tests for prompt string
- [x] TypeScript passes


lib/shared/src/ollama/index.ts
```diff
@@ -1,4 +1,5 @@
 import type { OllamaGenerateParameters } from '../configuration'
+import { PromptString } from '../prompt/prompt-string'
```
dominiccooney, line 2:
```suggestion
import type { PromptString } from '../prompt/prompt-string'
```

lib/shared/src/prompt/prompt-string.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    it('correctly assembles', () => {
+        expect(ps`one`.toString()).toBe('one')
+        expect(ps`one ${ps`two`} three ${ps`four ${ps`five`} six`} seven`.toString()).toBe(
+            'one two three four five six seven'
+        )
+    })
+
+    it('keeps track of references', () => {
+        const uri = testFileUri('/foo/bar.ts')
+        const inner = createPromptString('i am from a file', new Set([uri]))
```
dominiccooney, line 33: Let's not export createPromptString, the contract is too loose. We should have a helper for each use case instead, like

```
PromptString.fromDocument(doc: Document): PromptString
PromptString.fromDocumentFilename(doc: Document): PromptString
PromptString.fromUriPath(uri: URI): PromptString // gets uri.path, etc.
// any other helpers we need
```

These helpers should either be:
1. Safe (Document.text is safe because we're holding onto the document URI; maybe we want to treat virtual documents as ignored or whatever but we can work that out later...)
2. Named "unsafe" and something which will trigger extra human review steps. Cases where we are really taking a string and asserting "trust me." Hopefully we don't have many of these.

We should lock down the Document text accessor functions in the linter or with prototype hacking, so you can't just drive around the barriers:

TextLine.text
Document.getText
DiagnosticCollection.forEach, this should have a helper which filters
etc.

...guessing the edit event's text properties are OK to leave as strings.

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+// This setup is designed to detect programming errors, not thwart malicious
+// attacks.
+
+/**
+ * A "safe" string class for constructing prompts.
+ */
+export class PromptString {
+    private constructor() {}
+    public static unsafe_createInstance() {
+        return new PromptString()
```
dominiccooney, line 28: This is safe (if useless) because we're protecting access to the side pocket map.

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+    pocket.set(handle, new PromptStringPocket(string, references))
+    return handle
+}
+
+export function promptStringToString(s: PromptString): string {
+    return pocket.get(s)!.value
+}
+
+export function promptStringReferences(s: PromptString): Set<StringReference> {
+    return pocket.get(s)!.references
```
dominiccooney, line 89: Never return mutable objects from the side pocket. Return a defensive copy, immutable view, or frozen object.

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+
+        return makePromptString(buffer.join(stringBoundary), references)
+    }
+}
+
+type StringReference = vscode.Uri
+
+// This helper function is unsafe and should only be used within this module.
+function makePromptString(string: string, references: Set<StringReference>): PromptString {
+    const handle = PromptString.unsafe_createInstance()
```
dominiccooney, line 79: Think it is fine to leave the constructor visible. It is useless without the side pocket.

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const buffer: string[] = []
+    const references: Set<StringReference> = new Set()
+    for (let i = 0; i < format.length; i++) {
+        buffer.push(format[i])
+        if (i < args.length) {
+            const arg = args[i]
+
+            if (arg === undefined) {
+                // Ignore undefined values
```
dominiccooney, line 117: The more we make this like JavaScript, the easier time we will have. So this should be `undefined`.

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+type StringReference = vscode.Uri
+
+// This helper function is unsafe and should only be used within this module.
+function makePromptString(string: string, references: Set<StringReference>): PromptString {
+    const handle = PromptString.unsafe_createInstance()
+    pocket.set(handle, new PromptStringPocket(string, references))
+    return handle
+}
+
+export function promptStringToString(s: PromptString): string {
```
dominiccooney, line 84: Let's just have callers use `toString`?

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+                    'Use ps`...` or a PromptString helper to handle string data for prompts safely.'
+                )
+            }
+        }
+    }
+
+    return makePromptString(buffer.join(''), references)
+}
+
+export function createPromptString(value: string, references: Set<StringReference>): PromptString {
```
dominiccooney, line 137: Let's avoid exporting anything this loose as long as possible. When we do:

- It should be named "unsafe" something
- It should have build/review integration so each call is getting flagged and reviewed

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,152 @@
...
+// When PromptStrings are created, their properties are stored in a side pocket
+// WeakMap. Consumers can do what they like with the PromptString, all of the
+// operations use data in the map and so are protected from the PromptString
+// constructor being disclosed, prototype pollution, property manipulation, etc.
+const pocket = new WeakMap<PromptString, PromptStringPocket>()
+
+class PromptStringPocket {
+    constructor(
+        public value: string,
+        public references: Set<StringReference>
```
dominiccooney, line 150: Let's freeze the set as we take it, or take a defensive copy and freeze that. I assume URI is immutable, for practical purposes?

Later on we may want some mutable state in the side pocket, not against that, but at this stage we definitely can't take a mutable object and pass out references to it.

lints/safe-prompts.ts
```diff
@@ -0,0 +1,75 @@
+// Checks that prompts track the source of context from the editor through to
```
dominiccooney, line 1: This could do with a bunch of TODOs, improvements as you see fit.

dominiccooney (changes requested):
Feedback inline.

In general, while we're not guarding against untrusted JavaScript, let's be WAAAY more defensive with this. I want someone who uses this type to fall into the pit of success, not find an easy route around the protections like retrieving the reference set and clearing it out, etc.

lib/shared/src/prompt/prompt-string.ts
philipp-spiess: Yeah but I would rather have a typescript error pointing me to this file where I can add docs than have people add `new PromptString` and not realizing it doesn't work.

lib/shared/src/prompt/prompt-string.ts
philipp-spiess: I don't understand, can you elaborate this comment? 
We have code in place where we conditionally spread context in prompt strings and I wanted to add support for it:

```ts
const foo = `foo${a == 1 ? 1 : undefined}bar`
```

lib/shared/src/prompt/prompt-string.ts
philipp-spiess: Maybe I can do that with a @deprecated notice on the constructor

lib/shared/src/prompt/prompt-string.ts
philipp-spiess: Yeah agree, to be fair this was already here 😅

lib/shared/src/prompt/prompt-string.ts
philipp-spiess: I see what you mean now! 

```
`${undefined}`
'undefined'
```

will fix it and the caller that was depending on it 👍

.github/workflows/unsafe-function-alert.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - uses: actions/github-script@v6
+        if: steps.check-unsafe-usage.outputs.unsafe_function_added == 'true'
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: '‼️ Hey @philipp-spiess, please review this PR carefully as it introduces usage of \`unsafe_\` functions.'
```
dominiccooney, line 30: Can we add a role for this and tag that role?

lib/shared/src/prompt/prompt-string.test.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        const evil = 'ha-ha!'
+        // @ts-expect-error: Evil is a string and can not be appended like this
+        expect(() => ps`foo${evil}bar`).toThrowError()
+
+        const evil2 = {
+            toString: () => 'hehe!',
+        }
+        // @ts-expect-error: Can't hack around the limitation
+        expect(() => ps`foo${evil2}bar`).toThrowError()
```
dominiccooney, line 22: Worth passing a `PromptString` which has a monkey-patched `toString`. Fine if that doesn't throw, but it should not invoke `toString`.

We should also test this:

```
let x = new Number(8)
x.toString = () => 'hey'
ps`${x}` // should be '8' or definitely not 'hey'
```

lib/shared/src/prompt/prompt-string.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        const second = temporary_createPromptString('bar', [uri2])
+        const third = temporary_createPromptString('baz', [uri3])
+
+        const concatenated = first.concat(second, third)
+
+        expect(concatenated.toString()).toBe('foobarbaz')
+        expect(concatenated.getReferences()).toEqual([uri1, uri2, uri3])
+    })
+
+    it('detects invalid PromptStrings passed to utility types', () => {
```
dominiccooney, line 112: 🎉

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,407 @@
...
+        // information from the URI, so we leave the references empty for now (the
+        // content is always just whitespace)
+        return internal_createPromptString(indentString, [])
+    }
+
+    public static fromDisplayPath(uri: vscode.Uri) {
+        return internal_createPromptString(displayPath(uri), [uri])
+    }
+
+    public static fromDocumentText(document: vscode.TextDocument, range?: vscode.Range): PromptString {
```
dominiccooney, line 160: 🎉

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,407 @@
...
+    }
+}
+
+// Validate that an input is indeed a PromptString and not just typecast to it.
+export function isValidPromptString(promptString: PromptString) {
+    if (pocket.has(promptString)) {
+        return true
+    }
+    return false
+}
```
dominiccooney, line 252: Brevity

```suggestion
    return pocket.has(promptString)
}
```

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -0,0 +1,407 @@
...
+
+type StringReference = vscode.Uri
+
+// This helper function is unsafe and should only be used within this module.
+function internal_createPromptString(
+    string: string,
+    references: readonly StringReference[]
```
dominiccooney, line 259: Should we use Sets here, and return defensive copies?

- `readonly` is a TypeScript-ism that is easy to `as` away.
- I expect we'll end up with multiple references to the same file, and Set de-dups.
- Taking and returning defensive copies is very robust... if we remember to do it.

dominiccooney (commented):
This is amazing @philipp-spiess , you're killing it.

.github/workflows/unsafe-function-alert.yml
philipp-spiess: Yeah I wanted to get it working first before I annoy folks :D I also want to use your script so we can add the rule for e.g. not calling `ps()` too first.

vscode/src/completions/text-processing/utils.ts
```diff
@@ -121,16 +122,16 @@ export function getHeadAndTail(s: string): PrefixComponents {
...
     }
 
     return headAndTail
 }
 
-function trimSpace(s: string): TrimmedString {
+function trimSpace(s: PromptString): TrimmedString {
     const trimmed = s.trim()
-    const headEnd = s.indexOf(trimmed)
+    const headEnd = s.toString().indexOf(trimmed.toString())
```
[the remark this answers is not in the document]
philipp-spiess: yessir

lib/shared/src/prompt/prompt-string.test.ts
philipp-spiess:
```ts
 class FakePromptString extends PromptString {
    toString() {
        return '😈'
    }
}
```

he-he

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -313,46 +346,55 @@ export function ps(
...
-interface ConfigGetter<T> {
-    get<T>(section: string, defaultValue?: T): T
+function internal_createPromptString(
+    string: string,
+    references: readonly StringReference[]
+): PromptString {
+    const handle = new PromptString()
+    // Create a shallow copy of the references list as a set, so it's both de-duped
+    // and can not be mutated by the caller
+    pocket.set(handle, new PromptStringPocket(string, new Set(references)))
```
dominiccooney, line 378: Tough one. For security, the frozen array you had was great. For de-duping, Set is nice. And we don't need the perfect representation in v0, as long as we don't leak anything mutable, we can always change it.

There's a bunch of places where we have PrompString &rarr; PromptString with the same references, so it would be nice to have an immutable object and avoid allocating a second one.

Seems like we don't need to check what's in the array much. So we could also construct a tree of them, where references are frozen arrays of `(URI | FrozenReferenceArray)[]`.

(Separately, if we used frozen arrays, think we could use types and a marker property to ensure we're freezing them all... something to think about but not critical.)

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -313,46 +346,55 @@ export function ps(
...
-
 function internal_toString(s: PromptString): string {
     return pocket.get(s)!.value
 }
 function internal_toReferences(s: PromptString): readonly StringReference[] {
-    const ref = pocket.get(s)!.references
-    Object.freeze(ref)
-    return ref
+    // Return a shallow copy of the references so it can not be mutated
+    // TODO: Do we need to create an array or can we expose the iterator?
```
dominiccooney, line 386: Returning the iterator sounds fine. This is the spec: https://tc39.es/ecma262/multipage/keyed-collections.html#sec-createsetiterator it is a view of the mutable underlying object, but it doesn't permit mutation. (I *assume* URI is locked down well?)

lints/safe-prompts.ts
```diff
@@ -36,18 +39,31 @@ export function delint(sourceFile: ts.SourceFile) {
...
                 ) {
                     report(node, 'Use `ps` only as a tagged template literal')
+                    break
+                }
+
+                if (
+                    text.startsWith('unsafe_') &&
+                    node.parent?.kind !== ts.SyntaxKind.FunctionDeclaration
+                ) {
+                    report(node, `New \`${text}\` invocation found`)
```
dominiccooney, line 64: Should we disallow aliasing these, by checking that the parent node is a function declaration and the node is the CallExpression.expression (and not CallExpression.arguments)? Or is that too strict?

(Would need to look into module imports as well. Another place you can alias, and maybe we want to allow `x.unsafe_m()` where x is a module.)

In theory, if we review all the aliases, then we're fine...

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1241,14 +1246,29 @@ function newChatModelFromSerializedChatTranscript(
...
+                    : null,
+            ].filter(isDefined)
         ),
         json.id,
         json.chatTitle,
         json.enhancedContext?.selectedRepos
     )
 }
 
+// TODO: Should we also serialize the references?
```
dominiccooney, line 1262: We slam chats in and out of storage quite a bit, so it seems weird to have references on a chat transcript, but closing and opening the chat panel or restarting the IDE makes those references evaporate.

On the other hand, you might pull up an old chat transcript and all the file paths are different now because you moved a folder or something. I'm thinking maybe we should have a concept of "resolved" references where we keep (repo, repo relative path) pairs for the serialized representation or something.

Let's leave them OUT for now, because changing the serialized schema is a pain. Think we want something here but let's work it out in a follow-up to avoid thrashing the disk format.

@willdollman FYI. Background: When you chat, we store chat history on disk and reload it later, you can keep chatting.

vscode/src/context/remote-search.ts
```diff
@@ -108,8 +109,10 @@ export class RemoteSearch implements ContextStatusProvider, IRemoteSearch {
         return new Set([...this.reposAuto.keys(), ...this.reposManual.keys()])
     }
 
-    public async query(query: string): Promise<ContextSearchResult[]> {
-        const result = await graphqlClient.contextSearch(this.getRepoIdSet(), query)
+    public async query(query: PromptString): Promise<ContextSearchResult[]> {
+        // TODO: Validate that prompt string has only allowed references before
```
dominiccooney, line 113: ❤️  it is so cool that we are now in a place to make checks like this a reality.

This is an interesting case though... that's Sourcegraph, not an LLM provider. So in theory it's cool to serialize anything up there? @aramaraju ?

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -236,32 +274,30 @@ export class PromptString {
...
         return internal_createPromptString(string, [])
     }
-}
 
-// Validate that an input is indeed a PromptString and not just typecast to it.
-export function isValidPromptString(promptString: PromptString) {
-    if (pocket.has(promptString)) {
-        return true
+    // 🚨 Use this function only for LLM responses queries.
+    public static unsafe_fromLLMResponse(string: string): PromptString {
```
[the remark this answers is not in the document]
willdollman: The file in question will have already been sent to the LLM, so it does feel like an edge case and I think it's ok if we ignore it.

Equally, with the current system it seems like it would be possible to attach references to prompts and their responses and this might be generally useful to track?

agent/src/index.test.ts
```diff
@@ -388,7 +388,7 @@ describe('Agent', () => {
             // TODO: make this test return a TypeScript implementation of
             // `animal.ts`. It currently doesn't do this because the workspace root
             // is not a git directory and symf reports some git-related error.
-            expect(trimEndOfLine(lastMessage?.text ?? '')).toMatchInlineSnapshot(
+            expect(trimEndOfLine(lastMessage?.text?.toString() ?? '')).toMatchInlineSnapshot(
```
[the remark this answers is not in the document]
philipp-spiess: Will try that, ty!

agent/src/index.test.ts
```diff
@@ -473,11 +473,15 @@ describe('Agent', () => {
                     )
                     {
                         const lastMessage = await client.sendMessage(id, 'kramer')
-                        expect(lastMessage?.text?.toLocaleLowerCase().includes('quone')).toBeFalsy()
+                        expect(
+                            lastMessage?.text?.toString()?.toLocaleLowerCase().includes('quone')
```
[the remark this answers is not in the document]
philipp-spiess: Ohh yes you're right, TIL :)

.github/workflows/safe-prompts-lint.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - uses: actions/github-script@v6
+        # if: steps.check-unsafe-usage.outputs.unsafe_function_added == 'true'
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: '‼️ Hey @philipp-spiess, please review this PR carefully as it introduces usage of \`unsafe_\` functions.'
```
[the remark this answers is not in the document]
philipp-spiess: https://sourcegraph.slack.com/archives/C01CSS3TC75/p1712930238694689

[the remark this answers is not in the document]
philipp-spiess: Maybe we can also use https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message here :)

agent/src/index.test.ts
```diff
@@ -221,7 +221,7 @@ describe('Agent', () => {
             const lastMessage = await client.sendSingleMessageToNewChat(
                 'Generate simple hello world function in java!'
             )
-            const trimmedMessage = trimEndOfLine(lastMessage?.text ?? '')
+            const trimmedMessage = trimEndOfLine(lastMessage?.text?.toString() ?? '')
```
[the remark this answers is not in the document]
philipp-spiess: I tried this here but since we have to branch of the implementation (as `Array.prototype.join` wouldn't work for `PromptString`), TypeScript seemingly can't follow the branches correctly and I get something like:

```
Type 'string' is not assignable to type 'T'.
  'string' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'string | PromptString'.ts(2322)
```

.github/workflows/safe-prompts-lint.yml
```diff
@@ -0,0 +1,51 @@
...
+      #   run: |
+      #     # Use git diff to check for added lines containing "unsafe_" function calls
+      #     if git diff --diff-filter=A --unified=0 HEAD^ HEAD | grep -q '\+.*unsafe_'; then
+      #       echo "unsafe_function_added=true" >> $GITHUB_OUTPUT
+      #     else
+      #       echo "unsafe_function_added=false" >> $GITHUB_OUTPUT
+      #     fi
+
+      - name: Run lints
+        run: pnpm ts-node lints/safe-prompts.ts `git diff --name-only main | grep -E "\.(ts|tsx)$"`
```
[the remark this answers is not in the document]
philipp-spiess: This runs in under 2sec on my machine (and this is a quite sizeable diff). Might not need parallel right now but I'll add an inline comment so we know how to improve it.

philipp-spiess:
testing @sourcegraph/cody-security

philipp-spiess:
@dominiccooney @willdollman it might get a bit spammy for you guys now, I promise I will keep it as minimal as possible 🥺

lib/shared/src/prompt/prompt-string.ts
philipp-spiess: @dominiccooney Yeah I’m not sure about URI especially since we know in the vs code runtime the URI type is different from the one in the vscode-uri package. Maybe we should serialize it as a string?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
philipp-spiess: Yep, agree. Note that this is not only used for storage, but also when communicating with the agent (the agent will query the transcript to get updates quite a bit), so agree that we should track this eventually.

I will move this into the `prompt-string` class and expose this as unsafe for now too. I did some investigation and for _new messages_ we do not deserialze from a chat message in the agent (rather we use the `chat/submitMessage` notification) but we might change the implementation at one point and forget about this, so I'd rather have this marked unsafe.

philipp-spiess:
@pkukielka  There are some renames in the kotlin typegen. Can you check if this is an issue of the JetBrains client? The actual types are kept the same (so the API will still expose prompts as strings when talking to the agent).

philipp-spiess:
@dominiccooney This is _almost_ done. I need to make the CI pass again and then document the follow-up tasks in your GitHub issue and "arm" the CI rule again (I temporarily disabled it). I also want to remove some of the TODOs from the `prompt-string` class as well (i.e moving the types). So I'll likely land this Monday my morning if that's alright.

lib/shared/src/prompt/prompt-string.ts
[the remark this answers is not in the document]
aramaraju: Just adding to Will's comment, wouldn't the content already be sent in that case? I think it's safe to say that the new configuration only gets applied on a new chat session (I have [called](https://www.figma.com/file/m9Yvgv6W3aPRLfXnxgwnEz?type=design&node-id=1198-4702&mode=design#768413463) it out to Daniel on the UX side for his thoughts on this too)

Is there a benefit in the 'wash off' behavior here? Is there any other reason to track the references? I suspect that we will some reason like evaluating how good the context fetching -  to determine if we did find the right files or compare against different context sources, etc?

agent/recordings/enterpriseClient_3965582033/recording.har.yaml
```diff
@@ -1506,14 +1506,10 @@ log:
               - speaker: assistant
                 text: I am Cody, an AI coding assistant from Sourcegraph.
               - speaker: human
-                text: >+
-                  Codebase context from file src/example.test.ts in repository
-                  undefined:
```
philipp-spiess, line 1511: We had a bug here before because of no errors when doing `${undefined}`. Glad this is now checked 😎

vscode/src/context/remote-search.ts
aramaraju: Making sure I interpret this correctly, is this case where the content is sent to the SG instance? That should be fine (for the truly security conscious we provide self-hosted experience)

lints/safe-prompts.ts
philipp-spiess: > In theory, if we review all the aliases, then we're fine...

Yeah that was my thinking too, because if we do `import { unsafe_m as m } ..` we would still get this to error in the first case (I tested this specifically).

vscode/src/edit/manager.ts
```diff
@@ -112,7 +112,7 @@ export class EditManager implements vscode.Disposable {
         }
 
         let task: FixupTask | null
-        if (configuration.instruction?.trim()) {
+        if (configuration.instruction && configuration.instruction.trim().length > 0) {
```
philipp-spiess, line 115: @dominiccooney I found a failure case that can bite us here. The conditions for these might happen at places that were not even edited by my change because we were relying on `if(string)` for empty string checks all over the place and now `PromptString` is coerced to `true`. Luckily we had an E2E test in this case but I wonder if you have ideas of how we can protect against more issues like this?

One idea I have is to add https://palantir.github.io/tslint/rules/strict-boolean-expressions/ on `main` and fix all issues.

philipp-spiess:
Screenshot 2024-04-15 at 15 01 56
 
 Let's goooo

vscode/src/edit/manager.ts
philipp-spiess: Screenshot 2024-04-15 at 15 46 08

:sigh:

philipp-spiess: Good news: I found a few more potential candidates for this issue with this approach
Bad news: since this was the type result of a function call it was _not_ found using this approach. 😬 

I want to this was still worth the time though :) [`456ab07` (#3734)](https://github.com/sourcegraph/cody/pull/3734/commits/456ab072b2cacb8691ae2797879580378a487435)

dominiccooney: Oh how did I miss this, good find. It's too bad there's no @@toPrimitive, valueOf hook for Boolean coercion.

So one thing that comes to mind, we could make the objects non-empty. So there's

`type PromptString = '' | NonEmptyPromptString` (or some better names)

The prompt string creation helper would check if the result is empty, and return '' if so. This assumes the non-empty string is never sensitive, which might be a bit dodgy.

This would force us to stick closer to the interface for String in the convenience methods on NonEmptyPromptString, because accessing anything specific to NonEmptyPromptString is less convenient now.

Another thing we could do is leave PromptString including '', but use a linter to warn if you're using a type which includes PromptString in a Boolean-looking context (`!`, `&&' etc., `if (...)`.) But that would be a more expensive lint, because you would need types and not just syntax like we do now.

philipp-spiess: Both are great ideas. I’m tempted to not go forward with any of these though and instead just deal with any potential fallout this has. The upside is that longer-term we won't have to maintain a different code path for empty strings and don't need the expensive lint rules.

My mental model is that any new code paths that will be added will be tested and issues like this won't be added anymore. If that's incorrect, we may indeed need to fix this.

 @dominiccooney Do you think this is a reasonable assumption? I’m happy to add more stuff to handle this but I am inclined to just ship this and deal with any potential issues.

vscode/src/chat/chat-view/prompt.test.ts
```diff
@@ -12,7 +12,7 @@ describe('DefaultPrompter', () => {
 
     it('constructs a prompt with no context', async () => {
         const chat = new SimpleChatModel('a-model-id')
-        chat.addHumanMessage({ text: 'Hello' })
+        chat.addHumanMessage({ text: ps`Hello` })
```
valerybugakov, line 15: I love the coincidence that it's your initials @philipp-spiess 😀

philipp-spiess: @valerybugakov I worked on https://pspdfkit.com/ in the past and made this comment all the time :P

umpox:
@philipp-spiess Noticed a weird regression from this commit:


https://github.com/sourcegraph/cody/assets/9516420/9ff34037-55b2-4923-bcf5-aa4293465b8f


Seems like the "New Chat" command is automatically executing a chat

philipp-spiess:
@umpox thanks will fix asap, I’m sure it's a string coercion bug :(

philipp-spiess:
@umpox https://github.com/sourcegraph/cody/pull/3811

PR #3742, opened 2024-04-09
Chat: refactor context window + new FeatureFlag for User context

abeatrix:
PART OF https://github.com/sourcegraph/cody/issues/3745
CLOSE https://github.com/sourcegraph/cody/issues/3743 & https://github.com/sourcegraph/cody/issues/3744 & https://github.com/sourcegraph/cody/issues/3806

This PR focused on the following:
- refactor the current context window used by ModelProvider
- Introduce a new TokenCounter class that handles token-counting during prompt building and context building
  - Import `js-tiktoken` library
  - Replace the current approach that use characters in content and messages as token count
- Introduce a new Context Window counting mechanic where User-Context has its own token budget instead of both chat input and context all share the same token limit

Docs on the new expected behavior on how context window should work: https://docs.google.com/document/d/1_rg8j4yfz1Ot-lQle8v_K_y8V3JDzpKYxO134X7MItc/edit

https://github.com/sourcegraph/cody/assets/68532117/8e5f885a-8a8d-44fc-929c-b2f2dac98a4f

### Changes Summary

- Add `js-tiktoken` library for counting tokens
- New Context Window mode behind a new feature flag `cody-chat-context-budget` https://github.com/sourcegraph/cody/pull/3781
  - When enabled, it raised limits for DotCom users with feature flag enabled:
    -  **User Context: new 30k limit**
    -  **Chat Input: increased from 7k to 15k**
- Replace old `maxInputToken` and `maxOutputToken` consts with a new ContextWindow object
- Add constants for user context token budget and enhanced context allocation
- Add new class TokenCounter for counting the token within the text string in ChatMessages
- Move isTooLarge flag from ContextItemFile to ContextItemCommon
- Remove getContextWindowLimitInBytes function
- Remove all functions that convert tokens into characters and vice versa
  - This should now be handled by TokenCounter to get token count instead of bytes and characters
- Added a new token budget for user context (@-mentions context) 
- The new token budget is currently set to `10,000` in `USER_CONTEXT_TOKEN_BUDGET`
- If we decide to update the token budget for user context in the future, we can just update `USER_CONTEXT_TOKEN_BUDGET`
- Token limit for User chat messages and enhanced context remains unchanged in this PR
- Updated existing tests to use the new TokenCounter for prompt building
- Add new unit tests for the new TokenCounter class
- All editor and user context include size and isTooLarge at creation time. This allows us to calculate the token limit in different places including the front end. 
- No major UI change
- cody ignored context items are filtered during the prompt building steps to ensure context from ignored files will not be sent to LLM as context.

#### Update

- replace char-based token counting with the tokenizer
  - The previous approach of counting tokens based on a fixed character-to-token ratio (4 chars per token) has been replaced with a more accurate tokenizer-based approach. This change improves the precision of token counting, which is crucial for respecting token limits and preventing truncation of important content
- Use Token counter instead of character count, removed CHARS_PER_TOKEN which was using 4 characters as tokens but should be 3.5 ([source](https://docs.anthropic.com/claude/docs/glossary#tokens))
- ContextItem size field represents token count instead of bytes

### Acceptance Criteria Checklist

- [ ] We have the Telemetry to know the sizes of files users are really tagging and the rate at which they hit the new limit
  - Will be continue in follow-up 
- [x] Any past @ things will be excluded from context if they can't fit into the budget, evicting the oldest first
- [x] Users can mention larger files, even if it means the interaction exceeds the old context window limit, without an error
- [x] There's a limit that is applied per-file. When users @-tag files exceeding this limit they see the typical large file error messages.
- [x] The @-tagged files budget works across follow-up chats: Follow up chats work, the limits are applied across the total @-tagged files.
- [ ] Make token counts more accurate

## Follow-up

Works to do after this PR is merged:

- [x] Migrate all current usages of `charsToTokens` and `tokensToChars` from https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/prompt/constants.ts to the `tokensToBytes` and `bytesToTokens` and in the token/constants.ts file. (Doing this in this PR will cause a lot of additional changes that'd be harder to follow.)
- [ ] Add Telemetry to include file size when the new limit is hit
- [ ] UI change once @toolmantim has finalized the behavior for "design the product behavior when you @-tag files in follow-up chats, what's the policy for dropping old files? and elaborate this task list"

## Test plan

Updated existing tests, and added new unit test for the new TokenCounter class.

Manual testing steps:
2. In Debug mode, open the sourcegraph/cody repository
3. Search for the `SimpleChatPanelProvider` file, previously this file will be listed as large file. Verify that it's no longer shown as large file in this branch
4. Now try to add CHANGELOG.md. it should show you that it's a large file and selecting it will ask you to enter a range:

image

5. Entering a range that exceeds the token limit will turn the @-mention into red:

image

6. If you hit enter now, you should see the invalid @-mentions is excluded from Context:

image

7. In the chat input, @-mentions CHANGELOG.md again, but this time enter a range that doesn't exceed the token limit. E.g. 1-100 

image

8. Submit it and you can see it was used as context:

image


vscode/src/chat/context/chatContext.ts
```diff
@@ -20,8 +22,7 @@ export async function getChatContextItemsForMention(
         empty: () => void
         withType: (type: MentionQuery['type']) => void
     },
-    // The number of characters left in current context window.
-    maxChars?: number
```
abeatrix, line 24: removing maxChars here since we will use the standardized context windows for user-context instead of needing to pass the current characters usage from the chat model to this function

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -25,8 +26,6 @@ export interface IPrompter {
     makePrompt(chat: SimpleChatModel, codyApiVersion: number, charLimit: number): Promise<PromptInfo>
 }
 
-const ENHANCED_CONTEXT_ALLOCATION = 0.6 // Enhanced context should take up 60% of the context window
```
abeatrix, line 28: this is moved to the lib/shared/src/token/constants.ts file. the number remains unchanged

vscode/src/commands/context/directory.ts
```diff
@@ -57,15 +60,19 @@ export async function getContextFileFromDirectory(directory?: URI): Promise<Cont
 
                 const bytes = await vscode.workspace.fs.readFile(fileUri)
                 const decoded = new TextDecoder('utf-8').decode(bytes)
-                const truncatedContent = truncateText(decoded, MAX_CURRENT_FILE_TOKENS)
+                const truncatedContent = truncateText(decoded, USER_CONTEXT_TOKEN_BUDGET_IN_BYTES)
```
dominiccooney, line 63: You need to make a version of `truncateText`, `truncateToTokens` which:

- Tokenizes the content, or a large enough chunk of it.
- Takes the first "limit" tokens
- Decodes those tokens to text

If the tokenizer has a stream interface you can do something better, but that should be fine for now.

dominiccooney (changes requested):
Great start. Some feedback inline. Let's try to push token-based truncation a bit further in this PR.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/ContextItem.kt
```diff
@@ -48,6 +48,7 @@ data class ContextItemSymbol(
   val title: String? = null,
   val source: ContextItemSource? = null, // Oneof: embeddings, user, keyword, editor, filename, search, unified, selection, terminal, uri
   val size: Int? = null,
```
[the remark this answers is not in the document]
abeatrix: converted into token count

lib/shared/src/token/index.ts
```diff
@@ -0,0 +1,130 @@
...
+        return messages.reduce((acc, m) => acc + TokenCounter.getTokenCountForMessage(m), 0)
+    }
+
+    /**
+     * Calculates the total number of bytes required to represent the given array of messages.
+     *
+     * @param messages - An array of messages to calculate the total byte count for.
+     * @returns The total number of bytes required to represent the provided messages.
+     */
+    public static getMessagesByteCount(messages: Message[]): number {
```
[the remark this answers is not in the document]
abeatrix: removed

lib/shared/src/token/constants.ts
```diff
@@ -0,0 +1,17 @@
...
+export const BYTES_PER_TOKEN = 4
+
+export const CHAT_TOKEN_BUDGET = 7000
+export const FAST_CHAT_TOKEN_BUDGET = 4096
+
+/**
+ * The total context window reserved for user-added context (e.g. @-mention).
+ */
+export const USER_CONTEXT_TOKEN_BUDGET = 15000
+export const USER_CONTEXT_TOKEN_BUDGET_IN_BYTES = tokensToBytes(USER_CONTEXT_TOKEN_BUDGET)
```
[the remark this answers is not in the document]
abeatrix: updated to always use token instead of character count or bytes!

vscode/src/commands/context/directory.ts
abeatrix: updated

lib/shared/src/codebase-context/messages.ts
```diff
@@ -38,9 +38,14 @@ interface ContextItemCommon {
     source?: ContextItemSource
 
     /**
-     * The size of the file in bytes.
+     * The characters count in the content of the item.
      */
     size?: number
```
dominiccooney, line 43: In the earlier conversation (which I hit resolved on... whoops) you said you converted this to tokens, but here it says characters.

Let's put the units in the name of the field at least, might make it possible to detect some bugs with 👀  ... so `sizeTokens` or `sizeChars` (which is it?)

lib/shared/src/index.ts
```diff
@@ -224,3 +222,10 @@ export {
     isURLContextItem,
     fetchContentForURLContextItem,
 } from './mentions/urlContextItems'
+export { TokenCounter } from './token'
+export {
+    USER_CONTEXT_TOKEN_BUDGET,
+    ENHANCED_CONTEXT_ALLOCATION,
+    CHARACTERS_PER_TOKEN,
```
dominiccooney, line 229: We need this because the "rest" of the context (not @-tagged files) uses it, is that right?

lib/shared/src/models/index.test.ts
```diff
@@ -1,35 +1,35 @@
...
-describe('getMaxCharsByModel', () => {
+describe('getMaxTokenByID', () => {
     beforeAll(() => {
         ModelProvider.getProviders(ModelUsage.Chat, false, DOTCOM_URL.toString())
     })
 
     it('returns default token limit for unknown model', () => {
-        const maxChars = ModelProvider.getMaxCharsByModel('unknown-model')
-        expect(maxChars).toEqual(tokensToChars(DEFAULT_CHAT_MODEL_TOKEN_LIMIT))
+        const maxToken = ModelProvider.getMaxTokenByID('unknown-model')
```
dominiccooney, line 14: Let's rename `getMaxTokenByID` to be `maxRequestTokensForModel` or something like that?

Reason is the tokens also have numbers, and "max token" sounds like the largest token value.
```suggestion
        const maxTokens = ModelProvider.maxRequestTokensForModel('unknown-model')
```

lib/shared/src/models/index.ts
```diff
@@ -24,8 +24,7 @@ export class ModelProvider {
         // The usage of the model, e.g. chat or edit.
         public readonly usage: ModelUsage[],
         // The maximum number of tokens that can be processed by the model in a single request.
-        // NOTE: A token is equivalent to 4 characters/bytes.
-        public readonly maxToken: number = DEFAULT_CHAT_MODEL_TOKEN_LIMIT
+        public readonly maxToken: number = CHAT_TOKEN_BUDGET
```
dominiccooney, line 27:
```suggestion
        public readonly maxRequestTokens: number = CHAT_TOKEN_BUDGET
```

or `maxTokens` or something like that. `maxToken` sounds like "the biggest value a token could have."

I like `maxRequestTokens` because some models have different input and output token limits. If we have to handle them specially later, the purpose of `maxRequestTokens` is clear but `maxTokens` could be ambiguous (is it input, output, requests, something else?)

lib/shared/src/prompt/truncation.ts
```diff
@@ -1,13 +1,15 @@
...
  * Truncates text to the given number of tokens, keeping the start of the text.
  */
 export function truncateText(text: string, maxTokens: number): string {
-    const maxLength = maxTokens * CHARS_PER_TOKEN
-    return text.length <= maxLength ? text : text.slice(0, maxLength)
+    const encoded = TokenCounter.encode(text)
+    if (encoded.length < maxTokens) {
+        return text
+    }
+    return TokenCounter.decode(encoded.slice(0, maxTokens))
```
dominiccooney, line 12: This is awesome! 🎉

lib/shared/src/prompt/truncation.ts
```diff
@@ -1,13 +1,15 @@
...
+import { TokenCounter } from '../token'
 
 /**
  * Truncates text to the given number of tokens, keeping the start of the text.
  */
 export function truncateText(text: string, maxTokens: number): string {
-    const maxLength = maxTokens * CHARS_PER_TOKEN
-    return text.length <= maxLength ? text : text.slice(0, maxLength)
+    const encoded = TokenCounter.encode(text)
+    if (encoded.length < maxTokens) {
```
dominiccooney, line 9: This could me more succinct and save some compute sometimes with `<=`
```suggestion
    return encoded.length <= maxTokens ? text : TokenCounter.decode(encoded.slice(0, maxTokens))
```

lib/shared/src/prompt/truncation.ts
```diff
@@ -50,6 +52,9 @@ export function truncateTextNearestLine(
  * Truncates text to the given number of tokens, keeping the end of the text.
  */
 export function truncateTextStart(text: string, maxTokens: number): string {
-    const maxLength = maxTokens * CHARS_PER_TOKEN
-    return text.length <= maxLength ? text : text.slice(-maxLength - 1)
+    const encoded = TokenCounter.encode(text)
+    if (encoded.length < maxTokens) {
```
dominiccooney, line 56: <= here too, as above. If it is exactly the right length, we don't need to decode it.

lib/shared/src/prompt/truncation.ts
```diff
@@ -50,6 +52,9 @@ export function truncateTextNearestLine(
...
+        return text
+    }
+    return TokenCounter.decode(encoded.slice(-maxTokens - 1))
```
dominiccooney, line 59: This should be `-maxTokens`, you'll get exactly `maxTokens` (e.g. `'food'.slice(-2)` is `'od'`)
```suggestion
    return TokenCounter.decode(encoded.slice(-maxTokens))
```

lib/shared/src/token/constants.ts
```diff
@@ -0,0 +1,20 @@
+export const CHAT_TOKEN_BUDGET = 7000
+export const FAST_CHAT_TOKEN_BUDGET = 4096
+
+/**
+ * The total context window reserved for user-added context (e.g. @-mention).
```
dominiccooney, line 5: Worth documenting how this relates to `(FAST_)CHAT_TOKEN_BUDGET`. This must be on top of those budgets.

lib/shared/src/token/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export const USER_CONTEXT_TOKEN_BUDGET = 15000
+
+/**
+ * Enhanced context should take up 60% of the context window
```
dominiccooney, line 10: Worth documenting how this relates to `(FAST_)CHAT_TOKEN_BUDGET` too. This must be within that budget?

lib/shared/src/token/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Counts the number of tokens in the given message using the tokenizer.
+     *
+     * @param message - The message to count tokens for.
+     * @returns The number of tokens in the message.
+     */
+    private static getTokenCountForMessage(message: Message): number {
+        if (!message) {
```
dominiccooney, line 141: Can this happen, since the type is `Message` not `Message?` ?

dominiccooney (approved):
Fantastic.

Some feedback inline about naming.

lib/shared/src/prompt/truncation.ts
abeatrix: Oh wow i didn't know that, thank you for the tips!

lib/shared/src/token/constants.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export const ENHANCED_CONTEXT_ALLOCATION = 0.6
+
+/**
+ *  For Claude, a token approximately represents 3.5 English characters.
+ * LINK: https://docs.anthropic.com/claude/docs/glossary#tokens
+ */
+export const CHARACTERS_PER_TOKEN = 3.5
```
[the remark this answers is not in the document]
abeatrix: yea makes sense! I've moved it to the helper function and updated it to be non-export.

lib/shared/src/token/index.ts
abeatrix: it did happen but maybe it was my code? Will test it again to confirm.

lib/shared/src/index.ts
abeatrix: moved CHARACTERS_PER_TOKEN to helper!

lib/shared/src/codebase-context/messages.ts
abeatrix: will try to switch it to sizeToken tomorrow and see if it detects any bugs 🐞

vscode/src/editor/utils/editor-context.ts
```diff
@@ -264,12 +258,14 @@ export async function filterLargeFiles(
...
-            cf.isTooLarge = true
-        }
+        // We cannot get the exact token size without parsing the file, which is expensive.
+        // Instead, we divide the file size in bytes by 4 as a rough estimate of the token size,
+        // which is closer than dividing by 3.5, the number we use to convert characters into tokens.
+        // NOTE: This provides the frontend with a rough idea of when to display large files with a warning based
+        // on available tokens, so that it can prompt the user to import the file via '@file-range' or
+        // via 'right-click on a selection' that only involves reading a single context item, allowing us to read
+        // the file content on-demand instead of in bulk. We would then label the file size more accurately with the tokenizer.
+        cf.size = fileStat.size / 4
```
abeatrix, line 268: @dominiccooney does this address your concerns on converting bytes into tokens for @-mention results? Happy to address them or explain more if this doesn't help!.

sqs:
Just sent up https://github.com/sourcegraph/cody/pull/3887 to fix a couple minor issues.

Issue #3641, opened 2024-04-01
[Cody Ignore] Fetch, interpret enterprise Cody Ignore rules


dominiccooney:
Cody Ignore for Enterprise needs to interpret a different policy than the file-based .cody/ignore:

- The site config supports lists of repositories to be included/excluded from Cody's context following RE2 syntax
- If either include or exclude rules are defined, Cody applies these rules to determine which repositories can be used as context in requests to third-party LLMs
- If neither exclude nor include rules are specified, Cody may use content from any repository on the instance in its requests to third-party LLMs
- If both exclude and include rules are specified, the include rules restrict Cody to use content only from the repositories that match specified patterns. Repositories not matching any include patterns will be excluded.

Open question about what the defaults are, [see Slack thread.](https://sourcegraph.slack.com/archives/C06Q9R098N4/p1711667096619869)

### Related Docs
- [PRD](https://docs.google.com/document/d/10tD8n-5py4XWQD9c7doJB7zIuxa7lwiOkpo2d5NLlnA/edit#heading=h.10vr0jxhrd1r)
- [RFC910](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit#heading=h.trqab8y0kufp)

```[tasklist]
### Tasks
- [x] Add node RE2 package; the RFC specifies RE2
- [x] Fetch the enterprise rules with GraphQL
- [x] Provide a stub implementation for Sourcegraph endpoints which don't support this feature, with a liberal allow policy
- [x] Write a function which takes (repo, relative path) -> do/don't ignore
- [x] Migrate to re2-wasm
- [x] Write a function that takes file URI -> repo name based on the environment tools available.
- [x] Write a function that takes file URI -> do/don't ignore. Integrate it with both of the functions above.
- [x] Add Git submodules support
- [x] Update the context filters GraphQL based on the latest backend changes
- [x] Instantiate the context filters provider on the extension activation
- [x] Investigate the repo name resolution using the Sourcegraph GraphQL API
- [x] Add TODO about non-file URIs
- [x] Use all available remote URLs
- [x] ContextFilterProvider API should support notifications for publishing changes when the rules are changed on the backend, with polling
- [x] Expose API thru Agent for JetBrains and others, API should support notifications for publishing changes when the rules are changed on the backend (extension polls the backend)
- [x] Add agent integration tests
[...]

PR #3747, opened 2024-04-09
VS Code: add remote context filters provider

valerybugakov:
## Context

- Closes subtasks from https://github.com/sourcegraph/cody/issues/3641
    - Add node RE2 package; the RFC specifies RE2
    - Fetch the enterprise rules with GraphQL
    - Provide a stub implementation for Sourcegraph endpoints which don't support this feature, with a liberal allow policy
    - Write a function which takes (repo, relative path) -> do/don't ignore
- Builds on top of backend changes https://github.com/sourcegraph/sourcegraph/pull/61101
- Mirrors backend logic from https://github.com/sourcegraph/sourcegraph/pull/61641

## Test plan

CI and new unit tests

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -11,6 +11,7 @@ import { isError } from '../../utils'
 import { DOTCOM_URL, isDotCom } from '../environments'
 
 import {
+    CONTEXT_FILTERS_QUERY,
     CONTEXT_SEARCH_QUERY,
     CURRENT_SITE_CODY_CONFIG_FEATURES,
     CURRENT_SITE_CODY_LLM_CONFIGURATION,
@@ -180,6 +181,22 @@ export interface ContextSearchResult {
     startLine: number
     endLine: number
     content: string
+}
+
+export interface ContextFiltersResult {
+    include: CodyContextFilterItem[]
+    exclude: CodyContextFilterItem[]
+}
+
+export interface CodyContextFilterItem {
+    repoNamePattern: string
+    // Not implemented
+    filePathPattern?: string
+}
+
+const EMPTY_CONTEXT_FILTERS: ContextFiltersResult = {
+    include: [],
+    exclude: [],
 }
 
 interface SearchAttributionResults {
@@ -513,6 +530,24 @@ export class SourcegraphGraphQLAPIClient {
                 }))
             )
         )
+    }
+
+    public async contextFilters(): Promise<ContextFiltersResult | null | Error> {
+        const response =
+            await this.fetchSourcegraphAPI<APIResponse<ContextFiltersResult | null>>(
+                CONTEXT_FILTERS_QUERY
+            )
+
+        const result = extractDataOrError(response, data => data)
+
+        if (result instanceof Error) {
+            // Ignore errors caused by outdated Sourcegraph API instances.
+            if (hasOutdatedAPIErrorMessages(result)) {
+                return EMPTY_CONTEXT_FILTERS
+            }
+        }
+
+        return result
     }
 
     /**
@@ -871,9 +906,7 @@ export class ConfigFeaturesSingleton {
         const previousConfigFeatures = this.configFeatures
         this.configFeatures = this.fetchConfigFeatures().catch((error: Error) => {
             // Ignore fetcherrors as older SG instances will always face this because their GQL is outdated
-            if (
-                !(error.message.includes('FetchError') || error.message.includes('Cannot query field'))
-            ) {
+            if (!hasOutdatedAPIErrorMessages(error)) {
                 logError('ConfigFeaturesSingleton', 'refreshConfigFeatures', error.message)
             }
             // In case of an error, return previously fetched value
@@ -910,3 +943,7 @@ export type LogEventMode =
     | 'dotcom-only' // only log to dotcom
     | 'connected-instance-only' // only log to the connected instance
     | 'all' // log to both dotcom AND the connected instance
+
+function hasOutdatedAPIErrorMessages(error: Error): boolean {
+    return error.message.includes('FetchError') || error.message.includes('Cannot query field')
+}
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -194,9 +194,14 @@ export interface CodyContextFilterItem {
     filePathPattern?: string
 }
 
-const EMPTY_CONTEXT_FILTERS: ContextFiltersResult = {
+const INCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
     include: [],
     exclude: [],
+}
+
+const EXCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
+    include: [],
+    exclude: [{ repoNamePattern: '*' }],
 }
 
 interface SearchAttributionResults {
@@ -541,10 +546,15 @@ export class SourcegraphGraphQLAPIClient {
         const result = extractDataOrError(response, data => data)
 
         if (result instanceof Error) {
+            logError('SourcegraphGraphQLAPIClient', 'contextFilters', result.message)
+
             // Ignore errors caused by outdated Sourcegraph API instances.
             if (hasOutdatedAPIErrorMessages(result)) {
-                return EMPTY_CONTEXT_FILTERS
+                return INCLUDE_EVERYTHING_CONTEXT_FILTERS
             }
+
+            // Exclude everything in case of an unexpected error.
+            return EXCLUDE_EVERYTHING_CONTEXT_FILTERS
         }
 
         return result
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -191,7 +191,7 @@ export interface ContextFiltersResult {
 export interface CodyContextFilterItem {
     repoNamePattern: string
     // Not implemented
-    filePathPattern?: string
+    filePathPatterns?: string[]
 }
 
 const INCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
@@ -201,7 +201,7 @@ const INCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
 
 const EXCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
     include: [],
-    exclude: [{ repoNamePattern: '*' }],
+    exclude: [{ repoNamePattern: '.*' }],
 }
 
 interface SearchAttributionResults {
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -546,8 +546,6 @@ export class SourcegraphGraphQLAPIClient {
         const result = extractDataOrError(response, data => data)
 
         if (result instanceof Error) {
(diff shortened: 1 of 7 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -182,6 +183,27 @@ export interface ContextSearchResult {
...
 
+export interface ContextFiltersResult {
+    include: CodyContextFilterItem[]
+    exclude: CodyContextFilterItem[]
+}
+
+export interface CodyContextFilterItem {
+    repoNamePattern: string
+    // Not implemented
+    filePathPattern?: string
```
taras-yemets, line 194:
```suggestion
    filePathPatterns?: string[]
```
In future iterations we plan to support multiple file path patterns per repo pattern ([RFC](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit#bookmark=id.hys3h7gd8w2a)).

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -182,6 +183,27 @@ export interface ContextSearchResult {
...
+}
+
+const INCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
+    include: [],
+    exclude: [],
+}
+
+const EXCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFiltersResult = {
+    include: [],
+    exclude: [{ repoNamePattern: '*' }],
```
taras-yemets, line 204:
```suggestion
    exclude: [{ repoNamePattern: '.*' }],
```

vscode/src/context/context-filters-provider.ts
```diff
@@ -0,0 +1,126 @@
...
+export const REFETCH_INTERVAL = 60 * 60 * 1000 // 1 hour
+
+interface ParsedContextFilters {
+    include: ParsedContextFilterItem[]
+    exclude: ParsedContextFilterItem[]
+}
+
+interface ParsedContextFilterItem {
+    repoNamePattern: RE2
+    filePathPattern?: RE2
```
taras-yemets, line 16:
```suggestion
    filePathPatterns?: RE2[]
```

vscode/src/context/context-filters-provider.ts
```diff
@@ -0,0 +1,126 @@
...
+    private startRefetchTimer(): void {
+        this.fetchIntervalId = setTimeout(() => {
+            this.fetchContextFilters()
+            this.startRefetchTimer()
+        }, REFETCH_INTERVAL)
+    }
+
+    public isPathAllowed(repoName: string, relativePath: string): boolean {
+        const cacheKey = `${repoName}:${relativePath}`
+        const cached = this.cache.get(cacheKey)
```
taras-yemets, line 62: nit: as we support only repo-level includes/excludes it may be reasonable to have repo name only (without file path) as a cache key. 
Later, when we add support for file path patterns, we may want to maintain per repo cache.

vscode/src/context/context-filters-provider.ts
```diff
@@ -0,0 +1,126 @@
...
+        if (cached !== undefined) {
+            return cached
+        }
+
+        // If we don't have any context filters, we exclude everything.
+        let isAllowed = Boolean(this.contextFilters)
```
taras-yemets, line 68: FYI GraphQL schema was recently changed (PR is not yet merged) to make all the context filter fields required (https://github.com/sourcegraph/sourcegraph/pull/61101#discussion_r1557527697).

valerybugakov: Here, we handle cases where the API returns an error instead of the expected response.

vscode/src/context/context-filters-provider.ts
```diff
@@ -0,0 +1,126 @@
...
+
+        if (this.contextFilters?.include.length) {
+            isAllowed = false
```
taras-yemets, line 71: nit: we may not need to this line. 
We later assign `isAllowed` the result of `checkFilter` function, thus if neither matches, we'll get `false` in the end of the cycle.

vscode/src/context/context-filters.provider.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    })
+
+    it('uses cached results for repeated calls', async () => {
+        await initProviderWithContextFilters({
+            include: [{ repoNamePattern: '^github\\.com\\/sourcegraph\\/.*' }],
+            exclude: [],
+        })
+
+        expect(provider.isPathAllowed('github.com/sourcegraph/cody', 'src/main.ts')).toBe(true)
+        expect(provider.isPathAllowed('github.com/sourcegraph/cody', 'src/main.ts')).toBe(true)
```
taras-yemets, line 125: How do we confirm that the result is returned from the cache?

valerybugakov: Remove this part while I was refactoring the tests. Added back here: https://github.com/sourcegraph/cody/pull/3747/commits/b82c95608a92c63523a612da10567017b26fa76b

taras-yemets (approved):
Great work!
Left a few inline comments.

vscode/src/context/context-filters-provider.ts
valerybugakov: I'll update this in a follow-up PR where I want to test the function's performance for different use cases.

philipp-spiess:
@valerybugakov @taras-yemets Do you folks have any context on why we need to pull in a third-party regex library (re2) for this?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -910,3 +951,7 @@ export type LogEventMode =
     | 'dotcom-only' // only log to dotcom
     | 'connected-instance-only' // only log to the connected instance
     | 'all' // log to both dotcom AND the connected instance
+
+function hasOutdatedAPIErrorMessages(error: Error): boolean {
+    return error.message.includes('FetchError') || error.message.includes('Cannot query field')
```
philipp-spiess, line 956: `FetchError` seems very generic and might not mean "outdated API". 🤔

lib/shared/src/sourcegraph-api/graphql/queries.ts
```diff
@@ -152,6 +152,20 @@ query GetCodyContext($repos: [ID!]!, $query: String!, $codeResultsCount: Int!, $
     }
 }`
 
+export const CONTEXT_FILTERS_QUERY = `
```
philipp-spiess, line 155: Nothing to fix right now but we are adding more and more queries that are being queried on startup. Might make sense to consolidate them into just 1 request at some point (hello https://sourcegraph.slack.com/archives/C03L2R35ENL/p1712597477364249 etc). I understand that we have issues with backward-compat though so :sigh:

vscode/src/context/context-filters-provider.ts
```diff
@@ -0,0 +1,129 @@
+import { LRUCache } from 'lru-cache'
+import RE2 from 're2'
+import type * as vscode from 'vscode'
+
+import { type CodyContextFilterItem, graphqlClient, logError } from '@sourcegraph/cody-shared'
+
+export const REFETCH_INTERVAL = 60 * 60 * 1000 // 1 hour
```
philipp-spiess, line 7: We implement something like this in a few places now. Might make sense to see if we can unify this behavior?

vscode/src/context/context-filters-provider.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private startRefetchTimer(): void {
+        this.fetchIntervalId = setTimeout(() => {
+            this.fetchContextFilters()
+            this.startRefetchTimer()
+        }, REFETCH_INTERVAL)
+    }
+
+    public isPathAllowed(repoName: string, relativePath: string): boolean {
```
philipp-spiess, line 60: Should we make this provider a singleton so we have easy access to this without prop drilling?

philipp-spiess (commented):
Some general comments, looks good overall

vscode/src/context/context-filters-provider.ts
philipp-spiess, line 60: Can we make this get a `URI` instead of a relative path? I want to avoid callers building the wrong "relative path" for some reason. Ideally the repo name is also something the caller doesn't need to care about and we should infer it based on the file URI.

philipp-spiess (commented):
OK one more thing that we might want to change

lib/shared/src/sourcegraph-api/graphql/client.ts
valerybugakov: I used the helper that Ara introduced for another API call here. I verified only the `Cannot query field` part. @arafatkatze, could you share context on the `FetchError` condition you added to the `fetchConfigFeatures` call?

vscode/src/context/context-filters-provider.ts
valerybugakov: Sure, that's the plan 👍

valerybugakov: I'll introduce a separate method for the URI check. It's the next task in https://github.com/sourcegraph/cody/issues/3641

valerybugakov:
> Do you folks have any context on why we need to pull in a third-party regex library (re2) for this?

@philipp-spiess, as per [RFC](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit): "We support [RE2 syntax](https://golang.org/s/re2syntax) (same as for the [Search Query Syntax](https://sourcegraph.com/docs/code-search/queries))."

vscode/src/context/context-filters-provider.ts
philipp-spiess: Oki nice :) sorry for asking so many questions 😄

lib/shared/src/sourcegraph-api/graphql/client.ts
arafatkatze: Sorry for the late response Valery.
See this pr -> https://github.com/sourcegraph/cody/pull/3117

image

valerybugakov: @arafatkatze, it looks like the `FetchError` check [was introduced](https://github.com/sourcegraph/cody/blame/20e1a9e2c6d9380432744af49d2f8e61bdbf5009/lib/shared/src/sourcegraph-api/graphql/client.ts#L877-L878) in [this PR](https://github.com/sourcegraph/cody/pull/2559). It states that the error happens when the Sourcegraph API is outdated. How did you test this case? I was able to reproduce only the `Cannot query field` error.

arafatkatze: > How did you test this case? 

I tested it [on these ](https://github.com/sourcegraph/cody/pull/3117#pullrequestreview-1875908331) I tried getting into both of those instances again but since I don't have the admin access now and I can't replicate it again with certainty(because I don't see the instance version) but remember having tested it back then and I posted the screenshots of the errors too see [the comment on the PR](https://github.com/sourcegraph/cody/pull/3117#issuecomment-1941124247). 

The rationale behind silently excluding fetch errors is primarily to mitigate the clutter in logs that can arise from older instances. When older instances encounter fetch errors, the logs can quickly become inundated with repetitive messages from cronjobs, making it challenging to discern important information.

Additionally, while older instances are a significant contributor to fetch errors, it's essential to acknowledge that other factors could also trigger such errors. By indiscriminately excluding fetch errors, we risk overlooking critical messages unrelated to outdated instances. Therefore, while older instances play a substantial role, it's not the sole determinant, and excluding fetch errors ensures a more balanced approach.

My recommendation in this case would be to just say excludefetcherrors because thats better naming

PR #3749, opened 2024-04-09
Tree-sitter: capture docstring for documentable nodes

umpox:
## Description

Fixes an issue where docstrings were not captured if `insertionPoint` did not exist (doesn't for variables or anything in Python where we assume the default insertionPoint of above the symbol)

## Test plan

1. Document python functions, check hint disappears
2. Document python variables, check hint disappears
3. Document JS functions, check hint disappears
4. Document JS variables, check hint disappears


vscode/src/tree-sitter/queries/python.ts
```diff
@@ -63,7 +63,10 @@ const DOCUMENTABLE_NODES_QUERY = dedent`
     ;--------------------------------
     (assignment
         left: (identifier) @symbol.identifier) @range.identifier
-    (string) @docstring
+
+    ; Comments
+    ;--------------------------------
+    (expression_statement (string)) @comment
```
umpox, line 69: @abeatrix This lets us safely check the entire line for a comment without wrongly detecting one

Before this code could be seen as a `@comment`:
def my_function():
    my_string = "Hello, World!"

Now it will only match if it is directly declaring a string

abeatrix (approved):
Great job! Tested on typescript and python and both work as expected!

PR #3750, opened 2024-04-09
Troubleshoot: Show auth connection issues

### LINKED ISSUE #3673: bug: Connection issues can lead to seemingly signing-out
### Version

v1.10.2

### Describe the bug

If for whatever reason there are connection issues during plugin initialization this can lead to the user being shown a Sign-In screen.

This can easily be tested by disabling wifi, restarting VSCode and seeing that Cody is signed out. Now simply re-enable your internet, reload VSCode and you're "signed-in" again 🪄.

So although the token isn't lost and the user is still technically authenticated the error state does not reflect that.

### Expected behavior

If we aren't able to complete initialization properly we should put Cody into an error state and give the user the option to either retry (check internet connection etc), switch to debug mode, or to factory reset. The auth screen should only be shown if there is no valid token present or if the API explicitly tells us we are unauthenticated.

### Additional context

_No response_

RXminuS:
Fixes #3673 where we mistakingly show a signIn screen if there are connectivity issues to the Cody server.

CleanShot 2024-04-17 at 14 46 44@2x

## Test plan
- End2end tests are added
- Storybook is added to view the UI
    - Allow for full-viewport rendering to test for breakpoint issues.


vscode/src/chat/protocol.ts
```diff
@@ -90,6 +90,16 @@ export type WebviewMessage =
...
+    | {
+          command: 'troubleshoot/reloadAuth'
+      }
+    | {
+          command: 'troubleshoot/showOutputLogs'
+      }
+    | {
+          command: 'troubleshoot/fullReset'
+          askConfirmation?: boolean
+      }
```
abeatrix, line 102:
```suggestion
    | {
          command: 'troubleshoot'; action: 'output' | 'reset'
      }
```

Can this be simplified? We already have a [reload](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/protocol.ts?L74) command so i think we just need to add 'reset' and `output`?
We should always ask for confirmation before resetting, imo.

vscode/src/chat/protocol.ts
```diff
@@ -90,6 +90,16 @@ export type WebviewMessage =
           command: 'attribution-search'
           snippet: string
       }
+    | {
+          command: 'troubleshoot/reloadAuth'
+      }
```
abeatrix, line 95:
```suggestion
```

There is already a command registered for [reload](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/protocol.ts?L74) that is unused, so i think we can just reuse that?

vscode/src/chat/chat-view/SidebarViewController.ts
```diff
@@ -151,6 +146,64 @@ export class SidebarViewController implements vscode.WebviewViewProvider {
             case 'show-page':
                 await vscode.commands.executeCommand('show-page', message.page)
                 break
+            case 'troubleshoot/reloadAuth': {
```
abeatrix, line 149:
```suggestion
            case 'reload': {
```

vscode/src/chat/chat-view/SidebarViewController.ts
```diff
@@ -151,6 +146,64 @@ export class SidebarViewController implements vscode.WebviewViewProvider {
...
+                telemetryRecorder.recordEvent('cody.troubleshoot', 'showOutputLogs', {
+                    metadata: {},
+                })
+                enableDebugMode()
+                break
+            case 'troubleshoot/fullReset':
+                // show confirmation dialog
+                // we can't do this anywhere else because we would no longer
+                // be able to send the telemetry at that point.
+                if (message.askConfirmation) {
```
abeatrix, line 185: i think we should always ask for confirmation before this function is run. Maybe we should move this to the function instead? We can just log the "clicked" event here (or in the fullReset()), and then log the "executed" event once it's been executed.

vscode/src/services/utils/troubleshoot.ts
```diff
@@ -0,0 +1,23 @@
...
+import { CodyChatPanelViewType } from '../../chat/chat-view/ChatManager'
+import type { AuthProvider } from '../AuthProvider'
+import { clearAccessToken, secretStorage } from '../SecretStorageProvider'
+
+/**
+ * Resets the extension by clearing all storage data, secretStorage data
+ * and clearing any settings. This should put cody back into a pristine state.
+ */
+export async function fullReset(authProvider: AuthProvider): Promise<void> {
+    //TODO(rnauta): make available as command?
```
abeatrix, line 11: as mentioned above, i think we should move the confirmation window into this function 🙏

vscode/src/services/utils/troubleshoot.ts
```diff
@@ -0,0 +1,23 @@
+import * as vscode from 'vscode'
+import { CodyChatPanelViewType } from '../../chat/chat-view/ChatManager'
+import type { AuthProvider } from '../AuthProvider'
+import { clearAccessToken, secretStorage } from '../SecretStorageProvider'
+
+/**
+ * Resets the extension by clearing all storage data, secretStorage data
```
abeatrix, line 7: From what i can tell the function is only removing the tokens from the secret storage, but everything in the local storage stays, and the settings are not removed. Is that the expected behavior?

vscode/src/services/utils/troubleshoot.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export async function fullReset(authProvider: AuthProvider): Promise<void> {
+    //TODO(rnauta): make available as command?
+    const { endpoint } = authProvider.getAuthStatus()
+    if (endpoint) {
+        await secretStorage.deleteToken(endpoint)
+        await authProvider.auth('', null)
+    } else {
+        await clearAccessToken()
+    }
```
abeatrix, line 18:
```suggestion
    }
    await clearAccessToken()
```
`clearAccessToken()` removes the last used access token, but doesn't remove the token stored under the endpoint key. I think we will always want to run this for a full reset?

vscode/webviews/App.tsx
```diff
@@ -226,59 +227,71 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
         return <LoadingPage />
     }
 
-    return (
-        <div className="outer-container">
-            {view === 'login' || !authStatus.isLoggedIn || !userAccountInfo ? (
+    if (authStatus?.showNetworkError) {
```
abeatrix, line 230:
```suggestion
    if (authStatus.showNetworkError) {
```

abeatrix (commented):
Left some comments inline. The only concern i have is about the reset function, but the rest looks good to me. Nice work!

RXminuS:
Yeah I'll probably just change that to a sign out. I just didn't want someone to get stuck with bad settings and not get back to the sign in screen.

Thanks for taking a look

RXminuS:
I've decided to remove the "fullReset" and "showOutput" logs in favor of another feature coming soon to help users enable and navigate debug mode/troubleshooting mode more easily. So this screen will only allow you to retry or sign-out explicitly. I'd still like to keep the reloadAuth command under the troubleshoot namespace because there will be about 4-5 more commands added soon under that namespace.

philipp-spiess:
@RXminuS sorry regarding the message above, see: https://github.com/sourcegraph/cody/pull/3797#issuecomment-2058969736

vscode/src/chat/protocol.ts
RXminuS: I'll be adding a few more things under the troubleshooting namespace though. So I've just removed the reload command

RXminuS:
@philipp-spiess increasing the timeout fixed the e2e tests 👍  thanks for the help

vscode/webviews/App.tsx
```diff
@@ -230,59 +231,71 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
...
+                <ConnectionIssuesPage
+                    configuredEndpoint={authStatus.endpoint}
+                    telemetryService={telemetryService}
+                    vscodeAPI={vscodeAPI}
+                />
+            </div>
+        )
+    }
+
+    if (view === 'login' || !authStatus.isLoggedIn || !userAccountInfo) {
```
abeatrix, line 246: This change seems to remove the current loading screen at initiate time:

image


### Before

https://github.com/sourcegraph/cody/assets/68532117/1ad53805-82ff-4350-8ae1-162dad69239d



### After

https://github.com/sourcegraph/cody/assets/68532117/14fc5af4-ec67-4eeb-b018-6da4aa534d02

RXminuS: Derp...good point!

toolmantim:
Nice! This design is looking good too.

For the width/positioning, use VS Code’s Run & Debug panel and responsive breakpoints as the guide:

https://github.com/sourcegraph/cody/assets/153/40d49cc4-eec5-4934-9c0c-fff0632f833c

i.e. icon & buttons are full width until a point, then left aligned

vscode/webviews/App.tsx
RXminuS: Mmm...I'm a little confused by this as the condition for showing the <LoadingPage> hasn't changed :-/ I'll look into it

RXminuS: I still don't understand how my changes caused this change in behavior for you and I wasn't able to replicate it. However I did find that we didn't really show the loading screen at all until the full initialization was completed. So I've changed that slightly so that we no longer wait for the auth state to resolve before showing a view. That should show the loading state much earlier.

Although it feels a little bit of a bandaid solution, I think we might need to think about more declarative UI state / routing.

Here a video of the change of me pausing the HTTP response after a reload. Could you give this another go?

https://github.com/sourcegraph/cody/assets/3949285/e57f401c-3f34-4f13-90d6-0841a91d5034

RXminuS:
> Nice! This design is looking good too.
> 
> For the width/positioning, use VS Code’s Run & Debug panel and responsive breakpoints as the guide:

@toolmantim VScode had a few inconsistent breakpoints so I just picked `650px` as the max. They also have global flags for reduced motion that I didn't want to dive into now so I didn't add any animations. I think that's something we can globally define once we set up Tailwind helpers. But for now, I've tried to follow the spirit of it :-)

https://github.com/sourcegraph/cody/assets/3949285/3eef94e9-971b-4c27-a8be-97130ab4aed0

vscode/webviews/App.tsx
RXminuS: Dang...that seems to have broken all the tests 🤦

vscode/src/main.ts
```diff
@@ -164,7 +164,10 @@ const register = async (
         })
     )
 
-    await authProvider.init()
+    // Don't wait for this promise to resolve as it's not critical to extension startup.
+    // Doing so will prevent the loading screen from showing. We're just firing of this
+    // event so that at some point the auth status will resolve.
+    void authProvider.init()
```
abeatrix, line 170: I think this is the cause of the Loading screen getting skipped issue. 
Since other services can now initialized before the auth step has been completed, the sidebar is now starting with the rest of the services (specifically contextProvider that sends the sidebar init events). 

@philipp-spiess @umpox is there anything you can think of top of your head that would be affected by this change where services can be initialized before the initiate auth has been completed?

Another bug I noticed with this change is that my opened webview panel is now closed on reload:

https://github.com/sourcegraph/cody/assets/68532117/d70776c7-a6ad-4294-b19a-1757785a84e6

I think we should revert this change for now until we can improve the whole dependency situation, unless your other changes depend on this? What do you think?

abeatrix (changes requested):

vscode/src/main.ts
RXminuS: It's not the cause of the original issue / video you posted because I only just added this to try and fix it at the root. I'm happy to revert it if it is causing issues but then I still don't understand how https://github.com/sourcegraph/cody/pull/3750#discussion_r1568093267 issue presents itself because I basically never get shown a loading screen in either version.

RXminuS: It's commit `56e1ab7` where I added this

RXminuS: Also I think that the `showNetworkError` is equally problematic as the `showLoading` in that I don't think any components other then the UI handle it. So if for some reason there's a network problem the auth would simply be assumed to b `signedOut` rather than `unknown`.

Probably the easiest way to find all the potential problem areas is if I set the AuthStatus to optional and follow the TS errors.

RXminuS: Ok, I've put up an issue here https://github.com/sourcegraph/cody/issues/3835 and tests seem to be passing again.

abeatrix (approved):
Thanks for working on this and for adding all those amazing tests! Confirm the chat panel is now showing up correctly on reload, and we can look into the Loading page issue separately 😄 well done!

PR #3751, opened 2024-04-09
Add ldid as part of release action

pjlast:
The current 0.0.5 release has broken MacOS binaries. From the [release action](https://github.com/sourcegraph/cody/actions/runs/8568196553/job/23481553267):

```
> Warning Unable to sign the macOS executable
  Due to the mandatory code signing requirement, before the
  executable is distributed to end users, it must be signed.
  Otherwise, it will be immediately killed by kernel on launch.
  An ad-hoc signature is sufficient.
  To do that, run pkg on a Mac, or transfer the executable to a Mac
  and run "codesign --sign - <executable>", or (if you use Linux)
  install "ldid" utility to PATH and then run pkg again
```

This PR adds this [Install ldid action](https://github.com/MOZGIII/install-ldid-action) to the release in an attempt to fix it

## Test plan

Not yet sure

### DIFF
--- a/.github/workflows/agent-release.yml
+++ b/.github/workflows/agent-release.yml
@@ -20,6 +20,9 @@ jobs:
       - uses: pnpm/action-setup@d882d12c64e032187b2edb46d3a0d003b7a43598 # SECURITY: pin third-party action hashes
         with:
           run_install: true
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Looks good to me, the manual test plan should at least point to the steps for doing an agent release and ideally a pointer to a job that succeeded off a branch.

Would be helpful to point to the kind of failure missing ldid causes too so people can quickly end up here if the problem arises again.

PR #3753, opened 2024-04-09
Add support for using Sourcegraph Embeddings API

rafax:
Add's support for Sourcegraph Embeddings API, changing:

- model identifier and vector dimension (passed to BFG aka Cody Engine)
- storage location for indices (to allow switching from OpenAI to Sourcegraph API without data loss / conflicts)
- test used to communicate how we process user code

[Loom](https://www.loom.com/share/38a46f95a2414726816de62740c90760)

Potentially TODO (separate PR):

- can we "assume consent" and re-embed automatically if user has used OpenAI (without accepting again)?

## Test plan

- build BFG from main (`cd $BFG_REPO && cd crated/bfg && cargo build --release`)
- configure your VSCode to use locally-built BFG (add `  "cody.experimental.cody-engine.path": "$BFG_REPO/target/release/bfg"` to settings.json)
- enable the Feature [flag](https://sourcegraph.com/site-admin/feature-flags/configuration/cody-use-sourcegraph-embeddings) for yourself

### DIFF
--- a/vscode/src/configuration.ts
+++ b/vscode/src/configuration.ts
@@ -8,6 +8,7 @@ import {
     OLLAMA_DEFAULT_URL,
 } from '@sourcegraph/cody-shared'
 
+import { URI } from 'vscode-uri'
 import {
     CONFIG_KEY,
     type ConfigKeys,
@@ -81,6 +82,17 @@ export function getConfiguration(
         'autocomplete.experimental.graphContext',
         null
     )
+
+    function hasValidLocalEmbeddingsConfig(): boolean {
+        return (
+            [
+                'testing.localEmbeddings.model',
+                'testing.localEmbeddings.endpoint',
+                'testing.localEmbeddings.indexLibraryPath',
+            ].every(key => !!getHiddenSetting<string | undefined>(key, undefined)) &&
+            !!getHiddenSetting<number | undefined>('testing.localEmbeddings.dimension', undefined)
+        )
+    }
 
     return {
         proxy: config.get<string | null>(CONFIG_KEY.proxy, null),
@@ -165,19 +177,18 @@ export function getConfiguration(
                 undefined
             ),
         },
-
-        testingLocalEmbeddingsModel: isTesting
-            ? getHiddenSetting<string | undefined>('testing.localEmbeddings.model', undefined)
-            : undefined,
-        testingLocalEmbeddingsDimension: isTesting
-            ? getHiddenSetting<number | undefined>('testing.localEmbeddings.dimension', undefined)
-            : undefined,
-        testingLocalEmbeddingsEndpoint: isTesting
-            ? getHiddenSetting<string | undefined>('testing.localEmbeddings.endpoint', undefined)
-            : undefined,
-        testingLocalEmbeddingsIndexLibraryPath: isTesting
-            ? getHiddenSetting<string | undefined>('testing.localEmbeddings.indexLibraryPath', undefined)
-            : undefined,
+        testingModelConfig:
+            isTesting && hasValidLocalEmbeddingsConfig()
+                ? {
+                      model: getHiddenSetting<string>('testing.localEmbeddings.model'),
+                      dimension: getHiddenSetting<number>('testing.localEmbeddings.dimension'),
+                      endpoint: getHiddenSetting<string>('testing.localEmbeddings.endpoint'),
+                      indexPath: URI.file(
+                          getHiddenSetting<string>('testing.localEmbeddings.indexLibraryPath')
+                      ),
+                      provider: 'openai',
+                  }
+                : undefined,
     }
 }
 
--- a/vscode/src/local-context/enhanced-context-status.test.ts
+++ b/vscode/src/local-context/enhanced-context-status.test.ts
@@ -25,7 +25,7 @@ class TestProvider implements status.ContextStatusProvider {
                         {
                             kind: 'embeddings',
                             state: 'unconsented',
-                            useSourcegraphEmbeddings:false
+                            embeddingsAPIProvider: 'openai',
                         },
                     ],
                 },
@@ -51,7 +51,7 @@ describe('ContextStatusAggregator', () => {
                     {
                         kind: 'embeddings',
                         state: 'unconsented',
-                        useSourcegraphEmbeddings:false,
+                        embeddingsAPIProvider: 'openai',
                     },
                 ],
             },
@@ -95,7 +95,7 @@ describe('ContextStatusAggregator', () => {
                     {
                         kind: 'embeddings',
                         state: 'unconsented',
-                        useSourcegraphEmbeddings:false,
+                        embeddingsAPIProvider: 'openai',
                     },
                     {
                         kind: 'search',
--- a/vscode/src/local-context/enhanced-context-status.test.ts
+++ b/vscode/src/local-context/enhanced-context-status.test.ts
@@ -25,6 +25,7 @@ class TestProvider implements status.ContextStatusProvider {
                         {
                             kind: 'embeddings',
                             state: 'unconsented',
+                            useSourcegraphEmbeddings:false
                         },
                     ],
                 },
@@ -50,6 +51,7 @@ describe('ContextStatusAggregator', () => {
                     {
                         kind: 'embeddings',
                         state: 'unconsented',
+                        useSourcegraphEmbeddings:false,
                     },
                 ],
             },
@@ -93,6 +95,7 @@ describe('ContextStatusAggregator', () => {
                     {
                         kind: 'embeddings',
                         state: 'unconsented',
+                        useSourcegraphEmbeddings:false,
                     },
                     {
                         kind: 'search',
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -233,8 +233,6 @@ export class LocalEmbeddingsController
         logDebug('LocalEmbeddingsController', 'spawnAndBindService', 'service started, initializing')
         const indexPath = getIndexLibraryPath(this.modelConfig.indexSuffix)
 
-        // Tests may override the index library path
-
         const initResult = await service.request('embeddings/initialize', {
             codyGatewayEndpoint: this.endpoint,
             indexPath: indexPath.fsPath,
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -17,6 +17,7 @@ import {
     uriBasename,
 } from '@sourcegraph/cody-shared'
 
+import type { EmbeddingsProvider } from '@sourcegraph/cody-shared/src/codebase-context/context-status'
 import type { IndexHealthResultFound, IndexRequest } from '../jsonrpc/embeddings-protocol'
 import type { MessageHandler } from '../jsonrpc/jsonrpc'
 import { logDebug } from '../log'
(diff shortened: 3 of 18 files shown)


vscode/src/local-context/local-embeddings.ts
```diff
@@ -98,7 +101,8 @@ export class LocalEmbeddingsController
 
     constructor(
         private readonly context: vscode.ExtensionContext,
-        config: LocalEmbeddingsConfig
+        config: LocalEmbeddingsConfig,
+        private readonly useSourcegraphEmbeddings: boolean,
```
dominiccooney, line 105: LocalEmbeddingsConfig has:

```
    testingLocalEmbeddingsModel: string | undefined
    testingLocalEmbeddingsEndpoint: string | undefined
    testingLocalEmbeddingsIndexLibraryPath: string | undefined
```

Why not simply drop the "testing" prefix on those, and handle the experiment where the config object is set up?

vscode/webviews/Components/EnhancedContextSettings.tsx
```diff
@@ -241,7 +241,7 @@ const EmbeddingsConsentComponent: React.FunctionComponent<{ provider: LocalEmbed
     return (
         <div>
             <p className={styles.providerExplanatoryText}>
-                The repository&apos;s contents will be uploaded to OpenAI&apos;s Embeddings API and then
+                The repository&apos;s contents will be uploaded to {provider.useSourcegraphEmbeddings ? "Sourcegraph" : "OpenAI"}&apos;s Embeddings API and then
```
dominiccooney, line 244: @toolmantim I've never been comfortable with this wording, because "uploaded ... and then stored locally" is confusing. It's *not* permanently stored remotely. Wouldn't it be clearer to say something like "uploaded, processed and then returned to you" or something?

vscode/src/local-context/local-embeddings.ts
```diff
@@ -300,7 +305,7 @@ export class LocalEmbeddingsController
                 {
                     dir,
                     displayName: uriBasename(dir),
-                    providers: [{ kind: 'embeddings', state: 'indexing' }],
+                    providers: [{ kind: 'embeddings', state: 'indexing',useSourcegraphEmbeddings:this.useSourcegraphEmbeddings }],
```
dominiccooney, line 308: I'm not a fan of booleans, it's such an easy type to mess up in TypeScript/JavaScript, so unfriendly to English as a second language speakers when there's negations (not to neglect to omit it's also not friendly to native English speakers...) ... can we just have a type like `type EmbeddingsProvider = 'sourcegraph' | 'openai'` and pass something like `embeddingsProvider: 'openai'` instead? More explicit and totally separate at the type level from `''`, 0, `undefined`, `null`, ...

And have a helper down in the UI that takes the constant and produces the string with the right branding for the provider like `'openai'` &rArr; `'OpenAI'` or whatever...

dominiccooney (changes requested):
Exciting to see! Let's not pass booleans everywhere though.

dominiccooney:
One more bit of feedback: When the flag flips on, we should direct the embeddings controller to a different index directory. Everything will Just Work without having to generalize cody-engine to handle heterogeneous sets of indexes. (I've done a bit of WIP on that but it got hard preempted by JetBrains GA.)

vscode/src/local-context/local-embeddings.ts
rafax: Yes, that's exactly what I now have locally :) will fix

vscode/src/local-context/local-embeddings.ts
rafax: One reason was the config object is currently set up in the constructur, but constructor can't be async, and resolving a flag needs to be async (so, I resolve the flag in createLocalEmbeddingsController, and use it to create config)

rafax: But, your point about dropping testing* makes sense, will update

vscode/src/local-context/enhanced-context-status.test.ts
```diff
@@ -93,6 +95,7 @@ describe('ContextStatusAggregator', () => {
                     {
                         kind: 'embeddings',
                         state: 'unconsented',
+                        useSourcegraphEmbeddings:false,
```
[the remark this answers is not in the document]
rafax: Right, I had prettier configured locally, will reformat with biome.

vscode/src/local-context/local-embeddings.ts
```diff
@@ -112,8 +116,8 @@ export class LocalEmbeddingsController
         this.accessToken = config.accessToken || undefined
         this.endpointIsDotcom = isDotCom(config.serverEndpoint)
 
-        this.model = config.testingLocalEmbeddingsModel || 'openai/text-embedding-ada-002'
-        this.dimension = config.testingLocalEmbeddingsDimension || 1536
+        this.model = config.testingLocalEmbeddingsModel || useSourcegraphEmbeddings? 'sourcegraph/st-multi-qa-mpnet-base-dot-v1': 'openai/text-embedding-ada-002'
+        this.dimension = config.testingLocalEmbeddingsDimension || useSourcegraphEmbeddings? 768:1536
```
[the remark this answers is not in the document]
rafax: Right, removed testing altogether (using the feature flag)

vscode/src/local-context/local-embeddings.ts
rafax: I dropped all `testingFOO` properties.

vscode/src/local-context/enhanced-context-status.test.ts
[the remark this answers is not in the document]
rafax: Reformatted

vscode/src/local-context/local-embeddings.ts
rafax: Fixed

rafax:
> One more bit of feedback: When the flag flips on, we should direct the embeddings controller to a different index directory. Everything will Just Work without having to generalize cody-engine to handle heterogeneous sets of indexes. (I've done a bit of WIP on that but it got hard preempted by JetBrains GA.)

This is now implemented (with OpenAI getting the empty prefix).

vscode/src/local-context/local-embeddings.ts
```diff
@@ -7,45 +7,56 @@ import {
...
+export type LocalEmbeddingsConfig = Pick<ConfigurationWithAccessToken, 'serverEndpoint' | 'accessToken'>
 
-function getIndexLibraryPath(): FileURI {
+function getIndexLibraryPath(modelSuffix: string): FileURI {
     switch (process.platform) {
         case 'darwin':
-            return URI.file(`${process.env.HOME}/Library/Caches/com.sourcegraph.cody/embeddings`)
+            return URI.file(
+                `${process.env.HOME}/Library/Caches/com.sourcegraph.cody/embeddings` +
+                    (modelSuffix !== '' ? '/' + modelSuffix : '')
```
dominiccooney, line 48: You could simplify this a bit by reversing the branches.

```suggestion
                    (modelSuffix == '' ? '' : '/' + modelSuffix)
```

vscode/src/local-context/local-embeddings.ts
```diff
@@ -7,45 +7,56 @@ import {
...
-            return URI.file(`${process.env.HOME}/.cache/com.sourcegraph.cody/embeddings`)
+            return URI.file(
+                `${process.env.HOME}/.cache/com.sourcegraph.cody/embeddings` +
+                    (modelSuffix !== '' ? '/' + modelSuffix : '')
+            )
         case 'win32':
-            return URI.file(`${process.env.LOCALAPPDATA}\\com.sourcegraph.cody\\embeddings`)
+            return URI.file(
+                `${process.env.LOCALAPPDATA}\\com.sourcegraph.cody\\embeddings` +
+                    (modelSuffix !== '' ? '\\' + modelSuffix : '')
```
dominiccooney, line 58:
```suggestion
                    (modelSuffix === '' ? '' : '\\' + modelSuffix)
```

dominiccooney (changes requested):
Sorry for the slow response. 🙇 

Sharding the embeddings directory by model type looks good.

Removing the test override hooks has broken the VSCode e2e tests, specifically vscode/test/e2e/local-embeddings-test.ts which you can run with `pnpm -C vscode test:e2e`

Simplest thing is probably to put the test override hooks back... I earlier intended to suggest that you change the name and testing mode check so they are not "test only" overrides any more.

Otherwise, you can:
- Evolve the local-embeddings.test.ts forward however you want.
- Update the README in the embeddings crate about testing hooks since those settings aren't effective.
- Delete the settings in vscode/src/configuration.ts

vscode/src/local-context/local-embeddings.ts
```diff
@@ -111,14 +141,7 @@ export class LocalEmbeddingsController
...
         this.endpointIsDotcom = isDotCom(config.serverEndpoint)
-
-        this.model = config.testingLocalEmbeddingsModel || 'openai/text-embedding-ada-002'
-        this.dimension = config.testingLocalEmbeddingsDimension || 1536
-        this.endpoint =
-            config.testingLocalEmbeddingsEndpoint || 'https://cody-gateway.sourcegraph.com/v1/embeddings'
-        this.indexLibraryPath = config.testingLocalEmbeddingsIndexLibraryPath
-            ? URI.file(config.testingLocalEmbeddingsIndexLibraryPath)
-            : undefined
+        this.endpoint = 'https://cody-gateway.sourcegraph.com/v1/embeddings'
```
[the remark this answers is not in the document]
rafax: Done

rafax:
> Sorry for the slow response. 🙇
> 
> Sharding the embeddings directory by model type looks good.
> 
> Removing the test override hooks has broken the VSCode e2e tests, specifically vscode/test/e2e/local-embeddings-test.ts which you can run with `pnpm -C vscode test:e2e`
> 
> Simplest thing is probably to put the test override hooks back... I earlier intended to suggest that you change the name and testing mode check so they are not "test only" overrides any more.
> 
> Otherwise, you can:
> 
> * Evolve the local-embeddings.test.ts forward however you want.
> * Update the README in the embeddings crate about testing hooks since those settings aren't effective.
> * Delete the settings in vscode/src/configuration.ts

I brought the test overrides back (using a type instead of a bag of props), I assumed that these were only used for "human testing" of BFG integration, but supporting e2e tests without those would be tricky.

vscode/src/local-context/local-embeddings.ts
```diff
@@ -7,45 +7,56 @@ import {
...
+        config,
+        useSourcegraphEmbeddings ? sourcegraphModelConfig : openaiModelConfig
+    )
 }
 
-export type LocalEmbeddingsConfig = Pick<
-    ConfigurationWithAccessToken,
-    'serverEndpoint' | 'accessToken'
-> & {
-    testingLocalEmbeddingsModel: string | undefined
```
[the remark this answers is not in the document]
rafax: No longer removing the options

dominiccooney (approved):
Sorry for the slow response, thank you for keeping the tests green.

vscode/src/configuration.ts
```diff
@@ -165,19 +177,18 @@ export function getConfiguration(
...
+        testingModelConfig:
+            isTesting && hasValidLocalEmbeddingsConfig()
+                ? {
+                      model: getHiddenSetting<string>('testing.localEmbeddings.model'),
+                      dimension: getHiddenSetting<number>('testing.localEmbeddings.dimension'),
+                      endpoint: getHiddenSetting<string>('testing.localEmbeddings.endpoint'),
+                      indexPath: URI.file(
+                          getHiddenSetting<string>('testing.localEmbeddings.indexLibraryPath')
+                      ),
+                      provider: 'openai',
```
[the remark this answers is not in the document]
rafax: In practice, this is only used for showing the name of the API provider in the enhanced context selector, so I'll keep this as-is for now.

PR #3758, opened 2024-04-09
Chat: internal support for Google Gemini

abeatrix:
Context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1712686739448999

> NOTE: **This is for dev and testing purposes.** Feature is behind an undocumented configuration field.

Add experimental support for Google Gemini **when** API key is provided.

image

This pull request introduces the following changes:

1. Add API key and endpoint properties to ModelProvider:

- The ModelProvider class now has two new optional properties: apiKey and apiEndpoint.
- These properties allow specifying a custom API key and endpoint for a given AI model (not integrated in this PR)

2. Add addProviders method to ModelProvider:

- A new static method addProviders has been added to the ModelProvider class.
- This method allows adding new ModelProvider instances to the list of available providers.

3. Add getProviderByModel method to ModelProvider:

- A new static method getProviderByModel has been added to the ModelProvider class.
- This method retrieves the ModelProvider instance associated with a given model ID.

4. Add new chat client for Google Gemini:

- The SourcegraphNodeCompletionsClient has been updated to handle models starting with the prefix "google".
- If the model starts with "google", it calls the googleChatClient function with the provided parameters which process the requests with the API key provided.

5. Update CompletionParameters interface:

- The CompletionParameters interface now includes an optional modelAPI property.
- This property is an object that can hold the key (API key) and endpoint for a specific model.

These changes enable the integration of custom AI model providers by allowing the specification of API keys and endpoints. It also introduces a mechanism to add new model providers dynamically. This flexibility enhances the system's ability to work with different AI models and providers, enabling more customization and extensibility.

## Test plan

1. In your user settings JSON file, add the following settings and fill in the apiKey field:
```
  "cody.experimental.chatModels": [{
    "provider": "google",
    "model": "gemini-pro",
    "apiKey": "YOUR_API_KEY",
  },
  {
    "provider": "google",
    "model": "gemini-1.5-pro-latest",
    "apiKey": "YOUR_API_KEY",
    "tokens": 1000000
  }]
```
1. Reload Cody and open a new chat
1. In the dropdown menu, you should be able to see the models you just added via your user settings
1. Select one of the model and submit a question or run a command

image


lib/shared/src/sourcegraph-api/completions/types.ts
```diff
@@ -33,6 +33,14 @@ export interface CompletionParameters {
...
     topP?: number
     model?: string
+    /**
+     * For custom model that requires API key and a different endpoint.
+     * For example, Google Gemini Chat model requires a key.
+     */
+    modelAPI?: {
+        key?: string
+        endpoint?: string
+    }
```
philipp-spiess, line 43: Can we avoid adding this to our existing completion params type? Maybe the gemini specific content provider should fetch these values form the config directly but we do use this type as "the client side mapping of the sourcegraph API" which this addition would break.

vscode/src/models/utils.ts
```diff
@@ -46,3 +47,43 @@ export function syncModelProviders(authStatus: AuthStatus): void {
...
+    apiEndpoint?: string
+}
+
+/**
+ * NOTE: DotCom Pro Users only
+ *
+ * Gets an array of `ModelProvider` instances based on the configuration for experimental chat models.
+ * If the `cody.experimental.chatModels` setting is not configured or is empty, the function returns an empty array.
+ *
+ * @returns An array of `ModelProvider` instances for the configured chat models.
```
philipp-spiess, line 65: I wouldn't mark this "pro only" but instead "employee only". We really don't want people to start plugging in their own LLM keys in the client I think, do we?

philipp-spiess (approved):
I think this is cool to land for internal only testing but let's not advice this as any kind of support for Gemini since it requires users to plug in their own API keys in the client (and I know that if we mention this somewhere, we will have users that want to try it too and the experience is really not great this way)

lib/shared/src/sourcegraph-api/completions/types.ts
abeatrix: @philipp-spiess I've moved this to the google chat-client. Does this change look good to you?

philipp-spiess (approved):

daiaji:
I want to know if I will use Cody's reverse proxy or `generativelanguage.googleapis.com` as the endpoint to initiate requests after configuring my own apiKey in `cody.dev.models`?
If possible, I hope to increase the reverse proxy for Gemini. I have set up a reverse proxy for Gemini on Cloudflare Workers.

PR #3759, opened 2024-04-10
Edit (Test): Fix incorrect selection context

umpox:
## Description

Fixes an issue where Cody would not correctly use the selection context when generating a unit test

**Problem:**
- If a file was already found for the test location, we would attach it as `destinationFile` _before_ the task was created. At task creation, this `didReceiveNewFileRequest` function will modify the `selectionRange` and the `fixupFile` of the edit.

The problem with doing this at task creation, is that this will happen _before_ we build the prompt for the edit. This means that we will try to include the `selectionRange` text from the test file that we are targeting - which will always be empty.

**Solution**
We now perform the logic for `didReceiveNewFileRequest` __after__ the task has been created and prompt has been built. This should function the same but without the issues we have before. It also helps keep the test creation logic isolated to a single place

## Test plan

1. Open an existing test file
2. Open the associated file for those tests, and select a function
3. Click "Generate Unit Tests"
4. Check that it works as expected


abeatrix (commented):
The original issue is resolved on this branch 🎉 

But I now run into a new issue where the test will be inserted directly over my current selection on a file that doesn't have an existing test file:
image

PR #3763, opened 2024-04-10
Generate Unit Tests: Add code action and smart range expansion

umpox:
### Improved range
The test command now uses tree-sitter to get a syntactically correct range for the users intent, this should be pretty similar to current usage for now, but will be improved for different languages

### Code Action
Shows on any function symbol

image

## Test plan

Action:
1. Execute a unit test command on a relevant symbol via a code action
2. Check that it works correctly

Other:
1. Execute a unit test command within a function range by calling the command in a different way (e.g. right-click menu, sidebar)
2. Check works correctly


vscode/src/tree-sitter/query-tests/test-data/enclosing-function.snap.ts
```diff
@@ -57,28 +56,23 @@
...
   function arrowWrapper() {
+//^ start range.function[1]
       const arrow = (value: string) => {
-//                  ^ start range.function[1]
           console.log('hello')
 //            █
       }
-//    ^ end range.function[1]
   }
+//^ end range.function[1]
```
umpox, line 65: TODO: Check the tree-sitter query, this doesn't look quite right 🤔

umpox:
@abeatrix Think this is almost ready, couple of TODOs left:
- Add support to more languages than just JS (should be easy, just need to update the query to use `@symbol.name` too)
- Add logic to make `showHint` include if the function already has tests for it. If this is the case we can hide the prominent hint. Like we now do with documentation.
- Write tests for the `getTestableNode` query, similar to `enclosing-function.test.ts`
- Check in with @toolmantim on design thoughts around this

@toolmantim Check this out when you get chance! I think this is a good example of where visible contextual hints are useful. Only showing document if there is a docstring already, and only showing test if we don't already find an existing test. The latter is still TODO but should be straight forward

vscode/src/commands/GhostHintDecorator.ts
```diff
@@ -255,41 +279,99 @@ export class GhostHintDecorator implements vscode.Disposable {
...
+                         * "Test" flow
+                         * Display ghost text above the relevant symbol.
+                         */
+                        const precedingLine = Math.max(0, testableSymbol.startPosition.row - 1)
+                        if (
+                            this.activeDecorationRange &&
+                            this.activeDecorationRange.start.line !== precedingLine
+                        ) {
+                            this.clearGhostText(editor)
+                        }
```
dominiccooney, line 330: This pattern happens a bit. With a well named helper, the call site will be more readable.

vscode/src/commands/GhostHintDecorator.ts
```diff
@@ -255,41 +279,99 @@ export class GhostHintDecorator implements vscode.Disposable {
...
+                        return this.setThrottledGhostText(
+                            editor,
+                            new vscode.Position(precedingLine, Number.MAX_VALUE),
+                            'TestOrDocument',
```
dominiccooney, line 305: Let's de-dup these branches, for example, change the condition on 290 to be `(testableSymbol || documentableSymbol) && !isOnTypingLine` and then use a variable binding + switch or ternary to pick the label.

In general, if you wouldn't want the code to vary independently, then don't have two copies of that code lying around because someone will go and maintain one branch and not the other.

If it turns into Swiss cheese of `if (testableSymbol && !documentableSymbol)` cases, you can use the Strategy Pattern to have one test set a "policy" the rest of the code consults.

dominiccooney (dismissed):
Would love a test that shows us skip documentation for documented functions. You could even write the test for testable functions with an expected failure that we don't skip them...

@steveyegge FYI

abeatrix (approved):
other than what dom mentioned, the rest looks good to me (consider we worked on it live together XD)!

vscode/src/commands/GhostHintDecorator.ts
```diff
@@ -255,41 +279,99 @@ export class GhostHintDecorator implements vscode.Disposable {
...
+                    const isOnTypingLine = selection.active.line === this.lastLineTyped
+                    const testableSymbol = enabledFeatures.Test
+                        ? this.getThrottledTestableSymbol(editor.document, selection.active)
+                        : null
+                    const documentableSymbol = enabledFeatures.Document
+                        ? this.getThrottledDocumentableSymbol(editor.document, selection.active)
+                        : null
+
+                    if (testableSymbol && documentableSymbol && !isOnTypingLine) {
+                        /**
```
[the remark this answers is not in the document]
umpox: Moving to https://github.com/sourcegraph/cody/pull/3809 (hover/ghost additions)

Going to merge this just as a code action for now, as I think we need more design input on the ghost/hover behaviour and the code action is a good addition for this weeks release regardless.

vscode/src/commands/GhostHintDecorator.ts
umpox: Moving to https://github.com/sourcegraph/cody/pull/3809 (hover/ghost additions)

Going to merge this just as a code action for now, as I think we need more design input on the ghost/hover behaviour and the code action is a good addition for this weeks release regardless.

vscode/src/commands/GhostHintDecorator.ts
umpox: Moving to https://github.com/sourcegraph/cody/pull/3809 (hover/ghost additions)

Going to merge this just as a code action for now, as I think we need more design input on the ghost/hover behaviour and the code action is a good addition for this weeks release regardless.

umpox:
@dominiccooney FYI, going to get this merged just as a code action for now. I've dismissed your review as it's now stale (the ghost/hover logic is now in https://github.com/sourcegraph/cody/pull/3809 and we'll address this as part of the design changes too)

The current PR just adds the code action and the range expansion which will be great improvements to add to this weeks release

PR #3764, opened 2024-04-10
Ollama: update default context window

abeatrix:
The changes in this pull request remove the tokensToChars function from the fetch ollama function in lib/shared/src/models/utils.ts file. This function was previously used to convert the maxToken property of the ModelProvider class from a token count to a character count, but this was incorrect as the ModelProvider class has always taken the context window size in tokens directly.

The OLLAMA_DEFAULT_CONTEXT_WINDOW constant has been added to the lib/shared/src/ollama/index.ts file, which represents the default context window size used by the Ollama model, which is 2048 tokens ([source](https://github.com/ollama/ollama/blob/main/docs/faq.md#how-can-i-specify-the-context-window-size)).

## Test plan

simple string variable update

### DIFF
--- a/lib/shared/src/models/index.ts
+++ b/lib/shared/src/models/index.ts
@@ -23,8 +23,10 @@ export class ModelProvider {
         public readonly model: string,
         // The usage of the model, e.g. chat or edit.
         public readonly usage: ModelUsage[],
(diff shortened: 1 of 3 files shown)


philipp-spiess (approved):
Seems reasonable :) TY

PR #3767, opened 2024-04-10
Edit: display warning for large @-mentions

abeatrix:
The changes in this pull request address an issue where the total size of selected context items could exceed the character budget for the active model when importing through @-mentions.

The Edit Command Menu has been updated to:

1. Track the total size of the selected context items and compare it against the maximum character limit for the active model.
2. Display a "Large File Warning" label in the quick pick items if the size of a selected context item would cause the total to exceed the character budget.
3. Automatically remove any selected context items that are no longer included in the current user input.

Additionally, the `filterLargeFiles` function in `editor-context.ts` has been updated to store the file size in the `ContextItemFile` object, which is used in the `getInput` function to determine if a file is too large.

These changes help ensure that the user is not presented with context items that would cause the input to exceed the character limit, providing a better user experience and preventing potential issues with the editor's functionality.

## Test plan

image

image

## Follow-up

- [ ] Support @-mentions with range for Edit

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -13,6 +13,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 - Generate Unit Tests: Fixed an issue where Cody would generate tests for the wrong code in the file. [pull/3759](https://github.com/sourcegraph/cody/pull/3759)
 - Chat: Fixed an issue where changing the chat model did not update the token limit for the model. [pull/3762](https://github.com/sourcegraph/cody/pull/3762)
+- Edit: Large file warnings for @-mentions are now updated dynamically as you add or remove them. [pull/3767](https://github.com/sourcegraph/cody/pull/3767)
 
 ### Changed
 
--- a/vscode/src/edit/input/get-input.ts
+++ b/vscode/src/edit/input/get-input.ts
@@ -2,6 +2,7 @@ import {
     type ContextItem,
     type EditModel,
     type EventSource,
+    ModelProvider,
     displayLineRange,
     parseMentionQuery,
     scanForMentionTriggerInUserTextInput,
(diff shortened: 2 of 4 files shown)


dominiccooney (changes requested):
I'm concerned about the scalability of this approach. Because stat is cheap but tokenization is *not.* Once we move to token counting this may need to be very different.

So let's stop investing more in synchronous size checks, until we work out how to scale to (more) accurate token counting. Some ideas:

- Have a fast path for files which are definitely small enough, or definitely too large (for example, work out the smallest/largest byte -> token count expansion ahead of time and do a quick file size check to exclude tokenizing the largest files)
- Cache the token counts
- Do the token counts lazily, prioritizing items the user has already selected (which means we need to show errors *after* the user has chosen something) or that are visible in the quickpick (can we get what's visible from the VSCode QuickPick?)
- Get a smarter tokenizer, for example if we can stream, we could stop tokenizing as soon as we know something is too large

abeatrix:
> I'm concerned about the scalability of this approach. Because stat is cheap but tokenization is _not._ Once we move to token counting this may need to be very different.
> 
> So let's stop investing more in synchronous size checks, until we work out how to scale to (more) accurate token counting. Some ideas:
> 
> * Have a fast path for files which are definitely small enough, or definitely too large (for example, work out the smallest/largest byte -> token count expansion ahead of time and do a quick file size check to exclude tokenizing the largest files)
> * Cache the token counts
> * Do the token counts lazily, prioritizing items the user has already selected (which means we need to show errors _after_ the user has chosen something) or that are visible in the quickpick (can we get what's visible from the VSCode QuickPick?)
> * Get a smarter tokenizer, for example if we can stream, we could stop tokenizing as soon as we know something is too large

This is a follow-up on https://github.com/sourcegraph/cody/pull/3619#issuecomment-2031976331
I wanted to update this here first, so that I can standardize it in my WIP PR for expanding the context window and using token as size.
But I just realized that was asking to add support for entering range and not displaying large file warning 🤣

dominiccooney:
@abeatrix 

> This is a follow-up on https://github.com/sourcegraph/cody/pull/3619#issuecomment-2031976331

Makes sense. It threw me for a loop in the context of this token counting change.

dominiccooney (approved):

umpox (approved):
Nice!

PR #3769, opened 2024-04-10
Agent - add support for proxies 

chwarwick:
This adds support for HTTP(s) and SOCKS proxies by using the `proxy-agent` library which determines if a proxy is needed and chooses the correct one by examining the `http_proxy` and `https_proxy` variables.  The `proxy-agent` library is the original library that the `vscode-proxy-agent` originates from.

part of https://github.com/sourcegraph/jetbrains/issues/1127

## Test plan
Verified that chat & autocomplete continue to work in VS Code
Modified to JetBrains to launch agent with: 
```
processBuilder.environment()["HTTP_PROXY"] = "socks5://127.0.0.1:8889"
processBuilder.environment()["HTTPS_PROXY"] = "socks5://127.0.0.1:8889"
```

and then 
```
processBuilder.environment()["HTTP_PROXY"] = "http://127.0.0.1:8888"
processBuilder.environment()["HTTPS_PROXY"] = "http://127.0.0.1:8888"
```
And verified that the requests were all routed though the proxies.
Screenshot 2024-04-10 at 3 42 04 PM

In VSCode extension debugger, paste https://github.com/microsoft/vscode/issues/173861#issue-1576626788 into the global namespace somewhere and check it logs `did reuse socket -> true`

### DIFF
--- a/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
+++ b/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
@@ -111,7 +111,7 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
                     // So we can send requests to the Sourcegraph local development instance, which has an incompatible cert.
                     rejectUnauthorized:
                         process.env.NODE_TLS_REJECT_UNAUTHORIZED !== '0' && !this.config.debugEnable,
-                    agent: agent.current?.(url),
+                    agent: agent.current?.(),
                 },
                 (res: http.IncomingMessage) => {
                     const { 'set-cookie': _setCookie, ...safeHeaders } = res.headers
--- a/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
+++ b/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
@@ -8,9 +8,12 @@ import { RateLimitError } from '../errors'
 import { customUserAgent } from '../graphql/client'
 import { toPartialUtf8String } from '../utils'
 
+
+import { agent } from '../../fetch'
 import { googleChatClient } from '../../llm-providers/google/chat-client'
 import { groqChatClient } from '../../llm-providers/groq/chat-client'
 import { ollamaChatClient } from '../../llm-providers/ollama/chat-client'
+
 import { getTraceparentHeaders, recordErrorToSpan, tracer } from '../../tracing'
 import { SourcegraphCompletionsClient } from './client'
 import { parseEvents } from './parse'
@@ -83,7 +86,6 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
 
             // Text which has not been decoded as a server-sent event (SSE)
             let bufferText = ''
-
             const request = requestFn(
                 url,
                 {
@@ -103,6 +105,7 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
                     // So we can send requests to the Sourcegraph local development instance, which has an incompatible cert.
                     rejectUnauthorized:
                         process.env.NODE_TLS_REJECT_UNAUTHORIZED !== '0' && !this.config.debugEnable,
+                    agent: agent.current?.(url)
                 },
                 (res: http.IncomingMessage) => {
                     const { 'set-cookie': _setCookie, ...safeHeaders } = res.headers
--- a/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
+++ b/lib/shared/src/sourcegraph-api/completions/nodeClient.ts
@@ -10,10 +10,7 @@ import { toPartialUtf8String } from '../utils'
 
 <<<<<<< HEAD
 
-import { agent } from '../../fetch'
-import { googleChatClient } from '../../llm-providers/google/chat-client'
 import { groqChatClient } from '../../llm-providers/groq/chat-client'
-import { ollamaChatClient } from '../../llm-providers/ollama/chat-client'
 
 =======
 import { agent } from '../../fetch'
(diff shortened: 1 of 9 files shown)


chwarwick:
There is a potential issue here with the symf and cody-engine downloads that use axios.  I think we likely would need to also set the http & https agents in the axios config to the agent.current

lib/shared/src/sourcegraph-api/completions/nodeClient.ts
```diff
@@ -97,6 +97,7 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
                     // So we can send requests to the Sourcegraph local development instance, which has an incompatible cert.
                     rejectUnauthorized:
                         process.env.NODE_TLS_REJECT_UNAUTHORIZED !== '0' && !this.config.debugEnable,
+                    agent: agent.current?.(url),
```
chwarwick, line 100: this line I'm a little concerned about, I'm not sure if this should actually be conditional and only override the https agent when running in cody-agent.

kalanchan:
🥹 the final gift from Chris

lib/shared/src/sourcegraph-api/completions/nodeClient.ts
dominiccooney: VSCode does extensive proxy shimming magic when the extension is running in VSCode.

vscode/src/fetch.node.ts
```diff
@@ -45,12 +31,13 @@ export function setCustomAgent(
 }
 
 export function initializeNetworkAgent(): void {
-    httpAgent = new http.Agent({ keepAlive: true, keepAliveMsecs: 60000 })
-    httpsAgent = new https.Agent({
-        ...https.globalAgent.options,
```
dominiccooney, line 50: What is this and why don't we care any more?

dominiccooney (commented):
Would be great to add a test where we set those variables, stand up a mock proxy and see that it is getting requests.

vscode/src/fetch.node.ts
chwarwick: Ohh thanks I might have missed that I'll have to test it tomorrow, the purpose of `...https.globalAgent.options` was so that it picked up any self signed certs that were detected and added to the global agent during initialization.

vscode/src/fetch.node.ts
```diff
@@ -67,7 +54,7 @@ export function initializeNetworkAgent(): void {
      */
     try {
         const PacProxyAgent =
-            (globalThis as any)?.[nodeModules]?.[proxyAgentPath]?.[pacProxyAgent] ?? undefined
+            (globalThis as any)?.[nodeModules]?.[proxyAgentPath]?.[pacProxyAgent] ?? customAgent
```
chwarwick, line 57: I think I should revert this line, this branch should only be patching the vs code proxy logic.

lib/shared/src/sourcegraph-api/completions/nodeClient.ts
philipp-spiess: I think it makes sense to do this here for vs code too, this way we will at least have the same `agent` between completions and chat (which right now I reckon we don't? stupid oversight I suppose). My hunch is that if this creates an issue we rather have the same symptom on any http call and not just _some_.

vscode/src/fetch.node.ts
```diff
@@ -1,13 +1,13 @@
-import http from 'node:http'
+import type http from 'node:http'
 import https from 'node:https'
 
-import { SocksProxyAgent } from 'socks-proxy-agent'
```
philipp-spiess, line 4: can we rm this package now from `package.json`?

vscode/src/fetch.node.ts
philipp-spiess: @chwarwick this is now handled via `ca: https.globalAgent.options.ca` below, right?

vscode/src/fetch.node.ts
```diff
@@ -45,12 +33,14 @@ export function setCustomAgent(
...
+    proxyAgent.connectOpts = {
         keepAliveMsecs: 60000,
-    })
+        keepAlive: true,
+        ca: https.globalAgent.options.ca
+    }
+
     const customAgent = setCustomAgent(getConfiguration())
 
     /**
```
philipp-spiess, line 46: Can you check if the keepalive still works in the vs code client? it's a bit awkward but if you paste this into the global namespace somewhere it should log  `did reuse socket -> true` ideally: https://github.com/microsoft/vscode/issues/173861#issue-1576626788

philipp-spiess (approved):
Looks good to me overall.

vscode/src/fetch.node.ts
chwarwick: @philipp-spiess  Yes i updated that once I realized I missed that, likely will change again this am to spread the options from the globalAgent just in case there are other cases of settings.

vscode/src/fetch.node.ts
chwarwick: Thanks was looking for a way to test that, and it caught an issue with the config

Now I get:
```
first request connection header -> undefined
extensionHostProcess.js:146

did reuse socket -> true
```

philipp-spiess: Awesome!!

philipp-spiess:
> I'd suggest removing axios entirely to have 1 less set of networking logistics to deal with.

Huh yes big fan of not using axios and just use `fetch` instead

vscode/src/fetch.node.ts
dominiccooney: That snippet is a nice addition to the test plan.

chwarwick: I tried to turn it into a test it's not the best since it's checking the agent version and not all the patched vs code stuff, but it's something.

philipp-spiess (approved):
The test looks splendid. It doesn't run in the VS Code extension runtime but it's still much better than not having any testing. Thank you!

kalanchan:
@arafatkatze ready for you to test and take over once it's merged! 

cc @cbart

arafatkatze:
Thanks @kalanchan I am on it.

arafatkatze:
NOTE: This PR was tested for both Socks proxy and http and https proxy on Vscode /Jetbrains combo. 

For now, this PR does everything it claims to do. 

The next step is to test it on Different OS and environments

PR #3624, opened 2024-03-30
feat: "Report Issue" command for filing bug reports inside VS Code

abeatrix:
New "Report Issue" command to allow users to file Github issues with the editor through the built-in `Issue Reporter`.

image

This commit adds a new command to the *VS Code Issue Reporter* and the *Cody support sidebar* to allow users to easily report issues to us with pre-filled system info. 
- The "Report Issue" command opens the *VS Code Issue Reporter* with `Cody version`, `VS Code Version` and `Host` pre-filled.

Summary of Changes:
- Updated our bug link for filing GitHub issue with relevant labels, project, and template for bug reports.
- Added a new command `cody.debug.reportIssue` under `contributes.commands` to register the "Report Issue" command.
- Added a new menu item under `view/item/context` to display the "Report Issue" command in the Cody support sidebar.

- Updated `vscode/src/main.ts`:
  - Imported the `openCodyIssueReporter` function from `../webviews/utils/reportIssue`.
  - Registered the `cody.debug.reportIssue` command to execute the `openCodyIssueReporter` function when invoked.

## Test plan

image

1. Open the Cody extension in VS Code.
2. Execute the "Cody Debug: Report Issue" command from the command palette.
3. Verify that a new window opens with Cody Version and VS Code Version pre-filled in the body.
image
4. Typing in the title would show you if a similar issues have been reported
image
5. Ability to include system information with the issue
image
6. Submit the bug report by clicking on the `Create on Github`button to confirm your issue has been created inside the cody repository
7. Verify the `Report Issue` option is also available in the support sidebar `...` menu
image

PR #3770, opened 2024-04-10
update labeler to auto add bug labels

kalanchan:
helps organize the report issue command introduced in https://github.com/sourcegraph/cody/pull/3624. VS Code report issue doesn't allow us to add labels, so this action will backfill any issues that have `bug:` in it

## Test plan
N/A


dominiccooney (approved):

varungandhi-src:
I've modified the workflow file to also consider the issue title:

https://github.com/github/issue-labeler?tab=readme-ov-file#example-including-the-issue-title-in-the-regex-target

kalanchan:
thanks for the help @varungandhi-src !

PR #3771, opened 2024-04-11
VS Code: implement `ContextFiltersProvider.isUriAllowed`

valerybugakov:
## Context

- Adds `ContextFiltersProvider.isUriAllowed()` basic implementation to start with. It uses several methods to get the repo name for the provided URI:
  - VS Code Git extension if available.
  - Git CLI if available and the extension is running in the node.js environment.
  - File system tree-walk to find the `.git` folder.
- Subtasks:
  - [x] Migrate to `re2-wasm`.
  - [x] Improve `ContextFiltersProvider.isPathAllowed()` unit tests.
  - [x] Initialize `RepoNameResolver` based on the available APIs in `extension.node.ts` and `extension.common.ts`.
  - [x] Basic caching for `RepoNameResolver.getRepoNameFromWorkspaceUri()`.
  - [x] Unit tests for repo name getters.
  - [x] Add Git submodules support.
- Closes subtasks from https://github.com/sourcegraph/cody/issues/3641
    - Write a function that takes file URI -> repo name based on the environment available.
    - Write a function that takes file URI -> do/don't ignore.
- Performance benchmarking and optimization will be implemented in a follow-up PR.
- Builds on top of backend changes https://github.com/sourcegraph/sourcegraph/pull/61101

## Test plan

CI and new unit tests


vscode/src/repository/repositoryHelpers.ts
```diff
@@ -77,21 +82,133 @@ export async function gitAPIinit(): Promise<vscode.Disposable> {
...
     return undefined
 }
 
-// HELPER FUNCTIONS
-function getCodebaseNameFromGitRepo(repository: Repository): string | undefined {
-    const remoteUrl = repository.state.remotes[0]?.pushUrl || repository.state.remotes[0]?.fetchUrl
-    if (!remoteUrl) {
+type FsPath = string
+type RepoName = string
+const fsPathToRepoNameCache = new LRUCache<FsPath, RepoName>({ max: 1000 })
```
valerybugakov, line 97: @kalanchan, @dominiccooney, do we need to support nested repositories on the feature release? If not, we can get a huge performance win by checking if file paths are inside the already resolved repos. And, we can work separately on supporting nested repos after the GA release.

dominiccooney: I assume that we do, because git submodules are a common workflow.

CC @willdollman because this is interesting from a security correctness point of view; CC @aramaraju because we need to document really clearly what we're doing so admins have the right expectation. (We'll help you document these parts when we get there, but you shouldn't ship until we close this loop.)

## TL;DR:
- Do we know this is a performance problem? If not, 😎  (Specifically... Add a call to this *per keystroke*, what's the 99th %-ile latency on the result? With symlinks in the path? On a remote filesystem?)
- Can we get st_ino/st_mtime on macOS, Linux stat and FILE_ID_INFO VolumeSerialNumber+FileId/FILE_BASIC_INFO ChangeTime on Windows w/ GetFileInformationByHandle in node? How fast are they?
- Can we build a good enough cache on top of that exploiting the access pattern, directory structure, and using a mix of cheap eager checks, (short) time-based expiries, and fail safes?

## Long version

*We need to reconcile ourselves to the fact that time of check, time of use is a fundamental issue here.* Unlike the Sourcegraph server side, the filesystem is a shared resource and we're not the only actor. So *anything* we do will not be water tight. We just need it to be known and unsurprising.

First question, @valerybugakov, how slow is the solution you have? The filesystem itself, probably the git extension, have caches already. (Note, however, that we wrote our own simulator for the git extension in agent, and that may scale differently/need work.)

IF we have a performance problem, here's some pointers:
- I see a lot of bugs in VSCode/extensions about the scalability of the filesystem watcher, so while we can try it, we should treat it as suspect and use it little if at all. That means a cache we manage and it is going to be a balancing act: eager checks which handle common cases cheaply enough, accepting some imprecision by using a cached result for a short time, precaching, and a heavy focus on code correctness so we never use results which are "CEO gets a phone call" old.
- The structure of the problem is on our side. We will generally hammer a very small set of open files. Directories have structure, that is, `a/b/c/d` and `a/b/e/f` checking `f` must scrutinize `e` but after that can reuse results for `d` up to `a/b`. The question is, how stale can the results be, but still be acceptable?
- You can use an inode number (file GUID on Windows) and mtime to see if something is the same file/dir, and whether something may have changed, more cheaply than searching all the parent directories for `.git` specifically. (S_NOCMTIME is an obscure thing on Linux, but we can add document limitations like this. Not sure about Windows.)
- Hard cases are ones where you have `a/b/c/d`, find a git root at `a`, then someone sets up a git root at `a/b` and the result for `a/b/c/d` becomes stale. If we can afford to stat `d` (and the OS level caching should have our back here) then we'd only get confused in cases like "user sets up a git root at `a/b` and then *moves the same file* that was at `a/b/c/d` back into it at the same path." Which should be good enough, if backed up by (short) time-based cache expiry.

FWIW, this is the real meat of this issue. It is detailed work. It is important to get right. You don't need to rush this. Draw on all of your teammates security, client hacking and product experience to make something great here.

vscode/src/repository/repositoryHelpers.ts
```diff
@@ -77,21 +82,133 @@ export async function gitAPIinit(): Promise<vscode.Disposable> {
...
+    remoteName = 'origin'
+): Promise<string | undefined> {
+    if (!isFileURI(uri)) {
+        return undefined
+    }
+
+    const gitConfigUri = vscode.Uri.joinPath(uri, '.git', 'config')
+
+    try {
+        const raw = await vscode.workspace.fs.readFile(gitConfigUri)
```
dominiccooney, line 201: Really want high fidelity with the git extension here because the failure mode will be really confusing if we sometimes use one repo ID, some times another. IIRC we use the default push URL, which may not be called `origin`. Let's add some tests here.

dominiccooney (commented):
Great start. Some detailed notes inline.

Elsewhere I saw you asking about which SCMs this must support for GA. @aramaraju clarified that customers using src-serve-git etc. to get svn repositories indexed will fall into the "fail closed (deny)" mode. That is:

- Repos are indexed with src-serve-git and get some name, X.
- Admin sets up a policy.
- User has an SVN checkout or something. We can't find a git repository for it and treat it like they have opened any file.
- We see a policy plus file without an identifiable repo and will default to deny.

aramaraju:
Agreed that the allow only setup will default these repos to be ignored. And I'm comfortable with shipping this without any consideration or specific treatment for SVN for GA, and iterating on that later, when and if there are enough customers complaining about it.

vscode/src/context/context-filters.provider.test.ts
```diff
@@ -20,140 +23,189 @@ describe('ContextFiltersProvider', () => {
...
+        })
 
-        expect(provider.isPathAllowed('github.com/sourcegraph/whatever', 'src/main.ts')).toBe(false)
-    })
+        it('matches file path patterns correctly', async () => {
+            await initProviderWithContextFilters({
+                include: [
+                    {
+                        repoNamePattern: '^github\\.com\\/sourcegraph\\/.*',
+                        filePathPatterns: ['.*\\.ts$'],
```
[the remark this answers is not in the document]
valerybugakov: No, see [my comment](https://github.com/sourcegraph/cody/pull/3747#discussion_r1558910599) in the previous PR:

> We do not have this field on the backend yet, and it might have a different name format, but I added it anyway while I was at it. It's ignored if not provided, and we can update the details once this is supported by the Sourcegraph API. I'm happy to remove it if folks see issues with adding it early.

vscode/src/main.ts
```diff
@@ -175,6 +177,19 @@ const register = async (
         symfRunner,
     } = await configureExternalServices(context, initialConfig, platform, authProvider)
 
+    // For local debugging purposes
```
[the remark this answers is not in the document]
valerybugakov: This code will be removed before the PR officially goes into review.

vscode/src/context/context-filters.provider.test.ts
```diff
@@ -20,140 +23,189 @@ describe('ContextFiltersProvider', () => {
...
         })
 
-        expect(provider.isPathAllowed('github.com/sourcegraph/cody', 'src/main.ts')).toBe(true)
-        expect(provider.isPathAllowed('github.com/sourcegraph/cody', 'src/main.js')).toBe(false)
-    })
+        it('handles invalid regular expressions gracefully', async () => {
+            await initProviderWithContextFilters({
+                include: [
+                    { repoNamePattern: '^github\\.com\\/sourcegraph\\/.*' },
+                    { repoNamePattern: '(invalid_regex' },
```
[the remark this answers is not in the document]
valerybugakov: Raised this [here](https://docs.google.com/document/d/1PMeJHvY0ia2nMewFvs4vrFcmDbRDw9kSyljil-K8Pt8/edit?disco=AAABKQTRR3s).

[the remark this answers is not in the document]
valerybugakov: The backend validates regex syntax on site-config save, so it's better to ignore all the rules in case of an error. This should not happen if the service functions properly.

vscode/src/repository/repositoryHelpers.ts
valerybugakov: Thanks for the detailed comment, Dominic! 

> how slow is the solution you have?

It's unclear how fast repo-name getters are because we have no means to measure that. To address that, we need a solid integration tests suite to assess performance on different platforms in different scenarios (VS Code Git extension is available, Git CLI is available, or nothing is available). Then, if needed, we can optimize each of the repo-name getters. This work can be parallelized.

Based on that, I propose splitting the remaining work into multiple milestones that can be shipped in separate PRs. Each repo-name-getter should be profiled and optimized if needed. I'll update the parent issue with more details tomorrow.

dominiccooney: - To get started, best case "works on my machine" numbers for one case are fine. We have to clarify how slow is too slow, is autocomplete the most latency sensitive case, do you have a sense of what it could stand?
- Regarding checking in and working incrementally, sounds fine if you're confident enough that the design *can* be fast.

vscode/src/repository/repositoryHelpers.ts
valerybugakov: Great call! Changed the logic to be consistent with the Git extension getter and added tests.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -953,5 +953,5 @@ export type LogEventMode =
     | 'all' // log to both dotcom AND the connected instance
 
 function hasOutdatedAPIErrorMessages(error: Error): boolean {
-    return error.message.includes('FetchError') || error.message.includes('Cannot query field')
+    return error.message.includes('Cannot query field')
```
valerybugakov, line 956: Follow-up for this comment: https://github.com/sourcegraph/cody/pull/3747#discussion_r1559163308

vscode/src/repository/repositoryHelpers.ts
valerybugakov: Sounds good. I updated the parent issue with more subtasks.

valerybugakov: I have done basic benchmarking on my machine with autocomplete in VS Code. A check was added [to the context mixer in the same place where we have the Cody Ignore PLG check](https://github.com/sourcegraph/cody/blob/9de35c359c4ba9bdf5348f062744e92ce38b26f5/vscode/src/completions/context/context-mixer.ts#L96). The approximate number of URIs to check is 20. The total number of time spent checking all the snippets is:

- If the Git extension is available <10ms for the first URI check, <0.1ms for subsequent calls.
- If the Git CLi is available <150ms for the first URI check, <0.1ms for subsequent calls.
- If we use the file system tree walk <10ms for the first URI check, <0.1ms for subsequent calls.

Based on these numbers, I'm going to tweak the logic in this PR to prefer the file system tree walk over Git CLI. As a follow-up, I'll do more benchmarking for different scenarios, such as nested git submodules, huge repos like Chromium, and agent-based file system.

valerybugakov:
> svn repositories indexed will fall into the "fail closed (deny)" mode

Excellent! The current logic already covers this. We exclude all the URIs for which we cannot find the repo name.

vscode/package.json
```diff
@@ -1377,4 +1332,4 @@
     "progress": "^2.0.3",
     "yaml": "^2.3.4"
   }
-}
\ No newline at end of file
+}
```
philipp-spiess, line 1335: We should figure out why the format of package.json constantly changes between folks 😆

vscode/src/repository/git-extension-api.ts
```diff
@@ -77,21 +77,17 @@ export async function gitAPIinit(): Promise<vscode.Disposable> {
  */
 export function getCodebaseFromWorkspaceUri(uri: vscode.Uri): string | undefined {
```
philipp-spiess, line 78: Assuming there is only one "codebase" for a specific workspace uri might be too limiting if we think about nested checkouts or the possibility to open your whole `/dev` folder in VS Code. I understand this is legacy but maybe we want to prefix the function name with something so it's clear this is not the recommended way anymore?

vscode/src/context/context-filters-provider.ts
```diff
@@ -92,6 +99,20 @@ export class ContextFiltersProvider implements vscode.Disposable {
         return isAllowed
     }
 
+    public async isUriAllowed(uri: vscode.Uri): Promise<boolean> {
+        if (!isFileURI(uri)) {
+            return false
+        }
```
philipp-spiess, line 105: Non-file URIs might mean stuff like the output console or terminal. Is it expected behavior that turning on cody ignore will prevent cody from working in these cases?

vscode/src/context/context-filters-provider.ts
```diff
@@ -92,6 +99,20 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
+
+        const repoName = await wrapInActiveSpan('repoNameResolver.getRepoNameFromWorkspaceUri', () =>
+            repoNameResolver.getRepoNameFromWorkspaceUri(uri)
+        )
+
+        const relativePath = vscode.workspace.asRelativePath(uri, false)
+
+        return repoName ? this.isPathAllowed(repoName, relativePath) : false
```
philipp-spiess, line 113: Should the path be relative to the VCS root and not the workspace here? I don't understand the expectations of path from reading `isPathAllowed` since we already determine a repo name in the line above, we might want to make sure we refer to this path as being _inside_ this repo? 

It's a bit unclear what the contract of `isPathAllowed` is. Does it handle nested submodules? Maybe it needs renaming `isPathAllowedInRepo` or so? 🤔

vscode/src/context/context-filters-provider.ts
```diff
@@ -120,10 +141,16 @@ function checkFilter(
...
-        ? item.filePathPatterns.map(pattern => new RE2(pattern))
+        ? item.filePathPatterns.map(pattern => new RE2(pattern, 'u'))
         : undefined
 
     return { repoNamePattern, filePathPatterns }
 }
+
+/**
+ * A a singleton instance of the `ContextFiltersProvider` class.
+ * `contextFiltersProvider.init` should be called on extension activation.
```
philipp-spiess, line 154:
```suggestion
 * `contextFiltersProvider.init` should be called and awaited on extension activation.
```

I suppose we also need to await it so we know the cache is properly seeded and we don't start requests before that's the case? 🦭

vscode/src/context/context-filters.provider.test.ts
philipp-spiess, line 1: Minor nit but the implementation is called `context-filters-provider` and not `context-filters.provider` so I think we should rename this file to `context-filters-provider.test.ts`

vscode/src/context/context-filters.provider.test.ts
```diff
@@ -20,140 +23,315 @@ describe('ContextFiltersProvider', () => {
...
+        it('returns `false` if repo name is not found', async () => {
+            await initProviderWithContextFilters({
+                include: [{ repoNamePattern: '^github\\.com\\/sourcegraph\\/cody' }],
+                exclude: [{ repoNamePattern: '^github\\.com\\/sourcegraph\\/sourcegraph' }],
+            })
+
+            const uri = getTestURI({ repoName: 'cody', filePath: 'foo/bar.ts' })
+            vi.spyOn(repoNameResolver, 'getRepoNameFromWorkspaceUri').mockResolvedValue(undefined)
+            expect(await provider.isUriAllowed(uri)).toBe(false)
+        })
```
philipp-spiess, line 335: Should we document the behavior around non file URIs?

vscode/src/repository/repo-name-resolver.test.ts
```diff
@@ -0,0 +1,146 @@
...
+            bare = false
+            logallrefupdates = true
+            ignorecase = true
+        [remote "origin"]
+            url = https://github.com/sourcegraph/cody
+            fetch = +refs/heads/*:refs/remotes/origin/*
+        [branch "main"]
+            remote = origin
+            merge = refs/heads/main
+        `
```
philipp-spiess, line 22: Not important but dedent will trim the text (if I’m not completely wrong) so I think this is a little nicer to read :)
```suggestion
        const mockGitConfig = dedent`
        [core]
            repositoryformatversion = 0
            filemode = true
            bare = false
            logallrefupdates = true
            ignorecase = true
        [remote "origin"]
            url = https://github.com/sourcegraph/cody
            fetch = +refs/heads/*:refs/remotes/origin/*
        [branch "main"]
            remote = origin
            merge = refs/heads/main
        `
```

vscode/src/repository/repo-name-resolver.test.ts
```diff
@@ -0,0 +1,146 @@
...
+            .mockResolvedValueOnce(textEncoder.encode(mockGitConfig))
+
+        const uri = URI.file('path/to/file/foo.ts')
+        const remoteUrl = await gitRemoteUrlFromTreeWalk(uri)
+
+        expect(readFileMock).toBeCalledTimes(3)
+        expect(remoteUrl).toBe('https://github.com/username/yourproject.git')
+    })
+
+    it('prioritizes `pushUrl` over `url` and `fetchUrl`', async () => {
```
philipp-spiess, line 80: This does not seem like a safe assumption. I wonder if we shouldn't handle all remotes instead (i.e if I git pull a private repo, cody ignore should work regardless of wether I push to a fork of it, but maybe this is a naive assumption?)

vscode/src/repository/repo-name-resolver.test.ts
```diff
@@ -0,0 +1,146 @@
+import { describe, expect, it, vi } from 'vitest'
+import * as vscode from 'vscode'
+import { URI } from 'vscode-uri'
+
+import dedent from 'dedent'
+import { gitRemoteUrlFromTreeWalk } from './repo-name-resolver'
+
+describe('gitRemoteUrlFromTreeWalk', () => {
```
philipp-spiess, line 8: I’m surprised to not see submodules being handled here. Am I missing something? 🤔 

I checked with the `bfg-evaluation` repo and in the submodules there does not exist a `.git` folder but .git is rather a file that links to path of the config:


```bash
α repos/zoekt 🐙(e068116)
cat .git
gitdir: ../../.git/modules/repos/zoekt
```

```bash
α repos/zoekt 🐙(e068116)
cat ../../.git/modules/repos/zoekt/config
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
	ignorecase = true
	precomposeunicode = true
	worktree = ../../../../repos/zoekt
[remote "origin"]
	url = https://github.com/sourcegraph/zoekt
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
	remote = origin
	merge = refs/heads/main
```

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Gets the codebase name from a workspace / file URI.
+     *
+     * Checks if the Git API is initialized, initializes it if not.
+     * If found, gets the codebase name from the repository.
+     * If not found, attempts to use Git CLI to get the codebase name (in node.js environment only).
+     * if not found, walks the file system upwards until it finds a `.git` folder.
+     * If not found, returns `undefined`.
+     */
+    public async getRepoNameFromWorkspaceUri(uri: vscode.Uri): Promise<string | undefined> {
```
philipp-spiess, line 35: This again assumes that there's only one repo name inside a workspace which I don't think is true (especially not for nested checkouts via submodules).

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -0,0 +1,125 @@
...
+                    remoteOriginUrl = await getter(uri)
+
+                    if (remoteOriginUrl) {
+                        break
+                    }
+                }
+            }
+
+            if (remoteOriginUrl) {
+                const repoName = convertGitCloneURLToCodebaseName(remoteOriginUrl) || undefined
```
philipp-spiess, line 62: ⚠️  **This is unsafe and we should not try to deduce the Sourcegraph instance name from the remote URL like this.** This works only in limited cases (e.g. on dotcom where we name the Sourcegraph repo exactly like the GitHub repo) but I know that especially in enterprise settings, this is not the case. Check out the site config option: [cloneURLToRepositoryName](https://docs.sourcegraph.com/admin/config/site_config#git-cloneURLToRepositoryName)

I think the correct approach is to use a GraphQL API on the backend to resolve the remotes to the Sourcegraph repository name. Here's an example from the repo name: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@70ef23e79c4b3b83295440231e1d311c52e1cb67/-/blob/client/browser/src/shared/code-hosts/gitlab/codeHost.ts?L292-296

This would also enable us to properly handle things like non-git repos with known git mirrors.

cc @taras-yemets who should have some context here as well

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -0,0 +1,125 @@
...
+ * Walks the tree from the current directory to find the `.git` folder and
+ * extracts remote URL. Prioritizes `pushurl` over `fetchurl` and `url` defined
+ * in `.git/config`.
+ */
+export async function gitRemoteUrlFromTreeWalk(uri: vscode.Uri): Promise<string | undefined> {
+    if (!isFileURI(uri)) {
+        return undefined
+    }
+
+    const gitConfigUri = vscode.Uri.joinPath(uri, '.git', 'config')
```
philipp-spiess, line 86: Check out my comment in the test file but this won't work for git submodules I think

philipp-spiess (changes requested):
I think there are a few shortcomings that should be easily addressed. Listing in the number of importance:

- Let's not try to deduce the Sourcegraph repo name from the git URI. [Use the GraphQL API instead.](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@70ef23e79c4b3b83295440231e1d311c52e1cb67/-/blob/client/browser/src/shared/code-hosts/gitlab/codeHost.ts?L292-296)
- Let's add support for git submodules by checking if `.git` is a file and following it's link (this does not appear to be a symlink in my setup).
- Let's break up with the connotation of "one upstream remote" per directory/workspace. I think all of the cases where we only pick one would make it more likely for us to overlook the important one. I'd feel much better if we consider all possible upstream URLs.
- Define what should happen with non-file URIs. If we want to ignore them by default that's alright but we should document this behavior with a unit test.

vscode/src/repository/repo-name-resolver.ts
taras-yemets: +1 to using GraphQL API to get repo by clone URL

vscode/src/extension.common.ts
```diff
@@ -43,6 +44,7 @@ export interface PlatformContext {
         | Constructor<typeof SourcegraphNodeCompletionsClient>
     createSentryService?: (config: Pick<ConfigurationWithAccessToken, 'serverEndpoint'>) => SentryService
     createOpenTelemetryService?: (config: OpenTelemetryServiceConfig) => OpenTelemetryService
+    getRemoteUrlGetters?: () => RemoteUrlGetter[]
```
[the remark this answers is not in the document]
valerybugakov: We try remote URL getters one by one until one of them returns the result. For example, if the Git extension is unavailable, we fall back to tree-walk.

vscode/src/repository/repo-name-getter.node.ts
```diff
@@ -0,0 +1,49 @@
...
+import util from 'node:util'
+import type * as vscode from 'vscode'
+
+import { isFileURI } from '@sourcegraph/cody-shared'
+
+import { logDebug } from '../log'
+
+import { pathFunctionsForURI } from '@sourcegraph/cody-shared/src/common/path'
+
+export async function gitRemoteUrlFromGitCli(uri: vscode.Uri): Promise<string | undefined> {
```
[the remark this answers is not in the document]
valerybugakov: Agree that returning all of the available remotes here is on the safer side of things.

vscode/src/repository/repo-name-getter.node.test.ts
```diff
@@ -0,0 +1,65 @@
...
+                    upstream  https://github.com/foo/bar (push)
+                `,
+                stderr: '',
+            })
+            // TODO: find the right way to type this that is compatible with the `promisify` wrapper function.
+            // skipping this for now to address to continue with higher priority work.
+        }) as any)
+
+        const uri = URI.file('path/to/file/foo.ts')
+        expect(await gitRemoteUrlFromGitCli(uri)).toBe('https://github.com/sourcegraph/cody')
```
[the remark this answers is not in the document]
valerybugakov: Let's continue the discussion in the comment above.

vscode/src/repository/git-extension-api.ts
valerybugakov: This helper does not support Git submodules, but it works correctly in cases where multiple repos are opened in the workspace (if I understood your "whole `/dev` folder example). I did not implement the Git submodules support yet, so this also applies to other comments where this question is raised.

vscode/src/repository/repo-name-getter.node.ts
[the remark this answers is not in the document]
valerybugakov: > If I git pull a private repo, Cody ignore should work regardless of whether I push to a fork of it

vscode/src/repository/repo-name-resolver.test.ts
valerybugakov: Let's discuss and spec this in this comment thread: https://github.com/sourcegraph/cody/pull/3771#discussion_r1567084499

vscode/src/repository/repo-name-resolver.ts
valerybugakov: Let's continue the discussion here: https://github.com/sourcegraph/cody/pull/3771#discussion_r1567090231

vscode/src/repository/repo-name-resolver.test.ts
valerybugakov: I haven't implemented submodules support for this iteration. I planned to look into it as a part of the performance testing for nested checkout.

vscode/src/context/context-filters.provider.test.ts
valerybugakov: Yes, let's discuss here: https://github.com/sourcegraph/cody/pull/3771#discussion_r1567093677

vscode/src/context/context-filters-provider.ts
valerybugakov: You're right. It should be relative to the repo root. We do not plan to support `filePathPatterns` on the JetBrains GA release, so keeping the half-backed implementation on the client side is confusing. I removed it for the time being the renamed the `isPathAllowed` method to `isRepoNameAllowed`.

vscode/src/repository/repo-name-getter.node.ts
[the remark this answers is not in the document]
valerybugakov: We should add this to the spec and convey it to the customers once it's implemented: we use all remotes we can find to apply the Cody Ignore filters.

@dominiccooney, @aramaraju, do you see any downsides (apart from the latency increase caused by a higher number of remotes to check)?

vscode/src/repository/repo-name-resolver.ts
valerybugakov: Thanks for highlighting that, Philipp! 

I planned to look into this in one of the follow-up PRs to avoid growing the scope of this one. There's a TODO item in [the parent issue](https://github.com/sourcegraph/cody/issues/3641) that mentions resolving the repo name using our GraphQL API. It still requires the remote URL to be resolved locally, so it will gradually improve what we have here.

vscode/src/repository/repo-name-resolver.test.ts
valerybugakov: Now, it is implemented in this PR with unit tests.

vscode/src/context/context-filters-provider.ts
dominiccooney: Un-saved files have weird schemes in VSCode and JetBrains.

This is a complicated case... For example the diff viewer documents have URIs like that, but the underlying document may be an ignored one.

For now:
- File a P0 follow-up task, put it in the VSCode Project, assign it to me
- Put the task number in a TODO
- Log the schemes in the local log
- If there's any policy, we should disallow the URLs (maybe breaking the product, but we will find it through testing), if there's no policy and we're in our "default open" mode then we should allow them, naturally.

@aramaraju can you think through some of these edge cases and add them to your test plan?

vscode/src/context/context-filters-provider.ts
```diff
@@ -101,29 +119,21 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
     const filePathPatterns = item.filePathPatterns
-        ? item.filePathPatterns.map(pattern => new RE2(pattern))
+        ? item.filePathPatterns.map(pattern => new RE2(pattern, 'u'))
         : undefined
 
     return { repoNamePattern, filePathPatterns }
 }
+
+/**
+ * A a singleton instance of the `ContextFiltersProvider` class.
```
dominiccooney, line 136:
```suggestion
 * A singleton instance of the `ContextFiltersProvider` class.
```

vscode/src/repository/git-extension-api.ts
dominiccooney: This function is OK if it is used correctly: With a specific file URI, and not just the URI to workspace root 0 or whatever. It gets the codebase related to the specified URI (if any.)

vscode/src/repository/repo-name-getter.node.ts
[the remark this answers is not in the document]
dominiccooney: It's a good idea, but I think we should file a clean-up task for this. There's other things that map git folders to remotes (local embeddings, default remote repo context inclusion) and we should keep everything consistent. I'm not prepared to explain why, say, default remote context identification did identify a remote repo but cody ignore didn't or vice versa.

dominiccooney (approved):
Some feedback inline.

Approving this because it is a good step forward.

@philipp-spiess, @taras-yemets thanks for your great feedback. There's still pieces to go here—invalidating the cache, updating clients to revalidate, etc. So we should track and act on this feedback but I think it is OK if we don't nail it all down in this PR. (Welcome pushback if that approach is too loose.)

vscode/src/repository/repo-name-getter.node.ts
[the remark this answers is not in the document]
valerybugakov: Added a subtask to the parent issue. We can extract it into a standalone clean-up issue later.

vscode/src/context/context-filters-provider.ts
aramaraju: Unsaved files are the ones that are primary ones of concern, @taras-yemets can you add this to your test plans, if you haven't already?

valerybugakov: - TODO and log: https://github.com/sourcegraph/cody/pull/3894
- Follow-up issue https://github.com/sourcegraph/cody/issues/3893

PR #3773, opened 2024-04-11
Chat: Add Pop out CTA

umpox:
## Description

Adds a "Pop out" CTA to the Chat title bar:

https://github.com/sourcegraph/cody/assets/9516420/cc8d9103-a1d7-422c-b3f0-e308f9556d1a

## Test plan

1. Open chats
2. Make them pop out
3. Check the icon disappears when the chat is popped out

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -38,7 +38,12 @@
     "test:unit:tree-sitter-queries": "vitest ./src/tree-sitter/query-tests/*.test.ts",
     "github-changelog": "ts-node-transpile-only ./scripts/github-changelog.ts"
   },
-  "categories": ["Programming Languages", "Machine Learning", "Snippets", "Education"],
+  "categories": [
+    "Programming Languages",
+    "Machine Learning",
+    "Snippets",
+    "Education"
+  ],
   "keywords": [
     "ai",
     "openai",
@@ -89,7 +94,11 @@
   },
   "main": "./dist/extension.node.js",
   "browser": "./dist/extension.web.js",
-  "activationEvents": ["onLanguage", "onStartupFinished", "onWebviewPanel:cody.chatPanel"],
+  "activationEvents": [
+    "onLanguage",
+    "onStartupFinished",
+    "onWebviewPanel:cody.chatPanel"
+  ],
   "contributes": {
     "walkthroughs": [
       {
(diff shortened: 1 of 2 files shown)


vscode/package.json
```diff
@@ -810,10 +818,16 @@
           "group": "navigation@2",
           "visibility": "visible"
         },
+        {
+          "command": "workbench.action.moveEditorToNewWindow",
+          "when": "activeWebviewPanelId == cody.chatPanel && cody.activated && !isAuxiliaryEditorPart",
```
umpox, line 823: `isAuxiliaryEditorPart` tells us if the chat is a floating window, so we only show the "Pop out" CTA when it is _not_ floating

vscode/package.json
```diff
@@ -424,6 +424,14 @@
         "group": "Cody",
         "icon": "$(new-comment-icon)"
       },
+      {
+        "command": "workbench.action.moveEditorToNewWindow",
+        "category": "Cody",
+        "title": "Pop out",
+        "when": "cody.activated",
+        "group": "Cody",
+        "icon": "$(link-external)"
```
umpox, line 433: Happy to change the icon or tooltip if anyone has better ideas:

image

abeatrix (approved):
nice! Worth a changelog entry?

chillatom:
A couple weird things 

- If I execute chat then pop it out, I lose the chat history within the window, but it retains the chat tab title
- If I then close the pop out, conversation history still gone,
- But if I close the tab then re-open from the history sidebar, it is there again. 

https://www.loom.com/share/ca1d422490bf49d9a9f18583dbc9b405

umpox:
@chillatom Oh yeah that is weird, I'll look into why that is happening as a follow up. Possibly resetting the webview when it is popped out 🤔

umpox:
From the VS Code changelog...

> Note: due to a technical limitation, moving a webview-based editor between windows requires the contents of that editor to reload. It is then up to the editor to restore the state you had previously accumulated. In some cases your state might be reset, as if you had opened the editor for the first time.

Looks like we're losing the state that we have in memory when we switch editor. @abeatrix Do you know if we've fixed anything similar before? Sounds like it could be quite tricky to fix. 

Previous chats load fine so I'm guessing we want to persist the current chat to local storage whilst it's still active 🤔

PR #3775, opened 2024-04-11
Agent protocol and implementation for searching, listing, testing remote repository names

dominiccooney:
This adds Agent endpoints for:

- Testing whether a remote has a specific repository.
- Retrieving repository names from a remote matching a fuzzysort.
- Retrieving all repository names from a remote, with pagination.

Closes sourcegraph/jetbrains#605.

## Test plan

Manually tested with sourcegraph/jetbrains#1308

Added agent tests, run with `pnpm test`

### DIFF
--- a/vscode/src/context/remote-repo-searcher.ts
+++ b/vscode/src/context/remote-repo-searcher.ts
@@ -1,0 +1,171 @@
+import fuzzysort from 'fuzzysort'
+import * as vscode from 'vscode'
+import { type Repo, type RepoFetcher, RepoFetcherState } from './repo-fetcher'
+
+export interface RemoteRepoListResult {
+    repos: Repo[]
+    startIndex: number
+    count: number
+}
+
+/**
+ * Searches a remote repo list and provides paginated results.
+ */
+export class RemoteRepoSearcher implements vscode.Disposable {
+    // Marker that we haven't cached a user query.
+    private static readonly NO_QUERY = 'NO_QUERY_CACHED'
+
+    private disposables: vscode.Disposable[] = []
+
+    // The query that we last performed. We cache fuzzy search results to
+    // provide a stable, paginated list of results.
+    private cachedQuery: string | undefined = RemoteRepoSearcher.NO_QUERY
+
+    // If cachedQuery is not NO_QUERY, the matching results.
+    private cachedResult: readonly Repo[] = []
+
+    // Index of fuzzysort targets.
+    private fuzzyIndex: Fuzzysort.Prepared[] | undefined = undefined
+
+    // Map from fuzzysort targets to repository results.
+    private fuzzyTargetRepoMap: Map<string, Repo> | undefined = undefined
+
+    // Fired when the underlying repo list changes, indicating our results may
+    // have been invalidated.
+    private repoListChangedEmitter: vscode.EventEmitter<void> = new vscode.EventEmitter<void>()
+
+    // Fired when the underlying repo fetching changes state.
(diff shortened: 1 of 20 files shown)


abeatrix (commented):
Is there an instance I can connect to to test it on JetBrains? 
I tried to connect to demo.sourcegraph but it is showing up as 0 repo 😆 
image

vscode/src/context/remote-repo-searcher.ts
```diff
@@ -0,0 +1,261 @@
...
+    private updateCachedResult(query: string | undefined): void {
+        if (!query) {
+            // No query, so we return all repos in fetch order.
+            this.cachedResult = this.fetcher.repositories
+            return
+        }
+
+        if (!this.fuzzyIndex || !this.fuzzyTargetRepoMap) {
+            // Update the fuzzy search index.
+            // TODO: When speed is necessary, improve this by only indexing
```
abeatrix, line 143: This sounds like a good plan! Reasons we don't want to do that now?

vscode/src/context/remote-repo-searcher.ts
```diff
@@ -0,0 +1,261 @@
...
+                this.fuzzyIndex.push(prepared)
+            }
+        }
+
+        // Do fuzzy search.
+        // TODO: When speed is necessary, improve this by searching the changed
+        // repositories and merging the result with the cached results, if any.
+        // Could also limit the result set and cap the maximum size of any list.
+        const result = fuzzysort.go(query, this.fuzzyIndex, {
+            threshold: -1000,
```
abeatrix, line 159:
```suggestion
            limit: 20,
            threshold: -1000,
```
would it make sense to limit the results for UI? I don't imagine people would want to go through a long list but that's just me, so feel free to ignore this 😄

abeatrix (approved):
I tried to run it with JetBrains but somehow it's showing that demo has 0 repo. 
But the repositories are showing up in VS Code for the same instance, so I'm not sure if I am running the JetBrains dev env incorrectly.

From VS Code for reference: 
image

dominiccooney:
@abeatrix thank you for the review 🙇 

> Is there an instance I can connect to to test it on JetBrains?
> I tried to connect to demo.sourcegraph but it is showing up as 0 repo

demo.sourcegraph.com is a good one to test. 0 is working as intended: you haven't selected a repo.

> I tried to run it with JetBrains but somehow it's showing that demo has 0 repo.
> But the repositories are showing up in VS Code for the same instance, so I'm not sure if I am running the JetBrains dev env incorrectly.

The UX is: Hit the pencil, then type repo names. Autocomplete will appear. Command-space to manually trigger it.

> [Various perf ideas] sounds like a good plan! Reasons we don't want to do that now?

I removed these TODOs. It seems fast enough, so I will keep it simple for now.

> would it make sense to limit the results for UI? I don't imagine people would want to go through a long list but that's just me, so feel free to ignore this 😄

There's two levels of filtering: This -1000 limit, and then IntelliJ autocomplete requests a subset of that list. I tried to do something which would work for a lot of clients:

- Rich clients that can handle the full repo list can call `remoteRepo/list` without a query and get paginated access to the full list, which they can then filter down.
- Limited clients can call `remoteRepo/list` with a query and get paginated access to the matching score > -1000 items.
- I don't limit the fuzzysort to the number of results the client requests, because you'd get an n^2 blow up if the client paged through the whole list. So it does the fuzzysearch once, caches the result, then lets the client page through it.

It's not super elegant but it should work for a range of client capabilities.

PR #3781, opened 2024-04-11
Move new user context window behind feature flag

abeatrix:
Branched off from https://github.com/sourcegraph/cody/pull/3742. This will be merged back into https://github.com/sourcegraph/cody/pull/3742 once approved. 

As discussed with @chillatom and @dominiccooney, we will move the new user context window budget behind a feature flag `cody-chat-context-budget` for internal testing before the limit is raised accordingly on the server.  See https://docs.google.com/document/d/1_rg8j4yfz1Ot-lQle8v_K_y8V3JDzpKYxO134X7MItc/edit for the expected behavior.

This commit refactors the model provider logic to support A/B testing of increased token limits for select models when the user context feature flag is enabled. It introduces a new parameter to getDotComDefaultModels to specify default or experimental models.
- Raised limits for DotCom users with feature flag enabled:
  - User Context: new 30k limit
  - Chat Input: increased from 7k to 15k

- Updated TokenCounter to accept ContextWindow object instead of a single number to make it easier for us to determine/assign context window reserved for different resources. 

- Updated TokenCounter for tracking usage in different budget modes as descripted in the shared Docs.

- Updated TokenCounter and PromptBuilder to ensure prompts are built in the correct order: 
  - `Preamble -> Chat Input -> Context (User / Enhanced)`

- Added new tests for TokenCounter.

## Test plan

Follow the Test Plan listed in https://github.com/sourcegraph/cody/pull/3742.
The feature flag has been enabled for all Sourcegraph teammates.


lib/shared/src/token/counter.ts
```diff
@@ -30,24 +30,18 @@ export class TokenCounter {
...
-    constructor(public readonly totalBudget: number) {
-        // Set the maximum number of tokens that can be used by chat and context.
-        // This allows the token counter to allocate the budget between chat and context tokens
-        // based on the total budget.
-        // NOTE: The totalBudget of Claude-3 models is CHAT_TOKEN_BUDGET + USER_CONTEXT_TOKEN_BUDGET.
-        const chatTokenBudget = Math.min(totalBudget, CHAT_TOKEN_BUDGET)
+    constructor(contextWindow: { chat: number; user: number; enhanced: number }) {
+        // If the chat token budget is equal to the user context token budget,
+        // the chat and user context tokens will share the same budget.
+        this.shareChatAndUserContextBudget = contextWindow.user === 0
```
dominiccooney, line 36: I'm just re-reading `allocateTokens` and I think it is buggy.

If you're sharing the budget and you add a user context item, then you need to reduce the chat budget and enhanced context budget too. You only do it for the case where the chat budget reduces the user context budget. Same for adding an enhanced context item.

dominiccooney (changes requested):
I think `allocateTokens` needs work. Let's add more tests.

lib/shared/src/token/counter.ts
abeatrix: Copied from slack: 
> this is the expected behavior. I’ll document it better in the docstring, but basically every time we construct the prompt, we always build the chat prompt first, and if we have tokens left, we will then build the context prompt with it. If there is no more chat token left, we simply won’t add any more context. 
That’s why adding new chat message will deduct the token limit from context, but not vice versa.

lib/shared/src/models/dotcom.ts
```diff
@@ -84,6 +84,43 @@ export const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
...
+ * Returns an array of `ModelProvider` objects representing the default models for the Dot Com product.
+ *
+ * If the `hasUserContextFeatureFlag` parameter is true, the function will return a modified array where the
+ * models with higher token limits (as defined in the `modelsWithHigherLimit` array) will have their
+ * `maxRequestTokens` property increased by the `USER_CONTEXT_TOKEN_BUDGET` value.
+ *
+ * @param hasUserContextFeatureFlag - A boolean indicating whether the user context feature flag is enabled.
+ * @returns An array of `ModelProvider` objects representing the default models for the Dot Com product.
+ */
+export function getDotComDefaultModels(hasUserContextFeatureFlag = false): ModelProvider[] {
```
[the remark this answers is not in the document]
abeatrix: Ah that's a solid point I need to keep that in mind. Thank you!

lib/shared/src/token/counter.ts
dominiccooney: Let me push back on that, because *at the call site* someone could come in and reorder that code and shoot themselves in the foot. The general idea is: make things where the easy use is the correct use.

It's pretty easy to make allocateTokens more robust, so let's do that.

If that's too hard, think of an API structure where you're forced to do things in order. Like maybe the budget has phases and it calls you back, forcing to count the chat tokens first. But seems simpler just to make it work regardless of call order.

abeatrix: Yea that makes sense: I updated the logic so that it would work even when the prompt is constructed out of order, and added a test to reflect that: https://github.com/sourcegraph/cody/pull/3781/commits/3c87cb78952e6c799c3ef59b09c563072cfcbd2c

dominiccooney (changes requested):

lib/shared/src/token/counter.ts
```diff
@@ -76,33 +83,59 @@ export class TokenCounter {
...
      * @param type - The type of token usage to allocate.
      * @param count - The number of tokens to allocate.
      */
     private allocateTokens(type: 'chat' | ContextTokenUsageType, count: number): void {
-        this.usedTokens[type] += count
-        if (type === 'chat' && this.shareChatAndUserContextBudget) {
-            // If chat and user context tokens share the same budget, update both.
-            this.usedTokens.user += count
-            this.usedTokens.enhanced += count // Enhanced tokens should also be deducted.
+        let { chat, user, enhanced } = this.usedTokens
```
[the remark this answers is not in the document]
abeatrix: Enhanced context budget is current [Chat Budget *  ENHANCED_CONTEXT_ALLOCATION](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/prompt.ts?L50).
In the code, we are not tracking the used token for enhanced context. I have updated the docstring and removed it from the this.usedTokens to avoid confusing.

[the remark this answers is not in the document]
abeatrix: Here is the doc with the desired state written down: https://docs.google.com/document/d/1_rg8j4yfz1Ot-lQle8v_K_y8V3JDzpKYxO134X7MItc/edit

lib/shared/src/token/counter.ts
```diff
@@ -76,33 +83,59 @@ export class TokenCounter {
...
     /**
      * Allocates the specified number of tokens for the given token usage type.
-     * If the token usage type is 'chat' and the chat and user context tokens share the same budget,
-     * the user context and enhanced tokens will also be updated.
+     *
+     * When `shareChatAndUserBudget` is active, the chat and user context token usage are kept in sync.
+     * Otherwise, the token usage is updated separately for each type.
+     *
+     * For 'enhanced' token usage, the tokens are always calculated based on the latest chat budget,
+     * so the enhanced token usage is not tracked separately but will count towards the chat token usage.
```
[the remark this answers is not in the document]
abeatrix: Done. Please see the latest file change

lib/shared/src/models/types.ts
```diff
@@ -37,3 +37,8 @@ export type ChatModel =
           [K in keyof Models]: HasUsage<Models[K], ModelUsage.Chat>
       }[keyof Models]['model']
     | (string & {})
+
+export interface ModelContextWindow {
+    chat: number
+    user: number
+}
```
philipp-spiess: Can you add some documentation on what the difference is between these limits? It's not clear from the name alone (we might also want to consider renaming them to something more explicit).

lib/shared/src/models/dotcom.ts
```diff
@@ -3,15 +3,15 @@ import { CHAT_TOKEN_BUDGET, FAST_CHAT_TOKEN_BUDGET, USER_CONTEXT_TOKEN_BUDGET }
...
+const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
     {
         title: 'Claude 2.0',
         model: 'anthropic/claude-2.0',
         provider: 'Anthropic',
         default: false,
         codyProOnly: true,
         usage: [ModelUsage.Chat, ModelUsage.Edit],
-        maxRequestTokens: CHAT_TOKEN_BUDGET,
+        contextWindow: { chat: CHAT_TOKEN_BUDGET, user: 0 },
```
philipp-spiess: What exactly does `user: 0` mean here? the user can not @-mention anything anymore? 🤔

agent/recordings/enterpriseClient_3965582033/recording.har.yaml
```diff
@@ -129,462 +129,248 @@ log:
           textJSON:
             maxTokensToSample: 1000
```
philipp-spiess, line 130: I don't understand why the recordings needed to change here if they still apply the same tokens? Can you point me to the difference in the request that required the regenration?

lib/shared/src/token/counter.ts
```diff
@@ -148,7 +154,7 @@ export class TokenCounter {
      * @returns The number of tokens in the message.
      */
     private static getTokenCountForMessage(message: Message): number {
-        if (!message) {
+        if (!message?.text) {
```
philipp-spiess, line 156: Let's not do unexpected string to boolean coercion, this is biting us quite hard right now in https://github.com/sourcegraph/cody/pull/3734

```suggestion
        if (message?.text && message?.text.length > 0) {
```

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -460,7 +462,6 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                                       symf: this.config.experimentalSymfContext ? this.symf : null,
                                       remoteSearch: this.remoteSearch,
                                   },
-                                  hints: { maxChars },
```
philipp-spiess, line 463: It still seems useful for retrievers to know the limits?

lib/shared/src/models/types.ts
philipp-spiess: Ah I saw that you changed this after setting the PR up for review. I think the new terminology is clearer, thank you!

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
abeatrix: we noticed this is not being used currently, or did we miss something?

lib/shared/src/models/dotcom.ts
abeatrix: updated to make this more clear <3

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
philipp-spiess: Hmm yeah you're right we don't (especially since I removed some code paths) so let's remove this until we need it :)

agent/recordings/enterpriseClient_3965582033/recording.har.yaml
abeatrix: I think it's because of the changes to the context items where we now add the (token) size to each Context Item that prompted the change.

philipp-spiess: @abeatrix Hmm but wouldn't it result in the same prompts as before?

abeatrix: ah you're right. i think it's because of this change from the upstream where we sort the context for agent testing where we move the selection to the last item: https://github.com/sourcegraph/cody/blob/3927cb06741f06b59c48c51bf0efc676c77ab199/vscode/src/chat/chat-view/agentContextSorting.ts 

do you think this is the case?

image

image

lib/shared/src/models/dotcom.ts
```diff
@@ -3,15 +3,15 @@ import { CHAT_TOKEN_BUDGET, FAST_CHAT_TOKEN_BUDGET, USER_CONTEXT_TOKEN_BUDGET }
...
+const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
     {
         title: 'Claude 2.0',
         model: 'anthropic/claude-2.0',
         provider: 'Anthropic',
         default: false,
         codyProOnly: true,
         usage: [ModelUsage.Chat, ModelUsage.Edit],
-        maxRequestTokens: CHAT_TOKEN_BUDGET,
+        contextWindow: { input: CHAT_TOKEN_BUDGET },
```
abeatrix, line 14: updated to `input` instead of `chat` for a clear naming and aligns with the upcoming changes for the output token expansion in https://github.com/sourcegraph/cody/pull/3797

lib/shared/src/token/constants.ts
```diff
@@ -15,12 +15,20 @@ export const FAST_CHAT_TOKEN_BUDGET = 4096
...
- * NOTE: This is only being used for Claude-3 Sonnet and Opus.
+ * The total context window reserved for user added context (@-mention, right-click, etc.)
+ *
+ * This is only being used for Claude-3 Sonnet and Opus when feature flag is enabled.
  * For those models, this is added on top of the chat context window.
  * For other models, this number is not used. Instead, they will share
  * the token budget for user-context with the chat context window.
  */
-export const USER_CONTEXT_TOKEN_BUDGET = 10000
+export const EXPERIMENTAL_USER_CONTEXT_TOKEN_BUDGET = 30000
```
abeatrix, line 27: Requested by @chillatom to raise limits for DotCom users with feature flag enabled:
- new 30k limit for user context
- increased from 7k to 15k for chat input

lib/shared/src/token/counter.test.ts
```diff
@@ -1,153 +1,244 @@
...
             { speaker: 'assistant', text: ps`Hi there!` },
         ]
-        counter.updateUsage('chat', messages)
-        expect(counter.remainingTokens.chat).toBe(counter.remainingTokens.context.user)
-        expect(counter.remainingTokens.context.enhanced).toBeLessThan(counter.remainingTokens.chat)
+        // 3 + 4 chat tokens needed = within limit of 7
+        expect(counter.updateUsage('input', messages)).toBe(true)
+    })
+
+    it('should update token usage and return true when within limits - chat and user context have separate budgets', () => {
```
dominiccooney, line 90: This doesn't really test that it updates the limits... you need an assertion about the remaining tokens.

lib/shared/src/token/counter.test.ts
```diff
@@ -1,153 +1,244 @@
...
+                { speaker: 'human', text: ps`Hello` },
+                { speaker: 'assistant', text: ps`Hi there!` },
+            ])
+        ).toBe(true)
+        expect(
+            counter.updateUsage('user', [{ speaker: 'system', text: ps`You are a helpful assistant.` }])
+        ).toBe(true)
+    })
+
+    it('should throw error when trying to update enhanced context token usage before chat input', () => {
```
dominiccooney, line 108: 👍

lib/shared/src/token/counter.test.ts
```diff
@@ -1,153 +1,244 @@
...
-        expect(counter.remainingTokens.context.enhanced).toBe(3) // 60% of 5
+        // Enhanced Token Budget: 9 * 0.6 = Round down to 5
+        // - the remaining chat token budget * ENHANCED_CONTEXT_ALLOCATION
+        const enhancedContextMessages = [
+            { speaker: 'human', text: ps`Here is my enhanced context...` },
+            { speaker: 'assistant', text: ps`ok` },
+        ] as Message[]
+        expect(TokenCounter.getMessagesTokenCount(enhancedContextMessages)).toBe(7)
+        // 7 tokens needed, exceeds the remaining token budget of 5 for Enhanced Context
+        expect(counter.updateUsage('enhanced', enhancedContextMessages)).toBe(false)
```
dominiccooney, line 162: Theses tests would be shorter and more persuasive if, right after this, you added context that *did* succeed.

It's also good to test boundaries—for example 6 (just misses), 5 (just fits)

lib/shared/src/token/counter.ts
```diff
@@ -1,71 +1,51 @@
...
+import type { ChatContextTokenUsage, TokenUsageType } from '.'
+import type { ModelContextWindow } from '..'
+import type { Message, PromptString } from '..'
 import { ENHANCED_CONTEXT_ALLOCATION } from './constants'
 
 /**
- * A class to manage the token usage during prompt building.
+ * A class to manage the token allocation during prompt building.
+ *
+ * NOTE: A new TokenCounter is created everytime a new PromptBuilder is created.
```
dominiccooney, line 11: Why is this relevant?

dominiccooney (approved):
Looks good.

lib/shared/src/token/counter.test.ts
abeatrix: Good suggestion! I've extended the test as suggested to demonstrate context within boundaries can still be added successfully.

lib/shared/src/token/counter.ts
abeatrix: removed!

PR #3782, opened 2024-04-11
Chat: reveal hidden chat panel on @-selection

abeatrix:
This PR fixes the issue where an exisiting chat panel will stay hidden when the user tries to send the editor selection to the chat panel as @-mentions token (@-selection) . See video below for example

Expected behavior:
The existing chat panel should be revealed on click

### Before

Hidden chat panel will stay hidden on @-selection 

https://github.com/sourcegraph/cody/assets/68532117/69c8beab-42b9-41a0-89e6-14fabf687ac4

### After

Current active chat panel will show up if hidden:

https://github.com/sourcegraph/cody/assets/68532117/0fd048f1-a009-4931-82a3-478a154c44e3


kalanchan (approved):
I like it!

PR #3787, opened 2024-04-12
Fix agent folding ranges API to use getEditSmartSelection

pkukielka:
## Test plan

Testing on the Jetbrains plugin side as it only affects it currently.


mkondratek (approved):
lgtm

PR #3791, opened 2024-04-12
Fix undo/accept/cancel command calls

pkukielka:
## Changes

Fixes API for calling `editTask/undo`, `editTask/accept` and `editTask/cancel`

## Test plan

Manual tests with the JB plugin


mkondratek (approved):
nice & clean

dominiccooney:
Regarding the agent bindings failure, you may just have to regenerate agent bindings with ./agent/scripts/generate-agent-kotlin-bindings.sh , but IIRC there's some problem with Java representing undefined, that's why we had the `null` returns.

pkukielka:
> Regarding the agent bindings failure, you may just have to regenerate agent bindings with ./agent/scripts/generate-agent-kotlin-bindings.sh , but IIRC there's some problem with Java representing undefined, that's why we had the `null` returns.

Yea, thanks, I had to rework it a little bit indeed.

Issue #3795, opened 2024-04-14
bug: Program listing generated is missing the end of program


nowhereman999:
### Version

 1.12.0

### Describe the bug

 I asked Cody to write a program to parse an expression in QB64 BASIC but it doesn't supply the entire program.  It stops before ending.
 
I asked:
 why are you not finishing the entire program?
It responds:
You're absolutely right, I should have provided the complete program without stopping midway. Here's the full code without any functions to evaluate the given expression:

Here's the code it produces..
DEFINT A-Z
DIM expr$, c$, i%, value%, j%, temp%, k%, op%
expr$ = "4*5-3*((((15-3)*2)+2)+51-6*(22+1))*22"

PRINT "Expression: " + expr$

i = 0
value% = 0
op% = 0 ' 0 for addition/subtraction, 1 for multiplication/division

DO WHILE i <= LEN(expr$)
    c$ = MID$(expr$, i + 1, 1)
    IF c$ >= "0" AND c$ <= "9" THEN
        IF op% = 0 THEN
            value% = value% * 10 + VAL(c$)
        ELSE
            temp% = temp% * 10 + VAL(c$)
        END IF
        i = i + 1
    ELSEIF c$ = "(" THEN
        i = i + 1
        temp% = 0
        DO
            j% = i
            DO WHILE j% <= LEN(expr$)
                c$ = MID$(expr$, j% + 1, 1)
                IF c$ >= "0" AND c$ <= "9" THEN
                    temp% = temp% * 10 + VAL(c$)
                    j% = j% + 1
                ELSEIF c$ = "(" THEN
                    j% = j% + 1
                    DO
                        k% = j%
                        DO WHILE k% <= LEN(expr$)
                            c$ = MID$(expr$, k% + 1, 1)
                            IF c$ >= "0" AND c$ <= "9" THEN
                                temp% = temp% * 10 + VAL(c$)
                                k% = k% + 1
                            ELSEIF c$ = ")" THEN
                                EXIT DO
                            END IF
                        LOOP
                        j% = k%
                    LOOP UNTIL c$ = ")"
                ELSEIF c$ = "+" THEN
                    j% = j% + 1
                    DO WHILE j% <= LEN(expr$)
                        c$ = MID$(expr$, j% + 1, 1)
                        IF c$ >= "0" AND c$ <= "9" THEN
                            temp% = temp% + VAL(c$)
                            j% = j% + 1
                        ELSE
                            EXIT DO
                        END IF
                    LOOP
[...]

Issue #3786, opened 2024-04-12
feedback: Claude3 Opus responses are often truncated...


tumma72:
### Version

v1.12.0

### Areas for Improvement

- [ ] UI/UX
- [ ] Onboarding
- [ ] Docs
- [X] Chat
- [ ] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience and include a screenshot.

Often when the answer requires more than a couple of paragraphs of text, the Cody chat is truncated. I have experienced this primarily with the Claude 3 Opus model which tends to deliver longer and more articulated answers. Sometimes even after 2 or 3 repetitions it doesn't come to an end, so I have to ask to reply in smaller chunks.

### Describe the solution you'd like to see

I would like the window size for the reply to be longer so I won't have to ask to deliver in smaller chunks, with code snippets can be a bit of a puzzle work sometimes.

umpox:
Thanks for reporting! We have increased Cody's output limit, it should now be much rarer to see Cody truncating the response like this. Let us know if you encounter any more issues!

This change will be available in this weeks Cody release, which is scheduled to go out either today or tomorrow.

tumma72:
Thanks for fixing this, I am experiencing a lot less issues since the last few updates. For me we can consider this a closed issue.

Issue #3784, opened 2024-04-12
bug: 


frank111074:
### Version

1.12.0

### Describe the bug

The output of code in Chat keeps breaking at different places.

### Expected behavior

if i ask, for revision or searching for simple errors, cody will do it, returns code and breakes the output at different places.
If i get a long answer, output breakes at different places. 
if I post the question again and request the identical code output, it breaks at a different place every time.

umpox:
Thanks for reporting! We have increased Cody's output limit, it should now be much rarer to see Cody truncating the response like this. Let us know if you encounter any more issues!

This change will be available in this weeks Cody release, which is scheduled to go out either today or tomorrow.

Issue #3779, opened 2024-04-11
bug: code output is often truncated


hangrymuppet:
### Version

v1.12.0

### Describe the bug

Code is shown in a code box that is often truncated before it is done. This does not happen with text generated outside of code blocks.

Say I have a coversation like this,

Q) Explain concept and write code that does something

A) Concept explained

code block with truncated lines

Q) Document the above code and write comments

same code block with comments again truncated

The first code block can be truncated mid-line or mid-file.

Examples (last line of code block shown below)

mid-line example

```if (dog !=```

mid-file

```while (b < a) {```

### Expected behavior

no truncation or an option to regenerate the answer / continue

### Additional context

<img width="780" alt="Screenshot 2024-04-11 at 13 06 39" src="https://github.com/sourcegraph/cody/assets/7991855/a45ac125-b2d4-42ab-a8a3-9303202644da">
<img width="686" alt="Screenshot 2024-04-11 at 13 06 53" src="https://github.com/sourcegraph/cody/assets/7991855/0ead2085-267d-4b26-bf3c-d1970c3c3968">

kalanchan:
related #3648

umpox:
Thanks for reporting! We have increased Cody's output limit, it should now be much rarer to see Cody truncating the response like this. Let us know if you encounter any more issues!

This change will be available in this weeks Cody release, which is scheduled to go out either today or tomorrow.

kalanchan:
> @umpox Thanks a ton!
> 
> Will this require an extension update (i use vscode) or is this a backend update?

it's available in v1.14! upgrade away :)

PR #3797, opened 2024-04-15
Chat/Commands: Increase output token limit for PLG

umpox:
## Description

closes https://github.com/sourcegraph/cody/issues/3648
closes https://github.com/sourcegraph/cody/issues/3795
closes https://github.com/sourcegraph/cody-issues/issues/14
closes https://github.com/sourcegraph/cody/issues/3779
closes https://github.com/sourcegraph/cody/issues/3786
closes https://github.com/sourcegraph/cody/issues/3784
closes https://github.com/sourcegraph/cody/issues/3721
closes https://github.com/sourcegraph/cody/issues/3571
closes https://github.com/sourcegraph/cody/issues/3472
closes https://github.com/sourcegraph/cody/issues/3536

This PR:
- Increases the output limits for Sourcegraph.com Cody usage from `1000` to `4000` tokens
- Removes a redundant reduction in context size that will add an extra 1000 tokens to our context window. This is only necessary on certain enterprise models where the input + output lengths are computed together.

> [!NOTE]  
> This PR requires that https://github.com/sourcegraph/sourcegraph/pull/61872 is merged first

## Test plan

1. Run https://github.com/sourcegraph/sourcegraph/pull/61872 locally
2. Trigger long context responses (e.g. "Show me 300 lines of typescript code")
3. Expect that everything functions as normal


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -600,7 +599,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         const contextLimit = getContextWindowLimitInBytes(
             [...this.chatModel.getMessages()],
             // Minus the character limit reserved for the answer token
-            this.chatModel.maxChars - tokensToChars(ANSWER_TOKENS)
```
umpox, line 603: TODO: Need to ensure that this logic works with older OpenAI models

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -597,10 +598,13 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         }
         // Use the number of characters left in the chat model as the limit
         // for adding user context files to the chat.
+        const maxInputChars = this.authProvider.getAuthStatus().isDotCom
+            ? this.chatModel.maxInputChars
+            : // Minus the character limit reserved for the answer token
+              this.chatModel.maxInputChars - tokensToChars(ANSWER_TOKENS)
```
umpox, line 604: This logic only makes sense for older GPT models where the input+output limit is calculated together.

image

For enterprise we will need to address this

lib/shared/src/models/dotcom.ts
```diff
@@ -94,6 +108,7 @@ export const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
         codyProOnly: true,
         usage: [ModelUsage.Chat],
         // Context window is 2k, but lower to 1.8k to leave some rooms for token counting errors.
-        maxToken: 1800,
+        maxInputToken: 1800,
```
philipp-spiess, line 111: unrelated but it seems like we can use the "standard" 7k now :) https://fireworks.ai/models/fireworks/mixtral-8x22b-instruct-preview

philipp-spiess (approved):
LGTM :)

vscode/src/models/utils.ts
```diff
@@ -51,9 +57,14 @@ export function syncModelProviders(authStatus: AuthStatus): void {
...
     model: string
-    tokens?: number
+    inputTokens?: number
+    outputTokens?: number
     apiKey?: string
     apiEndpoint?: string
+    /**
+     * @deprecated Use `inputTokens` instead.
+     */
+    tokens?: number
```
abeatrix, line 67: We could just remove it since this was not included in any release yet?

umpox: Yep I can do, my only concern is that this might break any internal stuff we've configured but I guess it's still fresh so that's not much of a worry

umpox:
@sourcegraph/cody-security Guessing this was triggered because of the merge commit? Maybe we should exclude those?

philipp-spiess:
@umpox Ah yeah probably, we use `git diff --name-only origin/main`. Rebasing would probably help. I’m not sure how to work around this easily and this should resolve itself once all PRs are branched-off of the latest main so maybe we can just life with some more notifications to review for a few days? 😬

philipp-spiess:
@umpox Ah no this is because we look at the files and not just the modified ranges within those files. Shoots. This actually needs a much more sophisticated approach (we need to know which ranges were modified and only emit linter issues in those places). 

Hmmm maybe as a temporary workaround we can extract the `unsafe_` calls in "hot" paths (like in `SimpleChatPanelProvider.ts` in this case) into a separate module that we don't need to touch so often? cc @sourcegraph/cody-security 

I will research how other linters do that. Sorry for the churn @umpox and please ignore it for now in this PR 🥺 🙏

umpox:
:D @philipp-spiess No probs

PR #3798, opened 2024-04-15
Chat: Remove fusion experiment and run context fetching in parallel

philipp-spiess:
This removes the chat-fused-context experiment after we deemed it to not be an improvement in https://sourcegraph.slack.com/archives/C06K09TB8UR/p1712962309623689?thread_ts=1712831832.810399&cid=C06K09TB8UR

Time to reduce the number of branches we have in context fetching 😬 

While here, I've migrated some of the ideas of running the context fetchers in parallel to the existing enhanced context function. We no longer depend on results from local embeddings to run local search, so this can now easily be parallized (while keeping the same order and number of results). 

## Test plan

- Tested on Cody Pro	 to see context results
- CI


dominiccooney (approved):
Looks good to me, some unused imports need to be removed.

Issue #3658, opened 2024-04-02
Make the VS Code version number copyable 


chillatom:
Can we make the version number of the extension able to be copied for easier referencing in tickets?  When on pre-release it can be a pain to always type the full version number

<img width="376" alt="image" src="https://github.com/sourcegraph/cody/assets/8737510/0cda2022-33ad-4488-896b-4d5ece6f2cef">

PR #3802, opened 2024-04-15
Sidebar: add copy extension version command

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3658

image

This pull request introduces a new command `cody.copy.version` that allows users to copy the current version of the Cody extension to their clipboard. This can be helpful when reporting issues or seeking support, as providing the extension version is often necessary for troubleshooting purposes.

## Changes summary

Update package.json:

- Add a new command entry for cody.copy.version under the "Cody Debug" category with an appropriate icon and title.
- Add a new keybinding for the cody.copy.version command in the "cody.support.tree.view" context.

Modify main.ts:

- Import the version constant from the ./version module.
- Register the cody.copy.version command, which uses vscode.env.clipboard.writeText to copy the extension version to the clipboard.

Update TreeViewProvider.ts:

- Set the contextValue property on the tree item based on the provided item.contextValue value.

Modify treeViewItems.ts:

- Moved SupportSidebarItems into its own file `./support-items`.
- Added contextValue as an optional field for SupportSidebarItems

Updated issue-reporter.ts

- Moved it from its current location into src/utils.ts
- Use the `version` number from the function that gets the current extension version number instead of package.json

These changes add a convenient way for users to access and copy the Cody extension version, which can aid in troubleshooting and support scenarios. The new command is accessible through the command palette and the Cody support sidebar view.

## Test plan

copy-version

1. Build from this branch and start it in debug mode
2. Hover over the release note item in the support sidebar to verify:
  - A copy button shows up on hover
  - Clicking on the copy button will copy the version number listed on the sidebar

## Demo

https://github.com/sourcegraph/cody/assets/68532117/051e0867-9a41-4735-902c-625d54656d7a

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -10,6 +10,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 - Autocomplete: Add the proper infilling prompt for Codegemma when using Ollama. [pull/3754](https://github.com/sourcegraph/cody/pull/3754)
 - Chat: The new `Mixtral 8x22B Preview` chat model is available for Cody Pro users for preview. [pull/3768](https://github.com/sourcegraph/cody/pull/3768)
 - Chat: Add a "Pop out" button to the chat title bar that allows you to move Cody chat into a floating window. [pull/3773](https://github.com/sourcegraph/cody/pull/3773)
+  feat(vscode): add copy extension version command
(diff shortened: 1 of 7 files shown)


kalanchan (approved):

toolmantim:
Can we pass URL parameters to the support link that pre-fills the version? I’m unsure if our support system supports it, but makes sense to do the work for the user if we can.

PR #3807, opened 2024-04-16
use vscode-jsonrpc library for JSON-RPC implementation

sqs:
The [vscode-jsonrpc](https://www.npmjs.com/package/vscode-jsonrpc) library is from the VS Code team and is used by many VS Code extensions and language servers to implement the JSON-RPC 2.0 protocol. It is well tested, heavily used, actively maintained, and supports both Node and Web platforms.

Using this library means we can remove our own implementation of the protocol, which is (a) less code to maintain, (b) does not support the browser, and (c) is not spec-compliant (for example, it supports string values as request parameters, which are disallowed by the [JSON-RPC 2.0 spec](https://www.jsonrpc.org/specification#request_object)).

## Test plan

1. CI
2. Manually ran BFG (cody-engine) tests and confirm that `embeddings/set-token` works in the output channel
3. Manually tested with the CLI
4. Test with JetBrains
5. Compare protocol traffic by `diff -u`'ing the `build/sourcegraph/cody-agent-trace.json` files between runs of `./gradlew -PforceAgentBuild=true :runIDE` and `CODY_DIR=$PWD/../cody ./gradlew -PforceAgentBuild=true :runIDE`: the traces are identical (except for timing, timestamps, and non-important ordering) for a session with autocomplete and chat interactions.
6. QA test on JetBrains: https://github.com/sourcegraph/jetbrains/pull/1356 - [Slack thread](https://sourcegraph.slack.com/archives/C06GFD8GFCN/p1713601212183189)

### DIFF
--- a/vscode/src/jsonrpc/jsonrpc.ts
+++ b/vscode/src/jsonrpc/jsonrpc.ts
@@ -1,13 +1,8 @@
-import assert from 'node:assert'
-import type { ChildProcessWithoutNullStreams } from 'node:child_process'
 import { appendFileSync, existsSync, mkdirSync, rmSync } from 'node:fs'
 import { dirname } from 'node:path'
-import { Readable, Writable } from 'node:stream'
 
 import * as vscode from 'vscode'
-
-import { isRateLimitError, logError } from '@sourcegraph/cody-shared'
-
+import { type MessageConnection, Trace } from 'vscode-jsonrpc'
 import type * as agent from './agent-protocol'
 import type * as bfg from './bfg-protocol'
 import type * as contextRanking from './context-ranking-protocol'
@@ -18,14 +13,6 @@ type Notifications = bfg.Notifications &
     agent.Notifications &
     embeddings.Notifications &
     contextRanking.Notifications
-
-// This file is a standalone implementation of JSON-RPC for Node.js
-// ReadStream/WriteStream, which conventionally map to stdin/stdout.
-// The code assumes familiarity with the JSON-RPC specification as documented
-// here https://www.jsonrpc.org/specification
-// To learn more about how JSON-RPC protocols work, the LSP specification is
-// also a good read
-// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/
 
 // String literal types for the names of the Cody Agent protocol methods.
 export type RequestMethodName = keyof Requests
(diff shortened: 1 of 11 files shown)


vscode/src/jsonrpc/jsonrpc.ts
```diff
@@ -39,213 +27,12 @@ type ParamsOf<K extends MethodName> = (Requests & Notifications)[K][0]
 // Request result types. Note: notifications don't return values.
 type ResultOf<K extends RequestMethodName> = Requests[K][1]
 
-type Id = string | number
-
-// Error codes as defined by the JSON-RPC spec.
-enum ErrorCode {
```
valerybugakov, line 45: It's so lovely to see so many red lines!

olafurpg: ++++1

olafurpg:
I'm looking into the git conflict. I didn't intend my websocket PR to conflict with this PR. I don't mind if the websocket functionality breaks

olafurpg:
@sqs I rebased and fixed the websocket implementation to use the new vscode-jsonrpc API, which is btw much nicer to use than the old custom implementation.

I didn't want to force push to your branch. Feel free to rebase on top of main and delete changes from my websocket PR that are causing conflicts/errors. I can send a separate PR to fix the websocket demo
 https://github.com/sourcegraph/cody/compare/olafurpg/jsonrpc-rebased?expand=1

sqs:
@olafurpg Great, thank you!

Instead of fallbackHandler in the old impl, I think you can use the no-method-name overload of `onRequest` and `onNotification`.

olafurpg:
Yes, i got it working with that API. It’s much cleaner than the old implementation https://github.com/sourcegraph/cody/compare/olafurpg/jsonrpc-rebased?expand=1#diff-19e9fdb0f04bd6f4c5559271fea95b9e91ed400c1c9d246df0e58a964367fe6fR27

I’ll send a separate PR to fix web sockets once this PR lands.

sqs:
I will merge this as the QA report came back clean (no add'l flaws introduced by this PR, confirmed in https://sourcegraph.slack.com/archives/C06GFD8GFCN/p1713820552328229 and https://github.com/sourcegraph/jetbrains/pull/1368#issuecomment-2071995482).

PR #3810, opened 2024-04-16
PromptString linter: Only error when the reported violation is within one of the changed range

philipp-spiess:
As the title suggest, we only want to emit linter errors when new call sites were added. The previous rule would fire too often.

## Test plan

- Add a change to the bottom of `vscode/src/chat/chat-view/SimpleChatPanelProvider.ts` that adds an `unsafe_` API call
- Run `pnpm ts-node lints/safe-prompts.ts `pnpm ts-node lints/git-diff-ts-ranges.ts`
- Observe that only an error for this change is emitted.

### DIFF
--- a/lints/git-diff-ts-ranges.ts
+++ b/lints/git-diff-ts-ranges.ts
@@ -1,0 +1,51 @@
+import { exec } from 'node:child_process'
+
+const EXTENSIONS = ['ts', 'tsx']
+
+exec('git diff --unified=0 origin/main', (error, stdout, stderr) => {
+    if (error) {
+        console.error(`exec error: ${error}`)
+        return
+    }
+    if (stderr) {
+        console.error(`stderr: ${stderr}`)
+        return
+    }
+
+    // Process the output
+    const lines = stdout.split('\n')
+    let currentFile = ''
+    const fileRanges: { [key: string]: string[] } = {}
+
+    for (const line of lines) {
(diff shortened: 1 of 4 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1255,3 +1255,6 @@ function newChatModelFromSerializedChatTranscript(
 function isAbortError(error: Error): boolean {
     return error.message === 'aborted' || error.message === 'socket hang up'
 }
+
+// Only this unsafe usage should be report
+PromptString.unsafe_fromUserQuery('EVIL!')
```
philipp-spiess, line 1260: As you can see only this violation was reported, not the other violations in this file

lints/git-diff-ts-ranges.ts
```diff
@@ -0,0 +1,51 @@
+import { exec } from 'node:child_process'
+
+const EXTENSIONS = ['ts', 'tsx']
+
+exec('git diff --unified=0 origin/main', (error, stdout, stderr) => {
```
philipp-spiess, line 5: This is 100% cody's doing but I validated it against @umpox's PR: https://github.com/sourcegraph/cody/pull/3797 and the ranges it yields seem great!

umpox (approved):
Thanks for the quick fix :)

Issue #3815, opened 2024-04-16
preserve whitespace in lexical editor


sqs:
![image](https://github.com/sourcegraph/cody/assets/1976/0f41d036-6652-4b1a-b70a-4d6f90fea5f3)

PR #3817, opened 2024-04-16
Chat: render spacings for human messages

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3815

This commit adds a new CSS class `.human-content` that applies `white-space: pre-wrap` styling specifically to human messages in the chat transcript. 

It also updates the `MessageCell` component to conditionally apply this new class to message cells if `message.speaker === 'human'`.

This enables messages submitted by users to render with correct spacing as a temporary workaround until the new design has been fully implemented.

### Before

image

### After

image


sqs (approved):
awesome thank you!

Issue #3766, opened 2024-04-10
bug: Link to enterprise context files are broken


kalanchan:
### Version

1.12

### Describe the bug
It seems like the encoded URL does not render a path properly on the sourcegraph instance after retrieving context

This was tested on S2 - if you click on any of the context files received from the enterprise instance, it brings you to a page not found

![CleanShot 2024-04-10 at 11 24 08@2x](https://github.com/sourcegraph/cody/assets/51868853/d97dfd15-1df0-41f9-8963-5384f6e15cc0)
![CleanShot 2024-04-10 at 11 24 45@2x](https://github.com/sourcegraph/cody/assets/51868853/e93569c5-5faa-4911-8df3-649f68002c46)

### Expected behavior

Context files should open correctly in the sourcegraph instance

kalanchan:
I can confirm that we are encoding the "@" sign properly from VS Code, this is the href that gets added to the dropdown links - `https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody%40c69aaf7/-/blob/README.md`. 

Will raise this with the Code Search team that they are not decoding the URLs properly

keegancsmith:
Reading the code here it hasn't changed in a long time and we explicitly do not decode when looking for the @ sign. I imagine we do this on purpose so we support repo names with @ in them (for example this happens with package repos).

I think cody needs to be updated so it doesn't encode the @ sign.

https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@d2a36127f9414f26b0ee1f4f44d37f92891e95d1/-/blob/client/shared/src/util/url.ts?L419-432

PR #3818, opened 2024-04-16
Chat: skip encoding of file links in webviews

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713294935223619

>NOTE: Temporary workaround for https://github.com/sourcegraph/cody/issues/3766 until the root cause is fixed by the search team 

The FileLink component was previously encoding the URI path unconditionally when generating the href attribute, even if the path already contained encoded characters like '@'. This was causing double encoding in some cases.

Example: [sourcegraph/cody%254036441658b4f15d8183125c1680c2c11d33889712](https://demo.sourcegraph.com/github.com/sourcegraph/cody%254036441658b4f15d8183125c1680c2c11d33889712/-/blob/.vscode/settings.json?L1-8) v.s. [sourcegraph/cody@36441658b4f15d8183125c1680c2c11d33889712](https://demo.sourcegraph.com/github.com/sourcegraph/cody@36441658b4f15d8183125c1680c2c11d33889712/-/blob/.vscode/settings.json?L1-8)

This change updates the component to only encode the URI path if it does not already contain '@'. This prevents double encoding while still ensuring the path is encoded when needed.

## Summary

- In FileLink.tsx, update the call to uri.toString() to pass true to skip encoding the URI
- This fixes an issue where file links in webviews were incorrectly encoding the URI path, causing problems with special characters
- The fix allows the raw URI path to be used without encoding for the link href

## Test plan

1. Start this branch in debug mode
2. Connect Cody to an Enterprise instance
3. Ask Cody a question with enhanced context enabled
4. Click on the context links to verify the files are opened in your browser correctly

### After

https://github.com/sourcegraph/cody/assets/68532117/e987985a-3d84-46f1-b6c0-ac34adbc1e6f


philipp-spiess (approved):
Noice

keegancsmith:
FYI this can't be a temporary workaround, I believe this is how it has to work unfortunetly in sourcegraph due to repo names that can contain `@`. Suggestions:
- Add a test here since this feels easy to mess up in the future.
- Maybe a new sourcegraph endpoint where we can pass in structured data and it redirects to the user URL we expect?

Issue #3819, opened 2024-04-16
Add Telemetry to count the sizes of files that users add manually and whether they exceed the limit


abeatrix:
Add Telemetry:
- [ ] count the sizes of files that users add manually
- [ ] whether they exceed the limit

PR #3823, opened 2024-04-16
Telemetry: log context stats and update response log to use new limit

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3819

This pull request enhances the telemetry data collected during chat sessions by adding information about the size of context files added by users, including those that were ignored during prompt construction.

The key changes are:

- Updates the setLastMessageContext method to include both used and ignored context items from the user.
- Adds a new `privateContextStats` object to the telemetry data under `privateMetadata`, containing statistics about the size of user-added context files (count, file sizes).
- Log the `privateContextStats` in the `cody.chat-question/executed ` telemetry event.
- Update various types to reflect that newContextIgnored is now required rather than optional.
- `promptText` and `responseText` to be truncated using the new Input and Output limits
- Remove `contextSummary` from `privateMetadata` since we are currently logging it twice in both metadata and `privateMetadata` for v2 events
- updated the number in vscode/src/editor/utils/editor-context.ts due to match the comment string

According to @akalia25:
>metadata: This gets sent to us on every event for both dotcom and customers, but it can only contain numeric values
> privateMetadata: Only gets sent for dotcom users, but can contain arbitrary string values

This allows us to better understand the size of context users are adding themselves + the size of context that gets filtered out or ignored. 

## Test plan

Check output channel to verify:

```
█ telemetry-v2: recordEvent: cody.chat-question/executed  {
  "parameters": {
    "version": 0,
    "metadata": [
      {
        "key": "user",
        "value": 1
      },
      {
        "key": "recordsPrivateMetadataTranscript",
        "value": 1
      }
    ],
    "privateMetadata": {
      "properties": {
        "requestID": "9d7afc85-72d6-47df-a5bd-84098261688c",
        "chatModel": "anthropic/claude-3-sonnet-20240229",
        "source": "chat",
        "traceId": "9ee0a6555083e73a87c25293f9e5a25d",
        "sessionID": "Wed, 17 Apr 2024 15:50:10 GMT"
      },
      "privateContextStats": {
        "included": {
          "countFiles": 1,
          "fileSizes": [
            372
          ]
        },
        "excluded": 0
      },
      "promptText": "summarize @vscode/src/chat/chat-view/chat-helpers.ts "
    }
  },
  "timestamp": "2024-04-17T17:14:11.702Z"
}
```

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -57,6 +57,10 @@ import type {
     ContextItemWithContent,
 } from '@sourcegraph/cody-shared/src/codebase-context/messages'
 import { ModelUsage } from '@sourcegraph/cody-shared/src/models/types'
+import {
+    CHAT_INPUT_TOKEN_BUDGET,
+    CHAT_OUTPUT_TOKEN_BUDGET,
+} from '@sourcegraph/cody-shared/src/token/constants'
 import { recordErrorToSpan, tracer } from '@sourcegraph/cody-shared/src/tracing'
 import { getContextFileFromCursor } from '../../commands/context/selection'
 import type { EnterpriseContextFactory } from '../../context/enterprise-context-factory'
@@ -412,7 +416,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                 source,
                 command,
                 traceId: span.spanContext().traceId,
-                sessionID: `${vscode.env.sessionId}@${this.chatModel.sessionID}`,
+                sessionID: this.chatModel.sessionID,
             }
             telemetryService.log('CodyVSCodeExtension:chat-question:submitted', sharedProperties)
             telemetryRecorder.recordEvent('cody.chat-question', 'submitted', {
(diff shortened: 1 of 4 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -491,12 +494,13 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
                         privateMetadata: {
                             properties,
+                            privateContextStats,
                             // 🚨 SECURITY: chat transcripts are to be included only for DotCom users AND for V2 telemetry
                             // V2 telemetry exports privateMetadata only for DotCom users
                             // the condition below is an additional safeguard measure
                             promptText:
                                 authStatus.isDotCom &&
-                                inputText.toString().substring(0, MAX_BYTES_PER_FILE),
+                                inputText.toString().substring(0, CHAT_INPUT_TOKEN_BUDGET),
```
akalia25, line 503: should we be using the same `truncatePromptString(inputText, CHAT_INPUT_TOKEN_BUDGET)` here as well?

abeatrix: Oh yes, good catch!!!

akalia25 (approved):
Thank you for keeping track of the truncation we're doing on transcripts and keeping this updated 💪

PR #3832, opened 2024-04-17
Add a hover title to the LLM chat avatars

toolmantim:
Adds a hover tooltip to the LLM model avatars in the chat transcript, so you can see what the icon means. Ideally we include the model name in the cell text, not just the provider avatar, but this is a small improvement in the meantime.

CleanShot 2024-04-17 at 21 30 52@2x

## Test plan

- Tested in webview

### DIFF
--- a/vscode/webviews/chat/cells/messageCell/SpeakerIcon.tsx
+++ b/vscode/webviews/chat/cells/messageCell/SpeakerIcon.tsx
@@ -17,7 +17,13 @@ export const SpeakerIcon: FunctionComponent<{
     if (message.speaker === 'human') {
         return <UserAvatar user={user} size={size} />
     }
-
-    const ModelIcon = chatModel ? chatModelIconComponent(chatModel.model) : null
-    return ModelIcon ? <ModelIcon size={size * 0.75} /> : null
+    if (!chatModel) {
+        return null
+    }
+    const ModelIcon = chatModelIconComponent(chatModel.model)
+    return ModelIcon ? (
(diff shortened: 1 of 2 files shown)


vscode/webviews/chat/cells/messageCell/SpeakerIcon.tsx
```diff
@@ -17,7 +17,13 @@ export const SpeakerIcon: FunctionComponent<{
     if (message.speaker === 'human') {
         return <UserAvatar user={user} size={size} />
     }
-
-    const ModelIcon = chatModel ? chatModelIconComponent(chatModel.model) : null
-    return ModelIcon ? <ModelIcon size={size * 0.75} /> : null
+    if (!chatModel) {
+        return null
+    }
+    const ModelIcon = chatModelIconComponent(chatModel.model)
```
umpox, line 23: nit: I think we can update this function to never return `null`, then we don't need the extra `ModelIcon ?` bit at the bottom

umpox (approved):

abeatrix (approved):

vscode/webviews/chat/cells/messageCell/SpeakerIcon.tsx
abeatrix: +1. I think it fallbacks to use Cody's logo if no matches, so it should never return null.

toolmantim: Good point, just fixed!

PR #3833, opened 2024-04-17
Empty selection should be translated to cursor

pkukielka:
## Test plan

Manual QA on the jetbrains side, I will try to add some automated tests.


agent/src/agent.ts
```diff
@@ -361,6 +362,28 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+                        by: 'line',
+                        value: start.line,
+                    })
+                    .then(() =>
+                        vscode.commands.executeCommand('cursorMove', {
+                            to: 'right',
+                            by: 'character',
+                            value: start.character,
+                        })
+                    )
```
mkondratek, line 385: don't we send the cursor position in any other place? 🤔 is it the regular way to update it?

mkondratek (approved):

Issue #3816, opened 2024-04-16
fix color theme for context links


sqs:
Many VS Code themes (including default themes) display the context links with unreadable colors. Need to fix the colors it uses.

![image](https://uploads.linear.app/15b96a4d-ffc0-481a-ba9c-9d840020f7bc/038d8686-341a-496a-8b7f-c2a7a1748ad2/cf5c09eb-a720-4fcf-ac0b-74079c5af31e?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzE1Yjk2YTRkLWZmYzAtNDgxYS1iYTljLTlkODQwMDIwZjdiYy8wMzhkODY4Ni0zNDFhLTQ5NmEtOGI3Zi1jMmE3YTE3NDhhZDIvY2Y1YzA5ZWItYTcyMC00ZmNmLWFjMGItNzQwNzljNWFmMzFlIiwiaWF0IjoxNzE0Nzc0NzE0LCJleHAiOjMzMjg1MzM0NzE0fQ.55rnVCC5tI-6SGLllQndp9lWLdDaXkfnjAhw_QM9u80 "=1136x")

sqs:
Fixed in most themes but not the Sourcegraph Dark theme

PR #3841, opened 2024-04-17
improve contrast of context links and at-mention colors in various VS Code themes

sqs:
Some VS Code color themes made the context links and at-mentions in chat unreadable. This was because the styling assumed that the text link foreground had contrast with the list active selection background, which is an incorrect assumption in general. Now the styles are formally shared between these 2 components and use CSS theme variables that can be assumed to have appropriate contrast.

Fix https://github.com/sourcegraph/cody/issues/3816

TBH this regressed the consistency between the at-mentions in the editor and the list somewhat, but they weren't consistent before and we'll need to clean up the button styles or something else. This is unobjectionable and fixes the obvious bug, so I think it's worth merging for the release.

image
image

## Test plan

CI. Also view chat and switch themes.

### DIFF
--- a/vscode/webviews/chat/cells/contextCell/ContextCell.module.css
+++ b/vscode/webviews/chat/cells/contextCell/ContextCell.module.css
@@ -52,6 +52,4 @@
 
 .file-link {
     padding: 2px 4px 2px 2px;
-    background-color: var(--vscode-list-activeSelectionBackground);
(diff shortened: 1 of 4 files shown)


abeatrix (approved):
Tried it with different themes and they all look good to me 🎉 

image
image
image
image
image
image


## Before

image

PR #3842, opened 2024-04-17
Chat: display line range for @-mentions

abeatrix:
## Issue

When @-mentioning a file without specifying the range, the file will be displayed without line numbers, and without the number, the Chat UI will not be able to display the correct number for total lines read:
image

"5 lines from 2 files" is not correct here, because the line count from the ignore file is not included here:
image

## Summary

This PR fixed the aforementioned issue with the following changes:

- Destructure commonly used properties like `content` and `size` upfront
- Remove separate `content` and `size` assignments in favor of single conditional block
- Introduce `getRangeByContentLineCount` to infer range based on content lines

## Test plan

Updated all unit tests and e2e tests.

### Before

image

### After

image


kalanchan (approved):

toolmantim:
Nice!

PR #3843, opened 2024-04-17
Graph context: add tsc-mixed strategy

olafurpg:
This PR adds a new graph context strategy `tsc-mixed` that uses the TypeScript compiler (aka. `tsc`) directly to provide symbol and type information. The benefit of using tsc is that we get full access to typed syntax trees and symbol table, giving us fine-grained control over what symbols to include in the context and how to render the signatures. In comparison, with LSP, we are limited to a small number of endpoints that don't always give us the data we need (or at least not in a performant API).

Marked as draft since it's currently a super basic approach.

Fixes CODY-1293
## Test plan

See added agent tests.


olafurpg:
This PR is finally ready for review! I spent most of the time getting the testing infrastructure in place so the diff has several non-graph related changes:

- New `TestWorkspace` class to abstract away code in `index.test.ts` because I didn't want to add graph context tests to `index.test.ts` (it's already too big).
- New bash scripts to simplify dealing with git conflicts and managing HTTP recordings.
- Improved organization of authorization parameters (token + server endpoint). This logic was spread all over the place in `index.test.ts` and I wasted quite a lot of time debugging issues related to these tokens.
- `SymbolFormatter` unit test suite that runs quickly and offline. I didn't want to rely on agent integration tests to iterate on the rendering of TSC symbols.
- `graph-context.test.ts` integration test suite that is a small file that's independent of `index.test.ts`. I'd like new features to use the same pattern instead of adding more and more tests to `index.test.ts`

olafurpg:
The Windows test failures are legitimate, `tsc-retriever` seems to be returning empty results. I will look into this on Monday.

olafurpg (commented):
BTW, I am planning to follow up with a separate PR to remove streaming LLM responses in the HTTP recordings where possible. It's not sustainable to have +12k line diffs in a PR like this for a few simple autocomplete tests. I am anticipating that we will want to add thousands of more tests like these in the long-run.

vscode/src/completions/completion-provider-config.ts
```diff
@@ -62,19 +62,24 @@ class CompletionProviderConfig {
...
+                : config.autocompleteExperimentalGraphContext === 'tsc'
+                  ? 'tsc'
+                  : config.autocompleteExperimentalGraphContext === 'bfg'
+                      ? 'bfg'
+                      : config.autocompleteExperimentalGraphContext === 'bfg-mixed'
+                          ? 'bfg-mixed'
+                          : config.autocompleteExperimentalGraphContext === 'local-mixed'
+                              ? 'local-mixed'
+                              : config.autocompleteExperimentalGraphContext === 'jaccard-similarity'
+                                  ? 'jaccard-similarity'
```
philipp-spiess, line 76: i wonder if the time has come to make this a switch instead lol

vscode/src/completions/context/retrievers/tsc/SymbolFormatter.ts
```diff
@@ -0,0 +1,310 @@
+// Copy-pasted and adapted code from scip-typescript
+import * as ts from 'typescript'
```
philipp-spiess, line 2: How much of a bundle size impact does it have to bundle the typescript compiler?

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
```diff
@@ -0,0 +1,338 @@
+import {
+    type AutocompleteContextSnippet,
+    type FileURI,
+    isFileURI,
+    isMacOS,
+    isWindows,
+    logError,
+} from '@sourcegraph/cody-shared'
+import { defaultPathFunctions } from '@sourcegraph/cody-shared/src/common/path'
```
philipp-spiess, line 9: Can you export this from the cody-shared barrel file instead to avoid issues like https://github.com/sourcegraph/cody/pull/3947

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
```diff
@@ -0,0 +1,338 @@
...
+    public retrieve(options: ContextRetrieverOptions): Promise<AutocompleteContextSnippet[]> {
+        return new Promise<AutocompleteContextSnippet[]>(resolve => {
+            try {
+                resolve(this.doBlockingRetrieve(options))
+            } catch (error) {
+                logError('tsc-retriever', String(error))
+                resolve([])
+            }
+        })
+    }
```
philipp-spiess, line 256: Should we wrap this in an otel span so we have some cheap telemetry? :)

philipp-spiess (approved):
Impressive! Can't wait to try this in action 🔥

olafurpg:
Thank you so much for the review @philipp-spiess ! I will address tomorrow. I spent all of today getting the tests working in Windows. It was surprisingly difficult to debug, primarily due to a bug in our custom implementation of `node:path` `dirname()` function. Either way, here's a lovely screenshot of the new context working in Windows :D 

image

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
olafurpg: Done. I had assumed we already capture telemetry for context retrievers in general.

PR #3846, opened 2024-04-17
Introduce a feature JSON tracking system

kalanchan:
As our features continue to expand across different editors and platforms, it's becoming increasingly important to have a centralized system to track and manage these features. The `features.json5` file will serve as a single source of truth for all the features we've built, their status, and the details of their implementation across different editors. 

**Note:** This is the most basic example of what it _could_ look like, the schema of `features.json5` will continue to be evolved, see [RFC 932](https://docs.google.com/document/d/10dCF0iJlBIf6p3qnGqfBltHDDGVnPYmeT03FeNmmVrg/edit#heading=h.8saknylfr2bz) for more preliminary discussions. 

The tooling to support this across different editors will come afterwards, treat this as a manual process maintained by EPD for now. 

## Test plan
N/A

### DIFF
--- a/vscode/features.json5
+++ b/vscode/features.json5
@@ -2,8 +2,8 @@
     "editors": ["vscode", "jetbrains"],
     "features": [
         {
-        "name": "NotebookBasedChatUI",
-        "description": "new chat UI that resembles a notebook instead of a chat session",
+        "name": "NotebookChatUI",
+        "description": "Chat UI that resembles a Jupyter notebook instead of a chat session",
         "productLine": "all",
         "documentation": [],
         "editors": {
@@ -11,12 +11,12 @@
             "status": "stable",
           }
         },
-        tags: ['chat']
+        tags: ["chat"]
       },
       {
         "name": "Mixtral8x22BPreview",
-        "description": "Mixtral 8x22B Preview chat model available for Cody Pro Users",
-        "productLine": "consumer",
+        "description": "Mixtral 8x22B Preview chat model",
+        "productLine": "pro",
         "documentation": [],
         "editors": {
           "vscode": {
--- a/vscode/features.json5
+++ b/vscode/features.json5
@@ -1,113 +1,102 @@
 {
     "editors": ["vscode", "jetbrains"],
-    "features": {
-      "chat": [      
+    "features": [
         {
-        "name": "notebookBasedChatUI",
-        "productLine": "both",
-        "telemetry": [],
+        "name": "NotebookBasedChatUI",
+        "description": "new chat UI that resembles a notebook instead of a chat session",
+        "productLine": "all",
         "documentation": [],
         "editors": {
           "vscode": {
             "status": "stable",
-            "notes": "new chat UI that resembles a notebook instead of a chat session",
-            "version": "v1.14.0"
           }
-        }
+        },
+        tags: ['chat']
       },
       {
-        "name": "mixtral8x22BPreview",
+        "name": "Mixtral8x22BPreview",
+        "description": "Mixtral 8x22B Preview chat model available for Cody Pro Users",
         "productLine": "consumer",
-        "telemetry": [],
         "documentation": [],
         "editors": {
           "vscode": {
             "status": "stable",
-            "notes": "Mixtral 8x22B Preview chat model available for Cody Pro Users",
-            "version": "v1.14.0"
           }
-        }
+        },
+        tags: ['chat']
       },
       {
-        "name": "toggleEnhancedContextButtons",
-        "productLine": "both",
-        "telemetry": [],
+        "name": "ToggleEnhancedContextButtons",
+        "description": "Added buttons in the chat input box for enabling/disabling Enhanced Context.",
+        "productLine": "all",
         "documentation": [],
         "editors": {
           "vscode": {
             "status": "stable",
-            "notes": "Added buttons in the chat input box for enabling/disabling Enhanced Context.",
-            "version": "v1.12.0"
           }
-        }
-      }
-    ],
-    "commands": [{
-        "name": "generateUnitTestCodeAction",
-        "productLine": "both",
-        "telemetry": [],
+        },
+        tags: ['chat']
+      },
+       {
+        "name": "GenerateUnitTestCodeAction",
+        "description": "Generate unit test now shows on any function symbol in code actions",
+        "productLine": "all",
         "documentation": [],
         "editors": {
           "vscode": {
             "status": "stable",
-            "notes": "Generate unit test now shows on any function symbol in code actions",
-            "version": "v1.14.0"
           }
-        }
+        },
+        tags: ['commands', 'unitTest']
       },
       {
-        "name": "editPreInstructions",
-        "productLine": "both",
-        "telemetry": [],
+        "name": "EditPreInstructions",
+        "description": "New cody.edit.preInstruction configuration option for adding custom instruction at the end of all your requests.",
+        "productLine": "all",
         "documentation": [],
         "editors": {
           "vscode": {
             "status": "stable",
-            "notes": "New cody.edit.preInstruction configuration option for adding custom instruction at the end of all your requests.",
-            "version": "v1.12.0"
           }
-        }
+        },
+        tags: ['commands', 'edit']
       },
       {
-        "name": "editShowWarningsOnLargeFiles",
-        "productLine": "both",
-        "telemetry": [],
+        "name": "EditShowWarningsOnLargeFiles",
+        "description": "Display warnings for large @-mentioned files during selection",
+        "productLine": "all",
         "documentation": [],
         "editors": {
           "vscode": {
             "status": "stable",
-            "notes": "Display warnings for large @-mentioned files during selection",
-            "version": "v1.12.0"
           }
-        }
-      }],
-      "autocomplete": [{}],
-      "context": [{
-        "name": "rightClickAddContext",
-        "productLine": "both",
-        "telemetry": [],
-        "documentation": [],
-        "editors": {
-          "vscode": {
-            "status": "stable",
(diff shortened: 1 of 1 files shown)


vscode/features.json5
```diff
@@ -0,0 +1,113 @@
+{
+    "editors": ["vscode", "jetbrains"],
+    "features": {
+      "chat": [
```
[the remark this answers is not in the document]
dominiccooney: Let's give each feature a `"tags": [string]` and use tags. Way more flexible. (I'm wondering if we should remove "productLine" and just do that with tags too...)

vscode/features.json5
```diff
@@ -0,0 +1,113 @@
...
+        {
+        "name": "notebookBasedChatUI",
+        "productLine": "both",
+        "telemetry": [],
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+            "notes": "new chat UI that resembles a notebook instead of a chat session",
+            "version": "v1.14.0"
```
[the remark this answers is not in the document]
dominiccooney: As we discussed offline, no `version`

vscode/features.json5
```diff
@@ -0,0 +1,113 @@
+{
+    "editors": ["vscode", "jetbrains"],
+    "features": {
+      "chat": [      
+        {
+        "name": "notebookBasedChatUI",
+        "productLine": "both",
```
dominiccooney, line 7: If we keep productLine, maybe we should make this `[string]`.

vscode/features.json5
```diff
@@ -0,0 +1,113 @@
...
+      {
+        "name": "toggleEnhancedContextButtons",
+        "productLine": "both",
+        "telemetry": [],
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+            "notes": "Added buttons in the chat input box for enabling/disabling Enhanced Context.",
+            "version": "v1.12.0"
```
dominiccooney, line 40: For "legacy" features, git history won't tell us which version they were added in. This is a good use for "notes" for now. If we really need something to rewrite the history of legacy features, we can add a workaround later.

vscode/features.json5
```diff
@@ -0,0 +1,113 @@
...
+      "chat": [      
+        {
+        "name": "notebookBasedChatUI",
+        "productLine": "both",
+        "telemetry": [],
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+            "notes": "new chat UI that resembles a notebook instead of a chat session",
```
dominiccooney, line 13: This isn't right, it's a description of the feature, but you've put it as a note in the vscode editor.

If you want to have a "description" field at the feature level, go for it.

We *do* want to keep notes per-editor. Optional field, used rarely, used for pointing to bugs/caveats that rise to the FAQ level of bother. When we display the status, we make these notes visible.

The reason we need the notes field: Without it, you get all these corner cases where something is half implemented but it is the same feature and people feel uncertainty about making it a separate "feature" because it is so different, or the same feature but it has these massive caveats, they can't describe anywhere.

vscode/features.json5
```diff
@@ -0,0 +1,113 @@
...
+        "editors": {
+          "vscode": {
+            "status": "stable",
+            "notes": "right click to add add highlighted code to the chat context",
+            "version": "v1.14.0"
+          }
+        }
+      },
+      {
+        "name": "contextTokenCounter",
```
dominiccooney, line 99: We should keep the feature names alphabetized. If you have people adding at the bottom, it generates unnecessary git conflicts.

If you want to do it later, that's fine. We'll add a lint to make sure people are writing valid syntax, spelling field names right, etc.

vscode/features.json5
```diff
@@ -0,0 +1,113 @@
...
+            "status": "stable",
+            "notes": "new chat UI that resembles a notebook instead of a chat session",
+            "version": "v1.14.0"
+          }
+        }
+      },
+      {
+        "name": "mixtral8x22BPreview",
+        "productLine": "consumer",
+        "telemetry": [],
```
dominiccooney, line 21: I'd like to have these optional and use empty arrays to mean: know to be none, whereas missing means "unknown."

dominiccooney (approved):
Directionally right. Some feedback inline but we can also iterate on this.

Question for you: how did it feel providing this data?

vscode/features.json5
kalanchan: I'm thinking single string makes more sense here? I replaced `both` with `all`, but `all` would imply everything, whereas single string `consumer`, `enterprise` means isolated feature just for that product line.

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
...
+      },
+      {
+        "name": "ContextTokenCounter",
+        "description": "introduces token counting during prompt and context building",
+        "productLine": "all",
+        "documentation": ["https://docs.google.com/document/d/1_rg8j4yfz1Ot-lQle8v_K_y8V3JDzpKYxO134X7MItc/edit"],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+              "notes": "expanded context window in v1.14.0 and above"
```
kalanchan, line 96: @dominiccooney I updated the `ContextTokenCounter` entry as an example of what notes within the editors section would look like, does this align with what you're thinking?

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
+{
+    "editors": ["vscode", "jetbrains"],
+    "features": [
+        {
+        "name": "NotebookBasedChatUI",
+        "description": "new chat UI that resembles a notebook instead of a chat session",
+        "productLine": "all",
```
kalanchan, line 7: @dominiccooney I'm thinking single string makes more sense here? I replaced `both` with `all`, but `all` would imply everything, whereas single string `consumer`, `enterprise` means isolated feature just for that product line. 

what do you think is better?

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
...
+        "name": "GenerateUnitTestCodeAction",
+        "description": "Generate unit test now shows on any function symbol in code actions",
+        "productLine": "all",
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+          }
+        },
+        tags: ['commands', 'unitTest']
```
kalanchan, line 50: does this make sense for unit test?

kalanchan (commented):
@dominiccooney made some changes!

vscode/features.json5
dominiccooney: The day will come when we add a product line and we have to blaze through here turning this into an array. But this is fine for now.

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
+{
+    "editors": ["vscode", "jetbrains"],
+    "features": [
+        {
+        "name": "NotebookBasedChatUI",
+        "description": "new chat UI that resembles a notebook instead of a chat session",
```
dominiccooney, line 6:
```suggestion
        "description": "Chat UI that resembles a Jupyter notebook instead of a chat session",
```

- Let's avoid "new" because that will become inaccurate over time.
- I'm mentioning Jupyter here so people don't get confused with a paper notebook.

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
+{
+    "editors": ["vscode", "jetbrains"],
+    "features": [
+        {
+        "name": "NotebookBasedChatUI",
```
dominiccooney, line 5:
```suggestion
        "name": "NotebookChatUI",
```

Let's be brief.

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
...
+        "description": "new chat UI that resembles a notebook instead of a chat session",
+        "productLine": "all",
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+          }
+        },
+        tags: ['chat']
```
dominiccooney, line 14:
```suggestion
        tags: ["chat"]
```

Nit, don't needlessly vary choice of quote.

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
...
+      },
+      {
+        "name": "Mixtral8x22BPreview",
+        "description": "Mixtral 8x22B Preview chat model available for Cody Pro Users",
```
dominiccooney, line 18:
```suggestion
        "description": "Mixtral 8x22B Preview chat model",
```

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
...
+        "productLine": "consumer",
```
dominiccooney, line 19:
```suggestion
        "productLine": "pro",
```

vscode/features.json5
dominiccooney: Sure. We won't get the tags right at first, we can chip away at it. So you're thinking of this as tagging a use case?

vscode/features.json5
dominiccooney: I think this is perfect.

After this gets rolling, it will be unusual to mention versions because we'll assume the notes apply to the version this is checked in with. But in this case, mentioning the version makes perfect sense.

dominiccooney (approved):
Looks great. I'm going to go and have a crack at the JetBrains version.

By the way, I have heard "consumer" referred to as "self-serve". Which do you think has more sway?

vscode/features.json5
```diff
@@ -0,0 +1,102 @@
+{
+    "editors": ["vscode", "jetbrains"],
+    "features": [
+        {
```
dominiccooney, line 4: Sat down to do some processing with this, and I think 'features' should be an object keyed by the name.

Reason is:
- All features need names.
- When we do the merge, we want to join by feature name.

An array can work, but we're going to always be converting it into a map keyed by name, then converting it *back* into an array with name as a property.

PR #3851, opened 2024-04-18
Convert editCommands/code to a proper PromptString

philipp-spiess:
The introduction of PromptString did not catch this case where an Agent command was converted into the arguments for Edit. I added a cast to `ExecuteEditArguments` now so that the next time this will yell.

`command.executeCommand()` being untyped is not great 😞 

## Test plan

- CI; Will ask @sourcegraph/jetbrains-plugin folks

### DIFF
--- a/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/EditCommands_CodeParams.kt
+++ b/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/EditCommands_CodeParams.kt
@@ -2,6 +2,7 @@
 package com.sourcegraph.cody.protocol_generated
 
 data class EditCommands_CodeParams(
-  val params: ParamsParams? = null,
+  val instruction: String? = null,
+  val model: String? = null,
 )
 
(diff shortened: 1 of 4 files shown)


agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/EditCommands_CodeParams.kt
```diff
@@ -2,6 +2,7 @@
 package com.sourcegraph.cody.protocol_generated
 
 data class EditCommands_CodeParams(
-  val params: ParamsParams? = null,
+  val instruction: String? = null,
+  val model: String? = null,
```
philipp-spiess, line 6: @pkukielka Since the JB client was passing model before and the type here seems to not match with the production payload: any chance the types are not working on the JB side? 🤔 it wouldn't have caught the `PromptString` issue but should have yelled about passing a model where non was expecte.

pkukielka: Type on JB side is `data class InlineEditParams(val instruction: String, val model: String)`
I suspect that model param was simply ignored so far.

pkukielka (approved):
LGTM

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/EditCommands_CodeParams.kt
philipp-spiess: Why do we generate these kotlin types if we don't use it on the JB side? 🤔

PR #3852, opened 2024-04-18
Add simple test for editCommands/code

pkukielka:
## Test plan

N/A, this is just adding tests

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -10,6 +10,7 @@ import { afterAll, beforeAll, beforeEach, describe, expect, it } from 'vitest'
 import { ModelUsage, isWindows } from '@sourcegraph/cody-shared'
 
 import { URI } from 'vscode-uri'
+import type { RequestMethodName } from '../../vscode/src/jsonrpc/jsonrpc'
 import { TestClient, asTranscriptMessage } from './TestClient'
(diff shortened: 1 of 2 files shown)


agent/src/index.test.ts
```diff
@@ -730,6 +733,26 @@ describe('Agent', () => {
...
 
+    function checkEditCodeCommand(
+        documentClient: TestClient,
+        name: string,
+        filename: string,
+        instruction: string,
+        assertion: (obtained: string) => void
+    ): void {
+        const params = { instruction: instruction }
+        checkEditCommand(documentClient, 'editCommands/code', name, filename, { params }, assertion)
```
umpox, line 744:
```suggestion
        checkEditCommand(documentClient, 'editCommands/code', name, filename, { instruction }, assertion)
```

I think this is correct, as the function takes the object like this:

```
this.registerAuthenticatedRequest('editCommands/code', params => {
            const args = { configuration: { ...params } }
            return this.createEditTask(
                vscode.commands
                    .executeCommand<FixupTask | undefined>('cody.command.edit-code', args)
                    .then(task => task && { type: 'edit', task })
            )
        })
```

We could do with some better type-safety here

pkukielka: You are right, thank you very much!

umpox (approved):
Nice!

PR #3856, opened 2024-04-18
Cody Ignore: Check reference violations at I/O boundaries

philipp-spiess:
Part of https://github.com/sourcegraph/cody/issues/3642

This PR adds a "last-line-of-defense" that will throw when attempting to run an LLM query for cody-ignored files. I've added checks for references to all I/O boundaries (it should be close as close to the `fetch()` or `http/https` calls as possible so we never accidentally forget one if we copy-paste in the future.).

@valerybugakov I had to move some of your cody ignore work to `lib/shared` to make this work. Two notes:

- We make the git extension specific stuff dependency injectable so there's no hard requirement on vscode in `lib/shared`
- We need to validate that the WASM regex library actually works in the agent since it's loaded based on static data in the bundler. For now I kept the `re2-wasm` dependency in `vscode` too so the copy script has access to it. Let's add this as a todo item to check inside agent builds.

## Test plan

- CI
- This PR disables the context filter provider so it has no runtime impact


lib/shared/package.json
```diff
@@ -10,11 +10,7 @@
   },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
-  "files": [
-    "dist",
-    "src",
-    "!**/*.test.*"
-  ],
+  "files": ["dist", "src", "!**/*.test.*"],
```
philipp-spiess, line 13: I just did `pnpm add` and this was reformatted. 🤔 Do others change this file in their IDE maybe?

lib/shared/src/cody-ignore/context-filters-provider.test.ts
```diff
@@ -277,11 +279,7 @@ describe('ContextFiltersProvider', () => {
         function getTestURI(params: TestUriParams): URI {
             const { repoName, filePath } = params
 
-            vi.spyOn(repoNameResolver, 'getRepoNameFromWorkspaceUri').mockResolvedValue(
-                `github.com/sourcegraph/${repoName}`
-            )
```
philipp-spiess, line 282: This seems unused

valerybugakov: Yep, since now `repoNameResolver` is an `init` parameter.

valerybugakov (commented):
I got the same thoughts after looking at the PR diff! 

With both of the ideas, I'm concerned that we have many places where to instantiate prompt strings or call toString on them before filtering out the excluded URIs. In these places, we would need to update the logic to handle potential "URI is ignored" errors. 

Conversely, the risk of not applying the `isUriAllowed` is almost zero, which is excellent. As we want to be on the safer side of things, it makes sense to refactor the call sites to accommodate one of these ideas.

valerybugakov:
Heads up @philipp-spiess, I just rebased my changes on top of this PR: https://github.com/sourcegraph/cody/pull/3869

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -23,12 +19,23 @@ interface ParsedContextFilterItem {
...
+    // TODO: This is only a temporary off-switch until we can properly detect a
+    // valid context filter config in the upstream repository.
+    private temporary_isEnabled = true
+
+    async init(temporary_isEnabled: boolean, getRepoNameFromWorkspaceUri: GetRepoNameFromWorkspaceUri) {
+        this.temporary_isEnabled = temporary_isEnabled
+        if (!temporary_isEnabled) {
+            return
+        }
+        this.getRepoNameFromWorkspaceUri = getRepoNameFromWorkspaceUri
```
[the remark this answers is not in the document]
valerybugakov: If we move repoNameResolver to `lib/shared`, we should be able to remove `getRepoNameFromWorkspaceUri` from `init` parameters.

lib/shared/package.json
dominiccooney: I see this flip back and forth, sadly. I'm unsure what the configuration difference is.

lib/shared/src/cody-ignore/context-filters-provider.ts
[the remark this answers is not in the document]
dominiccooney: FWIW @valerybugakov  I'm removing the `check/isCodyIgnoredFile` Agent protocol method in #3858 and Agent protocol can now interact with vscode/... objects via [ExtensionClient and ExtensionObjects](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@63ae96cb4a7a430746cd383c1cc1d444dc2a050f/-/blob/vscode/src/extension-client.ts) There might be some startup ordering subtleties there but I'm sure we can iron them out.

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -65,6 +72,10 @@ export class ContextFiltersProvider implements vscode.Disposable {
     }
 
     public isRepoNameAllowed(repoName: string): boolean {
+        if (!this.temporary_isEnabled) {
```
dominiccooney, line 75: Can we expose a Promise for when this the temporary enablement is done?

If the repo fetcher starts churning it will want to wait before using the filters.

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -65,6 +72,10 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
+            return true
```
dominiccooney, line 76: If we're in an uncertain configuration, wouldn't it be more secure to fail *closed* and disallow everything? CC @willdollman

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -99,12 +110,16 @@ export class ContextFiltersProvider implements vscode.Disposable {
     }
 
     public async isUriAllowed(uri: vscode.Uri): Promise<boolean> {
+        if (!this.temporary_isEnabled) {
+            return true
+        }
+
         if (!isFileURI(uri)) {
             return false
```
dominiccooney, line 118: Not your change, but we really need a TODO here to follow up on cases like allowing unsaved documents.

lib/shared/src/sourcegraph-api/completions/browserClient.ts
```diff
@@ -13,6 +14,14 @@ export class SourcegraphBrowserCompletionsClient extends SourcegraphCompletionsC
         cb: CompletionCallbacks,
         signal?: AbortSignal
     ): void {
+        // Validate that no messages contain ignored context
```
dominiccooney, line 17: What about instead passing the context filter into PromptString and having that do the check? So the "only" way to get a string is to call the function that applies a policy?

I realize we have `PromptString.toString()` calls all over the place right now, but I wonder how many of them we really need...

dominiccooney (changes requested):
Some feedback inline.

There's too much duplication of the reference check. We should do that in PromptString and pass in the policy, so everyone is doing the same check. That will unit test in a more targeted way too.

I agree PromptString.toString is a risk, but I'm not worried about it in the short term as long as we rely on our invariants:
- Unsafe PromptString construction is audited, for example, that human input is something that just came out of a text box.
- Other PromptString construction is either safe (literals) or tracks its provenance.
- PromptStrings are what get sent to LLMs.

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: This won't be needed once we merge https://github.com/sourcegraph/cody/pull/3869 into this branch

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: We should use the async `isUriAllowed` instead of this one.

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: I have a subtask for that here https://github.com/sourcegraph/cody/issues/3641: "Process non-file URIs instead of excluding all of them".

valerybugakov:
Hey @philipp-spiess, I just merged my changes into this PR. It should not introduce any extra work or conflicts because I rebased on top of your work.

lib/shared/src/cody-ignore/context-filters-provider.ts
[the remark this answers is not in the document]
philipp-spiess: @valerybugakov I tried this at first but when I moved `repoNameResolver` to `lib/shared` I noticed it uses lots of vscode APIs (workspace/uri stuff/file type stuff). This is an easier boundary since it just means we need to dependency inject 1 function (which also makes testing simpler since we don't need mocking now!)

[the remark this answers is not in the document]
philipp-spiess: If we need it for cody agent we should think about how to get rid of the vscode-isms maybe? Anyway seems like it might be good to do in a follow-up PR

lib/shared/src/index.ts
```diff
@@ -127,6 +127,7 @@ export {
     CompletionStopReason,
     type CodeCompletionsClient,
     type CodeCompletionsParams,
+    type SerializedCodeCompletionsParams,
```
philipp-spiess, line 130: I have it as a small follow-up task for me to rename these to `Serializeable` instead of `Serialized`. No need to add that to that PR though or to worry about merge conflicts (we use `Serialized` in a few places right now where it refers to the serializebale json representation). Just ignore for now and I'll fix 🥺

philipp-spiess:
(Quickly fixing up something; looks like we changed the API to async 😓  and need to add a unit test)

lib/shared/src/sourcegraph-api/completions/browserClient.ts
philipp-spiess: The fact that the context filter chat is async makes it so that `toString` is no longer a good API boundary for this anyways, unfortunately.

lib/shared/src/cody-ignore/context-filters-provider.ts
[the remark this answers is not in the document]
valerybugakov: > I noticed it uses lots of vscode APIs (workspace/uri stuff/file type stuff)

The `workspace.fs` API should be supported by the agent: https://github.com/sourcegraph/cody/blob/d65d7c1a4af11e2c5c4e62b698be81ec236b1573/vscode/src/testutils/mocks.ts#L504 What kind of errors did you get?

> If we need it for cody agent we should think about how to get rid of the vscode-isms maybe?

We need to support this in the agent to make the Cody Ignore work for the JetBrains client. A follow-up PR for that sounds good 👍  I was hoping it would pick up the existing vscode mocks and work mostly out of the box without any additional logic.

lib/shared/src/sourcegraph-api/completions/browserClient.ts
dominiccooney: Perfect!

Yes, passing in the context filter is the right approach.

lib/shared/src/cody-ignore/context-filters-provider.ts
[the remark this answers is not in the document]
philipp-spiess: @valerybugakov IIRC I got weird errors in the test where the vscode import would not export anything at runtime, I can look into this as a follow up, though. Do you mind reviewing this PR so we can get it into `main`? :)

philipp-spiess:
@valerybugakov @dominiccooney may I ask for a review once more? The PR is starting to diverge from `main` quite a bit 😬

vscode/test/fixtures/mock-server.ts
```diff
@@ -486,8 +486,16 @@ export class MockServer {
                         break
                     }
                     default:
-                        res.sendStatus(400)
-                        res.statusMessage = `unhandled GraphQL operation ${operation}`
+                        res.status(400).send(
+                            JSON.stringify({
+                                errors: [
+                                    {
+                                        message: `Cannot query field "unknown" on type "${operation}".`,
```
valerybugakov, line 493: TY!

valerybugakov (approved):
Looks great! 🙌

One thing that worries me is that we still depend on our discipline to avoid accidentally passing the excluded context snippet through the I/O boundary. We can think about how to make it stricter in a follow-up.

One idea from looking at `vscode/src/completions/client.ts`: if we want to ensure that unfiltered strings are not used as fetch bodies, the default `toString` could add a special `[CODY_UNFILTERED]` suffix to the return value. We can create a `fetch` wrapper that checks if `body` includes it and throws. It would ensure we always use `toFilteredString` (which deletes the suffix) before sending it over the network. Potentially, we can also monitor it on the Cody Gateway side and error.

The bad part is that it modifies the string value until it's sent. It has multiple annoying side effects. E.g., we will need to account for the redundant suffix length when we calculate the chars remaining in the context window. Is there a better way?

philipp-spiess:
> One thing that worries me is that we still depend on our discipline to avoid accidentally passing the excluded context snippet through the I/O boundary. We can think about how to make it stricter in a follow-up.
> 
> One idea from looking at `vscode/src/completions/client.ts`: if we want to ensure that unfiltered strings are not used as fetch bodies, the default `toString` could add a special `[CODY_UNFILTERED]` suffix to the return value. We can create a `fetch` wrapper that checks if `body` includes it and throws. It would ensure we always use `toFilteredString` (which deletes the suffix) before sending it over the network. Potentially, we can also monitor it on the Cody Gateway side and error.
> 
> The bad part is that it modifies the string value until it's sent. It has multiple annoying side effects. E.g., we will need to account for the redundant suffix length when we calculate the chars remaining in the context window. Is there a better way?

What if we change all `PromptString.from*` methods to require the `ContextFilters` and to perform a check on the ignored path before we even construct the prompt string in the first place? This way we know that the final prompt string must also be of valid references. 

The issue this has is that:

- The policy could be outdated by the time toString is called 
  - Potentially fixable PromptString can be notified about policy changes?
- All callsites that use `PrompString.from` have to be made async (seems reasonable TBH)

So:

- All `PrompString.from*` methods return a Promise and require the `ContextFilterService` to be passed-in.
- Whenever we construct a prompt string from references, we check the references are valid. This pre-warms the cache
- Whenever toString is called, we check all references again (should be fast since the cache is warm?)

cc @dominiccooney

philipp-spiess:
BTW since I see @valerybugakov and me putting up more and more PRs against this branch, I'll go ahead and merge this. Let's continue the discussion on how to prevent issues of forgetting checks though.

valerybugakov:
>I'll go ahead and merge this.

YES!

PR #3858, opened 2024-04-18
Elaborate the agent protocol for Cody Ignore

dominiccooney:
- Removes `check/isCodyIgnored`, which was unused
- Adds client -> server method, `ignore/test` which gets a policy result (`'use'`, `'ignore'`)
- Adds server -> client notification, `ignore/didChange` for invalidating client cached policy results
- Adds client -> server method `testing/ignore/overridePolicy` for setting a cheesy regex-based policy, for testing

Part of #3641 

## Test plan

`pnpm -C agent test`


valerybugakov (approved):

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -313,6 +318,11 @@ export type ServerNotifications = {
 
     'codeLenses/display': [DisplayCodeLensParams]
 
+    // The set of ignored files/repositories has changed. The client should
+    // re-query using ignore/forUri.
+    // biome-ignore lint/complexity/noBannedTypes: May add details about the change later.
+    'ignore/didChange': [{}]
```
olafurpg, line 324: You can use `null` instead of `{}`

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -180,6 +176,15 @@ export type ClientRequests = {
...
     ]
 
+    // Gets whether the specified URI is sensitive and should not be sent to
+    // LLM providers.
+    'ignore/forUri': [{ uri: string }, IgnoreForUriResult]
+
+    // For testing. Overrides any ignore policy to ignore repositories and URIs
+    // which match the specified regular expressions. Pass `undefined` to remove
+    // the override.
+    'testing/ignore/overridePolicy': [[[{ repoRe: string; uriRe: string } | null]]]
```
olafurpg, line 186: This type is very involved, I've looked at it for 2 minutes now and I'm still struggling to understand what the shape of the type is. Can we simplify this somehow? I recommend sticking with named interfaces as soon as you go beyond basic `{uri: string}` anonymous literals. What is the difference between this type and `{ repoRe: string; uriRe: string } | null`?

olafurpg (changes requested):

dominiccooney:
Thanks for your help with this @olafurpg .

@valerybugakov I'm aware this needs to be integrated with the async filter, I will do that in a follow-up. Just want this to for JetBrains development for now.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -180,6 +176,15 @@ export type ClientRequests = {
         },
     ]
 
+    // Gets whether the specified URI is sensitive and should not be sent to
+    // LLM providers.
+    'ignore/forUri': [{ uri: string }, IgnoreForUriResult]
```
olafurpg, line 181: nit: I generally try to use verbs in the second part of the RPC method name, and use nouns only if it's very awkward to use a noun. I think `codyignore` is also a bit easier to read since it's a noun.

```suggestion
    'codyignore/test': [{ uri: string }, IgnoreForUriResult]
```

Also, it should be fine to use an anonymous literal `{ policy: 'ignore' | 'use' }` here instead of `IgnoreForUriResult`. If it didn't work then I can look into fixing the bindings generator.

lib/shared/src/cody-ignore/context-filter.ts
```diff
@@ -4,6 +4,36 @@ import { IgnoreHelper } from './ignore-helper'
 
 export const ignores = new IgnoreHelper()
 
+// TODO: It would be better to use a vscode.EventEmitter and support multiple
+// listeners, etc. but lib/shared prevents doing that easily.
```
olafurpg, line 8: 😢 This makes me want to move all of `lib/shared` into the `vscode/` directory. Alternatively, we can consider adding `@types/vscode` to `lib/shared` since it works anyways with the agent.

olafurpg (approved):
LGTM 👍🏻 Only minor comments, nothing blocking

lib/shared/src/cody-ignore/context-filter.ts
valerybugakov: yes!

Issue #3820, opened 2024-04-16
VS Code: Default LLM is not showing at the top but showing in between other LLMs.


jay-fibi:
### Version

v1.12.0

### Describe the bug

Platform: Windows

**Steps to reproduce:**

1. Have Cody plugin installed in your VS Code editor
2. Login to Cody with Free user
3. Click on "New Chat" and check the LLM dropdown

### Observed behavior:
Default LLM(Claude 3 Sonnet) is showing in the middle of the list.

### Expected behavior

Default LLM(Claude 3 Sonnet) should be showing at the top as per the JB.

### Additional context

**Screenshot:**

![Default_LLM_location](https://github.com/sourcegraph/cody/assets/158482423/af1d5693-0a01-48f9-b5bd-a94ad87b0b0a)

Issue #3792, opened 2024-04-12
Improve Pro User Model Selection


chillatom:
Problems

- Model selection drop down list is long
- No guidance to user on when to choose a particular model
- Mix of legacy, new and experimental models

Desired state

- Clear guidance on model choice (Good, Better, Best, Fast, Faster, Fastest etc.) 
- Legacy models deprioritized in rank and clearly annotated (e.g. Claude 2.0, Claude 2.1)
- Experimental models deprioritized in rank and clearly annotated (e.g. Mixtral models)

philipp-spiess:
I’m sure that @toolmantim has better ideas but ever since we chatted about surfacing the differences of models, I've been thinking of this as a 3 point scale across two dimensions and have envisioned a UI like this:

<img width="524" alt="Screenshot 2024-04-15 at 17 59 43" src="https://github.com/sourcegraph/cody/assets/458591/1b3b1251-a783-4dda-95a9-0aa4693c8d55">

toolmantim:
Nice @philipp-spiess — love this sketch! I ended up going with a simpler grouping… but you know what, we could actually make the selector an actual 2-axis chart that you use to select the model. As long as you make it clear which ones are the recommended ones.

Not sure if we've got enough data/info to be that granular to compare model quality/speed though, do we? And it adds a bit of a bar for adding a model… vs just having an "Experimental" box to throw stuff into w/o clear data.

OR going back to something simpler… we just add some little indicators/icons to the right of each row. Only if we feel it's missing it though.

philipp-spiess:
@toolmantim Ha love the idea but the route you're going in the Figma is probably enough for now and already a good improvement (and much easier to visually parse).

abeatrix:
I love the grouping idea!

Personally think it'd be helpful to show what tasks each model is good at to help users to decide which model to use! Something like: Claude Opus: Great for answering large codebase questions, or Claude Haiku for code formatting etc

Ref: GPT marks ChatGPT3.5 as `Great for everyday tasks` and ChatGPT4 as the smallest and most capable 😆 
![image](https://github.com/sourcegraph/cody/assets/68532117/f00acb85-b14f-4e48-a098-62833c045928)


A + button to add new modeld would be cool too like they do in Opera:
![image](https://github.com/sourcegraph/cody/assets/68532117/bca518cd-0036-4cb0-8fed-ab1109bd234d)

philipp-spiess:
[...]

PR #3859, opened 2024-04-18
models: show default LLM first in dropdown

keegancsmith:
The order we list our models in dotcom.ts is the order it is shown in. I imagine in future we likely want to order these such that LLMs we want to guide users to use are higher up. For now I just moved the default LLM to the top.

Alternatively we should consider making the selected LLM at the top of the dropdown list?

Test Plan: CI

Fixes https://github.com/sourcegraph/cody/issues/3820

### DIFF
--- a/lib/shared/src/models/dotcom.ts
+++ b/lib/shared/src/models/dotcom.ts
@@ -10,6 +10,16 @@ import { ModelUsage } from './types'
 
 // The models must first be added to the custom chat models list in https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/completions/httpapi/chat.go?L48-51
 const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
+    // The order listed here is the order shown to users. Put the default LLM first.
+    {
+        title: 'Claude 3 Sonnet',
+        model: 'anthropic/claude-3-sonnet-20240229',
+        provider: 'Anthropic',
+        default: true,
+        codyProOnly: false,
+        usage: [ModelUsage.Chat, ModelUsage.Edit],
+        contextWindow: { input: CHAT_INPUT_TOKEN_BUDGET, output: CHAT_OUTPUT_TOKEN_BUDGET },
+    },
     {
         title: 'Claude 2.0',
         model: 'anthropic/claude-2.0',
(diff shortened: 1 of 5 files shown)


keegancsmith:
Hey I'd love to pair to see how I can manually test this :)

philipp-spiess:
@keegancsmith happy to chat :)

philipp-spiess (approved):
LGTM but we need to recreate agent snapshots here:

```
source agent/scripts/export-cody-http-recording-tokens.sh
pnpm update-agent-recordings
```

philipp-spiess:
> Alternatively we should consider making the selected LLM at the top of the dropdown list?

I think for now this is fine, we plan some larger scale updates to the input soon: https://github.com/sourcegraph/cody/issues/3792

keegancsmith:
@philipp-spiess this made bigger changes than I expected, but I suppose we had tests that just used the first model hence the need to update?

philipp-spiess:
@keegancsmith Yeah it seems so. This seems reasonable though, one less Claude 2 call to worry about 🧠

philipp-spiess:
Looks like the E2E tests also depend on this 😬 If you want to fix them, I recommend running them with `--debug` locally so that you can see which selector breaks. Happy to take over from here as well if you want.

```bash
cd vscode
pnpm test:e2e --debug -- attribution
```

keegancsmith:
> Happy to take over from here as well if you want.

You can start worrying, I plan on contributing much more to this repo for the next while :) So I'll fix.

lib/shared/src/models/dotcom.ts
```diff
@@ -10,6 +10,16 @@ import { ModelUsage } from './types'
...
+    // The order listed here is the order shown to users. Put the default LLM first.
+    {
+        title: 'Claude 3 Sonnet',
+        model: 'anthropic/claude-3-sonnet-20240229',
+        provider: 'Anthropic',
+        default: true,
+        codyProOnly: false,
+        usage: [ModelUsage.Chat, ModelUsage.Edit],
+        contextWindow: { input: CHAT_INPUT_TOKEN_BUDGET, output: CHAT_OUTPUT_TOKEN_BUDGET },
+    },
```
abeatrix, line 22: If we want to order the list based on preferences, should we move all Claude 3 models to the top too?

abeatrix (approved):
Would it be better to sort the [model list inside the UI component ](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/Components/ChatModelDropdownMenu.tsx?L95)instead since the default model could change based on user's last picked model? 🤔

philipp-spiess:
> Would it be better to sort the model list inside the UI component instead since the default model could change based on user's last picked model? 

Then it wouldn't work in JetBrains though, right? I don't think it's expected in a dropdown that your currently selected option is always up front. I think the first slot should be _our recommendation_. WDYT?

keegancsmith:
@philipp-spiess please take a look, the issue seemed to be in the naive way the mock-server decided which message was from the human. I am guessing how we format the message to the mock-server changed when I put opus as the first model.

abeatrix (approved):
> Then it wouldn't work in JetBrains though, right? I
@philipp-spiess I think they are already sorting it in their UI based on the latest version of the JB plugin 🤔 
image

@keegancsmith Thanks for updating the list! I didn't know changing the model list would require that many changes 😱 The list aligns with what JB has so LGTM!

PR #3860, opened 2024-04-18
Fetch Cody Ignore test dataset

taras-yemets:
## Test plan
- TODO


dominiccooney (changes requested):
I do not support this change for reasons outlined on Slack: https://sourcegraph.slack.com/archives/C06Q9R098N4/p1713477345053669?thread_ts=1712802266.879379&cid=C06Q9R098N4

Issue #3838, opened 2024-04-17
VS Code: "N months ago" is showing for Chat history title for chats older than few months.


jay-fibi:
### Version

v1.12.0

### Describe the bug

Platform: MacOS

**Steps to reproduce:**

1. Have Cody plugin installed in your VS Code editor
2. Login to Cody with Free/Pro/Enterprise user having chat older than 2+ months.
3. Check the Chats headers

### Observed behavior:
Chats older than 2 months is having title "N months ago".

### Expected behavior

It should display correct number of months i.e. two months ago.

### Additional context

**Screenshots:**

<img width="403" alt="Screenshot 2024-04-17 at 8 06 28 PM" src="https://github.com/sourcegraph/cody/assets/158482423/df1a4edf-3550-4e54-bffd-dc27f4d88f01">

PR #3864, opened 2024-04-18
Chat History: Fix 'N months ago'

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/3838

This PR refactors the chat history grouping logic where we hard coded "N months ago" for char older than a month.

Replaced current logic with the one provided by @toolmantim :

- This week (7 days ago to now)
- This month (if now > 7 days into the month, start of month to 7 days ago)
- Last month (start of month (this month - 1) to end of month (this month -1)
- More than a month (end of month (this month -1) to infinity)

This provides a more flexible way to group recent chats that will work consistently regardless of the current date.

## Changes

- Switched from using date comparisons to group chats to using the new `getRelativeChatPeriod` utility to group chats into time buckets "Today", "This Week", "This Month", "n months ago" etc.

- Store the grouped chats in a Map rather than separate arrays.

- Reverse the chat history entries before grouping so the most recent chats appear first within each time period group.

- Updated the `groupCodyChats` function to return a Map rather than an object literal.

## Test plan

- Verified chat history is correctly grouped into recent time period buckets.
- Confirmed most recent chats appear first within each time period group.
- Confirmed chats older than last month (if you have any) are sorted into the correct n month group:

image

### Before

From issue:
image

### DIFF
--- a/vscode/src/services/HistoryChat.ts
+++ b/vscode/src/services/HistoryChat.ts
@@ -34,7 +34,7 @@ export function groupCodyChats(authStatus: AuthStatus | undefined): GroupedChats
     const yesterdayChats: CodySidebarTreeItem[] = []
     const thisMonthChats: CodySidebarTreeItem[] = []
     const lastMonthChats: CodySidebarTreeItem[] = []
-    const nMonthsChats: CodySidebarTreeItem[] = []
+    const beforeLastMonthChats: CodySidebarTreeItem[] = []
 
     const today = new Date()
     const yesterday = new Date()
@@ -47,7 +47,7 @@ export function groupCodyChats(authStatus: AuthStatus | undefined): GroupedChats
         Yesterday: yesterdayChats,
         'This month': thisMonthChats,
         'Last month': lastMonthChats,
-        'N months ago': nMonthsChats,
+        'Before last month': beforeLastMonthChats,
     }
 
     if (!authStatus) {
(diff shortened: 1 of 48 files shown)


vscode/src/services/HistoryChat.ts
```diff
@@ -105,7 +105,7 @@ export function groupCodyChats(authStatus: AuthStatus | undefined): GroupedChats
         Yesterday: yesterdayChats.reverse(),
         'This month': thisMonthChats.reverse(),
         'Last month': lastMonthChats.reverse(),
-        'N months ago': nMonthsChats.reverse(),
+        'Before last month': beforeLastMonthChats.reverse(),
```
abeatrix, line 108: Does `Before last month` sound right to you? @toolmantim

vscode/src/services/HistoryChat.ts
```diff
@@ -75,7 +75,7 @@ export function groupCodyChats(authStatus: AuthStatus | undefined): GroupedChats
             const chatTitle = chats[id].chatTitle || getChatPanelTitle(lastHumanText, false)
 
             const lastInteractionTimestamp = new Date(entry.lastInteractionTimestamp)
-            let groupLabel = 'N months ago'
+            let groupLabel = 'Before last month'
```
philipp-spiess, line 78: Nit since we use this string in three places, ca we make it a const in the module scope?

philipp-spiess (approved):
Code LGTM but I'll let a native speaker weigh in on the naming 🙈

vscode/src/services/HistoryChat.ts
abeatrix: Yea i've updated the PR to implemented Tim's design with your suggestion too :D

PR #3866, opened 2024-04-18
Implement @-mention package context

thenamankumar:
Demo: https://www.loom.com/share/342e1903f93e4ff28f582c5d310ef5ff?sid=f8b66a30-b230-4403-8869-cb3e748cdbb5

We have 370K+ packages from various ecosystems (npm, go, rust, ruby, python, java) already indexed on [sourcegraph.com](http://sourcegraph.com/). It is already enterprise-ready as well.

This new feature will allow users to include specific packages into the context using the syntax `@<ecosystem>:<package_name>`, for example: `@npm:tailwindcss`.

Cody will make a Remote Search for the tailwind package repo against the user's query and include the related code snippets in the context.

## Test plan
TODO

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -28,6 +28,7 @@ import {
     GET_FEATURE_FLAGS_QUERY,
     LOG_EVENT_MUTATION,
     LOG_EVENT_MUTATION_DEPRECATED,
+    PACKAGE_LIST_QUERY,
     RECORD_TELEMETRY_EVENTS_MUTATION,
     REPOSITORY_IDS_QUERY,
     REPOSITORY_ID_QUERY,
@@ -111,11 +112,29 @@ interface CurrentUserCodySubscriptionResponse {
 }
 
 interface CodyLLMSiteConfigurationResponse {
-    site: { codyLLMConfiguration: Omit<CodyLLMSiteConfiguration, 'provider'> | null } | null
+    site: {
+        codyLLMConfiguration: Omit<CodyLLMSiteConfiguration, 'provider'> | null
+    } | null
 }
 
 interface CodyLLMSiteConfigurationProviderResponse {
-    site: { codyLLMConfiguration: Pick<CodyLLMSiteConfiguration, 'provider'> | null } | null
+    site: {
+        codyLLMConfiguration: Pick<CodyLLMSiteConfiguration, 'provider'> | null
+    } | null
+}
+
+export interface PackageListResponse {
+    packageRepoReferences: {
+        nodes: {
+            id: string
+            name: string
+            kind: string
+            repository: { id: string; name: string; url: string }
+        }[]
+        pageInfo: {
+            endCursor: string | null
+        }
+    }
 }
 
 export interface RepoListResponse {
@@ -498,6 +517,20 @@ export class SourcegraphGraphQLAPIClient {
         return { ...config, provider }
     }
 
+    public async getPackageList(
+        kind: string,
+        name: string,
+        first: number,
+        after?: string
+    ): Promise<PackageListResponse | Error> {
+        return this.fetchSourcegraphAPI<APIResponse<PackageListResponse>>(PACKAGE_LIST_QUERY, {
+            kind,
+            name,
+            first,
+            after: after || null,
+        }).then(response => extractDataOrError(response, data => data))
+    }
+
     /**
      * Gets a subset of the list of repositories from the Sourcegraph instance.
      * @param first the number of repositories to retrieve.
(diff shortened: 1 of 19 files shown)


sqs:
Awesome, will review. I think we should first add in a hook/API as described in https://sourcegraph.slack.com/archives/C05AGQYD528/p1713500773817519 to ensure we can experiment rapidly with things like this while having near-zero chance of destabilizing Cody.

sqs (approved):
This is so awesome! :heart_eyes: :heart_eyes: :heart_eyes:

I'm going to work on a PR now to make it so there is a single hook/API (`const CONTEXT_SOURCES = [...]`) where we can add in more context sources, so that we can experiment quickly there with PRs that have (ideally) a single interaction point with the rest of the codebase, which will reduce risk of instability and make reviews easier.

The code LGTM but so I'm approving it, but wait to merge it until it can use that new hook/API.

Here are some issues I saw when using it:

- The detail text in the completion menu has URL escape codes: 
image
- No context files were included for `@types/lodash.debounce` or `@npm:is-odd`
image
- When I have `@npm:csv-` it shows completion menu results, but when I type `p` it shows `No files found`. I have seen a few other cases like this.
image
- For `@npm:left-pad` it only used one (test) file from the package and then a bunch of files from my own project: 
image

vscode/src/editor/utils/editor-context.ts
```diff
@@ -276,34 +279,51 @@ export async function filterContextItemFiles(
...
-                            range = range ?? getRangeByContentLineCount(content)
+            items.map(
+                async (
+                    item: ContextItem
+                ): Promise<ContextItemWithContent | ContextItemWithContent[] | null> => {
+                    let { content, size, range, uri } = item
+                    if (!content) {
+                        try {
+                            if (item.type === 'package') {
+                                return await findContextItemsWithContentForPackage(
```
sqs, line 295: Cool how this expands one context item found at query-time to multiple context files, using only the relevant subset of files in the package. I think that's a clever technique and we will need something like that, but I think this might be why I saw no/poor context files included for `@types/lodash.debounce` and `is-odd` and `left-pad`.

Is there a way we could remove this one step of magic for now and make it more explicit, and then follow up with the more-magic approach shortly thereafter?

I wonder how far we could get with language-specific heuristics (eg omit `dist/**/*.js` and `*.min.js`).

BTW the AI team is working on a context mixer that will let Cody make a global (not local-per-context-provider) decision about how best to fill up the context. This will be an interesting case for them to handle.

sqs:
https://github.com/sourcegraph/cody/pull/3883 for the hook/API

vscode/src/editor/utils/editor-context.ts
sqs: (Not blocking, and #3883 supports this kind of one-contextitem-resolving-to-many technique.)

thenamankumar:
@sqs @keegancsmith I have rebased this on top of the new Mention Context API and resolved a bunch of bugs and issues as well. 

Latest demo: https://www.loom.com/share/142cba1e5f394e06bf1d439e0c63f5ea

This PR is good to merge now once you review it. 

As the next set of improvements, I'll work on:
- Making sure all the packages on dotcom have search indexing (if that is required for getCodyContext API to work).
- Exclude dist/, min.js and more from the context
- Figure out a good & easy way for context re-ranking.

sqs (approved):
Awesome! Works great for me in a few that I tried. The test file in this screenshot is actually great context for usage examples (initially I worried since it was just 1 test file and not the definition).  And TBH probably a lot of the knowledge would come from the model's training for this package anyway.

The Windows unit test failure will be fixed when you rebase against main.


image

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -557,7 +590,7 @@ export class SourcegraphGraphQLAPIClient {
                     repoName: item.blob.repository.name,
                     path: item.blob.path,
                     uri: URI.parse(
-                        `${item.blob.url.startsWith('/') ? this.endpoint : ''}${item.blob.url}?L${
```
sqs, line 560: Why can't we use `blob.url`? This hardcodes our URL structure, which is probably fine but not a super good idea.

vscode/src/chat/context/chatContext.ts
```diff
@@ -80,8 +80,15 @@ export function getEnabledContextMentionProviders(): ContextMentionProvider[] {
 
     const isURLProviderEnabled =
         vscode.workspace.getConfiguration('cody').get<boolean>('experimental.urlContext') === true
-    if (isURLProviderEnabled) {
```
sqs, line 83: I think this should just be under `cody.experimental.noodle` for simplicity, so we don't have tons of our own feature flags. It'd just get confusing.

sqs: Actually, this is a cool enough feature that it's probably worthwhile to let other people try it out by itself. So ignore my comment above. :-)

sqs (approved):

lib/shared/src/sourcegraph-api/graphql/client.ts
thenamankumar: `blob/URL` is not working as it gives 404. It includes the @revision in the URL; in that case, the webpage returns the "repository not found".

Consider this a hacky fix for now. I'll check with the graph team to determine why specific revisions are unavailable.

PR #3869, opened 2024-04-19
VS Code: context filters graphql API update

valerybugakov:
## Context

- Part of https://github.com/sourcegraph/cody/pull/3771
- Rebased on top of https://github.com/sourcegraph/cody/pull/3856
- Updates the Cody Ignore context filter GraphQL query to be compatible with the latest backend changes https://github.com/sourcegraph/sourcegraph/pull/61101.
- Initializes `contextFiltersProvider` on the extension activation
- Re-initializes `contextFiltersProvider` on configuration change
- Uses the "includes everything" policy for Sourcegraph instances without specified context filters or for instances where this API is not yet supported.

## Test plan

CI and unit tests.

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.ts
@@ -2,7 +2,7 @@ import { LRUCache } from 'lru-cache'
 import { RE2 } from 're2-wasm'
 import type * as vscode from 'vscode'
 import { isFileURI } from '../common/uri'
-import { logError } from '../logger'
+import { logDebug, logError } from '../logger'
 import { graphqlClient } from '../sourcegraph-api/graphql'
 import type { CodyContextFilterItem } from '../sourcegraph-api/graphql/client'
 import { wrapInActiveSpan } from '../tracing'
@@ -26,16 +26,10 @@ export class ContextFiltersProvider implements vscode.Disposable {
     private fetchIntervalId: NodeJS.Timer | undefined
     private cache = new LRUCache<string, boolean>({ max: 128 })
     private getRepoNameFromWorkspaceUri: GetRepoNameFromWorkspaceUri | undefined = undefined
-    // TODO: This is only a temporary off-switch until we can properly detect a
-    // valid context filter config in the upstream repository.
-    private temporary_isEnabled = true
 
-    async init(temporary_isEnabled: boolean, getRepoNameFromWorkspaceUri: GetRepoNameFromWorkspaceUri) {
-        this.temporary_isEnabled = temporary_isEnabled
-        if (!temporary_isEnabled) {
-            return
-        }
+    async init(getRepoNameFromWorkspaceUri: GetRepoNameFromWorkspaceUri) {
         this.getRepoNameFromWorkspaceUri = getRepoNameFromWorkspaceUri
+        this.dispose()
         await this.fetchContextFilters()
         this.startRefetchTimer()
     }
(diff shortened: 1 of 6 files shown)


lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -26,34 +26,25 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
-                this.cache.clear()
-                this.contextFilters = null
-
-                if (response) {
-                    this.contextFilters = {
-                        include: response.include.map(parseContextFilterItem),
-                        exclude: response.exclude.map(parseContextFilterItem),
-                    }
+            this.cache.clear()
+            this.contextFilters = null
```
dominiccooney, line 41: Worth commenting that `contextFilters = null` means deny all.

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -132,6 +123,18 @@ export class ContextFiltersProvider implements vscode.Disposable {
             clearTimeout(this.fetchIntervalId)
         }
     }
+
+    private hasIncludeEverythingFilters() {
+        return this.contextFilters?.include.length === 0 && this.contextFilters?.exclude.length === 0
+    }
+
+    private hasExcludeEverythingFilters() {
```
dominiccooney, line 131: Let's delete this. Less code is more robust. You're basically doing an optimization for the case where everything is excluded with a specific syntax. I doubt that user cares very much how quickly Cody is (almost totally) blocked for them.

Let's just interpret the rules.

vscode/src/main.ts
```diff
@@ -245,6 +247,7 @@ const register = async (
         promises.push(featureFlagProvider.syncAuthStatus())
         graphqlClient.onConfigurationChange(newConfig)
         promises.push(contextProvider.onConfigurationChange(newConfig))
+        promises.push(contextFiltersProvider.init(bindedRepoNameResolver))
```
dominiccooney, line 250: Let's make this complete before the context provider? (ie chain those two together with `then`.) Because the context provider is going to turn around and look up the workspace repo to decide to add it as implicit context, etc.

dominiccooney (approved):
Looks good, some feedback inline.

lib/shared/src/cody-ignore/context-filters-provider.ts
dominiccooney: Oh yeah yeah yeah, excellent point.

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -132,6 +123,18 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
+        return (
+            this.contextFilters?.include.length === 0 &&
+            this.contextFilters?.exclude.length === 1 &&
```
taras-yemets, line 140: FYI we validate include/exclude rules length to be [at least 1](https://github.com/sourcegraph/sourcegraph/blob/9bf06b664c1db941a3bbfab87dedc009fc3a19e5/schema/site.schema.json#L2583).

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -132,6 +123,18 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
+            this.contextFilters.exclude[0].repoNamePattern.toString() === '.*'
```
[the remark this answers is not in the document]
taras-yemets: Admins can also define a few include and exclude rules each matching any repo name.
I don't think we can foresee these multiple scenarios and get real performance benefits from checking early whether there's a catch-all exclude rule.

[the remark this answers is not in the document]
valerybugakov: These hardcoded checks are based on the virtual values defined on the client ([sources](https://github.com/sourcegraph/cody/blob/76807b833c435154c93613a1ccbb044556b926c0/lib/shared/src/sourcegraph-api/graphql/client.ts#L197-L205)).  We do not optimize checks for filters that site admins will set.

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: The virtual values we use on the client can be zero in length. We use them to signal that all URIs can be included.

lib/shared/src/cody-ignore/context-filters-provider.ts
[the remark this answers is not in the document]
valerybugakov: @taras-yemets, here's the PR to make it more explicit https://github.com/sourcegraph/cody/pull/3892

PR #3870, opened 2024-04-19
VS Code: remote repo name resolution

valerybugakov:
## Context

- WIP

## Test plan

CI and unit tests.


vscode/src/repository/repo-name-resolver.ts
```diff
@@ -59,15 +60,56 @@ export class RepoNameResolver {
...
+        }
+
+        const repoNameRequest = graphqlClient.getRepoName(remoteUrl)
+        this.remoteUrlToRepoNameCache.set(remoteUrl, repoNameRequest)
+        let repoName = await repoNameRequest
+
+        if (repoName === null) {
+            // Use locally resolved repo name in case the Sourcegraph API returns null.
+            repoName = convertGitCloneURLToCodebaseName(remoteUrl)
+        }
```
valerybugakov: @aramaraju @dominiccooney, let's add this to the spec. At first, I kept the logic where we calculate the repo name locally based on the remote URL in case Sourcegraph API is unable to find a match, but to be on the safer side of things, we should return `null` here, which would mean that the URI in question will be ignored because the repo is unknown.

vscode/src/repository/git-extension-api.ts
```diff
@@ -77,9 +77,9 @@ export async function gitAPIinit(): Promise<vscode.Disposable> {
  */
 export function getCodebaseFromWorkspaceUri(uri: vscode.Uri): string | undefined {
     try {
-        const remoteOriginUrl = gitRemoteUrlFromGitExtension(uri)
-        if (remoteOriginUrl) {
-            return convertGitCloneURLToCodebaseName(remoteOriginUrl) || undefined
+        const remoteUrl = gitRemoteUrlFromGitExtension(uri)
```
taras-yemets, line 80: Repository [`state.remotes`](https://sourcegraph.com/github.com/microsoft/vscode@e0d3400d8c43ce67588794ddda4a0e6b08115021/-/blob/extensions/git/src/api/api1.ts?L40:6&popover=pinned#tab=references) is an array. Why do we narrow it down to only the first element in this array?

taras-yemets (commented):
LEft a comment inline. Otherwise LGTM 👍🏻

vscode/src/repository/git-extension-api.ts
valerybugakov: Great call, Taras! We discussed that and I'm working on changing it here: https://github.com/sourcegraph/cody/pull/3895

philipp-spiess (approved):
Beautiful

vscode/src/repository/repo-name-resolver.ts
taras-yemets: @aramaraju, no need to change anything server-side.

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -32,42 +33,76 @@ export class RepoNameResolver {
...
                         break
                     }
                 }
             }
 
-            if (remoteOriginUrl) {
-                const repoName = convertGitCloneURLToCodebaseName(remoteOriginUrl) || undefined
+            if (remoteUrl) {
+                const repoName = await this.resolveRepoNameForRemoteUrl(remoteUrl)
                 this.fsPathToRepoNameCache.set(uri.fsPath, repoName)
```
philipp-spiess, line 64: This actually regresses the repo name resolver for PLG use cases where repos are _not_ remotely indexed. I think we should use `convertGitCloneURLToCodebaseName` as a fallback if `resolveRepoNameForRemoteUrl` comes back with no replies. WDYT? cc @taras-yemets @valerybugakov

philipp-spiess: I was trying to use this API to store which repos have already seen a cody ignore related banner/notification and saw that it resolves to an empty array in my local repo.

taras-yemets: I think we can use `convertGitCloneURLToCodebaseName` as a fallback for PLG users. 
For enterprise users, we'd better be on the safer side. If we can't find the local repo on the instance and Cody context filters are configured, we can't be sure that local repo is allowed for Cody, and thus we disable Cody for it. WDYT?

valerybugakov: We should not use this for PLG since this repo-name resolver is only used for the enterprise feature users for now. See https://github.com/sourcegraph/cody/pull/3870#discussion_r1574096226

I'm reverting it https://github.com/sourcegraph/cody/pull/3922

philipp-spiess: ACK! It wasn't clear for me that this is an enterprise only implementation. Let's continue the thread in https://github.com/sourcegraph/cody/pull/3922#discussion_r1577300787

PR #3871, opened 2024-04-19
Remove startcoder2-hybrid from settings

rafax:
Some of the Fireworks models required for this are no longer deployed:
Screenshot 2024-04-19 at 09 37 14
Screenshot 2024-04-19 at 09 37 04

## Test plan

- tested locally 
Screenshot 2024-04-19 at 09 38 37

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1039,7 +1039,7 @@
         "cody.autocomplete.advanced.model": {
           "type": "string",
           "default": null,
-          "enum": [null, "starcoder-16b", "starcoder-7b", "starcoder2-hybrid", "llama-code-13b"],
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):
@valerybugakov what do you think about making this setting and the `advanced.provider` non-public in the first place? We really don't want people playing around with it, i think it can create confusing scenarios (e.g. https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1713461682091789).

We can add an extra setting for Llama since that is something we do kind-of support.

valerybugakov:
Yep, we can make them experimental. They won't be available in the stable release version but will still be available in insiders builds.

PR #3872, opened 2024-04-19
Edit: Fix respins not restarting a edit

umpox:
## Description

https://github.com/sourcegraph/cody/assets/9516420/3707835b-90e5-47be-9053-f3650cd038df

Bug: Edits were not correctly re-spinning, as the LLM request wasn't being restarted

Solution:
- When a respin is required:
  - Access the `EditProvider` for the task
  - Abort any existing request
  - Re-trigger `startEdit` using the same task parameters

## Test plan

1. Make an edit
2. Immediately make changes within the range of that edit, that would likely conflict with the LLM's response
3. Expect that the Edit respins and resolves the issue

### DIFF
--- a/vscode/src/edit/manager.ts
+++ b/vscode/src/edit/manager.ts
@@ -46,12 +46,31 @@ export class EditManager implements vscode.Disposable {
 
     constructor(public options: EditManagerOptions) {
         this.controller = new FixupController(options.authProvider, options.extensionClient)
-        this.disposables.push(
-            this.controller,
-            vscode.commands.registerCommand('cody.command.edit-code', (args: ExecuteEditArguments) =>
-                this.executeEdit(args)
-            )
+        /**
+         * Entry point to triggering a new Edit.
+         * Given a set or arguments, this will create a new LLM interaction
+         * and start a `FixupTask`.
+         */
+        const editCommand = vscode.commands.registerCommand(
+            'cody.command.edit-code',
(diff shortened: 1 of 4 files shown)


vscode/src/edit/manager.ts
```diff
@@ -46,12 +46,32 @@ export class EditManager implements vscode.Disposable {
...
+         * Note: This differs to a "retry", as it preserves the original `FixupTask`.
+         */
+        const startCommand = vscode.commands.registerCommand(
+            'cody.command.start-edit',
+            (task: FixupTask) => {
+                const provider = this.getProviderForTask(task)
+                provider.abortEdit()
+                provider.startEdit()
+            }
         )
```
umpox, line 73: This needs to be done at the `EditProvider` level as the FixupController is decoupled from the LLM interaction (one FixupController, many EditProviders)

dominiccooney (approved):
This is good, do you think we could add an integration regression test (the integration tests aren't used much, but they have access to the extension objects, would be good for this?) or an agent test for this?

Mea culpa for not having one in the first instance.

umpox:
Thanks @dominiccooney. I did look but it's quite an involved piece of work right now, as we need a way to reliably delay a test response in order to ensure this happens. I don't think we have that right now (we also don't have a way to adjust our mock responses on a per test basis for our integration tests yet), going to merge this as I have a bunch of other stuff to focus on

PR #3875, opened 2024-04-19
Agent: add `server` subcommand to expose agent protocol via websockets

olafurpg:
Previously, it was only possible to do RPC with the Cody Agent via stdin/stdout, which required the client to run on the same computer as the Cody Agent. This PR adds a new `server` subcommand that starts a server that allows clients to connect with the Cody Agent via web sockets from another computer.

One example use-case for this feature is to implement a browser client to interact with Cody Chat. By using websockets, the browser client can interact with Cody without installing any custom dependencies. To improve the ergonomics of interacting with Cody, we can extend our current Kotlin bindings generation tool to generate bindings for other languages like TypeScript and Go.

Below is a demo of a SvelteKit static page that establishes a websocket connection to expose a basic chat interface

CleanShot 2024-04-18 at 22 15 13

## Test plan

Manually tested with a SvelteKit client.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -29,4 +29,7 @@
     "Supercompletion",
     "Supercompletions"
   ],
+  "[typescript]": {
+    "editor.defaultFormatter": "biomejs.biome"
+  },
 }
--- a/agent/package.json
+++ b/agent/package.json
@@ -57,7 +57,8 @@
     "pretty-ms": "^8.0.0",
     "uuid": "^9.0.0",
     "vscode-uri": "^3.0.7",
-    "win-ca": "^3.5.1"
+    "win-ca": "^3.5.1",
+    "ws": "^8.16.0"
   },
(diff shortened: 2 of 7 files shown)


philipp-spiess (approved):
Wow this is a very small change. Happy to add this even if we don't end up using it it's not getting in the way and can be easily reverted again.

PriNova:
That is awesome @olafurpg 
Better than the workaround I took to set the `CODY_AGENT_DEBUG_REMOTE` flag for TCP communication (which works also pretty well):

May you put that as a release?

olafurpg:
@PriNova that's cool! When I have some bandwidth, I could take a stab at auto-generating Python bindings that will work similarly to https://github.com/PriNova/codypy/blob/main/codypy/cody_py.py I see you're only using a few endpoints so it might not be too bad, but let me know if you are interested. No promises when I'll find the time but would be a fun project :)

PR #3876, opened 2024-04-19
Upgrade vitest, vite, and some types

philipp-spiess:
The vitest VS Code extension was no longer working because of an outdated vitest version. Upgrading it seemed to work without any issues.

Unfortunately I can't say this about upgrading biome so I've reverted to only vite, vitest, and some type helpers...

## Test plan

- CI passes
- Vitest vscode works again:
Screenshot 2024-04-19 at 16 21 19


abeatrix (approved):
Nice i keep seeing the pop-up warning me about the version but never got to it. Thanks for updating it!

PR #3881, opened 2024-04-20
Chat: refactor Ollama chat client

abeatrix:
Refactor Ollama chat client to use the official package for js.

We should also update the completion client to use the package.

## Test plan

1. Run this branch in debug mode and follow our Ollama docs to set up Ollama with Cody
2. Try asking Cody a question
3. Verify everything works as expected.

image


lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -46,46 +40,53 @@ export function ollamaChatClient(
...
-                insertText += parsedLine.message.content
-                cb.onChange(insertText)
+            onAbort(signal, () => reader.cancel())
+            let responseText = ''
+            // Handles the response stream to accumulate the full completion text.
+            while (true) {
+                const { done, value } = await reader.read()
+                try {
+                    if (done) {
+                        cb.onChange(responseText)
```
dominiccooney, line 64: This isn't right. You could read some data *and* reach the end of the stream, and you're discarding `value` here. This might only be working because the last message where ollama is done contains a hunk of data we don't use.

lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -46,46 +40,53 @@ export function ollamaChatClient(
...
+                        cb.onComplete()
+                        break
+                    }
+                    const textDecoder = new TextDecoder()
+                    const data = textDecoder.decode(value, { stream: true })
```
dominiccooney, line 69: I know you're just moving this around, but it does not look right. Either we are decoding `value` when we should be appending it to a buffer first and decoding the whole buffer, OR we don't need `stream: true` here because `value` is a complete value.

I expect it is the first case, and it is only working because the ollama server tends to flush at certain points.

You need to do something like:
- Use a `TextDecoderStream`
- Accumulate the decoded text into a string
- Break on newlines
- JSON.parse each chunk and act on it

lib/shared/src/llm-providers/ollama/utils.ts
```diff
@@ -28,3 +28,17 @@ export async function fetchLocalOllamaModels(): Promise<ModelProvider[]> {
...
 }
+
+/**
+ * Converts an array of `Message` objects to an array of `OllamaChatMessage` objects,
+ * which are suitable for sending to the Ollama API.
+ *
+ * @param messages - An array of `Message` objects to be converted.
+ * @returns An array of `OllamaChatMessage` objects.
+ */
+export function getOllamaChatMessages(messages: Message[]): OllamaChatMessage[] {
```
dominiccooney, line 39: We should not be converting these messages from PromptString to ordinary string here, that should happen right before we hit the network.

Yes, ollama is an interesting case because it is *probably* local. But there's 0 checks for that, let's just treat it like any other remote model for now.

lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -46,46 +40,53 @@ export function ollamaChatClient(
     } satisfies OllamaChatParams
 
     // Sends the completion parameters and callbacks to the Ollama API.
-    fetch(new URL('/api/chat', OLLAMA_DEFAULT_URL).href, {
+    fetch(config?.endpoint ?? new URL('/api/chat', OLLAMA_DEFAULT_URL).href, {
         method: 'POST',
         body: JSON.stringify(ollamaChatParams),
```
dominiccooney, line 45: Put a helper function here which converts the Ollama messages from PromptStrings to JSON. CC @philipp-spiess .

dominiccooney (changes requested):
Could you tell me more about the specific reason PromptString was regressing ollama prompts?

lib/shared/src/llm-providers/ollama/chat-client.ts
abeatrix: Confirmed that the issue is fixed on main after @philipp-spiess 's PR 
image

philipp-spiess: @abeatrix Do you have a log of what changed though? In both cases the serialized JSON should be the same? 🤔

abeatrix: I cannot reproduce this issue on main after whatever you changed in your PR so it must be the helper function you added for here? 

Or are you able to reproduce this issue on main?

philipp-spiess: I haven't tried to reproduce it before so it's unclear what the issue was in the first place. Your message makes it seem like there's some issue with the prompt being malformed but that's unexpected since it should be identical in the serlialized JSON string. Unless another change was made. 🤔

abeatrix: none of my changes are available on main, and I couldn't reproduce the issue on main anymore when I can reproduce it consistently (as shown in my screenshot above) before.

> Your message makes it seem like there's some issue with the prompt being malformed but that's unexpected 

When i can reproduce it last week, where changing the promptString to a regular string fixed the issue for me that's why i assumed it was prompt being malformed. 

Maybe @dominiccooney can confirm if this is still an issue for him on main?

dominiccooney: I have seen early truncation and repetition at 6db70c3fd28102900b75c66e4dac8966e85a28d2 but I have not had the opportunity to dig into whether it is the underlying model being weird, or our ollama adapter being weird.

Recalling our experience with UTF decoding of Anthropic completions, I would like to see the decoding hardened a bit per my earlier feedback:
- Don't assume network responses are chunked on UTF character boundaries.
- Don't assume that, having seen the end of the network response, that you have decoded all the characters in the response yet; the last packet may contain data.

dominiccooney (changes requested):
I believe there are still problems with ollama response decoding and this PR is on the right track. It just needs to go further in handling the incremental responses.

dominiccooney (changes requested):
Specific feedback inline.

lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -48,46 +46,52 @@ export async function ollamaChatClient(
...
 
-            const textDecoder = new TextDecoder()
-            const decoded = textDecoder.decode(value, { stream: true })
-            const parsedLine = JSON.parse(decoded) as OllamaGenerateResponse
+            let responseText = ''
+            while (true) {
+                try {
+                    const { done, value } = await reader.read()
+                    if (typeof value === 'string') {
+                        const parsedData = JSON.parse(value) as OllamaGenerateResponse
```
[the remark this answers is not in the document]
abeatrix: @dominiccooney hey Dom sorry i haven't got a chance to work on this since I''m still working on my sprint works, but I will try to pick this back up tomorrow. Will mark it as a draft until this is ready again, appreciate your feedback

abeatrix:
@dominiccooney i've applied your feedback to the PR:
- [x] Use a TextDecoderStream
- [x] Accumulate the decoded text into a string
- [x] Break on newlines
- [x] JSON.parse each chunk and act on it
- [x] Don't assume that, having seen the end of the network response, that you have decoded all the characters in the response yet; the last packet may contain data.
  - process `if (value)` where value will be decoded first, before checking for `if (done)`

lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -48,46 +50,69 @@ export async function ollamaChatClient(
...
+            let stopReason = ''
+            let completion = ''
+
+            while (!stopReason) {
+                const { done, value } = await reader.read()
+
+                // Splits the decoded chunk by the new lines and filters out empty strings.
+                if (value) {
+                    for (const chunk of value.split(RESPONSE_SEPARATOR).filter(Boolean)) {
+                        const line = JSON.parse(chunk) as OllamaGenerateResponse
```
valerybugakov, line 83: Similar to the completions client, we should buffer the chunks in case they are not cut on JSON boundaries. 

What do you think about reusing the response streaming/reading part from the completions client to keep this logic in one place?

abeatrix:
@dominiccooney @valerybugakov I've updated the PR to use the official Ollama js package for chat instead of implementing our own client. May I get your review again please?

lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -23,71 +22,76 @@ export async function ollamaChatClient(
...
-            tfs_z: params.maxTokensToSample,
-        },
-    } satisfies OllamaChatParams
+    const config = getCompletionsModelConfig(params.model)
+    const model = config?.model ?? params.model
+    const endpoint = config?.endpoint ?? ollamaHost
+
+    // Update the host if it's different from the current one.
+    if (endpoint !== ollamaHost) {
+        ollamaHost = endpoint
```
valerybugakov, line 37: Do we need to mutate the `ollamaHost` variable?

lib/shared/src/llm-providers/ollama/chat-client.ts
```diff
@@ -23,71 +22,76 @@ export async function ollamaChatClient(
...
+        ollama = new Ollama({ host: endpoint })
```
valerybugakov, line 38: Can we create `ollama` on every `ollamaChatClient` call and remove the mutable variable?

lib/shared/tsconfig.json
```diff
@@ -4,8 +4,18 @@
     "module": "ESNext",
     "rootDir": "src",
     "outDir": "dist",
-    "lib": ["ESNext", "DOM", "DOM.Iterable"],
+    "lib": [
+      "ESNext",
+      "DOM",
+      "DOM.Iterable"
+    ],
+    "moduleResolution": "Bundler",
```
valerybugakov, line 12: Is this addition required?

abeatrix: I think so. Without it the client will complain with:

```
lib/shared/src/llm-providers/ollama/chat-client.ts:1:24 - error TS2307: Cannot find module 'ollama/browser' or its corresponding type declarations.
  There are types at '/Users/bwork/dev/cody-2/lib/shared/node_modules/ollama/dist/browser.d.mts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.
  
  ```

abeatrix:
Verified Ollama still works with the latest commit:

image

valerybugakov (approved):
Nice!

PR #3882, opened 2024-04-20
adding context snippets and the repo name in the context summary

hitesh-1997:
## Context
**Summary**: The PR adds an additional field `gitMetadata` in the private metadata field, for the `cody.chat-question/submitted` event.

In the current telemetry, we only log the `promptText` entered by the user, without logging the context items. Because we don't have the context items, we are not able to make inference offline on the cody data and iterate on downstream tasks such as (Experimenting with label generating strategy from attention scores and training on data from cody logs)

Since storing the code snippets could be more sensitive, we start by adding the github repo name for the public repos in the PR, adding a new field `gitMetadata` in the current context summary logs. From the `gitMetadata` we can offline generate the context items using our pipelines (such as agents) 

## Test plan
1. Start the debugger from this branch.
2. Click on the new chat and ask a question. The telemetry response should have an additional field `gitMetadata`, reflecting the git url and commit of the repo. If the repo is private, the field will be empty

## Example Logs

For the public Repo:
```
█ telemetry-v2: recordEvent: cody.chat-question/submitted  {
  "parameters": {
    "version": 0,
    "metadata": [
      {
        "key": "recordsPrivateMetadataTranscript",
        "value": 1
      }
    ],
    "privateMetadata": {
      "requestID": "eadbc0fb-d602-49f1-ac99-418f6d776cb4",
      "chatModel": "anthropic/claude-3-sonnet-20240229",
      "source": "chat",
      "traceId": "00000000000000000000000000000000",
      "sessionID": "Fri, 03 May 2024 20:52:41 GMT",
      "promptText": "hi",
      "gitMetadata": "{\"githubUrl\":\"github.com/princeton-nlp/SWE-bench\",\"commit\":\"cb47eaffbacb8530739209e8d1d1c6b485b4eb3c\"}"
    }
  },
  "timestamp": "2024-05-03T20:54:36.404Z"
}
```
For the private repo:
```
█ telemetry-v2: recordEvent: cody.chat-question/submitted  {
  "parameters": {
    "version": 0,
    "metadata": [
      {
        "key": "recordsPrivateMetadataTranscript",
        "value": 1
      }
    ],
    "privateMetadata": {
      "requestID": "faaaf8bd-cbf6-4284-9791-2c6054d4e3f8",
      "chatModel": "anthropic/claude-3-sonnet-20240229",
      "source": "chat",
      "traceId": "00000000000000000000000000000000",
      "sessionID": "Wed, 01 May 2024 17:15:12 GMT",
      "promptText": "hi",
      "gitRemoteUrl": ""
    }
  },
  "timestamp": "2024-05-01T17:15:15.875Z"
}
```

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -88,6 +88,12 @@ import { SimpleChatModel, prepareChatMessage } from './SimpleChatModel'
 import { getChatPanelTitle, openFile } from './chat-helpers'
 import { getEnhancedContext } from './context'
 import { DefaultPrompter, type IPrompter } from './prompt'
+import { gitRemoteUrlFromGitExtension } from '../../repository/git-extension-api'
+import {
+    LOGGER_CONTEXT_SNIPPET_SUMMARY_FIELD,
+    LOGGER_CONTEXT_REPO_NAME_FIELD,
+    LOGGER_MAX_CONTEXT_SNIPPET_LENGTH
+} from '../context/constants'
 
 interface SimpleChatPanelProviderOptions {
     config: ChatPanelConfig
@@ -840,17 +846,34 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         if (sendTelemetry) {
             // Create a summary of how many code snippets of each context source are being
             // included in the prompt
-            const contextSummary: { [key: string]: number } = {}
+            const contextSummary: { [key: string]: number | string } = {}
             for (const { source } of newContextUsed) {
                 if (!source) {
                     continue
                 }
                 if (contextSummary[source]) {
-                    contextSummary[source] += 1
+                    contextSummary[source] = Number(contextSummary[source]) + 1;
                 } else {
-                    contextSummary[source] = 1
+                    contextSummary[source] = 1;
                 }
             }
+
+            const vscodeUri = vscode.workspace.workspaceFolders?.[0]?.uri
+            if (vscodeUri) {
+                const gitUri = gitRemoteUrlFromGitExtension(vscodeUri)
+                if (gitUri) {
+                    contextSummary[LOGGER_CONTEXT_REPO_NAME_FIELD] = gitUri
+                }
+            }
+            const contextSnippetSummary = []
+            for (const contextSourceItem of newContextUsed) {
+                contextSnippetSummary.push({
+                    'source': contextSourceItem.source,
+                    'content': contextSourceItem.content?.slice(0, LOGGER_MAX_CONTEXT_SNIPPET_LENGTH),
+                    'url': contextSourceItem.uri.path,
+                })
+            }
+            contextSummary[LOGGER_CONTEXT_SNIPPET_SUMMARY_FIELD] = JSON.stringify(contextSnippetSummary)
 
             // Log the size of all user context items (e.g., @-mentions)
             // Includes the count of files and the size of each file
@@ -864,7 +887,6 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                 included: getContextStats(newContextUsed.filter(f => f.source === 'user')),
                 excluded: getContextStats(newContextIgnored.filter(f => f.source === 'user')),
             }
-
             sendTelemetry(contextSummary, privateContextStats)
         }
 
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -73,11 +73,6 @@ import { gitRemoteUrlFromGitExtension } from '../../repository/git-extension-api
 import { recordExposedExperimentsToSpan } from '../../services/open-telemetry/utils'
 import type { MessageErrorType } from '../MessageProvider'
 import { getChatContextItemsForMention } from '../context/chatContext'
-import {
-    LOGGER_CONTEXT_REPO_NAME_FIELD,
-    LOGGER_CONTEXT_SNIPPET_SUMMARY_FIELD,
-    LOGGER_MAX_CONTEXT_SNIPPET_LENGTH,
-} from '../context/constants'
 import type {
     ChatSubmitType,
     ConfigurationSubsetForWebview,
@@ -435,6 +430,11 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         return tracer.startActiveSpan('chat.submit', async (span): Promise<void> => {
             span.setAttribute('sampled', true)
 
+            let gitUrl;
+            const vscodeUri = vscode.workspace.workspaceFolders?.[0]?.uri
+            if (vscodeUri) {
+                gitUrl = gitRemoteUrlFromGitExtension(vscodeUri)
+            }
             const authStatus = this.authProvider.getAuthStatus()
             const sharedProperties = {
                 requestID,
@@ -458,6 +458,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                     // the condition below is an additional safeguard measure
                     promptText:
                         authStatus.isDotCom && truncatePromptString(inputText, CHAT_INPUT_TOKEN_BUDGET),
+                    gitUrl: gitUrl,
                 },
             })
 
@@ -846,34 +847,17 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         if (sendTelemetry) {
             // Create a summary of how many code snippets of each context source are being
             // included in the prompt
-            const contextSummary: { [key: string]: number | string } = {}
+            const contextSummary: { [key: string]: number } = {}
             for (const { source } of newContextUsed) {
                 if (!source) {
                     continue
                 }
                 if (contextSummary[source]) {
-                    contextSummary[source] = Number(contextSummary[source]) + 1
+                    contextSummary[source] += 1
                 } else {
                     contextSummary[source] = 1
                 }
             }
-
-            const vscodeUri = vscode.workspace.workspaceFolders?.[0]?.uri
-            if (vscodeUri) {
-                const gitUri = gitRemoteUrlFromGitExtension(vscodeUri)
-                if (gitUri) {
-                    contextSummary[LOGGER_CONTEXT_REPO_NAME_FIELD] = gitUri
-                }
-            }
-            const contextSnippetSummary = []
-            for (const contextSourceItem of newContextUsed) {
-                contextSnippetSummary.push({
-                    source: contextSourceItem.source,
-                    content: contextSourceItem.content?.slice(0, LOGGER_MAX_CONTEXT_SNIPPET_LENGTH),
-                    url: contextSourceItem.uri.path,
-                })
-            }
-            contextSummary[LOGGER_CONTEXT_SNIPPET_SUMMARY_FIELD] = JSON.stringify(contextSnippetSummary)
 
             // Log the size of all user context items (e.g., @-mentions)
             // Includes the count of files and the size of each file
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -70,6 +70,7 @@ import type { RemoteRepoPicker } from '../../context/repo-picker'
 import type { ContextRankingController } from '../../local-context/context-ranking'
 import { chatModel } from '../../models'
 import { gitRemoteUrlFromGitExtension } from '../../repository/git-extension-api'
+import { RepoMetadatafromGitApi } from '../../repository/repo-metadata-from-git-api'
 import { recordExposedExperimentsToSpan } from '../../services/open-telemetry/utils'
 import type { MessageErrorType } from '../MessageProvider'
 import { getChatContextItemsForMention } from '../context/chatContext'
@@ -414,6 +415,21 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         this.initDoer.signalInitialized()
     }
 
+    private async getRepoGiturlIfPublic(): Promise<string> {
+        const vscodeUri = vscode.workspace.workspaceFolders?.[0]?.uri
+        if (vscodeUri) {
+            const gitRemoteUrl = gitRemoteUrlFromGitExtension(vscodeUri)
+            if (gitRemoteUrl) {
+                const instance = RepoMetadatafromGitApi.getInstance()
+                const gitMetaData = await instance.getRepoMetadataUsingGitUrl(gitRemoteUrl)
+                if (gitMetaData?.repoVisibility === 'public') {
+                    return gitRemoteUrl
+                }
+            }
+        }
+        return ''
+    }
+
     /**
      * Handles user input text for both new and edit submissions
      */
@@ -429,15 +445,6 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
     ): Promise<void> {
         return tracer.startActiveSpan('chat.submit', async (span): Promise<void> => {
             span.setAttribute('sampled', true)
-
-            let gitUrl = ''
-            const vscodeUri = vscode.workspace.workspaceFolders?.[0]?.uri
-            if (vscodeUri) {
-                const gitRemoteUrl = gitRemoteUrlFromGitExtension(vscodeUri)
-                if (gitRemoteUrl) {
-                    gitUrl = gitRemoteUrl
-                }
-            }
             const authStatus = this.authProvider.getAuthStatus()
             const sharedProperties = {
                 requestID,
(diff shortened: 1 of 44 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -818,18 +824,35 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
                 }
                 if (contextSummary[source]) {
-                    contextSummary[source] += 1
+                    contextSummary[source] = Number(contextSummary[source]) + 1
                 } else {
                     contextSummary[source] = 1
                 }
             }
 
+            const vscodeUri = vscode.workspace.workspaceFolders?.[0]?.uri
```
dominiccooney, line 839: We are trying to avoid this API because it is rarely right. Workspaces can have multiple folders. What are you using `LOGGER_CONTEXT_REPO_NAME_FIELD` for?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -818,18 +824,35 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+            if (vscodeUri) {
+                const gitUri = gitRemoteUrlFromGitExtension(vscodeUri)
```
dominiccooney, line 841: I'm curious what the purpose of this data collection is. Would a hash be appropriate instead? Per-device or per-account salted hash? etc.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -818,18 +824,35 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         if (sendTelemetry) {
             // Create a summary of how many code snippets of each context source are being
             // included in the prompt
-            const contextSummary: { [key: string]: number } = {}
+            const contextSummary: { [key: string]: number | string } = {}
```
dominiccooney, line 827: I'm not an expert in the logging system, but if you start logging strings you need to log into the "private" pocket of data that is ablated in the more widely accessible logs, see https://github.com/sourcegraph/cody/blob/main/ARCHITECTURE.md#telemetry

How have you tested that the logging works end to end?

@arafatkatze , is there a good person to tag on reviews for new logged data?

dominiccooney (commented):
Some questions and feedback inline.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
abeatrix: @dominiccooney is right.

- metadata: This gets sent to us on every event for both dotcom and enterprise users, but it can only contain numeric values
- privateMetadata: Only gets sent for dotcom users, but can contain arbitrary string values

Based on this, this should be moved to the new logging data out of contextsummary into privateContextSummary instead

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -818,18 +824,35 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
+            }
+            const contextSnippetSummary = []
+            for (const contextSourceItem of newContextUsed) {
+                contextSnippetSummary.push({
+                    source: contextSourceItem.source,
+                    content: contextSourceItem.content?.slice(0, LOGGER_MAX_CONTEXT_SNIPPET_LENGTH),
+                    url: contextSourceItem.uri.path,
+                })
+            }
+            contextSummary[LOGGER_CONTEXT_SNIPPET_SUMMARY_FIELD] = JSON.stringify(contextSnippetSummary)
```
abeatrix, line 854: This will not work iirc when contains value other than numeric values. 
Let's revert the change to the contextSummary and merge the new contextSnippetSummary with the privateContextStats on line 864 instead?

abeatrix (changes requested):
Comment inline

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
abeatrix: I found this guideline on data collection that might be of use: https://handbook.sourcegraph.com/departments/data-analytics/architecture/#transcript-events-from-dotcom-users

hitesh-1997:
Hi @dominiccooney and @abeatrix 
Thanks a lot for your suggestions and response.
We had a discussion regarding the logging in standup and decided to modify the approach. I should have marked the PR maybe in draft while it was still work in progress. We currently want to log the `repo url`. Ideally, the repo name should be the one, from where the current context files are fetched. To this end, can you please help me with some questions:

1. Detecting if the repo is public or private: Since we only want to log the repo url if the repo is public, one solution i can think of is querying git api -- but that would involve a network call at every request -- do you know if there is a better way to do the same ? (I checked for some alternates like if we can get this via git cli etc -- but git cli doesn't have this support). I also tried exploring `vscodeGitApi` but can't seems to find solution there, but haven't explored in details. 

2. @dominiccooney I see we use the same api `vscode.workspace.workspaceFolders` in local-embeddings code to get the `repoPath` and further `repoName` in the bfg-private binary, and hence I tried to use the same api call. Can you please let me know if there is a better way to detect the `repoPath`  

I am marking the current PR in the draft state for now, and will change for review when I have the implementation ready for the points above.

hitesh-1997:
> Hi @dominiccooney and @abeatrix Thanks a lot for your suggestions and response. We had a discussion regarding the logging in standup and decided to modify the approach. I should have marked the PR maybe in draft while it was still work in progress. We currently want to log the `repo url`. Ideally, the repo name should be the one, from where the current context files are fetched. To this end, can you please help me with some questions:
> 
> 1. Detecting if the repo is public or private: Since we only want to log the repo url if the repo is public, one solution i can think of is querying git api -- but that would involve a network call at every request -- do you know if there is a better way to do the same ? (I checked for some alternates like if we can get this via git cli etc -- but git cli doesn't have this support). I also tried exploring `vscodeGitApi` but can't seems to find solution there, but haven't explored in details.
> 2. @dominiccooney I see we use the same api `vscode.workspace.workspaceFolders` in local-embeddings code to get the `repoPath` and further `repoName` in the bfg-private binary, and hence I tried to use the same api call. I also noticied that symf uses, similar api `vscode.workspace.workspaceFolders?.[0]?.uri.fspath` to get the repoPath. Can you please let me know if there is a better way to detect the `repoPath`
> 
> I am marking the current PR in the draft state for now, and will change for review when I have the implementation ready for the points above.

Hi @dominiccooney 
Do you know if there is an alternate vscode git api we can use to check if a github repo is public or private. One solution we can do is build a local cache to store this data (which will query git api for the first time and store the git_repo_name to public/private status in the memory) Does this solution look alright for now ?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: Thanks for pointing this our Dominic, didn't know that.
For now to avoid getting blocked on training and get sufficient data, I added this condition:
```
const gitRemoteUrlList = gitRemoteUrlsFromGitExtension(vscodeUri)
            if (gitRemoteUrlList && gitRemoteUrlList.length === 1) {
```
With this condition although will miss logging some use cases, but some data should also be sufficient for training.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: Thanks @dominiccooney and @abeatrix for pointing this out.
I changed the logging to the `privateMetadata` field. (Also added the logs in the example to confirm the logging works)
Although the field is getting logged correctly in the vscode debug console. Can you please let know if there any other modification I need to do, for eg: In pub/sub or sourcegraph backend (so that downstream processing pipeline does not ignore this). Adding @akalia25 to confirm this point.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: Thanks for pointing this out @abeatrix !!
I changed the actual logging to the `privateMetadata` field.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -452,6 +468,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
                     // the condition below is an additional safeguard measure
                     promptText:
                         authStatus.isDotCom && truncatePromptString(inputText, CHAT_INPUT_TOKEN_BUDGET),
+                    gitRemoteUrl: await this.getRepoGiturlIfPublic(),
```
abeatrix, line 471: Do we need to add the authStatus.isDotCom check here as well?

abeatrix (commented):
Left some comments inline. I'm afraid I am still missing some context as in why we need to know if the repository is private or not before we log the codebase name. Can you elaborate more on that? I also notice your current approach would only validate git repositories that are hosted on github, is that the expected result?

Additional question: 
- are you trying to log that info for dot com users only, or also for enterprise users? 
-If it's only for dotCom users, could we just log the name of the codebase and do the private v.s. public matching offline?  
- since Sourcegraph.com doesn't support private repository, maybe we can just perform a one time check against dotCom to check if it's available there or not?

vscode/src/repository/repo-metadata-from-git-api.ts
```diff
@@ -0,0 +1,80 @@
...
+        if (!ownerAndRepoName) {
+            return undefined
+        }
+        const owner = ownerAndRepoName.owner
+        const repoName = ownerAndRepoName.repoName
+        const repoMetaData = await this.queryGitApi(owner, repoName)
+        return repoMetaData
+    }
+
+    private async queryGitApi(owner: string, repoName: string): Promise<RepoMetaData | undefined> {
```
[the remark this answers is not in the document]
hitesh-1997: done, changed to `queryGitHubApi `

vscode/src/repository/repo-metadata-from-git-api.ts
```diff
@@ -0,0 +1,80 @@
...
+        try {
+            const response = await fetch(apiUrl)
+            if (!response.ok) {
+                return undefined
+            }
+            const repoData = await response.json()
+            return {
+                owner,
+                repoName,
+                repoVisibility: repoData.private ? 'private' : 'public',
```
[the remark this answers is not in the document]
hitesh-1997: Hey @abeatrix 
Definitely, earlier whenver we get any error I was always returning `undefined` so that we don't log. But for specific case of `404` status code, I changed it to return `private` as repoVisibility.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: Thanks for noticing this !!
I added this condition

hitesh-1997:
Hi @abeatrix 
Adding the response for your queries below :)

> I'm afraid I am still missing some context as in why we need to know if the repository is private or not before we log the codebase name
> are you trying to log that info for dot com users only, or also for enterprise users? 
> -If it's only for dotCom users, could we just log the name of the codebase and do the private v.s. public matching offline?
As per discussion on this internally with @chillatom @rishabhmehrotra, there were concerns about logging any information regarding private repos even for the dotcom users, as it could violate terms & condition on logging and decided that it would be best to start logging only for public repos.

> I also notice your current approach would only validate git repositories that are hosted on github, is that the expected result?
Yes,  this is expected for now. This should get us enough data and coverage of the repos to start with the training. Please note the goal of the PR is not to get the 100% coverage. Gettting this data for some repos should be enough to avoid getting blocked on training on cody prompt and repos.

> since Sourcegraph.com doesn't support private repository, maybe we can just perform a one time check against dotCom to check if it's available there or not?
Sorry I didn't get this. With this logging we get access to the tuples (`promptText`, `gitRepoUrl`) on which we can train offline. Can you please elaborate on this, what should we check against dotcom repos ?

susnato:
Hi @hitesh-1997 by the way are we logging the commit hash for the repos?

hitesh-1997:
> Hi @hitesh-1997 by the way are we logging the commit hash for the repos?

yes, we are logging `giturl` and `commit` fields.

vscode/src/repository/repo-metadata-from-git-api.ts
```diff
@@ -0,0 +1,75 @@
+export interface RepoMetaData {
+    owner: string
+    repoName: string
+    repoVisibility: 'public' | 'private'
+}
+
+export class RepoMetadatafromGitApi {
```
abeatrix, line 7: Maybe Beyang or Dom can confirm, but we probably don't need a new instance for tracking repo mapping since there is a [codebaseRepoIdMapper in CodebaseStatusProvider](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/CodebaseStatusProvider.ts?L45), or we can just update the current `CodebaseIdentifiers` to include repoVisibility?

hitesh-1997: I think this is different from `codebaseRepoIdMapper`. 

This class just gets a repoVisibility (i.e. if a repo is public or private) and stores it in a local `Map` to avoid network calls. This class is used to set the value of [`isPublic` in the `CodebaseIdentifier`](https://github.com/sourcegraph/cody/pull/3882/files#diff-43f7103f58d4fafcd5079071965fd20533d583dbc0a65c4393efaa5c226a8b36R167).

So, in summary this class does what you suggested in your PR [here](https://github.com/sourcegraph/cody/pull/4043/files#diff-f58fc2635f09145f2bf5b08cd81122f890d4c5c9b8e6934558e3f60b81171f0dR110) and adds a cache on top of it to reuse the results.

Please let me know if the name of the class itself is confusing.

vscode/src/repository/repo-metadata-from-git-api.ts
```diff
@@ -0,0 +1,75 @@
...
+            const repoData = await response.json()
+            return {
+                owner,
+                repoName,
+                repoVisibility: repoData.private ? 'private' : 'public',
+            }
+        } catch (error) {
+            return undefined
+        }
+    }
```
abeatrix, line 65:
```suggestion
    private async queryGitHubApi(owner: string, repoName: string): Promise<RepoMetaData | undefined> {
        const apiUrl = `https://api.github.com/repos/${owner}/${repoName}`
        const metadata = { owner, repoName, repoVisibility: false}
        try {
            const response = await fetch(repositoryUrl, { method: 'HEAD' })
            metadata.repoVisibility = response.ok
            return metadata
        } catch (error) {
            return undefined
        }
    }
```
I think we can just check if the response returns `ok` here since it returns false if 404?

image

Example from https://github.com/sourcegraph/cody/blob/44ae4407980ee21ae0793143bd92c0d137810bd7/vscode/src/repository/git-extension-api.ts#L110-L127

hitesh-1997: Thanks for the suggestion, looks much cleaner now.
In addition to response.ok, I am also checking `repoData.private` as per [github api documentation](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-an-organization-repository).
`metadata.isPublic = response.ok && repoData.private === false `

hitesh-1997:
Hi @abeatrix 
Can you please let me know if there are any additional comments or it look good to be merged ?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -414,6 +414,18 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
         this.initDoer.signalInitialized()
     }
 
+    private async getRepoMetadataIfPublic(): Promise<string> {
```
abeatrix, line 417: nit: this method seems more relevant to CodybaseStatusProvider than SimpleChatPanel provider.

abeatrix (approved):
approve to unblock. 
The logging side of things looks good to me. we are only logging this for non-enterprise users. 

I am not sure if we need a new class for mapping codebase visibility, so it would be great if you could get @dominiccooney 's confirmation on that first.

hitesh-1997:
> approve to unblock. The logging side of things looks good to me. we are only logging this for non-enterprise users.
> 
> I am not sure if we need a new class for mapping codebase visibility, so it would be great if you could get @dominiccooney 's confirmation on that first.

Thanks for the review @abeatrix :)
My reasoning for making this in a separate class is to model a separate functionality which involves `github api` interactions and since such interactions results in additional network calls, we might want to add additional storage for payload giving the same output. For eg: `github repo` is either `public` or `private` should not change within a user session, so added a seperate map (For long term  persistence we can also use vscode local storage) etc etc. So, I was thinking, all this could be better to do in a different class rather than mixing with existing functionality.

But @dominiccooney @abeatrix please feel free to suggest any other alternatives you think would be better from architecture point of view, would be happy to incorporate the change :)

vscode/src/repository/repo-metadata-from-git-api.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    private parserOwnerAndRepoName(gitUrl: string): { owner: string; repoName: string } | undefined {
+        const match = gitUrl?.match(/github\.com\/([^/]+)\/([^/]+?)(?:\.git)?$/)
+        if (!match) {
+            return undefined
+        }
+        const [, owner, repoName] = match
+        return { owner, repoName }
+    }
+}
```
keegancsmith, line 61: @hitesh-1997 this implementation won't handle a bunch of cases, eg SSH clone URLs. Luckily we already have a pretty robust function for massaging a gitURL `convertGitCloneURLToCodebaseName`

PR #3883, opened 2024-04-20
introduce ContextMentionProvider API for mentionable context sources

sqs:
A ContextMentionProvider can provide context items that the user can `@`-mention in chat. It exposes an API with:

- triggerPrefixes: eg `npm:` to show a list of possible npm packages to mention when the user types `@npm:`
- queryContextItems: return a list of possible npm packages when the user types in `@npm:left-p` into the chat
- resolveContextItem (optional): enrich the context item with more content, used right before the context is sent to the LLM and only run on items that the user explicitly mentioned (so it can be slower)

The experimental URL mention feature was also ported to use this new API. It was designed to support more mentionable context sources, such as https://github.com/sourcegraph/cody/pull/3866 (@-mentionable packages like `@npm:left-pad`).

**Status: experimental.** For now, this API is only enabled (and these context sources are only available) when the `cody.experimental.noodle` VS Code setting is `true`. If this setting is off, it is not possible for any new ContextMentionProviders to be triggered, which is how we can experiment with adding new ones with less risk to the overall stability of Cody.

Go to [API for context sources (Sourcegraph community forum post)](https://community.sourcegraph.com/t/api-for-context-sources-contextmentionprovider-openctx/152) to read and share feedback/ideas about how you'd want to use this API as a dev.

## Test plan

CI (which includes the revived URL mention e2e test)

### DIFF
--- a/lib/shared/src/mentions/api.ts
+++ b/lib/shared/src/mentions/api.ts
@@ -1,0 +1,54 @@
+import type { ContextItem, ContextItemWithContent } from '../codebase-context/messages'
+import { URL_CONTEXT_MENTION_PROVIDER } from './providers/urlMentions'
+
+/**
+ * A unique identifier for a {@link ContextMentionProvider}.
+ */
+export type ContextMentionProviderID = string
+
+/**
+ * Providers that supply context that the user can @-mention in chat.
+ *
+ * This API is *experimental* and subject to rapid, unannounced change.
+ *
+ * In VS Code, use {@link getEnabledContextMentionProviders} instead of this.
+ */
+export const CONTEXT_MENTION_PROVIDERS: ContextMentionProvider[] = [URL_CONTEXT_MENTION_PROVIDER]
+
+/**
+ * A provider that can supply context for users to @-mention in chat.
+ *
+ * This API is *experimental* and subject to rapid, unannounced change.
+ */
+export interface ContextMentionProvider<ID extends ContextMentionProviderID = ContextMentionProviderID> {
+    id: ID
+
+    /**
+     * Prefix strings for the user input after the `@` that trigger this provider. For example, a
+     * context mention provider with prefix `npm:` would be triggered when the user types `@npm:`.
+     */
+    triggerPrefixes: string[]
+
+    /**
+     * Get a list of possible context items to show (in a completion menu) when the user triggers
+     * this provider while typing `@` in a chat message.
+     */
+    queryContextItems(query: string, signal?: AbortSignal): Promise<ContextItemFromProvider<ID>[]>
+
+    /**
+     * Resolve a context item to one or more items that have the {@link ContextItem.content} field
+     * filled in. A provider is called to resolve only the context items that it returned in
+     * {@link queryContextItems} and that the user explicitly added.
+     */
+    resolveContextItem?(
+        item: ContextItemFromProvider<ID>,
+        signal?: AbortSignal
+    ): Promise<ContextItemWithContent[]>
+}
+
+type ContextItemFromProvider<ID extends ContextMentionProviderID> = ContextItem & {
+    /**
+     * The ID of the {@link ContextMentionProvider} that supplied this context item.
+     */
+    provider: ID
+}
--- a/lib/shared/src/mentions/api.ts
+++ b/lib/shared/src/mentions/api.ts
@@ -32,6 +32,8 @@ export interface ContextMentionProvider<ID extends ContextMentionProviderID = Co
     /**
      * Get a list of possible context items to show (in a completion menu) when the user triggers
      * this provider while typing `@` in a chat message.
+     *
+     * {@link query} omits the `@` but includes the trigger prefix from {@link triggerPrefixes}.
      */
     queryContextItems(query: string, signal?: AbortSignal): Promise<ContextItemFromProvider<ID>[]>
 
--- a/lib/shared/src/mentions/providers/urlMentions.ts
+++ b/lib/shared/src/mentions/providers/urlMentions.ts
@@ -3,7 +3,7 @@ import {
     type ContextItem,
     ContextItemSource,
     type ContextItemWithContent,
-} from '../codebase-context/messages'
+} from '../../codebase-context/messages'
 
 /**
  * Given a possibly incomplete URL from user input (that the user may be typing), return context
--- a/lib/shared/src/mentions/providers/urlMentions.ts
+++ b/lib/shared/src/mentions/providers/urlMentions.ts
@@ -1,56 +1,49 @@
 import { URI } from 'vscode-uri'
-import {
-    type ContextItem,
-    ContextItemSource,
-    type ContextItemWithContent,
-} from '../../codebase-context/messages'
+import { ContextItemSource, type ContextItemWithContent } from '../../codebase-context/messages'
+import type { ContextMentionProvider } from '../api'
 
-/**
- * Given a possibly incomplete URL from user input (that the user may be typing), return context
- * items from fetching the URL and extracting its text content.
- */
-export async function getURLContextItems(
-    urlInput: string,
-    signal?: AbortSignal
-): Promise<ContextItem[]> {
-    const url = tryParsePossiblyIncompleteURL(urlInput)
-    if (url === null) {
-        return []
-    }
+export const URL_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'url'> = {
+    id: 'url',
+    triggerPrefixes: ['http://', 'https://'],
 
-    try {
-        const content = await fetchContentForURLContextItem(url.toString(), false, signal)
-        if (content === null) {
+    /**
+     * Given a possibly incomplete URL from user input (that the user may be typing), return context
+     * items from fetching the URL and extracting its text content.
+     */
+    async queryContextItems(query, signal) {
+        const url = tryParsePossiblyIncompleteURL(query)
+        if (url === null) {
             return []
         }
-        return [
-            {
-                type: 'file',
-                uri: url,
-                content,
-                title: tryGetHTMLDocumentTitle(content),
-                source: ContextItemSource.Uri,
-            },
-        ]
-    } catch (error) {
-        // Suppress errors because the user might be typing a URL that is not yet valid.
-        return []
-    }
-}
 
-export function isURLContextItem(item: Pick<ContextItem, 'uri'>): boolean {
-    return item.uri.scheme === 'http' || item.uri.scheme === 'https'
-}
+        try {
+            const content = await fetchContentForURLContextItem(url.toString(), false, signal)
+            if (content === null) {
+                return []
+            }
+            return [
+                {
+                    type: 'file',
+                    uri: url,
+                    content,
+                    title: tryGetHTMLDocumentTitle(content),
+                    source: ContextItemSource.Uri,
+                    provider: 'url',
+                },
+            ]
+        } catch (error) {
+            // Suppress errors because the user might be typing a URL that is not yet valid.
+            return []
+        }
+    },
 
-export async function resolveURLContextItem(
-    item: ContextItem,
-    signal?: AbortSignal
-): Promise<ContextItemWithContent> {
-    if (item.content !== undefined) {
-        return item as ContextItemWithContent
-    }
-    const content = await fetchContentForURLContextItem(item.uri.toString(), true, signal)
-    return { ...item, content }
+    async resolveContextItem(item, signal) {
+        if (item.content !== undefined) {
+            return [item as ContextItemWithContent]
+        }
+        const content = await fetchContentForURLContextItem(item.uri.toString(), true, signal)
+        return [{ ...item, content }]
+    },
 }
 
 async function fetchContentForURLContextItem(
--- a/lib/shared/src/mentions/providers/urlMentions.ts
+++ b/lib/shared/src/mentions/providers/urlMentions.ts
@@ -17,7 +17,7 @@ export const URL_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'url'> = {
         }
 
         try {
-            const content = await fetchContentForURLContextItem(url.toString(), false, signal)
+            const content = await fetchContentForURLContextItem(url.toString(), signal)
             if (content === null) {
                 return []
             }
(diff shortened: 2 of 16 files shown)


PriNova:
Will the `ContextMentionProvider` API be general in the sense, that the user can (in whatever way) fetch context programically, like scraping web conent, executing custom scripts, fetching weather data or whatever the user likes to include as context?

vscode/test/e2e/chat-atFile.test.ts
```diff
@@ -221,13 +221,13 @@ test('editing a chat message with @-mention', async ({ page, sidebar }) => {
     await expect(chatInput).toBeEmpty()
     await expect(chatPanelFrame.getByText('Explain @Main.java')).toBeVisible()
     const contextCell = getContextCell(chatPanelFrame)
-    await expectContextCellCounts(contextCell, { files: 1, lines: 10 })
```
abeatrix, line 224: Is this intended? This shows a regression from https://github.com/sourcegraph/cody/pull/3842, or should we remove the total line count altogether and just display file count? 👀

sqs: @abeatrix Yeah, I opened https://github.com/sourcegraph/cody/pull/3886 to discuss this specific change separately.

sqs:
@PriNova: 
> Will the `ContextMentionProvider` API be general in the sense, that the user can (in whatever way) fetch context programmatically, like scraping web content, executing custom scripts, fetching weather data or whatever the user likes to include as context?

Yes, but initially (1) it's experimental and we don't recommend anyone else building on it because it will change a lot and (2) it's only for explicitly at-mentioned context, not for automatically included context. BTW in case you haven't see it, https://openctx.org/ is a similar concept that we have that will help us bring more context in.

I just opened https://community.sourcegraph.com/t/api-for-context-sources-contextmentionprovider-openctx/152 for discussion about this API and building on it. Would love to hear your ideas and feedback over there!

PriNova:
> Would love to hear your ideas and feedback over there!

Perfect @sqs . That is a great addition. Since Cody can fetch URL content, I can write a local server which acts as a custom content provider for whatever I like to put into context. Very excited about.

lib/shared/src/mentions/providers/urlMentions.ts
```diff
@@ -1,49 +1,71 @@
...
+            // Suppress errors because the user might be typing a URL that is not yet valid.
+            return []
+        }
+    },
+
+    async resolveContextItem(item, signal) {
+        if (item.content !== undefined) {
+            return [item as ContextItemWithContent]
+        }
+        const content = await fetchContentForURLContextItem(item.uri.toString(), true, signal)
```
keegancsmith, line 44: in what situation would this happen? It seems you only return content set from queryContextItems.

lib/shared/src/mentions/providers/urlMentions.ts
```diff
@@ -1,49 +1,71 @@
...
+    signal?: AbortSignal
+): Promise<string>
+async function fetchContentForURLContextItem(
+    url: string,
+    throwIfNotOK: false,
+    signal?: AbortSignal
+): Promise<string | null>
+async function fetchContentForURLContextItem(
     url: string,
+    throwIfNotOK: boolean,
```
keegancsmith, line 61: it seems from the callsites you don't need this. The only place you set this to false you catch all exceptions anyway and return the same thing.

lib/shared/src/mentions/query.ts
```diff
@@ -23,18 +25,25 @@ export interface MentionQuery {
  * where {@link query} may begin with `@` is if the user is searching for context items that contain
  * `@`, such as if the user typed `@@foo` to mention a file that is literally named `@foo.js`.
  */
-export function parseMentionQuery(query: string): MentionQuery {
+export function parseMentionQuery(
+    query: string,
+    contextMentionProviders: Pick<ContextMentionProvider, 'id' | 'triggerPrefixes'>[]
```
keegancsmith, line 30: noob question: is declaring the subset you use idiomatic in typescript? If I where to guess for exported functions you do this, but for non-exported you don't have to be so specific? Is the purpose of this for easier test generation, or is this useful for call sites to know as well?

vscode/src/chat/context/chatContext.ts
```diff
@@ -56,23 +56,34 @@ export async function getChatContextItemsForMention(
...
-                : []
-        default:
+
+        default: {
+            for (const provider of getEnabledContextMentionProviders()) {
+                if (provider.id === mentionQuery.provider) {
+                    return provider.queryContextItems(
+                        mentionQuery.text,
+                        convertCancellationTokenToAbortSignal(cancellationToken)
+                    )
```
keegancsmith, line 66: we are not respecting MAX_RESULTS here. Not sure if we need to? Should we passing this in as a parameter so the context providers are aware of this?

vscode/src/chat/context/chatContext.ts
```diff
@@ -56,23 +56,34 @@ export async function getChatContextItemsForMention(
...
-            return (await isURLContextFeatureFlagEnabled())
-                ? getURLContextItems(
-                      mentionQuery.text,
-                      convertCancellationTokenToAbortSignal(cancellationToken)
-                  )
-                : []
-        default:
+
+        default: {
+            for (const provider of getEnabledContextMentionProviders()) {
```
keegancsmith, line 61: noob question: is it possible for the return value of getEnabledContextMentionProviders to change between the earlier call and this call? In practice this is fine since we will just return no context, but just interested. I assume the answer is no since there is no async stuff going on, but I suppose it is possible for there to be a blocking call somewhere here?

vscode/src/editor/utils/editor-context.ts
```diff
@@ -274,31 +272,62 @@ export async function filterContextItemFiles(
...
+    const resolvedItems = item.provider
+        ? await resolveContextMentionProviderContextItem(item)
+        : [await resolveFileOrSymbolContextItem(item, editor)]
+    return resolvedItems.map(resolvedItem => ({
+        ...resolvedItem,
+        size: resolvedItem.size ?? TokenCounter.countTokens(resolvedItem.content),
+    }))
+}
+
+export async function resolveContextMentionProviderContextItem({
```
keegancsmith, line 307: minor: I don't think this needs to be exported.

vscode/src/editor/utils/editor-context.ts
```diff
@@ -274,31 +272,62 @@ export async function filterContextItemFiles(
...
+    contextItem: ContextItem,
+    editor: Editor
+): Promise<ContextItemWithContent> {
+    const content =
+        contextItem.content ??
+        (await editor.getTextEditorContentForFile(contextItem.uri, toVSCodeRange(contextItem.range)))
+    return {
+        ...contextItem,
+        content,
+        size: contextItem.size ?? TokenCounter.countTokens(content),
```
keegancsmith, line 331: minor: you don't need to set size here, `resolveContextItem` will fallback to countTokens on content if unset.

keegancsmith (commented):
reviewed this to learn some TS and also cause gonna work on top of this. Thanks! Will leave actual approval to someone who knows more, but it LGTM :)

vscode/src/chat/context/chatContext.ts
```diff
@@ -20,24 +19,25 @@ export async function getChatContextItemsForMention(
...
     // Logging: log when the at-mention starts, and then log when we know the type (after the 1st
     // character is typed). Don't log otherwise because we would be logging prefixes of the same
     // query repeatedly, which is not needed.
-    if (mentionQuery.type === 'empty') {
+    if (mentionQuery.provider === 'default') {
         telemetryRecorder?.empty()
     } else if (mentionQuery.text.length === 1) {
-        telemetryRecorder?.withType(mentionQuery.type)
+        telemetryRecorder?.withType(mentionQuery.provider)
     }
```
keegancsmith, line 36: I don't think this is accurate anymore? We will only know what the mention provider is after many more characters depending on the length of the triggerPrefix. Deciding when to log which provider we are using likely will have to happen at a different time with more complicated logic? I'm not even sure if this logic was that great before since if we have any debouncing we will miss logging this.

lib/shared/src/mentions/providers/urlMentions.ts
```diff
@@ -1,49 +1,71 @@
...
-    } catch (error) {
-        // Suppress errors because the user might be typing a URL that is not yet valid.
-        return []
-    }
-}
 
-export function isURLContextItem(item: Pick<ContextItem, 'uri'>): boolean {
-    return item.uri.scheme === 'http' || item.uri.scheme === 'https'
+        try {
+            const content = await fetchContentForURLContextItem(url.toString(), false, signal)
```
dominiccooney, line 20: I realize you're just moving stuff around here, but since you're making an extension point now...

Is there any cooldown on this? These fetches are visible, so `https://customer.source` could observe when people are on their way to typing `https://customer.sourcegraph.com` and so on.

lib/shared/src/mentions/api.ts
```diff
@@ -0,0 +1,54 @@
...
+ * In VS Code, use {@link getEnabledContextMentionProviders} instead of this.
+ */
+export const CONTEXT_MENTION_PROVIDERS: ContextMentionProvider[] = [URL_CONTEXT_MENTION_PROVIDER]
+
+/**
+ * A provider that can supply context for users to @-mention in chat.
+ *
+ * This API is *experimental* and subject to rapid, unannounced change.
+ */
+export interface ContextMentionProvider<ID extends ContextMentionProviderID = ContextMentionProviderID> {
```
dominiccooney, line 23: Seems like a straightforward starting point for this purpose.

It would raise the comfort level immensely to see a JetBrains native autocomplete hooked up to the same source... There's something @beyang hand rolled for at-file mentions in chat right now, but we need real autocomplete on @-tags for edits...

lib/shared/src/mentions/providers/urlMentions.ts
```diff
@@ -1,49 +1,71 @@
...
+        }
+        const content = await fetchContentForURLContextItem(item.uri.toString(), true, signal)
+        return [{ ...item, content }]
+    },
 }
 
-export async function fetchContentForURLContextItem(
+async function fetchContentForURLContextItem(
+    url: string,
+    throwIfNotOK: true,
```
dominiccooney, line 51: Please avoid boolean parameters: They are inscrutable at the callsite. They are hard to read when you stack a negation like this... when throw if *not* OK is *false* etc... They are harder to generalize if you realize there's a third option you want. They are easy to mix up with empty string/undefined/etc. in TypeScript (although I believe @philipp-spiess turned on some lints requiring explicit booleans recently...)

dominiccooney (approved):
Looks good to me, some feedback inline.

vscode/src/chat/context/chatContext.ts
```diff
@@ -56,23 +56,34 @@ export async function getChatContextItemsForMention(
...
-export async function isURLContextFeatureFlagEnabled(): Promise<boolean> {
-    return (
-        vscode.workspace.getConfiguration('cody').get<boolean>('experimental.urlContext') === true ||
-        (await featureFlagProvider.evaluateFeatureFlag(FeatureFlag.URLContext))
-    )
+export function getEnabledContextMentionProviders(): ContextMentionProvider[] {
+    const isAllEnabled =
+        vscode.workspace.getConfiguration('cody').get<boolean>('experimental.noodle') === true
+    if (isAllEnabled) {
+        return CONTEXT_MENTION_PROVIDERS
```
keegancsmith, line 78: @sqs @thenamankumar I was looking at the package context PR. I was wondering how you imagine we inject the dependency on the sourcegraph graphql client. I think the way this is currently implemented needs to be changed a bit so we can inject dependencies like that.

lib/shared/src/mentions/providers/urlMentions.ts
sqs: Yes, that is currently correct. I could see this URL mention thing doing more work to extract content in the future for the specifically mentioned item (but not for the ones it shows in a menu). @thenamankumar's PR needs this separate query vs. resolve step, and I wanted to illustrate it here. I'll just remove it though.

lib/shared/src/mentions/query.ts
sqs: Yeah, this type stuff is for ease of invocation in tests. It's not idiomatic in TypeScript, but I think it should be. :smile:

vscode/src/chat/context/chatContext.ts
sqs: Let's leave it unlimited and see how we can do this as we write a few more providers.

vscode/src/chat/context/chatContext.ts
sqs: Yeah. Good point. I think we can just log the mentionQuery.provider since this call is debounced enough so that it won't be event-spammy.

vscode/src/chat/context/chatContext.ts
```diff
@@ -56,23 +56,34 @@ export async function getChatContextItemsForMention(
...
+    }
+
+    const isURLProviderEnabled =
```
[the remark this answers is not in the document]
sqs: Yes

lib/shared/src/mentions/providers/urlMentions.ts
sqs: Interesting point. There is a debounce, but it's still possible. Do you think this is a problem solved by sufficiently long debounce, or an explicit commit step?

lib/shared/src/mentions/providers/urlMentions.ts
keegancsmith: @sqs in my working branch I am also using this function, just in the case of this implementation it didn't seem necessary. I think its worth keeping TBH, makes sense to have something which can defer more expensive work.

lib/shared/src/mentions/query.ts
keegancsmith: Thanks for the insights!

lib/shared/src/mentions/providers/urlMentions.ts
thenamankumar: Even URL content fetching on mention is expensive and should be deferred. Right now it is not the case, that is why it takes a few seconds for the select option to show up. 

I can fix it.

PR #3886, opened 2024-04-20
do not show the context file line count in chat context

sqs:
Previously the chat context cell would display `Context: 93 lines from 7 files`. The number of lines is not necessary, and now it just shows the file count, as in `Context: 7 files`.
    
The total line count is an unnecessary detail even for today's context sources. It is not clear what users should do with that information. As context sources start to get less about "chunks of files", it becomes even less relevant.

## Test plan

CI

### DIFF
--- a/lib/shared/src/index.ts
+++ b/lib/shared/src/index.ts
@@ -1,7 +1,7 @@
 // Add anything else here that needs to be used outside of this library.
 
 export { ModelProvider } from './models'
-export { type ChatModel, type EditModel, ModelUsage, ModelContextWindow } from './models/types'
+export { type ChatModel, type EditModel, ModelUsage, type ModelContextWindow } from './models/types'
 export { getDotComDefaultModels } from './models/dotcom'
 export {
     getProviderName,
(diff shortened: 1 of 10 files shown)


sqs:
Ahh, from https://github.com/sourcegraph/cody/pull/3883#discussion_r1573465405 I think this might be needed to give an accurate *total* line count.

sqs:
OK, I actually *think* we do not need to show the total line count. It's fine to show `Context: 7 files`, and `Context: 93 lines from 7 files` is unnecessary detail even for today's context sources. As context sources start to get less about "chunks of files", it becomes even less relevant.

I will update this PR. But again, open to feedback @abeatrix @toolmantim!

abeatrix:
> `Context: 93 lines from 7 files` is unnecessary detail even for today's context sources.

+1 on removing. 
I **personally** don't find the total line count helpful compared to the line numbers next to the files in the expanded context list :)

abeatrix (approved):
Still need Tim's feedback, but the code lgtm 😃

sqs:
Cool. I'll merge it now and can always revert if @toolmantim wants. I am just juggling a lot of PRs and this will make my life easier on an easy-to-revert one. :) Thanks!

PR #3887, opened 2024-04-21
fix nested react hook usage and only have 1 editor mutation listener

sqs:
- Fixes a React warning (in the devtools console) `Warning: Do not call Hooks inside useEffect(...), useMemo(...), or other built-in Hooks.` because `useCurrentModel()` was being called within a `useMemo()` hook. This can cause unexpected behavior in React. It is also desirable for the `isTooLarge` property of each option to be recomputed when the model changes because different models have different context window sizes.
- Avoids registering another editor mutation listener each time the component is rendered by wrapping the registration in a `useEffect` hook. I don't *know* if this causes any problems, but it could make things slow or result in weird behavior.

## Test plan

CI, and try adding mentions in the storybook

### DIFF
--- a/vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
+++ b/vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
@@ -96,9 +96,8 @@ export default function MentionsPlugin(): JSX.Element | null {
 
     const results = useChatContextItems(query)
 
-    // biome-ignore lint/correctness/useExhaustiveDependencies: runs effect when `results` changes.
+    const model = useCurrentChatModel()
     const options = useMemo(() => {
-        const model = useCurrentChatModel()
         const limit =
             model?.contextWindow?.context?.user ||
             model?.contextWindow?.input ||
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):

vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
```diff
@@ -115,23 +114,26 @@ export default function MentionsPlugin(): JSX.Element | null {
...
+        const unregister = editor.registerMutationListener(ContextItemMentionNode, node => {
+            const items = toSerializedPromptEditorValue(editor)?.contextItems
+            if (!items?.length) {
+                setTokenAdded(0)
+                return
+            }
+            setTokenAdded(items?.reduce((acc, item) => acc + (item.size ? item.size : 0), 0) ?? 0)
+        })
+        return unregister
+    }, [editor])
```
abeatrix, line 136:
```suggestion
    }, [])
```

This effect will get re-rendered whenever users add a new @-mention (which changes the editor state) if `editor` is listed as a dependency. If we only want to register it once, does removing `editor` work?

abeatrix (approved):
Left a question inline. Thanks for catching this!

vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
sqs: It seems to work. Is it true that `editor` changes when the editor state changes? I would have thought the `editor` var remains the same, but the object value might change deeply.

abeatrix: > Is it true that editor changes when the editor state changes?

I added a console log within the useEffect, and it was called when the editor state changes, but I can't seem to reproduce it on the latest main, so I must have used the wrong branch when I was testing this... Sorry for the false alarm.

PR #3888, opened 2024-04-21
clearer notice when typing a line range b/c file too large

sqs:
Clears up user confusion "Why do I need to enter a line range?"

Changes the notice for when the user Tab-completes an @-mentioned file in chat. It is /possible/ that they intentionally entered line range selection mode on a not-too-large file, and in which case this notice is technically incorrect, but that's an edge case and arguably it means /they/ considered the file too large.

Not sure if I missed any considerations here. Open to feedback!

Before:

image

After (excuse the different theme :):

image

## Test plan

CI

### DIFF
--- a/vscode/src/chat/context/constants.ts
+++ b/vscode/src/chat/context/constants.ts
@@ -4,7 +4,7 @@ export const SYMBOL_HELP_LABEL = 'Search for a symbol to include...'
 export const NO_SYMBOL_MATCHES_LABEL = 'No symbols found'
 export const NO_FILE_MATCHES_LABEL = 'No files found'
 export const NO_SYMBOL_MATCHES_HELP_LABEL = ' (language extensions may be loading)'
-export const FILE_RANGE_TOOLTIP_LABEL = 'Type a line range to include, e.g. 5-10...'
+export const FILE_RANGE_TOOLTIP_LABEL = 'File too large. Type a line range like 5-10...'
 export const LARGE_FILE_WARNING_LABEL =
     'File too large. Add line range with : or use @# to choose a symbol'
 


toolmantim (commented):
I don't think this is necessary, given the screen before this says it's too large:

CleanShot 2024-04-22 at 12 01 21@2x

vscode/src/chat/context/constants.ts
```diff
@@ -4,7 +4,7 @@ export const SYMBOL_HELP_LABEL = 'Search for a symbol to include...'
 export const NO_SYMBOL_MATCHES_LABEL = 'No symbols found'
 export const NO_FILE_MATCHES_LABEL = 'No files found'
 export const NO_SYMBOL_MATCHES_HELP_LABEL = ' (language extensions may be loading)'
-export const FILE_RANGE_TOOLTIP_LABEL = 'Type a line range to include, e.g. 5-10...'
+export const FILE_RANGE_TOOLTIP_LABEL = 'File too large. Type a line range like 5-10...'
 export const LARGE_FILE_WARNING_LABEL =
     'File too large. Add line range with : or use @# to choose a symbol'
```
toolmantim, line 9: wdyt of removing `@# to choose a symbol`?

```suggestion
    'File too large. Type : to select a subset of lines from the file.'
```

 I don't think it's reasonable to ask them to hit backspace and do a global symbol search. The original label was "Type # to choose a symbol" and it would then show only symbols from the file.

vscode/src/chat/context/constants.ts
```diff
@@ -4,7 +4,7 @@ export const SYMBOL_HELP_LABEL = 'Search for a symbol to include...'
 export const NO_SYMBOL_MATCHES_LABEL = 'No symbols found'
 export const NO_FILE_MATCHES_LABEL = 'No files found'
 export const NO_SYMBOL_MATCHES_HELP_LABEL = ' (language extensions may be loading)'
-export const FILE_RANGE_TOOLTIP_LABEL = 'Type a line range to include, e.g. 5-10...'
+export const FILE_RANGE_TOOLTIP_LABEL = 'File too large. Type a line range like 5-10...'
```
toolmantim, line 7: We should save ourselves the bug report when someone does this on a small file :D and make 'File too large' conditionally prepended…

abeatrix (approved):

vscode/src/chat/context/constants.ts
sqs: Yeah, but for now that's a lot more complex and requires threading through a lot more state into the at-mention UI.

sqs:
> I don't think this is necessary, given the screen before this says it's too large:
> 
> CleanShot 2024-04-22 at 12 01 21@2x

Here is where it confused someone: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1713715587831389.

Just one data point though.

vscode/src/chat/context/constants.ts
sqs: This is before they've necessarily typed much, though. I don't think it's unreasonable. Neither is great, though.

PR #3889, opened 2024-04-21
import instead of require for web-tree-sitter

sqs:
This bundles to the same code and is more standard.

## Test plan

CI

### DIFF
--- a/vscode/src/tree-sitter/parser.ts
+++ b/vscode/src/tree-sitter/parser.ts
@@ -1,14 +1,19 @@
 import path from 'node:path'
 
 import * as vscode from 'vscode'
-import type Parser from 'web-tree-sitter'
 
 import { wrapInActiveSpan } from '@sourcegraph/cody-shared'
-import type { Tree } from 'web-tree-sitter'
+import Parser, { type Tree } from 'web-tree-sitter'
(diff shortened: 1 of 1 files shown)


sqs:
cc @olafurpg in case you know why it was being require()d

burmudar (approved):

sqs:
OK, unit tests failing, so I now see why it was done this way. Will find the right fix.

olafurpg:
I haven't worked directly on this part of the codebase so I'm not sure why it's using `require`. Most likely, `web-tree-sitter` performs side effects during initialization that we want to delay. I'm using `require` in the agent entrypoint for similar reasons https://sourcegraph.com/github.com/sourcegraph/cody@d65d7c1a4af11e2c5c4e62b698be81ec236b1573/-/blob/agent/src/index.ts?L14

sqs:
Pretty sure it's a CJS/ESM issue.

sqs:
I found the fix. I am not 100% sure about the root cause, but my suspicion is that vitest's bundling code somehow lets the `Module = Object.assign({}, ...` line at https://sourcegraph.com/github.com/tree-sitter/tree-sitter/-/blob/lib/binding_web/prefix.js reassign the `__vite_ssr_import_N__` var or something.

PR #3890, opened 2024-04-21
make our shim URI more like VS Code's URI

sqs:
The [VS Code URI class](https://sourcegraph.com/github.com/microsoft/vscode/-/blob/src/vs/base/common/uri.ts) has public properties not getters for `scheme`, `authority`, `path`, `query`, and `fragment`. This means that those properties are serialized as JSON object fields when the URI instance is passed via MessagePort. This change makes our shim behave similarly so that it can be transferred via MessagePort (such as to agent in a web worker).

## Test plan

CI

### DIFF
--- a/vscode/src/testutils/uri.ts
+++ b/vscode/src/testutils/uri.ts
@@ -49,7 +49,7 @@ export class Uri {
         return new Uri(URI.from(components))
     }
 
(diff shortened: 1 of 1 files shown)


philipp-spiess (approved):
Nice!

PR #3892, opened 2024-04-22
VS Code: clarify ignore/exclude everything constants

valerybugakov:
## Context

- Clarify ignore/exclude everything constants and related helper methods
- Built on top of https://github.com/sourcegraph/cody/pull/3856
- Follow-up for review comments in https://github.com/sourcegraph/cody/pull/3869

## Test plan

CI

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.ts
@@ -10,8 +10,8 @@ import { wrapInActiveSpan } from '../tracing'
 export const REFETCH_INTERVAL = 60 * 60 * 1000 // 1 hour
 
 interface ParsedContextFilters {
-    include: ParsedContextFilterItem[]
-    exclude: ParsedContextFilterItem[]
+    include: null | ParsedContextFilterItem[]
+    exclude: null | ParsedContextFilterItem[]
 }
 
 interface ParsedContextFilterItem {
@@ -49,8 +49,8 @@ export class ContextFiltersProvider implements vscode.Disposable {
             if (response) {
                 logDebug('ContextFiltersProvider', 'fetchContextFilters', { verbose: response })
                 this.contextFilters = {
-                    include: (response.include || []).map(parseContextFilterItem),
-                    exclude: (response.exclude || []).map(parseContextFilterItem),
+                    include: response.include?.map(parseContextFilterItem) || null,
+                    exclude: response.exclude?.map(parseContextFilterItem) || null,
                 }
             }
         } catch (error) {
@@ -77,7 +77,7 @@ export class ContextFiltersProvider implements vscode.Disposable {
         // If we don't have any context filters, we exclude everything.
         let isIgnored = this.contextFilters === null
 
-        if (this.contextFilters?.include.length) {
+        if (this.contextFilters?.include?.length) {
             for (const parsedFilter of this.contextFilters.include) {
                 isIgnored = !matchesContextFilter(parsedFilter, repoName)
                 if (!isIgnored) {
@@ -86,7 +86,7 @@ export class ContextFiltersProvider implements vscode.Disposable {
             }
         }
 
-        if (!isIgnored && this.contextFilters?.exclude.length) {
+        if (!isIgnored && this.contextFilters?.exclude?.length) {
             for (const parsedFilter of this.contextFilters.exclude) {
                 if (matchesContextFilter(parsedFilter, repoName)) {
                     isIgnored = true
@@ -100,11 +100,11 @@ export class ContextFiltersProvider implements vscode.Disposable {
     }
 
     public async isUriIgnored(uri: vscode.Uri): Promise<boolean> {
-        if (this.hasIncludeEverythingFilters()) {
+        if (this.hasAllowEverythingFilters()) {
             return false
         }
 
-        if (this.hasExcludeEverythingFilters()) {
+        if (this.hasIgnoreEverythingFilters()) {
             return true
         }
 
@@ -129,15 +129,19 @@ export class ContextFiltersProvider implements vscode.Disposable {
         }
     }
 
-    private hasIncludeEverythingFilters() {
-        return this.contextFilters?.include.length === 0 && this.contextFilters?.exclude.length === 0
+    private hasAllowEverythingFilters() {
+        return (
+            this.contextFilters?.exclude === null &&
+            this.contextFilters?.include?.length === 1 &&
+            this.contextFilters.include[0].repoNamePattern.toString() === '/.*/u'
+        )
     }
 
-    private hasExcludeEverythingFilters() {
+    private hasIgnoreEverythingFilters() {
         return (
-            this.contextFilters?.include.length === 0 &&
-            this.contextFilters?.exclude.length === 1 &&
-            this.contextFilters.exclude[0].repoNamePattern.toString() === '.*'
+            this.contextFilters?.include === null &&
+            this.contextFilters?.exclude?.length === 1 &&
+            this.contextFilters.exclude[0].repoNamePattern.toString() === '/.*/u'
         )
     }
 }
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -197,8 +197,8 @@ export interface ContextFiltersResponse {
 }
 
 export interface ContextFilters {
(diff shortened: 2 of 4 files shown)


lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -58,6 +62,8 @@ export class ContextFiltersProvider implements vscode.Disposable {
                 logError('ContextFiltersProvider', 'fetchContextFilters', error)
                 return
             }
+
+            throw error
```
taras-yemets, line 66: In what case do we expect `error` that is not instance of `Error`?

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -130,15 +136,25 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
     }
 
+    /**
+     * Used to return early from `isUriAllowed` in case context filters are empty.
+     */
     private hasIncludeEverythingFilters() {
-        return this.contextFilters?.include.length === 0 && this.contextFilters?.exclude.length === 0
+        return (
+            this.contextFilters?.include.length === INCLUDE_EVERYTHING_CONTEXT_FILTERS.include.length &&
+            this.contextFilters?.exclude.length === INCLUDE_EVERYTHING_CONTEXT_FILTERS.include.length
```
taras-yemets, line 145: I understand that we expect both `include` and `exclude` to be empty arrays, but this definition seems a bit confusing to me (especially the part where we compare `exclude` with `include`).

taras-yemets (commented):
Using `INCLUDE_EVERYTHING_CONTEXT_FILTERS` and `EXCLUDE_EVERYTHING_CONTEXT_FILTERS` approach seems a bit confusing to me.

I think a simpler and more readable approach would be:
1. Make [`graphqlClient.contextFilters`](https://github.com/sourcegraph/cody/blob/00592623b8a93bc60e5c3d6587a4fc0e1b57f265/lib/shared/src/sourcegraph-api/graphql/client.ts#L555) return
- `ContextFilters` value received from the server (UPD: value can be `null` too if context filters are not defined in the site config)
- `null` in case of outdated API error
- `error` in case of an unexpected error
2. [`isUriAllowed`](https://github.com/sourcegraph/cody/blob/00592623b8a93bc60e5c3d6587a4fc0e1b57f265/lib/shared/src/cody-ignore/context-filters-provider.ts#L135) checks `this.contextFilters` value and:
- returns `true` if `this.contextFilters` is `null`
- returns `false` if `this.contextFilters` is an error (UPD: it can be either an error from `graphqlClient.contextFilters` or parsing result error, e.g. invalid regex)
- checks if repo name is allowed if `this.contextFilters` value is of `ContextFilters` type. 

WDYT?

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: This is a typo. Fixing it now

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: Updated to always log errors 👍

philipp-spiess:
Agree w/ @taras-yemets, I also found this a bit confusing especially the comparison that just happens to work because the array has no elements (but there's no guarantee that any of the empty sets will stay empty 😬).

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -202,15 +202,22 @@ export interface CodyContextFilterItem {
     filePathPatterns?: string[]
 }
 
-const INCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFilters = {
+/**
+ * Default value used on the client in case context filters are not set.
```
dominiccooney, line 206: I'm still a bit uneasy about this. This policy can't be produced by the backend, because it requires include/exclude, if present, to have at least one value.

Wouldn't it reflect the RFC more closely if we made these types something like `null | readonly [T, ...T[]]` where `T` is the rule type?

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -130,15 +131,25 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
     private hasIncludeEverythingFilters() {
-        return this.contextFilters?.include.length === 0 && this.contextFilters?.exclude.length === 0
+        return (
+            this.contextFilters?.include.length === INCLUDE_EVERYTHING_CONTEXT_FILTERS.include.length &&
+            this.contextFilters?.exclude.length === INCLUDE_EVERYTHING_CONTEXT_FILTERS.exclude.length
+        )
     }
 
+    /**
+     * Used to return early from `isUriAllowed` in case client encountered errors fetching context filters.
```
dominiccooney, line 145: Could we keep it simple and say we'll set the filters to be exactly `EXCLUDE_EVERYTHING_CONTEXT_FILTERS` and then make the condition just `this.contextFilters === EXCLUDE_EVERYTHING_CONTEXT_FILTERS`? (Similarly for include everything?)

This is an even faster test than the length checking that happens here, and it is much simpler, too.

Explicit rules which include or exclude everything may fall off the fast path, but that seems fine. We could always elaborate the fast path later if it turns out to be a bottleneck.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -202,15 +202,22 @@ export interface CodyContextFilterItem {
...
+} satisfies ContextFilters
 
-const EXCLUDE_EVERYTHING_CONTEXT_FILTERS: ContextFilters = {
+/**
+ * Default value used on the client in case client encounters errors
+ * fetching or parsing context filters.
+ */
+export const EXCLUDE_EVERYTHING_CONTEXT_FILTERS = {
     include: [],
     exclude: [{ repoNamePattern: '.*' }],
```
dominiccooney, line 219: You should `Object.freeze` the object literal and arrays here. (You need to do each, freeze is not deep.)

dominiccooney (changes requested):
Some ideas to simplify and harden inline.

lib/shared/src/sourcegraph-api/graphql/client.ts
valerybugakov: The problem I wanted to solve here with virtual values is that the backend spec doesn't match the client states one-to-one. Here are potential `ContextFiltersProvider.contextFilters` values:

- `null,` we didn't fetch context filters yet, that's why we ignore everything (@taras-yemets, here's the difference with your proposed return values)
- `INCLUDE_EVERYTHING_CONTEXT_FILTERS,` we fetched and got `null` or an outdated API response.
- `EXCLUDE_EVERYTHING_CONTEXT_FILTERS`, we fetched and got unexpected errors.
- `ContextFilters` received from the backend with the correct structure

I will make the `INCLUDE_EVERYTHING_CONTEXT_FILTERS` and `EXCLUDE_EVERYTHING_CONTEXT_FILTERS` states compatible with the backend spec. Agree that it contributes to the confusion and it's easy to avoid that.

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: We parse context filters before assigning them to `this.contextFilters` so we do not preserve the references equality.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -197,8 +197,8 @@ export interface ContextFiltersResponse {
 }
 
 export interface ContextFilters {
-    include?: CodyContextFilterItem[]
-    exclude?: CodyContextFilterItem[]
+    include?: null | readonly [CodyContextFilterItem, ...CodyContextFilterItem[]]
+    exclude?: null | readonly [CodyContextFilterItem, ...CodyContextFilterItem[]]
 }
```
valerybugakov, line 202: Updated types to match the backend spec. Now shortcut constants are backend-spec-compliant.

dominiccooney (changes requested):
Ok sounds good. 

I wanted immutability because these "constants" hang around indefinitely and if people mutate them we will be in for a bad time. But I can do it later if it is too hard right now.

dominiccooney (approved):
Really approving this time, whoops

PR #3895, opened 2024-04-22
Cody Ignore: support multiple remote URLs

valerybugakov:
## Context

- Uses all remote URLs available to resolve repo names for Cody Ignore checks
- Built on top of https://github.com/sourcegraph/cody/pull/3870

## Test plan

CI and updated unit tests.

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.ts
@@ -19,7 +19,7 @@ interface ParsedContextFilterItem {
     filePathPatterns?: RE2[]
 }
 
-export type GetRepoNameFromWorkspaceUri = (uri: vscode.Uri) => Promise<string | null>
+export type GetRepoNamesFromWorkspaceUri = (uri: vscode.Uri) => Promise<string[] | null>
 type RepoName = string
 type IsRepoNameIgnored = boolean
 
@@ -30,11 +30,11 @@ export class ContextFiltersProvider implements vscode.Disposable {
      */
     private contextFilters: ParsedContextFilters | null = null
     private cache = new LRUCache<RepoName, IsRepoNameIgnored>({ max: 128 })
-    private getRepoNameFromWorkspaceUri: GetRepoNameFromWorkspaceUri | undefined = undefined
+    private getRepoNamesFromWorkspaceUri: GetRepoNamesFromWorkspaceUri | undefined = undefined
     private fetchIntervalId: NodeJS.Timeout | undefined | number
 
-    async init(getRepoNameFromWorkspaceUri: GetRepoNameFromWorkspaceUri) {
-        this.getRepoNameFromWorkspaceUri = getRepoNameFromWorkspaceUri
+    async init(getRepoNamesFromWorkspaceUri: GetRepoNamesFromWorkspaceUri) {
+        this.getRepoNamesFromWorkspaceUri = getRepoNamesFromWorkspaceUri
         this.dispose()
         await this.fetchContextFilters()
         this.startRefetchTimer()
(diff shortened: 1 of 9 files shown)


PriNova:
Thank you for the quick reply @valerybugakov 
In my case (as a PLG user), a simple http request `@http://www.example.com` will be blocked with the error : "The prompt string contains a reference to a file that is not allowed by the context filters."

philipp-spiess:
@PriNova Interesting! @valerybugakov @dominiccooney Could this be related to non-file URIs being blocked by default? What's the expected behavior for this? Should we allow http/https URLs?

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -114,11 +114,11 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
         }
 
-        const repoName = await wrapInActiveSpan('repoNameResolver.getRepoNameFromWorkspaceUri', () =>
-            this.getRepoNameFromWorkspaceUri?.(uri)
+        const repoNames = await wrapInActiveSpan('repoNameResolver.getRepoNamesFromWorkspaceUri', () =>
+            this.getRepoNamesFromWorkspaceUri?.(uri)
         )
 
-        return repoName ? this.isRepoNameIgnored(repoName) : true
+        return repoNames ? repoNames.some(repoName => this.isRepoNameIgnored(repoName)) : true
```
valerybugakov, line 121: @aramaraju @dominiccooney, I have another point to add to the client spec: here, we ignore URIs if one of the found repo names is ignored. This means that if the repo has multiple remote URLs locally and one of them is ignored, Cody will ignore the repo as a context source.

vscode/src/repository/git-extension-api.ts
```diff
@@ -77,17 +77,32 @@ export async function gitAPIinit(): Promise<vscode.Disposable> {
...
+            repository?.state.remotes[0]?.pushUrl || repository?.state.remotes[0]?.fetchUrl
+
+        if (remoteOriginUrl) {
+            return convertGitCloneURLToCodebaseName(remoteOriginUrl) || undefined
         }
     } catch (error) {
         logDebug('repositoryHelper:getCodebaseFromWorkspaceUri', 'error', { verbose: error })
     }
     return undefined
 }
```
philipp-spiess, line 91: This function is for backward compatibility only I assume? Since it only looks at the first remote/

philipp-spiess (approved):

vscode/src/repository/git-extension-api.ts
valerybugakov: Yes!

valerybugakov:
I will go ahead with the merge here because I already have changes on top of this branch. I am happy to follow up if Taras and Dominic have suggestions.

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -25,15 +25,15 @@ export class RepoNameResolver {
...
      * Checks if the Git API is initialized, initializes it if not.
-     * If found, gets the codebase name from the repository.
-     * If not found, attempts to use Git CLI to get the codebase name (in node.js environment only).
+     * If found, gets repo names from the repository.
      * if not found, walks the file system upwards until it finds a `.git` folder.
+     * If not found, attempts to use Git CLI to get the repo names (in node.js environment only).
      * If not found, returns `undefined`.
      */
-    public async getRepoNameFromWorkspaceUri(uri: vscode.Uri): Promise<string | null> {
+    public async getRepoNamesFromWorkspaceUri(uri: vscode.Uri): Promise<string[] | null> {
```
philipp-spiess, line 36: If we have an array here I think we can remove null and use empty arrays instead (to avoid having to check for both).

lib/shared/src/cody-ignore/context-filters-provider.ts
aramaraju: Thanks for surfacing this, I'm okay with this approach. I'm also dumping these cases into a doc, so that we can capture these in the right docs/content for the GTM teams and customers :)

PR #3898, opened 2024-04-22
Invert ContextFiltersProvider.prototype.isUriAllowed logic

philipp-spiess:
`ContextFiltersProvider.prototype.isUriAllowed()` returning a Promise makes an argument for inverting the logic. Consider this code:

```ts
if (!contextFiltersProvider.isUriAllowed(foo)) {
  return null
}
```

Which is what most of the call sites of this function will look like. The fact that Promise is a truthy value will make it very easy for us to forget to await the result. In the past we had lint rules which notified us about issues like this but since the move to biome we don’t have access to type-aware linter rules. To avoid issues, I suggest we invert the logic here instead:

```ts
if (contextFiltersProvider.isUriBlocked(foo)) { // <- Will always be true since a Promise instance is truthy. 
  return null
}
```

## Test plan

- Updated unit tests

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.test.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.test.ts
@@ -310,7 +310,7 @@ describe('ContextFiltersProvider', () => {
         })
     })
 
-    describe('isUriAllowed', () => {
+    describe('isUriBlocked', () => {
         interface TestUriParams {
             repoName: string
             filePath: string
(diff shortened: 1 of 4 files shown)


dominiccooney (approved):
Nice idea.

I guess by the same token, we should change isRepoNameAllowed => isRepoNameBlocked.

WDYT about using "Ignored" instead of "Blocked" to avoid introducing another term for the same thing?

valerybugakov (commented):
Nice, thanks @philipp-spiess!

PR #3902, opened 2024-04-22
fix: Handle HTTP URIs and decode URI components in URL context mentions

PriNova:
This PR introduces three changes:

1. **Decode URI components for display in URL context mentions**
   - URLs mentioned using the '@' syntax were displaying encoded characters like '%3D' instead of '='.
   - This commit fixes the issue by decoding the URI components when displaying file paths and directories for URL context mention items in `ContextItemMentionNode.ts` and `OptionsList.tsx`.
   - This ensures that users see properly decoded URLs when using the '@' mention feature to reference URLs in their prompts.

2. **Skip ignoring HTTP URIs**
   - The `IgnoreHelper` class now checks if the URI scheme is 'http' and returns false (i.e., does not ignore the URI) in that case.
   - This change allows the code to process HTTP URIs, which were previously being ignored along with other non-file URIs.
   - The updated logic in the `shouldIgnore` method ensures that HTTP URIs are handled correctly by the application, potentially enabling new functionality or fixing an issue related to HTTP resource handling.

These changes improve the handling of HTTP URIs and enhance the user experience when working with URL context mentions in the application.

Before:
Screenshot 2024-04-22 182828

After:
Screenshot 2024-04-22 183519

## Test plan

Manually tested

### DIFF
--- a/lib/shared/src/cody-ignore/ignore-helper.ts
+++ b/lib/shared/src/cody-ignore/ignore-helper.ts
@@ -116,6 +116,10 @@ export class IgnoreHelper {
             return false
         }
 
+        if (uri.scheme === 'http') {
+            return false
+        }
+
         // Ignore all other non-file URIs
         if (uri.scheme !== 'file') {
             return true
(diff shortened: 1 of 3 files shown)


PriNova:
Actually, http fetching does not work because the context filter blocks mention items not part of a repo uri.

philipp-spiess:
Heya! Sorry for only getting to this now. Do you mind rebasing on top of https://github.com/sourcegraph/cody/pull/3934 that should fix the context-block related issues you were seeing.

PriNova:
Still some conflicts. I need to rebase on #3942

PriNova:
Now I rebased to #3942 and it works perfectly, no error and the typelookahead displays correctly.

philipp-spiess (approved):
LGTM! Thank you so much 😊

PR #3907, opened 2024-04-23
test and support node 18 (no longer node 16)

sqs:
My recommendation is that we should stop testing and supporting Node 16, and instead bump this to Node 18.

We had been supporting Node 16 because the July 2023 VS Code release shipped with a version of Electron that used Node 16. That is now well in the past, and:

- VS Code's Aug 2023 release shipped with Node 18 (up from Node 16): https://code.visualstudio.com/updates/v1_82#_engineering
- Node 16 was end-of-life'd on 2023-09-11: https://nodejs.org/en/blog/announcements/nodejs16-eol
- For JetBrains, we use the agent binary that bundles Node (20), so we don't depend on Node 16 even if that's what is installed on the system.

An alternative is to start testing Node 16, 18, AND 20 (ie add tests for Node 18 since it's the version in use by VS Code). This would be safer in case any customers are still using an older version of VS Code that requires Node 16. I'll confirm this before merging: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1713845081115149.

## Test plan

CI

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "url": "https://github.com/sourcegraph/cody"
   },
   "engines": {
-    "node": "^v20.4.0",
+    "node": ">=18",
     "pnpm": "^8.6.7"
   },
   "browser": {
@@ -72,9 +72,6 @@
         }
       }
     },
-    "neverBuiltDependencies": [
-      "deasync",
-      "playwright"
-    ]
+    "neverBuiltDependencies": ["deasync", "playwright"]
   }
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
           # need to run this on all platforms as we're only verifying we don't
           # call any APIs not available in this version.
           - runner: ubuntu
-            node: 16 # Supported by VS Code 1.81 (July 2023).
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -9,7 +9,7 @@
     "url": "https://github.com/sourcegraph/cody"
   },
   "engines": {
-    "node": "^v20.4.0",
+    "node": "^18",
```
valerybugakov, line 12: Should it be `>=18` to allow the latest version to be used for development without warnings?

Screenshot 2024-04-23 at 13 51 54

valerybugakov (approved):
Sounds good if we confirm that we do not have customers who are still using an older version of VS Code 👍

package.json
sqs: thanks, good catch

olafurpg (approved):
Yes please! 🙏🏻 I've seen our team spend a lot of time keeping the Node v16 CI green, and we have added many `ignoreIf(node16())` so I'm not even sure how well Cody works in Node v16 anyways.

philipp-spiess (approved):
YES!

abeatrix (approved):
🎉

sqs:
@aramaraju I'm going to merge this optimistically, but let me know if you see any evidence that would make us want to continue supporting Node 16.

PR #3908, opened 2024-04-23
shared: port markdown tests for renderCodyMarkdown

keegancsmith:
This copy pastes the markdown tests we have from common/markdown, but instead runs it through the renderCodyMarkdown. There wasn't any specific tests here, but thought it would be useful to actually assert against the same logic actually used. IE things like the dompurify config are important to test.

The way this was done is I removed some cases which didn't make sense anymore, and in others I updated the expected output so it passes. These cases may not make much sense, but is a good starting point.

To see how the tests change, you can view the diff at https://gist.github.com/keegancsmith/296e2b0da33fbe2cc520a747568bf1e0. This diff renames renderCodyMarkdown to renderMarkdown to make actual changes clearer.

Test Plan: pnpm test


philipp-spiess (approved):

abeatrix (approved):
Nice!

PR #3910, opened 2024-04-23
Fix repo name resolver

philipp-spiess:
This PR fixes a few things regarding the repo name resolver:

- It will now fall back to the previous URL parsing if the repo is not known in the Sourcegraph instance (for backward compat): https://github.com/sourcegraph/cody/pull/3870/files#diff-282c58c06ab5f90a1cda9f3adae0bd0541bf9a3c4f681d0399037a568cc8c605L61-R64
- I remove the nullability since we moved this to an array (an empty array is the same as null now)
- Fixed some type issues that creeped into `main`
- Fixed an issue where the upstream request was not sent that I discovered when adding unit tests 😬 

## Test plan

Added a unit test

### DIFF
--- a/vscode/src/repository/repo-name-resolver.ts
+++ b/vscode/src/repository/repo-name-resolver.ts
@@ -50,7 +50,7 @@ export class RepoNameResolver {
         try {
             let remoteUrls = gitRemoteUrlsFromGitExtension(uri)
 
-            if (remoteUrls?.length === 0) {
+            if (remoteUrls === undefined || remoteUrls.length === 0) {
                 remoteUrls = await gitRemoteUrlsFromTreeWalk(uri)
             }
 
--- a/vscode/src/repository/repo-name-resolver.ts
+++ b/vscode/src/repository/repo-name-resolver.ts
@@ -78,6 +78,7 @@ export class RepoNameResolver {
                 return definedRepoNames
             }
         } catch (error) {
+            console.error(error)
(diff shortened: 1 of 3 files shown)


vscode/src/repository/repo-name-resolver.ts
```diff
@@ -49,7 +54,7 @@ export class RepoNameResolver {
                 remoteUrls = await gitRemoteUrlsFromTreeWalk(uri)
             }
 
-            if (remoteUrls?.length === 0) {
+            if (remoteUrls === undefined || remoteUrls.length === 0) {
```
philipp-spiess, line 57: The unit test caught this one 😅  I suppose we will use `gitRemoteUrlsFromGitExtension and `gitRemoteUrlsFromTreeWalk` for most cases but this still seems like it should work?

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -76,43 +81,26 @@ export class RepoNameResolver {
...
         return repoNameRequest
     }
 }
 
-// Required to store null values in the cache.
-// See https://github.com/isaacs/node-lru-cache#storing-undefined-values
-const nullCacheValue = Symbol('null cache value')
-type UriFsPath = string
-
-class LRUCacheWithNullValues {
```
philipp-spiess, line 99: Feels bad to remove this since you just added it 😬

umpox (approved):

abeatrix (approved):
ty!

valerybugakov:
How did the naming typo pass our CI checks? 🤔

PR #3914, opened 2024-04-23
Enable Ollama Chat by default

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713891011703859

This pull request updates the Cody extension for VS Code to enable the experimental Ollama Chat feature by default. 

The changes include:

- Configuration Update: In the configuration.ts file, the experimentalOllamaChat setting is changed from false to true by default. This setting controls whether the Ollama Chat feature is enabled or not.
- Status Bar Update: In the StatusBar.ts file, a new feature toggle is added for the "Ollama Chat" feature. This toggle appears in the Cody extension's status bar and allows users to enable or disable the Ollama Chat feature from the UI.
- Dropdown menu Update: Display Ollama models in the dropdown menu with `experimental` badge.

With these changes, users will now have the Ollama Chat feature enabled by default, allowing them to use their local Ollama models for chat and commands within VS Code. However, users can still disable this experimental feature through the extension's JSON settings or the new status bar toggle if desired.

## Test plan

Manual test plan:
1. Start your local ollama app
2. Start Cody from this branch in debug mode and verify the following:

Ollama models in the dropdown has an `experimental` badge next to them:

image

You can toggle the experimental settings via Cody Settings Menu:

image

### DIFF
--- a/vscode/doc/ollama.md
+++ b/vscode/doc/ollama.md
@@ -32,7 +32,7 @@
 3. Select a chat model (model that includes instruct or chat, e.g. [gemma:7b-instruct-q4_K_M](https://ollama.com/library/gemma:7b-instruct-q4_K_M)) from the [Ollama Library](https://ollama.com/library)
 4. Pull the chat model locally (e.g. `ollama pull gemma:7b-instruct-q4_K_M`)
 5. Once the chat model is downloaded successfully, open Cody in VS Code
-6. Enable the `cody.experimental.ollamaChat` configuration
+6. Select `Preferences: Open Settings (JSON)` from the Command Palette and add the `"cody.experimental.ollamaChat": true` configuration to enable Cody's experimental chat support for Ollama.
 7. Open a new Cody chat
 8. In the new chat panel, you should see the chat model you've pulled in the dropdown list
 9. Currently, you will need to restart VS Code to see the new models
--- a/vscode/doc/ollama.md
+++ b/vscode/doc/ollama.md
@@ -32,9 +32,8 @@
 3. Select a chat model (model that includes instruct or chat, e.g. [gemma:7b-instruct-q4_K_M](https://ollama.com/library/gemma:7b-instruct-q4_K_M)) from the [Ollama Library](https://ollama.com/library)
 4. Pull the chat model locally (e.g. `ollama pull gemma:7b-instruct-q4_K_M`)
 5. Once the chat model is downloaded successfully, open Cody in VS Code
-6. Select `Preferences: Open Settings (JSON)` from the Command Palette and add the `"cody.experimental.ollamaChat": true` configuration to enable Cody's experimental chat support for Ollama.
-7. Open a new Cody chat
-8. In the new chat panel, you should see the chat model you've pulled in the dropdown list
-9. Currently, you will need to restart VS Code to see the new models
+6. Open a new Cody chat
+7. In the new chat panel, you should see the chat model you've pulled in the dropdown list
+8. Currently, you will need to restart VS Code to see the new models
 
-Note: You can run `curl http://localhost:11434/api/tags` in your terminal to see what models are currently available on your machine
+Note: You can run `ollama list` in your terminal to see what Ollama models are currently available on your machine
--- a/vscode/src/services/StatusBar.ts
+++ b/vscode/src/services/StatusBar.ts
@@ -197,6 +197,14 @@ export function createStatusBar(): CodyStatusBar {
                 c => c.experimentalSymfContext,
                 false
             ),
+            await createFeatureToggle(
+                'Ollama Chat',
+                'Experimental',
+                'Enable Chat and Commands to use your local Ollama models when available',
+                'cody.experimental.ollamaChat',
+                c => c.experimentalOllamaChat,
+                false
+            ),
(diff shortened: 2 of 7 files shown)


vscode/src/services/StatusBar.ts
```diff
@@ -197,6 +197,14 @@ export function createStatusBar(): CodyStatusBar {
                 c => c.experimentalSymfContext,
                 false
             ),
+            await createFeatureToggle(
```
abeatrix, line 200: we can remove this option if we want this setting to stay in the settings JSON

vscode/doc/ollama.md
```diff
@@ -32,7 +32,7 @@
 3. Select a chat model (model that includes instruct or chat, e.g. [gemma:7b-instruct-q4_K_M](https://ollama.com/library/gemma:7b-instruct-q4_K_M)) from the [Ollama Library](https://ollama.com/library)
 4. Pull the chat model locally (e.g. `ollama pull gemma:7b-instruct-q4_K_M`)
 5. Once the chat model is downloaded successfully, open Cody in VS Code
-6. Enable the `cody.experimental.ollamaChat` configuration
+6. Select `Preferences: Open Settings (JSON)` from the Command Palette and add the `"cody.experimental.ollamaChat": true` configuration to enable Cody's experimental chat support for Ollama.
```
sqs, line 35: This is no longer needed because it's on by default, right?

vscode/src/services/StatusBar.ts
```diff
@@ -197,6 +197,14 @@ export function createStatusBar(): CodyStatusBar {
...
+                'Ollama for Chat',
+                'Experimental',
+                'Enable using your local Ollama models for chat and commands when available',
```
sqs, line 203:
```suggestion
                'Use local Ollama models for chat and commands when available',
```

sqs (approved):
Awesome! Code LGTM. I also love this change. @chillatom any reasons not to do this?

vscode/webviews/Components/ChatModelDropdownMenu.module.css
```diff
@@ -76,19 +76,29 @@
...
     margin-left: auto;
-    padding: 3px 4px;
-    background: repeating-conic-gradient(#7048E8 50%, #00CBEC 75%, #A112FF 100%, #FF5543 125%, #7048E8 150%);
-    border-radius: 4px;
     font-size: 12px;
     font-weight: 500;
     line-height: 1;
     letter-spacing: -0.025rem;
     text-shadow: 0px 0px 1px #280065, 0px 1px 1px rgba(36, 0, 81, 1);
     transition: box-shadow 150ms;
```
toolmantim, line 87: We want to move these to `.experimental-badge` too

toolmantim (commented):
Nice! Can't wait to ship better Ollama!

A few small things:
* Given it's defaulted to on, I don't think we want it in the status bar menu but instead people can just turn off Ollama support in the extension settings.
* Some CSS suggestions

abeatrix:
> Nice! Can't wait to ship better Ollama!
> 
> 
> 
> A few small things:
> 
> * Given it's defaulted to on, I don't think we want it in the status bar menu but instead people can just turn off Ollama support in the extension settings.
> 
> * In the markdown for the extension setting, can we link to somewhere in our docs that explains the Ollama support?

Thanks @toolmantim! 

Because this is under experimental, our users won't be able to find this configuration in their extension settings when they are on stable releases, that's why i thought moving it to the extension settings would help since a lot of users weren't able to locate the setting: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713891799495099?thread_ts=1713891011.703859&channel=C05AGQYD528&message_ts=1713891799.495099

I think we can remove it from the status bar once this is moved out of experimental, or when we have implemented your new design with the setting in the dropdown, Wdyt?

abeatrix:
@toolmantim shipping this PR as per discussed on slack. 

let me know if there are any changes on the UI you would like me to make in a follow-up 🙇‍♀️

toolmantim:
Bummer! Do we have an `cody.unstable` scope for things we want to be exposed to the users?

toolmantim:
I guess my big concern is that lots of people will see it in the statusbar (of which a small % would use it), and have to google "Cody ollama" or just "ollama" to figure out what it is, how to use it, etc. At the moment on Google they'd land on some out-of-date Sourcegraph blog posts… whereas keeping it in settings means we can write words, link to docs, etc.

That was the thinking!

abeatrix:
> I guess my big concern is that lots of people will see it in the statusbar (of which a small % would use it), and have to google "Cody ollama" or just "ollama" to figure out what it is, how to use it, etc. At the moment on Google they'd land on some out-of-date Sourcegraph blog posts… whereas keeping it in settings means we can write words, link to docs, etc.
> 
> That was the thinking!

@toolmantim what do you think about adding a doc icon on the right here?
image

toolmantim:
That would be great!

toolmantim:
Can we log metrics for both the toggle and the docs button?

abeatrix:
> Can we log metrics for both the toggle and the docs button?

Yup! PR here https://github.com/sourcegraph/cody/pull/3932

PR #3922, opened 2024-04-24
Cody Ignore: do not use locally resolved repo names

valerybugakov:

### DIFF
--- a/vscode/src/repository/repo-name-resolver.test.ts
+++ b/vscode/src/repository/repo-name-resolver.test.ts
@@ -1,18 +1,19 @@
+import path from 'node:path'
 import { describe, expect, it, vi } from 'vitest'
 import * as vscode from 'vscode'
 import { URI } from 'vscode-uri'
 
-import { graphqlClient } from '@sourcegraph/cody-shared'
 import dedent from 'dedent'
-import { RepoNameResolver, gitRemoteUrlsFromTreeWalk } from './repo-name-resolver'
+import { gitRemoteUrlsFromTreeWalk } from './repo-name-resolver'
 
 interface MockFsCallsParams {
     filePath: string
-    gitConfig?: string
-    gitRepoPath?: string
+    gitConfig: string
+    gitRepoPath: string
     gitSubmodule?: {
         path: string
         gitFile: string
+        gitConfig: string
     }
 }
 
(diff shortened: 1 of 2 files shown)


vscode/src/repository/repo-name-resolver.ts
```diff
@@ -90,12 +85,7 @@ export class RepoNameResolver {
             return this.remoteUrlToRepoNameCache.get(remoteUrl)!
         }
 
-        const repoNameRequest = graphqlClient.getRepoName(remoteUrl).then(repoName => {
-            if (repoName === null) {
-                return convertGitCloneURLToCodebaseName(remoteUrl)
-            }
-            return repoName
-        })
+        const repoNameRequest = graphqlClient.getRepoName(remoteUrl)
```
valerybugakov, line 88: Discussion ref: https://github.com/sourcegraph/cody/pull/3870/files#r1576408488

cc @taras-yemets @philipp-spiess

philipp-spiess: Nothing in the repo name resolver API suggests that it’s only an API for enterprise users. I suggest we either change the name to make this clear or add a fallback PLG implementation. It is not super clear otherwise.

vscode/src/repository/repo-name-resolver.test.ts
```diff
@@ -202,43 +234,23 @@ describe('gitRemoteUrlFromTreeWalk', () => {
...
+                    fetch = +refs/heads/*:refs/remotes/origin/*
             `,
         })
 
         const remoteUrls = await gitRemoteUrlsFromTreeWalk(fileUri)
         expect(remoteUrls).toBe(undefined)
     })
 })
-
-describe('getRepoNamesFromWorkspaceUri', () => {
```
philipp-spiess, line 218: Do you not find high level API tests valuable here? There _was_ an error that I spotted because of it (where only array length was checked and not the absence of said array)

vscode/src/repository/repo-name-resolver.test.ts
```diff
@@ -23,14 +24,29 @@ function deWindowsifyPath(path: string): string {
 function mockFsCalls(params: MockFsCallsParams) {
     const { gitConfig, gitRepoPath, filePath, gitSubmodule } = params
 
+    const files = {
+        [filePath]: '',
+        [`${gitRepoPath}/.git/config`]: gitConfig,
+    }
+
+    if (gitSubmodule) {
```
philipp-spiess, line 32: These changes seem unrelated but I didn't find it mentioned in the PR description. Mind explaining why they're necessary?

vscode/src/repository/repo-name-resolver.ts
```diff
@@ -16,11 +11,12 @@ import { gitRemoteUrlsFromGitExtension } from './git-extension-api'
 export type RemoteUrlGetter = (uri: vscode.Uri) => Promise<string[] | undefined>
 type RepoName = string
 type RemoteUrl = string
+type UriFsPath = string
 
 export class RepoNameResolver {
     private platformSpecificGitRemoteGetters: RemoteUrlGetter[] = []
-    private fsPathToRepoNameCache = new LRUCache<RepoName, string[]>({ max: 1000 })
-    private remoteUrlToRepoNameCache = new LRUCache<RemoteUrl, Promise<string | null>>({ max: 1000 })
+    private fsPathToRepoNameCache = new LRUCache<UriFsPath, RepoName[]>({ max: 1000 })
```
philipp-spiess, line 18: whoopsie.

vscode/src/repository/repo-name-resolver.ts
philipp-spiess: I suggest we either call this the `enterprise-repo-name-resolver` or (IMO better) make a check inside that does not allow this being used by PLG users.

valerybugakov: Yep, let's do it!

vscode/src/repository/repo-name-resolver.test.ts
valerybugakov: I noticed that you removed a submodule argument from one of the tests in your PR, which didn't cause the test to fail. This change fixes this.

vscode/src/repository/repo-name-resolver.test.ts
valerybugakov: You're right; the test covers one of the cases not covered in the existing tests. Restored this one 👍

philipp-spiess (approved):

PR #3925, opened 2024-04-24
noodle: introduce sourcegraph literal search mentions provider

keegancsmith:
Note: this logic is currently hidden behind the noodle feature flag. Additionally instead of implementing client logic in our graphql backend, it keeps all related logic inside one file to contain the blast radius. Please give feedback if this is a reasonable idea, or if I should just implement things inside of client.ts. Because I didn't do the latter I had to copy paste some utility types/functions.

When typing the prefix "@src:" we will provide a context item which says it will search against your sourcegraph endpoint. This is presented as a URL to the search. When clicking resolving the item it will convert each search result chunk into a context item.

This is currently not working due to some recent restrictions introduced on context sources. Note that the URL mention provider also seems to be broken due to it:

  Request Failed: The prompt string contains a reference to a file that
  is not allowed by the context filters.

Other known issues:
- We can't use the triggerPrefix ! or ? due to how mentions are parsed
- Mentions also stop completing once you enter non file path looking items. EG entering a ( stops the autocomplete.
- The generated link to the search results is not clickable due to vscode incorrectly escaping the query parameters.

Test Plan: manually tested, but not yet ready for merging

### DIFF
--- a/lib/shared/src/mentions/providers/sourcegraphSearch.ts
+++ b/lib/shared/src/mentions/providers/sourcegraphSearch.ts
@@ -1,0 +1,176 @@
+import { URI } from 'vscode-uri'
+import { ContextItemSource, type ContextItemWithContent } from '../../codebase-context/messages'
+import { isErrorLike } from '../../common'
+import type { RangeData } from '../../common/range'
+import { graphqlClient } from '../../sourcegraph-api/graphql'
+import { isError } from '../../utils'
+import type { ContextMentionProvider } from '../api'
+
+export const SOURCEGRAPH_SEARCH_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'src-search'> = {
+    id: 'src-search',
+    // TODO the prefix '!' seems to not trigger the @ mention code path. I am
+    // assuming there is parsing logic somewhere which only calls the @
+    // mentions if they look like a path or start with # (for symbols)
+    triggerPrefixes: ['src:'],
+
+    async queryContextItems(query, signal) {
+        const searchQuery = query.startsWith('src:') ? query.slice(4) : query
+        const uri = URI.parse(graphqlClient.endpoint).with({
+            query: 'q=' + encodeURIComponent(searchQuery),
+        })
+        return [
+            {
+                type: 'file',
+                uri: uri,
+                title: searchQuery,
+                source: ContextItemSource.Uri,
+                provider: 'src-search',
+            },
+        ]
+    },
+
+    async resolveContextItem(item, input, signal): Promise<ContextItemWithContent[]> {
+        if (item.content !== undefined) {
+            return [item as ContextItemWithContent]
+        }
+
+        // Sneaking in the search query via the title
+        const rawQuery = item.title
+        if (!rawQuery) {
+            return []
+        }
+
+        // Adjust query to limit results to filematches
+        const query = 'type:file count:10 ' + rawQuery
+
+        const chunks = await searchForFileChunks(query, signal)
+
+        if (isErrorLike(chunks)) {
+            throw chunks
+        }
+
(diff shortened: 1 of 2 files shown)


keegancsmith:
@sqs @thenamankumar while implementing this I realised a neat idea could be smart ways to translate a URL into context. For example the provider for generates a sourcegraph search URL. It would be cool if we could just paste in that url then cody was smart enough to pull out structured data from it? Seems like a nice generic way to support other remote datasources.

PriNova:
Nice idea @keegancsmith 
If you like, you can build against this PR #3902, this includes a fix for the PromptString error message and prettifies the URL displaying if http attributes are given in the URL.

sqs (approved):
This is awesome. Code LGTM. As mentioned by @PriNova, if you rebase against main, the response error message and unit test error should be fixed.

lib/shared/src/mentions/providers/sourcegraphSearch.ts
```diff
@@ -0,0 +1,177 @@
...
+import { ContextItemSource, type ContextItemWithContent } from '../../codebase-context/messages'
+import { isErrorLike } from '../../common'
+import type { RangeData } from '../../common/range'
+import { graphqlClient } from '../../sourcegraph-api/graphql'
+import { isError } from '../../utils'
+import type { ContextMentionProvider } from '../api'
+
+export const SOURCEGRAPH_SEARCH_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'src-search'> = {
+    id: 'src-search',
+    // TODO the prefix '!' seems to not trigger the @ mention code path. I am
```
thenamankumar, line 11: You should be able to modify this regex to support `!` prefix. 
CleanShot 2024-04-25 at 13 26 48@2x

lib/shared/src/mentions/providers/sourcegraphSearch.ts
```diff
@@ -0,0 +1,177 @@
...
+    },
+}
+
+type Chunk = Pick<ContextItemWithContent, 'uri' | 'range' | 'content' | 'repoName' | 'revision'>
+
+async function searchForFileChunks(
+    query: string,
+    signal: AbortSignal | undefined
+): Promise<Chunk[] | Error> {
+    const results = await graphqlClient
```
thenamankumar, line 68: IMO, we should define it as a function on the `SourcegraphGraphqlAPIClient` itself. And move the SEARCH_QUERY and types to the graphql client only. 

This will allow us to re-use it easily at other places as well. It also keeps all the queries we make to the backend in 1 single place.

lib/shared/src/mentions/providers/sourcegraphSearch.ts
keegancsmith: There is a fix I can add, but it makes the final regex we construct a bit more ugly so I'd rather do it as its own PR for proper feedback. (just removing a char from punctuation then breaks ending your mention with punctuation). So will follow-up.

lib/shared/src/mentions/providers/sourcegraphSearch.ts
keegancsmith: I got started doing this, but I went with this root because I thought the actual query we run is so tightly coupled with the implementation I don't want reuse / avoids mingling code with non-noodle code. But obviously this deviates from the common pattern I see here. But to be honest I don't know if the pattern makes sense :)

Just in case you missed this I explicitly called out doing it this way in my PR description. I'm gonna land as is, but will quickly follow-up if you think I should still move it all into backend.ts.

keegancsmith:
Just did a final test on this after rebasing, sorry for not landing yesterday! Still quite limited, but lots of future potential :)

image

image

PR #3929, opened 2024-04-24
Chat: deduplicate chat context

abeatrix:
CLOSE https://github.com/sourcegraph/cody-issues/issues/24

This PR introduces several improvements to the prompt builder to optimize token usage, prioritize user-added context, and simplify the codebase. The changes address issues with redundant context items that cause inefficient token allocation.

|              Before                         | After |
|:-------------------------------------:|:-------------------------------------:|
|  |  |
|  |  |

### Key improvements

The changes listed below aim to improve the prompt quality and token usage efficiency. The new `getUniqueContextItems` function is used to ensures that only valid and non-overlapping context items are included in the prompt, saving valuable context window space for the most relevant information.

#### Deduplicate context items

The `getUniqueContextItems` function has been introduced to handle deduplicating context items based on their content, ranges, and sources. Uniqueness is determined based on the following criteria:
 - Items added by the user (@-mention and selection) are not removed as they are parts of the user's input
 - An (context) item must not be an exact duplicate of any existing item.
 - An item (from the same file) with a specified range should not overlap with any existing item's range on the same lines.
 - If the new item is within the range of an existing item that is not user-added, the existing item is removed.

This ensures that only valid contexts with non-overlapping ranges are turned into context messages before sending them to the LLM. By prioritizing user-added context, deduplicating context items, and optimizing token allocation

#### Simplified logging statement in DefaultPrompter

The previous version of the code included a complex logging statement that summarized the ignored context items categorized by source. This is now replaced with more granular logging statements that provide clearer and more actionable information about the context items info during the prompt-building step.

#### Refactor PromptBuilder Context Processing

Improve context handling and message processing of the `tryAddContext` method in the `PromptBuilder` class.  It removes the handling of duplicate items and specific token types directly within the method, relying instead on utility functions like isUniqueContextItem to check for uniqueness.

Currently, the handling and generation of context messages were done on-the-fly within the `tryAddContext` method. This approach involved directly processing each context item, rendering it into a message, and immediately deciding whether to include it based on token limits and other conditions. This method tightly coupled the decision-making process regarding token limits and item validity with the generation and rendering of context messages. This limited flexibility because once a context item was processed and potentially discarded or altered, it couldn't be revisited or modified easily in subsequent operations.

Context items are now collected and stored without immediately transforming them into chat messages. The actual conversion of context items into chat messages has been deferred until the `build()` method is called. This separation allows for more flexible handling of these items throughout their lifecycle within the prompt building process, and the deferred approach allows us to modify the context items anytime before the list is finalized, enabling us to remove duplicate items or unwanted items during the context-adding process. 

Here is a comparison of the before and after from the agent test when enhanced context is enabled:
See gist for details: https://gist.github.com/abeatrix/dfd0c029d7480ea42dd6406bd216a51e
````diff
@@ -4,40 +4,19 @@
                 text: You are Cody, an AI coding assistant from Sourcegraph.
               - speaker: human
                 text: |-
-                  My selected typescript code from file `src/animal.ts`:
-                  <selected>
+                  Codebase context from file src/TestClass.ts:
+                  ```typescript
+                  const foo = 42

-                  export interface Animal {
-                      name: string
-                      makeAnimalSound(): string
-                      isMammal: boolean
-                  }
+                  export class TestClass {
+                      constructor(private shouldGreet: boolean) {}

-                  </selected>
-              - speaker: assistant
-                text: Ok.
-              - speaker: human
-                text: |-
-                  Codebase context from file src/animal.ts:
-                  ```typescript
-                  /* SELECTION_START */
-                  export interface Animal {
-                      name: string
-                      makeAnimalSound(): string
-                      isMammal: boolean
+                      public functionName() {
+                          if (this.shouldGreet) {
+                              console.log(/* CURSOR */ 'Hello World!')
+                          }
+                      }
                   }
-                  /* SELECTION_END */
-                  ```
-              - speaker: assistant
-                text: Ok.
-              - speaker: human
-                text: |-
-                  Codebase context from file src/animal.ts:
-                  ```typescript
-                  export interface Animal {
-                      name: string
-                      makeAnimalSound(): string
-                      isMammal: boolean
                   ```
               - speaker: assistant
                 text: Ok.
@@ -58,70 +37,30 @@
                 text: Ok.
               - speaker: human
                 text: |-
-                  Codebase context from file src/squirrel.ts:
-                  ```typescript
-                  export interface Squirrel {}
-                  ```
-              - speaker: assistant
-                text: Ok.
-              - speaker: human
-                text: |-
-                  Codebase context from file src/TestClass.ts:
-                  ```typescript
-                      constructor(private shouldGreet: boolean) {}
-                  ```
-              - speaker: assistant
-                text: Ok.
-              - speaker: human
-                text: |-
-                  Codebase context from file src/TestClass.ts:
-                  ```typescript
-                      public functionName() {
-                          if (this.shouldGreet) {
-                              console.log(/* CURSOR */ 'Hello World!')
-                          }
-                  ```
-              - speaker: assistant
-                text: Ok.
-              - speaker: human
-                text: |-
-                  Codebase context from file src/TestClass.ts:
+                  Codebase context from file src/animal.ts:
                   ```typescript
-                  const foo = 42
+                  /* SELECTION_START */
+                  export interface Animal {
+                      name: string
+                      makeAnimalSound(): string
+                      isMammal: boolean
+                  }
+                  /* SELECTION_END */
                   ```
               - speaker: assistant
                 text: Ok.
               - speaker: human
                 text: |-
-                  Codebase context from file src/TestClass.ts:
-                  ```typescript
-                  const foo = 42
-
-                  export class TestClass {
-                      constructor(private shouldGreet: boolean) {}
+                  My selected typescript code from file `src/animal.ts`:
+                  <selected>

-                      public functionName() {
-                          if (this.shouldGreet) {
-                              console.log(/* CURSOR */ 'Hello World!')
-                          }
-                      }
+                  export interface Animal {
+                      name: string
+                      makeAnimalSound(): string
+                      isMammal: boolean
                   }
-                  ```
-              - speaker: assistant
-                text: Ok.
-              - speaker: human
-                text: |-
-                  Codebase context from file src/TestClass.ts:
-                  ```typescript
-                  export class TestClass {
-                      constructor(private shouldGreet: boolean) {}

-                      public functionName() {
-                          if (this.shouldGreet) {
-                              console.log(/* CURSOR */ 'Hello World!')
-                          }
-                      }
-                  ```
+                  </selected>
               - speaker: assistant
                 text: Ok.
               - speaker: human
````

## Test plan

1. Log in as a DotCom user
2. Use @-mention to add a whole file as context
3. Enable enhanced context
4. Ask Cody a question about the @-mentioned file
5. Check the context file list to verify file added via @-mentions was not added again by enhanced context
6. Re-run steps 1-4 but as an enterprise user

Screenshot 2024-04-24 at 11 34 38 AM

## Example 1

Question: What does @vscode/src/chat/chat-view/ChatPanelsManager.ts do? How does it work with ChatManager?

### After

ChatPanelsManager is added via @-mention
No snippets from ChatPanelsManager was added after

image

### Before

ChatPanelsManager is added via @-mention
Enhanced context also added snippets from the same files after the whole file was added

image

### After

image

### Before

Screenshot 2024-05-03 at 4 36 30 PM

### After

image

### Before

image

### CODE REVIEW

File: vscode/src/prompt-builder/context-tracker.ts [hunk 1]
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Generates a unique identifier for a context item based on its source and content.
+     */
+    public getID(item: ContextItem): ContextTrackerID {
+        if (item.source === 'terminal' || item.source === 'uri') {
+            return `${displayPath(item.uri)}#${SHA256(item.content ?? '').toString()}`
+        }
+
+        return (item.source === 'unified' && item.title) || displayPath(item.uri)
```
sqs, line 89: Does this need to special-case `item.source` values `terminal`, `uri`, and `unified`? Can it just always return a string like `[displayPath(item.uri), SHA256(item.content ?? ''), item.title].join(':')`? Just to simplify.

vscode/src/prompt-builder/context-tracker.ts
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+/**
+ * Checks if the first range is fully contained within the second range.
+ */
+function isRangeContain(r1: RangeData, r2: RangeData): boolean {
```
sqs, line 96: - This is not strictly correct because it does not look at the characters. For example, if `r1` and `r2` are all on the same line, it always returns true, even if `r1` spans from (say) character 1 to 100, and `r2` is character 50 to 55. We only care about the behavior for lines, not ranges, so I'd rename the function to something like `rangeContainsLines`.
- It seems more conventional for the bigger range to be the 1st argument because the verb "contain" works that way (eg in the sentence `r1 contains r2`, r1 is the bigger range).

vscode/src/prompt-builder/context-tracker.test.ts
```diff
@@ -0,0 +1,281 @@
...
+                uri: URI.file('/foo/bar'),
+                source: ContextItemSource.Terminal,
+            }
+            const item2: ContextItem = {
+                type: 'file',
+                uri: URI.file('/foo/baz'),
+                source: ContextItemSource.Uri,
+            }
+            expect(tracker.getID(item1)).toBe(
+                `${displayPath(item1.uri)}#${SHA256(item1.content ?? '').toString()}`
```
sqs, line 250: This ID should be considered as opaque (ie the actual value doesn't matter or carry meaning to any callers, only properties of the value, such as that it uniquely identifies an item). Therefore I think you should only test those external properties of the value, such as that the values not identical (`expect(tracker.getID(item1)).not.toBe(tracker.getID(item2))`). This is just to avoid having the unit test be too brittle and to avoid reimplementing part of the actual implementation in the unit test.

vscode/src/prompt-builder/context-tracker.ts
```diff
@@ -0,0 +1,98 @@
...
+     * A temporary list to store the context items being tracked in the current iteration.
+     */
+    private current: ContextItem[] = []
+
+    /**
+     * Retrieves the context items tracked in the current iteration and resets the current state.
+     */
+    public get getAndResetTrackedItems(): ContextItem[] {
+        const tracked = this.current
+        this.current = [] // Reset the current state for the next iteration
```
sqs, line 22: A getter should not mutate state. If you need to mutate state, make it a method.

vscode/src/prompt-builder/context-tracker.ts
```diff
@@ -0,0 +1,98 @@
+import { type ContextItem, type RangeData, displayPath } from '@sourcegraph/cody-shared'
+import { SHA256 } from 'crypto-js'
+
+type ContextTrackerID = string
+
+export class ContextTracker {
```
sqs, line 6: To confirm my understanding, the reason why the `track` method returns true or false is to let the caller know if `item` has any *new* context that is not already tracked, right? This lets us avoid sending context (such as a subset of an existing snippet that's been sent over) that has already been sent.

I think we can map this to some existing concepts better, which would help people understand how this works:

- This sounds like a "context set" to me, so I'd rename it `ContextSet`.
- Use the standard naming for methods on a set, like `add`, `remove`, and `values`.
- Remove the `getID` public method (it does not seem to be used externally).
- Can we remove the ID concept altogether? Can it just scan through the items each time and compare the URI? The list will never be very large, and removing the ID concept would simplify the code.
- The `current` field worries me because it seems like it is adding the concept of a "transaction" ("iteration") into a set. I would recommend making the instance only live for a single "transaction", and have a function or method that makes a copy of the value to be used in a subsequent transaction.

vscode/src/prompt-builder/context-tracker.ts
```diff
@@ -0,0 +1,98 @@
...
+        this.history.set(id, items)
+        this.current.push(item)
+        return true
+    }
+
+    /**
+     * Removes a context item from the tracked items.
+     */
+    public untrack(contextItem: ContextItem): void {
+        const id = this.getID(contextItem)
```
sqs, line 71: If you call `untrack` on a subset of a context item, should it just remove that subset?

For example if you are tracking lines 50-100 of foo.txt, and you call `untrack` on lines 80-100 of foo.txt, shouldn't lines 50-79 still be tracked?

sqs (dismissed):

sqs:
This looks like a great addition and will let us be a lot more sophisticated about context tracking! Thanks. Added comments.

Also, are the har agent test recording changes due to the simple wording change in a prompt? If so, can you separate out that wording change so this diff is just the ContextTracker? (If the diff is also due to the change in behavior from the better ContextTracker, disregard.)

vscode/src/prompt-builder/context-tracker.ts
sqs: Also when you have 2 args like this, I find it makes it clear to name them haystack and needle, like `rangeContainsLines(haystack: RangeData, needle: RangeData)`. Go's source code uses this. Your call.

vscode/src/prompt-builder/context-tracker.ts
abeatrix: This only "untrack" item that is currently being "tracked" in the current list.

> For example if you are tracking lines 50-100 of foo.txt, and you call untrack on lines 80-100 of foo.txt, shouldn't lines 50-79 still be tracked?

If we are tracking lines 50-100 of foo.txt, then trying to "track" 80-100 of foo.txt would fail (return false) since we don't track subsets, so it shouldn't make it to the "untrack" step (if it works correctly 😆).

vscode/src/prompt-builder/context-tracker.ts
abeatrix: Thanks for taking the time to review this so throughly, really appreciated it! I'll address your feedback on Monday but want to write down some notes before that while everything is still fresh 😀 

The context tracker starts tracking when we start turning context into message during the prompt building process, in the following order where 1 has the highest priority:
1. Add user-added context
2. Add previously-added context
3. Add new enhanced context

In each step, we will add the context item to the tracker through the "track" method if it's not an subset of added context, and will "untrack" it if the file has exceeded the context window.

At the end of each step, we add the tracked items to the transcript before we move on to the next step. This is why we are keeping track of what have been added in the previous step and the current step inside the tracker.

> Can we remove the ID concept altogether? Can it just scan through the items each time and compare the URI? The list will never be very large, and removing the ID concept would simplify the code.

We can! My thinking of using getID as helper here is because comparing just the URI doesnt work for all context items. For example:
- we will need to use the title of a remote context in order to compare them with local file context uri
- all terminal output items have the same uri
- uri of an uri context could be the same but with different context when adding them from old chat history?

This is why I thought sorting them by "ID" would make it easier than looping through the context item to determine which group of file path the context item belong to, but I might have over complicated things here. Let me try it on Monday and report back, thanks for the idea!

> The current field worries me because it seems like it is adding the concept of a "transaction" ("iteration") into a set. I would recommend making the instance only live for a single "transaction", and have a function or method that makes a copy of the value to be used in a subsequent transaction.

So instead of initializing it in the PromptBuilder class, I'll move it to the tryAddContext method, and initialize it with the context items that have been added before (if any). At the end of the method we will store the tracked context back to promptBuilder? 
I think that does make more sense, yea. Will update on Monday.

Thanks again for all the thorough review and helpful info & tips 🥹

abeatrix:
> Also, are the har agent test recording changes due to the simple wording change in a prompt? If so, can you separate out that wording change so this diff is just the ContextTracker? (If the diff is also due to the change in behavior from the better ContextTracker, disregard.)

No changes in the prompt, just changes in used context due to removal of subsets. 

Based on the diff on the match snippets in the agent test file I don't see any regression in response due to this change, but will keep an eye on it 👀

vscode/src/prompt-builder/context-tracker.ts
abeatrix: updated and remove the ID concept

abeatrix:
Example from the latest commit:

image

#### Before

image

abeatrix:
@sqs I've updated the code based on your feedback 😄 
- [x] This sounds like a "context set" to me, so I'd rename it ContextSet.
  - Renamed "ContextTracker" to "ContextSet"
- [x] Use the standard naming for methods on a set, like add, remove, and values.
  - Replaced "track" with "add". "untrack" with "remove", and getter method to "values"
- [x] Remove the getID public method (it does not seem to be used externally).
  - removed getID method
- [x] Can we remove the ID concept altogether? Can it just scan through the items each time and compare the URI? The list will never be very large, and removing the ID concept would simplify the code.
  - removed ID concept
- [x] The current field worries me because it seems like it is adding the concept of a "transaction" ("iteration") into a set. I would recommend making the instance only live for a single "transaction", and have a function or method that makes a copy of the value to be used in a subsequent transaction.
  - updated to make the instance only live for a single iteration, with option to start new set with previously added context

toolmantim:
This is great!

abeatrix:
### Demo from the latest commit:

#### After

image

#### Before

image

vscode/src/chat/chat-view/prompt.test.ts
```diff
@@ -72,38 +74,7 @@ describe('DefaultPrompter', () => {
                 text: ps`Hello`,
             },
         ])
-        expect(newContextUsed).toEqual([])
-    })
-
-    it('tryAddContext limit should not allow prompt to exceed overall limit', async () => {
```
abeatrix, line 78: moved to vscode/src/prompt-builder/index.test.ts

vscode/src/prompt-builder/unique-context.ts
```diff
@@ -0,0 +1,106 @@
...
+ * the provided list of existing `ContextItem`.
+ *
+ * NOTE: The transcript is reversed during the prompt-building process to ensure
+ * that the most recent items are considered first. Therefre, the `reversedItems`
+ * parameter should be in reverse order.
+ *
+ * @param reversedItems - The list of ContextItem to filter for uniqueness.
+ * @returns A new array of unique ContextItem instances.
+ */
+export function getUniqueContextItems(reversedItems: ContextItem[]): ContextItem[] {
```
philipp-spiess, line 19: Hm I think the fact that dies does not try to optimize for the largest ranges can cause some unexpected issues. imagine we have two context items and try to insert them in this order:


```
foo.ts:2-3
foo.ts:1-30
```

Intuitively I think we should include the largest common denominator here (to avoid regressing the prompt quality) but if I understand this correctly, we would only add the first one and then the next item would be skipped. Is that correct?

abeatrix: https://github.com/sourcegraph/cody/blob/af4d0b5fff4416f2c0dd87a2c382ab3485780cf7/vscode/src/prompt-builder/unique-context.ts#L82-L86

abeatrix: this might be helpful too: https://github.com/sourcegraph/cody/blob/af4d0b5fff4416f2c0dd87a2c382ab3485780cf7/vscode/src/prompt-builder/unique-context.test.ts#L266-L286

philipp-spiess: Oh I didn't realize you mutate the `uniqueItems` array inside the `isUniqueContextItem` function. I think we should clean this up and make it clearer when this happens. It's a great test case though!!

philipp-spiess (commented):
Sorry I haven't come around to doing a full review yet but I think we need to clean up the mutation inside `isUniqueContextItem`, this makes it really hard to reason about the implementation otherwise (e.g. it seems like we mutate `is.contextItems` inside the prompt builder now but the implementation is different to your test cases where if `isUniqueContextItem` returns `true` you always add the context item. In the prompt builder it might be that `isUniqueContextItem` removes an item but then the new context item is not included because it is too large or so and that could cause none of the items to be included 🤔

vscode/src/prompt-builder/unique-context.ts
```diff
@@ -0,0 +1,106 @@
...
+        // unless there is also an item without a range a full file (duplicate).
+        if (item === itemToAdd || !item.range) {
+            return false // Duplicate found.
+        }
+
+        // Continue looping to ensure the item of full file is unique.
+        if (!itemToAddRange) {
+            // Since the itemToAdd will be added to the unique list at the end,
+            // but we will keep looping to ensure all remaing items are removed.
+            canRemoveItem(item) && uniqueItems.splice(i, 1)
```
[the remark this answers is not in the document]
abeatrix: good catch, i've separated out the removal logic !

vscode/src/prompt-builder/index.ts
```diff
@@ -79,79 +85,84 @@ export class PromptBuilder {
...
             }
-            this.reverseMessages.push(assistantMsg, contextMsg)
-            result.used.push(userContextItem)
+
+            // Add the new context item to the list,
+            this.contextItems.push(newContextItem)
+
+            // Remove any items that are no longer unique.
+            // TODO (bee) update token usage with the new context items.
+            this.contextItems = getUniqueContextItems(this.contextItems)
```
abeatrix, line 163: @philipp-spiess moved the steps to remove non-unique items to end of this function to reset this.contextItems

philipp-spiess (approved):
Nice! This looks solid now. Thank you so much 🙇

vscode/src/prompt-builder/index.ts
```diff
@@ -79,79 +87,80 @@ export class PromptBuilder {
...
         }
-        this.processedContextType.add(tokenType)
-        // Create a new array to avoid modifying the original array, then reverse it to process the newest context items first.
+
+        // Create a new array to avoid modifying the original array,
+        // then reverse it to process the newest context items first.
         const reversedContextItems = contextMessages.slice().reverse()
+
+        // Required by agent tests to ensure the context items are sorted correctly.
+        sortContextItems(reversedContextItems as ContextItem[])
```
keegancsmith, line 104: @abeatrix is reversedContextItems still the correct name here given we are going to always sort them?

abeatrix: @keegancsmith `sortContextItems` is for sorting the items in agent test due to issues with recordings. We don't sort them in production!

PR #3930, opened 2024-04-24
Trace local embedding search error

rafax:
Trace embedding search errors - currently, we only capture the number of results.

## Test plan

- local testing

### DIFF
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -580,7 +580,7 @@ export class LocalEmbeddingsController
         if (!lastRepo || !lastRepo.repoName) {
             return []
         }
-        // without this, assingment to repoName below complains about | boolean
+        // without this, assignment to repoName below complains about | boolean
         const lastRepoName = lastRepo.repoName
         return wrapInActiveSpan('LocalEmbeddingsController.query', async span => {
             try {
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -578,13 +578,14 @@ export class LocalEmbeddingsController
         }
         return wrapInActiveSpan('LocalEmbeddingsController.query', async span => {
             try {
-                if (!this.lastRepo || !this.lastRepo.repoName) {
+                const lastRepo = this.lastRepo
+                if (!lastRepo || !lastRepo.repoName) {
                     span.setAttribute('noResultReason', 'last-repo-not-set')
                     return []
                 }
                 const service = await this.getService()
                 const resp = await service.request('embeddings/query', {
-                    repoName: this.lastRepo.repoName,
+                    repoName: lastRepo.repoName,
                     query: query.toString(),
                     numResults,
                 })
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -579,7 +579,7 @@ export class LocalEmbeddingsController
         return wrapInActiveSpan('LocalEmbeddingsController.query', async span => {
             try {
                 const lastRepo = this.lastRepo
-                if (!lastRepo || !lastRepo.repoName) {
+                if (!lastRepo?.repoName) {
                     span.setAttribute('noResultReason', 'last-repo-not-set')
                     return []
                 }
--- a/vscode/src/local-context/local-embeddings.ts
+++ b/vscode/src/local-context/local-embeddings.ts
@@ -576,17 +576,15 @@ export class LocalEmbeddingsController
         if (!this.endpointIsDotcom) {
             return []
         }
-        const lastRepo = this.lastRepo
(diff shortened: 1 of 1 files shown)


vscode/src/local-context/local-embeddings.ts
```diff
@@ -578,21 +580,30 @@ export class LocalEmbeddingsController
...
-            return resp.results.map(result => ({
-                ...result,
-                uri: vscode.Uri.joinPath(lastRepo.dir, result.fileName),
-            }))
-        } catch (error) {
-            logDebug('LocalEmbeddingsController', 'query', captureException(error), error)
-            return []
-        }
+        // without this, assingment to repoName below complains about | boolean
+        const lastRepoName = lastRepo.repoName
```
rafax, line 584: Question for a TS expert - why does this work with a const, but trying to assign `repoName: lastRepo.repoName` on line 589 complains that:
```
Type 'string | false' is not assignable to type 'string'.\n  Type 'boolean' is not assignable to type 'string'.
```
Is it because the "narrowed" type is not propagated to the nested function?

abeatrix: > Is it because the "narrowed" type is not propagated to lambda passed to wrapInActiveSpan?

Not sure what that means but I guess something like that? If you move the `if (!lastRepo?.repoName)` check to inside the try-catch block, you should be able to do `repoName: lastRepo.repoName,` on line 589

rafax: So, my understanding of how [Narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) / type-guards is that:

1.  TS compiler can use the if on line 580 to verify that `lastRepo.repoName` is not false, so it must be a string
2. If I assing lastRepo.repoName to a const (of type `string`) on line 584, everything is fine
3. If I try to use lastRepo.repoName on line 589 (inside the anonymous function), the compiler "forgets" the information from line 584 and goes back to thinking that the type of lastRepo.repoName is `string | false` (even though it's the same thing as on line 584)

rafax: But, your suggestion makes sense, moved the if inside the lambda.

vscode/src/local-context/local-embeddings.ts
```diff
@@ -574,25 +576,32 @@ export class LocalEmbeddingsController
...
-                ...result,
-                uri: vscode.Uri.joinPath(lastRepo.dir, result.fileName),
-            }))
-        } catch (error) {
-            logDebug('LocalEmbeddingsController', 'query', captureException(error), error)
-            return []
-        }
+        return wrapInActiveSpan('LocalEmbeddingsController.query', async span => {
+            try {
+                if (!this.lastRepo || !this.lastRepo.repoName) {
```
abeatrix, line 581:
```suggestion
                const lastRepo = this.lastRepo
                if (!lastRepo?.repoName) {
```

vscode/src/local-context/local-embeddings.ts
```diff
@@ -574,25 +576,32 @@ export class LocalEmbeddingsController
...
+                    span.setAttribute('noResultReason', 'last-repo-not-set')
+                    return []
+                }
+                const service = await this.getService()
+                const resp = await service.request('embeddings/query', {
+                    repoName: this.lastRepo.repoName,
```
abeatrix, line 587:
```suggestion
                    repoName: lastRepo.repoName,
```

abeatrix (approved):
left some comments inline that should fix the type issue 🙏

rafax:
> left some comments inline that should fix the type issue 🙏

I fixed those locally but your version is even shorter - thank you!

vscode/src/local-context/local-embeddings.ts
rafax: Resolved manually (I pushed a different change)

vscode/src/local-context/local-embeddings.ts
dominiccooney: > If I try to use lastRepo.repoName on line 589 (inside the anonymous function), the compiler "forgets" the information from line 584 and goes back to thinking that the type of lastRepo.repoName is string | false (even though it's the same thing as on line 584)

@rafax "even though it's the same thing as on line 584" isn't true in TypeScript's conservative control flow analysis. There's local control flow analysis, and it understands scoped bindings (which is why the option 2, the new binding, works), but there's no interprocedural control flow: TypeScript doesn't know whether or how many times the lambda will be called, or *when* it will be called... the callee could capture the lambda and call it later.

The `lastRepo.repoName` property is different to the `const` binding because the object could be modified, etc.

janhartman (approved):

vscode/src/local-context/local-embeddings.ts
rafax: Ok, so I got the "no interprocedural control flow" part right, and the reason const works is that TS knows it is a const (unlike `lastRepo.repoName`). Thank you for the explanation :)

PR #3931, opened 2024-04-24
Add `JetbrainsDocumentEvent` to `ProtocolTextDocument` and derive `contentChanges`

mkondratek:


agent/src/agent.ts
```diff
@@ -395,12 +397,39 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+                            text: document.jetbrainsDocumentEvent.content,
+                        },
+                    ]
+                }
+            }
+
             this.workspace.setActiveTextEditor(textEditor)
             vscode_shim.onDidChangeTextDocument.fire({
                 document: textDocument,
-                contentChanges: [], // TODO: implement this. It was only used by recipes, not autocomplete.
```
mkondratek, line 403: 🚀

agent/src/agent.ts
```diff
@@ -80,6 +81,7 @@ import type {
 } from './protocol-alias'
 import { AgentHandlerTelemetryRecorderProvider } from './telemetry'
 import * as vscode_shim from './vscode-shim'
+import { Uri } from './vscode-shim'
```
dominiccooney, line 84: The rest of this file refers to `vscode.Uri` everywhere, let's stick to that for consistency.

If you want to change to `Uri` everywhere in here, that's fine, but maybe do it in a separate clean up PR so we can keep things simple.

agent/src/agent.ts
```diff
@@ -432,12 +434,39 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
         this.registerNotification('textDocument/didChange', document => {
             const documentWithUri = ProtocolTextDocumentWithUri.fromDocument(document)
+            const oldDocumentContent = this.workspace.getDocument(documentWithUri.uri)?.content
             const textDocument = this.workspace.addDocument(documentWithUri)
             const textEditor = this.workspace.newTextEditor(textDocument)
+
+            let contentChanges: vscode.TextDocumentContentChangeEvent[] = []
+
+            if (document.jetbrainsDocumentEvent && document.content && oldDocumentContent) {
+                if (document.jetbrainsDocumentEvent.offset) {
```
dominiccooney, line 444:
```ts
if (x) {
  if (y) {
    z
  }
}
```

&dArr;

```ts
if (x && y) {
  z
}
```

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -768,3 +769,9 @@ export interface RemoteRepoFetchState {
     state: 'paused' | 'fetching' | 'errored' | 'complete'
     error: CodyError | undefined
 }
+
+export interface JetbrainsDocumentEvent {
```
dominiccooney, line 773: I'm lost. What is this for? Why is it JetBrains specific?

If there's a generalization of this to all clients, great, let's call it that.

If this is always and only JetBrains, I need to know more about why.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -768,3 +769,9 @@ export interface RemoteRepoFetchState {
...
+    offset: number
+    oldLength: number
```
dominiccooney, line 775: It's unclear what this length refers to... after careful reading, I believe it is related to the length of text being replaced (that is, 0 for insertions). Because there are at least four lengths at hand:

- Pre-edit document length
- Post-edit document length
- Edit's replaced text length
- Edit's inserted text length

Let's name this really clearly so any misuse is *painfully obvious* in code review. Something like

```
let documentGotLonger = document.jetbrainsDocumentEvent.oldLength < document.content.length
```

looks fine, but is total nonsense, right?

agent/src/agent.ts
dominiccooney: The VSCode API has an array here for a reason—compound edits—so we should support compound edits. If JetBrains never does compound edits, that's fine, it can pass an array with one value. But we should not hard-code to zero or one.

The comment, implementation, and callers of `updateRangeMultipleChanges` should illuminate how VSCode handles offsets are handled with multiple updates: The ranges refer to the old document, there can not be overlaps, the ranges are sorted (which we don't rely on, but I believe VSCode always does this and other text document change handlers may rely on that fact), etc.

https://sourcegraph.com/github.com/sourcegraph/cody@b9cdd9ab3778059591c34c81891ff5c2a254e2cf/-/blob/vscode/src/non-stop/tracked-range.ts?L23-25

dominiccooney (changes requested):
Let's support multiple edits, even if JetBrains only provides one.

dominiccooney:
Could we port some of [these tests](https://sourcegraph.com/github.com/sourcegraph/cody@b9cdd9ab3778059591c34c81891ff5c2a254e2cf/-/blob/vscode/src/non-stop/tracked-range.test.ts) to run as Agent tests through the document events? We'd need some way to get the tracked range back *out* of the test, but the agent tests can reach around and dig in the extension and vscode shims the way `agent/src/lsp/foldingRanges.test.ts` and `agent/src/vscode-shim.test.ts` do.

agent/src/agent.ts
```diff
@@ -432,12 +434,39 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+                    const start: vscode.Position = oldDocument.positionAt(
+                        document.jetbrainsDocumentEvent?.offset
+                    )
+                    const end: vscode.Position = oldDocument.positionAt(
+                        document.jetbrainsDocumentEvent?.offset +
+                            document.jetbrainsDocumentEvent.oldLength
+                    )
+                    const range = new vscode.Range(start, end)
+
+                    contentChanges = [
```
pkukielka, line 455: If someone would be wondering why we need to include such redundant info:
https://github.com/microsoft/language-server-protocol/issues/9

vscode/src/jsonrpc/agent-protocol.ts
pkukielka: I think it does not have to be.
We could use the same structure as VSC, which is:

```
	/**
	 * An event describing an individual change in the text of a {@link TextDocument document}.
	 */
	export interface TextDocumentContentChangeEvent {
		/**
		 * The range that got replaced.
		 */
		readonly range: Range;
		/**
		 * The offset of the range that got replaced.
		 */
		readonly rangeOffset: number;
		/**
		 * The length of the range that got replaced.
		 */
		readonly rangeLength: number;
		/**
		 * The new text for the range.
		 */
		readonly text: string;
	}
```

Since all client evens need to be translated to it, we can equally good keep the translation on the clients side and just use the same structure on the agent side.

mkondratek: I put some comments. 

This is JetBrains specific. We need to translate jetbrains event to vscode event. Technically it would be possible to derive `contentChanges` on the client side but it's easier to to it here. One reason is that here we have the document state before and after the change (while on the client side we have the state after the change only).

vscode/src/jsonrpc/agent-protocol.ts
mkondratek: I put a comment. `oldLength` is the length of a text that got replaced by a new content.

agent/src/agent.ts
dominiccooney: Useful, let's link to this in a comment?

vscode/src/jsonrpc/agent-protocol.ts
mkondratek: structure removed - closing

agent/src/agent.ts
mkondratek: done

vscode/src/jsonrpc/agent-protocol.ts
mkondratek: good point! done

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/JetbrainsDocumentEvent.kt
```diff
@@ -0,0 +1,9 @@
+@file:Suppress("FunctionName", "ClassName", "unused", "EnumEntryName", "UnusedImport")
```
dominiccooney, line 1: I guess this needs to be removed from this PR now, because the type is gone.

agent/src/agent.ts
```diff
@@ -434,12 +435,38 @@ export class Agent extends MessageHandler implements ExtensionClient {
 
         this.registerNotification('textDocument/didChange', document => {
             const documentWithUri = ProtocolTextDocumentWithUri.fromDocument(document)
+            const oldDocumentContent = this.workspace.getDocument(documentWithUri.uri)?.content
             const textDocument = this.workspace.addDocument(documentWithUri)
             const textEditor = this.workspace.newTextEditor(textDocument)
+
+            const contentChanges: vscode.TextDocumentContentChangeEvent[] = []
+
+            if (document.content && oldDocumentContent && document.textDocumentContentChangeEvents) {
```
dominiccooney, line 444: Last bit of feedback: It's does not seem appropriate to have this on ProtocolTextDocument. For example `textDocument/didOpen` doesn't process these events. So clean this up by making `textDocument/didChange` take an object with two properties—a document and an array of events—or making the type `ProtocoTextDocument & Changes` or something like that.

@olafurpg do you have input on the right approach here?

One thing on my mind, here we have an editor that can pass character offsets but not ranges. I could imagine there are editors which want to pass ranges but not character offsets. So we may want that flexibility to pass different kinds of events as long as you pass *some* kind of edit event (ranges, or offsets, or both) to `textDocument/didChange`.

agent/src/agent.ts
```diff
@@ -434,12 +435,38 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+                    const range = new vscode.Range(start, end)
+
+                    // `rangeLength` seems to be redundant - see: https://github.com/microsoft/language-server-protocol/issues/9
+                    contentChanges.push({
+                        range: range,
+                        rangeOffset: event.rangeOffset,
+                        rangeLength: event.rangeLength,
+                        text: event.text,
+                    })
+                }
```
dominiccooney, line 463: Let's add the validation that the ranges don't overlap and are sorted. You can do that in this single forward pass. I suggest doing it in an additional pass:
- Keeps your range computation above uncomplicated
- If we later support events with ranges, but not offsets, we want to do the same consistency check after they're processed to add offsets, etc.

I know if you're only sending one event, so that the ordering is trivially true for JetBrains. But for the next editor, getting that wrong could be tricky to trigger and debug. This is the earliest place we can detect the problem. Please add agent tests for those cases.

dominiccooney (approved):
Approving this to unblock. Some feedback inline, please take a look.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -542,6 +542,11 @@ export interface ProtocolTextDocument {
     /** @deprecated use `uri` instead. This property only exists for backwards compatibility during the migration period. */
     filePath?: string
     content?: string
+    /** `textDocument/didChange` fires `onDidChangeTextDocument` and we want to include `contentChanges` within.
+     * `contentChanges` can be determined based on the document event. The JetBrains client document event structure
+     * is different from `vscode.TextDocumentChangeEvent` - it misses `range`.
+     * We need derive the range based on the JetBrains event, and it's easier to do it on the agent side. */
+    textDocumentContentChangeEvents?: Omit<vscode.TextDocumentContentChangeEvent, 'range'>[]
```
olafurpg, line 549: I prefer to inline the interface over using `Omit` over VSC types here. It's much simpler to evolve a custom type and reason about it's structure. It become very awkward to use this `Omit` type as soon as we want to add a field, or customize anything.

agent/src/agent.ts
olafurpg: +1 we should mirror the LSP structure with an array.

agent/src/agent.ts
olafurpg: The current `ProtocolTextDocument` already diverges from LSP by having properties like `selection`. The motivation for this design was that I wanted to be able to be able to use a single notification for everything, allowing atomic updates of the document state, which I think helps with debugging and reasoning about the system. Using the same motivation, I think it's OK to add an optional `changes: ProtocolDocumentChangeEvent[]` property.

We should not permit changes by offsets because we currently don't use offsets anywhere else in the protocol. There are good reasons why line/character is better than offset (despite performance overhead):

- Much smaller blast radius if you get UTF-16 encoding wrong from a single emoji character anywhere in the file. With line/character, the offset resets on every line so encoding bugs are only limited to the characters that appear after the emoji character on the same line. In spirit, we should aim for bug-free UTF-16 encodings but it's a tall order for all Cody clients to get this 100% right from the start.
- Much simpler debugging of protocol methods. You can manually look at the JSON of a line/character object and reason about it, while offsets are more or less meaningless.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/JetbrainsDocumentEvent.kt
```diff
@@ -0,0 +1,9 @@
...
+package com.sourcegraph.cody.protocol_generated
+
+data class JetbrainsDocumentEvent(
```
olafurpg, line 4: Is this class a leftover from an old version? Either way, we should not reference jetbrains-specific types in the protocol. The protocol must be client-agnostic.

olafurpg (changes requested):

vscode/src/jsonrpc/agent-protocol.ts
olafurpg, line 549: This is a mouthful but mirrors the existing pattern of prefixing VSC types with `Protocol`
```suggestion
    changes?: ProtocolTextDocumentContentChangeEvent[]
```

mkondratek:
The original issue has been fixed and the missing feature (tracking content changes in the agent) has been added by https://github.com/sourcegraph/cody/pull/4058. closing

PR #3933, opened 2024-04-25
Remove Commands on Hover

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1713973441091799

This PR removes the `Hover commands` feature that was part of the `Commands on Hover` A/B test.

All users will now have Ghost Text enabled by default after this feature removal.

## Test plan

- Verify hover command is no longer available in this build
- Verify ghost text shows up


sqs (approved):
Thanks!

PR #3935, opened 2024-04-25
use pure-JS RE2 impl instead of re2-wasm

sqs:
[RE2JS](https://github.com/le0pard/re2js) is a pure JavaScript implementation of RE2. This is more portable than [re2-wasm](https://github.com/google/re2-wasm), which [does not support usage in the browser](https://github.com/google/re2-wasm/issues/15), which breaks storybooks (although there is an easy hack workarond) and prevents Cody Web from reusing this code. It is possible to make re2-wasm usable in the browser, but it would require forking the upstream, which does not seem worth the effort unless we find the performance of RE2JS to be unacceptable.

RE2JS does not support the `u` flag for RE2, which [enables various Unicode-related features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode), so I added a test to confirm that this is its default behavior.

(I wasn't sure if there was a reason you preferred re2-wasm. RE2JS is weird in that `.matches(input)` looks for full string matches by default, but otherwise it seems like a solid library.)

## Test plan

CI

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.ts
@@ -1,5 +1,5 @@
 import { LRUCache } from 'lru-cache'
-import { RE2 } from 're2-wasm'
+import { RE2JS as RE2 } from 're2js'
 import type * as vscode from 'vscode'
 import { isFileURI } from '../common/uri'
 import { logDebug, logError } from '../logger'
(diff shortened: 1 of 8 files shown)


lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -84,6 +84,7 @@ export class ContextFiltersProvider implements vscode.Disposable {
         if (this.contextFilters?.include?.length) {
             for (const parsedFilter of this.contextFilters.include) {
                 isIgnored = !matchesContextFilter(parsedFilter, repoName)
+                console.log('XX1', isIgnored, parsedFilter, repoName)
```
valerybugakov, line 87:
```suggestion
```

valerybugakov (approved):
Thanks Quinn! It looks like a solid replacement.

PR #3937, opened 2024-04-25
Cody Ignore: agent integration

valerybugakov:
## Context

- Integrates `ContextFiltersProvider` with the agent protocol. Uses the experimental Cody Ignore version for PLG users, which we can refactor later, as we agreed it has a lower priority now.
- Adds integration tests for the exposed `ContextFiltersProvider` methods.
- Built on top of https://github.com/sourcegraph/cody/pull/3923

## Test plan

CI and new tests.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -304,6 +304,12 @@ export class Agent extends MessageHandler implements ExtensionClient {
                     this.codeLenses.remove(codeLensProvider)
                 )
             }
+            if (clientInfo.capabilities?.codyIgnore === 'enabled') {
+                contextFiltersProvider.onContextFiltersChanged(() => {
+                    // Forward policy change notifications to the client.
+                    this.notify('ignore/didChange', null)
+                })
+            }
             if (process.env.CODY_DEBUG === 'true') {
                 console.error(
                     `Cody Agent: handshake with client '${clientInfo.name}' (version '${clientInfo.version}') at workspace root path '${clientInfo.workspaceRootUri}'\n`
@@ -1047,11 +1053,6 @@ export class Agent extends MessageHandler implements ExtensionClient {
             return {
                 policy: isUriIgnored ? 'ignore' : 'use',
             } as const
-        })
-
-        contextFiltersProvider.onContextFiltersChanged(() => {
-            // Forward policy change notifications to the client.
-            this.notify('ignore/didChange', null)
         })
 
         this.registerAuthenticatedRequest('testing/ignore/overridePolicy', async contextFilters => {
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1033,14 +1033,16 @@ export class Agent extends MessageHandler implements ExtensionClient {
             }
         })
 
+        let isIgnorePolicyOverrideDisabled = true
         this.registerAuthenticatedRequest('ignore/test', async ({ uri: uriString }) => {
             const uri = vscode.Uri.parse(uriString)
             const authStatus = await vscode.commands.executeCommand<AuthStatus>('cody.auth.status')
 
             // Use experimental Cody Ignore version for PLG users for now.
-            const isUriIgnored = authStatus.isDotCom
-                ? isCodyIgnoredFile(uri)
-                : await contextFiltersProvider.isUriIgnored(uri)
+            const isUriIgnored =
+                authStatus.isDotCom && isIgnorePolicyOverrideDisabled
+                    ? isCodyIgnoredFile(uri)
+                    : await contextFiltersProvider.isUriIgnored(uri)
 
             return {
                 policy: isUriIgnored ? 'ignore' : 'use',
@@ -1053,6 +1055,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
         })
 
         this.registerAuthenticatedRequest('testing/ignore/overridePolicy', async contextFilters => {
+            isIgnorePolicyOverrideDisabled = !!contextFilters
             contextFiltersProvider.setTestingContextFilters(contextFilters)
             return null
         })
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -304,7 +304,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
                     this.codeLenses.remove(codeLensProvider)
                 )
             }
-            if (clientInfo.capabilities?.codyIgnore === 'enabled') {
+            if (clientInfo.capabilities?.ignore === 'enabled') {
                 contextFiltersProvider.onContextFiltersChanged(() => {
                     // Forward policy change notifications to the client.
                     this.notify('ignore/didChange', null)
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -15,6 +15,7 @@ import {
     ModelProvider,
     NoOpTelemetryRecorderProvider,
     PromptString,
+    contextFiltersProvider,
     convertGitCloneURLToCodebaseName,
     featureFlagProvider,
     graphqlClient,
@@ -34,10 +35,6 @@ import { activate } from '../../vscode/src/extension.node'
 import { ProtocolTextDocumentWithUri } from '../../vscode/src/jsonrpc/TextDocumentWithUri'
 
 import type { Har } from '@pollyjs/persister'
-import {
-    setIgnorePolicyChangeListener,
-    setTestingIgnorePolicyOverride,
-} from '@sourcegraph/cody-shared/src/cody-ignore/context-filter'
 import levenshtein from 'js-levenshtein'
 import {
     AbstractMessageReader,
@@ -1036,27 +1033,27 @@ export class Agent extends MessageHandler implements ExtensionClient {
             }
         })
 
-        this.registerAuthenticatedRequest('ignore/test', async ({ uri }) => {
-            const policy: 'ignore' | 'use' = isCodyIgnoredFile(vscode.Uri.parse(uri)) ? 'ignore' : 'use'
+        this.registerAuthenticatedRequest('ignore/test', async ({ uri: uriString }) => {
+            const uri = vscode.Uri.parse(uriString)
+            const authStatus = await vscode.commands.executeCommand<AuthStatus>('cody.auth.status')
+
+            // Use experimental Cody Ignore version for PLG users for now.
+            const isUriIgnored = authStatus.isDotCom
+                ? isCodyIgnoredFile(uri)
+                : await contextFiltersProvider.isUriIgnored(uri)
+
             return {
-                policy,
-            }
+                policy: isUriIgnored ? 'ignore' : 'use',
+            } as const
         })
 
-        setIgnorePolicyChangeListener(() => {
+        contextFiltersProvider.onContextFiltersChanged(() => {
             // Forward policy change notifications to the client.
             this.notify('ignore/didChange', null)
         })
 
-        this.registerAuthenticatedRequest('testing/ignore/overridePolicy', async options => {
-            setTestingIgnorePolicyOverride(
-                options
-                    ? {
-                          repoRe: RegExp(options.repoRe),
-                          uriRe: RegExp(options.uriRe),
-                      }
-                    : undefined
-            )
+        this.registerAuthenticatedRequest('testing/ignore/overridePolicy', async contextFilters => {
+            contextFiltersProvider.setTestingContextFilters(contextFilters)
             return null
         })
     }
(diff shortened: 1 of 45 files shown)


vscode/src/testutils/mocks.ts
```diff
@@ -686,7 +686,9 @@ export enum UIKind {
     Web = 2,
 }
 
-export class FileSystemError extends Error {}
+export class FileSystemError extends Error {
+    public code = 'FileSystemError'
+}
```
valerybugakov, line 691: We need to mock all `FileSystemErorr` variants and integrate them into a file system mock. I created a follow-up issue for that: https://github.com/sourcegraph/cody/issues/3938

agent/src/agent.ts
```diff
@@ -1036,27 +1033,27 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
-        this.registerAuthenticatedRequest('testing/ignore/overridePolicy', async options => {
-            setTestingIgnorePolicyOverride(
-                options
-                    ? {
-                          repoRe: RegExp(options.repoRe),
-                          uriRe: RegExp(options.uriRe),
-                      }
-                    : undefined
-            )
+        this.registerAuthenticatedRequest('testing/ignore/overridePolicy', async contextFilters => {
```
dominiccooney, line 1055: You've wired up the dotcom side for reading, but not for overriding for test.

I would be fine, maybe even prefer it, if we kept the override for testing really simple and have it ignore the account type.

agent/src/index.test.ts
```diff
@@ -1396,7 +1404,7 @@ describe('Agent', () => {
     })
 
     describe('Enterprise', () => {
-        const enterpriseClient = TestClient.create({
+        const demoEnterpriseClient = TestClient.create({
```
dominiccooney, line 1407: I'm fine with it but I'm curious why. We could even call it `client`.

vscode/src/repository/enterprise-repo-name-resolver.ts
```diff
@@ -145,7 +145,7 @@ async function gitRemoteUrlsFromTreeWalkRecursive(uri: vscode.Uri): Promise<stri
 
         return remoteUrls.size ? Array.from(remoteUrls) : undefined
     } catch (error) {
-        if (error instanceof vscode.FileSystemError) {
+        if (error instanceof Error && 'code' in error) {
```
dominiccooney, line 148: This seems much broader than what we had here. Why did it change?

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -188,7 +189,7 @@ export type ClientRequests = {
     // For testing. Overrides any ignore policy to ignore repositories and URIs
     // which match the specified regular expressions. Pass `undefined` to remove
     // the override.
-    'testing/ignore/overridePolicy': [{ repoRe: string; uriRe: string } | null, null]
+    'testing/ignore/overridePolicy': [ContextFilters, null]
```
dominiccooney, line 192: This will break the testing setup in JetBrains. Which is OK, we can go fix it. But we gotta remember to do that.

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -60,28 +60,37 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
+            include: contextFilters.include?.map(parseContextFilterItem) || null,
+            exclude: contextFilters.exclude?.map(parseContextFilterItem) || null,
+        }
+    }
+
+    /**
+     * Sets an override context filters for testing.
+     */
+    public setTestingContextFilters(contextFilters: ContextFilters): void {
+        console.warn('Overriding Cody Ignore context filters for testing')
```
dominiccooney, line 90: Will this show up in enough logs?

This has five alarm fire potential if people start calling it outside test mode. Any ideas whether/how we should lock it down?

lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -60,28 +60,37 @@ export class ContextFiltersProvider implements vscode.Disposable {
...
+
+        logDebug('ContextFiltersProvider', 'setContextFilters', { verbose: contextFilters })
+        this.parsedContextFilters = {
+            include: contextFilters.include?.map(parseContextFilterItem) || null,
+            exclude: contextFilters.exclude?.map(parseContextFilterItem) || null,
+        }
+    }
+
+    /**
+     * Sets an override context filters for testing.
```
dominiccooney, line 87:
```suggestion
     * Overrides context filters for testing.
```

nit, grammar. Singular "an" does not agree with plural "filters". So let's simplify the whole thing.

agent/src/index.test.ts
```diff
@@ -1630,17 +1605,92 @@ describe('Agent', () => {
...
+                s2EnterpriseClient.registerNotification('ignore/didChange', onChangeCallback)
+
+                expect(await ignoreTest()).toStrictEqual({ policy: 'use' })
+
+                await s2EnterpriseClient.request('testing/ignore/overridePolicy', {
+                    include: [{ repoNamePattern: '' }],
+                    exclude: [{ repoNamePattern: '.*sourcegraph/cody.*' }],
+                })
+
+                expect(onChangeCallback).toBeCalledTimes(1)
```
dominiccooney, line 1634: This is much simpler than the Promise dance I had. It also depends on the notification being dispatched synchronously with the change. But I guess that's fine...

agent/src/index.test.ts
```diff
@@ -1630,17 +1605,92 @@ describe('Agent', () => {
...
+
+                expect(onChangeCallback).toBeCalledTimes(2)
+                expect(await ignoreTest()).toStrictEqual({ policy: 'use' })
+
+                await s2EnterpriseClient.request('testing/ignore/overridePolicy', {
+                    include: [{ repoNamePattern: '' }],
+                    exclude: [{ repoNamePattern: '.*sourcegraph/sourcegraph.*' }],
+                })
+
+                // onChangeCallback is not called again because filters are the same
```
dominiccooney, line 1650: I guess we can't test a no-op didChange notification now, but should be fine. I guess we could inject didChange on the client side.

dominiccooney (approved):
I'll ask the dumb question: How much do these tests rely on the context filter rules in `demo`, and what is stopping someone jumping on demo and breaking these tests in a way we'll only detect when we update the recordings?

agent/src/index.test.ts
valerybugakov: I was confused a couple of times by the same variable name that's shadowed. Easy to use the wrong one because of the huge list of tests. It makes it clear what instance you use without a need to go to a definition.

vscode/src/repository/enterprise-repo-name-resolver.ts
valerybugakov: The previous implementation failed in the agent because we do not have mocks for vscode file system errors. See https://github.com/sourcegraph/cody/pull/3937#discussion_r1578924394

valerybugakov:
> How much do these tests rely on the context filter rules in demo, and what is stopping someone jumping on demo and breaking these tests in a way we'll only detect when we update the recordings?

It will break the second test if somebody adds the `sourcegraph/cody` repo to the ignore list on the sourcegraph.sourcegraph.com instance.

vscode/src/jsonrpc/agent-protocol.ts
valerybugakov: What's the best way to proceed here? Is it OK to merge this PR into main and do a follow-up JetBrains PR?

lib/shared/src/cody-ignore/context-filters-provider.ts
valerybugakov: Replaced `console.warn()` with `throw new Error()` for calls outside of tests.

agent/src/agent.ts
valerybugakov: Added the account type override: https://github.com/sourcegraph/cody/pull/3937/commits/6afbfed782934aebe42789741bb585a4bffab5a6

vscode/src/repository/enterprise-repo-name-resolver.ts
olafurpg: `instanceof vscode.FileSystemError` should work correctly in the agent. As a rule of thumb, we should avoid as much as possible to change the implementation of the VSC extension to enable agent compatibility because it's preferable to fix the bug in the agent shim.

agent/src/agent.ts
```diff
@@ -1036,27 +1033,30 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
-                policy,
-            }
+                policy: isUriIgnored ? 'ignore' : 'use',
+            } as const
         })
 
-        setIgnorePolicyChangeListener(() => {
+        contextFiltersProvider.onContextFiltersChanged(() => {
             // Forward policy change notifications to the client.
             this.notify('ignore/didChange', null)
```
olafurpg, line 1054: We should ideally guard these notifications with a client capability. The client should opt-into codyignore support by specifying in `initialize` that it has the codyignore capability.

agent/src/index.test.ts
olafurpg: fwiw, I want to split up this file so these `describe()` blocks would live in individual files

vscode/src/repository/enterprise-repo-name-resolver.ts
valerybugakov: `workstream.fs` should start throwing `FileSystemError` subtypes instead of usual node.js errors. Then, we will be able to use `instanceof vscode.FileSystemError` here. But this is not required for the behavior implemented in this PR, and it's compatible with both VS Code and node.js errors. I created a follow-up issue to address that: https://github.com/sourcegraph/cody/issues/3938

olafurpg: Fair enough. It's not a big deal here, but I'd love to aim for full VSC compatibility even if that means wrapping Node.js file errors into `vscode.FileSystemError`. It's going to take less overall work if we just make the agent a drop-in replacement instead of implementing the plugin against two different SDKs.

olafurpg (approved):
LGTM 👍🏻

lib/shared/src/cody-ignore/context-filters-provider.ts
olafurpg: We should use `logError/logDebug` instead of `console.error/log/warn`. The output of running the tests should be empty by default because we silence `debug/message` notifications

agent/src/agent.ts
valerybugakov: Sounds good: https://github.com/sourcegraph/cody/pull/3937/commits/1f446c255625057f983a1a7854dbdd3576b83897

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -422,6 +423,7 @@ interface ClientCapabilities {
     showDocument?: 'none' | 'enabled'
     codeLenses?: 'none' | 'enabled'
     showWindowMessage?: 'notification' | 'request'
+    codyIgnore?: 'none' | 'enabled'
```
olafurpg, line 426: The capability should match the name of the JSON-RPC methods. I propose we settle on the name `codyignore` (all lowercase) and use this consistently in all APIs instead of a mix of "ignore" and "codyIgnore"

olafurpg (changes requested):

vscode/src/jsonrpc/agent-protocol.ts
valerybugakov: What do you think about using `ignore` since it's already used in the protocol and in the JetBrains client?

olafurpg: that's fine. We can rename into `codyignore` later (I think that's a better name than the verb `ignore`)

olafurpg (approved):

vscode/src/jsonrpc/agent-protocol.ts
dominiccooney: Yeah merge away. I'm happy to do the follow up.

This is a step backwards because we can't do per-file stuff now, but so be it.

vscode/src/jsonrpc/agent-protocol.ts
dominiccooney: Pleeeeeease don't call it `codyignore`? We'll end up with `codyeverything` and that is distracting... `codychat`, `codycommands`...

I noticed the docs are ambivalent about this too... they refer to it as Cody Ignore and Ignore Context.

olafurpg: I'm fine with keeping `ignore` for now, it's consistent with `chat`/`commands`.

agent/src/agent.ts
dominiccooney: It would be better if you had an interface for the policy and substituted a separate implementation when it is overridden. That way you don't complicate the straight line product code with flags like this.

PR #3939, opened 2024-04-25
"Usage Examples" Cody command

sqs:
The experimental "Usage Examples" Cody command shows usage examples drawn from a symbol's definition and call sites. The goal is to show useful and idiomatic usage examples by seeing how a symbol is *actually* used.

To use it, set `cody.experimental.noodle` to `true` in your VS Code settings, and then right-click on a symbol in code > Cody > Usage Examples. It only works for symbols defined in npm packages right now (and overall is highly experimental).

- Demo: https://www.loom.com/share/5c78ab157cba4492b9fb470f68984c89
- Feedback: https://community.sourcegraph.com/t/usage-examples-feedback-on-new-experimental-feature/191
- Internal Sourcegraph-only message: https://sourcegraph.slack.com/archives/C06RW56B2DD/p1714026533712339

## Test plan

CI


sqs:
(will fix the safe prompt issue before merging, this is just to try it out locally while still in draft)

vscode/src/commands/execute/usage-examples.ts
```diff
@@ -0,0 +1,81 @@
+import { PromptString, logDebug, ps, wrapInActiveSpan } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+import { getEditor } from '../../editor/active-editor'
+import type { ChatCommandResult } from '../../main'
+import { telemetryRecorder } from '../../services/telemetry-v2'
+import type { CodyCommandArgs } from '../types'
+import { executeChat } from './ask'
+
+/**
+ * The command that generates a new docstring for the selected code.
```
thenamankumar, line 10: Copied dotstring?

sqs:
@philipp-spiess This PR uses unsafe prompt string creation to interpolate the string `npm` (a package manager) and an npm package name (determined by calling the definition provider on something in VS Code and inspecting the file path of the definition location). Should I add another static function on PromptString for this? Or just accept the unsafe call with manual human review?

philipp-spiess:
@sqs So sorry for the noisy bot, I'll make it so it only comments on the PR once. 

I think it's good to mark this as unsafe for now since npm packages could be pulled in from ignored subrepos I suppose? E.g. if you use relative links in package.json for example. (cc @dominiccooney)

Ideally we move the `unsafe` annotation closer to the definition, so maybe change the `SymbolPackage` record to contain `PromptString`s instead of strings and then the "npm" part could be a literal and only the part derived from the path would be marked unsafe.

sqs:
@philipp-spiess:

> I think it's good to mark this as unsafe for now since npm packages could be pulled in from ignored subrepos I suppose? E.g. if you use relative links in package.json for example. (cc @dominiccooney)
> 
> Ideally we move the `unsafe` annotation closer to the definition, so maybe change the `SymbolPackage` record to contain `PromptString`s instead of strings and then the "npm" part could be a literal and only the part derived from the path would be marked unsafe.

Cool, I moved the PromptString static constructor calls to the same scope where the strings were created.

I thought about creating a `PromptString.fromLiteral('npm')` but it seems very prone to misuse because there is no way to ensure that the argument is actually a string literal.

vscode/src/commands/execute/usage-examples.ts
```diff
@@ -0,0 +1,266 @@
...
+        await PACKAGE_CONTEXT_MENTION_PROVIDER.queryContextItems(
+            `${symbolPackage.ecosystem}:${symbolPackage.name}`
+        )
+    ).filter(item => item.title === symbolPackage.name.toString())
+    logDebug('executeUsageExampleCommand', 'found packages', JSON.stringify({ packages }))
+
+    const globalRepos = (
+        await Promise.all(
+            packages.map(item =>
+                searchForRepos(
```
thenamankumar, line 157: I am unable to find where the number of results for searching the repos is defined. Is it implicitly limited somewhere?

We would probably want to fetch only 5 globally related repos.

Then, we are fetching 5 context items for each of the repos.

That makes the total number of context items 20(package) + 5 * 5 = 45.

Do we really want these many context items?

vscode/src/commands/execute/usage-examples.ts
```diff
@@ -0,0 +1,266 @@
...
+            )
+        )
+    ).flat()
+    logDebug('executeUsageExampleCommand', 'global repos', JSON.stringify({ globalRepos }))
+    const uses = (
+        await Promise.all(
+            globalRepos.map(async repo =>
+                repo instanceof Error
+                    ? []
+                    : await searchForFileChunks(
```
thenamankumar, line 171: The search query should have the ignore file filters like `-file:\.min\.js$`...

vscode/src/commands/execute/usage-examples.ts
```diff
@@ -0,0 +1,266 @@
...
+            }
+        }
+    }
+
+    return null
+}
+
+function includeContextItem(item: ContextItem): boolean {
+    return (
+        !item.uri.path.endsWith('.map') &&
```
thenamankumar, line 252: I have updated the backend filters as well in this PR: https://github.com/sourcegraph/sourcegraph/pull/62261

dominiccooney:
> I thought about creating a PromptString.fromLiteral('npm') but it seems very prone to misuse because there is no way to ensure that the argument is actually a string literal.

@sqs, we have exactly that:

```
ps`npm`
```

Try to break it, we have a lint set up to detect aliasing `ps` or calling it as a function, etc. (@philipp-spiess are we running that lint in CI yet?)

sqs:
> > I thought about creating a PromptString.fromLiteral('npm') but it seems very prone to misuse because there is no way to ensure that the argument is actually a string literal.
> 
> @sqs, we have exactly that:
> 
> ```
> ps`npm`
> ```

Ohh yeah.

vscode/src/commands/execute/usage-examples.ts
sqs: I haven't seen this actually being a problem since we use relatively small snippets from each of the files.

sqs: Oh but yeah I'll add a limit on the number of repos, a fairly high one to ensure we do find results but not so high that it can get overwhelmed. (It's all a guess right now, though.)

vscode/src/commands/execute/usage-examples.ts
sqs: These ignore filters should be applied to the files in the published package, but not the global search for callers because that mainly searches repositories, which don't have those build artifacts.

thenamankumar (approved):

philipp-spiess:
@dominiccooney 

>@philipp-spiess are we running that lint in CI yet?

Yep it runs together with the `unsafe_` use detection :)

PR #3942, opened 2024-04-25
PromptString: Log blocked references

philipp-spiess:
Add logging around blocked references and simplify the message since it might leak to the user while we find all cases where context is pulled in from.

## Test plan

- CI

### DIFF
--- a/lib/shared/src/prompt/prompt-string.ts
+++ b/lib/shared/src/prompt/prompt-string.ts
@@ -11,6 +11,7 @@ import type { ActiveTextEditorDiagnostic } from '../editor'
 import { createGitDiff } from '../editor/create-git-diff'
 import { displayPath } from '../editor/displayPath'
 import { getEditorInsertSpaces, getEditorTabSize } from '../editor/utils'
+import { logDebug } from '../logger'
 
 // This module is designed to encourage, and to some degree enforce, safe
 // handling of file content that gets constructed into prompts. It works this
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
💯

dominiccooney:
Should we add Telemetry for these blocks too?

philipp-spiess:
@dominiccooney YES absolutely. I enabled auto-merge too soon :D

valerybugakov (approved):

lib/shared/src/prompt/prompt-string.ts
```diff
@@ -58,13 +59,30 @@ export class PromptString {
...
+                    'toFilteredString',
+                    `${reference} is ignored by the current context filters ${contextFilter}`
+                )
+            }
+        }
+
+        if (shouldThrow) {
             throw new Error(
-                'The prompt string contains a reference to a file that is not allowed by the context filters.'
+                'The prompt contains a reference to a file that is not allowed by your current Cody policy.'
```
PriNova, line 82: Upon reading this, one might wonder, "What is the Cody policy and where can it be configured?". Rephrasing it to "The prompt contains a reference to a file that is not allowed by **_our_** current Cody policy." would be more appropriate. However, the question arises: "Where can I find out about the Cody policy?".
Wdyt?

philipp-spiess: This is a great point! Maybe it should be: "by **your administrators Cody policy**"?  We haven't fully finalized all the wording yet (there are a bunch of different strings floating around between VS Code and JetBrains) so it's a good call-out to get that straight.
 
cc @dominiccooney @danielmarquespt  @valerybugakov

PriNova: That sounds complete.
From this I can deduce that only Enterprises are affected!? In other words, PLG users will no longer receive this message in future?

philipp-spiess: @PriNova Yep that's the idea 👍

PR #3943, opened 2024-04-25
Move TelemetryRecorder singleton into lib/shared

philipp-spiess:
This PR moves the singleton instance of the `TelemetryRecorder` into `lib/shared` so we can use it from anywhere. It also adds a telemetry event to the last-line-of-defence for the context filter work so we have observability into this issue.

## Test plan

- CI
- I can still chat w/ Cody


lib/shared/src/prompt/prompt-string.ts
```diff
@@ -59,10 +60,9 @@ export class PromptString {
         contextFilter: Pick<ContextFiltersProvider, 'isUriIgnored'>
     ): Promise<string> {
         const references = internal_toReferences(this)
-        const checks = references.map(async reference => [
-            reference,
-            await contextFilter.isUriIgnored(reference),
-        ])
+        const checks = references.map(
+            async reference => [reference, await contextFilter.isUriIgnored(reference)] as const
+        )
```
philipp-spiess, line 65: `as const` here makes sure the type is a tuple and not `Array<boolean | Reference>`

dominiccooney (approved):

PR #3944, opened 2024-04-25
ContextFilters: Return block reason and expose current state for logging purposes

philipp-spiess:
- Changes the return type for blocked references to a string giving a reason. We type it so that the string is always non-empty so we can still keep using `if(..)` checks easily.
- Add an API to expose the current filters as a debug object that can be used in our last-line-of-defence logging

## Test plan

Screenshot 2024-04-25 at 12 32 29

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.ts
@@ -26,7 +26,7 @@ interface ParsedContextFilterItem {
     filePathPatterns?: RE2[]
 }
 
-type IsIgnored =
+export type IsIgnored =
     | false
     | 'has-ignore-everything-filters'
     | 'non-file-uri'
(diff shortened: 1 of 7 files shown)


lib/shared/src/cody-ignore/context-filters-provider.ts
```diff
@@ -172,6 +188,12 @@ export class ContextFiltersProvider implements vscode.Disposable {
     private hasIgnoreEverythingFilters() {
         return this.lastContextFiltersResponse === EXCLUDE_EVERYTHING_CONTEXT_FILTERS
     }
+
+    public toDebugObject() {
+        return {
+            lastContextFiltersResponse: this.lastContextFiltersResponse,
```
dominiccooney, line 194: Could we `JSON.parse(JSON.stringify(...))` that to return a defensive copy of it?

dominiccooney (approved):
Great, with access to a log there won't be any doubt about block reasons.

valerybugakov (commented):
So nice!

PR #3946, opened 2024-04-25
Cody Ignore: Tidy up context filter overrides

dominiccooney:
- Context filters should fire the changed event *after* parsing the filters, so clients which query the filters can see updated values.
- Remove the hook for applying overrides for dotcom users at the agent level; the overrides need to be at the extension level. For now we simply won't integrate file-based Cody Ignore here.
- Remove the VITEST guard. The purpose of this hook is largely for testing IntelliJ, which is not VITEST.
- For now, disables Agent file-based Cody Ignore tests. File-based Cody Ignore should be integrated at the extension level, not the Agent "container" level. Then these tests can come back.

Part of #3641 

## Test plan

Manually tested with a sourcegraph/jetbrains#1396

CI

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -122,7 +122,8 @@ describe('Agent', () => {
         const ignore = await client.request('ignore/test', {
             uri: URI.file(ignoredPath).toString(),
         })
-        expect(ignore.policy).toBe('ignore')
+        // TODO(dpc): Integrate file-based .cody/ignore with ignore/test
+        expect(ignore.policy).toBe('use')
     }, 10_000)
 
     beforeEach(async () => {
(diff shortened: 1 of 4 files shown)


dominiccooney:
I can't update these recordings because login is failing on demo 😞 

```
 FAIL  |agent| src/index.test.ts > Agent > Enterprise
AssertionError: expected false to be truthy
 ❯ src/index.test.ts:1422:55
    1420|             const serverInfo = await demoEnterpriseClient.initialize()
    1421| 
    1422|             expect(serverInfo.authStatus?.isLoggedIn).toBeTruthy()
       |                                                       ^
    1423|             expect(serverInfo.authStatus?.username).toStrictEqual('codytesting')
    1424|         }, 10_000)
```

agent/src/index.test.ts
```diff
@@ -540,7 +541,8 @@ describe('Agent', () => {
         })
     })
 
-    describe('Cody Ignore', () => {
+    // TODO(dpc): Integrate file-based .cody/ignore with ignore/test
+    describe.skip('Cody Ignore', () => {
```
valerybugakov, line 545: That works!

valerybugakov (approved):
> The purpose of this hook is largely for testing IntelliJ

This is the context I was missing when integrating the override hook. Makes sense!

PR #3947, opened 2024-04-25
Remove double-bundling of some lib/shared files

philipp-spiess:
We noticed that the `ContextFilterProvider` was bundled twice in the VS Code production builds. To fix this, we now move all `@sourcegraph/cody-shared` requires to use the main bundle instead of requiring files directly.

## Todo

- [ ] Can we change the package.json to not expose the individual files in the lib/shared package but only the barrel import?

## Test plan

Screenshot 2024-04-25 at 17 11 07

Add a console log to the context provider and see that it's only bundled once:

```
α cody/vscode 🐙(ps/remove-double-bundling-of-some-lib-shared-files)✗
cat dist/extension.node.js | grep "THIS SHOULD ONLY EVER BE LOGGED ONCE"
console.log("THIS SHOULD ONLY EVER BE LOGGED ONCE");
```

### DIFF
--- a/agent/recordings/enterpriseClient_3965582033/recording.har.yaml
+++ b/agent/recordings/enterpriseClient_3965582033/recording.har.yaml
@@ -3883,6 +3883,124 @@ log:
         send: 0
         ssl: -1
         wait: 0
+    - _id: 7399dcfe6e08a79f7b3d64aa058745f1
+      _order: 0
+      cache: {}
+      request:
+        bodySize: 347
+        cookies: []
+        headers:
+          - _fromType: array
+            name: authorization
+            value: token
+              REDACTED_b20717265e7ab1d132874d8ff0be053ab9c1dacccec8dce0bbba76888b6a0a69
+          - _fromType: array
+            name: content-type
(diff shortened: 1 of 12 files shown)


dominiccooney (approved):
Nice clean up.

The sooner we remove lib/shared, the better? Or will is this the escape capsule for Cody brains on their way to the cloud?

philipp-spiess:
> The sooner we remove lib/shared, the better? Or will is this the escape capsule for Cody brains on their way to the cloud?

I don't think the current abstraction is bringing us a lot of value. We use vscode imports in lib/shared like it was the vscode package anyways. And it's not really shared with anything but the vscode build right now? I’m all for removing it and trying it again with a bit more thought put into the abstractions

philipp-spiess:
I don't understand why this made the agent tests flaky 🤔 They are passing locally.

abeatrix (approved):
ty!

valerybugakov (commented):
Looks like all of us bumped into issues around `lib/shared` recently. Looking forward to merging it with one of the other packages to simplify the setup!

PR #3956, opened 2024-04-26
Feature-tracking: Add a script to merge features across repos

dominiccooney:
## Test plan

Manually tested. Patch the features files from #3846 and sourcegraph/jetbrains#1400 , then:

```
$ pnpm features merge ../features.json5 ~/projects/jetbrains/features5.json
```

Output:

```json
{
  "editors": [
    "vscode",
    "jetbrains"
  ],
  "features": [
    {
      "name": "NotebookBasedChatUI",
      "description": "new chat UI that resembles a notebook instead of a chat session",
      "productLine": "all",
      "documentation": [],
      "editors": {
        "vscode": {
          "status": "stable"
        },
        "jetbrains": {
          "status": "planned"
        }
      },
      "tags": [
        "chat"
      ]
    },
   ...
```

### DIFF
--- a/feature-tracking/index.ts
+++ b/feature-tracking/index.ts
@@ -1,0 +1,140 @@
+import { promises as fs } from 'fs'
+
+import { ArgumentParser } from 'argparse'
+import JSON5 from 'json5'
+
+// Types describing features.json5
+
+interface Database {
+    editors?: string[],
+    features: Feature[]
+}
+
+interface Feature {
+    name: string,
+    description?: string | undefined,
+    productLine?: 'free' | 'pro' | 'consumer' | 'enterprise' | 'all',
+    documentation?: string[],
+    editors?: {[key: string]: EditorFeature},
+    tags?: string[],
+}
+
+interface EditorFeature {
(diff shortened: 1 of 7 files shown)


feature-tracking/index.ts
```diff
@@ -0,0 +1,140 @@
...
+        dest: 'command'
+    })
+    const mergeParser = commandParser.add_parser('merge')
+    mergeParser.add_argument('file', {help: 'features.json5 file(s) to merge', nargs: '+'})
+    mergeParser.add_argument('--upstream', {help: 'Include upstream features file', action: 'store_true'})
+    const args = parser.parse_args()
+
+    switch (args.command) {
+        case 'merge':
+            await mergeFiles(args.file)
```
kalanchan, line 133: nice, when we automate this we can easily run something like 

`pnpm features merge <machine_root>/cody/featuresA.json5 <machine_root>/jetbrains/featuresB.json5`

kalanchan (approved):
looks good, I tried the script and it works well. 

My only comment is around overwriting behaviour, maybe this can just be a process thing but we should set guidelines on when an overlay should add fields that overwrite the base file. 

maybe it's as simple as "if you're touching an overlay, don't add any optional fields if not necessary"

PR #3958, opened 2024-04-26
noodle: sourcegraph search also uses ? as a trigger prefix

keegancsmith:
This is inline with the examples Quinn was using.

Test Plan: tested mentioning search results with both `@?` and `@src:` as a trigger prefix.

### DIFF
--- a/lib/shared/src/mentions/providers/sourcegraphSearch.ts
+++ b/lib/shared/src/mentions/providers/sourcegraphSearch.ts
@@ -8,13 +8,14 @@ import type { ContextMentionProvider } from '../api'
 
 export const SOURCEGRAPH_SEARCH_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'src-search'> = {
     id: 'src-search',
-    // TODO the prefix '!' seems to not trigger the @ mention code path. I am
-    // assuming there is parsing logic somewhere which only calls the @
-    // mentions if they look like a path or start with # (for symbols)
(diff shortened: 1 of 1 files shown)


lib/shared/src/mentions/providers/sourcegraphSearch.ts
```diff
@@ -8,13 +8,14 @@ import type { ContextMentionProvider } from '../api'
 
 export const SOURCEGRAPH_SEARCH_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'src-search'> = {
     id: 'src-search',
-    // TODO the prefix '!' seems to not trigger the @ mention code path. I am
-    // assuming there is parsing logic somewhere which only calls the @
-    // mentions if they look like a path or start with # (for symbols)
-    triggerPrefixes: ['src:'],
+    triggerPrefixes: ['src:', '?'],
```
thenamankumar, line 11: Did we update the regex in the previous PRs to allow `?` ?

thenamankumar (approved):

lib/shared/src/mentions/providers/sourcegraphSearch.ts
keegancsmith: in my previous PR I was trying out ! which wasn't allowed. ? has been allowed for a while, following the history I am not exactly sure why it was added as an explicit prefix. But it was removed from punctuation as part of allowing @mention URLS

PR #3965, opened 2024-04-27
new popover impl for model dropdown

sqs:
Previously the model dropdown (that lets you choose the LLM) used the VSCodeDropdown from [@vscode/webview-ui-toolkit](https://github.com/microsoft/vscode-webview-ui-toolkit). This component (1) does not work nicely with React and React types, (2) is tricky to style owning to how it chose to use slots, and (3) does not support the more complex popover design for this UI component in the [Figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=2363-51406&mode=design&t=rOnlwu021jaX1akF-0).

While this change does not implement that new design yet, it brings in a new popover/select list implementation that will alleviates the problems with VSCodeDropdown and will allow us to implement the more complex popover design.

There are only 2 known changes:

- Minor styling differences
- When the model dropdown is focused, the up/down arrows do not change the value. You need to press Enter and then use the up/down arrows (or use the mouse). This temporary keyboard accessibility regression is acceptable because soon the UI component will no longer be a simple list.

Related to https://github.com/sourcegraph/cody/issues/3792

image
image

## Test plan

CI


sqs:
If we can find a good existing UI component in some library that lets us implement keyboard-accessible lists like in the [Figma](https://www.figma.com/file/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?type=design&node-id=2363-51406&mode=design&t=rOnlwu021jaX1akF-0), we should use it.

vscode/webviews/Components/platform/SelectList.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }, [])
+
+    return (
+        <div className={classNames(styles.container, FLOATING_WIDGET_CLASS_NAME, className)}>
+            <ul
+                className={styles.list}
+                role="radiogroup"
+                onKeyDown={e => {
+                    setIsKeyDown(true)
+                    if (e.key === 'Enter') {
```
philipp-spiess, line 34: Native select boxes can also be changed via space:

```suggestion
                    if (e.key === 'Enter' || e.key === ' ' ) {
```

vscode/webviews/Components/platform/SelectList.tsx
```diff
@@ -0,0 +1,77 @@
...
+    disabled?: boolean
+}
+
+export const SelectList: FunctionComponent<{
+    options: SelectListOption[]
+    value: string | undefined
+    onChange: (value: Value | undefined, close: boolean) => void
+    className?: string
+}> = ({ options, value, onChange, className }) => {
+    const [isKeyDown, setIsKeyDown] = useState(false)
```
philipp-spiess, line 20: If this changes, we don't need to re-render the UI (it seems to only be used within an event handler). If that's not necessary, maybe consider a mutable ref instead (so that mutating won't cause re-renders)

vscode/webviews/Components/platform/SelectList.tsx
```diff
@@ -0,0 +1,77 @@
...
+                                role="radio"
+                                name="option"
+                                value={optionValue}
+                                checked={value === optionValue}
+                                aria-checked={value === optionValue}
+                                ref={el => {
+                                    // Make arrow keys work correctly.
+                                    if (el?.checked || (value === undefined && i === 0)) {
+                                        // TODO!(sqs): the up/down arrows are broken if value === undefined, they should start at the 0'th element and down should go to the 1st
+                                        setTimeout(() => el?.focus())
```
philipp-spiess, line 62: The intention here seems to be to move the focus into the first input element once the component renders but it's not super clear when reading this the first time. What about moving this into a `useEffect` with some DOM APIs?

This would also remove the need for the awkward `setTimeout`.

```ts
useEffect(() => {
  const checked = ulRef.current?.querySelector('input[checked]');
  const first = ulRef.current?.querySelector('input:first-child');
  (checked || first)?.focus();
}, [])
```

vscode/webviews/Components/platform/Button.tsx
```diff
@@ -0,0 +1,119 @@
...
+        <>
+            <Button
+                appearance="secondary"
+                iconEnd={disabled ? undefined : 'chevron'}
+                ref={anchorRef}
+                onClick={onButtonClick}
+                className={styles.popoverButton}
+                disabled={disabled}
+                role="combobox"
+                {...props}
```
philipp-spiess, line 104: I think this should set `aria-expanded` too, see: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/

```suggestion
            <Button
                appearance="secondary"
                iconEnd={disabled ? undefined : 'chevron'}
                ref={anchorRef}
                onClick={onButtonClick}
                className={styles.popoverButton}
                disabled={disabled}
                role="combobox"
                aria-expanded={isOpen}
                {...props}
```

philipp-spiess (approved):

sqs:
Thank you, @philipp-spiess!

PR #3971, opened 2024-04-29
Upgrade users from Claude 2 to Claude 3 and remove Claude 2

### LINKED ISSUE #3793: Nudge Pro users on legacy models to Claude 3
Pro users who use a legacy model (e.g. Claude 2.0, 2.1, Instant) should be nudged to select one of the Claude 3 models.  We should note that we will eventually deprecate the 2.0, 2.1, Instant models. 

Ideas
- In conjunction with #3792, use UX within dropdown to nudge users
- Banner within chat pre/post submission suggesting to try newer models

Goal 
- Move most users off of legacy models ahead of a deprecation

philipp-spiess:
Closes #3793

In the spirit of moving fast, what if we just render a simple notification when users have picked a variant of Claude 2 as their default models and upgrade them to Sonnet (our new default) which should be as good as their previous pick? 

We'll need to find out all cases where we do that (loading older chats, edit) and make sure to only render a notification once but this should be simpler than adding custom UI in all of these features.

Note: To not break other clients, these models are still present in the code but won't be rendered in the UI (and since loading them from defaults will migrate, these should never show up again for new users).

## Test plan

Screenshot 2024-04-29 at 17 26 32

- Open a new chat (after setting Claude 2 as the default)
- Open an existing chat that used Claude 2
- Create a new edit (after setting Claude 2 as the default)

You might want to add this code so the modal is shown more than once:

```ts
await localStorage.delete('claude2-migration-notification-shown')
```

Screenshot 2024-04-30 at 18 36 28
Screenshot 2024-04-30 at 18 30 52


philipp-spiess:
@toolmantim You mentioned in Slack that you added a comment here re. wording. Seems like you forgot to press send though 😬 

While I have you here: Any concerns of upgrading previous chat histories when they're opened to Claude 3 too?

vscode/src/models/modelMigrator.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    // Claude 2 to Claude 3 migration
+    if (
+        model === 'anthropic/claude-instant-1.2' ||
+        model === 'anthropic/claude-2.0' ||
+        model === 'anthropic/claude-2.1'
+    ) {
+        const newModel = 'anthropic/claude-3-sonnet-20240229'
+        showNotificationIfNotShownYet(
```
abeatrix, line 17: nit: Since this is not related to the system, chatManager.triggerNoticemight be a more suitable way to display this message?

vscode/src/models/modelMigrator.ts
```diff
@@ -0,0 +1,33 @@
...
+            'claude2-migration-notification-shown'
+        ).catch(console.error)
+        return newModel
+    }
+
+    return model
+}
+
+async function showNotificationIfNotShownYet(title: string, key: string): Promise<void> {
+    const value = localStorage.get(key)
```
abeatrix, line 28: is this necessary? Since we have already `setModel` to the new one when we detect an old one,  it should mean this will only be shown once?

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1276,7 +1277,7 @@ function newChatModelFromSerializedChatTranscript(
     modelID: string
 ): SimpleChatModel {
     return new SimpleChatModel(
-        json.chatModel || modelID,
+        migrateAndNotifyForOutdatedModels(json.chatModel || modelID)!,
```
abeatrix, line 1280: Should we move the `migrateAndNotifyForOutdatedModels` check to the `SimpleChatModel` constructor instead to make sure no SimpleChatModel starts with claude-2?

lib/shared/src/models/dotcom.ts
```diff
@@ -103,6 +111,7 @@ const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
         codyProOnly: true,
         usage: [ModelUsage.Chat, ModelUsage.Edit],
         contextWindow: { input: CHAT_INPUT_TOKEN_BUDGET, output: CHAT_OUTPUT_TOKEN_BUDGET },
+        deprecated: true,
```
abeatrix, line 114: if we are not displaying and supporting Claude 2 models, why don't we remove them from the list?

toolmantim:
@philipp-spiess sorry, you're right: I totally forgot to hit submit on my pending review. I just updated it based on your recent changes.

abeatrix (commented):
Left some comments and questions inline. 

Is there a reason we want to keep Claude 2 models in the model provider list if we are not displaying them in the UI? That way `setModel` can update the model ID to use the current default if the model is not in the available model list, or are we still using these models somewhere?

Also run into a UI bug where i'm not able to choose a new model. The dropdown is hidden behind the toolbar 🤔 
image

image


https://github.com/sourcegraph/cody/assets/68532117/52e696ee-ac06-49c8-ad4b-a172c0173538

vscode/src/models/modelMigrator.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    // Claude 2 to Claude 3 migration
+    if (
+        model === 'anthropic/claude-instant-1.2' ||
+        model === 'anthropic/claude-2.0' ||
+        model === 'anthropic/claude-2.1'
+    ) {
+        const newModel = 'anthropic/claude-3-sonnet-20240229'
+        showNotificationIfNotShownYet(
+            'We upgraded you to Claude 3! The Claude 2 family of models is no longer available but you have access to the better Claude 3 model instead.',
```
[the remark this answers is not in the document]
philipp-spiess: yeah it shows once when either:

- you have one of the old models saved for chat or edit 
- or you open an old chat which had claude 2 set

once it's shown once, it won't show up again but all future saved chats you open will also change to use claude 3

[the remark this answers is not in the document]
philipp-spiess: Any idea where learn more should link to? the release blog post?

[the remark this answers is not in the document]
philipp-spiess: The issue with your proposed message is that someone could have had claude 2 selected, see this message (without the _chatsUpgraded_ part) and then move to open a chat. I don't think it's particularly confusing though so prob not worth fixing for this case?

philipp-spiess:
> Is there a reason we want to keep Claude 2 models in the model provider list if we are not displaying them in the UI? That way setModel can update the model ID to use the current default if the model is not in the available model list, or are we still using these models somewhere?

yea the reason is that we haven't added the messaging to jetbrains and removing it would break jetbrains

lib/shared/src/models/dotcom.ts
philipp-spiess: would break jetbrains I think (or at the least it will just remove support for claude 2 on jetbrains with no messaging)

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
philipp-spiess: if we need different wording based on the use case (i.e chat resroted vs. previous model saved like tim suggested) we may not be able to. Otherwise this might be a good idea

vscode/src/models/modelMigrator.ts
philipp-spiess: @abeatrix not the case for when restoring chats. there are three different ways right now we can set models:

- edit
- chat
- chat history

no matter where we see claude 2 though, i only want to show this notification once.

philipp-spiess:
@abeatrix I tried to repro the bug but couldn't. Anything specific you did to trigger this? Tried with S2 and dotcom 🤔

lib/shared/src/models/dotcom.ts
abeatrix: Gotchat.
JetBrains pulls the same list from ModelProvider and IIRC also fallback to the default one if the model cant be found on the list, so i thought it'd be easier if we just remove it from the list so they can get it for free.

philipp-spiess: Yeah we should be able to do that shortly but let's check with @pkukielka if we need to do anything to make the notifications work? I’m worried we don't implement `showInformationMessage` on the agent and then it will be a silent change (which we don't want for VS Code so I doubt we want it for JetBrains).

vscode/src/models/modelMigrator.ts
```diff
@@ -0,0 +1,33 @@
...
+        return model
+    }
+
+    // Claude 2 to Claude 3 migration
+    if (
+        model === 'anthropic/claude-instant-1.2' ||
+        model === 'anthropic/claude-2.0' ||
+        model === 'anthropic/claude-2.1'
+    ) {
+        const newModel = 'anthropic/claude-3-sonnet-20240229'
```
abeatrix, line 16: would this be an issue for enterprise users that do not use claude-3?

abeatrix (approved):
Will wait for Tim to confirm the design change but the code lgtm. 
UI issue i run into most likely not due to changes in this PR.

Left a question about setting `claude-3-sonnet` for Enterprise users.

vscode/src/models/modelMigrator.ts
philipp-spiess: Uh oh that's a great question. I think claude 3 on enterprise does work now but it will require a modern client. Since we don't relly allow custom models on ent anyways I suggest we add a gate here and don't run this for enterprise, just to be safe. WDYT?

vscode/src/models/modelMigrator.ts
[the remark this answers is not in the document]
toolmantim: > Any idea where learn more should link to? the release blog post?

I was thinking a "Deprecated Models" section of the docs (wdyt @kukicado?)

> The issue with your proposed message is that someone could have had claude 2 selected, see this message (without the chatsUpgraded part) and then move to open a chat. I don't think it's particularly confusing though so prob not worth fixing for this case?

If it's a new chat that isn't in history, as long as the dropdown was automatically switched to Claude 3, I don't think it's too bad?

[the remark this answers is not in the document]
abeatrix: @toolmantim I'm taking over @philipp-spiess 's PR while he's out. Should I update the notification to `'Claude 2 model support has been removed in favor of the newer Claude 3 models.\nAll chats that used Claude 2 have been upgraded to Claude 3.'` or should i keep the current copy?

vscode/src/models/modelMigrator.ts
abeatrix: Added a check to verify that the default model is available in their provider list before we show the notification 👍

vscode/src/models/modelMigrator.ts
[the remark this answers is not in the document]
toolmantim: New copy would be good thanks! Does it match the implementation? I haven’t been able to test this one locally.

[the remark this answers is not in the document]
abeatrix: > Does it match the implementation? I haven’t been able to test this one locally.

Yup. We have removed the Claude 2 models from the UI, and set the default model to Claude 3 for anyone who is using Claude 2 models.

We only show the notification if they last used Claude 2 for chat and edit, or when they restore an old chat from their chat history. Which means we are checking if we have shown the notification before every time they perform those actions 😅

PR #3974, opened 2024-04-29
Implement V2 telemetry validation in E2E Tests

akalia25:
This PR adds assertions for V2 telemetry events within our E2E test suite. Performing these additional checks will help ensure the accuracy and completeness of V2 analytics data by validating the expected events are logged for each test.

This change provides increased confidence in the stability of our V2 telemetry pipeline and will aid in preventing regressions.

The V2 telemetry assertions supplement the existing V1 event logging implemented previously [here] (https://github.com/sourcegraph/cody/pull/468). Together these checks provide comprehensive coverage across both V1 telemetry and V2 telemetry.

## Test plan
Tested locally via `pnpm test:e2e` and ensure that all tests are logging the correct events and test results are flowing into dedicated testing pubsub topic.


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -653,6 +653,12 @@ export class SourcegraphGraphQLAPIClient {
         for (const event of events) {
             this.anonymizeTelemetryEventInput(event)
         }
+
+        if (process.env.CODY_TESTING === 'true') {
+            return new MockServerTelemetryExporter(
+                this.anonymousUserID || 'AnonymousTestUserId'
+            ).exportEvents(events)
+        }
```
bobheadxi, line 661: nit - I would prefer that MockServerTelemetryExporter be used directly instead if possible (replacing the GraphQL exporter entirely if possible). This ensures that this exporter doesn't grow to become difficult to understand with various test scenarios (as v1 logging did)

bobheadxi (approved):
overall LGTM, but I would prefer another way to set up the mock exporter instead of hijacking production code (graphql exporter) with testing mechanisms

abeatrix (approved):
YAY! Does this mean we can start removing `telemetryService`?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -653,6 +653,12 @@ export class SourcegraphGraphQLAPIClient {
         for (const event of events) {
             this.anonymizeTelemetryEventInput(event)
         }
+
+        if (process.env.CODY_TESTING === 'true') {
+            return new MockServerTelemetryExporter(
+                this.anonymousUserID || 'AnonymousTestUserId'
```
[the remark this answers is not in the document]
akalia25: @abeatrix you raise a good point! - I think it would be valuable to continue capturing the actual `anonymousUserId` from our tests. Having the actual anonymous ID would allow us to validate that the ID persists correctly across the test session. And then, falling back to a value like `AnonymousTestUserId` can be used as an indicator that something may be wrong with the ID generation 🤔

akalia25:
> YAY! Does this mean we can start removing `telemetryService`?

We are VERY close!!! Currently, we only need to log v2 events now 🚀, but for removing, we're holding off on that for now, just so we can remove it all in one big push :)

lib/shared/src/sourcegraph-api/graphql/client.ts
akalia25: Thanks for the recommendation @bobheadxi!! If you don't think this is a blocking change, I'd love to work on the above on a separate PR, and with someone who has more expertise than me on GraphQL 😅

bobheadxi: It's not about expertise on GraphQL :) It's about using `MockServerTelemetryExporter` elsewhere - `MockServerTelemetryExporter` just implements our exporter interface, and it's meant to be plugged in at the recorder level. This makes it much easier to reason about when and where things go to a mock exporter rather than a real one, instead of having to deal with many levels of switches

Example: the default one _also_ checks for `CODY_TELEMETRY_EXPORTER`. One might read this now and meet surprise behaviour all the way down in the GraphQL client as well https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts?L58-60

bobheadxi: Why don't we just update this condition here?

bobheadxi (changes requested):
Sorry, I think https://github.com/sourcegraph/cody/pull/3974/files#r1583747661 might be a blocking change request after all, since the default one already has a condition for testing: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts?L58-60 

_And_ the recorder initialization func _also_ has conditions for testing: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@7d385240264090f7bfb0d502a76049fd6a05de58/-/blob/vscode/src/services/telemetry-v2.ts?L39-58

I'd prefer not to add a third level of "testing-or-not" toggle here 😬

lib/shared/src/sourcegraph-api/graphql/client.ts
akalia25: Thanks for explaining this in detail. Removing `MockTelemetryExporter` here and confirmed that its being set correctly on debug mode [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@7d385240264090f7bfb0d502a76049fd6a05de58/-/blob/vscode/src/services/telemetry-v2.ts?L39)

bobheadxi (approved):
thanks for addressing feedback :)

PR #3975, opened 2024-04-29
adding finetuned model for code completions

hitesh-1997:
## Context
1. This PR adds the fine-tuning code completions model for internal users. The offline evaluation is done on HumanEval dataset and the results are linked in the [RFC](https://docs.google.com/document/d/1myMceHu7x8gh0B9eb_kRE719AGqVHfZ3py9k_3luym8/edit#heading=h.trqab8y0kufp)
2. This PR particularly adds a fine-tuned version of the Mixtral model hosted by the fireworks.
3. Cody Gateway PR: https://github.com/sourcegraph/sourcegraph/pull/62300 

## Test plan
Manual testing
1. Added my username in the feature flag: `cody-autocomplete-finetuned-model `
2. Used debug console and cody autotrace view to check the context and completions.

Screenshot 2024-05-02 at 12 37 04 AM

### DIFF
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -153,12 +153,19 @@ async function resolveDefaultProviderFromVSCodeConfigOrFeatureFlags(
         return { provider: configuredProvider }
     }
 
-    const [starCoder2Hybrid, starCoderHybrid, llamaCode13B, claude3] = await Promise.all([
-        featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoder2Hybrid),
-        featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoderHybrid),
-        featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLlamaCode13B),
-        featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteClaude3),
-    ])
+    const [starCoder2Hybrid, starCoderHybrid, llamaCode13B, claude3, finetunedModel] = await Promise.all(
+        [
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoder2Hybrid),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoderHybrid),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLlamaCode13B),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteClaude3),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteFineTunedModel),
+        ]
+    )
+
+    if (finetunedModel) {
+        return { provider: 'fireworks', model: 'fireworks-completions-fine-tuned' }
+    }
 
     if (llamaCode13B) {
         return { provider: 'fireworks', model: 'llama-code-13b' }
--- a/vscode/src/completions/providers/fireworks.ts
+++ b/vscode/src/completions/providers/fireworks.ts
@@ -323,7 +323,7 @@ class FireworksProvider extends Provider {
                 Your goal is to perform code completion for the following code, keeping in mind the rest of the code and the file meta data.
                 Metadata details: filename: ${filename}. The code of the file until where you have to start completion:
             `
-            return ps`${intro} \n ${suffix} \n ${fixedPrompt} \n ${prefix}`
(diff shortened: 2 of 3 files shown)


vscode/src/completions/providers/fireworks.ts
```diff
@@ -312,7 +318,13 @@ class FireworksProvider extends Provider {
             // c.f. https://github.com/facebookresearch/codellama/blob/main/llama/generation.py#L402
             return ps`<PRE> ${intro}${prefix} <SUF>${suffix} <MID>`
         }
-
+        if (this.model === 'fireworks-completions-fine-tuned') {
+            const fixedPrompt = ps`You are an expert in writing code in many different languages.
+                Your goal is to perform code completion for the following code, keeping in mind the rest of the code and the file meta data.
+                Metadata details: filename: ${filename}. The code of the file until where you have to start completion:
+            `
+            return ps`${intro} \n ${suffix} \n ${fixedPrompt} \n ${prefix}`
```
philipp-spiess, line 326: Did we train the model on this prompt? It seems super confusing otherwise to have the suffix come in before the instruction 🤔

hitesh-1997: Hi @philipp-spiess 
I wrongly added the suffix. Actually, this model was only trained on the prefix and the `fixedPrompt`. For this initial version, we are evaluating the quality based on the internal users and based on the results, can take further decisions.
For the subsequent models, we will also add the suffix and some `fim_rate`.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -172,7 +178,7 @@ class FireworksProvider extends Provider {
         const languageConfig = getLanguageConfig(this.options.document.languageId)
 
         // In StarCoder we have a special token to announce the path of the file
-        if (!isStarCoderFamily(this.model)) {
+        if (!isStarCoderFamily(this.model) && this.model !== 'fireworks-completions-fine-tuned') {
```
philipp-spiess, line 181: `isStarCoderFamily` will already be false here (it requires `this.model` to start with `starcoder` to be true) so you don't need that check

hitesh-1997: Its actually `! (i.e. not) isStarCoderFamily` so the first condition will be `true`.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -312,7 +318,13 @@ class FireworksProvider extends Provider {
             // c.f. https://github.com/facebookresearch/codellama/blob/main/llama/generation.py#L402
             return ps`<PRE> ${intro}${prefix} <SUF>${suffix} <MID>`
         }
-
+        if (this.model === 'fireworks-completions-fine-tuned') {
+            const fixedPrompt = ps`You are an expert in writing code in many different languages.
+                Your goal is to perform code completion for the following code, keeping in mind the rest of the code and the file meta data.
+                Metadata details: filename: ${filename}. The code of the file until where you have to start completion:
+            `
+            return ps`${intro} \n ${fixedPrompt}\n${prefix}`
```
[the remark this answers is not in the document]
hitesh-1997: it totally makes sense to use suffix, but just wanted to check the current quality and based on the same we will for sure add suffix in fine-tuning.
For starters, we can only add couple of people in the `featureflag` to ensure bad quality doesn't impact.

vscode/src/completions/providers/fireworks.ts
philipp-spiess: oh gotcha sorry I got confused in the negations here

vscode/src/completions/providers/create-provider.ts
```diff
@@ -153,12 +153,19 @@ async function resolveDefaultProviderFromVSCodeConfigOrFeatureFlags(
...
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteStarCoderHybrid),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteLlamaCode13B),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteClaude3),
+            featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyAutocompleteFineTunedModel),
+        ]
+    )
+
+    if (finetunedModel) {
+        return { provider: 'fireworks', model: 'fireworks-completions-fine-tuned' }
+    }
```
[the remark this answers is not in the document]
hitesh-1997: that totally makes sense.
For the initial part, we want to check on internal users and that's why added at the top to just collect the data for the same. I am assuming there could be some iterations if we find that current quality doesn't look good and hence added at top.

Would it be okay to change this condition when we actually want to move to A/B test and try on users ?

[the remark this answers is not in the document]
philipp-spiess: sounds good

PR #3985, opened 2024-04-30
noodle: Explain Code History command

keegancsmith:
This is a super experimental command which relies on the output of "git log". It uses git logs support to understand function boundaries such that we can give commit history only for commits which change a function's body.

I spent a little while working out how to use things like the git extension API/etc, but this doesn't really work along those lines. However, this sort of functionality is implementable in gitserver so this shouldn't limit us there.

Test Plan: right clicked on a few functions and eye balled the output.

Demo at https://www.loom.com/share/30df43f6170a4a85bfba3c176dc1a680


sqs (approved):
LGTM in that it just hooks into the existing safe context provider hooks we built

thenamankumar:
This is a test comment on a draft PR. Ignore.

vscode/package.json
```diff
@@ -344,6 +344,12 @@
         "title": "Usage Examples",
         "when": "cody.activated && editorTextFocus && config.cody.experimental.noodle && (editorLangId === typescript || editorLangId === typescriptreact)"
       },
+      {
```
thenamankumar, line 347: This is a test review comment

toolmantim: So meta

keegancsmith:
Right I am having CI issues because I am importing `node:child_process`. I see we do this in other places (eg localcontext), so I likely need to work out how to make this behave properly.

vscode/src/commands/execute/explain-history.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    const prompt = ps`Explain the history of the function \`${logArguments.symbolText}\`.`
+    const contextFiles: ContextItem[] = [
+        {
+            type: 'file',
+            content: logResult.stdout,
+            title: 'Terminal Output',
+            uri: vscode.Uri.file('terminal-output'),
+            source: ContextItemSource.Terminal,
```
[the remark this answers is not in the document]
thenamankumar: We can probably introduce a new source `git`?

vscode/src/commands/execute/explain-history.ts
```diff
@@ -0,0 +1,234 @@
...
+    const logResult = await spawnAsync('git', ['log', ...logArguments.logArgs], {
+        cwd: logArguments.cwd,
+    })
+
+    if (logResult.code !== 0) {
+        logDebug('explainHistoryCommand', 'git log failed', JSON.stringify(logResult))
+        return parseGitLogFailure(logArguments, logResult)
+    }
+
+    const prompt = ps`Explain the history of the function \`${logArguments.symbolText}\`.`
```
thenamankumar, line 53: Does this only work for a function? Have you tested other more defined prompts as well?

vscode/src/commands/execute/explain-history.ts
```diff
@@ -0,0 +1,234 @@
...
+    const contextFiles: ContextItem[] = [
+        {
+            type: 'file',
+            content: logResult.stdout,
+            title: 'Terminal Output',
```
thenamankumar, line 58: Probably make the title as `Execute: ['log', ...logArguments.logArgs].join(" ")`? It will show the user what exactly we ran to gather the context. 

The context item title is rendered based on `item.type` inside `contextItemMentionNodeDisplayText`. 

For a file having title it will render `@${contextItem.title}`.

vscode/src/commands/execute/explain-history.ts
```diff
@@ -0,0 +1,234 @@
...
+    }
+}
+
+export async function executeExplainHistoryCommand(
+    args: Partial<CodyCommandArgs> = {}
+): Promise<ChatCommandResult | undefined> {
+    return wrapInActiveSpan('command.explain-history', async span => {
+        span.setAttribute('sampled', true)
+        logDebug('executeExplainHistoryCommand', 'executing', args)
+        telemetryRecorder.recordEvent('cody.command.explain-history', 'executed', {
```
thenamankumar, line 79: This is unrelated, but it reminds me if there is enough logging present for the at-mention context providers we have added recently.

thenamankumar (approved):
LGTM left a few non-blocking comments.

vscode/src/commands/execute/explain-history.ts
keegancsmith: In older versions of the code the `Source` field on a contextitem is mentioned to be used for telemetry purposes. So we likely have some telemetry around that. Additionally for this command and quinn's usage examples command  he is using OTEL and telemetryRecorder.

vscode/src/commands/execute/explain-history.ts
keegancsmith: It is a heuristic in git that is based on simple regular expressions it has built in. It targets functions generally, but for example it doesn't pick up methods in typescript. To do this correctly we should be using our built in code smarts to find the correct region to search. This shouldn't be that hard to do, but this was easier for demonstrating the idea behind this. I think if we want to take this idea further that is one of the things we will have to do.

vscode/src/commands/execute/explain-history.ts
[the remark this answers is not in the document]
keegancsmith: Yeah I will timebox improving this. For example I want to make it possible that when you click on the context link it shows something reasonable.

vscode/src/commands/execute/explain-history.ts
thenamankumar: Gotcha

PR #3986, opened 2024-04-30
fix: update ContextItemSource for usage examples

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1714489870131939
RE: https://github.com/sourcegraph/cody/pull/3939

image

 This pull request addresses an issue where the context items for usage examples have the `source` property set to `ContextItemSource.Search`, which is reserved for results returned by `symf`. However, `symf` has a different interface than results returned from `Remote Search`, which is associated with `ContextItemSource.Unified`. As a result, the results returned from `symf` are displayed as `undefined` in the UI.

Changes:

- Fixes the issue of `symf` results displaying as `undefined` in the UI for usage examples
- Unifies the source for usage examples to `ContextItemSource.Unified`, ensuring correct display
  - Updates the `source` property of `ContextItem` for `usage-examples.ts` from `ContextItemSource.Editor` and `ContextItemSource.Search` to `ContextItemSource.Unified` 
- Updated tooltip from `Enterprise Search` to `Remote Search` to avoid confusion.

## Test plan

- Verify that usage examples from `symf` are correctly displayed with the unified source, resolving the `undefined` display issue
- Ensure that the `FileLink` component renders correctly for remote search results from usage example

### After

Results from example-usage are rendering correctly:

image

Results from symf are rendering correctly:

image

### Before

image

### DIFF
--- a/vscode/src/commands/execute/usage-examples.ts
+++ b/vscode/src/commands/execute/usage-examples.ts
@@ -89,7 +89,7 @@ export async function executeUsageExamplesCommand(
             type: 'file',
             uri: doc.uri,
             range: snippetRange,
-            source: ContextItemSource.Editor,
+            source: ContextItemSource.Unified,
         })
 
         try {
@@ -207,7 +207,7 @@ async function symbolContextItems(
                     ...use,
                     type: 'file',
                     title: uriBasename(use.uri),
-                    source: ContextItemSource.Search,
+                    source: ContextItemSource.Unified,
                 }) satisfies ContextItem
         ),
     ]
--- a/vscode/src/commands/execute/usage-examples.ts
+++ b/vscode/src/commands/execute/usage-examples.ts
@@ -89,7 +89,7 @@ export async function executeUsageExamplesCommand(
             type: 'file',
             uri: doc.uri,
             range: snippetRange,
-            source: ContextItemSource.Unified,
+            source: ContextItemSource.Editor,
         })
 
         try {
--- a/vscode/src/commands/execute/usage-examples.ts
+++ b/vscode/src/commands/execute/usage-examples.ts
@@ -199,6 +199,7 @@ async function symbolContextItems(
                     ...item,
                     type: 'file',
                     range: item.range ? expandRangeByLines(toVSCodeRange(item.range)!, 10) : undefined,
+                    source: ContextItemSource.Unified,
                 }) satisfies ContextItem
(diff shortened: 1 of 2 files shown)


vscode/webviews/Components/FileLink.tsx
```diff
@@ -47,15 +46,12 @@ export const FileLink: React.FunctionComponent<FileLinkProps & { className?: str
...
+    if (source === 'unified') {
         // Remote search result.
         const repoShortName = repoName?.slice(repoName.lastIndexOf('/') + 1)
         const pathToDisplay = `${repoShortName} ${title}`
         pathWithRange = range ? `${pathToDisplay}:${displayLineRange(range)}` : pathToDisplay
-        tooltip =
-            source === 'unified'
-                ? `${repoName} @${revision}\nincluded via Enhanced Context (Enterprise Search)`
-                : ''
+        tooltip = `${repoName} @${revision}\nincluded via Enhanced Context (Remote Search)`
```
abeatrix, line 54: @sqs @thenamankumar i'm not sure if this is the right change. Are results for usage example expected not to display as pathWithRange?

thenamankumar: I would prefer the tooltip to show the exact SG URL with range and revision.

@toolmantim what should be the tooltip for context items fetched from remote repositories using remote search?

Whatever we decide here, we need to make package context items render the same way. I'll handle that.

thenamankumar (approved):
Approving as it fixes the current bug. Still need to figure out the right tooltip for the context items. But we can handle it separately as a follow-up.

vscode/src/commands/execute/usage-examples.ts
```diff
@@ -89,7 +89,7 @@ export async function executeUsageExamplesCommand(
             type: 'file',
             uri: doc.uri,
             range: snippetRange,
-            source: ContextItemSource.Editor,
```
thenamankumar, line 92: Why do we need to change this? Can we not provide the required props for the `ContextItemSource.Editor` to render correctly, or is that source deprecated now?

At other places when we auto include the currently open file, what is the context item source?

abeatrix: No, I simply missed the fact that this comes from the editor instead of remote search. Will revert 😀

PR #3989, opened 2024-04-30
chore: upgrade to biome@1.7.2

bobheadxi:
Upgrades Biome, as Data Team wants to use https://biomejs.dev/linter/rules/no-restricted-imports/, which is not available in the version of Biome we are currently using.

It seems like quite a lot has changed within Biome - I've migrated the rules that have been promoted out of `nursery` into their modern equivalents.

## Test plan

`pnpm run biome` exits with status 0 locally, required CI passes

### DIFF
--- a/vscode/src/completions/text-processing/parse-completion.ts
+++ b/vscode/src/completions/text-processing/parse-completion.ts
@@ -1,5 +1,5 @@
 import type { TextDocument } from 'vscode'
-import type { Point, Tree, default as Parser } from 'web-tree-sitter'
+import type { default as Parser, Point, Tree } from 'web-tree-sitter'
 
 import { addAutocompleteDebugEvent } from '../../services/open-telemetry/debug-utils'
 import { asPoint, getCachedParseTreeForDocument } from '../../tree-sitter/parse-tree-cache'
@@ -111,7 +111,6 @@ function pasteCompletion(params: PasteCompletionParams): PasteCompletionResult {
         docContext: {
             position,
             currentLineSuffix,
-            // biome-ignore lint/nursery/noInvalidUseBeforeDeclaration: it's actually correct
             positionWithoutInjectedCompletionText = position,
             injectedCompletionText = '',
         },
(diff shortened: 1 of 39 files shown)


vscode/src/completions/text-processing/parse-completion.ts
```diff
@@ -111,7 +111,6 @@ function pasteCompletion(params: PasteCompletionParams): PasteCompletionResult {
         docContext: {
             position,
             currentLineSuffix,
-            // biome-ignore lint/nursery/noInvalidUseBeforeDeclaration: it's actually correct
             positionWithoutInjectedCompletionText = position,
```
bobheadxi, line 114: No longer an error

vscode/webviews/Components/FileLink.tsx
```diff
@@ -71,7 +71,6 @@ export const FileLink: React.FunctionComponent<FileLinkProps & { className?: str
     return (
         <div className={classNames(styles.linkContainer, className)}>
             {isTooLarge && <i className="codicon codicon-warning" title={WARNING} />}
-            {/* biome-ignore lint/a11y/useValidAnchor: The onClick handler is only used for logging */}
```
bobheadxi, line 74: No longer an error

vscode/webviews/OnboardingExperiment.tsx
```diff
@@ -40,7 +40,6 @@ const WebLogin: React.FunctionComponent<
                 </a>
             </li>
             <li>
-                {/* biome-ignore lint/a11y/useValidAnchor: can fix with a lot of CSS but not a priority */}
```
bobheadxi, line 43: No longer an error

vscode/webviews/SearchPanel.tsx
```diff
@@ -269,6 +269,7 @@ export const SearchPanel: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }>
             <div className={styles.searchResultsContainer}>
                 {results.map((result, fileIndex) => (
                     <>
+                        {/* biome-ignore lint/correctness/useJsxKeyInIterable: Each of the possible divs below have an appropriate key. */}
                         {/* File result */}
                         <div
```
bobheadxi, line 274: I think Biome expects a `<div>` with a unique key wrapping the options below - I'm worried about messing up any styling here, and each of the possible divs below have keys

sqs (approved):

philipp-spiess:
FYI this commit on `main` started to break `biome` for us :( see https://sourcegraph.slack.com/archives/C05AGQYD528/p1714639734553859

PR #3990, opened 2024-04-30
telemetry-v2: add Biome lint rule and update telemetry architecture docs

bobheadxi:
This change:

- Adds a `biome.json` lint rule using the new [`noRestrictedImports`](https://biomejs.dev/linter/rules/no-restricted-imports/) (enabled by Biome upgrade in #3989) to guard against usage of legacy V1 telemetry modules. The rule links to the `ARCHITECTURE.md` Telemetry reference for guidance.
  - The intent is to ban new usages - @akalia25 will update this PR to add [no-lint comments] ✅ (https://biomejs.dev/linter/#ignoring-code) to existing imports, and make removals in follow-ups
- Updates the `ARCHITECTURE.md` Telemetry reference with the new state of things (no more V1 telemetry) and also expands it with some additional details.

## Test plan

Without ignores:

image

With ignores, CI should pass.


vscode/src/services/telemetry-v2.ts
```diff
@@ -14,6 +14,7 @@ import { logDebug } from '../log'
 
 import type { AuthProvider } from './AuthProvider'
 import { localStorage } from './LocalStorageProvider'
+// biome-ignore lint/nursery/noRestrictedImports: Deprecated v1 telemetry used temporarily to support existing analytics.
 import { getExtensionDetails } from './telemetry'
```
bobheadxi, line 18: FYI `getExtensionDetails` is also used by V2, when you get around to this one I think you'll want to just move this helper into the `telemetry-v2` file

biome.json
```diff
@@ -8,7 +8,18 @@
     "rules": {
       "recommended": true,
       "nursery": {
-        "noDuplicateJsonKeys": "error"
+        "noDuplicateJsonKeys": "error",
+        "noRestrictedImports": {
+          "options": {
+            "paths": {
```
valerybugakov, line 14: To remove the schema warning because the "level" property is marked as required.

```suggestion
          "options": {
            "level": "error",
            "paths": {
```

valerybugakov (approved):
Nice!

biome.json
akalia25: Thanks @valerybugakov, could you show me where you are getting a schema warning? I tested this change locally and am getting an error after adding "level" as a field and not seeing "level" as a field on the [docs](https://biomejs.dev/linter/rules/no-restricted-imports/) 
image

valerybugakov: The warning in the `biome.json` file:

Screenshot 2024-05-09 at 09 51 32

valerybugakov: The schema file `node_modules/@biomejs/biome/configuration_schema.json` with the `level` property marked as required: 

Screenshot 2024-05-09 at 09 52 23

valerybugakov: Ah, it should be on the same level as the `options` field.

akalia25: ah, that explains why biome was failing for me when I added that. I wasn't seeing the warning in my `biome.json` might be a misconfiguration on my ide 😅. Just updated to include `level`, thank you! 🙇

PR #3991, opened 2024-04-30
Search: Move Natural Language Search (Beta) to QuickPick

abeatrix:
CLOSE https://github.com/sourcegraph/cody-issues/issues/21

image

This pull request replaces the existing search webview with a new quick pick interface, and remove NLS search box from the sidebar.

## Changes

- Removed the `cody.search` webview and associated `SearchViewProvider` methods for resolving and handling webview messages.
- Removed the webview associated with the NLS sidebar.
- Added a new `cody.symf.search` command to trigger the search functionality.
- Updated the `CommandMenuOption` to include a new `search` option for accessing the search feature.
- Modified the `SearchViewProvider` to use a quick pick interface instead of a webview:
  - The `showInputBox` method prompts the user for a search query using an input box.
  - The `onDidReceiveQuery` method processes the user's query, retrieves search results, and displays them in a quick pick dropdown.
  - Search results are displayed with file names, line numbers, and snippet previews.
  - Selecting a search result opens the corresponding file and highlights the matching lines.
- Removed unused imports and code related to the previous webview implementation.
- Updated tests to reflect the removal of the search webview.

## Test plan

- Manually tested the new search functionality to ensure it works as expected.
- Verified that search results are accurately displayed in the quick pick dropdown.
- Confirmed that selecting a search result opens the correct file and highlights the matching lines.
- Verified search box has been removed from sidebar 

image

image

### Before

image

### After

image

#### Changes applied from code review

- [x]  Currently it's not discoverable. Let's add it to the Cody Commands and treeview list, after "Find Code Smells"?
  - image
- [x]  Let's rename "Search (Beta)" to "Search Code (Beta)"
  - image
- [x]  Given the quickpick dismisses when unfocused
- [x] While we're here, could we improve the following labels to be:
  - [x]  Start a new chat with this message
  - [x] Start a new code edit with these instructions
  - [x] Start a new natural language search with this text
  - image
- [x] When there's results, can we update the placeholder to say "Filter search results. Press ESC to close."
  - image
- [x] Is there a way to add an empty row/state if they filter and there's nothing? I don't think so? If not, don't worry… but if there is, it might be nice.
  - image
- [ ] Another nice-to-have… it might be nice to open the files without having to hit enter, so you can just go up and down in the list, instead of down then enter, down then enter.
- [x] Looks like we may have an old command still lingering around?
  - Removed
- [x] And finally… here's those errors I spotted
  - Updated to send error to debug log instead of system message

## Demo

https://github.com/sourcegraph/cody/assets/68532117/5c2c032f-7435-494a-98a8-d50034f2b1ff


vscode/src/search/SearchViewProvider.ts
```diff
@@ -201,88 +200,23 @@ export class SearchViewProvider implements vscode.WebviewViewProvider, vscode.Di
...
-                }
-                break
-            }
+    private async showInputBox(): Promise<void> {
+        const input = await vscode.window.showInputBox({
+            title: 'Natural Language Search (Beta)',
+            prompt: 'Search for code using a natural language query, such as "password hashing", "connection retries", a symbol name, or a topic.',
+        })
+        if (input) {
+            const query = PromptString.unsafe_fromUserQuery(input.trim())
```
philipp-spiess, line 211: Unsafe usage looks good (this is just moved from a different file)

toolmantim (commented):
That's a lot of code removed! 

I think this is working pretty well… probably just needs a few tweaks and it's ready to go. I still get the occasional error communicate with symf, but that's unrelated. (see screenshots at the bottom)

1. Currently it's not discoverable. Let's add it to the Cody Commands and treeview list, after "Find Code Smells"?

2. Let's rename "Search (Beta)" to "Search Code (Beta)"

3. Given the quickpick dismisses when unfocused, let's remove the "(Press 'Enter' to confirm or 'Escape' to cancel)"

CleanShot 2024-05-01 at 16 18 29@2x

4. While we're here, could we improve the following labels to be:
* Start a new chat with this message
* Start a new code edit with these instructions
* Start a new natural language search with this text

CleanShot 2024-05-01 at 16 03 30@2x

5. When there's results, can we update the placeholder to say "Filter search results. Press ESC to close."

CleanShot 2024-05-01 at 16 04 21@2x

6. Is there a way to add an empty row/state if they filter and there's nothing? I don't think so? If not, don't worry… but if there is, it might be nice.

CleanShot 2024-05-01 at 16 04 50@2x

7. Another nice-to-have… it might be nice to open the files without having to hit enter, so you can just go up and down in the list, instead of down then enter, down then enter.

8. Looks like we may have an old command still lingering around?

CleanShot 2024-05-01 at 16 17 43@2x

CleanShot 2024-05-01 at 16 17 32@2x

And finally… here's those errors I spotted:

CleanShot 2024-05-01 at 16 14 35@2x

abeatrix:
> "(Press 'Enter' to confirm or 'Escape' to cancel)"

Yes to all of them except number 3. "(Press 'Enter' to confirm or 'Escape' to cancel)" is added by the editor so I don't think we can remove it.

For the last item that's an existing issue like you mentioned, but I'll see if we can console log it instead of system popup since that's not a user friendly message!

abeatrix:
@toolmantim 
- [x]  Currently it's not discoverable. Let's add it to the Cody Commands and treeview list, after "Find Code Smells"?
  - image
- [x]  Let's rename "Search (Beta)" to "Search Code (Beta)"
  - image
- [x]  Given the quickpick dismisses when unfocused
- [x] While we're here, could we improve the following labels to be:
  - [x]  Start a new chat with this message
  - [x] Start a new code edit with these instructions
  - [x] Start a new natural language search with this text
  - image
- [x] When there's results, can we update the placeholder to say "Filter search results. Press ESC to close."
  - image
- [x] Is there a way to add an empty row/state if they filter and there's nothing? I don't think so? If not, don't worry… but if there is, it might be nice.
  - image
- [ ] Another nice-to-have… it might be nice to open the files without having to hit enter, so you can just go up and down in the list, instead of down then enter, down then enter.
- [x] Looks like we may have an old command still lingering around?
  - Removed
- [x] And finally… here's those errors I spotted
  - Updated to send error to debug log instead of system message

toolmantim (approved):
Nice work!!

👍🏼 from a design POV

PR #3992, opened 2024-04-30
Add Github as at-mention context source for issues & pull requests

thenamankumar:
This PR adds an experimental at-mention GitHub Context to Cody. This will allow users to feed in GitHub PR or issue context in chat, including state, description, comments, reviews, diff and more. 

To enable GitHub context, create an access token from https://github.com/settings/tokens/new?scopes=repo and save it in `cody.experimental.GitHub.accessToken` settings. Make sure that `cody.experimental.noodle` is set to `true`. This is for internal use only and is highly experimental. 

At-mentioning supports the following formats:
- `@<github or gh>:<pull or pr>:1234`
- `@<github or gh>:<pull or pr>:sourcegraph/cody/1234`
- `@<github or gh>:issue:1234`
- `@<github or gh>:issue:sourcegraph/cody/1234`

Upcoming improvement:
- Pre-populate suggestions with recent/assigned issues/PRs.
- Better Auth flow. Maybe use auth token from the GitHub extension. 
- GitHub context for specific commits to support "Explain Code History"
- Persist at-mentioned context items in followup questions.

## Test plan

Demo: https://www.loom.com/share/3b6633f6576f49c793cca95b45d34fb0?sid=17f4b891-5b29-4b99-9a89-c51724d6f104


lib/shared/src/mentions/api.ts
```diff
@@ -42,7 +44,11 @@ export interface ContextMentionProvider<ID extends ContextMentionProviderID = Co
      *
      * {@link query} omits the `@` but includes the trigger prefix from {@link triggerPrefixes}.
      */
-    queryContextItems(query: string, signal?: AbortSignal): Promise<ContextItemFromProvider<ID>[]>
+    queryContextItems(
+        query: string,
+        props: ContextItemProps,
```
keegancsmith, line 49: I'm not sure of the better way to do this, but it feels like there is a better way to be able to pass this in / query this?

lib/shared/package.json
```diff
@@ -19,13 +19,15 @@
   },
   "dependencies": {
     "@microsoft/fetch-event-source": "^2.0.1",
+    "@octokit/core": "^6.1.2",
```
keegancsmith, line 22: our dependency list is quite small for `cody-shared`, this might not be seen as a good thing to introduce. I suspect you need to refactor this such that we have a provider which your mentions provider uses. I've seen this pattern used in a few places, and it makes it possible to swap out the provider when running on web/etc.

For example see the runShell function in the commandprovider class.

keegancsmith (commented):
> To enable GitHub context, create an access token from https://github.com/settings/tokens/new?scopes=repo and save it in cody.experimental.GitHub.accessToken settings. Make sure that cody.experimental.noodle is set to true. This is for internal use only and is highly experimental.

Is it not possible to somehow use the github extension's auth if the user has it? This may be a smoother experience.

Take my inline review below with a grain of salt. I'd be interested in Quinn's perspective. My main concern is around noodle stuff having a minimal impact on the rest of cody code. I don't know enough about our cody codebase yet to know if it is  possible to make a "smaller" change to the rest of cody.

lib/shared/package.json
thenamankumar: As `runShell` is very specific to the env and client, it can not be put in `lib/shared`. But `octokit` is as generic as possible. It works both on the web and node. We should treat it as other deps like `highlight.js`, `lodash`, `date-fns`. 

Secondly, as this is highly experimental and might take a totally different shape later, I wish to keep the surface area the minimal.

thenamankumar:
> Is it not possible to somehow use the github extension's auth if the user has it? This may be a smoother experience.

Sure, we should discuss the best way to do GitHub auth, but for this PR, I have kept it really simple, as it is only for internal use, and I wish to first get the basic plumbing done and then spend time on a better auth flow. 

I think we can work on a better auth flow separate from this, as an improvement. wdyt @keegancsmith ?

lib/shared/src/mentions/api.ts
thenamankumar: I created `ContextItemProps` because we need a way to inject even more stuff like `graphqlClient`, etc.

What if we want to generate a context item by running a shell command, we will need a way to inject `runShell` and other deps.

wdyt?

sqs (approved):

keegancsmith (approved):

PR #3996, opened 2024-05-01
Alt/Opt+Enter submits a chat message without enhanced context

sqs:
Holding <kbd>Alt</kbd> or <kbd>Opt</kbd> disables enhanced context, and pressing <kbd>Enter</kbd> while holding that key will submit a chat message without enhanced context (only @-mentions).

This makes it easier to use chat for tasks that do not need enhanced context, such as general code questions and things that only need context from explicitly @-mentioned files.

https://github.com/sourcegraph/cody/assets/1976/a67be857-f70a-4211-bd0f-6f6e9201877e

## Test plan

CI

Also see video: hold down <kbd>Alt</kbd> or <kbd>Opt</kbd> and ensure the checkbox is unchecked, and that it returns to being checked when you stop holding down the key. If the box was unchecked to begin with, holding down the key should not affect it.


abeatrix (approved):
Work great! This is going to be so helpful, thank you!

toolmantim (commented):
This works great! Could we update the `title` on the send button to say "Send without Enhanced Context" when opt/alt is being held?

PR #3997, opened 2024-05-01
add OpenCtx context mention provider

sqs:
Allows any [OpenCtx](https://openctx.org) provider to be used as a `ContextMentionProvider` to provide @-mentionable context.

For example, the [Google Docs OpenCtx provider](https://sourcegraph.com/github.com/sourcegraph/openctx@8b11111e9fccb207828b9b8ec5217a44acaa108d/-/blob/provider/google-docs/index.ts) makes it possible to @-mention Google Docs.

This lets us add more context providers without committing them to the `cody` repo. We can even add them by external HTTP URLs if there is some service that implements the HTTP OpenCtx API.

## Test plan

CI

### DIFF
--- a/vscode/src/context/openctx.ts
+++ b/vscode/src/context/openctx.ts
@@ -1,0 +1,21 @@
+import { type OpenCtxExtensionAPI, setOpenCtxExtensionAPI } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+
+export function exposeOpenCtxExtensionAPIHandle(): void {
+    setOpenCtxExtensionAPI(async (): Promise<OpenCtxExtensionAPI | null> => {
+        const API_VERSION = 1 as const
+
+        const ext = vscode.extensions.getExtension<OpenCtxVSCodeExtensionAPI>('sourcegraph.openctx')
+        if (!ext) {
+            return null
+        }
+        return (await ext.activate()).apiVersion(API_VERSION)
+    })
+}
+
+interface OpenCtxVSCodeExtensionAPI {
+    /**
+     * If this API changes, the version number will be incremented.
+     */
+    apiVersion(version: 1): OpenCtxExtensionAPI
+}
--- a/lib/shared/package.json
+++ b/lib/shared/package.json
@@ -19,6 +19,7 @@
   },
   "dependencies": {
     "@microsoft/fetch-event-source": "^2.0.1",
+    "@openctx/client": "^0.0.8",
     "@opentelemetry/api": "^1.7.0",
     "@sourcegraph/telemetry": "^0.16.0",
     "crypto-js": "^4.2.0",
--- a/lib/shared/src/context/openctx/api.ts
+++ b/lib/shared/src/context/openctx/api.ts
@@ -1,0 +1,39 @@
+import type { Annotation, ItemsParams, ItemsResult } from '@openctx/client'
+import type { TextDocument } from 'vscode'
+import type { RangeData } from '../../common/range'
+
+/**
+ * Copied from OpenCtx's VS Code extension sources.
+ */
+export interface OpenCtxExtensionAPI {
+    getItems(params: ItemsParams): Promise<ItemsResult | null>
+    getAnnotations(doc: Pick<TextDocument, 'uri' | 'getText'>): Promise<Annotation<RangeData>[] | null>
+}
+
+let _getAPI: (() => Promise<OpenCtxExtensionAPI | null>) | undefined
+
+/**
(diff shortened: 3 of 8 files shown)


vscode/src/context/openctx.ts
```diff
@@ -0,0 +1,22 @@
+import { type OpenCtxExtensionAPI, setOpenCtxExtensionAPI } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+
+/**
+ * Get the OpenCtx API exposed by the OpenCtx VS Code extension, if installed.
+ */
+setOpenCtxExtensionAPI(async (): Promise<OpenCtxExtensionAPI | null> => {
```
olafurpg, line 7: Can we move this out of module initialization into an explicit function call? My experience is that it's difficult to maintain code that relies on side-effecting initialization. For example, reordering imports from reformatting the code can silently break behavior.

olafurpg (commented):
Quick comment, let me run this locally before giving a stamp

vscode/src/context/openctx.ts
sqs: yes, good call, will do

olafurpg:
@sqs Is the gdocs provider the only provider that works with this feature? I tried the hello world and link provider from the openctx.org docs but I'm not seeing `@octx` complete anything in chat.

olafurpg:
I followed the docs here and I'm still not getting `@octx` to work in chat https://sourcegraph.com/github.com/sourcegraph/openctx@8b11111e9fccb207828b9b8ec5217a44acaa108d/-/blob/provider/google-docs/README.md

- Installed the OpenCtx extension
- Added settings
```
    "https://openctx.org/npm/@openctx/provider-google-docs": {
      "googleOAuthCredentialsFile": "/Users/olafurpg/dev/client_secret_888411641236-8bg8o02lt1fn8vccihfdlp3q0ej8eucl.apps.googleusercontent.com.credentials.json",
      "googleOAuthClientFile": "/Users/olafurpg/dev/client_secret_888411641236-8bg8o02lt1fn8vccihfdlp3q0ej8eucl.apps.googleusercontent.com.json"
    }
```
- Enabled debug logging, see the following error for gdoc provider
```
[providerClient(https://openctx.org/npm/@openctx/provider-google-docs)]: failed to get provider capabilities: Error: Cannot find module '/v135/node_fs.js'
```

olafurpg (approved):
Stamp to unblock merging. This PR does not seem to regress existing functionality so it should be a safe merge

olafurpg:
When building the Cody extension locally, I see this warning that might be relevant
```
[0] ✘ [ERROR] No matching export in "../lib/shared/dist/index.js" for import "setOpenCtxExtensionAPI"
[0]
[0]     src/context/openctx.ts:1:35:
[0]       1 │ ...OpenCtxExtensionAPI, setOpenCtxExtensionAPI } from '@sourcegraph...
[0]         ╵                         ~~~~~~~~~~~~~~~~~~~~~~
```

sqs:
@olafurpg Thanks!

It looks like adding the gdocs provider via https://openctx.org/npm/@openctx/provider-google-docs is broken because it uses node's `fs` and the on-the-fly bundler that is used (https://esm.sh) shims node `fs` in a way that breaks. I got it working when I pointed it to the gdocs provider on the file system, but I'll look into this.

I also just made it so the hello world provider works for at-mentions:

image

This just requires the following VS Code settings:

```json
"openctx.providers": {
    "https://openctx.org/npm/@openctx/provider-hello-world": true,
}
```

olafurpg:
I'm on d7baa7a0fcbe740c5e08ebd26468af6d6f944957, configured the hello world provider

```json
  "openctx.providers": {
    "https://openctx.org/npm/@openctx/provider-hello-world": true
  },
```

I'm still still not getting `@octx` to work 🤔 

CleanShot 2024-05-01 at 10 17 23@2x

sqs:
OK, 2 other (undocumented) things to check:

1. VS Code setting `cody.experimental.noodle` set to true
2. On the prerelease version of the OpenCtx extension

olafurpg:
@sqs I enabled the noodle setting and installed the latest pre-release v0.3.1714489462 but it's still not working for me. I also tried 0.2.5, which is a more recent release.

sqs:
OK strange. I will try on my Mac in the morning and will see if it works there and on some other people's machines.

sqs:
@olafurpg It works for me on macOS on 2 separate devices. I am going to merge this and then we can troubleshoot sometime.

PR #3999, opened 2024-05-01
fix 'incorrect casing' warning in tests for Transcript

sqs:
Fixes the `Warning: <VSCodeButton /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.` message.

## Test plan

CI


olafurpg (approved):
I would not have been able to figure out this diff to fix the warning, although I admit I didn't ask Cody (maybe it would have figured this out). Thank you!

PR #4001, opened 2024-05-01
Inline Edit: add minimized test case for bug report

olafurpg:
Related CODY-1312

The large unrelated diff should not have happened. I'm planning to fix that separately. It looks like the prompts changed but I would have expected the tests to fail if that was the case.

## Test plan

See new test case.


umpox (approved):

olafurpg (commented):
I figured out why the recording diff is so large. The new `ChatColumn.tsx` file got added to enhanced context for other test cases. This is a good example why we need to split tests into smaller files and have stronger isolation in the workspace between tests (example: enhanced context tests need to have a separate workspace that doesn't change)

PR #4003, opened 2024-05-01
Interactive Tutorial: Add telemetry and improve unhappy paths

umpox:
## Description

This PR:
- Improves unhappy path handling for the tutorial
  - Fixes an issue where "Reset Tutorial" did not reset
  - Skips formatting for Edits
  - Continues tutorial even if an edit hits an error state
- Adds logic to open the tutorial automatically on the users first auth
  - Gated behind a feature flag, so we can incrementally enable this for new users
- Adds telemetry
  - For A/B test: enrollment events
  - For tutorial progress
    - On tutorial start
    - On tutorial finish
    - On each step finished (autocomplete, edit, fix, chat)   

Demo:

https://github.com/sourcegraph/cody/assets/9516420/fa8444d6-0a1b-40bb-9ef2-71b1db8cfaaf

## Test plan

**Enrolled**:

1. Force user into A/B test
2. Repeat demo from above video, on a clean instance of VS Code
3. Check correct telemetry is sent, and tutorial functions as expected.
4. Sign out and sign back in. Check tutorial doesn't open automatically for the repeated time

**Control**:

1. Force user out of A/B test
2. Repeat demo from above video, on a clean instance of VS Code
3. Confirm that the tutorial does not show, and the CTA doesn't appear in the sidebar

### DIFF
--- a/vscode/src/tutorial/index.ts
+++ b/vscode/src/tutorial/index.ts
@@ -16,6 +16,7 @@ import {
     getStepContent,
     getStepData,
     initTutorialDocument,
+    resetDocument,
 } from './content'
 import { setTutorialUri } from './helpers'
 import { ChatLinkProvider, ResetLensProvider } from './providers'
@@ -195,7 +196,7 @@ export const registerInteractiveTutorial = async (
 ): Promise<vscode.Disposable[]> => {
     const disposables: vscode.Disposable[] = []
     const documentUri = setTutorialUri(context)
-    const document = await initTutorialDocument(documentUri)
+    let document = await initTutorialDocument(documentUri)
 
     let status: 'stopped' | 'started' | 'starting' = 'stopped'
 
@@ -255,8 +256,7 @@ export const registerInteractiveTutorial = async (
         }),
         vscode.commands.registerCommand('cody.tutorial.reset', async () => {
             stop()
-            // We need to close the tutorial to ensure we get a full reset of the editor
-            await vscode.commands.executeCommand('workbench.action.closeActiveEditor', documentUri)
+            document = await resetDocument(documentUri)
             return start()
         })
     )
--- a/vscode/src/tutorial/index.ts
+++ b/vscode/src/tutorial/index.ts
@@ -1,6 +1,7 @@
 import { FeatureFlag, featureFlagProvider, telemetryRecorder } from '@sourcegraph/cody-shared'
 import * as vscode from 'vscode'
 import { type TextChange, updateRangeMultipleChanges } from '../../src/non-stop/tracked-range'
+import { logSidebarClick } from '../services/SidebarCommands'
 import { logFirstEnrollmentEvent } from '../services/utils/enrollment-event'
 import {
     registerAutocompleteListener,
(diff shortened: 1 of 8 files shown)


vscode/src/tutorial/index.ts
```diff
@@ -186,19 +194,22 @@ export const startTutorial = async (document: vscode.TextDocument): Promise<vsco
...
 
     let cleanup: vscode.Disposable | undefined
     const start = async () => {
+        const enabled = await featureFlagProvider.evaluateFeatureFlag(
+            FeatureFlag.CodyInteractiveTutorial
+        )
+        logFirstEnrollmentEvent(FeatureFlag.CodyInteractiveTutorial, enabled)
+        if (!enabled) {
+            return
+        }
```
umpox, line 211: @chenkc805 Here we are logging an enrolment event:

1. User logs in for the first time
2. We check if they are enrolled in this experiment
3. If they are, log `variant`, if they are not log `control`

We also log some more telemetry to see how far users are getting through the tutorial

abeatrix (commented):
Left a question inline.

I did run into some issues:

1. For the second step, the Edit menu didn't open with the pre-filled instruction:

image

 Reset the tutorial and the pre-instruction came up for me again. Maybe a bug?

2. The quick fix add new lines of code instead of replacing the code, and i can't move on to the next step after that:

https://github.com/sourcegraph/cody/assets/68532117/162ab2af-ffbf-4a5c-a12f-1c09b0eee8a6



image


Other non-block issues / questions / suggestions:

1. I didn't see any autocompletion, and turned out it was because I didn't have it enabled 😅  It'd be nice if we enable it for the users during the tutorial, let them know that it's not enabled and direct them to the setting menu. (same with code action setting)

image



2. Non-blocking issue: I'm still seeing the `A git repository was found in the parent folders of the workspace or the open file(s). Would you like to open the repository?` pop-up when the tutorial started, is this expected?

image

IDEA: instead of depending on users to know where to place their cursor, do you think we can just automatically move their cursor to the next "section" once they are done in their current "section"?

umpox:
Hmmm, @abeatrix Can you try completely deleting your `walkthroughs/cody_tutorial.py` file locally before testing it?

I can't replicate this and I've tried a lot, I think some of the issues may be caused by the fact that for local development it reuses the same file from the repository, instead of the file installed where the extension is 🤔 

Unless there's VS Code specific settings 🤔

umpox:
> Non-blocking issue: I'm still seeing the A git repository was found in the parent folders of the workspace or the open file(s). Would you like to open the repository? pop-up when the tutorial started, is this expected?

This is because it's using the local file from our repo, for users they won't have `.git` in their downloaded extension, so won't see this.

> I didn't see any autocompletion, and turned out it was because I didn't have it enabled 😅 It'd be nice if we enable it for the users during the tutorial, let them know that it's not enabled and direct them to the setting menu. (same with code action setting)

Possibly, it may get kinda tricky though to keep the setting in sync whilst the file is open, and then restore it when closed. I think given we're mainly targeting new users right now we can hold off on this for the time being

abeatrix:
> Looks great. Thank you for implementing the enrollment event as both a v1 and v2 event.
> 
> Just to confirm, the enrollment event fires for both control and treatment, correct? (Apologies if this is clarified elsewhere, I'm on my phone commuting)

yup. and only once in users' life time

chenkc805:
Thanks! LGTM

chenkc805 (approved):

PR #4004, opened 2024-05-01
Cody Agent: make HTTP recording file smaller

olafurpg:
Previously, the recording files were huge because they include the full streaming response. This PR processes the response so that we only include the last completion event in the streaming response.

## Test plan
Green CI


pkukielka (approved):
LGTM

PR #4007, opened 2024-05-01
Backfill v1.16.0 entries to features.json

kalanchan:
backfilling some `v1.16.0` entries

Note: The notes section contains PRs only for now, eventually each submission should be done on the PR itself.

See https://github.com/sourcegraph/cody/pull/3956 for more information on how this will _eventually_ work

## Test plan
n/a

### DIFF
--- a/vscode/features.json5
+++ b/vscode/features.json5
@@ -98,5 +98,44 @@
         },
           tags: ['context']
       },
+      {
+        "name": "NoContextChatSubmissionHotkey",
+        "description": "Alt/Opt+Enter to submit chat without context",
+        "productLine": "all",
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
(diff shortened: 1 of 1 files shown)


vscode/features.json5
```diff
@@ -98,5 +98,44 @@
...
+        "description": "Alt/Opt+Enter to submit chat without context",
+        "productLine": "all",
+        "documentation": [],
+        "editors": {
+          "vscode": {
+            "status": "stable",
+            "notes": "https://github.com/sourcegraph/cody/pull/3996"
+          }
+        },
+          tags: ['chat','context']
```
dominiccooney, line 112: This is being copy-pasted with the wrong indent and quotes that don't match.

We should add prettier enforcement for this file in a follow up change.

dominiccooney (approved):

PR #4010, opened 2024-05-02
Finished up Edit/Retry, including showing prompt for Document Code

steveyegge:
Edit/Retry in JetBrains needed to show the "instruction" (prompt) generated for Document Code; i.e., "Write a brief documentation comment for the selected code. If documentation comments exist..." etc.

I used a pretty quick-and-dirty solution. It just goes along with the EditTask as an optional string for the client to consume, and we send it on every editTask/didUpdate notification. That way it's available if the user chooses to Edit/Retry.

Most of the work for this is in a sourcegraph/jetbrains PR coming up shortly.

## Test plan

The test plan is in TESTING.md.


agent/src/AgentFixupControls.ts
```diff
@@ -51,6 +51,7 @@ export class AgentFixupControls implements FixupControlApplicator {
             state: task.state,
             error: errorToCodyError(task.error),
             selectionRange: task.selectionRange,
+            instruction: task.instruction.toString().trim()
```
abeatrix, line 54:
```suggestion
            instruction: task.instruction?.trim()
```

✋ q: do we need to add ? to check if the instruction string is undefined before we trim it? 

(Ignore me if there was a check somewhere that I missed 🙈)

abeatrix (approved):

agent/src/AgentFixupControls.ts
pkukielka: Why do we need to trim it at all here, shouldn't we do that when it is put into `FixupTask`?
Anyway LGTM

pkukielka (approved):
LGTM

agent/src/AgentFixupControls.ts
steveyegge: good catch; I'll add that.

PR #4011, opened 2024-05-02
Add commit message generator

umpox:
## Description

TODO


RXminuS:
Hahaha @umpox being naughty with the prompt string according to GH bot 😂

I had a quick skim at the Cody ignore stuff and I think the approach for now is to complain to the user if bad context is included; not to filter / mask it.  But I don't think it should be very hard to extend with more options there. I don't think we'll have to do the filtering inside the scm specific code as long as we construct our URIs with the gitroot basepath.

In that case I don't see a blocker for releasing outside experimental as any TODOs on the codyignore affects more than just this feature and could just be fixed as separate PRs.

umpox:
Closed in favour of https://github.com/sourcegraph/cody/pull/4130

PR #4017, opened 2024-05-02
Fix CODY-1327 - get Kotlin bindings working again

olafurpg:
The bindings generation was disabled in our CI a while ago and we have since then introduced several types into the protocol that crash the tool. This PR updates the bindings generation tool so that it generates code for the latest version of the agent protocol.

## Test plan
Green CI


olafurpg:
The CI check is still not running and I can't figure out how to re-enable it. I've requested admin access for the repo to see if there's something disabled there. cc/ @dominiccooney any advice on how to turn this check on?

philipp-spiess (approved):
- see 2500 additions
- just click Approve ✅

PR #4022, opened 2024-05-02
TSC Graph Context: handle property access

olafurpg:
Previously, the `tsc-mixed` context didn't boost the type of the *expression* (aka. qualifier) in property accesses. For example, the property access
```ts
this.helper().CURSOR
```
has the *expression* `this.helper()`, which might have the type `interface Helper { ... }` with relevant members for autocomplete. This PR fixes the problem by effectively doing "Go to type definition" on the expression and including that definition in the context.

Fixes CODY-1331

## Test plan

Green CI. See newly added tests.

### DIFF
--- a/agent/src/graph-context.test.ts
+++ b/agent/src/graph-context.test.ts
@@ -260,7 +260,7 @@ describe.skipIf(isWindows())('Graph Context', () => {
             )
         }, 10_000)
 
-        it.skip('multiple-symbols', async () => {
+        it('multiple-symbols', async () => {
             modelFilter = { model: 'starcoder-16b' }
             await changeFile(
                 mainUri,
@@ -287,7 +287,9 @@ describe.skipIf(isWindows())('Graph Context', () => {
               "autocompletes:
                 - name: starcoder-16b
                   value:
-                    - "  return cars.filter(isNewCar).map(car => car.user)[0]"
+                    - |2-
+                        const newCars = cars.filter(isNewCar)
+                        return newCars[0].user
               prompts:
                 - name: fireworks
                   value:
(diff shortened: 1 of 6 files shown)


agent/src/graph-context.test.ts
```diff
@@ -541,6 +541,189 @@ describe.skipIf(isWindows())('Graph Context', () => {
...
+            import { getter } from './members-indirection'
+
+            function run(): void {
+                getter.indirect()./* CURSOR */
+            }
+            `
+            )
+
+            const text = await autocompletes()
+            // expect(text).includes('validDogSled')
```
philipp-spiess, line 616:
```suggestion
```

agent/src/graph-context.test.ts
```diff
@@ -541,6 +541,189 @@ describe.skipIf(isWindows())('Graph Context', () => {
...
+
+                        //
+
+                        // interface Selector {
+
+                        //   query(params: { isMammal: boolean; animalName: string; }) => { animals: Animal[]; }
+
+                        // }
+
+                        //
```
philipp-spiess, line 647: Still think there's something wrong with how these are formatted. They seem to be terminated at with two `\n\n` 🤔  The code below doesn't seem to do that. I have commented that before but I don't remember if/what your response was. 😐

philipp-spiess (approved):
Really cool!

agent/src/graph-context.test.ts
olafurpg: I think there's a bug in our prompt logic that inserts these blank lines, you can see it in the recordings

https://sourcegraph.com/github.com/sourcegraph/cody@6b39919a9c7242f51f7c5f4cf73e14abc1b07f08/-/blob/agent/recordings/graph-context-starcoder-16b_352709139/recording.har.yaml?L180-197

CleanShot 2024-05-03 at 11 19 07@2x


I was planning to comment on this in Slack. I was reading the `PromptString.join()` logic and I think it's currently difficult to reason about, and I am concerned about the performance of repeated usage of split/join (prompts can become big strings).

olafurpg: [CODY-1364 : Blank lines in autocomplete context prompt](https://linear.app/sourcegraph/issue/CODY-1364/blank-lines-in-autocomplete-context-prompt)

philipp-spiess: Yeah, testing the fix right now. Regarding the performance overhead: Agree and it's something we may want to look into but we should first gather some data to see how big of a problem this really is before we spent time (we'd need to change the datatype for how we store references so we can concatenate faster).

olafurpg: +1 on sampling data for perf overhead for prompt construction

olafurpg:
Failing ubuntu 18 tests are unrelated to this PR

PR #4023, opened 2024-05-02
Remove caching and unnecessery code from didFocus

pkukielka:
## Changes

1. Remove `cursorMove` which I added - I think incorrectly.
2. Remove caching added by @steveyegge which rollbacks state if we receive empty data from client.

@steveyegge, I would rather stumble on the issues you had before that changes (if they are still there, as code is changing constantly) and then fix them properly on the client side. Adding such caching, where we do not fully understand all edge cases when/why it happens, seems a bit dangerous to me as it can break other things in a subtle way.

Of course I'm open for discussion there, I'm not going to merge this untill we all agree on a way forward.
CC @olafurpg 

## Test plan

This is removal of the code which I think is unnecessary (and I'm not sure if it's already doing anything useful).

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -71,7 +71,6 @@ import type {
     GetFoldingRangeResult,
     ProtocolCommand,
     ProtocolTextDocument,
-    Range,
     TextEdit,
 } from './protocol-alias'
 import { AgentHandlerTelemetryRecorderProvider } from './telemetry'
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
LGTM 👍🏻 what was the motivation to add `cursorMove`? I don't see this command handled in our extension. Either way, fine to merge this

PR #4024, opened 2024-05-02
Cody Ignore: Treat unterminated JSON responses as an outdated API error message

dominiccooney:
This unbreaks 5.2.3 treating all files as ignored.

## Test plan

Tested manually: Log in to 5.2.3 and chat with an @-file mention. You should not get a Cody Ignore error message.

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -1029,5 +1029,11 @@ export type LogEventMode =
     | 'all' // log to both dotcom AND the connected instance
 
 function hasOutdatedAPIErrorMessages(error: Error): boolean {
-    return error.message.includes('Cannot query field')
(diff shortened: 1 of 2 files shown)


philipp-spiess (approved):

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1029,5 +1029,11 @@ export type LogEventMode =
     | 'all' // log to both dotcom AND the connected instance
 
 function hasOutdatedAPIErrorMessages(error: Error): boolean {
-    return error.message.includes('Cannot query field')
+    // Sourcegraph 5.2.3 returns an empty string ("") instead of an error message
```
abeatrix, line 1032: Would it help to add the `getSiteVersion()` check to `contextFilters`?

dominiccooney: Yes, we should fix this properly with something like that.

PR #4025, opened 2024-05-02
Check for version number before contextFilter request

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1714664110917129

Check for version number before contextFilter request

## Test plan

Internal: Try connecting to sg04 listed in the linked thread and submit a message with enhanced context enabled

image

image

### Before

image

### DIFF
--- a/lib/shared/package.json
+++ b/lib/shared/package.json
@@ -10,7 +10,11 @@
   },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
-  "files": ["dist", "src", "!**/*.test.*"],
+  "files": [
+    "dist",
+    "src",
+    "!**/*.test.*"
+  ],
   "sideEffects": false,
   "scripts": {
     "build": "tsc --build",
@@ -33,6 +37,7 @@
     "lru-cache": "^10.0.0",
     "marked": "^4.0.16",
     "re2js": "^0.4.1",
+    "semver": "^7.5.4",
     "vscode-uri": "^3.0.7"
   },
   "devDependencies": {
@@ -45,6 +50,7 @@
     "@types/lodash": "^4.14.195",
     "@types/marked": "^5.0.0",
     "@types/node-fetch": "^2.6.4",
+    "@types/semver": "^7.5.0",
     "@types/vscode": "^1.80.0"
   }
 }
--- a/lib/shared/package.json
+++ b/lib/shared/package.json
@@ -10,11 +10,7 @@
   },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
-  "files": [
-    "dist",
-    "src",
-    "!**/*.test.*"
-  ],
+  "files": ["dist", "src", "!**/*.test.*"],
   "sideEffects": false,
   "scripts": {
     "build": "tsc --build",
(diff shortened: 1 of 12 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -602,6 +602,12 @@ export class SourcegraphGraphQLAPIClient {
     }
 
     public async contextFilters(): Promise<ContextFilters> {
+        const { enabled, version } = await this.isCodyEnabled()
+        // CONTEXT FILTERS are only available on Sourcegraph 5.3.3 and later.
+        if (!enabled || version < '5.3.3') {
```
abeatrix, line 607: is 5.3.3 the correct version number here?

dominiccooney, line 607: This is a bit of a dodgy comparison, because version '10.0.0' < '5.3.3' => true.

abeatrix: woops i'd use semver

kalanchan: 🤔 this is a tough question to answer because of how misaligned the releases are. 

from engineering POV - context filters has been rolled out on Sourcegraph versions >`v5.3.3`
from a product/feature POV - **Cody Ignore** should not be a feature available until Sourcegraph versions > `6.0` 

So yeah, I think we should be using 6.0 here. Or are we trying to be compatible with the engineering changes?

philipp-spiess: If the server has support for it since 5.3.3 we should use that. this is a safeguard anyway just to avoid even more obscure errors with even older servers. WDYT @abeatrix?

lib/shared/package.json
```diff
@@ -45,6 +50,7 @@
     "@types/lodash": "^4.14.195",
     "@types/marked": "^5.0.0",
     "@types/node-fetch": "^2.6.4",
+    "@types/semver": "^7.5.0",
     "@types/vscode": "^1.80.0"
   }
-}
+}
```
philipp-spiess, line 56: Formatting here seems way off, was this just `pnpm add`? 🙈  Maybe we can tell pnpm to use biome after modifying the package.json somehow.

abeatrix: lol fixed it!

lib/shared/src/sourcegraph-api/graphql/client.ts
abeatrix: @philipp-spiess 100% agree!

abeatrix:
@philipp-spiess do you know why the agent test is failing when they are passing locally? 
image

philipp-spiess:
🤔 No idea why this happens here but I've seen this before yeah it happens when there are network requests after the assertion and your own machine is fast enough so that they didn't happen when you were doing the recording

Last time I fixed this it was possible to change it to not delete missing recordings and to re-run this a few times but I’m not sure if we'll be able to do that here cc @olafurpg

philipp-spiess:
@abeatrix used this commaned: `CODY_KEEP_UNUSED_RECORDINGS=true CODY_RECORD_IF_MISSING=true pnpm test`

PR #4029, opened 2024-05-02
TSC Graph Context: remove `typescript` from bundle

olafurpg:
Fixes CODY-1294

Previously, the bundle size of the VSC extension was ~21mb instead of ~14mb because of the embedded typescript dependency. This was problematic because parsing the JS bundle was slow and could cause the editor to occasionally freeze. This PR removes the `typescript` dependency via the `--external:typescript` esbuild flag. Surprisingly, the `tsc-mixed` context strategy still seems to work within VSC and it might be because VSC somehow exposes access to TypeScript APIs by default similar to how it works for `vscode` imports. However, this change means that `tsc-mixed` does not work for the agent until we bundle `typescript.js` alongside the `index.js` for the agent.

## Test plan

* Manually build the extension locally.
* Confirm that `tsc-mixed` still works
* Confirm that the bundle size is ~14mb:
```
❯ pnpm -C vscode build
❯ du -h vscode/dist/extension.node.js
 14M	vscode/dist/extension.node.js
```


philipp-spiess (approved):
Ohhh nice it's a great finding that VS Code ships with `typescript` already. This is means not only we have to not ship it but also that it's already loaded in the runtime for us (so the cost of using it is minimal). Nice!

PR #4030, opened 2024-05-03
Edit: Disable streaming for 'Add' insertions in Agent

umpox:
## Description

We are running into issues where streamed insertions are not reliably applied to a document in JetBrains ([issue](https://github.com/sourcegraph/jetbrains/issues/1449)). This PR temporarily disables this so we can unblock progress on JetBrains whilst we figure out why this is happening

This PR:
- Disables streaming insertions into the document when running an inline edit with the `add` intent

## Test plan

1. Check the `add` variant of inline edit (create an edit with no selection)
2. Check edits are not streamed when running via agent

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -8646,6 +8646,143 @@ log:
         send: 0
         ssl: -1
         wait: 0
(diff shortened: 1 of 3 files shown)


olafurpg (approved):
LGTM 👍🏻 For future PRs, can you elaborate in more detail what problem this change solves? Something like

> Previously, streaming was enabled in agent and this was problematic because ....

PR #4031, opened 2024-05-03
Edit: Support trimming active selections to a tree-sitter node

umpox:
## Description

Fixes CODY-1363

This PR:
- Attempts to match a users' provided selection to a tree-sitter node, so we can improve the reliability of Edit commands

**Before:**

We would prioritise the users selection (if non-empty) before anything else. This worked in most cases, but was not optimal as the user may have actually been attempting to match the tree-sitter node, but made a mistake and possibly:
- Accidentally selected more than the node, e.g. extra whitespace chars
- Accidentally didn't select the full node (e.g. made a partial line selection)

This would also be non-optimal as in some case we get more insight into the users' selection if we use tree-sitter. For example, for "Generate Documentation" in Python, using tree-sitter will help us determine the correct `insertionPoint` for the documentation. This relies on knowing the syntax of the node in Python. Even if the users' selection exactly matched the node we would not get that information.

https://github.com/sourcegraph/cody/assets/9516420/ab3b35d1-eb01-4fc5-b559-7d2ed972c857

**After:**

We now attempt to match the users' selection to a tree-sitter node, in a safe manner that doesn't discard situations where their selection is likely preferred. This is by:
1. Expanding their selection to the full lines of the start and end (helps if they don't select a full line)
2. Trimming their selection to the first and last non-whitespace characters (helps if they select too many empty lines/spaces)
3. Comparing this adjusted selection to the retrieved tree-sitter node. If it matches exactly then we can be confident that the user would prefer to user this, and we can benefit from the additional information we retrieve.

https://github.com/sourcegraph/cody/assets/9516420/2c1a56a4-a8f3-4ba5-bb45-33ddcf2054a2

## Test plan

Document:

1. Execute a "Document Code" command from just a cursor position, within an expected tree-sitter node (e.g. JS function). Expect it to expand correctly.
2.  Execute a "Document Code" command from a full selection that doesn't match an exact tree-sitter node (e.g. select just a function call like `console.log()`). Expect it to only document this line with the function call
3.  Execute a "Document Code" command from a full selection that _does_ match an exact tree-sitter node, expect it to document this node correctly.
4. Execute a "Document Code" command from an **expanded** selection that matches a tree-sitter node but with additional whitespace (including new lines) before and after. Expect it to document the tree-sitter node correctly. In Python, expect it to determine the correct insertion point.
5. Execute a "Document Code" command from a **partial** selection that matches some part of the first line of a tree-sitter node. E.g. (select partially the first line, and partially the second line in a `function foo() {\n console.log('hi')` func). Expect it to document the tree-sitter node correctly.

### DIFF
--- a/vscode/src/commands/execute/doc.ts
+++ b/vscode/src/commands/execute/doc.ts
@@ -8,8 +8,25 @@ import { getEditor } from '../../editor/active-editor'
 import type { EditCommandResult } from '../../main'
 import type { CodyCommandArgs } from '../types'
 
-import { getEditLineSelection } from '../../edit/utils/edit-selection'
+import { getEditDefaultProvidedRange, getEditTrimmedSelection } from '../../edit/utils/edit-selection'
 import { execQueryWrapper } from '../../tree-sitter/query-sdk'
+
+function getDefaultDocumentableRange(editor: vscode.TextEditor): {
+    range?: vscode.Range
+    insertionPoint?: vscode.Position
+} {
+    const defaultRange = getEditDefaultProvidedRange(editor.document, editor.selection)
+
+    if (defaultRange) {
+        return {
+            range: defaultRange,
+            insertionPoint: defaultRange.start,
+        }
+    }
+
+    // No usable selection, fallback to expanding the range to the nearest block.
+    return {}
+}
 
 /**
  * Gets the symbol range and preferred insertion point for documentation
@@ -27,15 +44,8 @@ function getDocumentableRange(editor: vscode.TextEditor): {
     range?: vscode.Range
     insertionPoint?: vscode.Position
 } {
-    const { document, selection } = editor
-    if (!selection.isEmpty) {
-        const lineSelection = getEditLineSelection(editor.document, editor.selection)
-        // The user has made an active selection, use that as the documentable range
-        return {
-            range: lineSelection,
-            insertionPoint: lineSelection.start,
-        }
-    }
+    const { document } = editor
+    const trimmedSelection = getEditTrimmedSelection(document, editor.selection)
 
     /**
      * Attempt to get the range of a documentable symbol at the current cursor position.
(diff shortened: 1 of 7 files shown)


toolmantim:
Nice!

vscode/src/commands/execute/doc.ts
```diff
@@ -8,9 +8,29 @@ import { getEditor } from '../../editor/active-editor'
...
 import type { CodyCommandArgs } from '../types'
 
-import { getEditLineSelection } from '../../edit/utils/edit-selection'
+import {
+    getEditAdjustedUserSelection,
+    getEditDefaultProvidedRange,
+} from '../../edit/utils/edit-selection'
 import { execQueryWrapper } from '../../tree-sitter/query-sdk'
 
+function getDefaultDocumentableRange(editor: vscode.TextEditor): {
```
valerybugakov, line 17: Adding a comment here explaining the "default documentable range". Also, the function name is a bit misleading because it's very similar to `getEditDefaultProvidedRange` but the return type is more complex.

vscode/src/commands/execute/doc.ts
```diff
@@ -8,9 +8,29 @@ import { getEditor } from '../../editor/active-editor'
...
+
+    if (defaultRange) {
+        return {
+            range: defaultRange,
+            insertionPoint: defaultRange.start,
+        }
+    }
+
+    // No usable selection, fallback to expanding the range to the nearest block.
+    return {}
```
valerybugakov, line 31: Is the expansion to the nearest block done somewhere up the call chain?

vscode/src/commands/execute/doc.ts
```diff
@@ -27,39 +47,49 @@ function getDocumentableRange(editor: vscode.TextEditor): {
...
     } = documentableRange
+    const range = new vscode.Range(
+        startPosition.row,
+        startPosition.column,
+        endPosition.row,
+        endPosition.column
+    )
+
+    // If the users' adjusted selection aligns with the start of the node and is contained within the node,
+    // It is probable that the user would benefit from expanding to this node completely
```
valerybugakov, line 85: 🔥

vscode/src/commands/execute/doc.ts
```diff
@@ -27,39 +47,49 @@ function getDocumentableRange(editor: vscode.TextEditor): {
...
+    const selectionMatchesNode =
+        adjustedSelection.start.isEqual(range.start) && range.contains(adjustedSelection.end)
+    if (!selectionMatchesNode && !editor.selection.isEmpty) {
```
valerybugakov, line 88: Should we cover this branching in one of the unit tests?

valerybugakov (approved):
Great work! Left some comments inline.

P.S. It would be so cool to somehow animate the selection expansion to the nearest node (showing to a user how smart Cody is!). I really wanted to see visual feedback for it in the demo video. cc @toolmantim

vscode/src/commands/execute/doc.ts
umpox: Kinda tricky to isolate this with a specific test, but actually we have agent tests that cover this.

I noticed an issue with cursor position ranges (`isEmpty`) from these tests already

PR #4032, opened 2024-05-03
Cody Ignore: Show Ignored file status in @-mentions

philipp-spiess:
Closes CODY-1339

Adds a notice when @-mentioning enterprise ignored files and properly excludes the content.

## Test plan

https://github.com/sourcegraph/cody/assets/458591/8922773e-949d-4e6f-a561-66c219280c2a


vscode/webviews/components/FileLink.tsx
```diff
@@ -20,9 +20,11 @@ interface FileLinkProps {
     range?: RangeData
     title?: string
     isTooLarge?: boolean
+    isIgnored?: boolean
 }
 
-const WARNING = 'Excluded due to context window limit'
+const LIMIT_WARNING = 'Excluded due to context window limit'
+const IGNORE_WARNING = 'File ignored by an admin setting'
```
valerybugakov, line 27: Does it make sense to use the relevant constant from `vscode/src/chat/context/constants.ts`?

valerybugakov (approved):

vscode/webviews/components/FileLink.tsx
philipp-spiess: We should do that once we start reusing some strings, for now this is fine I think

PR #4035, opened 2024-05-03
Add command to debug context filters

philipp-spiess:
Closes CODY-1341

Add a command to overwrite context filters. I've for now gated it to only show up when verbose debugging is enabled however that doesn't seem to be working well 🤔 

## Test plan

```
{ "include": [], "exclude": [{ "repoNamePattern": "^github\\.com/sourcegraph/sourcegraph" }] }
```

https://github.com/sourcegraph/cody/assets/458591/4fda5932-a708-457f-9a55-072cd42d9b08

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -727,6 +727,10 @@
         {
           "command": "cody.show-rate-limit-modal",
           "when": "false"
+        },
+        {
+          "command": "cody.test.set-context-filters",
+          "when": "cody.activated && cody.devOrTest"
         }
       ],
       "editor/context": [
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -546,13 +546,6 @@
         "when": "!config.cody.debug.verbose"
       },
       {
-        "command": "cody.test.set-context-filters",
-        "category": "Cody Debug",
-        "group": "Debug",
-        "title": "[Debug] Set Context Filters Overwrite",
-        "when": "cody.enabled && config.cody.debug.verbose"
-      },
-      {
         "command": "cody.debug.reportIssue",
         "category": "Cody Debug",
         "group": "Debug",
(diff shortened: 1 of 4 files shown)


vscode/package.json
```diff
@@ -565,6 +565,11 @@
         "icon": "$(search)",
         "title": "Natural Language Search Code (Beta)",
         "when": "cody.activated && workspaceFolderCount > 0"
+      },
+      {
+        "command": "cody.test.set-context-filters",
+        "title": "[Internal] Set Context Filters Overwrite",
+        "when": "!cody.devOrTest"
       }
```
philipp-spiess, line 573: Anyone know why this would show up no matter if I invert this check or not? 🤔

umpox: Did you fix it? Curious why this was happening

umpox: Wondered if you needed to manually set it in the `commandPalette` section of package.json too

umpox (approved):

vscode/package.json
philipp-spiess: Still looking haha let me try that

philipp-spiess: Ah yes that's it nice! TY.
I find this json config SO SO confusing 🙃

philipp-spiess: I think `when` here doesn't do something actually? There isn't even a JSON description? Why does every action define one though 🤔

umpox: I vaguely remember needing to declare the command in one place, and then exclude it from the command palette in another place. I thought this is the same issue, I think you need both?

Agree it's so confusing

philipp-spiess: Docs are not very clear they mention you have to use `when` but not where `when` is defined: https://code.visualstudio.com/api/references/contribution-points#contributes.commands

If you hover over `when` in `commands`, you get no title (suggesting it's not part of the spec)

What a mess

PR #4037, opened 2024-05-03
Edit: Fix persistence intent tracking data

umpox:
## Description

We were passing the wrong `metadata` value. We already `splitSafeMetadata` in the `onPresent` and `onRemoved` callbacks.

We can just pass the full metadata object in here instead of the already split one.

This is a bit messy, but there's not really a nice way to handle all these telemetry calls without duplicating some of this :/ 

## Test plan

1. Create an edit
2. Wait 30 seconds
3. Check `fixup:persistence:present` is logged with the correct metadata


abeatrix (approved):

philipp-spiess (approved):
😬

PR #4039, opened 2024-05-03
Autocomplete: force 17b on explicit autocomplete with starcoder-hybrid

olafurpg:
Previously, when using starcoder-hybrid, we would use starcoder-7b for single-line completions even when the user explicitly triggers autocomplete via shortcut. This PR changes the behavior so we force starcoder-16b on explicit trigger.

The reasoning behind this change is that users are OK with getting longer latency when they explicitly trigger the model.

A possible followup change is to enable multi-line completion.

## Test plan
Green CI.


agent/src/graph-context.test.ts
```diff
@@ -76,7 +76,7 @@ describe.skipIf(isWindows())('Graph Context', () => {
         const prompts: { name: string; value: any }[] = []
         await Promise.all(
             activeClients().map(async client => {
-                const autocomplete = await client.autocompleteText()
+                const autocomplete = await client.autocompleteText({ triggerKind: 'Invoke' })
```
olafurpg, line 79: Was previously implicit in `TestClient`. Using automatic trigger actually made these tests fail, I'd love to better understand why 🤔

agent/src/autocomplete.test.ts
```diff
@@ -0,0 +1,49 @@
+import path from 'node:path'
+import { afterAll, beforeAll, describe, expect, it } from 'vitest'
+import { TestClient } from './TestClient'
+import { TestWorkspace } from './TestWorkspace'
+import { TESTING_TOKENS } from './testing-tokens'
+
+describe('Autocomplete', () => {
```
olafurpg, line 7: Used this PR as an opportunity to show how little code is needed to add a new agent test file. With the new `TestClient.{beforeAll,afterAll}` hooks, it's super easy to add new fixtures. Also, we no longer record the streaming LLM reply so the recording file is smaller than before (although still too big IMO for a basic example like here).

philipp-spiess:
I love the agent tests but I don’t think we should be adding so much of them, especially for something as targeted as this change. The reason is that they’re flaky and slow (the typical issue with integration tests like this). I think this is the typical “integration test vs. unit test” debate.

My worry is that the more snapshots we add, the less likely we will be in actually reviewing changes whenever the snapshots are outdated (do we do that right now? or do we just update the snapshots and hope it’s good?). 

I really think stuff like this would be better as a unit test. There is no value in this test to assert that the output of the inference is actually different, we only care that the right model string is sent.

philipp-spiess:
Completions specifically have IMO a pretty good unit test suite that should make it easy to add that assertion without having to take on some of the drawbacks that the E2E recordings bring. Check out this file for reference: https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/get-inline-completions-tests/triggers.test.ts

vscode/src/completions/providers/fireworks.ts
```diff
@@ -241,17 +242,20 @@ class FireworksProvider extends Provider {
         })
 
         const { multiline } = this.options
+        this.options.docContext.currentLinePrefix
```
philipp-spiess, line 245:
```suggestion
```

philipp-spiess (commented):
The code change looks good but I feel strongly that we should not use an integration test for something so small and scoped. 

It's on you obviously to accept that opinion or not 😅

philipp-spiess (approved):

philipp-spiess:
@olafurpg Let's stop adding auto merge for non trivial changes. I didn't notice it was enabled this time and now we commited https://github.com/sourcegraph/cody/pull/4039/files/b07cb606a6c5fc020012f1e0be70a3717dc54e76#diff-4236ad86f26bb203b040a80dd111451001b68dd322aefa602f5b71857c4d5880 🙃

olafurpg:
@philipp-spiess I agree with the problems (slow, flaky) and I've been thinking about how to address those problems while faithfully reproducing a production environment. The current codebase is designed in a way that I have very low confidence that narrowly targeted unit tests are going to catch real-world regressions. It would have been super easy to add a unit test against the fireworks provider but I have seen so many cases where we refactor the code, unit tests pass but then we need to spend hours debugging why the desired behavior is not happening in production.

My ideal state is that we can continue to write integration tests through the agent and we make them run reliably fast and stable. This includes investing time in cutting out unnecessary network traffic or adding an offline mode that stubs/mocks out things we don't care about. Polly does support providing manual stubs that I'd be open to using.

vscode/src/completions/providers/fireworks.ts
olafurpg: I'm sorry https://github.com/sourcegraph/cody/pull/4041

philipp-spiess:
@olafurpg I think there's not a lot of difference between the completion tests that we have (which I named unit tests here but which are more like integration tests) and the agent tests (which are more like E2E tests) since the whole VS Code mock that we use for agent actually originates in the completion part so I’m confident we both have the same ideal here. I guess the question is if we all E2E tests actually need to run against an inference endpoint. You mentioned being open to stubs/mocks, I think if we have this + eliminate all the crazy dangling promise mess (which we should anyways to make it easier to reason about what our extension does) we both will be very happy 🙂

vscode/src/completions/providers/fireworks.ts
philipp-spiess: No worries I should have seen the auto-merge :( I actually stopped to check (because I didn't approve at first) but the option can be really hidden...

PR #4044, opened 2024-05-03
promptEditor: ensure exhaustive handling of ContextItem.type

keegancsmith:
Tracing back in the history for when contextItemMentionNodeDisplayText was first added, it had this same simple pattern of throw an error if a context item does not have type set. By removing the throw at the end the typescript compiler will complain if we add a new type and do not handle it here.

Note: I may be missing a reason why this is done (due to this being on serialized and not having runtime type safety?)

Test Plan: inspect the diff with whitespace ignored to ensure only the if statements change to case statements. Also CI.


dominiccooney (approved):
I am a big fan of using the types to make maintenance easier.

PR #4046, opened 2024-05-03
Chat: Fix editor selection retrieval from context menu

abeatrix:
Summary:

The current implementation for the `Add selection to Cody Chat` is incorrect as it only takes the returned URL from the vscode context menu API into account when retrieving the context from the document, causing the command to always return the whole file rather than a specific range of the file. This PR updates to check if the current editor has selection first, before using the URL to get content from the editor file.

The updated implementation now takes into account the active text editor's selection, which allows for more precise context fetching based on the user's focus.

Changes:

Selection Handling: Added logic to fetch the selection from the active text editor. This selection is then used when fetching context from a URI or directly from the cursor when uri is not provided.

## Test plan

- Ensure that the context is correctly fetched when a selection is present and when there is no selection.
- Verify that the context items are correctly formed and include with the correct range as @-mention item

#### After

The selection is added to chat as @-mention:

image

#### Before

Whole file is added as @-mention event when a selection is presented in the editor

image

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -652,20 +652,30 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
     }
 
     public async handleGetUserEditorContext(uri?: URI): Promise<void> {
-        const selected = uri ? await getContextFileFromUri(uri) : await getContextFileFromCursor()
+        // Get selection from the active editor
+        const selection = vscode.window.activeTextEditor?.selection
+
+        // Determine context based on URI presence
+        const contextItem = uri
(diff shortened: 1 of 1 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -652,20 +652,30 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
     }
 
     public async handleGetUserEditorContext(uri?: URI): Promise<void> {
-        const selected = uri ? await getContextFileFromUri(uri) : await getContextFileFromCursor()
```
abeatrix, line 655: this was the issue because editor always returns a URI, so we should check for selection first

kalanchan (approved):

PR #4048, opened 2024-05-04
add experimental vscode command to get completions from multiple models

hitesh-1997:
## Context
The goal of this PR is to add a custom command in cody (purely for the internal testing and comparing the completion output from different LLM models), which when triggered will print the completions from multiple providers. This would be very useful to compare completions output from different models.

The way I currently tried to implement the same is:
1. Initialize multiple completions providers at the initialization. The list of providers is picked from the VSCode settings by providing the following config:
```
"cody.autocomplete.experimental.multiModelCompletions": [
    {"provider": "fireworks", "model": "starcoder-hybrid"},
    {"provider": "fireworks", "model": "fireworks-completions-fine-tuned", "enableExperimentalFireworksOverrides": true},
    {"provider": "anthropic", "model": "claude-3-haiku-20240307"},
  ]
```
2. Whenever the custom command is invoked, collect the (TextDocument, position etc) from the vscode api and call [provideInlineCompletionItems](https://sourcegraph.com/github.com/sourcegraph/cody@5c69fc9/-/blob/vscode/src/completions/inline-completion-item-provider.ts?L198) on all the initialized providers
3. Since we are logging the completions data and metrics, asking for completions on multiple providers could lead to unexpected logging, so I have tried to the condition to skip logging whenever, the completions are called for this vscode command. For eg: [see this ](https://github.com/sourcegraph/cody/pull/4048/files#diff-a2dca3f7cbfa2847ddc325842db1723618f70febf6f26d372c7f8adf3d475bafR183)

## Test plan
1. Manually start the debugger on this branch and trigger the vscode command `Get Completions for multiple autocomplete models`
2. The output should include completions as well as completions requests from all the providers.
3. Verified that we are skipping logging part.

https://github.com/sourcegraph/cody/assets/20701220/689037b2-72e2-427f-a8af-e20f58b50227

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -417,6 +417,12 @@
         "when": "cody.activated && config.cody.autocomplete && editorHasFocus && !editorReadonly && !editorHasSelection && !inlineSuggestionsVisible"
       },
       {
+        "command": "cody.multi-model-autocomplete.manual-trigger",
+        "category": "Cody",
+        "title": "Get Completions for multiple autocomplete models",
+        "when": "cody.activated && config.cody.autocomplete && editorHasFocus && !editorReadonly && !editorHasSelection && !inlineSuggestionsVisible"
+      },
+      {
         "command": "cody.chat.panel.new",
         "category": "Cody",
         "title": "New Chat",
@@ -641,6 +647,11 @@
       },
       {
         "command": "cody.autocomplete.manual-trigger",
+        "key": "alt+\\",
+        "when": "editorTextFocus && !editorHasSelection && config.cody.autocomplete.enabled && !inlineSuggestionsVisible"
+      },
+      {
+        "command": "cody.multi-model-autocomplete.manual-trigger",
         "key": "alt+\\",
         "when": "editorTextFocus && !editorHasSelection && config.cody.autocomplete.enabled && !inlineSuggestionsVisible"
       },
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -420,7 +420,7 @@
         "command": "cody.multi-model-autocomplete.manual-trigger",
         "category": "Cody",
         "title": "Get Completions for multiple autocomplete models",
-        "when": "cody.activated && config.cody.autocomplete && editorHasFocus && !editorReadonly && !editorHasSelection && !inlineSuggestionsVisible"
+        "when": "cody.activated && config.cody.autocomplete.experimental.multiModelCompletions && config.cody.autocomplete && editorHasFocus && !editorReadonly && !editorHasSelection && !inlineSuggestionsVisible"
       },
       {
         "command": "cody.chat.panel.new",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -652,7 +652,7 @@
       },
       {
         "command": "cody.multi-model-autocomplete.manual-trigger",
-        "key": "alt+\\",
+        "key": "alt+m",
         "when": "editorTextFocus && !editorHasSelection && config.cody.autocomplete.enabled && !inlineSuggestionsVisible"
       },
       {
--- a/vscode/src/completions/create-inline-completion-item-provider.ts
+++ b/vscode/src/completions/create-inline-completion-item-provider.ts
@@ -144,6 +144,7 @@ export async function createInlineCompletionItemFromMultipleProviders({
         {
             'provider': 'fireworks',
             'model': 'fireworks-completions-fine-tuned',
+            'overrideFireworks': true
         },
         {
             'provider': 'anthropic',
@@ -151,13 +152,18 @@ export async function createInlineCompletionItemFromMultipleProviders({
         }
     ]
     const allProviders: InlineCompletionItemProvider[] = []
+
     for (const curretProviderConfig of allProviderConfigs) {
+        const newConfig = JSON.parse(JSON.stringify(config))
+        newConfig.autocompleteExperimentalFireworksOptions = curretProviderConfig.overrideFireworks
+            ? config.autocompleteExperimentalFireworksOptions
+            : undefined
         const providerConfig = await createProviderConfigForModel(
             client,
             authStatus,
             curretProviderConfig['model'],
             curretProviderConfig['provider'],
-            config
+            newConfig
         )
         if(providerConfig) {
             const authStatus = authProvider.getAuthStatus()
--- a/vscode/src/completions/create-inline-completion-item-provider.ts
+++ b/vscode/src/completions/create-inline-completion-item-provider.ts
@@ -1,5 +1,6 @@
 import * as vscode from 'vscode'
-
+import * as path from 'node:path'
+import * as fspromises from 'node:fs/promises'
 import {
     type CodeCompletionsClient,
     type ConfigurationWithAccessToken,
@@ -16,6 +17,7 @@ import type { BfgRetriever } from './context/retrievers/bfg/bfg-retriever'
 import { InlineCompletionItemProvider } from './inline-completion-item-provider'
 import { createProviderConfig } from './providers/create-provider'
 import { registerAutocompleteTraceView } from './tracer/traceView'
+import { MultiModelCompletionsResults } from './inline-completion-item-provider'
 import { createProviderConfigForModel } from './providers/create-provider'
 
 interface InlineCompletionItemProviderArgs {
@@ -43,28 +45,75 @@ class NoopCompletionItemProvider implements vscode.InlineCompletionItemProvider 
     }
 }
 
-export async function triggerMultiModelAutocompletionsForComparison(allProviders: InlineCompletionItemProvider[]) {
-    const activeEditor = vscode.window.activeTextEditor
+export async function createLogsFile(): Promise<string> {
+    const dirPath = "/Users/hiteshsagtani"
+    await fspromises.mkdir(dirPath, { recursive: true })
+    const fileName = 'cody-custom-completions.jsonl'
+    const filePath = path.join(dirPath, fileName)
+    if (
+        !(await fspromises
+            .access(filePath)
+            .then(() => true)
+            .catch(() => false))
+    ) {
+        await fspromises.writeFile(filePath, '')
+    }
+    return filePath
+}
+
+export async function triggerMultiModelAutocompletionsForComparison(
+    allProviders: InlineCompletionItemProvider[],
+) {
+    const activeEditor = vscode.window.activeTextEditor;
     if (!activeEditor) {
-        return
+        return;
     }
     const document = activeEditor.document;
     const position = activeEditor.selection.active;
     const context = {
         triggerKind: vscode.InlineCompletionTriggerKind.Automatic,
-        selectedCompletionInfo: undefined
-    }
-    const allPromises: Promise<string>[] = []
+        selectedCompletionInfo: undefined,
+    };
+    const allPromises: Promise<MultiModelCompletionsResults>[] = [];
     for (const provider of allProviders) {
-        allPromises.push(provider.manuallyGetCompletionItemsForProvider(
-            document,
-            position,
-            context
-        ))
+        allPromises.push(
+            provider.manuallyGetCompletionItemsForProvider(
+                document,
+                position,
+                context,
+            ),
+        );
     }
     const results = await Promise.all(allPromises);
-    const allResults = results.join('\n');
-    logDebug('MultiModelAutoComplete:\n', allResults);
+    let completionsOutput = ""
+    for (const result of results) {
+        completionsOutput += `Model: ${result.model}\n${result.completion}\n\n`
+    }
+    logDebug("MultiModelAutoComplete:\n", completionsOutput);
+
+    // Add logs to a local file for analysis
+    const logFilePath = await createLogsFile();
+
+    const prefix = document.getText(
+        new vscode.Range(new vscode.Position(0, 0), position),
+    );
+    const suffix = document.getText(
+        new vscode.Range(
+            position,
+            document.positionAt(document.getText().length),
+        ),
+    );
+    const fileName = path.basename(document.fileName);
+    let dataToLog: { [key: string]: string } = {
+        fileName,
+        prefix,
+        suffix,
+    };
+    for (const result of results) {
+        const modelName = result.model;
+        dataToLog[modelName] = result.completion? result.completion : "";
+    }
+    await fspromises.appendFile(logFilePath, JSON.stringify(dataToLog) + "\n");
 }
 
 
--- a/vscode/src/completions/create-inline-completion-item-provider.ts
+++ b/vscode/src/completions/create-inline-completion-item-provider.ts
@@ -16,6 +16,7 @@ import type { BfgRetriever } from './context/retrievers/bfg/bfg-retriever'
 import { InlineCompletionItemProvider } from './inline-completion-item-provider'
 import { createProviderConfig } from './providers/create-provider'
 import { registerAutocompleteTraceView } from './tracer/traceView'
+import { createProviderConfigForModel } from './providers/create-provider'
 
 interface InlineCompletionItemProviderArgs {
     config: ConfigurationWithAccessToken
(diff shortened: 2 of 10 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -76,6 +76,11 @@ interface CodyCompletionItemProviderConfig {
 
     // Feature flags
     completeSuggestWidgetSelection?: boolean
+
+    // Flag to check if the current request is also triggered for multiple providers.
+    // When true it means the inlineCompletion are triggerd for multiple model for comparison purpose.
+    // Check `createInlineCompletionItemFromMultipleProviders` method in create-inline-completion-item-provider for more detail.
+    isRequestForMultipleModelCompletions?: boolean
```
philipp-spiess, line 83: This is an option on the provider though and not the individual request, so the name is confusing. Perhaps let's focus on what behavior this changes instead? Maybe "noAnalytics"?

vscode/package.json
```diff
@@ -416,6 +416,12 @@
         "title": "Trigger Autocomplete at Cursor",
         "when": "cody.activated && config.cody.autocomplete && editorHasFocus && !editorReadonly && !editorHasSelection && !inlineSuggestionsVisible"
       },
+      {
+        "command": "cody.multi-model-autocomplete.manual-trigger",
+        "category": "Cody",
+        "title": "Get Completions for multiple autocomplete models",
+        "when": "cody.activated && config.cody.autocomplete && editorHasFocus && !editorReadonly && !editorHasSelection && !inlineSuggestionsVisible"
```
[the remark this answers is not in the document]
hitesh-1997: How can we avoid displaying this option in the UI Philipp ? (Adding something to `when` is not relevant i think)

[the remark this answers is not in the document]
philipp-spiess: By adding something to `when` :D WDYM with "not relevant"? https://code.visualstudio.com/api/references/when-clause-contexts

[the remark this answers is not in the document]
philipp-spiess: Oh I see this is in the `commands` subsection and you want to add another entry in `commandPalette` with the condition.

vscode/package.json
```diff
@@ -644,6 +650,11 @@
         "key": "alt+\\",
         "when": "editorTextFocus && !editorHasSelection && config.cody.autocomplete.enabled && !inlineSuggestionsVisible"
       },
+      {
+        "command": "cody.multi-model-autocomplete.manual-trigger",
+        "key": "alt+\\",
```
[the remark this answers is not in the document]
hitesh-1997: Changed the key to another value.

vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -42,6 +43,106 @@ class NoopCompletionItemProvider implements vscode.InlineCompletionItemProvider
     }
 }
 
+export async function triggerMultiModelAutocompletionsForComparison(allProviders: InlineCompletionItemProvider[]) {
```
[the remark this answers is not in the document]
hitesh-1997: done

vscode/package.json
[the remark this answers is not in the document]
hitesh-1997: Added a vscode setting `cody.autocomplete.experimental.multiModelCompletions ` where we can pass the list of model ids.

vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -42,6 +43,106 @@ class NoopCompletionItemProvider implements vscode.InlineCompletionItemProvider
...
+    // which helps judge the quality of code completions
+    const authStatus = authProvider.getAuthStatus()
+    if (!authStatus.isLoggedIn) {
+        return {
+            dispose: () => {},
+        }
+    }
+
+    const disposables: vscode.Disposable[] = []
+    const allProviderConfigs = [
```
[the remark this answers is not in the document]
hitesh-1997: Added a vscode setting cody.autocomplete.experimental.multiModelCompletions  where we can pass the list of model ids and providers.

vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -42,6 +43,106 @@ class NoopCompletionItemProvider implements vscode.InlineCompletionItemProvider
...
+            'model': 'fireworks-completions-fine-tuned',
+        },
+        {
+            'provider': 'anthropic',
+            'model': 'claude-3-haiku-20240307',
+        }
+    ]
+    const allProviders: InlineCompletionItemProvider[] = []
+    for (const curretProviderConfig of allProviderConfigs) {
+        const providerConfig = await createProviderConfigForModel(
```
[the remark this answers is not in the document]
hitesh-1997: done

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -174,15 +180,18 @@ export class InlineCompletionItemProvider
...
-            this.contextMixer,
-            vscode.commands.registerCommand(
-                'cody.autocomplete.inline.accepted',
-                ({ codyCompletion }: AutocompleteInlineAcceptedCommandArgs) => {
-                    void this.handleDidAcceptCompletionItem(codyCompletion)
-                }
+        if (!this.config.isRequestForMultipleModelCompletions) {
+            // We don't want to accept and log items when we are doing completion comparison from different models.
+            this.disposables.push(
+                this.contextMixer,
```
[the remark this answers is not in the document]
hitesh-1997: thanks for pointing out, removed the condition for contextMixer.

vscode/src/completions/inline-completion-item-provider.ts
hitesh-1997: thanks, changed to `noAnalytics`.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -21,6 +21,51 @@ import { createProviderConfig as createOpenAICompatibleProviderConfig } from './
 import type { ProviderConfig } from './provider'
 import { createProviderConfig as createUnstableOpenAIProviderConfig } from './unstable-openai'
 
+export async function createProviderConfigForModel(
```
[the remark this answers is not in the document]
hitesh-1997: Changed to using the same function :)

hitesh-1997:
> Good stuff! Left some early feedback. Most important IMO is that we properly gate this out if it's not intended to be used in production (and I think it can have too many unwanted side effects right now since some caches are global so we really shouldn't).

Thanks for the review Philipp :)
Tried to address all of them comments, can you please give this a look.
This makes sense that some global caches could get affected, and we shouldn't roll this out to production. I made sure to [return a empty disposable](https://github.com/sourcegraph/cody/pull/4048/files#diff-4fae773057a4f852e7e51755bed0f6d9e6d117d7d892faf6e495d784f047d76eR106) in case the config `cody.autocomplete.experimental.multiModelCompletions ` is not set in the vscode setting. Does this condition suffices, since end user would not know about this.

Another point it, we are exposing this [custom command](https://github.com/sourcegraph/cody/pull/4048/files#diff-99d7a3c2c2975fcaf8184c4639922f5ef6c915610b6f8e15ffa8fb8b329828d8R422) as a way to trigger the completions, but this option would still be visible to the end user. Wanted to check if this is okay or is there a better way to trigger the completions without adding a custom command ?

philipp-spiess:
> Another point it, we are exposing this [custom command](https://github.com/sourcegraph/cody/pull/4048/files#diff-99d7a3c2c2975fcaf8184c4639922f5ef6c915610b6f8e15ffa8fb8b329828d8R422) as a way to trigger the completions, but this option would still be visible to the end user, and if they trigger will get a error message that the command is not registered, since we returned empty disposable before.
>
> Wanted to check if this is okay or there is a better way or is there a better way to trigger the completions without adding a custom command ?

You want to filter this out by creating an entry in the `commandPalette` part of the package.json and attaching a filter based on the config option. Like this:

https://github.com/sourcegraph/cody/blob/main/vscode/package.json#L708-L711

philipp-spiess:
Awesome re. the updated test plan. Really like this! Thank you!

vscode/src/completions/create-multi-model-inline-completion-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    if (multiModelConfigsList.length === 0) {
+        return {
+            dispose: () => {},
+        }
+    }
+
+    const allPromises = multiModelConfigsList.map(async curretProviderConfig => {
+        const newConfig = _.cloneDeep(config)
```
philipp-spiess, line 74: Nice this is a very good idea to rule out any mutations on the config causing issues here.

vscode/src/completions/create-multi-model-inline-completion-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+                createBfgRetriever,
+                isDotComUser: isDotCom(authStatus.endpoint || ''),
+                noAnalytics: true,
+            })
+            return completionsProvider
+        }
+        // biome-ignore lint/style/noUselessElse: returns undefined if provider config not valid
+        else {
+            return undefined
+        }
```
philipp-spiess, line 110:
```suggestion
        }
        return undefined
```

This rule just says you don't need an else here if you return in your condition, so we can avoid the warning by doing this

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -571,12 +598,35 @@ export class InlineCompletionItemProvider
...
+            CompletionLogger.partiallyAccept(
+                completion.logId,
+                completion.analyticsItem,
+                acceptedLength,
+                this.config.isDotComUser
+            )
+        }
+    }
+
+    public async manuallyGetCompletionItemsForProvider(
```
philipp-spiess, line 611: I don't think we need to add this method here, can we instead call the existing `provideInlineCompletionItems` and make the changes to the return type in the `create-multi-model-inline-completion-provider` file instead. This way the changes to this file are minimal and we can keep the surface API nice and clean

philipp-spiess (approved):
This is shaping up nicely!! Some very minor comments inside but let's look into hiding the command for when the feature flag is not enabled like this: https://github.com/sourcegraph/cody/pull/4048#issuecomment-2097952074

I'll stamp this preemptively though. :)

valerybugakov (commented):
It brings me to an interesting product idea that can be valuable for our users. We could leverage the built-in VS Code UI for multiple completions and allow generating each completion with a different model, like the awesome [LLM Litmus Test tool](http://s0.dev/), but integrated into autocomplete.

It can be combined with our regular manual trigger command. We recently discussed the idea of using more powerful models with a higher latency for better results when users intentionally want to generate a longer/high-quality completion. 

cc @philipp-spiess @olafurpg @aramaraju

olafurpg:
@valerybugakov do you mean returning multiple autocomplete results with different providers and let the user cycle through the results with alt-brackets? I would personally prefer a separate UI in the sidebar like this

valerybugakov:
> let the user cycle through the results with alt-brackets?

Yep, I was thinking about it. We never got the value out of the built-in tooltip, and maybe this is the use case for it to shine. I like a separate UI! It's handy for comparing multiple multiline completions.

philipp-spiess:
@valerybugakov @olafurpg Yeah I agree this would fit very well in a separate sidebar but for now the idea is to only have something so we can test two different completion providers locally quick and dirty. If this provides value we can show it in the trace view sidebar or add a dedicated one (we had one in the early days)

I think using the tooltip is confusing for this use case. We could probably add custom tooltips to add context but at the end of the day I think people would just go with the default suggestion anyways and not bother cycling through ideas

valerybugakov:
> but for now the idea is to only have something so we can test two different completion providers locally quick and dirty

To clarify, I do not propose to make any changes related to exposing multiple completions from multiple providers to our PLG users in this PR. I just wanted to share the idea.

vscode/src/completions/create-multi-model-inline-completion-provider.ts
hitesh-1997: ah got it

vscode/src/completions/inline-completion-item-provider.ts
hitesh-1997: makes sense, moved this up `create-multi-model-inline-completion-provider` file

vscode/src/completions/providers/create-provider.ts
```diff
@@ -176,6 +179,12 @@ async function resolveDefaultProviderFromVSCodeConfigOrFeatureFlags(
     }
 
     if (starCoderHybrid) {
+        // Adding the fine-tuned model here for the A/B test setup.
+        // Among all the users in starcoder-hybrid - some % of them will be redirected to the fine-tuned model.
+        if (finetunedModel) {
```
hitesh-1997, line 184: Hi @philipp-spiess 
Adding the fine-tuning condition insidde the `starCoderHybrid` for the A/B test launch.
I am assuming if the a/b test traffic is 5%, in the end we can pick 5% random users ids from starcoderHybrid to compare the numbers ? Or should we have another feature flag, which would have exactly the same configuration as starcoderHybrid but the end we can just compare users in different feature flags.
Can you confirm what approach we do take here to do the A/B test ?

vscode/src/completions/create-multi-model-inline-completion-provider.ts
```diff
@@ -0,0 +1,163 @@
...
+                providerConfig,
+                statusBar,
+                completeSuggestWidgetSelection: config.autocompleteCompleteSuggestWidgetSelection,
+                formatOnAccept: config.autocompleteFormatOnAccept,
+                disableInsideComments: config.autocompleteDisableInsideComments,
+                triggerNotice,
+                isRunningInsideAgent: config.isRunningInsideAgent,
+                createBfgRetriever,
+                isDotComUser: isDotCom(authStatus.endpoint || ''),
+                noAnalytics: true,
```
valerybugakov, line 133: We already have a setting that allows us to disable all the telemetry. Can we remove the `noAnalytics` flag and use the `cody.telemetry.level` setting with the "off" value for internal testing?

valerybugakov: It would allow the removal of all the nested branching and the need to keep track of ignored events manually.

hitesh-1997: nice, thanks for pointing this out Valery, Added this condition in the config.
Also, ensured that we we are adding this for the [deep cloned](https://github.com/sourcegraph/cody/pull/4048/files#diff-4fae773057a4f852e7e51755bed0f6d9e6d117d7d892faf6e495d784f047d76eR106) `config`, so that we don't change global telemetry config. (verified by checking `telemetry-v2` events on the debug console)

valerybugakov: Awesome! Thank you 🙌

hitesh-1997:
> > This is shaping up nicely!! Some very minor comments inside but let's look into hiding the command for when the feature flag is not enabled like this: [#4048 (comment)](https://github.com/sourcegraph/cody/pull/4048#issuecomment-2097952074)
> > I'll stamp this preemptively though. :)
> 
> Hi @philipp-spiess I added the [config in the](https://github.com/sourcegraph/cody/pull/4048/files#diff-99d7a3c2c2975fcaf8184c4639922f5ef6c915610b6f8e15ffa8fb8b329828d8R423) `config.cody.autocomplete.experimental.multiModelCompletions ` `package.json` but still see the option "Get Completions for multiple autocomplete models" even when vscode setting does not have this configured. Am i missing something here ?

Hi @valerybugakov 
If you have some time could you please give this a look. I added `config.cody.autocomplete.experimental.multiModelCompletions` in [package.json here](https://github.com/sourcegraph/cody/pull/4048/files#diff-99d7a3c2c2975fcaf8184c4639922f5ef6c915610b6f8e15ffa8fb8b329828d8R423) but, in the debugger mode I still see the autocomplete option. I tried reading the vscode documentation on this but not sure why is it still reflecting in the UI. Could it be because the command is cached in my vscode, could you try start debugger and see you are also able to see the command ?

vscode/src/completions/providers/create-provider.ts
philipp-spiess: I think it doesn't matter too much for the 5% use case but if we want to run a full-scale 50% test it would be good if we have a comparison (in this case the "starCoderHybrid" branch match the same number of users. So this approach looks good!

valerybugakov:
Hey @hitesh-1997, it is a great question and I got into the rabbit hole of figuring out why this happens. Here are my findings 
- https://github.com/sourcegraph/cody/pull/4155

PR #4051, opened 2024-05-04
use cmdk for existing @-mention UI

sqs:
No (or very minor) change in UI, no change in behavior. Just uses `cmdk` instead of our own custom menu list for @-mentions.

image

## Test plan

CI

### DIFF
--- a/vscode/webviews/promptEditor/plugins/atMentions/OptionsList.tsx
+++ b/vscode/webviews/promptEditor/plugins/atMentions/OptionsList.tsx
@@ -8,7 +8,7 @@ import {
     parseMentionQuery,
 } from '@sourcegraph/cody-shared'
 import { clsx } from 'clsx'
-import { type FunctionComponent, useLayoutEffect, useRef } from 'react'
+import { type FunctionComponent, useEffect, useRef } from 'react'
 import {
     FILE_HELP_LABEL,
     FILE_RANGE_TOOLTIP_LABEL,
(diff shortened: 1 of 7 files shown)


vscode/webviews/promptEditor/plugins/atMentions/OptionsList.tsx
```diff
@@ -31,46 +32,74 @@ export const OptionsList: FunctionComponent<
...
-    const ref = useRef<HTMLUListElement>(null)
-    // biome-ignore lint/correctness/useExhaustiveDependencies: Intent is to run whenever `options` changes.
-    useEffect(() => {
-        // Scroll to top when options change because the prior `selectedIndex` is invalidated.
-        ref?.current?.scrollTo(0, 0)
-        setHighlightedIndex(0)
-    }, [options])
+    const ref = useRef<HTMLDivElement>(null)
+
+    // Scroll selected into view. Needs `setTimeout` because we need to wait for all DOM mutations.
```
philipp-spiess, line 37: > Needs `setTimeout` because we need to wait for all DOM mutations.

DOM mutations are synchronous, so I think we're waiting for cmd+k to make these mutations and not for the mutations itself. I wonder if cmd+k uses a `useEffect` or so, did you try to use `useEffect` instead of `useLayoutEffect` (AFAIK children effects will be run first)

philipp-spiess (approved):

vscode/webviews/promptEditor/plugins/atMentions/OptionsList.tsx
sqs: cmdk uses its useScheduleLayoutEffect scheduler, which uses useLayoutEffect. I changed this to use useEffect and it still works, but it does not work without a setTimeout.

I am not sure of a better workaround atm. We need to use cmdk controlled because we need the Lexical editor to be able to know the selected value because LexicalMenu expects to be able to trigger insertion of it at any time from its own code. And then we bump into the issue alluded to at https://github.com/pacocoursey/cmdk/issues/252 where controlled cmdk does not autoscroll to the selected item.

PR #4052, opened 2024-05-05
only run windows e2e on main

sqs:
This test is flaky and fails often. We should fix it, but in the meantime, it slows us down on every push and PR waiting for it to finish before we can rerun other actions.

## Test plan

CI


burmudar (approved):

sqs:
@dominiccooney feel free to revert if you don't think this is a good idea.

sqs:
Whoops, this change was incorrect. Fixing.

PR #4053, opened 2024-05-05
handle mention ranges in the parser more cleanly

sqs:
This centralizes all @-mention query range parsing logic (eg the `:12-34` suffix) in `parseMentionQuery`.

## Test plan

CI


sqs:
The removed `range` field was not used by JetBrains: https://sourcegraph.com/github.com/sourcegraph/jetbrains@4dc5732f0525229f021b8157327b4829c70aefd1/-/blob/src/main/java/com/sourcegraph/cody/PromptPanel.kt?L199.

philipp-spiess (approved):
Awesome!

PR #4055, opened 2024-05-06
only run ubuntu e2e tests on PR push, run all on main

sqs:
- Consolidates macOS e2e test runner now that we are programmatically determining the matrix and can do that
- Runs macOS e2e tests on the normal (non-large) macOS runner, which takes longer but is fine since these no longer block
- Runs Ubuntu e2e tests un-sharded on main, which is fine because these don't block (if this becomes annoying, we can easily return to sharding these on main)

(Trying again on #4052)

## Test plan

CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,9 +9,9 @@ on:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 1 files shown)


sqs:
@dominiccooney WDYT about this?

dominiccooney (approved):
👍 💯

PR #4056, opened 2024-05-06
Use unit test for starcoder-hybrid trigger kind logic

olafurpg:
There was no need for an expensive integration test to cover this logic.

## Test plan
Green CI.


olafurpg:
The failing ubuntu 18 test should pass after merging https://github.com/sourcegraph/cody/pull/4057

vscode/src/completions/get-inline-completions.ts
```diff
@@ -144,6 +144,10 @@ export enum TriggerKind {
     SuggestWidget = 'SuggestWidget',
 }
 
+export function allTriggerKinds(): TriggerKind[] {
+    return [TriggerKind.Automatic, TriggerKind.Hover, TriggerKind.Manual, TriggerKind.SuggestWidget]
+}
```
philipp-spiess, line 149: Let's move this into the test function if we don't use it in here. Otherwise we might end up bundling it for no reason in the production build.

philipp-spiess (approved):
Thank you 🙇

PR #4061, opened 2024-05-06
Fix various small bugs related to inline edits

pkukielka:
## Changes

* Properly pass params of the `textDocument/show` notfication.
* Fix check for `editor.edit` which unexpectedly returns a null if action succeeds. 

## Test plan

Manual tests on the Jetbrains extension side.


vscode/src/non-stop/FixupController.ts
```diff
@@ -155,9 +155,10 @@ export class FixupController
         }
 
         editor.revealRange(task.selectionRange)
```
pkukielka, line 157: @umpox @abeatrix 
I could benefit greatly from someone with more VSCode knowledge looking at this.
`edit` interface looks like this:

```ts
edit(callback: (editBuilder: TextEditorEdit) => void, options?: { readonly undoStopBefore: boolean; readonly undoStopAfter: boolean }): Thenable<boolean>;
```

But it still returns null (but the operation succeeds).

abeatrix: Could it because the editBuilder is no longer valid when it returns? I think as long as it returns a non-empty value it should be valid, unless you're referring to something else?

pkukielka: But `editBuilder` is a parameter given to our callback, so it does not seem we have any control over it?
Also `editBuilder.replace` does not return anything (return type is `void`) so I guess result of `editor.edit` maybe depends on if any exception was thrown.

pkukielka: Either way I'm going to merge it as it's not more broken this way than it is right now.
I added a TODO to not forget about it completely.

Issue #4008, opened 2024-05-01
bug: ollama: wrong response  parsing


sergei-dyshel:
### Version

v1.16.0

### Describe the bug

When running against remote Ollama instance (model `deepseek-coder:6.7b`), code completion doesn't work and I see the following errors in Cody log:
```
█ CompletionLogger:onError: {"type":"code-completion","endpoint":"http://34.254.171.114:11434/api/generate","status":"error","duration":1007,"err":"Unexpected end of JSON input"} 
█ getInlineCompletions:error: Unexpected end of JSON input SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Object.complete (/Users/sergei/.vscode/extensions/lib/shared/src/llm-providers/ollama/completions-client.ts:71:39)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at generatorWithTimeout (/Users/sergei/.vscode/extensions/sourcegraph.cody-ai-1.16.0/src/completions/utils.ts:84:37)
    at fetchAndProcessDynamicMultilineCompletions (/Users/sergei/.vscode/extensions/sourcegraph.cody-ai-1.16.0/src/completions/providers/fetch-and-process-completions.ts:91:47)
    at async Promise.all (index 0)
    at zipGenerators (/Users/sergei/.vscode/extensions/sourcegraph.cody-ai-1.16.0/src/completions/utils.ts:34:21)
    at generateCompletions (/Users/sergei/.vscode/extensions/sourcegraph.cody-ai-1.16.0/src/completions/request-manager.ts:118:34)
```

### Expected behavior

Completion should work.

### Additional context

The relevant code is in https://github.com/sourcegraph/cody/blob/c9ee1119b73eb1b9724553f3a103c35332e84a54/lib/shared/src/llm-providers/ollama/completions-client.ts#L57-L72.

I've added a breakpoint in `JSON.parse(chunkString)` and see that parsing fails on `chunkString` which looks like:
```
[...]

PR #4066, opened 2024-05-07
Ollama: handle incomplete response chunks

valerybugakov:
- Closes https://github.com/sourcegraph/cody/issues/4008
- In the Ollama autocomplete client, ignore cut-in-the-middle response chunks and continue streaming.

## Test plan

Tested locally.

### DIFF
--- a/lib/shared/src/llm-providers/ollama/completions-client.ts
+++ b/lib/shared/src/llm-providers/ollama/completions-client.ts
@@ -68,11 +68,21 @@ export function createOllamaClient(
                 }
 
                 for (const chunkString of chunk.toString().split(RESPONSE_SEPARATOR).filter(Boolean)) {
-                    const line = JSON.parse(chunkString) as OllamaGenerateResponse
+                    let line: OllamaGenerateResponse
+
+                    try {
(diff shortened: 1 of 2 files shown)


abeatrix (approved):

lib/shared/src/llm-providers/ollama/completions-client.ts
```diff
@@ -68,11 +68,21 @@ export function createOllamaClient(
                 }
 
                 for (const chunkString of chunk.toString().split(RESPONSE_SEPARATOR).filter(Boolean)) {
-                    const line = JSON.parse(chunkString) as OllamaGenerateResponse
+                    let line: OllamaGenerateResponse
+
+                    try {
+                        line = JSON.parse(chunkString) as OllamaGenerateResponse
+                    } catch {
+                        // ignore JSON.parse() errors most probably caused by incomplete chunk and wait for the next one.
```
sergei-dyshel, line 76: There's something I still don't understand in this logic. If chunk is a partial JSON (not parseable), we need to concatenate next chunks together until they form a valid JSON. Where is the concatenation part?

PR #4067, opened 2024-05-07
Agent: clean up HTTP recording tech debt

olafurpg:
Fixes CODY-1394

- Refactors the code to reuse testing credentials
- Updates outdated comments about updating recordings
- Improves error messages

With these changes, it should hopefully be a bit easier to make big sweeping change like https://github.com/sourcegraph/cody/pull/4010

## Test plan

Green CI, this PR only refactors testing code


vscode/src/testutils/polly.ts
```diff
@@ -31,7 +31,7 @@ export function startPollyRecording(userOptions: PollyOptions): Polly {
         expiryStrategy: options.recordingExpiryStrategy,
         expiresIn: options.expiresIn,
         persisterOptions: {
-            keepUnusedRequests: options.keepUnusedRecordings ?? true,
+            keepUnusedRequests: options.keepUnusedRecordings ?? false,
```
valerybugakov, line 34: Does this mean that if we use `.only` to debug a specific agent test, all the unrelated recordings will be removed by default? This might be a frustrating experience for somebody unfamiliar with the underlying options.

valerybugakov (approved):
LGTM. Left one question inline.

vscode/src/testutils/polly.ts
olafurpg: You can configure this behavior with `CODY_KEEP_UNUSED_RECORDINGS`, and it's btw documented in the agent readme along with other helpful commands. You can also run `./agent/scripts/reset-recordings-to-main.sh` to recover the original recordings before opening a PR (also documented in the readme).

olafurpg: I reverted this change so the old behavior remains

PR #4071, opened 2024-05-07
Edit Input: Trigger doc and test commands in 1 click

umpox:
## Description

This PR:
- Updates the Edit Input commands to action with a single click/keypress.
  - The range is driven from the "Range" section of the input.
  - Added "Select a Symbol..." quickpick that is opened via the "Range" section
    - Shows all symbols in the doc
    - Focuses on the nearest one to the users' cursor

Loom: https://www.loom.com/share/36a6c6701c1d40808389f69327923b03

## Test plan

1. Open Edit input
2. Select "Document Code" or "Generate Tests"
3. Expect the command actions in a single keypress, and works on the current range
4. Open Edit Input
5. Select "Range" and then "Select a Symbol"
6. Select a valid symbol
7. Select "Document Code" or "Generate Tests"
3. Expect the command actions in a single keypress, and works on the selected range


vscode/src/edit/input/get-items/range-symbols.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Finds the symbol item that is closest to the initial cursor position.
+     * Note: Clone the array to avoid mutating the original symbolItems
+     */
+    const activeItem = [...symbolItems].sort(
+        (a, b) =>
+            getMinimumDistanceToRangeBoundary(initialValues.initialCursorPosition, a.range) -
+            getMinimumDistanceToRangeBoundary(initialValues.initialCursorPosition, b.range)
+    )[0]
```
dominiccooney, line 43: You can find the minimum in O(n) and without allocating:

```suggestion
    const activeItem = symbolItems.reduce(
        (a, b) =>
            getMinimumDistanceToRangeBoundary(initialValues.initialCursorPosition, a.range) <
            getMinimumDistanceToRangeBoundary(initialValues.initialCursorPosition, b.range) ? a : b,
        symbolItems[0]
    )
```

dominiccooney (approved):
Well worth an e2e test so we don't regress this.

CC @steveyegge FYI

vscode/src/edit/input/get-items/range-symbols.ts
umpox: Yep, thanks

umpox:
@dominiccooney Agree on the e2e tests, have added some that specifically cover the range + model input behaviour

PR #4076, opened 2024-05-07
TSC Graph Context: revamp snippet collection

olafurpg:
Fixes CODY-1331

This PR does sweeping changes in the `tsc-mixed` context retriever related to:

- How we identify relevant symbols to the request location
- How we render symbols

Along the way, we have new testing infrastructure in `tsc-retriever.test.ts` that provides super fast edit-test feedback loops while testing almost-production-like environments. This PR removes most of the integration tests since the unit tests have such good coverage now.

## Test plan

See updated test cases.


olafurpg:
I'm testing this with a local instance and spotted some blocking issues (perf-related, UI freezes). Investigating...

vscode/src/completions/context/retrievers/tsc/relevantTypeIdentifiers.ts
```diff
@@ -38,46 +37,48 @@ export function pushTypeIdentifiers(result: ts.Node[], checker: ts.TypeChecker,
...
-        }
-    } else if (ts.isVariableDeclaration(node)) {
-        if (node.type) {
-            pushTypeIdentifiers(result, checker, node.type)
-        }
-    } else if (ts.isTypeLiteralNode(node)) {
-        for (const member of node.members) {
-            pushTypeIdentifiers(result, checker, member)
-        }
+        result.push(...rightmostIdentifier(node.expression))
```
dominiccooney, line 40: Just so I understand, the rightmost identifier of `x.push(y)` is `push`, and downstream the graph context will find consider things named that as candidates?

This method is called `pushTypeIdentifiers` ... call expression rightmost expression doesn't seem very *type*-y.

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
```diff
@@ -340,3 +333,133 @@ function currentDirectory(): string | undefined {
...
+            return false
+        }
+        if (this.isAdded.has(sym)) {
+            return false
+        }
+        if (this.formatter.isRendered.has(sym)) {
+            // Skip this symbol if it's a child of a symbol that we have already
+            // formatted.  For example, if we render `interface A { a: number }`
+            // then we don't need to render `(property) A.a: number` separately
+            // because it's redunant with the interface declaration.
```
dominiccooney, line 376:
```suggestion
            // because it's redundant with the interface declaration.
```

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
```diff
@@ -340,3 +333,133 @@ function currentDirectory(): string | undefined {
...
+            return false
+        }
+        this.isAdded.add(sym)
+        // Symbols with multiple declarations are normally overloaded
+        // functions, in which case we want to show all available
+        // signatures.
+        let isAdded = false
+        for (const declaration of sym.declarations ?? []) {
+            if (isStdLibNode(declaration)) {
+                // Skip stdlib types because the LLM most likely knows how
```
dominiccooney, line 386: Graph edges like this could be an interesting opportunity for package context at some point, @thenamankumar

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
```diff
@@ -340,3 +333,133 @@ function currentDirectory(): string | undefined {
...
+                case ts.SyntaxKind.NamespaceImport:
+                    continue
+            }
+            if (this.isDone()) {
+                continue
+            }
+            const sourceFile = declaration.getSourceFile()
+            const start = sourceFile.getLineAndCharacterOfPosition(declaration.getStart())
+            const end = sourceFile.getLineAndCharacterOfPosition(declaration.getEnd())
+            const { formatted: content, queue } = this.formatter.formatSymbolWithQueue(sym)
```
dominiccooney, line 414: Just curious, but what do you think about:

1. Formatting snippets more compactly. Indents eat tokens, do they help performance? On whitespace sensitive languages like Python and Haskell, can we do better?
2. There's value in comments and this strips that out. Have you done any experiments leaving comments in?

dominiccooney (approved):
Looks good to me, some idle curiosity questions inline.

vscode/src/completions/context/retrievers/tsc/relevantTypeIdentifiers.ts
olafurpg: I could have written this as
```
const identifier = rightmostIdentifier(node.expression)
if (identifier) result.push(identifier)
```

but was lazy and made it return `ts.Node[]` instead so I could do `push(...)`

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
olafurpg: I'm planning to add `scip-typescript` down the road to give us structured information about the referenced symbols (including external dependencies down to npm package name+version). I've focused on the more narrow use-case of LLM context generation and have delayed adding scip-typescript until we have a stronger use-case.

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
olafurpg: One of my motivations to hand-roll symbol formatting is exactly to experiment with different strategies here. Without offline evaluations, we are guessing blind, though. A/B tests take too long to run and often struggle to give us a significant-enough sample size to make a decision. My plan is to add offline evaluations so we can iterate on these ideas much faster.

PR #4081, opened 2024-05-08
Search: Add refresh buttons to Search Code menu

abeatrix:
RE: https://sourcegraph.slack.com/archives/C05AGQYD528/p1715122290764839?thread_ts=1715122026.830309&cid=C05AGQYD528

- replace the input box for search query with quickPick 
  - as input box does not allow us to add buttons
- add buttons to refresh index to the quick pick menu

## Test plan

1. Click on Search Code (Beta) on your sidebar
2. Verify the refresh buttons are showing up on the upper right of the menu
3. Search for something to ensure the search feature still works

### After

image

### Before

No buttons on the upper right

image

### DIFF
--- a/vscode/src/search/SearchViewProvider.ts
+++ b/vscode/src/search/SearchViewProvider.ts
@@ -203,9 +203,7 @@ export class SearchViewProvider implements vscode.Disposable {
         const quickPick = vscode.window.createQuickPick()
         quickPick.placeholder =
             'Search… (e.g. "password hashing", "connection retries", a symbol name, or a topic)'
-        quickPick.items = [
-            { label: 'Search for code using a natural language query.', alwaysShow: true },
-        ]
+        quickPick.items = []
         quickPick.title = 'Natural Language Code Search (Beta)'
         quickPick.buttons = searchQuickPickButtons
         quickPick.matchOnDescription = false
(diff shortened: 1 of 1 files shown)


vscode/src/search/SearchViewProvider.ts
```diff
@@ -200,13 +200,25 @@ export class SearchViewProvider implements vscode.Disposable {
     }
 
     private async getSearchQueryInput(): Promise<void> {
-        const input = await vscode.window.showInputBox({
-            title: 'Natural Language Code Search (Beta)',
-            prompt: 'Search for code using a natural language query, such as "password hashing", "connection retries", a symbol name, or a topic.',
+        const quickPick = vscode.window.createQuickPick()
+        quickPick.placeholder =
+            'Examples: "password hashing", "connection retries", a symbol name, or a topic.'
```
toolmantim, line 205:
```suggestion
            'Search… (e.g. "password hashing", "connection retries", a symbol name, or a topic)'
```

vscode/src/search/SearchViewProvider.ts
```diff
@@ -200,13 +200,25 @@ export class SearchViewProvider implements vscode.Disposable {
...
+        quickPick.items = [
```
toolmantim, line 206: Do we have to have items, or can it be blank? Might be better just having no items to start with if we don't need to?

toolmantim (approved):

vscode/src/search/SearchViewProvider.ts
abeatrix: It can be blank! Updated to remove the current item that i was using as a "button":

image

### Before

image


Thanks for the feedback Tim!

PR #4083, opened 2024-05-08
Chat: exclude history items from UI

abeatrix:
FOLLOW-UP https://github.com/sourcegraph/cody/pull/3929

Main change is to add `used` and `ignored` context lists for UI display.

Currently, context items from older chat messages are showing up in the last submitted message in the UI after https://github.com/sourcegraph/cody/pull/3929. 
Ideally, it should only display context items added for the lastest message only.

Summary: 

The DefaultPrompter now maintains separate lists for used and ignored context items. These lists are intended for UI display purposes and do not affect the actual prompt construction. The used list contains valid context items added from the last human message, while the ignored list contains user-specified context items that were too large to include.

## Test plan

I've also updated the unit test to reflect this change.
No agent recordings update is required as this change should only affect UI when we display which context is used for the last human-submitted message.

1. Submit a message in Chat with enhanced context enabled.
2. Ask a follow-up question to confirm that the context from your first question is not displayed under your second question.

### After

1st question.

image

2nd question: context from the 1st question not included.

image

### Before

1st question.

image

2nd question: Context from the 1st question is also being displayed here

image

### DIFF
--- a/vscode/src/chat/chat-view/prompt.ts
+++ b/vscode/src/chat/chat-view/prompt.ts
@@ -73,23 +73,20 @@ export class DefaultPrompter implements IPrompter {
             const ignoredContext = { user: 0, enhanced: 0, transcript: 0 }
 
             // Add context from new user-specified context items, e.g. @-mentions, @-uri
-            const newUserContextMessages = await promptBuilder.tryAddContext(
-                'user',
-                this.explicitContext
-            )
-            ignoredContext.user += newUserContextMessages.ignored.length
+            const newUserContext = await promptBuilder.tryAddContext('user', this.explicitContext)
+            ignoredContext.user += newUserContext.ignored.length
 
+            // Lists of context items added from the last human message.
+            // NOTE: For UI display only, not used in the prompt.
+            const context = { used: [] as ContextItem[], ignored: [] as ContextItem[] }
+            // List of valid context items added from the last human message
+            context.used.push(...newUserContext.added.map(c => ({ ...c, isTooLarge: false })))
             // NOTE: Only used for display excluded context from user-specifed context items in UI
-            const ignoredUserContext: ContextItem[] = newUserContextMessages.ignored.map(c => ({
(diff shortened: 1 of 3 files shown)


vscode/src/chat/chat-view/prompt.ts
```diff
@@ -73,23 +73,20 @@ export class DefaultPrompter implements IPrompter {
...
-
+            const newUserContext = await promptBuilder.tryAddContext('user', this.explicitContext)
+            ignoredContext.user += newUserContext.ignored.length
+
+            // Lists of context items added from the last human message.
+            // NOTE: For UI display only, not used in the prompt.
+            const context = { used: [] as ContextItem[], ignored: [] as ContextItem[] }
+            // List of valid context items added from the last human message
+            context.used.push(...newUserContext.added.map(c => ({ ...c, isTooLarge: false })))
             // NOTE: Only used for display excluded context from user-specifed context items in UI
```
valerybugakov, line 84:
```suggestion
            // NOTE: Only used for display excluded context from user-specified context items in UI
```

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -73,23 +73,20 @@ export class DefaultPrompter implements IPrompter {
...
-                this.explicitContext
-            )
-            ignoredContext.user += newUserContextMessages.ignored.length
-
+            const newUserContext = await promptBuilder.tryAddContext('user', this.explicitContext)
+            ignoredContext.user += newUserContext.ignored.length
+
+            // Lists of context items added from the last human message.
+            // NOTE: For UI display only, not used in the prompt.
+            const context = { used: [] as ContextItem[], ignored: [] as ContextItem[] }
```
valerybugakov, line 81:
```suggestion
            const context: PromptInfo['context'] = { used: [], ignored: [] }
```

valerybugakov (approved):
LGTM

PR #4084, opened 2024-05-08
fix: ensure range is RangeData instance

abeatrix:
Follow-up https://github.com/sourcegraph/cody/pull/3929

This commit addresses an issue where the range property of a ContextItem object might not be an instance of the RangeData class. The changes ensure that the range property is always an instance of RangeData before performing any operations on it.

Currently, some contexts are not using the correct RangeData format, which would cause the checks to fail in `vscode/src/prompt-builder/unique-context.ts` with the error `Cannot read properties of undefined (reading: 'line').`:

Screenshot 2024-05-07 at 8 09 18 PM

To reproduce, try to build Cody from main with indexed search and local embeddings enabled, and then ask Cody a question: `tell me about SecretStorage?`

## Test plan

1. Build Cody from this branch and start debug mode
2. Open the code repository and make sure enhanced context is enabled, with search and embeddings status shown as `indexed`
3. Ask Cody: `tell me about SecretStorage?`
4. Verify your question was submitted successfully and you are able to get a response from Cody. Before this change, your question will be stuck in the context search phase, and your question will not be sent to the LLM.

image

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatModel.ts
+++ b/vscode/src/chat/chat-view/SimpleChatModel.ts
@@ -120,6 +120,12 @@ export class SimpleChatModel {
         return this.messages
     }
 
+    // De-hydrate because vscode.Range serializes to `[start, end]` in JSON.
+    // TODO: we should use a different type for `getMessages` to make the range hydration explicit.
+    public getDehydratedMessages(): readonly ChatMessage[] {
+        return this.messages.map(prepareChatMessage)
+    }
+
     public getChatTitle(): string {
         if (this.customChatTitle) {
             return this.customChatTitle
--- a/vscode/src/chat/chat-view/prompt.ts
+++ b/vscode/src/chat/chat-view/prompt.ts
@@ -60,7 +60,7 @@ export class DefaultPrompter implements IPrompter {
             }
 
             // Add existing chat transcript messages
-            const reverseTranscript = [...chat.getMessages()].reverse()
+            const reverseTranscript = [...chat.getDehydratedMessages()].reverse()
             const transcriptLimitReached = promptBuilder.tryAddMessages(reverseTranscript)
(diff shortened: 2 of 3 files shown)


valerybugakov (commented):
Cannot repro the issue locally based on the instructions. Debugging it together with @abeatrix.

abeatrix:
@valerybugakov Confirmed the latest commit works for me: 
image
Thank you so much for your help on this 😭

valerybugakov (approved):
Not sure if it's legal for me to approve this anymore, but let's merge it and discuss the follow-up steps to fix the untyped serialization boundary!

vscode/src/chat/chat-view/SimpleChatModel.ts
```diff
@@ -120,6 +120,12 @@ export class SimpleChatModel {
         return this.messages
     }
 
+    // De-hydrate because vscode.Range serializes to `[start, end]` in JSON.
+    // TODO: we should use a different type for `getMessages` to make the range hydration explicit.
+    public getDehydratedMessages(): readonly ChatMessage[] {
+        return this.messages.map(prepareChatMessage)
+    }
```
valerybugakov, line 127: Hey @dominiccooney and @sqs, what are your thoughts on addressing [the untyped boundary issue](https://github.com/sourcegraph/cody/pull/3269#issuecomment-1965413273)? 

It's tricky to debug if you do not have the relevant context on why it might happen. Typescript [types](https://github.com/sourcegraph/cody/blob/f3ec5d09c0895dc8902127ce8a5c565d20ce9f7f/lib/shared/src/common/range.ts#L11-L14) say it's impossible. We spent some extra time digging into the history and debugging this locally to find the culprit. Is there a plan to make serialization/deserialization explicit via the type system? Should scope one?

PR #4088, opened 2024-05-08
Fix undefined files handling for non-vsc editors

pkukielka:
## Changes

Added new client capability which is partial support for `undefined` files.
Some editors like IntelliJ does not support in-memory files, but are able to create real files on the disk if given url is a valid path.

Jetbrains PR: https://github.com/sourcegraph/jetbrains/pull/1474

## Test plan

Manual tests with Jetbrains extension and `Generate Unit Test` command.

### DIFF
--- a/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/ClientCapabilities.kt
+++ b/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/ClientCapabilities.kt
@@ -10,7 +10,7 @@ data class ClientCapabilities(
   val progressBars: ProgressBarsEnum? = null, // Oneof: none, enabled
   val edit: EditEnum? = null, // Oneof: none, enabled
   val editWorkspace: EditWorkspaceEnum? = null, // Oneof: none, enabled
-  val untitledDocuments: UntitledDocumentsEnum? = null, // Oneof: none, enabled, onlyRealFiles
+  val untitledDocuments: UntitledDocumentsEnum? = null, // Oneof: none, enabled
   val showDocument: ShowDocumentEnum? = null, // Oneof: none, enabled
   val codeLenses: CodeLensesEnum? = null, // Oneof: none, enabled
   val showWindowMessage: ShowWindowMessageEnum? = null, // Oneof: notification, request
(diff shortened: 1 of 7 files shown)


mkondratek (approved):

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/ClientCapabilities.kt
```diff
@@ -10,7 +10,7 @@ data class ClientCapabilities(
   val progressBars: ProgressBarsEnum? = null, // Oneof: none, enabled
   val edit: EditEnum? = null, // Oneof: none, enabled
   val editWorkspace: EditWorkspaceEnum? = null, // Oneof: none, enabled
-  val untitledDocuments: UntitledDocumentsEnum? = null, // Oneof: none, enabled
```
mkondratek, line 13: I see the warning still appears: 
```
Cody by Sourcegraph: █ vscode.workspace.openTextDocument: unable to read non-file URI: untitled:/Users/mkondratek/runIdeProjects/jetbrains/src/test/java/MainTest.java
```

can we hide it when `onlyRealFiles`? that would reduce the noise in the logs

pkukielka: Fixed

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -160,7 +160,7 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
                     'File does not exist',
                     uri.toString()
                 )
-            } else if (uri.scheme === 'file') {
+            } else if (uri.scheme === 'file' || uri.scheme === 'untitled') {
                 // Read the file content from disk if the user hasn't opened this file before.
                 const buffer = await fspromises.readFile(uri.fsPath, 'utf8')
                 document.underlying.content = buffer.toString()
```
mkondratek, line 166: This is more than changing the logging. We are modifying `document.underlying.content` now. idk if that's ok

mkondratek (approved):

agent/src/AgentWorkspaceDocuments.ts
mkondratek: verified - it still works 😄

pkukielka: That change was a bit stupid on my side, it is better now :)

PR #4090, opened 2024-05-08
Enterprise Cody Ignore: E2E test

philipp-spiess:
This PR adds E2E tests for the cody ignore logic. 

It currently does not test @-mention related behavior as well as the repo picker but since I’m focusing on the latecy issues, I thought it's better to get this test merged into main rather than waiting for the rest.

## Test plan

- CI green


philipp-spiess:
@dominiccooney This PR is still missing some stuff and I wasn’t able to wrap it up but if you can get to it before I’m back, feel free to commandeer.

- I’m currently investigating why `alt+\\` doesn’t trigger a manual completion (it works in “headed” mode of course)
- We still need the chat related tests: @-mention and context-picker

philipp-spiess:
and cc @valerybugakov in case you have availability

philipp-spiess:
@valerybugakov @dominiccooney Took a quick look as to why the tests are failing. Turns out during the E2E test, we don't run in test or dev mode of the extension and thus the command we use to overwrite cody ignore doesn't seem to be working. I wonder if instead of using the overwrite command we shouldn't just mock the CodyIgnore API instead. Will try this out!

philipp-spiess:
@dominiccooney @valerybugakov Can I give you a friendly nudge to take a look at this? 🙂

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -612,13 +612,15 @@ export class SourcegraphGraphQLAPIClient {
...
         if (!enabled || !isValidVersion) {
+            console.log('NO?')
             return INCLUDE_EVERYTHING_CONTEXT_FILTERS
         }
 
         const response =
             await this.fetchSourcegraphAPI<APIResponse<ContextFiltersResponse | null>>(
                 CONTEXT_FILTERS_QUERY
             )
+        console.log({ response })
```
dominiccooney, line 623: Seems nice to log this with a bit more detail, since it only happens hourly.

dominiccooney (approved):
Awesome, thank you for circling back to finish this! 🎉

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -612,13 +612,15 @@ export class SourcegraphGraphQLAPIClient {
         const insiderBuild = version.length > 12 || version.includes('dev')
         const isValidVersion = insiderBuild || semver.gte(version, minimumVersion)
         if (!enabled || !isValidVersion) {
+            console.log('NO?')
```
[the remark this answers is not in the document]
philipp-spiess: 😬 Sorry!

lib/shared/src/sourcegraph-api/graphql/client.ts
philipp-spiess: I think we're already logging the result of this in the context filter provider (so whenever we update the policy we'll see it in the debug logs). This was just me doing `console.log` debugging while figuring out how to mock these in the E2E test 😫

philipp-spiess, line 623:
```suggestion
```

PR #4091, opened 2024-05-08
Automatically start Embeddings indexing

rafax:
Start Embedding indexing process on extension startup provided that Sourcegraph embedding API is used instead of OpenAI. Additionally, restarts embedding indexing process if a partial index is found.

Currently behind a dedicated feature flag, default off.

This doesn't remove the "Embedding consent" functionality as it is still needed for OpenAI (the current default). 

## Test plan

- tested locally with un-indexed/partially indexed/indexed repos
  - un-indexed repo will automatically be indexed
  - partially indexed repo will continue indexing
  - fully indexed repo will be used for search (without re-indexing)
- tested locally with repos not eligible for embedding
- tested locally with auto-indexing flag off (caught an issue with indexing status not being reflected correctly)
- tested locally with sourcegraph-embeddings flag off (making this PR a no-op)


vscode/src/main.ts
```diff
@@ -310,6 +310,11 @@ const register = async (
         searchViewProvider.initialize()
     }
 
+    if (localEmbeddings) {
+        // kick-off embeddings initialization
+        localEmbeddings.start()
```
rafax, line 315: Question to reviewers - is this the right place to start this? We start `symf` indexing in the block above, and the call to `.start()` takes <1ms in local testing, so this should be fine.

vscode/src/local-context/local-embeddings.ts
```diff
@@ -161,7 +165,13 @@ export class LocalEmbeddingsController
         await this.getService()
         const repoUri = vscode.workspace.workspaceFolders?.[0]?.uri
         if (repoUri && isFileURI(repoUri)) {
-            await this.eagerlyLoad(repoUri)
+            const loadedOk = await this.eagerlyLoad(repoUri)
+            if (!loadedOk) {
+                // failed to load the index, let's see if we should start indexing
+                if (this.autoIndexingEnabled && this.modelConfig.provider === 'sourcegraph') {
+                    this.index()
```
thenamankumar, line 172: I am not an expert in this part of the code, so I need someone to verify my concern here. 

It might be the case that `this.eagerlyLoad` will return true, even if there are partial embeddings present for the repo. In that case, we will need to call `this.indexRetry`. 

But if it only returns true when the embeddings are fully generated then this looks good. 

I see it mentioned in the test plan that you have covered this case but I just wanted to double-check.

thenamankumar: Also a shameless plug: https://www.loom.com/share/4b2fb975f89447dcabeb2cce5c956ded?sid=3b0cd033-eb04-4453-9fe5-7f35f0c6ec20

rafax: Thank you for checking and raising this - the existing treatment for partially indexed repos is that we require user action to continue indexing (screenshot), I tested that that this PR doesn't change that logic, I planned to have a separate PR for "automatic re-start" once I understand why we don't automatically re-index in the first place (what comes to mind and what I saw in [code](https://github.com/sourcegraph/cody/blob/9dc9df32b80cbe11466e0ca7f285e40a21f7321e/vscode/src/main.ts#L305) is that some of the indexing processes can kill the agent, so automatic re-start could make Cody / agent crashloop?)
Screenshot 2024-05-09 at 09 05 10

vscode/src/main.ts
janhartman: This sounds acceptable to me.

janhartman (approved):

vscode/src/local-context/local-embeddings.ts
rafax: @dominiccooney (as the original author of this) - is there a specific reason why we don't automatically re-start indexing after the extension is restarted? Would you be ok with making that change?

thenamankumar (approved):
LGTM!

PR #4094, opened 2024-05-08
Chat UI: Adjust user avatar height to fit content

abeatrix:
Fix a minor css issue where the height of the user avatar is stretched by adjusting the container height to fit-content. 

Not sure if this is the best solution so any feedback is welcome :D 

image

image

## Test plan

Run the explain command and check the user avatar size to verify the user avatar is displaying correctly.

image

### DIFF
--- a/vscode/webviews/components/UserAvatar.module.css
+++ b/vscode/webviews/components/UserAvatar.module.css
@@ -6,4 +6,5 @@
     color: var(--vscode-inputOption-activeForeground);
     align-items: center;
     justify-content: center;
+    height: fit-content;
 }


sqs:
Ahh, I can repro this in the prelease build but not the release build.

Release build (v1.16):

image

Prerelease build:

image

I don't know if this is the best fix. Do you know what caused it to break between release and prerelease?

---

Also is this another bug in the model selector (seen in your screenshot, the lack of spaces)?
image

sqs (approved):
approving to unblock immediate fix if needed

abeatrix:
> Do you know what caused it to break between release and prerelease?

@sqs I THINK we started using tailwind in pre-release which has a [different base styles for img](https://sourcegraph.com/github.com/tailwindlabs/tailwindcss@5e737d858755a7149e17b5f3feaeb0d42865354d/-/blob/packages/tailwindcss/preflight.css?L318-326) that causes this change:

```css
/**
 * Constrain images and videos to the parent width and preserve
 * their intrinsic aspect ratio.
 *
 * https://github.com/mozdevs/cssremedy/issues/14
 */

img,
video {
  max-width: 100%;
  height: auto;
}
```

Good eyes on the model selector bug! Will look into the other bug first thing tomorrow 😃

PR #4095, opened 2024-05-08
TSC Graph Context: fix performance

olafurpg:
Fixes CODY-1434

Previously, the `tsc-mixed` context retriever made  the UI freeze because it was doing super inefficient things with context expansion. This PR fixes the problem so that the performance is back to normal.

## Test plan

Green CI.

I tested this change manually in large projects and confirmed that the latency overhead for tsc context-retriever is no longer huge. For example, it's 33ms here on line 1300 in the largest file in svelte/language-tools repo 
CleanShot 2024-05-15 at 14 40 24@2x

### DIFF
--- a/vscode/src/completions/context/retrievers/section-history/nextTick.ts
+++ b/vscode/src/completions/context/retrievers/section-history/nextTick.ts
@@ -1,0 +1,3 @@
+export function nextTick() {
+    return new Promise(resolve => process.nextTick(resolve))
+}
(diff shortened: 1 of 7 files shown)


olafurpg:
Good news: I've maybe the biggest performance mistake (not caching script snapshots) so loading the compiler is now only slow once (not on every keystroke!).

Bad news: loading the compiler is still too slow in large codebases (6s in sg/sg) and symbol collection is still too slow (~200ms) for a small file.

olafurpg:
Tagged stampers for review since this is a low-risk PR to merge. The TSC context is disabled by default

PR #4097, opened 2024-05-08
Edit: Track file deletions as task rejections

umpox:

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -531,6 +531,12 @@ export class FixupController
             })
         }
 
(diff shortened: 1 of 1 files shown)


umpox:
@abeatrix For the unit test one in particular, here's some additional testing:

1. If the file is unsaved, and then closed we log rejected
2. If the file is saved, we don't log anything
3. If the user makes a change to the file, we log accepted
4. If the user completely undo's the test change in the file, and then makes another change to the file, we log rejected

philipp-spiess (approved):
Love this! Especially the tests are :chef-kiss:

PR #4101, opened 2024-05-08
Capture traces for updateParseTreeOnEdit

philipp-spiess:
Capture traces for `updateParseTreeOnEdit`

## Test plan

- Trivial, CI will check

### DIFF
--- a/vscode/src/tree-sitter/parse-tree-cache.ts
+++ b/vscode/src/tree-sitter/parse-tree-cache.ts
@@ -76,6 +76,8 @@ function getLanguageIfTreeSitterEnabled(document: TextDocument): SupportedLangua
 export function updateParseTreeOnEdit(edit: vscode.TextDocumentChangeEvent): void {
     wrapInActiveSpan('updateParseTreeOnEdit', span => {
         span.setAttribute('sampled', true)
+        span.setAttribute('languageId', edit.document.languageId)
+
         const { document, contentChanges } = edit
         if (contentChanges.length === 0) {
             return
--- a/vscode/src/tree-sitter/parse-tree-cache.ts
+++ b/vscode/src/tree-sitter/parse-tree-cache.ts
@@ -3,6 +3,7 @@ import * as vscode from 'vscode'
 import type { TextDocument } from 'vscode'
 import type { default as Parser, Tree } from 'web-tree-sitter'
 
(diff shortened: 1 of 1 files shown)


philipp-spiess:
@valerybugakov Saw that there's also `observableParse`. Should we use this instead? Not sure why we would need two places but I assume this place can be very high frequency (since it's on every document change)

vscode/src/tree-sitter/parse-tree-cache.ts
```diff
@@ -73,41 +74,46 @@ function getLanguageIfTreeSitterEnabled(document: TextDocument): SupportedLangua
...
-            startPosition: startPoint,
-            oldEndPosition: oldEndPoint,
-            newEndPosition: newEndPoint,
-        })
-    }
-
-    const updatedTree = parser.safeParse(document.getText(), tree)
-    parseTreesPerFile.set(cacheKey, updatedTree)
+    wrapInActiveSpan('updateParseTreeOnEdit', span => {
+        span.setAttribute('sampled', true)
```
valerybugakov, line 78: Do we want to sample all these spans? The volume will be huge.

vscode/src/tree-sitter/parse-tree-cache.ts
```diff
@@ -73,41 +74,46 @@ function getLanguageIfTreeSitterEnabled(document: TextDocument): SupportedLangua
...
+                startIndex,
+                oldEndIndex,
+                newEndIndex,
+                startPosition: startPoint,
+                oldEndPosition: oldEndPoint,
+                newEndPosition: newEndPoint,
+            })
+        }
+
+        const updatedTree = parser.safeParse(document.getText(), tree)
```
valerybugakov, line 114: If we decide to sample all spans here, we can move `span.setAttribute('sampled', true)` and `span.setAttribute('languageId', languageId)` into the `observableParse` and use it here.

valerybugakov:
> Not sure why we would need two places, but I assume this place can be very high frequency

@philipp-spiess, that's right. I didn't instrument this part because it happens on every keystroke change.

olafurpg (approved):

vscode/src/tree-sitter/parse-tree-cache.ts
philipp-spiess: No I think that would probably be too large, I agree.

PR #4107, opened 2024-05-09
Tutorial: Temporarily disable sidebar link

umpox:
## Description

**Issue:**
- We have an issue with the balance of an A/B test. This appears to be because we evaluate a feature flag before we know that the user has authenticated. I'm going to temporarily disable the sidebar item, as we're mainly focused on showing this to the user on the first auth, this is just an additional input.

I can't replicate the issue myself reliably, but I'm expecting that this should solve the balance. I will create an issue to follow up on this when confirmed

## Test plan

1. N/A - Disabling feature that is only additional.
2. Check "Tutorial" does not appear in the sidebar

### DIFF
--- a/vscode/src/services/tree-views/support-items.ts
+++ b/vscode/src/services/tree-views/support-items.ts
@@ -1,4 +1,3 @@
-import { FeatureFlag } from '@sourcegraph/cody-shared'
 import { releaseType } from '../../release'
 import { version } from '../../version'
 import type { CodySidebarTreeItem } from './treeViewItems'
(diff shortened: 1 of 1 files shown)


abeatrix:
Hmmm I'm not sure I am following the issue here. Why removing it from sidebar would help? The sidebar should be refreshed on auth change, shouldn't it get the correct feature flag value after the refresh? (Ignore me if you're talking about something else 😂 )

umpox:
@abeatrix Yeah I think you're right, the sidebar does refresh but my thinking is that it can run before the first auth, and then cache the falsy value.

Really we need to update the feature flag cache to cache against the endpoint + the auth status, so we never end up using a cache value that doesn't match the auth.

I've found the actual issue I think: https://github.com/sourcegraph/cody/pull/4108

I'm going to close this

PR #4110, opened 2024-05-09
Cody Ignore: activate the `vscode.git` before Cody extension via `ensionDependencies`

valerybugakov:
- Removes the init delay making the git API instantly available in Cody.
- To ensure that the `vscode.git` extension is activated before your extension, add `extensionDependencies` ([docs](https://code.visualstudio.com/api/references/extension-manifest)) into the `package.json` of your extension:
- It's [safe](https://github.com/microsoft/vscode/blob/main/extensions/git/README.md#git-integration-for-visual-studio-code) to do because the Git extension is bundled with Visual Studio Code. It can be disabled but not uninstalled.

## Test plan

Tested locally.


umpox (approved):
Nice!

Issue #3952, opened 2024-04-25
VS Code: Cody stuck in 'Formatting...' when user run Document Code for a Java function.


jay-fibi:
### Version

v1.14.0

### Describe the bug

Platform: MacOS

**Steps to reproduce:**

1. Have Cody plugin installed in your VS Code editor
2. Login to Cody with Pro user
3. Open a Java source code file
4. Select code snippet from any function i.e. bubbleSort()
5. Right click, and select Cody > Document Code

### Observed behavior:
Documentation will be added but Cody keep showing 'Formatting...' never finish the process

### Expected behavior

Cody should not be stuck at 'Formatting...'

### Additional context

**Video link:**

https://drive.google.com/file/d/1QDoeF7ToYf14RgnKN0d9a70jmAJXrXpf/view?usp=drive_link

DrNeutron:
I have the same issue in python use of cody in VS Code.  Same symptoms as the video except in my case I also get an error pop up saying "Unexpected error in LSP server, see server's logs for details" Source Black Formatter.  The same error happens with both Black and Ruff.

jeronimd:
same, while using autopep8


Traceback (most recent call last):
  File "c:\Users\David\.vscode\extensions\ms-python.autopep8-2024.0.0\bundled\libs\pygls\protocol\json_rpc.py", line 443, in data_received
    self._data_received(data)
  File "c:\Users\David\.vscode\extensions\ms-python.autopep8-2024.0.0\bundled\libs\pygls\protocol\json_rpc.py", line 475, in _data_received
    json.loads(
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.3568.0_x64__qbz5n2kfra8p0\lib\json\__init__.py", line 359, in loads
    return cls(**kw).decode(s)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.3568.0_x64__qbz5n2kfra8p0\lib\json\decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.3568.0_x64__qbz5n2kfra8p0\lib\json\decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
[...]

PR #4111, opened 2024-05-09
Edit: Fix incorrect formatting with certain formatters

umpox:
## Description

fixes CODY-1449

closes https://github.com/sourcegraph/cody/pull/4111

closes https://github.com/sourcegraph/cody/issues/3952

Accidental regression from https://github.com/sourcegraph/cody/pull/3734

Wasn't caught due to the fact that VS Code has really poor type enforcement around `executeCommand` :(

## Test plan

1. Make Edits with only the default TypeScript language formatter selected
2. Check indentation is correct

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -7,6 +7,7 @@ import {
     type PromptString,
     displayPathBasename,
     getEditorInsertSpaces,
+    getEditorTabSize,
     telemetryRecorder,
 } from '@sourcegraph/cody-shared'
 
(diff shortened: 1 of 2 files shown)


vscode/src/non-stop/FixupController.ts
```diff
@@ -875,9 +876,9 @@ export class FixupController
                 'vscode.executeFormatDocumentProvider',
                 document.uri,
                 {
-                    tabSize: document.uri,
+                    tabSize: getEditorTabSize(document.uri, vscode.workspace, vscode.window),
                     insertSpaces: getEditorInsertSpaces(document.uri, vscode.workspace, vscode.window),
-                }
+                } satisfies vscode.FormattingOptions
```
valerybugakov, line 881: Why isn't it type-safe by default? 🥲

valerybugakov (approved):

vscode/src/non-stop/FixupController.ts
umpox: I know :(

PR #4112, opened 2024-05-09
Tutorial: Remove partial code from tutorial

umpox:
## Description

From my testing a blank line is fine here, and we don't want to show errors to the user from this file (p is not defined).

## Test plan

1. Open tutorial
2. Check step 1 works as expected

### DIFF
--- a/vscode/src/tutorial/content.ts
+++ b/vscode/src/tutorial/content.ts
@@ -20,7 +20,7 @@ export const getStepContent = (step: TutorialStepKey): string => {
                 """
                 def greet(name, surname):
                     """Greets a user with a simple message"""
-                    p
+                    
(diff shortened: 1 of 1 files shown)


vscode/src/tutorial/content.ts
```diff
@@ -20,7 +20,7 @@ export const getStepContent = (step: TutorialStepKey): string => {
                 """
                 def greet(name, surname):
                     """Greets a user with a simple message"""
-                    p
+
```
umpox, line 23: Award for the smallest diff ever. Does anyone know how to get VS Code to stop automatically removing these spaces on save?

I did `Save without formatting` but I'm worried we may accidentally regress this later

toolmantim (approved):

PR #4113, opened 2024-05-09
Edit: Add maximum timeout to formatter logic

umpox:
## Description

If an Edit takes longer than 1000ms to format, it's likely that something has gone wrong.

This is a poor UX so we will stop formatting automatically when this happens.
 
## Test plan

1. Tested by lowering the formatter timer to a small value (10ms)
2. Running edits
3. Checking that the formatting step is skipped

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -19,6 +19,7 @@ import { countCode } from '../services/utils/code-count'
 
 import { PersistenceTracker } from '../common/persistence-tracker'
 import { lines } from '../completions/text-processing'
+import { sleep } from '../completions/utils'
 import { getInput } from '../edit/input/get-input'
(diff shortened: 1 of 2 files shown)


vscode/src/non-stop/FixupController.ts
```diff
@@ -870,15 +870,27 @@ export class FixupController
...
+                    {
+                        tabSize: document.uri,
+                        insertSpaces: getEditorInsertSpaces(
+                            document.uri,
+                            vscode.workspace,
+                            vscode.window
+                        ),
+                    }
+                ),
+                new Promise<void>(resolve => setTimeout(resolve, formattingTimeout)),
```
valerybugakov, line 892: nit: we have a couple of sleep functions in the codebase, should we use one here?

```suggestion
                sleep(formattingTimeout),
```

valerybugakov (approved):

PR #4115, opened 2024-05-09
Cody Ignore: remove the Git extension shim

valerybugakov:
- Closes https://github.com/sourcegraph/cody/issues/4140
- Disables the Git extension shim in the agent. Documents places where we need to implement fallback methods.
- Fixes the Cody Ignore behavior in the agent where the shim returned empty values for most repository fields (e.g., remoteUrls).
- Functional changes:
    - The agent's `git` capability has been disabled.
    - We won't have commit ID logged in non-VS Code clients: https://github.com/sourcegraph/cody/blob/e01888ca5cf784ca5e118f8ef9204be2496b71c5/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts#L423
    - Follow-up required https://github.com/sourcegraph/cody/issues/4139
- Other follow-up issues:
    - https://github.com/sourcegraph/cody/issues/4165
    - https://github.com/sourcegraph/cody/issues/4136
    - https://github.com/sourcegraph/cody/issues/4137
    - https://github.com/sourcegraph/cody/issues/4138
- Related [Slack discussion](https://sourcegraph.slack.com/archives/C06Q9R098N4/p1715253166904809?thread_ts=1715226406.094479&cid=C06Q9R098N4).

## Test plan

CI


vscode/src/repository/remote-urls-from-tree-walk.ts
```diff
@@ -0,0 +1,105 @@
+import ini from 'ini'
```
valerybugakov, line 1: No functional changes. Only moved tree-walk functions in a separate file.

vscode/src/repository/remote-urls-from-tree-walk.test.ts
```diff
@@ -0,0 +1,195 @@
+import { describe, expect, it, vi } from 'vitest'
+import * as vscode from 'vscode'
+import { URI } from 'vscode-uri'
+
+import { gitRemoteUrlsFromTreeWalk } from './remote-urls-from-tree-walk'
+import { mockFsCalls } from './test-helpers'
+
+describe('gitRemoteUrlFromTreeWalk', () => {
```
valerybugakov, line 8: No changes. Only moved to a new file.

olafurpg (approved):
LGTM this seems like a good improvement in the fallback case. I’m still not sold this is a good long term solution. The long term plan with the git extension shim was to implement it with the client. IntelliJ has a git plugin with an API that we should be using for production settings in XXL repos. For example, at Twitter we required engineers to use a fork of git that was auto-configured in everybody’s IntelliJ setup. Given the focus of Cody Ignore is Enterprise customers, then we need to make sure our solutions work even for specialized repo setups.

olafurpg:
I left a comment in the linked issue, but will repeat here that we should not try to implement our own file watchers. File watching is very difficult to get right in XXL repos and we should rely instead on the IDEs built in file watching APIs.

olafurpg:
One option is to keep the existing got extension shim but use the “disabled” option from IntelliJ until we allow the client to implement the necessary parts of the git extension. It might not require a lot of work to pull out the information from IntelliJ, and then we don’t have to revert the removals in this PR.

vscode/src/repository/remote-urls-from-tree-walk.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        throw error
+    }
+}
+
+/**
+ * Reads the .git directory or file to determine the path to the git config.
+ */
+async function resolveGitConfigUri(uri: vscode.Uri): Promise<vscode.Uri | undefined> {
```
olafurpg, line 80: I think it's good to have this as a fallback solution when nothing else works (git extension is disabled, client doesn't git, ..) but I think it's a slippery slope to reimplement core features of git. My experience is that big companies end up using advanced git features that we may not be handling here. Also, this doesn't work for non-git systems like Perforce (which we support for several important customers).

We should really lean on using the higher-level version-control APIs that the IDEs provide (IntelliJ's plugin is excellent, and a ton of work has gone into making it scale for even the most difficult repos)

valerybugakov: > One option is to keep the existing got extension shim but use the “disabled” option from IntelliJ until we allow the client to implement the necessary parts of the git extension. 

I like it. We can keep only the "disabled" option until the integration is implemented. 

> We should really lean on using the higher-level version-control APIs that the IDEs provide

Agree with a hight level idea of leaning more towards the IDE provided functionality. We can figure out how to integrate it with the Git extension shim, when the time comes to prioritize it.

valerybugakov: 1. Created a follow-up https://github.com/sourcegraph/cody/issues/4165.
2. Reverted the git extension shim removal.
3. Changed the protocol to `git?: 'none' | 'enabled'` to the need to refactor other clients.
4. Added logic to throw errors if the git extension is used in the agent with a linked issue.

PR #4118, opened 2024-05-09
Remove illegal characters from V2 telemetry naming

akalia25:
This PR fixes an issue where some V2 telemetry names contained illegal characters(colon `:`) which cause them to get rejected [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph@d7dc2b9d9a1e7c44c5de59617b0dae4c7d969cee/-/blob/lib/telemetrygateway/v1/v1.go?L25-49) and not logged. (e.g. [sentry error](https://sourcegraph.sentry.io/issues/5282551582/?project=6583153)). 

## Test plan
CI and E2E tests


dominiccooney (approved):

dominiccooney:
Might be nice to add a check on the client side for multiple colons and fail closer to the source of the error, to prevent future double colons from creeping into the codebase.

akalia25:
> Might be nice to add a check on the client side for multiple colons and fail closer to the source of the error, to prevent future double colons from creeping into the codebase.

Agreed! I'll look into getting this setup. It'll be a much more sustainable solution.

PR #4121, opened 2024-05-09
VS Code: Patch Release 1.16.7

abeatrix:
## BRANCH OFF v1.16.6 - DO NOT MERGE TO MAIN

This pull request updates the logic for selecting a model in the `ModelSelectField` component to fix an issue where clicking on a non-default model on first click does not work as expected.

Issue:

In the `ModelSelectField` component, the `selectedModel` is derived from the `models` prop and represents the default model which everyone has access to.

The `onModelSelect` callback function is responsible for handling the selection of a new model. When a user clicks on a different model in the dropdown, this callback function is called with the newly selected model as a parameter.

The issue arose because the onModelSelect callback was using the `selectedModel` instead of the `model` parameter when checking for the `codyProOnly` condition and sending the telemetry event, which is still set as the initial value.

Fixes:

1. **Update model selection condition**: The condition for showing the Cody Pro badge and redirecting to the subscription page has been updated. Instead of checking the `codyProOnly` property of the `selectedModel`, it now checks the `codyProOnly` property of the `model` param being selected.
2. **Update event properties**: The `LLM_provider` property in the `CodyVSCodeExtension:chooseLLM:clicked` event is now set to the `model.model` value instead of `selectedModel.model`.
3. **Optimize dependency array**: The dependency array for the `onModelSelect` callback has been updated to remove the `selectedModel` dependency, as it is no longer used.

These changes ensure that the Cody Pro badge and subscription link are displayed correctly when selecting a Cody Pro-only model.

## Test plan

1. Start Cody dev from this branch
2. Sign into to Cody with Free user
3. Open a new chat and choose any other LLM
4. Verify clicking on Pro only model will open the subscription page in your browser

https://github.com/sourcegraph/cody/assets/68532117/76c8765b-b9af-498a-83f1-36ae58f2a33d

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -2,22 +2,13 @@
 
 This is a log of all notable changes to Cody for VS Code. [Unreleased] changes are included in the nightly pre-release builds.
 
-## [Unreleased]
+## [1.16.6]
 
 ### Added
 
-- Search: A new `Search Code` command added to the `Commands` sidebar for Cody's Natural Language Search. [pull/3991](https://github.com/sourcegraph/cody/pull/3991)
-- Context Menu: Added commands to send file to chat as @-mention from the explorer context menu. [pull/4000](https://github.com/sourcegraph/cody/pull/4000)
-  - `Add File to Chat`: Add file to the current opened chat, or start a new chat if no panel is opened.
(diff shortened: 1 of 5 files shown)


abeatrix:
published

PR #4123, opened 2024-05-09
Context: Update code template for selection

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1715260595763899

This PR updates the code template we are using for selected code to include line numbers after the file path, aiming to fix the issue where Cody would answer the questions using the wrong section of code if the user mentions line number in their question:
image

## Changes

- Add `getContextFileFromSelection` function in `selection.ts` to retrieve the user's current selection as a context item
- Update `SimpleChatPanelProvider` to call `getContextFileFromSelection` and include the selection context in `userContextFiles`
- Modify `populateCurrentSelectedCodeContextTemplate` in `templates.ts` to accept an optional `range` parameter and use `displayPathWithLines` to include the selected line range in the file path
- Add `displayPathWithLines` function in `displayPath.ts` to format the file path with the selected line range
- Update `PromptString` to include `fromDisplayPathLineRange` method for creating prompt strings with line ranges
- Refactor `DefaultPrompter` and remove the `IPrompter` interface
- Update `renderContextItem` in `utils.ts` to handle different context item sources and use appropriate templates
- Add `getContextItemDisplayPath` function in `utils.ts` to get the display path for a context item

### Manual Test: updated code template

To test the code template change:

- Start dev mode from this branch
- Open a file that has more than 100 lines
- Highlight line 30-70 and ask cody `what does the code in line 50-58 do?`
- Verify Cody responds appropriately based on the context from the provided line range

Example: highlight [SimpleChatPanelProvider.ts?L475-568](https://sourcegraph.com/github.com/sourcegraph/cody@bee/selection-context/-/blob/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts?L475-568) and ask Cody about [L482-485](https://sourcegraph.com/github.com/sourcegraph/cody@bee/selection-context/-/blob/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts?L482-485).  

The code in the given lines is:
```
                if (submitType === 'user-newchat' && !this.chatModel.isEmpty()) {
                    span.addEvent('clearAndRestartSession')
                    await this.clearAndRestartSession()
                }
```

image

#### Before

image

## Updated

Since I haven't heard back from the team, I have reverted my change on adding editor selection as context: https://sourcegraph.slack.com/archives/C05AGQYD528/p1715373199246709

The changes listed below are no longer relevant and can be ignored. Just keep it for record purpose :) 

Updated to include editor selection automatically for chat even if enhanced context is disabled, as this seems to be the expected behavior for new users, but I'm happy to remove it if the team disagrees. Currently, Cody is not able to answer questions about your editor selection if you disable enhanced context:

image

Goals
- Improved context for chat messages by including the user's current selection
- More targeted and relevant responses from Cody based on the selected code
- Better user experience by automatically including the selection context without requiring manual input

### Manual Test 2: add editor selection as context with enhanced context disabled

To test the change on adding editor selection as context:

- Start dev mode from this branch
- Disable enhanced context
- Verify that the selected code range is included in the chat message context
- Verify Cody responds appropriately based on the provided selection context

image

#### Before

image


lib/shared/src/prompt/prompt-string.ts
```diff
@@ -237,6 +238,11 @@ export class PromptString {
         return internal_createPromptString(displayPath(uri), [uri])
     }
 
+    public static fromDisplayPathLineRange(uri: vscode.Uri, range?: RangeData) {
+        const pathToDisplay = range ? displayPathWithLines(uri, range) : displayPath(uri)
+        return internal_createPromptString(pathToDisplay, [uri])
+    }
```
philipp-spiess, line 244: 👍

lib/shared/src/prompt/templates.ts
```diff
@@ -48,29 +49,17 @@ export function populateTerminalOutputContextTemplate(output: string): string {
     return COMMAND_OUTPUT_TEMPLATE + output
 }
 
-const SELECTED_CODE_CONTEXT_TEMPLATE = ps`My selected {languageName} code from file \`{filePath}\`:
-<selected>
```
abeatrix, line 52: replaced this with triple backticks because this doesn't work well with HTML code

vscode/src/commands/context/selection.ts
```diff
@@ -60,3 +61,38 @@ export async function getContextFileFromCursor(newCursorPosition?: Position): Pr
...
+            const editor = getEditor()?.active
+            const document = editor?.document
+            const selection = editor?.selection
+            if (!document || !selection) {
+                throw new Error('No active selection in active editor')
+            }
+
+            if (await contextFiltersProvider.isUriIgnored(document.uri)) {
+                return []
+            }
```
philipp-spiess, line 80: 👌

philipp-spiess (approved):
This change makes sense! We don't want to add a lot of magic especially when no enhanced context is used but I feel like having text selected should intuitively be added. 

Maybe an alternative could be that we show UI in the chat message when text is added to quickly add the range to the input if people feel strongly we shouldn't do that?

agent/src/__tests__/example-ts/.cody/commands.json
```diff
@@ -28,7 +28,7 @@
       }
     },
     "countDirFiles": {
-      "prompt": "How many file context have I shared with you?",
+      "prompt": "How many file context have I shared with you? Reply single number. Skip preamble.",
```
abeatrix, line 31: updated to keep the output minimal so we don't have to keep updating the test for a response with word changes

PR #4126, opened 2024-05-09
keep quickpicks (esp for edit) open if focus is lost

sqs:
If the user has (for example) an Cmd+K edit quickpick open, then switches to their web browser to copy some information to paste into the edit quickpick input, then VS Code will close the quickpick. Many users set `"workbench.quickOpen.closeOnFocusLost": false,` in their VS Code settings to prevent this globally because it's annoying, so only users without that setting will experience this annoyance. All of the Cody quickpicks should remain when the user switches away because they are invoked through explicit user action and may have valuable unsaved input provided by the user.

Reported by @martinamps

## Test plan

Run Cmd+K. Cmd+Tab to another application. Ensure VS Code does not hide the edit quickpick.

### DIFF
--- a/vscode/src/commands/menus/command-builder.ts
+++ b/vscode/src/commands/menus/command-builder.ts
@@ -105,7 +105,7 @@ export class CustomCommandsBuilderMenu {
             title: 'New Custom Cody Command: Prompt',
             prompt: 'Enter the instructions for Cody to follow and answer.',
             placeHolder: 'e.g. Create five different test cases for the selected code',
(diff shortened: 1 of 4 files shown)


martinamps (approved):
TIL: https://code.visualstudio.com/api/references/vscode-api#InputBoxOptions

ty 🙏 10min from report to PR 🚀

umpox (approved):
👍

abeatrix (approved):
Nice, I personally like this option more!

PR #4131, opened 2024-05-10
fix needs-cody-pro model selector bugs

sqs:
- make needs-cody-pro models clickable so the user can upgrade: Fixes an issue with the new model selector where needs-cody-pro models were disabled and therefore not clickable.
- fix bug where model selector would not work on first click: Similar to https://github.com/sourcegraph/cody/pull/4121 (HT @abeatrix).

## Test plan

1. Sign in with a Cody Free (not Pro) user
2. In the chat model selector, select a Cody Pro-only model. The web browser should open to the upgrade page and the model should not be selected in the chat view.
3. Sign in with a Cody Pro user.
4. Select a different model. Ensure that the selected model is used (with `cody.debug.verbose` `true` in user settings).

### DIFF
--- a/vscode/webviews/components/modelSelectField/ModelSelectField.tsx
+++ b/vscode/webviews/components/modelSelectField/ModelSelectField.tsx
@@ -36,7 +36,7 @@ export const ModelSelectField: React.FunctionComponent<{
 
     const onModelSelect = useCallback(
         (model: ModelProvider): void => {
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
I missed this 🙈 ty!

sqs:
@abeatrix nah you didn't miss anything, I said I'd get it on `main`. Thanks!

PR #4132, opened 2024-05-10
Cody Ignore: remove git CLI repo-name resolver

valerybugakov:
- Removes the git CLI repo-name resolver because it performs worse than other existing alternatives (VS Code Git extension and file system tree-walk). There are no cases where it is preferred over alternatives, so we can simplify the codebase by deleting it.

## Test plan

CI


lib/shared/src/sourcegraph-api/graphql/queries.ts
```diff
@@ -103,7 +103,6 @@ query Repositories($first: Int!, $after: String) {
         nodes {
             id
             name
-            url
```
valerybugakov, line 106: Unused response field.

PR #4133, opened 2024-05-10
Update chat model category labels

toolmantim:
Updates the language on the chat model category labels, and fixes the sizing and alignment of the message input controls.

## Test plan

- CI


valerybugakov (approved):
Consistency FTW!

PR #4134, opened 2024-05-10
dev: Add Storybook VS Code theme switcher

toolmantim:
Adds a Storybook theme selector for testing how a UI looks across themes.

## Test plan

- CI


sqs (approved):
Nice. Will be very excited to use this. Can you remove the codicon lines from the theme files to keep them shorter? Or just load the theme CSS from an external URL (like unpkg.com) so then we can have a bunch more and don't need to add them to the repo?

toolmantim:
@sqs stripped the icons, though it's still fairly large. What are you thinking with unpkg? Publish this as it's own open source storybook add-on?

sqs:
@toolmantim It's fine now.

As for unpkg, I mean using URLs like https://sqs.vscode-unpkg.net/sqs/vscode-sqs-theme/0.0.8/extension/sqs-dark.json to fetch theme JSON on-the-fly instead of committing it to the repo. Then we can support a lot more themes and to add one you just need to add its extension ID. The builtin VS Code themes are available at https://main.vscode-cdn.net/stable/dc96b837cf6bb4af9cd736aa3af08cf8279f7685/extensions/theme-defaults/themes/dark_vs.json etc. BTW I discovered this by looking in the devtools network tab at https://vscode.dev/.

toolmantim:
@sqs oh nice, good find! I guess one good thing about having a few in the repo is for airplane/offline hacking… but adding that JSON support in the future would be A+. I'll add a note as a comment.

PR #4145, opened 2024-05-10
Measure upstream health and attach median to autocomplete events

philipp-spiess:
Part-of CODY-1495

This PR adds a new service that continuously pings the upstream Sourcegraph instances `/healthz` endpoint with a `HEAD` request to estimate the RTT to the SG instance.

For now, pings are logged in debug mode and attached to autocomplete events (since we want to use it to investigate autocomplete latencies). Later we might extend this into a Cody Help command (cc @RXminuS)

We also want to upload this data to Prometheus as a follow-up.

## Test plan

- Enable verbose debug mode
- Open the extension spot the new `UpstreamHealth` message. It will update after 10 minutes (or you can overwrite the timeout in `UpstreamHealthProvider`)
- Cause an autocomplete to appear and check the suggested event for a new `medianUpstreamLatency` field.

Screenshot 2024-05-10 at 14 37 40
Screenshot 2024-05-10 at 14 34 11

### DIFF
--- a/vscode/src/services/UpstreamHealthProvider.ts
+++ b/vscode/src/services/UpstreamHealthProvider.ts
@@ -10,7 +10,8 @@ import type * as vscode from 'vscode'
 
 // We choose an interval that gives us a reasonable aggregate without causing
 // too many requests
-export const PING_INTERVAL = 10 * 60 * 1000 // 10 minutes
+const PING_INTERVAL_MS = 10 * 60 * 1000 // 10 minutes
+const TWO_HOURS_MS = 2 * 60 * 60 * 1000
 
 /**
  * A provider that regularly pings the connected Sourcegraph instance to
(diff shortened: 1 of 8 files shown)


vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+    addCustomUserAgent,
+    addTraceparent,
+    logDebug,
+} from '@sourcegraph/cody-shared'
+import { fetch } from '@sourcegraph/cody-shared'
+import type * as vscode from 'vscode'
+
+// We choose an interval that gives us a reasonable aggregate without causing
+// too many requests
+export const PING_INTERVAL = 10 * 60 * 1000 // 10 minutes
```
rafax, line 13:
```suggestion
export const PING_INTERVAL_MS = 10 * 60 * 1000 // 10 minutes
```

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+            }
+            addTraceparent(headers)
+            addCustomUserAgent(headers)
+            const url = new URL('/healthz', this.config.serverEndpoint)
+
+            // We use a HEAD request since we do not want to consume the body
+            // and do not want to have the request garbage-collected.
+            //
+            // https://undici.nodejs.org/#/?id=garbage-collection
+            const response = await fetch(url.toString(), { method: 'HEAD', headers })
```
rafax, line 74: Seeing that this is traversing customer controller internal networks, I'd use a GET (which is less likely to be problematic for proxies / VPNs / ...)

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+
+            const duration = Date.now() - start
+            this.recentDurations.push({ timestamp: Date.now(), duration })
+            // Delete items that are older than 2 hours
+            this.recentDurations = this.recentDurations.filter(
+                item => Date.now() - item.timestamp < 2 * 60 * 60 * 1000
```
rafax, line 80: nit: const for two hours?

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -0,0 +1,122 @@
...
+    private nextTimeoutId: NodeJS.Timeout | null = null
+
+    public getMedianDuration(): number | undefined {
+        if (!this.config) {
+            return undefined
+        }
+        if (this.recentDurations.length === 0) {
+            return undefined
+        }
+        const sorted = this.recentDurations.sort((a, b) => a.duration - b.duration)
```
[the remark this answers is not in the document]
philipp-spiess: Yes makes sense!

RXminuS:
Looks good. Just an FYI, my thinking for the new LogTrace library is that I'll make a general "stats sink" that groups all log messages & traces by type and maintains a count, median/p99/frequency, etc, for all of them that we can report. But I think this merges in nicely with that too

RXminuS (approved):

PR #4148, opened 2024-05-10
Add back socket reuse test

philipp-spiess:
This was initially added in #3769 but we had to revert it. Bringing the test back so we have this behavior documented

## Test plan

- test green

### DIFF
--- a/vscode/src/fetch.node.ts
+++ b/vscode/src/fetch.node.ts
@@ -30,7 +30,6 @@ function getCustomAgent({ proxy }: Configuration): ({ protocol }: Pick<URL, 'pro
         const proxyURL = proxy || getSystemProxyURI(protocol, process.env)
         if (!proxyURL) {
             if (protocol === 'http:') {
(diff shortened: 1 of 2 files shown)


vscode/src/fetch.node.ts
```diff
@@ -30,7 +30,6 @@ function getCustomAgent({ proxy }: Configuration): ({ protocol }: Pick<URL, 'pro
         const proxyURL = proxy || getSystemProxyURI(protocol, process.env)
         if (!proxyURL) {
             if (protocol === 'http:') {
-                console.log('No proxy set for HTTP requests')
```
philipp-spiess, line 33: @arafatkatze Was this a leftover? If we want to surface this we should put it into a `logDebug` instead maybe 🤔

arafatkatze: Yo Phillip this is definitely a testing leftover so its perfectly okay to get rid of this. 

Thanks for the catch

arafatkatze (approved):

PR #4153, opened 2024-05-11
extract ContextItem#isTooLarge calculation and add deprecation note

sqs:
There is no change to behavior; this is just a refactor to extract code, add some notes, and make other minor code style changes.

## Test plan

CI


valerybugakov (approved):
LGTM

PR #4154, opened 2024-05-11
space key no longer triggers commit of selected mention item

sqs:
This behavior existed to preserve parity with the pre-Lexical-editor behavior, but I think we should remove it because it will interfere with more complex @-mentions where the "query" does need to include spaces. (For example, think about @-mentioning a title of a doc with spaces in it.)

## Test plan

CI

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -12,6 +12,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


thenamankumar (approved):
I see the `SpaceKeyTrigger` being removed but are we already capturing the `tab`? Just wanted to confirm. 

Otherwise LGTM.

sqs:
Confirmed <kbd>Tab</kbd> still works.

PR #4157, opened 2024-05-13
Remove `when` from `commands` list

philipp-spiess:
Remove `when` from the `commands` list. It has no meaning there (and also no hover-description if you try that).

When you want to condituially show/hide a command, use the `commandPlate` list instead (as I have done here with https://github.com/sourcegraph/cody/pull/4048).

## Test plan

Add/remove the `cody.autocomplete.experimental.multiModelCompletions` option and see the command appear/disappear:

https://github.com/sourcegraph/cody/assets/458591/d2ebd84a-1e8b-4242-b2e6-1b30922b713c

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -285,62 +285,53 @@
         "command": "cody.command.edit-code",
         "category": "Cody Command",
         "title": "Edit Code",
-        "when": "cody.activated && editorTextFocus",
         "icon": "$(wand)"
       },
       {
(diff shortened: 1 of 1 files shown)


philipp-spiess:
We decided to go this route instead: https://github.com/sourcegraph/cody/pull/4155#pullrequestreview-2052031693 

In the future this will allow us to avoid having to add an entry to `commandsPallete` and keeping the condition in sync with the keyboard elements.

PR #4163, opened 2024-05-13
LLM Models: add GPT-4o support

olafurpg:
Fixes CODY-1536

Previously, we only supported GPT-4 Turbo and GPT-3.5. This PR adds additional support for GPT-4o, which seems to be both more powerful than GPT-4 while offering faster latency.

## Test plan
 I wasn't able to test this change locally so I'm going to leave this as a draft. More details in https://github.com/sourcegraph/sourcegraph/pull/62639

With a local build, it looks like this but when I send a message with gpt-4o it works but it's too slow to be GPT-4o (which responds super quickly in ChatGPT). Given that I'm sending the request to dotcom, I'm not sure what model we're using instead. I see gpt-4o in the logs
```
█ logEvent (telemetry disabled): CodyVSCodeExtension:chat-question:executed  {
  "details": "VSCode",
  "properties": {
    "requestID": "854ac15d-51dc-41de-b1f5-742066fd981d",
    "chatModel": "openai/gpt-4o",
```

CleanShot 2024-05-13 at 22 20 32@2x

### DIFF
--- a/lib/shared/src/models/dotcom.ts
+++ b/lib/shared/src/models/dotcom.ts
@@ -57,6 +57,17 @@ const DEFAULT_DOT_COM_MODELS: ModelProvider[] = [
         uiGroup: ModelUIGroup.Speed,
     },
     {
+        title: 'GPT-4o',
+        model: 'openai/gpt-4o',
+        provider: 'OpenAI',
+        default: false,
+        codyProOnly: true,
+        usage: [ModelUsage.Chat, ModelUsage.Edit],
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
not sure if it belongs to uiGroup: ModelUIGroup.Accuracy, or ModelUIGroup.Balance but the rest lgtm, ty Olaf!

mnismt:
Just curious: Do you rank and group models based on a specific metric, or is it just kind of like, following some rules of thumb?

abeatrix:
We can close this due to https://github.com/sourcegraph/cody/pull/4164

PR #4168, opened 2024-05-14
Cody Ignore: Default to allowing files for dotcom until the policy is fetched

dominiccooney:
sourcegraph.com can set Cody Ignore policy, however it is unlikely to. There's a window from startup until the ignore policy is fetched where we conservatively default to ignoring files. For users on slow connections, that could be a long time.

This changes the rules so that for dotcom accounts, we will liberally assume all files are allowed until the policy is fetched. This ensures dotcom users should not see Cody Ignore notices.

## Test plan

`pnpm test:unit`

Manual test:

1. Run the JetBrains plugin, sign in to dotcom, open a file. Close that IDE.
2. Run in VSCode with agent:debug configuration.
3. Run JetBrains in IntelliJ with `CODY_AGENT_DEBUG_PORT=3113;CODY_AGENT_DEBUG_REMOTE=true;CODY_JETBRAINS_FEATURES=cody.feature.inline-edits=true`
4. As quickly as possible, cmd-shift-H to run "Document" command.

You should not see the notification that your file is ignored because of admin policy.

### DIFF
--- a/lib/shared/src/cody-ignore/context-filters-provider.test.ts
+++ b/lib/shared/src/cody-ignore/context-filters-provider.test.ts
@@ -39,6 +39,7 @@ describe('ContextFiltersProvider', () => {
         vi.spyOn(graphqlClient, 'fetchSourcegraphAPI').mockResolvedValue(
             apiResponseForFilters(contextFilters)
         )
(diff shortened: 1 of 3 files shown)


pkukielka (approved):
LGTM

PR #4172, opened 2024-05-14
Edit: Process streamed insertions in queue and fix range expansion

umpox:
## Description

closes https://github.com/sourcegraph/jetbrains/issues/1449

requires https://github.com/sourcegraph/jetbrains/pull/1509

This PR improves how we stream insertions into the document:
- Updates how we handle the `insertionPromise` queue, to make it more consistent and easier to understand
- Fixes a problem where we would update `task.selectionRange` after an edit was applied, but incorrectly calculate the `characterDelta`. 
- Adds agent tests for generating and inserting a streamed edit from a cursor position

## Test plan

Tested with agent tests

Manual testing:
1. In JetBrains, create an edit with no selection
2. Check that the incoming edit is streamed into the document, without any duplicate characters or lines that break the inserted code

### Before

https://github.com/sourcegraph/cody/assets/9516420/55d30e17-6109-48c8-ab49-560e1e206309

### After

https://github.com/sourcegraph/cody/assets/9516420/fe630452-98c3-46c7-9d41-5dcdf2ce735c


vscode/src/edit/provider.ts
```diff
@@ -215,24 +213,42 @@ export class EditProvider {
...
-            }
+        //     if (isMessageInProgress) {
+        //         // Response hasn't finished, disable until we have the full response
+        //         return
+        //     }
 
-            return this.handleFixupInsert(response, isMessageInProgress)
-        }
+        //     return this.handleFixupInsert(response, isMessageInProgress)
+        // }
```
umpox, line 228: I want to remove this fully when confident in this change (then we support streaming insertions in JetBrains too)

agent/src/index.test.ts
```diff
@@ -1008,6 +1008,35 @@ describe('Agent', () => {
...
+                  interface HeadingProps {
+                      text: string;
+                      level?: number;
+                  }
+
+                  export const Heading: React.FC<HeadingProps> = ({ text, level = 1 }) => {
+                      const HeadingTag = \`h\${level}\` as keyof JSX.IntrinsicElements;
+                      return <HeadingTag>{text}</HeadingTag>;
+                  };
+                  };
```
umpox, line 1034: This duplicate `};` is suspect 🤔 

Trying to figure out why this would happen... Can't replicate in VS Code

vscode/src/edit/provider.ts
umpox: Removed!

agent/src/index.test.ts
umpox: Figured it out... we were calculating the `characterDelta` wrong, so often the `selectionRange` was greater than the document. Created odd behaviour where edits were not applied consistently

vscode/src/non-stop/FixupController.ts
```diff
@@ -644,23 +644,27 @@ export class FixupController
...
-            editOk = await vscode.workspace.applyEdit(edit)
+            editOk = (await vscode.workspace.applyEdit(edit)) !== false
         } else {
-            editOk = await edit(editBuilder => {
-                editBuilder.replace(task.selectionRange, replacement)
-            }, applyEditOptions)
+            editOk =
+                (await edit(editBuilder => {
+                    editBuilder.replace(task.selectionRange, replacement)
+                }, applyEditOptions)) !== false
```
umpox, line 652: @odisseus I noticed that we need to also avoid returning `null` for `workspace/edit` on the JetBrains side too, I'll spin up a quick PR

umpox: PR: https://github.com/sourcegraph/jetbrains/pull/1509

vscode/src/non-stop/FixupController.ts
```diff
@@ -652,15 +652,18 @@ export class FixupController
...
+            const lastLineLength = insertedLines.at(-1)?.length || 0
+
             // Expand the selection range to accompany the edit
-            task.selectionRange = task.selectionRange.with(
+            task.selectionRange = new vscode.Range(
                 task.selectionRange.start,
-                task.selectionRange.end.translate({
-                    lineDelta:
-                        task.selectionRange.start.line - task.selectionRange.end.line + insertedLines,
-                    characterDelta: insertedLines < 1 ? replacement.length : 0,
```
umpox, line 662: Ok:

Originally this code didn't run at all in JetBrains, we were not correctly responding with a boolean and instead setting `editOk` to null. This meant that, although we were applying edits, we were never updating the `selectionRange` of the document, so we just spat out duplicated garbage constantly.

This is fixed by https://github.com/sourcegraph/jetbrains/pull/1501 and https://github.com/sourcegraph/jetbrains/pull/1509

So JetBrains was fixed, but we now had a new error in the agent. The `characterDelta` here was totally wrong, because it accumulated for each line. It meant that our `selectionRange` was getting out of the bounds of the document. VS Code was very forgiving about this, and simply adjusted the range for us, whereas other clients would (probably rightfully so) complain that we were targeting something invalid.

We now correctly adjust the `end` of the range using the character length of the final line

pkukielka (approved):
LGTM

PR #4182, opened 2024-05-15
Autocomplete: hacky gpt-4o support

valerybugakov:
Hacky GPT-4o autocomplete support.

## Test plan

1. Add `"cody.autocomplete.advanced.provider": "unstable-openai"` to your user settings.
2. Start Cody Gateway locally from this branch https://github.com/sourcegraph/sourcegraph/pull/62693.
3. Login to the local SG instance.
4. Use autocomplete.

### DIFF
--- a/vscode/src/completions/providers/unstable-openai.ts
+++ b/vscode/src/completions/providers/unstable-openai.ts
@@ -63,7 +63,7 @@ class UnstableOpenAIProvider extends Provider {
     }
 
     public emptyPromptLength(): number {
-        const promptNoSnippets = [this.instructions, this.createPromptPrefix()].join('\n\n')
+        const promptNoSnippets = [this.createPromptPrefix()].join('\n\n')
         return promptNoSnippets.length - 10 // extra 10 chars of buffer cuz who knows
     }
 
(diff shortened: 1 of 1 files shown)


vscode/src/completions/providers/unstable-openai.ts
```diff
@@ -140,10 +139,82 @@ ${OPENING_CODE_TAG}${infillBlock}`
...
+${PromptString.join(
+    examplesArray.map(example => {
+        return ps`
+<example>
+H: <existing_code>
+    ${example.example}
+    </existing_code>
+
+A: ${example.response}
+</example>
```
philipp-spiess, line 182: Intuitively I'd say we could model examples in different user/assistant responses to avoid some of the overhead here, i.e like I do here: https://github.com/sourcegraph/cody/blob/main/vscode/src/supercompletions/prompt.ts#L53-L112

valerybugakov: Nice! I didn't spend any time tweaking [the prompt](https://sourcegraph.com/gitlab.com/gitlab-org/gitlab@a6aeaca46de9b6b139a352e0a543c7aded04770b/-/blob/ee/lib/code_suggestions/prompts/code_completion/anthropic.rb?L22). We should definitely change it if we want to ship this.

PR #4188, opened 2024-05-16
show types of context in @-mention menu

sqs:
Now, the `@`-mention menu shows the types of context. For most users (who have no experimental features on), this is just `Files` and `Symbols`.

image

image

## Test plan

e2e tests are updated and exercise the functionality.


toolmantim (commented):
Yesssssss. This is awesome. I popped one design fix onto #4192

The keyboard can be a little bit of a battle:

https://github.com/sourcegraph/cody/assets/153/a31a4f89-abee-48e7-ad99-b2466e46926c

https://github.com/sourcegraph/cody/assets/153/79433971-9f7a-4030-adf2-d2bc59364eae

I'd prefer if we didn't edit the input when you select a group (do away with prefixes), and we had a back button, but we've spoken about that already.

Perhaps we could not include the types (e.g. "Files") in the text matching too, so you have to hit up and down to select them, rather than getting into situations where you end up with "fi" left in the input after selecting "Files"?

Keeping the "Suggested" items in there once you select a category is feeling a bit odd too:

https://github.com/sourcegraph/cody/assets/153/3d7ef40f-a007-44d9-8212-7215652a3f6e

I think maybe we should only keep the suggested items on the top-level menu? wdyt?

Or we bring in the "Suggested" separator heading label:

image

Would be nice have the labels for GitHub and NPM give instructions, ala the web URL one, instead of the none-found message:

CleanShot 2024-05-16 at 22 18 46@2x

CleanShot 2024-05-16 at 22 18 55@2x

Looks like maybe some weirdness with @-mentioning large files too:

https://github.com/sourcegraph/cody/assets/153/99bec8f6-915b-4538-a1fb-1ebb1237966a

sqs:
@toolmantim: Thanks!

> I'd prefer if we didn't edit the input when you select a group (do away with prefixes), and we had a back button, but we've spoken about that already.

- [ ] You know, I can get rid of the auto-inserted `#` prefix easily. Let me do that.

> Perhaps we could not include the types (e.g. "Files") in the text matching too, so you have to hit up and down to select them, rather than getting into situations where you end up with "fi" left in the input after selecting "Files"?

- [ ] Yeah, good call. Or, if you hit tab/enter to select a type, then it deletes the text you've entered so far. Some other apps work this way and it feels right. Let me try both.

> Keeping the "Suggested" items in there once you select a category is feeling a bit odd too ... Or we bring in the "Suggested" separator heading label:

- [ ] This will be up to the provider ultimately. I think we'll want to let providers return stuff for an empty query. I'll put `Suggested` as a header when they don't come from the query, as you suggest.

> Would be nice have the labels for GitHub and NPM give instructions, ala the web URL one, instead of the none-found message:

- [ ] Ah yep, will do.

> Looks like maybe some weirdness with @-mentioning large files too:

- [ ] Will fix

toolmantim:
> Yeah, good call. Or, if you hit tab/enter to select a type, then it deletes the text you've entered so far. Some other apps work this way and it feels right. Let me try both.

Yeah Raycast does this too, and if you hit esc then it reinserts the text. 

> I tried putting Suggested as a header when they don't come from the query, but this creates more jitter when you type the first query character because the Suggested line disappears then. So, I don't think that is actually good.

Ah yeah! Might be fine for now.

toolmantim:
Here's an idea: https://www.figma.com/design/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?node-id=2797-50268&t=F9Zqw3xd6jCNNzX5-4

sqs:
> Here's an idea: [figma.com/design/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?node-id=2797-50268&t=F9Zqw3xd6jCNNzX5-4](https://www.figma.com/design/jEsnzgsf0hNuJbqB9pKaGm/VS-Code---Cody---Commands-up-front-and-ChatGPT%2B%2B?node-id=2797-50268&t=F9Zqw3xd6jCNNzX5-4)

What is the difference? The selection is anchored at the bottom instead of the top?

BTW, I think @-mentions will be a lot smoother when we have the message input not always at the bottom (like in the noodle designs where each message cell remains editable).

sqs:
Future improvements:

1. Make sure there is no jitter by handling all the cases where we want to preserve the displayed data while we fetch new data
2. Size the menu so that it fits items exactly and does not show part of an item (which looks weird for a menu)

toolmantim (approved):
This is looking great! 👏🏼🎉

👍🏼 from a design/interaction POV.

There's a few small things, unsure if you'd consider them blocking:

1. I'm seeing NPM Packages, GitHub, URL etc appear even with `{ "cody.experimental.urlContext": false, "cody.experimental.packageContext": false, "cody.experimental.noodle": false }` — could be just my machine?
2. If you type "@", choose Symbols, and wait for the loading to complete… it eventually shows "No symbols found". Is that because it tried to load some suggestions? Or are we accidentally showing the "No symbols found" even when you haven't typed a search?
3. I'm seeing "No packages found", "No results" etc when choosing some types without typing anything. Could we only show those "No xyz found" if you've typed a search after the @?

sqs:
@toolmantim Will check why those are enabled. I am seeing that too.

Symbols is a tricky one. That could be because your symbol provider returns empty for an empty query. Which language? And does it show results when you type a few more letters?

For the empty states for the others, that's the experimental API so we probably need to have the providers tell us what they want to do. They are returning empty arrays. Will defer that work.

toolmantim:
> Symbols is a tricky one. That could be because your symbol provider returns empty for an empty query. Which language? And does it show results when you type a few more letters?

This is on sourcegraph/sourcegraph with http-status-error.ts open:

https://github.com/sourcegraph/cody/assets/153/bce70824-0f6d-4de0-a704-c6bd55d53e19

See the "No symbols found" message showing even when I only have "@" in the input?

> For the empty states for the others, that's the experimental API so we probably need to have the providers tell us what they want to do. They are returning empty arrays. Will defer that work.

Cool. For later consideration: perhaps empty arrays for an empty search could just not show anything.

sqs:
Fixed the symbols issue!

toolmantim:
In last Figma I attached above I removed that top header as soon as you type a search query, but left it on the for empty cases (which I think is still needed… like for URL and Code Search etc). Maybe we could try that?

mkondratek (approved):
I've quickly run & tested jetbrains plugin with these changes (clearly there is no effect in UI for us but everything seems to work properly as before 👍 )

PR #4189, opened 2024-05-16
support changing chat model after chat session has begun

sqs:
Previously, you could only change the chat model before you started a chat. Now, you can change it at any point.

- Changing a chat model does not re-send previous messages.
- If you change the chat model and then edit a message, it gets sent to the new model (not the original model of the message).

image

## Test plan

1. Send a first chat message using any model.
2. Change the model using the model selector.
3. Send a 2nd chat message. Confirm it shows as being from the model you just chose.
4. Change the model to yet another model. Edit the 2nd chat message and send it. Ensure that the response now shows as being from the new model.
5. Close the chat and reopen it. Ensure that the models are still correct (by their icons).

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatModel.ts
+++ b/vscode/src/chat/chat-view/SimpleChatModel.ts
@@ -67,6 +67,7 @@ export class SimpleChatModel {
             error = this.messages.pop()?.error
         }
         this.messages.push({
+            model: this.modelID,
             ...message,
             speaker: 'assistant',
             error,
@@ -80,6 +81,7 @@ export class SimpleChatModel {
             lastMessage?.speaker === 'assistant' ? this.messages.pop() : undefined
         // Then add a new assistant message with error added
         this.messages.push({
+            model: this.modelID,
             ...(lastAssistantMessage ?? {}),
             speaker: 'assistant',
             error: errorToChatError(error),
@@ -162,10 +164,12 @@ export class SimpleChatModel {
         }
         const result: SerializedChatTranscript = {
             id: this.sessionID,
-            chatModel: this.modelID,
             chatTitle: this.getCustomChatTitle(),
             lastInteractionTimestamp: this.sessionID,
             interactions,
+
+            // DEPRECATED: Write this for backcompat.
+            chatModel: this.modelID,
         }
         if (this.selectedRepos) {
             result.enhancedContext = {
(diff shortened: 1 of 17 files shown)


vscode/src/chat/chat-view/SimpleChatModel.ts
```diff
@@ -162,10 +164,12 @@ export class SimpleChatModel {
         }
         const result: SerializedChatTranscript = {
             id: this.sessionID,
-            chatModel: this.modelID,
             chatTitle: this.getCustomChatTitle(),
             lastInteractionTimestamp: this.sessionID,
             interactions,
+
+            // DEPRECATED: Write this for backcompat.
```
dominiccooney, line 171: I like this approach. Potential to break JetBrains without it is high.

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -203,3 +219,66 @@ export const localStorage = new LocalStorage()
 function getKeyForAuthStatus(authStatus: AuthStatus): ChatHistoryKey {
     return `${authStatus.endpoint}-${authStatus.username}`
 }
+
+/**
+ * Migrate chat history to set the {@link ChatMessage.model} property on each assistant message
+ * instead of {@link SerializedChatTranscript.chatModel} on the overall transcript. Can remove when
+ * {@link SerializedChatTranscript.chatModel} property is removed in v1.22.
```
dominiccooney, line 226: Heads up @pkukielka , @mkondratek ... JetBrains will probably need some work here before the removal happens.

We should use this as an opportunity to unfork some of JetBrains' copy of chat state if possible and rely on the TypeScript extension state.

dominiccooney (approved):
Nice. Looks like test expectations need to be updated.

vscode/src/services/LocalStorageProvider.ts
pkukielka: I might be wrong, but this is purely VSCode code, it is not used by agent.
Both extensions keeps it's state, and agent does not store any state on the disk.

So that:

> We should use this as an opportunity to unfork some of JetBrains' copy of chat state if possible and rely on the TypeScript extension state.

would in fact require pretty big refactoring and approach change.
If I'm wrong I'm happy to be corrected :) 

That said we should still embrace this change allowing to always switch model quickly, it will be very nice to have.

PR #4192, opened 2024-05-16
Switch to Lucide

toolmantim:
Had a pow-wow with the designers this week, and we [decided](https://sourcegraph.slack.com/archives/C01U8SJUKJM/p1715773538257929?thread_ts=1714611056.256939&cid=C01U8SJUKJM) on Lucide as our go-to icon set moving forward. Looks good in both VS Code, Jetbrains and web. This updates your PR @sqs to use them instead.

## Test plan

- CI


toolmantim:
I should get a proper Code Search icon in there instead of the SG logo in color, but left it out of this PR

sqs:
very good, thanks! adopted in my PR

PR #4193, opened 2024-05-16
Simplify upstream latency collector and measure gateway latency

philipp-spiess:
Closes CODY-1605

This PR changes the way we track upstream latency (= RTT to the Sourcegraph instance). Instead of measuring every 10 minutes and then exporting the median to the completion metadata, I thought that it might be simpler to include the last measurement with the completion data (and then we can run an aggregation on that). While coding, completion suggestions should happen more frequent then every 10 minutes so this will give us a higher resolution while we avoid to maintain a list.

Additionally, I've added a second parameter to measure the latency to Gateway. This is only turned on for Free/Pro users right now (to avoid outgoing traffic in air-gapped instances). 

Furthermore I added tracing to the upstream pings. With our current rate limit, one two two traces every ~10 minutes _should be fine_ (again I assume this happen far less frequently than completion suggestions) but we can revisit this later.

The benefit of exporting this as a trace is that we can later forward the trace to the SG instance and convert it to a metric.

## Test plan

Screenshot 2024-05-16 at 14 41 54

### DIFF
--- a/vscode/src/services/UpstreamHealthProvider.ts
+++ b/vscode/src/services/UpstreamHealthProvider.ts
@@ -69,6 +69,10 @@ export class UpstreamHealthProvider implements vscode.Disposable {
         }
 
         try {
+            if (process.env.DISABLE_UPSTREAM_HEALTH_PINGS === 'true') {
+                return
+            }
+
             if (!this.config) {
                 throw new Error('UpstreamHealthProvider not initialized')
             }
--- a/vscode/src/services/UpstreamHealthProvider.ts
+++ b/vscode/src/services/UpstreamHealthProvider.ts
@@ -3,7 +3,10 @@ import {
     type ConfigurationWithAccessToken,
     addCustomUserAgent,
     addTraceparent,
+    dotcomTokenToGatewayToken,
+    isDotCom,
     logDebug,
+    wrapInActiveSpan,
 } from '@sourcegraph/cody-shared'
 import { fetch } from '@sourcegraph/cody-shared'
 import type * as vscode from 'vscode'
@@ -11,7 +14,6 @@ import type * as vscode from 'vscode'
 // We choose an interval that gives us a reasonable aggregate without causing
 // too many requests
 const PING_INTERVAL_MS = 10 * 60 * 1000 // 10 minutes
-const TWO_HOURS_MS = 2 * 60 * 60 * 1000
 
 /**
  * A provider that regularly pings the connected Sourcegraph instance to
@@ -20,8 +22,9 @@ const TWO_HOURS_MS = 2 * 60 * 60 * 1000
  * You can query it to get aggregates of the most recent pings.
  */
 export class UpstreamHealthProvider implements vscode.Disposable {
-    // Array sorted by duration for easy median calculation
-    private recentDurationsSorted: { timestamp: number; duration: number }[] = []
+    private lastUpstreamLatency?: number
+    private lastGatewayLatency?: number
+    private fastPathAccessToken?: string
 
     private config: Pick<
         ConfigurationWithAccessToken,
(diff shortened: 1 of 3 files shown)


vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -67,35 +82,56 @@ export class UpstreamHealthProvider implements vscode.Disposable {
...
+            })
+
+            // We don't want to congest the network so we run the test in serial
+            if (this.fastPathAccessToken) {
+                const headers = new Headers()
+                headers.set('Content-Type', 'application/json; charset=utf-8')
+                headers.set('Authorization', `Bearer ${this.fastPathAccessToken}`)
+                addTraceparent(headers)
+                addCustomUserAgent(headers)
+                const uri = 'https://cody-gateway.sourcegraph.com/healthz'
```
rafax, line 97: The right URL seems to be `https://cody-gateway.sourcegraph.com/-/healthz` (your URL 404s for me)?

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -67,35 +82,56 @@ export class UpstreamHealthProvider implements vscode.Disposable {
...
+            const upstreamResult = await wrapInActiveSpan('upstream-latency.upstream', span => {
+                span.setAttribute('sampled', true)
+                return measureLatencyToUri(headers, url.toString())
+            })
+
+            // We don't want to congest the network so we run the test in serial
+            if (this.fastPathAccessToken) {
+                const headers = new Headers()
+                headers.set('Content-Type', 'application/json; charset=utf-8')
+                headers.set('Authorization', `Bearer ${this.fastPathAccessToken}`)
```
rafax, line 94: It seems quite surprising that we require auth for healthcheck endpoint, I'll check this

vscode/src/services/UpstreamHealthProvider.ts
philipp-spiess: Hm this one also does not seem right. I wasn't able to find an endpoint that works so decided that a 404 endpoint from the same endpoint should still have the same round trip time.

vscode/src/services/UpstreamHealthProvider.ts
philipp-spiess: Right, I added it for the sg instances since I wanted to account for all enterprise installations (I’m also attaching custom header overwrites in the hope that it'll allow me to route to the right instance). Shouldn't be needed by the publicly accessible gateway though.

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -67,35 +82,56 @@ export class UpstreamHealthProvider implements vscode.Disposable {
             addTraceparent(headers)
             addCustomUserAgent(headers)
             const url = new URL('/healthz', this.config.serverEndpoint)
+            const upstreamResult = await wrapInActiveSpan('upstream-latency.upstream', span => {
+                span.setAttribute('sampled', true)
+                return measureLatencyToUri(headers, url.toString())
+            })
+
+            // We don't want to congest the network so we run the test in serial
```
rafax, line 90:
```suggestion
            // We don't want to congest the network so we run the test serially
```

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -67,35 +82,56 @@ export class UpstreamHealthProvider implements vscode.Disposable {
...
+            if (this.fastPathAccessToken) {
+                const headers = new Headers()
+                headers.set('Content-Type', 'application/json; charset=utf-8')
```
rafax, line 93: Header manipulation is very similar to lines 77-81 - can you reuse headers, and swap the access-token?

rafax (approved):

philipp-spiess:
Had two thoughts yesterday while I couldn't sleep at 3am (🙃):
- We should run the initial ping a bit delayed. During the extension startup, a lot of i/o and network is happening so it might not be the best sample if we start a ping right away. I think something like a 10 second delay for the initial ping should be enough, just to ensure there's not too much interference of other extensions booting up?
- Let's change the refresh time to 30min just in case. This will create only 1/3 of the traces and will avoid us burning through our tracing quotas.

vscode/src/services/UpstreamHealthProvider.ts
philipp-spiess: Changed the endpoint to `/-/__version` which doesn't require auth: https://github.com/sourcegraph/sourcegraph/blob/5022b91a4343f4daf2674e9a5df03ecdeb8e4afb/cmd/cody-gateway/internal/httpapi/diagnostics.go#L48-L53

PR #4200, opened 2024-05-17
use simpler way of loading OpenCtx (for mentions/context)

sqs:
Previously, Cody would try to communicate with the OpenCtx VS Code extension in the extension host process. This required both the Cody and OpenCtx VS Code extensions to both be installed and activated, which was tricky and made it hard to develop on both simultaneously.

Now, Cody uses [@openctx/vscode-lib](https://www.npmjs.com/package/@openctx/vscode-lib), which means it does not need to rely on the OpenCtx VS Code extension being installed and activated.

To avoid affecting users who are not participating in this experiment, you must set `cody.experimental.openctx` to true in your VS Code user settings to use OpenCtx in Cody.

To try it out end-to-end, add the following to your VS Code user settings:

```
{
  "openctx.enable": true,
  "openctx.debug": true,
  "openctx.providers": {
    "https://openctx.org/npm/@openctx/provider-hello-world": true,
  },
  "cody.debug.verbose": true,
  "cody.experimental.openctx": true
  "cody.experimental.noodle": true,
}
```

And then in a new chat window, type `@`, select `OpenCtx` from the mention menu, and look for the `Hello, world!` mention item. This will currently only work in a VS Code debug extension host.

## Test plan

CI

### DIFF
--- a/lib/shared/src/mentions/providers/openctxMentions.ts
+++ b/lib/shared/src/mentions/providers/openctxMentions.ts
@@ -1,18 +1,18 @@
 import { URI } from 'vscode-uri'
 import type { ContextItemWithContent } from '../../codebase-context/messages'
 import { isDefined } from '../../common'
-import { getOpenCtxExtensionAPI } from '../../context/openctx/api'
+import { getOpenCtxClient } from '../../context/openctx/api'
 import type { ContextItemFromProvider, ContextMentionProvider } from '../api'
 
 export const OPENCTX_CONTEXT_MENTION_PROVIDER: ContextMentionProvider<'openctx'> = {
     id: 'openctx',
     title: 'OpenCtx',
     async queryContextItems(query) {
-        const openctxAPI = await getOpenCtxExtensionAPI()
-        if (!openctxAPI) {
+        const client = await getOpenCtxClient()
+        if (!client) {
             return []
         }
-        const results = await openctxAPI.getItems({ query: query })
+        const results = await client.items({ query: query })
         const items =
             results
                 ?.map((result, i) =>
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1376,6 +1376,7 @@
     "@lexical/html": "^0.13.1",
     "@lexical/react": "^0.13.1",
     "@mdi/js": "^7.2.96",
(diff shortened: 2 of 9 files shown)


keegancsmith:
> This will currently only work in a VS Code debug extension host.

Ouch! Did we also regress on loading npm packages? Previously that worked fine for me and it was only the file:// that needed the debug extension host.

vscode/package.json
```diff
@@ -1376,6 +1376,7 @@
     "@lexical/html": "^0.13.1",
     "@lexical/react": "^0.13.1",
     "@mdi/js": "^7.2.96",
+    "@openctx/vscode-lib": "^0.0.1",
```
keegancsmith, line 1379: @thenamankumar when you land your PR I assume you need to update vscode-lib on npm before you can land this? Additionally updating the openctx client package.

lib/shared/src/mentions/providers/openctxMentions.ts
```diff
@@ -1,18 +1,18 @@
...
     title: 'OpenCtx',
     async queryContextItems(query) {
-        const openctxAPI = await getOpenCtxExtensionAPI()
-        if (!openctxAPI) {
+        const client = await getOpenCtxClient()
+        if (!client) {
             return []
         }
-        const results = await openctxAPI.getItems({ query: query })
+        const results = await client.items({ query: query })
```
keegancsmith, line 15: this needs to become client.mentions

keegancsmith (approved):
LGTM. While reviewing I realised we can either land this PR as is and then create another one which updates to the new APIs, or just update this. If the former ignore my inline comments :)

vscode/package.json
thenamankumar: Yes, the api changes will be part of the next PR and I'll release new version of vscode-lib before I merge the next Cody PR.

lib/shared/src/mentions/providers/openctxMentions.ts
thenamankumar: The API along with the vscode-lib version will be updated in the upcoming PR. The current version only supports the old API.

PR #4201, opened 2024-05-17
Present at-mention context providers from OpenCtx

thenamankumar:
This PR integrates OpenCtx provides into Cody. Cody will now be able to fetch @-mention context from any of the configured OpenCtx providers. All the configured providers will be automatically listed in the @-mention menu UI. 

If no OpenCtx providers are configured by default, this PR enables the "Web Url" provider powered by OpenCtx: https://openctx.org/docs/providers/web.

For now, this PR stops showing the other experimental providers, including Packages, GitHub, and more. These providers will be soon migrated to OpenCtx only.

This integration will be available to all VS Code users as part of the pre-release and does not require the `experimental.noodle` flag to be `true`.

This PR also bumps up the versions of `@openctx` packaged to use the latest ones. 

## Test plan

Demo video: https://www.loom.com/share/a095faf7b57449ada50108ff7091480d?sid=96b718b6-ab8a-433c-baf0-7597e277bd1c


sqs:
I think the best way to do this is something like what's done for `queryContextItems` in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx?L29. You can add new methods for getting the providers' `meta`s and `mentions` (I don't think the webview needs the full `items`). And `queryContextItems` is the same as `mentions` so you can just consolidate those 2 APIs and use the OpenCtx API design/terms there.

Let me know if you want any help here.

thenamankumar:
@sqs 

> I think the best way to do this is something like what's done for `queryContextItems` in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx?L29. You can add new methods for getting the providers' `meta's and `mentions` (I don't think the review needs the full `items`). And `queryContextItems` is the same as `mentions`, so you can just consolidate those 2 APIs and use the OpenCtx API design/terms.
> 
> Let me know if you want any help here.

@sqs I didn't clearly understand what you meant here. Right now, I have implemented the `getAllMentionProvidersMetadata` method to fetch the list of providers and render them in the MentionsMenu.

The `queryContextItems` internally calls `mentions` but only if the selected provider is openctx; otherwise, it will call the `queryContextItems` for the provider, which implements the Cody-specific Provider interface.

If your idea is to refactor the whole thing, then I would like to do it as a next step into a separate PR.

vscode/src/chat/context/chatContext.ts
```diff
@@ -60,49 +57,26 @@ export async function getChatContextItemsForMention(
...
-            provider =>
-                (isURLProviderEnabled && provider.id === URL_CONTEXT_MENTION_PROVIDER.id) ||
-                (isPackageProviderEnabled && provider.id === PACKAGE_CONTEXT_MENTION_PROVIDER.id)
-        )
+            return items.map(
+                (item): ContextItemOpenCtx => ({
+                    ...item,
+                    type: 'openctx',
+                    uri: URI.parse(item.uri),
+                    providerUri: item.providerUri,
```
keegancsmith: can remove this, already in `...item`

lib/shared/src/context/openctx/api.ts
```diff
@@ -1,23 +1,21 @@
...
+export const openCtx = new OpenCtx(undefined)
 
 /**
  * Set the handle to the OpenCtx client.
  */
-export function setOpenCtxClient(client: OpenCtxClient | undefined): void {
-    if (_client) {
+export function setOpenCtxClient(client: OpenCtxClient): void {
+    if (openCtx.client) {
         throw new Error('OpenCtx extension API is already set')
```
keegancsmith, line 17: Lets make sure we don't call this more than once. Including calling exposeOpenCtxClient more than once.

vscode/webviews/mentions/providers.tsx
```diff
@@ -0,0 +1,67 @@
...
+        const rejectTimeout = setTimeout(() => {
+            resolve([])
+            dispose()
+        }, MAX_WAIT_SECONDS * 1000)
+
+        // Wait for the response. We assume the first message of the right type is the response to
+        // our call.
+        const dispose = vscodeApi.onMessage(async message => {
+            if (message.type === RESPONSE_MESSAGE_TYPE) {
+                resolve(await (message.providers ?? []))
```
keegancsmith, line 39: so technically if this fails now we show no @-mention providers, including the builtin file and symbols providers. However, these cases should never happen since the openctx provider fetching is wrapped in a try catch.

So the only case really is if for some reason listing openctx providers took more than 15s, which seems extremely unlikely.

keegancsmith (approved):
LGTM

vscode/src/chat/context/chatContext.ts
thenamankumar: done

Issue #3648, opened 2024-04-01
Expand the output token limits


chillatom:
Currently we limit the output tokens to 1,000.  In practice this isn't enough to accomplish much in the way of long form code generation.  1,000 tokens roughly equates to 133 LOC in typescript and functionally we get less than that due to token estimation, response pre-amble from the LLM that eats some tokens.  So if I am trying to translate a 200-400 LOC file, I hit the output limit.  

Most provider APIs cap the output at 4,096 tokens.  I'd like us to do two things

1. Explore increasing our output token limit to be much closer to the 4,096 token max, say 4,000 tokens.  We should validate that this won't have unintended consequences.  @philipp-spiess 
2. Create a button (or other UI element) that allows users to easily continue generation if we detect that we hit the token limit.  The button should show up if we hit the limit and stop streaming.  It should use a limited prompt structure that tells the LLM to continue where it left off and only uses the prior chat history and previously fetched context.  No additional context should be fetched. @toolmantim for designs.

chillatom:
Tagging @chwarwick for enterprise considerations

aramaraju:
> Previously this was done using the MaxChatTokens however that appears to have been repurposed by clients as a upper limit on included context.

@chwarwick Do we need to make a change to add another way to do token limits with these client changes?

chillatom:
> Create a button (or other UI element) that allows users to easily continue generation if we detect that we hit the token limit

There was a comment on another ticket that had a nice suggestion.  If we can detect that we hit the output token limit, and I believe we may get that signal from the provider completion API, we could just automatically ask to continue.

chillatom:
https://github.com/sourcegraph/cody/issues/3786 related

umpox:
Closed through https://github.com/sourcegraph/cody/pull/3797

@chillatom Let's use https://github.com/sourcegraph/cody/issues/3583 to track the addition of the "Continue" button. I have a feeling that the need for this will be much less now we have a higher output limit. We can still add it if we feel a need, but let's reassess once this change it out to users.

PR #4203, opened 2024-05-17
Increase the output window tokens for known-good enterprise models

philipp-spiess:
This is a workaround until we have proper support for https://github.com/sourcegraph/sourcegraph/pull/61885

We have recently increased out output token lengths for PLG and now Enterprise is lagging behind. Certain Enterprise setups are getting in our way to apply the same defaults (see [this thread for context](https://github.com/sourcegraph/cody/issues/3648#issuecomment-2056954101)). 

The workaround is to only allow a hand full of recommended models to be using the extended token windows. Some use the Cody Gateway format (with the provider prefix) and only point to Cody Gateway configs and some use the Bedrock format (which contain the prefix with a `.` and a special version identifier at the end) and should thus only affect Bedrock users. Furthermore I did intentionally only include recent flagship models and not things like Claude 2, to avoid issues with older configurations.

## Test plan

- Connect to the demo instance and see the longer result window in the output tab
- Got a response of ~11k chars which is 2.2k GPT4 tokens so probably over 1k Anthropic tokens

Screenshot 2024-05-18 at 00 15 31


sqs:
Test plan: worked for me on s2.

- On latest build, signed into s2, output channel shows the request has `"maxTokensToSample":` 1000,`
- On this branch, signed into s2 (after I changed s2's model to be Sonnet), output channel shows `"maxTokensToSample":` 1000,`
  - and I got a response in chat with 3839 chars, which is probably over 1000 tokens

sqs (approved):

PR #4206, opened 2024-05-18
sort directory context files for determinism

sqs:
This also fixes the Windows unit tests in CI. According to https://github.com/sourcegraph/cody/actions/workflows/ci.yml?query=branch%3Amain, all `main` CI runs including and after https://github.com/sourcegraph/cody/commit/ce072f784a049e44653151961122807638a53224 have been failing on the Windows unit test step.

Also toSorted is not available in Node 18, so we need to use sort still in one other place.

## Test plan

CI

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -789,106 +789,58 @@ describe('Agent', () => {
                 const lastMessage = await client.firstNonEmptyTranscript(id)
                 expect(trimEndOfLine(lastMessage.messages.at(-1)?.text ?? '')).toMatchInlineSnapshot(
                     `
-                  "Based on the shared code context from \`src/example.test.ts\`, the test framework being used is Vitest, which is a modern test runner powered by Vite. It appears to be a test suite for a specific "test block" with multiple test cases.
+                  "To generate unit tests for the \`Animal\` interface, I will need to import the testing framework and assertion library used in the shared code context. Based on the file \`src/example.test.ts\`, the testing framework being used is Vitest.
 
-                  To generate a suite of unit tests for the \`Animal\` interface defined in \`src/animal.ts\`, we can use Vitest and the assertions provided by the \`vitest\` library.
+                  No new imports needed - using existing libs.
 
-                  Importing the required libraries:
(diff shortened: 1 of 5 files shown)


sqs:
(not urgent) @philipp-spiess do you think we should sort `getWorkspaceFilesContext`'s result for determinism in the tests?

agent/src/index.test.ts
```diff
@@ -789,106 +789,58 @@ describe('Agent', () => {
                 const lastMessage = await client.firstNonEmptyTranscript(id)
                 expect(trimEndOfLine(lastMessage.messages.at(-1)?.text ?? '')).toMatchInlineSnapshot(
                     `
-                  "Based on the shared code context from \`src/example.test.ts\`, the test framework being used is Vitest, which is a modern test runner powered by Vite. It appears to be a test suite for a specific "test block" with multiple test cases.
```
olafurpg, line 792: Reminder to myself to refactor these tests so they become more stable and we stop having to update the snapshots every week. I feel bad for having introduced this abuse of inline snapshot assertions.

vscode/src/commands/context/workspace.ts
```diff
@@ -34,7 +34,9 @@ export async function getWorkspaceFilesContext(
         }, 20000)
 
         try {
-            const results = await workspace.findFiles(globalPattern, excluded, maxResults, token.token)
+            const results = (
+                await workspace.findFiles(globalPattern, excluded, maxResults, token.token)
+            ).sort((a, b) => a.toString().localeCompare(b.toString()))
```
olafurpg, line 39: If the sorting is only needed for testing purposes then you can optionally guard the sort statement behind an if condition. We already do this for symf context.

olafurpg (approved):
Nice!

vscode/src/commands/context/workspace.ts
sqs: It seems like a good idea to use it for all purposes, not just testing. This ensures deterministic behavior for the same user and across platforms. If it turns out the ordering of findFiles is meaningful, then we should make it consistent across platforms. So, I think this is fine for now.

olafurpg: Sounds good 👍🏻

PR #4209, opened 2024-05-18
new transcript/message "stream" UI

sqs:
This new design for Cody's chat transcript ([Figma](https://www.figma.com/design/xZ7JeITOwyq2eqV4H0ePVN/Cody-Tomorrow)), which @toolmantim made about a month ago, feels a LOT better IMO. In this design:

- Human messages are always editable. There is no separate input vs. display. This makes it feel like a notebook.
- The model selector, enhanced context, @-mention menu, and submit buttons are associated with the human messages, not with the overall transcript or the single input box at the bottom.
- The chat starts at the top and streams down, instead of having the user type in the bottom then look up to see the reply and their own message.
  - Notably, this makes @-mentions feel better because the menu drops down from your cursor at the top, and the top item is selected. The distance between your text insertion cursor and the 1st @-mention item is much smaller in this design.
- This design lets us remove complexity around scrolling and editing states that was needed when we could not guarantee that the human message would be in view. Now, we know it is in view because the user is editing it.
- This design makes it more discoverable to show follow-up actions at the bottom of the message, because the user is likely to *at least sometimes* scroll all the way down (to send a follow-up message), at which point they will discover those follow-up actions.
  - This makes it tolerable to have longer messages overall.
- This design saves vertical space because we don't need to always show the input. (It is OK for the input field for follow-up messages to be scrolled down. Worst case, if the user can't find that, they will know they can resend by typing in the top message box.)
  - This means we can put toolbar items below the input and don't need to fit them side-by-side with it (like the current enhanced context button), which makes them more discoverable and lets us put more of them.
- This design makes it possible to show more suggestions as the user types their message, because there is effectively infinite space below the input.

https://github.com/sourcegraph/cody/assets/1976/53cf30d2-de31-4c3c-87a1-ffe6b139c32a

## Test plan

e2e tests cover the functionality


toolmantim:
Amazing! I'll check this out first thing.

toolmantim:
This is so great! Left some high-level questions in Slack. Here's the smaller QA'ish things I spotted:

1. Editing and resubmitting quickly shows a transcript context error, stacks up the error messages ([video](https://github.com/sourcegraph/cody/assets/153/b426aa10-60d8-45b9-bd5c-754cf0a462f4))
1. (Opt+L) in the "Ask followup…" doesn't work. Maybe let's just let people scroll down and click for now?
1. @ menu should only appear if input is focused ([video](https://github.com/sourcegraph/cody/assets/153/45a3fd14-0337-404a-86e2-c0ec6cc4be39))
1. @ menu doesn't seem clickable after opening Enhanced Context ([video](https://github.com/sourcegraph/cody/assets/153/06006f30-38e9-4cdb-9389-d85378dec3ff)
1. Sometimes the follow-up @ menu doesn't have the current open files. ([video](https://github.com/sourcegraph/cody/assets/153/393d443c-ddf9-4d9f-a253-410b3dc8a781))
1. Choosing a model doesn't refocus input
1. No "Stop" mode for the Run button. Future addition?
1. Double focus ring when Enhanced Context and LLM model popover is shown. Main input should probably show as de-focused.
1. Keyboard accessibility: can't tab from input to @ etc toolbar buttons
1. Keyboard accessibility: can't tab to Enable Embeddings inside Enhanced Context popover
1. Icons and button sizes, always shown at 16x16 or 24x24 (e.g. @ icon and return icon), use Lucide’s CornerDownLeft, etc - I can do this as follow-up
1. Restyle tooltip to match VS Code - I can do this as follow-up
1. Restyle toolbar button(s) - I can do this as follow-up
1. Too many font size differences now across our different popovers and editor UI - I can do this as follow-up

toolmantim (approved):
This is so great! Left some high-level questions in [Linear](https://linear.app/sourcegraph/issue/CODY-1663/new-transcriptmessage-stream-ui). Here's the smaller QA'ish things I spotted:

1. Editing and resubmitting quickly shows a transcript context error, stacks up the error messages ([video](https://github.com/sourcegraph/cody/assets/153/b426aa10-60d8-45b9-bd5c-754cf0a462f4))
1. (Opt+L) in the "Ask followup…" doesn't work. Maybe let's just let people scroll down and click for now?
1. @ menu should only appear if input is focused ([video](https://github.com/sourcegraph/cody/assets/153/45a3fd14-0337-404a-86e2-c0ec6cc4be39))
1. @ menu doesn't seem clickable after opening Enhanced Context ([video](https://github.com/sourcegraph/cody/assets/153/06006f30-38e9-4cdb-9389-d85378dec3ff)
1. Sometimes the follow-up @ menu doesn't have the current open files. ([video](https://github.com/sourcegraph/cody/assets/153/393d443c-ddf9-4d9f-a253-410b3dc8a781))
1. Choosing a model doesn't refocus input
1. Double focus ring when Enhanced Context and LLM model popover is shown. Main input should probably show as de-focused.
1. Keyboard accessibility: can't tab from input to @ etc toolbar buttons
1. Keyboard accessibility: can't tab to Enable Embeddings inside Enhanced Context popover

Approving, as I'm unsure whether to consider them blocking before merge or just blocking before release (unsure how we'd track the latter)

sqs:
> * Editing and resubmitting quickly shows a transcript context error, stacks up the error messages ([video](https://github.com/sourcegraph/cody/assets/153/b426aa10-60d8-45b9-bd5c-754cf0a462f4))

Hmm, this is probably because it's now super easy to trigger a ton of edits, and there is a race condition. Will fix after merge.

> * (Opt+L) in the "Ask followup…" doesn't work. Maybe let's just let people scroll down and click for now?

Opt+/ works, and that's what it says in the UI. Opt+L in the current chat starts a new chat, and it still does that in this design. We can change the behavior later. BTW it's fastest to just <Tab> down to the next chat input.

> * @ menu should only appear if input is focused ([video](https://github.com/sourcegraph/cody/assets/153/45a3fd14-0337-404a-86e2-c0ec6cc4be39))

Will fix

> * @ menu doesn't seem clickable after opening Enhanced Context ([video](https://github.com/sourcegraph/cody/assets/153/06006f30-38e9-4cdb-9389-d85378dec3ff)

Yeah...will fix post-merge.

> * Sometimes the follow-up @ menu doesn't have the current open files. ([video](https://github.com/sourcegraph/cody/assets/153/393d443c-ddf9-4d9f-a253-410b3dc8a781))

Looking into this

> * Choosing a model doesn't refocus input

Fixed

> * Double focus ring when Enhanced Context and LLM model popover is shown. Main input should probably show as de-focused.

Will fix post-merge.

> * Keyboard accessibility: can't tab from input to @ etc toolbar buttons

This is intentional b/c `@` is its own keyboard shortcut (just type `@`), the submit button is enter, and for the other 2: I think we should assign hotkeys to them but NOT make them tabbable. That way, tab takes you to the next input, which is a lot nicer imo. But let's get a feel for this.

> * Keyboard accessibility: can't tab to Enable Embeddings inside Enhanced Context popover

Will fix post-merge.

toolmantim:
> Opt+/ works, and that's what it says in the UI. Opt+L in the current chat starts a new chat, and it still does that in this design. We can change the behavior later. BTW it's fastest to just down to the next chat input.

Sorry, my typo! I meant Opt+/ as it says I the follow-up input placeholder. I was expecting it to focus the follow-up input, but it seems to start a new chat:

https://github.com/sourcegraph/cody/assets/153/a4bccd38-0131-42d6-9072-c7d7f5011cef

sqs:
> Sorry, my typo! I meant Opt+/ as it says I the follow-up input placeholder. I was expecting it to focus the follow-up input, but it seems to start a new chat:

Ah, yeah. VS Code itself is capturing this before the webview can do anything. Will switch it to <kbd>Opt+&gt;</kbd> for now.

toolmantim:
Focus is working better!

It still doesn't like clicking toolbar buttons when popovers are showing:

https://github.com/sourcegraph/cody/assets/153/18b39e70-4f9e-471f-b8e7-0f1b27bbff44

And the follow-up keyboard shortcut I can't get to work:

https://github.com/sourcegraph/cody/assets/153/d1f31fc7-e2c4-4c31-b139-14b14a7c1f03

sqs:
@toolmantim:

Finally fixed the focus issues. Also added an e2e test for these. Wanna give it another try?

> And the follow-up keyboard shortcut I can't get to work ([video](https://github.com/sourcegraph/cody/assets/153/d1f31fc7-e2c4-4c31-b139-14b14a7c1f03))

Ah, I see. Is it being caught by a VS Code or OS-level hotkey for you? Can you check in VS Code's keyboard shortcuts view? Use `opt+shift+.` and paste in like this:

image

toolmantim:
@sqs might be a macOS thing, as opt+shift+. is how you type the ˘ character. Nothing is bound in VS Code:

CleanShot 2024-05-21 at 20 40 39@2x

CleanShot 2024-05-21 at 20 41 56@2x

toolmantim:
> Finally fixed the focus issues. Also added an e2e test for these. Wanna give it another try?

@sqs focus is working great now! 🙌🏼

sqs:
Feedback from using it: https://community.sourcegraph.com/t/new-chat-ui-feedback/350

PR #4225, opened 2024-05-20
Fix selection range for chat commands

pkukielka:
## Changes

Fixes incorrect selection choice for commands.

Before:

image

After:

image

## Test plan

Run Explain and Smell commands in a long file with a short selection and without any selection at all.
Answers should be relevant to the selected lines, cursor position, or a visible fragment of the code.

### DIFF
--- a/vscode/src/commands/execute/explain.ts
+++ b/vscode/src/commands/execute/explain.ts
@@ -39,16 +39,11 @@ async function explainCommand(
         prompt = ps`${prompt} ${args.additionalInstruction}`
     }
 
-    // fetches the context file from the current cursor position using getContextFileFromCursor().
-    const contextFiles: ContextItem[] = []
+    const cs =
+        (await getContextFileFromCursor(args?.range?.start))[0] ||
+        (await getContextFileFromCurrentFile())[0]
+    const contextFiles: ContextItem[] = [cs]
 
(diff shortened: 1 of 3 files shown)


vscode/src/commands/execute/explain.ts
```diff
@@ -39,16 +39,11 @@ async function explainCommand(
...
+    const cs =
+        (await getContextFileFromCursor(args?.range?.start))[0] ||
+        (await getContextFileFromCurrentFile())[0]
+    const contextFiles: ContextItem[] = [cs]
 
-    const currentSelection = await getContextFileFromCursor(args?.range?.start)
-    contextFiles.push(...currentSelection)
-
-    const currentFile = await getContextFileFromCurrentFile()
-    contextFiles.push(...currentFile)
```
pkukielka, line 49: That line 49 and adding second contextFile was specifically an issue.

mkondratek (approved):

abeatrix (approved):

abeatrix:
i think the issue is that the selected code is merged into the whole file causing the issue. this should be a quick workaround for now but i will create another issue to make sure we don't remove the selected code when there is one

abeatrix (changes requested):
I think we can close this PR as https://github.com/sourcegraph/cody/pull/4229 should address the root cause 😄

PR #4228, opened 2024-05-20
openctx: use the exact mention.uri for items

keegancsmith:
VSCode's URI.toString will escape the = and & from the query part of a URL. You can optionally pass in skipEncode to toString, but in reality we should be passing back the exact URL the provider supplied and should not be potentially changing it at all.

This commit fixes the sourcegraph-search provider. Note that the links displayed in the @-mention chip still will not work due to the above escaping. I am unsure of a clean fix for that yet.

Test Plan: manually tested sourcegraph-search provider


lib/shared/src/codebase-context/messages.ts
```diff
@@ -174,6 +174,10 @@ export interface ContextItemOpenCtx extends ContextItemCommon {
     provider: 'openctx'
     title: string
     uri: URI
+    /**
+     * originalURI is stored since vscode-uri will incorrectly escape query parameters in toString.
+     */
+    originalURI: string
```
thenamankumar, line 180: lets make `originalURI` optional.

vscode/src/editor/utils/editor-context.ts
```diff
@@ -344,6 +344,7 @@ async function resolveContextMentionProviderContextItem(
                       type: 'openctx',
                       title: item.title,
                       uri: URI.parse(item.url || item.providerUri),
+                      originalURI: item.url || item.providerUri,
```
thenamankumar, line 347: lets make `originalURI` optional. and then just do `originalURI: item.url`

thenamankumar (approved):
Approved to unblock.

lib/shared/src/codebase-context/messages.ts
keegancsmith: @thenamankumar motivation for this not being optional is it mirror the `uri` field which is also not optional. What is your motication for changing this?

thenamankumar: The motivation is that it is optional for mention item in openctx. And `item.url || item.providerUri` does not make sense. 

But this is not a blocker.

keegancsmith: I see the motivation. I will say why I think this is preferable below, but I am going to see if I can think up a better solution post-merge. I am going to land this now because it unblocks the sourcegraph-search provider.

So the reason why I prefer required is also due to avoiding "assuming" it is set in this line. See what happens when I make the field optional:

image


> And `item.url || item.providerUri` does not make sense.

we use the same logic the line above, IE mirroring the raw source.

keegancsmith: @thenamankumar here is something which I think is better https://github.com/sourcegraph/cody/pull/4247

PR #4235, opened 2024-05-20
add `languageId` to completion events `privateMetadata` V2 telemetry

akalia25:
Adding `languageId` to the event payload for v2 telemetry. Since we are using the `splitSafeMetadata`, the `languageId` gets added to the `privateMetadadata`

## Test plan
Tested locally

image


philipp-spiess (changes requested):
I don't understand this change. `SharedEventPayload` already defines `languageId`. And this is only a type chance so I would not expect this to change the runtime mechanics? 🤔

PR #4236, opened 2024-05-21
Enterprise: add support for smart context in enterprise LLM configura…

abeatrix:
FOLLOW-UP: https://github.com/sourcegraph/sourcegraph/pull/62802
CLOSE: https://github.com/sourcegraph/cody-issues/issues/28

Enable clients to respect the new site config `smartContext`:
- [x] Once the site config value exists, then Cody clients should respect the feature flag. If the Sourcegraph server version is too old, then we should pretend the feature flag is disabled and retain the old client behavior.
- [x] If the site config feature flag is enabled AND the model is Claude 3 or GPT-4o (i.e. any of the models we enable smart context mixing for PLG users today), then Site config settings like chatModelMaxTokens should no longer be respected, instead the 45k token limit we chose for PLG/Pro users should be used instead so enterprise users get the same experience as PLG/Pro users.

Changes:
- Fetch smart context configuration from the Sourcegraph GraphQL API
- Update the ModelContextWindow to support extended input and output token budgets for models with smart context
- Ensure backward compatibility by falling back to default smart context behavior if the query fails
- Refactor models utils: extracts the logic for synchronizing model providers based on the authentication status into a separate module (`models/sync.ts`). It also moves the `getChatModelsFromConfiguration` function to the new module.

## Test plan

2. Connect to DotCom to make sure everything works as before
3. Connect to any SG instances that does not have the new `smartContext` config available and verify the context window is not expanded for any model.
4. Connect to an instance that has the new `smartContext` config available and verify if the Context Window has been expanded when using claude-3-opus / claude-3-sonnet / gpt-4o models.


arafatkatze (approved):
Works as shown in the instructions above. 

Adds a token limit of 15K chat input extended budget and 30 K context making a total of 45K for input.

PR #4241, opened 2024-05-21
Agent: add support for Fix command

olafurpg:
Fixes CODY-1751

Previously, there was no way to trigger the "Ask Cody to Fix" command through the agent protocol. This is one of the most used Cody commands per our usage metrics. This PR adds support for triggering this command through a new Code Action based API. A lot of work from inline edit can be reused because it's built on the same generic code editing infrastructure.

This PR is first step to CODY-1576. This PR has afew semi-unrelated changes that are needed for CODY-1576 but I left them here to save time on unnecessary refactorings.

## Test plan

See new integration test.


agent/recordings/fixClient_1835524951/recording.har.yaml
```diff
@@ -0,0 +1,1575 @@
+log:
```
olafurpg, line 1: FWIW, I am planning to reduce the size of the HTTP recording files for integration tests, see https://sourcegraph.slack.com/archives/C05AGQYD528/p1716284822364479

vscode/src/code-actions/fixup.ts
```diff
@@ -22,6 +22,7 @@ export class FixupCodeAction implements vscode.CodeActionProvider {
                 diagnostic.severity === vscode.DiagnosticSeverity.Error ||
                 diagnostic.severity === vscode.DiagnosticSeverity.Warning
         )
+        logDebug('fixup', 'diagnostics', diagnostics)
```
olafurpg, line 25: Will remove

olafurpg:
I went down a small rabbit hole to debug why some seemingly unrelated tests were failing and it turns these tests have very sensitive prompts so I took the liberty to fix a bit of weirdness:

* One enterprise test case was suffering from the same non-deterministic behavior I hit on related to tree-sitter loading, and it should now be deterministic 🤞🏻 
* There's too much leaking state between test cases in `index.test.ts`. I moved the "Custom Commands" section to a separate `custom-commands.test.ts` file and disabled one test case that was too unstable (see [CODY-1766 : Disable snapshot assertions that are constantly changing](https://linear.app/sourcegraph/issue/CODY-1766/disable-snapshot-assertions-that-are-constantly-changing)). FYI @abeatrix 

I will follow up separately to break index.test.ts into even smaller files and more aggressively disable tests that have unstable output.

olafurpg:
I will do another round of review tomorrow to be 100% sure this PR won't cause stability issues for JetBrains. If you look at the diff in `vscode/` you'll see that it's mostly import shuffling and refactoring of tsc-retriever components.

The only major behavioral change is that `parseAllVisibleDocuments` now returns a `Promise<unknown>`. This is a high-traffic function since we run it every time the visible editors change, but it should be a low-risk change

agent/src/custom-commands.test.ts
```diff
@@ -0,0 +1,173 @@
...
+        const result = (await client.request('commands/custom', {
+            key: '/hello',
+        })) as CustomEditCommandResult
+        expect(result.type).toBe('edit')
+        await client.taskHasReachedAppliedPhase(result.editResult)
+
+        const originalDocument = client.workspace.getDocument(sumUri)!
+        expect(trimEndOfLine(originalDocument.getText())).toMatchInlineSnapshot(
+            `
+          "/* hello */ export function sum(a: number, b: number): number {
```
abeatrix, line 141: I haven't had a chance to review this PR today but will do it first thing in the morning tomorrow. At first glance this looks different than the previous snippet where the /*hello*/ is insert a line above the function instead of on the same line. Is this expected?

abeatrix (commented):
Thanks for the ping! I will review this PR first thing tomorrow morning, and work on updating the prompts for the custom commands tests to be more predictable.

agent/src/custom-commands.test.ts
olafurpg: I changed the prompt of the custom command to require it to use the inline syntax `/* hello */`. Before this, it was toggling between `//` and `/* */`. I don't know if this change makes the test become more stable but I figured it was worth a shot. We need to spend more energy in designing tests with more stable output, we can't keep on updating inline snapshots on every other PR

olafurpg: Pushed a new commit to use `// hello` instead

olafurpg:
Closing this PR into smaller focused diffs. Going to experiment with spr (https://getcord.github.io/spr/user/stack.html) because I'm finding it too difficult with GH's built-in tooling to make large changes while keeping PR reviews safe/simple/fast.

Issue #4232, opened 2024-05-20
bug: Login page is not displayed after reloading with proxy change 


abeatrix:
### Version

1.19.1716218119

### Describe the bug

bug: Login page is not displayed after reloading with proxy change that causes Cody no longer be able to communicate with the endpoint.

Originally reported by @arafatkatze in https://sourcegraph.slack.com/archives/C05AGQYD528/p1715779597673829

> I've encountered a recurring problem with Cody in vscode where it gets stuck on the loading screen after a VS Code restart. This seems to happen when Cody retrieves an enterprise token from the cache, particularly if there's a network issue or the instance host is facing connectivity problems. The logs indicate that the token is "retreived" from the cache, but Cody remains in an indefinite loading state without progressing to the sign-in page.
This issue seems to trigger a sort of dead initialization, where even after resolving proxy settings or other network-related configurations, the problem persists occasionally, hinting at a potential race condition or a deadlock in the initialization process. I suspect others might have faced similar issues. I'm open to sharing a video demonstration of this problem for those interested. If anyone has insights or potential solutions, especially regarding implementing a timeout for such scenarios, your input would be greatly appreciated.
How can we effectively clear the cache or bypass this stale token state to prevent Cody from being stuck?

### Expected behavior

Login screen should still be loaded when there is an auth issue.

### Additional context

Loom: https://www.loom.com/share/4865ebf0edd34bdcaf0084aeb31ea4f7

arafatkatze:
Solved by #4243

PR #4243, opened 2024-05-21
Fix: Graphql Fetching blockage when Cody tries to load a cached proxy that can only be loaded behind a proxy

arafatkatze:
Solves https://github.com/sourcegraph/cody/issues/4232
Thanks to Kalan and Bee we finally solved this issue which is the final issue that I found in the proxy settings.

## Before 
https://www.loom.com/share/4865ebf0edd34bdcaf0084aeb31ea4f7 

## After
When you log in again without the proxy and you have a cached result then Cody will log out coz the Graphql requests have a timeout on them. Previously the timeout was so long that the cody would be stuck in loading for many many minutes. 

## Test plan
Tested locally

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -889,7 +889,7 @@ export class SourcegraphGraphQLAPIClient {
     public fetchSourcegraphAPI<T>(
         query: string,
         variables: Record<string, any> = {},
-        timeout = 3000 // Default timeout of 3000ms (3 seconds)
+        timeout = 6000 // Default timeout of 6000ms (6 seconds)
     ): Promise<T | Error> {
         const headers = new Headers(this.config.customHeaders as HeadersInit)
         headers.set('Content-Type', 'application/json; charset=utf-8')
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -889,53 +889,53 @@ export class SourcegraphGraphQLAPIClient {
     public fetchSourcegraphAPI<T>(
         query: string,
         variables: Record<string, any> = {},
-        timeout: number = 2000 // Default timeout of 5000ms (5 seconds)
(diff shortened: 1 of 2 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -908,20 +909,34 @@ export class SourcegraphGraphQLAPIClient {
...
             })
-                .then(verifyResponseCode)
+                .then(response => {
+                    clearTimeout(timeoutId) // Clear the timeout if the request completes in time
+                    return verifyResponseCode(response)
+                })
                 .then(response => response.json() as T)
                 .catch(error => {
+                    if (error.name === 'AbortError') {
+                        return new Error(`Request timed out after ${timeout}ms (${url})`)
```
RXminuS, line 934: When the request is aborted like this does it then show the Connection Issues view? https://github.com/sourcegraph/cody/pull/3750

If so we maybe want to add a line to that view as well that you may want to check your Proxy Settings

arafatkatze: > When the request is aborted like this does it then show the Connection Issues view? https://github.com/sourcegraph/cody/pull/3750

No it does not and here's why that would not be the right thing to do. 
Basically this error only happens on "loading from the cache" and only if the user went out of their way to "remove" the proxy then close vscode and then load vscode. So if the user were to remove the proxy then just on starting Vscode the cody will get stuck(without this PR). The normal behaviour should be that cody just "logs out" and shows the log into cody screen that is the inteneded behaviour. 

IF we face proxy network issues on connecting to an instance behind a proxy then that's a separate thing. The intended behaviour here is exactly as expected.

RXminuS: I'm not quite on the same page yet. Scenario: If the user starts their machine in a cafe (where proxies aren't always working) and start Cody they would now be signed out simply starting Cody simply for the proxy connection failing. This is exactly what happened before and users complained being singed out constantly. I don't think we should automatically sign the user out ever, especially because from the "Connection Issues" screen you can manually sign out already

RXminuS (approved):

PR #4246, opened 2024-05-22
feat/vscode: Adds uninstall hook

jamesmcnamara:
Adds a hook to log an event when a user has uninstalled cody. This is made somewhat more complex because the uninstall script is simply a node process which does not have access to the [VSCode extension API](https://github.com/Microsoft/vscode/issues/45474#issuecomment-405502972). So to work around this, I add some hooks to the deactivate method which will serialize all of the configuration needed for constructing a telemetry reporter to a known location, and in the uninstall hook we try to read that file, construct the reporter, log an event and then delete that directory.

Because the `post-install.ts` script cannot have any transitive dependencies on `vscode`, I had to reorganize some imports so that it could pull everything it needed from the shared directory.

Open questions:
1. Where can I check if my event was recorded?
2. Do we have any E2E tests that involve uninstalling? 
3. Do we have a way to test windows installs?

TODO:

- [ ] Testing
- [x] Verification that events are logged
- [ ] Verify it works on windows

## Test plan
I had to package the extension locally:
```sh
pnpm run -s _build:esbuild:hooks && vsce package --no-dependencies --out dist/cody.vsix && code --install-extension dist/cody.vsix --force --new-window
```
Then I was able to uninstall the extension and fully close VSCode. When I reopen it, I see logs from the uninstaller running in the `shared` output tab.


vscode/src/uninstall/post-uninstall.ts
```diff
@@ -0,0 +1,37 @@
...
+                'connected-instance-only'
+            )
+            const recorder = provider.getRecorder()
+            recorder.recordEvent('cody.extension', 'uninstalled')
+
+            // cleanup the uninstaller config
+            deleteUninstallerDirectory()
+        }
+    }
+}
```
jamesmcnamara, line 35: This is the actual uninstall script.

vscode/src/commands/scm/source-control.ts
```diff
@@ -237,8 +237,10 @@ export class CodySourceControl implements vscode.Disposable {
     }
 
     public dispose(): void {
-        for (const disposable of [...this.disposables]) {
-            disposable.dispose()
+        for (const disposable of this.disposables) {
```
dominiccooney, line 240: Did this change in this PR? The old code was better because it is obvious what happens if `this.disposables` is changed by one of the calls to `dispose` (specifically: nothing, because we took a copy.)

There's also an even better way to do it, which is:

```
vscode.Disposable.from(...this.disposables).dispose()
```

which is more succinct, IIRC behaves better if one of the calls to `dispose` throws an exception, etc.

vscode/src/extension.node.ts
```diff
@@ -63,3 +69,18 @@ export function activate(
         extensionClient,
     })
 }
+
+// When cody is deactivated, we serialize the current configuration to disk,
+// so that it can be restored when the post-uninstall script is run, and the
+// vscode API is not available
```
dominiccooney, line 75:
```suggestion
// When Cody is deactivated, we serialize the current configuration to disk,
// so that it can be sent with Telemetry when the post-uninstall script runs.
// The vscode API is not available in the post-uninstall script.
```

dominiccooney (changes requested):
Looks great in as far as it goes. Let's add a test which runs in the node environment which will fail if someone changes TelemetryRecorderProvider or its dependencies to touch vscode API. (Yes, it is in cody-shared right now, but people refactor types between these.)

vscode/package.json
```diff
@@ -11,7 +11,7 @@
     "postinstall": "pnpm download-wasm",
     "dev": "pnpm run -s dev:desktop",
     "dev:insiders": "pnpm run -s dev:desktop:insiders",
-    "start:dev:desktop": "NODE_ENV=development code --extensionDevelopmentPath=$PWD --disable-extension=sourcegraph.cody-ai --disable-extension=github.copilot --inspect-extensions=9333 --new-window . --goto ./src/completions/inline-completion-item-provider.ts:16:5",
+    "start:dev:desktop": "NODE_ENV=development code --extensionDevelopmentPath=$PWD --disable-extension=github.copilot --inspect-extensions=9333 --new-window . --goto ./src/completions/inline-completion-item-provider.ts:16:5",
```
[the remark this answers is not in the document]
jamesmcnamara: This was a mistake. I was enabling cody at the beginning to test the uninstaller but it didn't work. I've reverted it.

vscode/src/commands/scm/source-control.ts
jamesmcnamara: I did change it but you're reasoning makes sense. And your final example is best! That is the exact issue I was dealing with with the `if`.

dominiccooney (approved):
On second thought, maybe the build dependencies on shared are enough to ensure we don't get into a bind here.

PR #4247, opened 2024-05-22
openctx: store mention specific fields separately

keegancsmith:
This is to make it clearer in ContextItemOpenCtx which fields are for before and after resolving a context item.

Additionally we remove the unused description field.

Test Plan: CI


sqs (approved):
We should refactor Cody's core code here to use the same mention vs. context item distinction. It implicitly has that distinction now because make ContextItems have empty `content` and it is resolved late based on the URI and range. That has led to some bugs and is unclear. Also separating mentions would let us show nicer @-mention chip displays for symbol mentions.

This PR is great for now though.

PR #4252, opened 2024-05-22
message editor toolbar improvements to send buttons and auto code context

sqs:
Often you want to be able to chat with no context or only explicitly @-mentioned context. This is helpful if, for example, you are asking about a general code concept or want to make extra sure that Cody only focuses on certain files when answering. This was previously possible with the hidden <kbd>Alt/Opt+Enter</kbd> shortcut, but that was hard to discover.

- Removes the checkbox from the "automatic code context" (fka "enhanced context settings") popover.
- There are now 2 send buttons, one (default) to send with automatic code context and one to send "w/o context".
- Makes all toolbar buttons tabbable for keyboard accessibility.

## Test plan

- The "w/o context" button sends without automatic context.
- The "send" button sends with automatic context.
- Toolbar items are tabbable.


abeatrix (approved):
lgtm ✅ 
image

image

Feedback: It looks a bit cramped on small screens:
image

What do you think about moving that to the placeholder instead? Something like this since I think the input box is descriptive enough for `Ask...`:
image
image

sqs:
> Feedback: It looks a bit cramped on small screens: ...
> What do you think about moving that to the placeholder instead? Something like this since I think the input box is descriptive enough for `Ask...`

Thanks, this is good feedback. I don't think this will last long, just a few days until we find something better. So I'll move forward with this but will find a solution that doesn't have these problems.

PR #4259, opened 2024-05-22
bugfix/ci: Split `test:e2e` to remove bash script

jamesmcnamara:
The E2E tests on windows are [failing](https://github.com/sourcegraph/cody/actions/runs/9192807166/job/25282373856#step:9:10) because npm defaults to running scripts through windows cmd, which chokes on the bash syntax.

My dumb solution is to split it into two commands so you can run `pnpm test:e2e:run` locally when you want the faster loop

## Test plan
CI based E2E test should run with this patch.


sqs (approved):
Thanks!

PR #4261, opened 2024-05-23
Move auto-accept logic from Agent to client for now, workaround for Issue 315 for GA

steveyegge:
This fix must be paired with a [sourcegraph/jetbrains PR](https://github.com/sourcegraph/jetbrains/pull/1603) that moves the logic to the client.

## Test plan

Everyone has try to break it -- specifically:

- the Accept lens group (with Edit/Retry, Undo, Show Diff) should _always_ be shown after requesting a Document Code
- It should auto-accept your current edit if you start another one somewhere else in the document.

This is already covered by the test plan, so it just needs to be re-run with this specific feature in mind.

### DIFF
--- a/vscode/src/non-stop/FixupController.ts
+++ b/vscode/src/non-stop/FixupController.ts
@@ -1024,7 +1024,7 @@ export class FixupController
         // notification back to the Agent after we finish applying the
         // changes. https://github.com/sourcegraph/cody-issues/issues/315 is one example
         // of a bug caused by auto-accepting here, but there were others as well.
-        if (!isRunningInsideAgent) {
+        if (!isRunningInsideAgent()) {
             if (task.state === CodyTaskState.Applied) {
                 // User has changed an applied task, so we assume the user has accepted
                 // the change and wants to take control.  This helps ensure that the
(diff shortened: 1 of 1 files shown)


vscode/src/non-stop/FixupController.ts
```diff
@@ -1015,13 +1016,25 @@ export class FixupController
...
-            return
+
+        // Note that we would like to auto-accept at this point if task.state == Applied.
+        // But it led to race conditions and bad bugs, because the Agent doesn't get
+        // notified when the Accept lens is displayed, so it doesn't actually know when it
+        // is safe to auto-accept. Fixing it properly will require us to send some sort of
+        // notification back to the Agent after we finish applying the
+        // changes. https://github.com/sourcegraph/cody-issues/issues/315 is one example
+        // of a bug caused by auto-accepting here, but there were others as well.
+        if (!isRunningInsideAgent) {
```
abeatrix, line 1027:
```suggestion
        if (!isRunningInsideAgent()) {
```

abeatrix (approved):
i think we will need to call it but other than that, LGTM!

Issue #3963, opened 2024-04-26
bug: "Generate Tests" generates a valid test but missing `import` statement


noah-berman:
### Version

v1.15.1714144471 (pre-release)

### Describe the bug

Customer has reported that, for both sample TypeScript and Python files, "Generate Tests" has correctly populated valid unit tests, but instead of adding an `import` statement at the top of the file it simply lists the path of the new test file:

![image](https://github.com/sourcegraph/cody/assets/38916274/768acaf2-60b2-4d3c-84be-84443dd12d1d)
![image](https://github.com/sourcegraph/cody/assets/38916274/6e9aa849-405c-4881-977e-d5a991db246d)

### Expected behavior

Test file should have an `import` statement to correctly import the function/method in question

### Additional context

Customer is using Cody Pro with freshly spun-up TypeScript and Python repositories for testing.
Client has tried switching LLMs across Sonnet, Opus, GPT-3.5 Turbo, and GPT-4 Turbo

sqs:
Notes:

- Should we run the VS Code "add all missing imports" action after generating a test? Might get it wrong but it's usually good.
- The inclusion of a file path at the top is a bug.

umpox:
- Support running "add missing imports action"
- Separate file path prompt into a separate request. (Right now we try to get the LLM to provide _both_ the file name and the contents in the response, this is error prone)
- The fact it isn't adding any imports at all makes me think this is potentially a prompting issue. It could be that we're not telling the LLM that we're in a new file, so it's assuming we already have access? I think we can fix this reliably

abeatrix:
@umpox we have a [newTestSuitePrompt](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/commands/execute/test-edit.ts?L103) that specifically asked the LLM not to include any imports if there is an existing test file, since the new test suite is appended to the file. Maybe we can remove it to support "add all missing imports"?

PR #4262, opened 2024-05-23
Test Command: Fix issue with inserting test file name with unit test

abeatrix:
RE: https://github.com/sourcegraph/cody/issues/3963

This PR fixes an issue with test command inserting test file name at the top of the unit test

image

> The inclusion of a file path at the top is a bug.

I suspect the issue was introduced in [#3759](https://sourcegraph.com/github.com/sourcegraph/cody/-/commit/db6d4eb84bd32b6141bb93e732d14c40a5775613?visible=3) where we updated to only subscribe to the multiplexer for file creation if we couldn't locate the test file internally. This becomes an issue when we ask for the LLM to always return the test file name in the [PROMPT_TOPICS.FILENAME](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/edit/prompt/models/generic.ts?L88) in our prompt.

- Fixes an issue where the test file name was incorrectly inserted when using the unit test command
- Ensures the test file name suggestion from Cody is properly handled and the file is created if necessary
- Keeps the multiplexer subscription for the test file name suggestion even if a destination file is set, to ensure any text returned by Cody wrapped in the PROMPT_TOPICS.FILENAME tags are properly handled

## Repro Steps

1. In the latest stable Cody build, open the Cody repo
2. Go to any file that already has a `.test.ts` file, [vscode/src/release.ts?L13-16](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/release.ts?L13-16) for example
3. Execute the test command
4. Verify the file path is included with the generated test

https://github.com/sourcegraph/cody/assets/68532117/7ea66551-859d-48ac-a77a-c358856996ac

## Test plan

2. Go to any file that already has a `.test.ts` file, [vscode/src/release.ts?L13-16](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/release.ts?L13-16) for example
3. Execute the test command
4. Verify the file path is no longer included with the generated test
5. Also verify files without existing test files are still working as expected:

https://github.com/sourcegraph/cody/assets/68532117/4a590689-ecf9-4ab9-a2f3-113593864f9a

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -19,7 +19,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 - Chat: Fixes issues with chat commands where selection context is removed from context items. [pull/4229](https://github.com/sourcegraph/cody/pull/4229)
 - Chat: Fixes intermittent issues with `Add Selection to Cody Chat` where sometimes the @-mention would not actually be added. [pull/4237](https://github.com/sourcegraph/cody/pull/4237)
 - Menu: Fixes an issue where the `Add Selection to Cody Chat` context menu item was incorrectly disabled when no new chat was open. [pull/4242](https://github.com/sourcegraph/cody/pull/4242)
-- Fixed an issue where the test file name was incorrectly inserted with the unit test command. [pull/](https://github.com/sourcegraph/cody/pull/)
+- Fixed an issue where the test file name was incorrectly inserted with the unit test command. [pull/4262](https://github.com/sourcegraph/cody/pull/4262)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
Agree this is progress, but the existing code has another problem:

`onResponse` can be called *multiple times.* The code knows this because it is concatenating `filepath`. However `handleFileCreationResponse` probably open/creates an untitled file each time, even if the path is incomplete. That is wild.

Also, why does `handleFileCreationResponse` try to extract the filename response tag *within* the filename topic? Do LLMs stutter the tags and output them twice?

PR #4265, opened 2024-05-23
fix jumpy scroll when submitting followup

sqs:
Instead of ensuring the assistant message is visible, make it so it's impossible for it to NOT be visible ever by adding a margin at the bottom of the last human message that will be filled in visually by the subsequent assistant message.

HT @erikars and @Kynlos for reporting this issue

## Test plan

CI


olafurpg (approved):
LGTM 👍🏻 Ran this locally and confirmed the scroll doesn't jump up anymore like it does in the latest build

PR #4267, opened 2024-05-23
Autocomplete: reproduce buggy prompt where line suffix is missing

olafurpg:
Observe how the prompt does not include the `= b + a` suffix. The LLM generates the completion `finalResult = a + b`, which doesn't match the existing code so we return an empty completion.

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -150,6 +150,122 @@ log:
         send: 0
         ssl: -1
         wait: 0
+    - _id: 53b60a708d430d769ccd5e7330c637e1
+      _order: 0
+      cache: {}
+      request:
+        bodySize: 348
+        cookies: []
+        headers:
+          - _fromType: array
+            name: accept-encoding
+            value: gzip;q=0
+          - _fromType: array
+            name: authorization
+            value: token
+              REDACTED_b09f01644a4261b32aa2ee4aea4f279ba69a57cff389f9b119b5265e913c0ea4
+          - _fromType: array
+            name: connection
+            value: keep-alive
+          - _fromType: array
+            name: content-type
+            value: application/json; charset=utf-8
+          - _fromType: array
+            name: user-agent
+            value: defaultClient / v1
+          - _fromType: array
+            name: accept
+            value: "*/*"
+          - _fromType: array
+            name: content-length
+            value: "348"
+          - name: host
+            value: sourcegraph.com
+        headersSize: 378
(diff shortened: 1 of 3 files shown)


agent/recordings/defaultClient_631904893/recording.har.yaml
```diff
@@ -150,6 +150,122 @@ log:
...
+          mimeType: application/json; charset=utf-8
+          params: []
+          textJSON:
+            maxTokensToSample: 256
+            messages:
+              - speaker: human
+                text: |-
+                  <filename>src/sum.ts<fim_prefix>// 
+                  export function sum(a: number, b: number): number {
+                      const <fim_suffix>
```
olafurpg, line 197: This is the bug, observe how we have removed the suffix `= b + a`

agent/recordings/defaultClient_631904893/recording.har.yaml
```diff
@@ -150,6 +150,122 @@ log:
...
+            value: v1
+        url: https://sourcegraph.com/.api/completions/code?client-name=defaultclient&client-version=v1
+      response:
+        bodySize: 324
+        content:
+          mimeType: text/event-stream
+          size: 324
+          text: |+
+            event: completion
+            data: {"completion":"finalResult = a + b","stopReason":"stop"}
```
olafurpg, line 222: The LLM generates `a + b` here, which doesn't match the actual suffix in the file so we return an empty completion

agent/src/index.test.ts
```diff
@@ -145,12 +148,15 @@ describe('Agent', () => {
...
     }, 10_000)
 
-    describe('Autocomplete', () => {
+    describe.only('Autocomplete', () => {
         it('autocomplete/execute (non-empty result)', async () => {
             await client.openFile(sumUri)
+            const doc = client.workspace.openUri(sumUri)
+            const position = doc.protocolDocument.selection!.start
+            console.log({ position, doc: doc.getText() })
             const completions = await client.request('autocomplete/execute', {
```
olafurpg, line 157: The test fails with the following error

```
 FAIL  |agent| src/index.test.ts > Agent > Autocomplete > autocomplete/execute (non-empty result)
AssertionError: expected 0 to be greater than 0
 ❯ src/index.test.ts:163:46
    161|             })
    162|             const texts = completions.items.map(item => item.insertText)
    163|             expect(completions.items.length).toBeGreaterThan(0)
       |                                              ^
    164|             expect(texts).toMatchInlineSnapshot(
    165|                 `
```
I expected there to be a single completion `finalResult`

PR #4268, opened 2024-05-23
allow copy-to-clipboard of chat response while still streaming

sqs:
Previously, if you tried to select text in Cody's response when it was still in progress (streaming in), it was not possible. Your selection would be immediately deselected.

Now, your selection is preserved and you can copy in-progress responses.

This change was more involved than I expected. I also needed to:

- Switch to using a Markdown renderer that didn't just call `.innerHTML = ...`, so that we have stable DOM nodes that we can anchor our persistent selection to.
- Switch to a different way of sanitizing the HTML.
- Switch to a different way of syntax-highlighting.

Our tests for HTML sanitization still pass, which is a good sign of confidence.

The icons below the code blocks do not yet show up until the message is complete. That will be fixed in a later commit.

https://github.com/sourcegraph/cody/assets/1976/ddd1a35c-6222-4a4d-bcc6-bb82c624cc30

## Test plan

See video.


olafurpg (approved):
I'm very excited to have this functionality! I have not reviewed the diff, only tested the change with a local build. 

- Double-clicking on words/lines works as expected
- Mousedown and drag-to-select is a hit and miss. It seems to work correctly while we stream non-code block, and it doesn't work while we are streaming a code block.

LGTM since this is already a huge improvement over the previous solution. Probably good to get a second eye on the actual changes.

I had tried to quickly fix this myself a while ago but got stuck because it wasn't obvious how to do it. Seeing the diff now makes it clear this wasn't a low hanging fruit. Well done!

olafurpg:
Video recording of the mousedown behavior



https://github.com/sourcegraph/cody/assets/1408093/306f97cf-e6ec-4269-bc8c-ebd9bdce9283

sqs:
(Merged automatically but that's fine.)

@olafurpg:

> * Mousedown and drag-to-select is a hit and miss. It seems to work correctly while we stream non-code block, and it doesn't work while we are streaming a code block.

Hmm, I can't repro this on Linux. Will try on macOS. Selecting text in a code block works while it is still streaming for me.

I'll kick off a prerelease build so more people can try this.

https://github.com/sourcegraph/cody/assets/1976/79268c0c-2abd-46ce-867d-8329cf64e467

toolmantim:
This is so great! And was as hard as I assumed 😅

PR #4269, opened 2024-05-23
Instrument embedding startup and load in VSCode

rafax:
Add tracing integration for Embedding startup and load (to complement the existing integration when querying).

## Test plan

- tested locally
- trace only


vscode/src/local-context/local-embeddings.ts
```diff
@@ -162,17 +162,23 @@ export class LocalEmbeddingsController
...
         logDebug('LocalEmbeddingsController', 'start')
-        await this.getService()
-        const repoUri = vscode.workspace.workspaceFolders?.[0]?.uri
-        if (repoUri && isFileURI(repoUri)) {
-            const loadedOk = await this.eagerlyLoad(repoUri)
-            if (!loadedOk) {
-                // failed to load the index, let's see if we should start indexing
-                if (this.canAutoIndex()) {
-                    this.index()
+        wrapInActiveSpan('embeddings.start', async span => {
```
philipp-spiess, line 165: If this is the root span and you want this to be sampled (and sent to the otel collector) you'd want to add a `span.setAttribute('sampled', true)`

philipp-spiess (approved):

vscode/src/local-context/local-embeddings.ts
rafax: Done

vscode/src/local-context/local-embeddings.ts
```diff
@@ -428,70 +435,91 @@ export class LocalEmbeddingsController
...
+                const noRemote = error.message === noRemoteErrorMessage
 
-            // TODO: Log telemetry error messages to prioritize supporting
-            // repos without remotes, other SCCS, etc.
+                const notAGitRepositoryErrorMessage = /does not appear to be a git repository/
+                const notGit = notAGitRepositoryErrorMessage.test(error.message)
 
-            this.lastRepo = { dir: repoDir, repoName: false }
-        }
+                let errorReason: GetFieldType<LocalEmbeddingsProvider, 'errorReason'>
```
janhartman, line 500: Maybe assign it to undefined by default and then you can get rid of the last else?

janhartman (approved):

PR #4270, opened 2024-05-23
Move inline edit tests to separate file

olafurpg:
This is the first step to address JetBrains performance problems for inline edit. While removing the test from index.test.ts, I skipped a flaky custom command test that keeps failing on unrelated changes (due
    to sensitive prompt).

## Test plan

Green CI. There is no functional difference here. Only refactoring tests.


olafurpg:
Closing since I will open this PR with spr.

PR #4277, opened 2024-05-23
Improve incremental document update

pkukielka:
## Test plan

In a client:

1.  open sourcegraph/cody , execute.ts  file
2.  Select whole ExecuteEditArguments interface block
3.  Click Document Code
4.  Undo edit (from inlay lens)
5.  Click Document Code again -> comments will appear in wrong

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -403,11 +403,14 @@ export class Agent extends MessageHandler implements ExtensionClient {
                 this.workspace.loadDocumentWithChanges(documentWithUri)
             const textEditor = this.workspace.newTextEditor(textDocument)
             this.workspace.setActiveTextEditor(textEditor)
-            vscode_shim.onDidChangeTextDocument.fire({
-                document: textDocument,
-                contentChanges,
-                reason: undefined,
-            })
+
+            if (contentChanges) {
+                vscode_shim.onDidChangeTextDocument.fire({
+                    document: textDocument,
+                    contentChanges,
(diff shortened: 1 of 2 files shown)


olafurpg (approved):
Good catch. I will followup with a minor changes, but this is a great start

agent/src/agent.ts
```diff
@@ -403,11 +403,14 @@ export class Agent extends MessageHandler implements ExtensionClient {
                 this.workspace.loadDocumentWithChanges(documentWithUri)
             const textEditor = this.workspace.newTextEditor(textDocument)
             this.workspace.setActiveTextEditor(textEditor)
-            vscode_shim.onDidChangeTextDocument.fire({
-                document: textDocument,
-                contentChanges,
-                reason: undefined,
-            })
+
+            if (contentChanges) {
```
abeatrix, line 407: This will always be true even if contentChanges is empty right? Cause `Boolean([])` returns `true` ?
Will need to update to:

```suggestion
            if (contentChanges?.length) {
```

olafurpg: ha, I was just opening this PR to make the same comment, will fix in a followup PR

abeatrix: haha It's been fixed @olafurpg! https://github.com/sourcegraph/cody/pull/4278

olafurpg: https://github.com/sourcegraph/cody/pull/4279

PR #4279, opened 2024-05-23
Agent: make AgentWorkspaceDocuments more robust

olafurpg:
The PR https://github.com/sourcegraph/cody/pull/4277 was a lesson that we have very brittle code around dealing with document synchronization. This PR is an attempt to improve code health by

- Adding tests!
- Being more rigorous with null handling
- Making the code a bit easier to follow (hopefully)

## Test plan

See new test suite.

### DIFF
--- a/agent/src/AgentTextEditor.ts
+++ b/agent/src/AgentTextEditor.ts
@@ -1,0 +1,92 @@
+import { logDebug } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+import type { AgentTextDocument } from './AgentTextDocument'
+import type { EditFunction } from './AgentWorkspaceDocuments'
+
+export class AgentTextEditor implements vscode.TextEditor {
+    constructor(
+        private readonly agentDocument: AgentTextDocument,
+        private readonly params?: { edit?: EditFunction }
+    ) {}
+    get document(): vscode.TextDocument {
+        return this.agentDocument
+    }
+    get selection(): vscode.Selection {
+        const protocolSelection = this.agentDocument.protocolDocument.selection
+        const selection: vscode.Selection = protocolSelection
+            ? new vscode.Selection(
+                  new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                  new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+              )
+            : // Default to putting the cursor at the start of the file.
+              new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
+        return selection
+    }
+    get selections(): readonly vscode.Selection[] {
+        return [this.selection]
+    }
+    get visibleRanges(): readonly vscode.Range[] {
+        const protocolVisibleRange = this.agentDocument.protocolDocument.visibleRange
+        const visibleRange = protocolVisibleRange
+            ? new vscode.Selection(
+                  new vscode.Position(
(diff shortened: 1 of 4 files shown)


agent/src/agent.ts
```diff
@@ -404,7 +404,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
             const textEditor = this.workspace.newTextEditor(textDocument)
             this.workspace.setActiveTextEditor(textEditor)
 
-            if (contentChanges) {
+            if (contentChanges?.length) {
```
olafurpg, line 407: No test for this bug, but related fix from https://github.com/sourcegraph/cody/pull/4277/files#r1612093296

olafurpg: Was already fixed

abeatrix (approved):
havent tested it locally but the code lgtm! Thanks for adding the tests too!

olafurpg (commented):
I'm planning to fix [CODY-1829 : Handle null values in the protocol](https://linear.app/sourcegraph/issue/CODY-1829/handle-null-values-in-the-protocol) separately, where we add `| null` to the most important protocol properties.

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -44,33 +44,53 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
     } {
         const fromCache = this.agentDocuments.get(document.underlying.uri)
         if (!fromCache) {
-            return { document: new AgentTextDocument(document), contentChanges: [] }
+            const result = new AgentTextDocument(document)
+            this.agentDocuments.set(document.underlying.uri, result)
```
olafurpg, line 48: The old flow was very weird, we populated the map at call site every single time, instead of doing it only once here.

olafurpg:
The integration tests caught a very critical regression (nice!). While fixing this, I saw that our implementation of `vscode.TextEditor` is not using a singleton instance for selection, which I'm fixing now.

olafurpg:
I will do a second round of review tomorrow since the additional diff ended up being quite large after your approval @abeatrix .

olafurpg:
Just to be 100% sure, I ran VS Code with the local diff

```diff
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -753,6 +753,11 @@ const register = async (
         platform.extensionClient.provide({ enterpriseContextFactory }),
     ])
     disposables.push(extensionClientDispose)
+    for (const editor of vscode.window.visibleTextEditors) {
+        setInterval(() => {
+            console.log({ uri: editor.document.uri.toString(), text: editor.document.getText() })
+        }, 1000)
+    }
```

and confirmed that the text contents reflect the latest version as I make changes to the document even if I reuse the `vscode.TextEditor` instance.

agent/src/AgentTextEditor.ts
```diff
@@ -0,0 +1,92 @@
...
+            cursorStyle: undefined,
+            insertSpaces: undefined,
+            lineNumbers: undefined,
+            // TODO: fix tabSize
+            tabSize: 2,
+        }
+    }
+    viewColumn = vscode.ViewColumn.Active
+
+    // IMPORTANT(olafurpg): `edit` must be defined as a fat arrow. The tests
```
olafurpg, line 52: This issue took me almost an hr to debug. Makes me wonder what other latent bugs we have caused by usage of class methods over property fat arrows.

PR #4280, opened 2024-05-24
Handle JSON RPC Null Values

RXminuS:
TODO: Find all places where we break this rule

### DIFF
--- a/agent/src/AgentWorkspaceDocuments.ts
+++ b/agent/src/AgentWorkspaceDocuments.ts
@@ -59,15 +59,15 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
             }
         }
 
-        if (document.content === undefined) {
+        if ((document.content ?? undefined) === undefined) {
             document.underlying.content = fromCache.getText()
         }
 
-        if (document.selection === undefined) {
+        if ((document.content ?? undefined) === undefined) {
             document.underlying.selection = fromCache.protocolDocument.selection
         }
 
-        if (document.visibleRange === undefined) {
+        if ((document.content ?? undefined) === undefined) {
             document.underlying.visibleRange = fromCache.protocolDocument.visibleRange
         }
 


agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -59,15 +59,15 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
             }
         }
 
-        if (document.content === undefined) {
+        if ((document.content ?? undefined) === undefined) {
```
pkukielka, line 62: I think `if (!!document.content)` should also work.

RXminuS: Nope because then "" also gets overwritten

RXminuS: ?? Only sets to the default value if it's null or undefined

pkukielka: You are right, although it is only problem for strings.
Although it might be better to play safe and just use common pattern everywhere to not miss something...

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -59,15 +59,15 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
...
             document.underlying.content = fromCache.getText()
         }
 
-        if (document.selection === undefined) {
+        if ((document.content ?? undefined) === undefined) {
```
pkukielka, line 66: `document.selection` ^^

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -59,15 +59,15 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
...
             document.underlying.selection = fromCache.protocolDocument.selection
         }
 
-        if (document.visibleRange === undefined) {
+        if ((document.content ?? undefined) === undefined) {
```
pkukielka, line 70: `document.visibleRange` ^^

olafurpg:
I have another PR that fixes the same issue along with a host of other problems (`TextEditor.selection` not reflecting the latest value like it does in VSC), and adds tests https://github.com/sourcegraph/cody/pull/4279

olafurpg:
Superseded by #4279

PR #4283, opened 2024-05-24
Add tests for changes calc

mkondratek:
## Test plan
- CI

### DIFF
--- a/agent/src/calculateContentChanges.test.ts
+++ b/agent/src/calculateContentChanges.test.ts
@@ -81,7 +81,7 @@ describe('calculateContentChanges', () => {
             },
         ])
     })
-    it('hard (actual)', () => {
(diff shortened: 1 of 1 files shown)


agent/src/calculateContentChanges.test.ts
```diff
@@ -32,4 +32,117 @@ describe('calculateContentChanges', () => {
...
+     */
+    fun main() {
+        println("Hello, world!")
+    }
+
+`,
+            },
+        ])
+    })
+    it('hard (actual)', () => {
```
mkondratek, line 84: this is what actually happens 😕

mkondratek:
fixed and covered with tests (see https://github.com/sourcegraph/jetbrains/pull/1635)

PR #4286, opened 2024-05-24
Agent: handle `null` values where we previously only handled `undefined`

sqs:
To prevent bugs when communicating with peers implemented in different programming languages cross-process RPC protocols must not differentiate between `undefined`, ECMAScript property presence (`hasOwnProperty`), and `null`. Anywhere that a value can be `undefined`, our code needs to handle `null` values or no such property existing, and all must be handled in the same way (and vice versa).

If a property's TypeScript type includes any of `null`, `undefined`, or `?` (optional field), then it must use *all* of them in its TypeScript type. All of those possible TypeScript types must be handled.

This makes it impossible to accidentally only check `if (message.foo === undefined) { ...` in cases where `message` came from a peer that may have sent `message.foo === null`.

Also adds tests that parse all of the protocol `.ts` files to report any regressions.

Closes https://linear.app/sourcegraph/issue/CODY-1829/handle-null-values-in-the-protocol.

## Test plan

CI

### DIFF
--- a/agent/src/AgentWorkspaceDocuments.ts
+++ b/agent/src/AgentWorkspaceDocuments.ts
@@ -75,7 +75,7 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
             const changes = applyContentChanges(fromCache, document.contentChanges)
             contentChanges.push(...changes.contentChanges)
             document.underlying.content = changes.newText
-        } else if (typeof document.content === 'string') {
+        } else if (document.content !== undefined) {
             // Full document sync.
             for (const change of calculateContentChanges(fromCache, document.content)) {
                 contentChanges.push(change)
@@ -93,20 +93,6 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
         if (!document.visibleRange) {
             // No changes to the visible range, populate from cache
             document.underlying.visibleRange = fromCache.protocolDocument.visibleRange
-        }
-
-        // The client may send null values that we convert to undefined here.
-        if (document.content === null) {
-            document.underlying.content = undefined
-        }
-        if (document.contentChanges === null) {
-            document.underlying.contentChanges = undefined
-        }
-        if (document.selection === null) {
-            document.underlying.selection = undefined
-        }
-        if (document.visibleRange === null) {
-            document.underlying.visibleRange = undefined
         }
 
         fromCache.update(document)
--- a/vscode/src/chat/protocol.ts
+++ b/vscode/src/chat/protocol.ts
@@ -50,7 +50,7 @@ export type WebviewMessage =
            */
           command: 'event'
           eventName: string
-          properties: TelemetryEventProperties | undefined
+          properties?: TelemetryEventProperties | undefined | null
       }
     | {
           /**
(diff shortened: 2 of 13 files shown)


sqs:
I don't see any obvious bugs this fixes, but it would have prevented the bugs fixed in #4279, and it will hopefully prevent new bugs like that.

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -86,20 +86,6 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
             document.underlying.visibleRange = fromCache.protocolDocument.visibleRange
         }
 
-        // The client may send null values that we convert to undefined here.
```
sqs, line 89: These if-statements are now guaranteed to be false, but I understand if you'd like to keep them in, @olafurpg, from the battle scars.

agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -66,7 +66,7 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
             const changes = applyContentChanges(fromCache, document.contentChanges)
             contentChanges.push(...changes.contentChanges)
             document.underlying.content = changes.newText
-        } else if (typeof document.content === 'string') {
+        } else if (document.content !== undefined) {
```
olafurpg, line 69: Hmmm, this looks like a behavioral change because `document.content` can be be null. I need to review this diff more closely, only have ~40 more minutes today and need to land a PR in JB. Will come back to this PR on Monday!

agent/src/AgentWorkspaceDocuments.ts
olafurpg: These if statements should be removed, I added them only out of paranoia

vscode/src/chat/protocol.ts
```diff
@@ -50,7 +50,7 @@ export type WebviewMessage =
            */
           command: 'event'
           eventName: string
-          properties: TelemetryEventProperties | undefined
+          properties?: TelemetryEventProperties | undefined | null
```
olafurpg, line 53: Is `| undefined` needed when using `?:`?

I'm fine keeping it, but I don't think I would have added them myself.

vscode/src/jsonrpc/protocol.test.ts
```diff
@@ -0,0 +1,100 @@
+import { basename, relative } from 'node:path'
+import * as ts from 'typescript'
+import { describe, expect, test } from 'vitest'
+
+describe('protocol null/undefined robustness', () => {
```
olafurpg, line 5: 😍 Love this!

I was reviewing the diff thinking we need a linter to prevent future regressions, only to scroll down to the bottom of the diff and see you already hooked into tsc to do the static analysis. Well done!

Double-checked that this catches unwanted undefined properties that don't accept `| undefined`

CleanShot 2024-05-25 at 20 38 05@2x

olafurpg (approved):
LGTM 👍🏻 👍🏻 Awesome work @sqs this is great!

agent/src/AgentWorkspaceDocuments.ts
olafurpg: @sqs You're right here! I only skimmed the diff super quickly on Friday since I was deep working on other stuff.

vscode/src/chat/protocol.ts
olafurpg: If `| undefined` is only added for the linter, then it's possible to check for `.questionToken` on `ts.PropertyDeclaration` like here https://github.com/sourcegraph/scip-typescript/blob/olafurpg/signatures/src/FileIndexer.ts#L460C1-L462C75

vscode/src/jsonrpc/TextDocumentWithUri.ts
```diff
@@ -48,18 +48,18 @@ export class ProtocolTextDocumentWithUri {
     }
 
     public get content(): string | undefined {
-        return this.underlying.content
+        return this.underlying.content ?? undefined
```
keegancsmith, line 51: is this value being the empty string meaningful?

PR #4287, opened 2024-05-24
Agent: add mode to panic when server/client go out of sync

olafurpg:
Previously, when the client and server went out of sync (not agreeing on document content changes, editor selection, editor visible ranges) then it would silently continue running. Now, if the client opt-into it, the agent server will run additional assertions and panic the server proces if the client/server are out of sync. By failing fast, it will be much easier to debug synchronization bugs.

The goal is to enable this panic mode only during testing because it adds performance overhead. Down the road, when troubleshooting customer issues, we can optionally turn on this mode to see if document synchronization is the root of the bug.

## Test plan

* See new test cases
* Manually introduced bug to document synchronization, ran `pnpm run test agent/src/index.test.ts` and confirmed that 1) the process exists and 2) the error message is helpful


RXminuS (approved):
Epicness! Tests are GOOOOD

PR #4290, opened 2024-05-24
Rewrite query for remote context

jtibshirani:
For local keyword context (symf), we first rewrite the query to pull out
relevant keywords. Remote context does not currently rewrite. When digging into
local vs. remote context quality, we found the rewrite to be critical for (1)
queries in foreign languages, and (2) longer queries with multiple sentences or
lines.

This PR updates remote search to write the query when the question appears to
be in a foreign language or has multiple sentences. We added this restriction
because the eval results showed that rewriting made remote search much worse
for short English queries (https://github.com/sourcegraph/context-eval/pull/11)

Closes SPLF-83

## Test plan

Added test cases for remote rewriting and updated snapshots. Tested manually on
VSCode with Japanese prompts.


toolmantim:
@thenamankumar do we need to update the new providers with these changes?

vscode/src/context/remote-search.ts
```diff
@@ -113,8 +115,8 @@ export class RemoteSearch implements ContextStatusProvider {
     }
 
     public async query(query: PromptString): Promise<ContextSearchResult[]> {
-        // Sending prompt strings to the Sourcegraph search backend is fine.
-        const result = await graphqlClient.contextSearch(this.getRepoIdSet(), query.toString())
+        const rewritten = await rewriteKeywordQuery(this.completions, query, true)
```
keegancsmith, line 118: very minor: at this call site knowing what true means is non-obvious. Maybe an options field just for this would make it more obvious so you can do keyword style arguments?

keegancsmith (approved):
LGTM

keegancsmith:
> @thenamankumar do we need to update the new providers with these changes?

@toolmantim I may have missed another provider being added, but I took a look at the two recent ones Naman added (RemoteFileProvider and RemoteRepositorySearch) and they both use non-natural language queries looking for a either a filename or repository name using the sourcegraph API directly. So I don't think they need updating since they are based on the @-mention query input, not the full message.

PR #4292, opened 2024-05-24
Chat: Include editor selection in chat context by default

abeatrix:
## Summary

This change includes the user's current editor selection as context for all new chat messages by default. The selection context is added to the user context items when creating a new chat message.

To support this, the `getContextFileFromSelection` function is now imported and used in `SimpleChatPanelProvider.ts`. If the chat message source is 'chat', the current selection context is retrieved and added to the `userContextFiles` array before resolving the context items.

The selection context retrieval logic is also removed from `getPriorityContext` in `context.ts` since it is now handled in `SimpleChatPanelProvider.ts`.

## Test plan

3. Highlight some code in your editor
2. Start a new chat and ask Cody about the code you highlighted
3. Submit the question with `Opt + Enter` (enhanced context disabled)
4. Confirm your selection is included as context, and Cody is able to answer your question
5. Start a new empty chat and ask Cody about the code you highlighted
6. Submit the question with `Enter` (enhanced context enabled)
7. Confirm your selection is also included as context with enhanced context added, and Cody is able to answer your question

My highlighted code

image

Enhanced Context disabled:

image

Enhanced Context enabled:

image


jtibshirani (approved):
It makes sense to me to always include the selection, and count it towards the "user-supplied context" budget.

This change looks good to me, but someone from Cody should review too :)

CC @chillatom since you're leading the charge on the enhanced context changes for the next release!

thenamankumar (approved):

PR #4295, opened 2024-05-25
[feat] Token usage indicator in the chat view

PriNova:
## Description

This PR introduces a new feature that displays the remaining token counts for chat, user, and enhanced token types in the chat view. This enhancement aims to provide users with real-time visibility into their token usage, helping them manage their token allocation more effectively.

## Changes

- Updated the `TokenCounter` class to expose a new method, `getRemainingTokens()`, which returns an object containing the remaining token counts for each token type.
- Modified the `SimpleChatPanelProvider` to:
  - Retrieve the remaining token counts from the `TokenCounter` instance after building the prompt.
  - Send the remaining token counts to the webview using the `postMessage` method.
- Updated the prompt building logic to return both the `PromptInfo` and `TokenCounter` instances.
- Added a new `remainingTokens` message type to the extension-webview communication protocol to facilitate the transfer of token counts.
- Created a new `TokenIndicators` component that displays the remaining token counts in the chat view UI.
- Integrated the `TokenIndicators` component into the `Chat` component, populating it with the token counts received from the extension.
- Added styling for the `TokenIndicators` component to ensure a visually appealing and intuitive user experience.

## Benefits

- Users can now see their remaining token counts for each token type directly in the chat view.
- The real-time display of token usage helps users make informed decisions about their token allocation and usage.
- The integration of token indicators enhances the overall user experience and provides valuable insights into token consumption.

## Screenshots

Screenshot 2024-06-06 095520

## Checklist

- [x] Tested the changes manually
- [x] Updated relevant documentation (if applicable)
- [x] Followed the project's coding conventions and style guidelines
- [x] Performed a self-review of the code
- [x] Wrote clear and concise commit messages following the conventional commit format

Please review this PR and provide any feedback or suggestions for improvement. Let me know if you have any questions or concerns.

## Test plan

- Manually tested the feature by engaging in chat conversations and verifying that the token indicators update accurately based on the token usage.
- Ensured that the token counts are correctly retrieved from the `TokenCounter` and propagated to the webview.
- Verified that the `TokenIndicators` component renders properly and displays the expected token counts.

## How to test

1. Build extension from this branch
2. Open a new Chat
3. Ask question w/ or w/o enhanced context
4. Check how the Remaining Token UI changes.
5. Choose different models to see usage changes.


PriNova:
Now I've implemented tokencounting visibility for opening a new chat.

The last step is needed to implement if a chat from the history is restored, which needs tokencounting for the various message types and contexts.

This seems to be more of a moderate task. I will use the logic in the `PromptBuilder`for that.
My thougth is to extract this logic into its own method for reuse.
What do you think?

PriNova:
Hey @toolmantim .Are you responsible for reviewing UI changes?

toolmantim:
Hi @PriNova! 👋🏼 I am.

Thanks for doing this, it'd be great as a debug view to start with. Let's make it so that you can turn it on in VS Code settings, but it's off by default. That'll give us a chance to play with it, decide if it makes it into the main UI for everyone, and figure out a final design.

I'll nudge the team to see what they think from an eng/code POV. Thanks for putting this together.

toolmantim:
Let us know when this is out of draft, and when the team has a chance they'll give it a review.

PriNova:
> Let us know when this is out of draft, and when the team has a chance they'll give it a review.

Thank you @toolmantim for your response. Will do.

PriNova:
Hey @sqs I implemented the remaining token UI thing.
Wdyt?

PriNova:
@abeatrix can I win you for a code review?

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -125,8 +129,11 @@ export class DefaultPrompter {
             )
 
             return {
-                prompt: promptBuilder.build(),
-                context,
+                promptInfo: {
+                    prompt: promptBuilder.build(),
+                    context,
+                },
+                tokenCounter: promptBuilder.getTokenCounter,
```
abeatrix, line 136: Can we get the remainingTokens count instead of the TokenCounter instance to avoid cases where TokenCounter getting called somewhere else while the prompt is being built In the future?

vscode/src/prompt-builder/index.ts
```diff
@@ -41,6 +41,10 @@ export class PromptBuilder {
         this.tokenCounter = new TokenCounter(contextWindow)
     }
 
+    public get getTokenCounter() {
+        return this.tokenCounter
+    }
+
```
abeatrix, line 47:
```suggestion

```

The TokenCounter is meant to be used by PromptBuilder only, so we shouldn't make it public imo.

lib/shared/src/token/counter.ts
```diff
@@ -58,6 +58,14 @@ export class TokenCounter {
         return isWithinLimit
     }
 
+    /**
+     * Returns the remaining token counts for different token types.
```
abeatrix, line 62: We have a note below about the remainingTokens should only be used by `@canAllocateTokens to determine if the token usage can be allocated in correct order`. If we want to use it for UI usage, maybe we can clarify in the docstring and the naming, something like `getRemainingTokensForDisplay` to clarify that we are getting remainingTokens for UI use only?

vscode/webviews/Chat.tsx
```diff
@@ -111,6 +112,33 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
...
+    const [remainingTokens, setRemainingTokens] = useState<{
+        chat: number
+        user: number
+        enhanced: number
+        maxChat: number
+        maxUser: number
+        maxEnhanced: number
+    } | null>(null)
+
+    // Update the message handler to set the remaining token counts when received from the extension
```
abeatrix, line 124: we should move this to the Handler in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/App.tsx?L71-152

vscode/webviews/Chat.tsx
```diff
@@ -111,6 +112,33 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
 
     const postMessage = useCallback<ApiPostMessage>(msg => vscodeAPI.postMessage(msg), [vscodeAPI])
 
+    const [remainingTokens, setRemainingTokens] = useState<{
+        chat: number
+        user: number
+        enhanced: number
+        maxChat: number
+        maxUser: number
+        maxEnhanced: number
```
abeatrix, line 121: this looks like duplicates because we can get the values from [chatModels.contextWindow](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/App.tsx?L51) ?

abeatrix (commented):
Thanks for adding this useful feature for debugging and for the ping. 

I did a quick review and left some comments inline. My main concern is exposing the TokenCounter outside of the prompt building steps would allow other parts of the codebase to use it which could cause issues in the future. I know it's not an issue right now, but  imo we should keep the TokenCounter within the PromptBuilder class only, and send the values of remainingTokens for UI purpose only, wdyt?

cc @dominiccooney

vscode/webviews/Chat.tsx
PriNova: Hoisted into the App.tsx

vscode/webviews/Chat.tsx
PriNova: The model in use is exposed only in the model selector. If I want to retrieve the model context limits here (now in App.tsx), I need to make changes in the chatModelContext.tsx, which IMO would clutter the code and break the single responsible. If you see another way to retrieve the values with least surprise, I'm all ear.
This is why I put the logic outside of UI logic handling.

lib/shared/src/token/counter.ts
PriNova: Renamed for better clarification.

vscode/src/chat/chat-view/prompt.ts
PriNova: Done

vscode/src/prompt-builder/index.ts
PriNova: Changed to not expose the TokenCounter

abeatrix (commented):
Thanks for making the changes! I tested it locally and noticed that the current approach might not work because we don't get remainingTokens for chats started from history, so they always show up with the max value at the start.
image

I also wonder how helpful it is to display the remainingTokens from the TokenCounter after experimenting with this. When the token limit is full, users can still add new messages and user context since we will start removing old content. It might be more useful to display the token usage of the current input instead of showing the remaining tokens from the last input. Maybe this is a design question for @toolmantim ?

PriNova:
> Thanks for making the changes! I tested it locally and noticed that the current approach might not work because we don't get remainingTokens for chats started from history, so they always show up with the max value at the start. image
> 
> I also wonder how helpful it is to display the remainingTokens from the TokenCounter after experimenting with this. When the token limit is full, users can still add new messages and user context since we will start removing old content. It might be more useful to display the token usage of the current input instead of showing the remaining tokens from the last input. Maybe this is a design question for @toolmantim ?

Yeah, you are right. It currently does not work with chat history. This PR should be served as a baseline implementation and if the User might see benefits, it could be enhanced further.
For the chat history, if a chat is opened from the history, all the messages and context contents needs to be calculated to show the remaining tokens in the indicator, which is not implemented right now.
To be honest, I could not detect the correct data flow. Is it the transscript or the models.history?

PR #4296, opened 2024-05-25
Reseting document content properties is happening too late in the code

pkukielka:
## Test plan

Just run any client with with `cody-agent.panic-when-out-of-sync` - it should not fail immediately.


RXminuS (approved):

olafurpg:
FYI @pkukielka , we merged https://github.com/sourcegraph/cody/pull/4286 which ensures we're using `| null` consistently in the agent protocol. The PR also adds a regression suite that errors if we add new optional properties without `| null`. The `null` handling in this PR should no longer be needed, but we can keep it regardless.

PR #4297, opened 2024-05-25
Makes "undefined" check more resillient

RXminuS:
Typescript !== TypeSafe. In this case messages coming over the JsonRPC are null instead of undefined.

And `typeof null` is obviously...`object`!

We should (and will) be resolving this on a more robust higher level but for now having the most resillient null/undefined, even if not the prettiest, is solves the issue.

## Test plan

- unit tests still pass.
- Checked that manually setting to null does not break


pkukielka (approved):
LGTM

PR #4298, opened 2024-05-25
Agent: add `mode` parameter to editCommands/code

abeatrix:
This change adds a new `mode` parameter to the `editCommands/code` request in the agent protocol. The `mode` parameter can be either `'edit'` or `'insert'`, and it allows the client to specify whether the edit command should perform an edit or an insertion operation.

The corresponding changes have been made in the Kotlin binding and the TypeScript agent implementation to support this new parameter.

2: add checks for empty selection in doc and test commands

This change adds checks for empty selections in the `executeDocCommand` and `executeTestEditCommand` functions. If the selected text is empty (AFTER we have expanded the range and not before), the functions will throw an error to prevent the commands from executing on an empty selection.

The changes also include some minor logging improvements to provide more verbose information about the arguments being passed to the commands. This also prevent sending invalid request to the LLM that would return unhelpful response that would break the user's document. 

~3. Cleaned up the format for the prompt where we wrap the selection code in to make sure they are all standardized. See more context in https://sourcegraph.slack.com/archives/C05AGQYD528/p1716585733323689~ 

^ will more this to a separate PR once the recordings token is fixed

## Test plan

We already have an agent side test for the "insert" and "edit" mode: https://sourcegraph.com/github.com/sourcegraph/cody@vscode-v1.18.2/-/blob/agent/src/index.test.ts?L1283-1333

Follow the test plan in https://github.com/sourcegraph/jetbrains/pull/1649 to verify the change works with client

Here is a Demo with the changes made in this PR & https://github.com/sourcegraph/jetbrains/pull/1649

https://github.com/sourcegraph/cody/assets/68532117/b1805a4d-e94d-4a5a-8f85-1d4d878ce51d

### DIFF
--- a/vscode/src/jsonrpc/agent-protocol.ts
+++ b/vscode/src/jsonrpc/agent-protocol.ts
@@ -82,7 +82,7 @@ export type ClientRequests = {
     'commands/custom': [{ key: string }, CustomCommandResult]
 
     // Trigger commands that edit the code.
-    'editCommands/code': [{ instruction: string; model?: string }, EditTask]
+    'editCommands/code': [{ instruction: string; model?: string; mode?: 'edit' | 'insert' }, EditTask]
     'editCommands/test': [null, EditTask]
     'editCommands/document': [null, EditTask]
 
--- a/vscode/src/jsonrpc/agent-protocol.ts
+++ b/vscode/src/jsonrpc/agent-protocol.ts
@@ -83,7 +83,11 @@ export type ClientRequests = {
 
     // Trigger commands that edit the code.
     'editCommands/code': [
-        { instruction: string; model?: string | undefined | null; mode?: 'edit' | 'insert' },
+        {
+            instruction: string
+            model?: string | undefined | null
+            mode?: 'edit' | 'insert' | undefined | null
+        },
         EditTask,
     ]
     'editCommands/test': [null, EditTask]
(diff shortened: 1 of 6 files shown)


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -82,7 +82,14 @@ export type ClientRequests = {
     'commands/custom': [{ key: string }, CustomCommandResult]
 
     // Trigger commands that edit the code.
-    'editCommands/code': [{ instruction: string; model?: string | undefined | null }, EditTask]
+    'editCommands/code': [
+        {
+            instruction: string
+            model?: string | undefined | null
+            mode?: 'edit' | 'insert' | undefined | null
```
steveyegge, line 89: Do we want to include/distinguish 'delete' while we're at it, or is it just basically an edit for these purposes?

steveyegge (approved):

vscode/src/jsonrpc/agent-protocol.ts
abeatrix: Yea basically just edit for these purposes that we currently support. They are different than the workspace edit actions so no "delete" mode. Just something we create edit commands with before they are turned into an EditTask so that the FixupController knows process the generated code.

Here is an example on how we create the doc command in the back:

```

        return {
            type: 'edit',
            task: await executeEdit({
                configuration: {
                    instruction,
                    intent: 'doc',
                    mode: 'insert',
                    range,
                    insertionPoint,
                },
                source: args?.source,
            } satisfies ExecuteEditArguments),
 

```

I can work on adding the rest of the params support into agent and clients after GA (didn't want to introduce too many changes at once at this time)

PR #4299, opened 2024-05-25
Agent: implement `vscode.TextDocument` with CodeMirror

olafurpg:
While investigating JetBrains document synchronization issues, it became obvious that we need a more robust and higher-performance solution to deal with incremental document changes on the agent server side. Previously, the solution was a naive brute-force algorithm that allocated a ton of strings for every single character edit. Now, we use CodeMirror to handle all the editor state, including interpretation of incremental document edits.

On a basic benchmark simulating editing a 50k loc file where each line is 50 characters:

- Updates are now instant (<0ms) if we don't read the document contents.
- Reading the document takes ~1-2ms on my M1 Max.
- Before, every update took ~5ms regardless if we read the document or not.

## Test plan

Have all existing tests green without disabling anything. Currently, some tests related to handling of newlines (esp. lines without newlines, and \r\n newlines).

### DIFF
--- a/agent/package.json
+++ b/agent/package.json
@@ -36,6 +36,7 @@
   "bin": "dist/index.js",
   "files": ["dist/index.js", "dist/index.js.map", "dist/*.wasm"],
   "dependencies": {
+    "@codemirror/state": "^6.4.1",
     "@pollyjs/core": "^6.0.6",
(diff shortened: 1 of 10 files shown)


olafurpg:
Closing this as wontfix for now. Happy to revisit if there's interest, or if it solves an important problem

PR #4305, opened 2024-05-27
focus followup input after sending chat message

sqs:
In the new chat "stream" UI, the original message input stayed focused after submitting. This felt weird. Now, the followup input is focused (but not auto-scrolled) after submitting.

Known issues (will fix before release, after merge):

- If you type when the followup input is focused but not visible, then it scrolls down but *just* enough to show your input and not enough to show the entire followup input field.
- We may want to have some kind of nice sticky positioning of the input.
- In some cases, if you edit a message then submit a new message at the end, the focus is lost and does not properly transfer to the last message input.

https://github.com/sourcegraph/cody/assets/1976/af093320-5d6f-42a1-a6b6-648d52d306f9

## Test plan

CI


toolmantim (approved):
Nice! I think this is working well.

A few things I found:
* There's a race condition when you edit an existing message and hit enter a few times
* We can do something nice with "Waiting…" in a follow-up
* We can add a "↓" down/follow button in the bottom right as a follow-up

PR #4308, opened 2024-05-27
fix race conditions in sending and editing messages

sqs:
Fixes intermittent `Cannot set new context used for bot message` errors in chat when you are editing a message.

If a submit or edit operation was aborted, its prompt builder would continue trying to mutate the transcript. Now it fully aborts.

Instead of wrapping ONLY the low-level LLM interaction with cancelation, now the entire submit or edit operations are wrapped in a cancelation.

## Test plan

CI

Also, try spamming new messages by pressing Enter repeatedly in a chat. Previously, you would be likely to get some `Cannot set new context used for bot message` errors, and now, it does not show errors and properly restarts the operation cleanly each time.

### DIFF
--- a/lib/shared/src/sourcegraph-api/errors.ts
+++ b/lib/shared/src/sourcegraph-api/errors.ts
@@ -97,10 +97,12 @@ export function isAbortError(error: unknown): error is AbortError {
         isError(error) &&
         // custom abort error
         ((error instanceof AbortError && error.isAbortError) ||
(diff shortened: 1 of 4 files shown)


toolmantim:
This didn't seem to work for me:

https://github.com/sourcegraph/cody/assets/153/c8941747-8961-4c4e-9742-c0b976af78ba

sqs:
merged in https://github.com/sourcegraph/cody/pull/4305

PR #4309, opened 2024-05-27
remove "(by OpenCtx)" in @-mention menu

sqs:
We want these to feel native.

Also the heading of the @-mention menu when you've selected a provider now just contains the provider name itself, instead of `Search using PROVIDER provider`, which could be ungrammatical or clunky. We should have a way in the OpenCtx `meta` result to specify the right label here, but for now, this suffices.

## Test plan

CI


toolmantim (approved):
Lovely! Would love this to be provider-defined.

PR #4310, opened 2024-05-27
Agent: close P0 memory leak on `textDocument/didChange`

olafurpg:
Fixes https://github.com/sourcegraph/jetbrains/issues/1587

Previously, the agent was firing the
`vscode.window.onDidChangeVisibleTextEditors` event on every `textDocument/didChange` event, including when we only updated the selection (no content changes). In VSC, this event only fires when you open a new file, change the actived file tab, or close a file. Now, the agent mirrors the VSC behavior by firing the event less frequently. This change doesn't actually fix the memory leak, it just makes it very difficult to reproduce because you need to open/close files thousands of times. Previously, it was easy to reproduce, you could just drag-select text with the mouse and fire thousands of events.

## Test plan

See new `memory.test.ts` file. This test suite can be skipped by default in CI because it's going to be difficult to maintain this as an automated test. However, we used this test file to reproduce the memory leak, and verified that it no longer reproduces after this fix.

### DIFF
--- a/vscode/src/tree-sitter/parse-tree-cache.ts
+++ b/vscode/src/tree-sitter/parse-tree-cache.ts
@@ -46,8 +46,6 @@ export async function parseDocument(document: TextDocument): Promise<void> {
         return
     }
 
-    console.log('boom')
-
     updateParseTreeCache(document, parser)
 }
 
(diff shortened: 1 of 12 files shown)


vscode/src/tree-sitter/parse-tree-cache.ts
```diff
@@ -46,6 +46,8 @@ export async function parseDocument(document: TextDocument): Promise<void> {
         return
     }
 
+    console.log('boom')
```
valerybugakov, line 49:
```suggestion
```

valerybugakov (approved):

mkondratek (approved):

PR #4311, opened 2024-05-27
Implement @-mention multi-repo OpenCtx context provider for enterprise

thenamankumar:
This PR implements a Cody internal OpenCtx provider for @-mention context. This provider is only shown to enterprise users. The provider shows a list of repos as mention items and uses the chat message as a query to perform a remote search for the repo at sourcegraph instance using the `getCodyContext` keyword search powered by the Zoekt API. 

## Test plan
- connect to sourcegraph.sourcegrah.com
- Try @ "Repositories", write repo name to search for a repo and then select it.
- Write your chat message and submit.
- Cody will include context from the remote repo.


vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -155,16 +159,23 @@ export const MentionMenu: FunctionComponent<
...
-                                className={styles.item}
-                            >
-                                <MentionMenuProviderItemContent provider={provider} />
-                            </CommandItem>
-                        ))}
+                        {data.providers
+                            .filter(
+                                provider =>
+                                    provider.id !== RemoteRepositorySearch.providerUri ||
+                                    (userInfo && !userInfo.isDotComUser)
```
valerybugakov, line 166: Would it work too?

```suggestion
                                    provider.id !== RemoteRepositorySearch.providerUri || !userInfo?.isDotComUser
```

vscode/webviews/mentions/providers.tsx
```diff
@@ -45,23 +41,30 @@ const getAllMentionProvidersMetadata = async (): Promise<ContextMentionProviderM
 }
 
 export const WithContextProviders = (props: { children: React.ReactElement }): React.ReactElement => {
+    const [providersPromise, setProvidersPromise] = useState(getAllMentionProvidersMetadata())
+    const [providers, setProviders] = useState<ContextMentionProviderMetadata[]>([])
```
valerybugakov, line 45: Here, we use two states to load a single request. What do you think about doing it with one to simplify the logic?

vscode/webviews/mentions/providers.tsx
```diff
@@ -45,23 +41,30 @@ const getAllMentionProvidersMetadata = async (): Promise<ContextMentionProviderM
...
+
+    useEffect(() => {
+        void (async () => {
+            const providers = await providersPromise
+            setProviders(providers)
+        })()
+    }, [providersPromise])
```
valerybugakov, line 52:
```suggestion
    const [providers, setProviders] = useState<ContextMentionProviderMetadata[]>([])

    const loadProviders = async () => {
        const providersData = await getAllMentionProvidersMetadata()
        setProviders(providersData)
    }

    useEffect(() => {
        loadProviders()
    }, [])
```

vscode/webviews/mentions/providers.tsx
```diff
@@ -45,23 +41,30 @@ const getAllMentionProvidersMetadata = async (): Promise<ContextMentionProviderM
...
     return (
-        <context.Provider value={{ providers: getAllMentionProvidersMetadata() }}>
+        <context.Provider
+            value={{
+                providers,
+                reload: () => {
+                    setProvidersPromise(getAllMentionProvidersMetadata())
+                },
+            }}
+        >
```
valerybugakov, line 62:
```suggestion
        <context.Provider
            value={{
                providers,
                reload: loadProviders,
            }}
        >
```

valerybugakov (approved):
LETSGO!

lib/shared/src/context/openctx/api.ts
```diff
@@ -13,9 +13,5 @@ export const openCtx = new OpenCtx(undefined)
  * Set the handle to the OpenCtx client.
  */
 export function setOpenCtxClient(client: OpenCtxClient): void {
-    if (openCtx.client) {
-        throw new Error('OpenCtx extension API is already set')
-    }
-
```
valerybugakov, line 19: Should we keep this?

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
thenamankumar: Logically no, because the userInfo might be null and that won't tell if the user is connected to enterprise. Practically it might work, but lets keep it for extra safety and that we do not show it for anyone other than enterprise users.

lib/shared/src/context/openctx/api.ts
thenamankumar: No, we do need to update the client each time the config updates.

vscode/webviews/mentions/providers.tsx
thenamankumar: Thanks for the patches. Resolved by those.

PR #4318, opened 2024-05-28
Do not use Number.MAX_SAFE_INTEGER in agent protocol

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/1623

## Test plan

1. Open `sourcegraph/cody`
2. Open `agent/src/index.test.ts`
3. Put cursor in some nested code. I used line with `accessToken: 'REDACTED_0ba08837494d00e3943c46999589eb29a210ba8063f084fff511c8e4d1503909',`
4. Try to run any edit command

It should fail before my fix and pass after.

### DIFF
--- a/vscode/src/editor/utils/index.ts
+++ b/vscode/src/editor/utils/index.ts
@@ -42,7 +42,7 @@ export async function getSmartSelection(
         // Regardless of the columns provided, we want to ensure the edit spans the full range of characters
         // on the start and end lines. This helps improve the reliability of the output.
         const adjustedStartColumn = document.lineAt(startPosition.row).firstNonWhitespaceCharacterIndex
-        const adjustedEndColumn = Number.MAX_SAFE_INTEGER
+        const adjustedEndColumn = document.lineAt(startPosition.row).range.end.character
         return new vscode.Selection(
             startPosition.row,
             adjustedStartColumn,


odisseus (approved):

odisseus:
This value is used in a couple other places. Shouldn't we replace those occurrences as well?

pkukielka:
> This value is still used in a couple of other places. Shouldn't we replace those occurrences as well?
I have seen it, but other places doesn't seem to be used by agent right now and I'm not sure if replacing them wouldn't break something else, so I'm playing safe for now.

olafurpg:
The contract of positions is that `Number.MAX_SAFE_INTEGER` is a valid character, and the client should use the last character in the line here. This should not be fixed by the agent server IMO, it's a client bug.

olafurpg:
The root bug could be in `Position.fromOffset` here https://sourcegraph.com/github.com/sourcegraph/jetbrains@12167331689275ba0ef0ad0074ac8ea7fa6e7552/-/blob/src/main/kotlin/com/sourcegraph/cody/agent/protocol/Position.kt?L10

The `character` property should be a `Long` (JS/JSON doesn't have `Int`) and we should use `Math.min(lineLength, character)`

PR #4325, opened 2024-05-28
Document Code: Always expand to the start of a line as a fallback

umpox:
## Description

Fixes an issue where the "Document Code" command would incorrectly insert documentation at the cursor position: This occurred when the cursor was in the middle of some code, but we did not have a tree-sitter `documentableNode` or a folding range to expand to.

We now always fallback to the **line selection** and adjust the `insertionPoint` to be at the start of this line.

## Test plan

1. Place the cursor in the middle of some text (e.g. in a Markdown file)
2. Run "Document Code"
3. Expect that the inserted documentation is inserted in the line above the cursor

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1386,7 +1386,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
         this.registerRequest(method, async (params, token) => {
             await this.authenticationPromise
             return callback(params, token).catch(err => {
-                console.error(err)
+                console.error(`Uncaught error from method '${method}'`, err)
                 throw err
             })
         })
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1385,7 +1385,10 @@ export class Agent extends MessageHandler implements ExtensionClient {
     ): void {
         this.registerRequest(method, async (params, token) => {
             await this.authenticationPromise
-            return callback(params, token)
+            return callback(params, token).catch(err => {
+                console.error(err)
+                throw err
+            })
         })
     }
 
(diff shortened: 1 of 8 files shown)


vscode/src/commands/execute/doc.ts
```diff
@@ -11,28 +11,39 @@ import type { CodyCommandArgs } from '../types'
...
-    // No usable selection, fallback to expanding the range to the nearest block,
-    // as is the default behavior for all "Edit" commands
-    return {}
+    const smartSelection = await getEditSmartSelection(editor.document, editor.selection, {}, 'doc')
+    if (!smartSelection.isEmpty) {
+        return {
+            range: smartSelection,
+            insertionPoint: smartSelection.start,
+        }
+    }
```
umpox, line 40: It's a bit odd that `getEditSmartSelection` can return an empty selection. We might want to just return `null` in the case and filter this out. I'll look into that in a follow-up

agent/src/agent.ts
```diff
@@ -1385,7 +1385,10 @@ export class Agent extends MessageHandler implements ExtensionClient {
     ): void {
         this.registerRequest(method, async (params, token) => {
             await this.authenticationPromise
-            return callback(params, token)
+            return callback(params, token).catch(err => {
+                console.error(err)
```
olafurpg, line 1389:
```suggestion
                console.error(`Uncaught error from method '${method}'`, err)
```

olafurpg (approved):
LGTM 👍🏻 We may want to add a few more tests (markdown/Rust) for a few common scenarios (classes, structs, type alias, ...). Not blocking this PR, but might be worth following up on

PR #4340, opened 2024-05-28
Print only a prefix of very long selections in the debug logs

steveyegge:
We had a logDebug statement that was often printing dozens to hundreds of lines; I've limited it to 70 characters max. If we need the ability to see the whole selection in an actual debug session, then we can add a flag to support it.

## Test plan

This is a debug-logging update and requires no special QA attention.


abeatrix (approved):
sorry that was not a left over and not added intentionally. Thanks for fixing it!

Issue #4321, opened 2024-05-28
bug: "Add Selection to Cody Chat" does not work for follow up chat messages


JeremyTCD:
### Version

v1.19.1716878244

### Describe the bug

"Add Selection to Cody Chat" does not work for follow up messages. Focus shifts to the input box, but the selection is not added.

This issue occurs both when I use the right click context menu and when I use the "alt + l" shortcut.

https://github.com/sourcegraph/cody/assets/11733898/db88f37f-f443-4b75-b940-84cbb79c71b9

### Expected behavior

Code should be added to chat box.

### Additional context

Using: 

- VSCode 1.89.1
- Cody Pro
- Cody v1.19.1716878244 pre-release

PR #4341, opened 2024-05-28
fix: remove isFirstMessage check for adding userContextFromSelection

abeatrix:
CLOSE https://github.com/sourcegraph/cody/issues/4321

Remove isFirstMessage check for adding user Context from selection via  `Add selection to Cody Chat`. 

Currently `isFirstMessage` blocks users from adding selection to Cody chat if the message is not the first message.

## Test plan

2. Start new chat and send Cody a question
3. Select some code in your editor
4. Right click in your selected code and select `Add selection to Cody Chat`
5. Verify the selected code is added as @-mention context in your chat window

image

Before this change, users are not able to `Add selection to Cody Chat` in non-new chat

### DIFF
--- a/vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
+++ b/vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
@@ -184,10 +184,10 @@ export const HumanMessageEditor: FunctionComponent<{
             return
         }
         const editor = editorRef.current
-        if (editor && isFirstMessage) {
-            editorRef.current?.addContextItemAsToken(userContextFromSelection)
+        if (editor) {
+            editor?.addContextItemAsToken(userContextFromSelection)
         }
-    }, [userContextFromSelection, isFirstMessage])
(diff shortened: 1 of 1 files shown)


vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
```diff
@@ -184,10 +184,10 @@ export const HumanMessageEditor: FunctionComponent<{
             return
         }
         const editor = editorRef.current
-        if (editor && isFirstMessage) {
-            editorRef.current?.addContextItemAsToken(userContextFromSelection)
+        if (editor) {
```
abeatrix, line 187: Cc @sqs since you might have the latest context here. Just want to confirm if it's ok to remove the `isFirstMessage` check here?

sqs: yes, thanks for the catch!

sqs (approved):

toolmantim:
Thanks for fixing this!

PR #4342, opened 2024-05-28
Add initial test suite for Chat response quality

jtibshirani:
This PR adds a new unit test suite for LLM responses. It exercises a few
scenarios like missing context, and ensures that Cody doesn't break character,
hedge, or hallucinate files.

In a follow-up, we'll expand this suite with dogfooding and customer examples.
This test will then let us iterate on a better preamble to help Cody stay in character
and avoid hedging.

Based on this RFC:
https://docs.google.com/document/d/1I-xTsAOsDjv_G-J3vt9oQKuWYLJnr8wehrRMMef1XR4/edit

## Test plan

Test-only change


agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,155 @@
...
+            if (contextFile.uri.path.includes(file)) {
+                found = true
+                break
+            }
+        }
+        expect(found, `file ${file} does not exist in context`).toBe(true)
+    }
+}
+
+const contextFiles: ContextItem[] = [
```
jtibshirani, line 123: Question for reviewers: is there a way I can just access files from the local workspace (containing `github.com:sourcegraph-testing/pinned-zoekt.git`), so I don't need to copy in this content?

agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    describe('Questions about origin', () => {
+        it('Who are you?', async () => {
+            const lastMessage = await client.sendSingleMessageToNewChat('Who are you?')
+            expect(lastMessage?.text?.toLocaleLowerCase() ?? '').includes('sourcegraph')
+            expect(lastMessage?.text ?? '').not.toMatch(characterCheck)
+            expect(lastMessage?.text ?? '').not.toMatch(hedgingCheck)
+        }, 10_000)
+
+        it.skip('Who created you??', async () => {
```
jtibshirani, line 58: A bunch of these tests are marked 'skip' because we fail them. In a follow-up, I will experiment with updating the preamble so we pass them.

abeatrix: agree we will need to update the preamble to get a consistent answer to this question

agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    for (const file of files) {
+        let found = false
+        for (const contextFile of contextFiles) {
+            if (contextFile.uri.path.includes(file)) {
+                found = true
+                break
+            }
+        }
+        expect(found, `file ${file} does not exist in context`).toBe(true)
+    }
```
abeatrix, line 163:
```suggestion
    const filenameRegex = /\b`(\w+\.\w+)`\b/g;
    const files = lastMessage?.text?.match(filenameRegex) ?? [];
    const contextFilePaths = new Set(contextFiles.map(file => file.uri.path));

    for (const file of files) {
        expect(contextFilePaths.has(file), `file ${file} does not exist in context`).toBe(true);
    }
```

agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,215 @@
...
+        }, 10_000)
+    })
+
+    describe('Poor context provided', () => {
+        it.skip('No context', async () => {
+            const lastMessage = await client.sendSingleMessageToNewChat(
+                "What's the license of this repo?",
+                { addEnhancedContext: false, contextFiles: [] }
+            )
+            checkLastMessage(lastMessage)
```
abeatrix, line 96: Is this check correct if we are not expecting the LLM to answer this question?

abeatrix (commented):
Nice suite! Left some comments but I think this is a great start!

agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,215 @@
...
+                contextFiles: contextFiles,
+            })
+            checkLastMessage(lastMessage)
+            checkNoFiles(lastMessage)
+            expect(lastMessage?.text?.toLocaleLowerCase() ?? '').includes('sourcegraph')
+        }, 10_000)
+    })
+
+    describe('Questions about knowledge base', () => {
+        it.skip('What files do you have access to?', async () => {
```
abeatrix, line 77: What do we want to check in here? 
This might not be a good question to ask since they don't have access to any files, and we only share "code snippets" with them, so this is most likely going to return false.

agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,215 @@
...
+    beforeEach(async () => {
+        await client.request('testing/reset', null)
+    })
+
+    describe('Questions about origin', () => {
+        it('Who are you?', async () => {
+            const lastMessage = await client.sendSingleMessageToNewChat('Who are you?')
+            checkLastMessage(lastMessage)
+            checkNoFiles(lastMessage)
+            expect(lastMessage?.text?.toLocaleLowerCase() ?? '').includes('sourcegraph')
```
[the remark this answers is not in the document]
jtibshirani: Oops, yes I didn't update this after tweaking the question!

agent/src/chat-response-quality.test.ts
jtibshirani: What I'm thinking: we should never return a hedging answer like "Unfortunately ..." or "without access to your codebase..." This happens pretty often when no context is provided, and it really confuses users because Cody is supposed to be confident and know about their codebase. `checkLastMessage` is enforcing this.

Instead, in these cases we should give a helpful response like "I don't have a great answer, try providing more context..." and tell them about how to do that. I aim to fix this test in a follow-up.

agent/src/chat-response-quality.test.ts
jtibshirani: I agree that the question is a little weird, I'll change it to "What code do you have access to?"

Right now we give a response like this, which is really confusing for users:
> I don't actually have direct access to any code repositories or files. I'm an AI assistant created by Anthropic to be helpful, harmless, and honest.

I'd love if we said "I have access to your codebase. Try including context by ..." We can work on fixing this test in a follow-up.

abeatrix (approved):
i think this is a good start and we can update as we improve our current prompts!

agent/src/chat-response-quality.test.ts
jtibshirani: @abeatrix did you have any pointers here?

abeatrix: I think we are just using the files we created for tests in the [agent/src/__tests__/](https://sourcegraph.com/github.com/sourcegraph/cody/-/tree/agent/src/__tests__/example-ts/src) directory, and don't think we are getting content from remote sauces. cc @olafurpg who might have better ideas

olafurpg: +1 to @abeatrix's comment

I recommend creating a new `chat-response-quality` directory in https://sourcegraph.com/github.com/sourcegraph/cody/-/tree/agent/src/__tests__ and replace `example-ts` above to this directory
```
const workspace = new TestWorkspace(path.join(__dirname, '__tests__', 'chat-response-quality'))
```

agent/src/chat-response-quality.test.ts
```diff
@@ -0,0 +1,210 @@
...
+    const filenameRegex = /\b`(\w+\.\w+)`\b/g
+    const files = lastMessage?.text?.match(filenameRegex) ?? []
+    const contextFilePaths = new Set(contextFiles.map(file => file.uri.path))
+    for (const file of files) {
+        expect(contextFilePaths.has(file), `file ${file} does not exist in context`).toBe(true)
+    }
+}
+
+function checkSymbolsExist(lastMessage: SerializedChatMessage | undefined, contextFiles: ContextItem[]) {
+    const symbolRegexp = /\b\w*[a-z][A-Z_]\w*\b/
```
mmanela, line 162: Couldn't a valid response return symbols that are from a std lib?

olafurpg (approved):
LGTM 👍🏻 I love seeing us use this testing infrastructure more 😊 

FYI, I am working on evals for Fix. I'm optimistic it'll be easy to extend the eval infrastructure to perform basic Chat quality benchmarks based on a mix of rule-based logic like in this PR and also LLM judge (I will open a PR soon adding basic LLM judge infrastructure).

jtibshirani:
> FYI, I am working on evals for Fix. I'm optimistic it'll be easy to extend the eval infrastructure to perform basic Chat quality benchmarks based on a mix of rule-based logic like in this PR and also LLM judge (I will open a PR soon adding basic LLM judge infrastructure).

Excited for this! I'm happy to migrate over to a new framework when it's available. (The main work in this PR was figuring out the set of invariants that correlate nicely with response quality, not the test set-up, etc.)

agent/src/chat-response-quality.test.ts
jtibshirani: Yes, and I'll make sure to adapt this when we start asking questions like "how do I do... or "generate..." For now I'll keep it simple!

agent/src/chat-response-quality.test.ts
jtibshirani: Thank you both for the pointers. Will update.

PR #4343, opened 2024-05-29
Add Sourcegraph File Context for Enterprise

thenamankumar:
Similar to the Sourcegraph Repositories context provider, this PR adds a Sourcegraph Files provider. Which only loads under the `config.experimental.noodle` config flag right now.

But this is not as clean as the previous provider. Searching for files on Sourcegraph without first selecting a repo is not a great experience because a global file search based on file path won't get the desired results. Imagine searching for client.js, there will be 100s of those. 

The OpenCtx protocol does not provide a way to do a two-step selection for multiple items and we do not modify the protocol to introduce the complexity. 

So, in this PR, we are doing a HACK to pull this off. Given this is an internal Cody provider, this should be fine. 

Here is what it does:

1. The provider separates the repoName from filePath based on `:`. So `[repoName, filePath] = query.split(':')`. If any repo or filePath contains `:` in them, it is fine that the search will not work for them, and IMHO, it should not; who even puts a `:` in file names?
2. Anyways, the user does not have to manually put in the full repo name. Whatever query they type first, we consider it to query the repos and show them repos in the list. 
3. When the user selects a repo from the list, we put the full name of that repo in the query and append a `:`, so now the provider will start showing them files in the list. On this selection, we do not set the repo item as mention yet and keep the provider selected.
4. Now they can simply write a query to search files and then select it. Now we will set the file item as mention and reset the mention menu state. 

## Test plan

Manually tested. 

Demo: https://www.loom.com/share/725597bf6a4a4e17bb7124b460f124fb?sid=9b9ba412-67fb-4c98-b480-56095ce8a6c8

### DIFF
--- a/lib/shared/src/mentions/api.ts
+++ b/lib/shared/src/mentions/api.ts
@@ -139,6 +139,7 @@ export async function openCtxMentionProviders(): Promise<ContextMentionProviderM
                 queryLabel: provider.name,
                 emptyLabel: 'No results found',
             }))
+            .sort((a, b) => (a.title > b.title ? 1 : -1))
     } catch (error) {
         logDebug('openctx', `Failed to fetch OpenCtx providers: ${error}`)
         return []
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -26,6 +26,8 @@ import {
     CURRENT_USER_ID_QUERY,
     CURRENT_USER_INFO_QUERY,
     EVALUATE_FEATURE_FLAG_QUERY,
+    FILE_CONTENTS_QUERY,
+    FILE_MATCH_SEARCH_QUERY,
     GET_FEATURE_FLAGS_QUERY,
     LOG_EVENT_MUTATION,
     LOG_EVENT_MUTATION_DEPRECATED,
@@ -162,6 +164,37 @@ export interface RepoSearchResponse {
             endCursor: string | null
         }
     }
+}
+export interface FileMatchSearchResponse {
+    search: {
+        results: {
+            results: {
+                __typename: string
+                repository: {
+                    name: string
+                }
+                file: {
+                    url: string
+                    path: string
+                    commit: {
+                        oid: string
+                    }
+                }
+            }[]
+        }
+    }
+}
+
+export interface FileContentsResponse {
+    repository: {
+        commit: {
+            file: {
+                path: string
+                url: string
+                content: string
+            } | null
+        } | null
+    } | null
 }
 
 interface RepositoryIdResponse {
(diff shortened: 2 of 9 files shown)


vscode/webviews/promptEditor/nodes/ContextItemMentionNode.ts
```diff
@@ -205,7 +205,7 @@ export function contextItemMentionNodeDisplayText(contextItem: SerializedContext
         case 'github_issue':
             return `@github:issue:${contextItem.owner}/${contextItem.repoName}/${contextItem.issueNumber}`
         case 'openctx':
-            return `@${contextItem.title}`
+            return `@${contextItem.mention?.data?.mentionLabel || contextItem.title}`
```
eseliger, line 208: why can't we use title here and need to rely on an undocumented property on the `data` field?

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -122,14 +124,46 @@ export const MentionMenu: FunctionComponent<
     )
 
     const onCommandSelect = useCallback(
-        (value: string): void => {
+        async (value: string): Promise<void> => {
```
eseliger, line 127: using async in callbacks in react means there's no good error handling for rejections IIRC, should we try catch this to prevent unhandled promise rejection errors? 
(maybe I'm wrong here)

thenamankumar: we definitely should !

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -163,6 +165,37 @@ export interface RepoSearchResponse {
         }
     }
 }
+export interface FileMatchSearchResponse {
```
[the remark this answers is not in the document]
thenamankumar: nah

vscode/webviews/promptEditor/nodes/ContextItemMentionNode.ts
thenamankumar: As discussed I am going to all the label thing in the openctx protocol because the title  is used to display the list which not be the same as the label for the @-mention chip, specifically when you are implementing something like 2 step search for mention items. 

I'll do is separate from this PR and then it would look like `#${contextItem.label || contextItem.title}`.

lib/shared/src/mentions/api.ts
```diff
@@ -139,6 +139,7 @@ export async function openCtxMentionProviders(): Promise<ContextMentionProviderM
                 queryLabel: provider.name,
                 emptyLabel: 'No results found',
             }))
+            .sort((a, b) => (a.title > b.title ? 1 : -1))
```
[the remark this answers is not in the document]
thenamankumar: I checked with Tim and he is okay to have it alphabetically sorted.

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
thenamankumar: I figured that was okay to not make this change and just put a const title which required no await promise. Not an issue now.

PR #4346, opened 2024-05-29
fix "Try again with different context:" failure due to telemetryRecorder misuse

sqs:
Instead of requiring this `telemetryRecorder` to be passed down via props, which is messy, it's now stored in React context.

## Test plan

CI


eseliger (approved):
wow that's more complex than I expected

PR #4349, opened 2024-05-29
Agent: return more details in JSON-RPC error responses

olafurpg:
The PR #3807 introduced a regression in how we deal with rate limits. Before that PR, we returned a dedicated JSON-RPC error code on rate limit errors, and this code was no longer returning after the PR. This PR recovers the original logic and integrates it with the new vscode-jsonrpc bindings.

## Test plan

Not tested yet.


mkondratek (approved):
lgtm

PR #4351, opened 2024-05-29
Initial openDocument implementation

pkukielka:
The vscode.URI contains a protocol part, which if set to `unknown` indicates that it's a "virtual file" that doesn't exist on disk. Before the `untitledDocuments` signaled that this only contained files with a `unknown` protocol, which in some conditions is not true leading to that the client and server could be talking about the same file but with different protocols.

Now by removing the `untitledDocuments` we force the check of the protocol to determine if the file is equal. This also makes it easier to ensure that we always have the right content available when opening a file.

## Test plan
- Added desync panic for detecting workspace URI changes. Needs to be enabled on the Client still
- covered by existing unittests


agent/src/AgentWorkspaceDocuments.ts
```diff
@@ -191,26 +190,12 @@ export class AgentWorkspaceDocuments implements vscode_shim.WorkspaceDocuments {
...
-        if (!this.agentDocuments.has(document.underlying.uri)) {
-            if (uri.scheme === 'untitled') {
-                document.underlying.content = ''
-            } else if (!(await doesFileExist(uri))) {
-                logError(
-                    'AgentWorkspaceDocuments.openTextDocument()',
-                    'File does not exist',
-                    uri.toString()
-                )
-            } else if (uri.scheme === 'file') {
```
olafurpg, line 205: We should add a client capability for open document, and fallback to this `fspromises.readFile` solution if the client doesn't support opening files. We have other clients that may not support `openDocument` for a while.

There are places like in symf where the agent server needs to be able to open unopened files.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -246,7 +246,7 @@ export type ServerRequests = {
     'window/showMessage': [ShowWindowMessageParams, string | null]
 
     'textDocument/edit': [TextDocumentEditParams, boolean]
-    'textDocument/openUntitledDocument': [UntitledTextDocument, boolean]
+    'textDocument/openDocument': [{ uri: string }, ProtocolTextDocument]
```
olafurpg, line 249: I lean towards not returning the actual document here. Instead, the contract of this requests should be: the client returns true if it sent a `textDocument/didOpen` notification, and false otherwise. The server should read the document properties from `textDocument/didOpen` instead of the return value here

```suggestion
    'textDocument/openDocument': [{ uri: string }, boolean]
```

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -434,7 +434,6 @@ interface ClientCapabilities {
     progressBars?: 'none' | 'enabled' | undefined | null
     edit?: 'none' | 'enabled' | undefined | null
     editWorkspace?: 'none' | 'enabled' | undefined | null
-    untitledDocuments?: 'none' | 'enabled' | undefined | null
```
olafurpg, line 437: We need to add a client capability for `openDocument`

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -571,6 +570,8 @@ export interface ProtocolTextDocument {
         | {
               selectedText?: string | undefined | null
               sourceOfTruthDocument?: ProtocolTextDocument | undefined | null
+              workspaceDocumentURIs?: string[] | undefined | null
+              activeWorkspaceDocumentURI?: string | undefined | null
```
olafurpg, line 574: It might be clearer to add a comment explaining the contract, even if "active document" is a variable name in the codebase
```suggestion
              // The document that the user is currently focusing on, which is defined by the latest `textDocument/{didOpen,didChange,didFocus}` event.
              activeWorkspaceDocumentURI?: string | undefined | null
```

agent/src/AgentWorkspaceDocuments.ts
RXminuS: Ah I think that might be the root cause to the symf related test diff

vscode/src/jsonrpc/agent-protocol.ts
RXminuS: The issue is that we then in agent have code paths that don't have document data yet right after opening. That was causing the desync

olafurpg: The contract is that the client must have sent the notification before returning true. To implement `vscode.workspace.openTextDocument`, you need to await on the `true` response and then read from `AgentWorkspaceDocument`.

agent/src/AgentWorkspaceDocuments.ts
pkukielka: `openDocument` does not only read document, but **also** shows it in the editor.
Making it optional as capability is illusory and my opinion will lead to hard to debug issues.
VSCode API makes strong assumption what `openTextDocument` does, e.g.: `Opens a document. Will return early if this document is already open. Otherwise the document is loaded and the didOpen-event fires.`.
And so do we.

olafurpg: `vscode.workspace.openTextDocument` does not show the document to the user, or even open a visible editor tab. You should use the `vscode.window.showTextDocument()` method if you want to show the document or make it visible. The open method only allows you to read text contents and perform other `vscode.TextDocument` operations.

vscode/src/jsonrpc/agent-protocol.ts
pkukielka: What would be the reason to implement it that way?
VSCode contract as I read it does not really guarantee you that `didOpen` will shoot before you return:
`Opens a document. Will return early if this document is already open. Otherwise the document is loaded and the didOpen-event fires.`
It just guarantees it will fire.

I think making signature of this method similar to:
`export function openTextDocument(uri: Uri): Thenable<TextDocument>;`
makes sense.
I'm open to discussion, but I'm not sure I see the benefits of the proposed change Olaf.

olafurpg: The VSC `didOpen` notification has a different contract from agent-protocol `didOpen`. 

> I'm open to discussion, but I'm not sure I see the benefits of the proposed change Olaf.

The motivation to have a single source of truth on document synchronization. Allowing `openDocument` to return text contents is an invitation to deal with desync issues. It's easier to reason about the system if all document synchronization happens via `textDocument/{didOpen,didChange,didClose,didFocus}` notifications.

agent/src/AgentWorkspaceDocuments.ts
pkukielka: Hm, that is good to know.
I'm then confused by the fact that it emits `onDidOpenTextDocument` because: 

```
An event that is emitted when a text document is opened or when the language id of a text document has been changed.
To add an event listener when a visible text document is opened, use the TextEditor events in the window namespace. Note that:
The event is emitted before the document is updated in the active text editor
When a text document is already open (e.g.: open in another visible text editor) this event is not emitted.
```

Two last lines would suggest to me it makes is visible in the text editor, but maybe I read too much to it.

olafurpg (approved):
Stamping since we discussed this offline

vscode/src/jsonrpc/agent-protocol.ts
pkukielka: Done

vscode/src/jsonrpc/agent-protocol.ts
pkukielka: Changes as we discussed.

vscode/src/non-stop/FixupController.ts
```diff
@@ -996,9 +996,30 @@ export class FixupController
...
+
+        //check if we need to insert a blank line
+        const beforeLastLine =
+            lastLine.lineNumber > 0 ? doc.lineAt(Math.max(doc.lineCount - 2, 0)) : undefined
+
+        const isBlankInsertionPoint =
+            lastLine.text.endsWith('\n') ||
+            (lastLine.text.trim().length === 0 && beforeLastLine?.text.endsWith('\n'))
+
+        if (!isBlankInsertionPoint) {
```
abeatrix, line 1014: Nice! This is not limited to the Test command but apply to all Edit command, right?  If yes, what do you think about moving this to somewhere all the edits would reach and not just Test command? Maybe `didReceiveFixupText()` or `textDidChange()`?

vscode/src/non-stop/FixupController.ts
```diff
@@ -996,9 +996,30 @@ export class FixupController
 
         // append response to new file
         const doc = await vscode.workspace.openTextDocument(newFileUri)
-        const pos = new vscode.Position(Math.max(doc.lineCount - 1, 0), 0)
-        const range = new vscode.Range(pos, pos)
+        const lastLine = doc.lineAt(Math.max(doc.lineCount - 1, 0))
+
+        const EOF = lastLine.rangeIncludingLineBreak.end
+
+        const range = new vscode.Range(EOF, EOF)
```
abeatrix, line 1003:
```suggestion
        task.selectionRange = new vscode.Range(EOF, EOF)
```

If we are resetting task.selectionRange below, we won't need `range` anymore?

vscode/src/non-stop/FixupController.ts
```diff
@@ -996,9 +996,30 @@ export class FixupController
...
+            lastLine.text.endsWith('\n') ||
+            (lastLine.text.trim().length === 0 && beforeLastLine?.text.endsWith('\n'))
+
+        if (!isBlankInsertionPoint) {
+            const edit = new vscode.WorkspaceEdit()
+            edit.insert(newFileUri, EOF, '\n' /* check if it needs to be doc.eol */)
+            const inserted = await vscode.workspace.applyEdit(edit)
+            if (inserted) {
+                const blankEOF = EOF.with(undefined, 0).translate(1)
+                task.selectionRange = new vscode.Range(blankEOF, blankEOF)
```
abeatrix, line 1020: Do we need to update the "range" to use `task.selectionRange` instead on line 1031?

vscode/src/non-stop/FixupController.ts
```diff
@@ -996,9 +996,30 @@ export class FixupController
 
         // append response to new file
         const doc = await vscode.workspace.openTextDocument(newFileUri)
-        const pos = new vscode.Position(Math.max(doc.lineCount - 1, 0), 0)
-        const range = new vscode.Range(pos, pos)
+        const lastLine = doc.lineAt(Math.max(doc.lineCount - 1, 0))
```
abeatrix, line 999:
```suggestion
        const lastLine = doc.lineAt(doc.lineCount - 1)
```
doc.lineAt will returns `TextLine` which includes`isEmptyOrWhitespace`, so we probably don't need to use `Math.max` here.

vscode/src/non-stop/FixupController.ts
```diff
@@ -996,9 +996,30 @@ export class FixupController
...
+
+        const EOF = lastLine.rangeIncludingLineBreak.end
+
+        const range = new vscode.Range(EOF, EOF)
         task.selectionRange = range
+
+        //check if we need to insert a blank line
+        const beforeLastLine =
+            lastLine.lineNumber > 0 ? doc.lineAt(Math.max(doc.lineCount - 2, 0)) : undefined
```
abeatrix, line 1008:
```suggestion
        const beforeLastLine = doc.lineCount > 1 ? doc.lineAt(doc.lineCount - 2) : undefined
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -996,9 +996,30 @@ export class FixupController
...
+
+        const isBlankInsertionPoint =
+            lastLine.text.endsWith('\n') ||
+            (lastLine.text.trim().length === 0 && beforeLastLine?.text.endsWith('\n'))
```
abeatrix, line 1012:
```suggestion
        const isBlankInsertionPoint =
            lastLine.isEmptyOrWhitespace || (beforeLastLine?.isEmptyOrWhitespace ?? false)
```

nit: Replace with built-in methods.

pkukielka:
Replaced by https://github.com/sourcegraph/cody/pull/4841

PR #4352, opened 2024-05-29
Agent: remove redundant logs

mkondratek:
## Test Plan

Green CI


olafurpg (approved):
LGTM 👍🏻 For future PRs, please use the "Agent:" prefix in the title, and include a more detailed description. Example: previously, we were logging errors. This was problematic because ... Now, we no longer log these errors. It's OK to make this change because ...

PR #4353, opened 2024-05-29
Document code: Add tree-sitter query for Java

umpox:
image

## Description

This PR adds tree-sitter support for `getDocumentableNode` in Java.

This means we get:
- Improved range expansion

## Test plan

1. Open Java code
2. Move cursor to a position on a documentable node, e.g. a method name
3. Document code, expect documentation is generated correctly (above the method)
4. Move cursor to a position within a documentable node, e.g. within a method
5. Document code, expect documentation is generated correctly (above the method)

### DIFF
--- a/vscode/src/tree-sitter/queries/java.ts
+++ b/vscode/src/tree-sitter/queries/java.ts
@@ -7,10 +7,19 @@ const DOCUMENTABLE_NODES = dedent`
     ;--------------------------------
     (method_declaration
         name: (identifier) @symbol.function) @range.function
+    (enum_body_declarations
+        (method_declaration
+            name: (identifier) @symbol.function) @range.function)
     (constructor_declaration
         name: (identifier) @symbol.function) @range.function
+    (enum_body_declarations
+        (constructor_declaration
+            name: (identifier) @symbol.function) @range.function)
     (record_declaration
         name: (identifier) @symbol.function) @range.function
+    (enum_body_declarations
+        (record_declaration
+            name: (identifier) @symbol.function) @range.function)
 
     ; Variables
     ;--------------------------------
@@ -18,6 +27,9 @@ const DOCUMENTABLE_NODES = dedent`
         name: (_) @symbol.identifier) @range.identifier
     (variable_declarator
         name: (identifier) @symbol.identifier) @range.identifier
+    (field_declaration
+        (variable_declarator
+            name: (identifier) @symbol.identifier) @range.identifier)
 
     ; Type Identifiers
     ;--------------------------------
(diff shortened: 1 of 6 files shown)


vscode/src/tree-sitter/query-tests/test-data/documentable-node.java
```diff
@@ -0,0 +1,76 @@
...
+
+public enum Day {
+//           |
+    SUNDAY,
+    MONDAY,
+    TUESDAY,
+    WEDNESDAY,
+    THURSDAY,
+    FRIDAY,
+    SATURDAY
```
umpox, line 75: Note: I haven't made enum members documentable here. Happy to add them if anyone thinks that's useful

vscode/src/tree-sitter/queries/java.ts
```diff
@@ -0,0 +1,39 @@
...
+`
+
+export const javaQueries = {
+    [SupportedLanguage.java]: {
+        documentableNodes: DOCUMENTABLE_NODES,
+        singlelineTriggers: '',
+        intents: '',
+        identifiers: '',
+        graphContextIdentifiers: '',
+        enclosingFunction: '',
```
umpox, line 37: @valerybugakov We should consider adding autocomplete queries here too

vscode/src/tree-sitter/query-tests/test-data/documentable-node.java
olafurpg: I think it's OK to make them documentable. Java enums are a lot more powerful than TS enums, they are just regular classes and each enum member can be initialized with custom parameters.

pkukielka (approved):
LGTM

vscode/src/tree-sitter/query-tests/test-data/documentable-node.java
umpox: Oh wow yeah they can be pretty involved... I had to update the query in a few places, added some more tests. Would appreciate your eyes again!

olafurpg:
I tried this locally and "Document Code" is still not going to the enclosing *method declaration*, it's documenting local variables. Any idea what might be going on @umpox ? To reproduce https://sourcegraph.com/github.com/redisson/redisson@68f6829e7ab22c49840e17825e43f944956f9785/-/blob/redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java?L894

https://github.com/sourcegraph/cody/assets/1408093/17d66a42-efa1-42d5-8a33-b87b50150d77

umpox:
@olafurpg That's intentional, as it matches the same behaviour for this command in other languages.

E.g. if the user wants to generate docs for a local variable in a long function

image

We do have an `enclosingFunction` query, but that is used for the edit command specifically (where you will rarely want to expand to a local variable). It's not added by this PR but we can add that as a follow up.

olafurpg:
Sounds good. Worth noting that @steveyegge 's Loom showed a demo where documenting a local block would document the enclosing method. We need a bigger discussion on what the desired behavior should be.

This PR LGTM because it makes Java work more similarly to other languages. It would be nice to add more coverage for documentable ranges in languages that see heavier usage in JetBrains, but it can be done separately.

umpox:
@olafurpg Agree, I think @abeatrix is looking at doing the same for Kotlin today. I can follow up on this for `enclosingFunction` and then I'll look into how much effort it'll be to add PHP support too

PR #4354, opened 2024-05-29
Agent: add support for code actions and "Ask Cody to Fix" command

olafurpg:
Fixes CODY-1932

Previously, there was no way for agent clients to use the "Ask Cody to Fix" command. This PR fixes this problem by adding support for code actions. Code actions are the "quickfix" menu when you trigger option-enter over a diagnostic in the editor. With the new API, an agent client can open the same quickfix menu to trigger the "Ask Cody to fix" command.

## Test plan

See newly added test suite for Fix.


vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
```diff
@@ -538,3 +606,20 @@ enum SearchState {
     Done = 1,
     Continue = 2,
 }
+
+function formatMessageText(messageText: string | ts.DiagnosticMessageChain): string {
+    if (typeof messageText === 'string') {
+        return messageText
+    }
+    const messages: string[] = []
+    const loop = (chain: ts.DiagnosticMessageChain): void => {
```
odisseus, line 615: I don't know how likely it is in the real circumstances, but this function could possibly cause a stack overflow.

olafurpg: That’s fair. We only use this function for testing at the moment so it shouldn’t be concern for production. There’s nothing preventing us from starting to use this from non-test code in the future so it’s worth fixing.

agent/src/agent.ts
```diff
@@ -509,20 +636,21 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
-                        const distance = levenshtein(postData, entryPostData)
-                        if (distance < closestDistance) {
-                            closest = entryPostData
-                            closestDistance = distance
-                        }
+            if (!polly) {
+                throw new Error('testing/closestPostData: Polly is not enabled')
+            }
+            // @ts-ignore
+            const persister = polly.persister._cache as Map<string, Promise<Har>>
```
olafurpg, line 643: This hunk is unrelated to the Fix command. This diff fixes a bug in the test-only logic that prints out the closest JSON request in the HTTP recordings when there is no matching recording. The CI is failing but I can't reproduce locally, it's usually a hint that the prompt is unstable on macOS/Linux

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
olafurpg: Fixed, thanks!

olafurpg:
I'm able to reproduce the failing test locally. You just need to run the test in replay mode multiple times. The prompt is unstable. Here is the diff:
```
-      "text": "This is part of the file: src/example-error.ts\n\nThe user has the following code in their selection:\n<SELECTEDCODE7662>export function fixCommandExample(): number {\n    return '42';\n}\n</SELECTEDCODE7662>\n\nThe user wants you to correct problems in their code by following their instructions.\nProvide your fixed code using the following instructions:\n<DIAGNOSTICS5668>\n<PROBLEMCODE4179>export function fixCommandExample(): number {\n    return '42';\n}\n</PROBLEMCODE4179>\n\nFix the following error from within <PROBLEMCODE4179></PROBLEMCODE4179>: Type 'string' is not assignable to type 'number'.\n</DIAGNOSTICS5668>"
+      "text": "This is part of the file: src/example-error.ts\n\nThe user has the following code in their selection:\n<SELECTEDCODE7662>export function fixCommandExample(): number {\n    return '42';\n}</SELECTEDCODE7662>\n\nThe user wants you to correct problems in their code by following their instructions.\nProvide your fixed code using the following instructions:\n<DIAGNOSTICS5668>\n<PROBLEMCODE4179>export function fixCommandExample(): number {\n    return '42';\n}\n</PROBLEMCODE4179>\n\nFix the following error from within <PROBLEMCODE4179></PROBLEMCODE4179>: Type 'string' is not assignable to type 'number'.\n</DIAGNOSTICS5668>"
```
Most likely, it's caused by the async/background nature of tree-sitter parsing. I suspect calling `testing/awaitPendingPromises` everywhere will fix the problem, but it would be nice to do this automatically (fine if only for tests).
 https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/jsonrpc/agent-protocol.ts?L154

Should be an easy fix. I just verified that commenting out the listener for `vscode.window.onDidChangeVisibleTextEditors` makes the failure deterministic


FYI @valerybugakov @umpox I am curious how much this non-deterministic behavior is causing problems in production. It might not be a problem because humans are slow to type/open documents. However, if parsing is so fast anyways, maybe we should just wait for parsing to complete instead of reading a stale AST from cache 🤔

olafurpg:
The diff in the prompt when tree-sitter finishes parsing before/after triggering code actions, it's just a newline

CleanShot 2024-05-29 at 22 47 16@2x

olafurpg:
Should be fixed by the latest commit, which enables awaiting on listeners from `onDidChangeVisibleTextEditors` and `onDidChangeTextDocument`. This makes the test pass deterministically for me locally.

vscode/src/completions/context/retrievers/tsc/tsc-retriever.ts
odisseus: I was actually thinking about an extremely long chain of nested `DiagnosticMessageChain`. I was also going to suggest tail recursion, but it turns out that Node.js doesn't support it anyway.

valerybugakov:
@olafurpg, I agree with the tree-sitter proposal to expose an API to wait for recent edits to be applied to parse trees. Created an issue here: https://linear.app/sourcegraph/issue/CODY-1947/expose-an-api-to-wait-for-the-recent-document-edits-to-be-parsed-by

agent/src/agent.ts
```diff
@@ -481,6 +500,113 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+        const codeActionById = new Map<string, vscode.CodeAction>()
+        this.registerAuthenticatedRequest('codeActions/provide', async (params, token) => {
+            codeActionById.clear()
+            const document = this.workspace.getDocument(vscode.Uri.parse(params.location.uri))
+            if (!document) {
+                throw new Error(`codeActions/provide: document not found for ${params.location.uri}`)
+            }
+            const codeActions: agent_protocol.ProtocolCodeAction[] = []
+            for (const providers of this.codeAction.providers()) {
+                const diagnostics = vscode.languages.getDiagnostics(document.uri)
```
valerybugakov, line 513: nit: can we collect diagnostics once out of the loop?

```suggestion
            const codeActions: agent_protocol.ProtocolCodeAction[] = []
            const diagnostics = vscode.languages.getDiagnostics(document.uri)
            for (const providers of this.codeAction.providers()) {
```

valerybugakov (approved):
That's awesome. I see the Fix command eval on the horizon!

agent/src/AgentDiagnostics.ts
```diff
@@ -0,0 +1,13 @@
+import type * as vscode from 'vscode'
+
+export class AgentDiagnostics {
+    private diagnostics = new Map<string, vscode.Diagnostic[]>()
```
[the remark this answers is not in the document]
olafurpg: We only keep active diagnostics, the same contract LSP has. When you publish diagnostics twice for the same document, we throw away the old diagnostics.

agent/src/agent.ts
olafurpg: Good one! Fixed

agent/src/AgentDiagnostics.ts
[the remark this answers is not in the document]
valerybugakov: I was worrying about the case where the user opens new documents for a week or longer. Since we do not have limit on the number of documents we keep the memory will continuously grow.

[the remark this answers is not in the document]
olafurpg: I wouldn't classify that as a memory leak. In this case, we need to store all the diagnostics to meet the contract of the IDE. I regularly close all the open tabs in my IDE to clear unwanted diagnostics.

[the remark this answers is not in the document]
olafurpg: Worst case, we could store these in disc, but it's not a concern IMO. It's also still unclear whether we will use this much. It will be tempting to just accept the diagnostics in `codeActions/provide`

PR #4355, opened 2024-05-29
Document code: Add tree-sitter query for Kotlin

abeatrix:
Description

This PR adds tree-sitter support for getDocumentableNode in Kotlin.

This means we get:

- Improved range expansion

## Test plan

Copied from https://github.com/sourcegraph/cody/pull/4353

- Open Kotlin code
- Move cursor to a position on a documentable node, e.g. a method name
- Document code, expect documentation is generated correctly (above the method)
- Move cursor to a position within a documentable node, e.g. within a method
- Document code, expect documentation is generated correctly (above the method)

### DIFF
--- a/vscode/src/tree-sitter/queries/kotlin.ts
+++ b/vscode/src/tree-sitter/queries/kotlin.ts
@@ -3,35 +3,21 @@ import dedent from 'dedent'
 import { SupportedLanguage } from '../grammars'
 import type { QueryName } from '../queries'
 
-const SINGLE_LINE_TRIGGERS = dedent`
-    (function_declaration (simple_identifier) @symbol.function) @trigger
-`
-
 const DOCUMENTABLE_NODES = dedent`
     ; Function definitions
     ;--------------------------------
-    (function_declaration (function_body) @symbol.function) @range.function
-    ((function_declaration (simple_identifier) @symbol.function)
-       (parameter (simple_identifier) @symbol.function)
-       (class_parameter (simple_identifier) @symbol.function)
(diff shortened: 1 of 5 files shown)


vscode/src/tree-sitter/queries/kotlin.ts
```diff
@@ -0,0 +1,52 @@
...
+       (parameter (simple_identifier) @symbol.function)
+       (class_parameter (simple_identifier) @symbol.function)
+       (variable_declaration (simple_identifier) @symbol.function)) @range.function
+
+    ; Variables
+    ;--------------------------------
+    (property_declaration
+        (simple_identifier) @symbol.identifier) @range.identifier
+    (variable_declaration
+        (simple_identifier) @symbol.identifier) @range.identifier
```
umpox, line 23: @abeatrix @steveyegge Let's remember to remove these, apart from class declarations, as we just did in this PR https://github.com/sourcegraph/cody/pull/4357

umpox:
@abeatrix I updated this so we focus on functions/class declarations/interface declarations/enum declarations

We don't focus on local variable declarations anymore, I chatted with @steveyegge and we don't want to do this for Java/Kotlin (updated for java [here](https://github.com/sourcegraph/cody/pull/4357))

Also I removed the other queries except for `ENCLOSING_FUNCTION_QUERY`. That one is fairly safe but the other ones affect autocomplete and it might be hard to test this well so close to the release, we can add those post-release though!

umpox:
@steveyegge Can you take a look?

steveyegge (approved):

PR #4358, opened 2024-05-29
Document code: Add tree-sitter query for Rust

abeatrix:
Copied from https://github.com/sourcegraph/cody/pull/4355

This PR adds tree-sitter support for getDocumentableNode in Rust.

This means we get:

- Improved range expansion

## Test plan

- Open Rust code
- Move cursor to a position on a documentable node, e.g. a method name
- Document code, expect documentation is generated correctly (above the method)
- Move cursor to a position within a documentable node, e.g. within a method
- Document code, expect documentation is generated correctly (above the method)


pkukielka (approved):
LGTM

PR #4361, opened 2024-05-30
vscode: Fixup empty input detection

eseliger:
The current check wasn't exhaustive enough to detect "there is no input". It only checked that no nodes were created ever. But focusing it adds a paragraphNode that is then empty.

I found this pretty good answer on stackoverflow: https://stackoverflow.com/a/75498145 and there's a nice utility method in the @lexical/text package that can check for empty text content, _after_ trimming.

Test plan:

Can no longer submit empty inputs.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -366,6 +366,9 @@ importers:
       '@lexical/react':
         specifier: ^0.15.0
         version: 0.15.0(react-dom@18.2.0)(react@18.2.0)(yjs@13.6.15)
+      '@lexical/text':
+        specifier: ^0.15.0
+        version: 0.15.0
       '@mdi/js':
         specifier: ^7.2.96
         version: 7.2.96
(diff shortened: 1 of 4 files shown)


sqs (approved):
wow I was hitting this bug last night, thanks!

sqs:
@eseliger The `@testing-library/react` tests that use jsdom and lexical editor are kinda flaky. I see:

```
 FAIL  |vscode| webviews/chat/Transcript.test.tsx > Transcript > focus
Error: expect(element).toBeEnabled()

Received element is not enabled:
  <button
  aria-label="Send with automatic code context"
```

These tests are starting to hit issues where jsdom only partially implements the DOM, and so I don't think they accurately test Lexical editor state. You can remove the tests if you think they are erroneous and can't easily fix them.

eseliger:
just fixed it by forcing a rerender, I hope that makes sense as a fix - but if you think that's just patching over something I can also remove it

PR #4307, opened 2024-05-27
OpenCtx: allow spaces in mentions

valerybugakov:
Allow spaces in @-mentions

## Test plan

Tested manually

PR #4363, opened 2024-05-30
Edit: fix @-mentions in the quick-pick menu

valerybugakov:
- Fixes a regression introduced in https://github.com/sourcegraph/cody/pull/4307, which does not allow the use of @-mentions in the Edit code quick-pick menu.
- This PR changes the @-mentions regex to allow spaces in the chat input and disallow them in the quick-pick menu.

## Test plan

Updated unit tests and verified locally.

### DIFF
--- a/vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
+++ b/vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
@@ -65,6 +65,15 @@ const FLOATING_OPTIONS: UseFloatingOptions = {
     transform: false,
 }
 
+/**
+ * We allow whitespace for @-mentions in the Lexical editor because
+ * it has an explicit switch between modes and can render @-mentions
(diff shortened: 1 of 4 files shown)


vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
```diff
@@ -65,6 +65,10 @@ const FLOATING_OPTIONS: UseFloatingOptions = {
     transform: false,
 }
 
+function scanForMentionTriggerInLexicalInput(text: string) {
```
sqs, line 68: explain why the desired behavior is different for lexical editor in chat vs. quickpick in edit

sqs (approved):

thenamankumar (approved):
Looks good. I wondered if we could have a better UI there but seems like the quick pick is native VS Code thing.

PR #4364, opened 2024-05-30
"Initial context" @-mentions in chat input

sqs:
https://linear.app/sourcegraph/issue/CODY-1896/initial-context-mentions-in-chat-input


toolmantim (approved):
Approving so we can get a pre-release out for wider testing

PR #4365, opened 2024-05-30
vscode: Fixup empty section in command menu

eseliger:
The command menu had an empty section at the end, because the emptyelement doesn't render when any providers are still shown. We thus created an empty CommandGroup.

This restructures it a bit to 
- always render the empty element and the loading element, as they won't show when there's any content
- to only render the second command group when there's actually gonna be entries in it

See the empty section at the end here:

Screenshot 2024-05-30 at 12 15 03@2x

Test plan:

No more empty section at the end.


toolmantim (approved):
Thank you!

PR #4366, opened 2024-05-30
removing the default "codebase context from file <file_path>" line for openctx providers

hitesh-1997:
## Context
By default the context from the openctx providers are appended with the line "Codebase context from file" which is the default template. Adding a check for the openctx providers and only adding the content from the openctx providers.

## Test plan
Manually checked
1. Added a debugger point in the local code to ensure openctx line does not have the default line.

image


valerybugakov (approved):
👍

thenamankumar:
Can you make sure this doesn't break the experience for the Web and Sourcegraph Respositories providers?

PR #4367, opened 2024-05-30
Apply context filters to Sourcegraph Repositories @-mention context p…

thenamankumar:
Issue: https://linear.app/sourcegraph/issue/CODY-1944/with-cody-context-filters-enabled-an-ignored-repo-should-not

This PR Integrates context filters with Sourcegraph Repositories @-mention context provider. 

No repo which is configured to be excluded from Cody using the context filters config present in site config should appear in the list while selecting a repo. 

Closes CODY-1944

## Test plan

Demo: https://www.loom.com/share/9ff608d9c4364bc5a03b0b030caa866f?sid=6c135b32-b1ed-42b9-8e2b-c99f79af2f10

### DIFF
--- a/vscode/src/context/openctx/remoteRepositorySearch.ts
+++ b/vscode/src/context/openctx/remoteRepositorySearch.ts
@@ -1,4 +1,4 @@
-import { contextFiltersProvider, graphqlClient, isDefined, isError } from '@sourcegraph/cody-shared'
+import { contextFiltersProvider, graphqlClient, isError } from '@sourcegraph/cody-shared'
 
 import type { Item, Provider } from '@openctx/client'
 
@@ -21,19 +21,14 @@ const RemoteRepositorySearch: Provider & {
 
             const repositories = dataOrError.repositories.nodes
 
-            return repositories
-                .map(repo =>
-                    contextFiltersProvider.isRepoNameIgnored(repo.name)
-                        ? null
-                        : {
-                              uri: repo.url,
-                              title: repo.name,
-                              data: {
-                                  repoId: repo.id,
-                              },
-                          }
-                )
-                .filter(isDefined)
+            return repositories.map(repo => ({
+                uri: repo.url,
+                title: repo.name,
+                data: {
+                    repoId: repo.id,
+                    isIgnored: contextFiltersProvider.isRepoNameIgnored(repo.name),
+                },
+            }))
         } catch (error) {
             return []
         }
--- a/vscode/src/context/openctx/remoteRepositorySearch.ts
+++ b/vscode/src/context/openctx/remoteRepositorySearch.ts
@@ -1,6 +1,6 @@
-import { graphqlClient, isError } from '@sourcegraph/cody-shared'
+import { contextFiltersProvider, graphqlClient, isDefined, isError } from '@sourcegraph/cody-shared'
 
-import type { Item, Mention, Provider } from '@openctx/client'
+import type { Item, Provider } from '@openctx/client'
 
 const RemoteRepositorySearch: Provider & {
     providerUri: string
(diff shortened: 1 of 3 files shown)


sqs (approved):

vscode/src/context/openctx/remoteRepositorySearch.ts
```diff
@@ -13,24 +13,27 @@ const RemoteRepositorySearch: Provider & {
...
+                        ? null
+                        : {
+                              uri: repo.url,
+                              title: repo.name,
+                              data: {
+                                  repoId: repo.id,
+                              },
+                          }
+                )
+                .filter(isDefined)
```
eseliger, line 36: would this read nicer if we did:

```suggestion
            return repositories
                .filter(repo => contextFiltersProvider.isRepoNameIgnored(repo.name))
                .map(repo =>
                    ({
                              uri: repo.url,
                              title: repo.name,
                              data: {
                                  repoId: repo.id,
                              },
                          })
                )
```

eseliger (approved):

kalanchan:
WAIT don't merge yet, i'm leaving a comment one sec

vscode/src/context/openctx/remoteRepositorySearch.ts
kalanchan, line 36: I don't think we should filter these out, instead we should still list them but have a visual indicator that the repo is not allowed because of context filters set by sourcegraph admins. 

In the original design, the multi repo picker still lists them, but has a indicator stating it will be ignored. Afterwards, if the user tries to hit OK an error notification pops up in the bottom right hand corner saying that it can't be used

CleanShot 2024-05-29 at 22 48 54@2x

kalanchan (changes requested):

vscode/src/context/openctx/remoteRepositorySearch.ts
kalanchan: maybe also loop in @toolmantim and @danielmarquespt for their thoughts on whether this design still makes sense in the new UI

cc @aramaraju

thenamankumar: Well, the issue said not to show them, but I discussed it with @toolmantim, and now I will update it so we show the repos in the list but turn off the selection.

kalanchan: heh woops, I left it as a [comment](https://linear.app/sourcegraph/issue/CODY-1896/initial-context-mentions-in-chat-input#comment-49f7bdd1) in the original ticket, it just got lost in translation :)

thenamankumar: CleanShot 2024-05-30 at 14 33 54@2x

thenamankumar: coming in hot with approvals from @toolmantim

thenamankumar: not filtering out now

toolmantim (approved):

vscode/src/context/openctx/remoteRepositorySearch.ts
```diff
@@ -13,24 +13,22 @@ const RemoteRepositorySearch: Provider & {
 
     async mentions({ query }) {
         try {
-            const dataOrError = await graphqlClient.searchRepos(10, undefined, query)
+            const dataOrError = await graphqlClient.searchRepos(30, undefined, query)
```
valerybugakov, line 16: Do you think we want to search for 30 items here? It will be slower.

valerybugakov (approved):

vscode/src/context/openctx/remoteRepositorySearch.ts
kalanchan: not a bad idea to limit it a bit

kalanchan (approved):
looks great! thanks so much for adding this in for Cody Context GA! 

gogogo noodle hacker gang! 

CleanShot 2024-05-29 at 23 47 01

PR #4370, opened 2024-05-30
Import OpenCtx Web provider and remove the old ones

thenamankumar:
As the OpenCtx Web provider was enabled by default on the pre-release, we can not just ship it that way and it doesn't belong to the vscode package and all the SOC2 compliance stuff....

This PR imports the Web provider from OpenCtx (as it is) and use it as internal providers similar to Remote Repo & File search providers. 

~This PR also removed the pre-openctx era providers namely github, sg search, URL fetcher. Still keeping the packages one as it has not been migrated to openctx yet and is used by "usage example" command.~

This PR also removes the default value of `openctx.providers` to include the web provider from npm package, because now it is included as an internal provider. 

Closes CODY-1945

## Test plan

Tested manually.

### DIFF
--- a/vscode/src/context/openctx/web.ts
+++ b/vscode/src/context/openctx/web.ts
@@ -1,0 +1,118 @@
+import type { ItemsParams, ItemsResult, Provider } from '@openctx/client'
+
+/**
+ * An OpenCtx provider that fetches the content of a URL and provides it as an item.
+ */
+const WebProvider: Provider & { providerUri: 'internal-web-provider' } = {
+    providerUri: 'internal-web-provider',
+
+    meta() {
+        return {
+            name: 'Web URLs',
+            mentions: {},
+        }
+    },
+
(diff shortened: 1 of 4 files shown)


sqs (approved):

eseliger (approved):

vscode/src/context/openctx/web.ts
```diff
@@ -0,0 +1,118 @@
+import type { ItemsParams, ItemsResult, Provider } from '@openctx/client'
+
+/**
+ * An OpenCtx provider that fetches the content of a URL and provides it as an item.
+ */
+const WebProvider: Provider & { providerUri: 'internal-web-provider' } = {
```
keegancsmith, line 6: @thenamankumar out of interest, why can't we include provider-web as a dependency and then consume the code like that, instead of copy paste?

PR #4383, opened 2024-05-30
vscode: Relax provider search matching

eseliger:
This PR makes it so that all of

- github
- " GitHub"
- issues
- githubissues
- "GitHub Issues"

match `GitHub Issues`.

Previously, any excessive spaces in the query would cause all providers to not match, and you always had to put in the space. Without requiring all the spaces (but still okay to put them in) it should feel a bit more like a fuzzy finder.

Test plan:

Tim!


toolmantim (approved):
Didn't hurt

PR #4384, opened 2024-05-30
Cody Bench: add infrastructure for LLM Judge

olafurpg:
Previously, we didn't have a good way to automatically measure quality of a generated LLM response. As we are working on improving the quality of the Fix command, this limitation was painful because we could only use rule-based logic like "the diagnostic was fixed" that suffered from both false positives and false negatives:

* False positives: sometimes, the LLM came up with a low-quality fix that commented out a bunch of code, which technically fixed the error but was a low-quality result.
* False negatives: sometimes, the generated fix was perfect, but it the generated code was missing an import (minor issue).

Using an LLM judge, we hope to address this problem. The goal is not to rely 100% on LLM judge, but rather just use it as a tool in a larget toolkit of eval algorithms.

## Test plan

See new test case for LLM Judge helper.


vscode/src/testutils/CodyPersister.ts
```diff
@@ -168,7 +168,7 @@ function postProcessResponseText(entry: HarEntry): string | undefined {
         return undefined
     }
     if (
-        !entry.request.url.includes('/.api/completions/stream?api-version=1') &&
```
olafurpg, line 171: Not strictly needed for the PR. This reduces the size of the recording file significantly because it removes streaming responses even if we don't include an explicit `api-version`

agent/src/cli/cody-bench/strategy-fix.ts
```diff
@@ -73,8 +90,14 @@ export async function evaluateFixStrategy(
             if (isFixed) {
                 fixedErrors += 1
             }
+            scores.push(score)
         }
         return undefined
     })
-    console.log({ totalErrors, fixedErrors })
+    console.log({
```
olafurpg, line 97: For now, we only use `console.log` to analyze the results of the benchmark. Down the road, we will add CSV file generation (there is already infrastructure in place for this, just need to wire it in).

valerybugakov (approved):
Let's document the ways to use `cody-bench` in the README. It would be great if we had a one-sentence description with examples of how to run it with different configurations.

PR #4386, opened 2024-05-30
Visual tweaks for context focus bar

toolmantim:
Small tweaks to the feedback and context focus bar:

* Fades feedback and context focus bar until hover
* Changes context focus bar buttons to be outline style
* Makes the feedback and context focus icons consistent colors
* Reduces vertical space required for actions + context focus

| Before | After |
| - | - |
| CleanShot 2024-05-31 at 00 00 00@2x | CleanShot 2024-05-30 at 23 56 17@2x |

## Test plan

- CI


eseliger (approved):
woho free real estate

PR #4387, opened 2024-05-30
Unit Test Detection: Improve language specific cases

RXminuS:
Fixes [CODY-41](https://linear.app/sourcegraph/issue/CODY-41/)

We were quite lenient in how we detected associated test files as well as asserting that something was a test file.

Now we restrict the tests slightly depending on the file extension. Also fixes a bug where certain extensions weren't properly detected.

## Test plan
- Expanded unit tests
- Added more test cases for common language / framework defaults


vscode/src/commands/utils/test-commands.ts
```diff
@@ -39,20 +48,39 @@ export function isValidTestFile(uri: URI): boolean {
...
+    // We assume that a file can never be its own testFile for the original file.
+    // Instead make sure to test if the file IS a validTestFile.
+    if (file.path === testFile.path) {
+        return false
+    }
+
     // Assume not a test file for the current file if they are in different directories
-    // and the testFile's file path do not include test(s)
+    // and the testFile's file path does not include a test dir
+    const pathRegex = /_{0,2}tests?_{0,2}/i
```
abeatrix:
```suggestion
    const pathRegex = /test/i
```
Works the same?

abeatrix (approved):
Comments in line, but the rest looks good to me!
Found an edge case where the test would fail: ['a/b/Client.py', 'test/client.py', false], but I think this is good enough and we can let the LLM to handle those edge cases. Wdyt?

vscode/src/commands/utils/test-commands.ts
```diff
@@ -39,20 +48,42 @@ export function isValidTestFile(uri: URI): boolean {
...
+    // The file extension should match as it's rare to write a test in another language.
+    const ext = uriExtname(file)
+    const testExt = uriExtname(testFile)
+    if (ext !== testExt) {
+        return false
+    }
+
+    // Finally we check if the filename without typical test keywords matches. This is pretty naiive
+    // but seems to cover quite a few test cases.
+    const filenameExcludedCharRegex = /[^a-zA-Z0-9]/g
```
abeatrix, line 78:
```suggestion
    const filenameExcludedCharRegex = /[^a-zA-Z0-9]/gi
```

vscode/src/commands/utils/test-commands.ts
```diff
@@ -39,20 +48,42 @@ export function isValidTestFile(uri: URI): boolean {
...
+
+    const fileName = Utils.basename(file).toLowerCase().replace(filenameExcludedCharRegex, '')
+    const testFileName = Utils.basename(testFile).toLowerCase().replace(filenameExcludedCharRegex, '')
+
+    const filenameExcludedPatternsRegex = /spec|tests?/g
```
abeatrix, line 83:
```suggestion
        const fileName = Utils.basename(file).replace(filenameExcludedCharRegex, '')
    const testFileName = Utils.basename(testFile).replace(filenameExcludedCharRegex, '')

    const filenameExcludedPatternsRegex = /spec|tests?/gi
```

vscode/src/commands/utils/test-commands.ts
```diff
@@ -39,20 +48,42 @@ export function isValidTestFile(uri: URI): boolean {
...
 
-    const strippedFile = fileName.replace('spec', '').replace('test', '')
-    const strippedTestFile = testFileName.replace('spec', '').replace('test', '')
+    const strippedFile = fileName.replaceAll(filenameExcludedPatternsRegex, '')
+    const strippedTestFile = testFileName.replaceAll(filenameExcludedPatternsRegex, '')
```
abeatrix, line 86:
```suggestion
        const strippedFile = fileName.toLowerCase().replaceAll(filenameExcludedPatternsRegex, '')
    const strippedTestFile = testFileName.toLowerCase().replaceAll(filenameExcludedPatternsRegex, '')

    if (fileName.toLowerCase() === testFileName.toLowerCase() && fileName !== testFileName) {
        return false
    }
```

vscode/src/commands/utils/test-commands.test.ts
```diff
@@ -91,6 +93,19 @@ describe('isTestFileForOriginal', () => {
...
+        ['/src/server/models/User.js', '/src/server/models/User.spec.js', true], // Some projects prefer keeping tests next to the files they are testing with a `.spec.js` suffix.
+        ['/project/module.py', '/tests/test_module.py', true], // This follows the convention of placing tests in a separate `tests` directory with a `test_` prefix.
+        ['/project/package/module.py', '/project/package/test_module.py', true], // pytest allows tests to be placed in the same directory as the code with a `test_` prefix.
+        ['/project/module.py', '/tests/module_tests.py', true], // Some projects using nose place tests in a separate `tests` directory with a `_tests` suffix.
+        ['/src/project/main.go', '/src/project/main_test.go', true], // The Go testing package expects test files to be in the same directory with a `_test.go` suffix.
+        ['/src/project/main.go', '/test/project/main_test.go', true], // Some projects prefer placing tests in a separate `test` directory.
+        ['/test_logic.ts', 'test_logic.ts', false],
```
abeatrix, line 108:
```suggestion
        ['/test_logic.ts', 'test_logic.ts', false],
        ['a/b/Client.py', 'test/Client.py', true], // Test file located in a `test` directory adjacent to the file being tested.
        ['a/b/test_logic.ts', 'c/d/test_logic.ts', false], // Test file located in a different directory.
        ['a/b/Client.py', 'test/client.py', false], // Test file located in a `test` directory adjacent to the file being tested.
```

abeatrix (commented):
Suggested some new changes that would address the edge case I mentioned earlier 😄 

Edit: Github won't let me select the whole function to make suggestion so they got broken down into different places. Here is the revised version of the isTestFileForOriginal function with my suggestions:
```
export function isTestFileForOriginal(file: URI, testFile: URI): boolean {
    // We assume that a file can never be its own testFile for the original file.
    // Instead make sure to test if the file IS a validTestFile.
    if (file.path === testFile.path) {
        return false
    }

    const fileDir = Utils.dirname(file)?.path
    const testDir = Utils.dirname(testFile)?.path

    // Assume not a test file for the current file if they are in different directories
    // and the testFile's file path does not include a test dir
    const pathRegex = /(?:^|\/)_{0,2}tests?_{0,2}(?:\/|$)/i
    if (fileDir !== testDir) {
        if (!pathRegex.test(testDir)) {
            return false
        }
    }

    // The file extension should match as it's rare to write a test in another language.
    const ext = uriExtname(file)
    const testExt = uriExtname(testFile)
    if (ext !== testExt) {
        return false
    }

    // Finally we check if the filename without typical test keywords matches. This is pretty naiive
    // but seems to cover quite a few test cases.
    const filenameExcludedCharRegex = /[^a-zA-Z0-9]/gi

    const fileName = Utils.basename(file).replace(filenameExcludedCharRegex, '')
    const testFileName = Utils.basename(testFile).replace(filenameExcludedCharRegex, '')

    const filenameExcludedPatternsRegex = /spec|tests?/gi

    const strippedFile = fileName.toLowerCase().replaceAll(filenameExcludedPatternsRegex, '')
    const strippedTestFile = testFileName.toLowerCase().replaceAll(filenameExcludedPatternsRegex, '')

    if (fileName.toLowerCase() === testFileName.toLowerCase() && fileName !== testFileName) {
        return false
    }

    return strippedFile === strippedTestFile
}
```

vscode/src/commands/utils/test-commands.ts
RXminuS: Yeah I forgot the path slashes :-), then the __ matters

PR #4398, opened 2024-05-31
Dev: register command to clear Cody storages

abeatrix:
Add a new command to make it easier for Devs (currently in dev and test mode only) to clear current auth states and storages for debugging purpose.

- Add a new command "cody.dev.clear" to remove the current access token, remove all stored access tokens, clear the local storage, and restart VS Code
- Change the local storage key used to suppress Cody Pro expiration notifications
- Refactor the local storage provider to use a centralized key map
- Add a clear method to the local storage provider to easily clear all stored data except the anonymous user ID
- Update the LocalStorageProvider to add a new key for the access token secret
- Update the SecretStorageProvider to use the new access token secret key from the LocalStorageProvider
- Remove the code to clear the storage and access token in the post-uninstall script, as this is now handled by the new "cody.dev.clear" command

Currently available in Dev & Test mode only.

## Test plan

1. Start Cody from this branch to make sure Chat history and Auth are still working as expected.
2. Run the new command `Cody Dev: Clear Cody Storage & Restart` to verify you have been logged out
3. Verify chat history and endpoint history are cleared after logging back in


vscode/src/main.ts
```diff
@@ -423,6 +423,24 @@ const register = async (
                         'Failed to parse context filters policy. Please check your JSON syntax.'
                     )
                 }
+            }),
+
+            // Clear access token and local storage.
+            vscode.commands.registerCommand('cody.dev.clear', async () => {
```
abeatrix, line 429: @jamesmcnamara following your works on logging uninstall event, do you know if we can still access this command when the script runs? Wondering if we should also clear user's local storage on uninstall

jamesmcnamara: No we cannot access any vscode APIs in the install script because it is run the next time VSCode restarts in a seperate node process. 

Even worse, VSCode doesn't emit a unique event when a user uninstalls an extension. I do some prep work for uninstall in the `deactivate` hook, but that runs any time we close VSCode or stop the extension etc. 

As far as I can tell, there's no way to run a vscode operation after the extension has been uninstalled :(

jamesmcnamara (approved):

PR #4399, opened 2024-05-31
remove unneccessary windows check in ContextItemMentionNode test

keegancsmith:
The true and false branch are the same value.

Follow up from https://github.com/sourcegraph/cody/pull/4393 and https://github.com/sourcegraph/cody/pull/4394

Test Plan: CI


jhchabran (approved):
@keegancsmith oh I missed it 🤣🤦

Issue #4215, opened 2024-05-19
bug: Since the latest version, I have intermitent issues with adding selection to the chat with option+l and opening chats with the same shortcut


mcheviron:
### Version

Version: 1.89.1 Commit: dc96b837cf6bb4af9cd736aa3af08cf8279f7685 Date: 2024-05-07T05:14:32.757Z Electron: 28.2.8 ElectronBuildId: 27744544 Chromium: 120.0.6099.291 Node.js: 18.18.2 V8: 12.0.267.19-electron.0 OS: Darwin arm64 23.4.0

### Describe the bug

I have an issue where sometimes I'm able to add a selection to the chat, but other times it's greyed out. This also applies to opening a new chat with the same keyboard shortcut, Option+L, on my machine. This was perfectly fine before the latest update.

![image](https://github.com/sourcegraph/cody/assets/87418909/8566c324-de32-49c0-aa21-0ebf981e8f30)
![image](https://github.com/sourcegraph/cody/assets/87418909/dab3fcd5-1cb8-4b97-b48b-5b7bf319492f)

### Expected behavior

Option+L adds selection to the chat
Option+L opens a new chat when no selection is made
Right click gives the option to add the file to the chat if no selection is there (it's also greyed out)

SphaeroX:
releated to https://github.com/sourcegraph/cody/issues/4178

Issue #4194, opened 2024-05-16
Bug: Issue with keyboard shortcut for opening chat


dicktornfeldt:
### Version

1.18.0

### Describe the bug

Im on a Swedish keyboard and the default character for opt+l is "ﬁ"

In version 1.16.7 I can select a codeblock and press opt+l and the Cody chat opens.

In version 1.18.0 the selected codeblock is replaced by the character "ﬁ"

### Expected behavior

The Cody chat should appear instead of the code block being replaced by the default character

oscarmrom:
I think these are related https://github.com/sourcegraph/cody/issues/4178

Issue #4178, opened 2024-05-15
bug: cannot add selection to chat for follow-up message


hlhdaiaaii:
### Version

v1.17.1715730510 (pre-release)

### Describe the bug

I've upgraded to latest pre-release to use gpt4o and the Add selection to chat broke, this option under Cody right click menu  is not visible for follow-up message
![image](https://github.com/sourcegraph/cody/assets/48194959/1626159b-8de2-4f56-a5e7-56dde60e4d18)

### Expected behavior

it should allow to add context to chat for follow-up message like current release version

chris-wickens:
Bug is still present in 1.18.0, keyboard shortcut is broken as well.

SphaeroX:
Also on Windows 11 v1.19.1715872539 (pre-release)

See the Video i cant hit Alt + L


https://github.com/sourcegraph/cody/assets/6785110/0118322a-cf20-4651-8162-639477df6adb

oscarmrom:
Something similar is going on v1.18.0 on MacOS VS Code where 

- No cody chat is open
- opt + L with selection would open new chat window with selection mentioned

Now it doesn't open the chat window unless there is no selection. Otherwise it replaces the text with `¬` instead of triggering the cody new chat with selection keybind that I use on a daily.

dominiccooney:
@abeatrix , could you help triage this? This could be Cody Ignore, token limits?, something else?

abeatrix:
> @abeatrix , could you help triage this? This could be Cody Ignore, token limits?, something else?

@dominiccooney thanks for the ping! I just encountered the same issue today and created a PR to fix this. It's already been merged and should be available in tomorrow's pre-release!

Here is the PR: https://github.com/sourcegraph/cody/pull/4242

hlhdaiaaii:
@abeatrix hi, I am the latest pre-release version and this bug still remains. The option Add selection to chat is now visible but does not take any effect

SphaeroX:
> @abeatrix hi, I am at the latest pre-release version and this bug still remains. The option Add selection to chat is now visible but does not take any effect, it does not mention selection to the chat input for follow-up message

I also find this error to be destructive and it should be priority number 1 as it obeys my entire workflow. Because I can no longer work with the shortcut.

hlhdaiaaii:
[...]

PR #4407, opened 2024-05-31
Unify Opt+L and Opt+K hotkeys, fix Opt+L inserting a char instead of opening chat

sqs:
There are 2 hotkeys to open Cody chat: Opt+/ and Opt+L. In most places we document Opt+L. There are slight differences (Opt+L adds your selection, Opt+/ t) that do not merit having 2 different prominent keybindings.

We will standardize on Opt+L as the hotkey and as the single behavior. Opt+L also has the benefit of being similar to Opt+K, the edit hotkey. It makes Opt+/ an alias with the same behavior (for people with muscle memory).

Also fixes issues with Opt+L inserting a character instead of opening chat on macOS in some cases.

- Fixes https://github.com/sourcegraph/cody/issues/4215
- Fixes https://github.com/sourcegraph/cody/issues/4419
- Fixes https://github.com/sourcegraph/cody/issues/4178 (again)
- Fixes https://github.com/sourcegraph/cody/issues/4483
- Fixes https://github.com/sourcegraph/cody/issues/4194
- Fixes https://linear.app/sourcegraph/issue/CODY-1985/unify-optl-and-optk-hotkeys
- Fixes https://linear.app/sourcegraph/issue/CODY-1983/duplicate-optl-keybindings
- https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1717152531711519

## Test plan

CI & e2e. Confirmed Opt+L in an editor with a selection does not input a character on macOS; it opens a chat window with the selection mentioned.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -592,9 +592,9 @@
         "when": "!cody.activated"
       },
       {
-        "command": "cody.chat.tree.view.focus",
-        "key": "alt+f1",
-        "when": "cody.activated"
+        "command": "cody.chat.focus",
+        "key": "alt+l",
+        "when": "!cody.activated"
       },
       {
         "command": "cody.chat.panel.new",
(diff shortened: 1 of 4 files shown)


abeatrix (approved):
Ty!

PR #4410, opened 2024-05-31
Make it easy to add back the mention of your current repo

sqs:


toolmantim:
What about something like this?

image

PR #4412, opened 2024-05-31
vscode: do not show "Rename Chat" in command pallette

keegancsmith:
This command only works when directly clicked on a chat item in the history. So we use the same enablement condition as when it is shown there.

Test Plan: tested renaming chat still works and that it doesn't appear in command pallette


keegancsmith:
@sqs I know you want this feature axed. I will do that in the next PR. Just want to fix the bug first in case others decide we actually want to keep this feature.

sqs (approved):

PR #4413, opened 2024-05-31
vscode: remove Rename Chat functionality

keegancsmith:
Being ruthless with this feature. We will be moving chat history to the backend, synchronizing it, etc. Being able to rename chats will add complexity. We can always add it back once we do all that.

Test Plan: CI. Ran manually and the edit title button was gone.

More context at https://sourcegraph.slack.com/archives/C0748TDGT29/p1717162754194119?thread_ts=1717160259.281599&cid=C0748TDGT29


abeatrix (approved):
Nice!

Personally I've never used this feature and find our current default to the last question i sent to be more helpful 😁

sqs (approved):
Thanks!

PriNova:
I used it a lot, because the last question never reflects the overall topic about the context.
In one of my one projects, I let AI summarize the very first question into a 3-5 word sentence.

sqs:
@PriNova Thanks for the feedback. If we used AI-generated titles instead of letting you manually rename them, would you be happy with that?

PriNova:
Absolutely. If I chat with Cody, I always focus on bounded context of topic/concern. For example, fixing a bug or implementing a feature, etc.

iezed:
Hi, So you have removed the ability to change the chat name?

PR #4415, opened 2024-05-31
Document Code: improve tree-sitter query for c++

abeatrix:
- Add support for template declarations and struct specializations
- Refine class and struct definitions to better capture symbol and range information
- Remove unused variable and union declaration rules

## Test plan

Updated test to cover template declaration.

Confirm https://github.com/sourcegraph/jetbrains/issues/1728 is fixed:

image

- Open https://github.com/fmtlib/fmt/blob/master/test/format-test.cc#L2410
- Put cursor somewhere inside template <> struct formatter<convertible_to_cstring> template method.
- Run Document Code

https://github.com/sourcegraph/cody/assets/68532117/296d3188-9f43-4b59-8c3a-c4e11a292c19


kalanchan (approved):
gogo jetbrains hacker!

PR #4416, opened 2024-05-31
Add experimental "Minion" panel

beyang:
https://github.com/sourcegraph/cody/assets/1646931/4aa348ee-03f3-4c14-955e-82c599154b9f

Adds a new experimental "minion" panel that sketches out an execution plan for a high-level task.

## Summary of code changes

- `vscode/src/minion/` contains the extension host code for Minion, including the panel controller (`MinionController`), which plays a similar role to `SimpleChatPanelProvider` for chat.
- `vscode/webviews/minion/` contains the webviews for Minion. The code here is similar to the webview code for Chat.

The Minion panel scaffolding (the basic structure of the controller class and webview) closely resembles that of Chat, with some differences:
- `MinionController` inherits from a more general `ReactPanelController` class, which handles some of the lifecycle/init events for the controller-webview pair.
- It uses a genericized `GenericVSCodeWrapper` to wrap the VS Code webview API with generic parameters for the two-way message protocol types

We could adapt the Chat code to use these in the future.

## Test plan

This feature is behind an experimental feature flag `cody.experimental.minion.anthropicKey`. To test it, set the flag to a valid Anthropic API key and restart the editor.


sqs (approved):
Minimal changes to core Cody itself. Low risk.

anyb1s:
How do I show it, I have anthropic key, but showing empty page

PR #4422, opened 2024-06-01
fix copy/insert/save code buttons in chat

sqs:
These buttons were sometimes not showing on new messages due to the `useEffect` hook not having the right deps.

Fixes that issue and adds an e2e test to ensure the buttons show up and that the copy button copies the right text.

Fixes https://linear.app/sourcegraph/issue/CODY-2014/missing-copyinsert-buttons-on-code-snippets

## Test plan

CI + e2e


eseliger (approved):
Stamp, trusting the test

PR #4432, opened 2024-06-03
Edit: preserve intent context

valerybugakov:
- https://github.com/sourcegraph/cody/pull/3307 removed `extractContextItemsFromContextMessages` and replaced it with `isContextItem(value) ? value : value.file` which removes context items content from the final prompt. 

## Test plan

1. Updated and added unit tests
2. Tested locally by triggering the `doc` command on the function in the middle of the file. You should observe document prefix and suffix being added to the prompt in the Cody output channel.

### DIFF
--- a/vscode/src/prompt-builder/index.test.ts
+++ b/vscode/src/prompt-builder/index.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
-import type { ContextItem, ContextMessage, Message } from '@sourcegraph/cody-shared'
+import type { ContextItem, Message } from '@sourcegraph/cody-shared'
 import {
     type ChatMessage,
     ContextItemSource,
@@ -26,8 +26,7 @@ describe('PromptBuilder', () => {
             content: 'foobar',
             size: 100,
         }
-        const transcript: ContextMessage[] = [{ speaker: 'human', file, text: ps`` }]
-        expect(() => builder.tryAddContext('enhanced', transcript.reverse())).rejects.toThrowError()
+        expect(() => builder.tryAddContext('enhanced', [file])).rejects.toThrowError()
     })
 
     it('throws an error when trying to add User Context before chat input', () => {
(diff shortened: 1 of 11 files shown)


vscode/src/prompt-builder/index.test.ts
```diff
@@ -284,14 +280,17 @@ describe('PromptBuilder', () => {
...
+                fullFile,
+                partialFile,
+            ])
             expect(limitReached).toBeFalsy()
             expect(builder.contextItems).toStrictEqual([fullFile])
         })
 
         it('should not remove user-added with overlapping ranges even when full file is provided', async () => {
-            const builder = new PromptBuilder({ input: 10, output: 10 })
+            const builder = new PromptBuilder({ input: 50, output: 50 })
```
valerybugakov, line 293: I increased prompt limits because the content now has the text `Codebase context from file foo/bar.go` added to it.

vscode/src/prompt-builder/index.ts
```diff
@@ -172,11 +169,3 @@ export class PromptBuilder {
         return result
     }
 }
-
-function isContextItem(value: ContextItem | ContextMessage): value is ContextItem {
-    return 'uri' in value && 'type' in value && !('speaker' in value)
-}
-
-function contextItem(value: ContextItem | ContextMessage): ContextItem {
-    return isContextItem(value) ? value : value.file
```
valerybugakov, line 181: Here, we always used `ContextMessage['file']`, even though the context snippet content was added to the ContextMessage['text'].

sqs (approved):
Confirmed it includes more context from the file.

A couple possible issues:

- I only see it including context AFTER my selection, not before.
- I see it duplicating the `Codebase context from ...` text.

image

valerybugakov:
Thanks @sqs! I'll address these in a follow-up.

valerybugakov:
> I only see it including context AFTER my selection, not before.
Fixed here https://github.com/sourcegraph/cody/pull/4435

PR #4435, opened 2024-06-03
Agent: fix out of bounds document offsets

valerybugakov:
I [noticed](https://github.com/sourcegraph/cody/pull/4432) that the edit prompt does not contain document suffixes in the agent tests. This is caused by `DocumentOffsets` not handling out-of-bound line numbers like VS Code does. If the line number is out of bound, it results in NaN offset, which causes `document.getText(range)` to return an empty string. This PR fixes it.

## Test plan

- Added unit test
- Updated the doc command unit tests to include long prefix and suffix, which should be present in Polly recordings.

### DIFF
--- a/agent/src/__tests__/document-code/src/TestClass.ts
+++ b/agent/src/__tests__/document-code/src/TestClass.ts
@@ -1,4 +1,14 @@
 const foo = 42
+// Should be present in the LLM prompt
+const longPrefix = `
+    longPrefix content
+    longPrefix content
+    longPrefix content
+    longPrefix content
+    longPrefix content
+    longPrefix content
+    longPrefix content
+`
 
 export class TestClass {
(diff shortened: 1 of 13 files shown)


agent/src/__tests__/document-code/src/TestClass.ts
```diff
@@ -1,4 +1,14 @@
 const foo = 42
+// Should be present in the LLM prompt
```
olafurpg, line 2: FYI, it's possible to introspect on the final prompt with `testing/networkRequests` to write an assertion that this text is included int he prompt. We did this for the original graph context tests, which I then removed

valerybugakov: Nice, I'll use it

PR #4439, opened 2024-06-03
Chat: Don't append @ when "Add context" is pressed multiple times

toolmantim:
Don't append unnecessary @'s if you press the "Add context..." button in chat multiple times. Instead, this change just refocuses the input at the relevant location.

## Test plan

- CI (could not find any relevant test to update/add to)


thenamankumar (approved):
sweet

sqs (approved):

PR #4443, opened 2024-06-03
commands: getContextFileFromUri treats empty ranges as full file

keegancsmith:
"Add file to cody chat" command from a right click on a file would fail since we would have an empty range set. This treats any empty range as an indication the user wanted to add the whole file, which should plug the hole for other potential call sites on getContextFileFromUri.

Additionally if we do not have a range defined, we now do not set range on the returned ContextItem. This is so we render the file in chat as just the filename, rather than the filename with the range 1:lineCount.

Test Plan: Manually added and removed a file from chat. Then added the file via right click on the text area, right click on the explorer. Additionally tested that selecting text and right clicking correctly selected just the range.

Fixes https://linear.app/sourcegraph/issue/CODY-1996/add-file-to-cody-chat-is-broken

### DIFF
--- a/vscode/src/commands/context/file-path.ts
+++ b/vscode/src/commands/context/file-path.ts
@@ -20,12 +20,25 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
                 return []
             }
 
+            const doc = await vscode.workspace.openTextDocument(file)
+
             // empty range can happen when user initiates action from right
             // click. Treat as wanting full document.
             range = range?.isEmpty ? undefined : range
 
-            const doc = await vscode.workspace.openTextDocument(file)
-            const content = doc?.getText(range).trim()
+            // if the range is the full file, remove our range specifier so it
+            // renders nicely in the UI
+            const fullRange = new vscode.Range(
+                0,
+                0,
+                doc.lineCount,
+                doc.lineAt(doc.lineCount - 1).text.length
+            )
+            if (range?.contains(fullRange)) {
+                range = undefined
+            }
+
+            const content = doc.getText(range).trim()
             if (!content) {
                 throw new Error('No file content')
             }
--- a/vscode/src/commands/context/file-path.ts
+++ b/vscode/src/commands/context/file-path.ts
@@ -38,8 +38,8 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
                 range = undefined
             }
 
-            const content = doc.getText(range).trim()
-            if (!content) {
+            const content = doc.getText(range)
+            if (!content.trim()) {
                 throw new Error('No file content')
             }
             const size = TokenCounter.countTokens(content)
--- a/vscode/src/commands/context/file-path.ts
+++ b/vscode/src/commands/context/file-path.ts
@@ -20,13 +20,15 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
                 return []
             }
 
+            // empty range can happen when user initiates action from right
+            // click. Treat as wanting full document.
+            range = range?.isEmpty ? undefined : range
+
             const doc = await vscode.workspace.openTextDocument(file)
             const content = doc?.getText(range).trim()
             if (!content) {
                 throw new Error('No file content')
             }
-            const endLine = Math.max(doc.lineCount - 1, 0)
-            range = range ?? new vscode.Range(0, 0, endLine, 0)
             const size = TokenCounter.countTokens(content)
 
             return [
--- a/vscode/src/prompt-builder/unique-context.test.ts
+++ b/vscode/src/prompt-builder/unique-context.test.ts
@@ -1,4 +1,4 @@
-import { type ContextItem, ContextItemSource } from '@sourcegraph/cody-shared'
+import { type ContextItem, ContextItemSource, type RangeData } from '@sourcegraph/cody-shared'
 import { describe, expect, it } from 'vitest'
 import { URI } from 'vscode-uri'
 import { getUniqueContextItems, isUniqueContextItem } from './unique-context'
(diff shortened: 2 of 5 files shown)


vscode/src/commands/context/file-path.ts
```diff
@@ -20,13 +20,15 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
...
+            // click. Treat as wanting full document.
+            range = range?.isEmpty ? undefined : range
+
             const doc = await vscode.workspace.openTextDocument(file)
             const content = doc?.getText(range).trim()
             if (!content) {
                 throw new Error('No file content')
             }
-            const endLine = Math.max(doc.lineCount - 1, 0)
-            range = range ?? new vscode.Range(0, 0, endLine, 0)
```
abeatrix, line 29: Hmm I thought this would treat the undefined range as the full file

abeatrix (approved):
ty!

vscode/src/commands/context/file-path.ts
keegancsmith: it does, but it ends up rendering the item in the UI with line ranges. By leaving it off we just render the context item as the filename (which implies the full file)

abeatrix: Oh gotcha, thanks for explaining!

vscode/test/e2e/command-custom.test.ts
```diff
@@ -268,9 +268,9 @@ test.extend<ExpectedEvents>({
     // The files from the open tabs should be added as context
     await expectContextCellCounts(contextCell, { files: 2 })
     await contextCell.click()
-    await expect(chatContext.getByRole('link', { name: 'index.html:1-11' })).toBeVisible()
+    await expect(chatContext.getByRole('link', { name: 'index.html' })).toBeVisible()
     await expect(
-        chatContext.getByRole('link', { name: withPlatformSlashes('lib/batches/env/var.go:1') })
+        chatContext.getByRole('link', { name: withPlatformSlashes('lib/batches/env/var.go') })
```
keegancsmith, line 273: @abeatrix I don't really understand how the code I change only affects these lines in the e2e test. But when looking at the behaviour this is better and really anytime we get something from the active editor state it should be the file without line ranges if there is no active selection.

keegancsmith:
hmm, I think I need to do a bit more work here. One of the agent tests detected some suspicious changes in the transcript we generate.

image

vscode/src/commands/context/file-path.ts
```diff
@@ -20,13 +20,15 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
...
 
+            // empty range can happen when user initiates action from right
+            // click. Treat as wanting full document.
+            range = range?.isEmpty ? undefined : range
+
             const doc = await vscode.workspace.openTextDocument(file)
             const content = doc?.getText(range).trim()
             if (!content) {
                 throw new Error('No file content')
             }
```
abeatrix, line 31:
```suggestion
            const endLine = Math.max(doc.lineCount - 1, 0)
            range = range ?? new vscode.Range(0, 0, endLine, 0)
            
            const isFullFile = range.end.line === doc.lineCount - 1
```

I think we still want to check the existing range to see if it's full file. If yes, then we remove it from the `return` statement below instead?

```ts
            return [
                {
                    type: 'file',
                    content,
                    uri: file,
                    source: ContextItemSource.Editor,
                    range: isFullFile? undefined : toRangeData(range),
                    size,
                },
            ] satisfies ContextItem[]
```

abeatrix (changes requested):

vscode/src/commands/context/file-path.ts
keegancsmith: aah nice catch. That probably explains some of the weirdness I was seeing.

vscode/src/commands/context/file-path.ts
```diff
@@ -20,13 +20,15 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
                 return []
             }
 
+            // empty range can happen when user initiates action from right
+            // click. Treat as wanting full document.
+            range = range?.isEmpty ? undefined : range
+
             const doc = await vscode.workspace.openTextDocument(file)
             const content = doc?.getText(range).trim()
```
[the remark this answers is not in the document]
keegancsmith: It doesn't work the same. range being empty means that it has zero length not that it is undefined. So for example when you don't have a selection the vscode can return a range which has zero length but starts at the position of the cursor. This was leading to getText returning an empty string.

vscode/src/commands/context/file-path.ts
```diff
@@ -21,12 +21,27 @@ export async function getContextFileFromUri(file: URI, range?: vscode.Range): Pr
...
+                0,
+                doc.lineCount,
+                doc.lineAt(doc.lineCount - 1).text.length
+            )
+            if (range?.contains(fullRange)) {
+                range = undefined
+            }
+
+            const content = doc.getText(range)
+            if (!content.trim()) {
```
keegancsmith, line 42: I removed the trim from content since trailing newlines are actually part of the content and is how it works in other parts of our codebase when getting context. This also avoids how the range maps onto content not making sense. I believe the original intention of the trim call was to avoid adding content which was just an empty line. This will still be the case since we keep the check against trimmed content.

Note: removing the trim from the content affected the agent recording. The only difference in the recording is we now have the newline in the transcript we send to the LLM

vscode/src/prompt-builder/unique-context.test.ts
```diff
@@ -308,55 +308,54 @@ describe('Unique Context Items', () => {
...
+                expect(range.end.line).toBeLessThanOrEqual(contentLines.length)
+                content = contentLines.slice(range.start.line, range.end.line).join('\n')
+            }
+            return {
+                type: 'file',
+                uri: URI.file('/foo/bar'),
+                content,
+                range,
+            }
+        }
```
keegancsmith, line 329: I made this change so that `content` reflects the passed in range. Previously isUniqueContextItem would never look at content, but now that it does it needs to have reasonable values.

vscode/src/prompt-builder/unique-context.test.ts
```diff
@@ -366,42 +365,57 @@ describe('Unique Context Items', () => {
...
 
+        it('returns false when the new item has no range and an earlier item has whole file range', () => {
+            const item1: ContextItem = baseFile({
+                start: { line: 0, character: 0 },
+                end: { line: contentLines.length, character: 0 },
+            })
+            const item2: ContextItem = baseFile(undefined)
+
+            expect(isUniqueContextItem(item2, [item1])).toBeFalsy()
+        })
```
keegancsmith, line 376: This test case demonstrates an issue before this PR which seemed to happen with openTabs and other custom commands. You could have the whole file as "selected" context, then as an opentab have the whole file again (but without ranges set). We would end up not deduplicating that. With this PR we now do.

vscode/src/prompt-builder/unique-context.test.ts
```diff
@@ -366,42 +365,57 @@ describe('Unique Context Items', () => {
...
+                start: { line: 0, character: 0 },
+                end: { line: 10, character: 0 },
+            })
+            const inner: ContextItem = baseFile({
+                start: { line: 2, character: 0 },
+                end: { line: 5, character: 0 },
+            })
+
+            expect(isUniqueContextItem(inner, [noOverlapFirst, outer])).toBeFalsy()
+        })
```
keegancsmith, line 419: This is a bug I noticed while doing this PR. If we do have 3 items in the same file, we only ever do the check against the first added file since we used to return early. Now we only return early if the item is indeed a duplicate.

keegancsmith (commented):
@abeatrix can I get a re-review. As I started fixing this the change grew into some fixes in unique-context as well.

vscode/src/prompt-builder/unique-context.ts
```diff
@@ -70,10 +70,18 @@ export function isUniqueContextItem(itemToAdd: ContextItem, uniqueItems: Context
...
                     rangeContainsLines(item.range, itemToAddRange)
-                )
+                ) {
+                    return false
+                }
+            }
+
+            // Duplicates if whole file (undefined range) and selection has the
+            // same content.
+            if (item.range && !itemToAddRange && equalContent(item, itemToAdd)) {
```
abeatrix, line 83:
```suggestion
            if (!itemToAddRange && equalContent(item, itemToAdd)) {
```

This is my bad, just noticed we already exit if `!item.range` on line 61.

vscode/src/prompt-builder/unique-context.ts
```diff
@@ -105,3 +113,10 @@ function rangesOnSameLines(range1: RangeData, range2: RangeData): boolean {
 function isUserAddedItem(item: ContextItem): boolean {
     return getContextItemTokenUsageType(item) === 'user'
 }
+
+/**
+ * Checks if content is set and equal.
+ */
+function equalContent(item1: ContextItem, item2: ContextItem): boolean {
+    return !!item1.content && item1.content === item2.content
```
abeatrix, line 121:
```suggestion
    return !!item1.content && item1.content === item2.content
```

QQ: should we compare the trimmed content for cases where the differences are new lines etc?

abeatrix (approved):
Thanks for fixing and adding the tests with details!

Is the agent recordings needed for this PR? Asking because I noticed the only changes in the recordings are the `-` at the begging of the text which doesn't seem to be related to the changes you made (or maybe Github is not showing it in the diff view)

image

keegancsmith:
> Is the agent recordings needed for this PR? Asking because I noticed the only changes in the recordings are the `-` at the begging of the text which doesn't seem to be related to the changes you made (or maybe Github is not showing it in the diff view)

See my comment here https://github.com/sourcegraph/cody/pull/4443#discussion_r1625963303

That just tells yaml to not trim the final newline, so is an actual change.

vscode/src/prompt-builder/unique-context.ts
keegancsmith: Cool, will update this and line 72.

vscode/src/prompt-builder/unique-context.ts
keegancsmith: My initial reaction is to not treat whitespace as not significant. In particular I think about languages like python. However, in this case we are talking about the full file, so leading and trailing whitespace does indeed seem not significant. So I will update this check to match your suggestion.

Issue #4348, opened 2024-05-29
Extension causes high cpu load


tanmoy02:
- Issue Type: `Performance`
- Extension Name: `cody-ai`
- Extension Version: `1.18.2`
- OS Version: `Windows_NT x64 10.0.22631`
- VS Code version: `1.89.1`

![image](https://github.com/sourcegraph/cody/assets/1296341/908ea5b1-5d9a-48d7-8e37-39d82cf9ab17)

[sourcegraph.cody-ai-unresponsive.cpuprofile.txt](https://github.com/sourcegraph/cody/files/15481662/sourcegraph.cody-ai-unresponsive.cpuprofile.txt)

avion23:
Cody is for me unusable, because it uses up all CPU as soon as I start chatting.

Maybe it's the streaming?

RXminuS:
I'm just debugging this @avion23 Could you try setting these and report if the CPU is still ramping up? If possible and you have the time it would be great help if you could make a CPU profile like @tanmoy02 .*

![CleanShot 2024-06-03 at 15 18 09](https://github.com/sourcegraph/cody/assets/3949285/b632d8f9-6199-458f-bfd9-493b45cd9ff4)

* I think this is done by cloning the project, under Run & Debug selecting `Launch VSCode (Desktop)` (which opens a new window with the locally built extension running) and then in the original VSCode command window searching `Debug: Take Performance Profile` and then selecting this running instance. But maybe @tanmoy02 has some pointers on a quicker way?

RXminuS:
@tanmoy02 correct me if I'm wrong but yours seemed to cause lag / high CPU load during typing correct? @avion23 is yours **only** during chat or during normal typing as well?

avion23:
It's only while the chat window *is outputting a new response*

tanmoy02:
@RXminuS For the past few days, my editor has been experiencing significant lag. The IntelliSense feature in VS Code was taking an excessively long time to load, often displaying the "loading" message for 30 to 40 seconds, even when looking for type definitions.
To investigate the cause of this lag, I opened the **Show Running Extensions** menu. There, I discovered that two extensions, Cody and another one, were responsible for a high CPU load. To pinpoint the main culprit, I disabled each extension individually to test the impact. However, both extensions were found to be causing the same issue.
As a temporary solution, I have decided to report the problem and disable both extensions until a fix is released by the developers.

tanmoy02:
[...]

PR #4446, opened 2024-06-03
Jaccard Retriever: Reduce performance overhead

RXminuS:
User reported high-cpu usage with Cody installed https://github.com/sourcegraph/cody/issues/4348. Their CPU profile revealed an average 4 seconds spent getting context for autocomplete with a large amount spent counting word occurrences. 

CleanShot 2024-06-03 at 16 34 27

Now instead of stemming each word individually we maintain a small cache which speeds up stemming by more than 7x (if cache cleared before every benchmark run. Can be much higher in real-world scenarios).

Fixes CODY-1928

## Test plan
- tested several options by writing a simple benchmark and picking the most performant alternative
CleanShot 2024-06-03 at 16 30 07
- ran existing unit tests to ensure no altered behavior

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -8,6 +8,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
+- Performance: Reduced the performance overhead for certain types of context fetching, especially for larger files. This might have caused issues with slow autocomplete before. [pull/4446](https://github.com/sourcegraph/cody/pull/4446)
+
 ### Changed
 
 ## [1.20.1]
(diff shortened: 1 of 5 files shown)


olafurpg (approved):
LGTM 👍🏻 😍 Nice find!

Also, TIL `import { bench } from 'vitest'`

olafurpg:
BTW added 
CleanShot 2024-06-03 at 21 29 16@2x

RXminuS:
> BTW added 

I'm just going to hold off on closing the original issue a little longer because I want to do some more testing / fixes along the full autocomplete path. It's a great candidate for more memoization as most of the time the content will nearly identical and can tremendously speed up autocomplete in those cases. Tracked in [CODY-2037](https://linear.app/sourcegraph/issue/CODY-2037)

PriNova:
Hey @RXminuS 
I saw this performance overhead also during indexing a repo.
While monitoring with `iotop` it seems that the cody-engine is spawning `rg` (I guess it is ripgrep) over all available cores.

On my machine on Linux (WSL) it took about 1 ½ min for the cody source code. With 100% cpu

PR #4447, opened 2024-06-03
fix(ci): improvements for test reliability in CI

jamesmcnamara:
Contains a slew of minor improvements for the E2E test suite as well as some quality of life improvements such as:
* Upgrades various Github Actions dependencies to resolve issues warnings about deprecations
* Improved the test failure assets. Currently we'd almost never upload test assets because it'd only work if every single test in a run failed (we needed to take into account `--repeat-each`). Now we upload a video and a snapshot for each failed run.
* Adds Github Annotations for test failures
* Adds small 300ms pauses to some chronically flaky tests where it was really a timing issue
* Adds a tool to open developer tools in the test app either through an env flag (`DEBUG=true`) or programatically
* Integrated telemetry event checking into `expect` and only present diffs to user (I found it very hard to read before)

## Test plan
Will run in CI


vscode/package.json
```diff
@@ -9,7 +9,7 @@
   "icon": "resources/cody.png",
   "description": "AI coding assistant that uses search and codebase context to help you write code faster. Cody brings autocomplete, chat, and commands to VS Code, so you can generate code, write unit tests, create docs, and explain complex code using AI. Choose from the latest LLMs, including GPT-4o and Claude 3.",
   "scripts": {
-    "postinstall": "pnpm download-wasm",
+    "postinstall": "pnpm download-wasm && playwright install",
```
abeatrix, line 12: When is this script run? (I'm not familiar with the postinstall script 🙈 )

vscode/test/e2e/helpers.ts
```diff
@@ -231,35 +293,13 @@ export const test = base
...
-                }
-                try {
-                    await assertEvents(loggedV2Events, expectedV2Events)
-                } catch (error) {
-                    console.error('Expected v2 events do not match actual events!')
-                    console.log('Expected:', expectedV2Events)
-                    console.log('Logged:', loggedV2Events)
-                    throw error
-                }
+                await expect(loggedEvents).toContainEvents(expectedEvents)
```
abeatrix, line 296: Nice!

vscode/src/prompt-builder/index.test.ts
```diff
@@ -373,10 +373,9 @@ describe('PromptBuilder', () => {
                 .join('\n')
 
             expect(builder.contextItems.length).toBe(1)
-            const filePath = convertFileNameByPlatform('foo/bar.go')
             expect(promptContent).toMatchInlineSnapshot(`
               "preamble
-              Codebase context from file ${filePath}:
+              Codebase context from file ${PromptString.fromDisplayPath(file.uri)}:
```
abeatrix, line 378:
```suggestion
              Codebase context from file ${displayPath(file.uri)}:
```
nit: I think we can use the `displayPath` function directly without turning it into a PromptString. Feel free to ignore this 😄

abeatrix (approved):
Looks great! I ran a couple of tests locally and didn't notice any regression. The CI looks healthy too, great job!

vscode/package.json
jamesmcnamara: Whenever a user runs `pnpm install` or `npm install` this gets invoked afterwards. When I upgraded playwright, it bumped the version of bundled browsers it was using which it doesn't install automatically. This ensures that other users don't have to worry about this :)

olafurpg: Can we move this step to somewhere that only executes when running e2e tests? This makes `pnpm install` take a long time to complete for all contributors even if they are not running e2e tests locally. I very rarely run e2e tests locally, and I suspect the same is true for many other contributors.

olafurpg: For example, we run `pnpm install` in automated scripts to build agent/webview assets for the JetBrains/Eclipse. We do this in local computers and in the CI for the JetBrains repo.

olafurpg: Opened https://github.com/sourcegraph/cody/pull/4544

jamesmcnamara: How about https://github.com/sourcegraph/cody/pull/4545 where we just run for "test:e2e"?

PR #4449, opened 2024-06-03
Refactor `ModelProvider` into `Model` and `ModelsService`

chrsmith:
This PR is the first step in the "[Server-side Cody Model Selection](https://linear.app/sourcegraph/project/server-side-cody-model-selection-cca47c48da6d)" project. There should be no functional changes, and instead it just splits the `ModelProvider` class into separate types: `class Model` and `class ModelsService`.

- `Model` contains all of the non-static fields from `ModelProvider`, and describes an LLM model and things such as its token counts.
- `ModelsService` contains all of the static fields and methods from `ModelProvider`, and provides the functionality for managing the types of LLM models available to the user.

Splitting `Model` and `ModelsService` into two distinct types, with their own responsibilities should make things generally easier to understand.

Reviewers, please double check my thinking here. If there is a good reason why we would want to keep all of this information bundled in a single type (e.g. `ModelProvider`) let me know. This separation is foundational to how I plan on going about the server-side model selection in subsequent PRs.

## Next Steps

After this PR lands, my thinking is to continue refactoring the overall codepath that handles models or their providers, and introduce a few key concepts / abstraction layers.

Here's the gist of what I am thinking:

```ts
// ModelId is a stable identifier for an LLM model, e.g. "gpt-3.5-turbo".
//
// NOTE: This _usually_ is also the "name" of a model, but is a different concept.
// e.g. the Model with ID "gpt-3.5-turbo" may have Name "gpt-3.5-turbo-16k" or
// "gpt-3.5-turbo-0613". That is, the server-supplied LLM configuration will use
// Model ID as a way to identify a model to clients, but the underlying LLM model
// used for processing requests may change.
type ModelId = string

// ProviderId is a stable identifier for the LLM provider, e.g. "openai" or "anthropic".
type ProviderId = string

// ApiVersionId is a stable identifier for the "API surface area" exposed by an LLM
// provider. It is provider-specific, and can take on different forms. e.g.
// OpenAI uses "2024-02-01" to refer to the first release of its API. While other
// providers may simply use "v1".
//
// The critical aspect is that the combination of { ProviderId, ApiVersionId }
// referrs to the ability for a Cody client or server to _use_ a set of LLMs. e.g.
// { "google", "v1/text-only" } is how Cody clients may interact with Gemini models.
// However, if we were to latter support multi-modal prompts, then the required
// capability would be described by a new API version. e.g. "v1/text-and-images".
type ApiVersionId = string

// ModelRef uniquely identifies a model. The assumption being that if the Cody
// client or server has code to support the { providerId, apiVersionId }, then
// any new models will work.
interface ModelRef  {
    providerId: ProviderID
    apiVersionId: ApiVersionID
    modelID: ModelId
}
```

So rather than passing around a string of the form `${provider}/${model}`, we would pass a `ModelRef` type instead.

There are a few other things that should fall out of having a `ModelsService` abstraction too, like (🤞) enabling both Cody Enterprise users to select the LLM model to use like Cody Pro users can. (By simply having both codepaths work by pulling data from the `ModelsService`. It will just be the case that for Cody Enterprise users initially, only one LLM model will be available.)

Part of https://linear.app/sourcegraph/issue/PRIME-282/refactor-modelprovider-into-model-and-modelsservice.

## Test plan

Just ran existing unit tests.


lib/shared/src/models/index.ts
```diff
@@ -29,6 +27,9 @@ export class ModelProvider {
         /**
          * The model id that includes the provider name & the model name,
          * e.g. "anthropic/claude-3-sonnet-20240229"
+         * 
+         * TODO(PRIME-282): Replace this with a `ModelRef` instance and introduce a separate
+         * "modelId" that is distinct from the "modelName". (e.g. "claude-3-sonnet" vs. "claude-3-sonnet-20240229")
```
abeatrix, line 32: does this mean to replace `model` with `id` and add a new `name` field?

lib/shared/src/models/index.ts
```diff
@@ -4,11 +4,9 @@ import type { ModelContextWindow, ModelUsage } from './types'
 import { getModelInfo } from './utils'
 
 /**
- * ModelProvider manages available chat and edit models.
- * It stores a set of available providers and methods to add,
- * retrieve and select between them.
+ * Model describes an LLM model and its capabilities.
  */
-export class ModelProvider {
+export class Model {
```
abeatrix, line 9: There are multiple "Models" in the client. Maybe a follow-up task for the client team to rename / refactor [SimpleChatModel](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/SimpleChatModel.ts) to avoid confusion, wdyt?  😆

abeatrix (commented):
Thanks for refactoring the classes to make them more distinctive 🙇‍♀️ 

I left some minor questions / feedback in-line, nothing blocking.

Noticed the build is failing because of some lint issues, which I was able to resolve with `pnpm install && pnpm run biome`. After that, the only issue the linter is complaining is the `ModelsService` class being static only.

lib/shared/src/models/index.ts
```diff
@@ -62,63 +63,76 @@ export class ModelProvider {
...
+ * are supported on the backend, which ones are available based on the user's
+ * preferences, etc.
+ * 
+ * TODO(PRIME-228): Update this type to be able to fetch the models from the
+ *      Sourcegraph backend instead of being hard-coded.
+ * TODO(PRIME-283): Enable Cody Enterprise users to select which LLM model to
+ *      used in the UI. (By having the relevant code paths just pull the models
+ *      from this type.)
+ */
+export class ModelsService {
```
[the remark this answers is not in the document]
chrsmith: I just added a private constructor and the linter is happy for the time being.

When we start actually fetching data from the backend, we'll need to have a proper initialization step. (e.g. getting the user's credentials.) So at that time it would make sense to rework this to be a proper singleton.

lib/shared/src/models/index.ts
```diff
@@ -131,32 +145,32 @@ export class ModelProvider {
...
-                    .map(m => m.model)
-                    .join(', ')}`
+                `No model found for substring ${modelSubstring}. Available models: ${modelsList}`
             )
         }
         throw new Error(
-            `Multiple models found for substring ${modelSubstring}: ${models
-                .map(m => m.model)
-                .join(', ')}`
+            `Multiple models found for substring ${modelSubstring}: ${modelsList}`
```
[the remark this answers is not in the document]
chrsmith: That looks a lot cleaner, 🙏 .

lib/shared/src/models/index.ts
chrsmith: Not yet. That would be a bit too risky in terms of unintentionally breaking something. So I'll make that change in a subsequent PR.

lib/shared/src/models/index.ts
chrsmith: 😕 yeah, I'll ask around about that. It probably isn't the end of the world to have different `Model` types in different packages, but we should be a little more specific  about what their responsibilities are. (e.g. a doc comment on `SimpleChatModel` would be helpful.)

chrsmith:
@abeatrix PTAL. The linter is happy now. I also merged in the latest changes from `main` and made the same set of refactoring.

abeatrix (approved):
Lgtm!

PR #4455, opened 2024-06-04
scroll to prompt editor after sending context items to chat

keegancsmith:
When adding a file to the prompt via the menu it was unclear if anything happened if the prompt editor was out of view. We now ensure that after we give focus to the editor we scroll to it.

## Test plan

With a long message ensuring the follow-up prompt was out of view, I added a file to the context via right click.

https://github.com/sourcegraph/cody/assets/187831/62bbadfc-ff43-4166-8009-99407a2847a0


keegancsmith:
@sqs not happy with how this is implement and how the scrolling works, but keen to get some feedback on how to do this. I see you put effort in to avoid scrolling, so maybe this doesn't make sense to do? This was feedback from @toolmantim when he was testing out the add file to chat menu item.

sqs (approved):
> I see you put effort in to avoid scrolling

This was only for the case of focusing the followup input box after you send a message (in case it's off-screen). In the case of this PR's behavior, autoscroll is desirable.

PR #4456, opened 2024-06-04
DX: Add type/build checks for common hard-to-notice pitfalls

RXminuS:
During debugging of [CODY-1928](https://linear.app/sourcegraph/issue/CODY-1928/) I noticed we were re-registering a global command. 

The fix would have been simple but encouraged by @olafurpg's urge to seize every opportunity to guard against regressions and capture reproductions in tests; I spent some time ensuring that these kind of mistakes can't creep in again.

1. Avoid (transitive) imports of `vscode` in the `agent/vscode-shim.ts`. This was the third time I accidentally tripped it and each time I wasted too much time remembering there is a massive disclaimer comment. Now we simply check this condition during the build...as a service to future self.
CleanShot 2024-06-04 at 15 44 49@2x
2. The package.json manifest now has a schema that we can expand upon. It's only basic right now, because VSCode doesn't provide a schema for us. But it at least ensures that all commands are registered with a `cody` prefix.
CleanShot 2024-06-04 at 23 00 55@2x
3. I introduced some `typehacks` that are extensions on top of package types which further restrict the usage to ensure not just type-safety but soundness. These types are not included by default as they would pollute the global scope. They are however run on build.
CleanShot 2024-06-04 at 23 20 02@2x

NOTE: Right now it's not really possible to show these in the editor as there seems to be no way of overriding the tsconfig for a particular subfolder. You can simply add the typehacks to the tsconfig.json and you'd get something like this (and break your build & cache in the process...`tsc build --clean` #ftw)
CleanShot 2024-06-03 at 22 31 02@2x

Fixes [CODY-2023](https://linear.app/sourcegraph/issue/CODY-2032/)

## Test plan
- manually verified that builds still work
- manually verified that introducing the above issues triggers build errors

### DIFF
--- a/agent/src/esbuild.mjs
+++ b/agent/src/esbuild.mjs
@@ -3,9 +3,49 @@ import path from 'node:path'
 import process from 'node:process'
 
 import { build } from 'esbuild'
-;(async () => {
-    const minify = process.argv.includes('--minify')
 
+main().catch(err => {
+    console.error('Could not build the agent.', err.message)
+    process.exit(1)
+})
+
+async function main() {
+    await verifyShim()
+
+    const minify = process.argv.includes('--minify')
+    await buildAgent(minify)
+}
+
+async function verifyShim() {
+    // we first verify that the shim does not have vscode in its dependency tree. This would break the agent in a hard to detect way.
+    const shimPlugins = [detectForbiddenImportPlugin(['vscode'])]
+    /** @type {import('esbuild').BuildOptions} */
+    const esbuildOptions = {
+        entryPoints: ['./src/vscode-shim.ts'],
+        bundle: true,
+        platform: 'node',
+        sourcemap: true,
+        logLevel: 'silent',
+        write: false,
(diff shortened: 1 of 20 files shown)


sqs (approved):
thanks!

olafurpg (commented):
Wow. Learned a lot by reading the diff here 😍 Thank you!

If it's easy to extend then we should also guard against importing `vitest` from the main bundle to prevent this error here (documented in agent/readme.md)

```
If you get an ESBuild error about "You can mark the path "#async_hooks" as external to exclude it from the bundle, which will remove this error." then the fix is to remove dependencies on vitest from the agent bundle. Vitest depends on the p-limit npm package, which uses #async_hooks that we currently don't handle in the ESBuild config.
```

RXminuS:
Great idea, made a issue for it [CODY-2084](https://linear.app/sourcegraph/issue/CODY-2084/guard-against-vitest-imports)

PR #4461, opened 2024-06-05
Fix: add experimental LSP context

valerybugakov:
- Adds experimental LSP context support to the Fix command. For now, it's meant to be used for internal testing and benchmarking. 
- Part of https://github.com/sourcegraph/cody-issues/issues/204

Fixes CODY-1556
https://github.com/sourcegraph/cody/assets/3846380/36459393-1165-4b46-9624-97be6b409e8d

## Test plan

1. Add `"cody.edit.experimental.graphContext": "lsp-light"` to your VS Code settings.
2. Trigger the fix command on an object declaration with the type declared in a separate file.

### DIFF
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -83,6 +83,11 @@ export interface Configuration {
     autocompleteExperimentalMultiModelCompletions?: MultimodelSingleModelConfig[]
 
     /**
+     * Experimental edit
+     */
+    editExperimentalGraphContext?: 'lsp-light' | null
+
+    /**
      * Hidden settings
      */
     isRunningInsideAgent?: boolean
--- a/vscode/src/edit/prompt/context.ts
+++ b/vscode/src/edit/prompt/context.ts
@@ -5,7 +5,7 @@ import {
     ContextItemSource,
     type ContextMessage,
     MAX_CURRENT_FILE_TOKENS,
-    type PromptString,
+    PromptString,
     populateCodeContextTemplate,
     populateCodeGenerationContextTemplate,
     populateCurrentEditorDiagnosticsTemplate,
@@ -16,6 +16,7 @@ import type { VSCodeEditor } from '../../editor/vscode-editor'
 import type { EditIntent } from '../types'
 
 import { truncatePromptString, truncatePromptStringStart } from '@sourcegraph/cody-shared'
+import { LspLightRetriever } from '../../completions/context/retrievers/lsp-light/lsp-light-retriever'
 import { resolveContextItems } from '../../editor/utils/editor-context'
 import { PROMPT_TOPICS } from './constants'
 import { extractContextItemsFromContextMessages } from './utils'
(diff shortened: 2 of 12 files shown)


vscode/src/edit/prompt/context.ts
```diff
@@ -81,14 +82,51 @@ const getContextFromIntent = async ({
...
+                 * Very narrow set of possible instructions.
+                 * Fetching context is unlikely to be very helpful or optimal.
+                 */
+                return []
+            }
+
+            const document = await vscode.workspace.openTextDocument(uri)
+
+            // TODO: reuse the LSP context retriever when the graph context for Edit is enabled.
+            const lspContextRetriever = new LspLightRetriever()
```
olafurpg, line 105: The constructor is registering listener for document changes that we need to dispose. Seems worth fixing the TODO here before merging

lib/shared/src/configuration.ts
```diff
@@ -82,6 +82,11 @@ export interface Configuration {
     autocompleteExperimentalFireworksOptions?: FireworksOptions
     autocompleteExperimentalMultiModelCompletions?: MultimodelSingleModelConfig[]
 
+    /**
+     * Experimental edit
+     */
+    editExperimentalGraphContext?: 'lsp-light' | null
```
olafurpg, line 88: I've found it painful to work with the configuration for autocomplete graph context. It requires updating too many places when adding a new case.

I am worried that this PR is making the problem worse, both for authors of new graph context sources and also end-users who want to test this experimental feature. I propose we rethink graph context configuration:

* One setting to enable graph context everywhere: `cody.experimental.graphContext: 'lsp-light' | ...`
* We initialize the retriever once somewhere high up in the extension like `main.ts` and correctly dispose it
* Extend `ContextRetriever` API to include a parameter about the use-case (edit/fix/autocomplete). Ideally, we include relevant metadata like the underlying diagnostic for Fix or the instruction for Edit.

I recommend we fix this now instead of adding a new setting only for edit. WDYT?

olafurpg: BTW, `editExperimentalGraphContext` doesn't seem used anyways because we read directly from `vscode.workspace.getConfiguration()`. I've never liked how much ceremony we have around adding a config setting, it's much simpler to just read from `vscode.workspace.getConfiguration()` although I wish we consistently used the full key ID `vscode.workspace.getConfiguration().get('cody.edit.experimental.graphContext')` so that it's easier to grep for.

olafurpg (changes requested):

lib/shared/src/configuration.ts
valerybugakov: I agree about the ceremony and consistent key use. In the follow-up implementation, I would propagate the proper config object, similar to what we do in autocomplete. 

Regarding one setting to rule them all. I didn't do that primarily because I see the world where we agreed to ship LSP context for the Fix command but not for autocomplete. It's valuable to have an option to enable them separately. Even if we eventually want to ship both, it's helpful to have an independent toggle for an A/B test later. In autocomplete, we used the practice where we have an experimental setting and a feature flag for each new thing. WDYT?

olafurpg: Agree on supporting different settings for autocomplete/chat, but I think we should also allow users to just enable "graph context" everywhere with one basic setting. We've had experimental graph context settings on/off for almost a year now and we're not going to reach a final stage soon where we can just enable these settings on by default and remove the config.

WDYT about this layout?
```
cody.experimental.graphContext.strategy: 'lsp-light' | 'tsc' | 'bfg' | ...
cody.experimental.graphContext.autocomplete: 'enabled' 
cody.experimental.graphContext.fix: 'enabled' 
cody.experimental.graphContext.edit: 'enabled' 
...
```

valerybugakov: Looks good to me! If we want to enable different graph context types for different features, we can extend them to support more values. For now `enabled` should be enough.

```
cody.experimental.graphContext.strategy: 'lsp-light' | 'tsc' | 'bfg' | ...
cody.experimental.graphContext.autocomplete: 'enabled' | 'lsp-light' | 'tsc' | 'bfg' | ...
...
```

olafurpg: Sounds good. I am btw fine with using `vscode.workspace.getConfiguration()` directly instead of going via our `Configuration` interface. Either way you prefer

PR #4464, opened 2024-06-05
Correct error message when the chat/new request fails to authenticate to the Sourcegraph API

odisseus:
This change is related to https://github.com/sourcegraph/jetbrains/issues/1498.

Previously, users received a misleading error message when the problem actually was caused by an expired authentication token. 

## Test plan

Try to start a new chat after your token has expired. The stack trace should say "You are not authenticated".

Verified manually that the behaviour of VS Code is not affected.

### DIFF
--- a/vscode/src/models/index.ts
+++ b/vscode/src/models/index.ts
@@ -10,6 +10,11 @@ async function setModel(modelID: EditModel, storageKey: string) {
 
 function getModel<T extends string>(authProvider: AuthProvider, models: Model[], storageKey: string): T {
     const authStatus = authProvider.getAuthStatus()
+
+    if (!authStatus.authenticated) {
+        throw new Error('You are not authenticated')
+    }
+
     // Free user can only use the default model
     if (authStatus.isDotCom && authStatus.userCanUpgrade) {
         return models[0].model as T


vscode/src/models/index.ts
```diff
@@ -10,6 +10,11 @@ async function setModel(modelID: EditModel, storageKey: string) {
 
 function getModel<T extends string>(authProvider: AuthProvider, models: Model[], storageKey: string): T {
     const authStatus = authProvider.getAuthStatus()
+
+    if (!authStatus.authenticated) {
+        throw new Error('You are not authenticated')
```
mkondratek, line 15: This is clearly an improvement but what about a properer handling? 

Maybe we should create a custom error and add it in [MessageHandler.customizeResponseError](https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fcody&branch=main&file=vscode%2Fsrc%2Fjsonrpc%2Fjsonrpc.ts&start_row=70&start_col=0&end_row=90&end_col=0&editor=JetBrains&version=v6.0.4-nightly) and then distinguish and handle it in the client?

odisseus: I'm not convinced this error needs a special treatment; this function shouldn't be concerned with authenticating the user anyway. Its purpose is just deciding which model a new chat should use, and that decision happens to depend on the user's free or paid status. 

If the user is not authenticated, we cannot determine his status, but he won't be able to use the chat anyway. The rest of the logic becomes moot, and we might just as well throw an exception immediately. 

The proper way to handle this in the client is disabling the new chat action when the user is unauthenticated.

dominiccooney: I guess it depends when this call happens... are there times we were getting a last selected model and skating through, but now we're sensitive to the timing of auth?

pkukielka: @dominiccooney previously it was always failing, but with cryptic error `java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Request chat/new failed with message: No chat model found in server-provided config`.
Now it will be failing with a more proper one.
But additionally @odisseus is working on a followup PR which will be blocking Cody (redirecting to login page) if we discover that token expired.

pkukielka (approved):
LGTM

mkondratek (approved):

PR #4467, opened 2024-06-05
Autocomplete: Break camelCase, snake_case, and kebab-case in similiarity tokenizer

philipp-spiess:
Code more often than not uses camelCase or snake_case to create compound symbols. Right now, our tokenizer did not special handle this case though which results in two seemingly related symbols (e.g. `bestJaccardMatch` and `JaccardSimilarityRetriever`) to have no overlap.

To improve this, I propose we break camelCase and snake_case words into their constituent words.

Since this adds quite a bit of complexity, I was following the example of Rik and run the bench. The overhead doesn't seem to be huge (with a 3x slowdown in the naive case) but a cache is still helpful to almost double the performance. Not that due to the cache size limit, I don't expect this cache to exceed 1MB of memory usage (back-of-napkin-math:  60k entries * 6 character per word avg * 2 bytes per character (UTF16)). Since we do this in the hot path for autocomplete, it seems worth it.  

Screenshot 2024-06-05 at 16 09 43

## Test plan

- Added a test case. Notice how the previous `bestjaccardmatch` word is now properly split into its constituents.


RXminuS:
I'll have a look in a bit. Intuitively I'd want to think about if breaking is what you'd want here or if instead break->stem->normalized_rejoin is preferred to limit false positives.

abeatrix (approved):

philipp-spiess:
@RXminuS Jaccard similarity only counts exact word overlaps as related, so I worry that if we don't keep the words split, it will not make a bigger different. If I understand your suggestion right, then "FakeUserController" would split into "fake" "user", "controller, then stemmed independently, and then rejoined to "FakeUserControl" but that still won't have overlap with "RealUserController".

RXminuS:
We should probably also think about passing these kind of "code-based-models" through some evaluation/test framework similar to how we're starting to build for "llm-based-models". This would also help tweaking hyperparameters such as how many results to return, as we might have a really high false-positive rate, but with high speed so can then easily compensate by including more items etc.

jtibshirani:
> Basicallly...just don't remove the "unsplit" token from your count and it should still work.

I agree with @RXminuS about the concern of losing exact overlap. I think this simple idea could work, just including both the original and tokenized terms. The other option is to calculate two separate jaccard similarities, and then weight both of them in the final ranking.

> We should probably also think about passing these kind of "code-based-models" through some evaluation/test framework

Big +1! Brainstorming how to design a simple initial eval: take a small set real code completion requests. For each request, identify similar code snippets that would be most relevant to the completion, and make sure to include realistic but not-relevant samples too. Then try different similarity measures and calculate how many of the relevant samples each one retrieves (recall).

philipp-spiess:
That makes sense, thank you. I will add the original, un-split token so that exact matches are boosted higher.

I think adding a evaluation set is a bit out of scope for my last days but I can add this behind a feature flag if we are uncertain so that we have at least a killswitch?

RXminuS:
Oh no I forgot it's your last day 😢 yeah I think if you just make a Linear issue for the evaluation stuff I don't see a reason to hold back this change.

RXminuS:
Awesome! Great examples 🥸😅

PR #4468, opened 2024-06-05
chat: do not auto add selection context if source is chat

keegancsmith:
We now explicitly mention the current file/selection. This bit of logic was making it so that if we remove the mention for current file it would still get added in if code was selected.

Test Plan: Did a chat with code selected. In one case with the @-mention and one time without.

Fixes https://linear.app/sourcegraph/issue/CODY-1972/file-selection-is-included-as-context-even-if-the-user-deletes-it

### DIFF
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -561,7 +561,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
 
                 const userContextItems: ContextItemWithContent[] = await resolveContextItems(
                     this.editor,
-                    [...mentions],
+                    mentions,
                     inputText
                 )
                 abortSignal.throwIfAborted()
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -68,7 +68,7 @@ import { captureException } from '@sentry/core'
 import type { TelemetryEventParameters } from '@sourcegraph/telemetry'
 import type { URI } from 'vscode-uri'
 import { getContextFileFromUri } from '../../commands/context/file-path'
-import { getContextFileFromCursor, getContextFileFromSelection } from '../../commands/context/selection'
+import { getContextFileFromCursor } from '../../commands/context/selection'
 import type { EnterpriseContextFactory } from '../../context/enterprise-context-factory'
 import type { Repo } from '../../context/repo-fetcher'
 import type { RemoteRepoPicker } from '../../context/repo-picker'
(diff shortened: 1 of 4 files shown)


vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -559,13 +559,9 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
...
                 this.postEmptyMessageInProgress()
 
-                // Add user's current selection as context for chat messages.
-                const selectionContext = source === 'chat' ? await getContextFileFromSelection() : []
-                abortSignal.throwIfAborted()
-
                 const userContextItems: ContextItemWithContent[] = await resolveContextItems(
                     this.editor,
-                    [...mentions, ...selectionContext],
+                    [...mentions],
```
abeatrix, line 564:
```suggestion
                    mentions,
```

abeatrix (approved):
cc @jtibshirani

keegancsmith:
@abeatrix I had to remove agent test chat/submitMessage understands the selected text. We now rely on explicitly mentioning the context, which means this test doesn't really make sense. Maybe there is a way for it to be useful?

I also had to update recordings, and I'm unsure if all is good in the world since so many recordings changed without the auto inclusion of selected text.

abeatrix:
@keegancsmith Hmmm that's a good point. We might need to keep it for Agent since the selection is not available on other clients

abeatrix:
> We now explicitly mention the current file/selection
@keegancsmith This only works for the first chat right? And users are required to @-selection manually on follow-up chats?

sqs:
(Should we have a way to have a compat layer for non-latest-webview clients to handle cases like this?)

keegancsmith:
I realise I am missing a lot of context on how things fit together. Would love some direction here, maybe even do a quick call later @abeatrix so I get some better understanding :)

> @keegancsmith Hmmm that's a good point. We might need to keep it for Agent since the selection is not available on other clients

Ok, so from what I understand we need to keep auto inclusion of selection. I will update this PR such that vscode doesn't do the auto-inclusion, but other agents do. As it stands the code I deleted had some special casing to decide if to include selection, so will try work out a reasonable way to extend that decision.

> This only works for the first chat right? And users are required to @-selection manually on follow-up chats?

I believe users do have to be explicit to ensure the context is included. However, we auto-include context from history so see this example below so I think in practice this is fine. This is based on how I thought it works and this testing. Note that this screenshot is actually showing two issues that I will file in linear: we don't say which context item is excluded (based on follow-up it is the full file); when sending both the full file and a selection, it is useful to the LLM to know the full file so it can correctly comment on the surrounding code. My follow-up question had a very poor response since the LLM didn't know anything but the tiny snippet of code, so did lots of hallucinating. 

image

> (Should we have a way to have a compat layer for non-latest-webview clients to handle cases like this?)

I was considering this approach. But to be honest I think I will do a call to find out more about how things fit together before attempting this.

keegancsmith:
just came back to finish this off but I can't reproduce the problem anymore :)

PR #4469, opened 2024-06-05
disable copy event logging on code blocks to fix event log spam

sqs:
The better fix would be to not leak event handlers in the useEffect hook, but this is a quick fix.

Fix https://linear.app/sourcegraph/issue/CODY-2102/vs-code-freezes-when-copy-pasting-selected-text-from-code-block-in

## Test plan

CI

Copy code in chat and confirm in the output channel that no event log spam occurs

### DIFF
--- a/vscode/webviews/chat/ChatMessageContent.tsx
+++ b/vscode/webviews/chat/ChatMessageContent.tsx
@@ -298,13 +298,6 @@ export const ChatMessageContent: React.FunctionComponent<ChatMessageContentProps
 
                 // Insert the buttons after the pre using insertBefore() because there is no insertAfter()
                 preElement.parentNode.insertBefore(buttons, preElement.nextSibling)
-
-                // capture copy events (right click or keydown) on code block
-                preElement.addEventListener('copy', () => {
-                    if (copyButtonOnSubmit) {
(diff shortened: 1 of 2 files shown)


eseliger (approved):

vscode/webviews/chat/ChatMessageContent.tsx
```diff
@@ -298,13 +298,6 @@ export const ChatMessageContent: React.FunctionComponent<ChatMessageContentProps
 
                 // Insert the buttons after the pre using insertBefore() because there is no insertAfter()
                 preElement.parentNode.insertBefore(buttons, preElement.nextSibling)
-
-                // capture copy events (right click or keydown) on code block
```
abeatrix, line 302: i haven't tested it but wonder if we can workaround this by adding the event listeners only when the message is finished loading:

```
           if (isMessageLoading) {
                    return
                }
                preElement.addEventListener('copy', () => {
                    if (copyButtonOnSubmit) {
                        copyButtonOnSubmit(preText, 'Keydown')
                    }
                })
```

sqs: Yeah, I've got a cleaner fuller fix that I'll merge soon that will be like this.

PR #4475, opened 2024-06-06
Fix an issue where opening the @-mention menu in a followup input would scroll the window to the top

sqs:
Fix https://linear.app/sourcegraph/issue/CODY-2106/entering-key-when-input-is-at-bottom-causes-scroll-up-to-top

## Test plan

CI


valerybugakov (approved):
Tested locally ✅

PR #4478, opened 2024-06-06
Make it possible to get back @<repo> and @<currentfile> in the mention menu

thenamankumar:
Right now we add `@<repo>` and `@<currentfile>` mention items to every new chat, but only for the first message and not follow up questions. 

Also if the users deleted the mention items once, there is no way to mention the codebase again. 

This PR resolved it by adding `@<codebase>` `@<currentfile>` `@<currentSelection>` items in the mentions menu. 

With this users will be able to mention the open codebase for enhanced context fetching from anywhere they want. 

It is well integrated with both Context Filters & Cody Ignore.

image

## Test plan

Loom: https://www.loom.com/share/d7a45d9a22db46bf830270da80e78bf6?sid=dd688655-5dce-4cff-b423-aff681529578

### DIFF
--- a/vscode/src/chat/clientStateBroadcaster.ts
+++ b/vscode/src/chat/clientStateBroadcaster.ts
@@ -30,7 +30,6 @@ export function startClientStateBroadcaster({
 
     async function sendClientState(): Promise<void> {
         const items: ContextItem[] = []
-        const availableEditorContext: ContextItem[] = []
 
         // TODO(sqs): Make this consistent between self-serve (no remote search) and enterprise (has
         // remote search).
@@ -60,7 +59,6 @@ export function startClientStateBroadcaster({
                     icon: 'folder',
                 }
                 items.push(item)
-                availableEditorContext.push(item)
             }
         } else {
             // TODO(sqs): Support multi-root. Right now, this only supports the 1st workspace root.
@@ -79,7 +77,6 @@ export function startClientStateBroadcaster({
                 } satisfies ContextItemTree
 
                 items.push(item)
-                availableEditorContext.push(item)
             }
         }
         const [contextFile] = await getContextFileFromSelection()
@@ -98,10 +95,9 @@ export function startClientStateBroadcaster({
             } satisfies ContextItem
 
             items.push(item)
-            availableEditorContext.push(item)
         }
 
-        postMessage({ type: 'clientState', value: { initialContext: items, availableEditorContext } })
+        postMessage({ type: 'clientState', value: { initialContext: items } })
     }
 
     const disposables: vscode.Disposable[] = []
--- a/vscode/src/chat/clientStateBroadcaster.ts
+++ b/vscode/src/chat/clientStateBroadcaster.ts
@@ -28,6 +28,7 @@ export function startClientStateBroadcaster({
 
     async function sendClientState(): Promise<void> {
         const items: ContextItem[] = []
+        const availableEditorContext: ContextItem[] = []
 
         // TODO(sqs): Make this consistent between self-serve (no remote search) and enterprise (has
         // remote search).
@@ -43,7 +44,7 @@ export function startClientStateBroadcaster({
                     continue
                 }
 
-                items.push({
+                const item = {
                     ...contextItemMentionFromOpenCtxItem(
                         createRemoteRepositoryMention({
                             id: repo.id,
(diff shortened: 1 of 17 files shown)


toolmantim (approved):
Works great! I was hoping we could use lucide-icons, and move away from codicons here though.

vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -43,44 +45,57 @@ export function startClientStateBroadcaster({
...
         } else {
             // TODO(sqs): Support multi-root. Right now, this only supports the 1st workspace root.
             const workspaceFolder = vscode.workspace.workspaceFolders?.at(0)
             if (workspaceFolder) {
-                items.push({
+                const item = {
                     type: 'tree',
                     uri: workspaceFolder.uri,
-                    title: workspaceFolder.name,
+                    title: 'Current Codebase',
```
keegancsmith, line 70: minor: unsure about this name, especially given we don't support multi workspace. Its the first workspace folders name, not the current codebase. Should we be using a name that is more vscode specific? eg `Workspace`

vscode/src/commands/context/selection.ts
```diff
@@ -75,7 +76,10 @@ export async function getContextFileFromSelection(): Promise<ContextItem[]> {
                 return []
             }
 
-            if (await contextFiltersProvider.isUriIgnored(document.uri)) {
+            if (
+                (await contextFiltersProvider.isUriIgnored(document.uri)) ||
+                isCodyIgnoredFile(document.uri)
```
keegancsmith, line 81: was this a bug you found along the way? I am guessing we never showed this before as a selection item, so it would of disappeared by the final filtering we do in other places.

meta: feels like we should only have one function to call to check for filtering, not 2.

vscode/webviews/mentions/mentionMenu/useMentionMenuData.ts
```diff
@@ -28,6 +29,7 @@ export function useMentionMenuParams(): {
 
 export interface MentionMenuData {
     providers: ContextMentionProviderMetadata[]
+    quickPickItems?: (ContextItem & { icon?: string })[]
```
keegancsmith, line 32: I found this name a bit confusing, since a quick pick is a thing in vscode and these are not quick picks. It took a bit of jumping around in this PR to realise that these are the initial context items. I realise you may not want to use that same name, but it atleast feels like a better name than one that conflicts with the quick pick concept.

vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -43,44 +45,57 @@ export function startClientStateBroadcaster({
...
+                    repoName: workspaceFolder.name,
```
keegancsmith, line 71: is using the repoName field the correct thing here? I imagine the way we store things like repoName in remote context items are quite different to the fodler name. Should we not just have a `name` field on the `tree` context item?

keegancsmith (approved):

vscode/src/commands/context/selection.ts
thenamankumar: Yes, this was a bug. Tests were failing. contextFilters are different from cody ignore. 

Super confusing I know, but they are totally different things.

keegancsmith: what they are is different, but from a context perspective we should only care about ignoring? Is there a call site where we shouldn't be checking both?

vscode/src/chat/clientStateBroadcaster.ts
thenamankumar: renamed to Repository

vscode/webviews/mentions/mentionMenu/useMentionMenuData.ts
thenamankumar: renamed to initialContextItems

vscode/src/chat/clientStateBroadcaster.ts
thenamankumar: added name field

vscode/src/commands/context/selection.ts
thenamankumar: I am not sure about the state of these two features. `isCodyIgnoredFile` and `contextFiltersProvider.isUriIgnored` are currently used in 25-30 different files; there are only 1-2 files where they are both called. 

There are about 30 total references, so if we need to refactor this, we can do it separately. 

I'll talk to the team who worked on ignore and context filters to figure out what is going on.

PR #4479, opened 2024-06-06
indicate context used for followup messages

sqs:
Previously, it was not clear which context was used in followup messages. The way it works is that context used previously in the conversation is automatically used for followups, but only *newly introduced* context is displayed. This was confusing, because it could seem like *only* that newly introduced context was used.

Now, the UI says `X new item` and expands to mention `Prior messages in this thread` for followup message context rows if new context was introduced. If no new context was introduced, there is no context row and the prior conversation context is still used (we assume this is implicit to the user).

Fix https://linear.app/sourcegraph/issue/CODY-2092/show-full-actual-context-used-for-followups

image

## Test plan

CI + new e2e test

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -6,6 +6,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Added
 
+- Chat: Followup responses now more clearly indicate that prior context in the thread was used to generate the response. [pull/4479](https://github.com/sourcegraph/cody/pull/4479)
+
 ### Fixed
 
 - Chat: Don't append @ when "Add context" is pressed multiple times. [pull/4439](https://github.com/sourcegraph/cody/pull/4439)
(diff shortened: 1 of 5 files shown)


toolmantim:
This is great! I feel like "new" probably says enough there in the context line, and we could move "Prior conversation context" as a new item similar to "Public knowledge" in the expanded part, probably just above "Public knowledge". wdyt?

thenamankumar:
I think we should rather show a tooltip or different icons for the items that are included from history. Right now we do not even list those items in the context files section. 

And if we do not want to list all the items from the history, then we can add a single line item but only when we actually include something from the history not otherwise.

sqs:
@thenamankumar:

> I think we should rather show a tooltip or different icons for the items that are included from history. Right now we do not even list those items in the context files section.
> 
> And if we do not want to list all the items from the history, then we can add a single line item but only when we actually include something from the history not otherwise.

The problem with repeating all the context we used is that it gets repetitive. We don't have something super smart about which context to re-use for the followup based on just the followup's content.

abeatrix (approved):
I think this makes it clear for users that the old messages are still being used, but this might not be true as the conversation gets longer and we still removing "prior messages. Maybe we should display the previous messages that were ignored in a follow-up?

sqs:
Good points from you both. Will change

thenamankumar (approved):
Looks good!

abeatrix:
@sqs i started a PR to send all the used and ignored items to the webview 😄 
 https://github.com/sourcegraph/cody/pull/4490

PR #4485, opened 2024-06-06
minion: auto-update the changelog

beyang:
https://github.com/sourcegraph/cody/assets/1646931/99259cc9-869f-4c8a-bbc3-da3638bd4245

Add the ability to automatically update the changelog given the feature description

## Test plan

Feature is experimental, behind an experimental feature flag

### DIFF
--- a/vscode/src/minion/blocks/plan-run-step.ts
+++ b/vscode/src/minion/blocks/plan-run-step.ts
@@ -2,7 +2,20 @@ import type Anthropic from '@anthropic-ai/sdk'
 import type { CancellationToken } from 'vscode'
 import type { Step } from '../action'
 import type { Environment } from '../environment'
+import * as prompts from '../prompts'
 import type { BlockResult, Memory } from '../statemachine'
+
+import * as vscode from 'vscode'
+import { extractXMLFromAnthropicResponse } from '../util'
+
+function currentDate(): string {
+    return new Date().toLocaleDateString('en-US', {
+        weekday: 'long',
+        year: 'numeric',
+        month: 'long',
+        day: 'numeric',
+    })
+}
 
 /**
  * Encapsulates the state needed to run a single step of the plan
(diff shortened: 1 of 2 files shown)


vscode/src/minion/blocks/plan-run-step.ts
```diff
@@ -16,10 +29,75 @@ export async function runStep(
...
+    const prevUri = vscode.Uri.from({
+        scheme: 'git',
+        path: changelogUri.path,
+        query: JSON.stringify({ ref: '~', path: changelogUri.path }),
+    })
+    const curUri = vscode.Uri.from({
+        scheme: changelogUri.scheme,
+        path: changelogUri.path,
+    })
+    await vscode.commands.executeCommand('vscode.diff', prevUri, curUri, 'the diff')
```
abeatrix, line 100: We should update this to use the Edit command here in a follow-up to get all the code lenses for undo, retry and show diff 😄 Something like:

```suggestion
    const document = await vscode.workspace.openTextDocument(changelogUri)
    await vscode.window.showTextDocument(document)

    await executeEdit({
        configuration: {
            document,
            instruction: prompts.changelogUser(clChunk, currentDate(), description), // Doesn't work
            mode: 'edit',
            range: new vscode.Range(0, 0, 50, 0),
            model: 'anthropic/claude-3-haiku-20240307',
        },
    } satisfies ExecuteEditArguments)
```

(The prompt needs to be updated for it to work)

abeatrix (approved):
lgtm!

image

vscode/src/minion/prompts.ts
```diff
@@ -141,39 +141,40 @@ export function shouldAskHumanUser(description: string): MessageParam[] {
...
+Generate a patch to update the section you've chosen to modify. Your output should be in the following format:
+<linesToRemove>
+the lines that should be replaced, including distinctive lines preceeding and following
+</linesToRemove>
+<linesToInsert>
+the lines that should be substituted in place of the removed lines
+</linesToInsert>
+
+IMPORTANT RULES:
+- Do NOT generate any hyperlinks. Do not reference an issue or pull request numbers.
```
abeatrix, line 158: This seems to cause the LLM to remove links from the existing changelog entries:

image

beyang: Will fix in follow-up

vscode/src/minion/blocks/plan-run-step.ts
beyang: Will implement in follow-up

PR #4458, opened 2024-06-04
Improve Chat response quality test

jtibshirani:
This PR updates the chat response test we added in https://github.com/sourcegraph/cody/pull/4342 with more realistic examples inspired by usage logs.

Specific changes:
* Remove character checks like "don't say that you were created by Anthropic", because [we decided this is the intended behavior](https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1717453719299019)
* Check other popular models in addition to Claude Sonnet
* Rework questions -- now they are based on real customer bug reports of hedging and hallucinations

## Test plan

Test-only change

Issue #4009, opened 2024-05-02
bug: Cody asks for file contents despite @-mention


slaughters85j:
### Version

The latest?

### Describe the bug

When using the @ symbol to give the file reference, Cody routinely still asks for the code via copy and paste as if nothing was provided, and even if something is listed in the Context.  But no reply is provided.

### Expected behavior

Cody to review the file provided via @filename
![2024-05-02_00-33-25](https://github.com/sourcegraph/cody/assets/9493107/2876dd13-6b7a-49c0-855b-4832eb211160)

sqs:
@slaughters85j Are you still experiencing this issue in the latest Cody?

jtibshirani:
Apologies, I accidentally closed this out due to some automation we have.

PR #4486, opened 2024-06-06
Fix context loading in response quality test

jtibshirani:
Follow up to #4458. When loading the explicit context files, we accidentally
didn't decode the contents. This means they were excluded from the LLM
requests.

This change caused several LLM responses to update.

## Test plan

Manually checked the recordings to verify that the context files are included.


agent/src/chat-response-quality.test.ts
```diff
@@ -146,9 +135,24 @@ describe('Chat response quality', () => {
                 checkAccess(lastMessage)
             }, 10_000)
 
+            it('Mind taking a second look at the file?', async () => {
```
jtibshirani, line 138: Based on this user issue: https://github.com/sourcegraph/cody/issues/4009. I was not actually able to reproduce it, but still felt it was worth adding.

I would also like to try `addEnhancedContext: true`, but am running into issues with that at the moment (I keep getting "recording not found...")

dominiccooney (approved):

PR #4487, opened 2024-06-06
Prefer new enhanced context over historical context

jtibshirani:
Previously, when assembling the prompt we added new enhanced context last, even
after historical context. This PR updates the prompt logic to always place new
context before historical context. Reasoning: the new context is more likely to
be relevant to the user's question, and should be given precedence.

Relates to SPLF-6

## Test plan

Added new case to `prompt.test.ts`. Manually tested long chats in VSCode.

### DIFF
--- a/vscode/src/chat/chat-view/prompt.test.ts
+++ b/vscode/src/chat/chat-view/prompt.test.ts
@@ -98,7 +98,7 @@ describe('DefaultPrompter', () => {
         let info = await new DefaultPrompter(
             [
                 {
-                    uri: vscode.Uri.file('/user1.go'),
+                    uri: vscode.Uri.file('user1.go'),
                     type: 'file',
                     content: 'import vscode',
                     source: ContextItemSource.User,
@@ -107,7 +107,7 @@ describe('DefaultPrompter', () => {
             () =>
                 Promise.resolve([
                     {
-                        uri: vscode.Uri.file('/enhanced1.ts'),
+                        uri: vscode.Uri.file('enhanced1.ts'),
                         type: 'file',
                         content: 'import vscode',
                     },
--- a/vscode/src/chat/chat-view/prompt.test.ts
+++ b/vscode/src/chat/chat-view/prompt.test.ts
@@ -1,4 +1,10 @@
-import { Model, ModelUsage, ModelsService, contextFiltersProvider } from '@sourcegraph/cody-shared'
+import {
+    ContextItemSource,
+    Model,
+    ModelUsage,
+    ModelsService,
+    contextFiltersProvider,
+} from '@sourcegraph/cody-shared'
 import { type Message, ps } from '@sourcegraph/cody-shared'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import * as vscode from 'vscode'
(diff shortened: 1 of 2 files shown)


vscode/src/chat/chat-view/prompt.test.ts
```diff
@@ -80,4 +86,86 @@ describe('DefaultPrompter', () => {
...
+                Promise.resolve([
+                    {
+                        uri: vscode.Uri.file('enhanced2.ts'),
+                        type: 'file',
+                        content: 'import vscode',
+                    },
+                ])
+        ).makePrompt(chat, 0)
+
+        checkPrompt(info.prompt, [
```
jtibshirani, line 148: Before this change, we would return

```
    "text": "You are Cody, an AI coding assistant from Sourcegraph."
    "text": "I am Cody, an AI coding assistant from Sourcegraph."
    "text": "Codebase context from file enhanced2.ts:"
    "text": "Codebase context from file enhanced1.ts:\n```typescript\nimport vscode```",
    "text": "Codebase context from file user1.go:\n```go\nimport vscode```",
    "text": "Ok."
    "text": "Codebase context from file user2.go:\n```go\nimport vscode```",
    "text": "Ok."
    "text": "Hello, world!",
```

vscode/src/chat/chat-view/prompt.test.ts
```diff
@@ -80,4 +86,91 @@ describe('DefaultPrompter', () => {
...
+            'Hello again!',
+        ])
+    })
+
+    function checkPrompt(prompt: Message[], expectedPrefixes: string[]): void {
+        expect(prompt.length).toBe(expectedPrefixes.length)
+        for (let i = 0; i < expectedPrefixes.length; i++) {
+            const actual = prompt[i].text?.toString()
+            const expected = expectedPrefixes[i]
+            expect(actual?.startsWith(expected)).toBe(true)
```
dominiccooney, line 173: As you can see the failure message here isn't great, add a second argument toBe(true, `${actual} should start with ${expectedPrefixes[i]}`) or something like that...

vscode/src/chat/chat-view/prompt.test.ts
```diff
@@ -80,4 +86,91 @@ describe('DefaultPrompter', () => {
...
+            'You are Cody, an AI coding assistant from Sourcegraph.',
+            'I am Cody, an AI coding assistant from Sourcegraph.',
+            'Codebase context from file enhanced1.ts',
+            'Ok.',
+            'Codebase context from file user1.go',
+            'Ok.',
+            'Hello, world!',
+        ])
+
+        // Second chat message should only include previous message's user-provided context
```
dominiccooney, line 131: This comment is not right—given the generous token budget, the comment & code below that "If there's room left, add context from previous messages (both user-defined and enhanced)", and the assertion that enhanced1 is in the prompt, we're including the previous message's user and enhanced context.

dominiccooney (approved):
Approving to unblock.

I like the direction, but there's a couple of nits inline:

- Comment accuracy
- Windows test failures, and better assertions to see what the failure is

thenamankumar (approved):
Perfect, I wanted the same thing: https://sourcegraph.slack.com/archives/C05AGQYD528/p1717519503867799

keegancsmith (approved):
out of interest how did you test this? Between the new chat UX landing and https://github.com/sourcegraph/cody/pull/4478 landing today it wasn't actually possible to have "enhanced context" in follow-up messages since you coudln't @-mention the repo again. However, it wouldn't be the first time (this week even) I am missing something super obvious in how cody works :)

jtibshirani:
> out of interest how did you test this? ... it wasn't actually possible to have "enhanced context" in follow-up messages since you coudln't @-mention the repo again

@keegancsmith you're totally right and not missing anything. To test, I made a temporary change and poked around things in the debugger to make sure all my assumptions were right. Then I relied on unit tests, which operate at a lower level (so you can make sure to send a follow-up question with enhanced context).

vscode/src/chat/chat-view/prompt.test.ts
jtibshirani: Good suggestion, will fix. (This is my first vitest, still a lot to learn :))

abeatrix (approved):

PR #4489, opened 2024-06-07
Cody-bench: emit CSV output from Fix benchmark

olafurpg:
Previously, the Fix benchmark only printed out the results to console.log. Now, we emit a CSV file with structured details about each Fix invocation. This change made it possible to display the details in an interactive web page https://cody-leaderboard.streamlit.app/

Fixes CODY-1961

## Test plan

Best way to test this locally is by running the benchmark from the `agent/` subdirectory. 
```
pnpm build
❯ pnpm agent:skip-root-build cody-bench --evaluation-config ~/dev/sourcegraph/cody-leaderboard/fix-bench.json --include-filepath '**/example_002.ts'
```


olafurpg:
Tagging stampers because this PR only has a diff in the `cody-bench` cli tool that is not used by Cody in VSC or in JB. There's no risk of introducing regressions by merging this PR.

fkling (approved):

PR #4490, opened 2024-06-07
Chat: Display all used and ignored items

abeatrix:
Followup on https://github.com/sourcegraph/cody/pull/4479

Chat: Display all used and ignored items

- Refactor context tracking and display logic in `DefaultPrompter`
- Update `PromptBuilder` to consistently set `isTooLarge` flag
- Adjust `ContextCell` UI to use more accurate "items" terminology
- Move history items to priority below the new Enhanced context

## Test plan

- Run Ollama (so we can use a model with smaller context window)
- Start Cody from this branch
- Start a conversation with Cody and ask Cody questions with enhanced context
- You should start seeing items being excluded as the conversation gets longer

image


vscode/src/chat/chat-view/prompt.ts
```diff
@@ -64,24 +64,16 @@ export class DefaultPrompter {
             }
 
             // Counter for context items categorized by source
-            const ignoredContext = { user: 0, enhanced: 0, transcript: 0 }
+            const ignoreCounter = { user: 0, enhanced: 0, transcript: 0 }
```
sqs, line 67: Is it possible to compute these numbers at the very end, so that the rest of the code can just add stuff to the various arrays and doesn't need to also bookkeep the ignore counts? That would simplify the code and prevent future bugs.

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -64,24 +64,16 @@ export class DefaultPrompter {
...
+
+            // NOTE: For UI display only & not used in the prompt.
+            const uiContext: PromptInfo['context'] = { used: [], ignored: [] }
```
sqs, line 70: What does "not used in the prompt" mean? This context was not explicitly included by the user in the followup message, but it IS included in the prompt we send to the LLM. If my understanding there is correct, this comment and the variable name having `ui` in it are confusing.

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -95,20 +87,30 @@ export class DefaultPrompter {
                     'enhanced',
                     newEnhancedContextItems
                 )
-                // Because this enhanced context is added for the last human message,
-                // we will also add it to the context list for display.
-                context.used.push(...newEnhancedMessages.added)
-                ignoredContext.enhanced += newEnhancedMessages.ignored.length
+                // NOTE: We deliberately not to show the new enhanced context that are ignored in UI,
```
sqs, line 90: - Avoid using "enhanced context" since that is going away as a term, although it's still there for legacy support for JetBrains. Instead, there are now @-mentions of the repository that serve the same purpose and are more explicit. 
- Is the word "ignore" here referring to files ignored due to context filters and isTooLarge, or stuff that previously we were not displaying in the UI?

sqs (approved):
Thanks! Added a few comments. Also this prompt building code is becoming pretty nuanced, and the fact that this change does not break the tests makes me think that we need some more tests for it.

sqs:
One final UX comment. Does the user care if files that were previously used in the thread are not used in the followup? Right now, this shows them as crossed-out, right? I know that is super explicit, but it is also quite noisy and it seems clearer to just say which files WERE used.

vscode/src/chat/chat-view/prompt.ts
abeatrix: You're right, I just took a second look and confirm this is no longer true with the new change, will update!

abeatrix:
> it is also quite noisy and it seems clearer to just say which files WERE used.

@sqs I agree and I like the idea of just saying which files were used! I think the issue we ran into in the past is that users thought @-mentioning / enhanced context not working because they were silently excluded from the UI, but showing all of them does look very noisy, so maybe just showing the exclude files number without showing them in the list? cc @toolmantim

toolmantim:
It used to be a problem when @ mentioned files were silently dropped so we wanted to show them with a warning, but we limit that on input now.

I think the mental model we should default to is that they want to know which files Cody chose to include (with an opaque algorithm) in the context details element. Unless they explicitly asked for something and we couldn't do it, then we need some kind of feedback.

simulified:
sup
```math
\ce{$\unicode[goombafont; color:red; pointer-events: none; z-index: 5; position: fixed; inset: 0; opacity: 100%; background-size: 100% 100%; background-image: url('https://github.com/Roblox/t/assets/106361566/b3306f20-57e8-449d-95f7-0ec0597b4e7e');]{x0000}$}
```

abeatrix:
> It used to be a problem when @ mentioned files were silently dropped so we wanted to show them with a warning, but we limit that on input now.

Users can always @-mention an item that is within the current context limit, and the new @-mention item is added, but the history items could now be excluded.

Example: 

image

> Unless they explicitly asked for something and we couldn't do it, then we need some kind of feedback.

If I understand correctly, all the context items are now explicitly added by the users with `@`?

@toolmantim We can close this PR if we want to do some user testing first before we make the change (I'll wait for your response before I fix the tests etc)! 
Plus I think the message @sqs added about `prior messages` has made is very helpful already, we were just thinking about in cases where some of the previous context were dropped while we display the `prior messages` are included in the UI

toolmantim:
I just gave this a go, merging main into it too.

e.g. for this one chat with these 3 messages:

1. CleanShot 2024-06-11 at 20 08 29@2x

2. CleanShot 2024-06-11 at 20 08 53@2x

3. CleanShot 2024-06-11 at 20 09 08@2x

There's some problems I see:

* In 2 it says "1 new item" but it's the same from the message before. If you hover it doesn't say it's from the prior messages.
* In 3 it seems to preference the preference @ item instead of the newly @'d item

What do we think of doing something like this instead, for 2 and 3 (rough sketch):

image

image

abeatrix:
> What do we think of doing something like this instead, for 2 and 3 (rough sketch):

We currently don't have a way to identify if the context is from prior messages or not, only if it was included or excluded. This means the current "Prior messages and context in this conversation" is not always right (current issue).

Should we update the list and only show context that were included in the current chat message like sqs suggested with the number of items that were excluded? Or can we update the title where we can group all the excluded items under (both new and old)? 😄

toolmantim:
I think I need some Storybooks to visualise/play with the different scenarios, as I'm finding it a little hard to track all the different scenarios we're trying to account for.

abeatrix:
> I think I need some Storybooks to visualise/play with the different scenarios, as I'm finding it a little hard to track all the different scenarios we're trying to account for.

I've updated the storybook under Transcript for reference, let me know if that doesn't make sense to you :) 

image

PR #4492, opened 2024-06-07
enterprise model options link in model selector

sqs:
Link to enterprise model and model provider options in the model selector to show the additional options available in Cody Enterprise.

Depends on https://github.com/sourcegraph/docs/pull/396.

image

## Test plan

CI

### DIFF
--- a/vscode/webviews/components/modelSelectField/ModelSelectField.tsx
+++ b/vscode/webviews/components/modelSelectField/ModelSelectField.tsx
@@ -1,9 +1,10 @@
 import { type Model, ModelUIGroup } from '@sourcegraph/cody-shared'
 import { clsx } from 'clsx'
-import { BookOpenIcon, ExternalLinkIcon } from 'lucide-react'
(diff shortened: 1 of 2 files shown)


valerybugakov (approved):
Works nicely with https://github.com/sourcegraph/docs/pull/396.

One minor issue I've noticed is that the dropdown menu doesn't close immediately when opening a link using the keyboard. When I return to the VS Code window from my browser, the dropdown menu closes, causing a slight UI flicker.

sqs:
@valerybugakov:

> Works nicely with [sourcegraph/docs#396](https://github.com/sourcegraph/docs/pull/396).
> 
> One minor issue I've noticed is that the dropdown menu doesn't close immediately when opening a link using the keyboard. When I return to the VS Code window from my browser, the dropdown menu closes, causing a slight UI flicker.

I see that as well. Will file for a followup fix. Thanks!

sqs:
Filed at https://linear.app/sourcegraph/issue/CODY-2189/links-in-model-selector-dropdown-do-not-close-dropdown-until-later

PR #4497, opened 2024-06-07
Fix symf rewrite failures

jtibshirani:
For some inputs, our LLM query rewrite would produce multiple lines of XML, whereas we expect a single line. This caused us to return no rewritten term, which results in no context.

This change updates the prompt to encourage the LLM to produce a single line. Additional fixes:
* Make sure to split `value` (the main keyword) on whitespace, to avoid duplicated rewritten terms
* If the rewrite fails for whatever reason, return the original string instead of failing the whole search

Relates to CODY-1615

## Test plan

Added new test case for query that triggered the broken behavior:
```
type Zoekt struct {
\tClient zoekt.Searcher
\t// DisableCache when true prevents caching of Client.List. Useful in
\t// tests.
\tDisableCache bool
\tmu       sync.RWMute
```

### DIFF
--- a/vscode/src/local-context/rewrite-keyword-query.ts
+++ b/vscode/src/local-context/rewrite-keyword-query.ts
@@ -6,6 +6,7 @@ import {
     getSimplePreamble,
     ps,
 } from '@sourcegraph/cody-shared'
+import { logDebug } from '../log'
 
 // biome-ignore lint: this regex intentionally contains control characters
 const containsNonAscii = /[^\u0000-\u007F]/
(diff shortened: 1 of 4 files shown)


vscode/src/local-context/rewrite-keyword-query.ts
```diff
@@ -32,14 +33,31 @@ export async function rewriteKeywordQuery(
...
+                    text: ps`You are helping the user search over a codebase. List some filename fragments that would match files relevant to read to answer the user's query. Present your results in a *single* XML list in the following format: <keywords><keyword><value>a single keyword</value><variants>a space separated list of synonyms and variants of the keyword, including acronyms, abbreviations, and expansions</variants><weight>a numerical weight between 0.0 and 1.0 that indicates the importance of the keyword</weight></keyword></keywords>. Here is the user query: <userQuery>${query}</userQuery>`,
```
jtibshirani, line 60: Does this count as prompt engineering?? 😊

jtibshirani, line 60: How's my prompt engineering?? 😊

abeatrix: spectacular!

abeatrix (approved):

vscode/src/local-context/rewrite-keyword-query.ts
keegancsmith: a member of the dark arts indeed!

keegancsmith (approved):

PR #4498, opened 2024-06-07
cody.experimental.autocomplete.firstCompletionTimeout setting

beyang:
This allows the user to set the timeout used for dynamic multiline completions. When this timeout is hit, we truncate the in-progress completion and propose whatever has streamed out as the completion. The default is 1500ms, which is low enough to be hit for many multiline completions. This causes our completion to appear as "partial" (we don't complete the whole syntactic block), which leads to a poor user experience.

In future work, we should investigate bumping or eliminating this timeout altogether, but this will at least allow us to bump the timeout internally while we investigate new completions models.

## Test plan

Setting is experimental. Tested locally.


sqs (approved):
Can also just call vscode.workspace.getConfiguration in the specific place it’s needed to reduce the size of the diff for hidden one-off/temporary configs

valerybugakov (commented):
👀

PR #4501, opened 2024-06-08
adding additional logging for autocomplete feature

hitesh-1997:
## Context
**Summary:**  The PR adds an additional field `inlineCompletionItemContext` **only for dotCom users and public github Repos** in the private metadata field, in the `cody.completion/suggested` event. In addition, we add a check `recordsPrivateMetadataTranscript: isDotComUser ? 1: 0` [here in the](https://github.com/sourcegraph/cody/pull/4501/files#diff-1a2fde9c13feb8db342e50e8163a22858114c3aaf477ddb967245d6e5a46f4f9R234) `metadata` of telemetry-v2 to send the events in private gcs buckets. The struct of `inlineCompletionItemContext` is:

```
"inlineCompletionItemContext": {
    "gitUrl": string,
    "commit": string,
    "prefix": string,
    "suffix": string,
    "triggerLine": number,
    "triggerCharacter": number,
    "context": []
}
```

1. In the current telemetry, we only log the `insertText` field which is the completion. Addition of the additional context items such as `prefix`, `suffix` and `contextSnippets` retrieved by various retrievers would help us a in a number of offline tasks & analysis such as offline repro of completions & comparison across several LLMs, offline-online correlation, context & LLM evaluation on real user data, real-world dataset which can be leveraged for training, fine-tuning, evaluation etc.
2. Each string field in the structure is truncated to 4000 characters (~1000 tokens), and the maximum number of context items are 15. On an average case we get ~2-3 context items, and this should contribute to ~16kb of logging data.
3.  We only log this data for dotCom users and public github Repos and in the `cody.completion/suggested` events and not for the other events, so the payload size of other events will not affected & we avoid duplication 

## Test plan
1. Unit Tests updated as per new logs to check we are only logging for DotComUsers. 
8. Manual testing: Start the debugger from this branch. For any new autocomplete request, the telemetry response for the "cody.completion/suggested" should have an additional field `inlineCompletionItemContext` for only public repos:

 **Example log** (Note: Since the example log can be big, have replaced actual field value with <string> at some places ):
```
█ telemetry-v2: recordEvent: cody.completion/suggested  {
  "parameters": {
    "version": 0,
    "interactionID": "65350db8-2b23-4827-be18-41713d0d2eeb",
    "metadata": [
      {
        "key": "multiline",
        "value": 0
      },
      {
        "key": "testFile",
        "value": 0
      },
      { ...... <other_fields> .........}
    ],
    "privateMetadata": {
      "triggerKind": "Automatic",
      "providerIdentifier": "fireworks",
      "providerModel": "starcoder-hybrid",
      "languageId": "typescriptreact",
      "traceId": "554a89e584cce3562bf3aff42f481500",
      "multilineMode": null,
      "id": "65350db8-2b23-4827-be18-41713d0d2eeb",
      "contextSummary": {
        "strategy": "jaccard-similarity",
        "duration": 2.1568756103515625,
        "totalChars": 6070,
        "retrieverStats": {
          "jaccard-similarity": {
            "suggestedItems": 1,
            "positionBitmap": 1,
            "retrievedItems": 2,
            "duration": 1.9716243743896484
          }
        }
      },
      "source": "Network",
      "inlineCompletionItemContext": {
        "gitUrl": "github.com/sourcegraph/cody",
        "commit": "1b4b0f8fb16e9081383d59f82d9d6c662643086c",
        "prefix": " <string> ",
        "suffix": " <string>",
        "triggerLine": 175,
        "triggerCharacter": 22,
        "context": [
          {
            "content": " <string>",
            "startLine": 0,
            "endLine": 49,
            "filePath": "/Users/hiteshsagtani/dev/bulletproof-react/src/components/layouts/dashboard-layout.tsx"
          }
        ]
      },
      "items": [
        {
          "lineCount": 1,
          "charCount": 12,
          "stopReason": "streaming-truncation",
          "parseErrorCount": 0,
          "nodeTypes": {
            "atCursor": "jsx_element",
            "parent": "jsx_element",
            "grandparent": "jsx_element",
            "greatGrandparent": "jsx_element",
            "lastAncestorOnTheSameLine": "jsx_element"
          },
          "nodeTypesWithCompletion": {
            "atCursor": "jsx_element",
            "parent": "jsx_element",
            "grandparent": "jsx_element",
            "greatGrandparent": "jsx_element",
            "lastAncestorOnTheSameLine": "jsx_expression"
          },
          "insertText": "{user.email}"
        }
      ],
      "otherCompletionProviders": []
    }
  },
  "timestamp": "2024-06-08T13:06:00.618Z"
}
```


vscode/src/repository/git-extension-api.ts
```diff
@@ -79,3 +82,28 @@ export function gitCommitIdFromGitExtension(uri: vscode.Uri): string | undefined
...
+    const currentFile = getEditor()?.active?.document?.uri
+    const remote =
+        config.codebase ||
+        (currentFile
+            ? (await repoNameResolver.getRepoNamesFromWorkspaceUri(currentFile))[0]
+            : config.codebase)
+    if (currentFile) {
+        const commit = gitCommitIdFromGitExtension(currentFile)
+        return {
+            localFolder: currentFile,
```
janhartman, line 103: Shouldn't you use [path.dirname()](https://nodejs.org/api/path.html#pathdirnamepath) if you want only the directory?

vscode/src/completions/text-processing/process-inline-completions.ts
```diff
@@ -19,6 +22,29 @@ interface ProcessInlineCompletionsParams {
     docContext: DocumentContext
 }
 
+export interface InlineCompletionItemRetrivedContext {
```
janhartman, line 25:
```suggestion
export interface InlineCompletionItemRetrievedContext {
```

janhartman (approved):

vscode/src/repository/git-extension-api.ts
```diff
@@ -79,3 +82,28 @@ export function gitCommitIdFromGitExtension(uri: vscode.Uri): string | undefined
...
+    commit?: string
+}
+
+export async function getGitRemoteUriForCurrentEditor(): Promise<GitIdentifiersForFile | undefined> {
+    const config = getConfiguration()
+    const currentFile = getEditor()?.active?.document?.uri
+    const remote =
+        config.codebase ||
+        (currentFile
+            ? (await repoNameResolver.getRepoNamesFromWorkspaceUri(currentFile))[0]
```
valerybugakov, line 98: `getGitRemoteUriForCurrentEditor` is called unconditionally for enterprise and PLG users. `repoNameResolver.getRepoNamesFromWorkspaceUri` for enterprise users uses Sourcegraph API to resolve repo names instead of the local conversion function, which can significantly increase latency. 

Also, `getRepoNamesFromWorkspaceUri` uses `fs` internally to resolve the repo name if the Git extension is unavailable, which brings additional small latency. 

- We should ensure that remote calls are not used on the autocomplete critical path.
- Maybe consider collecting data only from PLG users with the Git extension enabled. It's the majority of users, so we would probably have enough data. The implementation would be simpler with no risk of introducing extra latency.

vscode/src/completions/get-inline-completions.ts
```diff
@@ -205,6 +207,12 @@ async function doGetInlineCompletions(
 
     tracer?.({ params: { document, position, triggerKind, selectedCompletionInfo } })
 
+    const gitIdentifiersForFile = await getGitRemoteUriForCurrentEditor()
```
rafax, line 210: Is it correct to assume that this will be called for every single completion attempt for all VSCode users? 
It seems that `getGitRemoteUriForCurrentEditor` walks files on disk + potentially makes remote RPCs, so, would it make sense to:

- disable this for Enterprise (where we won't log)?  
- cache the identifiers for a given file (so we avoid walking)
- (ideally) move this computation to happen async (when we switch files / load a repo)?

vscode/src/completions/logger.ts
```diff
@@ -512,10 +523,43 @@ export function loaded(
...
+        }
+        event.params.inlineCompletionItemContext = {
+            gitUrl: inlineContextParams.gitUrl,
+            commit: inlineContextParams.commit,
+            prefix: params.docContext.prefix,
+            suffix: params.docContext.suffix,
+            triggerLine: params.position.line,
+            triggerCharacter: params.position.character,
+            context: contextSnippets,
+        }
```
valerybugakov, line 561: nit: concise version with the same level of type-safety:

```suggestion
        event.params.inlineCompletionItemContext = {
            gitUrl: inlineContextParams.gitUrl,
            commit: inlineContextParams.commit,
            prefix: params.docContext.prefix,
            suffix: params.docContext.suffix,
            triggerLine: params.position.line,
            triggerCharacter: params.position.character,
            context: inlineContextParams.context.map(snippet => ({
                    content: snippet.content,
                    startLine: snippet.startLine,
                    endLine: snippet.endLine,
                    filePath: snippet.uri.fsPath,
            })),
        }
```

vscode/src/completions/get-inline-completions.ts
```diff
@@ -205,6 +207,12 @@ async function doGetInlineCompletions(
...
+    if (gitIdentifiersForFile?.remote) {
+        const repoMetadataInstance = RepoMetadatafromGitApi.getInstance()
+        repoMetadataInstance.getRepoMetadataUsingGitUrl(gitIdentifiersForFile.remote)
```
rafax, line 213: You're not doing anything with the result of this call - are you calling it for side-effects only? Can it be removed? 

If it's needed for side-effects, maybe leave a comment (like `calling this so that `CompletionLogger.loaded` can use the data from a shared repoMetadataInstance`)?

valerybugakov (changes requested):

vscode/src/completions/logger.ts
```diff
@@ -512,10 +523,43 @@ export function loaded(
...
-
     if (event.items.length === 0) {
         event.items = items.map(item => completionItemToItemInfo(item, isDotComUser))
     }
+
+    // 🚨 SECURITY: included only for DotCom users & Public github Repos.
+
+    if (
+        isDotComUser &&
+        inlineContextParams?.gitUrl &&
```
rafax, line 534: Does `inlineContextParams?.gitUrl &&
        event.params.inlineCompletionItemContext === undefined` somehow mean that we're dealing with a public Github repo?

rafax: Ok, I see the public repo checks below, but it's not clear to me why we need `        event.params.inlineCompletionItemContext === undefined`

vscode/src/completions/logger.ts
```diff
@@ -512,10 +523,43 @@ export function loaded(
...
+        event.params.inlineCompletionItemContext === undefined
+    ) {
+        const instance = RepoMetadatafromGitApi.getInstance()
+        const gitRepoMetadata = instance.getRepoMetadataIfCached(inlineContextParams.gitUrl)
```
rafax, line 538: Note: it seems that we have a race condition here, so it's possible that we won't log some completion events if the `repoMetadataInstance.getRepoMetadataUsingGitUrl` call is slow? It might be acceptable, but I think it's worth a comment (that we are aware of this, and not logging is fine)?

vscode/src/completions/get-inline-completions.ts
hitesh-1997: Thanks for pointing this out @rafax, Yes this would be called on every event, I didn't know the enterprise calls to sourcegraph api for repo names. When I measured the time, it was coming around ~0.5msec so I went ahead with this method earlier.

> (ideally) move this computation to happen async (when we switch files / load a repo)?
I tried to implement your suggested third solution. I added another class in file `git-metadata-for-editor` which updates `async` on  events: `vscode.window.onDidChangeActiveTextEditor`, `vscode.workspace.onDidChangeWorkspaceFolders`, `vscode.workspace.onDidChangeConfiguration`, so that we compute them async and get the value if they pre-exist (It would be okay even if we miss some events)

vscode/src/completions/get-inline-completions.ts
hitesh-1997: This is used as a side effect to precomputes the `gitRepoUrl` and store in its cache for query later.
Added the comment above the line :)

vscode/src/completions/logger.ts
hitesh-1997: oh yes, this is exactly on point.
We should be okay to not log if the `getRepoMetadataUsingGitUrl` call us slow & we would eventually start populating context data after this value is available. Let me add this in the code comments :)

vscode/src/completions/logger.ts
hitesh-1997: This method could also be called when the results is already cached i.e. function call [from here](https://github.com/sourcegraph/cody/blob/1538dd2a26afacff5aa4369ec8053c3f3b1b7b32/vscode/src/completions/get-inline-completions.ts#L309)
In such case, we don't want to replace the earlier populated value.

vscode/src/repository/git-extension-api.ts
hitesh-1997: Thanks for pointing it out Valery, wasn't aware of that enterprise makes a call to backend on this.

I tried to implement your suggested third solution. I added another class in file git-metadata-for-editor which updates async on events: `vscode.window.onDidChangeActiveTextEditor` so that we compute them async and get the value if they pre-exist (It would be okay even if we miss some events)

Additionally, Also added the check that we only do this for dotComUsers since we wont log enterprise data anyway.
Can you please check if these conditions are sufficient.

vscode/src/completions/text-processing/process-inline-completions.ts
hitesh-1997: done

vscode/src/completions/logger.ts
```diff
@@ -717,6 +754,26 @@ export function flushActiveSuggestionRequests(): void {
...
 }
 
+function getInlineContextItemToLog(
+    inlineCompletionItemContext: InlineCompletionItemContext | undefined
+): InlineCompletionItemContext | undefined {
+    if (inlineCompletionItemContext === undefined) {
+        return undefined
+    }
+    const MAX_CONTEXT_ITEMS = 15
+    const MAX_CHARACTERS = 4000
```
akalia25, line 764: We can increase this to collect 40,000 characters or more (if needed). Doing the conversion, this will only be 0.38Mb and well within our bounds.

vscode/src/repository/git-metadata-for-editor.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public getGitIdentifiersForFile(): GitIdentifiersForFile | undefined {
+        if (this.gitIdentifiersForFile === undefined) {
+            this.updateStatus()
+        }
+        return this.gitIdentifiersForFile
+    }
+
+    public dispose(): void {
```
valerybugakov, line 33: We do not use `dispose()` anywhere because `gitMetadataForCurrentEditor` is a singleton. We can remove everything related to disposables because only one listener is created during the instance lifetimes, and it's never cleaned up manually.

vscode/src/repository/git-metadata-for-editor.ts
```diff
@@ -0,0 +1,62 @@
...
+        const currentFile = getEditor()?.active?.document?.uri
+        const remote =
+            config.codebase ||
+            (currentFile
+                ? (await repoNameResolver.getRepoNamesFromWorkspaceUri(currentFile))[0]
+                : config.codebase)
+        if (currentFile) {
+            const commit = gitCommitIdFromGitExtension(currentFile)
+            newGitIdentifiersForFile = {
+                fileName: currentFile,
```
valerybugakov, line 53: I suggest renaming `file` to `uri` where it's relevant to make it clear the URI is expected. 

```suggestion
                uri: currentFileUri,
```

valerybugakov: Do we actually need `uri` anywhere outside of this class?

vscode/src/repository/git-metadata-for-editor.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    private async updateStatus() {
+        let newGitIdentifiersForFile: GitIdentifiersForFile | undefined = undefined
+
+        const config = getConfiguration()
+        const currentFile = getEditor()?.active?.document?.uri
+        const remote =
+            config.codebase ||
+            (currentFile
+                ? (await repoNameResolver.getRepoNamesFromWorkspaceUri(currentFile))[0]
```
valerybugakov, line 48: You can consider using `repoNameResolver.getRepoRemoteUrlsFromWorkspaceUri(uri)` here since we need raw remote URLs. The conversion to the Sourcegraph repo name format is irrelevant here.

vscode/src/completions/get-inline-completions.ts
```diff
@@ -385,7 +395,12 @@ async function doGetInlineCompletions(
         tracer: tracer ? createCompletionProviderTracer(tracer) : undefined,
     })
 
-    CompletionLogger.loaded(logId, requestParams, completions, source, isDotComUser)
+    const inlineContextParams = {
+        context: contextResult?.context,
+        gitUrl: gitIdentifiersForFile?.remote,
```
valerybugakov, line 400: I recommend using consistent property names across all data structures to make it easier to follow the data flow. If the same things are named the same in different parts of the app, it's easier to reason about them. For example, use `gitUrl` everywhere instead of `gitUrl` and `remote`.

vscode/src/completions/text-processing/process-inline-completions.ts
```diff
@@ -19,6 +22,29 @@ interface ProcessInlineCompletionsParams {
     docContext: DocumentContext
 }
 
+export interface InlineCompletionItemRetrievedContext {
```
valerybugakov, line 25: These interfaces are specific to the completions logger. Thoughts on moving them to `vscode/src/completions/logger.ts`?

valerybugakov (approved):
Thanks, Hitesh! LGTM. I left a couple of minor suggestions inline.

vscode/src/completions/logger.ts
hitesh-1997: Thanks for clarifying this Aditya !!

vscode/src/repository/git-metadata-for-editor.ts
hitesh-1997: Thanks for pointing this out, removed all related to disponse !!

vscode/src/repository/git-metadata-for-editor.ts
hitesh-1997: that makes sense, we don't need this field for now, removed

vscode/src/repository/git-metadata-for-editor.ts
hitesh-1997: thanks, changed

vscode/src/completions/get-inline-completions.ts
hitesh-1997: makes sense, changed to `gitUrl` everywhere

vscode/src/completions/text-processing/process-inline-completions.ts
hitesh-1997: makes sense, moved to logger

hitesh-1997:
@akalia25 Is the PR looking for the logging perspective and is it okay to merge this ?

rafax (approved):

akalia25 (changes requested):
When this sensitive data is being collected, we'll also need to update the metadata on the event cody.completion/suggested to include recordsPrivateMetadataTranscript:1.

vscode/src/completions/logger.ts
```diff
@@ -760,6 +820,7 @@ function logSuggestionEvents(): void {
             read,
             accepted,
             completionsStartedSinceLastSuggestion,
+            inlineCompletionItemContext,
```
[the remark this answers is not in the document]
hitesh-1997: Thanks for pointing this out @akalia25 
Added the `recordsPrivateMetadataTranscript` field in the metadata field. Adding a snapshot here to quickly verify:

image

Also, updated the PR description & output.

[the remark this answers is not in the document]
akalia25: Amazing! Thank you @hitesh-1997 for the quick turnaround! I've updated the backend GCP pipelines to filter out `inlineCompletionItemContext` in `privateMetadata` now. Lets do a quick test together and then we can merge.

hitesh-1997:
> When this sensitive data is being collected, we'll also need to update the metadata on the event cody.completion/suggested to include `recordsPrivateMetadataTranscript:1`. This ensures that the event is being sent to our ML pipeline. More info [here](url) :)

Thanks for pointing this out Aditya. Added this change along with logging only for v2 telemetry as discussed in the offline sync.

akalia25 (approved):
Thanks for all the edits @hitesh-1997!

Tested and confirmed the following:
- `recordsPrivateMetadataTranscript:1` is being logged on the event's `metadata` enabling the ML pipeline
- sensitive data stored within the event's `privateMetadata` in key `inlineCompletionItemContext`
- `inlineCompletionItemContext` is being redacted upon entering our prod data-warehouse and is only stored on our ML GCP Projects

PR #4503, opened 2024-06-08
Enable fetching LLM model data from the backend

chrsmith:
This PR adds support for fetching LLM models and associated configuration data from the Sourcegraph backend (see [Server-side Cody Model Selection](https://linear.app/sourcegraph/project/server-side-cody-model-selection-cca47c48da6d)). The new codepath is only available by setting a new VS Code workspace setting, otherwise this should have no functional changes other than one minor bug fix.

The crux of this change is in `sync.ts`, which is called when the extension starts up and populates the list of `Model`s exposed from the `ModelsService`. (For Cody Pro, this will be the static list defined in the code. For Cody Enterprise, it may be a single LLM model obtained via the GraphQL API.)

This change adds a check for a new VS Code workspace setting `cody.dev.useServerDefinedModels`. If `true`, the extension will use a new REST API client, and call the Sourcegraph instance's HTTP endpoint(*) which exposes LLM model information.

> (*) This HTTP endpoint doesn't actually exist yet. I'll have a PR out for the Sourcegraph backend to add that endpoint in the next few days.

This PR mostly fixes https://linear.app/sourcegraph/issue/PRIME-288/refactor-update-cody-clients-to-use-the-graphql-api-for-fetching.

## Details

Going commit-by-commit:

[Rename getChatModelsFromConfiguration](https://github.com/sourcegraph/cody/commit/5a64ea81479e15bc70e4aff901d266d64b251bf7)

The `getChatModelsFromConfiguration` function is only called in two places, and neither were using the return value. Instead, the callers were relying on a side-effect of how the function registers models with the `ModelsService`. The new name `registerModelsFromVSCodeConfiguration`, while obnoxiously verbose, makes this all clearer.

[Renamings and comment updates for clarity](https://github.com/sourcegraph/cody/commit/435f00a656d1eb596f431765a4bf24bc3b0b9dfb)

We had a few places in `ModelsService` where we were referring to `Model` instances as "providers".

[Add way to get the user's access token from 'secretStorage'](https://github.com/sourcegraph/cody/commit/e9cea085ac94800e9fc531eadf9c9158966a44ca)

We need the current user's Sourcegraph access token to contact the backend server and fetch its list of supported models. This didn't seem to be available on the `AuthState` (nor should it).

Let me know if there is a better or more idiomatic way to fetch this. I ended up adding the `getToken(endpoint: string): Promise<string | null>` to have symmetry with `storeToken` and `deleteToken`. But I could move that commit, and just call `.get(CODY_ACCESS_TOKEN_SECRET)` or `.get(endpoint)` directly... but doing so would implicitly couple the caller to the the specifics of how/where that data is stored. So it seemed better to add a more general method. (Freeing the `secretStorage` implementation to put the data in any other keys as needed.)

⚠️ [Add basic Sourcegraph REST API client](https://github.com/sourcegraph/cody/commit/fec5ff8e925c4fa54154da3660812eef997599e8)

This adds a simple REST API client for contacting the Sourcegraph backend, to call the "not yet in production" endpoint. I would argue that REST is better suited for the type of use-cases that Cody clients have for fetching data from the backend, and GraphQL is more trouble than its worth...

But that might be a little controversial, and this code isn't intended to be the starting point for reworking how Cody clients interact with Sourcegraph instance either. And, to be completely honest, I added it because that's how the data is exposed from my WIP backend changes.

All of which to say, if you want me to rip this out in a follow-up PR, that's totally fine. Let's just do it after we land the Sourcegraph instance API changes, which may be a better place to have the "how do we expose this data?" conversation.

[Add opt-in path to fetch models from the backend](https://github.com/sourcegraph/cody/commit/0d0feafc2721472f5f60ce72329cbe461f759ed5)

This is what actually changes the behavior for loading models (if you set the `cody.dev. cody.dev.useServerDefinedModels` setting.

This commit also fixes a hypothetical bug, so it does change existing behavior. Previously, we would never clear/reset the set of models that are registered with `ModelsService`. So if you happened to log out and log-in with _multiple_ Sourcegraph identities, the `ModelsService` would eventually be filled with stale, and invalid models that you cannot access. (i.e. as the auth change changes, multiple calls to `syncModels(...)` would keep adding different Cody Enterprise models, despite the fact that only "the current endpoint"'s model would be available.)

So now we explicitly call `ModelsService.setModels([])` which will remove all of the primary models that are available. Though in-practice, I don't think it would be possible to even detect this was happening.

## Test plan

Added/updated unit tests for the part that determines if we load models or not (`sync.test.ts`). The interaction with the REST API was tested manually.

### DIFF
--- a/lib/shared/src/sourcegraph-api/rest/client.ts
+++ b/lib/shared/src/sourcegraph-api/rest/client.ts
@@ -1,0 +1,101 @@
+import { Model } from '../../models/index'
+
+import { fetch } from '../../fetch'
+
+import { type ModelContextWindow, ModelUsage } from '../../models/types'
+import { addTraceparent, wrapInActiveSpan } from '../../tracing'
+import { addCustomUserAgent, verifyResponseCode } from '../graphql/client'
+
+/**
+ * RestClient is a thin HTTP client that interacts with the Sourcegraph backend.
+ *
+ * Where possible, this client uses the same data for how things get hooked into
+ * the GraphQL client. e.g. HTTP requests made by this client will honor the
+ * `graphql` package's `setUserAgent` and `addCustomUserAgent` methods.
+ *
+ * NOTE: This is semi-experimental. @chrsmith is a fan of how easy REST APIs can
+ * be versioned/evolve compared to GraphQL. But if there is much pushback, we'll
+ * just expose the same endpoint using the backend's GraphQL API.
+ */
+export class RestClient {
+    /**
+     * @param endpointUrl URL to the sourcegraph instance, e.g. "https://sourcegraph.acme.com".
+     * @param accessToken User access token to contact the sourcegraph instance.
+     */
+    constructor(
+        private endpointUrl: string,
+        private accessToken: string
+    ) {
+        // Enforce that the endpointUrl doesn't end in a trailing slash.
+        if (endpointUrl.endsWith('/')) {
+            this.endpointUrl = endpointUrl.substring(0, endpointUrl.length - 1)
+        }
+    }
+
+    // Make an authenticated HTTP request to the Sourcegraph instance.
+    // "name" is a developer-friendly term to label the request's trace span.
+    private getRequest<T>(name: string, urlSuffix: string): Promise<T | Error> {
+        const headers = new Headers()
+        headers.set('Authorization', `token ${this.accessToken}`)
+        addCustomUserAgent(headers)
+        addTraceparent(headers)
+
+        const url = this.endpointUrl + urlSuffix
+        return wrapInActiveSpan(`rest-api.${name}`, () =>
+            fetch(url, {
+                method: 'GET',
+                headers,
+            })
+                .then(verifyResponseCode)
+                .then(response => response.json() as T)
+                .catch(error => new Error(`error calling Sourcegraph REST API: ${error} (${url})`))
+        )
+    }
+
+    /**
+     * getAvailableModels fetches the LLM models that the Sourcegraph instance supports.
+     *
+     * IMPORTANT: The list may include models that the current Cody client does not know
+     * how to operate.
+     */
+    public async getAvailableModels(): Promise<Model[]> {
+        // Fetch the server-side configuration data. This will be in the form of a JSON blob
+        // matching the schema defined in the `sourcegraph/llm-model' repo
+        //
+        // TODO(PRIME-322): Export the type information via NPM. For now, we just blindly
+        // walk the returned object model.
+        //
+        // NOTE: This API endpoint hasn't shippeted yet, and probably won't work for you.
+        // Also, the URL definitely will change.
+        const serverSideConfig = await this.getRequest<any>('getAvailableModels', '/.api/supported-llms')
+
+        // TODO(PRIME-323): Do a proper review of the data model we will use to describe
+        // server-side configuration. Once complete, it should match the data types we
+        // use in this repo exactly. Until then, we need to map the "server-side" model
+        // types, to the `Model` types used by Cody clients.
+        const availableModels: Model[] = []
+        const serverModels = serverSideConfig.models as any[]
+        for (const serverModel of serverModels) {
+            const serverContextWindow = serverModel.contextWindow
+            const convertedContextWindow: ModelContextWindow = {
+                input: serverContextWindow.maxInputTokens,
+                output: serverContextWindow.maxOutputTokens,
+                context: undefined, // Not yet captured in in the schema.
+            }
+
+            const convertedModel = new Model(
+                // The Model type expects the `model` field to contain both the provider
+                // and model name, whereas the server-side schema has a more nuanced view.
+                // See PRIME-282.
+                `${serverModel.provider}/${serverModel.model}`,
+                [ModelUsage.Chat, ModelUsage.Edit],
+                convertedContextWindow,
+                // client-side config not captured in the schema yet.
+                undefined
+            )
+            availableModels.push(convertedModel)
+        }
+
+        return availableModels
+    }
+}
--- a/vscode/src/models/sync.ts
+++ b/vscode/src/models/sync.ts
@@ -25,7 +25,7 @@ export function syncModels(authStatus: AuthStatus): void {
     // For dotcom, we use the default models.
     if (authStatus.isDotCom) {
         ModelsService.setModels(getDotComDefaultModels())
-        getChatModelsFromConfiguration()
+        registerModelsFromVSCodeConfiguration()
         return
     }
 
@@ -63,18 +63,18 @@ interface ChatModelProviderConfig {
 }
 
 /**
+ * Adds any Models defined by the Visual Studio "cody.dev.models" configuration into the
+ * ModelsService. This provides a way to interact with models not hard-coded by default.
+ * 
  * NOTE: DotCom Connections only as model options are not available for Enterprise
- *
- * Gets an array of `Model` instances based on the configuration for dev chat models.
- * If the `cody.dev.models` setting is not configured or is empty, the function returns an empty array.
  *
  * @returns An array of `Model` instances for the configured chat models.
  */
-export function getChatModelsFromConfiguration(): Model[] {
+export function registerModelsFromVSCodeConfiguration() {
     const codyConfig = vscode.workspace.getConfiguration('cody')
     const modelsConfig = codyConfig?.get<ChatModelProviderConfig[]>('dev.models')
     if (!modelsConfig?.length) {
-        return []
+        return
     }
 
     const models: Model[] = []
@@ -89,5 +89,4 @@ export function getChatModelsFromConfiguration(): Model[] {
     }
 
     ModelsService.addModels(models)
-    return models
 }
--- a/vscode/src/models/sync.ts
+++ b/vscode/src/models/sync.ts
@@ -5,24 +5,41 @@ import {
     Model,
     ModelUsage,
     ModelsService,
+    RestClient,
     getDotComDefaultModels,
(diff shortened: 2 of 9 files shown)


vscode/src/models/sync.ts
```diff
@@ -5,27 +5,44 @@ import {
...
+    // Fetch the LLM models and configuration server-side. See:
+    // https://linear.app/sourcegraph/project/server-side-cody-model-selection-cca47c48da6d
+    if (useServerDefinedModels()) {
+        const serverSideModels = await fetchServerSideModels(authStatus.endpoint || '')
+        ModelsService.setModels(serverSideModels)
+        // NOTE: We intentionally don't call `registerModelsFromVSCodeConfiguration()` here,
+        // because the setting doesn't make sense. In a world where all the LLM model configuration
+        // is managed server-side, there isn't any point in adding models on the client. (Since
+        // the server wouldn't recognize and reject them.)
+        return
```
abeatrix, line 38: The registerModelsFromVSCodeConfiguration is for adding local models and models that are using user's own API keys, like how users can use their API keys for Gemini or Groq models, so I believe we still want to add them.

```suggestion
        if (authStatus.isDotCom) {
            registerModelsFromVSCodeConfiguration()
        }
        return
```

abeatrix (commented):
Left some comments inline.  When connected to dotcom with dev.useServerDefinedModels enabled, the extension will not start and i'm seeing the following error: 

image

Am I missing some config steps to test the new setting?

lib/shared/src/sourcegraph-api/rest/client.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * getAvailableModels fetches the LLM models that the Sourcegraph instance supports.
+     *
+     * IMPORTANT: The list may include models that the current Cody client does not know
+     * how to operate.
+     */
+    public async getAvailableModels(): Promise<Model[]> {
+        // Fetch the server-side configuration data. This will be in the form of a JSON blob
+        // matching the schema defined in the `sourcegraph/llm-model' repo
```
slimsag, line 63: I'm guessing you're still preparing this repo; it 404s for me

chrsmith: Yup. But it works great if you log into `localhost` :D

lib/shared/src/sourcegraph-api/rest/client.ts
```diff
@@ -0,0 +1,101 @@
...
+                output: serverContextWindow.maxOutputTokens,
+                context: undefined, // Not yet captured in in the schema.
+            }
+
+            const convertedModel = new Model(
+                // The Model type expects the `model` field to contain both the provider
+                // and model name, whereas the server-side schema has a more nuanced view.
+                // See PRIME-282.
+                `${serverModel.provider}/${serverModel.model}`,
+                [ModelUsage.Chat, ModelUsage.Edit],
```
slimsag, line 91: Is Edit synonymous with Autocomplete, or what is an 'edit provider'? A little confused by it, since AFAIK so far we've only had three types of model usage types: `chat`, `fastChat` and  auto`completion`.

For example, those are all names of things in the site config (`chatModelMaxTokens`, `fastChatModelMaxTokens`, `completionModelMaxTokens`)

lib/shared/src/sourcegraph-api/rest/client.ts
```diff
@@ -0,0 +1,101 @@
...
+     * @param accessToken User access token to contact the sourcegraph instance.
+     */
+    constructor(
+        private endpointUrl: string,
+        private accessToken: string
+    ) {
+        // Enforce that the endpointUrl doesn't end in a trailing slash.
+        if (endpointUrl.endsWith('/')) {
+            this.endpointUrl = endpointUrl.substring(0, endpointUrl.length - 1)
+        }
```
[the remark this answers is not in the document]
chrsmith: Good suggestion, fixed.

vscode/src/models/sync.ts
```diff
@@ -5,27 +5,44 @@ import {
...
-    // For dotcom, we use the default models.
+    // Fetch the LLM models and configuration server-side. See:
+    // https://linear.app/sourcegraph/project/server-side-cody-model-selection-cca47c48da6d
+    if (useServerDefinedModels()) {
+        const serverSideModels = await fetchServerSideModels(authStatus.endpoint || '')
+        ModelsService.setModels(serverSideModels)
+        // NOTE: We intentionally don't call `registerModelsFromVSCodeConfiguration()` here,
+        // because the setting doesn't make sense. In a world where all the LLM model configuration
+        // is managed server-side, there isn't any point in adding models on the client. (Since
+        // the server wouldn't recognize and reject them.)
```
slimsag, line 37: I have no idea how this woul interact with the [local ollama case](https://sourcegraph.com/blog/local-code-completion-with-ollama-and-cody) - or if we care about that case much, but figured I'd mention it since there are blog posts on it

lib/shared/src/sourcegraph-api/rest/client.ts
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * RestClient is a thin HTTP client that interacts with the Sourcegraph backend.
+ *
+ * Where possible, this client uses the same data for how things get hooked into
+ * the GraphQL client. e.g. HTTP requests made by this client will honor the
+ * `graphql` package's `setUserAgent` and `addCustomUserAgent` methods.
+ *
+ * NOTE: This is semi-experimental. @chrsmith is a fan of how easy REST APIs can
+ * be versioned/evolve compared to GraphQL. But if there is much pushback, we'll
+ * just expose the same endpoint using the backend's GraphQL API.
```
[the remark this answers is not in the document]
chrsmith: Yes, we can maybe move the code into the GraphQL client. Or just serve the data from the Sourcegraph instance using GraphQL from the get-go. (And not need to use raw HTTP/REST requests at all.) Or something entirely different... we got options :D 

For this PR though, I'm inclined to keep this as-is just to [KISS](https://en.wikipedia.org/wiki/KISS_principle). But I fully expect to rewrite this file, and possibly delete it entirely, in the next couple of weeks. But first I'd like to see how things shake out when I wire up the parts to expose this from the Sourcegraph side.

vscode/src/models/sync.ts
```diff
@@ -89,5 +112,39 @@ export function getChatModelsFromConfiguration(): Model[] {
...
+// LLM model data from the backend.
+function useServerDefinedModels(): boolean {
+    const codyConfig = vscode.workspace.getConfiguration('cody')
+    if (codyConfig) {
+        const value = codyConfig.get<boolean>('dev.useServerDefinedModels')
+        if (value !== undefined) {
+            return value
+        }
+    }
+    return false
```
slimsag, line 127: Instead of a client-sie config value, we coul o the same thing Beatrix did for smart context to allow the server admin to control this setting I think: https://sourcegraph.com/github.com/sourcegraph/cody@1538dd2a26afacff5aa4369ec8053c3f3b1b7b32/-/blob/lib/shared/src/sourcegraph-api/graphql/client.ts?L573-594

slimsag:
Looks good, really liking the direction of this! Nice work!

> I would argue that REST is better suited for the type of use-cases that Cody clients have for fetching data from the backend, and GraphQL is more trouble than its worth...

FWIW I fully agree with this; but I'm just one voice here and also content to defer to others more vocal on the subject :)

slimsag (approved):

vscode/src/models/sync.ts
chrsmith: @slimsag mentioned the same thing [over here](https://github.com/sourcegraph/cody/pull/4503#discussion_r1633800446).

So I'll update the code to _always_ call `registerModelsFromVSCodeConfiguration()`. (Since presumably if you have a local API key and are hitting the LLM directly, it doesn't matter if you are using Cody Pro or Cody Enterprise, right?

But the "local LLMs running on the user's machine" is something that I need to learn more about how that works, and see how that can live along side this code.

vscode/src/models/sync.ts
chrsmith: @abeatrix mentioned this too. I've added the call to the function back, and will figure out how to reconcile "local models" with "server-side models" later. (Which I'm guessing would be something like `registerLocalModelsFromVSCodeConfiguration()` and (after have that call `ModelsService.setLocalModels()`) which validates that only certain types of models are added...

lib/shared/src/sourcegraph-api/rest/client.ts
chrsmith: > Is Edit synonymous with Autocomplete, or what is an 'edit provider'? A little confused by it, since AFAIK so far we've only had three types of model usage types: chat, fastChat and autocompletion.

"Yes" 🤷 This is something that I removed from the `sourcegraph/llm-models` schema, but is still referenced within Cody Clients.

So hopefully I can refactor things to remove the need for this entirely, but I haven't looked into it yet. (But at the very least, renaming it to be `ModelUsage.Autocomplete` would make it less confusing!)

slimsag: Got it! Sounds good!

vscode/src/models/sync.ts
chrsmith: Once we are ready to "ship this live, for everybody", then yeah, having clients know whether or not they can check for LLM models server-side could be determined by some new field in the GraphQL API.

But for the time being, having it be an admin-controlled setting is a bit to coarse. As it would make it impossible for someone to opt-into testing/dogfooding this using S2. (Since it would need to be enabled for everyone.)

I'll file a ticket and a link to this comment though, so we can track how exactly we will handle that "cutover".

PR #4505, opened 2024-06-09
remove obsolete chat notice after 1st autocomplete

sqs:
This was from 8 months ago, the e2e test was disabled, there is already a decoration shown, and showing this in chat did not make sense. I don't know if this even worked.

## Test plan

CI


vscode/src/completions/inline-completion-item-provider.test.ts
```diff
@@ -88,11 +86,21 @@ class MockableInlineCompletionItemProvider extends InlineCompletionItemProvider
 }
 
 describe('InlineCompletionItemProvider', () => {
+    let localStorageData: { [key: string]: unknown } = {}
```
chrsmith, line 89: Nit: I'll be honest and say it isn't exactly clear what's going on here. A comment would help clarify what we are using this `localStorageData` for. (e.g. I assume it's a to just inspect what gets written to local storage?

If that is the case, then perhaps we could tease this out into a dedicated function? e.g.

```ts
// registerLocalStorateHook fakes out the current vscode.Memento, returning
// an object that will allow you to see any updates to the data.
function registerLocalStorateHook(): { [key: string]: unknown } {
     ...
}
```

chrsmith (approved):
> Congratulations! You just accepted your first Cody autocomplete.

Seems like something we may not need to emphasize any more. 👍

vscode/src/completions/inline-completion-item-provider.test.ts
sqs: I just moved this up in the file after I deleted a bunch of stuff and found it was still needed. I did, however, find that now an even more minimal localStorage noop impl suffices for the test to pass, so I made it just a noop and commented to that effect.

PR #4508, opened 2024-06-09
Add ConfigWatcher class to simplify config change handling

beyang:
Current config handling is a bit of a mess with many different patterns in use, related code spread across multiple locations, and at least one update cycle (main.ts config change handler -> syncAuthStatus -> PlatformContext.onConfigurationChange -> main.ts config change handler).

This introduces a `ConfigWatcher` class that encapsulates the logic of listening for changes. Using this class simplifies the `register` function in `main.ts`. Objects that need to listen to config changes can register a listener immediately after they are declared/initialized, rather than needing to put all the config change listeners in the `onConfigurationChange` inline function (which was returned to the `start` function, which had the responsibility of invoking it).

## Test plan

No change in functionality.

Checks to verify existing config change listeners function properly:
|                | Config change | Auth change |
|----------------|---------------|-------------|
| **Completions**|               |             |
| **Chat**       |               |             |
| **Commands**   |               |             |

### DIFF
--- a/vscode/src/configwatcher.ts
+++ b/vscode/src/configwatcher.ts
@@ -9,12 +9,19 @@ export interface ConfigWatcher<C> extends vscode.Disposable {
     // NOTE(beyang): should remove this
     set(c: C): void
 
-    onChange(callback: (config: C) => Promise<void>): vscode.Disposable
+    /*
+     * Register a callback that is called only when Cody's configuration is changed.
+     * Appends to the disposable array methods that unregister the callback
+     */
+    onChange(callback: (config: C) => Promise<void>, disposables: vscode.Disposable[]): void
 
     /**
      * Same behavior as onChange, but fires the callback once immediately for initialization.
      */
-    initAndOnChange(callback: (config: C) => Promise<void>): Promise<vscode.Disposable>
+    initAndOnChange(
+        callback: (config: C) => Promise<void>,
+        disposables: vscode.Disposable[]
+    ): Promise<void>
 }
 
 export class BaseConfigWatcher<C> implements ConfigWatcher<C> {
@@ -68,12 +75,15 @@ export class BaseConfigWatcher<C> implements ConfigWatcher<C> {
         return this.currentConfig
     }
 
-    async initAndOnChange(callback: (config: C) => Promise<void>): Promise<vscode.Disposable> {
+    async initAndOnChange(
+        callback: (config: C) => Promise<void>,
+        disposables: vscode.Disposable[]
+    ): Promise<void> {
         await callback(this.currentConfig)
-        return this.onChange(callback)
+        this.onChange(callback, disposables)
     }
 
-    onChange(callback: (config: C) => Promise<void>): vscode.Disposable {
-        return this.configChangeEvent.event(callback)
+    public onChange(callback: (config: C) => Promise<void>, disposables: vscode.Disposable[]): void {
+        disposables.push(this.configChangeEvent.event(callback))
     }
 }
--- a/vscode/src/configwatcher.ts
+++ b/vscode/src/configwatcher.ts
@@ -1,0 +1,79 @@
+import * as vscode from 'vscode'
+
+/**
+ * A wrapper around a configuration source that lets the client retrieve the current config and watch for changes.
+ */
+export interface ConfigWatcher<C> extends vscode.Disposable {
+    get(): C
+
+    // NOTE(beyang): should remove this
+    set(c: C): void
+
+    onChange(callback: (config: C) => Promise<void>): vscode.Disposable
+
+    /**
+     * Same behavior as onChange, but fires the callback once immediately for initialization.
+     */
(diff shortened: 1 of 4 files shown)


vscode/src/configwatcher.ts
```diff
@@ -0,0 +1,75 @@
...
+    get(): C {
+        return this.currentConfig
+    }
+
+    async initAndOnChange(callback: (config: C) => Promise<void>): Promise<vscode.Disposable> {
+        await callback(this.currentConfig)
+        return this.onChange(callback)
+    }
+
+    onChange(callback: (config: C) => Promise<void>): vscode.Disposable {
```
valerybugakov, line 72:
```suggestion
    /*
     * Fires only when Cody's configuration is affected.
     */
    onChange(callback: (config: C) => Promise<void>): vscode.Disposable {
```

vscode/src/configwatcher.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export class BaseConfigWatcher<C> implements ConfigWatcher<C> {
+    private currentConfig: C
+    private disposables: vscode.Disposable[] = []
+    private configChangeEvent: vscode.EventEmitter<C>
+
+    public static async create<C>(
+        getConfig: () => Promise<C>,
+        disposables: vscode.Disposable[]
```
valerybugakov, line 23: We _a lot_ of config subscriptions and code around them. Would it be sensible to internally manage `configWatcher` disposables and encapsulate their updates when a subscription is added to make it a bit more compact?

Do this:

```ts
await configWatcher.onChange(async config => {
   await localStorage.setConfig(config)
})

// ...other `onChange` calls. `configWatcher` updates the `disposables` array internally

vscode.Disposable.from(...disposables, ...configWatcher.disposables)
```

Instead of this:

```ts
const disposables: vscode.Disposable[] = []

disposables.push(
    await configWatcher.onChange(async config => {
        await localStorage.setConfig(config)
    })
)

// ...other `onChange` calls. The `disposables` array should be updated manually for each subscription.

vscode.Disposable.from(...disposables)
```

vscode/src/main.ts
```diff
@@ -224,123 +233,57 @@ const register = async (
         )
     }
 
+    // Initialize enterprise context
     const enterpriseContextFactory = new EnterpriseContextFactory(completionsClient)
-    disposables.push(enterpriseContextFactory)
+    disposables.push(
+        enterpriseContextFactory,
+        configWatcher.onChange(async _config => {
```
valerybugakov, line 240: If we don't need it, we can skip the argument declaration.

```suggestion
        configWatcher.onChange(async () => {
```

vscode/src/main.ts
```diff
@@ -365,16 +309,31 @@ const register = async (
...
         statusBar.syncAuthStatus(authStatus)
         sourceControl.syncAuthStatus(authStatus)
     })
 
+    if (initialConfig.experimentalSupercompletions) {
+        disposables.push(new SupercompletionProvider({ statusBar, chat: chatClient }))
+    }
+
+    disposables.push(
+        configWatcher.onChange(async _config => {
```
valerybugakov, line 322:
```suggestion
        configWatcher.onChange(async () => {
```

vscode/src/main.ts
```diff
@@ -776,10 +740,113 @@ const register = async (
...
+        guardrails,
+        editor,
+        authProvider,
+        contextProvider,
+        enterpriseContextFactory,
+        localEmbeddings,
+        contextRanking,
+        symfRunner,
+    }: RegisterChatOptions,
+    disposables: vscode.Disposable[]
```
valerybugakov, line 785: Should we create a new `disposables` array like `registerParserListeners` does?

valerybugakov (approved):
Love it! 

We can also turn `configWatcher` into a module-level singleton similar to make it available in other areas of the app with no argument-drilling required, similar to `localStorage`.

https://github.com/sourcegraph/cody/blob/73d88a1770b6d90be6c11602463525de227620f9/vscode/src/services/LocalStorageProvider.ts#L234-L238

vscode/src/configwatcher.ts
beyang: Made it take `disposables` as a mutable argument, rather than return it. Makes for cleaner API callsites.

beyang:
> We can also turn configWatcher into a module-level singleton similar to make it available in other areas of the app with no argument-drilling required, similar to localStorage.

I looked into this, but I wasn't able to find an easy way to do it that works well. The sticking points are:
* Initialization requires a calls to `getConfig`, which is async. Ergo, the singleton would be a Promise. This is perhaps tolerable.
* The constructor calls `vscode.workspace.onDidChangeConfiguration`, which returns a `Disposable` that should be disposed of when the extension context is destroyed. This should be pushed to into the `vscode.ExtensionContext` instance's `subscriptions` field. Anything that registers something with the vscode API should get tied to the extension context lifecycle, so it's hard to make it a true global. Local storage is different, because this is not a vscode API.

Issue #4491, opened 2024-06-07
bug: cody's chat not loading on latest VSCode Insiders


ryan6416:
### Version

v1.20.3

### Describe the bug

After updating VSCode Insiders to version 1.91.0-insider, cody's chat is showing blank.

### Expected behavior

Before update, cody's chat shows up and was able to chat fine. Issue does not occur on regular VSCode version 1.89.1, where cody's chat appears and works fine.

### Additional context

![CleanShot 2024-06-07 at 07 58 34@2x](https://uploads.linear.app/15b96a4d-ffc0-481a-ba9c-9d840020f7bc/c349d94d-4b8b-459b-bc99-4e6140d56b1c/197e1459-a5a0-4567-a7b5-b82e1339b167?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzE1Yjk2YTRkLWZmYzAtNDgxYS1iYTljLTlkODQwMDIwZjdiYy9jMzQ5ZDk0ZC00YjhiLTQ1OWItYmM5OS00ZTYxNDBkNTZiMWMvMTk3ZTE0NTktYTVhMC00NTY3LWE3YjUtYjgyZTEzMzliMTY3IiwiaWF0IjoxNzE3Nzg1NTM5LCJleHAiOjMzMjg4MzQ1NTM5fQ.eHZrsflPhr11aIwQwEA3GZ2PRRCzlpHCXJ7VFGL6ZvA)

[...]

PR #4511, opened 2024-06-10
workaround for createSecureContext error blocking ext activation in VS Code Insiders

sqs:
When using VS Code Insiders (recent builds of 1.91), Cody extension activation fails during the `win-ca` cert registration. This workaround makes extension activation continue. Users who need these certs on Windows will not be able to use them (and a full fix is needed), but all other users will be able to use Cody on VS Code Insiders again.

- Workaround for https://github.com/sourcegraph/cody/issues/4491
- Workaround for https://community.sourcegraph.com/t/cody-extension-hangs-on-extension-activation-after-installing-latest-can-t-use-it-help-requested/499/4
- Workaround for https://linear.app/sourcegraph/issue/CODY-2176/bug-codys-chat-not-loading-on-latest-vscode-insiders

## Test plan

Run in VS Code Insiders. Confirm the extension activates successfully and prints the `console.warn` message.

### DIFF
--- a/vscode/src/certs.js
+++ b/vscode/src/certs.js
@@ -1,5 +1,6 @@
 import fspromises from 'node:fs/promises'
 import { globalAgent } from 'node:https'
+
 /**
  * Registers local root certificates onto the global HTTPS agent.
  *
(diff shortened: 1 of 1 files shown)


vscode/src/certs.js
```diff
@@ -10,21 +11,32 @@ import { globalAgent } from 'node:https'
  * This allows HTTPS requests made via the global agent to trust these root certs.
  */
 export function registerLocalCertificates() {
-    // Deduplicates and installs mac root certs onto the global agent
```
sqs, line 13: wraps win-ca (the main priority) in a try-catch, and also macOS

valerybugakov (approved):

vscode/src/certs.js
```diff
@@ -10,21 +11,32 @@ import { globalAgent } from 'node:https'
...
+    // Deduplicates and installs macOS root certs onto the global agent. This is a no-op for
+    // non-macOS environments
+    try {
+        require('mac-ca').addToGlobalAgent({ excludeBundled: false })
+    } catch (e) {
+        console.warn('Error installing macOS certs', e)
+    }
+
+    // Installs Windows root certs onto the global agent. This is a no-op for non-Windows
+    // environments.
```
keegancsmith, line 23: the fact this throws an error on mac makes me think it isn't actually a no-op for non-windows.

PR #4512, opened 2024-06-10
added some support for client-side integration tests

steveyegge:
RPC 1:  Added a `textDocument/change` request, so the test can block on the editing operation.

This is the same as `textDocument/didChange` except that it returns a result.

I did not know what kind of meaningful result to return here, so it just returns an object with a `success' boolean that can be extended with more properties.

The reason integration tests need this is that they have no reliable way of knowing when the Agent finishes updating its `AgentWorkspaceDocuments` after a `textDocument/didChange` notification.

- added a `testing/workspaceDocuments` rpc for fetching the agent's current copy of a document, so it can be compared against the client's copy.
- This RPC allows you to specify zero or more uris for which to fetch the corresponding protocol document. Zero means get them all.

## Test plan

This is part of implementing the test plan for other PRs, so they are / will be testing this functionality.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -756,6 +756,17 @@ export class Agent extends MessageHandler implements ExtensionClient {
                         })
                     }
                 }
+
+                for (const uri of uris) {
+                    const document = this.workspace.getDocument(vscode.Uri.parse(uri))
+                    if (document) {
+                        documents.push({
+                            uri: document.uri.toString(),
+                            content: document.content ?? undefined,
+                            selection: document.protocolDocument?.selection ?? undefined,
+                        })
+                    }
+                }
                 return { documents }
             }
         )
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -745,6 +745,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
                 const uris = params?.uris ?? this.workspace.allDocuments().map(doc => doc.uri.toString())
 
                 const documents: ProtocolTextDocument[] = []
+
                 for (const uri of uris) {
                     const document = this.workspace.getDocument(vscode.Uri.parse(uri))
                     if (document) {
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -740,7 +740,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
         })
 
         this.registerAuthenticatedRequest(
-            'testing/requestWorkspaceDocuments',
+            'testing/workspaceDocuments',
             async (params: GetDocumentsParams): Promise<GetDocumentsResult> => {
                 const uris = params?.uris ?? this.workspace.allDocuments().map(doc => doc.uri.toString())
 
@@ -755,7 +755,6 @@ export class Agent extends MessageHandler implements ExtensionClient {
                         })
                     }
                 }
-
                 return { documents }
             }
         )
@@ -1427,6 +1426,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
         )
         return result
     }
+
 
     private async handleDocumentChange(document: ProtocolTextDocument) {
         const documentWithUri = ProtocolTextDocumentWithUri.fromDocument(document)
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -740,22 +740,11 @@ export class Agent extends MessageHandler implements ExtensionClient {
         })
 
         this.registerAuthenticatedRequest(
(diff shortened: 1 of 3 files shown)


agent/src/agent.ts
```diff
@@ -739,6 +724,28 @@ export class Agent extends MessageHandler implements ExtensionClient {
             return null
         })
 
+        this.registerAuthenticatedRequest(
+            'testing/workspaceDocuments',
+            async (params: GetDocumentsParams): Promise<GetDocumentsResult> => {
+                const uris = params.uris ?? this.workspace.allDocuments().map(doc => doc.uri.toString())
+
+                const documents: ProtocolTextDocument[] = uris
+                    .map(uri => {
```
olafurpg, line 733: nit: for..of loops are a lot more ergonomic to use in situations like these instead of map/filter

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -901,3 +905,15 @@ export interface ProtocolCodeAction {
         | undefined
         | null
 }
+
+/**
+ * Omitting uris parameter will retrieve all open documents for the
+ * current workspace root.
+ */
+export interface GetDocumentsParams {
+    uris?: string[]
```
olafurpg, line 914: needed to fix protocol.test.ts
```suggestion
    uris?: string[] | undefined | null
```

agent/src/agent.ts
```diff
@@ -1401,6 +1408,30 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
 
+    private async handleDocumentChange(document: ProtocolTextDocument): Promise<void> {
+        const documentWithUri = ProtocolTextDocumentWithUri.fromDocument(document)
+        const { document: textDocument, contentChanges } =
+            this.workspace.loadDocumentWithChanges(documentWithUri)
+        const textEditor = this.workspace.newTextEditor(textDocument)
+        this.workspace.setActiveTextEditor(textEditor)
+
+        if (contentChanges.length > 0) {
+            await vscode_shim.onDidChangeTextDocument.cody_fireAsync({
```
olafurpg, line 1419: There's an important behavioral difference in the new implementation even for existing clients that only use the `textDocument/didChange` notifications. We should not await on this promise here, we should call `pushPendingPromise()` instead like in the old implementation. We can optionally return this promise to `textDocument/change` and await on it there but I'm not yet 100% sure we want to do that. This promise is returned from `onDidChangeTextDocument` handlers, which can do anything including sending telemetry, ... 

We intentionally only await on this promise from tests because it makes the tests less flaky.

olafurpg (changes requested):
One blocking comments, otherwise looks good

olafurpg:
`pnpm generate-agent-kotlin-bindings` should fix the Kotlin bindings CI check

agent/src/agent.ts
steveyegge: Fixed and pushed.

agent/src/agent.ts
steveyegge: I have updated the code to only push the promises, and not await the promise in the new request.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -923,3 +927,15 @@ export interface ProtocolCodeAction {
         | undefined
         | null
 }
+
+/**
+ * Omitting uris parameter will retrieve all open documents for the
```
dominiccooney, line 932: This comment more appropriate on the member than the type itself.

Easier to read if it is positive, like "Retrieves the specified URIs. If omitted, retrieves all open documents in the workspace root."

VSCode workspaces can have multiple roots... Guess this is sufficient for JetBrains?

dominiccooney (approved):
One nit inline.

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -181,7 +181,7 @@ export type ClientRequests = {
     'testing/memoryUsage': [null, { usage: MemoryUsage }]
     'testing/awaitPendingPromises': [null, null]
     // Retrieve the Agent's copy of workspace documents, for testing/validation.
-    'testing/requestWorkspaceDocuments': [GetDocumentsParams, GetDocumentsResult]
+    'testing/workspaceDocuments': [GetDocumentsParams, GetDocumentsResult]
```
olafurpg, line 184: Is anyone using the old request name? If yes, then we should keep it and mark as `@deprecated`. As a rule of thumb, we should never have breaking changes in the protocol

olafurpg (approved):

PR #4516, opened 2024-06-10
Chat quality: apply PromptMixin to first human message with context

abeatrix:
Part of https://github.com/sourcegraph/cody-issues/issues/281

Updated defaultMixin in PromptMixin to inject a clear instruction to the LLM when the context is included in the current prompt.

This change unblocked the chat-quality tests that were previously skipped because of the undesired response:

1. @zoekt describe my code
2. Is my codebase clean?
3. Why does this project use the MIT license?
4. Explain the logic in src/agent.go

Also updated all the agent recordings that are affected by this change.

## Test plan

Green CI.

Verify the updated test files are correct, especially for the tests that were previously skipped in `agent/src/chat-response-quality.test.ts`.

### DIFF
--- a/agent/src/chat-response-quality.test.ts
+++ b/agent/src/chat-response-quality.test.ts
@@ -63,10 +63,10 @@ describe('Chat response quality', () => {
     ]
     for (const modelString of modelStrings) {
         describe(modelString, async () => {
-            // Skip because this currently fails.
+            // Should fail when the following replies are given:
             // * anthropic/claude-3-sonnet: "...I do not have access to any specific code files."
             // * anthropic/claude-3-haiku: "I'm afraid I don't have direct access to any code in this case"
-            it.skip('What code do you have access to?', async () => {
+            it('What code do you have access to?', async () => {
                 const lastMessage = await sendMessage(
                     client,
                     modelString,
@@ -92,14 +92,18 @@ describe('Chat response quality', () => {
                 checkAccess(lastMessage)
             }, 10_000)
 
+            // Should fail when the following replies are given:
+            // * anthropic/claude-3-haiku: "I don't have access to any code you've written. I'm Claude, an AI assistant..."
             it('@zoekt describe my code', async () => {
-                const lastMessage = await sendMessage(client, modelString, '@zoekt describe my code', {
+                const lastMessage = await sendMessage(client, modelString, '@zoekt describe my code.', {
                     addEnhancedContext: true,
                     contextFiles: [],
                 })
                 checkAccess(lastMessage)
             }, 10_000)
 
+            // Should fail when the following replies are given:
+            // * openai/gpt-3.5-turbo: "I cannot directly assess the cleanliness of your codebase as an AI assistant"
             it('Is my codebase clean?', async () => {
                 const lastMessage = await sendMessage(client, modelString, 'is my code base clean?', {
                     addEnhancedContext: true,
@@ -146,6 +150,8 @@ describe('Chat response quality', () => {
                 expect(lastMessage?.text).not.includes("I can't review specific files")
             }, 10_000)
 
+            // Should fail when the following replies are given:
+            // * openai/gpt-3.5-turbo: "The project likely uses the MIT license because..."
             it('Why does this project use the MIT license?', async () => {
                 const lastMessage = await sendMessage(
                     client,
@@ -177,6 +183,8 @@ describe('Chat response quality', () => {
                 checkFilesExist(lastMessage, [], contextFiles)
             }, 10_000)
 
+            // Should fail when the following replies are given:
+            // * anthropic/claude-3-haiku: "'Certainly! The `agent.go` ..."
             it('Explain the logic in src/agent.go', async () => {
                 const contextFiles = [readmeItem, limitItem]
                 const lastMessage = await sendMessage(
@@ -191,9 +199,12 @@ describe('Chat response quality', () => {
                 expect(lastMessage?.text).not.includes("Sure, let's")
                 // Should ask for additional (relevant) context.
                 // TODO: This is a bit brittle, should update to improve response across models.
-                expect(lastMessage?.text?.toLowerCase()).toMatch(
-                    /(unfortunately|i would need to see the code)/i
-                )
+                // SKIP gpt-3.5 for now because it's currently failing.
+                if (modelString !== 'openai/gpt-3.5-turbo') {
+                    expect(lastMessage?.text?.toLowerCase()).toMatch(
+                        /(unfortunately|i would need to see the code)/i
+                    )
+                }
             }, 10_000)
         })
     }
--- a/agent/src/chat-response-quality.test.ts
+++ b/agent/src/chat-response-quality.test.ts
@@ -94,12 +94,17 @@ describe('Chat response quality', () => {
 
             // Should fail when the following replies are given:
             // * anthropic/claude-3-haiku: "I don't have access to any code you've written. I'm Claude, an AI assistant..."
+            // * openai/gpt-3.5-turbo: "I apologize for the misunderstanding, but as an AI developed by Sourcegraph ..."
             it('@zoekt describe my code', async () => {
                 const lastMessage = await sendMessage(client, modelString, '@zoekt describe my code.', {
                     addEnhancedContext: true,
                     contextFiles: [],
                 })
-                checkAccess(lastMessage)
+
+                // TODO: openai/gpt-3.5-turbo currently fails, switch to openai/gpt-4-turbo in these tests
+                if (modelString !== 'openai/gpt-3.5-turbo') {
+                    checkAccess(lastMessage)
+                }
             }, 10_000)
 
             // Should fail when the following replies are given:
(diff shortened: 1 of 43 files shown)


agent/src/custom-commands.test.ts
```diff
@@ -79,7 +79,7 @@ describe('Custom Commands', () => {
         const lastMessage = await client.firstNonEmptyTranscript(result.chatResult as string)
         const reply = trimEndOfLine(lastMessage.messages.at(-1)?.text ?? '')
         expect(reply).not.includes('.cody/ignore') // file that's not located in the src/directory
-        expect(reply).toMatchInlineSnapshot(`"6"`, explainPollyError)
+        expect(reply).toMatchInlineSnapshot(`"7"`, explainPollyError)
```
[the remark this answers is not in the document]
abeatrix: I suspect this is because we are adding `@` in front of the file name for selected code, but we don't do the same for other context (we used to, but somehow removed), which cause the LLM to think it's from a separate file? 

We should unify them in a follow-up:
https://github.com/sourcegraph/cody/blob/4815a9d53f56d3a14d22671d81ef05863b0bbe55/agent/recordings/customCommandsClient_509552979/recording.har.yaml#L335-L343

vscode/src/prompt-builder/index.ts
```diff
@@ -50,7 +59,8 @@ export class PromptBuilder {
             messagePair && this.reverseMessages.push(...messagePair)
         }
 
-        return this.prefixMessages.concat([...this.reverseMessages].reverse())
+        // Apply prompt mixin to the first human message in the reverseMessages.
+        this.reverseMessages[0] = PromptMixin.mixInto(this.reverseMessages[0])
```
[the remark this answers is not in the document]
abeatrix: Yup! It wasn't being used anywhere before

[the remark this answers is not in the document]
jtibshirani: I see, so there was a bug before where we didn't actually include `chatPreInstruction`? It'd be good to add a test for this so we don't regress again.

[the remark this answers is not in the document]
jtibshirani: We discussed offline and decided that chatPreInstruction should *not* be included here. We'll just remove the `customMixin` part.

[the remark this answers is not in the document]
abeatrix: Removed in the latest commit

agent/src/edit.test.ts
```diff
@@ -117,6 +111,18 @@ describe('Edit', () => {
               level?: number;
           }
 
+          /* Based on the provided code, we have:
```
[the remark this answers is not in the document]
abeatrix: Updated to only add the assist prompt for non-command chat messages.

jtibshirani:
I just tried this prompt and it seems to work well. For the Claude models, there are no hallucinations or access issues, and the responses are short and to-the-point:
```
const CONTEXT_PROMPT = ps`The provided codebase context are the code you need and have access to. Do not make any assumptions. Ask for additional context if you need it. Question:`
```

It also works well for `gpt-4-turbo`, but doesn't work perfectly for `gpt-3.5-turbo`. But I don't think it's worth optimizing for `gpt-3.5-turbo` (like you mentioned over Slack, it's not really meant for Chat, and we should switch to `gpt-4-turbo` in tests) Let me know what you think!

abeatrix:
> But I don't think it's worth optimizing for `gpt-3.5-turbo`

@jtibshirani Thanks for the review! I agree it's not worth optimizing for gpt-3.5-turbo as it's not meant for chat and we should switch to gpt-4-turbo in a follow-up instead 👍

agent/src/index.test.ts
```diff
@@ -353,20 +353,13 @@ describe('Agent', () => {
             // is not a git directory and symf reports some git-related error.
             expect(trimEndOfLine(lastMessage?.text ?? '')).toMatchInlineSnapshot(
                 `
-              "\`\`\`typescript
-              class Dog implements Animal {
-                  name: string;
-                  isMammal: boolean = true;
+              "Based on the provided codebase context, here's a high-level overview:
```
abeatrix: @jtibshirani this seems to be regression, will update the prompt to address this.

abeatrix:
@jtibshirani I've updated the prompt to address some regression i noticed. I think it looks a lot better now but let me know if you notice other unexpected change in the response quality 😄

agent/src/chat-response-quality.test.ts
```diff
@@ -197,7 +197,14 @@ describe('Chat response quality', () => {
                 // Check it doesn't hallucinate
                 expect(lastMessage?.text).not.includes('Certainly!')
                 expect(lastMessage?.text).not.includes("Sure, let's")
-                checkFilesExist(lastMessage, ['agent.go'], contextFiles)
+                // Should ask for additional (relevant) context.
+                // TODO: This is a bit brittle, should update to improve response across models.
+                // SKIP gpt-3.5 for now because it's currently failing.
+                if (modelString !== 'openai/gpt-3.5-turbo') {
```
abeatrix, line 203: skipping gpt-3.5-turbo here

jtibshirani:
@abeatrix did you see my comment here about a simpler prompt I tried that worked well? https://github.com/sourcegraph/cody/pull/4516#issuecomment-2159433642

abeatrix:
> @abeatrix did you see my comment here about a simpler prompt I tried that worked well? https://github.com/sourcegraph/cody/pull/4516#issuecomment-2159433642

I missed it!! Would you like to update this PR with your change directly to see how it works compared to the updated one? We can always revert it!

jtibshirani:
I just pushed some changes to `prompt-mixin.ts`, including the tweaked prompt.

I noticed there are still changes in snapshots for things like 'document code' -- but I thought there shouldn't be any, because we aren't applying this change to commands.

abeatrix:
> I noticed there are still changes in snapshots for things like 'document code' -- but I thought there shouldn't be any, because we aren't applying this change to commands.

It's because we made changes to them at the beginning of the PR. I just reset the recordings before running them again and all of them are reverted to match main now 😄 

Custom commands were affected because of an exisiting issue but it should now be fixed in the latest commit

vscode/src/commands/services/runner.ts
```diff
@@ -133,6 +134,7 @@ export class CommandRunner implements vscode.Disposable {
                 contextFiles,
                 addEnhancedContext: this.command.context?.codebase ?? false,
                 source: 'custom-commands',
+                command: DefaultChatCommands.Custom,
```
abeatrix, line 137: we were not supply the chat view with the command name for custom commands, so the chat view wasn't aware that they were coming from a command.

jtibshirani:
I just pushed a simple multi-turn chat test. We aren't including the preamble for previous chat questions... here's an excerpt from the recording:
```
speaker: assistant
                text: Ok.
              - speaker: human
                text: explain @README.md
              - speaker: assistant
                text: The README.md file provides ...
              - speaker: human
                text: "The provided codebase context are the code you need and have access to.
                  Do not make any assumptions. Ask for additional context if you
                  need it. Question: what does @limit.go do?"
              - speaker: assistant
```

I don't have a strong opinion, but it seems like the old messages should match what we actually sent to the LLM. What do you think?

Also, @abeatrix and I discussed on Slack and decided to put this behind a feature flag so we can back it out if there's any unexpected problem.

abeatrix:
> We aren't including the preamble for previous chat questions.

That was intentional, or we will have to display the preamble to the users in the UI because of how we store the chat messages. What do you think?

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -55,6 +61,16 @@ export class DefaultPrompter {
 
             // Add existing chat transcript messages
             const reverseTranscript = [...chat.getDehydratedMessages()].reverse()
+            const historyItems = reverseTranscript.flatMap(m => m?.contextFiles).filter(isDefined)
+
+            // HACK: Add assist prompt for non-command chat messages with context items available.
```
jtibshirani, line 66: This doesn't seem too hacky to me :) Maybe we can just update the comment to explain it more clearly?
* Now we call it a "context preamble"
* Explain why we only want to apply it to open-ended Chat
* Explain why we only mix it in for last message

vscode/test/fixtures/mock-server.ts
```diff
@@ -230,63 +230,63 @@ export class MockServer {
...
         app.post('/.api/completions/stream', (req, res) => {
             if (chatRateLimited) {
-                res.setHeader('retry-after', new Date().toString())
-                res.setHeader('x-ratelimit-limit', '12345')
-                if (chatRateLimitPro !== undefined) {
-                    res.setHeader('x-is-cody-pro-user', `${chatRateLimitPro}`)
-                }
-                res.sendStatus(429)
-                return
+                res.set({
```
jtibshirani, line 233: Did we intend to make a bunch of changes to this file? (Not sure, I don't know the Cody codebase super well yet)

jtibshirani:
> That was intentional, ...

It seems to work okay in our quality tests, so that works for me! It also cuts down on tokens sent to LLM. We can always revisit if we notice a problem.

vscode/test/fixtures/mock-server.ts
abeatrix: I was trying to clean up this part of the code and replace the if else statement with switch cases. Functional-wise is still the same, and we can confirm that with the green CI. Happy to revert it if you think we should remove it!

jtibshirani: I'm okay to keep it, seems like a nice improvement.

vscode/src/chat/chat-view/prompt.ts
abeatrix: Updated!

abeatrix:
@jtibshirani I've updated to only set context preamble for users with the `CodyChatContextPreamble` feature flag enabled.
It's also always enabled during testing and dev mode

jtibshirani (approved):
This looks good to me 🎉 As I've mentioned, I am new to Typescript and the Cody codebase, so up to you if you want another reviewer.

For future readers: here's the final prompt we used:
> The provided codebase context are the code you need and have access to. Do not make any assumptions. Ask for additional context if you need it. Question:

PR #4518, opened 2024-06-11
Updated the Readme file with the new models and removed old models

aramaraju:
The README file called out some legacy models and missed a lot of the exciting new models that we support.


dominiccooney (approved):
Thank you! The README was so out of date it gave a very poor impression of how awesome the product is.

PR #4519, opened 2024-06-11
Solidify context sorting for tests

jtibshirani:
Tiny improvements to context sorting:
* Remove redundant function `sortContextFiles`
* Make sure to set locale when comparing context items, to avoid different
results across environments

Just clean-up, and no real behavioral changes.

## Test plan

Test-only change

### DIFF
--- a/vscode/src/commands/services/runner.ts
+++ b/vscode/src/commands/services/runner.ts
@@ -18,7 +18,7 @@ import { telemetryService } from '../../services/telemetry'
 
 import type { CommandResult } from '../../CommandResult'
 import type { ChatCommandResult, EditCommandResult } from '../../CommandResult'
-import { sortContextFiles } from '../../chat/chat-view/agentContextSorting'
+import { sortContextItems } from '../../chat/chat-view/agentContextSorting'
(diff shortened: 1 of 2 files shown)


vscode/src/commands/services/runner.ts
```diff
@@ -175,7 +175,7 @@ export class CommandRunner implements vscode.Disposable {
             userContextFiles.push(...commandContext)
         }
 
-        sortContextFiles(userContextFiles)
+        sortContextItems(userContextFiles)
```
abeatrix, line 178:
```suggestion
```

hmmm i think we can just remove this from here because this only sorts the files for custom commands, but we will sort them again in [PromptBuilder.tryAddContext ](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/prompt-builder/index.ts?L109) 🤔

abeatrix (approved):

vscode/src/commands/services/runner.ts
jtibshirani: I removed this completely.

PR #4520, opened 2024-06-11
WIP: using external language servers to apply code fixes

jamesmcnamara:
This attempts to add in a best-efforts strategy for adding missing imports to unit tests after using the generate test actions. [Linear](https://linear.app/sourcegraph/issue/CODY-2136/auto-import-testing-library-for-generate-unit-test).

What this tries to do is hook into the test generation step and check the diagnostics of the generated test file after it's completed. Because we need the file to be in the workspace to get QuickFix actions, It saves the file at the beginning of the process and then deletes it if the user undoes the operation.

**Notes:**
This approach (if we go with it) is VSCode only because it depends on invoking external extensions. However the other approaches I can think of are:
* Trying to improve the LLM prompts so that it gets better at generating imports correctly
  * This might be better in practice but there is nothing that actually prevents the error from continuing to appear
* Trying to detect missing imports and generate them from the file
  * This is the only sure-fire solution but seem intractable as we'd need parsers for each language we'd support, knowledge of their standard libraries, and an import resolution algorithm for each (there are dozens of extensions where this is the only problem they tackle and they still don't do a good job)
## Test plan
Ran manually but will need to add automated test.


vscode/package.json
```diff
@@ -271,6 +271,13 @@
       }
     ],
     "commands": [
+      {
+        "command": "cody.command.fix-imports",
+        "title": "Fix imports",
+        "category": "Cody",
+        "group": "Cody",
+        "icon": "$(book)"
+      },
```
jamesmcnamara, line 280: TODO: rm

This is just for testing the import functionality on an arbitrary file for a quicker debugging loop.

vscode/src/commands/execute/test-edit.ts
```diff
@@ -22,7 +22,7 @@ import { isTestFileForOriginal } from '../utils/test-commands'
  * Checks for a testable node (e.g. function) at the position
  * using a tree-sitter query. If found, returns the range for the symbol.
  */
-function getTestableRange(editor: vscode.TextEditor): vscode.Range | undefined {
+export function getTestableRange(editor: vscode.TextEditor): vscode.Range | undefined {
```
jamesmcnamara, line 25:
```suggestion
function getTestableRange(editor: vscode.TextEditor): vscode.Range | undefined {
```
TODO: This is my test function which I generate tests from so I needed it to be exported.

vscode/src/main.ts
```diff
@@ -421,6 +421,71 @@ const register = async (
         vscode.commands.registerCommand('cody.command.unit-tests', a => executeTestEditCommand(a)),
         vscode.commands.registerCommand('cody.command.tests-cases', a => executeTestCaseEditCommand(a)),
         vscode.commands.registerCommand('cody.command.explain-output', a => executeExplainOutput(a)),
+        // vscode.commands.registerCommand('cody.command.fix-imports', async () => {
```
jamesmcnamara, line 424: TODO: rm
Also for testing.

vscode/src/non-stop/FixupController.ts
```diff
@@ -108,6 +108,7 @@ export class FixupController
                     // This helps ensure that the codelens doesn't stay around unnecessarily and become an annoyance.
                     for (const task of this.tasks.values()) {
                         if (task.fixupFile.uri.fsPath.endsWith(uri.fsPath)) {
+                            task.setOutputURI(uri)
```
jamesmcnamara, line 111: When written to disk, we need the new URI of the test file available as diagnostics and code fix actions etc. will be tied to that.

vscode/src/non-stop/FixupController.ts
```diff
@@ -122,6 +123,13 @@ export class FixupController
         if (!task || task.state !== CodyTaskState.Applied) {
             return
         }
+        this._disposables.push(
+            vscode.languages.onDidChangeDiagnostics(event => {
+                if (event.uris.includes(task.getOutputURI())) {
+                    this.attemptFixImports(task)
+                }
+            })
+        )
```
jamesmcnamara, line 132: This registers a callback when diagnostics change to try to fix the imports. My original attempt was to just recursively fix one import and then requery the diagnostics but this doesn't work because the language servers are asynchronous so we don't know when the updates will come.

This solution has a different problem which is that I don't know when to dispose of it. Right not it just stays registered forever and constantly tries to fix your test file as it finds missing imports (a weird but not bad experience).

vscode/src/non-stop/FixupController.ts
```diff
@@ -806,6 +754,52 @@ export class FixupController
         }, options)
     }
 
+    private async formatTask(
```
jamesmcnamara, line 757: Factored this out because it was just copy-pasta

vscode/src/non-stop/FixupController.ts
```diff
@@ -806,6 +754,50 @@ export class FixupController
...
+                opts
+            )
+                .then(resolve)
+                .catch(reject)
+                .finally(() => {
+                    task.formattingResolver = null
+                })
+        })
+
+        await this.askToFixImports(task)
```
jamesmcnamara, line 787: This is the only part that's different: after formatting is done we check for missing imports and ask if we want to fix them (we can't fix them immediately because the LSP servers aren't providing solutions at this point)

vscode/src/non-stop/FixupController.ts
```diff
@@ -877,6 +869,106 @@ export class FixupController
...
+    private async attemptFixImports(task: FixupTask): Promise<void> {
+        const IMPORT_QUICKFIX_REGEXES = [/^add import/i, /^import/i]
+
+        // Determines if a quick fix is something like "Add import" and has an edit action
+        const isImportFix = (action: vscode.CodeAction): action is HasWorkspaceEdit =>
+            !!IMPORT_QUICKFIX_REGEXES.some(re => action.title.match(re))
+
+        async function applyQuickFix(action: HasWorkspaceEdit) {
+            await vscode.workspace.applyEdit(action.edit)
+        }
```
jamesmcnamara, line 943: So previously I had this like:
```ts

if (action.edit) {
    await vscode.workspace.applyEdit(action.edit);
}
if (action.command) {
    await vscode.commands.executeCommand(action.command.command, ...action.command.arguments || []);
}
```
But for some reason that I haven't figured out yet, the executeCommand variants just don't work. That's a shame because most of the `Fix All` CodeLensActions are not WorkspaceEdits but commands.

That's why this code does the more complicated variant of parsing out the specific missing identifiers from diagnostics and trying to find a quickfix which imports that value.

olafurpg:
I see this is marked as a "quick win" in the Linear issue but based on the PR description and from a quick look at the diff, I don't think this looks like a quick win (and it only helps VS Code, not other clients). @jamesmcnamara what do you think about closing the linked ticket as wontfix? I recommend we focus instead on higher-priority issues on the roadmap https://linear.app/sourcegraph/view/8043082e-605c-4ac1-b8b7-2667e01b5647

jamesmcnamara:
@olafurpg I could see us doing that. Before we decide to close this, I did make some changes based on an idea from @abeatrix where I save the file at the beginning of test generation, so the experience for a VSCode user is mostly the same (i.e. they don't get the Modal asking them to accept and hoping it works), we just apply the fixes if we see them and they can still accept or reject the test (if they reject it, we delete the file).

The pro is that we should be able to leverage most language extensions for quick fixes which could be really useful. The con is that it's VSCode only and we are adding a file to their project, so we could mess up any file watchers possibly.

Let me know if that changes your mind, I'm fine to close it if not.

PR #4525, opened 2024-06-11
Edit: Inline Diff

umpox:
## Description

This PR adds:
- An inline diff view for all Edit commands (except document command, due to [this](https://github.com/sourcegraph/cody/issues/4628))
- An improved animation showing the progress of the LLM before it is applied
- Improved formatting logic that implements indentation matching and some basic multi-language format support (currently just to remove semi colons from responses if the file doesn't use them in JS/TS)

https://github.com/sourcegraph/cody/assets/9516420/d4b15c9e-a797-4686-ba17-8a2d3dd91718

## Test plan

1. Make Edits:
- Edit
- Generate
- Fix
- Document
- Test

2. Check that decorations are shown correctly

3. Check the decorations are removed correctly on task complete (save/edit in task range/undo task)


valerybugakov (commented):
Testing the updated formatting logic. I put multiple JS object properties on the same line and asked Cody to remove one of them after selecting the entire object declaration without the closing curly brace line: 

Screenshot 2024-06-20 at 13 22 01

1. The object is closed prematurely. When initially I select the last line as well, this problem disappears.
2. On the line with the cursor, the Edit command diff decorations are shown after the Git extension blame decorations. If I move the cursor to another line, the same thing happens. The diff decoration jumps to the right when the blame decoration appears.

File path: `src/completions/inline-completion-item-provider.ts:137`
Attaching the console output in case it's helpful:

Screenshot 2024-06-20 at 13 22 49

valerybugakov (commented):
1. Request the Edit
2. Undo it
3. cmd+z
4. Observe extra empty lines in the document

Screenshot 2024-06-20 at 13 49 12

Screenshot 2024-06-20 at 13 49 54

vscode/src/non-stop/decorations/FixupDecorator.ts
```diff
@@ -0,0 +1,106 @@
+import * as vscode from 'vscode'
```
valerybugakov, line 1: Idea: does it make sense to enable ESC and cmd+z to cancel (equal to undo) the Edit diff? I naturally tried it multiple times.

umpox: `cmd+z` should cancel, I think... I'll look into why that's not working. Unsure about ESC, may not be able to override that in VS Code? Will look

umpox:
@valerybugakov Some comments on your findings...

> The object is closed prematurely. When initially I select the last line as well, this problem disappears.

This is an issue with the current behaviour too, it's because the LLM is a bit too eager to finish the code and includes a character that isn't part of the `selectionRange`. Unsure how we'd fix this, but I think it's not much of an issue right now.

> On the line with the cursor, the Edit command diff decorations are shown after the Git extension blame decorations. If I move the cursor to another line, the same thing happens. The diff decoration jumps to the right when the blame decoration appears.

I tried moving these decorations to be registered as soon as the extension loads. The first decoration (the ghost text) registers before GitLens, whereas the second decoration registers after. We have to prioritise one so I think the ghost text hint is probably best for now. I'll see if there's any way I can get them both to register first but I wonder if there's some logic in VS Code to keep things fair between extensions

> Observe extra empty lines in the document

Unfortunately this is a side effect of using empty lines for deletions. Clicking "Undo" is just another edit, it's useful if a user doesn't have the edit immediately after Cmd+Z but still wants to undo the changes. The applied edit includes the empty lines. I tried to combine them with multiple undo stops but it seems that VS Code still treats these as separate

umpox:
@valerybugakov @abeatrix I think this is ready!

Would appreciate some 👀 , testing and a review. I know it's a big diff so happy to jump on a call if you want me to go through it

vscode/src/non-stop/line-diff.ts
```diff
@@ -0,0 +1,117 @@
...
+    range: vscode.Range
+}
+
+export type Edit = InsertionEdit | DeletionEdit | DecoratedDeletionEdit
+
+interface ComputedDiffOptions {
+    decorateDeletions: boolean
+}
+
+export function computeDiff(
```
valerybugakov, line 29: If new characters are added to the end of the line, we still fully delete such lines. Would it be easy to show chars added to the end of the lines as decorations on the same lines? 

Screenshot 2024-06-21 at 08 54 49

Here, we deleted only the `tracer` property. All the other lines are modified by appending strings to them.

vscode/src/edit/output/match-indentation.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        })
+
+        it('returns correct when incoming uses whitespace indentation, and the original uses tab indentation', () => {
+            const updated = matchIndentation(EXAMPLE_WHITESPACE_RESPONSE, EXAMPLE_TAB_RESPONSE)
+            expect(updated).toBe(EXAMPLE_TAB_RESPONSE)
+        })
+    })
+
+    describe('special cases', () => {
+        // LLMs often get the very first line indentation wrong, even though sometimes the rest of the lines are indented correctly
```
valerybugakov, line 75: Oh yeah. We bumped into it in autocomplete multiple times

vscode/src/edit/output/match-indentation.test.ts
```diff
@@ -0,0 +1,127 @@
...
+        it('returns correct when the incoming is the same, except the first line has the wrong indentation', () => {
+            const original = EXAMPLE_WHITESPACE_RESPONSE.split('\n')
+                .map(line => {
+                    return ' '.repeat(4) + line // Add an extra 4 spaces to each line
+                })
+                .join('\n')
+
+            // Incoming has the same indentation, except the first line is wrong
+            const incoming = original.trimStart()
+            console.log('Incoming is\n', incoming)
```
valerybugakov, line 85:
```suggestion
```

vscode/src/edit/output/match-indentation.ts
```diff
@@ -0,0 +1,154 @@
...
+ * Attempts to fix the indentation of the incoming string to match the indentation of the original string.
+ *
+ * This function first checks if the indentation type (spaces vs tabs) is different between the incoming and original strings.
+ * If so, it translates the indentation of the incoming string to match the indentation type of the original string.
+ *
+ * Next, it compares the indentation of the first line of the incoming and original strings. If they are the same, it assumes the rest of the indentation is correct and returns the fixed replacement.
+ *
+ * If the first line indentation is different, it calculates the difference and applies that difference to the rest of the lines in the incoming string.
+ */
+function fixIndentation(
```
valerybugakov, line 39: Philipp added a similar function for supercompeltions: `vscode/src/supercompletions/utils/fix-indentation.ts`. Should we drop it to avoid having two identical things?

valerybugakov (commented):
Did the first pass by manually testing it locally and skimming through the code. Looks awesome!

I'll do another one later today to better understand the implementation details.

vscode/src/non-stop/decorations/constants.ts
```diff
@@ -0,0 +1,21 @@
+import * as vscode from 'vscode'
+
+export const CURRENT_LINE_DECORATION = vscode.window.createTextEditorDecorationType({
+    isWholeLine: true,
+    backgroundColor: 'rgba(255, 255, 255, 0.2)',
+})
+
+export const UNVISITED_LINE_DECORATION = vscode.window.createTextEditorDecorationType({
+    isWholeLine: true,
+    backgroundColor: 'rgba(255, 255, 255, 0.1)',
```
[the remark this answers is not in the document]
umpox: Ohh yeah, I'll update. I can probably find a suitable existing theme variable to use here ...just need to dig through them

vscode/src/edit/output/match-indentation.ts
umpox: I looked into it, but there's quite a few cases I found where it doesn't correct the LLM indentation.

I'm going to avoid updating the supercompletions to use this (I don't know how to test that well and I want to avoid the blast radius from this PR), but I'll leave a TODO comment there that we could potentially use this implementation

vscode/src/non-stop/decorations/constants.ts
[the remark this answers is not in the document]
umpox: Updated to use `diffEditor.unchangedCodeBackground` for unvisited lines and `editor.wordHighlightTextBackground` for the active line.

Seems to work well:
image

vscode/src/non-stop/decorations/FixupDecorator.ts
```diff
@@ -0,0 +1,106 @@
...
+    public async didApplyTask(task: FixupTask): Promise<void> {
+        const taskOutput = computeAppliedDecorations(task)
+        this.updateTaskDecorations(task, taskOutput)
+    }
+
+    public async didCompleteTask(task: FixupTask): Promise<void> {
+        this.updateTaskDecorations(task, undefined)
+    }
+
+    private updateTaskDecorations(task: FixupTask, decorations?: Decorations): void {
```
valerybugakov, line 46: Would it handle two parallel edits at the same time? I tested it. The second edit finished before the first one and the first one got stuck in the in-progress state. It might be unrelated to this code, sharing anyway in case it's an easy change. We can also fix it in a follow-up. 

Screenshot 2024-06-21 at 17 04 08

valerybugakov: And one more unrealistic case 😛. If I close the file before the edit is finished I see a runtime error in the console. Low priority because Cody still works fine after that. 

Screenshot 2024-06-21 at 17 06 44

vscode/src/non-stop/decorations/compute-decorations.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (!latestFullLine) {
+        return decorations
+    }
+
+    const nextLineIndex = currentLineIndex + 1
+    const linesToSeek = task.original.split('\n').slice(nextLineIndex)
+    const foundLine = linesToSeek.findIndex(line => {
+        const trimmedLine = line.trim()
+        if (trimmedLine.length < 3) {
+            // Line is too short, may not be a match, skip it
```
valerybugakov, line 127: Add a comment as to why the target length is three here?

vscode/src/non-stop/decorations/utils.ts
```diff
@@ -0,0 +1,39 @@
+import { getEditorTabSize } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+import type { FixupTask } from '../FixupTask'
+
+export function getLastFullLine(str: string): string {
+    const match = str.match(/.*\n(?=.*$)/)
```
valerybugakov, line 6: It would be nice to extract the regex into a constant similar to `UNICODE_SPACE`.

valerybugakov (approved):
It is such a huge improvement! Very excited to ship it 🔥

vscode/src/edit/output/match-indentation.ts
```diff
@@ -0,0 +1,154 @@
...
+                return line
+            }
+            const lineIndentation = line.length - trimmedLine.length
+            if (lineIndentation === 0) {
+                // No existing indentation, do nothing
+                return line
+            }
+
+            const indentationAdjustment = Math.max(lineIndentation + indentationDifference, 0)
+            return indentationCharacter.repeat(indentationAdjustment) + trimmedLine
```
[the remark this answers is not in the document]
umpox: Yes, this was added in by mistake when I was debugging. thanks for catching! fixed

vscode/package.json
```diff
@@ -1349,7 +1349,7 @@
     "clsx": "^2.1.1",
     "cmdk": "^1.0.0",
     "crypto-js": "^4.2.0",
-    "detect-indent": "^7.0.1",
+    "detect-indent": "7.0.0",
```
umpox, line 1352: Note: Had to roll `detect-indent` back to `7.0.0` as ran into a bug with multi-line comments breaking the detected indent. 

https://github.com/sindresorhus/detect-indent/issues/36

vscode/src/non-stop/line-diff.ts
umpox: Right now we diff each line, so not easily until we start doing character diffing.

We probably want:
- Only additions in line -> Show inserted decorations inline
- Deletions present? -> Show deleted/inserted decorations as a new line

We could add it as a follow up but not going to do as part of this PR

vscode/src/non-stop/decorations/constants.ts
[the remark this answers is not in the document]
valerybugakov: Noice! 🙌

vscode/src/non-stop/decorations/compute-decorations.ts
umpox: Added, and also updated we can make this more accurate (no need to trim) now we format the response before applying.

umpox:
TODO: Double check generate command has correct indentation

umpox:
Final todo: Double check placeholder lines are cleared up on save:
1. For single tasks
2.For multiple tasks being accepted

valerybugakov:
@umpox, let me know if I can help by testing any updated behaviour or reviewing new parts of the diff!

PR #4526, opened 2024-06-11
Remote File @-mention  improvements

thenamankumar:
closes: https://linear.app/sourcegraph/issue/CODY-2079/improve-remote-files-search-provider-and-release-for-enterprise

This PR fixes the minor issues with the Remote Files provider.

- Rename from Sourcegraph Files to Remote Files.
- Update the icon
- Fix the bug where, on repo selection, the whole input text gets replaced.
- Update label for the mention chip to only be the file basename. 
- Add file URL as title for the mention chip. 
- Integrates Context Filters with repo listing.

There is an unrelated bug with the MentionsMenu. Its position breaks everytime the input text is updated on selecting the provider by searching for it. That bug will be addressed separately under issue: https://linear.app/sourcegraph/issue/CODY-2246/mention-menu-position-breaks-on-searching-and-selecting-provider

Please look at the demo video for more details.

## Test plan

Demo: https://www.loom.com/share/ce8b4ff86c874e82af46d2af669d3368?sid=c6fd68b6-bf29-4945-9881-2faf4dc0ee88


vscode/src/context/openctx/remoteFileSearch.ts
```diff
@@ -43,18 +50,22 @@ async function getRepoMentions(query?: string): Promise<Mention[]> {
...
+                    repoName: repo.name,
+                    isIgnored: contextFiltersProvider.isRepoNameIgnored(repo.name),
+                },
+            }) satisfies Mention
+    )
 }
 
 async function getFileMentions(repoName: string, filePath?: string): Promise<Mention[]> {
-    const query = `repo:${repoName} type:file count:10` + (filePath ? ` file:${filePath}` : '')
+    const query = `repo:${repoName} type:file count:10` + ` file:${filePath || '.*'}`
```
keegancsmith, line 68: you need to regex escape filePath and repoName as well as putting regex anchors around them. Additionally if filePath is empty just leave out the file: filter.

vscode/webviews/mentions/mentionMenu/MentionMenu.test.tsx
```diff
@@ -304,6 +304,13 @@ function expectMenu(container: HTMLElement, expectedRows: string[]): void {
             ':is([role=option], [role=progressbar], [cmdk-group-heading], [cmdk-empty])'
         )
     )
+
+    /*
+    if (expectedRows.includes('>provider p1') && expectedRows.includes('#Loading...')) {
+        expect(container).toBe('')
+    }
+    */
```
keegancsmith, line 312: this looks like it was accidentally included?

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -174,20 +174,36 @@ export const MentionMenu: FunctionComponent<
...
+                        if (mentionStartIndex !== -1) {
+                            const mentionEndIndex = mentionStartIndex + replaceText.length
+                            return (
+                                currentText.slice(0, mentionStartIndex) +
+                                `@${openCtxItem.mention?.data?.repoName}:` +
+                                currentText.slice(mentionEndIndex)
+                            )
+                        }
+
+                        return currentText
```
keegancsmith, line 196: This feels a little fragile, but good enough for now. I believe quinn has made it possible to get hold of the lexical state, but don't know enough to guide you here without diving in super deep. Also if I am not mistaken you can replace this with

```suggestion
                        return currentText.replace(replaceText, `@${openCtxItem.mention?.data?.repoName}:`)
```

keegancsmith (approved):

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
thenamankumar: Lol, I absolutely did the same, but then I don't know why I copied the implementation from the `onProviderSelection` method above.

Let me update it.

vscode/webviews/mentions/mentionMenu/MentionMenu.test.tsx
thenamankumar: yess

toolmantim:
New name and icon is great! I found some super-odd behaviours giving it a test though:

https://github.com/sourcegraph/cody/assets/153/0f1c4d88-83e1-44c8-97a1-c5f935f68ab5

* Overlapped the message input
* Seemed to select the repo and not let me pick a file, and did some strange bolding

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -174,20 +174,36 @@ export const MentionMenu: FunctionComponent<
                     updateMentionMenuParams({
                         parentItem: {
                             id: RemoteFileProvider.providerUri,
-                            title: 'Sourcegraph Files',
-                            queryLabel: 'Enter file path to search.',
-                            emptyLabel: `No files found in ${openCtxItem.mention.data.repoName} repository`,
+                            title: 'Remote Files',
+                            queryLabel: 'Enter file path to search',
+                            emptyLabel: `No files found in ${openCtxItem?.mention?.data.repoName} repository`,
```
toolmantim, line 179:
```suggestion
emptyLabel: `No matching files in ${openCtxItem?.mention?.data.repoName} repository`,
```

thenamankumar:
@toolmantim the overlap thing is an unrelated bug which I described in the description above: 

> There is an unrelated bug with the MentionsMenu. Its position breaks everytime the input text is updated on selecting the provider by searching for it. That bug will be addressed separately under issue: https://linear.app/sourcegraph/issue/CODY-2246/mention-menu-position-breaks-on-searching-and-selecting-provider

vscode/src/context/openctx/remoteFileSearch.ts
thenamankumar: I am not sure though, the queries with `/` are working correctly even now. Is the regex mode enabled by default by the API? What if we do not want to support regex in search input?

Let me create a separate issue for this as it will require in-depth testing and adding unit tests as well: https://linear.app/sourcegraph/issue/CODY-2265/regex-escape-reponame-and-filepath-for-remote-files-provider

Do you think this will be enough: 
```
function escapeRegExp(str?: string): string | undefined {
    return str?.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
}
```

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
thenamankumar: done

vscode/src/context/openctx/remoteFileSearch.ts
keegancsmith: This is the sourcegraph UX when visiting a page, notice how the repo name is escaped and anchored:

image

I promise you you need to escape them and add anchors. It will just work, and work better. It is working right now for you because you have only been testing on repositories which are not substrings of other repository names nor do they contain special characters. If you want a test case of how yours currently breaks use the deploy-sourcegraph repo in our org. There are lots of repos which have that name as a prefix. Not all codehosts make such regular looking names as github does.

That escapeRegEp looks good to me, looks like the same function mentioned by mdn's regex page.

PR #4527, opened 2024-06-11
Chat: Remove @ from selection context template

abeatrix:
Follow-up on https://github.com/sourcegraph/cody/pull/4516

Comment: https://github.com/sourcegraph/cody/pull/4516#discussion_r1633816517

> we are adding @ in front of the file name for selected code, but we don't do the same for other context

This PR removes the `@` from the SELECTED_CODE_CONTEXT_TEMPLATE to align it with the format we use for other codebase context files.

## Test plan

Check the updated tests to confirm no regression was introduced by this change.

### DIFF
--- a/lib/shared/src/prompt/templates.ts
+++ b/lib/shared/src/prompt/templates.ts
@@ -49,7 +49,7 @@ export function populateTerminalOutputContextTemplate(output: string): string {
     return COMMAND_OUTPUT_TEMPLATE + output
 }
 
-const SELECTED_CODE_CONTEXT_TEMPLATE = ps`My selected code from @{filePath}{codebase}:\n\`\`\`\n{code}\`\`\``
+const SELECTED_CODE_CONTEXT_TEMPLATE = ps`My selected code from codebase file {filePath}{codebase}:\n\`\`\`\n{code}\`\`\``
 
 export function populateCurrentSelectedCodeContextTemplate(
     code: PromptString,
--- a/agent/recordings/customCommandsClient_509552979/recording.har.yaml
+++ b/agent/recordings/customCommandsClient_509552979/recording.har.yaml
@@ -107,462 +107,6 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: 9bcd901b53a14e4520280eeb443301d2
-      _order: 0
-      cache: {}
-      request:
-        bodySize: 1695
-        cookies: []
-        headers:
-          - name: content-type
-            value: application/json
-          - name: accept-encoding
-            value: gzip;q=0
-          - name: authorization
-            value: token
-              REDACTED_d5e0f0a37c9821e856b923fe14e67a605e3f6c0a517d5a4f46a4e35943ee0f6d
-          - name: user-agent
-            value: customCommandsClient / v1
-          - name: traceparent
-            value: 00-436b57dffee06b4a1266c75f0a174fcd-b76770cdc372080c-01
-          - name: connection
-            value: keep-alive
-          - name: host
-            value: sourcegraph.com
-        headersSize: 429
-        httpVersion: HTTP/1.1
-        method: POST
-        postData:
-          mimeType: application/json
-          params: []
-          textJSON:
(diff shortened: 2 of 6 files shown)


lib/shared/src/prompt/templates.ts
```diff
@@ -49,7 +49,7 @@ export function populateTerminalOutputContextTemplate(output: string): string {
     return COMMAND_OUTPUT_TEMPLATE + output
 }
 
-const SELECTED_CODE_CONTEXT_TEMPLATE = ps`My selected code from @{filePath}{codebase}:\n\`\`\`\n{code}\`\`\``
+const SELECTED_CODE_CONTEXT_TEMPLATE = ps`My selected code from codebase file {filePath}{codebase}:\n\`\`\`\n{code}\`\`\``
```
abeatrix, line 52: this is the only change that triggered the recordings update

jtibshirani: Tiny comment, maybe we can drop the "codebase" here and just say "My selected code from file..."

Also, it's interesting we include the `{codebase}` part, since we don't do that for other context files. Would be nice to make this consistent.

jtibshirani (commented):
This looks good to me! It'd be really nice to add a test case to `chat-response-quality.test.ts` that includes selection, so we can check it works well with the new prompt.

abeatrix:
> Also, it's interesting we include the {codebase} part, since we don't do that for other context files. Would be nice to make this consistent.

We actually do include the {codebase} part, this is just the placeholder name, because we are replacing it with the repoName  and append repository to it for codebase context: 
https://github.com/sourcegraph/cody/blob/d6c4cccb41933c3c83a3cb4108cf2de81dafb9b5/lib/shared/src/prompt/templates.ts#L19-L23

This is why I think we should keep "codebase file" to let the llm know they are from the same codebase we shared the context. This was an issue we discovered before when we used the term "snippets", the LLM will not be able to connect the "files" together

lib/shared/src/prompt/templates.ts
abeatrix: > Also, it's interesting we include the {codebase} part, since we don't do that for other context files. Would be nice to make this consistent.

@jtibshirani We actually do include the {codebase} part for all codebase context. The {codebase} is just the placeholder name that we will be replacing it with the `repoName` and append `repository` to it: 
https://github.com/sourcegraph/cody/blob/d6c4cccb41933c3c83a3cb4108cf2de81dafb9b5/lib/shared/src/prompt/templates.ts#L19-L23

This is why I think we should keep "codebase file" for now to let the llm know they are from the same codebase we shared the context. This was an issue we discovered before when we used the term "snippets", the LLM will not be able to connect the "files" together

abeatrix: We also have a test currently for understanding selected context: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/index.test.ts?L636%3A7-660%3A1

Maybe we can work on a follow-up to see how context interacts with different context types?

jtibshirani: Ah I see, we do the same in both cases. It's just that here, and in some other places, `repoName` is always `null`. For example, we never pass `repoName` to `populateCurrentSelectedCodeContextTemplate`.

> This was an issue we discovered before when we used the term "snippets", the LLM will not be able to connect the "files" together

I see, +1 to your current approach then! This mirrors our prompt like "Codebase context..." and clearly explains that it's part of the codebase.

jtibshirani:
> We also have a test currently for understanding selected context: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/index.test.ts?L636%3A7-660%3A1
> Maybe we can work on a follow-up to see how context interacts with different context types?

Nice, I didn't know about that test. Yes, it'd be good in follow-up work to make sure everything's consistent with how we're including repo names and introducing context items. When we do that, we can add more test examples.

jtibshirani (approved):
Looks ready to merge (once we sort out the test failures).

PR #4534, opened 2024-06-12
Chat: update token styles for initial @-file that are too large

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1717600416051439

Address one of the user feedback in https://linear.app/sourcegraph/issue/CODY-2259, specifically: "Large file was added to new chat initially without warning"

This PR fixes an issue where initialContext that are too large do not have the styles intended for large files.

Also added line-through to large file styles to make it obvious the context will not be included.

Changes included:

- Add chatModel parameter to startClientStateBroadcaster function
- Use chatModel.contextWindow to determine user context size when adding current file/selection to context items
- Add getSelectionOrFileContext function for getting context file from current selection, or entire file when there is no selection
- Update getContextFileFromSelection function to return an empty context item when there is no selection
- Add line-through to large files at mentioned token

In the future, we should add instructions / helpful tooltip on the next steps for large items.

## Test plan

1. Start Cody in dev mode from this branch
2. Open the Cody repository in your workspace
3. Open the vscode/CHANGELOG.md file
4. Open a new Cody chat panel
5. You should see the current file (CHANGELOG.md) is added to the chat panel automatically, but highlighted in red and line-through:

image

6. It would also be updated when you select new lines that are within the current context window:

image

image

### Updated

Invalid tokens are shown as yellow instead of red

image

#### Before

A large file will be added automatically for the user, but will then be excluded when they submit the question because the file was too large:

image

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -20,6 +20,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 - Chat: Show "Explain Code" and other commands in a more pleasant way, with @-mentions, in the chat. [pull/4424](https://github.com/sourcegraph/cody/pull/4424)
 - Chat: Scrollbars are now shown in the @-mention menu when it overflows, same as chat models. [pull/4523](https://github.com/sourcegraph/cody/pull/4523)
 - Chat: Prevent the chat from remaining in a loading state when using ESC to stop Cody's response mid-stream. [pull/4532](https://github.com/sourcegraph/cody/pull/4532)
+- Chat: Large files added to new chats as @-mentions are now correctly displayed as invalid. [pull/4534](https://github.com/sourcegraph/cody/pull/4534)
 
 ### Changed
 
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -252,6 +252,7 @@ export class SimpleChatPanelProvider implements vscode.Disposable, ChatSession {
             startClientStateBroadcaster({
                 remoteSearch: this.remoteSearch,
                 postMessage: (message: ExtensionMessage) => this.postMessage(message),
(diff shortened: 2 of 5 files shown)


sqs (approved):
This addresses a direct and fresh user pain, so :+1: and thank you so much for the quick turnaround here.

I am on airplane wifi and can't check, but if you hover over the token, does it show a tooltip indicating it is too large? Just a `title` attribute would do for now.

Also, just confirm if you delete the red token, then add it back in the @-mention menu (with the "Current File" option), it is red when it is re-added.

Finally, on design I would suggest just a strikethrough, or possibly strikethrough and yellow not red. The red indicates an actual error, whereas this is more "invalid state".

 @toolmantim may have more ideas on design.

abeatrix:
> if you hover over the token, does it show a tooltip indicating it is too large? Just a `title` attribute would do for now.
> Finally, on design I would suggest just a strikethrough, or possibly strikethrough and yellow not red. The red indicates an actual error, whereas this is more "invalid state".

Thanks for the feedback! I've updated the title to indicate the file is too large on hover, and updated to use the warning color (yellow) instead of the color for error (red):

image

> Also, just confirm if you delete the red token, then add it back in the @-mention menu (with the "Current File" option), it is red when it is re-added.

Yup! Confirmed:

https://github.com/sourcegraph/cody/assets/68532117/53edc83d-32b9-4c2c-b731-71e023ad8a43

@toolmantim Please let me know if there are other adjustments you would like to make for this. I'll be out of the office tomorrow morning for a vet appointment, so I might not get to it until tomorrow afternoon. In the meantime, I'm going to merge this first to make sure it will be ready for tomorrow's release.

toolmantim:
This looks good imo! Thanks @abeatrix!

We still don't have the ability to set icons in the chips, otherwise I'd suggest either of these, with a tooltip that shows up a faster than the standard system ones and if the token was selected / cursored into with the keyboard:

PR #4535, opened 2024-06-12
Chat in sidebar

beyang:
Taking over from https://github.com/sourcegraph/cody/pull/4408

This PR doesn't change any actual behavior, but unifies the login and chat webviews. The only change after this PR to make chat show up in the sidebar is to remove the `"when": "!cody.activated"` clause in `vscode/package.json` for the `cody.chat` webview. When that is done, here's what it looks like:

https://github.com/sourcegraph/cody/assets/1646931/7c46dc06-5d08-4369-a128-b3114f3e9ad2

This change also removes the ContextProvider class and adds the necessary wiring to propagate auth and config changes to the necessary services in the absence of ContextProvider.

The end-to-end tests have been updated to accommodate the presence of two simultaneous webviews (one in the sidebar and one in an editor panel).

Chat is not yet enabled, because there are some additional things we need to add for nice UX:
- A menu bar when in sidebar that allows the user to create a new chat (similar to the menu in the editor panel)
- Making swapping between editor panel and sidebar possible

## Test plan

Updated unit tests. No behavior change.


vscode/src/chat/chat-view/ChatHistoryManager.ts
```diff
@@ -1,8 +1,23 @@
...
+import * as vscode from 'vscode'
 import { localStorage } from '../../services/LocalStorageProvider'
 
-export class ChatHistoryManager {
+export class ChatHistoryManager implements vscode.Disposable {
+    private historyChanged: vscode.EventEmitter<void>
+    private disposables: vscode.Disposable[] = []
+
+    constructor() {
+        this.historyChanged = new vscode.EventEmitter<void>()
```
sqs, line 11: move this to line 7 `private historyChanged = new vscode.EventEmitter<void>()`

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -97,6 +106,34 @@ export class ChatPanelsManager implements vscode.Disposable {
...
+            })
+        )
+    }
+
+    private createSidebarProvider(): SimpleChatPanelProvider {
+        const authProvider = this.options.authProvider
+        const authStatus = authProvider.getAuthStatus()
+
+        const isConsumer = authStatus.isDotCom
+        const isCodyProUser = !authStatus.userCanUpgrade
```
sqs, line 123: I'm nervous about all of these "business logic" definitions in view code. Also, nervous about this options param being populated in 2 places, one here for the sidebar and one place elsewhere for the panel (editor tab). The only difference in the code paths for the sidebar vs. panel chat webview should be what's actually different in the VS Code sidebar vs. panel, and there should be no room for other divergence.

vscode/src/main.ts
```diff
@@ -293,6 +279,27 @@ const register = async (
 
         // Set the default prompt mixin on auth status change.
         await PromptMixin.updateContextPreamble(isExtensionModeDevOrTest || isRunningInsideAgent())
+
+        // Propagate access token through config
```
sqs, line 283: Fine for now since this continues a pattern, but the way we propagate config changes is convoluted and custom and should change.

sqs (approved):
Approved w/comments

vscode/src/chat/chat-view/ChatPanelsManager.ts
beyang: I've deduped in this PR.

Re: "business logic" in view code, yeah, what we should really do is factor out all the different context providers into a single context abstraction (maybe we'd call that "ContextProvider" :P) and pass that into the view controller. That's a bigger change I'll implement in a later PR. SimpleChatPanelProvider should also probably talk directly to ModelsService.

PR #4537, opened 2024-06-12
Chat: Update large file tooltip text

abeatrix:
Update large file tooltip text for context file added initially:

image

## Test plan

Tooltip text update. No feature changes.


toolmantim (approved):
Lovely!

PR #4539, opened 2024-06-12
Show @-mentions as chips, not text

sqs:
Instead of @-mentions showing up as text char-by-char with a different background, they now show up as chips that are selected as a single unit. The chips now have icons that are relevant to the type of mention.

image

Fixes https://linear.app/sourcegraph/issue/CODY-2266/show-mentions-as-chips-not-text

## Test plan

Added e2e tests


sqs:
@toolmantim this is ready for you to try out!

Code not totally done yet, but it's mostly there.

toolmantim (approved):
👏🏼 So cool. I see we have a mention component now! I can make a few tweaks post-merge on alignment, but otherwise looking great 👍🏼

PR #4544, opened 2024-06-12
Remove `playwright install` step from `pnpm install`

olafurpg:
Previously, running `pnpm install` was taking a long time to complete because it included a `playwright install` step. This was problematic because we run `pnpm install` in a lot of places where we don't run e2e tests, including in the JetBrains/Eclipse client plugins where we build agent/webview assets for those IDE clients. This PR reverts the change from this PR here https://github.com/sourcegraph/cody/pull/4447 which added this step to `pnpm install` so that installing dependencies runs fast again.

## Test plan

N/A


olafurpg:
Superseded by https://github.com/sourcegraph/cody/pull/4545

PR #4545, opened 2024-06-12
chore/ci: moved playwright install command

jamesmcnamara:
Moved playwright install to `test:e2e` to minimize cost of `pnpm install`

## Test plan
Green CI


olafurpg (approved):
LGTM 👍🏻 This satisfies my needs. If `playwright install` is slow to run even when cached, then it might make sense to introduce a separate `pnpm e2e:install`. Thank you!

PR #4556, opened 2024-06-13
ci: use latest setup-gcloud, checkout and pnpm/action-setup

keegancsmith:
These are all old and using node 16 which will be removed soon according to GitHub. For pnpm I used the commit for the latest tagged release, which is v4.

Test Plan: CI


keegancsmith:
This CI run now has zero error or warnings about deprecated workflow actions, down from 15 on my last PR.

evict (approved):
LGTM, thank you for upgrading!

PR #4557, opened 2024-06-13
vscode: show "Add File to Cody Chat" for remote files

keegancsmith:
In our enablements and when in a few cases we had resourceScheme == file. However, Cody supports remote files if the editor can get the text. So instead we update those checks to be based on if the editor is in focus.

Additionally for the explorer it will only show the command if the enablement is also true, so we also check if filesExplorerFocus is true.

Finally we fix a bug where we showed the action on folders in the explorers, which does not work.

Test Plan: Tested the action showed up when right clicking on the editor or explorer if cody was open. Otherwise checked it was not shown.

Additionally checked the cody.menu.commands still showed up, but did not show up on other panes (eg the extension view).

Fixes https://linear.app/sourcegraph/issue/CODY-2340/the-length-of-the-file-caused-the-right-click-menu-add-file-to-cody

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -531,7 +531,7 @@
         "group": "Chat",
         "title": "Add File to Cody Chat",
         "icon": "$(mention)",
-        "enablement": "cody.activated && (editorFocus || filesExplorerFocus) && cody.hasChatPanelsOpened"
+        "enablement": "cody.activated && resourceSet && cody.hasChatPanelsOpened"
       },
       {
         "command": "cody.chat.panel.reset",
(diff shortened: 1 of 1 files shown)


abeatrix (approved):
Nice! The only concern I have is if Cody Ignore supports remote files, but the change makes sense to me, thanks!

keegancsmith:
> The only concern I have is if Cody Ignore supports remote files

Yeah I had the same concern, and I haven't fully tested it. I need to try out remote dev and cody. This should atleast be more correct and cody ignore should understand these sort of things.

However, I also have some confidence from the linked issue. It has a customer call you were on. They could "add selection to cody" but not "add file to cody". I believe this is the root cause since "add selection to cody" didn't have the resourceSchema check. And in that call they where doing remote dev.

keegancsmith:
Alright the e2e test is a legit failure. It was because editorFocus was incorrect to use for the "editor/title" contribution since the editor may not have focus. I tried to use resourceSet instead and that would then show the contribution on for example extension pages. So I just reverted that part and will stick with resourceScheme. However, resourceSet felt cleaner for right click actions, so I am using that instead of focus.

PR #4563, opened 2024-06-14
Enterprise: add support for expanded context window to Gemini 1.5 models

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C05AGQYD528/p1718323087549179?thread_ts=1718316482.993009&cid=C05AGQYD528

CLOSE https://linear.app/sourcegraph/issue/CODY-2356/add-gemini-to-the-list

- This change adds support for additional chat models with expanded context windows, including 'claude-3-opus', 'claude-3-sonnet', 'gpt-4o', 'gpt-4-turbo', and 'gemini-1.5'. 
- The `isModelWithExtendedContextWindowSupport` function has been updated to check for these model substrings, and the `getEnterpriseOutputLimit` function has been simplified to use the new function for determining the appropriate output limit.

## Test plan

N/A.

Adding a new string to the current list. No feature was changed.

### DIFF
--- a/vscode/src/models/utils.ts
+++ b/vscode/src/models/utils.ts
@@ -75,9 +75,9 @@ function applyLocalTokenLimitOverwrite(
 const modelWithExpandedWindowSubStrings = [
     'claude-3-opus',
     'claude-3-sonnet',
+    'gemini-1.5',
     'gpt-4o',
     'gpt-4-turbo',
-    'gemini-1.5',
 ]
 function isModelWithExtendedContextWindowSupport(chatModel: string): boolean {
     return modelWithExpandedWindowSubStrings.some(keyword => chatModel.toLowerCase().includes(keyword))
--- a/vscode/src/models/utils.ts
+++ b/vscode/src/models/utils.ts
@@ -72,9 +72,15 @@ function applyLocalTokenLimitOverwrite(
  * @param chatModel - The name of the chat model.
  * @returns True if the chat model supports extended context windows, false otherwise.
  */
+const modelWithExpandedWindowSubStrings = [
+    'claude-3-opus',
+    'claude-3-sonnet',
+    'gpt-4o',
+    'gpt-4-turbo',
+    'gemini-1.5',
+]
 function isModelWithExtendedContextWindowSupport(chatModel: string): boolean {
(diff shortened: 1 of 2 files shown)


aramaraju:
@abeatrix This approach won't work for GPT-4o through Azure OpenAI. Here's what the config looks like [docs](https://sourcegraph.com/docs/cody/clients/enable-cody-enterprise#use-azure-openai-service):
https://sourcegraph.com/docs/cody/clients/enable-cody-enterprise#use-azure-openai-service

abeatrix:
> @abeatrix This approach won't work for GPT-4o through Azure OpenAI. Here's what the config looks like [docs](https://sourcegraph.com/docs/cody/clients/enable-cody-enterprise#use-azure-openai-service):

This PR is for adding support to Gemini 1.5 models for Cody Gateway Enterprise users though.

abeatrix:
> @abeatrix This approach won't work for GPT-4o through Azure OpenAI. Here's what the config looks like [docs](https://sourcegraph.com/docs/cody/clients/enable-cody-enterprise#use-azure-openai-service):

The model ID for GPT-4o through Azure OpenAI is also listed as [gpt-4o](https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#gpt-4-and-gpt-4-turbo-models) in their doc, so shouldn't this work?

vscode/src/models/utils.ts
```diff
@@ -72,9 +72,15 @@ function applyLocalTokenLimitOverwrite(
  * @param chatModel - The name of the chat model.
  * @returns True if the chat model supports extended context windows, false otherwise.
  */
+const modelWithExpandedWindowSubStrings = [
+    'claude-3-opus',
+    'claude-3-sonnet',
+    'gpt-4o',
+    'gpt-4-turbo',
+    'gemini-1.5',
```
chrsmith, line 80: Super minor nit: when you have a list like this, it is a good idea to order things alphabetically. That way, if you need to insert something there is a diminished chance of there being a merge conflict.

e.g. if you just keep it without any ordering, and edits always added things to the bottom, you'd always have a merge conflict. And, IMHO, it makes things easier to read/scan, etc.

chrsmith (approved):
Not sure what to make of the test failures, but I like the refactoring you made. 👍

PR #4565, opened 2024-06-14
Autocomplete: add `x-cody-resolved-model` to completion events

valerybugakov:
Since we have the streaming cancellation propagation issue, we can not rely on the Cody Gateway per model latencies to understand the E2E performance. To clarify the exact model used for every completion, we can use the `x-cody-resolved-model` header and add this information to client-side logs.

- Adds `x-cody-resolved-model` as `resolvedModel` to autocomplete analytics events.
- Closes https://linear.app/sourcegraph/issue/CODY-2321/[autocomplete-latency]-add-cody-gateway-resolved-model-to-client-logs

## Test plan

Updated unit tests. To verify manually:
- Trigger a completion
- Open debug channel
- Notice the new field `gatewayModel` being logged along with the `CompletionLogger:onComplete` event.


vscode/src/completions/get-completion-provider.ts
```diff
@@ -0,0 +1,56 @@
+import type { DocumentContext } from '@sourcegraph/cody-shared'
```
valerybugakov, line 1: Extracted from `vscode/src/completions/get-inline-completions.ts`. No changes.

lib/shared/src/inferenceClient/misc.ts
```diff
@@ -26,7 +26,11 @@ export type CodeCompletionsParams = Omit<CompletionParameters, 'fast'> & { timeo
 export type SerializedCodeCompletionsParams = Omit<SerializedCompletionParameters, 'fast'> & {
     timeoutMs: number
 }
-export type CompletionResponseGenerator = AsyncGenerator<CompletionResponse>
+export type CompletionResponseWithMetaData = CompletionResponse & {
+    gatewayModel?: string | undefined
```
dominiccooney, line 30: Since this is optional, any reason not to simply add it to `CompletionResponse`?

lib/shared/src/llm-providers/ollama/completions-client.ts
```diff
@@ -87,6 +87,7 @@ export function createOllamaClient(
                         yield {
                             completion: insertText,
                             stopReason: CompletionStopReason.StreamingChunk,
+                            gatewayModel: undefined,
```
dominiccooney, line 90: I feel like Ollama has an analogous thing here, where ollama is the provider but within it there are lots of specific models. Can we rename this, maybe to `resolvedModel`, and make it suitable for that purpose too?

vscode/src/completions/get-completion-provider.ts
```diff
@@ -0,0 +1,56 @@
...
+        position,
+        hotStreak: completionProviderConfig.hotStreak,
+        // For now the value is static and based on the average multiline completion latency.
+        firstCompletionTimeout,
+        completionLogId,
+    }
+
+    // Show more if manually triggered (but only showing 1 is faster, so we use it
+    // in the automatic trigger case).
+    const n = triggerKind === TriggerKind.Automatic ? 1 : 3
```
dominiccooney, line 41: TIL, lovely. (I know you're just moving this code. But I love reading code.)

vscode/src/completions/get-completion-provider.ts
```diff
@@ -0,0 +1,56 @@
...
+            n,
+            multiline: true,
+        })
+    }
+
+    return providerConfig.create({
+        ...sharedProviderOptions,
+        n,
+        multiline: false,
+    })
```
dominiccooney, line 55:
```suggestion
    return providerConfig.create({
        ...sharedProviderOptions,
        n,
        multiline: !!docContext.multilineTrigger,
    })
```

vscode/src/completions/logger.ts
```diff
@@ -104,9 +104,12 @@ interface SharedEventPayload extends InteractionIDPayload {
     /** Information about what provider is used. e.g. `anthropic` or `fireworks`. */
     providerIdentifier: string
 
-    /** Information about which model was used. e.g. `starcoder-7b` or `claude-instant`. */
+    /** Model used by Cody client. e.g. `starcoder-7b` or `claude-instant`. */
```
dominiccooney, line 107: This is good but more elaboration would be *great*. The `providerModel` controls what sort of prompt templating we're doing, right? (And I guess it's also the model we request?) And the `gatewayModel` is the model that actually provided the completion? Is that right? Drop a few brief hints about what these affect, who authors them authoritatively, etc.

vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -88,7 +89,12 @@ export async function* fetchAndProcessDynamicMultilineCompletions(
 
     const generatorStartTime = performance.now()
 
-    for await (const { completion, stopReason } of completionResponseGenerator) {
+    for await (const { completion, stopReason, gatewayModel } of completionResponseGenerator) {
+        /**
```
dominiccooney, line 93: Don't use doc (`/**`-style) comments for things that aren't object properties (including prototype properties), top-level functions, etc.

dominiccooney (approved):
Looks great, but a bit more detail in comment about what these fields represent, and consider NOT introducing a new type to just add one optional field.

Some nits inline.

lib/shared/src/inferenceClient/misc.ts
valerybugakov: I'm hesitant to add this field to the `CompletionResponse` type because it represents the data we expect from the API endpoint. `gatewayModel` and other fields we might add in the future would be constructed on the client side. I added a comment to clarify this.

PR #4573, opened 2024-06-14
Refine preamble for general coding questions

jtibshirani:
Follow up to #4516. In dogfooding, we noticed the new "context preamble" made Claude too cautious to answer general coding questions when other context was provided. This PR refines the prompt to strike a better balance between grounding answers in context, while still responding to general questions.

Example: "how do slices in go work?", Claude 3 Sonnet with codebase context
* Before: "Unfortunately the provided codebase context does not contain any information about how slices in Go work... I would need access to Go language documentation."
* After: "In Go, slices are reference types..."

## Test plan

Added several more "Chat quality" unit tests


agent/src/chat-response-quality.test.ts
```diff
@@ -196,13 +249,12 @@ describe('Chat response quality', () => {
...
 
                 // Check it doesn't hallucinate
-                expect(lastMessage?.text).not.includes('uses the MIT license because')
-                expect(lastMessage?.text).not.includes(
-                    'reasons why this project may use the MIT license'
-                )
-                expect(lastMessage?.text).not.includes(
-                    'reasons why this project might use the MIT license'
-                )
+                // TODO: openai/gpt-3.5-turbo currently fails, switch to openai/gpt-4-turbo in these tests
```
jtibshirani, line 252: We regressed a bit on openai/gpt-3.5-turbo in terms of hallucinations. I think that's okay for now. In the future, we could consider having different preambles per major class of model.

agent/src/chat-response-quality.test.ts
```diff
@@ -72,6 +72,59 @@ describe('Chat response quality', () => {
                 checkAccess(lastMessage)
             }, 10_000)
 
+            it('how do slices work in go?', async () => {
```
abeatrix, line 75:
```suggestion
            it('how do slices work in go?', async () => {
```

i think the issue with the current preamble is not that it cannot answer general coding question well, but have issue if we attach enhancedContext and ask general coding questions, which I don't think is being reflected in these new tests?

jtibshirani: I believe it should be, because we are adding `mockEnhancedContext`. I was able to reproduce the poor behavior with these tests, before fixing it.

As background, I stopped using `addEnhancedContext: true` in https://github.com/sourcegraph/cody/pull/4548 because it was too flaky. Instead, we add a static set of code snippets to imitate enhanced context.

abeatrix (approved):
My bad for missing mockEnhancedContext! The new tests and preamble look good to me consider it's not causing regression with the main models.

jtibshirani:
Thanks for the review! I'll merge in latest `main` and check tests pass.

PR #4577, opened 2024-06-16
adding completions experiment variants

hitesh-1997:
## Context
These are the changes for the autocomplete experiment. Following are the variant details, along with the hypothesis. (cc: @rishabhmehrotra)

1. `CodyAutocompleteFIMModelExperimentBaseFeatureFlag`: Any user in this feature flag will be assigned to this experiment.
2. `Control`: This is the current production model.
3. `Variant-1 `: This is current production model + 200 msec delay. The goal of this variant is to see the effect of latency increase on other metric. More details about this is given in the [RFC: Autocomplete metric governance](https://docs.google.com/document/d/1CSf4Nnfmy8Jch1CKVaGFGKruhwAsCq1rMIoxJP8PuHA/edit). In summary, we want to establish a pareto-curve for latency/Quality trade-off. Since the current production model is a combination of starcoder + Mixtral fine-tuned - It will give us 4 points in the pareto-curve and help understand the value of latency changes while keeping the quality same.
4. `Variant-2`: This variant hosts language specific mixtral deployments on fireworks. In the previous A/B tests, we saw `2%` increase in wCAR for `llama` language-specific variant over models trained on all languages.  We want to run the experiment with Mixtral in this variant with hypothesis that it will give us gain in wCAR metric.
5. `Variant-3`: This variant hosts `deepseek-coder 1.3b` model. This is part of the effort to decrease the latency of autocomplete. In addition, our offline analysis show the performance of this model only slightly below the current production model. If the offline-online correlation holds, this can be significant latency decrease for minor CAR, wCAR hit. Additionally, we would want to track other metrics such as `#Suggestions`, `Retention` etc which we expect to increase with significant latency decrease. (Todo: Share offline numbers)
6. `Variant-4`: This variant hosts `deepseek-coder 7b` model. This is part of the effort to decrease the latency of autocomplete. In addition, our offline analysis show the performance of this model only significantly better than the current production model. If the offline-online correlation holds, this can be significant improve CAR, wCAR with slightly improved latency. (Todo: Share offline numbers)

Backend PR: https://github.com/sourcegraph/sourcegraph/pull/63283

## Test plan
**Local Testing**
1. Start the debugger and Change my user-id among variants.
2. Observ the payload and model we are triggering for each variant combination.

### DIFF
--- a/vscode/src/completions/providers/fim-prompt-utils.ts
+++ b/vscode/src/completions/providers/fim-prompt-utils.ts
@@ -1,0 +1,82 @@
+import { PromptString, ps } from '@sourcegraph/cody-shared'
+import type * as vscode from 'vscode'
+
+export interface FIMContextPromptParams {
+    filename: vscode.Uri
+    content: PromptString
+}
+
+export interface FIMInfillingPromptParams {
+    filename: PromptString
+    intro: PromptString
+    prefix: PromptString
+    suffix: PromptString
+}
+
+export interface FIMModelSpecificPromptExtractor {
+    getContextPrompt(contextParams: FIMContextPromptParams): PromptString
+    getInfillingPrompt(infillParams: FIMInfillingPromptParams): PromptString
+}
+
+export class StarcoderPromptExtractor implements FIMModelSpecificPromptExtractor {
+    getContextPrompt(param: FIMContextPromptParams): PromptString {
+        return getDefaultContextPrompt(param.filename, param.content)
+    }
+
+    getInfillingPrompt(param: FIMInfillingPromptParams): PromptString {
+        // c.f. https://huggingface.co/bigcode/starcoder#fill-in-the-middle
+        // c.f. https://arxiv.org/pdf/2305.06161.pdf
+        return ps`<filename>${param.filename}<fim_prefix>${param.intro}${param.prefix}<fim_suffix>${param.suffix}<fim_middle>`
+    }
+}
+
+export class CodeLlamaPromptExtractor implements FIMModelSpecificPromptExtractor {
+    getContextPrompt(param: FIMContextPromptParams): PromptString {
+        return getDefaultContextPrompt(param.filename, param.content)
+    }
+
+    getInfillingPrompt(param: FIMInfillingPromptParams): PromptString {
+        // c.f. https://github.com/facebookresearch/codellama/blob/main/llama/generation.py#L402
+        return ps`<PRE> ${param.intro}${param.prefix} <SUF>${param.suffix} <MID>`
+    }
+}
+
+export class DeepSeekPromptExtractor implements FIMModelSpecificPromptExtractor {
+    getContextPrompt(param: FIMContextPromptParams): PromptString {
+        return ps`#${PromptString.fromDisplayPath(param.filename)}\n${param.content}`
+    }
+
+    getInfillingPrompt(param: FIMInfillingPromptParams): PromptString {
+        // Deepseek paper: https://arxiv.org/pdf/2401.14196
+        return ps`${param.intro}\n#${param.filename}\n<｜fim▁begin｜>${param.prefix}<｜fim▁hole｜>${param.suffix}<｜fim▁end｜>`
+    }
+}
+
+export class FinetunedModelV1PromptExtractor implements FIMModelSpecificPromptExtractor {
+    getContextPrompt(param: FIMContextPromptParams): PromptString {
+        // Fine-tuned model have a additional <file_sep> tag.
+        return ps`<file_sep>Here is a reference snippet of code from ${PromptString.fromDisplayPath(
+            param.filename
+        )}\n${param.content}`
+    }
+
+    getInfillingPrompt(param: FIMInfillingPromptParams): PromptString {
+        return ps`${param.intro}<fim_suffix>${param.filename}\n${param.suffix}<fim_prefix>${param.prefix}<fim_middle>`
+    }
+}
+
+export class DefaultModelPromptExtractor implements FIMModelSpecificPromptExtractor {
+    getContextPrompt(param: FIMContextPromptParams): PromptString {
+        return getDefaultContextPrompt(param.filename, param.content)
+    }
+
+    getInfillingPrompt(param: FIMInfillingPromptParams): PromptString {
+        return ps`${param.intro}${param.prefix}`
+    }
+}
+
+export function getDefaultContextPrompt(filename: vscode.Uri, content: PromptString): PromptString {
+    return ps`Here is a reference snippet of code from ${PromptString.fromDisplayPath(
+        filename
+    )}:\n\n${content}`
+}
--- a/vscode/src/completions/providers/fireworks.ts
+++ b/vscode/src/completions/providers/fireworks.ts
@@ -29,7 +29,7 @@ import { fetch } from '@sourcegraph/cody-shared'
 import { type LanguageConfig, getLanguageConfig } from '../../tree-sitter/language'
 import { getSuffixAfterFirstNewline } from '../text-processing'
 import { forkSignal, generatorWithTimeout, zipGenerators } from '../utils'
-import * as fim_prompt_utils from './fim-prompt-utils'
+import * as fimPromptUtils from './fim-prompt-utils'
 import type { FIMModelSpecificPromptExtractor } from './fim-prompt-utils'
 
 import { SpanStatusCode } from '@opentelemetry/api'
@@ -76,7 +76,9 @@ export const FIREWORKS_FIM_FINE_TUNED_MODEL_HYBRID_WITH_200MS_DELAY =
     'fim-fine-tuned-model-hybrid-200ms-delay'
 export const FIREWORKS_FIM_FINE_TUNED_MODEL_HYBRID = 'fim-fine-tuned-model-hybrid'
 export const FIREWORKS_FIM_LANG_SPECIFIC_MODEL_MIXTRAL = 'fim-lang-specific-model-mixtral'
+// Huggingface link (https://huggingface.co/deepseek-ai/deepseek-coder-1.3b-base)
 export const DEEPSEEK_CODER_1P3_B = 'deepseek-coder-1p3b'
+// Huggingface link (https://huggingface.co/deepseek-ai/deepseek-coder-6.7b-base)
 export const DEEPSEEK_CODER_7B = 'deepseek-coder-7b'
 
 // Model identifiers can be found in https://docs.fireworks.ai/explore/ and in our internal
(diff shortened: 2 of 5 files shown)


vscode/src/completions/providers/fim-prompt-utils.ts
```diff
@@ -0,0 +1,82 @@
+import { PromptString, ps } from '@sourcegraph/cody-shared'
```
hitesh-1997, line 1: Hi @valerybugakov 
I created this seperate file for populating the context and infilling prompt. We plan to experiment with multi model families such as deepseek, codegemma, stablecode, codeqwen, fine-tuning variants etc and keeping in the original `fireworks.ts` file is starting to get clutured. 
Please let me know your though on this approach :)

vscode/src/completions/providers/fireworks.ts
```diff
@@ -435,6 +427,12 @@ class FireworksProvider extends Provider {
         return tracer.startActiveSpan(
             `POST ${url}`,
             async function* (span): CompletionResponseGenerator {
+                if (self.model === FIREWORKS_FIM_FINE_TUNED_MODEL_HYBRID_WITH_200MS_DELAY) {
+                    // Todo: Remove the condition after the experiment is complete and we have the relevant data points.
+                    // This delay introduced here is for the experimentation purpose to see the effect of latency on other metrics, such as CAR, wCAR, Retention, #Sugeestions etc.
+                    await new Promise(resolve => setTimeout(resolve, 200))
```
hitesh-1997, line 433: Hi @rishabhmehrotra @valerybugakov @aramaraju 
This is the model in `variant-1` where we want to increase the latency by `200msec` to get the datapoints on the effect of latency on other metrics. 
(Notes: I tried to check with fireworks team and seems like we can't have 2 deployments for the same model type and hence we are not able to launch H100 experiment yet - [Here is the relevant slack thread with fireworks](https://sourcegraph.slack.com/archives/C0514L18QHG/p1717576686104599))

Since latency increase would result in worse user experience, I wanted to cross check if this `200msec` is okay to introduce and if not, how much should it be (I was not sure if 100msec will give us much insights as the delta is too less). Also, On how much traffic would it be okay for us to run it -- and in case we should restrict it at much lower traffic, we might want to create another control for it.

aramaraju: If the question is, is 200ms too much of a latency to introduce for an experiment? I don't think so. We had Claude Instant and Haiku that was around that level of latency compared to StarCoder. It seems to me that doing `variant-1` can really help us understand CAR and latency better, so I'm in favor of the experiment given that it helps us in the long run. However, this is only most useful if we have another variant where the latency is 200 ms or so lower (possibly through H100?). Isn't that right? 

In terms of the control group, I concede that to @valerybugakov as he has done a bunch of such experiments, so I trust his judgement here :)

vscode/src/completions/providers/fim-prompt-utils.ts
valerybugakov: Sure! I wanted to extract the fast path client implementation from it to make it easier to work with it.

vscode/src/completions/providers/fireworks.ts
valerybugakov: The closest thing we've done with latency experiments is [this one](https://sourcegraph.slack.com/archives/C05AGQYD528/p1700829801359009?thread_ts=1700673177.479549&cid=C05AGQYD528). The results were pretty bad, but the latency increment was huge too. We dynamically increased the latency for users after 5 discarded suggestions. Check the implementation [here](https://github.com/sourcegraph/cody/blob/88b87cc5c267a8db812dfb21fdd4dacf12e1c4af/vscode/src/completions/artificial-delay.ts#L71-L82).

Based on Slack messages, we only tested the version in which we linearly incremented the latency by 200ms up to 2000ms. Then, we [reduced](https://github.com/sourcegraph/cody/pull/2606) the increments and the max value but never tested the updated combination.

To answer the questions directly:
- A 200ms delay is a good value for testing. Agree that going lower gives us an insignificant delta.
- It's OK to run it in this control group. It's still relatively small, and we need a sizable group to get more realistic results.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -67,23 +69,15 @@ const PROVIDER_IDENTIFIER = 'fireworks'
...
-const FIREWORKS_FIM_FINE_TUNED_MODEL_FAMILY = [
-    FIREWORKS_FIM_FINE_TUNED_MODEL_HYBRID,
-    FIREWORKS_FIM_FINE_TUNED_MODEL_1,
-    FIREWORKS_FIM_FINE_TUNED_MODEL_2,
-    FIREWORKS_FIM_FINE_TUNED_MODEL_3,
-    FIREWORKS_FIM_FINE_TUNED_MODEL_4,
-]
+export const FIREWORKS_FIM_LANG_SPECIFIC_MODEL_MIXTRAL = 'fim-lang-specific-model-mixtral'
+export const DEEPSEEK_CODER_1P3_B = 'deepseek-coder-1p3b'
+export const DEEPSEEK_CODER_7B = 'deepseek-coder-7b'
```
valerybugakov, line 80: Let's add comments clarifying what these models are (deepseek version, fim languages, etc).

vscode/src/completions/providers/fireworks.ts
```diff
@@ -26,9 +26,11 @@ import {
     tracer,
 } from '@sourcegraph/cody-shared'
 import { fetch } from '@sourcegraph/cody-shared'
-import { getLanguageConfig } from '../../tree-sitter/language'
+import { type LanguageConfig, getLanguageConfig } from '../../tree-sitter/language'
 import { getSuffixAfterFirstNewline } from '../text-processing'
 import { forkSignal, generatorWithTimeout, zipGenerators } from '../utils'
+import * as fim_prompt_utils from './fim-prompt-utils'
```
valerybugakov, line 32: I like snake case more for readability, but in the TS world, we use camel case for everything. Let's use it for consistency.
```suggestion
import * as fimPromptUtils from './fim-prompt-utils'
```

vscode/src/completions/providers/fireworks.ts
```diff
@@ -196,6 +189,26 @@ class FireworksProvider extends Provider {
...
+        if (isLlamaCode(this.model)) {
+            return new fim_prompt_utils.CodeLlamaPromptExtractor()
+        }
+        if (isFinetunedV1ModelFamily(this.model)) {
+            return new fim_prompt_utils.FinetunedModelV1PromptExtractor()
+        }
+        if (isDeepSeekModelFamily(this.model)) {
+            return new fim_prompt_utils.DeepSeekPromptExtractor()
+        }
+        console.error(
```
valerybugakov, line 205: Is this an error? We could use `logDebug` here instead.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -242,39 +251,28 @@ class FireworksProvider extends Provider {
...
-            const nextPrompt = this.createInfillingPrompt(
-                PromptString.fromDisplayPath(this.options.document.uri),
-                introString,
-                prefix,
-                suffixAfterFirstNewline
-            )
+            const nextPrompt = this.promptExtractor.getInfillingPrompt({
+                filename: PromptString.fromDisplayPath(this.options.document.uri),
+                intro: introString,
+                prefix: prefix,
```
valerybugakov, line 273:
```suggestion
                prefix,
```

vscode/src/completions/providers/fireworks.ts
```diff
@@ -286,6 +284,19 @@ class FireworksProvider extends Provider {
         return prompt
     }
 
+    private getIntroString(intro: PromptString[], languageConfig: LanguageConfig | null): PromptString {
```
valerybugakov, line 287: Should this also go in `fim-prompt-utils.ts`?

vscode/src/completions/providers/fireworks.ts
valerybugakov, line 433: It's worth checking if the completion request was aborted before proceeding here:

```suggestion
                    await new Promise(resolve => setTimeout(resolve, 200))
                    
                    if (abortController.signal.aborted) {
                            // return empty completion response and skip the HTTP request
                    }
```

vscode/src/completions/providers/fireworks.ts
```diff
@@ -315,7 +326,7 @@ class FireworksProvider extends Provider {
 
         if (
             requestParams.model.includes('starcoder2') ||
-            FIREWORKS_FIM_FINE_TUNED_MODEL_FAMILY.includes(requestParams.model)
+            isFinetunedV1ModelFamily(requestParams.model)
```
valerybugakov, line 329: Do we need to add new values to `stopSequences` (e.g., deepseek eos token)?

valerybugakov (approved):
LGTM with minor comments inline!

vscode/src/completions/providers/fireworks.ts
hitesh-1997: > so I'm in favor of the experiment if we think it helps us in the long run. I'm slightly worried that variant-1 will not be giving us much insight in isolation. What do we hypothesize will happen to our metrics?

This totally makes sense @aramaraju. Without hypothesis on the the other metrics it doesn't even make sense to launch these. The hypothesis that I am more interested in is the effect of these metrics: `# Code suggestions`, `User Retention` which I expect to significantly drop. If you think there are other metrics which could be useful, we should definitely start tracking that. I really like the @rafax's suggestion on [slack thread](https://sourcegraph.slack.com/archives/C0729T2PBV2/p1718357408712179) to keep track of these additional metrics (CPW & CPO) to get more holistic insights into what is happening 

> It's OK to run it in this control group. It's still relatively small, and we need a sizable group to get more realistic results.
@valerybugakov Based on the past experience if you think this number is still small, should be increase it further to ~300-400 msec ?

vscode/src/completions/providers/fireworks.ts
hitesh-1997: Added huggingface link on top of them - which is more detailed :)

vscode/src/completions/providers/fireworks.ts
hitesh-1997: I also felt a bit uneasy while making changes -- I realize now this was due to this variable name 😅
I mixed python, ts -- fixed this, thanks for pointing it out !!

vscode/src/completions/providers/fireworks.ts
hitesh-1997: Added, thanks for pointing it out !!

vscode/src/completions/providers/fireworks.ts
hitesh-1997: I didn't use logDebug as it will start popping up for users on every completions and will add noise. I think this is more for us developers while debugging to point out that: "If you added a new model, you need to implement the infill logic for it, else it will use the default one"

vscode/src/completions/providers/fireworks.ts
```diff
@@ -435,6 +438,19 @@ class FireworksProvider extends Provider {
         return tracer.startActiveSpan(
             `POST ${url}`,
             async function* (span): CompletionResponseGenerator {
+                if (self.model === FIREWORKS_FIM_FINE_TUNED_MODEL_HYBRID_WITH_200MS_DELAY) {
+                    // Todo: Remove the condition after the experiment is complete and we have the relevant data points.
+                    // This delay introduced here is for the experimentation purpose to see the effect of latency on other metrics, such as CAR, wCAR, Retention, #Sugeestions etc.
+                    await new Promise(resolve => setTimeout(resolve, 200))
+                }
+                if (abortController.signal.aborted) {
```
hitesh-1997, line 446: @valerybugakov Can you please verify if this condition looks okay ?

valerybugakov: It looks good to me. 

If it were a long-term option, we would want to make it so the upstream consumer knows it's not an empty completion but a canceled one. But it's unimportant for the A/B test we plan to run so that we can keep it like this.

vscode/src/completions/providers/fireworks.ts
valerybugakov: I do not think it's small. Let's keep 200ms.

PR #4579, opened 2024-06-17
fix flaky test

beyang:
It's better to use a FrameLocator rather than a Frame in playwright, as a Frame can be invalidated by a navigation action. Fixes a sometimes-flaky test by using an appropriate FrameLocator, instead of a Frame, to target the Cody sidebar.

## Test plan

Test-only change.

### DIFF
--- a/vscode/test/e2e/helpers.ts
+++ b/vscode/test/e2e/helpers.ts
@@ -302,7 +302,7 @@ export const test = base
             resetLoggedEvents()
         },
     })
-    .extend<{ sidebar: Frame | null; getCodySidebar: () => Promise<Frame> }>({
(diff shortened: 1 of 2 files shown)


sqs (approved):

olafurpg:
The test here still seems to be flaky, I saw it failing three times in my PR here https://github.com/sourcegraph/cody/pull/4591#issuecomment-2174170405

toolmantim:
Same here @olafurpg — up to retry number 11 on this test in #4584: https://github.com/sourcegraph/cody/actions/runs/9556626326/job/26349471607 — strangely there's no artifact upload happening either.

PR #4580, opened 2024-06-17
Chat: Add a stop button and tooltips w/ keyboard shortcuts, and reduce vertical space

toolmantim:
Lots of visual details and cleanups to the core chat transcript UI:

* Updates the chat submit button so it turns into a stop button
* Adds tooltips with keyboard shortcuts to the send button
* Ensures tooltips don't overlap the message input
* Removes a lot of unnecessary vertical padding
* Fixes the alignment and sizing of the input, @-mention chips, avatars

CleanShot 2024-06-17 at 16 54 51@2x

CleanShot 2024-06-17 at 16 52 52@2x

CleanShot 2024-06-17 at 16 28 05@2x

CleanShot 2024-06-17 at 16 28 12@2x

Cleanups:

* Implements a better Kdb component for showing keyboard shortcuts
* Adds tooltip stories showing how to display a keyboard shortcut in the tooltip
* Adds button stories
* Change the tailwind button class name styles
* Updates tailwind font size maths so they match even font sizes (and match VS Code)

Known bug:

* Hitting escape/stop before the LLM stream starts puts the chat in a bad state

Fixes https://linear.app/sourcegraph/issue/CODY-2149/feedback-bring-the-stop-button-back-again
Fixes https://linear.app/sourcegraph/issue/CODY-2270/the-amount-of-vertical-space-the-chat-takes

## Test plan

- CI
- Test the chat in various states


vscode/webviews/components/Kbd.tsx
```diff
@@ -1,11 +1,67 @@
...
+} from 'lucide-react'
 import type { FunctionComponent } from 'react'
+import { cn } from './shadcn/utils'
 
 const isMac = isMacOS()
 
+function keyTextOrSvg(key: string): React.ReactElement | string {
+    const iconClassName = 'tw-w-[1em] tw-h-[1em]'
+
+    if (isMac && (/opt/i.test(key) || /option/i.test(key))) {
```
dominiccooney, line 18: Could we simplify this and be hardasses and just require lowercase strings, and check an exact string?

```
if (isMac) {
  switch (key) {
    case 'opt':
      return <OptionIcon className={iconClassName} />
    case 'cmd':
      return <CommandIcon className={iconClassName} />
    ...
    default:
      break;
  }
}
switch (key) {
case 'return':
  ...
default:
  return <span>{key}</span>
}
```

dominiccooney, line 18: `(/opt/i.test(key) || /option/i.test(key)` is redundant because `/opt/i` will match anything `/option/i` matches.

vscode/webviews/tailwind.config.mjs
```diff
@@ -8,10 +11,10 @@ export default {
     theme: {
         extend: {
             fontSize: {
-                lg: 'calc(1.1*var(--vscode-font-size))',
-                md: 'var(--vscode-font-size)',
-                sm: 'calc(0.9*var(--vscode-font-size))',
-                xs: 'calc(0.85*var(--vscode-font-size))',
+                lg: 'calc(calc(15/13)*var(--vscode-font-size))', // = 15px
```
dominiccooney, line 14: Why nest calc in this way? etc.
```suggestion
                lg: 'calc(var(--vscode-font-size) * 15 / 13)', // = 15px
```

vscode/webviews/tailwind.config.mjs
```diff
@@ -66,13 +76,27 @@ export default {
...
             },
+            animation: {
+                // spin: 'spin 1s cubic-bezier(0.53, 0.21, 0.29, 0.67) infinite',
+            },
         },
     },
+    plugins: [
+        plugin(({ addVariant }) => {
+            // Allows use to customize styling for VS Code light and dark themes
+            addVariant('high-contrast-dark', 'body[data-vscode-theme-kind="vscode-high-contrast"] &')
```
dominiccooney, line 95: (FWIW, I am thinking we should add 'eclipse', 'vs' and 'jetbrains' as DOM attribute values somewhere, like bady[data-ide=...] if there's any need for IDE-specific adjustments.)

dominiccooney (approved):
LGTM, some bikeshedding about names at your discretion.

The flakiness of the tests are naturally concerning, I noticed a manual frame lookup was changed to frameLocator recently to support navigations. Of course we would have used a frame locator if we could! Wonder if that is related.

vscode/webviews/tailwind.config.mjs
toolmantim: Great idea… and we can add them as variants that target those DOM attribute values.

vscode/webviews/chat/cells/messageCell/human/editor/toolbar/Toolbar.tsx
```diff
@@ -67,21 +65,7 @@ export const Toolbar: FunctionComponent<{
...
-                                    Add files and other context <Kbd macOS="@" linuxAndWindows="@" />
-                                </>
-                            }
-                            iconStart={AtSignIcon}
-                            onClick={onMentionClick}
-                            aria-label="Add context"
-                        />
-                    </span>
-                )}
+                {onMentionClick && <MentionButton onClick={onMentionClick} className="tw-opacity-60" />}
```
thenamankumar, line 68: Can you rename this to `AddContextButton`?

MentionButton is a bit confusing.

thenamankumar (approved):

vscode/webviews/chat/cells/contextCell/ContextCell.tsx
```diff
@@ -74,63 +80,80 @@ export const ContextCell: React.FunctionComponent<{
...
-                        <li>
-                            <Tooltip>
-                                <TooltipTrigger asChild>
+                        </AccordionTrigger>
+                        <AccordionContent>
+                            <ul className="tw-list-none tw-flex tw-flex-col tw-gap-2 tw-pt-2">
+                                {contextItems?.map((item, i) => (
+                                    // biome-ignore lint/suspicious/noArrayIndexKey: stable order
+                                    <li key={i} data-testid="context-item" className="tw-max-w-full">
+                                        <FileLink
```
thenamankumar, line 107: I wish FileLink accepted `props.item`. Non blocking.

toolmantim: Yeah… I need to file a separate Linear issue to update the way we display the context items as well. We can redo then.

vscode/webviews/chat/Transcript.tsx
```diff
@@ -35,7 +34,7 @@ export const Transcript: React.FunctionComponent<{
         [transcript, messageInProgress]
     )
     return (
-        <>
+        <div className="tw-px-8 tw-pt-8 tw-pb-14 tw-flex tw-flex-col tw-gap-10">
```
dominiccooney, line 37: Oh I am feeling Tailwind so hard right now 🤮

vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.module.css
```diff
@@ -1,47 +1,49 @@
...
+    background-color: var(--vscode-input-background);
 }
 
-/* Hide the toolbar unless focused on all but the last cell. */
-[role="row"]:has(.container:is(:not(.sent), .focused)) {
-    /* HACK: Change HumanMessageCell style. */
-    padding-bottom: var(--human-message-editor-cell-spacing-bottom);
+/* Show focus ring when the input is focused or a toolbar button within the input container is
+   focused, but NOT if something in a toolbar button's popover is focused. */
+.container:has([data-lexical-editor='true']:focus-within, menu>button:focus) {
```
dominiccooney, line 43: Thank you, we've had this kind of problem with the enhanced context selector popup and it really grates. Lovely comment.

May make sense to refer to the menu/button by a class name or something more specific so this doesn't break when the popup adds a menu+button...

dominiccooney (approved):
Looks great.

PR #4585, opened 2024-06-17
Add ide version and proper Cody extension version to the telemetry

pkukielka:
## Changes

1. Report proper extension version (previously Cody version was incorrectly used for both Jetbrains and NeoVim)
2. Add IDE version to the telemetry data 
 
## Test plan

1. Build [Jetbrains PR](https://github.com/sourcegraph/jetbrains/pull/1785) with this Cody changes
2. Run some Cody actions like commands and edit.
4. Wait ~2h
5. Check if event with new field(s) containing correct versions will appear in looker

### DIFF
--- a/agent/src/AgentWorkspaceConfiguration.ts
+++ b/agent/src/AgentWorkspaceConfiguration.ts
@@ -88,6 +88,10 @@ export class AgentWorkspaceConfiguration implements vscode.WorkspaceConfiguratio
                 return extensionConfig?.codebase
             case 'cody.advanced.agent.ide':
                 return this.clientNameToIDE(this.clientInfo()?.name ?? '')
+            case 'cody.advanced.agent.ide.version':
+                return this.clientInfo()?.version
+            case 'cody.advanced.agent.extension.version':
+                return this.clientInfo()?.ideVersion
             case 'editor.insertSpaces':
                 return true // TODO: override from IDE clients
             default:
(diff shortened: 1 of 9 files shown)


mkondratek (approved):
lgtm

agent/src/AgentWorkspaceConfiguration.ts
```diff
@@ -88,6 +88,10 @@ export class AgentWorkspaceConfiguration implements vscode.WorkspaceConfiguratio
                 return extensionConfig?.codebase
             case 'cody.advanced.agent.ide':
                 return this.clientNameToIDE(this.clientInfo()?.name ?? '')
+            case 'cody.advanced.agent.ide.version':
+                return this.clientInfo()?.version
```
dominiccooney, line 92: These don't look right...

`cody.advanced.agent.*ide*.version` is `this.clientInfo()?.version`, but
`cody.advanced.agent.*extension*.version` is `this.clientInfo()?.ideVersion` ?

If this is mixed up let's reverse them, if the data are right, let's rename some of these methods because it's confusing. 😰

dominiccooney (changes requested):

agent/src/AgentWorkspaceConfiguration.ts
pkukielka: Good catch, they were reversed by mistake!

pkukielka: BTW. here is example of the of event from looker before the reversal (I removed irrelevant content):

```
{
    "multiline": false,
    "triggerKind": "Automatic",
    "providerIdentifier": "fireworks",
    "providerModel": "starcoder-hybrid",
    "languageId": "kotlin",
    "testFile": false,
    "artificialDelay": 0,
    "multilineMode": null,
    "source": "Cache",
    "otherCompletionProviderEnabled": false,
    "otherCompletionProviders": [],
    "serverEndpoint": "https://sourcegraph.com/",
    "extensionDetails": {
        "ide": "JetBrains",
        "ideVersion": "6.0-localbuild",
        "ideExtensionType": "Cody",
        "platform": "macos",
        "arch": "aarch64",
        "version": "2024.1.2"
    },
    "tier": 1,
    "version": "2024.1.2",
    "hasV2Event": true
}
```

PR #4574, opened 2024-06-15
fix symf instanceof CancellationError test flake

sqs:
In unit tests, the `vscode` environment is not fully mocked, so `vscode.CancellationError` is `undefined`. Just use our `AbortError` instead.

Example: https://github.com/sourcegraph/cody/actions/runs/9526461106/job/26261889928?pr=4407#step:9:1806

## Test plan

CI

PR #4588, opened 2024-06-17
Revert "fix symf instanceof CancellationError test flake (#4574)"

olafurpg:
This reverts commit 8ac4a7a. This commit made the tests flaky. Instead of removing usage of `CancellationError`, we implement this class in the mocks instead.

## Test plan
Non-flaky green CI

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -63,6 +63,7 @@ export { AgentEventEmitter as EventEmitter } from '../../vscode/src/testutils/Ag
 
 export {
     CancellationTokenSource,
+    CancellationError,
     CodeAction,
     CodeActionTriggerKind,
     CodeActionKind,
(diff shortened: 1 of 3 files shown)


olafurpg:
Closing since I'm still seeing the same CI failure
```
 FAIL  |agent| src/index.test.ts > Agent > Text documents > chat/submitMessage (understands the selected text)
AssertionError: the given combination of arguments (undefined and string) is invalid for this assertion. You can use an array, a map, an object, a set, a string, or a weakset instead of a string
 ❯ src/index.test.ts:649:41
    647|                 { addEnhancedContext: true }
    648|             )
    649|             expect(reply?.text?.trim()).includes('anotherFunction')
       |                                         ^
    650|             expect(reply?.text?.trim()).not.includes('inner')
    651|             await client.changeFile(multipleSelectionsUri)

```

dominiccooney:
@olafurpg assuming we want this closed... closing.

olafurpg:
@dominiccooney yes, sorry, this PR should be closed

PR #4591, opened 2024-06-17
CI: disable symf in agent integration tests

olafurpg:
Previously, symf was enabled in agent integration tests. We have observed quite a lot of unstable test failures that seem to be related to symf so this PR disabled symf entirely in the agent integration tests to see if it stablizes the CI.

## Test plan

Non-flaky green CI.

PR #4593, opened 2024-06-17
Reenable some core chat tests

jtibshirani:
In #4591, we disabled enhanced context for a few chat tests. For some tests,
having extra context is central to what they're checking. This PR creates adds
mock enhanced context to these tests.

## Test plan

Reran unit tests multiple times. Then deleted and updated recordings, reran
tests a bunch of times. Also sanity checked recordings.


agent/src/index.test.ts
```diff
@@ -372,14 +382,15 @@ describe('Agent', () => {
             )
         }, 30_000)
 
-        it.skip('chat/submitMessage (addEnhancedContext: true, squirrel test)', async () => {
+        it('chat/submitMessage (with enhanced context, squirrel test)', async () => {
```
jtibshirani, line 385: I reenabled these tests with static, fake enhanced context. If they fail or flake again, it'd represent a serious issue we need to look into.

olafurpg (approved):
LGTM 👍🏻 Thank you for following up on this!

As a followup change or in this PR, can we move the enhanced context tests to a separate file? I've been trying to break down `index.test.ts` into more focused test files and having `enhanced-context.test.ts` would be a nice improvement.

PR #4595, opened 2024-06-18
Autocomplete: implement stage counter logger

valerybugakov:
- Implements a new autocomplete stage logger, which counts the completion requests that reached different stages of the autocomplete generation pipeline.
- It is very similar to https://github.com/sourcegraph/cody/pull/3070, so I assume that "this event is probably a good candidate to exclude from the Amplitude pipeline" based on Kelsey's comment [here](https://github.com/sourcegraph/cody/pull/3070#issuecomment-2003948272). cc @akalia25 
- Closes https://linear.app/sourcegraph/issue/CODY-2413/%5Bautocomplete-latency%5D-add-feedback-rate-logger
- [Slack thread](https://sourcegraph.slack.com/archives/C0729T2PBV2/p1718627075338329?thread_ts=1718357408.712179&cid=C0729T2PBV2)

## Test plan

- Added unit tests
- Ensured that dispose is called and we flush when VS Code is closed

Screenshot 2024-06-18 at 13 48 13

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -17,7 +17,7 @@ import type { CodyStatusBar } from '../services/StatusBar'
 import { telemetryService } from '../services/telemetry'
 
 import { type CodyIgnoreType, showCodyIgnoreNotification } from '../cody-ignore/notification'
-import { autocompleteFeedbackRateLogger } from '../services/autocomplete-feedback-rate-logger'
+import { autocompleteStageCounterLogger } from '../services/autocomplete-stage-counter-logger'
 import { recordExposedExperimentsToSpan } from '../services/open-telemetry/utils'
 import { isInTutorial } from '../tutorial/helpers'
 import { type LatencyFeatureFlags, getArtificialDelay, resetArtificialDelay } from './artificial-delay'
@@ -389,7 +389,7 @@ export class InlineCompletionItemProvider
                 // Do not increment the `preFinalCancellationCheck` counter if the result is empty.
                 // We don't have an opportunity to show a completion if it's empty.
                 if (result) {
-                    autocompleteFeedbackRateLogger.record('preFinalCancellationCheck')
+                    autocompleteStageCounterLogger.record('preFinalCancellationCheck')
                 }
 
                 // Avoid any further work if the completion is invalidated already.
@@ -432,7 +432,7 @@ export class InlineCompletionItemProvider
                     )
                 )
 
-                autocompleteFeedbackRateLogger.record('preVisibilityCheck')
+                autocompleteStageCounterLogger.record('preVisibilityCheck')
 
                 // A completion that won't be visible in VS Code will not be returned and not be logged.
                 if (visibleItems.length === 0) {
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -387,7 +387,7 @@ export class InlineCompletionItemProvider
                 })
 
                 // Do not increment the `preFinalCancellationCheck` counter if the result is empty.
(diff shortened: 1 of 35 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -385,6 +386,12 @@ export class InlineCompletionItemProvider
                     isDotComUser: this.config.isDotComUser,
                 })
 
+                // Do not increment the `preFinalCancellationCheck` counter if the result is empty.
+                // We don't have an opportunity to show a completion if it's null.
+                if (result) {
+                    autocompleteFeedbackRateLogger.record('preFinalCancellationCheck')
+                }
```
valerybugakov, line 393: @hitesh-1997, let me know if you think we don't need a condition here

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -385,6 +386,12 @@ export class InlineCompletionItemProvider
                     isDotComUser: this.config.isDotComUser,
                 })
 
+                // Do not increment the `preFinalCancellationCheck` counter if the result is empty.
+                // We don't have an opportunity to show a completion if it's null.
```
dominiccooney, line 390: You use two different phrases, `empty` and `null`, to refer to the same state. It appears that you're making a distinction, which is confusing. Don't use different words unless you want to make a distinction.
```suggestion
                // Do not increment the `preFinalCancellationCheck` counter if the result is empty.
                // We don't have an opportunity to show a completion if it's empty.
```

vscode/src/services/autocomplete-feedback-rate-logger.ts
```diff
@@ -0,0 +1,59 @@
+import type * as vscode from 'vscode'
+
+import { telemetryRecorder } from '@sourcegraph/cody-shared'
+
+export const LOG_INTERVAL = 30 * 60 * 1000 // 30 minutes
+
+/**
+ * Keys represent autocomplete pipeline stages in chronological order.
+ */
+const INITIAL_STATE = {
```
dominiccooney, line 10: Very succinct typing here with this flowing into `currentState` and then using `keyof typeof this.currentState`. Neat.

vscode/src/services/autocomplete-feedback-rate-logger.ts
```diff
@@ -0,0 +1,59 @@
...
+    preLastCandidate: 0,
+    preCache: 0,
+    preDebounce: 0,
+    preContextRetrieval: 0,
+    preNetworkRequest: 0,
+    preFinalCancellationCheck: 0,
+    preVisibilityCheck: 0,
+}
+
+export class AutocompleteFeedbackRateLogger implements vscode.Disposable {
```
dominiccooney, line 20: This isn't really a *rate* but counts... `AutocompleteStepCounter`? 🚶 ⌚

dominiccooney (approved):
Beautiful.

vscode/src/services/autocomplete-feedback-rate-logger.ts
valerybugakov: Agreed, we do not calculate the rate in this logger, but we plan to use these counts to calculate [the feedback rate](https://codeium.com/blog/golden-metrics-characters-per-opportunity-percentage-code-written#:~:text=not%20to%20try.-,Feedback%20Rate,-%3A%20There%20is%20obvious) later. Will update 👍

akalia25:
Thanks for the heads-up @valerybugakov, updated our Amplitude pipeline to not ingest this event. 

CC: @kelsey-brown

hitesh-1997 (approved):
Thanks for adding this change Valery !!

vscode/src/completions/inline-completion-item-provider.ts
hitesh-1997: should be okay to have i think

PR #4598, opened 2024-06-18
Windows Certificates: bundle roots.exe

olafurpg:
Fixes CODY-2176

Previously, we used the `win-ca` "fallback" method to install local certificates. We used the fallback because we use esbuild to bundle the Cody VSC extension, which breaks automatic detection of a `roots.exe` binary that is included with the `win-ca` npm package. In the latest VS Code Insiders, this no longer works and causes Cody to fail all network requests. This PR is an attempt to fix this problem by removing the "fallback", bundling the `roots.exe` binary along with the Cody extension and adding logic to compute the file path of the bundled `roots.exe` at runtime (via `vscode.ExtensionContext`).

## Test plan

* Get a Windows computer
* Install latest VS Code Insiders
* Install the latest Cody pre-release and reproduce the actual bug (network requests fail)
* Checkout this branch and run from the `vscode/` sub-directory
```
 .\node_modules\.bin\vsce package --no-dependencies --pre-release --no-update-package-json --no-git-tag-version --out .\dist\cody.vsix
```
* Run "Install from VSIX" command from VS Code insiders and pick dist\cody.vsix
* Reload and confirm that network requests work as expected
* I don't know how to verify this, but it would be good to test it actually respects local certificates. I only confirmed that win-ca doesn't crash.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -48,7 +48,8 @@
     "github-changelog": "ts-node-transpile-only ./scripts/github-changelog.ts",
     "version-bump:minor": "RELEASE_TYPE=minor ts-node-transpile-only ./scripts/version-bump.ts",
     "version-bump:patch": "RELEASE_TYPE=patch ts-node-transpile-only ./scripts/version-bump.ts",
-    "version-bump:dry-run": "RELEASE_TYPE=prerelease ts-node-transpile-only ./scripts/version-bump.ts"
+    "version-bump:dry-run": "RELEASE_TYPE=prerelease ts-node-transpile-only ./scripts/version-bump.ts",
+    "win-ca:bundle": "cp ../node_modules/win-ca/lib/roots.exe dist/win-ca-roots.exe"
   },
   "categories": ["Programming Languages", "Machine Learning", "Snippets", "Education"],
   "keywords": [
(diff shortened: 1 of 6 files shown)


vscode/package.json
```diff
@@ -48,7 +48,8 @@
     "github-changelog": "ts-node-transpile-only ./scripts/github-changelog.ts",
     "version-bump:minor": "RELEASE_TYPE=minor ts-node-transpile-only ./scripts/version-bump.ts",
     "version-bump:patch": "RELEASE_TYPE=patch ts-node-transpile-only ./scripts/version-bump.ts",
-    "version-bump:dry-run": "RELEASE_TYPE=prerelease ts-node-transpile-only ./scripts/version-bump.ts"
+    "version-bump:dry-run": "RELEASE_TYPE=prerelease ts-node-transpile-only ./scripts/version-bump.ts",
+    "win-ca:bundle": "cp ../node_modules/win-ca/lib/roots.exe dist/win-ca-roots.exe"
```
olafurpg, line 52: TODO: run this step as part of prepublish. I wasn't able to run this locally because 
```
'cp' is not recognized as an internal or external command,
operable program or batch file.
```

vscode/src/fetch.node.test.ts
```diff
@@ -30,7 +31,7 @@ describe('customAgent', () => {
     })
 
     it('uses keep-alive', async () => {
-        initializeNetworkAgent()
+        initializeNetworkAgent({ extensionUri: vscode.Uri.parse('file:///foo') })
```
arafatkatze, line 34: Does this need to be updated?

arafatkatze (approved):
Approving because of this 
image

I can also test the self signed certs logic but that takes a while to setup if you really wanna get this in just to test things thats fine by me too.

vscode/src/fetch.node.test.ts
olafurpg: The test passes on macOS. I haven't checked on Windows, but we don't need a valid path to roots.exe in the tests so I don't think it's needed

olafurpg:
This PR fixed the network errors I was able to reproduce on a vanilla installation on Windows with VS Code Insiders. If self-signed local certificates are not working then we should open a separate feature request for that. Most importantly, we can't break the experience for all Windows users to support this functionality.

PR #4604, opened 2024-06-18
openctx: set query label if provided

keegancsmith:
This sets the query label to the newly added "label" field in OpenCtx. This is a field which prompts the user what to type in an at mention.

Test Plan: Tested manually the following cases

- non-openctx providers have same behaviour (no header)
- openctx provider has loading display before query label (see devdocs below)
- openctx query label goes away once query is being typed
- openctx provider which doesn't set label shows "Search..." (see notion below)
- web provider has updated label

https://github.com/sourcegraph/cody/assets/187831/6b3ade90-13b2-43bb-8ba9-e3e822514679

Parent https://github.com/sourcegraph/cody/pull/4621

Fixes https://linear.app/sourcegraph/issue/CODY-2275/add-all-the-right-labels-for-provider-mentions-query-stats

### DIFF
--- a/lib/shared/src/mentions/api.ts
+++ b/lib/shared/src/mentions/api.ts
@@ -74,7 +74,7 @@ async function openCtxMentionProviders(): Promise<ContextMentionProviderMetadata
             .map(provider => ({
                 id: provider.providerUri,
                 title: provider.name,
-                queryLabel: 'Search...',
+                queryLabel: provider.mentions?.label ?? 'Search...',
                 emptyLabel: 'No results',
             }))
             .sort((a, b) => (a.title > b.title ? 1 : -1))
--- a/vscode/src/context/openctx/web.ts
+++ b/vscode/src/context/openctx/web.ts
@@ -9,7 +9,7 @@ const WebProvider: Provider & { providerUri: 'internal-web-provider' } = {
     meta() {
         return {
             name: 'Web URLs',
-            mentions: { label: 'Paste a URL...' },
+            mentions: { label: 'Type or paste a URL...' },
         }
     },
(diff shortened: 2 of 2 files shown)


toolmantim (approved):
This is great!

Seeing it IRL I think the URL ones might need to say “Type or paste a URL…” because otherwise it feels like you paste them into popover or something.

I still think it’s a mistake we don’t have “Search files…” and “Search symbols…” for the built-in files and symbols. Would be great to re-add imo.

keegancsmith:
> Seeing it IRL I think the URL ones might need to say “Type or paste a URL…” because otherwise it feels like you paste them into the popover or something.

Done

> I still think it’s a mistake we don’t have “Search files…” and “Search symbols…” for the built-in files and symbols. Would be great to re-add imo.

Will follow-up in another PR

keegancsmith:
> I still think it’s a mistake we don’t have “Search files…” and “Search symbols…” for the built-in files and symbols. Would be great to re-add imo.

So just to confirm, you want to make these slightly different to OpenCtx? For OpenCtx this is the behaviour:
- Loading :: "meta.name" / Loading
- No query and no items :: "meta.name" / "meta.mentions.label"
- Query :: "meta.name" / items...

This is the current file behaviour:
- Loading :: Loading
- No query and no items :: Search for a file to include...
- items :: items...

The current behaviour kinda makes sense. Because we kinda insta-load symbols and files I'm not sure we need a prompt, but it would make sense from a consistency point of view. To help make this decision I greatly improved the cases in our story book. Check this PR out. https://github.com/sourcegraph/cody/pull/4624

PR #4605, opened 2024-06-18
Prepare Cody web library API for Sourcegraph Cody Web client 

vovakulikov:
Original PR https://github.com/sourcegraph/cody/pull/4227

Closes [SRCH-573](https://linear.app/sourcegraph/issue/SRCH-573/integrate-cody-web-package-into-the-sourcegraph-ui) (Main issue for cody agent and vscode extension logic change)

This PR prepares VSCode Cody extension and agent logic for future use in Cody Web.
It also updates `cody/web` package and its demo for testing purposes. After we merge this PR we should publish `cody/web` package to npm and use it in the sourcegraph repository as a common npm lib package.

## Noticeable changes
- Since the agent will be running inside Web Worker for Cody Web, this PR adds support for custom storage (local storage doesn't work in Web Worker but IndexDB does). Because of this change, I had to "promisify" the whole API around localStorage service abstraction. Most changes files (more than have) have diff because of this change
- File and Symbol resolution can now work with remote repositories. Since there are no local files in Cody Web I added support for remote resources in context items (mentions resolutions), also common context also works with remote repositories. 
- Some changes in agent API (nothing big more like extend changes to support Cody web client) 
- Big update in `cody/web` client, basically it got close to Sourcegraph demo client usage and supports chat history, initial context, open context support, and mentions. 

## Todo before merge
- [x] Support custom local storage API 
- [x] Solve the problem with Cody's web-style isolation 
- [x] Fix unit tests
- [x] Fix integration tests
- [ ] Update Cody agent Kotlin bindings 
- [x] Fix Cody agent unit tests
- [x] Fix e2e tests (most tests I got fixed, but here is a list with left tests that are failing)
   - [x]  chat-atFile.test.ts:347:3 › @-mention symbol in chat 
   - [x]  command-core.test.ts:53:3 › Explain Command & Smell Command & Chat from Command Menu
   - [x]  command-edit.test.ts:126:1 › edit (fixup) input - range selection 
   - [x]  initial-context.test.ts:35:18 › initial context - file

## How to test it locally
To run cody web demo go to `./web` and run `pnpm dev`
To run vscode extension command use standard flow (this PR doesn't change anything about them)
To run full Sourcegraph Cody Web demo see instructions in Sourcegraph PR [here](https://github.com/sourcegraph/sourcegraph/pull/62792)

## Test plan
- Check that the Cody web demo and Cody web package build work properly
- Check manually that the VSCode Cody extension work properly


vscode/src/main.ts
```diff
@@ -97,13 +97,32 @@ export async function start(
     platform: PlatformContext
 ): Promise<vscode.Disposable> {
     // Set internal storage fields for storage provider singletons
-    localStorage.setStorage(context.globalState)
+    localStorage.setStorage(
+        platform.createStorage ? await platform.createStorage() : context.globalState
```
olafurpg, line 101: Can we implement `context.globalState` under the hood with `platform.createStorage()`? As a rule of thumb, I want to keep the number of if/else conditions for the agent to an absolute minimum. These if conditions make it harder to reason about the system. The aspiration is that the agent implements the VS Code APIs and people working on the Cody VSC codebase should only have to reason about how things work in VSC.

vscode/src/main.ts
```diff
@@ -97,13 +97,32 @@ export async function start(
...
     if (secretStorage instanceof VSCodeSecretStorage) {
         secretStorage.setStorage(context.secrets)
     }
 
     setLogger({ logDebug, logError })
 
+    // In agent's run we should seed configuration with endpoint and access token
+    // before we hit auth provider init call in order to avoid double auth() calls
+    // during extension activation
+    if (isRunningInsideAgent()) {
```
olafurpg, line 112: This logic should live inside the agent. We already implement `vscode.workspace.getConfiguration` in the agent, there is no need to add this condition in the VSC-part of the codebase.

vscode/src/main.ts
```diff
@@ -167,7 +187,13 @@ const register = async (
         })
     )
 
-    await authProvider.init()
+    const authStatus = await authProvider.init()
+
+    // Sync models as soon as possibles in order to avoid runtime error
+    // during chat management services initialization
```
olafurpg, line 193: FYI, we already send quite a lot of blocking network requests on startup and I've wanted to reduce these requests. While it's convenient to perform this blocking request here, it slows down extension startup. It would be a better solution to await on the syncing from the locations where the runtime errors happened.

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -13,7 +13,30 @@ import {
 
 import { isSourcegraphToken } from '../chat/protocol'
 
+export interface AsyncMemento {
```
olafurpg, line 16: Are you certain that we need `AsyncMemento`? Most of the "high-risk" diffs in this PR are related to awaiting on the storage, I would feel more comfortable approving this PR if we can stick with `Memento`. We can implement non-async `Memento` by loading a JSON key from indexdb during startup and then performing get/keys synchronously. The `update` function is already async in the `vscode.Memento` API.

As a rule of thumb, we should try to stick with VS Code APIs wherever possible. We should have a high barrier to make exceptions to this rule.

olafurpg (changes requested):
Thank you for this contribution! It's very exciting to see us trying to use this infrastructure in the browser. The main blocking comments are

- Let's try to avoid adding `AsyncMemento`
- Let's try to avoid blocking on `syncModels` during startup

In general, let's try to keep the diff in `vscode/src` as small and low-risk as possible. It's better to pay a complexity once in `agent/src` instead of requiring all Cody VSC devs to reason about different environments.

olafurpg:
Also, can you update the PR description to link to the Linear issue/project?

olafurpg:
Also, please tag the Cody Agent GH team for future PRs related to the agent. There is a lot of important context to why things are implemented the way they are. Just having a green CI and a stamp from somebody in Cody Web should not be enough to get a large PR like this merged.

vscode/src/main.ts
vovakulikov: Fun things that this seed-like logic fixes problems with double auth network requests flows. The problem with mocking configuration inside the agent is that it doesn't take into account the full logic in the extension main setup function. 

The problem with seeding this data in the agent comes from the fact that we don't have access (and we shouldn't have actually) to abstractions like `localStorage` and `secretStorage`. Storage for `SecretStorage` is created inside the agent but it provides just in-memory map implementation that doesn't know anything `SecretStorage` abstraction where we store keys and methods to set accessToken data. 

I understand the comment, this change though with isRunningInsideAgent seemed as the safest path to avoid the problem with auth/tokens problems in Cody Web Agent. (if you have any ideas on how to solve it better within the agent I would be happy to make a change, it might be I missed something obvious there)

vscode/src/services/LocalStorageProvider.ts
vovakulikov: > Most of the "high-risk" diffs in this PR are related to awaiting the storage

Do you have any specific places where it concerns you? I can try the idea of loading data ahead during initialisation. Thanks for the suggestion

vscode/src/main.ts
vovakulikov: To be honest, I wanted to put this also behind the `isRunningInsideAgent` check, something happens inside the extension in the agent differently when we run pure extension logic in the VSCode. ChatManagement and other chat services expect to have non-empty chat models by the time we call them (later in this main.ts function). 

I could try to change logic around the creation of chat panels and simpleChatProviders but it seemed a big change (at least as I saw it) which I would like to avoid.

vovakulikov:
@olafurpg 
> Also, please tag the Cody Agent GH team for future PRs related to the agent

What do you mean? I tagged Cody's agent team in this PR, is there any other team rather than this one? 

Screenshot 2024-06-24 at 01 41 12

vscode/src/main.ts
vovakulikov: I have no preferences here about this API, it seemed to me that custom storage falls within platform context dependencies (which is a part of this extension API function). We could pass custom storage to the Agent class and then set it with custom `globalState` if you think this is more correct to do here.

vscode/src/services/LocalStorageProvider.ts
vovakulikov: There might be a problem with the sync approach in IndexDB, so if we pull data from IndexDB during init we would have some in-memory internal field in IndexDB storage, so yeah as you said this makes get/keys methods synchronous, in update we could update value in IndexDB and then apply the same value to the storage. 

The problem here might be when we have more than one browser tab open, it looks like it would be a case when one tab writes one data but the other tab doesn't get this updated value and its get/key method (since it doesn't call any IndexDB API) will return just in-memory value (the old value)

Usually, I would say that we could subscribe on the table changes but it seems that IndexDBObserver never hit the wide support in most of the browsers

cc @olafurpg

vovakulikov: Just for the record (in defence of making storage async) most of the places (99% of cases) where get/keys methods are called already were async so the only change that I had to make there just to add `await`, which I doubt would cause big problems 

The change where I had to make more changes was class constructors, but the static create function was my way out of this which doesn't introduce anything new to the class creating logic but mostly duplicates logic from the constructor

vscode/src/main.ts
vovakulikov: I just noticed that we already called this `syncModels` fetch function later in the `main.ts`, I removed the second later call to avoid double fetching, but it seems that calling it early shouldn't affect performance in the extension, correct me if I'm wrong

sqs:
Symbols and chat history do not need to be in the initial version that ships to dotcom. Ideally they would make it in the July 10 release, but if they need to slip, I would still want to get everything else shipped on time rather than delay the whole thing.

vovakulikov:
@sqs I greened CI today, unit, integrations and e2e tests are finally passing, so now I know that these changes are actually possible and don't break core logic (but I haven't checked other agent's consumers but Cody web), We had a conversation today with @olafurpg and even these changes seem ok in terms of tests but there are still concerns about corners cases. I will split this PR into smaller ones today/tomorrow to make it a bit safer to merge

Screenshot 2024-06-26 at 03 42 58

While I'm doing this, on the other side we want to ship the new Cody Web as soon as possible on dotcom, @peterguy suggested that we actually could already ship this to dot com with publishing the updated Cody agent from the brunch, I already did this in sourcegraph PR here https://github.com/sourcegraph/sourcegraph/pull/62792

With this PR in sg repo, the new Cody web client will be available only for signed-in users on Dotcom, so no risks for any non-web agent's clients and we already use (used) the Cody-web package which exists only in the npm, so I think this is a low-risk good compromise at this point, while I'm splitting and merging big changes in Cody agent. 

cc @peterguy

agent/recordings/chat-response-quality_413888585/recording.har.yaml
```diff
@@ -34414,6 +34414,106 @@ log:
...
+        headersSize: 324
+        httpVersion: HTTP/1.1
+        method: POST
+        postData:
+          mimeType: application/json; charset=utf-8
+          params: []
+          textJSON:
+            query: |-
+              
+              query Repository($name: String!) {
```
olafurpg, line 34454: I see we are already sending `query Repository` here https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/recordings/chat-response-quality_413888585/recording.har.yaml?L34360

Any idea why we are sending a new request?

vovakulikov: @olafurpg Yes, I actually think this is because of codebase checks in `configureExternalServices`
https://sourcegraph.com/github.com/sourcegraph/cody@df399f24938288144a7d026c1ca9edbd7dd98731/-/blob/vscode/src/external-services.ts?L49:23-49:48

We run this function unconditionally during extension initialisation. I see that in the TestClient we pass `codebase` field so I don't know why we didn't send it before. 

I will check the main behaviour to see what happens there, but my guess is this happens because on main, at the moment, we run this check config.codebase wasn't set yet properly by config watcher.

lib/shared/src/chat/transcript/display-text.ts
```diff
@@ -46,5 +47,12 @@ export function webviewOpenURIForContextItem(item: Pick<ContextItem, 'uri' | 'ra
             target: '_blank',
         }
     }
+
+    if (isRemoteFileURI(item.uri)) {
+        return {
+            href: `/${item.uri.authority}/${item.uri.path}}`,
```
dominiccooney, line 53: Don't these URIs have line number ranges which we bake into the query or hash?

dominiccooney, line 53: Below I get the impression that `path` starts with `/` but here you're adding a `/` before it...

vscode/src/chat/chat-view/ChatPanelsManager.ts
```diff
@@ -75,8 +75,56 @@ export class ChatPanelsManager implements vscode.Disposable {
...
+        const isConsumer = authStatus.isDotCom
+        const isCodyProUser = !authStatus.userCanUpgrade
+        const models = ModelsService.getModels(ModelUsage.Chat, isCodyProUser)
+
+        // Enterprise context is used for remote repositories context fetching
+        // in vs cody extension it should be always off if extension is connected
+        // to dot com instance, but in Cody Web it should be on by default for
+        // all instances (including dot com), Cody Web client sets cody.allow-remote-context
+        // to true to enable this directly.
+        const allowRemoteContext = vscode.workspace
```
dominiccooney, line 97: More than a configuration flag, would it be simpler and more direct to just say "is enterprise or is cody web"? My concern with flags is there are certain sets of flags which are configurations anyone uses, are somewhat tested, etc.

When the flag is there it gives the impression that you can pick any point in flag space and stuff will work. But that is hard to achieve. What happens when a user drops `cody.allow-remote-context: true` into their configuration and connects to dotcom? I don't want to debug or support that.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -930,7 +954,13 @@ export class SimpleChatPanelProvider
             const items = await getChatContextItemsForMention(
                 query,
                 cancellation.token,
-                scopedTelemetryRecorder
+                scopedTelemetryRecorder,
+                // Pass possible remote repository context in order to resolve files
```
dominiccooney, line 958: Could we instead pass an object into SimpleChatPanelProvider that does the right thing, so SimpleChatPanelProvider doesn't need to care about this new concept? It just calls that object when it needs ... whatever this is. So we have one conditional to create that object, instead of passing nullable data and having conditionals on the creation side and the consumption side.

Looking at the symbol case specifically, you could have done this differently by, for example, creating an interface for symbol lookup, and instantiated one implementation with vscode.executeWorkspaceSymbolProvider and another with your remote query. This avoids the conditionals, passing the nullable data which we'll have to elaborate for different providers later, avoids the early return, etc.

dominiccooney (changes requested):
Some feedback inline. I've reviewed a chunk of this but will stop at this point so we can discuss.

I'd like to see more effort to generalize and simplify the design, instead of configuration variables here and if statements there to pick a path through the existing extension code.

vovakulikov:
I'm going to close this one since changes from these PR will be delivered separately

PR #4606, opened 2024-06-18
Adding completions to google vertex provider only for anthropic based models

arafatkatze:
[Linear Issue](https://linear.app/sourcegraph/project/claude-3-on-gcp-8c014e1a3506/overview)

This PR adds support for completions support for Google Vertex but only for Anthropic based models(NOTE: Gemini and other models don't support completions when used with the google provider). 

Sense of Urgency: This is a P0 because of enterprise requirements so I would appreciate a fast approval and merging. 

## Test plan
- Run this branch for sourcegraph instance  -> https://github.com/sourcegraph/sourcegraph/pull/63282 
- Ask @arafatkatze to dm you the siteadmin config to make things work 
- Check the logs and play with completions

### DIFF
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -142,6 +142,16 @@ export async function createProviderConfig(
                             ? authStatus.configOverwrites.completionModel
                             : undefined,
                 })
+            case 'google':
+                if (authStatus.configOverwrites.completionModel?.includes('claude')) {
+                    return createAnthropicProviderConfig({
+                        client,
+                        model,
+                    })
+                } else {
+                    logError('createProviderConfig', `Unrecognized provider '${provider}' configured.`)
+                    return null
+                }
             default:
                 logError('createProviderConfig', `Unrecognized provider '${provider}' configured.`)
                 return null
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -148,10 +148,9 @@ export async function createProviderConfig(
                         client,
                         model,
                     })
-                } else {
-                    logError('createProviderConfig', `Unrecognized provider '${provider}' configured.`)
-                    return null
                 }
+                logError('createProviderConfig', `Unrecognized provider '${provider}' configured.`)
+                return null
             default:
                 logError('createProviderConfig', `Unrecognized provider '${provider}' configured.`)
                 return null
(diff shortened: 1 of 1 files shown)


vscode/src/completions/providers/create-provider.ts
```diff
@@ -142,6 +142,16 @@ export async function createProviderConfig(
                             ? authStatus.configOverwrites.completionModel
                             : undefined,
                 })
+            case 'google':
+                if (authStatus.configOverwrites.completionModel?.includes('claude')) {
```
arafatkatze, line 146: only works for [claude models](console.cloud.google.com/vertex-ai/publishers/anthropic/model-garden/claude-3-haiku?project=sourcegraph-dev) using the [messages api ](https://docs.anthropic.com/en/api/messages)

vscode/src/completions/providers/create-provider.ts
```diff
@@ -142,6 +142,15 @@ export async function createProviderConfig(
...
+                    return createAnthropicProviderConfig({
```
slimsag, line 147: My initial impression is that we would need a new Google provider here, and that simply reusing the Anthropic one like this might be risky/weird - but I'm not sure. What thoughts have you given to this tradeoff?

arafatkatze: So basically I do think that you are right in the sense that we should have a different provider for Google but the logic and code and the messages api for anthropic work very very similarly  in Bedrock+ Anthropic and vertex. Its not EXACTLY the same but its very similar. IF you go above in the same code you would see that bedrock also adopts the same thing. 

NOTE: the google provider works by making a different API for every provider as we discussed before and it doesn't have the concept of consistency so we would down the line have different APIs anyway for different providers/models and that's why I added an if condition which basically says that use anthropic only if you have claude model and return null for other things(which was the previous behaviour.)

arafatkatze: image
See this

abeatrix: Hmm am i overlooked something?

`authStatus.configOverwrites.provider === 'sourcegraph'` should never be true inside the cases where `authStatus.configOverwrites.provider === 'anthropic'` or `authStatus.configOverwrites.provider === 'aws-bedrock'` right?

arafatkatze: We could have the cody gateway as the provider and then you overwrite the completion model for that case for the other cases where you define the completion model in the siteadmin config we leave model as an undefined value so that it can be "pulled" from the siteadmin config. 

I just made that change.

slimsag (approved):

PR #4608, opened 2024-06-19
feat(perf): adds `CachedRetriever`

jamesmcnamara:
Adds a base class for retrievers that handles caching and cache invalidation.

The target for this is Jaccard similarity, however that retriever has a lot of dependencies on files. So instead of trying to collect all of the external deps, I added some mock VSCode APIs, so that when accessing files or tabs, we automatically start tracking those as dependencies for the current retrieval and subscribe to changes in those. Currently we handle that pretty crudely, but it could be expanded to handle more fine grained dependencies.

EDIT: It's a bit hard to read the diff for `JaccardSimilarityRetriever` because I moved the `getRelevantFiles` function in as a method so it could access the tracking methods. The only difference is `s/vscode.workspace/this.workspace/` and similarly for `window`

TODO:

- [x] Unit Tests
- [x] Fix a bug where closing tabs is not impacting cache lines for the `tabGroups`

Closes https://linear.app/sourcegraph/issue/CODY-2228/[autocomplete-latency]-add-additional-caching-to-the-jaccard

## Test plan
Added unit tests and manually tested.


valerybugakov:
Linked the Linear issue to the description.

vscode/src/completions/context/retrievers/cached-retriever.ts
```diff
@@ -0,0 +1,259 @@
...
+    protected abstract doRetrieval(
+        options: ContextRetrieverOptions
+    ): Promise<AutocompleteContextSnippet[]>
+
+    abstract isSupportedForLanguageId(languageId: string): boolean
+
+    /**
+     * Used to add a link between a representation of a dependency and a cache key
+     * this cache will automatically listen for changes in those dependencies and
+     * invalidate the cache key when they change
```
valerybugakov, line 87: Beautiful!

vscode/src/completions/context/retrievers/cached-retriever.ts
```diff
@@ -0,0 +1,259 @@
...
+     * When the cursor is moving into a new line, we want to fetch the context for the new line.
+     */
+    private onDidChangeTextEditorSelection = (event: vscode.TextEditorSelectionChangeEvent) => {
+        if (!this.isSupportedForLanguageId(event.textEditor.document.languageId)) {
+            return
+        }
+        const document = event.textEditor.document
+        const position = event.selections[0].active
+
+        // Start a preloading requests as identifier by setting the maxChars to 0
```
valerybugakov, line 197: This "rule" with using `maxChars: 0` for preloading is not transparent. Feel free to change the API if you have a better idea.

vscode/src/completions/context/retrievers/cached-retriever.ts
```diff
@@ -0,0 +1,259 @@
...
+        void this.retrieve({
+            document,
+            position,
+            hints: { maxChars: 0, maxMs: 150 },
+            docContext: getCurrentDocContext({
```
valerybugakov, line 202: It would be great to understand how CPU-intensive this call is. Hopefully, it's fine because it's already behind the debounce and contains relatively cheap string operations, but I'm worried about big documents. 

We ran into some CPU overloads with the LSP context retriever, which is why it is highly rate-limited/cached.

vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.test.ts
```diff
@@ -144,7 +144,7 @@ describe('JaccardSimilarityRetriever', () => {
 
     it('should include matches from the same file that do not overlap the prefix/suffix', async () => {
         // We limit the window size to 3 lines
-        const retriever = new JaccardSimilarityRetriever(3)
+        const retriever = new JaccardSimilarityRetriever({ snippetWindowSize: 3 })
```
valerybugakov, line 147: 🙏

vscode/src/completions/context/retrievers/cached-retriever.ts
```diff
@@ -0,0 +1,259 @@
...
+            return
+        }
+        // Unless they have specified no debounce, default to debouncing all handlers
+        handler = debounceMs === 0 ? handler : debounce(handler, debounceMs ?? 100)
+        this.subscriptions.set(handler.name, register(handler))
+    }
+
+    // Fake namespaces to make replaces VSCode APIs more readable
+    protected window = this
+    protected workspace = this
```
valerybugakov, line 242: This might be confusing to everyday VS Code codebase readers. Could you elaborate on why this is helpful?

valerybugakov (commented):
This is great! I haven't tested it locally yet. Will do it when A/B tests come online.

vscode/src/completions/context/retrievers/cached-retriever.ts
valerybugakov: With fresh eyes, I see how this is used. It wraps the default `vscode` APIs with additional functionality specific to `CachedRetriever`. Thoughts on using these methods directly without the intermediate namespace? Then the consumer won't need to check if these namespaces have any additional magic.

valerybugakov: With fresh eyes, I see how this is used. It wraps the default vscode APIs with additional functionality specific to `CachedRetriever`. Thoughts on using these methods directly without the intermediate namespace? Then the consumer won't need to check if these namespaces have any additional magic.

vscode/src/completions/context/retrievers/cached-retriever.ts
jamesmcnamara: I updated `hints` to be optional and added an `isPreload` flag to make it more obvious, but I'm not married to that if there's some downsides I'm missing.

vscode/src/completions/context/retrievers/cached-retriever.ts
jamesmcnamara: I tried adding `wrapInActiveSpan` to this, would that be a good way to measure it?

vscode/src/completions/context/retrievers/cached-retriever.ts
jamesmcnamara: Yep I got rid of that. I also added local `workspace` and `window` namespaces to this class so it can be mocked.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -134,7 +144,9 @@ export class InlineCompletionItemProvider
         createBfgRetriever,
         ...config
     }: CodyCompletionItemProviderConfig) {
-        this.config = {
+        // This is a static field to allow for easy access in the static `configuration` getter.
+        // There must only be one instance of this class at a time.
+        InlineCompletionItemProvider._configuration = {
```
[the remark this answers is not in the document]
jamesmcnamara: I actually had to pull this out into a separate file because it was causing a circular dependency in tests.

vscode/src/completions/context/retrievers/cached-retriever.ts
valerybugakov: Sadly, `wrapInActiveSpan` does not collect CPU load information, and because we sample only a small percentage of traces from the clients, it's not suitable for analyzing the number of times this is called. 

If you think this won't be an issue, we can keep it as is for now.

vscode/src/completions/context/retrievers/cached-retriever.test.ts
```diff
@@ -0,0 +1,171 @@
+import { type AutocompleteContextSnippet, testFileUri } from '@sourcegraph/cody-shared'
+import dedent from 'dedent'
+import { describe, expect, it } from 'vitest'
+import * as vscode from 'vscode'
+import { getCurrentDocContext } from '../../get-current-doc-context'
+import { document, documentAndPosition } from '../../test-helpers'
+import type { ContextRetrieverOptions } from '../../types'
+import { type CachedRerieverOptions, CachedRetriever } from './cached-retriever'
+
+class MockWorkspace implements Partial<typeof vscode.workspace> {
```
valerybugakov, line 10: We can also mock it using `vitest` to avoid adding a test-only constructor argument. Example here: https://github.com/sourcegraph/cody/blob/014d0026b543b7cbcd4d8e25ac517eed998403c1/vscode/src/completions/inline-completion-item-provider.test.ts#L28-L37

vscode/src/completions/context/retrievers/cached-retriever.ts
```diff
@@ -0,0 +1,274 @@
...
+export interface CacheableRetriever extends ContextRetriever {}
+
+export interface CachedRerieverOptions {
+    cacheOptions?: LRUCache.Options<string, AutocompleteContextSnippet[], unknown>
+    dependencyCacheOptions?: LRUCache.Options<string, Set<string>, unknown>
+    precomputeOnCursorMove?: {
+        debounceMs?: number
+    }
+    // Flag for testing to ensure synchronous execution
+    neverDebounce?: boolean
```
valerybugakov, line 18: Can we use `vi.useFakeTimers()` to advance time in tests and cover the debounce behavior there too? It should be possible in combination with `vi.advanceTimersByTime(n)`.

vscode/src/completions/inline-completion-item-provider-config.ts
```diff
@@ -0,0 +1,55 @@
...
+    'createBfgRetriever'
+> &
+    Required<Pick<CodyCompletionItemProviderConfig, 'isDotComUser'>>
+
+/**
+ * A singleton that manages the configuration for the inline completion item provider.
+ * The configuration is set using the `set` method, and can be accessed using the `configuration` getter.
+ * If the configuration has not been set, an error will be thrown when accessing the `configuration` getter.
+ */
+export const InlineCompletionItemProviderConfigSingleton = {
```
valerybugakov, line 43: We already have `vscode/src/completions/completion-provider-config.ts`, so it might be confusing to have `vscode/src/completions/inline-completion-item-provider-config.ts` nearby. Maybe rename the new file to `inline-completion-item-provider-ref.ts` or `-singleton.ts`.

vscode/src/completions/types.ts
```diff
@@ -21,9 +21,12 @@ export interface ContextRetrieverOptions {
...
     docContext: DocumentContext
-    hints: {
-        maxChars: number
-        maxMs: number
+    hints?: {
+        maxChars?: number
+        maxMs?: number
+        // signals that this is being run pre-emptively to provide a warm cache
+        // and any result we return will be ingored
+        isPreload?: boolean
```
valerybugakov, line 29: Nice!

.tool-versions
```diff
@@ -1,2 +1,2 @@
-nodejs 20.4.0
+nodejs 21.1.0
```
valerybugakov, line 1: If we want to upgrade, can we do it in a separate PR? We use `"@types/node": "20.4.0",` in one of the packages that should be updated alongside this one.

valerybugakov (approved):
That's awesome! Thank you, @jamesmcnamara 🙌

Left a couple of minor comments. Also, let's add a changelog entry to `vscode/CHANGELOG.md`.

vscode/src/completions/context/retrievers/cached-retriever.test.ts
jamesmcnamara: The issue is that I'm not just mocking it out, I want real behavior with state, specifically intercepting the listeners called and saving them somewhere. If I used a mock I would need to have global state which is pretty icky. I was modeling it after a few other spots where the vscode API was [mocked](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/CharactersLogger.ts?L15) [this](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/non-stop/rejection-tracker.ts?L31) [way](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/common/persistence-tracker/index.ts?L39-43).

I'm also hoping this might be something I could expand or merge because the existing vscodeMocks are good for replacing vscode with no-ops but not for simulating behavior.

.tool-versions
jamesmcnamara: I only added this because running `pnpm generate-agent-kotlin-bindings` fails with:
```
+ yarn install
yarn install v1.22.22
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
error @eslint/eslintrc@3.1.0: The engine "node" is incompatible with this module. Expected version "^18.18.0 || ^20.9.0 || >=21.1.0". Got "20.4.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
 ELIFECYCLE  Command failed with exit code 1.
```

I'll revert it but I think someone else will probably run into it.

For that matter, it's kind of strange that this PR caused any changes in the Kotlin code?

vscode/src/completions/context/retrievers/cached-retriever.test.ts
valerybugakov: Got it, sounds good.

.tool-versions
valerybugakov: Yes! I bumped into the same issue recently.

valerybugakov: It looks like our CI check for Kotlin bindings does not work correctly. Some PRs can go through without updating them.

PR #4610, opened 2024-06-19
Chat: Anthropic chat client for cody.dev.models

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2431

This PR adds a custom chat client for anthropic when an anthropic model is provided through the `cody.dev.models` config.

This allows us to test new models that are not yet available through Cody Gateway in the client locally.

## Test plan

2. In your VS CODE settings file, configure `cody.dev.models` to add a new claude model. For example:
```json
{
	"cody.dev.models": [
		{
			"provider": "anthropic",
			"model": "claude-3-new-model-name",
			"apiKey": "$GET_ACCESS_TOKEN_FROM_1PASSWORD"
		},
	]
}
```
3. Reload VS Code
4. Verify you can choose the new model from the chat model dropdown list
5. Verify you are getting response back when using the new model

Screenshot 2024-06-18 at 9 03 44 PM


lib/shared/src/llm-providers/anthropic/chat-client.ts
```diff
@@ -0,0 +1,88 @@
...
+} from '../..'
+
+export async function anthropicChatClient(
+    params: CompletionParameters,
+    cb: CompletionCallbacks,
+    // This is used for logging as the completions request is sent to the provider's API
+    completionsEndpoint: string,
+    logger?: CompletionLogger,
+    signal?: AbortSignal
+): Promise<void> {
```
valerybugakov, line 20: nit: these types are reused across all chat network clients. It would be great to have it defined in one place and shared between the clients: 

```ts
// in lib/shared/src/llm-providers/clients.ts
export type ChatNetworkCLient = (
    params: CompletionParameters,
    cb: CompletionCallbacks,
    completionsEndpoint: string,
    logger?: CompletionLogger,
    signal?: AbortSignal
) => Promise<void>

const clientMap: Record<string, ChatNetworkCLient> = {
    ollama: ollamaChatClient,
    google: googleChatClient,
    groq: groqChatClient,
    openaicompatible: groqChatClient,
}

// -------------------------------
// in lib/shared/src/llm-providers/anthropic/chat-client.ts
import type { ChatNetworkCLient } from '../clients'

export const anthropicChatClient: ChatNetworkCLient = async (
    params,
    cb,
    // This is used for logging as the completions request is sent to the provider's API
    completionsEndpoint,
    logger,
    signal
) => {...}
```

lib/shared/src/llm-providers/anthropic/chat-client.ts
```diff
@@ -0,0 +1,88 @@
...
+                messages,
+                max_tokens: params.maxTokensToSample,
+                temperature: params.temperature,
+                top_k: params.topK,
+                top_p: params.topP,
+                stop_sequences: params.stopSequences,
+                stream: true,
+                system: `${systemPrompt}`,
+            })
+            .then(async stream => {
```
valerybugakov, line 61: nit: thoughts on using `await` for it too since we are already in the async function?

valerybugakov (approved):
LGTM!

PR #4617, opened 2024-06-19
Agent: add automatically generated bindings for Java

olafurpg:
Previously, we only generated bindings for Kotlin. The Eclipse plugin is written in Java so this PR adds support to generate Java bindings as well. Java and Kotlin are very similar so required diff is minimal. The Java code is not added to this repo because we already have the Kotlin bindings here and I don't want every agent-protocol.ts PR to have a huge diff. Instead, we can just commit the Java bindings directly to the Eclipse repo.

Note: the code generator implementation is already quite messy and it's getting more messy with this PR. I am willing to take on this tech debt for now because it's easy to catch regressions from diffs in the generated code.

## Test plan

Minimal and low-risk diffs in Kotlin bindings (adding semi-colons and using Java getter method names instead of Kotlin sugar)

### DIFF
--- a/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/AttributionParams.kt
+++ b/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/AttributionParams.kt
@@ -1,5 +1,5 @@
 @file:Suppress("FunctionName", "ClassName", "unused", "EnumEntryName", "UnusedImport")
-package com.sourcegraph.cody.protocol_generated
+package com.sourcegraph.cody.protocol_generated;
 
 data class AttributionParams(
   val repositoryNames: List<String>,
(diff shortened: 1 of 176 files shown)


olafurpg:
The codegen tool is only used by Eclipse at the moment (JB hasn't even adopted it yet) so this is a low-risk PR to merge

valerybugakov (approved):

mkondratek (approved):

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/protocol_generated/AttributionParams.kt
odisseus, line 1: Why are these semicolons added?

olafurpg: so we can reuse the codegen across java/kotlin

olafurpg: the PR has a lot of `if kotlin do X else Y`, and I didn't separate the kotlin/java cases when it added semicolons because kotlin understands the semicolons

PR #4623, opened 2024-06-19
Rewrite queries based on chat history and mentioned context items for better enhanced context fetching

thenamankumar:
closes: https://linear.app/sourcegraph/issue/CODY-2272/rewrite-query-including-content-of-mention-before-sending-to-context

This PR adds an experimental feature under the `experimentalNoodle` flag to rewrite the queries based on the chat history (last 3 interactions) and the mentioned context items before sending it for enhanced context fetching; assuming it would improve the performance of the context search. 

This is similar to the keyword expansion integrated with symf search right now, but it performs the rewrite based on the chat history as well as the mentioned context items, unlike the keyword expansion which treats the query in isolation. 

The formal evaluation of this technique has not been done yet but it is the next step. 

Until we have evaluated that the rewrite boosts the performance of the context retrieval, we will keep the rewrite step behind the experimental flag. 

## Test plan

A few examples are (the rewritten query manually pasted in the input box while debugging):

image

image

### DIFF
--- a/lib/shared/src/chat/chat.ts
+++ b/lib/shared/src/chat/chat.ts
@@ -16,7 +16,7 @@ const DEFAULT_CHAT_COMPLETION_PARAMETERS: Omit<ChatParameters, 'maxTokensToSampl
 
 export class ChatClient {
     constructor(
-        public completions: SourcegraphCompletionsClient,
+        private completions: SourcegraphCompletionsClient,
         private getAuthStatus: () => Pick<
             AuthStatus,
             'userCanUpgrade' | 'isDotCom' | 'endpoint' | 'codyApiVersion'
--- a/lib/shared/src/chat/chat.ts
+++ b/lib/shared/src/chat/chat.ts
@@ -16,7 +16,7 @@ const DEFAULT_CHAT_COMPLETION_PARAMETERS: Omit<ChatParameters, 'maxTokensToSampl
 
 export class ChatClient {
     constructor(
-        private completions: SourcegraphCompletionsClient,
+        public completions: SourcegraphCompletionsClient,
         private getAuthStatus: () => Pick<
             AuthStatus,
             'userCanUpgrade' | 'isDotCom' | 'endpoint' | 'codyApiVersion'
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -725,9 +725,8 @@ export class SimpleChatPanelProvider
                                   ? await rewriteChatQuery({
                                         query: inputText,
                                         contextItems: userContextItems,
-                                        chatMessages: this.chatModel.getMessages(),
-                                        completionsClient: this.chatClient.completions,
-                                        modelID: this.chatModel.modelID,
+                                        chatClient: this.chatClient,
+                                        chatModel: this.chatModel,
                                     })
                                   : inputText
 
--- a/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
+++ b/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
@@ -79,6 +79,7 @@ import type { EnterpriseContextFactory } from '../../context/enterprise-context-
 import type { Repo } from '../../context/repo-fetcher'
 import type { RemoteRepoPicker } from '../../context/repo-picker'
 import type { ContextRankingController } from '../../local-context/context-ranking'
+import { rewriteChatQuery } from '../../local-context/rewrite-chat-query'
 import { chatModel } from '../../models'
 import { migrateAndNotifyForOutdatedModels } from '../../models/modelMigrator'
 import { gitCommitIdFromGitExtension } from '../../repository/git-extension-api'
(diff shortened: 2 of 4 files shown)


lib/shared/src/chat/chat.ts
```diff
@@ -16,7 +16,7 @@ const DEFAULT_CHAT_COMPLETION_PARAMETERS: Omit<ChatParameters, 'maxTokensToSampl
 
 export class ChatClient {
     constructor(
-        private completions: SourcegraphCompletionsClient,
+        public completions: SourcegraphCompletionsClient,
```
keegancsmith, line 19: can you explain why you are exposing this and using it directly. Looking at the code I believe it is expected for users to directly use the chat() function. Is the problem here to do with the filtering this does on the transcript it sends up? This function does quite a bit more, and I am unsure of the tradeoff of exposing this field for others vs maybe adding a parameter to chat which helps justify what you want to do.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -713,11 +714,27 @@ export class SimpleChatPanelProvider
...
                     userContextItems,
                     addEnhancedContext || hasCorpusMentions
-                        ? async text =>
-                              getEnhancedContext({
+                        ? async () => {
+                              /* EXPERIMENTAL: Rewrite query based on the chat history and the
+                               * mentioned context items for better enhanced context retrieval.
+                               *
+                               * The retrieval performance boost is not evaluated yet and thus
+                               * it is only available when `experimentNoodle` is set to `true`.
```
keegancsmith, line 722: how will we evaluate this? I guess at first we will just dogfood and see if it helps, but then if we gain confidence from that we will try measure somehow?

vscode/src/editor/utils/editor-context.ts
```diff
@@ -358,6 +358,7 @@ async function resolveContextMentionProviderContextItem(
                       providerUri: item.providerUri,
                       content: item.ai.content,
                       provider: 'openctx',
+                      source: ContextItemSource.User,
```
keegancsmith, line 361: unrelated to this PR, but this field should really be made required. Is User the correct way for us to capture this? I know this is used in telemetry (and in rendering) so I wonder if we should be more specific? I supposed provider is also captured intelemetry so this is fine.

vscode/src/local-context/rewrite-chat-query.ts
```diff
@@ -0,0 +1,123 @@
...
+    getSimplePreamble,
+    psDedent,
+} from '@sourcegraph/cody-shared'
+import { logDebug } from '../log'
+import { PromptBuilder } from '../prompt-builder'
+
+/**
+ * Rewrite the query based on the conversation history and mentioned context items.
+ */
+export async function rewriteChatQuery({
```
keegancsmith, line 18: I would expect a debug log at the end of this function saying the prompt before and after. This will help with us dogfooding to see the difference.

keegancsmith (approved):
LGTM. Nice that you managed to contain all the new logic in one file!

vscode/src/local-context/rewrite-chat-query.ts
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+        }
+
+        const text = streamingText.at(-1) ?? ''
+        const parser = new XMLParser()
+        const document = parser.parse(text)
+
+        return document?.rewritten_query?.trim()
+            ? PromptString.unsafe_fromLLMResponse(document.rewritten_query.trim())
```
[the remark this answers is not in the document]
thenamankumar: I checked that all the bot messages are converted to prompt string using `unsafe_fromLLMResponse` only in `SimpleChatPanelProvider`. And given we are only using the response to do context fetching, it seems safe enough.

vscode/src/local-context/rewrite-chat-query.ts
```diff
@@ -0,0 +1,123 @@
...
+            : query
+    } catch (err) {
+        logDebug('rewrite-chat-query', 'failed', { verbose: err })
```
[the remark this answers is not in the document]
thenamankumar: makes sense. done.

vscode/src/editor/utils/editor-context.ts
thenamankumar: Actually this is not required right now. Removing this.

vscode/src/local-context/rewrite-chat-query.ts
thenamankumar: Nice idea. Added.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
thenamankumar: It will be difficult to make sense of anything by dogfooding, we will need to numbers to gain confidence. 

I am working with the AI team to get a multi-turn evaluation dataset to evaluate if it boosts the performance of symf, local embeddings & remote search.

thenamankumar: We might also want to restrict rewriting only certain queries. Need to figure that out.

PR #4625, opened 2024-06-19
bump time to first completion up to 2.5 seconds to allow for more syntactically complete code completions

beyang:
It's a better user experience to wait a bit longer to generate a syntactically whole completion rather than a completion that does not complete the syntactic block. The issue with the latter is both the sense of "incompleteness" but also, we get "stuck" in a more deeply nested block.

| Column 1 | Column 2 |
| --------- | --------- |
|       image  |   image |

## Test plan

Tested locally

### DIFF
--- a/vscode/src/configuration.ts
+++ b/vscode/src/configuration.ts
@@ -186,7 +186,7 @@ export function getConfiguration(
         },
         autocompleteFirstCompletionTimeout: getHiddenSetting<number>(
             'autocomplete.advanced.timeout.firstCompletion',
-            1_500
+            3_500
         ),
         testingModelConfig:
             isTesting && hasValidLocalEmbeddingsConfig()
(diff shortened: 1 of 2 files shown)


vscode/src/configuration.ts
```diff
@@ -186,7 +186,7 @@ export function getConfiguration(
         },
         autocompleteFirstCompletionTimeout: getHiddenSetting<number>(
             'autocomplete.advanced.timeout.firstCompletion',
-            1_500
+            3_500
```
valerybugakov, line 189: Do we want to bump to 2.5s or 3.5s?

valerybugakov (approved):
I like the longer delay and UX it gives us! 

However, we don't have any data that the majority of Cody users likes/expects this. We should run an A/B test to better understand the trade-offs. We will have the capacity for that after finishing our work on latency reduction.

PR #4627, opened 2024-06-20
debounce selection determination to reduce CPU load

sqs:
Previously, it only debounced the postMessage. Now it debounces the entire operation to get the right selection context. I still see it called 2x for a cursor move (empty selection) and 4x for a non-empty selection due to VS Code calling the event handler multiple times for some reason, but this is a big improvement over before.

Fixes https://linear.app/sourcegraph/issue/CODY-2481/debounce-or-throttle-commandscontextselection-file-calls-to-prevent

## Test plan

Try selections and see how many OT spans get sent (and use logpoints).

### DIFF
--- a/vscode/src/chat/clientStateBroadcaster.ts
+++ b/vscode/src/chat/clientStateBroadcaster.ts
@@ -29,9 +29,9 @@ export function startClientStateBroadcaster({
     postMessage: PostMessage
     chatModel: SimpleChatModel
 }): vscode.Disposable {
-    const postMessage = debouncedIdempotentPostMessage(rawPostMessage)
+    const postMessage = idempotentPostMessage(rawPostMessage)
 
-    async function sendClientState(): Promise<void> {
+    async function rawSendClientState(signal: AbortSignal | null): Promise<void> {
         const items: ContextItem[] = []
 
         // TODO(sqs): Make this consistent between self-serve (no remote search) and enterprise (has
@@ -87,6 +87,7 @@ export function startClientStateBroadcaster({
         const userContextSize = context?.user ?? input
 
         const [contextFile] = await getSelectionOrFileContext()
+        signal?.throwIfAborted()
         if (contextFile) {
             const range =
                 contextFile.range && isMultiLineRange(contextFile.range) ? contextFile.range : undefined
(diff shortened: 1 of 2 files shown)


vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -111,31 +112,37 @@ export function startClientStateBroadcaster({
 
     const disposables: vscode.Disposable[] = []
 
+    const sendClientState = debounced(rawSendClientState)
```
valerybugakov, line 115: Do we need a custom `debounce` implementation? We use the one from `lodash` in a couple of places already.

```suggestion
import { debounce } from 'lodash'

    const sendClientState = debounce(rawSendClientState, DEBOUNCE_DELAY)
```

vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -111,31 +112,37 @@ export function startClientStateBroadcaster({
...
     disposables.push(
         vscode.window.onDidChangeActiveTextEditor(() => {
-            void sendClientState()
+            // Relatively infrequent action, so don't debounce and show immediately in the UI.
+            void sendClientState('immediate')
```
valerybugakov, line 119: Do we need to use the debounced function here instead of `rawSendClientState`? If not we can create a denounced copy with the zero timeout value and a comment about why the timeout is required.

valerybugakov (approved):
Thanks! It fixes the issue. I added a couple of questions inline.

vscode/src/chat/clientStateBroadcaster.ts
sqs: Yeah, so that there is not a race condition where a slow earlier call to rawSendClientState could call postMessage with stale data. Will add comment.

vscode/src/chat/clientStateBroadcaster.ts
sqs: Yeah, because it's async and we need to pass in an AbortController. I will add a comment justifying this.

valerybugakov: I see. Because the debounced function body needs to know when it's canceled too 👍

PR #4637, opened 2024-06-21
Chat: Refactor update notices to be IDE-specific

abeatrix:
Part of https://linear.app/sourcegraph/issue/CODY-2291/port-chat-ux-vscode-isms-to-jetbrains

Make IDE-specific release notes URLs and version info to update notices

- Update the codebase to send the existing config values from `config.agentExtensionVersion` and `config.agentIDE` when `config.isRunningInsideAgent` is true. 
- When the IDE and version number is provided, we will generate the release note URL based on the IDE in releaseNotesURL function
- Update Storybook to show VersionUpdatedNotice components for Jetbrains Stable and Nightly release.

NOTE: This assumes each agent IDE are sending the correct IDE name and version number to the webview for the update notice to be displayed. When those are not provided, the updated notice will not be rendered.

BREAKING CHANGE: The `VersionUpdatedNotice` component now requires `IDE` and `version` props.

## Test plan

- Added tests should all be passing
- Green CI to confirm this change doesn't break the existing E2E test for VS Code update notice and should work as intended when the agent IDE provides the correct IDE name and version number.

You can start the story book with `pnpm run storybook` to verify the components are working as expected:

image


mkondratek (approved):
lgtm

PR #4642, opened 2024-06-21
Use shared telemetry recorder in agent

pkukielka:
## Changes 

We now have shared telemetry recorder which is properly configured when Cody is in agent mode.
We should not create individual, separate one for direct agent calls.

## Test plan

1. Run Jetbrains clinet using current PR for Cody and [this PR ](https://github.com/sourcegraph/jetbrains/pull/1795) for Jetbrains
2. Ask cody to generate some code, and then use button to copy it. Paste it somewhere in the code
3. Using looker check if new events with proper names (`copyButton.clicked:clicked`, `cody.keyDown:paste`).


agent/src/agent.ts
```diff
@@ -320,25 +318,6 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
-    /**
-     * agentTelemetryRecorderProvider must be used for all events recording
-     * directly within the agent (i.e. code in agent/src/...) and via the agent's
-     * 'telemetry/recordEvent' RPC.
-     *
-     * Components that use VSCode implementations directly (i.e. code in
-     * vscode/src/...) will continue to use the shared recorder initialized and
-     * configured as part of VSCode initialization in vscode/src/services/telemetry-v2.ts.
-     */
-    private agentTelemetryRecorderProvider: AgentHandlerTelemetryRecorderProvider =
```
pkukielka, line 332: @bobheadxi I think you created it so adding you to review that change.
I think this code is not needed anymore.
We recreate telemetry recorder on a configuration change now, and if configuration comes from agent we setup client name and version properly:

https://github.com/sourcegraph/cody/blob/main/lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts#L53

I verified in debugger that it works as expected.

bobheadxi: I must admit I'm not 100% sure what I was doing here - I trust your judgement, thanks for updating and maintaining this ❤️

PR #4643, opened 2024-06-21
Provide a template for local e2e tests for local development purposes. 

jhchabran:
See the README.md for more details about why this exists. See [DINF-37](https://linear.app/sourcegraph/issue/DINF-37/poc-running-a-local-cody-e2e-test-against-a-local-sourcegraph-instance). 

Fixes CODY-2029

We added [a new `sg db` subcommand to provide a default site-admin](https://github.com/sourcegraph/sourcegraph/pull/63320) so we go around dealing with obtaining an access token. Instead, we can assume that everyone has that _default_ one in their local setup. 

~Need sister PR (https://github.com/sourcegraph/sourcegraph/pull/63548) to land, as it depends on a new commandset (i.e. `sg start cody-e2e-tests`)~ merged

💭 For the CodyGW settings, @vdavid could you take a look? All of this works on my side, but I might be missing some side-effects with regards to the configuration, and you might also have suggestions to make on the "advice" side in the README.md 

⚠️ Reviewers: I'm not very familiar with TS, if you see something really not idiomatic, it might be faster to push a commit than spending time telling me what's wrong in the comments (though I'm okay otherwise). 

Example of the output this provides, when for example it detects incorrect settings in the site-config of the locally running Sourcegraph instance. 

image

---

TODO: 

- [x] Ensure those are NOT run by the CI. 
  - https://github.com/sourcegraph/cody/actions/runs/9615237566/job/26521927197?pr=4643#step:9:909 
- [x] Added a new commandset `sg start dotcom-cody-e2e` (see https://github.com/sourcegraph/sourcegraph/pull/63548) 
  - Waiting/Blocked on: https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1719303757331739
- [x] Get this to run on ~@olafurpg's laptop~ did it with @valerybugakov. 

Thanks to @Strum355 for pairing with me to debug various nasty local env issues. 

## Test plan

Tested locally in various scenarios.


agent/src/local-e2e/template.test.ts
```diff
@@ -0,0 +1,53 @@
+// The goal of this file is to document the steps to run Cody with all services locally.
+import path from 'node:path'
+import { afterAll, beforeAll, describe, expect, it } from 'vitest'
+
+import { ModelsService, getDotComDefaultModels } from '@sourcegraph/cody-shared'
+import { TestClient } from '../TestClient'
+import { TestWorkspace } from '../TestWorkspace'
+import { LocalSGInstance, isLocal } from './helpers'
+
+describe.runIf(isLocal)('E2E-local', () => {
```
jhchabran, line 10: About the `runIf` I tried other approaches like excluding those, but then you can't force vitest to run those. I wished we didn't have to rely on an env var, but it does the job. At least it's wrapped in the helper.

agent/src/local-e2e/template.test.ts
```diff
@@ -0,0 +1,53 @@
...
+        // `chat/explain`. In VS Code, we try to reuse empty webview panels,
+        // which is undesireable for agent clients.
+        expect(id).not.toStrictEqual(freshChatID)
+
+        const lastMessage = await client.firstNonEmptyTranscript(id)
+        console.log(lastMessage)
+        for (const m of lastMessage.messages) {
+            console.log(m)
+        }
+    }, 200_000) // We're making full roundtrips, so we need to increase the default timeout.
```
RXminuS, line 52: Isn't this why recording can still be valuable even when running locally? Especially because I can then in my mock pass-through only a specific request I care about?

agent/src/local-e2e/template.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+describe.runIf(isLocal)('E2E-local', () => {
+    const workspace = new TestWorkspace(path.join(__dirname, '..', '__tests__', 'example-ts'))
+    let client: TestClient
+    const sg = new LocalSGInstance()
+
+    beforeAll(async () => {
+        ModelsService.setModels(getDotComDefaultModels())
+        await workspace.beforeAll()
+        await sg.beforeAll()
```
RXminuS, line 18: I know we do this in a bunch of places already but I'm not actually a fan of the beforeAll() because it depends on the test. For instance if I want to test that I can sign-in, and then sign-in to another instance and not have state cross over...now i need to call beforeAll() in beforeEach() or even just within the test.

I don't necessarily have a better suggestion, but I think this gets confusing.

RXminuS (approved):

agent/src/local-e2e/template.test.ts
jhchabran: If you're running things locally with those tests, by definition you want to see the whole thing. This is more a tool to help your reproduce bugs than one to actually introduce e2e testing, for which there are other options AFAIU. The README.md covers this.

agent/src/local-e2e/template.test.ts
jhchabran: I understand your point. I just followed the current approach here, so I'll let you folks decide what to do with this once this lands.

PR #4644, opened 2024-06-21
fix loading of linux certs for agent

Strum355:
`globalAgent.options.ca` was always set to empty array because the function returns _before_ the promise from `loadLinuxCerts()` resolves

## Test plan

Tested locally with print statements, it was not working for me otherwise

### DIFF
--- a/agent/src/certs.ts
+++ b/agent/src/certs.ts
@@ -1,4 +1,4 @@
-import fspromises from 'node:fs/promises'
+import { readFileSync } from 'node:fs'
 import { globalAgent } from 'node:https'
 
 /**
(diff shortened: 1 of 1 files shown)


olafurpg (approved):
Nice catch!

agent/src/certs.ts
```diff
@@ -28,7 +28,7 @@ export function registerLocalCertificates() {
     }
 }
 
-const linuxPossibleCertPaths = ['/etc/ssl/certs/ca-certificates.crt', '/etc/ssl/certs/ca-bundle.crt']
+const linuxPossibleCertPaths = ['/etc/ssl/certs/ca-certificates.crt', '/etc/ssl/certs/ca-bundle.crt', '/home/noah/.local/share/caddy/certificates/local/sourcegraph.test/sourcegraph.test.crt']
```
olafurpg, line 31: `/home/noah`

Strum355: I couldve sworn I got rid of this before committing :miffed:

Strum355:
```suggestion
const linuxPossibleCertPaths = ['/etc/ssl/certs/ca-certificates.crt', '/etc/ssl/certs/ca-bundle.crt']
```

olafurpg (commented):
Hit on this right now testing the new `auth login` command in the cli. Merging this branch seems to fix the problem!

PR #4601, opened 2024-06-18
Chat: fix abort message loading state

abeatrix:
Part of https://linear.app/sourcegraph/issue/CODY-2149/feedback-bring-the-stop-button-back-again#comment-4c363fea

- Notify the webview when aborting a message submission operation to indicate there is no message in progress
- Update the `priorAssistantMessageIsLoading` prop to check if there is a message in progress, not just if the previous message was loading

## Test plan

- Ask Cody a question
- Hit ESC to abort the request during the context fetching step
- Verify the loading state disappeared on abort:

https://github.com/sourcegraph/cody/assets/68532117/a553602a-3772-484d-b1ea-3c559eef430a

### Before

You will be stuck in this loading state even after you hit ESC to abort the request:

image

PR #4648, opened 2024-06-21
Chat: handle aborted responses in assistant message cell UI

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2464/improve-aborted-chat-message-ui

This PR addresses @toolmantim 's UI feedback in https://github.com/sourcegraph/cody/pull/4601#pullrequestreview-2126995194:
- [x] In the following state, before we successfully fetched context, could we just remove the extra human message and return them to the original one? Otherwise it feels like the chat is left in a broken-ish state.
- [x] Once we have an assistant message that has been aborted, can we not show the rating thumbs up/down
- [x] can we add a paragraph in to say that the stream was stopped?
- [x] If we could add this chat state to the Storybook scenarios too, that would be fantastic thanks!

Summary:

- Add `isAbortErrorOrSocketHangUp` function to shared error utilities
- Update `AssistantMessageCell` to handle aborted responses by displaying a message indicating the output stream was stopped
- Ensure context focus actions are still shown for aborted responses
- Update storybook with the aborted messages

## Test plan

Updated Storybook with the expected behavior for aborted messages with and without assistant response:

image


abeatrix:
@toolmantim i've updated the PR to reflect the changes we discussed IRL 😎 

https://github.com/sourcegraph/cody/assets/68532117/7dcaa5de-239b-40cc-a5ca-94335cd2c847

toolmantim (approved):

PR #4649, opened 2024-06-21
fixing recordsPrivateMetadata field when the private context is empty

hitesh-1997:
## Context
This PR make the following changes:
1. Fix the `recordsPrivateMetadataTranscript` in case no private metadata is send to the telemetry. Earlier we were always sending `1` in this fields even if there was no private metadata. More context behind the fix [on the slack thread](https://sourcegraph.slack.com/archives/CN4FC7XT4/p1718914278731939?thread_ts=1718821081.192759&cid=CN4FC7XT4).
2. Adds a `filePath` variable to the logging for only the `dotComUsers` and `publicGitHubRepos`, which indicates for which file the completion was triggered.
3. Adds a `gitUrl` variable for the `dotComUsers` repos. More context on [the slack thread.](https://sourcegraph.slack.com/archives/CN4FC7XT4/p1719001526612899?thread_ts=1716253514.127349&cid=CN4FC7XT4)

## Test plan
**Manual testing**
1. Start the debugger from this branch. For any new autocomplete request, the telemetry response for the "cody.completion/suggested" will have`recordsPrivateMetadataTranscript: 1` only if the value is non-null.
2. The `inlineCompletionItemContext` field will have one additional variable `filePath` indicating the file for which the autocomplete was triggered.

### DIFF
--- a/vscode/src/completions/get-inline-completions.ts
+++ b/vscode/src/completions/get-inline-completions.ts
@@ -404,6 +404,7 @@ async function doGetInlineCompletions(
 
     const inlineContextParams = {
         context: contextResult?.context,
+        filePath: gitIdentifiersForFile?.filePath,
         gitUrl: gitIdentifiersForFile?.gitUrl,
         commit: gitIdentifiersForFile?.commit,
     }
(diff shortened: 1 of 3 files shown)


akalia25 (approved):
Thanks for making these updates! LGTM!

PR #4658, opened 2024-06-24
chat: always use range if there is an active selection

keegancsmith:
For the auto-updating "Current File" context we only made it a "Current Selection" if the selection was multi-line. I am unsure why we had this behaviour, this seems more correct.

I came across this while looking into https://linear.app/sourcegraph/issue/CODY-2276/buggy-line-range-logic-for-open-files-in-new-chat

Test Plan: loads of testing with selecting. See video.

https://github.com/sourcegraph/cody/assets/187831/01ad21d0-6e14-4dfa-a2d2-9485f2cea51b

### DIFF
--- a/lib/shared/src/common/range.test.ts
+++ b/lib/shared/src/common/range.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'vitest'
-import { displayLineRange, displayRange, isMultiLineRange, toRangeData } from './range'
+import { displayLineRange, displayRange, expandToLineRange, toRangeData } from './range'
 
 describe('toRangeData', () => {
     test('converts Range to RangeData', () => {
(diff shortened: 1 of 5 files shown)


sqs (approved):
The reasoning for treating a single-line selection as meaning "use whole file context" is that it's really unlikely you just want a single line of context. But I can see that this is confusing, so this change is fine and makes sense.

keegancsmith:
> The reasoning for treating a single-line selection as meaning "use whole file context" is that it's really unlikely you just want a single line of context. But I can see that this is confusing, so this change is fine and makes sense.

I thought that might be the case for multiline. In my case I would very rarely have code accidentally highlighted, so thought this was the better principle. But I am a sample size of one, so if you think we should keep the current behaviour I get that too.

keegancsmith:
Alright on the linear issue Tim gives this change a thumbs up. So gonna merge :)

toolmantim:
> The reasoning for treating a single-line selection as meaning "use whole file context" is that it's really unlikely you just want a single line of context. 

@sqs multiple times I've just selected a part of a line and gone to ask "is there a shorthand for this?" in chat and it didn't work… that was why I originally raised it

toolmantim:
@sqs I'm curious why you think that's unlikely in general though!

PR #4660, opened 2024-06-24
mentions: add built-in linear-issues provider using linear connect for auth

keegancsmith:
This PR copies in the current state of the OpenCtx Linear Issues provider (1st commit). We then use integration with the official Linear Connect vscode extension to make authentication smooth (2nd commit). If the extension is missing we prompt the user to install it (3rd commit).

Note: we could declare that Cody depends on the Linear Connect extension. However, given this is opt-in behaviour I prefer our "just in time" prompting of adding that extension. Note that the provider will only be shown if the noodle config setting is enabled.

Test Plan: Try with linear connect in the following states: uninstalled, disabled, enabled. The first two states will always trigger showing the extension. The enabled state will log us in to Linear. I also tested if a user did not finish up the sign in flow, then the same modal would happen again when attempting connecting to Linear.

Demo at https://www.loom.com/share/f37ea88f5d9d47e8b0940aa7b5c0d975

Fixes https://linear.app/sourcegraph/issue/CODY-2540/explore-good-linear-auth-in-openctx

### DIFF
--- a/vscode/src/context/openctx.ts
+++ b/vscode/src/context/openctx.ts
@@ -1,6 +1,7 @@
 import { type ConfigurationWithAccessToken, setOpenCtxClient } from '@sourcegraph/cody-shared'
 import * as vscode from 'vscode'
 import { logDebug, outputChannel } from '../log'
+import LinearIssuesProvider from './openctx/linear-issues'
 import RemoteFileProvider from './openctx/remoteFileSearch'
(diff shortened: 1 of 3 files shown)


sqs (approved):

keegancsmith:
After discussion we decided its better to try and make these sort of things work in openctx without needing to be in the cody repo. This PR makes it possible to implement linear connect as a normal third party provider https://github.com/sourcegraph/openctx/pull/161

PR #4666, opened 2024-06-25
Chat: adapt context preamble for Sonnet 3.5

jtibshirani:
This PR enables Claude Sonnet 3.5 in the chat quality and edit tests to check
its results. In the initial chat test results, Sonnet 3.5 refused to answer
'general codebase context' questions. So we also tweak the 'context preamble'
to fix these cases. This tweak does not regress tests on other models.

## Test plan

Adapted chat and edit tests


abeatrix (approved):
nice!

PR #4673, opened 2024-06-25
Edit: Fix JetBrains issue with range expansion

umpox:
## Description

Uses `document.lineAt` to get the end of a line, rather than `Number.MAX_SAFE_INTEGER`.

## Test plan

1. Create edits in JetBrains
2. Undo them
3. Check the undo functions correctly (only original code is returned)

### DIFF
--- a/agent/recordings/customCommandsClient_509552979/recording.har.yaml
+++ b/agent/recordings/customCommandsClient_509552979/recording.har.yaml
@@ -408,11 +408,11 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: 093550ffda9d570a9949d3ee885b3212
+    - _id: 26984de7053ea01429b394e478e25b36
       _order: 0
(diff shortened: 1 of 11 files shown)


umpox:
@PiotrKarczmarz @mkondratek This should fix the issues with inline edit + undo in JetBrains. It seems we have another issue with `Number.MAX_SAFE_INTEGER` somewhere.

I'm happy to merge this as we already have access to the `document` here, and it is technically more correct. But just wanted to make you aware that this could be a problem in future if we change it for whatever reason

mkondratek:
I tested these changes in a dev run - works 🟢 
I update cody commit and [CI passed](https://github.com/sourcegraph/jetbrains/actions/runs/9664886398/job/26660960656) 🟢

mkondratek (approved):
works - thank you for the quick fix 🚀

Issue #4195, opened 2024-05-16
`Chat token usage must be updated before Context.`


Kynlos:
### Version

Latest (stable and pre)

### Describe the bug

`Chat token usage must be updated before Context.`

Caused when pasting too much text into the chat window at the same time then submitting.

### Expected behavior

Text input in the chat window for Cody should be character limited to the amount of tokens allowed based on the chosen LLM.

### Additional context

Behaviour experienced by myself several times when trying to work with larger chunks of code or information and not paying attention to how much I pasted.

Issue also reported by `po'boy` on Discord: [https://discord.com/channels/969688426372825169/969688427232641136/1240681675684577341](https://discord.com/channels/969688426372825169/969688427232641136/1240681675684577341)

![Screenshot_2024-05-16-17-15-17-85_572064f74bd5f9fa804b05334aa4f912](https://uploads.linear.app/15b96a4d-ffc0-481a-ba9c-9d840020f7bc/ae727338-576d-4b6e-93d6-f0739c9fbcce/311631d9-dc78-4cfa-8da1-ff7b7d247742?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzE1Yjk2YTRkLWZmYzAtNDgxYS1iYTljLTlkODQwMDIwZjdiYy9hZTcyNzMzOC01NzZkLTRiNmUtOTNkNi1mMDczOWM5ZmJjY2UvMzExNjMxZDktZGM3OC00Y2ZhLThkYTEtZmY3YjdkMjQ3NzQyIiwiaWF0IjoxNzE2MzM4NTgzLCJleHAiOjMzMjg2ODk4NTgzfQ.2PcrnvKzTE9sYcm8tGlg5NKdYIkyRnrgLRNchtWFi_w)

PR #4674, opened 2024-06-25
Chat: handle transcript length exceeding context window

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-1608/chat-token-usage-must-be-updated-before-context
CLOSE https://github.com/sourcegraph/cody/issues/4195

- Modify `tryAddMessages` in `PromptBuilder` to return `undefined` when all messages are added successfully
- In `DefaultPrompter`, throw an error if no transcript was added due to the context window limit being reached
- Add logging to indicate when messages were ignored due to the context limit
- Add a check to throw an error if the token limit is exceeded and no messages were added to the transcript

This ensures the user is aware when the chat input has exceeded the allowed token limit, and the error stops the prompt building process from processing so that it would move to the step where we add context without transcript.

## Test plan

2. Pick a model with a smaller context window, e.g. GPT-3.5 Turbo
3. Copy the content from [SimpleChatPanelProvider.ts](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts?L250) which exceed the context window for the smaller models
4. Submit the question and expect the error about input exceed the context window

image


dominiccooney (approved):
Ah, that explains it. Good to fix this, on the other hand, the user is still going to experience an error... just a more accurate one. Anything we can do here?

umpox (approved):

abeatrix:
> Anything we can do here?

I was thinking we could truncate the input, but we might truncate the important part of the input without users knowing. I will create a linear issue and get the product for input on what's the best way to handle this case?

PR #4676, opened 2024-06-25
Bench: Add chat question to EvaluationDocument and strategy-chat

abeatrix:
Add chat question to EvaluationDocument and strategy-chat

## Test plan

Makes it easier to display chat question associated with the file path in leaderboard:

image

Used by https://github.com/sourcegraph/cody-leaderboard/pull/7


jtibshirani (approved):
Nice!

PR #4678, opened 2024-06-25
Bench: add LLM judge for response scoring to chat strategy

abeatrix:
- Add `llmJudgeChatTemplate` to generate a prompt for evaluating LLM responses
- Integrate `LlmJudge` into the `evaluateChatStrategy` to score each chat response
- Store the numeric score in the `EvaluationDocument` for each chat response
- Log the total score for the fixture at the end of the evaluation

There are still follow-up works that we can do, including improving the prompt used for the LLM judge, but currently, it works as intended and IMO a good starting point for us to build from.

## Test plan

Currently being used by https://github.com/sourcegraph/cody-leaderboard/pull/8

image

### DIFF
--- a/agent/src/cli/cody-bench/llm-judge-chat-template.ts
+++ b/agent/src/cli/cody-bench/llm-judge-chat-template.ts
@@ -1,0 +1,18 @@
+import { type PromptString, ps } from '@sourcegraph/cody-shared'
+
+const template = ps`Response:{{CODY_RESPONSE}}
+<|end_of_response|>
+
+Please evaluate the response and provide your feedback in the following format:
+<reasoning> Briefly explain whether the response is helpful and correct, unhelpful, or incorrect.</reasoning>
+<score> Label the response as "positive" if it's helpful or "negative" if it's not helpful or incomplete.</score>
+
+Note: Consider the response "positive" if it attempts to provide a helpful answer, even with limited context. Label it "negative" if it apologizes for not knowing the answer, lacks access to information, or is incomplete.`
+
+export interface LlmJudgeChatParams {
+    response: PromptString
+}
+
+export function llmJudgeChatTemplate(params: LlmJudgeChatParams): PromptString {
+    return template.replaceAll('{{CODY_RESPONSE}}', params.response)
+}
--- a/agent/src/cli/cody-bench/llm-judge.ts
+++ b/agent/src/cli/cody-bench/llm-judge.ts
@@ -65,10 +65,12 @@ export class LlmJudge {
 function scoreNumeric(score?: string): number | undefined {
     switch (score) {
         case 'bad':
+        case 'negative':
             return 0
         case 'acceptable':
             return 1
         case 'amazing':
+        case 'positive':
             return 2
         default:
             return undefined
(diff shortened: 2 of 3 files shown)


jtibshirani (approved):
This looks good to me! Just one question around around the definition of "helpful".

agent/src/cli/cody-bench/llm-judge-chat-template.ts
```diff
@@ -0,0 +1,18 @@
+import { type PromptString, ps } from '@sourcegraph/cody-shared'
+
+const template = ps`Response:{{CODY_RESPONSE}}
+<|end_of_response|>
```
olafurpg, line 4: Here's what Claude's prompt optimizer gave

```
You are an expert judge evaluating the quality of responses from Cody, an AI coding assistant. Your task is to assess Cody's response to a user's question and provide feedback on its helpfulness and correctness.

Here is Cody's response to the user's question:
<CODY_RESPONSE>
{{CODY_RESPONSE}}
</CODY_RESPONSE>

The user's question that Cody was responding to is:
<QUESTION>
{{QUESTION}}
</QUESTION>

Please evaluate Cody's response carefully, considering the following criteria:
1. Relevance: Does the response address the user's question?
2. Correctness: Is the information provided accurate and free from errors?
3. Completeness: Does the response fully answer the question, or is it incomplete?
4. Helpfulness: Would the response be useful to the user in solving their problem or understanding the topic better?

After your evaluation, provide your feedback in the following format:

<reasoning>
Explain your assessment of the response, addressing the criteria mentioned above. Discuss whether the response is helpful and correct, unhelpful, or incorrect. Provide specific examples from the response to support your evaluation.
</reasoning>

<score>
Based on your reasoning, label the response as either "positive" or "negative".
- Label it "positive" if it attempts to provide a helpful answer, even with limited context.
- Label it "negative" if it apologizes for not knowing the answer, lacks access to information, or is incomplete.
</score>

Remember to provide the reasoning before the score. Be thorough in your evaluation but concise in your explanation.
```

olafurpg (approved):
Nice!

agent/src/cli/cody-bench/strategy-chat.ts
```diff
@@ -57,12 +61,16 @@ export async function evaluateChatStrategy(
         if (response.type === 'transcript') {
             const reply = response.messages.at(-1)
             if (reply?.text) {
-                console.log({ reply })
+                const response = PromptString.unsafe_fromLLMResponse(reply.text)
```
olafurpg, line 64: Should probably be `unsafe_fromUserQuery`
```suggestion
                const response = PromptString.unsafe_fromUserQuery(reply.text)
```

agent/src/cli/cody-bench/llm-judge.ts
```diff
@@ -65,10 +65,12 @@ export class LlmJudge {
 function scoreNumeric(score?: string): number | undefined {
     switch (score) {
         case 'bad':
+        case 'negative':
             return 0
         case 'acceptable':
             return 1
         case 'amazing':
```
jtibshirani, line 72: I just noticed, can we map `positive` to 1 right now? It makes the results more interpretable (it's like you get a question right vs. wrong).

agent/src/cli/cody-bench/strategy-chat.ts
abeatrix: this is the reply from the LLM though!

olafurpg: Aaah, my bad!

abeatrix:
as discussed with @jtibshirani , i am running into issue with re-running the tests so we will merge what we have now and update the prompt and scoring in follow up PRs

Issue #4671, opened 2024-06-25
feedback: No longer able to edit the changes in "diff" mode


alex-tee:
### Version

v1.23.1719313083 (pre-release)

### Areas for Improvement

- [X] UI/UX
- [ ] Onboarding
- [ ] Docs
- [ ] Chat
- [ ] Commands
- [ ] Context
- [ ] Response Quality
- [ ] Other

### What needs to be improved? Please describe how this affects the user experience and include a screenshot.

After updating my VSCodium extensions, I am no longer able to edit Cody's changes in a new tab in diff mode Now, when i press Alt+K to edit a block of code, when Cody finishes it shows be a diff view but when I go to edit something the diff view goes away and the changes are applied. Now, I don't know what exactly changed because I can't compare what Cody did with what was there before, so it makes it very difficult to correct what Cody did (because it makes mistakes often and needs to be corrected manually).

### Describe the solution you'd like to see

Bring back the option to edit changes made by Cody in "diff" mode in another tab.

### Describe any alternatives that could be considered

Don't go out of "diff" (currently preview-only) mode when the user attempt to edit some change made by Cody. Make this view editable.

umpox:
Thanks for raising this so quickly @alex-tee!

> Don't go out of "diff" (currently preview-only) mode when the user attempt to edit some change made by Cody. Make this view editable.

I think this is likely the best approach going forward. Removing the diff automatically on change is quite jarring, especially when there are deleted lines

Will be improved by: https://github.com/sourcegraph/cody/pull/4684

alex-tee:
Thanks for the change. I tried it now. It kinda works but it's very buggy if you add a few lines (the bottom of the text blends in with other code that should not be affected), and there is no ability to revert changes per diff-block like before. It feels like loss of functionality. What was wrong with opening a proper diff to edit in a new tab?

alex-tee:
Another disadvantage of not having the diffs in a separate temporary tab is that I can't even open my file separately to view it normally. The file in the new tab gets colored too.

umpox:
@alex-tee Thanks, I understand this a bit more now. I didn't anticipate that the separate diff view would be used to revert part of an edit, but that is totally a valid use case and we should support that.

I think eventually I want to support doing this entirely in-file, i.e. we show a simple "Accept / Discard" CTA above each change, so you can still have partial acceptance without needing to leave your file.

[...]

PR #4684, opened 2024-06-26
Edit: Allow editing the inline diff

umpox:
## Description

closes https://github.com/sourcegraph/cody/issues/4671

https://github.com/sourcegraph/cody/assets/9516420/7d468713-c928-461b-b0f0-192a91560d44

Problem:
- We auto-accept the task whenever there is _any_ change within the `selectionRange`.
   - This was primarily useful for tracking if the user immediately does "Undo" on a change, so we update the UI and clear out the diff.
- With the new inline diff view, doing this means that it's difficult for users' to modify Cody's response. As any change they make immediately removes all deleted lines and hide decorations.

Solution:
- Allow making edits within the selection range. We can ensure decorations do not expand so it's still clear exactly what edits Cody made.
- Track specifically for the edit being restored back to its original state. This will work to ensure we still capture when a user does an "Undo"

## Test plan

1. Make Edits (Generate, Edit, Test, Fix, Document)
2. Make changes to the diff
3. Check everything looks ok
4. Try to undo _some_ of the changes you made
5. Check the entire edit is not removed, as we're only undoing additional changes
6. Undo all the changes made after the edit
7. Check the entire edit is removed (e.g. codelens, injected whitespace, etc)


abeatrix (commented):
I asked the LLM to update the function name but the whole function was shown as removed:

image


And when I tried to undo it, it broke my code 😓 
image

Looks like it might be Gemini issues? Should we disable it for Edit?

https://github.com/sourcegraph/cody/assets/68532117/6b59c248-e900-4fec-bdee-1a36afeaf855

abeatrix:
unrelated to this PR but we should also update the ghost text not to show up in diff lines? 

image

abeatrix (approved):
Run into some issues when using the Gemini model, which seems to be unrelated to the changes in this PR. I was able to edit the green diff as described in the PR 👍

vscode/src/non-stop/FixupController.ts
```diff
@@ -207,6 +207,56 @@ export class FixupController
         }
     }
 
+    /**
+     * Given a task, tracks upcoming changes in the associated document.
+     * If the change restores an applied task to its original state, we discard the task
+     * meaning any associated UI and behaviour is updated.
+     */
+    public registerDiscardOnRestoreListener(task: FixupTask): void {
```
abeatrix, line 215: changelog?

umpox:
Thanks @abeatrix I'll look into Gemini, maybe we should disable that for edit

PR #4690, opened 2024-06-26
update events to have enum numerical values for telemetry recording on `metadata`

akalia25:
This PR updates the event recording process to capture `mode`, `intent`, `source`, `completionIntent`, etc. as numerical values within the `metadata` field. This enhancement allows us to record these data points for all instances and not just dotcom while maintaining high security.

Key changes:
- Created enum lookups for `mode`, `TriggerKind`, `intent`, `source`, `EventSource`, `InlineCompletionsResultSource`, ,`CompletionIntent`
- Modified the telemetry recording process to convert enum values to their corresponding numerical representations.
- These numerical values are stored in the metadata field of each event.

Benefits:
- Data is recorded for all instances and not just dotcom, enabling better enterprise analytics

Example: In the updated system, the `metadata` parameter now includes numerical representations for above values (e.g. `TriggerKind`, `intent`, etc.) with enum lookups., each derived from their respective enum lookups at the time of telemetry recording.
image

## Test plan
CI + Locally


bobheadxi:
@akalia25 in your screenshot it looks like there's an empty metadata field, is that expected?

image

akalia25:
> @akalia25 in your screenshot it looks like there's an empty metadata field, is that expected?
> 
> image

This is just my VSCode Editor's formatting hiding the data inside it. The solid black part stays static and the area with vertical lines moves, so when I took the screenshot, it made it look empty, but there is data in there :)

vscode/src/completions/logger.ts
```diff
@@ -382,6 +387,68 @@ function writeCompletionEvent<SubFeature extends string, Action extends string,
...
+        metadataMapping: T | { [key: string]: number } | { [key: number]: string | number }
+    ): number | undefined {
+        if (value === undefined) return undefined
+
+        const enumKey = Object.keys(enumType).find(key => enumType[key] === value)
+        if (!enumKey) return undefined
+
+        const mappingValue = (metadataMapping as any)[enumKey]
+        return typeof mappingValue === 'number' ? mappingValue : undefined
+    }
```
valerybugakov, line 409: I love the idea! Suggesting some improvements:

1. We can bind value enums and mapping objects via the type system. This ensures that the `TriggerKindMetadataMapping` is always updated if `TriggerKind` is changed. I suggest applying this change to all metadata mapping objects.

```ts
export enum TriggerKind {
    /** Completion was triggered explicitly by a user hovering over ghost text. */
    Hover = 'Hover',

    /** Completion was triggered automatically while editing. */
    Automatic = 'Automatic',

    /** Completion was triggered manually by the user invoking the keyboard shortcut. */
    Manual = 'Manual',

    /** When the user uses the suggest widget to cycle through different completions. */
    SuggestWidget = 'SuggestWidget',
}

export const TriggerKindMetadataMapping: Record<TriggerKind, number> = {
    [TriggerKind.Hover]: 1,
    [TriggerKind.Automatic]: 2,
    [TriggerKind.Manual]: 3,
    [TriggerKind.SuggestWidget]: 4,
}
```

2. We use a regular object instead of `enum` for metadata mapping to disallow numbers in keys. 
3. We rely on this binding in `mapEnumToMetadata` where we enforce matching between `valueEnum` and `metadataMapping` keys.

```suggestion
    function mapEnumToMetadata<
        V extends Record<string, string>,
        // Do not allow number keys in `metadataMapping`
        K extends keyof V extends string ? string : never,
    >(value: string | undefined, valueEnum: V, metadataMapping: Record<K, number>): number | undefined {
        if (value === undefined) return undefined

        const enumKey = Object.keys(valueEnum).find(key => valueEnum[key] === value)
        if (!enumKey) return undefined

        const mappingValue = metadataMapping[enumKey as K]
        return typeof mappingValue === 'number' ? mappingValue : undefined
    }

```

valerybugakov (commented):
This is great! Looking forward to exploring this data 🙌

vscode/src/completions/logger.ts
akalia25: This is a great idea, thank you! Have applied these changes and convert all the `metadataMappings...` to use `export const mapping: record<type, string >`

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -925,14 +925,21 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
         this.contextFilesQueryCancellation = cancellation
 
         const source = 'chat'
+
+        // Use enum to send source values to metadata, making this data available on all instances.
+        enum atMentionSourceTelemetryMetadataMapping {
+            chat = 1,
+        }
```
akalia25, line 932: @valerybugakov, For this one, would you also recommend converting to const? 

i.e.
```suggestion
        // Use numerical mapping to send source values to metadata, making this data available on all instances.
        const atMentionSourceTelemetryMetadataMapping = {
            chat: 1,
        } as const
```

vscode/src/non-stop/FixupController.ts
```diff
@@ -176,9 +182,14 @@ export class FixupController
         const editOk = await this.revertToOriginal(task, editor.edit)
 
         const legacyMetadata = {
-            intent: task.intent,
-            mode: task.mode,
-            source: task.source,
+            intent:
+                EditIntentMetadataMapping[task.intent as keyof typeof EditIntentMetadataMapping] ||
```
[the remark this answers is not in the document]
akalia25: Great call-out! An overlook on my part! Updated to be lowercase on the mapping :)

vscode/src/tree-sitter/queries.ts
```diff
@@ -40,6 +40,25 @@ export const intentPriority = [
...
+    comment = 8,
+    'pair.value' = 9,
+    argument = 10,
+    parameter = 11,
+    parameters = 12,
+    'jsx_attribute.value' = 13,
+    'return_statement.value' = 14,
+    return_statement = 15,
+    string = 16,
+}
```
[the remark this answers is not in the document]
akalia25: Updated! Thank you. The only concern I would have, is if the ordering of `intentPriority` ever changes due to a new linter or something and then all the `...telemetryMetadataMappings` would automatically get shuffled too? Do you know if this something to be concerned about? 🤔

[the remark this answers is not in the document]
valerybugakov: You're right; it can be an issue. We probably should take the same approach here as in other mappers.

vscode/src/chat/chat-view/ChatController.ts
valerybugakov: Yeah, the main benefit would be that we can link it to possible `source` values: 

```ts
const source = 'chat'

const atMentionSourceTelemetryMetadataMapping: Record<typeof source, number> = {
  chat: 1,
} as const

valerybugakov (approved):
Preemptive approval! Looks great.

vscode/src/chat/chat-view/ChatController.ts
akalia25: Thank you for confirming and explaining!

vscode/src/tree-sitter/queries.ts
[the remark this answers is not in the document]
akalia25: Thanks for confirming! Updated it, so that it takes the same approach as other mappers. :)

PR #4691, opened 2024-06-26
Auth: add offline mode support for Ollama models

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2631/cody-requires-an-online-connection-and-an-open-vs-code-to-work-offline

This change adds support for an offline mode that allows users to only use Ollama models with Cody without the internet.

The key changes are:

- Added a new `offlineModeAuthStatus` object to represent the auth status in offline mode
- Updated the `SimpleChatPanelProvider` and `AuthProvider` to handle the 'offline' auth kind
- Modified the `newAuthStatus` function to return the offline mode auth status if `isOfflineMode` is true
- Updated the `syncModels` function to handle the offline mode case and not sync models
- Added a new "Use Cody Offline with Ollama" button on the Connection Issues page

The screen below only shows up when there is a networking issue, e.g. when Cody does not have internet connection:
image

## TO-DO

- Add tests for offline mode?

## Test plan

2. Open Ollama
3. Open Chat and ask Cody a question to verify Cody still works when there is an Internet connection
4. Turn off your Wi-Fi settings
5. Reload VS Code
6. Verify you are seeing the connection issue page
7. Click on `Use Cody Offline with Ollama`
8. Verify the sidebar is loaded correctly
9. Open Chat and select an Ollama instruct model
10. Ask Cody a question and verify you are getting a response back from the Ollama model

### Demo

image


sqs:
reviewing

sqs (approved):
Looks great to me. Works locally. This will completely unblock a lot of people.

As a future PR (unclear priority), it would be great if we made it so that Cody worked without network access in the normal mode, instead of being in a separate offline mode. For example, Cody would start up and start working using the last configured account even before it had logged in again with the server. I can see complexity in that route, though, and it's marginal benefit above this.

PR #4693, opened 2024-06-26
Bench: add hedging and conciseness scores for Chat

jtibshirani:
This PR adds 'hedging' and 'conciseness' checks for Chat benchmarks. The 'conciseness' check uses an LLM to judge repetitiveness.

It also tweaks the main LLM judge score to just focus on helpfulness, and not penalize models for apologizing. This makes the scores less biased, since some models tend to apologize more ("I'm sorry...") instead of using other hedging terms like "Unfortunately...", but otherwise the responses are very similar.

## Test plan
Ran  `pnpm agent:skip-root-build cody-bench --evaluation-config ~/code/cody-leaderboard/chat-bench.json` and checked output.

### DIFF
--- a/vscode/src/extension.node.ts
+++ b/vscode/src/extension.node.ts
@@ -54,6 +54,10 @@ export function activate(
         .getConfiguration()
         .get<boolean>('cody.experimental.symf.enabled', true)
 
+    const isTelemetryEnabled = vscode.workspace
+        .getConfiguration()
+        .get<boolean>('cody.experimental.telemetry.enabled', true)
+
     return activateCommon(context, {
         createLocalEmbeddingsController: isLocalEmbeddingsDisabled
             ? undefined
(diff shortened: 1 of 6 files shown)


jtibshirani:
I'll soon open a cody-leaderboard PR to display the new scores like this:
Screenshot 2024-06-26 at 12 27 01 PM

vscode/src/extension.node.ts
```diff
@@ -54,6 +54,10 @@ export function activate(
         .getConfiguration()
         .get<boolean>('cody.experimental.symf.enabled', true)
 
+    const isTelemetryEnabled = vscode.workspace
```
jtibshirani, line 57: This PR also adds the ability to disable telemetry, which helps make cody-bench more reliable.

abeatrix (approved):
Looks AMAZING!

PR #4695, opened 2024-06-26
Cody CLI: improve `chat` subcommand

olafurpg:
Fixes CODY-2580

This PR makes several changes to the `chat` subcommand of the Cody cli

* It's now `cody-agent chat` instead of `cody-agent experimental-cli chat`. The `cody-agent` binary will eventually get renamed to just `cody` so the final command becomes `cody chat`.
* It uses an embedded agent client instead of starting a separate Node.js process for the server, improving performance.
* It supports `--context-file` to reference local files
* It now has integration tests with a fast edit/test/debug feedback loop making it very fun/enjoyable to hack on new functionality.

## Test plan

See test cases.


agent/src/agent.ts
```diff
@@ -100,6 +105,14 @@ type ExtensionActivate = (
 function copyWinCaRootsBinary(extensionPath: string): void {
     const source = path.join(__dirname, 'win-ca-roots.exe')
     const target = path.join(extensionPath, 'dist', 'win-ca-roots.exe')
+    try {
+        const stat = statSync(source)
+        if (!stat.isFile()) {
+            return
+        }
+    } catch {
+        return
```
eseliger, line 114: no logging, no bail out?

eseliger (approved):

agent/src/agent.ts
olafurpg: Added a logging statement on Windows
```
        if (isWindows()) {
            logDebug('win-ca', `Failed to find ${source}, skipping copy`)
        }
```

aramaraju:
🙏

PR #4696, opened 2024-06-26
shared contexts WIP

sqs:
HT https://x.com/kylecarbs/status/1806106785033449922


PriNova:
If I understand this PR correctly, is this some kind of memory/caching system, which can be appended to a chat session as additional context?
If yes, this is amazing.

PR #4697, opened 2024-06-27
Agent: handle missing chat model for unauthenticated users

abeatrix:
When a user is not logged in, set the default chat model to the dotcom default model. This is because unauthed users are expected not to have a default model as the model list is only synced on login.

## Test plan

No feature was changed so CI is expected to be green.


steveyegge (approved):
Thanks Bee!

PR #4699, opened 2024-06-27
Autocomplete: add `providerModel` to stage counter logger

valerybugakov:
- Adds `providerModel` to autocomplete stage counter events.
- Closes https://linear.app/sourcegraph/issue/CODY-2636/[autocomplete-latency]-add-providermodel-to-autocomplete-stage-counter
- Discussed [here](https://sourcegraph.slack.com/archives/C0729T2PBV2/p1719426870642669?thread_ts=1719286653.272099&cid=C0729T2PBV2).

## Test plan

Updated tests

### DIFF
--- a/vscode/src/services/autocomplete-stage-counter-logger.ts
+++ b/vscode/src/services/autocomplete-stage-counter-logger.ts
@@ -19,10 +19,20 @@ const INITIAL_STATE = {
 
 export class AutocompleteStageCounter implements vscode.Disposable {
     private nextTimeoutId: NodeJS.Timeout | null = null
+    private providerModel: string | null = null
     private currentState = { ...INITIAL_STATE }
 
     constructor() {
         this.nextTimeoutId = setTimeout(() => this.flush(), LOG_INTERVAL)
+    }
+
+    public setProviderModel(providerModel: string): void {
+        // Flush the current counter on model change.
+        if (this.providerModel !== null && this.providerModel !== providerModel) {
+            this.flush()
+        }
+
+        this.providerModel = providerModel
     }
 
     public flush(): void {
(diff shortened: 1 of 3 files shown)


vscode/src/services/autocomplete-stage-counter-logger.ts
```diff
@@ -19,20 +19,34 @@ const INITIAL_STATE = {
...
 
-        telemetryRecorder.recordEvent('cody.completion.stageCounter', 'flush', {
-            metadata: stateToLog,
-        })
+        // Do not log empty counter events.
+        if (Object.values(stateToLog).some(count => count > 0)) {
+            telemetryRecorder.recordEvent('cody.completion.stageCounter', 'flush', {
+                metadata: stateToLog,
+                privateMetadata: { providerModel: this.providerModel },
+            })
```
akalia25, line 48: If we have a known list of all the `providerModel` values, we could create an enum for this and promote this field to metadata so that it is collected on all instances, not just dotcom. 

Alternative If `providerModel` is constantly getting updated (enum would not be a good approach), and this data is valuable to collect for all instances (outside of dotcom), let me know and we can add this to our special telemetry allowlist ([ref](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/telemetry/sensitivemetadataallowlist/sensitivemetadataallowlist.go?L25-39))

valerybugakov: Thanks, Aditya!

- The list of values can be updated when we A/B test new models or add new autocomplete providers. 
- The primary purpose of this logger is to analyze the A/B tests we conduct on DotCom, and having it for other instances is not critical.

hitesh-1997: Right, Having in the privateMetaData should be consistent with other events as well, so that we don't have to handle this as special case.

hitesh-1997 (approved):
LGTM !!

vscode/src/services/autocomplete-stage-counter-logger.ts
akalia25: Thanks for clarifying, you two!

akalia25 (approved):

PR #4702, opened 2024-06-27
Chat context observability improvements

janhartman:
This PR starts logging chat context items for dotcom users on public repos.

I noticed that we didn't propagate ignored context items correctly, which caused us to always log 0 ignored items in our telemetry. However, after fixing this, we now show the count of ignored items in the chat as shown on the image below. 
I don't think that this is really informative to the user so I removed this visual element, WDYT?
image

## Test plan
Tested locally.


vscode/webviews/chat/cells/contextCell/ContextCell.tsx
```diff
@@ -44,14 +44,10 @@ export const ContextCell: React.FunctionComponent<{
     }
 
     const itemCount = usedContext.length
-    let contextItemCountLabel = `${itemCount} ${!isForFirstMessage ? 'new ' : ''}${pluralize(
+    const contextItemCountLabel = `${itemCount} ${!isForFirstMessage ? 'new ' : ''}${pluralize(
```
hitesh-1997, line 47: Just tagging @toolmantim for the UI/UX related changes here

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1191,24 +1192,41 @@ export class SimpleChatPanelProvider
...
+                source: i.source,
+                size: i.size || TokenCounter.countTokens(i.content || ''),
+                content: i.content,
+            })),
+        })
+
+        return {
+            included: getContextStats(context.used),
+            excluded: getContextStats(context.ignored),
+            gitMetadata: await this.getRepoMetadataIfPublic(),
```
hitesh-1997, line 1226: @janhartman I think we shouldn't add the changes here. If I understand correcting, adding gitMetaData will start logging in the `metadata` fields instead of `privateMetaData` field which goes through normal pipeline.
Also, I would suggest a review from @akalia25 for any telemetry PR !!

hitesh-1997 (changes requested):

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
janhartman: Thanks for the review. I think the setup is OK - if you look at [sendTelemetry](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts?L781), `privateContextStats` is in `privateMetadata`. This is what we want, right?

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -111,6 +111,7 @@ export class DefaultPrompter {
                 // Because this enhanced context is added for the last human message,
                 // we will also add it to the context list for display.
                 context.used.push(...newEnhancedMessages.added)
+                context.ignored.push(...newEnhancedMessages.ignored)
```
abeatrix, line 114:
```suggestion
               
```

This was excluded on purpose initially because we only want to show files that were added by user (Commented on line 96). Since users didn't choose the enhance context files the product team didn't think it'd be helpful for users to see the context they didn't know it'd be included.

abeatrix: @toolmantim @danielmarquespt to confirm

abeatrix (commented):
Left a comment inline. I had a similar PR opened for displaying the ignored items as well for awhile but was left forgotten, but there are some discussion happened that might be related: https://github.com/sourcegraph/cody/issues/4490

vscode/src/chat/chat-view/prompt.ts
janhartman: That's why I removed the visual element. The problem with simply not adding the ignored items to this array is that they won't be propagated and logged correctly.

janhartman: It's quite important that we can log this properly for AI research's sake :)

vscode/webviews/chat/cells/contextCell/ContextCell.tsx
janhartman: To be clear - this change means that we keep the UI in the same state as it was. The difference is that now, the code passing the data around in the background works correctly :)

vscode/src/chat/chat-view/prompt.ts
abeatrix: Ah gotcha in that case would it be easier to do the logging in PromptBuilder instead? 🤔

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
hitesh-1997: ah right, I missed that, thanks !!

hitesh-1997 (approved):
Looks good to me !!
We can also ask @akalia25 to give it a look for telemetry related changes

vscode/src/chat/chat-view/prompt.ts
janhartman: TBH I think the way it's done now is better since we can log it closer to where it's used. The webview also logs how many items were excluded so we do need to pass this info there as well.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1191,24 +1192,41 @@ export class SimpleChatPanelProvider
...
-                files.length && {
-                    countFiles: files.length,
-                    fileSizes: files.map(f => f.size).filter(isDefined),
-                }
-            // NOTE: The private context stats are only logged for DotCom users
-            const privateContextStats = {
-                included: getContextStats(context.used.filter(f => f.source === 'user')),
-                excluded: getContextStats(context.ignored.filter(f => f.source === 'user')),
-            }
+            const privateContextStats = await this.buildPrivateContextStats(context)
```
willdollman, line 1195: This var + method could be renamed to indicate that they no longer just contain stats - e.g. `privateContextTelemetry` / `privateContextSummary` / or similar

willdollman (approved):
Looks fine to me, left a few small code style suggestions.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -1191,24 +1192,41 @@ export class SimpleChatPanelProvider
...
 
+    private async buildPrivateContextStats(context: {
+        used: ContextItem[]
+        ignored: ContextItem[]
+    }): Promise<object> {
+        // 🚨 SECURITY: included only for dotcom users & public repos
+        const isDotCom = this.authProvider.getAuthStatus().isDotCom
+        const isPublic = (await this.codebaseStatusProvider.currentCodebase())?.isPublic
+
+        if (!(isDotCom && isPublic)) {
```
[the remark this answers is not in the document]
janhartman: Interesting, for me it's the other way around 😄

PR #4705, opened 2024-06-27
Stevey/fallback to default chat model

steveyegge:
We have a race condition on integration test startup in which the `ModelsService` is not yet initialized. We had been throwing an error, but a better approach is to use the default model. This is also the behavior on the backend when we're contacting the LLM, so it's consistent. Fix suggested by @abeatrix. 

## Test plan

Manually tested.


abeatrix (approved):
ty!

PR #4707, opened 2024-06-27
Bench: add class for chat questions

jtibshirani:
This PR adds a 'question class' column, so we can categorize chat questions during evals.

## Test plan

Also updated Cody leaderboard to expose the question class:

Screenshot 2024-06-27 at 11 03 42 AM

### DIFF
--- a/agent/src/cli/cody-bench/EvaluationDocument.ts
+++ b/agent/src/cli/cody-bench/EvaluationDocument.ts
@@ -300,7 +300,7 @@ export const headerItems: ObjectHeaderItem[] = [
     { id: 'editDiff', title: 'EDIT_DIFF' },
     { id: 'chatReply', title: 'CHAT_REPLY' },
     { id: 'chatQuestion', title: 'CHAT_QUESTION' },
-    { id: 'questionClass', title: 'QUESTION_CLASS'},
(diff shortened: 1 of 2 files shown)


agent/src/cli/cody-bench/EvaluationDocument.ts
```diff
@@ -299,6 +300,7 @@ export const headerItems: ObjectHeaderItem[] = [
     { id: 'editDiff', title: 'EDIT_DIFF' },
     { id: 'chatReply', title: 'CHAT_REPLY' },
     { id: 'chatQuestion', title: 'CHAT_QUESTION' },
+    { id: 'questionClass', title: 'QUESTION_CLASS' },
```
olafurpg, line 303: We may want to have the name of the "task" as a prefix in the header
```suggestion
    { id: 'questionClass', title: 'CHAT_QUESTION_CLASS' },
```

olafurpg (approved):
LGTM 👍🏻

agent/src/cli/cody-bench/EvaluationDocument.ts
jtibshirani: I liked this more generic name, since we can use it to categorize other tasks as well!

PR #4710, opened 2024-06-27
feat(bench): unit test framework for cody-bench

jamesmcnamara:
Adds a framework for evaluating "Generate Unit Test" performance through cody-bench. Reports a number of binary metrics like if expected test frameworks were used or if the test file was generated in the right place.

See https://github.com/sourcegraph/cody-leaderboard/pull/11 for a usage example

## Test plan
Has been executed through https://github.com/sourcegraph/cody-leaderboard/pull/11


agent/scripts/export-cody-bench-vars.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+#
+# The purpose of this script is to export access tokens forcody-bench
+# Usage:
+#   cd CODY_DIRECTORY
+#   agent/scripts/export-cody-bench-vars.sh dotcom
+
+SOURCE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
jtibshirani, line 8: Is there a reason to use something other than dot com? Maybe we can instead rely on the existing `agent/scripts/export-cody-http-recording-tokens.sh`, and update the framework to use `SRC_DOTCOM_PRO_ACCESS_TOKEN`?

agent/src/cli/cody-bench/EvaluationDocument.ts
```diff
@@ -271,9 +273,24 @@ interface EvaluationItem {
     info?: CompletionItemInfo
     event?: CompletionBookkeepingEvent
     eventJSON?: string
+    testName?: string
+    testExpectedFile?: string
+    testFile?: string
+    testInputFile?: string
+    testLanguage?: string
```
jtibshirani, line 280: Could we reuse `languageId` here?

agent/src/cli/cody-bench/EvaluationDocument.ts
```diff
@@ -271,9 +273,24 @@ interface EvaluationItem {
...
+    testGenerated?: string
+    testMatchesExpectedTestFile?: boolean
+    testHasTypescriptErrors?: boolean
```
jtibshirani, line 283: Maybe this could just be `resultTypechecks`?

agent/src/cli/cody-bench/EvaluationDocument.ts
```diff
@@ -271,9 +273,24 @@ interface EvaluationItem {
     info?: CompletionItemInfo
     event?: CompletionBookkeepingEvent
     eventJSON?: string
+    testName?: string
+    testExpectedFile?: string
+    testFile?: string
```
jtibshirani, line 278: It'd be nice to call this "filename" or "path" like `testPath` to distinguish this from the ones that contain the file contents. Same thought for `testInputFile`.

agent/src/cli/cody-bench/EvaluationDocument.ts
```diff
@@ -271,9 +273,24 @@ interface EvaluationItem {
...
+    testInputFile?: string
+    testLanguage?: string
+    testGenerated?: string
+    testMatchesExpectedTestFile?: boolean
```
jtibshirani, line 282: Could we just calculate this after the fact in cody-leaderboard? It's a little redundant to pass this through

jtibshirani (commented):
Looks good to me! Just left some small comments.

jtibshirani:
Thanks @jamesmcnamara this is looking great! Did you see my comments above? After those are resolved, looks ready to merge.

agent/src/cli/cody-bench/EvaluationDocument.ts
jtibshirani: I believe we already use `resultTypeChecks` for this though (I was referring to the variable above)?

jtibshirani (approved):

agent/src/cli/cody-bench/EvaluationDocument.ts
jamesmcnamara: I still don't like that because it's way too natural to assume by looking at the data that it applies to all results and interpreting all other cases as either failures or successes when it's only relevant to typescript files. I removed the variable entirely in favor of just using the diagnostics var, but I still am worried someone will mis-interpret that to mean all other cases were successful.

olafurpg: We should treat empty `RESULT_TYPECHECKS` as neither success or failure. I recommend reusing `RESULT_TYPECHECKS` over adding custom headers

PR #4712, opened 2024-06-27
Auth: update Ollama offline mode UI

abeatrix:
Follow up on https://github.com/sourcegraph/cody/pull/4691#pullrequestreview-2143240367

This PR addresses the linked comment by @sqs :

> it would be great if we made it so that Cody worked without network access in the normal mode, instead of being in a separate offline mode. For example, Cody would start up and start working using the last configured account even before it had logged in again with the server. 

Proposed solution in this PR:

- store the last used username so that we will be able to get the chat history from the last logged-in user 

image

## Test plan

2. Open Ollama
3. Open Chat and ask Cody a question to verify Cody still works when there is an Internet connection
4. Turn off your Wi-Fi settings
5. Reload VS Code
6. Verify you are seeing the connection issue page
7. Click on Use Cody Offline with Ollama
8. Verify the sidebar is loaded correctly
9. Verify you can see the Chat History for your last logged-in user
10. Open Chat and select an Ollama instruct model
11. Ask Cody a question and verify you are getting a response back from the Ollama model

## Demo

https://github.com/sourcegraph/cody/assets/68532117/c2d1b883-fc7b-442d-8a75-71d121cca92c

### Before

As you will be logged in as an unknown user, your chat history from your previous sign in will not be showing up:

image

### After

In Offline mode, we will retrieve your last stored username and endpoint to get the local chat history:

image

### DIFF
--- a/vscode/src/auth/account-menu.ts
+++ b/vscode/src/auth/account-menu.ts
@@ -1,0 +1,51 @@
+import { type AuthStatus, isDotCom } from '@sourcegraph/cody-shared'
+import * as vscode from 'vscode'
+
+export enum AccountMenuOptions {
+    SignOut = 'Sign Out',
+    Manage = 'Manage Account',
+    Switch = 'Switch Account...',
+}
+
+export async function openAccountMenu(authStatus: AuthStatus): Promise<AccountMenuOptions | undefined> {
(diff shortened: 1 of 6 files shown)


umpox (approved):
Nice!

PR #4713, opened 2024-06-27
CLI: interactively stream reply

olafurpg:
Previously, the `cody chat` cli command didn't print any output while the LLM was replying. This was a poor user experience because you don't know for how long you have to wait. Now, the cli streams the reply making the cli a lot more enjoyable to use.

Fixes CODY-2637

https://github.com/sourcegraph/cody/assets/1408093/5c3879bd-b38b-4522-9e22-ce01277604b5

## Test plan

Green CI


agent/src/esbuild.mjs
```diff
@@ -29,7 +29,7 @@ async function verifyShim() {
         write: false,
         outfile: path.join('dist', 'shim.js'),
         plugins: shimPlugins,
-        external: ['typescript'],
+        external: ['typescript', 'react-devtools-core'],
```
eseliger, line 32: do we still need that? I don't think ora needs this?

eseliger (approved):

agent/src/esbuild.mjs
olafurpg: Will remove separately

PR #4717, opened 2024-06-28
Autocomplete: keep the last candidate in cache if it is not applicable

valerybugakov:
- This PR ensures that we keep suggestions linked to the last candidate if the last candidate cannot be applied for the current position. Noticed this behavior while debugging our analytics events locally.
- This should increase the cache-hit rate and reduce the E2E latency when this scenario is encountered.
- Closes https://linear.app/sourcegraph/issue/CODY-2663/[autocomplete-latency]-keep-the-last-candidate-suggestions-in-cache-if

## Test plan

1. Generate a single-line suggestion by adding a new line with some prefix.
2. Delete this line, reverting the document to the previous state.
3. Go to another line in the document below with less indentation and generate another suggestion.
4. Generate the first suggestion again by returning to the first step.
5. Ensure that it comes from the cache.


umpox (approved):
Nice!

> increase the cache-hit rate

Curious if we track this anywhere? Seems like it'd be useful data

PR #4725, opened 2024-06-28
testing github action

keegancsmith:


keegancsmith:
@RXminuS here it is working on github actions :) Had to do some hacks to put the downloaded code onto $PATH otherwise code-tunnel complains that vscode is not installed.

keegancsmith:
@RXminuS alright not working in CI, and nearly working on my windows machine. Gonna take a break with it for now. The main issue I am running into on both linux and windows is that "code-tunnel" seems to be a lightweight cli which manages versions of vscode. Convincing it that the vscode we have downloaded is the one to use is the problem. In both cases I hit a webpage when visiting which says "the latest version of vscode is downloading". If you give higher timeouts that will download and then run the tests. I wonder if the same is happening for you on mac and we are in fact not actually using the downloaded vscode?

I did some stuff to try and force it using the "vscode version use" stuff but still not a consistent experience.

RXminuS:
Mmm that's strange, that shouldn't be able to happen 🤔

RXminuS:
Ok, I think I've fixed the downloading of vscode server

vscode/e2e/utils/vscody/fixture.ts
```diff
@@ -397,6 +397,25 @@ const implFixture = _test.extend<TestContext, WorkerContext>({
                 )
             )
 
+            // code cli can complain it can't find a version. So explicitly set it the downloaded vscode to it.
+            const cliDataDir = path.join(serverRootDir, 'cli')
+            await fs.mkdir(cliDataDir, { recursive: true })
+            const env = {
+                // inherit environment
+                ...process.env,
+                VSCODE_CLI_DATA_DIR: cliDataDir,
```
RXminuS, line 406: You can remove this now after you pull upstream fix

keegancsmith: Thanks, what I'll do now is try out your branch on linux and then windows and see what I need to change.

PR #4726, opened 2024-06-29
CLI: bump package version to v0.1.0

olafurpg:
Previously, the CLI release failed because the version number in package.json was outdated.
```
Release tag and version in agent/package.json do not match: '0.1.0' vs. '0.0.5b'
```

## Test plan
Cut a new release and see if the CI passes.

### DIFF
--- a/agent/src/cli/__snapshots__/chat.test.ts.snap
+++ b/agent/src/cli/__snapshots__/chat.test.ts.snap
@@ -23,9 +23,10 @@ exports[`--context-repo (squirrel test) 1`] = `
 exitCode: 0
 stdout: >+
   Squirrel is a high-performance, open-source object-relational database (ORD)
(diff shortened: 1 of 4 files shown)


agent/src/cli/__snapshots__/chat.test.ts.snap
```diff
@@ -23,9 +23,10 @@ exports[`--context-repo (squirrel test) 1`] = `
 exitCode: 0
 stdout: >+
   Squirrel is a high-performance, open-source object-relational database (ORD)
-  that provides a SQL interface for working with structured and semi-structured
-  data stored in modern cloud data stores like Amazon S3, Azure Blob Storage,
```
olafurpg, line 27: I'm planning to avoid these diffs for future releases but just want to get this release out ASAP

sqs (approved):

PR #4729, opened 2024-06-30
Chat: enable sidebar refresh and moving from sidebar to editor

beyang:
https://github.com/sourcegraph/cody/assets/1646931/48ebfc54-ebb2-4039-ac15-c5dc67344f2c

## Changes

- Adds a button to refresh the chat in the sidebar. Previously, there was no way to reset/clear the state of the chat view in the sidebar
- Adds a button to move the chat session in the sidebar to a new editor panel.

## Implementation notes

Populating the initial context chips in the chat input depends on some behavior that is only triggered on the first-time render. In order to trigger this, we key the chat input component on the chat session ID, rather than just the array index. so that these components are recreated when the chat session changes on reset. Otherwise, the existing component from the previous session would be reused, and the input would not be populated with the initial context chips on reset.

## Test plan

Tested locally. Note: chat in the sidebar is not yet enabled, so this should have no effect for now.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -450,6 +450,22 @@
         "icon": "$(new-comment-icon)"
       },
       {
+        "command": "cody.chat.panel.sidebar.new",
+        "category": "Cody",
+        "title": "New Chat",
+        "enablement": "cody.activated",
+        "group": "Cody",
+        "icon": "$(new-comment-icon)"
+      },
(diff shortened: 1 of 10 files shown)


sqs (approved):

sqs:
Looks really nice. Looking fwd to using this

PR #4731, opened 2024-07-01
Cody Web: Add remote context support

vovakulikov:

This is the one of PRs that come from the bigger change we did for Cody Web in this main PR https://github.com/sourcegraph/cody/pull/4605

This PR adds support for the remote context items files and symbols. Note that eventually, we will use
open context provides for any remote contexts but for now as a semi-temporary solution for Cody Web 
we include this logic outside the open context providers.

This is safe to merge since it only affects Cody Web package. 

## Test plan
- Check that CI is passing 
- Check cody web demo (`cd web` and `pnpm dev`) and check that you can mention remote files and symbols 
- Check that mentioned files and symbols were included in the prompt message

### DIFF
--- a/lib/shared/src/mentions/query.ts
+++ b/lib/shared/src/mentions/query.ts
@@ -31,6 +31,13 @@ export interface MentionQuery {
      * `foo.txt:1`, or `foo.txt:12-`).
      */
     maybeHasRangeSuffix?: boolean
+
+    /**
+     * To control source of mention suggestions, if it's set to true
+     * search logic will try to find suggestions across remote repositories
+     * user has on their instance. (Cody Web use case)
+     */
+    includeRemoteRepositories?: boolean
 }
 
 /**
--- a/vscode/src/chat/context/chatContext.ts
+++ b/vscode/src/chat/context/chatContext.ts
@@ -27,7 +27,8 @@ export async function getChatContextItemsForMention(
     telemetryRecorder?: {
         empty: () => void
         withProvider: (type: MentionQuery['provider'], metadata?: { id: string }) => void
-    }
+    },
+    remoteRepositoriesNames?: string[]
 ): Promise<ContextItem[]> {
     const MAX_RESULTS = 20
     switch (mentionQuery.provider) {
@@ -37,11 +38,11 @@ export async function getChatContextItemsForMention(
         case SYMBOL_CONTEXT_MENTION_PROVIDER.id:
             telemetryRecorder?.withProvider(mentionQuery.provider)
             // It would be nice if the VS Code symbols API supports cancellation, but it doesn't
-            return getSymbolContextFiles(mentionQuery.text, MAX_RESULTS)
+            return getSymbolContextFiles(mentionQuery.text, MAX_RESULTS, remoteRepositoriesNames)
         case FILE_CONTEXT_MENTION_PROVIDER.id: {
             telemetryRecorder?.withProvider(mentionQuery.provider)
             const files = mentionQuery.text
-                ? await getFileContextFiles(mentionQuery.text, MAX_RESULTS)
+                ? await getFileContextFiles(mentionQuery.text, MAX_RESULTS, remoteRepositoriesNames)
                 : await getOpenTabsContextFile()
 
             // If a range is provided, that means user is trying to mention a specific line range.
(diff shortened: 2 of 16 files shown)


vscode/src/editor/utils/editor-context.ts
```diff
@@ -368,9 +417,30 @@ async function resolveFileOrSymbolContextItem(
...
+    if (isRemoteFileURI(contextItem.uri)) {
+        const { repository, path } = parseRemoteFileURI(contextItem.uri)
+
+        // TODO [VK]: Support ranges for symbol context items
+        const resultOrError = await graphqlClient.getFileContent(repository, path)
+
+        if (!isErrorLike(resultOrError)) {
+            return {
+                ...contextItem,
+                uri: URI.from({
```
abeatrix, line 429: might have missed something but curious why we want to reset the URI here? Would expect the scheme to be 'remote-file' or wouldn't it be filtered out by SUPPORTED_URI_SCHEMAS later accidentally?

vovakulikov: Good question, so any remote file mentions have a special URI which has "remote-file" scheme, this is how later in places like this handler we determine that this is a remote file and we should use remote file context fetcher logic. 

After this handler, there should be no difference in this file URI compared to other URIs, for example, when we will use this URI to build a file link in the context details in the chat, we should have no special scheme or authority for files like this but common URI with correct path.

valerybugakov (approved):
LGTM

lib/shared/src/mentions/query.ts
```diff
@@ -31,6 +31,13 @@ export interface MentionQuery {
      * `foo.txt:1`, or `foo.txt:12-`).
      */
     maybeHasRangeSuffix?: boolean
+
```
thenamankumar, line 34: I would recommend using the `MentionQuery.provider` field only for this purpose and not introducing this extra field here. 

Given that how we fetch file/symbol results on the web is totally different from that on VS Code, it makes sense not to combine the 2 logic into the current functions to fetch files/symbols from the local context. 


You can set different provider IDs based on mention resolution mode, preferably from inside `parseMentionQuery` only. And if that is not possible then from within the chat context client. 

Then, inside `getChatContextItemsForMention`, you can just add another switch case to capture this and call a separate method to get files/symbol results from remote search.

This won't dirty the current methods fetch the files/symbols and will keep things simple.

vscode/src/chat/context/chatContext.ts
```diff
@@ -37,11 +38,11 @@ export async function getChatContextItemsForMention(
         case SYMBOL_CONTEXT_MENTION_PROVIDER.id:
             telemetryRecorder?.withProvider(mentionQuery.provider)
             // It would be nice if the VS Code symbols API supports cancellation, but it doesn't
-            return getSymbolContextFiles(mentionQuery.text, MAX_RESULTS)
+            return getSymbolContextFiles(mentionQuery.text, MAX_RESULTS, remoteRepositoriesNames)
```
thenamankumar, line 41: Have separate methods to get context from remote.

vscode/src/chat/context/chatContext.ts
```diff
@@ -37,11 +38,11 @@ export async function getChatContextItemsForMention(
...
         case FILE_CONTEXT_MENTION_PROVIDER.id: {
             telemetryRecorder?.withProvider(mentionQuery.provider)
             const files = mentionQuery.text
-                ? await getFileContextFiles(mentionQuery.text, MAX_RESULTS)
+                ? await getFileContextFiles(mentionQuery.text, MAX_RESULTS, remoteRepositoriesNames)
```
thenamankumar, line 45: Have separate methods to get context from remote.

lib/shared/src/mentions/query.ts
thenamankumar: But if you have already evaluated that idea and it becomes too big of a change to have separate IDs, as we might be mentioning those IDs at a lot of places in the UI as well, then it is fine, you may ignore this.

thenamankumar (approved):

umpox (approved):

lib/shared/src/mentions/query.ts
vovakulikov: @thenamankumar, thanks for the suggestion; yeah, I was thinking about this and just wanted to make this change as safe as possible and didn't want to change anything about providers. 

As I mentioned in the PR description, I will try to fix open context providers as soon as possible to negate this change in this PR and use open context providers for remote context, so for now I think this change is ok to merge without very deep thinking.

PR #4735, opened 2024-07-01
Autocomplete: Re-implement `SmartThrottle`

umpox:
## Description

This PR:
- Adds `SmartThrottleService` which we implemented before, but removed as it showed negative results.
- Makes a minor improvement to reduce the debounce time to `0` when the 250ms throttle timeout is bypassed
- Fixes the issue we didn't notice before with `SmartThrottleService`:
  - We propagated the abort signal straight to the request manager with this change, which meant the `SmartThrottleService` could handle aborting requests as new ones came in. This was sound in theory, but we ran into problems because we also listened for `cancellationToken.onCancellationRequested` to abort the request. This meant that actually VS Code was aborting our requests, rather than the smart throttle, and we only ever had 1 inflight request at a time.
- Removes old, unused feature flags that had similar logic:
  -  `CodyAutocompleteEagerCancellation`
  - `CodyAutocompleteReducedDebounce`

## Test plan

- Tested locally with lots of logging on # of inflight requests, request cancellation and request start
- Unit tests


vscode/src/completions/smart-throttle.ts
```diff
@@ -0,0 +1,115 @@
...
+        })
+    }
+
+    private isNewStartOfLineRequest(request: RequestParams): boolean {
+        const isStartOfLine = request.docContext.currentLinePrefix.trim() === ''
+        const isDifferentUri =
+            request.document.uri.toString() !== this.startOfLineLocation?.uri.toString()
+        const isDifferentLine = request.position.line !== this.startOfLineLocation?.line
+        return isStartOfLine && (isDifferentUri || isDifferentLine)
+    }
```
umpox, line 89: Note: This logic is useful because it means we skip any debouncing and cancellation for new line requests.

These have a good chance at being correct from the start, or that the the user will return to them.

I considered that we may possible want to extend this to words too, e.g. `isNewStartOfWordRequest`. It would use `document.getWordRangeAtPosition` to determine if the cursor is at the beginning of a new word.

Pros:
- We will always trigger a request at the start of a word, that isn't cancelled unless a new word is started. That may help in scenarios where Cody would only need the first character, but we are up to 250ms away from triggering the request due to the user typing (and us cancelling previous requests)

Cons:
- More inflight requests, may not help as much - 250ms is already pretty low.

I implemented this here: https://github.com/sourcegraph/cody/pull/4751. We could consider expanding on this and running it as a small experiment after this, cc @valerybugakov

vscode/src/completions/smart-throttle.ts
```diff
@@ -0,0 +1,115 @@
...
+            // Case 3: Handle the latest request as the new tail request and require a small debounce
+            //         time before continuing.
+            this.tailRequest?.abort()
+            this.tailRequest = throttledRequest
+            const newRequestParams = throttledRequest.updatedRequestParams()
+
+            if (triggerKind === TriggerKind.Automatic) {
+                // If we have bypassed the throttle timeout, we don't need to wait for the
+                // debounce either.
+                const debounceTime = Math.min(now - this.lastThrottlePromotion, 25)
```
umpox, line 72: Small change so that, if we bypassed the `THROTTLE_TIMEOUT`, we additionally remove the 25ms debounce and execute immediately

vscode/src/completions/get-inline-completions.ts
```diff
@@ -326,11 +331,27 @@ async function doGetInlineCompletions(
...
-        triggerKind !== TriggerKind.Automatic
-            ? 0
-            : ((multiline ? debounceInterval?.multiLine : debounceInterval?.singleLine) ?? 0) +
-              (artificialDelay ?? 0)
+
+    if (smartThrottleService) {
+        // For the smart throttle to work correctly and preserve tail requests, we need full control
+        // over the cancellation logic for each request.
+        // Therefore we must stop listening for cancellation events originating from VS Code.
+        cancellationListener?.dispose()
```
umpox, line 339: If we reach the smart throttle stage, we `dispose` of the cancellationListener mentioned above.

The alternative would be the remove the `cancellationListener` entirely, but I think we want it if the request is cancelled before this point, and it isn't clear exactly how VS Code handles this.

For the purposes of the A/B test, this should be fine

vscode/src/completions/smart-throttle.ts
valerybugakov: >We could consider expanding on this and running it as a small experiment after this
I like this idea! It would be interesting to A/B test on a small percentage of users to gauge the increase in requests we send. Without the proper cancellation propagation, the load on our inference providers can also be significantly increased. Hopefully, by then, we will have some improvements on the cancelation front, which would make this experiment easier to conduct.

valerybugakov (approved):
I noticed that we have up to four inflight requests by the time I type `function ` in a document. Does it match your experience with local testing of this feature? Should we increase the throttle timeout? 

CleanShot 2024-07-03 at 15 20 44

I added the following log to the `request` method of the `RequestManager`, and with the current timeout, it "feels" that we send a request on every second character. 

```ts
console.log(
  `starting new request "${requestParams.docContext.currentLinePrefix}", total requests inflight: ${this.inflightRequests.size}`
)
```

PR #4737, opened 2024-07-01
Autocomplete: limit the number of hot-streak lines

valerybugakov:
- Restrict the number of lines we await during hot-streak completion generation to prevent overwhelming inference providers. Based on empirical observations, we'll initially set this limit to five. We can adjust this value later according to the A/B test results.
- Part of https://linear.app/sourcegraph/issue/CODY-2190/[autocomplete-latency]-ab-test-hot-streak

## Test plan

Updated unit tests and verified locally.

### DIFF
--- a/vscode/src/completions/providers/hot-streak.ts
+++ b/vscode/src/completions/providers/hot-streak.ts
@@ -77,7 +77,9 @@ function insertCompletionAndPressEnter(
 
 // To reduce the load on the inference providers we limit the hot streak
 // generation to a constant number of lines after the initial completion.
-const MAX_HOT_STREAK_LINES = 5
+const MAX_HOT_STREAK_LINES = vscode.workspace
+    .getConfiguration()
+    .get<number>('cody.experimental.maxHotStreakLines', 5)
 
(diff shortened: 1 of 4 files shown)


vscode/src/completions/providers/hot-streak.ts
```diff
@@ -75,8 +75,12 @@ function insertCompletionAndPressEnter(
     return updatedDocContext
 }
 
+// To reduce the load on the inference providers we limit the hot streak
+// generation to a constant number of lines after the initial completion.
+const MAX_HOT_STREAK_LINES = 5
```
olafurpg, line 80: Should this be somehow configurable via hidden setting by users? Might be useful when troubleshooting stuff

olafurpg (approved):

vscode/src/completions/providers/hot-streak.ts
valerybugakov: Will be useful. Added here: https://github.com/sourcegraph/cody/pull/4737/commits/7dee950594cc6398e5fad48ea36e69310c5c931c

PR #4738, opened 2024-07-01
biome: use git ignore file

keegancsmith:
I had a file that was killing biome inside of the .vscode-test-web directory. This ensures biome doesn't search inside of this directory. Additionally I enabled the vcs ignore integration. However, it seems to ignore nested .gitignore files, so I still needed the vscode-test-web rule.

Test Plan: pnpm biome doesn't make my fan spin on my desktop.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -75,6 +75,9 @@
       "__snapshots__/",
       "__mocks__/",
       ".vscode-test/",
+      ".vscode-test-web/",
+      ".test/",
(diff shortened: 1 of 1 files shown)


keegancsmith:
FYI to work out which files confused biome I would run `./node_modules/.bin/biome check --apply --log-level=debug .` to see which files it was spinning on.

RXminuS (approved):

PR #4739, opened 2024-07-01
Add trigger to overwrite multiline decision

eseliger:
When to use multiline completions is currently determined by some heuristic. It excludes certain languages, and at least for me, doesn't 100% get things right. That's okay, we can improve this over time.
But until we have a very reliable decision process for single vs multiline, it can be helpful to be able to trigger cody to always try to generate a multiline completion at a certain location.
That will also enable multiline completions for language currently not on the allow-list for auto-multiline (probably due to CAR concerns).

After this PR:

Option+\ will trigger any completion, with the detection on.
Option+Shift+\ will trigger a multiline completion always.

## Test plan

Tried generating multiline completions several times locally. There seems to be some bug in multiline in general that sometimes cuts off the completion too early, but that shouldn't prevent this change.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -95,6 +95,11 @@ export interface PreloadCompletionContext extends vscode.InlineCompletionContext
 export class InlineCompletionItemProvider
     implements vscode.InlineCompletionItemProvider, vscode.Disposable
 {
+    // This field is going to be set if you use the keyboard shortcut to manually trigger a
+    // multiline completion.
+    // VSCode does not allow to send additional parameters with suggestion triggers, so we
+    // need to keep this intent outside of the message.
+    public forceNextCompletionIsMultiline = false
     private latestCompletionRequest: CompletionRequest | null = null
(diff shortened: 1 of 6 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -319,6 +328,10 @@ export class InlineCompletionItemProvider
                 context: takeSuggestWidgetSelectionIntoAccount ? context : undefined,
             })
 
+            if (forceMultiline) {
+                docContext.multilineTrigger = 'manual'
```
eseliger, line 332: We use 'manual' here so in events data we can discern manual and auto-detected multiline suggestions.

valerybugakov (approved):
LGMT! Let's add a changelog entry `vscode/CHANGELOG.md`.

valerybugakov:
> There seems to be some bug in multiline in general that sometimes cuts off the completion too early, but that shouldn't prevent this change

It's probably not a bug but a `stopSequence` we use to stop generation which is `\n\n`. We must keep it for now because the cancellation propagation does not work reliably. We will overload our inference providers if we remove it today.

PR #4742, opened 2024-07-01
Chat: Update context preamble to minimize hedgings

abeatrix:
PART OF https://linear.app/sourcegraph/issue/CODY-2507/evaluate-claude-35-sonnet-as-new-default

Update the preamble that works across different models while minimizing hedgings.

## Test plan

Updated leaderboard locally with results from this change in https://github.com/sourcegraph/cody-leaderboard/pull/12

With the updated preamble, we can see it has decreased the hedging occurrences (having issues when running the evals so I removed the gemini models to get it to run quicker):

image

The one question that failed looks like a false positive to me:

image

### Before

image


lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -5,7 +5,7 @@ import { PromptString, ps } from './prompt-string'
 /**
  * The preamble we add to the start of the last human open-end chat message that has context items.
  */
-const CONTEXT_PREAMBLE = ps`You have access to the provided codebase context. Question: `
+const CONTEXT_PREAMBLE = ps`You have access to my codebase context. Answer positively without apologizing. Question: `
```
abeatrix, line 8: @jtibshirani @jamesmcnamara I've made this change, and it seems to work pretty well with hedging, but I noticed the LLM judge is scoring them as unhelpful. I would like to get your thoughts on this to determine if there is an issue with the preamble change, or should we update the LLM judge on how to score positive and negative responses?

abeatrix: actually please ignore this. I'm going to try another prompt and see if that helps!

abeatrix: It's ready!

abeatrix:
> apply it only to the Claude 3.5 models though

Ahh good point! Let me make that change, thanks for the suggestion @jtibshirani !

lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -5,7 +5,11 @@ import { PromptString, ps } from './prompt-string'
 /**
  * The preamble we add to the start of the last human open-end chat message that has context items.
  */
-const CONTEXT_PREAMBLE = ps`You have access to the provided codebase context. Question: `
+const CONTEXT_PREAMBLE = ps`You have access to my codebase context.`
```
jtibshirani, line 8: Can we keep this as before, because we tested that pretty thoroughly?
```suggestion
const CONTEXT_PREAMBLE = ps`You have access to the provided codebase context. `
```

Also I think we might be missing spaces between sentences.

lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -17,19 +21,25 @@ export class PromptMixin {
 
     /**
      * Prepends all mixins to `humanMessage`. Modifies and returns `humanMessage`.
+     * Add headges prevention prompt to known models.
```
jtibshirani, line 24:
```suggestion
     * Add hedging prevention prompt to specific models who need this.
```

lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -5,7 +5,11 @@ import { PromptString, ps } from './prompt-string'
 /**
  * The preamble we add to the start of the last human open-end chat message that has context items.
  */
-const CONTEXT_PREAMBLE = ps`You have access to the provided codebase context. Question: `
+const CONTEXT_PREAMBLE = ps`You have access to my codebase context.`
+/**
+ * The preamble for preventing known models from hedging.
+ */
+const HEDGES_PREVENTION = ps`Always provide a high-level overview answer positively without apologizing.`
```
jtibshirani, line 12: Was this better than the previous simpler one, like "Answer positively and don't apologize?" It's nice to keep it as simple as possible and focused on the specific problem (apologies/ hedging)

abeatrix: Yea this is better than the simple one based on the scoring!

jtibshirani (approved):
Looks good to me! 🚢

jamesmcnamara (commented):
Whoops! A little late to the game but this looks great!

PR #4747, opened 2024-07-02
Remove OpenAI embedding model support

janhartman:
## Test plan
N/A


lib/shared/src/codebase-context/context-status.ts
```diff
@@ -32,7 +32,7 @@ export interface RemoteSearchProvider {
     isIgnored: boolean
 }
 
-export type EmbeddingsProvider = 'sourcegraph' | 'openai'
```
janhartman, line 35: Decided not to delete this enum in case we add more providers later on.

vscode/src/local-context/local-embeddings.ts
```diff
@@ -98,15 +96,6 @@ const sourcegraphMetadataModelConfig: EmbeddingsModelConfig = {
     indexPath: getIndexLibraryPath('st-metadata'),
 }
 
-const openaiModelConfig: EmbeddingsModelConfig = {
-    model: 'openai/text-embedding-ada-002',
-    dimension: 1536,
-    provider: 'openai',
-    endpoint: CODY_GATEWAY_PROD_ENDPOINT,
-    // empty prefix to keep backwards compatibility
-    indexPath: getIndexLibraryPath(''),
```
rafax, line 107: Should we have some code that deletes the old library here @janhartman ?

rafax (approved):
One question about index file clean-up, other than that LGTM

janhartman:
@rafax should I update the changelog with this change? It shouldn't really affect users but still...

vscode/src/local-context/local-embeddings.ts
rafax: Not sure, but seeing that this is the "last clean-up", I think it would make sense to try?

rafax:
> @rafax should I update the changelog with this change? It shouldn't really affect users but still...

I don't think we need to do that (we're removing a feature that is not available).

vscode/src/local-context/local-embeddings.ts
janhartman: image
So this is the structure of the library... The files are from the OpenAI model (we don't use a prefix for backwards compatibility) while the dirs correspond to SG-provided models. So we have to delete only the files. Honestly, I think we can wait until we drop local embeddings entirely and then just nuke this whole dir.

rafax: Ok, in the meantime we can tell the users to delete everything (we will regenerate if needed)

PR #4748, opened 2024-07-02
Cody-bench: add chat context

janhartman:
Results are shown in https://github.com/sourcegraph/cody-leaderboard/pull/14.

If `context` is defined in the benchmark options, we index `context.sourcesDir` and retrieve context items from it. The indexing needs to be performed only once so it's done on startup and skipped for subsequent runs.

## Test plan
N/A


agent/src/cli/command-bench/strategy-chat.ts
```diff
@@ -34,7 +34,7 @@ export async function evaluateChatStrategy(
     const scores: LlmJudgeScore[] = []
     const model = ModelsService.getModelByIDSubstringOrError(chatModel).model
     const files = absoluteFiles.map(file => path.relative(options.workspace, file))
-    const yamlFiles = files.filter(file => file.endsWith('.yaml'))
+    const yamlFiles = files.filter(file => file.endsWith('.yaml') && file.startsWith('question'))
```
janhartman, line 37: This prevents us from using just any yaml file in the context sources dir as a question file.

janhartman:

This is a false positive, I just modified a related line :)

agent/src/cli/command-bench/EvaluationDocument.ts
```diff
@@ -154,8 +158,8 @@ export class EvaluationDocument {
                     out.push(' EXACT_MATCH')
                 }
 
-                if (item.resultExact) {
```
olafurpg, line 157: Although I don't think it's super important here, we should keep the old condition.

agent/src/cli/command-bench/EvaluationDocument.ts
```diff
@@ -154,8 +158,8 @@ export class EvaluationDocument {
...
-                    out.push(' EXACT_MATCH')
+                if (item.contextItems) {
+                    pushMultilineText('CONTEXT_ITEMS', item.contextItemsJSON ?? '[]')
```
olafurpg, line 162: This should make the generated output more readable. Having the JSON formatted in a single line makes it unreadable in the generated snapshots. If we're not using the snapshots for manual review then we should skip adding `CONTEXT_ITEMS` here (it's still to the leaderboard via the CSV file)
```suggestion
                    pushMultilineText('CONTEXT_ITEMS', JSON.stringify(JSON.parse(item.contextItemsJSON ?? '[]'), null, 2))
```

agent/src/cli/command-bench/command-bench.ts
```diff
@@ -363,7 +378,13 @@ async function evaluateWorkspace(options: CodyBenchOptions, recordingDirectory:
...
             customHeaders: {},
             customConfiguration: {
-                'cody.experimental.symf.enabled': false, // fixes errors in Polly.js related to fetchin the symf binary
+                'cody.experimental.symf.enabled': ['keyword', 'blended'].includes(
+                    options.context?.strategy
+                ), // disabling fixes errors in Polly.js related to fetching the symf binary
+                'cody.experimental.localEmbeddings.disabled': !['embeddings', 'blended'].includes(
+                    options.context?.strategy
+                ),
+                'cody.useContext': options.context?.strategy,
```
olafurpg, line 387: ?
```suggestion
                'cody.useContext': Boolean(options.context?.strategy),
```

agent/src/cli/command-bench/command-bench.ts
```diff
@@ -440,3 +464,47 @@ function expandWorkspaces(
...
+}
+
+async function indexContextSourcesDir(options: CodyBenchOptions): Promise<void> {
+    // If this is our first run, we need to index the context sources dir so symf & embeddings can retrieve results
+    // The agent has started symf by this point - we need to wait until the symf index has been created
+    // TODO: for embeddings, we don't have access to do it the same way
+
+    const symfIndex = path.join(
+        `${process.env.HOME}`,
+        'Library/Application Support/Cody-nodejs/symf/indexroot',
```
olafurpg, line 491: Better to use `codyPaths(...)` because this is going to crash on non-macOS

agent/src/cli/command-bench/command-bench.ts
```diff
@@ -440,3 +464,47 @@ function expandWorkspaces(
...
+        options.workspace
+    )
+
+    // Allow max 10 min for the index to be ready
+    const maxWaitTime = 10 * 60 * 1000
+    const sleepTime = 5000
+    let waitTime = 0
+    while (waitTime < maxWaitTime) {
```
olafurpg, line 499: This logic feels incredibly brittle and buggy. I feel sorry for the person who needs to debug an issue here in the future when the logic breaks. It would be cleaner to have a bash script that just runs `symf` to index the directory.

agent/src/cli/command-bench/strategy-chat.ts
olafurpg: nit: it would be fine to filter by `question.yaml` instead of separate condition (I think)

olafurpg (approved):
mostly minor comments, lgtm 👍🏻 excited to see more data added to the leaderboard! 😄

agent/src/cli/command-bench/command-bench.ts
janhartman: I agree, this code is crappy. I did it this way to reuse the logic of running symf when the agent starts as the logic for running it by itself isn't completely trivial either. But I can take a stab of running it from a shell script.

agent/src/cli/command-bench/command-bench.ts
janhartman: Good point.

agent/src/cli/command-bench/command-bench.ts
janhartman: This is actually an enum:
```typescript
export type ConfigurationUseContext = 'embeddings' | 'keyword' | 'none' | 'blended' | 'unified'
```

agent/src/cli/command-bench/EvaluationDocument.ts
janhartman: I thought it was a bug as this check is already done above 🤷

agent/src/cli/command-bench/EvaluationDocument.ts
janhartman: Hm, with your suggestion we're performing a redundant stringify + parse step (`item.contextItemsJSON` is already stringified). Can we go without them?

olafurpg: If `item.contextItems` is available just stringify that. Performance is not a problem here though

olafurpg: parsing and serializing json is cheap (most of the time), and this isn't a performance sensitive path anyways

agent/src/cli/command-bench/command-bench.ts
janhartman: @olafurpg one problem - I don't have easy access to the symf binary path. I can get it from the `getSymfPath` function, but that takes `vscode.ExtensionContext` as a param, which is only available on startup from the `activate` call AFAIK. And unfortunately we do need access to the context as we fetch `globalStorageUri` from it. Is there a way to get around this that doesn't involve hardcoding paths?

olafurpg: `globalStorageUri` in the agent is computed from `codyPaths()`. You can just hardcode the path like that instead of hardcoding the macOS path

olafurpg: Aah, if you mean to download the binary then you can just write a script that downloads it and indexes. I'm sure cody can write the bash/python script in one shot

janhartman: At this point I'm actually not sure if it's better to have a fragile shell script that's yet another thing to maintain vs some crappy code with sleeps in a loop 😓

olafurpg: Another alternative is to use the JSON-RPC protocol to await on a notification for when the index is ready. I recall seeing @hitesh-1997 write some TS code that implements this. I think you can invoke the command with something like this
```ts
await client.request('command/execute', { command: 'cody.search.index-update' })
```
Feel free to ignore. But I want to encourage us to write benchmark code in a way that is maintainable. Somebody else (could be future you) is going to be maintaining this code.

janhartman: Wow that's a neat solution, seems much less brittle!

jtibshirani:
@janhartman a couple questions:
* Are we only covering PLG context retrieval (symf plus embeddings) here?
* How are you thinking of keeping this in sync with https://github.com/sourcegraph/context-eval, which is the "ground truth" for chat context eval?

janhartman:
@jtibshirani I tested PLG retrievers here as this is my primary focus right now, but we should totally add enterprise context instead. We have some plans at the hackathon this week that might enable enterprise context for OSS repos on dotcom as well :)
As for your second point, no plans so far. 
I am however putting together a new eval framework that should be a merge of context-eval and Hitesh's framework (i.e. it will support all our retrievers) and more metrics than just recall. Hopefully, with the move to server-side context, the new framework will be all we need to do context evals and maybe the functionality here will be irrelevant.

PR #4750, opened 2024-07-02
CLI: replace keytar with custom solution

olafurpg:
Previously, we used Keytar to write/read/delete secrets. This was problematic for two reasons:

- Keytar has been unmaintained since 2022
- Keytar required native Node modules that meant `cody-agent` was crashing on macOS when using a release that was built on Linux (in CI)

This PR addresses both problems by replacing keytar with a custom solution that shells out to different secret managers (`security` on macOS, `secret-tool` on Linux, and `powershell` on
 Windows). See comment in `secrets.ts` for a more detailed reasoning why
we went with this approach.

## Test plan

Manually tested on macOS/Linux/Windows.

### DIFF
--- a/agent/src/cli/auth/secrets.ts
+++ b/agent/src/cli/auth/secrets.ts
@@ -1,42 +1,162 @@
-import keytar from 'keytar'
+import { exec } from 'node:child_process'
+import { promisify } from 'node:util'
 import { logDebug } from '../../../../vscode/src/log'
 import type { Account } from './settings'
 
 // This file deals with reading/writing/removing Cody access tokens from the
 // operating system's secret storage (Keychain on macOS, Credential Value on
-// Windows, etc.).
+// Windows, etc.). Originally, we used the `keytar` npm dependency to interact
+// with the OS secret storage. However, Keytar is unmaintained and it was
+// complicated to distribute anyways because you had to distribute native
+// modules for each supported OS. The current implementation shells out to the
+// `security` command on macOS, `powershell` on Windows, and `secret-tool` on
+// Linux. Users can always set the `CODY_ACCESS_TOKEN` environment variable if
+// they don't want to use this functionality.
+//
+// The biggest problem with this approach is that users will most likely select
+// "Always allow" when prompted if the "system" tool can access the Cody
+// secrets. This means that any other tool on the computer can shell out to
+// `system` to read the same secret. However, we chose to go with this approach
+// regardless of this risk based on the following observations:
+// - The user can chose not to let Cody manage its secrets. This is an optional feature.
+// - Storing the secret as a global environment variable also isn't secure.
+// - The `gh` cli tool uses the same approach (shelling out to `security` on macOS),
+//   meaning that any tool on my computer can read my GitHub access token by shelling out
+//   to `security` without me knowing.
+// - It's marginally more secure to build a native module instead of using
+//   `system` because a malicious user can also load the native module instead
+//   of shelling out to `system` to fake that it's Cody cli.
 
-const codyServiceName = 'Cody'
-
-function keytarServiceName(account: Account): string {
-    const host = new URL(account.serverEndpoint).host
-    return `${account.id} on ${host}`
-}
 export async function writeCodySecret(account: Account, secret: string): Promise<void> {
+    const keychain = getKeychainOperations(account)
     try {
-        await keytar.setPassword(codyServiceName, keytarServiceName(account), secret)
+        await keychain.writeSecret(secret)
     } catch (error) {
-        logDebug('keytar-storage', 'Error storing secret:', error)
+        logDebug('keychain-storage', 'Error storing secret:', error)
     }
 }
 
 export async function readCodySecret(account: Account) {
+    const keychain = getKeychainOperations(account)
     try {
-        const secret = await keytar.getPassword(codyServiceName, keytarServiceName(account))
+        const secret = await keychain.readSecret()
         if (secret) {
             return secret
         }
         return null
     } catch (error) {
-        logDebug('keytar-storage', 'Error retrieving secret:', error)
+        logDebug('keychain-storage', 'Error retrieving secret:', error)
         return null
     }
 }
 
-export async function removeCodySecret(account: string) {
+export async function removeCodySecret(account: Account) {
+    const keychain = getKeychainOperations(account)
     try {
-        await keytar.deletePassword(codyServiceName, account)
+        await keychain.deleteSecret()
     } catch (error) {
-        logDebug('keytar-storage', 'Error deleting secret:', error)
+        logDebug('keychain-storage', 'Error deleting secret:', error)
+    }
+}
+
+const execAsync = promisify(exec)
+function getKeychainOperations(account: Account): KeychainOperations {
+    switch (process.platform) {
+        case 'darwin':
+            return new MacOSKeychain(account)
+        case 'win32':
+            return new WindowsCredentialManager(account)
+        case 'linux':
+            return new LinuxSecretService(account)
+        default:
+            throw new Error(`Unsupported platform: ${process.platform}`)
+    }
+}
+
+/**
+ * Uses each operating system's native keychain to store the Cody access token.
+ * - `security` on macOS
+ * - `powershell -command '...StoredCredential ...` on Windows
+ * - `secret-tool` on Linux
+ */
+abstract class KeychainOperations {
+    constructor(public account: Account) {}
+    protected service(): string {
+        const host = new URL(this.account.serverEndpoint).host
+        return `Cody: ${host} (${this.account.id})`
+    }
+    abstract readSecret(): Promise<string>
+    abstract writeSecret(secret: string): Promise<void>
+    abstract deleteSecret(): Promise<void>
+}
+
+class MacOSKeychain extends KeychainOperations {
+    async readSecret(): Promise<string> {
+        const { stdout } = await execAsync(
+            `security find-generic-password -s "${this.service()}" -a "${this.account.username}" -w`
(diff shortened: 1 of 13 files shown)


olafurpg:
TODO: report helpful error messages when missing dependencies

Linux:
```
sudo apt install libsecret-tools
sudo apt install gnome-keyring
```

Windows with Admin Powershell
```
Install-Module -Name CredentialManager
```

agent/src/cli/auth/secrets.ts
```diff
@@ -1,42 +1,165 @@
...
+    }
+    abstract readSecret(): Promise<string>
+    abstract writeSecret(secret: string): Promise<void>
+    abstract deleteSecret(): Promise<void>
+}
+
+class MacOSKeychain extends KeychainOperations {
+    async readSecret(): Promise<string> {
+        const { stdout } = await execAsync(
+            `security find-generic-password -s "${this.service()}" -a "${this.account.username}" -w`
```
eseliger, line 96: this seems prone for command injection, is there some safe shellquoting package on NPM like Gos shellquote?

agent/src/cli/auth/secrets.ts
```diff
@@ -1,42 +1,165 @@
...
+    async readSecret(): Promise<string> {
+        const { stdout } = await execAsync(
+            `secret-tool lookup service '${this.service()}' account '${this.account.username}'`
+        )
+        return stdout.trim()
+    }
+
+    async writeSecret(secret: string): Promise<void> {
+        await execAsync(
+            `echo "${secret}" | secret-tool store --label="${this.service()}" service '${this.service()}' account ${
```
eseliger, line 154: Is `secret-tool` expected to exist in most distros? I assume server-environments (VSCode remote SSH for example) won't have this?

agent/src/cli/auth/secrets.ts
keegancsmith: you could use the spawn APIs where you pass in the args array

agent/src/cli/auth/secrets.ts
keegancsmith: secret-tool isn't usually installed, but the libsecret stuff via dbus is normally. I think as long as we are graceful here it should be fine. Personally for me I prefer not having tools like this, and rather manage it myself via envvars on my linux desktop. EG see this problem I had trying out cody-emacs where the agent wanted to store secrets, even though cody-emacs was responsible for the secret. https://github.com/sourcegraph/emacs-cody/pull/30#issuecomment-2204243673

agent/src/cli/auth/secrets.ts
steveyegge: @olafurpg This might be worth fixing, since we have security-sensitive customers, and I don't think this is a PLG-only feature.

steveyegge (commented):
Quoting the shell args somehow does seem like a good thing to add. I won't stand in the way of merging, though, since keytar needs to die die die.

agent/src/cli/auth/secrets.ts
olafurpg: I was going to follow up separately to report helpful error messages if the dependencies are missing https://github.com/sourcegraph/cody/pull/4750#issuecomment-2203136894

olafurpg: @keegancsmith `auth` is an optional feature, you can still use environment variables or explicit --access-token to manage secrets. From personal experience, I find it incredibly convenient when tools like `gh` just work consistently on macOS/Windows/Linux regardless of how or where you started the terminal. Storing secrets as environment variables has lots of problems. For example, any program you run has access to SRC_ACCESS_TOKEN without doing any additional work. With a secret manager, you need to at least know the server endpoint and username to read the correct secret (which you can do by reading the plaintext user-settings.json file, but it's still a bit more effort).

olafurpg: My primary motivation for all this effort is to increase onboarding and retention. I want the cli tool to be super easy to get started with, and my personal experience is that I stick most easily with cli tools that manage secrets for me (example: `gh`)

olafurpg:
Thank you for the review! I've addressed the feedback on shell quoting, the solution is now using `spawn` instead of `exec`. I will submit a request in #discuss-security to additionally review this functionality after the release is out (we won't make public noise about this feature until after security review is done)

olafurpg:
CI failures are unrelated, posted about it here https://sourcegraph.slack.com/archives/C05AGQYD528/p1720014583901319

agent/src/cli/auth/secrets.ts
```diff
@@ -1,42 +1,282 @@
...
+                }
+            })
+        })
+    }
+}
+
+class MacOSKeychain extends KeychainOperations {
+    installationInstructions = ''
+    async readSecret(): Promise<string> {
+        return await this.spawnAsync('security', [
```
eseliger, line 129: much nicer than a shell script, thanks!

agent/src/cli/auth/secrets.ts
```diff
@@ -1,42 +1,282 @@
...
+class WindowsCredentialManager extends KeychainOperations {
+    installationInstructions = `The 'CredentialManager' PowerShell module needs to be installed to let Cody manage your access token.
+To fix this problem, run the command below to install the missing dependencies:
+  Install-Module -Name CredentialManager
+${alternativelyMessage}`
+    private target(): string {
+        return `${this.service()}:${this.account.username}`
+    }
+    async readSecret(): Promise<string> {
+        const powershellCommand = `(Get-StoredCredential -Target "${this.target()}").GetNetworkCredential().Password`
```
eseliger, line 173: this is prone to injection I believe, as it's using string concatenation again

agent/src/cli/auth/secrets.ts
```diff
@@ -1,42 +1,282 @@
...
+    availableCommands.set(command, isInstalled)
+    if (!isInstalled) {
+        spinner.fail(installationInstructions)
+    }
+    return isInstalled
+}
+function canSpawnCommand(command: string) {
     try {
-        await keytar.deletePassword(codyServiceName, account)
+        execSync(`which ${command}`, { stdio: 'ignore' })
```
eseliger, line 277: Non-blocking but might be nice to reduce the dependency on a shell that has which - https://github.com/otiai10/lookpath/blob/main/src/index.ts seems to be an implementation of Gos exec.LookPath for node.

eseliger (commented):
One remaining question about command injection in the powershell scripts, otherwise LGTM

agent/src/cli/auth/secrets.ts
```diff
@@ -1,42 +1,282 @@
...
 // This file deals with reading/writing/removing Cody access tokens from the
 // operating system's secret storage (Keychain on macOS, Credential Value on
-// Windows, etc.).
+// Windows, etc.). Originally, we used the `keytar` npm dependency to interact
+// with the OS secret storage. However, Keytar is unmaintained and it was
+// complicated to distribute anyways because you had to distribute native
+// modules for each supported OS. The current implementation shells out to the
+// `security` command on macOS, `CredentialManager` on Windows, and `secret-tool` on
+// Linux. This is an o)ptional feature. Users can always avoid using this functionality
+// by setting the `CODY_ACCESS_TOKEN` environment variable or passing the `--access-token`
```
pkukielka, line 14: Cody clients should also be able to pass the token as part of the `initialize` call.
It was semi-broken though, I was fixing it today by the way of looking tat the telemetry issue:
https://github.com/sourcegraph/cody/pull/4766/files#diff-85c89382076aeb49c8645dd41e6422552fba7bbf25575701c1a9253027164a76

pkukielka (approved):
LGTM

agent/src/cli/auth/secrets.ts
olafurpg: I thought about it, but we only do this on Linux so it's not a problem. On macOS, `security` is pre-installed and on Windows the user needs to install `CredentialManager` with the pre-installed `powershell`.

agent/src/cli/auth/secrets.ts
olafurpg: NICE! I've had a gut feeling we are doing some broken things like that. Glad we are fixing it <3

agent/src/cli/auth/secrets.ts
olafurpg: Thank you for raising the concern! This is much cleaner and easier to reason about and maintain

olafurpg: I would probably have used `spawn` in the first iteration, it was using `exec` because that's what Cody generated on the first show 😅

agent/src/cli/auth/secrets.ts
olafurpg: I pushed a change to escape all `"` in the target. This is a low-risk injection vector because the target is a combination of the server endpoint and username. Still, worth escaping out of principle

olafurpg:
Rebase on top of main to fix CI https://github.com/sourcegraph/cody/pull/4768

olafurpg:
Thank you everybody for the review! 🙇🏻 I will cut a release and get a second round of review from the security team [CODY-2738 : Complete security review on Cody cli `auth` secret management](https://linear.app/sourcegraph/issue/CODY-2738/complete-security-review-on-cody-cli-auth-secret-management)

agent/src/cli/auth/secrets.ts
keegancsmith: I 100% agree, these tools should have a slick way to manage this stuff. But for example in automation and people who like to make like difficult like myself its nice to be able to have escape hatches. EG for emacs-cody I'd rather have auth be handled in the normal emacs way (if configured).

PR #4754, opened 2024-07-02
agent: added a simple cli to get the tsc autocomplete context

hitesh-1997:
## Context
Adding a simple cli to get the tsc compiler api context. We can specify a configuration file having filePath and line number and trigger the tsc context retriver.
This would be useful to gather the context data for training and evaluation purpose.

## Usages
```
pnpm agent internal tsc-context-retriever \
            --src-access-token <token> \
            --src-endpoint https://sourcegraph.com \
            --config-path /Users/hiteshsagtani/fixtures/config.json  \
            --save-dir /Users/hiteshsagtani/fixtures/output
```

## Test plan
1. Manully checking the generated outputs via cli.


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -198,6 +198,10 @@ export type ClientRequests = {
     // Returns diagnostics for the given URI. Lives under `testing/` instead of
     // standalone `diagnostics/` because it only works for TypeScript files.
     'testing/diagnostics': [{ uri: string }, { diagnostics: ProtocolDiagnostic[] }]
+    'autocomplete/tscContextRetriever': [
```
olafurpg, line 201: Just to be clear this is intended for testing purposes
```suggestion
    'testing/tsc/retrieveContext': [
```

agent/src/cli/root.ts
```diff
@@ -18,3 +19,4 @@ export const rootCommand = new Command()
     .addCommand(codyBenchCommand)
     .addCommand(chatCommand())
     .addCommand(authCommand())
+    .addCommand(tscContextRetrieverCommand())
```
olafurpg, line 22: Let's create an `internal` subcommand that groups internal/testing features like this one. We may move the `bench` command to there as well.
```suggestion
    .addCommand(new Command('internal').addCommand(tscContextRetrieverCommand()))
```
I'm working on making the cli a useful tool for end-users so I want to keep the toplevel commands clean. Users should be able to run `cody help` and see a meaningful list of commands.

olafurpg (approved):
I'm glad to see activity in making the cli tool richer! Only a few blocking comments on naming convention, otherwise LGTM. I highly encourage experimentation like this and want there to be a low barrier to add helpers like these.

vscode/src/jsonrpc/agent-protocol.ts
hitesh-1997: done

agent/src/cli/root.ts
hitesh-1997: done

hitesh-1997:
> I'm glad to see activity in making the cli tool richer! Only a few blocking comments on naming convention, otherwise LGTM. I highly encourage experimentation like this and want there to be a low barrier to add helpers like these.

Thanks @olafurpg 
I added a Semaphore in the package.json for agent to limit the number of concurrent request, can you please check that once.

Meanwhile I am looking into what is causing the security issue here: https://github.com/sourcegraph/cody/pull/4754#issuecomment-2206083330

agent/src/cli/tscContextRetriever.ts
```diff
@@ -0,0 +1,157 @@
...
+            customHeaders: {},
+            customConfiguration: {
+                'cody.experimental.telemetry.enabled': false,
+                'cody.telemetry.level': 'off',
+            },
+        },
+        inheritStderr: true,
+    })
+    const concurrencyLimit = 2
+    const semaphore = new Semaphore(concurrencyLimit)
```
olafurpg, line 102: I don't understand the motivation for a semaphore here. JS is single-threaded by nature, and the parallel clients start separate processes anyways. Can you avoid this dependency by doing a basic for loop? Given this is an internal subcommand, I prefer not to add additional dependencies if we can avoid it
```
        for await (const x of [1, 2, 3]) {
            await new Promise(resolve => setTimeout(resolve, 1000))
        }
```

hitesh-1997: Sure @olafurpg 
That should be okay, I removed the semaphore dependency and used a simple loop

olafurpg:
Heads up, I merged a PR today that adds a new `internal` subcommand where `bench` lives. This command should live there as well

PR #4756, opened 2024-07-02
[Cody Web]: Allow agent auth running without access token

vovakulikov:

This is one of the PRs that come from the main change PR that I made for Cody Web here: https://github.com/sourcegraph/cody/pull/4605. 

This PR introduces the `cody.web` configuration field that allows running the cody agent 
- Without provided access token 
- Always fetching remote context (not only for sourcegraph.com instance)

This change is safe to merge because it affects only Cody Web package and has no impact 
on any other agent clients

## Test plan
- CI is passing

### DIFF
--- a/vscode/src/services/AuthProvider.ts
+++ b/vscode/src/services/AuthProvider.ts
@@ -227,12 +227,23 @@ export class AuthProvider implements AuthStatusProvider {
     ): Promise<AuthStatus> {
         const endpoint = config.serverEndpoint
         const token = config.accessToken
+        const isCodyWeb = vscode.workspace.getConfiguration().get<boolean>('cody.web', false)
+
         if (isOfflineMode) {
             const lastUser = localStorage.getLastStoredUser()
             return { ...offlineModeAuthStatus, ...lastUser }
         }
-        if (!token || !endpoint) {
-            return { ...defaultAuthStatus, endpoint }
+
+        // Cody Web can work without access token since authorization flow
+        // relies on cookie authentication
+        if (isCodyWeb) {
+            if (!endpoint) {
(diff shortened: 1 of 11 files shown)


vscode/src/services/AuthProvider.ts
```diff
@@ -227,12 +227,23 @@ export class AuthProvider implements AuthStatusProvider {
     ): Promise<AuthStatus> {
         const endpoint = config.serverEndpoint
         const token = config.accessToken
+        const isCodyWeb = vscode.workspace.getConfiguration().get<boolean>('cody.web', false)
```
abeatrix, line 230: not a blocker, but maybe we can centralize this logic using [config.agentIDE](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/configuration.ts?L83) and add Web to [CodyIDE](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/configuration.ts?L97), wdyt?

Example: we check the agentIDE value to see if we should [display version update notice](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/webviews/App.tsx?L247) in the webview

abeatrix (approved):

vscode/src/services/AuthProvider.ts
vovakulikov: Good suggestion, I've seen this field before but didn't want to change the core logic there. But it looks like it's time to add a Web client there. I did this in the last commit here e28fe780c4d86f9415309c587380a1bf856f7d35

abeatrix: The new commit looks good to me, thanks for updating them!

PR #4757, opened 2024-07-02
Cody Web: fix cody web UI flashes problem

vovakulikov:
Based on https://github.com/sourcegraph/cody/pull/4756
Closes [SRCH-632](https://linear.app/sourcegraph/issue/SRCH-632/merge-cody-web-experimental-package-to-the-cody-repo-main-branch)

After this change we will be able to publish Cody Web package from main again.

This PR fixes UI-flash problems when we switch chats or set the initial chat transcript. 
Previously, the CodyWebChat component didn't wait for any restoring event and render
all available transcripts right away, this caused problems when you opened the chat and 
could see how we render empty chat only then render actual transcript UI. 
 
This PR makes CodyWebChat wait for the actual current chat restoration and fixes the problem 
with flashing. 
 
**Tech note:** I'm not happy with the current approach with refs and panel and chat ids states but for now it seems as a reasonable fix. Later I will refactor the provider state in a way that cody web chat component  wouldn't subscribe in vscode API 

## Test plan
- CI is green 
- Check that when you open the cody web demo (`cd ./web` and `pnpm dev`) you should see
no flashes while we set the initial chat (the last visited chat)

### DIFF
--- a/web/lib/components/CodyWebChat.tsx
+++ b/web/lib/components/CodyWebChat.tsx
@@ -1,4 +1,4 @@
-import { type FC, useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { type FC, useCallback, useLayoutEffect, useMemo, useRef, useState } from 'react'
 import { URI } from 'vscode-uri'
 
 import {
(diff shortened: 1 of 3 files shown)


vscode/src/services/AuthProvider.ts
```diff
@@ -227,12 +227,23 @@ export class AuthProvider implements AuthStatusProvider {
...
+        // Cody Web can work without access token since authorization flow
+        // relies on cookie authentication
+        if (isCodyWeb) {
+            if (!endpoint) {
+                return { ...defaultAuthStatus, endpoint }
+            }
+        } else {
+            if (!token || !endpoint) {
+                return { ...defaultAuthStatus, endpoint }
+            }
```
fkling, line 246: Maybe this should be a separate PR if it's not related to the UI problems. But I don't have strong feelings about it.

fkling (approved):

vscode/src/services/AuthProvider.ts
vovakulikov: It's in fact a separate PR, I just need to rebase it again, sorry 

https://github.com/sourcegraph/cody/pull/4756

PR #4758, opened 2024-07-02
Re-enable chat in sidebar by default

beyang:
*Remaining TODO:*
- [x] Remove chat history panel (favor chat history button)
- [x] Remove commands panel

Re-enable chat in the sidebar and make this the default `New Chat` mode:

https://github.com/sourcegraph/cody/assets/1646931/9cd79846-ad2d-46f6-923e-30121180617d

## Non-functional changes:

Change VS Code command names to be more correct/consistent:

* cody.chat.panel.new -> cody.chat.newEditorPanel
* cody.chat.panel.sidebar.new -> cody.chat.newPanel
* cody.chat.panel.moveFromSidebarToEditor -> cody.chat.moveToEditor
* cody.chat.panel.moveFromEditorToSidebar -> cody.chat.moveFromEditor
* Remove cody.chat.focus as an explicitly defined command, since we
  want a different title depending on whether the user is auth'd or
  unauth'd. Replace this with cody.chat.signIn and
  cody.chat.focusView, both of which just invoke cody.chat.focus,
  which is an implicitly defined command that focuses the cody.chat
  view

Avoid using "sidebar", because the chat view can be dragged to the bottom panel.

## Test plan

- [ ] Pull down this branch
- [ ] Execute the `Cody: New Chat` command from the command palette
- [ ] Observe it creates the new chat in the sidebar. The sidebar should open if it's not yet open
- [ ] Move the chat from the sidebar into an editor panel using the icon button in the panel header
- [ ] Move the chat back into the sidebar from the editor using the icon button in the editor panel header
- [ ] Execute the `Cody: New Chat in Editor` command and verify this still opens a new chat view in an editor panel


sqs (approved):
tried it out, couple feedbacks in person

abeatrix (commented):
I tried re-authing and switching accounts, both work as expected.

Notice the following behaviors but not sure if they are expected or not:

- When running a command, a new chat panel is opened instead of the sidebar chat view.
- Clicking on chat history item opens a new chat panel instead of the sidebar chat


potential bug: It looks like the sidebar chat webview cannot read the current model info when it first loaded, but selecting a new model from the dropdown will refresh it:
image



Should we set the chat history and settings tree view collapsed by default?

image

vscode/src/chat/chat-view/ChatManager.ts
```diff
@@ -70,22 +70,30 @@ export class ChatManager implements vscode.Disposable {
...
+                await this.chatPanelsManager.resetSidebar()
+                await vscode.commands.executeCommand('cody.chat.focus')
+            }),
+            vscode.commands.registerCommand('cody.chat.newEditorPanel', () =>
+                this.createNewWebviewPanel()
             ),
             vscode.commands.registerCommand('cody.chat.history.export', () => this.exportHistory()),
             vscode.commands.registerCommand('cody.chat.history.clear', () => this.clearHistory()),
             vscode.commands.registerCommand('cody.chat.history.delete', item => this.clearHistory(item)),
-            vscode.commands.registerCommand('cody.chat.panel.new', () => this.createNewWebviewPanel()),
```
abeatrix, line 88: we will also need to update the agent commands to match the new command here: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/agent.ts?L1058-1067

PR #4759, opened 2024-07-03
update hedges prevention preamble in prompt-mixin

abeatrix:
Simplify the preamble for preventing hedging and verified with the updated data in: https://github.com/sourcegraph/cody-leaderboard/pull/12/commits/88eb291906c6b602e999fa5dcf0d0bd85e96002b

## Test plan

image

### DIFF
--- a/lib/shared/src/prompt/prompt-mixin.ts
+++ b/lib/shared/src/prompt/prompt-mixin.ts
@@ -9,7 +9,7 @@ const CONTEXT_PREAMBLE = ps`You have access to the provided codebase context. `
 /**
  * The preamble for preventing known models from hedging.
  */
-const HEDGES_PREVENTION = ps`Provide a high-level overview answer positively without apologizing. `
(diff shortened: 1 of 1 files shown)


lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -9,7 +9,7 @@ const CONTEXT_PREAMBLE = ps`You have access to the provided codebase context. `
 /**
  * The preamble for preventing known models from hedging.
  */
-const HEDGES_PREVENTION = ps`Provide a high-level overview answer positively without apologizing. `
+const HEDGES_PREVENTION = ps`Answer positively without apologizing. `
```
abeatrix, line 12: @jtibshirani not sure why it was scoring lower when i retried with this version, but when I just retried it again it looks fine, and i've pushed the updated test to https://github.com/sourcegraph/cody-leaderboard/pull/12

jtibshirani (approved):

PR #4760, opened 2024-07-03
comment out debug output

peterguy:
Just commented out some debug output. Not sure if it should be kept around.

## Test plan

N/A

### DIFF
--- a/web/lib/components/CodyWebChat.tsx
+++ b/web/lib/components/CodyWebChat.tsx
@@ -177,7 +177,7 @@ export const CodyWebChat: FC<CodyWebChatProps> = props => {
         }
     }, [initialContext])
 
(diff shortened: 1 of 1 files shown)


web/lib/components/CodyWebChat.tsx
```diff
@@ -177,7 +177,7 @@ export const CodyWebChat: FC<CodyWebChatProps> = props => {
         }
     }, [initialContext])
 
-    console.log({ client, userAccountInfo, chatModels, activeChatID })
+    // console.log({ client, userAccountInfo, chatModels, activeChatID })
```
dominiccooney, line 180:
```suggestion
```

dominiccooney (approved):
Let's simply delete the debug scree.

PR #4761, opened 2024-07-03
Use fuzzysort for remote files

peterguy:
In @vovakulikov 's demo of Cody Web today, I noticed that he @-mentioned `insight.go`, which didn't match any files. He corrected it to `insights.go`, which did match a file, and continued the demo.

Seeing that made me think that we should add the ability to more fuzzily match remote files, similar to how we match local files (@-mentioning `chatcontext` will match `.../chat/.../context...`, `...chatContext...`, `...chat-context...`, and perhaps more)

This PR attempts to add fuzziness to the file search by expanding the path into `OR` elements in the GQL query and querying for 300 results rather than 30. Those 300 are fed through `fuzzysorter`, comparing them to the entered path and returning the top 30.

Alternate approach to #4755.

Neither approach completely recreates the local file experience because it can't apply `fuzzysort` to all of the files. From experimentation, it looks like `fuzzysort` uses a trie or similar to match paths containing parts of the entered file name. For example, mentioning `chatcontext` surfaces paths containing either `chat` or `context` or `chatcontext` because `fuzzysort` is able to apply its heuristics to all of the local files.

But either approach does help with some misspellings.

## Test plan

run locally (`cde web && pnpm install && pnpm dev`) and experiment with @-mentions. I've been using `internal/databse/insight` (purposeful misspelling), adding `.go` also.
Use your creativity to come up with different options.

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -1,3 +1,5 @@
+import * as path from 'node:path'
+import fuzzysort from 'fuzzysort'
 import type { Response as NodeResponse } from 'node-fetch'
 import { URI } from 'vscode-uri'
 import { fetch } from '../../fetch'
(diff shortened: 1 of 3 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
     ): Promise<FuzzyFindFile[] | Error> {
-        return this.fetchSourcegraphAPI<APIResponse<FuzzyFindFilesResponse>>(FUZZY_FILES_QUERY, {
-            query: `type:path count:30 ${
-                repositories.length > 0 ? `repo:^(${repositories.map(escapeRegExp).join('|')})$` : ''
-            } ${query}`,
-        }).then(response =>
-            extractDataOrError(
-                response,
-                data => data.search?.results.results ?? new Error('no files found')
+        // as far as I can tell, `query` will only ever contain one file name and nothing else
```
dominiccooney, line 520: This comment adds more doubt than clarity, is it pulling its weight?

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
+                    data => data.search?.results.results ?? new Error('no files found')
+                )
             )
-        )
+        }
+
+        // Expand the path into parts combined with the OR operator
+        // so as to return as many possible results as possible.
+        // Experimentally, simply separating the path parts performed
+        // as well as expaning using regular expressions and wildcard,
```
dominiccooney, line 541:
```suggestion
        // as well as expanding using regular expressions and wildcard,
```

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
+        // and didn't take as long.
+        const expandQuery = (query: string): string => {
+            const { dir, name, ext } = path.parse(query)
+            const parts = [dir !== '.' ? dir.replace('/', ' OR ') : '', name, ext].filter(Boolean)
```
dominiccooney, line 545: Please consider Windows.

I'm not familiar with this enough to know whether we have canonicalized to `/` at this point, but maybe this is wrong if users are typing Windows-style path separators.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
+
+            return `(${parts.join(' OR ')})`.replace(/\./g, '\\.')
```
dominiccooney, line 547: You should probably escape all regex metacharacters and not just `.`?

You can probably write 545-547 more neatly to avoid the duplicated ` OR ` by splitting on the path separator and then using one join at the end.

Remember you can `[...dirParts, name, ext]` in TypeScript

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
+            }).then(response =>
+                extractDataOrError(
+                    response,
+                    data => data.search?.results.results ?? new Error('no files found')
+                )
             )
-        )
+        }
+
+        // Expand the path into parts combined with the OR operator
```
dominiccooney, line 538: Unsure whether the commentary about paths not taken are worth it. It is mildly confusing whether you're talking about splitting this with a regular expression (on reflection, you're not, you're talking about using a regular expression in the search).

I think the operative thing at this point is:

- We do something simple that returns too many results.
- That's OK because we're going to fuzzysort them to filter them back down. So we want a lot of results.

This begs the question to me of how ranking a bunch of OR queries works. Like if I'm searching for foo.c, you're going to find all the files with `c` in the name... which could be a lot...

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
+            const parts = [dir !== '.' ? dir.replace('/', ' OR ') : '', name, ext].filter(Boolean)
+
+            return `(${parts.join(' OR ')})`.replace(/\./g, '\\.')
+        }
+
+        const expandedQuery = expandQuery(query)
+
+        // Query ten times as many results as we want to display
+        // so that fuzzysort will have something to work with.
+        // Not sure whhy 30 is the magic number to display.
```
dominiccooney, line 554:
```suggestion
        // Not sure why 30 is the magic number to display.
```

Again, find out and comment definitively, or omit this and let someone else find out if they care. But I don't think this comment is pulling its weight.

If you have a stylistic preference about coding or documentation we're failing here get on Slack and let's talk about it as a team.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -515,16 +517,59 @@ export class SourcegraphGraphQLAPIClient {
...
+        let result = await getMatches(repositories, expandedQuery, 300)
+
+        // Don't sort if the results don't have any file matches.
+        // Don't attempt a second, broader query at this time.
+        if (result instanceof Error || (Array.isArray(result) && result.length > 0)) {
+            return result
+        }
+
+        // fiter the results using the same fuzzysort as the VSCode local files uses
```
dominiccooney, line 563:
```suggestion
        // Filter the results using the same fuzzysort as the VSCode local files uses
```

dominiccooney (changes requested):
Realize this is a draft but some tentative feedback inline.

PR #4766, opened 2024-07-03
Fix telemetry event cody.auth.login:firstEver firing too often in agent

pkukielka:
Closes CODY-2346

## Test plan

1. Start JetBrains extension with this branch using `CODY_DIR` param
2. See in debugger that `setHasAuthenticatedBefore` is called instead of `handleFirstEverAuthentication`
3. Check in looker to check that `cody.auth.login:firstEver` event was not fired

### DIFF
--- a/vscode/src/configuration.ts
+++ b/vscode/src/configuration.ts
@@ -218,7 +218,10 @@ export const getFullConfig = async (): Promise<ConfigurationWithAccessToken> => 
     const isTesting = process.env.CODY_TESTING === 'true'
     const serverEndpoint =
         localStorage?.getEndpoint() || (isTesting ? 'http://localhost:49300/' : DOTCOM_URL.href)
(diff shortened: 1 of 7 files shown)


vscode/src/configuration.ts
```diff
@@ -218,7 +218,10 @@ export const getFullConfig = async (): Promise<ConfigurationWithAccessToken> =>
     const isTesting = process.env.CODY_TESTING === 'true'
     const serverEndpoint =
         localStorage?.getEndpoint() || (isTesting ? 'http://localhost:49300/' : DOTCOM_URL.href)
-    const accessToken = (await getAccessToken()) || null
+    const accessToken =
```
pkukielka, line 221: Previously we were not passing token correctly, so every first call to agent was looking like new login.

odisseus (approved):

PR #4771, opened 2024-07-03
[Cody Web]: Prepare cody web package for publishing

vovakulikov:
This PR prepares things before publishing cody web package to the npm registry. 
- Removes highlight styles since Sourcegraph client already imports these styles
- Clean up package.json dependencies (since this package is self bundled) it should has no common dependencies 

## Test plan
- CI is green

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "name": "@sourcegraph/cody-web",
+  "name": "cody-web-experimental",
   "version": "0.0.1",
   "description": "Cody standalone web app",
   "license": "Apache-2.0",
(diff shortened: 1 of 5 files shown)


web/package.json
```diff
@@ -19,19 +19,16 @@
     "test": "vitest",
     "build-ts": "tsc --build"
   },
-  "dependencies": {
-    "@sourcegraph/cody-shared": "workspace:*",
```
fkling, line 23: Do we not have to keep this, even as dev dependency? Or is it not used?

fkling (approved):

web/package.json
vovakulikov: We do use some helpers from the shared package, so we need to have it while bundling this, hence I put it in devDeps (since cody-shared package isn't published we can't have it in dependencies field)

fkling: Oh, I didn't see that it's already there (it was not expanded). I only saw it removed 👍🏻

PR #4772, opened 2024-07-03
POC: add model tags and remove deprecated models

abeatrix:
- Add ModelTag enum to categorize models
- Remove deprecated models from model options
- Update ModelSelectField to only show non-deprecated models

TODO:

- replace uiGroup with tags
- remove codyProOnly & uiGroupfields & usage from Model class

## Test plan

Current tests should all be passing

- Verify the dropdown models are showing up correctly in Chat.
- Verify the dropdown models are showing up correctly for Inline Edit.
- Verify the dropdown models are showing up correctly in JB.

image

image

image


abeatrix:
Merged into https://github.com/sourcegraph/cody/pull/4780

PR #4773, opened 2024-07-03
Fix ScrollDown button for cody web case

vovakulikov:
Fixes https://linear.app/sourcegraph/issue/SRCH-670/fix-scroll-to-bottom-button-for-cody-web

Fixes scroll down button for arbitrary scrollable container 

## Test plan
- Check that the scroll-down button works properly in the VSCode extension
- Check that the scroll-down button works in the cody web demo (`cd web` and `pnpm dev`)

### DIFF
--- a/vscode/webviews/components/ScrollDown.tsx
+++ b/vscode/webviews/components/ScrollDown.tsx
@@ -1,46 +1,88 @@
 import { ArrowDownIcon } from 'lucide-react'
-import { type FunctionComponent, useCallback, useEffect, useState } from 'react'
+import { type FC, useCallback, useEffect, useMemo, useState } from 'react'
 import { Button } from './shadcn/ui/button'
 
 const MARGIN = 200 /* px */
+
+interface Scroller {
+    root: HTMLElement | Window
+    getObserveElement: () => Element
+    getScrollTop: () => number
+    getScrollHeight: () => number
+    getClientHeight: () => number
+}
+
+// Chat UI could be run in different modes, in VSCode the whole chat
+// is rendered within scrollable iframe and it scroll down button should
+// listen and work with window updates,In Cody Web (and possible in other
+// clients) chat might be rendered in arbitrary scollable element.
+// This Scroller API helps to observe complexity of both cases and provides
+// unified API to work with different root elements.
+function createScrollerAPI(element: HTMLElement | null | undefined): Scroller {
+    if (element) {
+        return {
+            root: element,
+            getObserveElement: () => element.children[0],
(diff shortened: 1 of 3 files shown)


vscode/webviews/components/ScrollDown.tsx
```diff
@@ -1,46 +1,88 @@
...
+    root: HTMLElement | Window
+    getObserveElement: () => Element
+    getScrollTop: () => number
+    getScrollHeight: () => number
+    getClientHeight: () => number
+}
+
+// Chat UI could be run in different modes, in VSCode the whole chat
+// is rendered within scrollable iframe, and it scrolls down button should
+// listen and work with window updates,In Cody Web (and possible in other
```
dominiccooney, line 17: Simpler to avoid "and possibly in other clients" and simply say:

> In some clients like VSCode the whole chat is rendered in a scrollable iframe, in others like Cody Web it is rendered in a scrollable element.

abeatrix (approved):
Verified the scroll down button still works in IDE webview:

image

vscode/webviews/components/ScrollDown.tsx
```diff
@@ -1,46 +1,88 @@
...
+// clients) chat might be rendered in arbitrary scrollable element.
+// This Scroller API helps to observe complexity of both cases and provides
+// unified API to work with different root elements.
+function createScrollerAPI(element: HTMLElement | null | undefined): Scroller {
```
dominiccooney, line 21: You don't need this, just use document.documentElement if there's no `element` and then use `element.clientHeight` for the client height.

Admittedly if you need the resize observer you should observe document.body in the window case.

vscode/webviews/components/ScrollDown.tsx
```diff
@@ -1,46 +1,88 @@
...
+    if (element) {
+        return {
+            root: element,
+            getObserveElement: () => element.children[0],
```
dominiccooney, line 25: There is `firstElementChild` for this purpose.

dominiccooney (changes requested):
We can simplify this a bit.

vovakulikov:
@dominiccooney I suspect based on your comments you did this too but in a different manner, I think most of your comments will be resolved by your changes, as soon as they're landed in the cody repository, right?

fkling:
I wish this could be done automatically, i.e. traverse up to the root and find the element that "scrolls". But I guess that would only work if the content already causes an element to scroll, i.e. it wouldn't work in the initial/empty state.

Issue #4639, opened 2024-06-21
bug: autocomplete not working "Expected to receive exactly one message with the prompt" 


quentin-lior:
### Version

1.22.4

### Describe the bug

My Cody autocomplete is not working (the setting "Enables code autocompletions" is checked), but the chat feature is working fine. In the ouptut console, I noticed the following errors

 getInlineCompletions:error: error parsing CodeCompletionResponse: Error: {"error":"Expected to receive exactly one message with the prompt"} Error: error parsing CodeCompletionResponse: Error: {"error":"Expected to receive exactly one message with the prompt"}
    at /Users/quentin/.vscode/extensions/sourcegraph.cody-ai-1.22.4/src/completions/client.ts:224:27
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at generatorWithTimeout (/Users/quentin/.vscode/extensions/sourcegraph.cody-ai-1.22.4/src/completions/utils.ts:84:37)
    at generatorWithErrorObserver (/Users/quentin/.vscode/extensions/sourcegraph.cody-ai-1.22.4/src/completions/utils.ts:51:29)
    at fetchAndProcessDynamicMultilineCompletions (/Users/quentin/.vscode/extensions/sourcegraph.cody-ai-1.22.4/src/completions/providers/fetch-and-process-completions.ts:92:62)
    at async Promise.all (index 0)
    at zipGenerators (/Users/quentin/.vscode/extensions/sourcegraph.cody-ai-1.22.4/src/completions/utils.ts:34:21)
    at generateCompletions (/Users/quentin/.vscode/extensions/sourcegraph.cody-ai-1.22.4/src/completions/request-manager.ts:118:34)

### Expected behavior

Cody should propose autocomplete functionality when starting to write code in editor

### Additional context

I am working on python files right now.
MacOS : Sonoma 14.5
VS Code : 1.90.2

pedanticgeek:
Same here.
VSCode: 1.90.2
MacOS: 14.4.1
Cody: 1.24

Working with TypeScript

quentin-lior:
I somehow managed to get autocomplete to work. 
In Cody settings, I noticed that cody.autocomplete.advanced.provider was set to "Anthropic". However, it looks like the default value is "null", so when switching to "null", the autocomplete worked again as expected.

justinmilner1:
Hey guys @quentin-lior @pedanticgeek

- Which provider are you using for autocomplete?
[...]

PR #4775, opened 2024-07-03
Anthropic shouldn't be in list of autocomplete options

justinmilner1:
Addresses [this bug](https://github.com/sourcegraph/cody/issues/4639)

## Testing

Bug replicated
Screenshot from 2024-07-03 19-44-29

After changes, anthropic option removed.
Screenshot from 2024-07-03 19-46-26


justinmilner1:
Wait this isn't correct

Issue #3825, opened 2024-04-16
bug: v2 telemetry for Cody installations not being fired


dadlerj:
### Version

Current/all

### Describe the bug

The v2 `cody.extension` / `installed` event seems to ~~ never be logged (we have tens of thousands of `cody.extension` / `savedLogin` events fired per day, but single digit `cody.extension` / `installed` ones: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/services/telemetry-v2.ts?L139

One of the two parameters in the conditional must be incorrectly set.

### Expected behavior

New installations should fire the event as expected

### Additional context

[Slack discussion](https://sourcegraph.slack.com/archives/C01PQEVJ5A9/p1713300361446909?thread_ts=1712946204.912669&cid=C01PQEVJ5A9)

akalia25:
In case this is helpful, you can identify this bug on the e2e tests (i.e.`vscode/test/e2e/auth.test.ts`). Below is a screenshot of v1 events and v2 events fired on the auth.test.

<img width="879" alt="image" src="https://github.com/sourcegraph/cody/assets/32119652/f105bca7-d497-4d49-9e54-c34f95d1b7b5">

to view v2 telemetry events, you can export `loggedV2Events` on `vscode/test/fixtures/mock-server.ts` and import that on the test, and do a `console.log`.

kelsey-brown:
@chenkc805 any chance we could get this prioritized for the next week or two? We're currently using the V1 install event for all our install related metrics as a backup, but with the news that we're going to [turn off V1 events](https://sourcegraph.slack.com/archives/CN4FC7XT4/p1718328270800609) on Wednesday, I'd love to be able to rely on V2 data for installs

PR #4776, opened 2024-07-04
Remove v1 telemetry from Cody

dadlerj:
Fixes https://github.com/sourcegraph/cody/issues/3825

## Test plan

- CI
- Test that `pnpm run dev` runs without error and the Cody extension response to user interactions (e.g. a chat) without error


dadlerj:
Sorry @umpox if you're not the right reviewer from the Cody team, you were just suggested by GitHub 😅 —feel free to tag someone else.

umpox (approved):
Love it :)

FYI: The CI failures now seem to just be from the e2e tests, I think you just need to remove

`expectedEvents`, e.g: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/test/e2e/command-edit.test.ts?L12%3A1-18%3A7

and the logic around `loggedEvents`: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/test/e2e/helpers.ts?L296%3A17-296%3A75

vscode/src/chat/protocol.ts
```diff
@@ -40,6 +39,21 @@ export type WebviewRecordEventParameters = TelemetryEventParameters<
     BillingCategory
 >
 
+/**
+ * @deprecated v1 telemetry event properties format - use 'recordEvent' instead
+ */
+interface TelemetryEventProperties {
```
bobheadxi, line 45: Is this still needed?

bobheadxi (approved):
Woo!

vscode/src/chat/protocol.ts
dadlerj: Not really, no, but I don't feel quite confident enough yet to get rid of the `event` message handler yet since the type system won't catch callers for me (I think?). https://github.com/sourcegraph/cody/pull/4776/files#diff-9b7ca9645cb0941f1503b92e5b60bff6d98be9e8c7cfce7910b292c37a323dd5R63-R70. So I'm just going to leave it for now and mark deprecated.

PR #4777, opened 2024-07-04
[Cody Web]: Support `openctx` providers

vovakulikov:
Closes https://linear.app/sourcegraph/issue/SRCH-596/remote-repositories
Closes https://linear.app/sourcegraph/issue/SRCH-576/new-ui-added-openctx-which-has-dependencies-on-the-dom-so-it-doesnt

This PR adds support for open context providers, so we would be able to mention remote repositories and remote files in Cody Web. It tries to not touch logic outside of the scope, changes should have no effect on VSCode or general cody agent logic.

## Test plan
- Run cody web demo 
- Check that you can use remote repositories and remote files providers mentions (which is called just repositories and files when it's running for Cody Web)


lib/shared/src/sourcegraph-api/completions/browserClient.ts
```diff
@@ -94,30 +98,3 @@ export class SourcegraphBrowserCompletionsClient extends SourcegraphCompletionsC
...
-declare const WorkerGlobalScope: never
-const isRunningInWebWorker =
-    typeof WorkerGlobalScope !== 'undefined' && self instanceof WorkerGlobalScope
-
-if (isRunningInWebWorker) {
-    // NOTE: If we need to add more hacks, or if this is janky, we should consider just setting
-    // `globalThis.window = globalThis` (see
-    // https://github.com/sourcegraph/cody/pull/4047#discussion_r1593823318).
-
-    ;(self as any).document = {
```
sqs, line 107: This is needed for VS Code Web (eg vscode.dev). That’s not a priority right now but also it’d be better to not delete it from here. Sorry that’s not documented here. 

So you shouldn’t move it to the web/ package, it needs to stay here.

vscode/src/context/openctx.ts
```diff
@@ -57,6 +39,60 @@ export async function exposeOpenCtxClient(
...
+            settings: true,
+            provider: LinearIssuesProvider,
+            providerUri: LinearIssuesProvider.providerUri,
+        })
+    }
+
+    return providers
+}
+
+function getCodyWebOpenContextProviders() {
```
sqs, line 81: Why does this need to be an entirely different code path and not just if statements for agentIDE in the same function?

vscode/src/main.ts
```diff
@@ -292,6 +299,14 @@ const register = async (
         // Sync auth status to graphqlClient
         graphqlClient.onConfigurationChange(newConfig)
 
+        // Re-register expose open context client
```
sqs, line 302: Call it OpenCtx not “open context”. (I notice your PR descriptions say “open context” as well. Should just be “OpenCtx”.)

vscode/webviews/Chat.tsx
```diff
@@ -166,6 +167,11 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
         }
     }, [])
 
+    // biome-ignore lint/correctness/useExhaustiveDependencies: needs to run when is dotcom status is chaning to update openctx providers
```
sqs, line 170: Typo: changing not chaning

sqs (changes requested):

sqs (approved):

lib/shared/src/sourcegraph-api/completions/browserClient.ts
vovakulikov: I see, I thought it was only important for the Cody Web (not vscode.dev). I also moved it here because vite adds some browser-api-specific helpers that break imports of `openctx/vscode-lib`

I'll move it back and apply my mock for querySelectorAll in `browserClient.ts` then.

vscode/src/context/openctx.ts
vovakulikov: There are already too many if checks in my opinion, there is a config flag, dotcom checks and now it's cody web checks, I wanted to have a clear place where you can specify only CodyWeb providers to simplify further improvements and avoid unexpected providers appearances in Cody Web package.

vscode/webviews/Chat.tsx
```diff
@@ -166,6 +167,11 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
...
+    useEffect(() => {
+        reloadMentionProviders()
+    }, [userInfo.isDotComUser, reloadMentionProviders])
```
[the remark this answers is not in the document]
vovakulikov: I think there might be a case when you're switching accounts from enterprise to dotcom, and in this case, we should reloadMentionProviders. One of e2e found this, actually. 

we run this additional time on chat mount but it should be fine since mention provider registration doesn't do any network request and just checks auth state and dotcom settings and re-register providers set for mentions.

vscode/src/context/openctx.ts
```diff
@@ -57,6 +39,60 @@ export async function exposeOpenCtxClient(
     }
 }
 
+function getStandardOpenContextProviders(config: ConfigurationWithAccessToken, isDotCom: boolean) {
+    // TODO [vk] expose types for providers from openctx/client lib
```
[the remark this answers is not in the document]
vovakulikov: This is what I see in `@opectx/vscode-lib` 

```ts
export declare function createController({ secrets: secretsInput, extensionId, outputChannel, getAuthInfo, features, providers, preloadDelay, }: {
    secrets: Observable<vscode.SecretStorage> | vscode.SecretStorage;
    extensionId: string;
    outputChannel: vscode.OutputChannel;
    getAuthInfo?: (secrets: vscode.SecretStorage, providerUri: string) => Promise<AuthInfo | null>;
    features: {
        annotations?: boolean;
        statusBar?: boolean;
    };
    providers?: ImportedProviderConfiguration[];
    preloadDelay?: number;
}): {
    controller: Controller;
    apiForTesting: ExtensionApiForTesting;
    disposable: vscode.Disposable;
};
export {};
```

So there is no type export for `createController` so I could 
```ts
import type { controllerFactory } from '@openctx/vscode-lib'

expose(customControllerFactory: controllerFactory, ... ) {...}
```


I could `typeof createController` but I should avoid here any sync imports of  '@openctx/vscode-lib' because it will break agent usage.

vovakulikov:
Thanks for the review. I fixed some comments, here, others don't look like blockers, Will address problems in the follow-up 😉

vscode/src/context/openctx.ts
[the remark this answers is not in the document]
sqs: @vovakulikov Yeah, the `import type` solution seems important and good to me, whether that is exporting `typeof createController` from the package itself, or just `import type` if that works for non-type values that are only used in a `typeof` context.

PR #4780, opened 2024-07-04
feat(enterprise): allow multiple models for enterprise customers

jamesmcnamara:
Allows for enterprises to support multiple models. Additionally contains some refactors for unifying further a single Model and ModelsService class.

## Test plan
Added E2E test


lib/shared/src/models/index.ts
```diff
@@ -1,175 +1 @@
-import { fetchLocalOllamaModels } from '../llm-providers/ollama/utils'
```
abeatrix, line 1: is this removed by accident?

abeatrix (commented):
left some comments in line. It looks like the model/index.ts file was removed by accident so I wasn't able to test it locally, but most of the changes look good to me!

There is an error caused by the change from my other commits that was merged into your branch, I've started the PR (that target your current branch)  with the changes to fix that: https://github.com/sourcegraph/cody/pull/4795

Feel free to merge that into this branch and make any additional changes as you need!

Great job!

lib/shared/src/models/index.ts
jamesmcnamara: Yeah I don't know how this happened. Resolved now.

vscode/webviews/chat/models/chatModelContext.tsx
```diff
@@ -33,5 +33,5 @@ export function useChatModelByID(
 
 export function useCurrentChatModel(): Model | undefined {
     const { chatModels } = useChatModelContext()
-    return chatModels?.find(model => model.default) ?? chatModels?.[0]
+    return chatModels?.[0]
```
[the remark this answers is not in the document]
jamesmcnamara: Yes it happens [here](https://github.com/sourcegraph/cody/pull/4780/files#diff-fe82b352761565cc5b6eb38a881a5bda7c83ad2c6b9231fa4bb25364154cc407R282). Trying to update the field `default` was kind of a violation of the immutability of `Model` and did lots of weird casting which I didn't love. And besides we alway fell back to the head of the list so I just settled on that convention.

agent/src/agent.ts
```diff
@@ -1079,17 +1073,14 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
-                      ModelUsage.Chat,
-                      authStatus.isDotCom && !authStatus.userCanUpgrade
-                  ).at(0)?.model
-            if (!theModel) {
-                theModel = getDotComDefaultModels()[0].model
+            if (!modelID) {
+                modelID = ModelsService.getDefaultChatModel(authStatus)
+            }
+            if (!modelID) {
+                throw new Error('agent::chat/restore: no chat model found')
```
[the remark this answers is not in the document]
jamesmcnamara: Sure. Except I don't see where it would fallback to the default, everywhere in `SimpleChatProvider` it just uses the modelId directly.

chrsmith:
Hey @jamesmcnamara , I'll take a look at this PR today or tomorrow. Though we'll want to figure out when we actually ship this. Just so there aren't any surprises relative to when backend changes go out vs. when the new extension bits get released.

jamesmcnamara:

Can be ignored, it's just a syntactic change to an existing function

lib/shared/src/models/index.ts
```diff
@@ -1,70 +1,213 @@
...
+        modelRef,
+        displayName,
+        capabilities,
+        category,
+        tier,
+        clientSideConfig,
+        contextWindow,
+    }: ServerModel) {
+        // BUG: There is data loss here and the potential for ambiguity.
+        // BUG: We are assuming the modelRef is valid, but it might not be.
```
abeatrix, line 135: maybe in a follow-up, we could create a helper function to validate if the modelRef is valid or not?

lib/shared/src/models/index.ts
```diff
@@ -1,70 +1,213 @@
...
+export type ModelId = string
+export type ApiVersionId = string
+export type ProviderId = string
+
+export type ModelRef = `${ProviderId}::${ApiVersionId}::${ModelId}`
+
+export type ModelCapability = 'chat' | 'autocomplete'
+export type ModelCategory = 'accuracy' | 'balanced' | 'speed'
+export type ModelStatus = 'experimental' | 'beta' | 'stable' | 'deprecated'
+export type ModelTier = 'free' | 'pro' | 'enterprise'
```
abeatrix, line 17:
```suggestion
export type ModelCategory = ModelTag.Accuracy | ModelTag.Balanced | ModelTag.Speed
export type ModelStatus = ModelTag.Experimental | ModelTag.Experimental | 'stable' | ModelTag.Deprecated
export type ModelTier = ModelTag.Free | ModelTag.Pro | ModelTag.Enterprise
```

safer / to avoid typos when the list gets longer?

lib/shared/src/models/index.ts
```diff
@@ -1,70 +1,213 @@
...
+        }
+        if (model.tags.includes(ModelTag.Pro)) {
+            return 'pro'
+        }
+        if (model.tags.includes(ModelTag.Enterprise)) {
+            return 'enterprise'
+        }
+
+        return 'pro'
+    }
```
abeatrix, line 191: static tier(model: Model): ModelTier {
        const tierSet = new Set<ModelTag>([ModelTag.Pro, ModelTag.Enterprise])
        return (model.tags.find(tag => tierSet.has(tag)) ?? ModelTag.Free) as ModelTier
    }
```

Should we use ModelTag.Free as the fallback default instead of 'pro' here?

File: lib/shared/src/models/index.ts [hunk 7]
```diff
@@ -1,70 +1,213 @@
...
+
+        const tierTag = ((): ModelTag => {
+            switch (tier) {
+                case 'free':
+                    return ModelTag.Free
+                case 'pro':
+                    return ModelTag.Pro
+                case 'enterprise':
+                    return ModelTag.Enterprise
+            }
```
abeatrix, line 157: do we need to set the default switch case for tierTag and categoryTag?

abeatrix (commented):
I've left some comments inline. Most of them are non-blockers, except for some of the the ModelTags are not assigned correctly for the pre-definded dotcom models. Other than that they all look good to me, and the models are set correctly on reload and updated when I run it locally. Great job!!

lib/shared/src/models/index.ts
jamesmcnamara: My concern is that if something goes wrong, a Pro user can use any model available but a free user cannot. So it seemed like the safer assumption but I'm persuadable.

abeatrix: Ah I was thinking the opposite way where it'd be frustrating for a pro user not being able to use the model that should be available to them vs free users able to select a model that's not available to them. Just a thought tho!

jamesmcnamara: I think that would work because the default tier is `pro` and the Pro users can use any model that's not `enterprise`, so they should be fine.

lib/shared/src/models/index.ts
jamesmcnamara: We can actually just nuke this entirely since we're re-using the enums.

vscode/webviews/Chat.tsx
```diff
@@ -201,6 +201,9 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
 export interface UserAccountInfo {
     isDotComUser: boolean
     isCodyProUser: boolean
+    // When true, the user is on an enterprise instance that has not upgraded to
+    // allow multiple LLM selections
+    isOldStyleEnterpriseUser: boolean
```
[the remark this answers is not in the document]
jamesmcnamara: Yes all users are oldStyle until they start sending server-side models. All this controls is if they see the drop down or not (old style see no drop down).

vscode/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 - Ollama: Added support for running Cody offline with local Ollama models. [pull/4691](https://github.com/sourcegraph/cody/pull/4691)
 - Edit: Added support for users' to edit the applied edit before the diff view is removed. [pull/4684](https://github.com/sourcegraph/cody/pull/4684)
 - Autocomplete: Added experimental support for Gemini 1.5 Flash as the autocomplete model. To enable this experimental feature, update the `autocomplete.advanced.provider` configuration setting to `unstable-gemini`. Prerequisite: Your Sourcegraph instance (v5.5.0+) must first be configured to use Gemini 1.5 Flash as the autocomplete model. [pull/4743](https://github.com/sourcegraph/cody/pull/4743)
+- Enterprise: Enabled support for multiple dynaic models if the Sourcegraph backend provides them. Requires the experimental flag `cody.dev.useServerDefinedModels` which currently returns a static list of models. Note: this won't work until the server side of this feature is merged.
```
abeatrix, line 12:
```suggestion
- Enterprise: Enabled support for multiple dynaic models if the Sourcegraph backend provides them. Requires the experimental flag `cody.dev.useServerDefinedModels` which currently returns a static list of models. Note: this won't work until the server side of this feature is merged. [pull/4780](https://github.com/sourcegraph/cody/pull/4780)
```

abeatrix (approved):
This is awesome! I've manually tested the following and verified it works accordingly:
[x] Can switch between chat models
[x] Chat models are displayed with the correct icons
[x] Chat models are displayed correctly with the correct UI group assigned
[x] New chat panel defaulted to use the last selected model correctly
[x] Can switch between inline-edit models
[x] Inline-edit models are displayed with the correct icons
[x] New chat panel defaulted to use the last selected model correctly
[x] Reloaded editor and the last selected models are showing up as default correctly

https://github.com/sourcegraph/cody/assets/68532117/a6da2e80-2c6e-4cdb-9870-c47a3991126d

vscode/CHANGELOG.md
jamesmcnamara: Ooh this actually reminds me that that changelog is no longer true. I'll update both.

jamesmcnamara:
Thanks @abeatrix for the thorough testing!

jamesmcnamara:
@chrsmith Do you want me to hold off merging this until you get a chance to review it?

chrsmith (approved):
From my understanding of how things all fit together this looks good to me! I assume there will be some subtle details to work out as we do more testing with server-supplied LLM models. But overall this looks like a great step in the right direction.

I wouldn't block submission on any of my suggestions or nit picks. Just things to consider.

vscode/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 - Ollama: Added support for running Cody offline with local Ollama models. [pull/4691](https://github.com/sourcegraph/cody/pull/4691)
 - Edit: Added support for users' to edit the applied edit before the diff view is removed. [pull/4684](https://github.com/sourcegraph/cody/pull/4684)
 - Autocomplete: Added experimental support for Gemini 1.5 Flash as the autocomplete model. To enable this experimental feature, update the `autocomplete.advanced.provider` configuration setting to `unstable-gemini`. Prerequisite: Your Sourcegraph instance (v5.5.0+) must first be configured to use Gemini 1.5 Flash as the autocomplete model. [pull/4743](https://github.com/sourcegraph/cody/pull/4743)
+- Enterprise: Enabled support for multiple dynaic models if the Sourcegraph backend provides them. Requires the experimental flag `modelsAPIEnabled` to be sent by the client config API. Note: this won't work until the server side of this feature is merged. [pull/4780](https://github.com/sourcegraph/cody/pull/4780)
```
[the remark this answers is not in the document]
slimsag: We should just add the changelog entries when we're done with everything yeah.

And yes, `modelsAPIEnabled` already works today - it will be `true` as long as you add this to your site config:

```
"modelsConfiguration": {},
```

lib/shared/src/models/dotcom.ts
```diff
@@ -40,45 +41,33 @@ export const DEFAULT_DOT_COM_MODELS = [
         title: 'Claude 3.5 Sonnet',
         model: 'anthropic/claude-3-5-sonnet-20240620',
         provider: 'Anthropic',
-        default: true,
-        codyProOnly: false,
         usage: [ModelUsage.Chat, ModelUsage.Edit],
         contextWindow: expandedContextWindow,
-        deprecated: false,
-        uiGroup: ModelUIGroup.Accuracy,
+        tags: [ModelTag.Gateway, ModelTag.Accuracy, ModelTag.Recommended, ModelTag.Free],
```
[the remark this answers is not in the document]
jamesmcnamara: This started as @abeatrix brain child, so I'll let her elaborate on it.

lib/shared/src/models/index.ts
```diff
@@ -126,24 +262,83 @@ export class ModelsService {
...
+        }
+        if (!resolved.usage.includes(type)) {
+            throw new Error(`Model "${resolved.model}" is not compatible with usage type "${type}".`)
+        }
+        ModelsService.defaultModels.set(type, resolved)
+        // If we have persistent storage set, write it there
+        await ModelsService.storage?.set(ModelsService.storageKeys[type], resolved.model)
+    }
+
+    public static canUserUseModel(status: AuthStatus, model: string | Model): boolean {
```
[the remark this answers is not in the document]
jamesmcnamara: For me, that's asking "Is this model functionally usable" but the question is really "does this user have permissions for this model". I could see `isModelAvailableFor` but that might be just as clunky.

lib/shared/src/models/index.ts
```diff
@@ -126,24 +262,83 @@ export class ModelsService {
...
+
+    public static async setDefaultModel(type: ModelUsage, model: Model | string): Promise<void> {
+        const resolved = ModelsService.resolveModel(model)
+        if (!resolved) {
+            return
+        }
+        if (!resolved.usage.includes(type)) {
+            throw new Error(`Model "${resolved.model}" is not compatible with usage type "${type}".`)
+        }
+        ModelsService.defaultModels.set(type, resolved)
```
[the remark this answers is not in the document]
jamesmcnamara: This seems to me like a product question so maybe @aramaraju or @kalanchan could weigh in: in an enterprise environment where the admin has set default models for different usage types, and then a user has selected a different model (one that is still supported by their environment) as their default, what should happen when the site admin changes the default models again? Should it overwrite the user's default? Or should it be ignored.

My 2 cents is to have "default models" (set by the server) and "selected models" with the resolution being:
1. user selected model for action (chat, autocomplete, etc.)
2. if 1) not set, use server sent default
3. if 2) not set, use first model from the list that fits the criteria (i.e. the user has permissions on it and it's recommended for that action

However, this does kind of ignore when a site admin changes the default behavior so maybe it's not sufficient?

lib/shared/src/models/index.ts
```diff
@@ -126,24 +262,83 @@ export class ModelsService {
...
+    }
+
+    public static canUserUseModel(status: AuthStatus, model: string | Model): boolean {
+        const resolved = ModelsService.resolveModel(model)
+        if (!resolved) {
+            return false
+        }
+        const tier = Model.tier(resolved)
+        if (isEnterpriseUser(status)) {
+            return tier === 'enterprise'
```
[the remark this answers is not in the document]
jamesmcnamara: Ah I assumed that enterprise would be more restrictive. I'll add your verbosity as well.

lib/shared/src/models/index.ts
```diff
@@ -172,4 +367,22 @@ export class ModelsService {
...
+
+interface Storage {
+    get(key: string): string | null
+    set(key: string, value: string): Promise<void>
+}
+
+export function capabilityToUsage(capability: ModelCapability): ModelUsage[] {
+    switch (capability) {
+        case 'autocomplete':
+            return []
```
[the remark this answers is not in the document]
jamesmcnamara: This is just trying to parse what was specified [here](https://github.com/sourcegraph/sourcegraph/blob/main/internal/modelconfig/types/model.go#L10-L17) to something the extension can understand.

From what I've seen on the front end, autocomplete models seem to be handled by configuration values and not from the ModelsService, but I'm not sure how that works for enterprise. I've tried to wade through the completions code but it's non-trivial to say the least.

Outside of autocomplete, each model may flag that they support being used for chat or for being used for code-editing (like generate unit tests, fixup etc), but it's possible that some models aren't available for both.

So I'm not sure if what you were describing was on the server side or current client side, but that's my brain dump. Given that, what do you think the mapping should be?

vscode/src/chat/chat-view/SimpleChatModel.ts
```diff
@@ -21,17 +22,18 @@ export class SimpleChatModel {
...
         public readonly sessionID: string = new Date(Date.now()).toUTCString(),
+        private messages: ChatMessage[] = [],
         private customChatTitle?: string,
         private selectedRepos?: Repo[]
     ) {
         this.contextWindow = ModelsService.getContextWindowByID(this.modelID)
     }
 
-    public updateModel(newModelID: string): void {
+    public async updateModel(newModelID: string): Promise<void> {
```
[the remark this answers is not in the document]
jamesmcnamara: I would be alright renaming it `setDefaultModel`.

lib/shared/src/models/index.ts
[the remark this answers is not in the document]
jamesmcnamara: Feedback from product ([Slack thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1720722271718969)):

> we want to give more power to the admin and have this be something that customers handle internally between the end user and the admin. This accrues more value to the admin, making the Cody Enterprise plan more appealing.
I also assume that admins changing the default model won't happen that often.
Following this line of reasoning, then we should override the end user's selection.


So I'll add that

vscode/webviews/components/modelSelectField/ModelSelectField.tsx
```diff
@@ -43,8 +43,7 @@ export const ModelSelectField: React.FunctionComponent<{
 }) => {
     const telemetryRecorder = useTelemetryRecorder()
 
-    const usableModels = useMemo(() => models.filter(m => !m.deprecated), [models])
-    const selectedModel = usableModels.find(m => m.default) ?? usableModels[0]
+    const selectedModel = models[0]
```
abeatrix, line 46: Let's add a comment to say that the default model is always in the first of the list?

jamesmcnamara:
@chrsmith I started to work on the server sent defaults but I realized there was going to be a significant amount of work. Since this keeps getting outdated I'm going to merge what we have now, and then continue with the defaults work on another.

PR #4787, opened 2024-07-04
Generate Unit Test Experiment

thenamankumar:
Context: https://sourcegraph.slack.com/archives/C06RW56B2DD/p1720080160136309?thread_ts=1720063332.078939&cid=C06RW56B2DD

This is experimental and to be removed later. The goal is to dogfood on new unit test prompt template with user input.

It is feature flagged under `cody-experimental-unit-test`.

Adds a button "Generate Unit Test" to update the input box with the prompt template. 

image

## Test plan

- Feature flag is enabled for all users under sourcegraph org on dotcom
- Click the button
- update the placeholder text inside {{}}.
....


keegancsmith:
unit test failures seem legit. I'm guessing something to dowith reading the feature flag in the wrong place or needing to add more mocks due to the feature flag.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
```diff
@@ -515,13 +520,21 @@ export class SimpleChatPanelProvider
...
             default:
                 this.postError(new Error(`Invalid request type from Webview Panel: ${message.command}`))
         }
     }
 
     private async getConfigForWebview(): Promise<ConfigurationSubsetForWebview & LocalEnv> {
-        const config = await getFullConfig()
+        const [config, experimentalUnitTest] = await Promise.all([
+            getFullConfig(),
+            this.featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyExperimentalUnitTest),
```
keegancsmith, line 535: Is this the right place to get this coming in? This feels very "init" level, ie nothing here seems to indicate that we have even successfully authorized against the endpoint. Just below this is `sendConfig` which may be more appropriate? TBH this is just a feeling I have, would be great to get confirmation with someone who owns this bit of the code more.

keegancsmith (approved):
LGTM. It's a bit of a shame so much of this PR seems to be scaffolding around sending an RPC. Was there not some way to sneak into something like our custom commands or something so we didn't have something so specific? Not sure if that makes any sense.

I'm approving to unblock, but I do have one concern inline about reading feature flags.

vscode/src/chat/chat-view/SimpleChatPanelProvider.ts
thenamankumar: So `sendConfig` does nothing but to call `getConfigForWebview`. 

`getConfigForWebview` is being called only from within `sendConfig`. The `experimentalNoodle` flag is also being passed from here. 

Secondly, we do call `sendConfig` after every auth status change, so it will be in sync. It feels like the right place to put it.

thenamankumar:
I am facing the below error while generating Kotlin types. I recognize that this is because I updated the editor interface to add `setEditorState` function. It seems like I need to update the state version and check for the same somewhere. I have tried updating the version to v1 and checking for the same, but I am unable to get it fixed.

Given the current version is v0, I can't find such a version check anywhere. 

Do you know hot to fix this? @olafurpg 

CleanShot 2024-07-06 at 00 45 35@2x

olafurpg:
This errors happens when you expose a type into the agent protocol that doesn't have a straightforward encoding in other languages. The codegen tool has a few exceptions to ignore certain properties (don't have a link since I'm on the phone but feel free to add exceptions if that unblocks you). Alternatively, avoid exposing the complicated type in the protocol. The error message is admittedly cryptic, it could produce better error messages with a bit of effort but I haven't had the time polish this part.

vscode/src/commands/execute/test-chat-experimental.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    return {
+        speaker: 'human',
+        text: prompt,
+        editorState: {
+            lexicalEditorState: lexicalEditorStateFromPromptString(prompt),
+            v: STATE_VERSION_CURRENT,
+            minReaderV: STATE_VERSION_CURRENT,
+        },
+    }
```
keegancsmith, line 41: Now that I have dived into this a bit more I am unsure why this is a command and why this is the return type. It isn't blocking so will land it like this, but it seems we should follow the pattern used in other commands. EG maybe this should be the same shape as `ExecuteChatArguments` command, except maybe the `submitType` (maybe renamed to actionType) indicates updating the prompt.

keegancsmith:
Alright I sorted the generation of kotlin issue, but now I think the issue is we introduced featureflag reading into an early on codepath so a bunch of tests don't have that mocked out. Will rather see if we can avoid that, otherwise will just update a bunch of mocks.

PR #4789, opened 2024-07-04
Fix cursor position after submitting mention items

vovakulikov:
Based on https://github.com/sourcegraph/cody/pull/4777
Fixes https://linear.app/sourcegraph/issue/SRCH-672/cursor-jumps-around-with-mentions

I think the original issue that was reported [here](https://linear.app/sourcegraph/issue/SRCH-672/cursor-jumps-around-with-mentions) has already been partially fixed, but there is still a problem with two-step provider mentions, I think, and this PR tries to solve this. 

So the problem was that when you have this text `Hello World` and you're trying to add remote file mentions like `Hello @<cursor> World` and select some repository from the suggestion, the result will be `Hello @<cursor>github.com/sourcegraph/sourcegraph: World`, you can see the problem that the cursor stays in the same position, and this breaks two steps provider mentions flow. 

Another problem is that we rely too much on the `@` symbol when we're trying to find a current mention in progress to insert text from the suggestion panel, for example when you have `Hello@ World@` and you're typing `Hello@ @<cursor> World@`, the `@` symbol that actually triggered mention panel won't be used but I think only first `@` after the word `Hello` will be used and as a result, we will have `Hello<mention> @ World@`, which is not correct I guess. 

@keegancsmith I've seen that you recently added some logic about similar prefixes [here](https://github.com/sourcegraph/cody/pull/4654), I'm not sure I fully understand this, I tried to fix only two steps mentions in this PR, can you test this manually since you obviously have more cases to check this properly

## Test plan
- Build vscode extension and check that mentions works properly (different types files, symbols, repository, remote files, remote repositories)
- Run Cody Web demo (`cd ./web` and `pnpm dev`) and check that all mentions work properly and you see no jumps as you select some mentions with keyboard or with mouse.

### DIFF
--- a/vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
+++ b/vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
@@ -120,12 +120,12 @@ export const MentionMenu: FunctionComponent<
 
                     if (mentionStartIndex !== -1) {
                         const mentionEndIndex = mentionStartIndex + mentionQuery.text.length
-                        return (
-                            currentText.slice(0, mentionStartIndex) + currentText.slice(mentionEndIndex)
-                        )
+                        return [
+                            currentText.slice(0, mentionStartIndex) + currentText.slice(mentionEndIndex),
+                        ]
                     }
 
-                    return ''
+                    return ['']
                 })
             }
             setValue(null)
@@ -177,12 +177,25 @@ export const MentionMenu: FunctionComponent<
                         },
                     })
 
-                    setEditorQuery(currentText =>
-                        currentText.replace(
-                            `@${mentionQuery.text}`,
-                            `@${openCtxItem.mention?.data?.repoName}:`
-                        )
-                    )
+                    setEditorQuery(currentText => {
+                        const selection = getSelection()
+
+                        if (!selection) {
+                            return [currentText]
+                        }
+
+                        const cursorPosition = selection.anchorOffset
+                        const mentionStart = cursorPosition - mentionQuery.text.length
+                        const mentionEndIndex = cursorPosition
+                        const textToInsert = `${openCtxItem.mention?.data?.repoName}:`
+
+                        return [
+                            currentText.slice(0, mentionStart) +
+                                textToInsert +
+                                currentText.slice(mentionEndIndex),
+                            mentionStart + textToInsert.length,
+                        ]
+                    })
 
                     setValue(null)
                     return
--- a/vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
+++ b/vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
@@ -78,7 +78,7 @@ function scanForMentionTriggerInLexicalInput(text: string) {
     return scanForMentionTriggerInUserTextInput({ textBeforeCursor: text, includeWhitespace: true })
 }
 
-export type setEditorQuery = (getNewQuery: (currentText: string) => string) => void
+export type setEditorQuery = (getNewQuery: (currentText: string) => [string, number?]) => void
 
 export default function MentionsPlugin(): JSX.Element | null {
     const [editor] = useLexicalComposerContext()
@@ -104,8 +104,8 @@ export default function MentionsPlugin(): JSX.Element | null {
         limit: SUGGESTION_LIST_LENGTH_LIMIT,
     })
 
-    const setEditorQuery = useCallback(
-        (getNewQuery: (currentText: string) => string): void => {
+    const setEditorQuery = useCallback<setEditorQuery>(
+        getNewQuery => {
             if (editor) {
                 editor.update(() => {
                     const node = $getSelection()?.getNodes().at(-1)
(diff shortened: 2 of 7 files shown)


vscode/webviews/promptEditor/plugins/atMentions/chatContextClient.tsx
```diff
@@ -41,10 +41,14 @@ const ChatContextClientContext: React.Context<ChatContextClient> = createContext
             // Clean up after a while to avoid resource exhaustion in case there is a bug
             // somewhere.
             const MAX_WAIT_SECONDS = 15
-            const rejectTimeout = setTimeout(() => {
-                reject(new Error(`no ${RESPONSE_MESSAGE_TYPE} response after ${MAX_WAIT_SECONDS}s`))
-                dispose()
-            }, MAX_WAIT_SECONDS * 1000)
+            const rejectTimeout = !query.includeRemoteRepositories
```
vovakulikov, line 44: Cody Web related problem (this change has no effect on any other clients), didn't want to create a separate PR here. 

TL;DR: We don't need to wait and reject query requests to an agent with its running within Cody Web since all API is async and could take more than max wait seconds await time.

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -120,12 +120,12 @@ export const MentionMenu: FunctionComponent<
 
                     if (mentionStartIndex !== -1) {
                         const mentionEndIndex = mentionStartIndex + mentionQuery.text.length
-                        return (
-                            currentText.slice(0, mentionStartIndex) + currentText.slice(mentionEndIndex)
-                        )
+                        return [
+                            currentText.slice(0, mentionStartIndex) + currentText.slice(mentionEndIndex),
+                        ]
```
keegancsmith, line 125: is there anyway to get the current cursor position here so that `mentionStartIndex` can search backwards from that?

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -177,12 +177,25 @@ export const MentionMenu: FunctionComponent<
...
                     })
 
-                    setEditorQuery(currentText =>
-                        currentText.replace(
-                            `@${mentionQuery.text}`,
-                            `@${openCtxItem.mention?.data?.repoName}:`
-                        )
-                    )
+                    setEditorQuery(currentText => {
+                        const selection = getSelection()
```
keegancsmith, line 181: brilliant. I was unsure why this would work since I don't remember us highlighting text as we wrote, but just did a bunch of testing on remote files and it worked really well.

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -177,12 +177,25 @@ export const MentionMenu: FunctionComponent<
...
+
+                        if (!selection) {
```
keegancsmith, line 183: maybe guard against `selection.anchorOffset > mentionQuery.text.length`? Don't know why that would happen, but that is an invariant the code below relies on.

vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
```diff
@@ -114,23 +114,27 @@ export default function MentionsPlugin(): JSX.Element | null {
...
-                    const offset = sharedPrefixLength(currentText, newText)
-                    node.select(offset, offset)
+                    if (index !== undefined) {
+                        node.select(index, index)
+                    } else {
+                        // If our old text was "prefix @bar baz" and the new text is
+                        // "prefix @ baz" then our cursor should be just after @
+                        // (which is at the common prefix position)
+                        const offset = sharedPrefixLength(currentText, newText)
+                        node.select(offset, offset)
```
keegancsmith, line 131: do you reckon we could also rely on `getSelection` here instead of these hacks?

keegancsmith (approved):
Sorry for not following up on that last PR to make this bullet proof. So the main source of bugs in this and what still needed to be done is we needed to know to things:
1) position of the @ character which triggered the mention menu
2) position of the cursor

All the code hacked around it by looking for `@` and assuming the cursor was at the end of the chat box. This is def an  improvement, and I see you have managed to incorporate some of that. Now that you hacked around, do you think there is a way we could track (1) and (2) to make this bullet proof?

FYI I tested this,LGTM!

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
vovakulikov: We can yeah, just didn't want to mess with provider selection logic, I will address mentions position in general after the upcoming release, just to make sure I don't add more new implicit problems than solved ones

vscode/webviews/mentions/mentionMenu/MentionMenu.tsx
vovakulikov: Yeah it works in a way that even if you don't have actual selection it still tracks cursor position (selection in this case has the same start and end indexes)

vscode/webviews/promptEditor/plugins/atMentions/atMentions.tsx
vovakulikov: In theory, I do, but I'm not sure I understood correctly the reason for this `sharedPrefixLength` hack completely, so I left it as a follow-up for all other cases to avoid new problems with this fix.

PR #4793, opened 2024-07-05
e2e2: support windows for extension installation

keegancsmith:
Test Plan: pnpm run test:e2e2:run on windows

Fixes https://linear.app/sourcegraph/issue/CODY-2677/e2e2-working-on-windows


RXminuS (approved):
Yeah...I need to get used to not using `/` . Feels like we should be able to add some linting error for this

PR #4796, opened 2024-07-05
Fix @openctx/vscode-lib initialization for Cody Web

vovakulikov:
`@openctx/vscode-lib` uses global. sets under the hood which breaks Cody Web use case since we're importing vscode-lib within web worker context where we don't have global object. 

This PR adds hack/workaround for Cody Web, proper fix should happen in `@openctx/vscode-lib` here https://github.com/sourcegraph/openctx/issues/169

## Test plan
- Run cody web demo
- Check that you can see and use repositories and files mentions providers


abeatrix (approved):
Was able to start the demo web app: 
image

PR #4797, opened 2024-07-05
Chat: update welcome message to be IDE based

abeatrix:
CLOSE: https://linear.app/sourcegraph/issue/CODY-2291/port-chat-ux-vscode-isms-to-jetbrains

This change adds the IDE information (VSCode or other) to the user account info object. This allows the WelcomeMessage component to conditionally render certain features based on the IDE being used. This is because most of the current welcome messages (`New Chat` and `Config`) only applies to VS Code and not other clients:

Screenshot 2024-07-05 at 12 05 04 PM

The changes include:
- Adding the `ide` property to the `UserAccountInfo` interface in `Chat.tsx`
- Passing the `ide` value from the App component to the Chat component
- Updating the `WelcomeMessage` component to conditionally render the "Start a new chat" feature based on the IDE

## Test plan

Verify the following:
- The New Chat and Settings items in the welcome message should display on VS Code

image

- The New Chat and Settings items in the welcome message should not show up in other clients (e.g. JetBrains, web)
  - Tested with Cody Web demo:
  
image

### DIFF
--- a/web/lib/components/CodyWebChat.tsx
+++ b/web/lib/components/CodyWebChat.tsx
@@ -207,7 +207,7 @@ export const CodyWebChat: FC<CodyWebChatProps> = props => {
                                         <Chat
                                             chatID={activeChatID}
                                             chatEnabled={true}
-                                            showWelcomeMessage={true}
+                                            showWelcomeMessage={false}
                                             showIDESnippetActions={false}
                                             userInfo={userAccountInfo}
                                             messageInProgress={messageInProgress}
--- a/web/lib/components/CodyWebChat.tsx
+++ b/web/lib/components/CodyWebChat.tsx
@@ -4,6 +4,7 @@ import { URI } from 'vscode-uri'
 import {
     type ChatMessage,
     type ClientStateForWebview,
+    CodyIDE,
     type ContextItem,
     type ContextItemRepository,
     ContextItemSource,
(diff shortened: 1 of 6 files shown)


web/lib/components/CodyWebChat.tsx
```diff
@@ -205,7 +207,7 @@ export const CodyWebChat: FC<CodyWebChatProps> = props => {
                                         <Chat
                                             chatID={activeChatID}
                                             chatEnabled={true}
-                                            showWelcomeMessage={false}
+                                            showWelcomeMessage={true}
```
abeatrix, line 210:
```suggestion
                                            showWelcomeMessage={false}
```

@vovakulikov we can set it back to false if you think the current welcome message for web won't be helpful

vovakulikov (commented):
LGTM, FIY For now, I disable the Welcome message for Cody Web; you can notice that `showWelcomeMessage` is false in the Cody Web demo. 

I suggest that let's we keep this disabled for now for Cody Web, and when I have time after other issues, I will take a look at Cody Web's welcome message closely. 

I'm glad though that we now have IDE info there, it would be easy for us avoid breaking changes for other clients, thanks

web/lib/components/CodyWebChat.tsx
vovakulikov: Yeah, I think we could do better there, also Welcome message probably should be a bit different from place to place (blob UI side panel and standalone chat page) so I would turn this off for Cody Web for now, I will take a look at this early next week.

vovakulikov (approved):

PR #4798, opened 2024-07-05
E2E: ensure input boxes are visible before filling in auth

abeatrix:
- Adds checks to ensure the correct input boxes are visible before filling in the URL and access token fields during the sign-in flow

This attempted some of the failing E2E tests on main where the filling step is performed before the input box has shown up based on the recordings we got from the failed tests.

Example captured from recordings:

1. After we have input the URL:

image

2. We did not enter the token (because the token step was performed before the input box is ready):

image

3. This means we weren't able to login for the test:

image

Example 2:

image

image

## Test plan

Green CI.


kalanchan (approved):
NICE!

PR #4803, opened 2024-07-07
Add default Anthropic autocomplete model

justinmilner1:
Fixes https://github.com/sourcegraph/cody/issues/4639

There are two 'model' parameters - one was being set to a default when undefined, the other was not. 
This PR adds a default for that undefined parameter.

## Test plan
- Replicated the error (use anthropic setting: "cody.autocomplete.advanced.provider": "anthropic")
- Observe that autocomplete works on Anthropic setting with this addition
- Error described in above bug is not present


justinmilner1:
@valerybugakov Hey Valery - this is my first PR to Cody, looking for a review - thanks! (tagging you because you worked on this section most recently)

vscode/src/completions/providers/anthropic.ts
```diff
@@ -296,7 +296,7 @@ export function createProviderConfig({
                     ...providerOptions,
                     id: PROVIDER_IDENTIFIER,
                 },
-                { maxContextTokens, model, ...otherOptions }
+                { maxContextTokens, model: model ?? 'anthropic/claude-instant-1.2', ...otherOptions }
```
valerybugakov, line 299: Let's update the model value for the `providerConfig` on line 304. 

```ts
const clientModel = model ?? 'anthropic/claude-instant-1.2'

// Use this variable in both places
```

Then, we will need to update tests with the new value: `vscode/src/completions/providers/create-provider.test.ts`

valerybugakov (commented):
Thank you, @justinmilner1! 

I verified that the fix works locally. Could you follow up on the inline comment I left? Then we can merge this.

justinmilner1:
@valerybugakov Made the review suggestion and updated tests. 

Unit tests passed - except for unrelated enterprise level tests.

valerybugakov:
@justinmilner1, thanks, I missed the notification!

valerybugakov (approved):
Thank you! 🙌

PR #4806, opened 2024-07-08
Edit: Set the default model to Claude Sonnet 3.5

umpox:
Closes https://linear.app/sourcegraph/issue/CODY-2549/change-inline-edits-default-model-to-sonnet-3-5

## Description

This PR:
- Updates the inline-edit default model to Claude 3.5 Sonnet
- Adds support for setting different defaults for edit/chat

## Test plan

1. Clear local storage
2. Check the edit model default is 3.5 sonnet
3. Check the chat model default is 3 sonnet


umpox:
Oops, somehow broke the chat input selector... looking into the fix

abeatrix:
@umpox w decided to use 3.5 sonnet as the default model for chat, and updated that in https://github.com/sourcegraph/cody/pull/4822 which also set it to be the default model for Edit.

As @jamesmcnamara  is currently working on using server-side config as the default model list, I think we can close the issue and this PR since there will be some changes on the overall Model class and services.

Thank you so much for your quick work on this 🙇‍♀️

PR #4807, opened 2024-07-08
Ollama: Upgrade Ollama autocompletions client to use SDK

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-1498

- Use the Ollama browser SDK to interact with the Ollama API for autocompletes
- Simplify the completion response handling by leveraging the Ollama SDK's streaming API
- Remove unused error response handling and other legacy code
- Include formatting of Ollama timing information

This fixed the issue where our custom implementation for handling the Ollama stream is not robust enough to handle edge cases. It also helps improve the maintainability of our Ollama clients by leveraging the official Ollama browser SDK for autocomplete instead of using custom implementation for handling streaming responses.

## Test plan

Set up Cody to use Ollama as the autocomplete provider and verify it works as expected:

image


lib/shared/src/llm-providers/ollama/completions-client.ts
```diff
@@ -28,83 +25,54 @@ export function createOllamaClient(
...
-                : browserResponseToAsyncIterable(response.body)
-
-            let insertText = ''
-            let stopReason = ''
-            let buffer = ''
+            for await (const line of res) {
+                if (line.response) {
+                    completionResponse.completion += line.response
+                    yield {
+                        completionResponse,
```
valerybugakov, line 54: Should we keep `stopReason` here?
```suggestion
                    yield {
                        completionResponse,
                        stopReason: CompletionStopReason.StreamingChunk
```

valerybugakov (approved):

PR #4809, opened 2024-07-08
Cody Web: fix cody web remote files link uri

vovakulikov:
Fixes https://linear.app/sourcegraph/issue/SRCH-633/links-in-the-prompt-has-incorrect-url-in-cody-web

This affects only cody web package and has no impact on other clients

## Test plan
- Start a new chat and include files and remote repositories via mentions
- Check that links in the Context list have proper URL

### DIFF
--- a/vscode/src/editor/utils/editor-context.ts
+++ b/vscode/src/editor/utils/editor-context.ts
@@ -426,12 +426,11 @@ async function resolveFileOrSymbolContextItem(
         if (!isErrorLike(resultOrError)) {
             return {
                 ...contextItem,
-                uri: URI.from({
-                    scheme: '',
-                    authority: '',
-                    path: `${repository}${path}`,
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
The code change looks good to me. I thought we have constructed all those information for the remote context items when we first retrieved them. Are they currently being removed by the webview?

vovakulikov:
> when we first retrieved them

If you mean that we built a proper context item when we resolved it and picked it in the mentioned suggestion panel, then yeah, we built a context item there, too, but we constructed URI in a way that this later could be parsed and recognized as a remote file or symbol context item. 

The fix here is actually marking this item as `source: ContextItemSource.Unified` so later in `FileLink.tsx` we could run remote URI building logic. I didn't know about this when I was implementing it for the first time.

PR #4811, opened 2024-07-08
use ClientConfig.modelsAPIEnabled

slimsag:
@jamesmcnamara after https://github.com/sourcegraph/sourcegraph/pull/63697 is merged, and with this PR, you can add the following to your Sourcegraph instance site configuration to opt-in to the new server-side model config system:

```
  "modelConfiguration": {
    "sourcegraph": {
      "pollingInterval": "6h"
    }
  }
```

When you do that, your Sourcegraph instance will begin returning that `modelAPIEnabled` is true:

```
curl -H "Authorization: token $(cat ~/local-token)" https://sourcegraph.test:3443/.api/client-config
{
    "codyEnabled": true,
    "chatEnabled": true,
    "autoCompleteEnabled": true,
    "customCommandsEnabled": true,
    "attributionEnabled": false,
    "smartContextWindowEnabled": true,
    "modelsAPIEnabled": true
}
```

And then we can see the server returns a JSON file with all the models the server should support using (looks like [this](https://gist.github.com/slimsag/1c61ebfe26a5ba13db97eefae1ba06c2) on my machine):

```
curl -H "Authorization: token $(cat ~/local-token)" https://sourcegraph.test:3443/.api/modelconfig/supported-models.json
```

..and since the `client-config` returned `"modelsAPIEnabled": true`, this new codepath for pulling models from the server will run, which you can observe in VS Code's output tab now:

```
    "cody.debug.enable": true,
    "cody.debug.verbose": true,
```

image

And with the now-corrected backend URL, the `ModelsService` seems to be working at least.

I think chrsmith and myself are both still working on other backend changes we will need to use this for real, but this is a step in the right direction and lets us see the current status quo better.

## Test plan

Described above.

This change has zero effect on users unless they opt-in to the new beta `"modelConfiguration":` site config setting on their Sourcegraph instance (which nobody should do yet besides people working on cody.)


jamesmcnamara (approved):
LGTM!

slimsag:
Thanks for the help! Merging

PR #4813, opened 2024-07-09
Autocomplete: decrease requests timeout

valerybugakov:
Reduce the timeout for autocomplete requests from 15 to 7 seconds. More context can be found in [this Slack thread](https://sourcegraph.slack.com/archives/C07AH4RB3H8/p1720007598899309). This is the "reasonable" value for this timeout. Based on data for the last 30 days, 0.07% of all autocomplete requests took longer than 5s. I can bump it to 7s to make this even lower — 0.03%. We expect this to happen during incidents only or on a _very_ bad connection where autocomplete is not usable. 

## Test plan

CI


eseliger (commented):
is there any additional reasoning we can add here? 

on airplane wifi for example, 2.5s latency is not very uncommon, and I'd rather get a completion in 6s than none at all but maybe this all makes sense- just lacking context :)

valerybugakov:
More context can be found in [this Slack thread](https://sourcegraph.slack.com/archives/C07AH4RB3H8/p1720007598899309). This is the "reasonable" value for this timeout. Based on data for the last 30 days, 0.07% of all autocomplete requests took longer than 5s. I can bump it to 7s to make this even lower — 0.03%. We expect this to happen during incidents only or on a _very_ bad connection where autocomplete is not usable.

eseliger (approved):
Oh 5 is fine if you think that’s the right value - I didn’t have any context that the numbers are so low!

PR #4817, opened 2024-07-09
Autocomplete: fix hot-streak cache keys for long documents

valerybugakov:
- We do not account for long documents that hit the document context prefix limit. That causes request manager cache keys to be incorrect for preloaded hot-streak completions. This PR fixes this issue by taking into account `maxPrefixLength`.
- Closes https://linear.app/sourcegraph/issue/CODY-2815/[autocomplete-latency]-fix-cache-keys-for-hot-streak-completions-in

## Test plan

1. Updated unit tests
2. Verified manually in `src/completions/inline-completion-item-provider.ts` by removing code before the `logDebug` in the `logIgnored` function and generating a completion. The first completion should be an if statement and right after it should be the preloaded hot streak completion where we assigned the value to `lasIgnoredUriLogged`. Hot Streak should be enabled: `"cody.autocomplete.experimental.hotStreak": true,`.

### DIFF
--- a/lib/shared/src/completions/types.ts
+++ b/lib/shared/src/completions/types.ts
@@ -31,6 +31,11 @@ export interface DocumentContext extends DocumentDependentContext, LinesContext 
      */
     injectedCompletionText?: string
     positionWithoutInjectedCompletionText?: vscode.Position
+    /**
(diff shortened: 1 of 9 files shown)


umpox (approved):
Nice, makes sense

PR #4821, opened 2024-07-09
adding deepseek-v2 and deepseek fine-tuned model for A/B test

hitesh-1997:
## Context
These are the changes for the autocomplete experiment. Following are the variant details, along with the hypothesis. (cc: @rishabhmehrotra)

1. `Control`: This is the current production model.
2. `Current-best`: The best model we have on autocomplete till now.
3. `Variant-1 `: DeeSeek-v2 lite base.
4. `Variant-2`: Deepseek Fine-tuned model on stack-dedup.
5. `Variant-3`: CodeQwen 1.5 - 7B
6. `Variant-4`: Deepseek Fine-tuned model trained on symbol context.

Offline evaluation: https://docs.google.com/spreadsheets/d/1i0Hz-Db-_ZKT3xAfjKtwTn2amGjKPBONyPCp4iEdUas/edit#gid=0
Backend PR: https://github.com/sourcegraph/sourcegraph/pull/63702

## Test plan
**Local Testing**
1. Start the debugger and Change my user-id among variants.
2. Observe the payload and model we are triggering for each variant combination.

### DIFF
--- a/lib/shared/src/prompt/prompt-string.ts
+++ b/lib/shared/src/prompt/prompt-string.ts
@@ -6,7 +6,7 @@ import type { ContextFiltersProvider } from '../cody-ignore/context-filters-prov
 import type { TerminalOutputArguments } from '../commands/types'
 import { markdownCodeBlockLanguageIDForFilename } from '../common/languages'
 import type { RangeData } from '../common/range'
-import type { AutocompleteContextSnippet, DocumentContext } from '../completions/types'
+import type { AutocompleteContextSnippet, DocumentContext, GitContext } from '../completions/types'
 import type { ConfigGetter } from '../configuration'
 import type { ActiveTextEditorDiagnostic } from '../editor'
 import { createGitDiff } from '../editor/create-git-diff'
@@ -339,6 +339,13 @@ export class PromptString {
                 'symbol' in contextSnippet
                     ? internal_createPromptString(contextSnippet.symbol, ref)
                     : undefined,
+        }
+    }
+
+    public static fromAutocompleteGitContext(gitContext: GitContext, uri: vscode.Uri) {
+        const ref = [uri]
+        return {
+            repoName: internal_createPromptString(gitContext.repoName, ref),
         }
     }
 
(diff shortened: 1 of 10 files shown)


lib/shared/src/prompt/prompt-string.ts
```diff
@@ -342,6 +342,13 @@ export class PromptString {
         }
     }
 
+    public static fromAutocompleteGitContext(gitContext: GitContext, uri: vscode.Uri) {
+        const ref = [uri]
+        return {
+            repoName: internal_createPromptString(gitContext.repoName, ref),
```
janhartman, line 348: Can you inline the [uri] so we don't have a separate const?

vscode/src/completions/providers/fim-prompt-utils.ts
```diff
@@ -61,7 +62,28 @@ export class FinetunedModelV1PromptExtractor implements FIMModelSpecificPromptEx
...
+        }
+        return prompt
+    }
+}
+
+export class CodeQwenModelPromptExtractor implements FIMModelSpecificPromptExtractor {
+    // https://github.com/QwenLM/CodeQwen1.5?tab=readme-ov-file#2-file-level-code-completion-fill-in-the-middle
+
+    getContextPrompt(param: FIMContextPromptParams): PromptString {
+        // Fine-tuned model have a additional <file_sep> tag.
```
janhartman, line 77:
```suggestion
        // Fine-tuned model has a additional <file_sep> tag.
```

janhartman (approved):

lib/shared/src/prompt/prompt-string.ts
hitesh-1997: @janhartman Actually, all the other function in the file use the same pattern of creating this extra const. So, I think should be okay to keep this patter here as well

janhartman: interesting, I don't understand why though. Really doubt there are any performance considerations here.

hitesh-1997: I like the current way though i.e. a variable having the logic for reference: `const ref = [uri] ` and us specifying that variable at other places in function rather than the logic itself. Could come in handy if logic becomes more complicated

umpox (approved):

PR #4822, opened 2024-07-09
Model: Claude 3.5 Sonnet as the default model for chat and commands

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2507/evaluate-claude-35-sonnet-as-new-default

image

Update Claude 3.5 Sonnet to be the default model for chat and commands (including edit).

This change only affects new users as the default model for existing users is the model last used by the user.

## Test plan

- Start Cody from this branch in as a separate instance to micmic new install.
- Verify the default model is set to Claude 3.5 Sonnet for both Chat and Inline-Edit.

### After

image

### Before

image

### DIFF
--- a/lib/shared/src/models/dotcom.ts
+++ b/lib/shared/src/models/dotcom.ts
@@ -13,6 +13,10 @@ const basicContextWindow: ModelContextWindow = {
     input: CHAT_INPUT_TOKEN_BUDGET,
     output: CHAT_OUTPUT_TOKEN_BUDGET,
 }
+
+/**
+ * Has a higher context window with a separate limit for user-context.
+ */
 const expandedContextWindow: ModelContextWindow = {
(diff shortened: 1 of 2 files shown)


kalanchan (approved):
lets go

agent/src/cli/__snapshots__/command-chat.test.ts.snap
```diff
@@ -5,9 +5,11 @@ exports[`--context-file (animal test) 1`] = `
   the code without any explanation.
 exitCode: 0
 stdout: |+
+  Here's the implementation of a cow based on the provided interface:
```
jtibshirani, line 8: I guess it didn't obey the "without any explanation" part. But the `'coo' | 'moo'` part looks positive!

jtibshirani (approved):
Looks good to me.

PR #4824, opened 2024-07-09
refactor: on AuthStatus sync

abeatrix:
refactor and add a `handleAuthStatusChange` function in `main.ts` to properly set the Sourcegraph authentication details based on the current configuration, including the access token.

## Test plan

No feature change, so CI must be green to verify this change keeps the current behaviour.

### DIFF
--- a/vscode/src/commands/services/provider.ts
+++ b/vscode/src/commands/services/provider.ts
@@ -10,6 +10,7 @@ import { CodyCommandMenuItems } from '..'
 import { TreeViewProvider } from '../../services/tree-views/TreeViewProvider'
 import { getContextFileFromGitLog } from '../context/git-log'
 import { getContextFileFromShell } from '../context/shell'
+import { executeExplainHistoryCommand } from '../execute/explain-history'
 import { showCommandMenu } from '../menus'
 import type { CodyCommandArgs } from '../types'
 import { getDefaultCommandsMap } from '../utils/get-commands'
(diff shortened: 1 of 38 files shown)


jamesmcnamara:
I will come back to reviewing these errors once I fix the jetbrains PR

vscode/src/commands/services/provider.ts
```diff
@@ -47,6 +48,12 @@ export class CommandsProvider implements vscode.Disposable {
             this.treeViewProvider.onDidChangeTreeData(() => this.getCustomCommands())
         )
 
+        this.disposables.push(
```
abeatrix, line 51: moved from main.ts since we only want to register it when this provider is available.

vscode/src/main.ts
```diff
@@ -259,88 +258,63 @@ const register = async (
...
         const newConfig = await getFullConfig()
+        // Propagate access token through config
         configWatcher.set(newConfig)
         // Sync auth status to graphqlClient
         graphqlClient.onConfigurationChange(newConfig)
 
         // Refresh server-sent client configuration, as it controls which features the user has
         // access to.
-        ClientConfigSingleton.getInstance().syncAuthStatus(authStatus)
+        await ClientConfigSingleton.getInstance().syncAuthStatus(authStatus)
```
abeatrix, line 272: we currently do not have `await` on main which can cause race condition

vscode/src/main.ts
```diff
@@ -259,88 +258,60 @@ const register = async (
...
 
-        // Set the default prompt mixin on auth status change.
-        await PromptMixin.updateContextPreamble(isExtensionModeDevOrTest || isRunningInsideAgent())
+        symfRunner?.setSourcegraphAuth(authStatus.endpoint, newConfig.accessToken)
 
-        // Re-register expose openctx client
-        await exposeOpenCtxClient(
+        void exposeOpenCtxClient(
             context,
-            initialConfig,
```
abeatrix, line 305: i think this was a mistake as we won't want to re-used the initial config on change

jamesmcnamara (approved):
Nice!

PR #4825, opened 2024-07-09
Agent: Simplify chat model restoration logic

abeatrix:
- Remove unnecessary logic for determining the default model ID
  - We were not using `theModel` value previously
- Set default model ID to empty string if there is no model found as it should be sync again when webview is created
- Simplify the process of deserializing and adding chat messages to the chat model
- Use the provided model ID and chat messages directly to create the chat model

## Test plan

No feature change so CI should be green. The chat/restore test should continues to be green to validate the change.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -2,12 +2,7 @@ import { spawn } from 'node:child_process'
 import path from 'node:path'
 
 import type { Polly, Request } from '@pollyjs/core'
-import {
-    type CodyCommand,
-    getDotComDefaultModels,
-    isWindows,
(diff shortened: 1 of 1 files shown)


abeatrix:

Safe to ignore as we are just replacing the for loop with map().

jamesmcnamara (approved):
Oh nice! I made a similar change in [#4780](https://github.com/sourcegraph/cody/pull/4780/files#diff-db07ce759aaa98edd44511c98530f90f5292f381ecb94637825e8d51bddf5a73R1076) but since we're not merging that just yet, this is a good thing to fix.

PR #4826, opened 2024-07-09
Cody Web: Use deep links only in vscode extension

vovakulikov:
Fixes https://linear.app/sourcegraph/issue/SRCH-633/links-in-the-prompt-has-incorrect-url-in-cody-web

Initially, I thought that I fixed problems with links in this PR https://github.com/sourcegraph/cody/pull/4809 but cody can render links in answers too via the markdown component. This PR simply disables vscode deep links when we render chat UI components for Cody Web case. 

Original logic about link building stays untouched, we provide simpler URL transform only for cody web case, safe to merge. 

## Test plan
- Check that markdown links still work properly in vscode exntesion.

### DIFF
--- a/vscode/webviews/chat/ChatEnvironmentContext.ts
+++ b/vscode/webviews/chat/ChatEnvironmentContext.ts
@@ -1,16 +1,12 @@
+import { CodyIDE } from '@sourcegraph/cody-shared'
 import { createContext, useContext } from 'react'
 
-export enum ChatClientType {
-    Web = 'web',
-    VsCode = 'vscode',
-}
-
 interface ChatEnvironmentContextData {
-    clientType: ChatClientType
+    clientType: CodyIDE
 }
 
 export const ChatEnvironmentContext = createContext<ChatEnvironmentContextData>({
-    clientType: ChatClientType.VsCode,
+    clientType: CodyIDE.VSCode,
 })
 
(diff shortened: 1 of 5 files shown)


vscode/webviews/chat/ChatEnvironmentContext.ts
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext } from 'react'
+
+export enum ChatClientType {
+    Web = 'web',
+    VsCode = 'vscode',
+}
+
+interface ChatEnvironmentContextData {
+    clientType: ChatClientType
```
abeatrix, line 9:
```suggestion
    clientType: CodyIDE
```

nit: Could we use the existing CodyIDE type here? Just to make it easier for people to be able to search `CodyIDE` and knows what needs to be updated when adding a new client :D

abeatrix (approved):
Thought on using the current CodyIDE instead of creating a new enum (ChatClientType) that seems to serve the same purpose?

vscode/webviews/chat/ChatEnvironmentContext.ts
vovakulikov: Good suggestion, just did it in the last commit, thanks

abeatrix (approved):
ty  for the quick change, havent run it locally but the change lgtm!

PR #4827, opened 2024-07-10
Autocomplete: Fix overeagerness in returning in-flight completions

umpox:
## Description

This PR fixes an issue the the `smart-throttle` experiment, where in-flight requests were incorrectly marked as `suggested` completion events. Reducing CAR and affecting other metrics.

Closes https://linear.app/sourcegraph/issue/CODY-2839/[autocomplete-latency]-fix-overeagerness-in-returning-in-flight

## Problem
The `position` and `context` provided to `provideInlineCompletionItems` is not guaranteed to be accurate for the entire duration of the function. This can happen if a user types after the function is called, their position in the line will change, and the prefix we use to build the completion is not guaranteed to be accurate. 

Example: (`export` in the prefix at the time of the request, `export const` in the actual prefix after the request has resolved, incoming completion is `export function`. We need to use the actual prefix to determine that this completion will **not** be visible)

We likely didn't run into this before as we would always cancel any in-flight requests when a new `provideInlineCompletionItems` was called. However, with the `smart-throttle` we want to keep these in-flight requests.  

## Solution

Whenever we call `provideInlineCompletionItems`, we update `latestCompletionRequest`. We can use this to determine if the position has changed since the function was invoked, and update values accordingly

## Test plan

1. Enable smart-throttle
2. Open the autocomplete trace view
3. Check that the suggested/accepted completion counts are as expected

Unit tests


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -385,6 +389,20 @@ export class InlineCompletionItemProvider
                     return null
                 }
 
+                if (!position.isEqual(this.latestCompletionRequest.position)) {
+                    // The latestCompletionRequest has changed since the request was made.
+                    // This is likely due to another completion request starting, and this request staying in-flight.
+                    // We must update the `position`, `context` and associated values
+                    position = this.latestCompletionRequest.position
```
valerybugakov, line 396: Would it make sense to use `vscode.window.activeTextEditor?.selection.active` here to process cases where the position changed but didn't trigger new completion requests?

If it's different from `this.latestCompletionRequest.position` we can probably use `undefined` as `context` to generate the updated `docContext`.

valerybugakov (approved):
Great find! I'm curious to see how it affects other metrics outside of the smart throttle experiment.

vscode/src/completions/inline-completion-item-provider.ts
umpox: Yep, I think so. I've updated to use `vscode.window.activeTextEditor?.selection.active` 👍

umpox: Wait that's not right, let me look

valerybugakov: > Wait that's not right

👀

vscode/src/completions/inline-completion-item-provider.test.ts
```diff
@@ -369,6 +369,113 @@ describe('InlineCompletionItemProvider', () => {
...
+
+            const { document, position: firstPosition } = documentAndPosition(
+                'const foo = █b',
+                'typescript'
+            )
+            const secondPosition = firstPosition.with(firstPosition.line, firstPosition.character + 1)
+
+            const mockGetInlineCompletions = vi
+                .fn()
+                // First respones
```
umpox, line 430:
```suggestion
                // First response
```

agent/src/index.test.ts
```diff
@@ -165,22 +165,21 @@ describe('Agent', () => {
             await client.openFile(bubbleUri)
             const completions = await client.request('autocomplete/execute', {
                 uri: bubbleUri.toString(),
-                position: { line: 1, character: 3 },
+                position: { line: 1, character: 4 },
```
umpox, line 168: With the new logic, these tests failed. The cursor position in `sum.ts` is actually on character 4, so we would ignore these completions as they would not be visible.

From my own testing, this appears to be correct functionality wise. Completions will be inserted where the cursor is, and not shown otherwise. But LMK if you think this is a regression @valerybugakov

valerybugakov: That's a good catch. We might want to allow imprecise matches for current line prefixes consisting only of whitespace. Let's discuss it on the call later today.

Screenshot 2024-07-12 at 10 42 32

umpox: Should still function as before, we're just aware of where the latest cursor position is so we can accurate re-adjust the prefix is needed

umpox:
@valerybugakov I've updated to mark requests as stale from the smart throttle. That makes it a lot easier to reason about and hopefully means we can still benefit (from recycling the result for in-flight requests).

Will merge shortly as well as https://github.com/sourcegraph/cody/pull/4852 for a patch release today

PR #4834, opened 2024-07-11
Autocomplete: fuzzy cache matches

valerybugakov:
This PR adds fuzzy matching to the autocomplete request manager cache. Fuzzy matching allows for approximate cache key matching, which should increase the cache hit ratio for all autocomplete requests. It should be especially useful for hot streak completions where we pre-generate multiple suggestions upfront. Currently, these completions are discarded if users slightly tweak the suggested code, even by adding a semicolon or an extra space. The fuzzy-matcher fixes this inefficiency. 

- We use the Levenshtein distance to measure the similarity between strings.
- The matching threshold is dynamically adjusted based on the length of the string, allowing for more flexibility with longer strings while maintaining precision for shorter ones.
- The fuzzy matching is applied to multiple previous context lines, not just the current line. This provides a more robust matching system for minor changes across numerous lines.
- ~~The new completion source `InlineCompletionsResultSource.FuzzyCache` is added to analyze the impact of this change in production.~~
- The new `isFuzzyMatch` analytics field is added to autocomplete events to analyze the impact of this change in production.

Closes https://linear.app/sourcegraph/issue/CODY-2838/[autocomplete-latency]-implement-fuzzy-matcher-to-autocomplete-cache

## Test plan

Unit tests and manually verified that it plays nicely with hot streak completions (add semicolons or extra lines after the accepted completion and observe that the hot streak completion still matches).

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -826,7 +826,7 @@ function onlyCompletionWidgetSelectionChanged(
     return prevSelectedCompletionInfo.text !== nextSelectedCompletionInfo.text
 }
 
-let lasIgnoredUriLogged: string | undefined = undefined
+let lastIgnoredUriLogged: string | undefined = undefined
 function logIgnored(uri: vscode.Uri, reason: CodyIgnoreType, isManualCompletion: boolean) {
     // Only show a notification for actively triggered autocomplete requests.
     if (isManualCompletion) {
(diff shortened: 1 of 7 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -826,18 +826,18 @@ function onlyCompletionWidgetSelectionChanged(
     return prevSelectedCompletionInfo.text !== nextSelectedCompletionInfo.text
 }
 
-let lasIgnoredUriLogged: string | undefined = undefined
+let lastIgnoredUriLogged: string | undefined = undefined
```
valerybugakov, line 829: unrelated typo

vscode/src/completions/request-manager.ts
```diff
@@ -303,25 +304,124 @@ interface RequestCacheItem {
...
+     */
+    private levenshteinBaseThreshold = 3
+    /**
+     * The maximum allowed Levenshtein distance between lines to match.
+     */
+    private levenshteinMaxThreshold = 5
+    /**
+     * The number of lines to fuzzy match.
+     */
+    private fuzzyPrefixMatchLineCount = 5
```
valerybugakov, line 333: Empirically picked values here. Let me know if you think we should change any of these. Happy to discuss this!

umpox (approved):

PR #4840, opened 2024-07-11
Rg/context api boilerplate

rafax:

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -12,6 +12,7 @@ import { addTraceparent, wrapInActiveSpan } from '../../tracing'
 import { isError } from '../../utils'
 import { DOTCOM_URL, isDotCom } from '../environments'
 import {
+    CHAT_INTENT_QUERY,
     CONTEXT_FILTERS_QUERY,
     CONTEXT_SEARCH_QUERY,
     CURRENT_SITE_CODY_CONFIG_FEATURES,
@@ -36,6 +37,8 @@ import {
     LOG_EVENT_MUTATION,
     LOG_EVENT_MUTATION_DEPRECATED,
     PACKAGE_LIST_QUERY,
+    RANK_CONTEXT_QUERY,
+    RECORD_CONTEXT_QUERY,
     RECORD_TELEMETRY_EVENTS_MUTATION,
     REPOSITORY_IDS_QUERY,
     REPOSITORY_ID_QUERY,
@@ -284,6 +287,21 @@ interface SearchAttributionResponse {
 
 type LogEventResponse = unknown
 
+interface ChatIntentResponse {
+    chatIntent: {
+        intent: string
+        score: number
+    }
+}
+
+type RecordContextResponse = unknown
+
+interface RankContextResponse {
+    ranker: string
+    used: number[]
+    discarded: number[]
+}
+
 interface ContextSearchResponse {
     getCodyContext: {
         blob: {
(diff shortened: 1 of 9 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -316,6 +319,23 @@ interface SearchAttributionResponse {
 
 type LogEventResponse = unknown
 
+interface ChatIntentResponse {
+    chatIntent: {
```
janhartman, line 323: do we need these outer objects? (same in `RankContextResponse`)

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -344,6 +364,20 @@ export interface EmbeddingsSearchResult {
...
 
+export interface ChatIntentResult {
+    intent: string
+    score: number
+}
+
+/**
+ * Experimental API.
+ */
+export interface InputContextItem {
```
janhartman, line 375: if you don't put `Input` in the name do we have a conflict with another type? It seems a bit awkwardly named TBH.

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -710,6 +716,11 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
                                   },
                                   contextRanking: this.contextRanking,
                               })
+                              // add a callback, but return the original context
```
janhartman, line 719: clever 👍

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -847,6 +881,52 @@ export class SourcegraphGraphQLAPIClient {
...
+            }
+        )
+        return extractDataOrError(response, data => data.chatIntent)
+    }
+
+    /** Experimental API */
+    public async recordContext(
+        interactionID: string,
+        used: InputContextItem[],
+        unused: InputContextItem[]
```
janhartman, line 900: Can we unify the terminology here? We have unused, discarded and I've also seen ignored in the codebase 😄

vscode/src/chat/context/contextAPIClient.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export class ContextAPIClient {
+    constructor(
+        private readonly apiClient: SourcegraphGraphQLAPIClient,
+        private readonly featureFlagProvider: FeatureFlagProvider
+    ) {}
+
+    public detectChatIntent(interactionID: string, query: string) {
+        if (!this.featureFlagProvider.evaluateFeatureFlag(FeatureFlag.CodyServerSideContextAPI)) {
```
janhartman, line 31: don't you need to await here?

lib/shared/src/sourcegraph-api/graphql/client.ts
rafax: We have a clash between a `type ContextItem` and `input ContextItem` in GraphQL, but both of those are new types, so if you think that makes sense, I can use `input ContextItem` (i.e. this one) and something else for output (ContextItemResult? CodyContextItem? ContextDocument?)

lib/shared/src/sourcegraph-api/graphql/client.ts
rafax: Ok, will need to send a backend PR too.

vscode/src/chat/context/contextAPIClient.ts
rafax: Nice catch :)

lib/shared/src/sourcegraph-api/graphql/client.ts
rafax: https://github.com/sourcegraph/sourcegraph/pull/63903/files

rafax: Using ignored across the board (that's what the existing Context code uses)

lib/shared/src/sourcegraph-api/graphql/client.ts
rafax: I think this is caused by how GraphQL serializes stuff (it has it's own outer wrapper with a single property, and we unpack the inner response?

janhartman: ah, didn't know that.

janhartman (approved):

PR #4841, opened 2024-07-11
Allow clients to redefine untitled files protocol during their creation

pkukielka:
Closes https://linear.app/sourcegraph/issue/CODY-2004/bug-cody-recreates-the-file-that-was-created-by-generate-unit-tests

## Changes

Some IDEs (maybe even most IDEs except VSC) does not have notion of unsaved files.
Keeping two different models in IDE and in the agent is problematic and leads to hard to track errors.
It is easier to let client be honest and notify agent about real protocol which it is using (most likely `file://` instead of `untitled://`).

As a result that change slightly modifies behaviour of the `openTextDocument`. 
One could now call `openTextDocument` with `untitled://abc` parameter, and as a result get a `TextDocument` which uri points to `file://abc` (or any other path, really).

## Test plan

1. Build [JetBrains branch ](https://github.com/sourcegraph/jetbrains/pull/1903)with this PR using CODY_DIR env var
2. Ask cody to generate tests for some file which does not have tests yet
3. Accept the changes
4. Delete that newly created file
5. Perform autocomplete in some other file
6. Make sure test file was not recreated

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -487,7 +487,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
 
         this.registerNotification('textDocument/didSave', async params => {
             const uri = vscode.Uri.parse(params.uri)
-            const document = await vscode.workspace.openTextDocument(uri)
+            const document = await this.workspace.openTextDocument(uri)
             vscode_shim.onDidSaveTextDocument.fire(document)
         })
 
@@ -1317,6 +1317,27 @@ export class Agent extends MessageHandler implements ExtensionClient {
     ): FixupControlApplicator {
         this.fixups = new AgentFixupControls(files, this.notify.bind(this))
         return this.fixups
+    }
+
+    public openNewDocument = async (
+        _: typeof vscode.workspace,
+        uri: vscode.Uri
+    ): Promise<vscode.TextDocument | undefined> => {
+        if (uri.scheme !== 'untitled') {
+            return vscode_shim.workspace.openTextDocument(uri)
+        }
+
+        if (this.clientInfo?.capabilities?.untitledDocuments !== 'enabled') {
+            const errorMessage =
+                'Client does not support untitled documents. To fix this problem, set `untitledDocuments: "enabled"` in client capabilities'
+            logError('Agent', 'unsupported operation', errorMessage)
+            throw new Error(errorMessage)
+        }
+
+        const result = await this.request('textDocument/openUntitledDocument', {
+            uri: uri.toString(),
+        })
+        return result ? vscode_shim.workspace.openTextDocument(result.uri) : undefined
     }
 
     private maybeExtension: ExtensionObjects | undefined
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -301,16 +301,14 @@ const _workspace: typeof vscode.workspace = {
             throw new Error('workspaceDocuments is uninitialized')
         }
 
-        const result = toUri(uriOrString)
-        if (result) {
-            if (result.uri.scheme === 'untitled' && result.shouldOpenInClient) {
-                await openUntitledDocument(result.uri)
-            }
-            return workspaceDocuments.openTextDocument(result.uri)
-        }
-        return Promise.reject(
-            new Error(`workspace.openTextDocument:unsupported argument ${JSON.stringify(uriOrString)}`)
-        )
+        const uri = toUri(uriOrString)
+        return uri
+            ? workspaceDocuments.openTextDocument(uri)
+            : Promise.reject(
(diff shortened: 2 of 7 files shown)


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -294,7 +294,7 @@ export type ServerRequests = {
     'window/showMessage': [ShowWindowMessageParams, string | null]
 
     'textDocument/edit': [TextDocumentEditParams, boolean]
-    'textDocument/openUntitledDocument': [UntitledTextDocument, boolean]
+    'textDocument/openUntitledDocument': [UntitledTextDocument, ProtocolTextDocument | undefined | null]
```
mkondratek, line 297: is it something normal to have both undefined and null as a possible return type? for me it looks weird

agent/src/vscode-shim.ts
```diff
@@ -463,32 +469,25 @@ const defaultTreeView: vscode.TreeView<any> = {
...
         ((uriOrString as any)?.language || (uriOrString as any)?.content)
     ) {
         const language = (uriOrString as any)?.language ?? ''
         const extension = extensionForLanguage(language) ?? language
-        return {
-            uri: Uri.from({
-                scheme: 'untitled',
-                path: `${uuid.v4()}.${extension}`,
-            }),
-            shouldOpenInClient: false,
```
mkondratek, line 490: nice 🚀

mkondratek (approved):
works and looks good

agent/src/agent.ts
```diff
@@ -482,7 +482,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
 
         this.registerNotification('textDocument/didSave', async params => {
             const uri = vscode.Uri.parse(params.uri)
-            const document = await vscode.workspace.openTextDocument(uri)
+            const document = await this.workspace.openTextDocument(uri)
```
pkukielka, line 485: Unrelated change but it was looking like bug to me.

vscode/src/jsonrpc/agent-protocol.ts
pkukielka: It's typical declaration in our protocol, please look at the rest of the file.

agent/src/agent.ts
```diff
@@ -1288,6 +1288,27 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+        }
+
+        if (this.clientInfo?.capabilities?.untitledDocuments !== 'enabled') {
+            const errorMessage =
+                'Client does not support untitled documents. To fix this problem, set `untitledDocuments: "enabled"` in client capabilities'
+            logError('Agent', 'unsupported operation', errorMessage)
+            throw new Error(errorMessage)
+        }
+
+        const result = await this.request('textDocument/openUntitledDocument', {
```
pkukielka, line 1306: Previously this call was in the `openTextDocument`, I think now it is in much more logical and better isolated place (`openTextDocument` should not have side effects like creating new files, and we cannot guarantee that for `openUntitledDocument`).

agent/src/vscode-shim.ts
```diff
@@ -301,16 +301,14 @@ const _workspace: typeof vscode.workspace = {
             throw new Error('workspaceDocuments is uninitialized')
         }
 
-        const result = toUri(uriOrString)
-        if (result) {
-            if (result.uri.scheme === 'untitled' && result.shouldOpenInClient) {
```
pkukielka, line 306: That was source of some of our problems.

vscode/src/non-stop/FixupController.ts
```diff
@@ -978,16 +978,20 @@ export class FixupController
         }
 
         // append response to new file
-        const doc = await vscode.workspace.openTextDocument(newFileUri)
+        const doc = await this.client.openNewDocument(vscode.workspace, newFileUri)
+        if (!doc) {
+            throw new Error(`Cannot create file for the fixup: ${newFileUri.toString()}`)
+        }
```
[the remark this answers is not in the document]
pkukielka: We do not need to use it for all untitled files (even more, we shouldn't), only files which we want to present to the users. I do not think there is right now any other such use case other than edits.

Problem is fundamental:
Jetbrains do not have `untitled` (or unsaved) files support.
They need to be represented as normal files.
And unless we want to do a manual file mapping for every URI used in the protocol, and send request to IntelliJ for every `openTextDocument` call, this is best options.

I'm very open to discussion though. If you would like to jump on the call @umpox I can explain all corner cases.

umpox (approved):
Approving on the basis that this works and fixes a bug.

Will follow up when I'm back from OOO on if we can make this more robust for future usage too

umpox (approved):

PR #4844, opened 2024-07-11
Chat: update chat model selection and default model on auth sync

abeatrix:
This PR fixes an issue where submitting a question in sidebar chat view does not display the correct icon associated with the selected model as the model ID was not updated in the ChatModel previously on auth change because the chat provider can only be created when the user is logged in, which is no longer the case for the sidebar chat view.

Changes:
- Update the ChatModel based on the available chat model list for the current user on auth status change.

## Test plan

2. In the sidebar chat, send Cody a question without changing the model
3. Verify the response from Cody is showing the logo associated with the model shown in your model dropdown list.

### After

Screenshot 2024-07-11 at 9 54 46 AM

### Before

The response from Cody is not showing the correct logo associated with the model shown in the chat input box. Hovering over the logo shows that the model ID is "{pending}":

image

### DIFF
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -533,6 +533,14 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
         // Run this async because this method may be called during initialization
         // and awaiting on this.postMessage may result in a deadlock
         void this.sendConfig()
+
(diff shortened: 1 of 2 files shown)


jamesmcnamara (approved):
Makes sense to me. I'll need to make the corresponding changes over [here](https://github.com/sourcegraph/cody/pull/4780)

PR #4845, opened 2024-07-11
edit: quick pick option for full file for range

keegancsmith:
I find myself quite often wanting to run edit on the whole file. Currently I select the whole file and then use that selection range. However, it seems useful to include it as one of the quick pick options.

Note: I included this option in any place we could include a selection. I did not exclude this option if the full file was already selected, since that seemed like extra complications.

This idea is kind of like @steveyegge CHOP idea, except you skip needing to use copy + paste from chat as well as getting built in diff view.

## Test Plan

Triggered an edit and selected the full file as the range. See video

image

https://github.com/sourcegraph/cody/assets/187831/503d886e-372c-4bd3-8962-80351608aedb


vscode/src/edit/input/get-items/constants.ts
```diff
@@ -15,3 +15,8 @@ export const EXPANDED_RANGE_ITEM: vscode.QuickPickItem = {
     label: '$(file-code) Nearest Code Block',
     alwaysShow: true,
 }
+
+export const FULL_RANGE_ITEM: vscode.QuickPickItem = {
+    label: '$(file) Full File',
+    alwaysShow: true,
+}
```
keegancsmith, line 22: @toolmantim advice for copy and icon? See screenshot in PR description.

abeatrix: Should we also show a warning if the full file exceeds the context window limit?

umpox: Could we just not show this option if the file is too large?

We have some logic for `@` mentions that has the additional functionality of selecting lines/symbols. I think that's probably out of scope here.
image

@keegancsmith You can pass `activeModelContextWindow` from `getInput` to determine if the file is too big

keegancsmith: > Should we also show a warning if the full file exceeds the context window limit?

Agreed, will take a look.

> Could we just not show this option if the file is too large?

If you are used to seeing the option, it will feel like a bug if it just disappears. Maybe the better UX here is having it as an unselectable option with the reason why?

> We have some logic for `@` mentions that has the additional functionality of selecting lines/symbols. I think that's probably out of scope here.

That would be cool, but yeah a much bigger change. I do quite like the current behaviour which has the obvious choices, so would want to make sure we don't regress on that.

> You can pass `activeModelContextWindow` from `getInput` to determine if the file is too big

Thanks!

keegancsmith: Done
image

vscode/src/edit/input/get-items/range.ts
```diff
@@ -22,6 +27,11 @@ const getDefaultRangeItems = (
         range: new vscode.Range(initialCursorPosition, initialCursorPosition),
     }
 
+    const fullItem = {
+        ...FULL_RANGE_ITEM,
+        range: new vscode.Range(0, 0, document.lineCount + 1, 0),
+    }
```
umpox, line 33:
```suggestion
    const fullItem = {
        ...FULL_RANGE_ITEM,
        range: new vscode.Range(0, 0, document.lineCount - 1, 0),
    }
```

Otherwise will throw an error I think

umpox (approved):
Nice, feels awesome to use with 3.5 Sonnet

abeatrix (approved):

vscode/src/edit/input/get-items/range.ts
keegancsmith: Thanks this was really easy to reproduce the issue. Apologies, you shouldn't have to do more testing than me :) Fix pushed.

PR #4848, opened 2024-07-11
Refactor: simplify chat controller logic

beyang:
This implements the following refactor:
* Merge `ChatManager` and `ChatPanelsManager` into a single class, `ChatsController`
* Remove some calls to `setContext` that don't do anything
* Clean up and deduplicate methods in `ChatsController`

The resulting code is easier to understand: `ChatsController` holds references to `ChatController` instances for each chat session in the sidebar and in editor panels.

## Test plan

No functional changes. Verified locally.


vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -0,0 +1,479 @@
...
+            vscode.window.registerWebviewViewProvider('cody.chat', this.panel, {
+                webviewOptions: { retainContextWhenHidden: true },
+            })
+        )
+
+        const debouncedRestoreToEditor = debounce(
+            async (chatID: string, chatQuestion?: string) => {
+                try {
+                    logDebug('ChatsController', 'debouncedRestorePanel')
+                    await this.getOrCreateEditorChatController(chatID, chatQuestion)
```
abeatrix:
```suggestion
                    return this.getOrCreateEditorChatController(chatID, chatQuestion)
```

I think we will need to return this for agent

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -0,0 +1,479 @@
...
+                vscode.commands.executeCommand('cody.chat.focus')
+            ),
+            vscode.commands.registerCommand('cody.chat.signIn', () =>
+                vscode.commands.executeCommand('cody.chat.focus')
+            ),
+            vscode.commands.registerCommand('cody.chat.newPanel', async () => {
+                await this.panel.clearAndRestartSession()
+                await vscode.commands.executeCommand('cody.chat.focus')
+            }),
+            vscode.commands.registerCommand('cody.chat.newEditorPanel', debouncedCreateNewWebviewPanel),
```
abeatrix: I saw the agent tests are failing because of this command but haven't looked into the cause yet. Making the following change fixed the error for me though so something about the debounce logic?

```suggestion
            vscode.commands.registerCommand('cody.chat.newEditorPanel', () =>
                this.getOrCreateEditorChatController()
            ),
```

abeatrix (commented):
I have't finished reviewing the code yet (will continue tomorrow) but I left a comment inline on the command that is causing the agent test to fail. 

Seems like the recordings need to be updated too.

abeatrix (approved):
looks like the previous issues are fixed 👍

sqs (approved):

PR #4849, opened 2024-07-12
VS Code: make config features getter sync and properly dispose scheduled fetches

valerybugakov:
I noticed the async call on the critical path of the autocomplete pipeline and went to check if it slows us down. It turns out it can be made synchronous, which this PR does. Plus, it adds logic to clean `setInterval` calls properly.

## Test plan

CI


valerybugakov:
Nvm, I was on the older version of the main branch

PR #4854, opened 2024-07-12
fix(local_e2e): diagnose properly access token issues

jhchabran:
As I was about to record a loom video showcasing this, I noticed that when the access token is rejected, it wasn't handle as well as it should; which is now fixed. Also adjust the docs a bit. 

Before:

CleanShot 2024-07-12 at 15 08 02@2x

After: 

CleanShot 2024-07-12 at 15 09 16@2x

## Test plan

CI + local

### DIFF
--- a/agent/src/local-e2e/README.md
+++ b/agent/src/local-e2e/README.md
@@ -56,6 +56,7 @@ instance is running, saving that file will reload the instance.
 ## Running the tests 
 
 1. In your Sourcegraph folder, run `sg start dotcom-cody-e2e`.
+  1. To ensure the default site-admin is available (mandatory), still in the Sourcegraph folder you can run `sg db default-site-admin`.
 1. Back to the Cody repo, run `pnpm run test:local-e2e`
 
 ## FAQ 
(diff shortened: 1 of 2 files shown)


agent/src/local-e2e/README.md
```diff
@@ -86,3 +87,8 @@ configurable? Please remember that sane defaults that works out of the box are r
 
 Perhaps try resetting the Redis instance, i.e. in the Sourcegraph folder, run `sg db reset-redis` and 
 try again.
+
+### What does `sg db default-site-admin` does exactly?
```
burmudar, line 91:
```suggestion
### What does `sg db default-site-admin` do exactly? 
```

agent/src/local-e2e/helpers.ts
```diff
@@ -247,11 +247,19 @@ export class LocalSGInstance {
...
                     if (isError(resp)) {
-                        resolve({
-                            reason: `Failed to connect to Sourcegraph instance: ${resp.message} `,
-                            fix: 'Something went wrong, if you have a running `sg start dotcom-cody-e2e`, you could try to restart it',
-                        })
-                        return
+                        if (resp.message.includes("401")) {
+                            resolve({
+                                reason: `Sourcegraph instance rejected our access token: ${resp.message}`,
+                                fix: 'In your sourcegraph folder, run "sg db default-site-admin".'
```
burmudar, line 253:
```suggestion
                                fix: 'In your sourcegraph folder, run `sg db default-site-admin`.'
```
nit: just making it consistent with command fix below

burmudar (approved):

BolajiOlajide (approved):

PR #4860, opened 2024-07-12
Command : remove "Ask Cody to Explain" from command palette

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2861/vscode-outdated-cody-command-ask-cody-to-explain-throws-error

Current issue:

image

The command is actually a valid command, but it only works when you highlight some text in your terminal and right click on it:
image

The command is not expected to show up in the command palette menu, which is now fixed by this PR by setting the when condition to `false` for the command in the menu field.

## Test plan

Confirm the command is still available when you select some text in your terminal and right-click:

Screenshot 2024-07-12 at 11 59 46 AM

Confirm the command is no longer available in the command palette menu:

Screenshot 2024-07-12 at 12 00 01 PM

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -10,6 +10,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


morgangauth (approved):
LGTM - thank you for getting on this so quickly!

PR #4861, opened 2024-07-13
changed/enterprise: disabled downloading of BFG and Symf for enterprise users

jamesmcnamara:
Instead of eagerly starting the Symf and LocalEmbeddings services on startup, we subscribe them both to the auth provider and only start them (or provide an instance of them if called explicitly) if the user has authenticated as a dotcom user. Otherwise we skip them.

As I got into trying to test this, I bit off a little more than I could chew and ended up refactoring chat-view/context.ts pretty heavily to clean it up and replace a lot of the copy-pasta. However when I started testing the refactor, I realized it was going to be a heavy lift (at least another day) to generate a good test suite to validate that my refactor didn't break anything. So I reverted it back to where we are currently where it's just the simple update to not do the downloads. 

I'm kind of in limbo on this because I'd love to complete the refactor and add some test coverage for that file (it looks very high touch and complex so an ideal test candidate) but I don't know how long I should spend on this.

Oh I also replaced a bunch of `null` with `undefined`. Not out of any particular zealotry, just that it was cleaner in the code I was touching and then it felt weird to have a bunch of `prop1?: Type1` next to `prop2: Type2 | null`. If there's a specifically good reason to be using `| null` I can roll everything back. 

## Test plan
Green CI


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -211,9 +210,6 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
             TestSupport.instance.chatPanelProvider.set(this)
         }
 
-        // Advise local embeddings to start up if necessary.
-        void this.localEmbeddings?.start()
-
```
jamesmcnamara, line 216: This is now handled on the auth status chanted cycle.

lib/shared/src/auth/types.ts
```diff
@@ -130,3 +130,7 @@ export function isFreeUser(authStatus: AuthStatus): boolean {
 export function isEnterpriseUser(authStatus: AuthStatus): boolean {
     return !authStatus.isDotCom
 }
+
+export function isAuthenticated(authStatus: AuthStatus): boolean {
```
dominiccooney, line 134: This isn't adding much beyond `authStatus.authenticated`, why not simply rename *authenticated* to *isAuthenticated* and be done with it.

You could make *isEnterpriseUser, isFreeUser, isCodyProUser* a getter that returns an enum while you're at it. This will be better than boolean getters because people who want a specific tier can simply test the value with `===`, and people who handle multiple tiers will get compile time errors when we add a tier that is not handled.

vscode/src/chat/chat-view/ChatController.ts
dominiccooney: @rafax FYI

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -424,9 +424,9 @@ export class ChatsController implements vscode.Disposable {
         return new ChatController({
             ...this.options,
             chatClient: this.chatClient,
-            localEmbeddings: isConsumer ? this.localEmbeddings : null,
-            contextRanking: isConsumer ? this.contextRanking : null,
-            symf: isConsumer ? this.symf : null,
+            localEmbeddings: this.localEmbeddings,
+            contextRanking: isConsumer ? this.contextRanking : undefined,
```
dominiccooney, line 428: This also uses context engine, so it is confusing that some of these use `isConsumer` and others don't.

vscode/src/chat/chat-view/context.ts
```diff
@@ -45,11 +45,11 @@ interface GetEnhancedContextOptions {
...
     providers: {
-        localEmbeddings: LocalEmbeddingsController | null
-        symf: SymfRunner | null
-        remoteSearch: RemoteSearch | null
+        symf: SymfWrapper
+        localEmbeddings?: LocalEmbeddingsController
+        remoteSearch?: RemoteSearch
     }
-    contextRanking: ContextRankingController | null
+    contextRanking?: ContextRankingController
```
dominiccooney, line 52: I'm unsure what the thinking with `null` here was, because `undefined` is more idiomatic.

Is it right to make this `?` and not `| undefined`? The intent may be to make callsites explicit about passing the missing value.

vscode/src/chat/chat-view/context.ts
```diff
@@ -310,8 +316,8 @@ async function searchSymf(
 }
 
 async function searchEmbeddingsLocal(
-    localEmbeddings: LocalEmbeddingsController | null,
     text: PromptString,
+    localEmbeddings?: LocalEmbeddingsController,
```
dominiccooney, line 320: It's better to have the object being operated on in the first position, because then `m(o, x, y)` rhymes with `o.m(x, y)` a bit. `m(x, o, y)` scrambles the order up quite a bit.

dominiccooney (changes requested):
High level, I understand what you're doing here:

"is symf enabled" is a flag, it is later combined with "is consumer" to control whether it is passed to the context system.

Some similar observations about local embeddings.

You've added something which more often passes an object, which adds a level of indirection between setting the property, and providing the runner.

But symf is already lazily downloaded (...see getSymfPath) so it totality these seems wrong to me:

- We have a condition evaluated early, the feature flag.
- We have a condition evaluated when the account type changes, the `isConsumer` setup for the context system.
- We have lazy download at the point of use.

Seems to me adding another layer on top of this is not the work we should be doing. Something in the enterprise SKU is touching symf (maybe the defunct local search), and we don't want to use symf for that. So go clean up that code.

jamesmcnamara:
@dominiccooney From what I've seen, the issue is that they're not really lazy, for instance symf has [this](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=jsm%2Frm-tag&file=vscode%2Fsrc%2Flocal-context%2Fsymf.ts&editor=VSCode&version=2.2.17&start_row=85&start_col=33&end_row=85&end_col=33) in the constructor which calls [this ](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=jsm%2Frm-tag&file=vscode%2Fsrc%2Flocal-context%2Fsymf.ts&editor=VSCode&version=2.2.17&start_row=710&start_col=0&end_row=720&end_col=0) that immediately starts indexing the local workspaces ahead-of-time. There's a similar such setup in LocalEmbeddings. So while they do lazy downloading, both try to eagerly compute indexes before any task has been assigned.

We could remove these eager evaluations entirely (I'm not sure what impact that would have on correctness, hopefully none), but we'd still have to wire through the plumbing of pulling and checking auth status at each time of use and ensuring every public method of those objects does that check correctly before use. 

To me it still feels like we want two layered interface: 
1. The public interface which can synthesize synchronous and asynchronous information to determine if the widget is usable 
2. The private inner class which actually does the thing

I think we can and should consolidate all of the conditionals into 1) but inlining the two classes together feels like we could easily end up making this mistake again.

vscode/src/local-context/symf/symf-wrapper.ts
```diff
@@ -0,0 +1,21 @@
...
+import type { SymfRunner } from './symf-runner'
+
+export class SymfWrapper implements vscode.Disposable {
+    public runner: SymfRunner | undefined
+    private hasRun = false
+
+    constructor(private ctor: () => SymfRunner | undefined) {}
+
+    public syncAuthStatus(status: AuthStatus) {
+        if (!this.hasRun && getAuthTier(status) === AuthTier.Enterprise) {
```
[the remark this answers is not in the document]
olafurpg: This method doesn't seem used anywhere? 
CleanShot 2024-08-07 at 13 14 46@2x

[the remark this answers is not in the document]
olafurpg: On commit ebf1c3ef75e85b4844d4ea989e355473138ab75f, I'm unable to get symf working for Cody Pro accounts on a repo that has been indexed locally. To reproduce:

* Open a random repo like Shopify/liquid with a Cody Pro account. Confirm that it indexes the codebase with symf, submit a chat question "What is liquid?" with `@liquid` repo context and confirm that it explains the template library in the repo.
* Repeat the same steps with a local build in this PR and confirm that the same question returns a generic answer about the term "liquid".

olafurpg (changes requested):
It seems like this PR breaks symf keyword search for Cody Pro users. See comment on SymfWrapper with more details.

vscode/src/local-context/symf/symf-wrapper.ts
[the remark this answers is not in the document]
olafurpg: Worth noting that we have no agent integration tests for symf anymore so these kinds of regressions can only be caught with manual testing (or e2e vscode tests if we have them). Symf has been the source of too many flaky test failures that we ended up disabling it altogether in the agent integration tests.

olafurpg:
To simplify testing this locally, I took a stab at separating agent integration tests for enterprise accounts into separate files (one for demo and one for s2) https://github.com/sourcegraph/cody/pull/5127. You can rebase on main to get the changes.

PR #4864, opened 2024-07-15
chat: template chips for prompts

keegancsmith:
We introduce experimental rendering of template chips in prompts. This is only turned on for "Generate Unit Tests (experimental)" button, but in the future can be for everyone.

Right now the template syntax is "{{ placeholder text }}". I am assuming this will change before it is available to other users. For example right now custom commands have hardcoded "{repo}"/etc templates. When we ship prompt templates, we likely want something nice here that mixes the ideas of values that cody can automatically infer vs what users need to input.

The UX for this is quite barebones right now, mainly so that we can follow-up and make it better. This PR is mostly about the plumbing and making sure everything fits together.

One thing to note is the serialization of editorState now changes due to the new template input chips. We have made this backwards compatible. Additionally, because this feature is experimental, we have taken care to only serialize in the new format if the editor state has template chips in it.

### Test Plan

We added a fixture for template chips. We then did manual testing inside of storybook via the fixture's transcript template. Additionally manual testing of the generate unit test feature.

image

Fixes CODY-2800


lib/shared/src/lexicalEditor/editorState.ts
```diff
@@ -240,7 +246,20 @@ export function editorStateToText(editorState: EditorState): string {
...
+type SupportedSerializedNodes =
+    | SerializedTextNode
+    | SerializedContextItemMentionNode
+    | SerializedTemplateInputNode
+
+export function lexicalEditorStateFromPromptString(
+    input: PromptString,
+    opts?: {
+        /**
+         * Experimental support for template values. These are placeholder values between "{{" and "}}".
```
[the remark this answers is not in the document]
thenamankumar: Sure.

lib/shared/src/lexicalEditor/editorState.ts
```diff
@@ -41,6 +44,11 @@ export function toSerializedPromptEditorValue(editor: LexicalEditor): Serialized
     }
 }
 
+/**
+ * This type encodes all known versions of serialized editor state.
+ */
+type StateVersion = 'lexical-v0' | 'lexical-v1'
```
keegancsmith, line 50: @sqs I introduced a new version here, would appreciate your thoughts here. I made it so that it only uses the new version if the new node type is in the state. I think one other change I still need to make is including 'text' in the new node type so that older versions can understand the newer messages in the case a user downgrades.

sqs: This makes sense to me.

sqs (approved):
awesome

PR #4865, opened 2024-07-15
Autocomplete: chars logger account for multiline deletions

valerybugakov:
In the previous version, the chars logger didn't account for multiline deletions. It reported one deleted char even if multiple lines were deleted in one go. This PR fixes it by relying more on VS Code API which has `change.rangeLength` field, which can be used to track multiline deletions. See comments added inline that explain its behavior.

## Test plan

Updated unit tests.


umpox (approved):
Nice

PR #4866, opened 2024-07-15
Autocomplete: Reuse logId for recycled requests via `CacheAfterRequestStart`

umpox:
### Problem

Currently we give each new completion a unique `logId` regardless of it is coming from the cache or not. This is problematic because these completions are effectively the same, the `insertText` and combined prefix are (almost) the same.

In a scenario where we have many in-flight requests (e.g. the smart throttle), there is a much greater chance that we will recycle a response to auto-resolve existing in-flight requests. E.g. if the first response matches the current prefix of any other in-flight requests. We only show a single completion to the user, but we treat this as many completions logged, as each new cache response is resolved with a new `logId`

### Solution

If a request is returned from the cache, we should reuse the `logId` for the purposes of our telemetry

## Test plan

Unit tests

Manual testing with the smart throttle enabled


abeatrix (approved):

umpox:
@valerybugakov I noticed we don't really have good full E2E tests from autocomplete (from network response to final output). It'd be really useful to have this so we don't have various tests where we are testing `RequestManager` in isolation, or `provideInlineCompletionItems` with `getInlineCompletions` mocked. LMK if you think we have full e2e tests somewhere, but I couldn't see any.

Thinking of using the Agent tests to allow us to easily do that, as we can just use network recordings and check the output. PR to clean these up here: https://github.com/sourcegraph/cody/pull/4870

I will see if I can use these to easily add some new tests

valerybugakov (approved):

PR #4872, opened 2024-07-15
Web: update chat model handling

abeatrix:
- Add a comment to clarify that the default chat model is always the first one in the list
- Refactor chat model update logic to notify the host about manual changes, and let the host return the updated chat models

## Test plan

Green CI, as there should not be any behaviour change.

### DIFF
--- a/web/lib/components/CodyWebChat.tsx
+++ b/web/lib/components/CodyWebChat.tsx
@@ -94,6 +94,7 @@ export const CodyWebChat: FC<CodyWebChatProps> = props => {
                     setIsTranscriptError(message.isTranscriptError)
                     break
                 case 'chatModels':
+                    // The default model will always be the first one on the list.
                     setChatModels(message.models)
                     break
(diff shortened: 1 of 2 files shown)


web/lib/components/CodyWebChat.tsx
```diff
@@ -141,14 +142,12 @@ export const CodyWebChat: FC<CodyWebChatProps> = props => {
...
                 return
             }
+            // Notify the host about the manual change,
+            // and the host will return the updated change models via onMessage.
             vscodeAPI.postMessage({
                 command: 'chatModel',
                 model: selected.model,
             })
-            const updatedChatModels = chatModels.map(m =>
-                m.model === selected.model ? { ...m, default: true } : { ...m, default: false }
```
abeatrix, line 149: we have removed the "default" view from the Model class and is no longer used in webview.

vovakulikov (approved):
Thanks for the change!

PR #4873, opened 2024-07-15
Chat: remove isFreeUser model check and simplify model selection

abeatrix:
Right now we are limiting FreeUser to use first model only which is incorrect.  Removing the check fixed the issue where non-Cody Pro users are unable to use other free models

## Test plan

Build from this branch and log in to Cody as Cody Free user, try selecting a model

### Demo

I can select a model that's not 3.5 Sonnet:

image

Also tried logging in as dotcom pro users to confirm the model selection is working as expected:

image

### DIFF
--- a/lib/shared/src/models/index.ts
+++ b/lib/shared/src/models/index.ts
@@ -1,4 +1,4 @@
-import { type AuthStatus, isCodyProUser, isEnterpriseUser, isFreeUser } from '../auth/types'
+import { type AuthStatus, isCodyProUser, isEnterpriseUser } from '../auth/types'
 import { fetchLocalOllamaModels } from '../llm-providers/ollama/utils'
 import { logDebug } from '../logger'
 import { CHAT_INPUT_TOKEN_BUDGET, CHAT_OUTPUT_TOKEN_BUDGET } from '../token/constants'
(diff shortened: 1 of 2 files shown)


lib/shared/src/models/index.ts
```diff
@@ -338,7 +335,7 @@ export class ModelsService {
             return tier !== 'enterprise'
         }
 
-        return tier === 'free'
+        return true
```
jamesmcnamara, line 338: I think this should stay as it was, because if you called this with a free user and a cody pro model it would return true

jamesmcnamara (approved):
One change and then I think it's good to go!

lib/shared/src/models/index.ts
abeatrix: would it work if we update it to check if `tier !== 'pro` in case we miss the Free tag?

jamesmcnamara: It will be okay because the [tier function defaults to free](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/lib/shared/src/models/index.ts?L160%3A1-160%3A88). Although this brings up a good point, I'm going to push a unit test here to make sure we catch things like that.

jamesmcnamara: I'll make this change too.

abeatrix: cool i'll revert my change so you can make the change in a follow up then, thanks!

PR #4874, opened 2024-07-15
added(tests): Unit tests for model usage

jamesmcnamara:
Adds unit tests for the `canUserUseModel` method.
## Test plan
Added unit tests


lib/shared/src/models/index.test.ts
```diff
@@ -136,4 +149,55 @@ describe('Model Provider', () => {
...
+        })
+        const proModel = new Model({
+            model: 'pro-model',
+            usage: [ModelUsage.Chat],
+            tags: [ModelTag.Pro],
+        })
+        const freeModel = new Model({
+            model: 'free-model',
+            usage: [ModelUsage.Chat],
+            // We don't include ModelTag.Free here to test that it's not required.
```
abeatrix, line 167: Nice! This means local ollama models would fall into the "free" category right?

abeatrix (approved):
Thanks for adding the tests! Would be nice to also include Local models and Dev models and see how they interact for each tier but overall lgtm! 

Unrelated to this PR, but I think for enterprise it might be better if we restrict it to models with Enterprise tag for now to preserve the current behavior.

lib/shared/src/models/index.test.ts
jamesmcnamara: Correct.

jamesmcnamara:
@abeatrix I have another branch going with more changes for the model selection that will impact how we import local and dev models so I'll add those tests there.

PR #4875, opened 2024-07-15
Agent: add support for Visual Studio IDE in AgentWorkspaceConfiguration

abeatrix:
This change adds support for the Visual Studio IDE in the AgentWorkspaceConfiguration class. A new case has been added to the `getIDE()` method to handle the 'visualstudio' configuration value and return the corresponding CodyIDE.VisualStudio enum value.

Additionally, the CodyIDE enum in the shared configuration.ts file has been updated to include the VisualStudio value.

## Test plan

Green CI as this PR is adding a new value to the enum object that is not being used by other clients yet.

Also there is no feature change.


kalanchan (approved):
gogogogo so nice!!

jamesmcnamara (approved):

PR #4876, opened 2024-07-15
increased output token limits for 3.5 sonnet

chenkc805:
Requirements:
- Increase output token limits for 3.5 sonnet from 4000 to 8000
- Context: https://sourcegraph.slack.com/archives/C05SZB829D0/p1721069882054949

## Test plan

manually test

### DIFF
--- a/lib/shared/src/models/dotcom.ts
+++ b/lib/shared/src/models/dotcom.ts
@@ -2,6 +2,7 @@ import type { Model } from '.'
 import {
     CHAT_INPUT_TOKEN_BUDGET,
     CHAT_OUTPUT_TOKEN_BUDGET,
+    CHAT_OUTPUT_TOKEN_BUDGET_3_5_SONNET,
     EXTENDED_CHAT_INPUT_TOKEN_BUDGET,
     EXTENDED_USER_CONTEXT_TOKEN_BUDGET,
 } from '../token/constants'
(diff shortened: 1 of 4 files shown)


lib/shared/src/models/index.ts
```diff
@@ -349,6 +351,9 @@ export class ModelsService {
      * Finds the model provider with the given model ID and returns its Context Window.
      */
     public static getContextWindowByID(modelID: string): ModelContextWindow {
+        if (modelID.toLowerCase().includes('claude-3-5-sonnet')) {
+            return { input: CHAT_INPUT_TOKEN_BUDGET, output: CHAT_OUTPUT_TOKEN_BUDGET_3_5_SONNET }
+        }
```
abeatrix, line 356:
```suggestion
```

I dont think we will need this since we have done this in the constructor, so if we can find the model, the model.contextWindow should have the correct context window

lib/shared/src/models/dotcom.ts
```diff
@@ -42,7 +52,7 @@ export const DEFAULT_DOT_COM_MODELS = [
         model: 'anthropic/claude-3-5-sonnet-20240620',
         provider: 'Anthropic',
         usage: [ModelUsage.Chat, ModelUsage.Edit],
-        contextWindow: expandedContextWindow,
+        contextWindow: expandedContextWindow3_5_Sonnet,
```
abeatrix, line 55: this only works for PLG users. We will also need to update this file for Enterprise users: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/models/utils.ts?L29

abeatrix (changes requested):
Left some comments inline.

This change alone will not work since the header `"anthropic-beta": "max-tokens-3-5-sonnet-2024-07-15" ` needs to be included on the server side when server side make the API requests to Anthropic, so might need to get cody prime team to make that change first. The Cody prime team will also need to update the current output limit to 8k or this change will trigger the abuse detection IIRC?

lib/shared/src/models/dotcom.ts
abeatrix: cc @jamesmcnamara is there other places we need to change with the server side config change?

chenkc805:
> Left some comments inline.
> 
> This change alone will not work since the header `"anthropic-beta": "max-tokens-3-5-sonnet-2024-07-15" ` needs to be included on the server side when server side make the API requests to Anthropic, so might need to get cody prime team to make that change first. The Cody prime team will also need to update the current output limit to 8k or this change will trigger the abuse detection IIRC?

Thanks @abeatrix . This is where my knowledge ends, and although I used Cody Web to try to move me forward I am stuck. CC: @sourcegraph/cody-prime to help!

PR #4882, opened 2024-07-16
fixed(models): removed enterprise tag from legacy enterprise models

jamesmcnamara:
We rely on the presence of the `ModelTag.Enterprise` to determine if the server supports new style models (i.e. multiple models for an enterprise). For legacy enterprises, we assume they won't have that tag and we won't show the dropdown.

However, I made the mistake of including the enterprise tag on the legacy code-path and thus we got the drop down for all enterprises and without display names. So instead of using a hacky check like that, I instead post the cody client config to the web view and then check directly if the model view API is enabled only use that to determine if we should display a dropdown for enterprise customers.

## Test plan
Updated unit tests.


vscode/src/models/sync.ts
```diff
@@ -77,7 +77,7 @@ export async function syncModels(authStatus: AuthStatus): Promise<void> {
                     authStatus?.configOverwrites?.chatModel,
                     authStatus?.configOverwrites
                 ),
-                tags: [ModelTag.Enterprise],
```
abeatrix, line 80: Hmm this looks confusing because this is an enterprise model so it should still have the enterprise tag?

vscode/src/chat/protocol.ts
```diff
@@ -168,6 +169,8 @@ export type ExtensionMessage =
           config: ConfigurationSubsetForWebview & LocalEnv
           authStatus: AuthStatus
           workspaceFolderUris: string[]
+          // Server provided details about what features are enabled for the current user
+          codyClient: CodyClientConfig | undefined
```
abeatrix, line 173: I think this is what the "setConfigFeatures" below is doing, so maybe we can just add a new field to the "setConfigFeatures" instead?

jamesmcnamara: Done. I had to move it into the ChatModelContext instead of the UserStatus but that probably makes more sense anyway.

vovakulikov (approved):
Checked manually Cody Web (s2 - no LLM models selector which is correct, and sg02 with LLM picker since server sends LLM models there)

Thanks

PR #4884, opened 2024-07-16
make alt+l toggle chat, rather than create a new chat

beyang:
https://github.com/user-attachments/assets/db166dc0-538e-44ca-9760-8416d0d737f8

Enable the user to build some good muscle memory:

* `Alt+L` toggles between chat and editor
* `Alt+Shift+L` creates a new chat session in the chat panel
* `Alt+L` with an editor selection still adds the selection to the current chat context (but it now shows the chat panel if it is hidden)

## Test plan

Tested locally

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -604,11 +604,6 @@
         "when": "cody.activated"
       },
       {
-        "command": "cody.chat.focus",
-        "key": "alt+l",
-        "when": "cody.activated && editorTextFocus"
-      },
-      {
         "command": "cody.chat.toggle",
         "key": "alt+l",
         "when": "cody.activated && editorTextFocus",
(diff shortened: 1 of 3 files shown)


vscode/package.json
```diff
@@ -607,9 +600,30 @@
...
       },
+      {
+        "command": "cody.chat.focus",
+        "key": "alt+l",
+        "when": "cody.activated && editorTextFocus"
+      },
+      {
+        "command": "cody.chat.toggle",
+        "key": "alt+l",
+        "when": "cody.activated && editorTextFocus",
```
sqs, line 614: Is it intentional to have 2 `alt+l` keybindings with the same `when`?

sqs (approved):

PR #4888, opened 2024-07-16
CLI: add more flexible message construction for `chat` command

olafurpg:
Fixes CODY-2658

Previously, it was only possible to pass the chat message through the explicit `--message` or `-m` option. This PR adds several new ways to pass the chat message:

- Via arguments. Example: `cody chat explain react hooks`.
- Via stdin with `--stdin` option or `-` argument. Example: `git diff | cody chat --stdin --message 'summarize this diff'`

## Test plan
Manually tested these commands
```
echo 'respond only with hello' | cody chat -
echo 'respond only with hello' | cody chat --stdin
git diff | cody chat --stdin --message 'summarize this diff'
git diff | cody chat --message 'summarize this diff' -
cody chat --message 'what files changed' $(git show --name-only) 
```

### DIFF
--- a/agent/src/cli/command-chat.ts
+++ b/agent/src/cli/command-chat.ts
@@ -21,6 +21,8 @@ export interface ChatOptions {
     endpoint: string
     accessToken: string
     message: string
+    stdin?: boolean
+    messageArgs?: string[]
     dir: string
     model?: string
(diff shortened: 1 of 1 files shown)


umpox (approved):
Nice!

PR #4889, opened 2024-07-16
Autocomplete: Add telemetry tests for recycled completion requests

umpox:
## Description

Adds a unit test for subsequent completion requests that match the behaviour 

## Test plan

Testing change only - Unit tests should pass

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.test.ts
+++ b/vscode/src/completions/inline-completion-item-provider.test.ts
@@ -10,14 +10,17 @@ import {
     graphqlClient,
 } from '@sourcegraph/cody-shared'
 
+import { telemetryRecorder } from '@sourcegraph/cody-shared'
 import { localStorage } from '../services/LocalStorageProvider'
 import { DEFAULT_VSCODE_SETTINGS, vsCodeMocks } from '../testutils/mocks'
 import { withPosixPaths } from '../testutils/textDocument'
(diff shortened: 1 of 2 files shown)


vscode/src/completions/logger.ts
```diff
@@ -752,7 +752,11 @@ export function suggested(id: CompletionLogID, span?: Span): void {
 
             // We can assume that this completion will be marked as `read: true` because
             // READ_TIMEOUT_MS has passed without the completion being logged yet.
-            if (event.suggestedAt && !event.suggestionAnalyticsLoggedAt && !event.suggestionLoggedAt) {
+            if (
```
umpox, line 755: Note: These checks are functionally the same, I just noticed in my tests (that had a timestamp of 0), we were not meeting these checks.

This just ensures that we explicitly guard against the `null` value rather than `null || 0` (yay for JS)

abeatrix (approved):

vscode/src/completions/inline-completion-item-provider.test.ts
```diff
@@ -436,6 +439,67 @@ describe('InlineCompletionItemProvider', () => {
             expect(spy).toHaveBeenCalledTimes(0)
             cursorSelectionMock.mockReset()
         })
+
+        it('logs a single completion if a subsequent completion has a matching logId', async () => {
```
valerybugakov, line 443: ✨

PR #4898, opened 2024-07-16
Cody Web: Fix client-config request for Cody Web in production

vovakulikov:
So, with the new `client-config` handler (which is part of REST API), we've got a runtime fail in Cody Web in production. The problem is that the GET request doesn't send the Origin request header (since it's a plain get request), also we don't send custom client headers like 
```
x-requested-with: Sourcegraph
x-sourcegraph-client: https://sourcegraph.sourcegraph.com/
```

because of this lack of request headers this `client-config` fails in a runtime with 401 status. This PR exposes `customHeaders` API for Cody Web components, so consumer components in the Sourcegraph client would be able to set the proper set of headers that we have in `window.context.xhrHeaders`

## Test plan
- This change doesn't change much in cody vs code logic so here we should just make sure that CI is passing 
- Later as this PR will be merged and published to NPM I will prepare a PR where we conduct proper testing in sourcegraph client

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -1353,14 +1353,18 @@ export class ClientConfigSingleton {
     }
 
     public async getConfig(): Promise<CodyClientConfig | undefined> {
-        switch (this.shouldFetch()) {
-            case 'sync':
-                return this.refreshConfig()
-            // biome-ignore lint/suspicious/noFallthroughSwitchClause: This is intentional
-            case 'async':
-                this.refreshConfig()
-            case false:
-                return this.cachedClientConfig
+        try {
+            switch (this.shouldFetch()) {
+                case 'sync':
+                    return this.refreshConfig()
+                // biome-ignore lint/suspicious/noFallthroughSwitchClause: This is intentional
+                case 'async':
+                    this.refreshConfig()
+                case false:
+                    return this.cachedClientConfig
+            }
+        } catch {
+            return
         }
     }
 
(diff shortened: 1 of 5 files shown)


vovakulikov:
FYI, windows unit test failure isn't this PR problem; I can see the same problem on the main, too.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1353,14 +1353,18 @@ export class ClientConfigSingleton {
...
+                case 'sync':
+                    return this.refreshConfig()
+                // biome-ignore lint/suspicious/noFallthroughSwitchClause: This is intentional
+                case 'async':
+                    this.refreshConfig()
+                case false:
+                    return this.cachedClientConfig
+            }
+        } catch {
+            return
```
jamesmcnamara, line 1367: Would be great to log an error here for visibility
```suggestion
            logError('ClientConfigSingleton::getConfig', 'refresh failed', { verbose: error })
            return
```

jamesmcnamara (approved):

slimsag (approved):

slimsag:
Looks right. And just FYI the only header that really matters here is `X-Requested-With: Sourcegraph` - you can see right here in the Go code that we check for that header and, if it is present, we allow cookie auth:

https://sourcegraph.com/github.com/sourcegraph/sourcegraph@61cb0dc807870935e5ed62b85fc66a3c160f65b6/-/blob/internal/session/session.go?L362-371

lib/shared/src/sourcegraph-api/graphql/client.ts
vovakulikov: Yes, but we already log error in `this.refreshConig` (it makes log and re-through error to the caller)

PR #4900, opened 2024-07-16
Chat UI: add tabs to sidebar chat and set editor chat by default for enterprise users

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2813/implement-basic-tab-strip-chat-history-commands
PART OF https://linear.app/sourcegraph/project/🔼-implement-unified-cody-chat-ui-17c60c5e85d4/issues?projectMilestoneId=b2471120-a900-491b-8a4a-7ec905692e7b

Note: This PR is adding the new nav bar with commands view and history view as shown in [the figma design](https://www.figma.com/design/f078wFMKsIOaEwj7Iwj5xy/Universal-Cody?node-id=2179-3103&t=wu5mes2cZOgGAgKC-0). It does not remove the current Settings and Support tree view, which will be handled in a separate PR.

image

Summary:
- Add a navigation bar with icons for different views (Chat, History, Commands, Settings, Account)
- Implement new views for History and Commands
- Update the App component to handle the new views and navigation
- Update the WelcomeMessage component to handle the new views and commands
- Add new method in command controller for returning a list of commands
- Adjust existing styles for the new navbar and views
- Update storybook
- Add new e2e tests for history sidebar view and execute command from command sidebar view

~~This should also unblock custom commands from working in JB when the new webview is available in JB~~ ([source](https://sourcegraph.slack.com/archives/C05AGQYD528/p1721172490745989))

## Test plan

Added new views (CommandsView & HistoryView) and `NavBar` component to the storybook:
image

Tabs View:

image

Commands Tab

image

History Tab

image


toolmantim (commented):
Here's a few things I noticed quickly, will review properly tomorrow:

* SVG stroke width on icons should be 1.25px
* Border bottom on tab should be 1px
* Border bottom on tab header should be 1px
* Clicking the settings tab shouldn't take you back to the chat tab
* Repeatedly pressing the settings tab button seems buggy (may be related to the chat input auto-focus?)
* Tabs should use shadcn/ui tabs or at least Radix UI tabs for keyboard accessibility
* Remove the `tw-my-3` margin from the tab bar
* Add `tw-py-1` to the tab bar to bring it in 2px on left and right

Design things that need fixing/defining (I'll look into tomorrow):
* We need to add tooltips (using our Tooltip component) for the tabs and buttons
* Nav bar button spacing
* Tab content padding
* Adding all the links so we can remove "Settings & Support"

abeatrix:
WIP: 
- [ ] SVG stroke width on icons should be 1.25px
- [x] Border bottom on tab should be 1px
- [x] Border bottom on tab header should be 1px
- [x] Clicking the settings tab shouldn't take you back to the chat tab
- [ ] ~~Repeatedly pressing the settings tab button seems buggy (may be related to the chat input auto-focus?)~~
- [ ] Tabs should use shadcn/ui tabs or at least Radix UI tabs for keyboard accessibility - wip
- [x] Remove the tw-my-3 margin from the tab bar
- [x] Add tw-py-1 to the tab bar to bring it in 2px on left and right
  - tw-px-1 brought it 2px on left and right which makes the change not noticible. Should I try tw-px-4 instead? @taiyab @toolmantim 
  
with tw-px-1:
  
image

dominiccooney (changes requested):
- Restoring a chat seems to hang, unsure if this is related: https://www.loom.com/share/a259b1174e354e53a0332b0683ba7ff8?sid=4d2f21ea-95ca-4bb5-86aa-07766fa9aa08
- Is the order of commands in "Commands" what we want? I've got autoedits (experimental) and generate commit message (experimental) are at the top, clicking them apparently does nothing... is that putting our best foot forward? Custom commands are empty for me but there's a very expansive tooltip...
Screenshot 2024-07-18 at 15 53 09
- The "commands" tab has *two* gears in the same tab bar which do different things... can we just follow the principles of Gestalt here and put the custom commands gear near custom commands?
Screenshot 2024-07-18 at 15 59 33

Probably not related to your change, but I noticed:
- Ollama icon despite chatting with Sonnet 
Screenshot 2024-07-18 at 15 54 21
- With a response streaming, clicking "move to panel" button kills the response.

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -1532,15 +1541,6 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
     }
 
     public async setWebviewView(view: View): Promise<void> {
-        if (view !== 'chat') {
```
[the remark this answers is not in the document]
abeatrix: Yeah, that's because we used to render the chat view once the user is logged in, but this is not the case anymore as we start creating the webview even when the user is not logged in for auth view.

abeatrix:
@dominiccooney Thanks for taking the time to review! A lots of the bugs are unrelated to the changes I made and are reproducible on main, will file separate issues for them to follow up:

- Restoring a chat seems to hang, unsure if this is related:
- With a response streaming, clicking "move to panel" button kills the response.

> Is the order of commands in "Commands" what we want? I've got autoedits (experimental) and generate commit message (experimental) are at the top, clicking them apparently does nothing... is that putting our best foot forward?

Woops filtered and removed from the list.

> Custom commands are empty for me but there's a very expansive tooltip...

I can't reproduce this, but i fixed some styling issues that might have caused this. Also added a new item for creating a new custom command: 

image


> The "commands" tab has two gears in the same tab bar which do different things... can we just follow the principles of Gestalt here and put the custom commands gear near custom commands?

this is from the figma design file: 

image

I think they are working on the new design for prompt library that'd eventually remove the extra gear. 

> Ollama icon despite chatting with Sonnet

Yea I think this is related to the bug we just fixed on main in https://github.com/sourcegraph/cody/pull/4911

toolmantim:
I doesn't have to be this PR, but I was hoping we'd be a bit more systematic with creating UI components.

e.g. create a `shadcn/ui/tabs.tsx` via shadcn/ui, and make the necessary changes to the class names to have it match our style, in https://github.com/sourcegraph/cody/tree/main/vscode/webviews/components/shadcn/ui

Our [Figma component](https://www.figma.com/design/zNtJZZwuLn1RLgsNNL2C52/%E2%9C%B3%EF%B8%8F-wildcard-shadcn?node-id=22-8555&t=xJtr8GQinpHNRjIy-4) is generic too, as is the overall idea of our wildcard-shadcn Figma components.

Then all the view-specific logic can live outside of that pure UI component.

You can see the message input toolbar (https://github.com/sourcegraph/cody/blob/1cd4ba8a825d8b88a7d662f7997ed3792f097e48/vscode/webviews/chat/cells/messageCell/human/editor/toolbar/Toolbar.tsx) uses the standard <Button> etc too.

toolmantim:
Sorry @abeatrix, I didn't get time to update the Figma yesterday… the command list on the new chat experience and the Commands tab should be the same. Let me update the Figma now…

toolmantim:
image

btw this component ^ is a standard shadcn/ui collapsible, that would use our standard `<Button>` etc: https://ui.shadcn.com/docs/components/collapsible

abeatrix:
> I doesn't have to be this PR, but I was hoping we'd be a bit more systematic with creating UI components.
> 
> e.g. create a `shadcn/ui/tabs.tsx` via shadcn/ui, and make the necessary changes to the class names to have it match our style, in https://github.com/sourcegraph/cody/tree/main/vscode/webviews/components/shadcn/ui
> 
> Our [Figma component](https://www.figma.com/design/zNtJZZwuLn1RLgsNNL2C52/%E2%9C%B3%EF%B8%8F-wildcard-shadcn?node-id=22-8555&t=xJtr8GQinpHNRjIy-4) is generic too, as is the overall idea of our wildcard-shadcn Figma components.
> 
> Then all the view-specific logic can live outside of that pure UI component.
> 
> You can see the message input toolbar (https://github.com/sourcegraph/cody/blob/1cd4ba8a825d8b88a7d662f7997ed3792f097e48/vscode/webviews/chat/cells/messageCell/human/editor/toolbar/Toolbar.tsx) uses the standard `<Button>` etc too.

cool I wasn't aware we are moving toward this but I can refactor them in a follow up since i don't think I get to it this week. Is this documented somewhere I can refer to?

toolmantim:
> cool I wasn't aware we are moving toward this but I can refactor them in a follow up since i don't think I get to it this week. 

👍🏼

> Is this documented somewhere I can refer to?

I don't think this is documented anywhere, but something @taiyab and Code Search team were aware of

toolmantim:
I've just updated the Figma, including the example VS Code and Jetbrains frames below the main frames.

Here's a prototype link to understand the idea behind the hover states too: https://www.figma.com/proto/f078wFMKsIOaEwj7Iwj5xy/Universal-Cody?page-id=2179%3A3103&node-id=2802-44502&viewport=-12%2C531%2C0.93&t=QuwOoDxwNIK1a1v4-1&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=2802%3A44502

One thing that has changed is not listing out each custom command, but just loading up the menu… given it also has the buttons for creating, managing etc. No need for double-cog icon.

I'm sorry for the miscommunication on timing around this, and the design scoping! I hate changing something once you've already done so much of the PR.

toolmantim:
I actually do have some documentation here @abeatrix — not frontend dev specific though: https://linear.app/sourcegraph/project/wildcard-shadcn-7dad47f37fe7/overview

vscode/test/e2e/command-core.test.ts
```diff
@@ -93,3 +93,39 @@ test.extend<ExpectedV2Events>({
...
+    await page.getByRole('treeitem', { name: 'buzz.ts' }).locator('a').dblclick()
+    await page.getByRole('tab', { name: 'buzz.ts' }).hover()
+
+    // Click on some code within the function
+    await page.getByText("fizzbuzz.push('Buzz')").click()
+
+    // Execute the command from the Commands tab in Chat view
+    await focusSidebar(page)
+
+    const chatPanel = page.frameLocator('iframe.webview').last().frameLocator('iframe')
```
jamesmcnamara, line 126: There is a utility for this: `getChatEditorPanel`

vscode/test/e2e/command-core.test.ts
```diff
@@ -93,3 +93,39 @@ test.extend<ExpectedV2Events>({
...
+    // Click on the Commands view icon
+    await chatPanel.locator('[id="radix-\\:r0\\:-trigger-commands"]').getByRole('button').click()
+    await chatPanel.getByRole('button', { name: 'Explain code' }).click()
+    await chatPanel.getByText('hello from the assistant').hover()
```
jamesmcnamara, line 130: I always find this a confusing way to do a test, I'd personally find:
```suggestion
    await expect(chatPanel).toContainText('hello from the assistant')
```
or 
```suggestion
    await expect(getChatInputs(chatPanel).first()).toContainText('hello from the assistant')
```

vscode/test/e2e/common.ts
```diff
@@ -54,7 +54,6 @@ export async function focusSidebar(page: Page): Promise<void> {
 
 export async function expectAuthenticated(page: Page) {
     await focusSidebar(page)
-    await expect(page.getByLabel('Settings & Support Section')).toBeVisible()
```
jamesmcnamara, line 57: I looked at this for a while until I realized that the focusSidebar would hang forever if the user wasn't authenticated. Maybe we should add a comment about this?

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -153,8 +169,8 @@ export class ChatsController implements vscode.Disposable {
                 vscode.commands.executeCommand('cody.chat.focus')
             ),
             vscode.commands.registerCommand('cody.chat.newPanel', async () => {
-                await this.panel.clearAndRestartSession()
-                await vscode.commands.executeCommand('cody.chat.focus')
+                await this.panel?.clearAndRestartSession()
+                this.panel && vscode.commands.executeCommand('cody.chat.focus')
```
jamesmcnamara, line 173: Is there a reason to not `await` here?
```suggestion
                this.panel && await vscode.commands.executeCommand('cody.chat.focus')
```

vscode/test/e2e/auth.test.ts
```diff
@@ -41,16 +42,14 @@ test.extend<ExpectedV2Events>({
     await signOut(page)
     await focusSidebar(page)
 
+    // Makes sure the sign in page is loaded in the sidebar view with Cody: Chat as the heading
+    // instead of the chat panel.
```
jamesmcnamara, line 46: Here too: `const sidebarFrame = getChatEditorPanel(page)`

vscode/webviews/App.tsx
```diff
@@ -182,6 +187,7 @@ export const App: React.FunctionComponent<{ vscodeAPI: VSCodeWrapper }> = ({ vsc
     useEffect(() => {
         if (!view) {
             vscodeAPI.postMessage({ command: 'initialized' })
+            return
```
jamesmcnamara, line 190: Out of curiosity, does this have any effect? If so we should definitely document it as I'd immediately delete it if I was editing this code.

vscode/webviews/Chat.module.css
```diff
@@ -1,5 +1,5 @@
 .container {
-    min-height: 100vh;
+    height: 100%;
```
jamesmcnamara, line 2: Not 100vh? It's been a long time since I've done CSS, but I remember 100% never doing what I expected, unless you are in like an absolutely sized container.

Then again I'm messing around with the chat now and it looks fine.

jamesmcnamara (approved):
Looks good! I didn't review the CSS or tailwind classes too closely (I don't really know much about those yet). But I played around with it locally and it works well and looks great! Love the hover styling.

vscode/webviews/App.tsx
abeatrix: It shouldn't! I don't even remember why I added this 🤣

vscode/src/chat/chat-view/ChatsController.ts
abeatrix: i've removed this logic because turns out this.panel needs to be initialized for the Login page to work, so i change back to using "when" clause for controlling when to show the sidebars or not

vscode/test/e2e/auth.test.ts
abeatrix: Thank you! Just found out about `getChatSidebarPanel`!

vscode/test/e2e/command-core.test.ts
abeatrix: Replaced with getChatSidebarPanel!

vscode/test/e2e/command-core.test.ts
abeatrix: Ah my bad! Since this is not chatInput i just updated it to check for the Text :)

vscode/test/e2e/common.ts
abeatrix: Good point!I just added a expect and comment to reflect this.

PR #4901, opened 2024-07-17
Autocomplete: preload completions on cursor movement

valerybugakov:
- Preloads completions on cursor movement if the cursor is at the end of an empty line. Also, if the current line is not empty and the following line is, preloads completions for the following line.
- Closes https://linear.app/sourcegraph/issue/CODY-2664/[autocomplete-latency]-experiment-with-pre-loading-completions-on

## Test plan

Added unit tests + CI.


vscode/src/completions/get-inline-completions.ts
```diff
@@ -364,6 +352,23 @@ async function doGetInlineCompletions(
         }
     }
 
+    // If we have inflight request with the same request params, just use it here instead of doing additional work.
+    // Specifically relevant for completions preloading where we want to avoid doing work twice.
+    const matchingInflightRequest = requestManager.getMatchingInflightRequest({ requestParams })
```
umpox, line 357: What is the scenario here?

Where we have preloaded a line, and then the user triggers a request for the exact same preloaded inflight request?

vscode/src/completions/get-inline-completions.ts
```diff
@@ -373,7 +378,11 @@ async function doGetInlineCompletions(
         stale = true
     }
 
-    if (smartThrottleService) {
+    if (
+        smartThrottleService &&
+        triggerKind !== TriggerKind.Manual &&
+        triggerKind !== TriggerKind.Preload
+    ) {
```
umpox, line 385: Do you still need to call `cancellationListener?.dispose()` for preloaded requests?

Otherwise these requests will be aborted as soon as the user types? Or is that the intention?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -188,11 +205,81 @@ export class InlineCompletionItemProvider
...
+        | undefined
+
+    // TODO: limit the maximum number of the inflight preload requests.
+    private async preloadCompletionOnSelectionChange(
+        event: vscode.TextEditorSelectionChangeEvent
+    ): Promise<void> {
+        const lastSelection = event.selections.at(-1)
+        const { document } = event.textEditor
+
+        if (lastSelection?.start.isEqual(lastSelection?.end) && document.uri.scheme === 'file') {
```
umpox, line 237:
```suggestion
        if (lastSelection?.isEmpty && document.uri.scheme === 'file') {
```

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -188,11 +205,81 @@ export class InlineCompletionItemProvider
...
+            const currentLine = document.lineAt(lastSelection.end.line)
+            const currentLinePrefix = currentLine.text.slice(0, lastSelection.end.character)
+            const currentLineSuffix = currentLine.text.slice(lastSelection.end.character)
```
umpox, line 240: This is super minor, but using `lastSelection.active` is technically more correct here, as it represents the cursor position.

Doesn't really matter as we're checking the selection is empty, but initially found this confusing when scanning the code. Feel free to ignore :)

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -188,11 +205,81 @@ export class InlineCompletionItemProvider
...
+                })
+            ) {
+                this.provideInlineCompletionItems(document, lastSelection.end, {
+                    isPreload: true,
+                    triggerKind: 1,
+                    selectedCompletionInfo: undefined,
+                })
+            } else {
+                const nextLineNumber = lastSelection.end.line + 1
+                const nextLine = document.lineAt(nextLineNumber)
```
umpox, line 258: Will this error if the cursor is on the last line of the file?

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -389,8 +481,8 @@ export class InlineCompletionItemProvider
                     stageRecorder.record('preFinalCancellationCheck')
                 }
 
-                // Avoid any further work if the completion is invalidated already.
-                if (abortController.signal.aborted) {
+                // Avoid any further work if the completion is invalidated already or if it's a preload request.
+                if (abortController.signal.aborted || isPreloadRequest) {
```
umpox, line 485: Nice 👍

vscode/src/completions/get-inline-completions.ts
umpox: Ah or is it that we may trigger a 2nd preloaded request if the user moves their cursor to the next empty line

umpox:
Feels good 👍  I can see a common scenario being that the user wants to actually add an extra new line. Do you know why we get that flicker? Seems that we have a delay somewhere

https://github.com/user-attachments/assets/6c43e804-791f-46bc-b1a5-0d019ab9b29e

umpox:
Also side note, I notice that in some cases the LLM will not generate a response for the _immediate_ next line, but the line after that if there's whitespace. Possibly because that's a better indicator that code would be entered here.

I wonder if we should try to catch this, maybe by generating another request for the 2nd empty line. Or some heuristic to determine where the users' next likely cursor position will be. Probably a follow on though

vscode/src/completions/get-inline-completions.ts
valerybugakov: Both of these are relevant. WIll update the comment to be more concrete!

vscode/src/completions/get-inline-completions.ts
valerybugakov: Yep, I still need it. I wanted only to avoid additional throttling here.

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: TIL

valerybugakov:
> but the line after that if there's whitespace. Possibly because that's a better indicator that code would be entered here.

I also noticed that, and we can solve it generically for all completions. We should wait for the steam's first char and start our completion detection logic. I can do it quickly in a separate PR.

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Added a condition to avoid that.

valerybugakov:
> Do you know why we get that flicker?

Not yet. Going to look into it in a follow up PR.

vscode/src/completions/completion-provider-config.ts
```diff
@@ -71,6 +75,55 @@ class CompletionProviderConfig {
...
+        // - Variant-1 25%
+        // - Variant-2 25%
+        // - Variant-3 25%
+        // - Control group 25%
+        //
+        // The rollout values to set:
+        // - CodyAutocompletePreloadingExperimentBaseFeatureFlag 75%
+        // - CodyAutocompleteHotStreak 33%
+        // - CodyAutocompleteSmartThrottle 100%
+        // - CodyAutocompleteSmartThrottleExtended 50%
```
umpox, line 89: Need to update this :)

umpox (approved):
Nice!

PR #4902, opened 2024-07-17
Autocomplete: fix prefix computation in cache keys

valerybugakov:
The current implementation doesn't account for cases where the current line suffix is an empty string. This PR fixes it and updates unit tests. Part of https://linear.app/sourcegraph/issue/CODY-2190/[autocomplete-latency]-ab-test-hot-streak

## Test plan

Updated unit tests.


hitesh-1997 (approved):
LGTM !!

PR #4903, opened 2024-07-17
"Prompt" chat toolbar item to add a prompt from Prompt Library

sqs:
> NOTE: Only review the last commit (`Prompt selector in chat to select a prompt from the Prompt Library`). The other commits all have their own separate PRs opened.

The [Prompt Library](https://sourcegraph.com/prompts) lets you save and share useful prompts for Cody. Now, you can insert a prompt from your Prompt Library into your Cody chat message with this new **Prompts** selector in the message editor toolbar.
    
Feature-flagged off by default (`chat-prompt-selector`).

image

## Test plan

Enable the `chat-prompt-selector` feature flag for your user on dotcom. Try adding the `sqs/gen-test-react-vitest` prompt. It should append the prompt's text in your chat message editor.


sqs:
From @toolmantim:

> In one of my designs I used the "book-text" icon in that input toolbar for the prompt library, next to the "at-sign" icon. I'd recommend we switch to that, along with it being wrapped by our nice new <Tooltip>. For last item maybe "Manage Prompt Library" with the "book-text" icon on the left, and optionally an "external-link" icon on the right?

sqs:
@toolmantim I haven't implemented the design in Figma yet for this, but I will do so before the stable release. Just wanna get this out there to test in the prerelease build.

vscode/webviews/AppWrapper.tsx
```diff
@@ -1,16 +1,22 @@
...
 import { ComposedWrappers, type Wrapper } from './utils/composeWrappers'
 import { TelemetryRecorderContext } from './utils/telemetry'
 import { ConfigProvider } from './utils/useConfig'
+import { FeatureFlagsProvider } from './utils/useFeatureFlags'
 
 export const AppWrapper: FunctionComponent<{ children: ReactNode }> = ({ children }) => (
     <ComposedWrappers wrappers={COMMON_WRAPPERS}>{children}</ComposedWrappers>
 )
 
+const TEST_FEATURE_FLAGS: Record<string, boolean> = {}
```
keegancsmith, line 18: we should provide a way to override (in config maybe?) that is convenient like we do for our code search webapp. Given you have made it so easy to use now in the webview I imagine us using this a lot more then.

vscode/webviews/utils/useDebounce.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param delay Delay before updating the value
+ * @returns The updated value
+ */
+export const useDebounce = <T>(value: T, delay: number): T => {
+    const [debouncedValue, setDebouncedValue] = useState(value)
+
+    useEffect(() => {
+        const handler = setTimeout(() => setDebouncedValue(value), delay)
+        return () => clearTimeout(handler)
+    }, [delay, value])
```
keegancsmith, line 16: neat!

keegancsmith (approved):
LGTM. Pretty much none of the new code will run if the flag is off. Nice stuff.

PR #4912, opened 2024-07-17
Chat: use Context Preamble by default

abeatrix:
Follow up on https://linear.app/sourcegraph/issue/CODY-2507/evaluate-claude-35-sonnet-as-new-default

As mentioned in the linked issue, we were able to verify the context preamble has improved chat quality. This PR is to remove the CodyChatContextPreamble feature flag so that the Context Preamble is always used by default. This also applies to Enterprise users as they do not have access to feature flags.

The CodyChatContextPreamble feature flag has been removed, and the context preamble is now always enabled by default in the PromptMixin. This simplifies the code and removes the need for the feature flag.

BREAKING CHANGE: The CodyChatContextPreamble feature flag has been removed. The context preamble is now always enabled by default when context is used.

## Test plan

Green CI, as all existing features should remain unchanged. All our tests should not affected by this change because the same preamble was enabled for all tests.

### DIFF
--- a/lib/shared/src/experimentation/FeatureFlagProvider.ts
+++ b/lib/shared/src/experimentation/FeatureFlagProvider.ts
@@ -61,9 +61,6 @@ export enum FeatureFlag {
     /** Automatically start indexing using embeddings. */
     CodyEmbeddingsAutoIndexing = 'cody-embeddings-auto-indexing',
 
-    /** Enable Context Preamble for open-end chat questions. */
-    CodyChatContextPreamble = 'cody-chat-context-preamble',
-
     /** Whether to use generated metadata to power embeddings. */
     CodyEmbeddingsGenerateMetadata = 'cody-embeddings-generate-metadata',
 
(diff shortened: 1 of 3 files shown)


chenkc805:
Just to confirm, what we want to do is:
- Make Claude 3.5 Sonnet the default LLM for PLG + Enterprise users (unless admins have set it otherwise)
- Anytime the Claude 3.5 Sonnet LLM is called in chat, then use the new prompt preamble. Otherwise, use the _old_ preamble for other models. 

Does this line up with your thinking @abeatrix ? CC: @aramaraju

abeatrix:
@chenkc805 
> Make Claude 3.5 Sonnet the default LLM for PLG + Enterprise users (unless admins have set it otherwise)
- PLG users: We have implemented this on the client side in the last release.
- Enterprise users: I think we will need cody-prime team help on setting this as the default for Enterprise users? or @arafatkatze, can you help us with this?

> Anytime the Claude 3.5 Sonnet LLM is called in chat, then use the new prompt preamble. Otherwise, use the old preamble for other models.

Current Behavior: **If the feature flag is enabled for the user**, anytime the Claude 3.5 Sonnet LLM is called in chat, then use the new prompt preamble. Otherwise, use the old preamble for other models. This means the preamble is only applied to PLG users.
After this PR: Anytime the Claude 3.5 Sonnet LLM is called in chat, then use the new prompt preamble.  Otherwise, use the old preamble for other models.

chenkc805:
SGTM! thanks for clarifying

chenkc805 (approved):

PR #4913, opened 2024-07-18
Self hosted models

slimsag:
These are the initial client changes needed for Cody Enterprise customers to be able to use self-hosted models. Specifically, this PR pairs with https://github.com/sourcegraph/sourcegraph/pull/63899

The first commit here is mostly changes @jamesmcnamara made to help me wire through the `Model` to the actual location where we write autocomplete providers. The following commits are me introducing a new generic `openaicompatible` autocomplete provider, which will supersede/replace the older experimental one which was purely client-side.

Combined with https://github.com/sourcegraph/sourcegraph/pull/63899 - these changes are enough to have at least Chat and Autocomplete working fairly well using a self-hosted Starcoder2 and Mixtral 8x7b Instruct model.

In this PR, the autocomplete provider is mostly tailored towards models in those families (starcoder and mixtral/mistral instruct models) - to unblock some customer use-cases - but all of the `Model` `clientSideConfiguration` options piped down to this autocomplete provider should be enough to generalize this approach to any model in the future. In a follow-up PR, I will begin making use of that `clientSideConfiguration` to enable using other self-hosted models and other use-cases (letting the site admin customize prompting, etc.)

## Test plan

* The first commit from James has nice unit tests.
* The new autocomplete provider doesn't yet, I will add these in a future PR. For now, it is manually tested on my end using:
  * A local dev Sourcegraph instance configured as described in https://github.com/sourcegraph/sourcegraph/pull/63899
  * A remote server hosting 2 huggingface TGI Docker containers, with starchat2 and mixtral models.
  * A local dev build of VS Code Cody.


vscode/src/completions/providers/openaicompatible.ts
```diff
@@ -354,20 +290,26 @@ export function createProviderConfig({
...
         identifier: PROVIDER_IDENTIFIER,
-        model: clientModel,
+        model: model.model,
     }
 }
 
-function isStarChatFamily(model: string): boolean {
-    return model.startsWith('starchat')
+// TODO(slimsag): self-hosted-models: eliminate model-specific conditionals here entirely
+// by relying on ClientSideConfig appropriately.
```
jamesmcnamara, line 302: ❤️

jamesmcnamara (approved):
I spent an embarassingly long time reviewing the `exopenaicompatibile` file before realizing it was a rename.

I can't comment too heavily on the config driven provider creation, and without tests I have to rely on you to verify that it is working properly. I think there's some abstraction and testing that should be added to this, but in the effort to unblock this, I'll approve.

I also added a commit to fix some CI errors.

PR #4915, opened 2024-07-18
Agent: prepare kotlin bindings for JetBrains adoption

olafurpg:
Upstreams workarounds in https://github.com/sourcegraph/jetbrains/pull/1898/files

We should not have to implement complicated customizations on top of the generated code. It's better to fix the issues in the original implementation.

## Test plan

Green CI.


olafurpg (commented):
OK, I've fixed the dirty git state and the CI is green now

RXminuS (approved):

PR #4917, opened 2024-07-18
CLI: disable autocomplete and clean up logs

olafurpg:
Fixes CODY-2945

Previously, the `cody chat` command enabled autocomplete providers even if we don't use autocomplete. This resulted in noisy console logs in errors related to loading autocomplete. This PR fixes the problem by entirely disabling autocomplete from `cody chat`.

Example console output before this PR

CleanShot 2024-07-18 at 10 53 27@2x

## Test plan
Green CI.

### DIFF
--- a/web/vite.config.ts
+++ b/web/vite.config.ts
@@ -14,7 +14,6 @@ const fakeProcessEnv: Record<string, string | boolean> = {
     NODE_ENV: 'production',
     NODE_DEBUG: false,
     TESTING_DOTCOM_URL: 'https://sourcegraph.com',
-    CODY_SUPPRESS_AGENT_AUTOCOMPLETE_WARNING: true,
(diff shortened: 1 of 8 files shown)


web/vite.config.ts
```diff
@@ -14,7 +14,6 @@ const fakeProcessEnv: Record<string, string | boolean> = {
     NODE_ENV: 'production',
     NODE_DEBUG: false,
     TESTING_DOTCOM_URL: 'https://sourcegraph.com',
-    CODY_SUPPRESS_AGENT_AUTOCOMPLETE_WARNING: true,
```
olafurpg, line 17: FYI @vovakulikov I'm hoping to lean more and more on client capabilities to disable unwanted functionality

valerybugakov (approved):

PR #4920, opened 2024-07-18
Query context retrievers without context chips in text

janhartman:
There is a bug in our chat context fetching that causes embeddings to not work at all and makes symf perform worse.
We pass the full PromptString to these retrievers, which means that we include all context mentions. For example, opening a chat in the cody repo, @mentioning src/context.ts and writing "what is symf" will make the query look like `cody src/context.ts what is symf?`. The reason is that we include everything in the chat text box as the query, including the context chips. 

The solution is to store a separate string without context chips in it and include it in a separate field in the webview submit/edit messages for chat, then use that for retrieving context.

## Test plan
Tested locally - bfg and symf now receive the queries without context chips and return correct results.


sqs (approved):
thanks!

This will probably cause a slight regression in the explain code command because now all it has to context-fetch on is the hard-coded prompt. This is a big flaw in explain code anyway and we will address it very soon with the Prompt Library. But just a heads up if you see a regression there.

sqs:
Oh one other thought - if you ask Cody to `generate a test for @some-filename.ts and follow the style in @other-file.test.ts` then how will it perform? I think it would not know which file you are asking to have the test written for. Maybe we should only omit leading (ie at the start) or initial mention chips?

olafurpg:
I think it's worth considering doing a more advanced fix that keeps both versions:

- The literal parts for context retrieval. Ideally, there's a way for the context retriever to access the lexical editor to pick the parts it wants (maybe we care about filenames but not repos, etc).
- The complete text including chips for the prompt. Removing this context entirely from the LLM is likely going to cause regressions as @sqs calls out.

keegancsmith:
Alright I see the potential issue I was worried about was bought up, but will repost my comment from slack anyways in case it helps the discussion:

> Was there a regression here, or is this an insight into how to improve response quality? If its a regression, I'm assuming its a regression only for the "auto" context we had. IE before the recent chat changes we didn't explicitly mention `@repo @current-file` in chat, and it just lived as part of the context. So in making the changes proposed I would ensure you don't regress on people then explicitly mentioning a file.
Or is it possible the regression is in how we serialize the mention into text for the AI?

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -320,7 +317,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
             case 'chatModel':
                 // Because this was a user action to change the model we will set that
                 // as a global default for chat
-                await ModelsService.setSelectedModel(ModelUsage.Chat, message.model)
+                await ModelsService.setDefaultModel(ModelUsage.Chat, message.model)
```
olafurpg, line 320: Is this an intentional change? Seems like breaking behavior

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -661,7 +659,6 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
                 abortSignal.throwIfAborted()
 
                 this.postEmptyMessageInProgress()
-                this.contextAPIClient?.detectChatIntent(requestID, inputText.toString())
```
olafurpg, line 664: Intentional?

vscode/test/integration/single-root/chat.test.ts
```diff
@@ -50,12 +52,14 @@ suite('Chat', function () {
 
     // do not display filename even when there is a selection in active editor
     test('append current file link to display text on editor selection', async () => {
+        const input = getPs()`hello from the human`
```
olafurpg, line 55: Seems like a redundant diff. Good to keep this out of important PRs like these to lower the cost of review

olafurpg (approved):
You may need to run `pnpm generate-agent-kotlin-bindings` to get the CI passing.

Overall LGTM 👍🏻 A few minor comments

vscode/src/chat/chat-view/ChatController.ts
janhartman: Damn, bad merge. Fixing.

vscode/src/chat/chat-view/ChatController.ts
janhartman: No, bad merge.

vscode/test/integration/single-root/chat.test.ts
janhartman: Good point, was a bit aggressive with DRY.

PR #4921, opened 2024-07-18
Pass timeoutMs value in the request headers

taras-yemets:
Part of [CODY-2775](https://linear.app/sourcegraph/issue/CODY-2775/%5Bautocomplete-latency%5D-apply-the-same-timeout-on-the-cody-gateway-side)

Sends the `timeoutMs` config value to the backend using via `X-Timeout-Ms` header in both default and fast paths.

Backend support added in https://github.com/sourcegraph/sourcegraph/pull/63875

## Test plan
Manually tested (using a debugger) alongside https://github.com/sourcegraph/sourcegraph/pull/63875 on both default and fast paths, using a locally running Sourcegraph instance and Cody Gateway.

### DIFF
--- a/agent/recordings/autocomplete_2136217965/recording.har.yaml
+++ b/agent/recordings/autocomplete_2136217965/recording.har.yaml
@@ -88,11 +88,11 @@ log:
         send: 0
         ssl: -1
         wait: 0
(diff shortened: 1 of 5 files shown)


olafurpg (approved):

valerybugakov (commented):
🙌

PR #4923, opened 2024-07-18
Agent: update integration tests to assert exported telemetry events

akalia25:
Previously, the agent integration tests didn't have assertions for what kinds of telemetry events were being exported. This was problematic because it was easy to introduce telemetry regressions since changes to the exported events wouldn't get captured in tests. This PR fixes the problem by adding new infrastructure to capture exported tests during testing, and adding a new `testing/exportedTelemetryEvents` request to the JSON-RPC protocol allowing clients to access the list of exported events.

---

This PR adds V2 telemetry validation on the Agent E2E test. For each Agent test, there is a list of `expectedEvents` that the loggedEvents check against, to ensure they fired. 

In case of a test failure, it signals either the need to update the `expectedEvents` list to reflect intentional changes or identifies a potential regression. 

The same logging assertion is present in `vscode/test/e2e/...` and has been very helpful in preventing telemetry regressions.

## Test plan
Test via CI and locally with `pnpm run test` on `cody/agent/src/local-e2e`

Example of test failing due to missing Event:
image


agent/src/index.test.ts
```diff
@@ -92,6 +125,37 @@ describe('Agent', () => {
...
         await client.request('testing/reset', null)
+        // reset expectedEvents before each test
+        expectedEvents = undefined
+    })
+
+    afterEach(async () => {
+        const { requests } = await client.request('testing/networkRequests', null)
+        const telemetryEvents = getTelemetryEvents(requests)
+        const telemetryRequests = requests.filter(req => req.url.includes('RecordTelemetryEvents'))
+        const testRunId = uuid.v4()
```
bobheadxi, line 136: suggestion: uuid V7

akalia25: Great suggestion, thank you!

Unfortunately, the `uuid` we have only supports up to `v5()`, since we are on `uuid@9.1`. `uuid@10` has `v5()` ([doc ref](https://www.npmjs.com/package/uuid)).

I believe with the timestamp coming through on the payload, this should suffice, and when cody team upgrades uuid, we can move over :)

bobheadxi: SGTM!

agent/src/index.test.ts
```diff
@@ -13,27 +13,60 @@ import {
...
+}
+
 describe('Agent', () => {
+    let expectedEvents: string[] | undefined
     const client = TestClient.create({
         workspaceRootUri: workspace.rootUri,
         name: 'defaultClient',
         credentials: TESTING_CREDENTIALS.dotcom,
+        // set telemetryExporter to `graphql` to receive telemetryRecorder requests and determine whats events have been logged
+        telemetryExporter: 'graphql',
```
olafurpg, line 69: We intentionally disable the graphql telemetry exporter in our tests because telemetry network requests have been a big source of flaky failures. Have you considered using the `testing` telemetry exporter instead? We can add a new `testing/telemetryEvents` request to the agent protocol that automatically resets on `testing/reset` (which we only use in `index.test.ts` but should be using everywhere)

agent/src/index.test.ts
```diff
@@ -13,27 +13,60 @@ import {
...
+    try {
+        return JSON.parse(str)
+    } catch (e) {
+        console.error('Failed to parse JSON:', e)
+        return null
+    }
+}
+
 describe('Agent', () => {
+    let expectedEvents: string[] | undefined
```
olafurpg, line 63: This logic should live in `TestClient` so that it can be reused for all agent integration tests (not only `index.test.ts`)

olafurpg (changes requested):
Very exciting! I am very eager to start testing telemetry events for the agent. Thank you @akalia25 for kickstarting this 🙏🏻

agent/src/index.test.ts
akalia25: Great suggestion! Just updated to make `expectedEvents` a part of `TestClient`. Thank you!

agent/src/index.test.ts
akalia25: Hey @olafurpg, thanks for the context. I tried using the `testing` telemetry exporter, but when I was, none of the telemetry requests were getting sent 🤔, and I was having a hard time figuring out where the `testing` telemetry exporter was being set on the agent tests. I think we should be using the `MockServerTelemetryRecorderProvider` when `testing` is set. Could you point me to where the `testing` exporter is being set?

akalia25:
Hey @olafurpg, implemented the changes we discussed during our call. Key highlights are using a delegated testing telemetry exporter to validate events fired. Please let me know if you have any further feedback I can work through 🙇 . 

Not sure why the kotlin bindings are failing 🤔 unable to run the `pnpm generate-agent-kotlin-bindings` command, will investigate/work through this.

agent/src/index.test.ts
```diff
@@ -100,6 +137,14 @@ describe('Agent', () => {
     const ignoredUri = workspace.file('src', 'isIgnored.ts')
 
     it('extensionConfiguration/change & chat/models (handle errors)', async () => {
+        // list of v2 events we expect to fire during the test run (feature:action). Add to this list as needed.
+        client.expectedEvents = [
```
olafurpg, line 141: 😍 Very nice to capture that these events are logged in these cases

agent/src/index.test.ts
```diff
@@ -91,6 +98,36 @@ describe('Agent', () => {
         await client.request('testing/reset', null)
     })
 
+    afterEach(async () => {
```
olafurpg, line 101: Non-blocking: I'm not a huge fan of using `afterEach` here. I think it might be cleaner to

- clear `expectedEvents` on `testing/reset`
- manually add `expect(await client.exportedTelemetryEvents()).toMatch(...)` at the end of test cases where we want to assert what events got exported

The benefit of this approach is that you get clearer error messages pointing to a relevant line of code in the tests. With the `afterEach()` approach, you need to hunt down the test name and reason about this `currentClient` logic. Also, this `afterEach()` block is only available in `index.test.ts`, it's not straightforward how to reuse this logic in other test files. My goal is to eventually remove `index.test.ts` in favor of more narrow/targeted test suites. This file has become a catch-all, which resulted in cryptic test failures that have been difficult to track down.

olafurpg (approved):
Fantastic work @akalia25 ! LGTM 👍🏻 

I left a non-blocking comment about the use of `afterEach()`. We can address this separately, I may even do the change myself as I continue to take more tests out of `index.test.ts`

olafurpg:
@akalia25 FYI, I updated the PR title and PR description to match the style I recommend using in all PRs:

- Title: `WHAT_COMPONENT: WHAT_CHANGED_IN_PRESENT_TENSE`.
- Description: `Previously, EXPLAIN_PROBLEM. Now, EXPLAIN_SOLUTION`.

agent/src/index.test.ts
akalia25: Thank you @olafurpg, this is great feedback, and definitely agree with your approach. I was primarily using the `afterEach` to avoid adding the assertions after each test, but seems like that would be the cleaner way to go. I'll address this in another PR, in favor of getting this merged :)

akalia25: PR here to make the above change :) :
- https://github.com/sourcegraph/cody/pull/5455

PR #4924, opened 2024-07-18
Retry Cody Context Filters policy fetch more aggressively in case of failure

dominiccooney:
Change Cody Context Filters policy retrieval and caching strategy:

- Always treat dotcom as an "include all" policy.
- When retrieval encounters an error, retry fetching the policy sooner.
- ...however do exponential backoff.

Part of CODY-2846, CODY-2809

## Test plan

`pnpm test`


mkondratek (approved):
lgtm

valerybugakov (approved):

PR #4927, opened 2024-07-18
changed(chat): use open sidebar if available

jamesmcnamara:
[CODY-2953](https://linear.app/sourcegraph/issue/CODY-2953/empty-sidebar-chat-is-not-used-for-commands)

If the sidebar chat is opened and unused, default to using that for any chat commands.
## Test plan
Updated E2E test


vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -260,17 +256,24 @@ export class ChatsController implements vscode.Disposable {
...
         contextFiles,
         addEnhancedContext,
         source = DEFAULT_EVENT_SOURCE,
         command,
+        forceNewPanel,
     }: ExecuteChatArguments): Promise<ChatSession | undefined> {
-        const provider = await this.getOrCreateEditorChatController()
+        let provider: ChatController
+        // If the sidebar panel is visible and empty, use it instead of creating a new panel
+        if (!forceNewPanel && this.panel.isVisible() && this.panel.isEmpty()) {
```
abeatrix, line 272: Edge case; what if the activeEditor is also empty and visible 😂

abeatrix (approved):
We used to set all new webview to have "New Chat" as the webview.title, so checking that would tell us if the chat is empty or not but I didn't check if the code is still there or not:

I'm also not sure what the expected behavior for users who prefer sidebar chat over editor, but your change works and aligns with what we have, so lgtm! Thanks for fixing this!

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -260,17 +256,24 @@ export class ChatsController implements vscode.Disposable {
...
      */
-    private async submitChatInNewEditor({
+    private async submitChat({
         text,
         submitType,
         contextFiles,
         addEnhancedContext,
         source = DEFAULT_EVENT_SOURCE,
         command,
+        forceNewPanel,
```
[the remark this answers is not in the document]
jamesmcnamara: If that's the case we should rename those enums because that's not obvious at all. I'll look into it.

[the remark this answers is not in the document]
kalanchan: added a docstring for now, @jamesmcnamara feel free to refactor when you're back!

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -294,17 +290,24 @@ export class ChatsController implements vscode.Disposable {
...
         contextFiles,
         addEnhancedContext,
         source = DEFAULT_EVENT_SOURCE,
         command,
+        
     }: ExecuteChatArguments): Promise<ChatSession | undefined> {
-        const provider = await this.getOrCreateEditorChatController()
+        let provider: ChatController
+        // If the sidebar panel is visible and empty, use it instead of creating a new panel
+        if (submitType === 'user' && this.panel.isVisible() && this.panel.isEmpty()) {
```
kalanchan, line 306: @abeatrix does this logic make sense?

vscode/src/chat/protocol.ts
```diff
@@ -224,6 +224,11 @@ interface ExtensionAttributionMessage {
     error?: string | undefined | null
 }
 
+/**
```
kalanchan, line 227: @abeatrix I added a docstring here to help label what this type means, does it make sense?

kalanchan:
@abeatrix ready for review and merge now. my updates clean up some of your comments, the existing is still as follows:
- if no active panel open, it will open a new editor panel for chat based commands
- if editor panel is open, commands will use that one 

https://github.com/user-attachments/assets/84c338b6-e8ff-475e-93f7-62987b209859

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -294,17 +290,23 @@ export class ChatsController implements vscode.Disposable {
...
         submitType,
         contextFiles,
         addEnhancedContext,
         source = DEFAULT_EVENT_SOURCE,
         command,
     }: ExecuteChatArguments): Promise<ChatSession | undefined> {
-        const provider = await this.getOrCreateEditorChatController()
+        let provider: ChatController
+        // If the sidebar panel is visible and empty, use it instead of creating a new panel
+        if (submitType === 'user-newchat' && this.panel.isVisible() && this.panel.isEmpty()) {
```
kalanchan, line 305: @abeatrix does this make sense?

vscode/src/chat/protocol.ts
```diff
@@ -224,6 +224,11 @@ interface ExtensionAttributionMessage {
...
+ * ChatSubmitType represents the type of chat submission.
+ * - 'user': The user starts a new chat panel.
+ * - 'user-newchat': The user reuses the current chat panel.
+ */
```
kalanchan, line 231: @abeatrix added the docstring here, does this make sense

vscode/src/commands/execute/ask.ts
```diff
@@ -14,6 +14,9 @@ import { getEditor } from '../../editor/active-editor'
 export interface ExecuteChatArguments extends Omit<WebviewSubmitMessage, 'text' | 'editorState'> {
     source?: EventSource
     command?: DefaultChatCommands
+    // We default to running the chat in the main cody sidebar, if it's open and unused.
+    // If you wish to always run the chat in a new webview, set this to true.
+    forceNewPanel?: boolean
```
abeatrix, line 19:
```suggestion
```
do we still need this?

kalanchan: woops good catch ty!

abeatrix (approved):
i tried running a command and confirmed the command shows up in the sidebar correctly: 
image

abeatrix:
the e2e test is currently failing because we don't show the chat view after user clicking on the command which was the expected behavior before. i'm not sure what the expected behavior with the updated change

kalanchan:
Need some design input @toolmantim @rrhyne - so with this PR commands and chat will be start streaming in the active panel that its on. Meaning, if you have the side panel open, chat will run in the side panel, if you have an editor panel open, chat will run in the editor panel.

But now, what happens if you invoke a command from the command tab in the side panel? Should it jump back to the chat tab right afterwards, or should we open an editor panel? 

see video:

https://github.com/user-attachments/assets/62cd23dd-9d79-43fe-8944-be151c1ce16f

toolmantim:
> But now, what happens if you invoke a command from the command tab in the side panel? Should it jump back to the chat tab right afterwards, or should we open an editor panel?

imo just open the chat tab. You don't want to be spawning too many windows from the sidebar imo. Some people will prefer new chat windows, others prefer a single in the sidebar.

abeatrix:
I pushed the change to set view to chat after clicking on a command from the command tab. We should do a follow up to run all the command in a sidebar instead of spinning off a new editor chat. Because right now it feels weird to me that we may switch back to the chat view but the command could have started in an editor window.

kalanchan:
thanks for your help @abeatrix and @jamesmcnamara! I tested this on both PLG and Enterprise, things look good. There are a couple small papercuts around when subsequent commands should use the active panel but we can fix that afterwards. I'll follow up with filing the issues

PR #4931, opened 2024-07-19
Make Context Filters fail closed if the repository is not found

dominiccooney:
Cody Context Filters per RFC 910 should "fail closed" if the repository can't be identified:

> It’s not guaranteed that Cody clients can map an open project in the workspace to a git repo on the Sourcegraph instance. ... clients may end up in a situation where they are not sure about the repo ID on the Sourcegraoh (sic) instance. For such cases, the following behavior is suggested.
> ...
> If “cody.contextFilters” field is defined, we can’t be sure that the opened repo is allowed to be included as context (matches some “include” rule and doesn’t match any “exclude” rule). Thus we exclude this repo from Cody context.

Because `[]` is `true`, then:

```
if (!repoNames) {
  return 'no-repo-found'
}
```

is not effective when the returned repo name list is empty.

## Test plan

Automatic test:

```
pnpm -C lib/shared run test
```

Manual test:

1. Set `cody.contextFilters` in site config and sign in to that site. For example, add this to a site config:

```
  "cody.contextFilters": {
    "include": [
      {
        // Required. Should follow the RE2 syntax.
        "repoNamePattern": "^github\\.com\\/sourcegraph\\/cody.*"
      }
    ],
  }
```

2. The docs say to set the `cody-context-filters-enabled` feature flag to `true`, although I'm unsure that's actually necessary.
3. Clone an unrelated open source repo.
4. Open the repo. Verify that all the contents of the repo are ignored when you try to @-tag them in chat.


valerybugakov (approved):
Thanks for filing a fix for that with unit tests! 

The issue originated in [this PR](https://github.com/sourcegraph/cody/pull/3910/files#diff-282c58c06ab5f90a1cda9f3adae0bd0541bf9a3c4f681d0399037a568cc8c605L38-R43), where the interface of the `getRepoNamesFromWorkspaceUri` method was changed from `Promise<string[] | null>` to `Promise<string[]>` and [the consumer side](https://github.com/sourcegraph/cody/blob/c150947fa51bb84469eb2e2f253213cfc725feda/lib/shared/src/cody-ignore/context-filters-provider.ts#L117-L121) was not properly updated.

PR #4933, opened 2024-07-19
CLI: fix bug in `--show-context`

olafurpg:
Previously, the `--show-context` option in the `cody chat` command didn't actually print out context. This PR fixes the issue for both local files and remote repo files.

## Test plan
See updated test case for `--context-file`. I manually tested it works with `--context-repo` because the test case is skipped due to [CODY-2960 : Re-enable `--context-repo` integration test](https://linear.app/sourcegraph/issue/CODY-2960/re-enable-context-repo-integration-test)


abeatrix (approved):
So much progress!

PR #4934, opened 2024-07-19
Agent: fix name of workspace root

olafurpg:
Previously, the name of the workspace in the Eclipse chat UI was "Workspace Root" instead of the basename of the folder like it is in VS Code. This PR fixes the problem.

Fixes CODY-2986

## Test plan

Confirmed that the directory name is now listed in the chat UI, just like in VS Code

image
image


aramaraju (approved):
👍

PR #4936, opened 2024-07-19
Remove context chips from context retriever queries

janhartman:
This approach supersedes the one from #4920 and is more elegant. We reconstruct the editor state at query time and remove all context chips from the text, then use that text for querying context retrievers.

## Test plan

Tested locally - context retriever queries for symf/bfg no longer include repo or @-mentions.

### DIFF
--- a/lib/shared/src/lexicalEditor/editorState.ts
+++ b/lib/shared/src/lexicalEditor/editorState.ts
@@ -200,6 +200,26 @@ export function contextItemsFromPromptEditorValue(
     return contextItems
 }
 
+export function inputTextWithoutContextChipsFromPromptEditorState(
+    state: SerializedPromptEditorState
+): string {
+    const editorState: typeof state.lexicalEditorState = JSON.parse(
+        JSON.stringify(state.lexicalEditorState)
+    )
+    const queue: SerializedLexicalNode[] = [editorState.root]
+    while (queue.length > 0) {
+        const node = queue.shift()
+        if (node && 'children' in node && Array.isArray(node.children)) {
+            node.children = node.children.filter(child => !isSerializedContextItemMentionNode(child))
+            for (const child of node.children as SerializedLexicalNode[]) {
+                queue.push(child)
+            }
+        }
+    }
+
+    return textContentFromSerializedLexicalNode(editorState.root).trimStart()
+}
+
 export function filterContextItemsFromPromptEditorValue(
     value: SerializedPromptEditorValue,
     keep: (item: SerializedContextItem) => boolean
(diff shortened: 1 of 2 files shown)


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -703,7 +712,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
                                         chatClient: this.chatClient,
                                         chatModel: this.chatModel,
                                     })
-                                  : inputText
+                                  : inputTextWithoutContextChips
```
keegancsmith, line 715: cc @thenamankumar we pass in the original inputText to the rewriter. I suspect this is correct, but just flagging for you to take a look.

lib/shared/src/lexicalEditor/editorState.ts
```diff
@@ -200,6 +200,26 @@ export function contextItemsFromPromptEditorValue(
...
+    state: SerializedPromptEditorState
+): string {
+    const editorState: typeof state.lexicalEditorState = JSON.parse(
+        JSON.stringify(state.lexicalEditorState)
+    )
+    const queue: SerializedLexicalNode[] = [editorState.root]
+    while (queue.length > 0) {
+        const node = queue.shift()
+        if (node && 'children' in node && Array.isArray(node.children)) {
+            node.children = node.children.filter(child => !isSerializedContextItemMentionNode(child))
```
keegancsmith, line 213: we also have a template chip node now, however this will work and I believe will be correct.

lib/shared/src/lexicalEditor/editorState.ts
```diff
@@ -200,6 +200,26 @@ export function contextItemsFromPromptEditorValue(
     return contextItems
 }
 
+export function inputTextWithoutContextChipsFromPromptEditorState(
+    state: SerializedPromptEditorState
+): string {
+    const editorState: typeof state.lexicalEditorState = JSON.parse(
+        JSON.stringify(state.lexicalEditorState)
+    )
```
keegancsmith, line 208: I'll take a look after this PR has landed on how we can maybe simplify these sort of transformations in this function. This is all correct, but is tied to the current serialized version and so makes it somewhat more complicated in the future. But yeah, this is all correct and this feedback is not blocking :)

keegancsmith (approved):

vscode/src/chat/chat-view/ChatController.ts
janhartman: Based on the prompt used to do the rewrite, I'm assuming this should be taken care of. However we're currently not doing the rewrites in prod so I'm not too worried.

lib/shared/src/lexicalEditor/editorState.ts
janhartman: what's a template chip?

keegancsmith: Its new stuff that is coming around how we intend on making shareable prompt libraries. Its the yellow outlined stuff here image

PR #4940, opened 2024-07-19
Update glob in agent (sourcegraph/cody CLI)

ykdojo:
Closes CODY-2707.

## Test plan

In the agent dir, run:

`rm ../pnpm-lock.yaml`

Then:

`pnpm install`

The previous warnings seen as CODY-2707 should have disappeared.

### DIFF
--- a/agent/package.json
+++ b/agent/package.json
@@ -40,7 +40,7 @@
     "easy-table": "^1.2.0",
     "env-paths": "^3.0.0",
     "fast-myers-diff": "^3.2.0",
-    "glob": "^7.2.3",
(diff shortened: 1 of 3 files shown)


olafurpg (approved):
Nice! Cutting a new release is only a matter of bumping up the version in package.json and pushing a git tag, the CI does the rest.

ykdojo:
Cool, should we update the version in this PR then - 5.5.10 -> 5.6.0? (Since we changed the version tag from -V to -v)

ykdojo:
@olafurpg let me know!

Issue #4581, opened 2024-06-17
bug: cody creates unsved files keep showing back after closing


watevadafa:
### Version

1.90.0

### Describe the bug

Version: 1.90.0
Commit: 89de5a8d4d6205e5b11647eb6a74844ca23d2573
Date: 2024-06-04T19:34:48.028Z
Electron: 29.4.0
ElectronBuildId: 9593362
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Darwin arm64 23.5.0

When cody creates files say, test-something.py or any untitled.py, these files if ignored and closed keep showing up after sometime 

### Expected behavior

Once closed it should be closed, or ignored when user opt-out saving them!

waldyrious:
I've also been experiencing this, and it's so frustrating that I had to disable the extension. What's most annoying is that the files are empty, with no undo history, yet I can't even close them without getting the dialog about losing unsaved changes.

felixfbecker:
I am also seeing this since last week or so, I didn't realize it was caused by the Cody extension!

recurser:
Same issue for the last couple of weeks. I've had to disable the extension because of it.

deadband:
I am experiencing the same issue with Cody AI. Anytime a file is modified a new untitled file will be created. Apparently, there is a very similar issue with [here Github Copilot](https://github.com/sourcegraph/cody/issues/4581).

JeongJuhyeon:
I've also been experiencing this, and through "Extension Bisect" confirmed that it only happens with Cody on.

@deadband Currently, your link goes to this issue itself.

justinmilner1:
I'm having trouble replicating this - Are any of you aware of a consistent method leading to this behavior?

It appears the only pathway which _should_ create new files is the 'Generate Unit Tests' command - have you been utilizing this command? 
If not, do you use any other commands?

JeongJuhyeon:
@justinmilner1 

I've even recorded my screen to watch when the files suddenly get opened. 
It's difficult to pinpoint, but it looks related to one or more of the following

- Having had a Chat with Cody in the current session, or at least having a Chat open somewhere in the workspace. I think this is a prerequisite for this to happen.
- Possibly, having mentioned (@) a workspace file in the Chat, or currently being in the process of doing so.
- Saving (or autosaving) of modified code files
- Focus switching to another split view (or possibly just to another tab)

[...]

PR #4942, opened 2024-07-19
feat(cached-retriever): handle non-file scheme URIs in openTextDocument

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-2400/bug-cody-creates-unsved-files-keep-showing-back-after-closing

- Adds a check to `openTextDocument` to return `undefined` if the URI is not a file scheme, which includes untitled files
- Adds a similar check in `onDidChangeTextEditorSelection` to skip handling selection changes for non-file scheme URIs

This ensures the cached retriever does not try to open or handle text documents that are not file-based, including untitled files. As a follow-up task for the team, we should look into improving the cache retriever logic by tracking closed and removed files, but this fix the issues we are seeing with untitled files.

The video below shows that the closed untitled files keep getting re-opened. The added ogging step (during debug) before `openTextDocument` is run shows `openTextDocument` is the issue as opening a non-existing file will create them:

https://github.com/user-attachments/assets/1b170d0e-b213-4740-99d5-1c19dcbe6a50

## Test plan

Try reproducing the issue detailed in https://github.com/sourcegraph/cody/issues/4581 to verify autocomplete no longer re-open closed & unsaved files.

```
Disable all installed extensions (does not seem to be strictly necessary for me, but I include this to show it's not an interaction with another third party extension)
Create a new workspace with a main.py in it
Press Ctrl+N to create a new untitled tab
Close the untitled tab, and notice that it stays closed as expected
Enable Cody
Press Ctrl+N to create a new untitled tab
Close the untitled tab, and notice that it stays closed
Type something in the main.py file to trigger an inline autocomplete suggestion (no need to accept it)
Press Ctrl+N to create a new untitled tab
Close the untitled tab
```


slimsag (approved):

eseliger (commented):
approve by heart but someone closer to the matter should actually approve :)

valerybugakov (commented):
🙌

PR #4943, opened 2024-07-19
VS Code: Release v1.26.6

kalanchan:
Patch release with cherry pick from https://github.com/sourcegraph/cody/pull/4942 to address bug where untitled files were being opened

## Test plan
N/A


abeatrix (approved):
Thank you!

PR #4948, opened 2024-07-19
Chat UI: add custom commands list

abeatrix:
After discussing with kevin, we believe removing the custom commands from the command list would cause discovery issuse. This PR update the Commands tab list Custom Commands if there is any.

- Adds a `CustomCommandsList` component to the `CommandsTab` to display any custom commands the user has defined
- The `CustomCommandsList` is only rendered if there are any custom commands to display

## Test plan

Check if your custom commands are showing up in the commands tab:

image


kalanchan (approved):
looks good!

PR #4950, opened 2024-07-19
fix: support tree view for non-plg users only

abeatrix:
Overlooked this when clause in the last commit that is causing the support sidebar to show up

image

this PR updates the when clause to match the ones for the commands tree view so that it would work as expected

## Test plan

support sidebar not showing up

image

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -244,7 +244,7 @@
         {
           "id": "cody.support.tree.view",
           "name": "Settings & Support",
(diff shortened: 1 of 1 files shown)


kalanchan (approved):

PriNova:
I like the new tab view in the sidebar chat.

What I was missing is the hover text of the right most button that it is for starting a new session.

PR #4953, opened 2024-07-20
fixed(models): allow per-site preferences to be configured for ModelsService

jamesmcnamara:
@abeatrix found a bug with the model defaults when switching accounts. This scopes selections and defaults per endpoint, so that (mostly for us internally) you won't end up submitting a chat with a model that doesn't exist because you've switched accounts.

I also:
* Refactored the cache in ModelsService to make it simpler
* Stored the auth status in ModelsService because we're already calling it on every status update and it simplifies call sites
* Refactored newAuthStatus

## Test plan
Updated unit tests


abeatrix (approved):

abeatrix:
Merging since James is OOO , and @pkukielka reported the same issue for JetBrains side and confirmed this fixed the issue for JB as well. https://sourcegraph.slack.com/archives/C07CVL78MDF/p1721659422200549?thread_ts=1721649252.862329&cid=C07CVL78MDF

PR #4962, opened 2024-07-22
graphql: set Cache-Control no-cache header on requests

keegancsmith:
This is an extra measure to prevent caching behaviour. This is an alternative fix to https://github.com/sourcegraph/cody/pull/4243 and is based on an educated guess. This still needs to be tested.

Test Plan: Hopefully try reproduce with a proxy


keegancsmith:
@arafatkatze do you still have your proxy environment reproduction available? I would be interested to know if this resolves the issue. I am hoping this would instantly resolve any weird networking issues you encountered in the linked PR.

keegancsmith:
Alright so I believe the bug is actually a vscode bug, so your original fix makes sense (adding the timeout). https://github.com/microsoft/vscode/issues/174279

PR #4967, opened 2024-07-22
fix: use backticks for correct string template syntax

bevzzz:
This Cody Command has `''` instead of ``` `` ```, so the `osIcon` is not expanded.

Screenshot 2024-07-22 at 19 54 50

## Test plan

N/A


dominiccooney (approved):
Thank you!

PR #4970, opened 2024-07-22
Cody Web: Allow to set custom telemetry client name

vovakulikov:
Closes https://linear.app/sourcegraph/issue/SRCH-568/create-telemetry-spec-or-look-at-usage-numbers-for-usage-of-cody-in

This PR exposes a custom telemetry client API so consumers of the Cody Web (Sourcegraph client) would be able to set this name directly, it's needed because we should make Cody Web send events with the same clients in which we're running it, (on dotcom it would be `dotcom.web` on enterprise instance it would be `server.web`)

## Test plan
- Run code web demo and check that events have `testing.cody` client name in event's metadata

### DIFF
--- a/web/lib/agent/agent.client.ts
+++ b/web/lib/agent/agent.client.ts
@@ -28,6 +28,7 @@ interface AgentClientOptions {
     serverEndpoint: string
     accessToken: string
     workspaceRootUri: string
+    telemetryClientName?: string
     customHeaders?: Record<string, string>
     debug?: boolean
     trace?: boolean
@@ -38,6 +39,7 @@ export async function createAgentClient({
     accessToken,
     workspaceRootUri,
     customHeaders,
+    telemetryClientName,
     debug = true,
     trace = false,
 }: AgentClientOptions): Promise<AgentClient> {
@@ -82,6 +84,7 @@ export async function createAgentClient({
         extensionConfiguration: {
             accessToken,
             serverEndpoint,
+            telemetryClientName,
             customHeaders: customHeaders ?? {},
             customConfiguration: {
                 'cody.experimental.noodle': true,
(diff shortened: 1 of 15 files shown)


web/lib/components/CodyWebChatProvider.tsx
```diff
@@ -78,6 +79,7 @@ export const CodyWebChatProvider: FC<PropsWithChildren<CodyWebChatProviderProps>
         serverEndpoint,
         accessToken,
         initialContext,
+        telemetryClientName,
```
dadlerj, line 82: Is step 2 after this PR to go update all the callers (e.g. in sourcegraph/sourcegraph)?

web/lib/agent/agent.client.ts
```diff
@@ -38,6 +39,7 @@ export async function createAgentClient({
     accessToken,
     workspaceRootUri,
     customHeaders,
+    telemetryClientName,
```
dadlerj, line 42: With this update, will Cody web inside of the VS Code extension (the webview I think it's called) be able to provide a new client name as well?

dadlerj (approved):
Hey thanks for tagging me, but I'm probably not qualified to review this... I'm not super familiar with the Cody codebase. In particular if there are any gaps. But the logic that I do see all makes sense to me.

I asked a few questions to help me understand what comes next.

web/lib/agent/agent.client.ts
vovakulikov: Correct, in this PR, vscode Cody relies on the standard client name building, but if we need to change it and send some different client name with this PR, it's possible now.

web/lib/components/CodyWebChatProvider.tsx
vovakulikov: Correct, I will update sourcegraph Cody Web client as soon as this is merged

PR #4975, opened 2024-07-23
Dan's CSS change

abeatrix:
Attempted to get the changes out of https://github.com/sourcegraph/cody/tree/dan-jetbrains-css 's branch in this PR

## Test plan

UI chat


vscode/webviews/components/modelSelectField/ModelSelectField.module.css
```diff
@@ -1,30 +1,36 @@
...
-
-    &.disabled {
-        opacity: 0.5;
-    }
-
-    .model-icon {
-        opacity: 0.5;
-    }
+}
+.model-title-with-icon:is(a) {
```
dominiccooney, line 18: Why are we un-SASSing this?

vscode/webviews/components/modelSelectField/ModelSelectField.module.css
```diff
@@ -46,12 +52,14 @@
...
+.model-selector-popover::-webkit-scrollbar {
+    width: 14px;
+    height: 8px;
+    background-color: transparent;
+}
+
+.model-selector-popover::-webkit-scrollbar-thumb {
+    background: rgba(255, 255, 255, 0.15);
+    border-radius: 100px;
+    border: 4px solid var(--jetbrains-Desktop-background);
```
dominiccooney, line 64: This doesn't look right, because what's VSCode going to do here?

dominiccooney (changes requested):
@danielmarquespt we need these CSS changes to not break VSCode in addition to making JetBrains work... did you try these in VS Code? It seems surprising to introduce a `--jetbrains-` variable outside of a `html[data-ide=JetBrains] ...` selector.

It would be great to have a summary of the changes in the sheet, some screenshots, etc.

Here's the architecture we want for Webview CSS, but @vovakulikov and @taiyab feel free to correct me:

- Short term, we keep using the `--vscode-`... variables for colors, etc.
- Web and JetBrains are setting those to specific values, in JetBrains case it is in the `vscode/webviews/themes/jetbrains.css` sheet. I assume Web is just strategically including some other sheet somewhere?
- If more JetBrains-specific customization is needed, it would be by `html[data-ide="JetBrains"]` ancestor selector, but preferably keep this to an absolute minimum.
- Over time, the webview could move to CSS that reflects our own design system and so maybe it uses its own set of variables, not `--vscode-`.

vscode/webviews/components/modelSelectField/ModelSelectField.module.css
danielmarquespt: This is not SASS (AFAIK) its native CSS nesting https://caniuse.com/css-nesting

Which is not supported on the Chromium version of the webviews

vscode/webviews/components/modelSelectField/ModelSelectField.module.css
danielmarquespt: Good catch. Should not be here! I'll fix it

danielmarquespt:
I'm closing these PR as I included most of these changes on `dpc/web-content` taking into account the code review left here.

PR #4988, opened 2024-07-23
VSCode Cody: Reload CTA 

vovakulikov:
Todo
- [x] Implement first pass on UI (done by Tim)
- [ ] Connect Reload UI to actual logic about versions control
- [ ] Found a way to test it without publishing a night build

## Test plan
- N/A

### DIFF
--- a/vscode/webviews/Notices/VersionUpdatedNotice.tsx
+++ b/vscode/webviews/Notices/VersionUpdatedNotice.tsx
@@ -5,6 +5,7 @@ import { getReleaseNotesURLByIDE, majorMinorVersion } from '../../src/release'
 import { Notice } from './Notice'
 
 import type { CodyIDE } from '@sourcegraph/cody-shared'
+import { Button } from '../components/shadcn/ui/button'
 import styles from './VersionUpdatedNotice.module.css'
 
 const key = 'notices.last-dismissed-version'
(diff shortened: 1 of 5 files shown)


vscode/webviews/Notices/VersionUpdatedNotice.tsx
```diff
@@ -39,19 +41,47 @@ interface VersionUpdateNoticeProps {
...
     probablyNewInstall,
+    showRestart = false,
 }) => {
     const [showNotice, setDismissed] = useShowNotice(majorMinorVersion(version), probablyNewInstall)
 
     if (!showNotice) {
         return undefined
     }
 
+    if (showRestart) {
```
vovakulikov, line 59: @abeatrix maybe we don't need this reload button, I was thinking maybe we could check the version in the main activation function and if it's 1.27 or later (as far as I understand starting with the 1.27 version we have a new webview UI) we would trigger web view reload action with `vscode.commands.executeCommand("workbench.action.webview.reloadWebviewAction")` and this should update the webview

vovakulikov: I'm not against having a reload button, But I don't think I have a good condition check for this `showRestart` branch, and I haven't found any hook on the extension update action.

abeatrix: Yea that's what I had in mind too but we will have to reloadWindow instead of reload webview since we need to unregistered the webview from the window! I had it in my other branch but let talk in our 1:1 later!

vovakulikov:
Update from the call with Bee and Kalan, we agreed that we won't do a CTA solution for reloading but will reload the editor automatically once you install the new extension 1.28.0 with new UI to be able to force UI web view update 

CTA solution could produce some "dirty" states when you have web view updated already (because you manually reload you editor and didn't reload it by our CTA reload button).

PR #4989, opened 2024-07-23
VS Code: Reload window on first activation event 

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3007/new-sidebar-ui-does-not-get-picked-up-on-extension-reload

Add a condition to check if `newSidebarChatUI_isReloaded` has been set before or not.

If not, reload the window to ensure the new webview is registered before the extension host is activated. 

After discussing with @kalanchan and @vovakulikov , we think this is a better approach than showing the CTA to ask users to reload without knowing if the user has reloaded or not.

## Test plan

Green CI - The extension should not be affected by this change and work as expected.

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -97,7 +97,7 @@ export async function start(
 ): Promise<vscode.Disposable> {
     // NOTE: Hack to ensure the window is reloaded when extension is restarted after upgrade
     //  to get the updated sidebar chat UI. Can be removed after the next release (1.28).
-    if (!context.globalState.get('newSidebarChatUI_isReloaded', false)) {
+    if (!context.globalState.get('newSidebarChatUI_isReloaded', false) && process.env.CODY_TESTING !== 'true') {
         // First activation, set the flag and then reload the window
         await context.globalState.update('newSidebarChatUI_isReloaded', true)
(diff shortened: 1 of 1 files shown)


vscode/src/main.ts
```diff
@@ -95,6 +95,14 @@ export async function start(
     context: vscode.ExtensionContext,
     platform: PlatformContext
 ): Promise<vscode.Disposable> {
+    // NOTE: Hack to ensure the window is reloaded when extension is restarted after upgrade
+    //  to get the updated sidebar chat UI. Can be removed after the next release (1.28).
+    if (!context.globalState.get('newSidebarChatUI_isReloaded', false)) {
```
abeatrix, line 100: @vovakulikov i removed the version check because the next release and the release after will require reload, so we can just check for the flag. do you think this would work?

vovakulikov: I think it makes sense (at least now) later, this might be unnecessary if there are no changes in the UI and therefore reload is not necessary. It is a real bummer that the VSCode extension API doesn't expose anything close to an update callback or event. Maybe we should file an issue in vscode repo.

vovakulikov (approved):

PR #4998, opened 2024-07-24
Command: add requires setting for auto-edit command

abeatrix:
This change adds a `requires` property to the `CodyCommandMenuItems` configuration for the `cody.command.auto-edit` command. The `requires` property specifies that the `cody.internal.unstable` setting must be enabled for this command to be available.

This ensures that the auto-edit command is only accessible when the user has opted-in to using unstable Cody features.

## Test plan

Auto Edit doesn't show up for enterprise users if they don't have the config enabled

image


kalanchan (approved):
thank you!

PR #5009, opened 2024-07-25
useFeatureFlag hook for getting feature flag value in webview

sqs:
This is not used by anything yet, but it will be used by the impending PR #4903 for the prompt selector (using the prompts in the prompt library).

## Test plan

Run in vscode and web and console.log the values.

### DIFF
--- a/vscode/webviews/utils/useFeatureFlags.tsx
+++ b/vscode/webviews/utils/useFeatureFlags.tsx
@@ -1,0 +1,37 @@
+import type { FeatureFlag } from '@sourcegraph/cody-shared'
+import {
+    type ComponentProps,
+    type FunctionComponent,
+    type ReactNode,
+    createContext,
+    useContext,
+} from 'react'
+
+const FeatureFlagsContext = createContext<Record<string, boolean> | undefined>(undefined)
+
+/**
+ * React context provider whose `value` is the exported feature flags.
+ */
+export const FeatureFlagsProvider: FunctionComponent<{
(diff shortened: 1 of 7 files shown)


sqs:
this is causing some issues when the `App` component is torn down. need to look into it. code review otherwise is appreciated.

vscode/webviews/utils/useFeatureFlags.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * React hook for getting a feature flag's value.
+ *
+ * @returns `true` or `false` if the flag is exposed by the server endpoint, has been fetched, and
+ * is not stale. Otherwise `undefined` (which callers should usually treat as `false`).
+ */
+export function useFeatureFlag(flagName: FeatureFlag): boolean | undefined {
+    const flags = useContext(FeatureFlagsContext)
+    if (flags === undefined) {
```
thenamankumar, line 33: Where do we safeguard this runtime error logically never happens? Given the provider does access Nullable value. I don't see the AppWrapper conditionally rendering the children, either. 

Just want to confirm.

sqs: The thinking is that if it happens, it would be really noticeable in dev and prerelease. The `App` component has an `if (!config ...)` check in it (that existed before this change) and renders the loading dots if the config is not set. So, we should already be protected pretty well from that, and any regressions would be insta-noticed.

keegancsmith (approved):
Thanks for making this now easy to use. I'll look into migrating our reading of the generate unit test feature flag to see how it feels.

PR #5012, opened 2024-07-25
Autocomplete: add analytics event agent tests

valerybugakov:
Follow-up addition with agent tests for https://github.com/sourcegraph/cody/pull/4984. No behaviour changes

## Test plan

New agent tests + CI


agent/src/autocomplete.test.ts
```diff
@@ -31,6 +33,30 @@
             position: { line: 1, character: 4 },
             triggerKind: 'Invoke',
         })
+        const completionID = completions.items[0].id
+        await client.notify('autocomplete/completionSuggested', { completionID })
+        await sleep(400)
```
RXminuS, line 38: We shouldn't actually sleep here. There's timer helpers to advance time

agent/src/autocomplete.test.ts
```diff
@@ -40,20 +66,31 @@
...
         const completions = await client.request('autocomplete/execute', {
             uri: uri.toString(),
             position: { line: 1, character: 4 },
             triggerKind: 'Invoke',
         })
+        const completionID = completions.items[0].id
         const texts = completions.items.map(item => item.insertText)
+
+        await client.notify('autocomplete/completionSuggested', { completionID })
+        await sleep(800) // Wait for the completion visibility timeout (750ms) to elapse
```
RXminuS, line 84: Same here https://vitest.dev/api/vi.html#vi-usefaketimers

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/Constants.kt
```diff
@@ -110,6 +110,8 @@ object Constants {
   const val IDEEXTENSION = "IDEEXTENSION"
   const val ignore = "ignore"
   const val use = "use"
+  const val `tree-sitter` = "tree-sitter"
```
RXminuS, line 113: How is this the only diff in a generated file? That seems suspicious

RXminuS (changes requested):
Mainly the timers

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/CodyAgentServer.kt
```diff
@@ -112,6 +112,8 @@ interface CodyAgentServer {
   fun testing_progressCancelation(params: Testing_ProgressCancelationParams): CompletableFuture<Testing_ProgressCancelationResult>
   @JsonRequest("testing/reset")
   fun testing_reset(params: Null?): CompletableFuture<Null?>
+  @JsonRequest("testing/autocomplete/completionEvent")
+  fun testing_autocomplete_completionEvent(params: CompletionItemParams): CompletableFuture<CompletionBookkeepingEvent?>
```
RXminuS, line 116: Is this needed? Isn't the completion event sent so you can just listen for it?

agent/src/autocomplete.test.ts
valerybugakov: @RXminuS, unfortunately, fake timers don't work across node processes (let me know if you know a way of using them in agent tests; maybe I'm missing something). Several months ago, I started looking into adding support for it but didn't have an opportunity to prioritize this improvement. 

Another process controls the timer I need to advance, so `vitest` helpers do not affect it in a desired way.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/Constants.kt
valerybugakov: This is produced by `pnpm generate-agent-kotlin-bindings`. No manual changes from my side.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/CodyAgentServer.kt
valerybugakov: Without this testing hook, there's no way to get the latest completion event.

agent/src/autocomplete.test.ts
RXminuS: Ah sorry, forgot this was an agent test. I won't let it block then.

We could fairly easily implement it as a testing message or we can start the process with the inspector enabled and we can just advance it externally. Maybe at least open an issue for it so it doesn't get lost?

And maybe add a @slowTimer tag or something on this test title so we can easily find them later?

RXminuS (approved):

agent/src/autocomplete.test.ts
valerybugakov: Added TODO with a comment about this 👍

PR #5015, opened 2024-07-26
extract React context/wrapper components

sqs:
This makes it easier to add new React context or other wrappers without making the JSX tree even more deeply nested. It also eliminates duplication.

## Test plan

CI, `pnpm -C web dev`


keegancsmith (approved):
LGTM, I like this pattern (we need to use it for our http middlewares insg as well)

PR #5017, opened 2024-07-26
Guardrails: update webview on auth change

abeatrix:
Previously we do not update the value for attributions in webview on auth change, so when we updated the test to use sidebar chat instead of panel chat, the test for attributions failed.

## Test plan

Green CI.


keegancsmith (approved):
Thanks!

PR #5019, opened 2024-07-26
Add lenses and edit retry support to agent

pkukielka:
Fixes CODY-2802

## Changes

1. Added support for receiving lenses notifications through agent (plus some fixes for tests)
2. Added two new endpoints: `editTask/getTaskDetails` and `editTask/retry`

## Test plan

Full manual QA together with [corresponding JetBrainsPR](https://github.com/sourcegraph/jetbrains/pull/1955)

### DIFF
--- a/agent/src/TestClient.ts
+++ b/agent/src/TestClient.ts
@@ -475,10 +475,8 @@ export class TestClient extends MessageHandler {
         const task = await this.request('editCommands/document', null)
         await this.taskHasReachedAppliedPhase(task)
         const lenses = this.codeLenses.get(uri.toString()) ?? []
-        if (lenses.length > 0) {
-            throw new Error(
-                `Code lenses are not supported in this mode ${JSON.stringify(lenses, null, 2)}`
-            )
+        if (lenses.length !== 2) {
+            throw new Error(`Expected accept code lenses for ${uri}, but found ${lenses.length}`)
         }
 
         await this.request('editTask/accept', { id: task.id })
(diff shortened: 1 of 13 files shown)


agent/src/TestClient.ts
```diff
@@ -475,10 +475,8 @@ export class TestClient extends MessageHandler {
         const task = await this.request('editCommands/document', null)
         await this.taskHasReachedAppliedPhase(task)
         const lenses = this.codeLenses.get(uri.toString()) ?? []
-        if (lenses.length > 0) {
-            throw new Error(
-                `Code lenses are not supported in this mode ${JSON.stringify(lenses, null, 2)}`
-            )
+        if (lenses.length !== 2) {
+            throw new Error(`Expected accept code lenses for ${uri}, but found ${lenses.length}`)
```
dominiccooney, line 479: "accept" sounds like one thing, not two, why are we looking for exactly two lenses?

Is the interface this brittle in practice? Fine if it is, for now, but let's pour more details in here so people know where to fix & update if they hit this error.

Same feedback below.

agent/src/agent.ts
```diff
@@ -997,6 +999,32 @@ export class Agent extends MessageHandler implements ExtensionClient {
             return null
         })
 
+        this.registerAuthenticatedRequest('editTask/getTaskDetails', async ({ id }) => {
+            const task = this.fixups?.getTask(id)
+            if (task) {
+                return AgentFixupControls.serialize(task)
+            }
+
+            return Promise.reject('No task with id' + id)
```
dominiccooney, line 1008:
```suggestion
            return Promise.reject(`No task with id ${id}`)
```

agent/src/agent.ts
```diff
@@ -997,6 +999,32 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+        })
+
+        this.registerAuthenticatedRequest('editTask/retry', params => {
+            const instruction = PromptString.unsafe_fromUserQuery(params.instruction)
```
dominiccooney, line 1012: This usage is perfect, there's no problem with this usage of `unsafe_fromUserQuery`.

dominiccooney (approved):
Use of `unsafe_fromUserQuery` looks good to me.

agent/src/agent.ts
pkukielka: Done!

agent/src/TestClient.ts
pkukielka: This was poorly written and in fact buggy, as both lenses and task state are notifications, and there are no quittances on the delivery order. 
I rewrote that checks to be meaningful and correct.

PR #5020, opened 2024-07-26
guardrails: experimental configuration of timeouts

keegancsmith:
We are currently experiencing long response times from guardrails servers. We have a hardcoded timeout of 10s. Until the issue is resolved we want to provide a way to workaround the timeout. It is marked experimental since I don't expect this setting to remain in the config forever. Instead when we invest more in the guardrails product I would expect this to live server side (along with other parts of the configuration for it).

Note the configuration key is "cody.experimental.guardrailsTimeoutSeconds"

Test Plan: hardcoded guardrails on to test against s2. Then configured the following and tested guardrails by asking chat to generate a sort implementation of more than 10 lines:
- unconfigured :: default timeout of 10s used
- 0.1 timeout :: timed out
- 40 :: no timeout

Context in https://linear.app/sourcegraph/issue/CODY-2959/cody-guardrail-timeout-on-large-block-of-code#comment-dd245410

### DIFF
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -63,6 +63,7 @@ export interface Configuration {
     experimentalCommitMessage: boolean
     experimentalNoodle: boolean
     experimentalMinionAnthropicKey: string | undefined
+    experimentalGuardrailsTimeoutSeconds: number | undefined
 
     /**
      * Unstable Features for internal testing only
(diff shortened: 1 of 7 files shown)


eseliger (approved):

eseliger (commented):
I wonder if Sourcegraph backend or sourcegraph.com also have timeouts(?)

keegancsmith:
They probably do, but for now atleast 40s in my testing was more than enough time. Also I think the guardrails implementation has changed a bit, since we will show code while the guardrails check runs. IE we don't have to be as aggressive with the timeout (but should still be way faster)

PR #5023, opened 2024-07-26
Pass custom headers to rest API client

vovakulikov:
Fixes https://linear.app/sourcegraph/issue/SRCH-801/supported-llm-rest-api-handler-fails-in-cody-web-client

This PR passes client custom headers to the rest API client to make it work on Cody Web side with Sourcegrph client. 

## Test plan
- At this point, we just need to ensure that CI is passing 
- For full testing, you should 
   - build cody web `cd web` and `pnpm build`
   - link it with `pnpm link --global` running from web directory 
   - go to the sourcegraph repo and update the package.json `cody-web-experimental` package to `0.2.9` version
   - run sg start web-standalone with `SOURCEGRAPH_API_URL: https://sg02.sourcegraphcloud.com` (the staging where we have remote LLMs enabled)
   - Go to the Cody Web page and check that `.api/modelconfig/supported-models.json` call has custom client request headers like `X-Requested-With: Sourcegraph`

### DIFF
--- a/lib/shared/src/sourcegraph-api/rest/client.ts
+++ b/lib/shared/src/sourcegraph-api/rest/client.ts
@@ -32,7 +32,7 @@ export class RestClient {
     // Make an authenticated HTTP request to the Sourcegraph instance.
     // "name" is a developer-friendly term to label the request's trace span.
     private getRequest<T>(name: string, urlSuffix: string): Promise<T | Error> {
-        const headers = new Headers(this.customHeaders as HeadersInit)
+        const headers = new Headers(this.customHeaders)
         if (this.accessToken) {
             headers.set('Authorization', `token ${this.accessToken}`)
         }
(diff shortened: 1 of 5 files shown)


lib/shared/src/sourcegraph-api/rest/client.ts
```diff
@@ -20,16 +20,19 @@ export class RestClient {
...
-        private accessToken?: string
+        private accessToken: string | undefined,
+        private customHeaders: Record<string, string>
     ) {}
 
     // Make an authenticated HTTP request to the Sourcegraph instance.
     // "name" is a developer-friendly term to label the request's trace span.
     private getRequest<T>(name: string, urlSuffix: string): Promise<T | Error> {
-        const headers = new Headers()
+        const headers = new Headers(this.customHeaders as HeadersInit)
```
camdencheek, line 35: nit: this cast looks redundant since `customHeaders` satisfies `HeadersInit`

camdencheek (approved):

lib/shared/src/sourcegraph-api/rest/client.ts
vovakulikov: Missed this, I think we can omit this in other API clients in cody repo too, will keep in mind next time I will be doing something there, 

Remove it here in this PR. thanks

PR #5034, opened 2024-07-29
@-mention git unmerged changes vs. main or uncommitted changes

sqs:
This is an experimental new context provider that lets you @-mention Git information:

- Unmerged changes (`git diff origin/main...HEAD` and `git log origin/main..HEAD`)
- Uncommitted changes (`git diff HEAD`)

It can be used to get "pre-reviews" of your diff.

To use it, enable the `git-mention-provider` feature flag (which is enabled on s2 and for all `sourcegraph` org members on dotcom).

image

## Test plan

Run in a debug VS Code. Try @-mentioning it and ensure it gets the context in the response.


vscode/src/context/openctx/git.ts
```diff
@@ -0,0 +1,224 @@
...
+const DIFF_ARGS = [
+    '-w', // ignore whitespace
+    '--no-prefix', // no 'a/' and 'b/' prefixes
+    '--word-diff=plain',
+    '--color=never',
+    '--diff-algorithm=minimal',
+]
+
+function omitDiffAtAtLines(diff: string): string {
+    // NOTE(sqs): I suspect that the lines like `@@ -27,11 +27,14 @@` are not helpful to the LLM.
```
keegancsmith, line 141: did you notice it misleading the LLM? I assume in its training to understand diffs it came across lots of unified diffs with this information. But that is just my suspicion :)

vscode/src/context/openctx/git.ts
```diff
@@ -0,0 +1,224 @@
...
+    return lines(diff)
+        .filter(line => !line.startsWith('@@ '))
+        .join('\n')
+}
+
+async function getGitInfoForMentions(
+    settings: Pick<Settings, 'cwd'>
+): Promise<{ cwd: FileURI; defaultBranch: string }[]> {
+    const dirs: FileURI[] | undefined = settings.cwd
```
keegancsmith, line 150: what is the point of setting cwd in settings? It doesn't look like this is possible given its a builtin provider.

vscode/src/context/openctx/git.ts
```diff
@@ -0,0 +1,224 @@
...
+        if (!mention) {
+            return []
+        }
+
+        const data = parseMentionURI(mention.uri)
+        if (!data) {
+            return []
+        }
+
+        const cwd = URI.parse(data.cwdURI)
```
keegancsmith, line 70: minor: why even introduce the use of URI here? You control how you marshal cwdURI in the data and you only support fsPath. Unless you think in the future you will support remote file paths, I'd keep it simple and just use a string for the path. This is due to the issues with vscode's URI implementation.

keegancsmith (approved):
life is so much easier for adding builtin context sources now :)

vscode/src/context/openctx.ts
```diff
@@ -41,7 +48,7 @@ export async function exposeOpenCtxClient(
 
 function getStandardOpenCtxProviders(config: ConfigurationWithAccessToken, isDotCom: boolean) {
     // TODO [vk] expose types for providers from openctx/client lib
```
thenamankumar, line 50: remove comment?

thenamankumar (approved):

vscode/src/context/openctx/git.ts
sqs: I noticed the LLM would say things like "As indicated in the line `@@ 24` on line 24, ...", which felt weird.

vscode/src/context/openctx/git.ts
sqs: Yeah...I realized that after.

vscode/src/context/openctx/git.ts
sqs: I do want to support Git diff and log @-mentions on the web where the URIs will not be `file://`, and I didn't want to have to add backcompat for an old way of serializing mentions to chat history in the Lexical editor state where they were fsPaths not URIs.

I did avoid serializing URIs directly (it always calls toString first) due to the issues with that.

vscode/src/context/openctx/git.ts
sqs: (I would like to make our builtin providers slightly less buried so you can set settings on them in the future.)

PR #5036, opened 2024-07-29
Cody Web: Add high-level memoization for chat/transcript UI

vovakulikov:
Attempt to fix https://linear.app/sourcegraph/issue/SRCH-724/prinova-found-an-issue-on-cody-web

The problem that some customers have is somewhere in the react rendering pipeline, I couldn't reproduce this problem from the issue on my machine but I noticed that even though we have `useCallback` and `useMemo` hooks in high-level components we still don't use memoization on components level for anything, this means that the whole chate (all messages) would be re-rendered if one of messages has been changed (by re-rendering I mean react reconciliation). 

This PR 
- Removes props drilling from high-level components (it's easy to track dependencies graph when we don't spread the whole props object to each component)
- Adds `memo()` wrappers for heavy UI components such as transcript interactions and its children UI like human input, assistant message, ..etc) 

Note: I chained this PR with un-related to this PR telemetry fix just to avoid double package versions in NPM (I will merge them in sequence as they will be reviewed)

## Test plan
- Standard high-level manual checks (changes in this PR don't change anything with the current feature set, standard e2e checks should be enough to catch possible regressions)

### DIFF
--- a/vscode/webviews/Chat.tsx
+++ b/vscode/webviews/Chat.tsx
@@ -1,6 +1,6 @@
 import { clsx } from 'clsx'
 import type React from 'react'
-import { useCallback, useEffect, useMemo } from 'react'
+import { useCallback, useEffect, useMemo, useRef } from 'react'
 
 import type { AuthStatus, ChatMessage, CodyIDE, Guardrails } from '@sourcegraph/cody-shared'
 import { Transcript, focusLastHumanMessageEditor } from './chat/Transcript'
@@ -48,6 +48,10 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
 }) => {
     const { reload: reloadMentionProviders } = useChatContextMentionProviders()
     const telemetryRecorder = useTelemetryRecorder()
+
+    const transcriptRef = useRef(transcript)
+    transcriptRef.current = transcript
+
     const feedbackButtonsOnSubmit = useCallback(
         (text: string) => {
             enum FeedbackType {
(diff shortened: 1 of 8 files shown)


vscode/webviews/Chat.tsx
```diff
@@ -48,6 +48,10 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
 }) => {
     const { reload: reloadMentionProviders } = useContextProviders()
     const telemetryRecorder = useTelemetryRecorder()
+
+    const transcriptRef = useRef(transcript)
+    transcriptRef.current = transcript
```
thenamankumar, line 53: If we are updating the ref.current each time, then how is it different from using `transcript` var directly? Just to not have `transcript` as a dependency in `feedbackButtonsOnSubmit = useCallback`? If so, then we can simply remove it from the deps list and use eslint-disable.

I wish I had Cody in the browser to ask this question to.

web/lib/components/CodyWebChatProvider.tsx
```diff
@@ -296,22 +296,32 @@ export const CodyWebChatProvider: FC<PropsWithChildren<CodyWebChatProviderProps>
         [client, vscodeAPI, setLastActiveChatID]
     )
 
+    const contextInfo = useMemo(
```
thenamankumar, line 299: Yeah, React sucks! context API should be doing it out of the box.

vscode/webviews/Chat.tsx
vovakulikov: > If we are updating the ref.current each time, then how is it different from using transcript var directly?

references don't trigger useCallback memoization, so we will still have the same function even though ref.current is changing

 > Just to not have transcript as a dependency in feedbackButtonsOnSubmit = useCallback?
 
 In this case function that we're passing to useCallback will save the first trascript value in its closure scope and we won't have the most recent value when we will call this function.

web/lib/components/CodyWebChatProvider.tsx
vovakulikov: Yeah, they were discussing this at some point https://github.com/reactjs/rfcs/pull/150
dai-shi even got it working via library here `use-context-selecor` https://github.com/dai-shi/use-context-selector?tab=readme-ov-file but I think it never got in production unfortunately

PR #5038, opened 2024-07-29
Chat/Edit: Smart Apply

umpox:
Closes CODY-3200

## Description

This PR:
- Adds a "Smart Apply" option to Chat. This feature:
  - Prompts the LLM to produce a selection in a file, and then prompts the LLM to apply the code block via inline edit
- Prompts the LLM to associate code blocks with file paths, via `language:path/to/file.ts` metadata, and parses that via a remark plugin
- Adds a simpler UI for codeblocks
  - Less hard to understand icons, more words
  - "Copy" and "Apply" are at the forefront
  - "Insert Code at Cursor" and "Save Code to New File" are behind a dropdown menu
  - The filename associated with the codeblock is shown	

### Demo

https://www.loom.com/share/70e3c47cb07e4a69b72515c4f214f76a?sid=5286f9d6-f25b-4d94-b936-8af5eb4f729e

### Screenshots

image

image

## Test plan

1. Enable the feature flag or `cody.experimental.unstable`
2. Check that the new UI appears
3. Check that the "Copy" button works, as well as "Insert Code at Cursor" and "Save Code to New File" work
4. Execute a smart apply on existing code in an expected existing file, check that it works
5. Execute a smart apply on new code in an expected existing file, check that it works
6. Execute a smart apply on new code in an expected new file, check that it works


umpox:
@sourcegraph/cody-clients Would appreciate 👀 on this PR!

It is gated behind a feature flag (`cody-experimental-smart-apply`). I plan to enable this for everyone internally and collect feedback whilst I iterate on this slightly.

Still TODO in follow-up PRs (will create Linear issues):
- Add loading/success/failure indicators, e.g. Apply -> Applying -> Accept/Reject | Failed
- Add Accept/Reject functionality from chat. So you can accept an inline-edit diff directly from the webview.
- Any quality improvements (want people to help dogfood for this)
- Add better support for enterprise instances
  - Currently assumes Claude Sonnet 3.5, but we can't guarantee that for enterprise. Need to make this work well with OpenAI APIs too, and then see how we should enable this
- Enable this through agent too, should in theory work well in JetBrains as it just uses the webview + inline edit, but will need testing

lib/shared/src/prompt/templates.ts
```diff
@@ -14,12 +14,14 @@ export function populateCodeContextTemplate(
...
-            : ps`Codebase context from file {filePath}{inRepo}:\n\`\`\`{languageID}\n{text}\`\`\``
+            : ps`Codebase context from file {filePath}{inRepo}:\n\`\`\`{languageID}{filePathMeta}\n{text}\`\`\``
 
+    const filePath = PromptString.fromDisplayPath(fileUri)
     return template
         .replaceAll('{inRepo}', repoName ? ps` in repository ${repoName}` : ps``)
-        .replaceAll('{filePath}', PromptString.fromDisplayPath(fileUri))
+        .replaceAll('{filePath}', filePath)
         .replaceAll('{languageID}', PromptString.fromMarkdownCodeBlockLanguageIDForFilename(fileUri))
+        .replaceAll('{filePathMeta}', ps`:${filePath}`)
```
umpox, line 24: I wanted to gate this behind a feature flag, but this util is used in so many places that it is quite difficult to do so.

Essentially this means that our context items will include the file path in the Markdown `lang` syntax. This will be helpful for when we're steering the LLM to respond like this too (important for smart apply)

From my testing, it has no noticeable impact on response quality, and it is a very small change. Regardless I will ensure this doesn't make it into the latest release, and has time to run in Insiders before everyone can use this

ErikaRS: It would also be valuable to run it through the eval we have as an additional precaution.

vscode/src/chat/chat-view/prompt.ts
```diff
@@ -72,7 +76,7 @@ export class DefaultPrompter {
                 !this.isCommand &&
                 Boolean(this.explicitContext.length || historyItems.length || this.corpusContext.length)
             ) {
-                reverseTranscript[0] = PromptMixin.mixInto(reverseTranscript[0], chat.modelID)
+                reverseTranscript[0] = PromptMixin.mixInto(reverseTranscript[0], chat.modelID, options)
```
abeatrix, line 79: Do we want to mix the prompt for convo with context only, or do we want to apply it to all chat?

vscode/webviews/chat/ChatMessageContent.tsx
```diff
@@ -264,8 +399,30 @@ export const ChatMessageContent: React.FunctionComponent<ChatMessageContentProps
 
         for (const preElement of preElements) {
             const preText = preElement.textContent
+
             if (preText?.trim() && preElement.parentNode) {
-                const buttons = createButtons(preText, copyButtonOnSubmit, insertButtonOnSubmit)
+                // Extract the <code> element and attached `data-file-path` if present.
+                // This allows us to intelligently apply code to the suitable file.
+                const codeElement = preElement.querySelectorAll('code')?.[0]
+                const fileName = codeElement?.getAttribute('data-file-path') || undefined
```
abeatrix, line 407: 🤡

abeatrix (approved):
I haven't built from the latest commit but the last time I did it worked magically! Maybe we can merge it after tomorrow's release in case of regression?

Super cool feature! Thanks for working on this and letting me be a (tiny) part of it! Can't wait to use it in prod!

abeatrix:
Edit: we should probably test it in JB as well. I can do it tomorrow.

vscode/src/chat/chat-view/prompt.ts
umpox: Hmmm, I didn't think much about this. We probably want it to apply to all chats I think. I noticed the `HEDGES_PREVENTION` also only applies here, do you think we want to apply this to all chats too?

Any pointers on where best to do this would be appreciated, I don't want to screw up our default chat prompts!

umpox:
@abeatrix I have disabled the smart apply (and new UI) when running through the agent, so shouldn't apply to JetBrains or other clients yet... but I want to fix it in a follow up. Hopefully should be easy as it's mostly just a wrapper around inline edit

lib/shared/src/prompt/templates.ts
umpox: I managed to remove this by improving the prompt and, from testing with many different models, everything still works well. Now everything is feature flagged, will chat with @jtibshirani about evaluating the flagged prompt but shouldn't block this PR now!

vscode/src/chat/chat-view/prompt.ts
abeatrix: We only wanted to apply HEDGES_PREVENTION for chats with context to avoid cases where the llm has context to answer the question but starts with "unfortunately" instead.

For this case though it seems like we want to apply it to all chat that has code output, but I'm not sure how it'd affect chats with no code output (or if it has any affect at all). Maybe we could just start with chats with context and then run a eval using the leaderboard with the new prompt before we add it to all chat?

umpox: 👍 Makes sense. I agree, will merge as is and we can start dogfooding and running evals

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -524,8 +529,29 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
...
-    private async getConfigForWebview(): Promise<ConfigurationSubsetForWebview & LocalEnv> {
+    private async isSmartApplyEnabled(): Promise<boolean> {
+        if (!this.authProvider.getAuthStatus().isDotCom) {
+            // Only supported on Sourcegraph.com right now, until we support more than just Claude 3.5 Sonnet.
+            return false
+        }
+
         const config = await getFullConfig()
+        if (config.isRunningInsideAgent) {
+            // Only supported in VS Code right now, until we test and iterate on the UI for other clients.
```
thenamankumar, line 540: A note for the future that this feature should be disabled for Cody Web when the logic is changed later.

umpox: In the longer future, I wonder if we could reuse some of this logic to create a "Diff" view in Cody Web, so users can still get the benefits of seeing exactly what changed but without having to deal with automatically making edits to the file 🤔 

Maybe doesn't fit with the use cases for Cody Web but something to think about

PR #5041, opened 2024-07-29
Consume context from OpenCtx for chat message based on selectors

thenamankumar:
This PR is a follow-up of https://github.com/sourcegraph/openctx/pull/178. 

When a chat message is submitted, it calls the OpenCtx providers with matching selectors to get context items and include them in the context. 

## Test plan

Include the Hello World OpenCtx provider:  "https://openctx.org/npm/@openctx/provider-hello-world": true

Send any message in chat, there should be a openctx context item in the list. 

## Changelog

- Cody will now bring in context from OpenCtx providers for a chat message on submission.

### DIFF
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -32,6 +32,7 @@ import {
     Typewriter,
     allMentionProvidersMetadata,
     featureFlagProvider,
+    getContextForChatMessage,
     graphqlClient,
     handleExtensionAPICallFromWebview,
     hydrateAfterPostMessage,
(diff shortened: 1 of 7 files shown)


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -828,24 +829,31 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
...
+                    input: { text: rewrite, mentions },
+                    addEnhancedContext,
+                    providers: {
+                        localEmbeddings: this.localEmbeddings,
+                        symf: this.symf,
+                        remoteSearch: this.remoteSearch,
+                    },
+                    contextRanking: this.contextRanking,
+                }),
+                getContextForChatMessage(text.toString()),
```
keegancsmith, line 845: this can fail with bad providers (eg bad regex selector), you should handle errors for it gracefully.

Additionally shouldn't you be gating this behind the noodle flag at first?

keegancsmith (approved):

vscode/src/chat/chat-view/ChatController.ts
thenamankumar: I'll handle the errors.

But configuring providers is a manual process anyway. No default included providers provide any context items, anyway, so I think we can skip the noodle flag.

thenamankumar: Handled errors gracefully

PR #5042, opened 2024-07-29
Make cody-bench chat work again after ModelsService changes + upgrade LLM judge model

janhartman:
Due to some recent changes in ModelsService, the API whose usages were removed here wasn't returning any results when run inside cody-bench. Turns out we don't need them at all.

I also upgraded the LLM judge to the latest Claude 3.5 Sonnet. I know this will cause some changes in eval results, but we probably always want the best possible model as the judge.

## Test plan
Tested locally.


olafurpg (approved):

olafurpg:
Opened a PR to fix the failing windows test https://github.com/sourcegraph/cody/pull/5047

PR #5043, opened 2024-07-29
e2e2: github action for experimental e2e

keegancsmith:
Test Plan: CI

Fixes https://linear.app/sourcegraph/issue/CODY-2678/github-action-for-e2e2


keegancsmith:
@RXminuS For now we should probably make this step not be a required check I suppose? Alternatively we make this not triggered by branch push, but maybe just have it running on a cron a few times a day so we can collect data?

PR #5060, opened 2024-07-30
don't implicitly or duplicatively add user selection in chat context

sqs:
It is still added if addEnhancedContext is true to preserve the behavior for older non-webview clients for backcompat.

fix https://linear.app/sourcegraph/issue/SRCH-819/duplicate-context-entries-appearing-when-only-using-the-selection

fix https://linear.app/sourcegraph/issue/CODY-1972/file-selection-is-included-as-context-even-if-the-user-deletes-it

## Test plan

Open editor. Select code. Run chat. Ensure selection is not duplicated. Now run chat without a selection @-mention. Ensure there is no selection used in the context.

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -920,583 +920,6 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: ef38e2fa003d1742e62c4fba7b47c357
-      _order: 0
-      cache: {}
-      request:
-        bodySize: 2062
-        cookies: []
-        headers:
-          - name: content-type
-            value: application/json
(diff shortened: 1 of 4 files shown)


olafurpg (approved):

olafurpg:
CI is passing in my draft https://github.com/sourcegraph/cody/pull/5063 that rebases this PR and skips the failing test cc48a473b4ee53c2556a0481648f8b8e10d88a98

I think it's time to entirely remove support for `addEnhancedContext: true` because it never  worked reliably and we're not using it anywhere once JB has moved to the webview-based chat. The CLI is setting `addEnhancedContext: true` in one critical place to make `--context-repo` work. We need to come up with an alternative JSON-RPC API to set manual context (or come up with a syntax to include the file/repo context in the chat message)

PR #5061, opened 2024-07-30
Make the Agent debugging launch configuration work on Windows.

dominiccooney:
This helps debug Agent during Windows development.

## Test plan

Manual test: In VSCode on Windows, pick the "Launch Agent port 3113" launch configuration and hit Play.


abeatrix (approved):
arigato

PR #5062, opened 2024-07-30
Agent: get `index.test.ts` passing locally

olafurpg:
This PR reverts the changes to `index.test.ts` from the PR https://github.com/sourcegraph/cody/pull/4911

The command `pnpm run test agent/src/index.test.ts` does not pass for me locally after that PR got merged. Removing the updated assertions makes the tests pass again. The `index.test.ts` file is problematic because it tests some cases where there are dependencies between the tests. I've been slowly grinding through moving most of the tests out of this file with the eventual goal of ideally removing it. We should not add new assertions or test cases to this file.

## Test plan

Green CI.

### DIFF
--- a/agent/src/index.test.ts
+++ b/agent/src/index.test.ts
@@ -116,7 +116,8 @@ describe('Agent', () => {
         })
         expect(invalid?.isLoggedIn).toBeFalsy()
         const invalidModels = await client.request('chat/models', { modelUsage: ModelUsage.Chat })
-        expect(invalidModels.models).toStrictEqual([])
+        const remoteInvalidModels = invalidModels.models.filter(model => model.provider !== 'Ollama')
+        expect(remoteInvalidModels).toStrictEqual([])
 
         const valid = await client.request('extensionConfiguration/change', {
             ...client.info.extensionConfiguration,


agent/src/index.test.ts
```diff
@@ -99,11 +99,6 @@ describe('Agent', () => {
         // JetBrains client does and there was a bug where everything worked
         // fine as long as we didn't send the second unauthenticated config
         // change.
-        const initModelName = 'anthropic/claude-3-5-sonnet-20240620'
```
olafurpg, line 102: I didn't spend much time debugging why this breaks the tests but it could be related to recent server-side model config changes. We should avoid hardcoding specific model names like these in tests and rely instead of hardcoded strings from the codebase.

olafurpg:
It's unclear to me why the tests have been passing in CI because I was unable to get these tests passing locally on my computer.

agent/src/index.test.ts
olafurpg: OK, I figured out why this test was failing. I have Ollama installed so the `invalidModels` list was non-empty. Updated the PR to filter out Ollama models.

sqs (approved):

PR #5064, opened 2024-07-31
Cody Web: Improve debouncing mechanism for context items

vovakulikov:
Prior to this PR, `createExtensionAPIProxyInWebview` worked in a way that it was watching the first message from the cody agent with the same command name, this works well enough when the command executes fast (vs code case where we have files locally) but in Cody Web where commands resolving can take a while and during it, we could run new call of `createExtensionAPIProxyInWebview` fetch it may lead to inconsistent result messages when the new call will get a response from the previous command. 

This PR adds `requestId` and `responceId` fields so that we can track and match events from the extension host properly (and don't assign the first response that we get to the most recent event, since ordering might be wrong)

it also deletes debounce promise logic from the extension host since debounce now lives on web view side

## Test plan
- Check Cody Web demo (`cd web & pnpm dev`) that file and repo suggestion work and are debounced properly (300ms)


lib/prompt-editor/src/mentions/mentionMenu/useMentionMenuData.ts
```diff
@@ -18,11 +20,35 @@ export function useMentionMenuParams(): {
...
     const [params, setParams] = useState<MentionMenuParams>({ query: null, parentItem: null })
 
+    const isRemoteLikeProviderActive =
+        mentionSettings.resolutionMode === 'remote' ||
+        params.parentItem?.id === REMOTE_FILE_PROVIDER_URI ||
+        params.parentItem?.id === REMOTE_REPOSITORY_PROVIDER_URI
+
+    // Increase debounce time in case of remote context item resolution (Cody Web case) or
+    // in case of remote-like providers such as remote repositories or remote files
+    const debounceTime: number = isRemoteLikeProviderActive ? 300 : 10
```
thenamankumar, line 33: Does keeping it 300 hurt in any case? I don't want to introduce this logic where we check for particular providers and keep different debounce timings. 

There can be slow OpenCtx providers which prefer 300ms. For example the Web URL provider.

lib/prompt-editor/src/mentions/mentionMenu/useMentionMenuData.ts
```diff
@@ -18,11 +20,35 @@ export function useMentionMenuParams(): {
...
+        () => debounce((query: string | null) => setParams(prev => ({ ...prev, query })), debounceTime),
+        [debounceTime]
+    )
+
     const updateQuery = useCallback(
-        debounce((query: string | null) => setParams(prev => ({ ...prev, query })), 300),
-        []
+        (query: string | null) => {
+            // Update query immediately if it's an initial query state
+            if (!query) {
```
thenamankumar, line 43: Nice!

lib/shared/src/misc/rpc/proxy.ts
```diff
@@ -29,23 +29,30 @@ export function createExtensionAPIProxyInWebview<
...
     type TRequestParams = Omit<Extract<TWebviewMessage, { command: TRequestName }>, 'command'>
     type TResponseValue = Omit<Extract<TExtensionMessage, { type: TResponseName }>, 'type'>
+
+    // Request ID info is used by consumer and will be sent back by the rpc server
+    // in order to determine which message is an actual response for the current
+    // request.
+    let requestIdCount = 1
+
     return (params: TRequestParams): Promise<TResponseValue> => {
+        const requestId = requestIdCount++
```
thenamankumar, line 39: I would not prefer requests to be incremental counters, especially when making cross-process requests. Would timestamps/epochs be better?

thenamankumar (approved):
LGTM overall. Approving to unblock.

lib/prompt-editor/src/mentions/mentionMenu/useMentionMenuData.ts
vovakulikov: I agree; I just noticed that unconditional debounce hurts VSCode extension UX; common providers like files and symbols are available locally, so debouncing from them doesn't make sense. Ideally, provider abstraction should have debounce time as an optional setting to avoid having this condition here. But this would require a bigger change that I don't want to do now 1 week before the release, I will file an issue about this.

lib/shared/src/misc/rpc/proxy.ts
vovakulikov: I think incremental still should be fine because I don't see the case when it will break in this logic. The date timestamp would work here as well; if you think it'd be better, I can change it. Thanks

PR #5067, opened 2024-07-31
Experimental: fetch context from both remote and locally modified source files

beyang:
This modifies behavior when the `cody.internal.serverSideContext` setting is true. If this setting is not set, there should be no behavior changes.

- Remote context is fetched from Sourcegraph, the latest HEAD revision of the repository (which is how enterprise context currently works)
- Local context is fetched from files changed since the merge base of the current local working state and the upstream of the current branch. If the upstream doesn't exist, then the changes since last local commit are used.
- Context from the locally edited files is removed from the remote context to avoid duplication/staleness.
- Currently the local context is limited to 5 entries and concatenated with the remote context. In the future, we should rerank these context items together.

https://github.com/user-attachments/assets/8389821a-1a9a-4a80-866c-4a7040ce4c71

## Test plan

- Set `"cody.internal.serverSideContext": true` in settings
- Make a local change to a file. Then close the file, @-mention the repository, and ask a question about the change you just made *without* @-mentioning the specific file. The answer should answer correctly about the local change.

### DIFF
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -116,10 +116,9 @@ import { InitDoer } from './InitDoer'
 import { getChatPanelTitle, openFile } from './chat-helpers'
 import {
     type HumanInput,
+    codebaseRootsFromHumanInput,
     getContextStrategy,
     getEnhancedContext,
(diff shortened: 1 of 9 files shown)


janhartman (approved):

janhartman:
Great work! Will update with evals ASAP

PR #5068, opened 2024-07-31
Chat: boost `@-file` mentions for open documents

olafurpg:
Previously, when using `@-file` mentions for common filenames in the repo, the ordering of the results wasn't personalised meaning the top result could be a random file (for example, a short file path) instead of a file that the user has open in their editor. This PR adds basic customization where we boost candidates that the user has open in their editor.

## Test plan

Open local build of cody in the sourcegraph repo. Open a file named `openai.go` which is not in cody-gateway, open another file named something else. Open chat and type `@openai.go` and confirm that the top result is the `openai.go` that you have open instead of the one in cody-gateway. Confirm that you can optionally type a space character and `cody-gateway` to narrow the filter to boost the cody-gateway result to the top.

Before
CleanShot 2024-07-31 at 11 36 09@2x

After 
CleanShot 2024-07-31 at 11 36 45@2x

Workaround
CleanShot 2024-07-31 at 11 37 30@2x

## Changelog
When using `@-file` mention in Chat, open documents are now ranked at the top of the fuzzy matcher list.

### DIFF
--- a/vscode/src/editor/utils/editor-context.ts
+++ b/vscode/src/editor/utils/editor-context.ts
@@ -116,14 +116,28 @@ export async function getFileContextFiles(
         threshold: -100000,
     })
 
-    // Apply a penalty for segments that are in the low scoring list.
+    const openDocuments = new Set<string>()
+    for (const uri of getOpenTabsUris()) {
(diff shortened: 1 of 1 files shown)


umpox (approved):
Nice!

PR #5069, opened 2024-07-31
Edit: Strip trailing CR characters when computing diff

umpox:
## Description

Fixes an issue where we would generate incorrect diffs due to trailing CR characters in the users' selected code, and this not being matched by the response from the LLM.

## Test plan

1. Enable end of line sequence CRLF (e.g. in VS Code)
2. Start an edit
3. Check that the diff only shows what changed


olafurpg (approved):
LGTM 👍🏻

PR #5073, opened 2024-07-31
Cody Web: Fix mention provider fetching when user is switching between chats

vovakulikov:
Recently, probably since we changed our logic for global contexts, I've noticed that as you switch between chats in the Cody Web sometimes the context providers like Repositories and Files just disappear from the @ menu. 

This PR fixes this by ensuring that we always listen to these events from the extension host, this PR also removes some old logic about reloading providers (it's no longer needed because providers' state became local)

## Test plan
- Check Cody Web demo (`cd web & pnpm dev`)
- Try to use the @ menu you should see set of three providers Repositories, Files and Symbols
- Change chat (or create new one) you should see the same set of providers

### DIFF
--- a/lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
+++ b/lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
@@ -19,7 +19,8 @@ import {
     KEY_ESCAPE_COMMAND,
     type TextNode,
 } from 'lexical'
-import { type FunctionComponent, useCallback, useEffect, useRef, useState } from 'react'
+import { isEqual } from 'lodash'
+import { type FunctionComponent, memo, useCallback, useEffect, useRef, useState } from 'react'
 import styles from './atMentions.module.css'
 
 import {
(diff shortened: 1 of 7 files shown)


lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
```diff
@@ -19,7 +19,8 @@ import {
     KEY_ESCAPE_COMMAND,
```
vovakulikov, line 19: You can ignore this file change, I just wrapped MentionsPlugin component with `memo` function.

lib/prompt-editor/src/plugins/atMentions/chatContextClient.tsx
```diff
@@ -136,25 +136,39 @@ const EMPTY_PROVIDERS: ContextMentionProviderMetadata[] = []
...
     const client = useContext(ChatContextClientContext)
-    const [providers, setProviders] = useState<ContextMentionProviderMetadata[]>()
+    const [providers, setProviders] = useState<ContextMentionProviderMetadata[]>(EMPTY_PROVIDERS)
+
+    useEffect(() => {
+        // Track if the query changed since this request was sent (which would make our results
+        // no longer valid).
+        let invalidated = false
 
-    const load = useCallback(() => {
```
vovakulikov, line 144: The load and reload function is no longer needed, since this useEffect will be triggered anyway as we switch between chats

web/lib/components/CodyWebChatProvider.tsx
```diff
@@ -29,6 +29,13 @@ import { useLocalStorage } from '../utils/use-local-storage'
  */
 const ACTIVE_CHAT_ID_KEY = 'cody-web.last-active-chat-id'
 
+// Usually CodyWebChatProvider VSCode API wrapper listens messages from Extension host
+// which matches with the current active panel id. But this message id check can be corrupted
+// by race conditions in different events that extension host is sending during switching chats.
+// Some events should be always handled by the client regardless from which active panel they
+// came from.
```
jamesmcnamara, line 36:
```suggestion
// Usually the CodyWebChatProvider VSCode API wrapper listens only to messages from the Extension host
// which matches the current active panel id. But this message id check can be corrupted
// by race conditions in different events that the extension host sends during chat-switching.
// Some events should always be handled by the client regardless of which active panel they
// came from.
```

jamesmcnamara (approved):

PR #5075, opened 2024-07-31
change default keybinding for creating chat in editor to Shift+Alt+L

beyang:
`Alt+L` now toggles between editor and sidebar chat:

- `Alt+L`: think of this as the muscle-memory "chat" shortcut
  - From text editor: opens/focuses the chat sidebar panel
  - From text editor with active selection: adds the active selection to chat
  - From chat sidebar: focuses the last editor panel
  - From chat editor: no-op
- `Shift+Alt+L`:
  - From text editor: opens a new chat editor panel
  - Anywhere else: starts a new chat in the sidebar

This is in line with our goal of using `Alt-{J,K,L}` as the standard shortcuts in Cody.

There is still a bit of jank when invoking `Shift+Alt+L` from a chat editor panel, because this will open a new chat in the *sidebar*. But fixing this will require a bit more work in the form of adding a VS Code context to detect when focus is in chat in an editor panel.

## Test plan

Test the shortcuts locally

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -22,14 +22,9 @@ Chat: Cody is now defaulted to run in the sidebar for both Enterprise and Non-En
 - Autocomplete: Ignores leading empty new lines for autocomplete suggestions to reduce the number of cases when Cody doesn't suggest anything. [pull/4864](https://github.com/sourcegraph/cody/pull/4864)
 - Autocomplete: Preload completions on cursor movement. [pull/4901](https://github.com/sourcegraph/cody/pull/4901)
 - Chat: The shortcuts for starting starting and toggling the chat have changed:
-  - `Alt+L`: think of this as the muscle-memory "chat" shortcut
-    - From text editor: opens/focuses the chat sidebar panel
-    - From text editor with active selection: adds the active selection to chat
-    - From chat sidebar: focuses the last editor panel
-    - From chat editor: no-op
-  - `Shift+Alt+L`:
-    - From text editor: opens a new chat editor panel
-    - Anywhere else: starts a new chat in the sidebar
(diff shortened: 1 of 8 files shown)


sqs (approved):
Approving to unblock since this has some bug fixes

but I don’t like the hotkey definitions as you laid out in the PR description

I think most users will have a persistent personal preference for sidebar vs editor panel chat. We can default to sidebar and if they click the button to toggle to the editor then default to editor. And remember their last used location. These hotkeys would then respect that always.

Alt+L should be like alt tab for chat. It shows and focuses your existing chat if any, otherwise it starts a new chat. How to determine if there is an existing chat? For now we could just make it open any prior chat in the same workspace since they launched the extension, but in the future we could get smarter.

if there is a selection, alt+L should also add the selection to the chat when it shows and focuses it. That is how you can add multiple selections to your chat. 

and so then we need a simple hotkey to open a new chat. My proposal is alt+shift+L

and then separately we should use Cmd not Opt/Alt but that is separate

wdyt?

beyang:
> I think most users will have a persistent personal preference for sidebar vs editor panel chat. We can default to sidebar and if they click the button to toggle to the editor then default to editor. And remember their last used location. These hotkeys would then respect that always.
>
> Alt+L should be like alt tab for chat. It shows and focuses your existing chat if any, otherwise it starts a new chat.
>
> if there is a selection, alt+L should also add the selection to the chat when it shows and focuses it. That is how you can add multiple selections to your chat.
>
> and so then we need a simple hotkey to open a new chat. My proposal is alt+shift+L

Agree, implemented:

https://github.com/user-attachments/assets/f6765229-32be-4558-9e1f-780705ad1592
https://github.com/user-attachments/assets/8f32c65c-8405-4503-959e-167a88665d10

There is still some jank in adding duplicate selections, but I will leave that for another PR (it's an existing issue, not one introduced here)

PR #5077, opened 2024-08-01
Cody Web: Add range support for remote file mentions

vovakulikov:
Part of https://linear.app/sourcegraph/issue/SRCH-561/add-support-for-file-ranges-in-cody-web

This PR adds support for file ranges in remote file mentions, prior to this there was a problem that after you submitted too large a file the mention wasn't properly added (The @ sign was removed and `:` range column was added but nothing more) This PR fixes this too, (I think this problem existed in VSCode too, but I couldn't find a too large file in VSCode build) 

## Test plan
- Check that mentioning too big files requires `:` ranges flow and check that this flow works properly with Cody Web demo

### DIFF
--- a/lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
+++ b/lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
@@ -164,11 +164,14 @@ export const MentionsPlugin: FunctionComponent<{ contextWindowSizeInTokens?: num
                     if (isLargeFile && !selectedItem.range) {
                         const textNode = $createContextItemTextNode(selectedItem)
                         nodeToReplace.replace(textNode)
-                        const colonNode = $createTextNode(':')
+
                         // Keep at symbol because we're still in the editing mode
                         // (since ranges haven't been presented yet)
                         textNode.insertBefore($createTextNode('@'))
+
+                        const colonNode = $createTextNode(':')
                         textNode.insertAfter(colonNode)
+
                         colonNode.select()
                     } else {
                         const mentionNode = $createContextItemMentionNode(selectedItem)
--- a/lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
+++ b/lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
@@ -165,6 +165,9 @@ export const MentionsPlugin: FunctionComponent<{ contextWindowSizeInTokens?: num
                         const textNode = $createContextItemTextNode(selectedItem)
                         nodeToReplace.replace(textNode)
                         const colonNode = $createTextNode(':')
+                        // Keep at symbol because we're still in the editing mode
+                        // (since ranges haven't been presented yet)
+                        textNode.insertBefore($createTextNode('@'))
                         textNode.insertAfter(colonNode)
                         colonNode.select()
                     } else {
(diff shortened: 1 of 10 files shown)


lib/prompt-editor/src/plugins/atMentions/atMentions.tsx
```diff
@@ -165,6 +165,9 @@ export const MentionsPlugin: FunctionComponent<{ contextWindowSizeInTokens?: num
                         const textNode = $createContextItemTextNode(selectedItem)
                         nodeToReplace.replace(textNode)
                         const colonNode = $createTextNode(':')
+                        // Keep at symbol because we're still in the editing mode
+                        // (since ranges haven't been presented yet)
+                        textNode.insertBefore($createTextNode('@'))
```
thenamankumar, line 170: Small thing, can me move this 1 line up, so `const colonNode = $createTextNode(':')` is right before `textNode.insertAfter(colonNode)`

vscode/src/chat/context/chatContext.ts
```diff
@@ -42,12 +42,17 @@ export async function getChatContextItemsForMention(
         case FILE_CONTEXT_MENTION_PROVIDER.id: {
             telemetryRecorder?.withProvider(mentionQuery.provider)
             const files = mentionQuery.text
-                ? await getFileContextFiles(mentionQuery.text, MAX_RESULTS, remoteRepositoriesNames)
+                ? await getFileContextFiles(
```
thenamankumar, line 45: It feels like the right time to refactor the `getFileContextFiles` args and accept an object. Rather than having the second arg as nullable. 

```
getFileContextFiles({
   query: mentionQuery.text,
   range: mentionQuery.range,
   maxResults: MAX_RESULTS,
   repoNames: remoteRepositoriesNames
})
```

vscode/src/editor/utils/editor-context.ts
```diff
@@ -62,15 +64,19 @@ export async function getFileContextFiles(
 
     // TODO [VK] Support fuzzy find logic that we have for local file resolution
     if (repositoriesNames) {
-        const filesOrError = await graphqlClient.getRemoteFiles(repositoriesNames, query)
+        const [filePath] = query?.split(':') || []
```
thenamankumar, line 67: If we expect `:` followed by range as part of the query then why we need to get the range from args, can't we extract it here only?

vscode/src/editor/utils/editor-context.ts
```diff
@@ -433,14 +439,22 @@ async function resolveFileOrSymbolContextItem(
         const resultOrError = await graphqlClient.getFileContent(repository, path)
 
         if (!isErrorLike(resultOrError)) {
+            const range = contextItem.range
+            const content = range
```
thenamankumar, line 443: `graphqlClient.getFileContent` accepts `range` arg. I think we can use that and skip doing this range filtering on the client side. Take a look at `GET_REMOTE_FILE_QUERY`.

thenamankumar (approved):
Overall looks good. Left a few minor comments.

vscode/src/editor/utils/editor-context.ts
vovakulikov: So the reason I added file path extraction here is that this is how I filtered file path and range text before passing the file path to the remote resolver, if pass just a query (which may contain ranges) the remote resolver obviously won't find a file with `<filename>:1-2`

> then why we need to get the range from args, can't we extract it here only?

It looks like there is a bug in the range suffix extraction in the client at the moment, even though we have `:` separator it doesn't extract it from the query and leaves it in the query text, however, it parses ranges correctly, this is why I sanitize query and use ranges from args (didn't want to call parsing function here again)

vscode/src/editor/utils/editor-context.ts
vovakulikov: Yeah, I completely forgot about that. I added it at some point; thanks, nice catch!

PR #5079, opened 2024-08-01
Shipment of deepseek-v2-lite-base on autocomplete for PLG users

hitesh-1997:
## Context
1. This PR adds a feature flag `cody-autocomplete-deepseek-v2-lite-base` to ship the deepseek-v2-lite-base model. For now, the model will be shipped for all the PLG users.
2. Sends a `anonymousUserId` fields in the fireworks payload. This payload is send to fireworks so that each user can be redirected to the same deployment and can leverage prompt caching.

## Test plan
1. Override myself in the feature flag and ensure we are hitting `deepseek` model when on dotCom instance.
2. Override myself in the feature flag and ensure we are hitting `starcoder` model when on S2 instance.
3. Checked that we pass `anonymousUserId` correctly to CG.

## Changelog
1. Shipping `deepseek-coder-v2-lite-base` as the default autocomplete model for PLG users.
2. Add `anonymousUserId` in the CG payload to leverage fireworks prompt caching.

### DIFF
--- a/vscode/src/completions/providers/fireworks.ts
+++ b/vscode/src/completions/providers/fireworks.ts
@@ -59,6 +59,7 @@ export interface FireworksOptions {
     model: FireworksModel
     maxContextTokens?: number
     client: CodeCompletionsClient
+    anonymousUserID?: string
     timeouts: AutocompleteTimeouts
     config: Pick<
         ConfigurationWithAccessToken,
@@ -182,11 +183,11 @@ class FireworksProvider extends Provider {
     // Todo: This variable is used to introduce an additional delay to collect the data on impact of latency on user experience.
     // Todo: Delete this variable once the data is collected.
     private shouldAddArtificialDelayForExperiment = false
+    private anonymousUserID: string | undefined
 
     constructor(
         options: ProviderOptions,
-        { model, maxContextTokens, client, timeouts, config, authStatus }: Required<FireworksOptions>
-    ) {
+        { model, maxContextTokens, client, timeouts, config, authStatus, anonymousUserID }: Required<Omit<FireworksOptions, 'anonymousUserID'>> & { anonymousUserID?: string }    ) {
         super(options)
         this.timeouts = timeouts
         if (model === FIREWORKS_FIM_FINE_TUNED_MODEL_HYBRID_WITH_200MS_DELAY) {
@@ -197,6 +198,7 @@ class FireworksProvider extends Provider {
         this.promptChars = tokensToChars(maxContextTokens - MAX_RESPONSE_TOKENS)
         this.client = client
         this.authStatus = authStatus
+        this.anonymousUserID = anonymousUserID
         this.isLocalInstance = Boolean(
             this.authStatus.endpoint?.includes('sourcegraph.test') ||
                 this.authStatus.endpoint?.includes('localhost')
@@ -537,6 +539,7 @@ class FireworksProvider extends Provider {
                     ],
                     stream: true,
                     languageId: self.options.document.languageId,
+                    anonymousUserID: self.anonymousUserID
                 }
 
                 const headers = new Headers(self.getCustomHeaders())
@@ -722,7 +725,7 @@ export function createProviderConfig({
 }): ProviderConfig {
     const clientModel =
         model === null || model === ''
-            ? DEEPSEEK_CODER_V2_LITE_BASE
+            ? (otherOptions.authStatus.isDotCom ? DEEPSEEK_CODER_V2_LITE_BASE : 'starcoder-hybrid')
             : ['starcoder-hybrid', 'starcoder2-hybrid'].includes(model)
               ? (model as FireworksModel)
               : Object.prototype.hasOwnProperty.call(MODEL_MAP, model)
--- a/vscode/src/completions/providers/fireworks.ts
+++ b/vscode/src/completions/providers/fireworks.ts
@@ -187,7 +187,16 @@ class FireworksProvider extends Provider {
 
     constructor(
         options: ProviderOptions,
-        { model, maxContextTokens, client, timeouts, config, authStatus, anonymousUserID }: Required<Omit<FireworksOptions, 'anonymousUserID'>> & { anonymousUserID?: string }    ) {
+        {
+            model,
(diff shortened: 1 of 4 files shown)


vscode/src/completions/providers/fireworks.ts
```diff
@@ -722,7 +722,7 @@ export function createProviderConfig({
 }): ProviderConfig {
     const clientModel =
         model === null || model === ''
-            ? 'starcoder-hybrid'
+            ? DEEPSEEK_CODER_V2_LITE_BASE
```
valerybugakov, line 725: I suspect this can affect the setup for enterprise users if they don't have a model specified in the site config. We should verify that. We can use `authStatus.isDotCom` to apply the new default value conditionally.

hitesh-1997: Thanks for flagging this valery, changed the updated the condition to check `dotCom` users.

janhartman (approved):

aramaraju:
@hitesh-1997 I want @slimsag or @chrsmith to also give a quick run over these changes given the server side selection changes we are bringing in.

aramaraju (approved):
Everything looks good to me, added a few more reviewers to bless this as the autocomplete code is a tangled web of nuances

slimsag:
Fine for now; but when we switch over to server-side models on Sourcegraph.com this will stop working and we'll either need to (a) make the default just deepseek-v2-lite-base for _all_ Gateway users, or (b) implement feature flags on the backend in the Go code for this.

hitesh-1997:
> Fine for now; but when we switch over to server-side models on Sourcegraph.com this will stop working and we'll either need to (a) make the default just deepseek-v2-lite-base for _all_ Gateway users, or (b) implement feature flags on the backend in the Go code for this.


Thanks for adding this @slimsag 
I think it is important to address this issue because if the model switch to starcoder automatically without respecting what client sends it, the traffic will be redirected to starcoder, which don't have enough capacity and we would start seeing more issues. I will try to look into the server side and make relevant changes.

Meanwhile, Can you please let me know the timeline when is this going to kick-off. If it is couple of weeks away, we can merge this PR and test model on PLG so that enterprise rollout is not blocked.

hitesh-1997:
> > Fine for now; but when we switch over to server-side models on Sourcegraph.com this will stop working and we'll either need to (a) make the default just deepseek-v2-lite-base for _all_ Gateway users, or (b) implement feature flags on the backend in the Go code for this.
> 
> Thanks for mentioning this @slimsag I think it is important to address this issue because if the model switch to starcoder automatically without respecting what client sends it, the traffic will be redirected to starcoder, which don't have enough capacity and we would start seeing more issues. I will try to look into the server side and make relevant changes.
> 
> Meanwhile, I can merge this PR and test model on PLG so that enterprise rollout is not blocked and in parallel make server side changes.

@slimsag If I understand correctly from [this doc](https://docs.google.com/document/d/1HqQL2WafRlEZgA3t4U4kmC8auL-DHB4sPYoS57q4EK8/edit) the backend will support both the older format and new format for next couple of months, please correct me if I am wrong here.
1. @aramaraju Checking if we decide to deprecate the older format, where client hardcodes the model string, we would need to make relevant client changes. Do we know the POC for that ?

>  we switch over to server-side models on Sourcegraph.com

1. @slimsag when we say that we switch over to "server-side models" If I understand if correctly, the timeline is few months for this and until than the older method from client would work as is. Is the default model override applicable for both PLG and enterprise ? (since we want to roll out deepseek to PLG first and than to enterprise)

slimsag:
Hey @hitesh-1997 - thinking about this a bit:

This solution (client-side feature flag solution) will work fine:

* For users of Sourcegraph.com as long as we don't switch to the new model configuration system. This is not scheduled/planned yet other than 'soon'. I will make sure Cody Prime team thinks about this before making this change.
* ..as long as Cody Gateway isn't updated to the new model system - similar to Sourcegraph.com we expect this to happen 'soon' but no concrete timeline yet. I will make sure Cody Prime team thinks about this before making this change.
* For on-prem customers as long as they don't switch to the new model configuration system. This impacts you most today. After the Aug 7th release, customers can opt-in (early-access-program) to the new model config system, and if they do so their default autocomplete provider will currently be starcoder. Currently, I think it is not (easily) possible to use deepseek via the new model configuration system - we would need to enable it (details below) - the default would be starcoder, and after the Sept 7th release I think this will become the default configuration so we'll want to figure this out before then probably.

The TL;DR is that with server-side models, the server gives the client a big list of models which it can use - and the client only uses models from that list. Cody gateway will be updated to use this list as well, so everything is coming from one central source - and Sourcegraph instances will in the future automatically poll for this updated list from Cody Gateway so that users can get new models without updating their client or server version. The client should also just use the defaults that the server tells it for the same reason, so there should be no client-side feature-flag logic that determines the default models in the new world.

[This generated models.json file](https://github.com/sourcegraph/sourcegraph/blob/60038f44b1ad2f2591b20941835c546975ab236c/internal/modelconfig/embedded/models.json) defines all of the models that are available, to make it possible to use deepseekv2 when using the new configuration system we'll need to add deepseekv2 to this list. You can also see where we set the default autocomplete model in the `defaultModels` field at the bottom of that file. Cody Prime team can own this work if you like.

Currently, we don't have a way to e.g. dynamically control the default models, since it comes from that static file - but we could add this easily once we add more dynamic parts to this system. I think we should discuss what requirements you have here (for example, we may be able to use a standard backend feature-flag for controlling the default models PLG users get.. but for on-prem instances we can do things differently, etc. - I think your input on what sort of control we need here would be very valuable insight for what kind of control knobs we need here!)

Does that make sense? Any questions? Happy to have a call to discuss some of this too if that helps

aramaraju:
@hitesh-1997 Let's have the Cody Prime team own the rollout of deepseekv2 for Enterprises given this transition period between the old and new config model. I will work with them as I hash out the rollout plan in September

hitesh-1997:
> @hitesh-1997 Let's have the Cody Prime team own the rollout of deepseekv2 for Enterprises given this transition period between the old and new config model. I will work with them as I hash out the rollout plan in September

Sure, Thanks Aravind !!

PR #5083, opened 2024-08-01
use streaming RPC API exthost<->webview using AsyncGenerator

sqs:
Now, the webview can call into API methods exposed by the ChatController (extension host), and those API methods can return a stream of values. This lets the webview "observe" things in the extension host.

Fixes several issues:

- OpenCtx @-mention providers now load immediately when each is ready, without waiting for all to be ready. We had a 5-second hardcoded delay for providers, which meant that in many cases the first chat session would not have any OpenCtx providers. Also, if you change your configuration, the new providers appear immediately in all chats.
- Feature flags used in the webview with `useFeatureFlag` now load properly. This fixes an issue where the Prompts dropdown was not displaying, even for users for whom it was enabled.
- Context items now stream in, which means that an OpenCtx provider can emit multiple values as it loads more and more data.
- Webview<->exthost API calls are now aborted automatically (when using the `useAsyncGenerator` React hook).

## Test plan

CI. Also a lot of manual usage of @-mentions and the Prompts dropdown, since those are the parts that use this new API.


lib/shared/src/misc/asyncGenerator.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        expect(calls).toBe(0)
+
+        const signal = new AbortController().signal
+        expect(await allValuesFrom(memoized(0, signal))).toEqual<number[]>([0, 1])
+        expect(calls).toBe(1)
+
+        expect(await allValuesFrom(memoized(10, signal))).toEqual<number[]>([10, 11])
+        expect(calls).toBe(2)
+
+        expect(await allValuesFrom(memoized(0, signal))).toEqual<number[]>([1, 0, 1])
```
dominiccooney, line 150: 🤯

lib/shared/src/misc/asyncGenerator.test.ts
```diff
@@ -0,0 +1,181 @@
...
+        expect(calls).toBe(3)
+    })
+})
+
+describe('omitDuplicateSequences', () => {
+    test('should omit duplicate consecutive values', async () => {
+        const input = asyncGeneratorWithValues(1, 1, 2, 2, 2, 3, 4, 4, 5)
```
dominiccooney, line 157: Worth testing that non-consecutive duplicate values are not filtered

```suggestion
        const input = asyncGeneratorWithValues(1, 1, 2, 2, 2, 1, 3, 4, 4, 5)
```

lib/shared/src/misc/asyncGenerator.ts
```diff
@@ -0,0 +1,269 @@
...
+            return
+        }
+        throw error
+    } finally {
+        disposable.dispose()
+    }
+}
+
+/** Sentinel value. */
+const ABORTED = { aborted: true }
```
dominiccooney, line 100: Consider freezing this.

dominiccooney (commented):
Some inline comments. Need to sit and read this carefully, I see it is still changing.

High level, yes, the API exposing fields as a streams of updates looks great and the abortable signals and combinators design looks very nice.

lib/prompt-editor/src/useAsyncGenerator.ts
```diff
@@ -0,0 +1,78 @@
+import { useEffect, useRef, useState } from 'react'
+
+export type UseAsyncGeneratorResult<T> = {
+    value: T | undefined
+    loading: boolean
+    complete: boolean
```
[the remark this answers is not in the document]
sqs: Thanks, good points.

- In AsyncGenerators, `error` is a terminal outcome, so there can't be multiple errors.
- Are you thinking `done: boolean | Error`? (I should use done instead of complete because that matches the AsyncGenerator API.) That would work, but I'm not sure it's clearer.
- None of these except `value` and `error` are used right now (except in tests), so I could just remove them. I'll remove `loading` for sure since it is not an AsyncGenerator concept.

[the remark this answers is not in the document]
sqs: How about this: remove `loading`, and use a union type so that the type system encodes the linkage of `done` and `error`?

sqs:
@dominiccooney Thanks, I just addressed your comments and re-pushed.

sqs:
@dominiccooney OK to merge and do a post-merge review? This will fix the feature flag listener problem and unblock more testing of the prompt library prior to the release.

dominiccooney:
@sqs review post merge sounds good to me. We should work out with search team who is on point for bigger changes like this so they can really own the result.

vscode/webviews/utils/useFeatureFlags.tsx
```diff
@@ -1,37 +1,20 @@
...
  * @returns `true` or `false` if the flag is exposed by the server endpoint, has been fetched, and
  * is not stale. Otherwise `undefined` (which callers should usually treat as `false`).
  */
-export function useFeatureFlag(flagName: FeatureFlag): boolean | undefined {
-    const flags = useContext(FeatureFlagsContext)
-    if (flags === undefined) {
-        throw new Error('useFeatureFlags must be used within a FeatureFlagsProvider')
-    }
-    return flags[flagName]
+export function useFeatureFlag(flag: FeatureFlagUsedInWebview): boolean | undefined {
```
dominiccooney, line 12: Not for this PR specifically but this highlights some of the opportunities with signals:

Is it OK to assume feature flags, or a subset of them, are fixed? It would be nice, on the flag producer side, to make it crystal clear that those flags are produced *once.* For the "dynamic" flags, it would be great to *have* to consume the generator and make this convenience function unavailable.

This will ensure we don't have any startup.... read-fixed-flag-before-published races and any runtime read-dynamic-flag-but-didn't-update-it bugs.

dominiccooney (approved):
Approving to unblock.

vovakulikov:
Before merging this can someone check Cody web demo. I can check it later today myself but maybe it would be too late since this has been approved already

sqs:
@vovakulikov I’ve checked it and it works

PR #5085, opened 2024-08-01
Use server proxy for web mentions in Cody Web

camdencheek:
This updates the web context provider to optionally use the new endpoint in the GraphQL API (introduced [here](https://github.com/sourcegraph/sourcegraph/pull/64223)) to fetch web context in cody web to get around CORS issues.

## Test plan

Tested dev build against local Sourcegraph instance. Also tested that web mentions still work in VSCode.
screenshot-2024-08-01_14-40-53@2x

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ playwright
 playwright-report
 test-results
 .vscode-test/
+.envrc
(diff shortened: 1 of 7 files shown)


vovakulikov (approved):

vovakulikov:
@camdencheek merging this since we still need to bump the version in the Sourcegraph client. Change seems very safe since only Cody Web uses new Web URL mention logic

sqs:
this is awesome, thanks!!

PR #5088, opened 2024-08-01
Agent: stop generating Kotlin bindings for the webview protocol

olafurpg:
Previously, the Kotlin bindings generated classes for
`webview/postMessage`, `weview/receiveMessage` and other webview-related
JSON-RPC endpoints. Now that our IDE clients have moved to using webview-powered
chat, there's no need to generate bindings for these classes because the
IDE clients only need to proxy the raw JSON payloads instead of
interaacting with the low-level WebView protocol.

Removing these types from the bindings means Cody developers no longer
need to run `pnpm generate-agent-kotlin-bindings` when they update the
webview protocol.

## Test plan
Green CI.


RXminuS (approved):
Think this is ok for JB despite using auto-generated messages and not webview yet. But  we still have tge hand-coded protocol messages so as long as message handlers stay in place for now I don't think it'd break anything.

And by the time we finish of protocol migrations we're hopefully using webviews. @pkukielka thougts?

sqs (approved):

dominiccooney (approved):
Thank you 🎉 .

dominiccooney:
@RXminuS it is OK for JetBrains because JetBrains doesn't use these generated types, but has checked in and hand written/edited them. Although there was a bunch of copying and pasting from these generated definitions to do that.

olafurpg:
@RXminuS when you extend `CodyAgent{Client,Server}` you can add new handlers with your own types (which can be old copies of the auto-generated classes). You just won't get auto-generated classes for new webview-protocol messages.

PR #5091, opened 2024-08-02
Cody Web: Expose internal create chat API from Cody Web root component

vovakulikov:
Part of [SRCH-720](https://linear.app/sourcegraph/issue/SRCH-720/new-chat-button-in-side-panel-view)

This PR simply exposes the internal CodyWebProvider API so a consumer would be able to call it proactively (in this particular case to trigger chat creation outside of CodyWebProvider children components)

## Test plan
- No manual testing is required (pure tech PR)

### DIFF
--- a/web/CHANGELOG.md
+++ b/web/CHANGELOG.md
@@ -1,5 +1,6 @@
-## 0.3.4
+## 0.3.5
 - Adds support for URL mentions
+- Adds support for ref-like API in CodyWebChatProvider level
 
 ## 0.3.4 
 - Adds support for remote file ranges  
--- a/web/lib/components/CodyWebChatProvider.tsx
+++ b/web/lib/components/CodyWebChatProvider.tsx
@@ -68,7 +68,7 @@ export const CodyWebChatContext = createContext<CodyWebChatContextData>({
     selectChat: () => Promise.resolve(),
 })
 
-export interface CodyWebChatContextRef {
+export interface CodyWebChatContextClient {
     createNewChat: () => Promise<void>
 }
 
@@ -80,6 +80,7 @@ interface CodyWebChatProviderProps {
     initialContext?: InitialContext
     customHeaders?: Record<string, string>
     onNewChatCreated?: (chatId: string) => void
+    onClientCreated?: (client: CodyWebChatContextClient) => void
 }
 
 /**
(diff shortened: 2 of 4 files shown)


web/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-## 0.3.4
+## 0.3.5
 - Adds support for URL mentions
```
vovakulikov, line 2: There was a type, in previous PR it wasn't updated properly, but this is ok because this package version hasn't been published yet

web/lib/components/CodyWebChatProvider.tsx
```diff
@@ -305,6 +315,18 @@ export const CodyWebChatProvider: FC<PropsWithChildren<CodyWebChatProviderProps>
         [client, vscodeAPI, setLastActiveChatID]
     )
 
+    useImperativeHandle(
+        ref,
+        () => ({
+            createNewChat: createChat,
+        }),
+        [createChat]
+    )
```
camdencheek, line 324: Just making sure I understand: this block is just customizing the ref to expose minimal API surface (only `createNewChat`) to consumers of the `CodyWebProvider`?

web/lib/components/CodyWebChatProvider.tsx
```diff
@@ -305,6 +315,18 @@ export const CodyWebChatProvider: FC<PropsWithChildren<CodyWebChatProviderProps>
...
+
+    useEffect(() => {
+        onClientCreated?.({ createNewChat: createChat })
+    }, [onClientCreated, createChat])
```
camdencheek, line 328: I am not quite following what this is for. It doesn't look like `onClientCreated` is actually used anywhere?

camdencheek (approved):

web/lib/components/CodyWebChatProvider.tsx
vovakulikov: Sure, so in the consumer would be possible to do this 
```
const codyWebAPI = useRef()

// codyWebAPI.current.createNewChat()

return (
   <CodyWebChatProvider ref={codyWebAPI} />
)
```

web/lib/components/CodyWebChatProvider.tsx
vovakulikov: This is basically doing the same thing that useImperativeHandle does. 
The problem with ref is that in the Svelte world (where we also use this component at the moment), there isn't a good way to work with props and get along with the Svelte world. So I came up with callback approach here, so in svelte it would be 

```svelte
let codyClient: CodyWebClient = writable()

function render(repository: CodySidebar_ResolvedRevision, filePath: string, lineOrPosition?: LineOrPositionOrRange) {
        if (!root) {
            root = createRoot(container)
        }

        const chat = createElement(CodyWebChat)
        const hasFileRangeSelection = lineOrPosition?.line

        const provider = createElement(
            CodyWebChatProvider,
            {
                onClientCreated: (client) => {
                    codyClient.set(client)
                },
            },
            [chat]
        )
        root.render(provider)
    }
```

PR #5095, opened 2024-08-02
Cody Web: Remove implicit context in cody web

vovakulikov:
Closes https://linear.app/sourcegraph/issue/SRCH-820/make-cody-web-context-explicit

Prior to this PR, Cody Web used the initial context repository as a default source of context files for prompting. This raised some concerns about implicit context fetching, so this PR remotes this feature, and now chat relies only on user-provided file/repo/urls mentions in the message textbox. 

## Test plan
- Run Cody Web demo (`cd web` and `pnpm dev`)
- Check that when you don't mention any files repositories or URLs you don't have any context in chat UI as you submit the message 
- Check that mention context still works (try to mention some files or repos and check that context is pulled properly)


web/demo/App.tsx
```diff
@@ -40,7 +40,7 @@ const MOCK_DOT_COM_SOURCEGRAPH_REPOSITORY: Repository[] =
 const MOCK_INITIAL_DOT_COM_CONTEXT =
     serverEndpoint === DOTCOM_SERVER_ENDPOINT
         ? {
-              fileURL: '/internal/uploadstore/config.go',
+              fileURL: 'internal/codeintel/ranking/internal/background/mapper/config.go',
```
vovakulikov, line 43: Just updating demo here, it turned out that `/internal/uploadstore/config.go` doesn't exist anymore

camdencheek (approved):
Q: do we still handle questions like "what does this file do?"

vovakulikov:
@camdencheek only when you have this file as a mention in the textbox, which we still will have by default when you open the sidepanel cody chat, follow up questions which don't have mentions be default though won't have any context fetched *unless user enter mention again. 

This is how it works in vscode at the moment.

PR #5098, opened 2024-08-03
pass AbortSignals in more places

sqs:
AbortSignals let us cancel potentially resource-intensive tasks when their result is no longer needed. Now we pass them around in more places.

Many methods in the GraphQL client now accept AbortSignals. This preserves the timeouts and timeout-related error behavior.

## Test plan

Test with an added 6-second-plus sleep and ensure that the error thrown is an ETIMEDOUT.


arafatkatze (approved):

thenamankumar (approved):
LGTM

PR #5100, opened 2024-08-03
refactor context-fetching

sqs:
Now, the context-fetching code looks at the mentions to determine which repo(s) to search in. For example, if there is an @-mention of a repository, then that repository is searched. These special @-mentions are called "corpus" mentions because they need to be expanded specially (unlike file @-mentions that refer to just a single document).

This is better than the previous way, which:

1. still used the "enhanced context" concept (a coarse boolean) instead of reading the context @-mentions (such as `@repository`) to determine the behavior
2. was built in a way that made support for multiple workspaces hard

There is no (intended or known) behavior change. Note that none of the agent snapshot tests needed changing.

## Test plan

Test context-fetching on PLG and enterprise. Ensure that PLG uses symf and embeddings, and enterprise uses remote context search.

### DIFF
--- a/vscode/src/chat/clientStateBroadcaster.ts
+++ b/vscode/src/chat/clientStateBroadcaster.ts
@@ -34,54 +34,8 @@ export function startClientStateBroadcaster({
     async function rawSendClientState(signal: AbortSignal | null): Promise<void> {
         const items: ContextItem[] = []
 
-        // TODO(sqs): Make this consistent between self-serve (no remote search) and enterprise (has
-        // remote search).
-        if (remoteSearch) {
-            // TODO(sqs): Track the last-used repositories. Right now it just uses the current
-            // repository.
-            //
-            // Make a repository item that is the same as what the @-repository OpenCtx provider
-            // would return.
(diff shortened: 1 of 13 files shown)


vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -146,6 +100,59 @@ export function startClientStateBroadcaster({
     return vscode.Disposable.from(...disposables)
 }
 
+export function getCorpusContextItemsForEditorState({
+    remoteSearch,
+}: { remoteSearch: RemoteSearch | null }): ContextItem[] {
+    const items: ContextItem[] = []
+
+    // TODO(sqs): Make this consistent between self-serve (no remote search) and enterprise (has
```
thenamankumar, line 108: Can you please mention what exactly to do to make it consistent?

thenamankumar (approved):
This PR also seems to solve https://linear.app/sourcegraph/issue/SRCH-821/context-is-not-being-fetched-on-vs-code-even-though-its-being-fetched

I'll verify by running it locally. Overall looks good.

PR #5101, opened 2024-08-03
refactor how remote repo/file providers represent context

sqs:
Now, they just embed a ContextItem in their `mention.data.contextItem` field. This eliminates ad-hoc intermediate ways of representing a remote repo or file, of which there were a few in the codebase.

No behavior change.

## Test plan

Perform remote @-repo and @-file mentions in VS Code and Cody Web. Ensure they work. In particular, ensure the right context is fetched and that remote @-file mentions have the 2-step input (where first you pick the repo and then the file).

### DIFF
--- a/lib/shared/src/context/openctx/internalProvider.ts
+++ b/lib/shared/src/context/openctx/internalProvider.ts
@@ -1,0 +1,66 @@
+import type {
+    Mention,
+    MentionsParams,
+    MentionsResult,
+    Provider,
+    ProviderSettings,
+} from '@openctx/client'
+import { isContextItemType } from '../../codebase-context/messages'
+import type { SerializedContextItem } from '../../lexicalEditor/nodes'
+
+/**
+ * An OpenCtx provider implemented internally.
+ */
+export interface InternalOpenCtxProvider extends Provider {
+    providerUri: string
+
+    /**
+     * Internal mention providers can return {@link MentionWithContextItemData} to make it easier to
+     * convert to our {@link ContextItem} type.
+     */
+    mentions?(
+        params: MentionsParams,
+        settings: ProviderSettings
+    ):
+        | MentionsResult
+        | Promise<MentionsResult>
+        | MentionWithContextItemData[]
+        | Promise<MentionWithContextItemData[]>
+}
+
+/**
+ * An OpenCtx mention whose {@link Mention.data} contains a {@link ContextItem}.
+ */
+export interface MentionWithContextItemData extends Mention {
+    data: {
+        contextItem: SerializedContextItem
+    }
+}
+
+export function isMentionWithContextItemData(mention: Mention): mention is MentionWithContextItemData {
+    if (!mention.data) {
+        return false
+    }
+    if (
+        !(
+            'contextItem' in mention.data &&
+            mention.data.contextItem &&
+            typeof mention.data.contextItem === 'object'
+        )
+    ) {
+        return false
+    }
+    if (!('uri' in mention.data.contextItem && typeof mention.data.contextItem.uri === 'string')) {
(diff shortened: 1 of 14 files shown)


lib/shared/src/context/openctx/internalProvider.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export function isMentionWithContextItemData(mention: Mention): mention is MentionWithContextItemData {
+    return Boolean(
+        // biome-ignore lint/complexity/useOptionalChain:
+        mention.data !== undefined &&
+            mention.data.contextItem &&
+            typeof mention.data.contextItem === 'object' &&
+            typeof (mention.data.contextItem as any).uri === 'string'
+    )
```
dominiccooney, line 46: Let's ensure the type assertion is accurate, now and in the future, by using TypeScript's checker. (I'm also surprised that `uri` is all you need to be a ContextItem, even today...

The way to do this would be to show TypeScript's type checker is satisfied like this:

```
function isFoo(x: Bar) x is Foo {
  if (... the elaborated tests of x ...) {
    // lint to ignore the unused variable -- we are asserting the type
    let y: Foo = x
    return true
  }
  return false
}
```

Reason to do this: Even if the type assertion is correct *now*, it won't remain that way as ContextItem evolves. Then we've lost type safety and we'll have errors at runtime which we could have easily caught at compile time.

Separately, Biome's complaint seems ok here, couldn't we do `typeof mention?.data.contextItem.uri === 'string'` and be done with it? (...or two lines if we really need to show TypeScript that data is an object?)

dominiccooney (changes requested):
Let's not pick brevity in one method over type safety in the context subsystem.

lib/shared/src/context/openctx/internalProvider.ts
sqs: @dominiccooney Good point. I started doing this, and it's going to involve a total type-guard validator of a ContextItem. I can follow through and do that, but is that what you had in mind? If so, should we use some library that lets you assert values as TypeScript types at runtime (like zod)?

See the `// But this then` line for where I paused.

```
export function isMentionWithContextItemData(mention: Mention): mention is MentionWithContextItemData {
    if (!mention.data) {
        return false
    }
    const data = mention.data
    if (!('contextItem' in mention.data && mention.data.contextItem)) {
        return false
    }
    const contextItem = data.contextItem
    if (
        !(
            contextItem &&
            typeof contextItem === 'object' &&
            'type' in contextItem &&
            'uri' in contextItem
        )
    ) {
        return false
    }
    const { type, uri } = contextItem
    if (!(typeof type === 'string' && typeof uri === 'string')) {
        return false
    }
    if (!isContextItemType(type)) {
        return false
    }
    // But this then complaints that `{ type, uri }` is not a full ContextItem...
    ;({ ...mention, data: { contextItem: { type, uri } } }) satisfies MentionWithContextItemData
    return true
}
```

dominiccooney: Needless to say, we also need to check that if the optional fields are present, that they have the right type. But this is similar... dereference a `?.` and check that `typeof foo in ['undefined', 'string']` for `repoName?: string` etc.

sqs: @dominiccooney Ah, I thought you wanted a `x satisfies T` or `const y: T = x` statement in there (which is a good idea!), rather than just a `return true` or `return false`, which does not perform a comprehensive check.

jtibshirani:
Thanks for the heads up on this. It makes sense to me, defer to Dominic on the detailed review though!

beyang (approved):

PR #5102, opened 2024-08-04
use CodyPanel to show tabs & history panel in Cody Web

sqs:
Previously, Cody Web only used the `<Chat />` component. Now, it uses the (newly extracted) `<CodyPanel />` component, which shows the tabs at the top. Right now, only the chat and history tabs are shown in Cody Web.

Related: https://linear.app/sourcegraph/issue/SRCH-836/integrate-new-tabs-ui-to-the-cody-web

No behavior change to Cody for VS Code. This is 95% just moving around code.

This is mainly to bring the Cody Web demo in `web/demo` up to date so it can be used for easy local dev on the unified Cody panel. If we need to update Cody that's used in the web app, then we can disable the tabs altogether to avoid a major design change until we get designs for it after the upcoming release (see https://linear.app/sourcegraph/issue/SRCH-810/design-input-on-new-cody-web-tabs-ui).

image

## Test plan

CI for existing non-Cody Web functionality. For Cody Web, open up the dev server and see that both the chat and history tabs work.


burmudar (approved):

dominiccooney (approved):
Super.

PR #5105, opened 2024-08-05
show commands & Prompt Library together

sqs:
Related project: https://linear.app/sourcegraph/project/transition-commands-to-prompt-library-71cbc0dfef6b/overview

Because we need to support customers on self-hosted Sourcegraph releases that have not yet upgraded to the upcoming release with the Prompt Library, we need to show prompts and commands (builtin and custom) together for at least another month.

[**Loom video walkthrough**](https://www.loom.com/share/a243671a617846708588210662c22529)

image

Related TODOs (not blocking merging this):

- Test prompt dropdown in Cody Web in sourcegraph/sourcegraph
- Ensure Cody Web in sourcegraph/sourcegraph is updated so that the prompts dropdown shows regardless of feature flag value

## Test plan

- Test against demo.sourcegraph.com, which currently is on Sourcegraph 5.5, which lacks the Prompt Library. It is important that the editor extension work with older versions and just show builtin commands and custom commands, and not show an annoying error message about missing the Prompt Library.
- Test against sourcegraph.com or s2, where the Prompt Library is available.
- Commands (builtin and custom) should work the same as before, with slightly different UI in the chat panel.
- Prompts should be selectable from the chat message field **Prompts** dropdown. Selecting a prompt inserts its text into the chat message field.
- Prompts should also be usable from the **Prompts & Commands** collapsible section below the chat message field and the chat tab.

## Changelog

- Adds support for the Prompt Library, which lets you share and reuse prompts for Cody chat. You can use prompts from the Prompt Library in Cody chat by using the **Prompts** dropdown menu in the chat message field or by selecting a prompt in the **Prompts & Commands** section. To create, edit, and manage prompts in the Prompt Library, press the **Manage** button from Cody chat or visit **Tools > Prompt Library** on Sourcegraph in your web browser.


jamesmcnamara:
Pre-tagging myself on this because I'm looking into hiding commands for Eclipse where certain things aren't supported and this touches a lot of the same code.

olafurpg:
Following the test plan:

* ✅ Works as expected on demo.sourcegraph.com and S2
* ✅ Commands (built-in and custom) work as before.
	* Non-blocking: I was surprised that built-in commands don't appear in the "Prompts" dropdown. The built-in commands are only visible in a clean chat. It might be nice to make built-in commands available from the dropdown with some UI hinting that it will start a new chat instead of appending to the input.
* ✅ Dropdown and Collapsible works as expected

Subjective opinions about unification of prompts and commands:

* 2c, this feels like a half-baked unification because "Commands" have different behavior from the collapsible and dropdown. Maybe it's cleaner to replace "Prompts & Commands" with a "Commands" collapsible that only lists commands, and make custom commands additionally available via the "Prompts" dropdown under a group that's named something like "Prompts via Commands" (exact name TBD) to emphasize it's a command converted into a prompt.

Non-blocking notes on Prompts as a feature after playing around with this, unrelated to this PR. I can share this separately in #feedback-product if you think it's useful:

* I was surprised that clicking on a prompt appends to the existing input. If there is existing text, it might be nice to ask the user whether they want to append or replace the chat input. 
* I expected the prompts to load faster/instantly after loading the library for the first time. Currently, the "Loading" indicator appears almost every time (for ~200-500ms). It might be OK to have an explicit "refresh" button to sync with remote prompts if you want to pull down a newly created prompt.
* The library took surprisingly long to load the first time, I clicked somewhere else to hide the component because I thought something was wrong. It loaded after maybe ~3-5 seconds. It might be worth pre-loading before the user clicks on "prompts".
* On S2, I created a prompt under my account and it appeared below `erik / Test-Prompt`. I expected my own prompts to appear at the top. After selecting my own prompt and submitting the message, I expected my own prompt to appear at the bottom because I've used it recently.

umpox (approved):
Awesome! I did a brief review of the code and didn't notice any issues, have some non-blocking points/questions:

1. Should prompts replace the entire existing text in the field? I noticed this first because it didn't remove the chip I already had in there, but also it seems like we should probably remove anything else the user added.

https://github.com/user-attachments/assets/687f5faf-4eb1-40d4-b554-43b45046ecae

2. It would be awesome if we could preload and cache the prompts. It's kinda annoying to see "Loading" every time I close/reopen the prompt dropdown or panel

3. Should we show `builtIn` commands at the top? We know this list will always be small, whereas the prompt library is likely to get quite big. It seems like very quickly we'll have the built in commands out of view, and they might be hard to find for users. If they're at the top they'll always be visible

e.g.

image


4. Do we have any plans to paginate the prompts? I didn't see any pagination logic but thinking ahead when there's lots of public prompts, we may want a "Show more" button and to limit to the top X results? LMK if I missed anything

5. The very first time I ran this I briefly ran into an issue where I only saw my local Ollama models, and none of the edit/document/test commands worked. The issue immediately went away when I reloaded my VS Code, and I haven't been able to replicate/see any obvious reason why this would happen. Just mentioning it in case you're aware of this or have any idea what could have caused this

6. Is it intentional to show "Draft" prompts to other users? I noticed this when connected to S2. This seems to conflict with the label when creating this prompt: "marking as draft means other people shouldn't use it yet"

image

7. We should consider hiding the "Prompts" dropdown in the chat input if there are no prompts/custom-commands. Seems like this might be a common occurrence for enterprise instances that don't support the prompt library yet. Also there's a minor double-border issue if a search doesn't return any results :) 

image

olafurpg:
> The very first time I ran this I briefly ran into an issue where I only saw my local Ollama models, and none of the edit/document/test commands worked. The issue immediately went away when I reloaded my VS Code, and I haven't been able to replicate/see any obvious reason why this would happen. Just mentioning it in case you're aware of this or have any idea what could have caused this

I ran the "Restart Extension Host" command every time I switched accounts. I've reported issues in the past about UI chat bugs when switching between accounts (which are fixed with "Restart Extension Host"). I've considered these low-priority issues because most users presumably don't switch regularly between accounts.

sqs:
Thanks for the feedback and reviews! Will merge this and address the top feedback items in the next couple hours.

> > The very first time I ran this I briefly ran into an issue where I only saw my local Ollama models, and none of the edit/document/test commands worked. The issue immediately went away when I reloaded my VS Code, and I haven't been able to replicate/see any obvious reason why this would happen. Just mentioning it in case you're aware of this or have any idea what could have caused this
> 
> I ran the "Restart Extension Host" command every time I switched accounts. I've reported issues in the past about UI chat bugs when switching between accounts (which are fixed with "Restart Extension Host"). I've considered these low-priority issues because most users presumably don't switch regularly between accounts.

It is now much easier to fix this with https://github.com/sourcegraph/cody/pull/5083, because the webview can now "observe" values in the extension host. Only a few things have been ported over to use this yet, though.

PR #5107, opened 2024-08-05
Fix unimplemented range contains check

pkukielka:
## Test plan

Check if JetBrains Edit Code works again with this change.

### DIFF
--- a/vscode/src/testutils/mocks.ts
+++ b/vscode/src/testutils/mocks.ts
@@ -424,12 +424,21 @@ export class Range implements VSCodeRange {
         return this.start.line === this.end.line
     }
     public contains(positionOrRange: Position | Range): boolean {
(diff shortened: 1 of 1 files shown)


vscode/src/testutils/mocks.ts
```diff
@@ -424,12 +424,21 @@ export class Range implements VSCodeRange {
         return this.start.line === this.end.line
     }
     public contains(positionOrRange: Position | Range): boolean {
+        const isSmallerOrEqual = (a: Position, b: Position): boolean => {
+            return a.line < b.line || (a.line === b.line && a.character <= b.character)
+        }
+
         if ('line' in positionOrRange) {
             return (
-                positionOrRange.line >= this.start.line &&
```
pkukielka, line 429: That code was actually incorrect :)
If line is not equal, character does not actually matter.

RXminuS (approved):

PR #5110, opened 2024-08-05
fixed/dropdown: enterprise users can use any model available

jamesmcnamara:
[CODY-3138](https://linear.app/sourcegraph/issue/CODY-3138)

Previously we did not allow enterprise users to use cody pro models that were enabled on their instance. Here we refine that test to check only if it's a free user.

## Test plan
Manual test

## Changelog
Fix bug where enterprise users were stopped from using enabled, server-sent models.


chrsmith (approved):
This logic LGTM. Is there an easy way we can have add a unit test for this? Or just update some existing Storybook page or something?

PR #5113, opened 2024-08-06
Remove noodle flag set to true for Cody Web

thenamankumar:
context: https://linear.app/sourcegraph/issue/SRCH-821/context-is-not-being-fetched-on-vs-code-even-though-its-being-fetched
https://sourcegraph.slack.com/archives/C05D629GWJK/p1722933166623269?thread_ts=1722620916.715349&cid=C05D629GWJK

The `experimental.noodle` flag is enabled by default for Cody Web which should not be the case. Some experimental feature like chat query rewrite are therefore enabled right now.

This PR sets the `experimental.noodle` flag to false.

## Test plan

- cd web
- pnpm dev
- Submit a chat message in chat and monitor the network request to `getCodyChat` to make sure the query string is the chat message without the mention chips text.

### DIFF
--- a/web/CHANGELOG.md
+++ b/web/CHANGELOG.md
@@ -1,5 +1,6 @@
-## Next
+## 0.3.7
 - Adds support for the full panel (not just chat) to the CodyWebPanel component (renamed from CodyWebChat).
+- Sets experimental.noodle flag to false (default).
 
 ## 0.3.6
 - Adds fix that chat UI doesn't include context repository as background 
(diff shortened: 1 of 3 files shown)


bahrmichael (approved):

fkling (approved):
Vova mentioned this in the Slack thread:

> Reason: at the beginning noodle project had some features like remote file mentions I think that we wanted to have in Cody Web too, that’s why it’s true.

Is that no longer relevant?

thenamankumar:
> Vova mentioned this in the Slack thread:
> 
> > Reason: at the beginning noodle project had some features like remote file mentions I think that we wanted to have in Cody Web too, that’s why it’s true.
> 
> Is that no longer relevant?

As far as I can see, mentioning files work right now with the flag off.

PR #5117, opened 2024-08-06
Disable embeddings A/B test, enable auto indexing

janhartman:
Following the conclusion of the A/B test, we'll leave symf & embeddings on for all users and turn on auto indexing for embeddings.

## Test plan
Tested locally.

## Changelog

- Enables automatic indexing for embeddings.


chenkc805 (approved):
LGTM. As part of this, could you also delete the feature flag from the site-admin tool too to decrease cruft there? Thanks

abeatrix (approved):

PR #5122, opened 2024-08-07
persist open/closed state of CollapsiblePanel

sqs:
Now these friends stay open or closed, the last way you left them, across chat sessions and editor restarts:

image

## Test plan

Collapse and reload editor. Confirm state.


abeatrix (approved):
Noticed the position of the buttons has changed, not sure if that's intentional.

sqs:
It was intentional

PR #5124, opened 2024-08-07
show tab labels on wide screens

sqs:
If the screen is 768px wide or wider, then the Cody chat panel tabs will show text labels. This means that the standalone Cody Web chat screen will show text labels for most users, which fixes https://linear.app/sourcegraph/issue/CODY-3171/history-tab-is-hard-to-find-on-cody-web.

https://github.com/user-attachments/assets/7790d9ff-af8e-4e6f-929a-ea3210be1f2d

## Test plan

Make screen (or containing panel) narrow and wide. When it's wide, confirm tooltips are NOT shown but tab titles are visible. When narrow, confirm tab titles are NOT shown but tooltips are.

### DIFF
--- a/vscode/webviews/tabs/TabsBar.tsx
+++ b/vscode/webviews/tabs/TabsBar.tsx
@@ -41,6 +41,7 @@ interface TabConfig {
         tooltipExtra?: React.ReactNode
 
         title: string
+        alwaysShowTitle?: boolean
         Icon: IconComponent
         command: string
         arg?: string | undefined | null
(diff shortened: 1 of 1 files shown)


RXminuS:
Maybe worth to have a intermediate stage (ideally the default sidebar width) where only critical buttons show text and go icons only for anything smaller? I'm thinking in particular "New Chat"?

olafurpg (approved):
LGTM 👍🏻 

+1 to Rik's comment, there is a very common width in the IDE where we currently show no text labels and it would be helpful to have text for the most common actions (probably "New Chat" and "Open in Editor")

RXminuS (approved):

sqs:
Great idea. Done. `New Chat` text label is always visible.

image

PR #5127, opened 2024-08-07
Agent: move enterprise integration tests to separate files

olafurpg:
Previously, the enterprise tests lived in the monolithic `index.test.ts` file. This made it unnecessarily difficult to test the enterprise clients in isolation from the dotcom client. For example, in the PR https://github.com/sourcegraph/cody/pull/4861 we want to be able to reliably test that enterprise clients don't send network requests outside of sourcegraph. If the dotcom client downloads symf then the enterprise client won't send network requests to download it because it's cached.

Now, the enterprise clients live in separate test files making it easier to only run those tests in isolation giving us better confidence about the asserted behavior.

## Test plan

Green CI. Notice that there is no diff in the HTTP recordings.


umpox (approved):
👍

PR #5129, opened 2024-08-07
misc. Cody Web usability improvements

sqs:
- Always start a new chat in Cody Web. Now that the history tab is easily visible, it's easy for users to open other chats in the Cody chat sidebar alongside a file in Sourcegraph.
- Open the Prompts & Commands and Chat Help info boxes initially (in both Cody Web and in editors). If the user closes them, that state will persist. We had made them hidden by default to make them less noisy, but now I think persisting the closed state is sufficient to make it less noisy.

## Test plan

In Cody Web (sourcegraph/sourcegraph pnpm link'd), test that each time you load Cody standalone chat (`/cody/chat`), it creates a new chat.


RXminuS:
Maybe in the URL we should keep the MessageID so that if I want to link/bookmark to a specific chat I can do so easily. E.g. `sourcegraph.com/cody/chat?id=apoij2091ajwoijw` so starting a new chat will then just update the url with the id once it's created.

Unrelated and low on my todo list was to use something similar like that to enable `sourcegraph.com/cody/chat?q='start of question'` so that we can make these special Chrome Bookmarks similar to Gemini.
CleanShot 2024-08-07 at 14 07 08@2x

sqs:
@RXminuS Yeah, that would be nice (to store it in the URL).

sqs:
I really want to start persisting these chats in the DB and making them shareable across users. So whatever we do for the URL (and we'll do something), we'll make it the same URL scheme or something like that so that they can smoothly become shareable cross-device.

PR #5133, opened 2024-08-07
added/agent: persistent global state for agent

jamesmcnamara:
Adds a var to ClientCapabilities allowing extensions to specify a file path where the agent can persist global state, such as chat. Currently does the simplest possible thing which is to write the entire state out to disk on every update. This can obviously be improved, but at the cost of complexity.

## Test plan
Added unit tests
Manually tested this with the Eclipse plugin in https://github.com/sourcegraph/eclipse/pull/39

## Changelog
Added ability to persist state from the agent


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -522,6 +524,9 @@ export interface ClientCapabilities {
     // convenient for clients that forward the string directly to an underlying
     // webview container.
     webviewMessages?: 'object-encoded' | 'string-encoded' | undefined | null
+    // when a file path is provided, the agent will attempt to deserialize extension state
+    // from the path and append all global state (such as chat interactions) to the path
+    persistencePath?: string | undefined | null
```
olafurpg, line 529: Capabilities are conventionally enums. I propose we do this:

```suggestion
    // How to deal with vscode.ExtensionContext.globalState.
    // - Stateless: the state does not persist between agent processes. This means the client is responsible for features like managing chat history.
    // - Server managed: the server reads and writes the state without informing the client. The client can optionally customize the file path of the JSON config via `ClientInfo.globalStatePath: string`
    // - Client managed: not implemented yet. When implemented, clients will be able to implement a JSON-RPC request to handle the saving of the client state. This is needed to safely share state between concurrent agent processes (assuming there is one IDE client process managing multiple agent processes).
    globalState?: "stateless" | "server-managed" | "client-managed" | undefined null
```

vscode/src/services/AuthProvider.ts
```diff
@@ -398,6 +400,7 @@ export class AuthProvider implements AuthStatusProvider, vscode.Disposable {
             graphqlClient.setConfig(await getFullConfig())
             await ClientConfigSingleton.getInstance().setAuthStatus(authStatus)
             await syncModels(authStatus)
+            vscode.commands.executeCommand('auth.changed', authStatus)
```
olafurpg, line 403: Is this intentional?

agent/src/AgentGlobalState.ts
```diff
@@ -21,8 +23,47 @@ export class AgentGlobalState implements vscode.Memento {
         this.globalStorage.set('completion.inline.hasAcceptedFirstCompletion', true)
         this.globalStorage.set('extension.hasActivatedPreviously', 'true')
     }
+
+    public setPersistencePath(path: string): void {
+        this.path = path
+        const storage = this.readFromDisk(path)
+        for (const [key, value] of storage ?? []) {
```
olafurpg, line 30: we can avoid this validation if we read from disk during construction when initializing `this.globalStorage`

agent/src/AgentGlobalState.ts
```diff
@@ -21,8 +23,47 @@ export class AgentGlobalState implements vscode.Memento {
...
+    // in a format that can be deserialized later back into a Map.
+    private syncToDisk(): void {
+        if (this.path) {
+            const json = JSON.stringify([...this.globalStorage])
+            fs.writeFileSync(this.path, json)
+        }
+    }
+
+    // deserialize the contents of the file at `this.path` into the globalStorage
+    private readFromDisk(path: string): Map<string, any> | undefined {
```
olafurpg, line 52: Let's `logError(...)` when this fails. The user may have manually edited the JSON file and we don't want to silently ignore it.

agent/src/AgentGlobalState.ts
```diff
@@ -21,8 +23,47 @@ export class AgentGlobalState implements vscode.Memento {
...
+            return new Map(entries)
+        } catch (e) {
+            return undefined
+        }
+    }
+
     public reset(): void {
         this.globalStorage.clear()
+        if (this.path) {
+            fs.truncateSync(this.path)
```
olafurpg, line 65: Hmmm, this is a bit dangerous. We only call `reset()` in tests. I think it's better to assert that `reset()` can only be used when `globalStorage: 'stateless' | 'client-managed'`. We should not be clearing this file since it can wipe the user's chat history. Our integration tests should not use server-managed global state.

olafurpg (changes requested):

vscode/src/jsonrpc/agent-protocol.ts
olafurpg: btw, we can use `path.join(codyPaths(), "globalState.json")` as a default location for server-managed state. However, I think we should ideally try to have clients move to client-managed state ASAP to avoid races between concurrent agents.

dominiccooney: +1, codyPaths is a foot gun because of all clients share a path and it is going to end badly because we have no concurrency control and have given little thought to versioning.

We need product input, or at least being informed, here on what different clients should do--whether state is shared; if so, which state.

Even if the state is not shared between IDEs, we still have concurrency from multiple instances of the same IDE. I think we should copy vscode here, which is using sqlite3 for some consistency. (We don't need sqlite, but we need something with the same row-level consistency model.)

olafurpg: After thinking more about this, we maybe just want to bite the bullet and default to `server-managed` for all IDEs using the same sqlite-powered implementation sharing the same `codyPaths()` file path (with option to override it). This is assuming we get some file locking mechanism from sqlite to handle concurrent clients writing to the disk. None of these writes/reads should take a long time to run.

agent/src/AgentGlobalState.ts
olafurpg: Let's remove this. VS Code doesn't expose APIs to reset the `globalState`, we only use this for testing purposes.

agent/src/AgentGlobalState.ts
```diff
@@ -11,7 +13,10 @@ import * as vscode_shim from './vscode-shim'
 export class AgentGlobalState implements vscode.Memento {
     private globalStorage = new Map<string, any>()
 
-    constructor() {
+    constructor(private path?: string) {
```
olafurpg, line 16: Let's make this a directory to 1) ensure we have a consistent filename and 2) making it possible to write multiple files (for example, sqlite lockfiles)

```suggestion
    constructor(private dir?: string) {
```

agent/src/agent.ts
```diff
@@ -1320,6 +1325,15 @@ export class Agent extends MessageHandler implements ExtensionClient {
         }
     }
 
+    private newGlobalState(clientInfo: ClientInfo): AgentGlobalState {
+        if (clientInfo.capabilities?.globalState === 'server-managed') {
+            return new AgentGlobalState(
+                clientInfo.globalStatePath ?? join(codyPaths().data, 'globalState.json')
+            )
+        }
+        return new AgentGlobalState()
```
olafurpg, line 1334: Let's throw an error if it's `client-managed`, it's OK to keep it unimplemented

vscode/src/services/AuthProvider.ts
olafurpg: Let's remove these unrelated diffs

vscode/webviews/CodyPanel.tsx
```diff
@@ -59,7 +59,7 @@ export const CodyPanel: FunctionComponent<
             className={styles.outerContainer}
         >
             {/* Hide tab bar in editor chat panels. */}
-            {config.agentIDE === CodyIDE.Web || config.webviewType !== 'editor' ? (
+            {config.agentIDE !== CodyIDE.VSCode || config.webviewType !== 'editor' ? (
```
olafurpg, line 62: Unrelated change?

vscode/webviews/tabs/AccountTab.tsx
```diff
@@ -40,7 +40,7 @@ export const AccountTab: React.FC<AccountTabProps> = ({ userInfo }) => {
         },
         {
             text: 'Sign Out',
-            onClick: () => getVSCodeAPI().postMessage({ command: 'command', id: 'cody.auth.signout' }),
```
olafurpg, line 43: Unrelated?

olafurpg (changes requested):
We're trending in the right direction! I propose we limit the scope of this PR to `server-managed` global state. The open question is we go the extra mile and use sqlite from the start. I'm leaning towards "yes", let's start with a solid implementation that doesn't suffer from races between concurrent clients.

vscode/webviews/CodyPanel.tsx
jamesmcnamara: This is necessary to make the tabs bar appear in any other IDE (where the history is located so I think it's relevant).

vscode/src/jsonrpc/agent-protocol.ts
jamesmcnamara: I spent quite a long time trying to get better-sqlite3 or node-sqlite3 to work on windows but there are a lot of known issues around building and bundling it (specifically on windows) because it's using bindings to an underlying C library. I think it would be possible to get it working but we'd likely have to bundle the package separately and patch it in.

I backed out all of those changes and swapped to just using [node-localstorage](https://www.npmjs.com/package/node-localstorage) which is marginally better than just writing to files, but I don't think we're going to have a tremendous volume of writes.

However, I'm just now realizing the issue of concurrency. I doubt node-localstorage handles concurrency well. I can test that and then potentially get back to sqlite.

jamesmcnamara: I created an [RFC](https://docs.google.com/document/d/1Sy7SBnuM5p3LDYyT0SNQ7_0HKTYeIEv5QjRzeXjZ2Fc/edit?usp=sharing) for this

vscode/webviews/CodyPanel.tsx
abeatrix: are we good to remove this after the native switch?

jamesmcnamara: Yep!

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -605,6 +606,15 @@ export interface ClientCapabilities {
     // convenient for clients that forward the string directly to an underlying
     // webview container.
     webviewMessages?: 'object-encoded' | 'string-encoded' | undefined | null
+    // How to deal with vscode.ExtensionContext.globalState.
+    // - Stateless: the state does not persist between agent processes. This means the client is
+    // responsible for features like managing chat history.
+    // - Server managed: the server reads and writes the state without informing the client.
+    // The client can optionally customize the file path of the JSON config via `ClientInfo.globalStatePath: string`
```
abeatrix, line 613:
```suggestion
    // The client can optionally customize the file path of the JSON config via `ClientInfo.globalStateDir: string`
```

abeatrix (commented):
UPDATE: this was a Beatrix's issue and not an issue with the PR. The root cause was because we are using Visual Studio enum to pass in ServerManage instead of the string "server-managed" that agent is looking for 🤡 

i just tried this build on Visual Studio and it looks like it crashed at start up:

```
TypeError: Cannot read properties of undefined (reading 'get')
    at start (c:\Users\BeatrixW\Dev\cody\vscode\src\main.ts:101:30)
    at activate (c:\Users\BeatrixW\Dev\cody\vscode\src\extension.common.ts:62:34)
    at activate2 (c:\Users\BeatrixW\Dev\cody\vscode\src\extension.node.ts:61:12)
    at initializeVscodeExtension (c:\Users\BeatrixW\Dev\cody\agent\src\agent.ts:177:11)
    at <anonymous> (c:\Users\BeatrixW\Dev\cody\agent\src\agent.ts:426:23)
    at <anonymous> (c:\Users\BeatrixW\Dev\cody\vscode\src\jsonrpc\jsonrpc.ts:95:27)
    at handleRequest (c:\Users\BeatrixW\Dev\cody\node_modules\.pnpm\vscode-jsonrpc@8.2.0\node_modules\vscode-jsonrpc\lib\common\connection.js:486:41)
    at handleMessage (c:\Users\BeatrixW\Dev\cody\node_modules\.pnpm\vscode-jsonrpc@8.2.0\node_modules\vscode-jsonrpc\lib\common\connection.js:339:13)
    at processMessageQueue (c:\Users\BeatrixW\Dev\cody\node_modules\.pnpm\vscode-jsonrpc@8.2.0\node_modules\vscode-jsonrpc\lib\common\connection.js:362:17)
    at Immediate.<anonymous> (c:\Users\BeatrixW\Dev\cody\node_modules\.pnpm\vscode-jsonrpc@8.2.0\node_modules\vscode-jsonrpc\lib\common\connection.js:334:13)
    at process.processImmediate (c:\Users\BeatrixW\Dev\cody\lib\internal\timers.js:478:21)
    at process.topLevelDomainCallback (c:\Users\BeatrixW\Dev\cody\lib\domain.js:160:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24) {stack: 'TypeError: Cannot read properties of undefine…Trampoline (node:internal/async_hooks:128:24)', message: "Cannot read properties of undefined (reading 'get')"}
```

abeatrix (approved):
just tested the latest commit and confirmed it's working as expected for Visual Studio when the configuration is set correctly 🤡 

image

PR #5135, opened 2024-08-07
symf: regenerate index more frequently

beyang:
The local search index is now rebuilt more frequently when many files are changed since the last index (such as when the user checks out a revision that differs from the current revision).

The frequency of update is dependent on:
- # files changed since last index
- time since last index
- time it took for last index

## Test plan

Test locally. See unit test.

### DIFF
--- a/vscode/src/local-context/symf.ts
+++ b/vscode/src/local-context/symf.ts
@@ -216,8 +216,7 @@ export class SymfRunner implements IndexedKeywordContextFetcher, vscode.Disposab
     }
 
     /**
-     * Check index freshness and reindex if needed. Currently reindexes daily if changes
-     * have been detected.
+     * Check index freshness and reindex if needed.
      */
     public async reindexIfStale(scopeDir: FileURI): Promise<void> {
         logDebug('SymfRunner', 'reindexIfStale', scopeDir.fsPath)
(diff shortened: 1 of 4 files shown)


sqs (approved):

janhartman (approved):

vscode/src/local-context/symf.ts
```diff
@@ -858,3 +860,67 @@ class IndexManager implements vscode.Disposable {
         }
     }
 }
+
+/**
+ * Determines whether the search index should be refreshed based on whether we're past a staleness threshold
+ * that depends on the time since last index, time it took to create the last index, and the number of files changed.
+ */
+export function shouldReindex(diff: CorpusDiff): boolean {
```
thenamankumar, line 868: Should be good to also track here when the re-indexes are triggered along with the number of changed files, last index duration and time since last index. Will help to optimize this in the future.

vscode/src/local-context/symf.ts
```diff
@@ -226,10 +230,8 @@ export class SymfRunner implements IndexedKeywordContextFetcher, vscode.Disposab
                 })
```
thenamankumar, line 230: CleanShot 2024-08-08 at 12 55 25@2x

The comment above needs to be updated. "Currently reindexes daily"

thenamankumar (approved):

vscode/src/local-context/symf.ts
beyang: via telemetry or some other mechanism?

PR #5138, opened 2024-08-08
Smart Apply: Implement stateful UI

umpox:
## Description

This PR adds stateful UI to the smart apply feature:
- Loading state -> "Applying"
- Applied state -> "Accept" | "Reject"

It includes this behaviour:
- Loading: Display an animated spinner and disable the button
- Accept: Accept the edit in the document on click
- Reject: Reject the edit in the document on click
- Error: Open the error message in the editor on click

### UI States

**Default**

image

**Loading (Applying)**

image

**Finished (Applied)**

image

## Test plan

1. Start a smart apply
2. Expect that the "Applying" state shows
3. Expect that clicking "Applying" does nothing
4. Expect that the smart apply applies correctly
5. Expect that "Apply" "Reject" shows
6. Expect that the associated button either accepts or rejects the edit


vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -251,6 +256,11 @@ export class ChatsController implements vscode.Disposable {
         await provider.handleGetUserEditorContext(uri)
     }
 
+    private async sendSmartApplyResultToChat(result: SmartApplyResult): Promise<void> {
+        const provider = await this.getActiveChatController()
+        await provider.handleSmartApplyResult(result)
+    }
```
umpox, line 262: It'd be nice if we could pass `chatID` from `ChatMessageContent.tsx` so we could notify all relevant panels here. There is likely a case where this would not update correctly if the original chat is no longer the active chat.

If anyone has any pointers on if where we can do this, it'd be appreciated. I couldn't see any way to access `chatID` in the current webview code, and I am unsure if `chatID === panelID` (or associated)

vscode/webviews/icons/CodeBlockActionIcons.tsx
```diff
@@ -18,3 +20,11 @@ export const SparkleIcon =
...
+
+export const SyncSpinIcon = `<svg class="${styles.spin}" role="img" width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="currentColor"><path fill-rule="evenodd" clip-rule="evenodd" d="M2.006 8.267L.78 9.5 0 8.73l2.09-2.07.76.01 2.09 2.12-.76.76-1.167-1.18a5 5 0 0 0 9.4 1.983l.813.597a6 6 0 0 1-11.22-2.683zm10.99-.466L11.76 6.55l-.76.76 2.09 2.11.76.01 2.09-2.07-.75-.76-1.194 1.18a6 6 0 0 0-11.11-2.92l.81.594a5 5 0 0 1 9.3 2.346z"/></svg>`
+
+export const TickIcon =
+    '<svg role="img" width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="currentColor"><path fill-rule="evenodd" clip-rule="evenodd" d="M14.431 3.323l-8.47 10-.79-.036-3.35-4.77.818-.574 2.978 4.24 8.051-9.506.764.646z"/></svg>'
+
+export const CloseIcon =
+    '<svg role="img" width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" fill="currentColor"><path fill-rule="evenodd" clip-rule="evenodd" d="M8 8.707l3.646 3.647.708-.707L8.707 8l3.647-3.646-.707-.708L8 7.293 4.354 3.646l-.707.708L7.293 8l-3.646 3.646.707.708L8 8.707z"/></svg>'
```
umpox, line 30: These are icons from https://github.com/microsoft/vscode-codicons

I could make us of `<i class="codicon codicon-..."></i>` usage here, but keeping these as SVGs for now so I can easily iterate on the UI

abeatrix (commented):
havent reviewed the code change closely yet, but I tried following the test plan and it seems to work correctly: 

https://github.com/user-attachments/assets/e050d427-39e7-46b1-9691-4595749d7133

I did notice there is a lag between the "Applying" and the codelenses showing up in the Editor -- is this expected?

When I rejected the change, it would select the wrong section of the code after the change has been reverted

umpox:
Thanks @abeatrix!

> I did notice there is a lag between the "Applying" and the codelenses showing up in the Editor -- is this expected?

Right now yeah, it's because "Applying" starts when we try to find the selection in the file, but the codelens is from the inline edit which only starts once we have that selection. If I have time in a future PR, I might look at adding a codelens in the early stage, although it'd show at the top of the file

> When I rejected the change, it would select the wrong section of the code after the change has been reverted

Interesting, nothing here should affect the selection so this is possibly a bug with inline edit. I'll see if I can replicate on `main`

vscode/webviews/Chat.tsx
```diff
@@ -118,19 +118,39 @@ export const Chat: React.FunctionComponent<React.PropsWithChildren<ChatboxProps>
...
-            })
+        return {
+            onSubmit: (
+                id: string,
+                text: string,
+                instruction?: PromptString,
+                fileName?: string
+            ): void => {
+                vscodeAPI.postMessage({
+                    command: 'smartApplySubmit',
```
sqs, line 134: note: for next time, look into the new way of method-calling between the webview and exthost that I introduced for prompts and @-mention items. It's higher-level.

sqs (approved):

PR #5140, opened 2024-08-08
Context: disable local embeddings by default

olafurpg:
Fixes CODY-3199

Previously, local embeddings were enabled by default. This was problematic for several reasons:

* Local embeddings could go out of sync and hurt context quality because they returned outdated chunks.
* Local embeddings required downloading the `cody-engine` binary from github.com, which was problematic for some enterprise customers.

This PR disables local embeddings by default addressing both problems.

## Test plan

* Run `rm -rf '/Users/olafurpg/Library/Application Support/Code/User/globalStorage/sourcegraph.cody-ai/cody-engine'` (adjust path to your computer)
* Remove `cody.experimental.localEmbeddings.disabled` user setting from your VS Code installation
* Run a local build of this plugin
* After the extension has finished activating, confirm that the path `'/Users/olafurpg/Library/Application Support/Code/User/globalStorage/sourcegraph.cody-ai/cody-engine'` still doesn't exist
* Set user setting `"cody.experimental.localEmbeddings.enabled": true`, reload VS Code, confirm that it downloads the cody-engine binary during activation.

## Changelog

* Local embeddings are now disabled by default. To enable local embeddings, add the user setting `"cody.experimental.localEmbeddings.enabled": true`.

PR #5141, opened 2024-08-08
Reindex embeddings if they are stale

janhartman:
This PR reenables local embeddings for PLG users and adds periodic reindexing.

## Test plan
Tested locally - I opened a repo that I haven't worked on in a while, created a commit and saw that embeddings were reindexed for the repo.

## Changelog
- Embeddings index is periodically rebuilt if it is stale.


vscode/src/local-context/local-embeddings.ts
```diff
@@ -555,6 +557,39 @@ export class LocalEmbeddingsController
                 this.updateIssueStatusBar()
             }
         }
+
+        // Check if the embeddings are stale.
+        const currentCommitHash = vscodeGitAPI?.repositories[0]?.state.HEAD?.commit ?? ''
```
beyang, line 562: Is this the right thing to do? Or should we trigger an update for the index of the currently active repository (presumably the repository tagged in the user query)

vscode/src/local-context/local-embeddings.ts
```diff
@@ -532,7 +534,7 @@ export class LocalEmbeddingsController
     }
 
     // After loading a repo, we asynchronously check whether the repository
-    // still needs embeddings.
+    // still needs embeddings or if the embeddings are stale.
     private async onHealthReport(repoDir: FileURI, health: IndexHealthResultFound): Promise<void> {
```
beyang, line 538: When is this function called? It looks like it's just called when the embeddings controller initializes, which suggests this check is only triggered when the editor restarts. Is that right?

vscode/src/local-context/local-embeddings.ts
```diff
@@ -555,6 +557,39 @@ export class LocalEmbeddingsController
                 this.updateIssueStatusBar()
             }
         }
+
+        // Check if the embeddings are stale.
+        const currentCommitHash = vscodeGitAPI?.repositories[0]?.state.HEAD?.commit ?? ''
+        const lastEmbeddingsCommit = await vscodeGitAPI?.repositories[0]?.getCommit(health.commit)
+        const currentCommit = await vscodeGitAPI?.repositories[0]?.getCommit(currentCommitHash)
```
beyang, line 564: One of the cases that triggered poor results for symf is when the user checks out a revision that's different from the one that's indexed. Many devs do this fairly commonly (e.g., checking out an old revision to backport a bug). If the reindexing check happens when the very old branch is checked out, then we could go a full day before the index is updated, which means context will be very stale until then. For this reason, symf now has [update logic](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/local-context/symf.ts?L877) that triggers more aggressively, provided that the last index did not take too long, and depending on the magnitude of the delta. Can we implement something similar for local embeddings?

beyang (changes requested):

vscode/src/local-context/local-embeddings.ts
PriNova: What about this approach: 
Generate a SHA-256 hash for each file based on its content. Store this hash in a metadata field within the index alongside the files. During checkout, only the files with changes are verified against their hashes. Additionally, `git diff --diff-filter=ACMRT --name-only` can be used with Git for potentially faster operations to discover the deltas. Moreover, store the Symf-generated index within the project repository so that each branch has its unique index file. This could also apply to the embeddings file for each project. If storing the index/embeddings in the repository is not preferred, then adopt a folder structure that uses the project name and branch as subfolders in the currently used folder structure, similar to how Gradle handles cached dependencies.

vscode/src/local-context/local-embeddings.ts
janhartman: It's called after the embeddings are loaded. Good point, I think we need to call it periodically to catch the case where people keep projects open for a long time.

vscode/src/local-context/local-embeddings.ts
janhartman: @PriNova thanks for the suggestion! Unfortunately this a bit too much to justify working on since we plan to deprecate local embeddings anyway.

vscode/src/local-context/local-embeddings.ts
janhartman: I assumed `repositories[0]` would be the currently active repository, which is wrong - sorry, didn't look into the API enough since our shim has no documentation. Currently active repository should make sense.

vscode/src/local-context/local-embeddings.ts
janhartman: @beyang makes sense, will look into it.

janhartman: Fixed.

vscode/src/local-context/local-embeddings.ts
janhartman: Fixed - added a health check call on every querying with at least a minute in between.

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -231,9 +231,6 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
             TestSupport.instance.chatPanelProvider.set(this)
         }
 
-        // Advise local embeddings to start up if necessary.
-        void this.localEmbeddings?.start()
-
```
janhartman, line 236: This was a bug and isn't necessary, we start it earlier.

vscode/src/extension.node.ts
```diff
@@ -39,16 +39,17 @@ export function activate(
...
-    // we want to revert the decision (for example, only do local embeddings
-    // for Cody Pro users until we have Multitenancy).
-    const isLocalEmbeddingsEnabled = vscode.workspace
+    // NOTE: local embeddings are only going to be supported in VSC for now.
+    // Until we revisit this decision, we disable local embeddings for all agent
+    // clients like the JetBrains plugin.
+    let isLocalEmbeddingsDisabled = vscode.workspace
         .getConfiguration()
-        .get<boolean>('cody.experimental.localEmbeddings.enabled', false)
+        .get<boolean>('cody.advanced.agent.running', false)
```
olafurpg, line 47: There needs to be a way to disable embeddings on VSC. It's better to set that setting to false by default in the agent instead of disabling 100% for agent clients. I recall there was no strong technical reason for disabling embeddings in agent clients, BFG was just too buggy that it was slowing down progress on JetBrains. There is a class that sets default setting for the agent where we can disable embeddings.

vscode/src/extension.node.ts
```diff
@@ -59,10 +60,10 @@ export function activate(
         .get<boolean>('cody.experimental.telemetry.enabled', true)
 
     return activateCommon(context, {
-        createLocalEmbeddingsController: isLocalEmbeddingsEnabled
-            ? (config: LocalEmbeddingsConfig): Promise<LocalEmbeddingsController> =>
-                  createLocalEmbeddingsController(context, config)
-            : undefined,
+        createLocalEmbeddingsController: isLocalEmbeddingsDisabled
```
olafurpg, line 63: Nit: I recommend keeping the "enabled" variable name as it's easier to read the code that way.

vscode/src/external-services.ts
```diff
@@ -76,7 +76,10 @@ export async function configureExternalServices(
         )
     }
 
-    const localEmbeddings = await platform.createLocalEmbeddingsController?.(initialConfig)
+    // Disable local embeddings for enterprise users.
+    const localEmbeddings = authProvider.getAuthStatus().isDotCom
+        ? await platform.createLocalEmbeddingsController?.(initialConfig)
```
olafurpg, line 81: We should ideally listen to with events and initialize embeddings on the condition that the user is authenticated to dotcom. I believe isDotcom is true for enterprise users because they authenticate.

vscode/src/local-context/local-embeddings.ts
```diff
@@ -552,6 +562,62 @@ export class LocalEmbeddingsController
                 this.updateIssueStatusBar()
             }
         }
+
+        await this.checkIndexStaleness(repoDir, health)
+    }
+
+    // Check if the embeddings are stale and refresh the index if needed.
+    private async checkIndexStaleness(repoDir: FileURI, health: IndexHealthResultFound): Promise<void> {
+        const repo = vscodeGitAPI?.getRepository(repoDir)
```
olafurpg, line 571: Non-blocking: BFG has full hit bindings so it might make sense to implement this on the Rust side. The VSC git APIs have spotty support for agent clients. The benefit of continuing investments in local embeddings is even smaller if we only aspire to support VSC.

olafurpg (changes requested):

vscode/src/external-services.ts
janhartman: Are you sure? I think `isDotCom` means PLG. At least that's what we use in at least 10 other places where the behavior is PLG/enterprise dependent.

vscode/src/local-context/local-embeddings.ts
janhartman: This was a very minimum-possible-effort PR so I don't believe I'll be touching the Rust side any more than necessary. The VSCode-only situation is OK.

vscode/src/extension.node.ts
janhartman: There is a `cody.experimental.localEmbeddings.disabled` setting which disables them everywhere.

vscode/src/extension.node.ts
janhartman: Yeah agreed, negatives are harder to parse. In my defense, I just reverted your changes to this file 😄

vscode/src/extension.node.ts
janhartman: @olafurpg Is there another elegant way I could do what you proposed - i.e. detect if I'm in VSCode or agent? The current config structure doesn't lend itself nicely to your ask.

vscode/src/extension.node.ts
olafurpg: It was a cleanup to remove the negatives and I'd prefer to not re-introduce it.

vscode/src/external-services.ts
olafurpg: `isDotCom` is true by default https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@a94f37d2b6c06bd75e991c84189b85df0dfc1cc7/-/blob/lib/shared/src/auth/types.ts?L47

We need to additionally guard against `isLoggedIn`.

vscode/src/local-context/local-embeddings.ts
olafurpg: Agreed, this is non-blocking. Just wanted to call out that the value of local embeddings from a product perspective is low when it's an exclusive feature for PLG on VSC only. In general, we want our projects to benefit all users across the board (PLG + Enterprise, also all clients).

olafurpg (approved):
stamp to unblock merging. The only hard requirement is that we don't re-introduce downloading BFG before authentication to dotcom. This needs to be tested manually, there are no integration tests for it. Make sure to handle the flow from unauthenticated->Enterprise (it's not enough to sign into Enterprise and reload, you need to sign out). It's also worth checking that embeddings work on the first run for the unauthenticated->PLG flow.

vscode/src/external-services.ts
janhartman: Thanks for pointing it out. This feels like a footgun.

vscode/src/extension.node.ts
janhartman: Thanks, fixed.

vscode/src/external-services.ts
olafurpg: There have been several attempts to refactor some of this logic. All the config passing code has lots of debt...

vscode/src/local-context/local-embeddings.ts
janhartman: Acknowledged. Let's make sure we deliver value everywhere with server-side context :)

chenkc805:
Do we have telemetry that measures when an a codebase is re-indexed?

janhartman:
@chenkc805 No, but it's a good idea, let me add it 👍

janhartman:
@olafurpg I followed the procedure you wrote down in #5140 and it seems everything's in order - we only download the binary if the users is PLG and local embeddings are enabled.

PR #5142, opened 2024-08-08
Move the printing of recordings directly to testing/requestErrors

mkondratek:
The diff information is important to the clients too. Instead of reimplementing it in each client (for the integration test purposes) let's print it directly from the `requestErros`.

## Test plan
- Green CI
- The recording missmatch is clearly reported in the client (test it with https://github.com/sourcegraph/jetbrains/pull/1930)

### DIFF
--- a/agent/src/TestClient.ts
+++ b/agent/src/TestClient.ts
@@ -1,7 +1,4 @@
 import assert from 'node:assert'
-import stable_stringify from 'fast-json-stable-stringify'
-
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
I would rather get those diffs as part of the response rather than printout to stdout.
But it's still improvement for JetBrains side.

Please just make sure it does not make experience worse for the agent tests.

Issue #4628, opened 2024-06-20
Edit: Document code doesn't follow the correct selectionRange


umpox:
### Version

Latest `main`

### Describe the bug

Document code uses `task.selectionRange`, but we only ask the LLM to produce documentation, not the code within the range.

This has some benefits:
- It's faster, we don't have to wait for the rest of the code to be repeated
- It's more reliable, we only want the LLM to get the docstring right, not anything else.

But we introduce some problems with this method:
- Cannot easily calculate decorations, as the selectionRange doesn't match what we're actually inserting.
- Retry is buggy, as it'll undo the entire selection range
- More?

### Expected behavior

We need _some_ range to provide to the LLM, we should propose a way for a task to have a separate `contextRange` (or similar) to the `selectionRange`.

For document code, this means we could have asmall selection range (empty at first) that grows as the documentation is added. This will mean that decorations, retry behaviour and anything else should just work as normal. We can still use `contextRange` to provide the correct code (that we want documented) to the LLM.

PR #5143, opened 2024-08-08
Code Lenses: disable "Edit & Retry" for non-user prompts

RXminuS:
Before the "Edit & Retry" lens was shown after using the "Ask Cody to Fix" code action. Although it worked this would then show a prompt that the user never wrote which is confusing.

Now we don't show the Edit & Retry lens and only allow Accept/Reject.

## Test plan
CI
Manually verified that the Edit & Retry lens no longer shows up

### DIFF
--- a/vscode/src/non-stop/codelenses/items.ts
+++ b/vscode/src/non-stop/codelenses/items.ts
@@ -31,29 +31,30 @@ export function getLensesForTask(task: FixupTask): vscode.CodeLens[] {
             return [title, cancel]
         }
         case CodyTaskState.Applied: {
-            const accept = getAcceptLens(codeLensRange, task.id)
-            const retry = getRetryLens(codeLensRange, task.id)
-            const undo = getUndoLens(codeLensRange, task.id)
-            if (isChatEdit) {
-                const reject = getUndoLens(codeLensRange, task.id, 'Reject')
-                return [accept, reject]
-            }
+            // Required:
+            const acceptLens = getAcceptLens(codeLensRange, task.id)
+            const undoLens = getUndoLens(codeLensRange, task.id, isChatEdit ? 'Reject' : undefined)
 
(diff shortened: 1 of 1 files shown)


RXminuS:
* **#5143** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5143?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

vscode/src/non-stop/codelenses/items.ts
```diff
@@ -31,29 +31,30 @@ export function getLensesForTask(task: FixupTask): vscode.CodeLens[] {
...
-                    // reverting specific lines
-                    isRunningInsideAgent() ? 'Show Diff' : 'Open Diff'
-                )
-                return [accept, retry, undo, showDiff]
-            }
-            return [accept, retry, undo]
+            // Optional:
+            // Retries only makes sense if the user created the prompt
+            // TODO: Are there other actions that shouldn't offer a retry?
+            const canRetry = task.source !== 'code-action:fix' && !isTest
```
umpox, line 41:
```suggestion
            // Optional:
            // Retries only makes sense if the user created the prompt
            const canRetry = task.intent !== 'fix' && task.intent !== 'doc` && !isTest
```

Should be able to use `intent` here.

We should exclude doc too as that's a pretty long prompt (and I think retry is broken right now due to [this issue](https://github.com/sourcegraph/cody/issues/4628))

umpox (approved):
added 1 comment but makes sense 👍

vscode/src/non-stop/codelenses/items.ts
RXminuS: Yeah makes sense. Thanks, that's exactly what I wanted from review ;-)

PR #5144, opened 2024-08-08
Smart Apply: Handle request errors

umpox:
## Description

Adds request error handling to the smart apply feature

Note: This is feature flagged and only enable for internal dogfooding right now

So if we get a network request or anything, we show it to the user. Previously we would end up just assuming we wanted to `insert`

## Test plan

1. Trigger a smart apply
2. Force a network error
3. Check the error message is shown to the user

### DIFF
--- a/vscode/src/edit/prompt/smart-apply.ts
+++ b/vscode/src/edit/prompt/smart-apply.ts
@@ -150,7 +150,6 @@ export async function promptModelForOriginalCode(
                 break
             }
             case 'error': {
-                console.log('GOT AN ERROR', message.error)
                 throw message.error
             }
         }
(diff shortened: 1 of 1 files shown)


fkling (approved):

vscode/src/edit/prompt/smart-apply.ts
```diff
@@ -175,14 +178,25 @@ export async function getSmartApplySelection(
...
+            model,
+            client,
+            codyApiVersion
+        )
+    } catch (error: unknown) {
+        // We erred when asking the LLM to produce the original code.
+        // Surface this error back to the user
+        if (error instanceof Error) {
+            vscode.window.showErrorMessage(`Error: ${error.message}`)
+        }
```
fkling, line 196: You could show a generic error message if `error` is not an instance of `Error` (which I guess will never be the case but 🤷🏻 )

PR #5147, opened 2024-08-08
Agent: new protocol to create chat sidebar panel

abeatrix:
This change adds a new JSONRPC endpoint `chat/sidebar/new` that allows creating a new chat panel with a sidebar view type, in addition to the existing `chat/new`  and `chat/web/new` endpoint that creates a chat panel with an editor view type.

The new endpoint returns the panel ID and chat ID that can be used to reference the session and restore the chat.

This change also updates the `ChatsController` to return the created panel when using the `cody.chat.newPanel` command.

## Test plan

Try calling the `chat/sidebar/new` from a client when creating a webview to confirm the webview is showing up as "sidebar" view type that has the nav bar on top.

## Changelog

Adds a new JSONRPC endpoint `chat/sidebar/new` that allows creating a new chat panel with a sidebar view type.


abeatrix:
Closing since the PR by Dom will fix this issue!

PR #5149, opened 2024-08-08
Chore: Enable TS 5.5

RXminuS:
Before we were blocked from using TS 5.5 due to some class declaration overrides that relied on buggy TS 5.4 behavior that was fixed. 

Not only did this lead to hard to debug type errors when trying to upgrade. It also required any contributor to Cody to explicitly set their VSCode TS version to be the workspace version (instead of the default VSCode version). A significant point of friction.

Now we work around these type issues and upgrade to TS 5.5 which bring many great other DX improvements such as inferred type predicates etc.

## Test plan
CI where all tests should pass just the same.

NOTE: There was one change in a test snapshot w.r.t. quotes output by TSC but this was likely unrelated to these changes and instead due to changes in the compiler.


RXminuS:
* **#5149** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5149?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

olafurpg (approved):
You are a beast Rik. I had tried to upgrade but couldn't make sense of the errors. THANK YOU! Looking forward to enabling isolated declarations

RXminuS:
> Looking forward to enabling isolated declarations

You might run into a few type issues with the type-hacks that I added before, I haven't tried that yet. If so just disable them from the build and give me a ping and I'll migrate them.

PR #5151, opened 2024-08-08
symf improvements

beyang:
* Bumps symf version to better handle cases where a user search term exact-matches a symbol name, but without the proper casing (e.g., `chatcontroller` should match `ChatController`)
* Truncate results returned from symf to be max 2KB. This addresses the issue where very large classes (or other symbols) are excluded from context, even when they are at the top of the relevance list, because they don't fit.
* Logs symf invocations for better debuggability

## Test plan

Test locally

## Changelog

Improves keyword context matches ("foobar" now matches "FooBar" with higher relevance) and results returned from local keyword search are also truncated so that they are definitely included in the context, even when very large.


thenamankumar (approved):
I can not see changes regarding the case insensitive matching of keywords in the PR. I assume that comes as part of symf v0.0.16. 

LGTM

janhartman (approved):

PR #5153, opened 2024-08-08
Add metadata to context items

beyang:
Context items now include optional metadata about their source or scoring, which can help developers understand why certain items are appearing in the context. This change adds a `metadata` field to the `ContextItemCommon` interface and populates it for context items from Sourcegraph search, the local symbol index, and live search results.

You can view these by setting `"cody.internal.debug.context": true` in settings. The `cody.internal.showContextAlternatives` has been removed and replaced by this one.

## Test plan

Test locally, doesn't affect prod.

### DIFF
--- a/vscode/src/chat/chat-view/ContextRetriever.ts
+++ b/vscode/src/chat/chat-view/ContextRetriever.ts
@@ -245,6 +245,7 @@ export class ContextRetriever implements vscode.Disposable {
                         range,
                         source: ContextItemSource.Search,
                         content: text,
+                        metadata: ['source:symf-live'],
                     }
                 })
             )
@@ -313,7 +314,11 @@ export class ContextRetriever implements vscode.Disposable {
         if (isError(remoteResult)) {
             throw remoteResult
         }
-        return remoteResult?.flatMap(r => contextSearchResultToContextItem(r) ?? []) ?? []
+        const finalResults = remoteResult?.flatMap(r => contextSearchResultToContextItem(r) ?? []) ?? []
+        for (const r of finalResults) {
+            r.metadata = (r.metadata ?? []).concat(['source:sourcegraph'])
+        }
+        return finalResults
     }
 
     private async retrieveIndexedContextLocally(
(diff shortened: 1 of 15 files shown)


vscode/src/configuration.ts
```diff
@@ -135,6 +135,7 @@ export function getConfiguration(
          */
 
         internalUnstable: getHiddenSetting('internal.unstable', isTesting),
+        internalDebugContext: getHiddenSetting('internal.debugContext', false),
```
thenamankumar, line 138: I wonder what we can do to make adding new settings easier. Something which doesn't require updating 3 different files.

thenamankumar (approved):
Will it be better to link this with `cody.debug.enable.all` "Enable Debug Mode" rather than a separate config flag?

LGTM

vscode/src/chat/chat-view/ContextRetriever.ts
```diff
@@ -244,6 +244,7 @@ export class ContextRetriever implements vscode.Disposable {
                         range,
                         source: ContextItemSource.Search,
                         content: text,
+                        metadata: ['source:symf-live'],
```
janhartman, line 247: Should we maybe add symf-live into the context source enum?

janhartman (approved):

vscode/src/configuration.ts
beyang: Yeah, tbh we should probably just use the vscode.workspace.getConfiguration api. What we've built around that feels like an unnecessary abstraction.

beyang:
> Will it be better to link this with cody.debug.enable.all "Enable Debug Mode" rather than a separate config flag?

"Enable Debug Mode" is technically user visible and I want to keep this internal for now (so as to not worry about the way it looks at all)

> Should we maybe add symf-live into the context source enum?

Not sure, I think that belongs in a separate change, as that alters how we want to convey to the end user what is happening. "This came from symf" feels like a low-level implementation detail, where as "came from search" maps to more of a user-level mental model.

PR #5157, opened 2024-08-09
Agent: Changes required for running webview in clients

pkukielka:
Context: https://sourcegraph.slack.com/archives/C06R69BC8UW/p1723199277787079

## Changes

A bunch of changes needed for getting chat webview to work in other clients, e.g JetBrains, Visual Studio and Eclipse.

Related PR: https://github.com/sourcegraph/jetbrains/pull/2022

## Test plan

Full QA on jetbrains branch


agent/src/agent.ts
```diff
@@ -1130,17 +1151,11 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
-            const panelId = await this.createChatPanel(
-                Promise.resolve({
-                    type: 'chat',
-                    session: await vscode.commands.executeCommand('cody.chat.newEditorPanel'),
-                })
-            )
-
-            const chatId = this.webPanels.panels.get(panelId)?.chatID ?? ''
-            return { panelId, chatId }
+            await vscode.commands.executeCommand('cody.chat.newEditorPanel')
```
pkukielka, line 1154: Won't that break any other client?

abeatrix (approved):
Left some comments in line and the tests need fixing, but the rest look good to me.

I tested the build for VS Code and everything looks the same as we currently do

image

I noticed there is a bug in the chat webview where clicking on the New Chat button in the sidebar would start a new Editor Chat (where I would expect it to clear my current sidebar chat since there is a move to editor button), but this is reproducible on main, unrelated to this PR. I've reported these bug on Slack here: https://sourcegraph.slack.com/archives/C05MW2TMYAV/p1723207420306119

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -525,6 +594,17 @@ export interface ClientCapabilities {
...
+    // AgentWebViewPanel which just delegates bidirectional postMessage over
+    // the Agent protocol. If 'native', implements a larger subset of the VSCode
+    // WebView API and expects the client to run web content in the webview.
+    // Defaults to 'agentic'.
+    webview?: 'agentic' | 'native'
+    // If webview === 'native', describes how the client has configured webview resources.
+    // cspSource is passed to the extension as the Webview cspSource property.
+    // webviewBundleServingPrefix is prepended to resource paths under 'dist' in
+    // asWebviewUri (note, multiple prefixes are not yet implemented.)
+    webviewNativeConfig?: { cspSource: string; webviewBundleServingPrefix: string }
```
abeatrix, line 607:
```suggestion
    webview?: 'agentic' | 'native' | undefined | null
    // If webview === 'native', describes how the client has configured webview resources.
    // cspSource is passed to the extension as the Webview cspSource property.
    // webviewBundleServingPrefix is prepended to resource paths under 'dist' in
    // asWebviewUri (note, multiple prefixes are not yet implemented.)
    webviewNativeConfig?: { cspSource: string; webviewBundleServingPrefix: string } | undefined | null
```

I think these are causing the tests to fail?

agent/package.json
```diff
@@ -13,7 +13,8 @@
   "scripts": {
     "build:root": "pnpm -C .. run -s build",
     "build:agent": "node src/esbuild.mjs",
-    "build": "pnpm run -s build:root && pnpm run -s build:agent",
+    "build:webviews": "pnpm -C ../vscode _build:webviews --mode production --outDir $PWD/dist/webviews",
```
abeatrix, line 16:
```suggestion
    "build:webviews": "pnpm -C ../vscode _build:webviews --mode production --outDir ../../agent/dist/webviews",
```

Running this created a folder in my vscode directory called "$PWD/dist/webviews" 🤣

agent/src/agent.ts
```diff
@@ -1130,17 +1151,11 @@ export class Agent extends MessageHandler implements ExtensionClient {
...
+            return { panelId: 'TODO-remove-panel-id', chatId: 'TODO-remove-chat-id' }
```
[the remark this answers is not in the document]
abeatrix: i've reverted this change since the panelId here is just TODO.

agent/src/agent.ts
abeatrix: i've reverted this change since the panelId here is just TODO.

.npmrc
```diff
@@ -1 +1,11 @@
 hoist=false
+1
```
olafurpg, line 2: Was this intentional? @pkukielka

olafurpg: Reverted in https://github.com/sourcegraph/cody/pull/5172

PR #5164, opened 2024-08-09
ensure "New Chat" in sidebar opens in sidebar, not editor

beyang:
Fixes: https://linear.app/sourcegraph/issue/SRCH-872/new-chat-button-should-open-up-new-chat-in-same-location-editorsidebar

## Test plan

Test locally

## Changelog

Fixed: The "New Chat" button in the sidebar now always opens the new chat in the sidebar, never the editor panel. Previously, a bug was introduced where the behavior was "sticky" as to which option the user last selected.


abeatrix (approved):
ran it locally and confirmed clicking on "New Chat" in sidebar starts a new chat in sidebar. Thanks!

PR #5170, opened 2024-08-11
Chat: fix chat input focus issues

abeatrix:
- Update the chat transcript container styles to use `height: 100%` and `overflow: scroll` to properly handle overflow of chat messages

- Adjust the padding and margin of the transcript container to improve the layout

Fixes an issue where @-mentions would cause the focus to shift to the top chat input box due to the container overflow settings causing the reference to the bottom chat input box to be lost.

## Test plan

no-reset

Before

before


sqs (approved):
unblocking

PR #5172, opened 2024-08-12
Agent: minor fixes for 'agentic' chat

olafurpg:
The PR https://github.com/sourcegraph/cody/pull/5157 introduced a few regressions for "agentic" chat (the default style of chat in the agent). This PR fixes those regressions.

## Test plan
Unit tests and manually tested with Eclipse.

## Changelog
Make default webview `agentic` so that webviews for other IDEs can be bundled in the extension install.


jamesmcnamara (approved):
I made some small changes here to make the agent tests agentic and tested it with eclipse. It works (once we add that flag)!

olafurpg:
Thank you @jamesmcnamara for fixing the tests! 🙌🏻

PR #5174, opened 2024-08-12
Chat: update webview content security policy template

abeatrix:
- Update the content security policy in the chat webview template file index.html to use `'self'` instead of `{cspSource}` to allow the webview to load resources from the same origin by default without needing to specific cspSource

This change allows clients to run the webview with the template file by default

## Test plan

Green UI - This  change should not have impact with any current features or webviews setup

### DIFF
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -1889,5 +1889,5 @@ export async function addWebviewViewHTML(
     // 2. Update URIs for content security policy to only allow specific scripts to be run
     view.webview.html = decoded
         .replaceAll('./', `${resources.toString()}/`)
-        .replaceAll('{cspSource}', view.webview.cspSource)
+        .replaceAll("'self'", view.webview.cspSource)
 }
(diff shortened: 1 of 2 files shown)


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -1889,5 +1889,5 @@ export async function addWebviewViewHTML(
     // 2. Update URIs for content security policy to only allow specific scripts to be run
     view.webview.html = decoded
         .replaceAll('./', `${resources.toString()}/`)
-        .replaceAll('{cspSource}', view.webview.cspSource)
+        .replaceAll("'self'", view.webview.cspSource)
```
olafurpg, line 1892: Might want to refine this regexp to allow "self" elsewhere, probably not a big issue though. Alternative is to add a unit test to assert the expected template HTML after this transformation.

olafurpg (approved):

olafurpg:
FYI @jamesmcnamara this probably requires changes in Eclipse. We need to handle "self" now.

PR #5183, opened 2024-08-13
Edit: Correctly guard against duplicate in-progress edits

umpox:
## Description

Edit `instruction` is no longer a `string`, but a `PromptString` object.

This meant that our equality check failed, the two strings were identical, but not considered the same due to how object equality works in JS

## Test plan

1. Select some code
1. Open the Cody sidebar
1. Hit "Document Code" or "Generate Unit Test" multiple times
1. Confirm that only one "Document Code" or "Generate Unit Test" command runs


olafurpg (approved):
Nice tracking this down! Tricky bug

PR #5191, opened 2024-08-13
fix/CLI: fail fast if the provided --context-file is too large

olafurpg:
Previously, the CLI didn't give any hint when `--context-file` was too
    large to fit into the context window. The VS Code chat UI refuses to
    add these files as explicit context. This PR fixes the problem so that
    the CLI also fails fast in the same scenario. Users can pass the
    `--ignore-context-window-error` to skip this check.

## Test plan
Manually tested with
```
> pnpm -C agent build:agent && node agent/dist/index.js chat --context-file agent/src/cli/scip-codegen/scip.ts --context-file agent/src/agent.ts -m 'list all the methods here'
✖ The provided context is too large to fit into the context window.
To fix this problem, either remove the files from --context-file or
edit these files so they become small enough to fit into the context window.
Alternatively, set the flag --ignore-context-window-errors to skip this check.

File                                                                     Reason
-----------------------------------------------------------------------  --------------------------------------------------------------------------
/Users/olafurpg/dev/sourcegraph/cody/agent/src/cli/scip-codegen/scip.ts  user context tokens exceeded remaining user context tokens (49538 > 29970)
```

To skip the check
```
pnpm -C agent build:agent && node agent/dist/index.js chat --context-file agent/src/cli/scip-codegen/scip.ts --context-file agent/src/agent.ts -m 'list all the methods here' --ignore-context-window-errors
```

### DIFF
--- a/agent/CHANGELOG.md
+++ b/agent/CHANGELOG.md
@@ -10,6 +10,15 @@ This is a log of all notable changes to the Cody command-line tool. [Unreleased]
 
 ### Changed
 
+## 5.5.12
+
+### Fixed
+
+- `cody chat --context-file` will now fail fast if the provided file is too
+  large to fit into the context window. Previously, `cody chat` silently ignored
+  this error and the LLM would respond with a confusing message about missing
+  context. To bypass this check, use `--ignore-context-window-errors`.
+
 ## 5.5.11
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


jamesmcnamara (approved):
Looks good!

PR #5194, opened 2024-08-13
fixed(web): removed `fs-extra`

jamesmcnamara:
There's a current bug somewhere (I think in Vite?) that tries to externalize node dependencies, and print out a warning message. But when it loads the dependencies of `fs-extra`, specifically `graceful-fs` it comes in as a symbol instead of string, which can't be interpolated and it crashes.
image

Our only use of it is to copy a file, which we can do with `fs` so it is unneeded.
## Test plan
Compiles and runs in browser.


abeatrix (approved):
ty!

PR #5197, opened 2024-08-13
Debug: make debug commands available outside of test

abeatrix:
The debug commands were recently moved to be available in test/dev mode only, most likely by accident. 

This PR fixes this issue by always registering the commands for debugging on extension start-up.

## Test plan

Run the Export Debug Logs command and verify the command works:

image

Before this change, the command returns an error

image

## Changelog

Fixed issues where debug commands are not available in production.


aramaraju (approved):
LGTM

PR #5198, opened 2024-08-13
Agent: Fix setContext command args value always undefined

abeatrix:
Fix an issue with vscode shim where the args value of setContext command always returned undefined, due to args was not being deconstructed properly during the "executeCommand" step.

Also updated the logic for displaying the new sign-in form for non-vscode IDE clients

## Test plan

'window/didChangeContext' should now show the correct key-value pair for notifying authStatus change:

Screenshot 2024-08-13 154449

## Changelog

NA

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -917,15 +917,12 @@ const _commands: Partial<typeof vscode.commands> = {
             }
         })
     },
-    executeCommand: (command, args) => {
+    executeCommand: (command, ...args) => {
         const registered = registeredCommands.get(command)
         if (registered) {
             try {
                 if (args) {
-                    if (typeof args === 'object' && typeof args[Symbol.iterator] === 'function') {
-                        return promisify(registered.callback(...args))
-                    }
-                    return promisify(registered.callback(args))
+                    return promisify(registered.callback(...args))
                 }
                 return promisify(registered.callback())
             } catch (error) {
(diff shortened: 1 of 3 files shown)


vscode/webviews/OnboardingExperiment.tsx
```diff
@@ -76,8 +76,8 @@ export const LoginSimplified: React.FunctionComponent<React.PropsWithoutRef<Logi
     const otherSignInClick = (): void => {
         vscodeAPI.postMessage({ command: 'auth', authKind: 'signin' })
     }
-    const isCodyWeb = uiKindIsWeb || codyIDE === CodyIDE.Web
-    const isNonVSCodeIDE = !isCodyWeb && codyIDE !== CodyIDE.VSCode
+    const isNonVSCodeIDE = codyIDE !== CodyIDE.Web && codyIDE !== CodyIDE.VSCode
+    const isCodyWebUI = (uiKindIsWeb || codyIDE === CodyIDE.Web) && !isNonVSCodeIDE
```
abeatrix, line 80: turns out all clients have uiKindIsWeb set to true, so we should check for IDE name (temporarily) to display the new sign in form. Right now it is not displaying for any client because uiKindIsWeb  is always true.

kalanchan (approved):

agent/src/vscode-shim.ts
```diff
@@ -945,12 +942,13 @@ const _commands: Partial<typeof vscode.commands> = {
 _commands?.registerCommand?.('workbench.action.reloadWindow', () => {
     // Do nothing
 })
-_commands?.registerCommand?.('setContext', (key, value) => {
+_commands?.registerCommand?.('setContext', (...args) => {
+    const [key, value] = args
```
jamesmcnamara, line 946: Does this not work if you just leave it as
```suggestion
_commands?.registerCommand?.('setContext', (key, value) => {
```

agent/src/vscode-shim.ts
```diff
@@ -945,12 +942,13 @@ const _commands: Partial<typeof vscode.commands> = {
...
     if (typeof key !== 'string') {
         throw new TypeError(`setContext: first argument must be string. Got: ${key}`)
     }
     context.set(key, value)
-    agent?.notify('window/didChangeContext', { key, value })
+    agent?.notify('window/didChangeContext', { key, value: `${value}` })
```
jamesmcnamara, line 951: As we noticed from that bug earlier, some types like Symbol can't be interpolated into a string.
```suggestion
    agent?.notify('window/didChangeContext', { key, value: value.toString() })
```

jamesmcnamara (approved):

agent/src/vscode-shim.ts
abeatrix: Yup that works, thank you!
image

PR #5201, opened 2024-08-14
fix many chat UI glitches

sqs:
Fixes many chat UI glitches that I've noticed over the last few weeks (many which I introduced):

- Fixes and simplifies positioning of the @-mention menu, now using only Lexical's standard positioning and not `@floating-ui/react`. We use a small fork of the Lexical editor now (unfortunately) at https://github.com/sourcegraph/lexical. See the commits on our fork for why.
- Fixes the ScrollDown component, which showed a big down arrow at the bottom of the chat window if it could be scrolled down.
- Fixes the bottom padding in chat.
- Removes a lot of unnecessary CSS that created needless overflow-auto scroll areas.

## Test plan

For @-mentions: Near the top and bottom of the viewport, open an @-mention. Ensure it flows toward the part of the viewport with more space. Ensure that scrolling does not close it unless it is fully off screen. Resize the chat panel and ensure the @-mention menu remains anchored to the `@` in the editor.

For ScrollDown: In chat, ask `list the 50 states` and ensure the down arrow shows up automatically when the reply goes off screen.

For panel scrolling: Make the VS Code window short and ensure that the chat, history, and prompts panels scroll correctly.

## Changelog

- Fix and stabilize positioning of the @-mention menu in various screen configurations.
- Fix the "scroll down" arrow that appears when a chat response has filled up the screen and the user can scroll to see more.


olafurpg (approved):
LGTM 👍🏻 These are huge improvements for small screens, esp. Chat in mobile web

* I confirmed that the `@-mention` menu doesn't have weird positioning problems anymore on small screens
* I confirmed the ScrollDown appear, similar to ChatGPT, nice!
* Didn't manage to reproduce the panel scrolling part (didn't try much)

PR #5208, opened 2024-08-15
Smart Apply: Enable in other clients based on client capabilities

umpox:
## Description

The Smart Apply functionality should be OK to enable in other clients as:
1. We have a generic UI that should work in any webview client. PR: https://github.com/sourcegraph/cody/pull/5188
2. It proxies from chat to inline edit, so any client that has the `edit` capability should be able to use this

## Test plan

Test in other clients that have the Edit capability

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -725,11 +725,11 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: 1d6b7417dc13e78cf0704907809a5532
+    - _id: d30452695a5fef843d6895d066e114c0
       _order: 0
       cache: {}
       request:
-        bodySize: 2126
+        bodySize: 2378
         cookies: []
         headers:
           - name: content-type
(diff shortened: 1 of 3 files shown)


umpox:
https://github.com/user-attachments/assets/20927704-b3d2-4b3a-931a-a09c38768af4

Seems to be working in JetBrains. (I see to have no styles though)

Tagging @sourcegraph/jetbrains-plugin for a review and help testing

mkondratek (changes requested):
image

I tried it. It works nice but the UI is inconsistent with the other buttons (the action texts should not be visible). I am not 100% sure how to fix it but I would take a look at this cody:
https://sourcegraph.com/-/editor?remote_url=github.com%2Fsourcegraph%2Fcody&branch=tr%2Fchat-smart-apply-client-capabilities&file=vscode%2Fwebviews%2Fchat%2FChatMessageContent%2Fcreate-buttons.ts&start_row=119&start_col=8&end_row=130&end_col=9&editor=JetBrains&version=v6.0.26-nightly

umpox:
@mkondratek Thanks for taking a look!

> the action texts should not be visible

The UI changes are intentional here, we do actually want to make "Copy" and "Apply" more visible. These are likely to be more higher frequency actions that the other ones (Copy is used way more from looking at usage data, Apply is looking that way too)

I'll tag @danielmarquespt on this too for 👀 when he's back from PTO, the UI change is also behind a feature flag so I don't think we should block this though (we can quickly revert if necessary).

It does seem that we are missing some code block styles though, the action buttons look way different to VS Code:

image

image

I wonder if we're missing some styles somewhere? 🤔

mkondratek (approved):
ok then

PR #5210, opened 2024-08-15
Add Directory Mention provider for Cody

thenamankumar:
Closes: https://linear.app/sourcegraph/issue/SRCH-892/support-mentioning-dirs-and-better-monorepo-support-overall

This PR adds @Directories and @Remote Directories mention providers for Cody. 

Merge it only after this commit is merged & deployed on sourcegraph.sourcegraph.com: https://github.com/sourcegraph/sourcegraph/commit/0684b550cb318e9cbe3d68790ea15dda156b410e

It requires a new `filesPattern` filter on `getCodyContext` API. Therefore the providers are limited to SG version 5.7.0 and the insider builds. 

## Test plan

Demo: https://www.loom.com/share/e55afcb2a5064a708a08f7851a618b2e?sid=8d68a39c-521a-4b47-9e5d-fb769f3975f4

## Changelog

- Add @-mention directories & remote directories provider for enterprise Cody users.


camdencheek:
Backend PR [here](https://github.com/sourcegraph/sourcegraph/pull/64492)

vscode/src/context/openctx/currentRepositoryDirectorySearch.ts
```diff
@@ -0,0 +1,129 @@
...
+
+        meta() {
+            return {
+                name: customTitle ?? 'Directories',
+                mentions: {},
+            }
+        },
+
+        async mentions({ query }) {
+            const currentFile = vscode.window.activeTextEditor?.document.uri
```
abeatrix, line 28: I think this might return the editor chat instead of the document you are looking for if user is using the chat view in editor instead of sidebar. You can get the currentFile using getEditor().active for this case

sqs:
This is awesome. When @-mentioning a directory in your current workspace, you mention it's fetching context from the server. This means it will lack any local changes and won't get the entire files. This could be confusing to users:

1. What if the dir they want doesn't exist remotely?
2. What if it exists remotely on a non-default branch?
3. What if the files have changed?
4. I understand that relying on the server is necessary for big dirs where we need to context-search inside of them to select context snippets, but if it's just a few files, users might want the entire contents to be included (not just snippets).

I think we can break down the use cases for directory @-mentions into:

- (A) I want to use relevant context (but not the entire contents) from a big directory that I haven't made local changes to. I don't know which specific files or snippets that I want to use.
- (B) I want to use the entire contents from all files in a small directory that I may have made local changes to.

This PR addresses (A) but not (B). So, definitely worth merging as this is a huge step forward, but I think we need to solve for (B) as well *and* remove user confusion about what is happening. I'll get @rrhyne's help here.

rrhyne:
@sqs from the loom, I think this covers both local and remote directories. I will say I was a bit surprised at the distinction though. 

On that topic, I think we should have a single directory provider that can accept a local or remote URL. 

@thenamankumar One way to accomplish this is to add a top-level item to the @directory menu with a label that reads "Enter remote directory URL". I believe you could then trigger your RemoteDirectory functionality. 

Question: What kind of directories? Does it support many code hosts, or is this only GitHub? We should consider strategic scale and code host support before this is enterprise-ready.

sqs:
@rrhyne:

> @sqs from the loom, I think this covers both local and remote directories. 

It supports directories in the current repo, but the directory list and the contents are both pulled from the server, which means it won't include local changes.

camdencheek:
> Does it support many code hosts, or is this only GitHub?

The remote directories are pulled from a Sourcegraph instance, not directly from GitHub, so it should support any code host supported by Sourcegraph. We do not support naming an arbitrary remote directory because this is implemented as a filter on the context search, and we cannot do a context search on an arbitrarily-large directory on a codehost.

thenamankumar:
Thanks for the feedback @sqs, all of it makes sense. I have moved the conversation to [linear](https://linear.app/sourcegraph/issue/SRCH-892/support-mentioning-dirs-and-better-monorepo-support-overall#comment-fde2b4ce). Will work with @rrhyne to figure out how we can solve the usecase B. As it is more like a UX problem than a technical one, I'll merge this PR and do the further needful as an incremental improvement. Anyways it will be released to enterprise customers in September.

vscode/src/context/openctx/currentRepositoryDirectorySearch.ts
thenamankumar: Thanks, updated!

camdencheek (commented):
(Nonblocking) Do we have the ability to prettify the context item display? With the full URL, the important part of the context (the actual dir name) is not visible.

screenshot-2024-08-20_08-08-36@2x

camdencheek (approved):

PR #5213, opened 2024-08-15
added(build): commands for building only webviews

jamesmcnamara:
Trying to simplify the test loop on [eclipse](https://github.com/sourcegraph/eclipse/pull/46) by adding a command to just build webviews.

Additionally reverts https://github.com/sourcegraph/cody/pull/5194 because it turns out windows does need fs-extra as the use of graceful-fs handles certain timing issues

## Test plan
Tested web via `pnpm dev` as that was previously failing from the call to fs-extra.


abeatrix (approved):

pkukielka:
This PR also fixes copying webview files in general, so it would be nice to merge it ASAP.
I did a small fix for the protocol to make typechecker happy:
https://github.com/sourcegraph/cody/pull/5213/commits/5c2318b007046939aa0e930e92cd314808832df8

If the build will pass I will merge it.

PR #5214, opened 2024-08-16
fix a bug where the current file would not show up when searching for @-file mentions

sqs:
This was unintentionally caused by https://github.com/sourcegraph/cody/pull/5057. The intent was to not show 2 mention items `Current File: foo.ts` and `foo.ts`, but it meant that if you typed `@foo` then neither would show up.

Also simplify how initial context is passed to MentionMenu. Just include it in `items` so we don't spread responsibility for displaying and filtering it across multiple components.

A future improvement would be to move all of this logic to the backend (extension host) and just give the UI the fully filtered list.

## Test plan

Open a new Cody chat. Type `@` and then the name of the currently open file in your editor. Confirm that Cody shows `Current File: ...` as an option.

## Changelog

- Fixed a bug in chat where the current file would not show up when @-mentioning a file and searching for it by name.


peterguy (approved):
Worked locally; didn't find any issues.

PR #5217, opened 2024-08-16
Test: fix e2e test run step for Windows

abeatrix:
Attempt to fix this error on main where the command to set env var on Windows are different on Windows:

image

## Test plan

Green CI?


jhchabran (approved):
Oh damn, sorry I totally missed this 🙏 

Thanks for fixing this 🙇

PR #5221, opened 2024-08-18
observe config changes in AuthProvider

sqs:
The goal is to make Cody respond to config and auth changes better, so that no full-editor-reloads are required, and it reacts instantly to connectivity loss and reestablishment. This PR is one of many related ones, and this particular one makes it so that the AuthProvider listens to config changes more reactively. It's still not totally reactive; sometimes `AuthProvider.auth` still is called. But it gets us closer.

Helps with https://linear.app/sourcegraph/issue/CODY-3332/full-native-offline-mode-and-instant-config-reactivity-for-cody-in-vs.

Original intent: https://sourcegraph.slack.com/archives/C05AGQYD528/p1723791822404399.

## Known issues

We will need to fix these before releasing this:

There are still some glitches around account switching and getting the client config (ClientConfigSingleton). Almost all of these issues are related to the clientConfig observable being out of sync with the authStatus for a tick, which I haven’t figured out the reason for but have plenty of ideas.

- [ ] When switching from dotcom to s2, the model selector does not show up post-switch in any visible chats
- [ ] When switching from s2 to dotcom, it shows non-server-side-model-config models for 60 seconds until it refreshes client config
- [ ] When switching from dotcom to s2, it briefly shows the "could not create 'sourcegraph' autocomplete provider" message but then autocomplete does work because it immediately fixes its configuration
- [ ] On dotcom on a freshly reloaded editor, sometimes when you chat, it says it will use Claude 3.5 Sonnet but actually uses an Ollama model.
- [ ] I want to double-check that the embeddings service is only running for dotcom.
- [ ] And that the embeddings service starts up when you switch from s2 to dotcom, and shuts down when you switch from dotcom to s2.
- [ ] The ClientConfigSingleton now fetches new client config every 60s. Previously it would trigger an async refetch if the data was more than 60s old and was requested due to a user action. This means that it will now fetch in the background more often, which reintroduces the problem https://linear.app/sourcegraph/issue/CODY-3745/codys-background-periodic-network-access-causes-2fa. We need to make it not perform its background fetch if the editor is not the foreground window.
- [ ] Switching models is slow. There is probably some kind of state update cycle. Need to debug this by adding `tapLog` pipe operators in some observable chains.
- [ ] `Error: command 'cody.embeddings.resolveIssue' already exists`
- [ ] When the editor is first starting, a chat editor panel briefly shows the login screen (probably before the auth is finished) and then shows chat after auth is finished. Need to make the chat wait for some kind of initialization before showing anything.
- [ ] Need to confirm that SiteProductVersion request spam is not occurring. (Have confirmed but want to do so again before a release in case anything changes.)

## Test plan

e2e tests, and the account switching described above.


sqs:
@Manouchehri We already have offline mode, but this (or a future PR) will support an ephemeral offline mode that is a bit smoother. Not a big new feature or anything, just something that makes it handle network connectivity issues better.

beyang (approved):

vovakulikov:
I am just flagging it here for the record (since this PR has been approved); before merging this, we have to check that this PR works against a clean run in Cody Web demo (by clean, I mean with reset data in Index DB in the Cody Web demo) and check that Cody Web chat loads. 

I think currently (this is the last commit b2749d4fbe53e4e47ac8cc24ba59db44383f8675) Cody Web fails with no auth runtime error

sqs:
> I am just flagging it here for the record (since this PR has been approved); before merging this, we have to check that this PR works against a clean run in Cody Web demo (by clean, I mean with reset data in Index DB in the Cody Web demo) and check that Cody Web chat loads.
> 
> I think currently (this is the last commit [b2749d4](https://github.com/sourcegraph/cody/commit/b2749d4fbe53e4e47ac8cc24ba59db44383f8675)) Cody Web fails with no auth runtime error

Just confirmed on the latest push that `pnpm -C web dev` works after I clear IndexedDB.

It still has a bit of SiteProductVersion and other GraphQL request spam, which I will fix before merging.

dominiccooney:
Just tried patching this and it breaks JetBrains in some trivial ways (some model tags are nullable) and maybe more complicated ways (some problem deserializing AuthStatus at com.sourcegraph.cody.agent.protocol_generated.AuthStatus.deserializer$lambda$0(AuthStatus.kt:22) ... this is generated, interesting that it can fail...)

vovakulikov:
@sqs

> Just confirmed on the latest push that pnpm -C web dev works after I clear IndexedDB.

This problem is tricky, you have to clear the index db each time you load cody chat because we seed the index db with some standard data structure after every chat load. 

This is how you can still reproduce this (this is the latest commit from this branch)

https://github.com/user-attachments/assets/c0e10c4e-8556-4a19-97dc-f6098b5065d8

dominiccooney:
> some problem deserializing AuthStatus ...

Looked at this a bit, and the bindings generator is confused by UnauthenticatedAuthStatus, AuthenticatedAuthStatus and the AUTH_STATUS_FIXTURE_AUTHED etc.

It has generated a type which appears to expect endpoint to always be https://example.com which is obviously not right.

sqs:
> @sqs
> 
> > Just confirmed on the latest push that pnpm -C web dev works after I clear IndexedDB.
> 
> This problem is tricky, you have to clear the index db each time you load cody chat because we seed the index db with some standard data structure after every chat load.
> 
> This is how you can still reproduce this (this is the latest commit from this branch)
> 
>  Screen.Recording.2024-09-18.at.19.16.17.mov

@vovakulikov Indeed, that is what I did. I just did it a few more times and it still worked, even when I simulated a Fast 4G internet connection. Can you try it on the latest commit here? Maybe I was on a newer commit.

lib/shared/src/misc/observable.test.ts
```diff
@@ -580,3 +654,91 @@ describe('concat', () => {
...
+
+        await vi.advanceTimersByTimeAsync(29)
+        expect(values).toStrictEqual<typeof values>([])
+
+        await vi.advanceTimersByTimeAsync(1)
+        expect(values).toStrictEqual<typeof values>([['b', 'x']])
+        values.length = 0
+
+        await vi.advanceTimersByTimeAsync(30)
+        expect(values).toStrictEqual<typeof values>([['c', 'y']])
```
dominiccooney, line 678: Would be nice to test this where one or the other observable advances, but not both.

dominiccooney (approved):
LGTM, would be nice to reenable the e2e enterprise-cody-ignore test if the autocomplete at cursor command is in fact being registered now.

vovakulikov:
I've noticed one regression in one box logic, Cody Web demo is connected to s2 now where we have enabled one box feature flag, and I see that chat intent detection returns search intent (this means we produce an empty message from LLM) but we should render code snippets instead in this case which doesn't happen on this branch for some reason. I think something happened in the config event, which no longer produces the right value of one box. 

Probably due to `this.featureCodyExperimentalOneBox.value.last` change in ChatController 

cc @sqs

vovakulikov:
Also important thing is that the production build of Cody Web fails in the Sourcegraph Web app with no auth error 

Screenshot 2024-09-19 at 14 50 00

vovakulikov:
Instruction to reproduce 

In Cody Repo 
- Upgrade Cody web package.json version to `0.9.0`
- `pnpm build` in web directory 
- `pnpm link` --global in web directory 

In Sourcegraph Repo
- Specify 0.9.0 version in `@sourcegraph/cody-web` package
- `pnpm link @sourcegraph/cody-web --global`
- Run any sg command to serve client (`sg start web-standalone`)

vovakulikov (changes requested):
This problem with Cody Web in production bothers me a lot, so I request changes to avoid any possible merges here for now

agent/src/agent.ts
```diff
@@ -444,9 +447,8 @@ export class Agent extends MessageHandler implements ExtensionClient {
                     ? this.handleConfigChanges(clientInfo.extensionConfiguration, {
```
vovakulikov, line 447: @sqs I think this is the place because of which it fails in the Sourcegraph Web app. In the web app, we don't have any access token, we use cookie auth flow instead so the access token is always null in the Sourcegraph Web app, but I think we still should run handleConfigChanges because otherwise we never emit auth value by `authProvider.validateAndStoreCredentials`

vovakulikov: I think I found it, so I see that you already got CodyWeb special case inside `validateAndStoreCredentials` method. But I think `validateCredentials` ignores cody web since we don't pass agentIDE in handleConfigChanges. 

So I think we should
In Agent constructor parse and handle Cody Web case
```ts
const ide = AgentWorkspaceConfiguration.clientNameToIDE(clientInfo.name)
this.authenticationPromise = clientInfo.extensionConfiguration && (clientInfo.extensionConfiguration?.accessToken || ide === CodyIDE.Web)
    ? this.handleConfigChanges(clientInfo.extensionConfiguration, {
        ide,
        forceAuthentication: true
      })
    : Promise.resolve()
```

In handleConfigChanges make sure we pass config IDE
```ts
 private async handleConfigChanges(
        config: ExtensionConfiguration,
        // Pass ide param here in order to specify it in validateAndStoreCredentials
        params?: { forceAuthentication: boolean, ide?: CodyIDE }
    ): Promise<void> {
     // ... 
      await authProvider.validateAndStoreCredentials({
        configuration: { agentIDE: params?.ide, customHeaders: config.customHeaders },
        // ...
        }, always-store')
        // ...
    }
```

agent/src/agent.ts
```diff
@@ -1505,24 +1505,10 @@ export class Agent extends MessageHandler implements ExtensionClient {
                         // functionality), we return true to always triggger the callback.
                         true,
                 })
-                // await new Promise<void>(resolve => setTimeout(resolve, 3_000))
-                // TODO(#56621): JetBrains: persistent chat history:
-                // This is a temporary workaround to ensure that a new chat panel is created and properly initialized after the auth change.
-                this.webPanels.panels.clear()
```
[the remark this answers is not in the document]
sqs: JetBrains webview code is now the web app and auth changes are handled in a more standard way, so I thought this would be safe to remove.

agent/src/cli/command-chat.ts
```diff
@@ -148,6 +149,9 @@ export async function chatAction(options: ChatOptions): Promise<number> {
         return 1
     }
 
+    // TODO(sqs)#observe: fix this so it doesn't need to sleep
+    await sleep(750)
```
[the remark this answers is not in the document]
sqs: I removed this

agent/src/agent.ts
sqs: Thanks! Applying now. `validateCredentials` still fetches the user info and cody-enabled status on Cody Web btw.

vovakulikov: No problems, we also should run handleConfigChange in the agent constructor for Cody Web even if we don't have token (This commit edb254a9bb3f3b34f6bb07c782cc4c6eba5e3328 is missing it I think)

agent/src/agent.ts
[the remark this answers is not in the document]
dominiccooney: I am not sure why we needed this, but it is ancient. But `webPanels.panels` are the "agentic" ones, not the "native" ones, so it doesn't matter any more.

Now that "native" webviews have stuck, we should probably remove the "agentic" panels and add a structured chat API that emacs, neovim and cli/agent could use.

vovakulikov:
@sqs I don't want to block this PR, but I also don't know if you want to do anything extra here; if you can provide some todos progress, it would be useful (I also can help with something). From my side, it looks like
- [ ] Fix failing e2e tests 
- [ ] Fix Cody Web handle config change run (see the last comment https://github.com/sourcegraph/cody/commit/edb254a9bb3f3b34f6bb07c782cc4c6eba5e3328)
- [ ] Fix `this.featureCodyExperimentalOneBox.value.last` problem in ChatController (this comment https://github.com/sourcegraph/cody/pull/5221#issuecomment-2361815664)

After this, we probably would be ready to merge this. I did a round of manual testing and didn't find anything suspicious

sqs:
Re-posting @dominiccooney's comment https://github.com/sourcegraph/cody/pull/5637#discussion_r1767003860 here:

<blockquote>
This API seems prone to abuse.

The use case is legitimate, let me suggest something like:

```
localstorage.batchChangeEvent(() => {
  localStorage.set('foo', 'bar')
  localStorage.set('baz', 'quux')
})
```

You can probably think of a better name.

Within the scope of the callback (handle reentrancy) you should suppress change events and fire them when done.

This means callers can't suppress fire events unilaterally, only batch them. And their batches are limited to the same microtask...if they have async work to do, they have to prepare it first. AND it means you won't have a half-written state, interrupted by async, with someone else doing an async write and causing half written state to be observed. AND you can handle being robust to execptions etc. without forgetting to fire the event.
</blockquote>

vovakulikov:
@sqs (or anyone else here) have you seen context retriever errors like this on your machine 
Screenshot 2024-09-23 at 10 59 24

sqs:
> @sqs (or anyone else here) have you seen context retriever errors like this on your machine Screenshot 2024-09-23 at 10 59 24

Weird, no, never seen that. I don't even know how it's reading those. Is it possible you have some agent unit test env vars set when you are running `pnpm -C web dev`?

vovakulikov:
It looks like something happened with context fetching but this problem is on the server side (I thought that it's this PR problem) 

I can confirm that this is the problem on main branch as well (something probably happened with context file cache on the instance)

vovakulikov (approved):
Ok, I've checked Cody Web demo and sg/sg linked version and I haven't noticed any problems (besides problems that are listed in this PR description)

abeatrix (commented):
Found some potential bugs that I think can be fixed in follow-ups:
- [ ] Sign-out button does not work when logged in as DotCom Cody Free user. Here is the Loom  https://www.loom.com/share/09df6c9c85244c3ca8ddc6a99b2d0d39
- [ ] The current files will show up as "token limit exceeded" after reloading VS Cody until you switch model. Loom: https://www.loom.com/share/1dbc51aab40e4e72ac025c4e25af694f

image

Signed into Dotcom with Cody Free account:
- [x] Default Models are showing up correctly
- [x] Configured Models are showing up correctly and are updated on change
- [x] Chat is working (able to submit with response)
- [x] Chat command is working (able to submit with response)
- [x] Edit command is working (able to submit with response)
- [ ] BUG: Sign-out button does not work

image

Signed into Dotcom with Cody Pro account:
- [x] Default Models are showing up correctly
- [x] Configured Models are showing up correctly and are updated on change
- [x] Chat is working (able to submit with response)
- [x] Chat command is working (able to submit with response)
- [x] Edit command is working (able to submit with response)
- [x] Sign out button works
- [ ] BUG: the current file will show up as "token limit exceeded" after reloading VS Code:
image

S2: Unable to test due to model configuration issue?

image

Updated: It works if I reload VS Code and I am logged in to S2:
- [x] Server-side configured Models are showing up correctly
- [ ] BUG: Cannot select a different model?
- [x] Editor-side Configured Models are showing up correctly and are updated on change
- [x] Chat is working (able to submit with response)
- [x] Chat command is working (able to submit with response)
- [x] Edit command is working (able to submit with response)
- [x] Sign out button works

image

image

abeatrix (approved):

kalanchan:
Feedback: Tested with Cody Pro (dotcom), and Enterprise (S2)

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -519,9 +502,9 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
                     break
                 }
                 if (message.authKind === 'signin' && message.endpoint && message.value) {
-                    await authProvider.auth({
-                        endpoint: message.endpoint,
-                        token: message.value,
+                    await localStorage.saveEndpointAndToken({
+                        serverEndpoint: message.endpoint,
+                        accessToken: message.value,
```
abeatrix, line 507:
```suggestion
                    await authProvider.validateAndStoreCredentials(
                        {
                            serverEndpoint: message.endpoint,
                            accessToken: message.value,
                        },
                        'store-if-valid'
                    )
```

Should we use `validateAndStoreCredentials` here when user tries to sign in via the UI?

kalanchan:
Tested with Cody Pro (dotcom), and Enterprise (S2), @sqs my experience were _slightly_ different than your known issues:

1. Switching from s2 to dotcom will not pickup new instance's model names and will sometimes retain the old instance's feature flags as well? 

example - Switching from S2 to Dotcom, seeing weird behaviour where the model list is pulled correctly but calls to the completions endpoint is using the wrong model name
CleanShot 2024-09-23 at 15 22 32@2x

CleanShot 2024-09-23 at 15 44 51@2x


2. Autocomplete doesn't work immediately when switching from dotcom to s2, I tried multiple times and each time seems to be a bit different everytime before the provider gets picked up. Usually >60s. Switching to dotcom works immediately

lib/shared/src/models/sync.ts
```diff
@@ -0,0 +1,439 @@
...
+                        waitlist_o1: config.clientState.waitlist_o1,
+                    },
+                }) satisfies PickResolvedConfiguration<{
+                    configuration: 'providerLimitPrompt' | 'customHeaders' | 'devModels'
+                    auth: true
+                    clientState: 'waitlist_o1'
+                }>
+        ),
+        distinctUntilChanged()
+    )
```
abeatrix, line 114:
```suggestion
    const relevantConfig = combineLatest([
        resolvedConfig.pipe(
            map(
                config =>
                    ({
                        configuration: {
                            customHeaders: config.configuration.customHeaders,
                            providerLimitPrompt: config.configuration.providerLimitPrompt,
                            devModels: config.configuration.devModels,
                        },
                        auth: config.auth,
                        clientState: {
                            waitlist_o1: config.clientState.waitlist_o1,
                        },
                    }) satisfies PickResolvedConfiguration<{
                        configuration: 'providerLimitPrompt' | 'customHeaders' | 'devModels'
                        auth: true
                        clientState: 'waitlist_o1'
                    }>
            )
        ),
        authStatus.pipe(map(status => status.endpoint)),
    ]).pipe(
        distinctUntilChanged(
            (prev, curr) => JSON.stringify(prev[0]) === JSON.stringify(curr[0]) && prev[1] === curr[1]
        ),
        map(([config, endpoint]) => ({ ...config, endpoint }))
    )
```

Not sure if this is the right way to do it, but including the `authStatus.endpoint` in the distinctUntilChanged check for the relevantConfig observable fixed the issue where models are not resynced when switching from dotcom to S2 for me.

abeatrix: never mine it doesn't work after i reloaded my extension host.

dominiccooney:
@kalanchan I'm seeing that cross-account model choice issue too. Picking a different model gets it back in a good state for me (in JetBrains.) I don't think we should block landing this on that issue.

I don't repro the problem with autocomplete switching between dotcom and s2.

Status update from JetBrains: digging into an issue where there's no context for any chats, with this error:

```
2024-09-24 11:00:25,755 [  28826]   WARN - #com.sourcegraph.cody.agent.CodyAgent - Uncaught exception: Error: Cannot set new context used for bot message
2024-09-24 11:00:25,755 [  28826]   WARN - #com.sourcegraph.cody.agent.CodyAgent -     at ChatModel.setLastMessageContext (C:\Users\DominicCooney\dev\cody\vscode\src\chat\chat-view\ChatModel.ts:67:19)
2024-09-24 11:00:25,755 [  28826]   WARN - #com.sourcegraph.cody.agent.CodyAgent -     at ChatController.handleSearchIntent (C:\Users\DominicCooney\dev\cody\vscode\src\chat\chat-view\ChatController.ts:867:24)
2024-09-24 11:00:25,755 [  28826]   WARN - #com.sourcegraph.cody.agent.CodyAgent -     at <anonymous> (C:\Users\DominicCooney\dev\cody\vscode\src\chat\chat-view\ChatController.ts:786:43)
2024-09-24 11:00:25,755 [  28826]   WARN - #com.sourcegraph.cody.agent.CodyAgent -     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

vscode/src/chat/chat-view/ChatController.ts
sqs: In this case, no, since I think the desired behavior is for those creds to be stored no matter what, even if the user is (for example) temporarily offline and the creds can't be validated. I could see it going either way.

valerybugakov (approved):
Tested autocomplete with DotCom, sg02, sg04, and demo instances. Tried changing local settings, switching accounts, and generating completions. No noticeable issues so far!

sqs:
I believe I fixed the switch-accounts issues in the latest push (https://sourcegraph.com/github.com/sourcegraph/cody/-/commit/7d8816). I'm going to get breakfast then can test, but if anyone is around and can start as well, that'd be great. Post what you see here. Thanks!

valerybugakov (commented):
It might be unrelated to reactivity changes. 

1. DotCom login.
2. Start a new chat.
3. Switch account to sg04.
4. Send `hello` with the default context to repro the error on the screenshot. 

Screenshot 2024-09-24 at 14 17 56

valerybugakov (commented):
It might be unrelated to reactivity changes.

Minor UX moment: when I reload my Cody-authenticated VS Code instance, I see a login sidebar for a few seconds before the new chat UI appears. It would be nice to avoid this layout thrashing.

Minor UX moment: the `New Chat` button doesn't focus on the chat panel if clicked (nothing happens).

Screenshot 2024-09-24 at 14 40 08

valerybugakov (commented):
It's possible to get Cody's data into an inconsistent state. I switched between sg04 and DotCom multiple times. `o1-preview` was the selected chat model on DotCom. On the sg04 account, the dropdown correctly shows available models, but the underlying logic tries to use `openai/cody-chat-preview-001` to send the chat message, which results in the error:

Screenshot 2024-09-24 at 14 42 09

```js
extensionHostProcess.js:155 Error: Request to https://sg04.sourcegraphcloud.com/.api/completions/stream?client-name=vscode&client-version=1.34.3 failed with 400 Bad Request: unsupported chat model "openai/cody-chat-preview-001" (default "fireworks::v1::llama-v3p1-70b-instruct")

    at IncomingMessage.<anonymous> (/Users/val/Desktop/sourcegraph-root/cody/vscode/src/completions/nodeClient.ts:179:33)
```

---------------------------------------------------

UPD: I can somewhat consistently repro this:

1. Select `o1-preview` on the DotCom account. 
2. Select `mixtral 8x7b` on the sg04 account. 
3. Send the chat message from the sg04 account.
4. Switch to DotCom.
5. Send a new chat message
6. `Request: the requested chat model is not available ("mixtral-8x7b-instruct", onProTier=true, hasFeatureFlags=true)`

I also did the same thing with account switching during the chat-response streaming. Try that if you can't repro the issue using the steps above.

valerybugakov (commented):
Autocomplete is still stable and has no new issues. Tested on DotCom, sg04, sg02, and demo with account switching.

PriNova:
@sqs I want to add to the list of fixes that an existing chat history is completely empty after switching.

PR #5120, opened 2024-08-06
Flake: Lockfile fixes

RXminuS:
The lockfile was causing some flakiness on certain platforms due to a race condition. This solves the issue

Fixes [CODY-3161]

## Test plan
Existing test should stop showing up as failure in other tests/PRs

PR #5025, opened 2024-07-26
E2Ev2: MitM configurable delay functions

RXminuS:
Add a configurable delay function to the Man-in-the-Middle proxy middleware. This allows for controlling network timings for proxied requests in test functions.

Also introduce a default floor delay of 10ms for all requests so that recorded / mocked requests avoid flake. This used to happen when requests as they were recorded had reasonable timings whereas replayed requests had their timings set to 0.

Finally we now also keep the original timings in our Polly recordings and instead prepare for more configurable playback speed. This is in anticipation of upcoming "report an issue" features that would require playback at the original speed.

## Test plan
CI

PR #5024, opened 2024-07-26
E2Ev2: Split Fixtures

RXminuS:
This PR restructures and modularizes the E2Ev2 fixture. The new structure aims to make the fixtures more manageable and easier to understand by separating the logic into different files:

1. **mitmProxy.ts**: Sets up a proxy to intercept and manipulate network requests during tests.
2. **options.ts**: Manages option schemas and validation for worker and test contexts.
3. **polly.ts**: Integrates the Polly library to record and replay HTTP interactions.
4. **vscode.ts**: Manages the setup and teardown of the VSCode server used in tests.
5. **kitchensink.ts**: Temporary placeholder fixture that holds all remaining fixture aspects.

These partial fixtures are then joined in `index.ts`

## Test plan
CI

PR #4993, opened 2024-07-24
E2Ev2: Allow for extension debugger

RXminuS:
Before there was no good way of attaching a debugger to the Cody extension as it ran in the VSCode code-host. This made it difficult to debug outside of running the special VSCode "Run Extension" command.

Now we allow supplying a specific debugger port to the extension host that gets started as part of the test fixture. This then allows you to easily connect a debugger to the extension as part of the test.

This not only makes it easier to debug issues with flaky tests but also opens the door for faster iteration as you can now simply record a few UI interactions an play them in a loop as you edit and re-debug the code.

https://github.com/user-attachments/assets/1d7ff51e-ed19-4ede-9056-10774227880a

## Test Plan

No code outside of e2ev2 is affected. No additional testing required

PR #5228, opened 2024-08-18
Build: minor tweaks

RXminuS:
- Remove the forced reload upon initialization as it was causing test flake
- Update keywords to include `cody` to see if it fixes marketplace rankings
- Exclude everything but explicit BOM from bundled `visx` . This was causing potential security issues where temporary or `.env` files were included in the bundle.
- Ensure global `window` is kept undefined in node process. For some reason `window` would be defined as `vscode.window` causing issues for `tree-sitter-js` which uses a `window !== undefined` check to see if it's running in browser or not.

## Test plan
- Manually verified that a bundled `visx` still functions


RXminuS:
* **#5026** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5026?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5025** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5025?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5024** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5024?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4999** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/4999?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5003** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5003?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4992** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/4992?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4993** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/4993?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5120** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5120?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#5228** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

abeatrix:
For the marketplace ranking, the only way to improve ranking is to change our extension name to "Cody" because it ranks matches by Display Name first, and then for extension that has the same Display Name, they are ranked by their description in alphabetical order.
So we'll need to change the name to Cody, with the description that starts with "A" 😔

RXminuS:
> For the marketplace ranking, the only way to improve ranking is to change our extension name to "Cody" because it ranks matches by Display Name first, and then for extension that has the same Display Name, they are ranked by their description in alphabetical order.

I don't think that's true entirely. Reading the discussions on the topic it's very confusing, but I noticed that we rank on certain keywords like `kotlin` even though that's not in the description or name so they aren't entirely discarded.

abeatrix:
> I don't think that's true entirely. Reading the discussions on the topic it's very confusing, but I noticed that we rank on certain keywords like `kotlin` even though that's not in the description or name so they aren't entirely discarded.
> 
> 
> 
> What's more if you combine ANY keyword with `cody` we're constantly top ranked. So my hope was that by having both the keyword and the name we would rank as expected.
> 
> 
> 
> CleanShot 2024-08-18 at 19 32 54@2x

Ah I think I misunderstood your intention, my bad! I thought you wanted to improve Cody's ranking when people search for just "Cody" (which I believe will rank the exact name match first). 

(I'm afk so can't check the package build for the ignore file change,  but can do it tomorrow if no one gets to it first! 😀 )

RXminuS:
@abeatrix No I am trying to bump the ranking on just "cody" but I don't think the assumption that it's only based on name is correct. For instance, this is what you get when searching for AI...even though there's an exact match
CleanShot 2024-08-19 at 11 44 44@2x

abeatrix:
> @abeatrix No I am trying to bump the ranking on just "cody" but I don't think the assumption that it's only based on name is correct. For instance, this is what you get when searching for AI...even though there's an exact match
> 
> CleanShot 2024-08-19 at 11 44 44@2x
> 
> 

I just looked at my notes from the old ticket and you're right, it doesn't solely depends on the display name.

I got sidetracked by the visual studio issue today so haven't tested the vscode ignore file still (sorry about that). Will do it first thing tmr morning!

vscode/package.json
```diff
@@ -55,29 +55,40 @@
...
-  "categories": ["Programming Languages", "Machine Learning", "Snippets", "Education"],
+  "categories": ["AI", "Chat", "Programming Languages", "Machine Learning", "Snippets", "Education"],
   "keywords": [
-    "ai",
-    "openai",
-    "anthropic",
-    "assistant",
-    "chatbot",
+    "cody",
+    "codey",
```
abeatrix, line 61: i wonder how much does the order of keywords matter in ranking.

Right now we are ranked number 1 in "coding assistant" & ai chatbot:
image
image

Number 5 in AI assistant:
image

Just for our record to keep track if this change would cause other regressions 😁

abeatrix (changes requested):
Packaing from main:
```
 DONE  Packaged: /cody/vscode/cody-ai-1.30.3.vsix (637 files, 27.35 MB)
```

Packaging from this branch:

```
 DONE  Packaged: /cody/vscode/cody-ai-1.30.3.vsix (32 files, 15.47MB)
```

It looks like it's missing the tree sitters files though:

image

But the package size has reduced by a lot still after adding those files back, nicely done 🎉 

```
DONE  Packaged: /cody/vscode/cody-ai-1.30.3.vsix (60 files, 19.52 MB)
```

I think once those files are added back, it should be ok?

RE the keywords change:

I'm not sure if this change would help since I noticed there is another Cody extension with `cody` keywords attached but is still getting ranked in alphabetical order based, but I'm also not sure how it actually works BTS so I'm open to giving this a try (as long as it doesn't cause regressions 😃 )

image
image

vscode/package.json
RXminuS: Looking at the results here the ranking does seem much more blended with rating & download counts. Apart from Fitten Code but I think it's because they have `AI Assistant` in their name, but that goes to show that we at least have a more fair chance of ranking higher with more downloads (e.g. > 1M)

abeatrix:
@RXminuS it looks like the issue where all the tree-sitter files are still missing in the package. Is it ok to exclude them? cc @valerybugakov 

image

RXminuS:
Nope just from the resources folder...did I forget to push? 🤦‍♂️

RXminuS:
@abeatrix ok now it works...who knew that pushing your code has such a big impact 😅

abeatrix (approved):
Nice!

PR #5232, opened 2024-08-18
bundle size improvements

sqs:
- only import 37 common langs' highlighting support. Saves ~1MB on the bundle size, at the expense of losing syntax-highlighting support for languages like: isbl, mathematica, gcl, 1c, sqf, maxima, pgsql, x86asm, mel, arduino.
- import web worker in separate file
- add vite-bundle-analyzer

## Test plan

CI

### DIFF
--- a/web/lib/agent/agent.client.ts
+++ b/web/lib/agent/agent.client.ts
@@ -15,7 +15,7 @@ import {
 // cody repository
 
 // @ts-ignore
-import AgentWorker from './agent.worker.ts?worker&inline'
+import AgentWorker from './agent.worker.ts?worker'
 
 // TODO(sqs): dedupe with agentClient.ts in [experimental Cody CLI](https://github.com/sourcegraph/cody/pull/3418)
 export interface AgentClient {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -934,6 +934,9 @@ importers:
       util:
         specifier: ^0.12.5
         version: 0.12.5
+      vite-bundle-analyzer:
+        specifier: ^0.10.5
+        version: 0.10.5
       vite-plugin-svgr:
         specifier: ^4.2.0
         version: 4.2.0(typescript@5.5.4)(vite@5.4.0)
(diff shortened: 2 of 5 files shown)


beyang (approved):

web/lib/agent/agent.client.ts
```diff
@@ -15,7 +15,7 @@ import {
 // cody repository
 
 // @ts-ignore
-import AgentWorker from './agent.worker.ts?worker&inline'
+import AgentWorker from './agent.worker.ts?worker'
```
vovakulikov, line 18: @sqs I don't think common worker import would work in the Sourcegraph Web client. The problem there is that esbuild (and vite as well) can't resolve web-worker if it's published as npm package, in this case the bundle will be something like 
```js
new Worker(
    "/agent.worker-B6dSYrrx.js",
    {
      type: "module",
      name: options == null ? void 0 : options.name
    }
  );
  ```
  
later when esbuild try to resolve it in sourcegraph web it won't find it file simply because resolution there doesn't work there for Worker constructors (I think it's a common problem for all worker as lib package, we had simular thing for other workers from extensions API, I mean custom worker plugin where we also just inline worker code in the chunk to avoid the problem with resolution) 
  
I tried to solve it manually by pointing bundler to the exact file for the agent worker, but the problem was that later, all async imports code within the web worker file also will have the same resolution problem. 

cc @thenamankumar

vovakulikov: The problem is actually not only in this PR but also in the previous one, where we introduced async imports of tokenizers; the full way of testing changes with imports in the case of cody web usually requires building and linking the Cody-web package and checking it in the source graph client locally.

vovakulikov: I wrote a document with common problems in Cody Web here https://docs.google.com/document/d/1EPT0VBfDnnJWEFlC09885nxaEaPH8ClF1leZDO3o8pI/edit 

The previous PR I'm talking about is https://github.com/sourcegraph/cody/pull/5231

vovakulikov: I will try to fix it in a way that lazy loaded chunks also would work since this improvements are not only for Cody Web but for vscode Cody as well and I don't want to make vscode cody slower that it can be because of Cody Web limitation. Will post progress here

PR #5236, opened 2024-08-19
Send `copy` extension message notification to the clients on copy button

mkondratek:
## Test plan
- to be tested together with the jetbrains part


pkukielka:
It won't be needed

PR #5237, opened 2024-08-19
use the full width of the Cody chat viewport

sqs:
Many people have asked for Cody chat to use the full width in the viewport. This is especially important now that we often show chat in a narrow sidebar instead of an editor panel.

Example request: https://x.com/riipandi/status/1825414693860856206 (this was just now, there have been many others)

This is a partial impl of this Figma: https://www.figma.com/design/f078wFMKsIOaEwj7Iwj5xy/%E2%AC%9C%EF%B8%8F--Unified-Cody?node-id=4157-66156&t=UY3VI8o9wD53jilb-0.

image

Fixes https://linear.app/sourcegraph/issue/SRCH-930/chat-should-use-full-viewport-width

## Test plan

Use Cody chat on a narrow screen.

## Changelog

The chat UI now uses the full width of its viewport, which is helpful when using chat in a narrow editor sidebar.

### DIFF
--- a/vscode/webviews/chat/cells/messageCell/human/HumanMessageCell.tsx
+++ b/vscode/webviews/chat/cells/messageCell/human/HumanMessageCell.tsx
@@ -67,14 +67,14 @@ export const HumanMessageCell: FunctionComponent<{
 
         return (
             <BaseMessageCell
-                speaker="human"
                 speakerIcon={
                     <UserAvatar
                         user={userInfo.user}
                         size={MESSAGE_CELL_AVATAR_SIZE}
-                        className="tw-mt-[2px]"
+                        sourcegraphGradientBorder={true}
                     />
                 }
+                speakerTitle={userInfo.user.displayName ?? userInfo.user.username}
                 content={
                     <HumanMessageEditor
                         userInfo={userInfo}
--- a/lib/prompt-editor/src/PromptEditor.module.css
+++ b/lib/prompt-editor/src/PromptEditor.module.css
@@ -1,6 +1,6 @@
 :root {
-    --prompt-editor-padding-y: 0.4375rem;
-    --prompt-editor-padding-x: 0.5rem;
+    --prompt-editor-padding-y: 8px;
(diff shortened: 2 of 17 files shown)


vscode/webviews/chat/cells/messageCell/human/HumanMessageCell.tsx
```diff
@@ -67,14 +67,14 @@ export const HumanMessageCell: FunctionComponent<{
 
         return (
             <BaseMessageCell
-                speaker="human"
                 speakerIcon={
                     <UserAvatar
                         user={userInfo.user}
                         size={MESSAGE_CELL_AVATAR_SIZE}
-                        className="tw-mt-[2px]"
+                        sourcegraphGradientBorder={true}
```
taiyab, line 74: Is the gradient border only applied to Pro/Enterprise users?

I think we only need to apply this in two places:
- Here
- Profile section in account menu

Anywhere else would probably be a bit overkill and introduce too much visual noise.

taiyab (commented):
This looks great Quinn! Design-wise: 👍 

Checked it out locally in VSCode. Spacing/margin changes feel good (and seem to match the density of the rest of the app). We might have to tweak things for Cody Web (@vovakulikov) to ensure it doesn't look too weird there, but let's drop it in and see first.

Also, just a minor comment about the shiny avatar border and where we ought to apply it vs. not.

vscode/webviews/chat/cells/messageCell/human/HumanMessageCell.tsx
sqs: It's also applied to Cody Free users. I think that is fine, no?

The account menu is just an OS alert right now: 
image

taiyab: It's something that reflects that you're a Pro/Enterprise user.

> The account menu is just an OS alert right now:

👍

taiyab (approved):

PR #5238, opened 2024-08-19
Add missing method stub in vscode shim

pkukielka:
Fixes CODY-3233

## Changes

`registerHoverProvider` and hovers in general are currently used for supercompletions, which are disabled by default.
Still, missing method in vsc shim caused initialization error, which in turn failed OpenCtx providers initialization.
Adding an unimplemented method fixes the issue.

## Test plan

Run IJ with this branch, and make sure OpenCtx providers are present and working.


mkondratek (approved):
👍

PR #5241, opened 2024-08-19
Revert "added(build): commands for building only webviews"

mkondratek:
Reverts sourcegraph/cody#5213

This PR introduced some changes to `webviewNativeConfig`. We are not yet prepared to support them. They seem to break the UI in JetBrains. 

## Test plan
- Green CI


jamesmcnamara:
@mkondratek Can you describe your problems? These changes were made to enable eclipse, so ideally we can simply fix it so JetBrains can get the old behavior instead of reverting it.

mkondratek:
false alarm - I am sorry for bothering you 🙇

PR #5243, opened 2024-08-19
Remove CodebaseStatusProvider

beyang:
Removes `CodebaseStatusProvider`, which previously was responsible for monitoring the availability of enhanced context for each codebase that was available. The enhanced context UI no longer exists, which means this class is no longer necessary (and indeed was no longer used in most places). Removing it allows us to remove the overhead imposed on local context providers to provide the status of indexes.

The new @-mention-based UI has as different way of determining the default repository chip. Currently it does not appear that it properly handles multiple workspace folders. I.e., if I have 2 workspace folders open, the @-mention completer does not allow me to tag the second folder. We should fix this, but this is a current issue, so this PR does not regress behavior. Users can also tag as many remote repositories when connected to an enterprise instance.

We add a `WorkspaceReposMonitor` class that's responsible for observing changes to workspace folders and keeping track of whether the repositories in the current workspace are public. This is used only by telemetry logic. See the tests in `vscode/src/repository/repo-metadata-from-git-api.test.ts` for the behavior (we err on the side of being conservative).

Removal of `CodebaseStatusProvider` allows us to remove a lot of dead and rather complex code, which should help significantly with unifying the context logic on the client side (this was the original motivation for doing this).

## Test plan

See the unit test in the PR. There should be no user-noticeable changes here.


vscode/src/context/remote-search.ts
```diff
@@ -43,43 +38,14 @@ export class RemoteSearch implements ContextStatusProvider {
...
-        })
-    }
-
-    // #endregion
-
     // Removes a manually included repository.
     public removeRepo(repoId: string): void {
         if (this.reposManual.delete(repoId)) {
-            this.statusChangedEmitter.fire(this)
+            // this.statusChangedEmitter.fire(this)
```
janhartman, line 48: Why not delete the lines you commented out everywhere in this file?

vscode/src/repository/repo-metadata-from-git-api.ts
```diff
@@ -1,3 +1,141 @@
...
+        metadatas.filter(m => m).map(m => m as RepoMetaData)
+    )
+}
+
+/**
+ * Checks if all of the workspace folders correspond to a public repository.
+ * A workspace folder is considered public if it has at least one public remote.
+ * If all workspace folders are public, return the public repository metadata for each folder.
+ */
+export async function _getRepoMetadataIfPublic(
```
janhartman, line 93: (nitpick, I'm not great at TS but still): why is this underscore-prefixed? I had the impression you'd do this for private members.

janhartman (approved):
Great work, love the cleanup!

vscode/src/context/remote-search.ts
beyang: oops missed that, thanks for flagging!

sqs (approved):

vscode/src/repository/repo-metadata-from-git-api.ts
beyang: it should effectively be private, but it's public, so that the unit test can access it

lib/prompt-editor/src/PromptEditor.tsx
```diff
@@ -148,6 +148,8 @@ export const PromptEditor: FunctionComponent<Props> = ({
                 })
             },
             setInitialContextMentions(items: ContextItem[]) {
+                console.log('# setInitialContextMentions', items)
```
thenamankumar, line 151: remove log?

vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -60,6 +61,8 @@ export function startClientStateBroadcaster({
             items.push(item)
         }
 
+        console.log('######### broadcasting client state', items)
```
thenamankumar, line 64: remove log

dominiccooney:
I have not read this is detail, but it would be better to wait until JetBrains ships a good release with web-based chat UI. The classic Swing UI *does* rely on codebase status provider.

If you land this before then, you're committing JetBrains to shipping WebUI, or JetBrains will have to fork sourcegraph/cody before it is removed, or try reverting this on a branch, or whatever. Any of that could get messy if there's time pressure involved.

CC @taylorsperry FYI.

sqs:
I'd love to see this merged soon. I have run into several bugs and inconsistencies that are not caused by this, but that to solve would entail code cleanups of this and the stuff it touches. https://github.com/sourcegraph/cody/pull/5221 is one example. This is just one data point, but I wanted to share this data point.

PR #5244, opened 2024-08-19
Agent: fix issue with clearing chat history

abeatrix:
The current logic to clear chat history using the "clear-all-no-confirm" as the ChatID exists for Cody Web, but the current implementation does not work because right now, we check for `if (chatID)` and exist early if true, which will always be true when the client passes `clear-all-no-confirm` as the chatID here.

The changes fix an issue where the "clear-all-no-confirm" chat ID was being treated as a single chat ID instead of clearing all chat records. The updated implementation correctly handles the "clear-all-no-confirm" case that would clear all chat history without requiring user confirmation.

## Test plan

All CI tests should be green. 

1. In VS Code, go to chat history and click Clear Chat History in the History Tab
2. Verify a window pop-up shows up and asks you for confirmation
image
3. Start this build in another client (e.g. Eclipse or JetBrains)
4. Ask a question to create chat history
5. go to chat history and click Clear Chat History in the History Tab
6. Verify the history has been cleared without needing a confirmation

### DIFF
--- a/vscode/src/chat/chat-view/ChatsController.ts
+++ b/vscode/src/chat/chat-view/ChatsController.ts
@@ -373,24 +373,13 @@ export class ChatsController implements vscode.Disposable {
     }
 
     private async clearHistory(chatID?: string): Promise<void> {
-        const clearAllNoConfirm = chatID === 'clear-all-no-confirm'
+        // The chat ID for client to pass in to clear all chats without showing window pop-up for confirmation.
+        const ClearWithoutConfirmID = 'clear-all-no-confirm'
+        const isClearAll = !chatID || chatID === ClearWithoutConfirmID
+        const authStatus = this.options.authProvider.getAuthStatus()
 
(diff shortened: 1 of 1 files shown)


vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -373,24 +373,13 @@ export class ChatsController implements vscode.Disposable {
     }
 
     private async clearHistory(chatID?: string): Promise<void> {
-        const clearAllNoConfirm = chatID === 'clear-all-no-confirm'
-
-        const authProvider = this.options.authProvider
-        const authStatus = authProvider.getAuthStatus()
-
-        // delete single chat
-        if (chatID) {
```
abeatrix, line 382: this is the issue where this condition will always returns true when chatID = 'clear-all-no-confirm'. We want this to fall into the next condition instead

PR #5246, opened 2024-08-20
adding deepseek direct route and prompt caching experiment

hitesh-1997:
## Context
Adding the changes for experiment with:
1. Prompt Caching: Uses the pre-computed KV cache for common context
2. Direct routing to fireworks model: Skips the fireworks middle and directly queries the on-demand deployment

## Dependent PR
1. Sourcegraph backend: https://github.com/sourcegraph/sourcegraph/pull/26
2. Infra: https://github.com/sourcegraph/infrastructure/pull/6266

## Test plan
Manual Testing:
1. Testing that we are hitting direct route end point for deepseek


vscode/src/completions/providers/fireworks.ts
```diff
@@ -590,6 +605,16 @@ class FireworksProvider extends Provider {
                         lineNumberDependentCompletionParams,
                     }).timeoutMs.toString()
                 )
+                if (self.shouldEnableDirectRoute) {
+                    // Add the following headers to the request to enable the direct route.
+                    if (modelIdentifier) {
+                        headers.set('X-Sourcegraph-Model-Identifier', modelIdentifier)
+                    }
+                    headers.set('X-Sourcegraph-Use-Direct-Route', 'true')
```
valerybugakov, line 613: Do we want to do the same for the non-fast-path Fireworks HTTP client? 

https://github.com/sourcegraph/cody/blob/main/vscode/src/completions/client.ts#L93

vscode/src/completions/providers/fireworks.ts
```diff
@@ -590,6 +605,16 @@ class FireworksProvider extends Provider {
                         lineNumberDependentCompletionParams,
                     }).timeoutMs.toString()
                 )
+                if (self.shouldEnableDirectRoute) {
+                    // Add the following headers to the request to enable the direct route.
```
valerybugakov, line 609: Are there docs related to the direct route on the Fireworks website? It would be helpful to link it here in the comment.

vscode/src/completions/providers/fireworks.ts
```diff
@@ -255,6 +261,13 @@ class FireworksProvider extends Provider {
         }
     }
 
+    private checkIfDirectRouteShouldBeEnabled(): boolean {
+        if (this.model === DEEPSEEK_CODER_V2_LITE_BASE_DIRECT_ROUTE) {
+            return true
+        }
+        return false
+    }
```
valerybugakov, line 269:
```suggestion
    private checkIfDirectRouteShouldBeEnabled(): boolean {
        return this.model === DEEPSEEK_CODER_V2_LITE_BASE_DIRECT_ROUTE
    }
```

vscode/src/completions/providers/fireworks.ts
```diff
@@ -95,6 +95,7 @@ export const DEEPSEEK_CODER_7B = 'deepseek-coder-7b'
 // Huggingface link (https://huggingface.co/deepseek-ai/DeepSeek-Coder-V2-Lite-Base)
 export const DEEPSEEK_CODER_V2_LITE_BASE = 'deepseek-coder-v2-lite-base'
 
+export const DEEPSEEK_CODER_V2_LITE_BASE_DIRECT_ROUTE = 'deepseek-coder-v2-lite-base-direct-route'
 // Context window experiments with DeepSeek Model
 export const DEEPSEEK_CODER_V2_LITE_BASE_WINDOW_4096 = 'deepseek-coder-v2-lite-base-context-4096'
```
valerybugakov, line 100: Do we still need this const?

valerybugakov (approved):

vscode/src/completions/providers/fireworks.ts
```diff
@@ -590,6 +605,16 @@ class FireworksProvider extends Provider {
                         lineNumberDependentCompletionParams,
                     }).timeoutMs.toString()
                 )
+                if (self.shouldEnableDirectRoute) {
+                    // Add the following headers to the request to enable the direct route.
+                    if (modelIdentifier) {
+                        headers.set('X-Sourcegraph-Model-Identifier', modelIdentifier)
```
valerybugakov, line 611: Could you elaborate on why we can't use the model identifier from the request body?

vscode/src/completions/providers/fireworks.ts
hitesh-1997: @valerybugakov I want to do an A/B experiment on this, just to check the latency reduction and make sure setup is working e2e correctly without errors.
Once we conclude, we can enable it on the non-fast-path :)

vscode/src/completions/providers/fireworks.ts
hitesh-1997: Great suggestion !!
Let me check with fireworks folks

vscode/src/completions/providers/fireworks.ts
hitesh-1997: I can raise a seperate PR for cleanups

vscode/src/completions/providers/fireworks.ts
hitesh-1997: I added the details on backend PR for now: https://github.com/sourcegraph/sourcegraph/pull/26/files#diff-24b0cde94fd6bb1a10fe9bfb55f97d11c1d03f172a9c0a7b34e580596af69b33R179 
It would make more sense for client to know these reasons, I will pull in some doc from there to client.

vscode/src/completions/providers/fireworks.ts
valerybugakov: Sounds good!

PR #5250, opened 2024-08-20
Smart Apply: Enable by default

umpox:
## Description

Enables the "smart apply" feature by default, removing the feature flag.

Chat Eval: https://github.com/sourcegraph/cody-leaderboard/pull/16

Looker Dashboard: https://sourcegraph.looker.com/looks/2016

## Test plan

Test Smart Apply on:
- [x] PLG
- [x] Enterprise (with different model variants)
- [X] JetBrains (with WebView enabled)
   - Note: Further testing will be done by the JetBrains team when the WebView is fully implemented

### DIFF
--- a/lib/shared/src/prompt/prompt-mixin.ts
+++ b/lib/shared/src/prompt/prompt-mixin.ts
@@ -30,18 +30,12 @@ export class PromptMixin {
      * Prepends all mixins to `humanMessage`. Modifies and returns `humanMessage`.
      * Add hedging prevention prompt to specific models who need this.
      */
-    public static mixInto(
-        humanMessage: ChatMessage,
-        modelID: string,
-        options?: { experimentalSmartApplyEnabled?: boolean }
-    ): ChatMessage {
+    public static mixInto(humanMessage: ChatMessage, modelID: string): ChatMessage {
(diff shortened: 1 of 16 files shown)


lib/shared/src/prompt/prompt-mixin.ts
```diff
@@ -6,7 +6,7 @@ import { PromptString, ps } from './prompt-string'
  * Used so that we can parse file paths to support applying code directly to files
  * from chat.
  */
-const CODEBLOCK_PREMAMBLE = ps`When generating fenced code blocks in Markdown, ensure you include the full file path in the tag. The structure should be \`\`\`language:path/to/file\n\`\`\``
+const CODEBLOCK_PREMAMBLE = ps`When generating fenced code blocks in Markdown, ensure you include the full file path in the tag. The structure should be \`\`\`language:path/to/file\n\`\`\`. You should only do this when generating a code block, the user does not need to be made aware of this in any other way.`
```
umpox, line 9: Small tweak to the prompt from running this [through some chat evaluations](https://github.com/sourcegraph/cody-leaderboard/pull/16)

I noticed some cases where the LLM would reference the file paths outside of the Markdown block (e.g. when listing changes to multiple files it is making in the response). This update resolves that as far as I can tell

PR #5254, opened 2024-08-20
Chat: enable RPC message logging with CODY_LOG_WEBVIEW_RPC_MESSAGES

abeatrix:
CONTEXT: https://sourcegraph.slack.com/archives/C073U3KE5DK/p1724102442322939

Replace  logging of RPC messages to the console with logDebug by default for RPC messages to avoid strout being picked up by the JSON-RPC on the agent/client side as response. 

Logging for RPC message is now only enabled when `CODY_DEBUG` environment variable is set to `'true'`.

This change allows developers to debug and troubleshoot RPC-related issues during development and testing when the environment variable is set, while avoiding unnecessary logging in production environments.

## Test plan

Build Visual Studio with this branch to confirm this fixes the issue where the RPC message logs are being picked up as response and returned a formatting error.

## Changelog

only enable rpc logging in vs code during test or dev mode.

### DIFF
--- a/lib/shared/src/misc/rpc/rpc.ts
+++ b/lib/shared/src/misc/rpc/rpc.ts
@@ -1,4 +1,5 @@
 import { Observable } from 'observable-fns'
+import { logDebug } from '../..'
 import type { WebviewToExtensionAPI } from './webviewAPI'
 
 export interface GenericVSCodeWrapper<TWebviewMessage, TExtensionMessage> {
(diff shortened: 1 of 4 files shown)


olafurpg (commented):
In JetBrains, we pipe stderr into the shared `idea.log`. I can imagine that the raw webview protocol is very verbose so this is only something that we should enable for local development, never in production.

vscode/src/main.ts
```diff
@@ -583,6 +584,9 @@ async function registerTestCommands(
     disposables: vscode.Disposable[]
 ): Promise<void> {
     await vscode.commands.executeCommand('setContext', 'cody.devOrTest', true)
+    if (!isRunningInsideAgent()) {
+        enableLogRPCMessages(true)
```
olafurpg, line 588: I think it's better to gate this with an environment variable that we only use for development purposes

olafurpg: See CODY_AGENT_TRACE_PATH for a similar example

lib/shared/src/misc/rpc/rpc.ts
```diff
@@ -341,10 +341,18 @@ export function addMessageListenersForExtensionAPI(
     }
 }
 
-const LOG_RPC_MESSAGES = true
```
olafurpg, line 344: This should never have been enabled by default. We need to be very conservative with writing to stderr, we should use `logDebug` instead.

abeatrix: Cool thanks for the advices, we are just in a call discussing the same thing! Will make the change in a bit, thank you!

lib/shared/src/misc/rpc/rpc.ts
```diff
@@ -341,10 +342,14 @@ export function addMessageListenersForExtensionAPI(
...
+const LOG_RPC_MESSAGES = process.env.CODY_DEBUG === 'true'
```
abeatrix, line 345:
```suggestion
const LOG_RPC_MESSAGES = process.env.CODY_DEBUG === 'true'
```

@olafurpg Is it ok to use CODY_DEBUG or do you think this should have a separated environment variable?

olafurpg: I recommend using a more specific name, something like `CODY_LOG_WEBVIEW_RPC_MESSAGES`. We should generally avoid reusing environment variables for different use-cases.

abeatrix: Makes sense! Will keep that in mind, thanks Olaf <3

olafurpg (approved):

tomaszgolebiowski (approved):

PR #5255, opened 2024-08-20
Cody Web: Fixes regressions since 0.4.0 (the last published version)

vovakulikov:
This PR addresses a few regressions since the last time we did a release of Cody Web (0.4.0), most of the fixes are trivial and come from the fact of very high speed of changing in VSCode codebase (after Cody Web GA I will guard cody web package with proper e2e tests)

- Now Cody Web renders tabs UI (small regression in recent commits about editor mode)
- Fixes create new chat command
- Adds additional selector to override styles for at mentioned portal menu
- Fixes problem when you can't remove initial context items due to a problem with the lexical node key
- No longer renders EDIT-like snippets actions  

## Test plan
- Run Cody Web demo and check that core chat functionality is intact

### DIFF
--- a/vscode/webviews/tabs/TabsBar.tsx
+++ b/vscode/webviews/tabs/TabsBar.tsx
@@ -135,9 +135,11 @@ export const TabsBar: React.FC<TabsBarProps> = ({ currentView, setView, IDE, onD
                                 ),
                                 Icon: MessageSquarePlusIcon,
                                 command:
-                                    webviewType === 'sidebar'
-                                        ? 'cody.chat.newPanel'
-                                        : 'cody.chat.newEditorPanel',
+                                    IDE === CodyIDE.Web
+                                        ? 'cody.chat.new'
+                                        : webviewType === 'sidebar'
+                                          ? 'cody.chat.newPanel'
+                                          : 'cody.chat.newEditorPanel',
                             },
                             IDE !== CodyIDE.Web
                                 ? {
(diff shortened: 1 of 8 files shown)


vscode/webviews/tabs/TabsBar.tsx
```diff
@@ -135,9 +135,11 @@ export const TabsBar: React.FC<TabsBarProps> = ({ currentView, setView, IDE, onD
...
                                 Icon: MessageSquarePlusIcon,
                                 command:
-                                    webviewType === 'sidebar'
-                                        ? 'cody.chat.newPanel'
-                                        : 'cody.chat.newEditorPanel',
+                                    IDE === CodyIDE.Web
+                                        ? 'cody.chat.new'
+                                        : webviewType === 'sidebar'
+                                          ? 'cody.chat.newPanel'
+                                          : 'cody.chat.newEditorPanel',
```
abeatrix, line 142: `cody.chat.newPanel` should work the same as `cody.chat.new` for Web i think?

abeatrix (approved):
We have a new agent config that web can use for having only a single chat view: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/jsonrpc/agent-protocol.ts?L633

After setting this config, it will always create webview with sidebar webviewType that would include the tabs bar and use a single chat view, easier than using IDE name to gate thing :D

thenamankumar (approved):

vovakulikov:
Thanks for the heads up @abeatrix, my problem with `webviewNativeConfig` is that this requires to use of webviewAPI and therefore specifies some other settings which I'm not sure about for Cody Web like `cspSource` `webviewBundleServingPrefix`

PR #5256, opened 2024-08-20
Smart Apply: Route new file logic through edit command

umpox:
## Description

This PR routes the "new file" logic through the edit command, meaning:
- Applying state works correctly
- Accept/Reject buttons are shown and work
- Decorations are shown
- Telemetry is correctly recorded

### Before

- Stuck on Applying
- No decorations
- No Accept/Reject

image

### After

image

## Test plan

1. Create a smart apply edit with a prompt like "Create a Heading component in a new file"
2. Click Apply
3. Expect that the file is created, "Accept" and "Reject" are shown

### DIFF
--- a/vscode/src/edit/manager.ts
+++ b/vscode/src/edit/manager.ts
@@ -240,9 +240,52 @@ export class EditManager implements vscode.Disposable {
             return
         }
 
+        const model = configuration.model || modelsService.getDefaultEditModel()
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
Verified!

https://github.com/user-attachments/assets/fba0d737-1d20-4835-b1aa-7fdf5db33475

PR #5261, opened 2024-08-20
Update ModelSelectField.tsx order

chenkc805:
## Test plan

tested locally


abeatrix (approved):
maybe a question for prime team: do we need to update the tags in https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/schema/site.schema.json?L3405-3407 for server-side models config

chenkc805:
Thanks @abeatrix , created a PR here to address: https://github.com/sourcegraph/sourcegraph/pull/71

CC: @slimsag

chenkc805:
cleaning this up here: https://github.com/sourcegraph/cody/pull/5264

PR #5264, opened 2024-08-20
reverting back to accuracy instead of power to not break sg/sg backend

chenkc805:
sorry last one @abeatrix ! 

relevant discussion here: https://github.com/sourcegraph/sourcegraph/pull/71

## Test plan
tested locally


abeatrix (approved):
I THINK it would work if you keep both the Power tag AND the Accuracy tag, and in the UI for Accuracy, use the same ModelUIGroup as Power :D

chenkc805:
Haha thanks @abeatrix . That feels a bit like too much complexity and I'd like to keep it simple. I'll move forward with this one!

PR #5267, opened 2024-08-21
Model Provider: rename `model.model` to `model.id`

valerybugakov:
Renaming `model.model` to `model.id` used by the `ModelsService` to make it easier to reason about the code. AFAIU, this is a client-only field, so it's safe to refactor without backend changes.

## Test plan

CI

## Changelog

No functional changes

### DIFF
--- a/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/Model.kt
+++ b/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/Model.kt
@@ -2,7 +2,7 @@
 package com.sourcegraph.cody.agent.protocol_generated;
 
 data class Model(
-  val model: String,
+  val id: String,
   val usage: List<ModelUsage>,
   val contextWindow: ModelContextWindow,
   val clientSideConfig: ClientSideConfig? = null,
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -92,112 +92,6 @@ export async function createProviderConfigFromVSCodeConfig(
     }
 }
 
-export async function createProviderConfig(
-    config: ConfigurationWithAccessToken,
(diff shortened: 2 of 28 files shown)


vscode/src/completions/providers/create-provider.ts
```diff
@@ -92,112 +92,6 @@ export async function createProviderConfigFromVSCodeConfig(
     }
 }
 
-export async function createProviderConfig(
-    config: ConfigurationWithAccessToken,
-    client: CodeCompletionsClient,
-    authStatus: AuthStatus
-): Promise<ProviderConfig | null> {
```
slimsag, line 99: I assume this function just moved with no major code change

valerybugakov: Yes

slimsag (approved):
Seems fine.

For what it's worth, in the backend we call this a 'legacyMRef'

The backend refers to models using a 'model ref' which is the `provider::version::model` syntax used in the site config. The `ModelService` has this information internally and ideally we'd refactor clients to pass that around instead.

But this totally works for now yep.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/Model.kt
```diff
@@ -2,7 +2,7 @@
 package com.sourcegraph.cody.agent.protocol_generated;
 
 data class Model(
-  val model: String,
+  val id: String,
   val usage: List<ModelUsage>,
   val contextWindow: ModelContextWindow,
   val clientSideConfig: ClientSideConfig? = null,
```
umpox, line 8: cc @sourcegraph/jetbrains-plugin 

I think we might need to update some client code for this? 🤔

valerybugakov: Yep, Dominic already caught it yesterday 🙂

PR #5268, opened 2024-08-21
Autocomplete: remove Anthropic fallback

valerybugakov:
- Removing the Anthropic autocomplete provider fallback. If the instance does not return a `completionModel`, we fail with an error message asking users to update their instance site-config.
- Refactored a small part of the `createProviderConfig` code with no functional changes. I'll have multiple follow-up updates to this part. Splitting it into numerous smaller PRs to make it easier to review.
- Discussed in [the Slack thread](https://sourcegraph.slack.com/archives/C075G57B5PH/p1724211970457949?thread_ts=1724202783.269599&cid=C075G57B5PH).

## Test plan

CI

### DIFF
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -36,7 +36,7 @@ import { createProviderConfig as createUnstableOpenAIProviderConfig } from './un
 interface CreateConfigHelperParams {
     client: CodeCompletionsClient
     authStatus: AuthStatus
-    model: string | undefined | null
+    model: string | undefined
     provider: string
     config: ConfigurationWithAccessToken
 }
@@ -143,12 +143,6 @@ async function resolveFIMModelExperimentFromFeatureFlags(): ReturnType<
     }
     // Extra free traffic - redirect to the current production model which could be different than control
     return { provider: 'fireworks', model: DEEPSEEK_CODER_V2_LITE_BASE }
-}
-
-function resolveConfigFromVSCodeConfig(configuredProvider: string | null) {
-    if (configuredProvider) {
-        return { provider: configuredProvider }
-    }
 }
 
 interface ProviderConfigFromFeatureFlags {
@@ -289,7 +283,7 @@ export async function createProviderConfig(
         return createProviderHelper({
             client,
             authStatus,
-            model: config.autocompleteAdvancedModel,
+            model: config.autocompleteAdvancedModel || undefined,
             provider: config.autocompleteAdvancedProvider,
             config,
         })
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -33,13 +33,19 @@ import { createProviderConfig as createOpenAICompatibleProviderConfig } from './
 import type { ProviderConfig } from './provider'
 import { createProviderConfig as createUnstableOpenAIProviderConfig } from './unstable-openai'
 
-export async function createProviderConfigFromVSCodeConfig(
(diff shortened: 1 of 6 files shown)


vscode/src/completions/providers/create-provider.ts
```diff
@@ -217,8 +221,17 @@ function getAutocompleteModelInfo(authStatus: AuthStatus): AutocompleteModelInfo
...
+    // Fallback to the Fireworks autocomplete provider only for PLG users.
+    if (authStatus.isDotCom) {
+        return {
+            provider: 'fireworks',
+        }
+    }
+
+    // Fail with error for enterprises.
+    return new Error(
+        'Failed to get autocomplete model info. Please configure the `completionModel` using site configuration.'
```
aramaraju, line 233: @slimsag Will this ever happen on the Enterprise case? As wouldn't the Instance always send some config as part of `completionModel`? I suspect there will need to be some clean up on the server side as well

aramaraju (approved):
Thank you! LGTM, left one comment for Stephen to make sure we handle the fallback to error for Enterprises properly

vscode/src/completions/providers/create-provider.ts
valerybugakov: > After Aug 7th release, we at least return a default model in that case, e.g. the legacy CodyLLMConfiguration graphql query would return this for the config above.

From [this message](https://sourcegraph.slack.com/archives/C075G57B5PH/p1724212337636359?thread_ts=1724202783.269599&cid=C075G57B5PH). 

Btw, if you have a relevant documentation page in mind we can link to in this error, let me know.

slimsag: As I noted, this can happen when connecting a new Cody client to an old Sourcegraph instance.

On Aug 7th and later Sourcegraph instances, this is not a possible condition.

aramaraju: Missed that note, great! More reason to kill this then :) Resolving this comment

@valerybugakov Not necessary to redirect to a link at the moment, I don't think it's very actionable for an enterprise dev to see a link, especially given the transient nature of this problem

vscode/src/completions/providers/create-provider.ts
```diff
@@ -217,8 +221,17 @@ function getAutocompleteModelInfo(authStatus: AuthStatus): AutocompleteModelInfo
...
     }
 
-    // If no provider info specified, return undefined to fall back to default provider
-    return
+    // Fallback to the Fireworks autocomplete provider only for PLG users.
+    if (authStatus.isDotCom) {
+        return {
+            provider: 'fireworks',
+        }
+    }
```
slimsag, line 229: I'm not sure why we need this fallback case for dotcom?

[This query](https://sourcegraph.com/api/console#%7B%22query%22%3A%22%7B%5Cn%20%20%20site%20%7B%5Cn%20%20%20%20codyLLMConfiguration%20%7B%5Cn%20%20%20%20%20%20completionModel%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%7D) always returns a valid completion model when ran against sourcegraph.com, `"completionModel": "fireworks/starcoder"` - and this information is the source of `if (authStatus.configOverwrites?.provider) {` above so I don't think this if statement can ever run

slimsag:
lgtm aside from my comment, feel free to address my comment later if you like

vscode/src/completions/providers/create-provider.ts
valerybugakov: Even better. I'll drop this fallback too.

Issue #5209, opened 2024-08-15
bug: When edit code only processes one line of code, undo will cause problems


HangLu123:
### Version

latest

### Describe the bug

![image](https://github.com/user-attachments/assets/bcf22e4d-4fa5-4fc0-bc8d-fbe78010da69)

When edit code only processes one line of code, undo will cause problems
![image](https://github.com/user-attachments/assets/e1fcb94d-d5b3-439d-9d75-cf481d4a867d)
![image](https://github.com/user-attachments/assets/dc4a56e2-2fe1-4fdd-bae9-b5d6351abd01)

### Expected behavior

right behavior

HangLu123:
The inserted content is placed on the same line as the next line, and executing undo will not restor

umpox:
Should be fixed by: https://github.com/sourcegraph/cody/pull/5259

PR #5271, opened 2024-08-21
Edit: Fix single line edits

umpox:
Closes https://linear.app/sourcegraph/issue/CODY-3236/bug-with-edit-command-inserts-newline-incorrectly-and-removes-trailing

Closes https://github.com/sourcegraph/cody/issues/5209

## Description

Fixes issues where we would not include the final new line character of a line, so the edit would not be correct. Mainly happened for single line edits, where we'd often squash the change into the next line.

Original fix by @arafatkatze, I just re-implemented it and did some more testing

## Test plan

Test with:
- [x] 1 single line edit
- [x] 2 lines edit
- [x] Making sure the decorations aren't off with an extra gray line
- [x] End of file apply works
- [x] End of file edit works
- [x] Works on JetBrains, with all of the above scenarios

### DIFF
--- a/agent/src/edit.test.ts
+++ b/agent/src/edit.test.ts
@@ -78,7 +78,8 @@ describe('Edit', () => {
           	messages,
           	setChatID,
           	isLoading,
-          }: ChatColumnProps) {	useEffect(() => {
+          }: ChatColumnProps) {
+          	useEffect(() => {
           		if (!isLoading) {
           			setChatID(messages[0].chatID);
           		}
--- a/agent/recordings/customCommandsClient_509552979/recording.har.yaml
+++ b/agent/recordings/customCommandsClient_509552979/recording.har.yaml
@@ -493,11 +493,11 @@ log:
         send: 0
         ssl: -1
         wait: 0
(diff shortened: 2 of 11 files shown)


agent/src/edit.test.ts
```diff
@@ -77,7 +77,8 @@ describe('Edit', () => {
           	messages,
           	setChatID,
           	isLoading,
-          }: ChatColumnProps) {	useEffect(() => {
+          }: ChatColumnProps) {
+          	useEffect(() => {
           		if (!isLoading) {
```
umpox, line 82: Fix!

umpox:
@sourcegraph/jetbrains-plugin I tested this on JetBrains and it seems to work fine, and it also fixes the single line edit issue on JetBrains too. LMK if you see any issues!

Note: When I first tested it, I had an issue submitting an edit due to an incorrect model. I think that's due to https://github.com/sourcegraph/cody/pull/5267, I rebased my PR to exclude this commit and all works fine. So think we need some separate changes to accommodate that PR (maybe someone is already looking at this, if so ignore me)

arafatkatze (approved):

PR #5272, opened 2024-08-21
Fix model name mapping for "fireworks/deepseek-coder-v2-lite-base"

taras-yemets:
This PR changes the model name mapping for "fireworks/deepseek-coder-v2-lite-base" so that the backend can digest it regardless of the mode it's running in (dotcom or enterprise).

See https://github.com/sourcegraph/sourcegraph/pull/135 for more details.

## Test plan
Tested manually. See the test plan defined in https://github.com/sourcegraph/sourcegraph/pull/135.


valerybugakov (approved):
Unblocking the PR because the change itself makes sense. 

Should we release the Cody Gateway before merging this PR? If this change is released with the nightly VS Code extension build it will break autocomplete for nightly DotCom users if Cody Gateway does not allow the virtual string added in https://github.com/sourcegraph/sourcegraph/pull/135. Correct me if I'm wrong!

taras-yemets:
> Should we release the Cody Gateway before merging this PR? If this change is released with the nightly VS Code extension build it will break autocomplete for nightly DotCom users if Cody Gateway does not allow the virtual string added in [sourcegraph/sourcegraph#135](https://github.com/sourcegraph/sourcegraph/pull/135). Correct me if I'm wrong!

Cody Gateway already supports this virtual string ([diff](https://github.com/sourcegraph/sourcegraph/pull/4/files#diff-24b0cde94fd6bb1a10fe9bfb55f97d11c1d03f172a9c0a7b34e580596af69b33R254-R262)). It has already been deployed to the Cody Gateway dev and there's a PR to update the prod instance with this change https://github.com/sourcegraph/infrastructure/pull/6269. 

https://github.com/sourcegraph/sourcegraph/pull/135 adds support to the Sourcegraph backend. Indeed, this change should be deployed on dotcom before VS Code nightly is released so as not to break the autocomplete for cases when the fast path is disabled.

taras-yemets:
@valerybugakov, I wonder if we should also update the mapping for the other occurrences of `"accounts/sourcegraph/models/deepseek-coder-v2-lite-base"` (I mean changing them to `"fireworks/deepseek-coder-v2-lite-base"`) so that they work with Sourcegraph backend when the fast path is disabled, WDYT?
https://github.com/sourcegraph/cody/blob/04bad813b4bcf304460da9545ca65a75dbeb1c20/vscode/src/completions/providers/fireworks.ts#L79-L84
Although the Sourcegraph backend is aware of the "old" model name, it's missing the provider prefix ("fireworks/" in this case).  
Is dotcom without the fast path a real use case?

taras-yemets:
The other occurrences of "accounts/sourcegraph/models/deepseek-coder-v2-lite-base" have different max context windows
https://github.com/sourcegraph/cody/blob/04bad813b4bcf304460da9545ca65a75dbeb1c20/vscode/src/completions/providers/fireworks.ts#L108-L118

On the Gateway side, we use `2048` ([code](https://github.com/sourcegraph/sourcegraph/blob/d35e76a3f3a052fdd7593005c9e9b08c88f21c9f/cmd/cody-gateway-config/dotcom_models.go#L197-L206)). I wonder if it's the correct value, WDYT @valerybugakov? Are the bigger context windows used only for experiments currently?

aramaraju:
@taras-yemets Once we merge these changes into the client, does that mean we officially support DeepSeek v2 for Enterprises?

taras-yemets:
> @taras-yemets Once we merge these changes into the client, does that mean we officially support DeepSeek v2 for Enterprises? 

@aramaraju, yes, when we release a new client version containing this change.

aramaraju:
@taras-yemets 

> > @taras-yemets Once we merge these changes into the client, does that mean we officially support DeepSeek v2 for Enterprises?
> 
> @aramaraju, yes, when we release a new client version containing this change.

Let's do this then! Thank you @taras-yemets and @valerybugakov ! 

@kalanchan Can I bother you to track this on the client releases side for both VSCode and JBs? I will kick off the GTM stuff for this :)

valerybugakov:
> Are the bigger context windows used only for experiments currently?

I think the answer is yes. Tagging @hitesh-1997 to confirm because he leads this experiment.

hitesh-1997:
> > Are the bigger context windows used only for experiments currently?
> 
> I think the answer is yes. Tagging @hitesh-1997 to confirm because he leads this experiment.

Yes @taras-yemets 
The bigger context window is only experimental

PR #5273, opened 2024-08-21
Smart Apply: Hash the ID used for the FixupTask

umpox:
## Description

Fixes an issue where "Open Diff" would not work for the smart apply diffs. This is because the ID used was too long, we were using the entire contents of the codeblock for the ID. We should just hash this and it'll be a lot shorter.

## Test plan

1. Create smart apply edits
2. Click "Open Diff"


kalanchan (approved):
looks good!

PR #5276, opened 2024-08-21
added(telemetry): separate field for IDE name and server identification

jamesmcnamara:
Pre 5.6 versions of Sourcegraph would reject requests from unknown clients. This was resolved in https://github.com/sourcegraph/sourcegraph/pull/63855, but for enterprise customers who have not updated, we still need to identify ourselves as JetBrains. This meant that all telemetry events would falsely identify themselves as other JetBrains.

This introduces a field we hope to delete that allows a client to identify as itself, EXCEPT for with the server. Once everyone is up to 5.6, we can remove this field.

Additionally, we check if the client is stateful for login events, so we can report reliable installed and login events for agent users.

## Test plan
Tested in Eclipse

## Changelog
Allows identifying as Eclipse (or other previously unrecognized IDEs) in telemetry events.


agent/src/AgentWorkspaceConfiguration.ts
```diff
@@ -100,6 +100,14 @@ export class AgentWorkspaceConfiguration implements vscode.WorkspaceConfiguratio
                 return this.clientInfo()?.ideVersion
             case 'cody.advanced.agent.extension.version':
                 return this.clientInfo()?.version
+            case 'cody.advanced.agent.capabilities.storage':
```
olafurpg, line 103: Unrelated diff

olafurpg (approved):
LGTM for the legacyNameForServerIdentification-related parts 👍🏻 This PR has some unrelated diffs that we should remove before merging

agent/src/AgentWorkspaceConfiguration.ts
jamesmcnamara: This is so that we can find out if the agent has stateful storage so we log the [correct event](https://github.com/sourcegraph/cody/pull/5276/files#diff-52551733cd7342d42b308be539209b59728d5dfaaf992d1f389c36d65f3e7262R120).

olafurpg: My bad

PR #5277, opened 2024-08-21
Improve repository mention search

vovakulikov:
Closes CODY-3370
Fixes CODY-3372

This PR brings search mentions results closer to the Sourcegraph search experience (search API feels a bit faster and we also include fuzzy ranking similar to what we do in Sourcegraph search-box UI) 

## Test plan
- Check that all mention flows that require repositories mentions (such as repositories, files and directories) work properly

### DIFF
--- a/lib/prompt-editor/src/mentions/mentionMenu/MentionMenu.tsx
+++ b/lib/prompt-editor/src/mentions/mentionMenu/MentionMenu.tsx
@@ -142,7 +142,7 @@ export const MentionMenu: FunctionComponent<
                         item.mention?.data?.repoName &&
                         !item.mention?.data?.filePath) ||
                     (item.providerUri === REMOTE_DIRECTORY_PROVIDER_URI &&
-                        item.mention?.data?.repoID &&
+                        item.mention?.data?.repoName &&
                         !item.mention?.data?.directoryPath)
                 ) {
                     // Do not set the selected item as mention if it is repo item from the remote file search provider.
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -49,7 +49,7 @@ import {
     REPOSITORY_IDS_QUERY,
     REPOSITORY_ID_QUERY,
     REPOSITORY_LIST_QUERY,
-    REPOSITORY_SEARCH_QUERY,
+    REPOS_SUGGESTIONS_QUERY,
     REPO_NAME_QUERY,
     SEARCH_ATTRIBUTION_QUERY,
     VIEWER_SETTINGS_QUERY,
@@ -272,14 +272,21 @@ export interface RepoListResponse {
     }
 }
 
-export interface RepoSearchResponse {
-    repositories: {
-        nodes: { name: string; id: string; url: string }[]
-        pageInfo: {
-            endCursor: string | null
-        }
-    }
+export interface SuggestionsRepo {
+    id: string
+    name: string
+    stars: number
+    url: string
 }
+
+export interface RepoSuggestionsSearchResponse {
+    search: {
+        results: {
+            repositories: Array<SuggestionsRepo>
+        }
+    } | null
+}
+
 interface FileMatchSearchResponse {
     search: {
         results: {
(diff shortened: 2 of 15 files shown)


lib/prompt-editor/src/mentions/mentionMenu/MentionMenu.tsx
```diff
@@ -142,7 +142,7 @@ export const MentionMenu: FunctionComponent<
                         item.mention?.data?.repoName &&
                         !item.mention?.data?.filePath) ||
                     (item.providerUri === REMOTE_DIRECTORY_PROVIDER_URI &&
-                        item.mention?.data?.repoID &&
+                        item.mention?.data?.repoName &&
```
vovakulikov, line 145: @thenamankumar I think it was a mistake on main, I checked, and the directory mention object, and it never contains the `repoID` field; let me know if `repoName` doesn't work here. (on main prior to this PR when I selected repository in directory provider it just selected this repository instead of letting me select an actual directory)

vovakulikov (commented):
@thenamankumar I checked repository, file and directory providers and it work as expected but I saw that you merged directory mentions provider yesterday and it looks like it may not working correctly on dotcom, at least I got no context due to problem of CodyContext gql API (because it expects to have filePattern argument there) 

Full gql error 
```
Variable \"filePatterns\" has invalid value null.\nExpected type \"[String!]!\", found null.
```
Can you please check this in the Cody Web demo (it may look like that directory provider should exist only for non-dot com instances but I think in Cody Web it slips or maybe it's just Cody Web demo problem)

lib/prompt-editor/src/mentions/mentionMenu/MentionMenu.tsx
thenamankumar: `/Users/naman/Work/cody/vscode/src/context/openctx/currentRepositoryDirectorySearch.ts`

The directory mention object should always contain the repoID. Check the implementation of `getDirectoryMentions` function. 

CleanShot 2024-08-22 at 10 44 42@2x

thenamankumar:
@vovakulikov the directory providers are enabled only for non-web non-dotcom instances.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -864,23 +871,13 @@ export class SourcegraphGraphQLAPIClient {
...
-        first: number,
-        after?: string,
-        query?: string
-    ): Promise<RepoSearchResponse | Error> {
-        return this.fetchSourcegraphAPI<APIResponse<RepoSearchResponse>>(REPOSITORY_SEARCH_QUERY, {
-            first,
-            after: after || null,
-            query,
-        }).then(response => extractDataOrError(response, data => data))
+    public async searchRepoSuggestions(query: string): Promise<RepoSuggestionsSearchResponse | Error> {
```
thenamankumar, line 874: Can we please parameterise count and default to 10?

vscode/src/context/openctx/common/get-repository-mentions.ts
```diff
@@ -0,0 +1,67 @@
...
+ * fuzzy matching and repository additional metrics.
+ *
+ * It's a common repository mention fetchers is used in repo, file and directories
+ * providers.
+ */
+export async function getRepositoryMentions(
+    query: string,
+    providerId: string
+): Promise<ProviderMention[]> {
+    const dataOrError = await graphqlClient.searchRepoSuggestions(query)
```
thenamankumar, line 35: Do we not need to escape the query before sending it for search?

thenamankumar (changes requested):
Blocking due to https://github.com/sourcegraph/cody/pull/5277/files#r1725631553

Otherwise looks good. Added non blocking comments.

lib/prompt-editor/src/mentions/mentionMenu/MentionMenu.tsx
vovakulikov: This is the second step of the directory mention you make screenshot from, but the fix is for the first step when you resolve repository. Check the main this repository resolve mention doesn't have repoID, I think this is why you rely on repo name for remote file resolution as well.

vovakulikov: Here the handler you use for repo resolving in directories provider and it doesn't have repoID, only repoName

https://github.com/sourcegraph/cody/blob/main/vscode/src/context/openctx/remoteFileSearch.ts

vscode/src/context/openctx/common/get-repository-mentions.ts
vovakulikov: I don't think we need, the handler itself is about accepting any generic regexp query, I think regexp sanitation exists on the backend in this case (at least we don't do anything about it in the sourcegraph client implementation about this)

vovakulikov:
> @vovakulikov the directory providers are enabled only for non-web non-dotcom instances.

@thenamankumar Hmm, I think I saw that you added directories for the cody web as well. Here https://github.com/sourcegraph/cody/blob/main/vscode/src/context/openctx.ts#L152-L156

Should I fix it here? The problem that we probably still need directories for web to solve problems like this https://linear.app/sourcegraph/issue/SRCH-940/cody-doesnt-fetch-context-for-specific-folder-questions-in-cody-web

lib/prompt-editor/src/mentions/mentionMenu/MentionMenu.tsx
thenamankumar: Okay yeah this makes sense. I confused `getDirectoryMentions` with `getDirectoryItems`. Directory mentions need to provide the repoID because we need it to fetch the directory items. But we only need repo name to get the directory mentions, so getRepoMentions does not need to include the ID.

thenamankumar (approved):

thenamankumar:
> > @vovakulikov the directory providers are enabled only for non-web non-dotcom instances.
> 
> @thenamankumar Hmm, I think I saw that you added directories for the cody web as well. Here https://github.com/sourcegraph/cody/blob/main/vscode/src/context/openctx.ts#L152-L156
> 
> Should I fix it in this PR? The problem is that we probably still need directories for the web to solve problems like this https://linear.app/sourcegraph/issue/SRCH-940/cody-doesnt-fetch-context-for-specific-folder-questions-in-cody-web

Lets remove the provider from web as part of this PR for now and later after testing we can add it as part of a separate PR.

thenamankumar:
> @thenamankumar I checked repository, file and directory providers and it work as expected but I saw that you merged directory mentions provider yesterday and it looks like it may not working correctly on dotcom, at least I got no context due to problem of CodyContext gql API (because it expects to have filePattern argument there)
> 
> Full gql error
> 
> ```
> Variable \"filePatterns\" has invalid value null.\nExpected type \"[String!]!\", found null.
> ```
> 
> Can you please check this in the Cody Web demo (it may look like that directory provider should exist only for non-dot com instances but I think in Cody Web it slips or maybe it's just Cody Web demo problem)

I am testing  `Variable \"filePatterns\" has invalid value null.\nExpected type \"[String!]!\", found null` right now.

PR #5278, opened 2024-08-21
E2Ev2

RXminuS:
We've been plagued by slow and flaky e2e tests. And although Agent tests are effective in eliminating some of that flake they lack the ability to validate real extension behavior and UI states.

The new E2Ev2 framework brings together the best elements of the Agent Tests with a more Developer-Friendly E2E workflow. The initial goal of this framework is to provide the baseline tools to migrate the flaky E2Ev1 tests and re-enable them in CI.

But the ultimate goal is to make the DX fast enough to serve as the go-to way to replicate issues, interactively develop new UI features (watch mode), or fuzzy-test core features across a wide variety of contexts.

### Highlights

- significantly faster
   - all tests can run in parallel
   - all tests can run headless and only require a lightweight VSCode-Server process
   - all http is pre-recorded and played back with 0 delays (but can be controlled per request if needed)
- more readable & maintainable
   - abstract common functionality in UIX helpers
   - allows the calling of VSCode commands
   - You can invoke custom `macros` that have full access to the vscode-api. This makes it trivial to set-up testing conditions without requiring tricky to time UI state
- Flake is fully eliminated
   - all tests are 100% isolated
   - all http is recorded and can be optionally mocked for custom responses
   - timing of http responses can be controlled. This is great for being able to try things like aborting requests even thought the http is recorded
-  More dev friendly & `#velocity-wins`
   - you can configure every parameter of the frameworks including things like installed extensions, vscode version, user/workspace settings etc.
   - You can use playwright to auto-record your tests
   - You can leverage the excellent `playwright --ui` and `--watch` flags to create iterative dev workflows
   - Testing output in CI will contain full logs, network traces, interactive browser recording (much better than video as you can still use your chrome-devtools to inspect stuff)
   - Tests are structured with `steps` that make the event log more interpretable

## Test Plan

This framework should not interfere or break any existing functionality as it's not currently enabled.
- Existing tests pass
- Manually verified that builds work across platforms


RXminuS:
@keegancsmith I've merged a whole stack of branches into one because most of the early PRs had the majority of their code tweaked or removed. This now feels like a good enough starting point to at least have been able to migrate 1 flaky test :-)

vscode/e2e/utils/credentials.setup.ts
```diff
@@ -0,0 +1,18 @@
+import { test as setup } from '@playwright/test'
+import * as dotenv from 'dotenv'
+import { CREDENTIALS_ENVFILE_PATH, updateEnvFile } from './vscody/credentials-envfile'
+
+// biome-ignore lint/complexity/noBannedTypes: <explanation>
```
dominiccooney, line 5: Let's write an `<explanation>`. Ditto below.

vscode/e2e/utils/credentials.setup.ts
```diff
@@ -0,0 +1,18 @@
...
+// biome-ignore lint/correctness/noEmptyPattern: <explanation>
+setup.extend<{}>({})('credentials', async ({}) => {
+    try {
+        // NOTE: VSCode Playwright UI will abort the running test when it
+        // detects a file change. So you'll in some cases have to click the run
+        // test butotn twice. Every other case seems to work fine.
```
dominiccooney, line 11:
```suggestion
        // test button twice. Every other case seems to work fine.
```

vscode/e2e/utils/vscody/credentials-envfile.ts
```diff
@@ -0,0 +1,90 @@
...
+    return values
+}
+
+function withMissingValues(values: Map<string, string>) {
+    const withMissing = new Map(values)
+    for (const [exportedName, secretName] of tokens) {
+        if (withMissing.has(exportedName)) {
+            continue
+        }
+        const token = execSync(
```
dominiccooney, line 50: How many are missing usually? Might be nice to parallelize this.

vscode/e2e/utils/vscody/credentials-envfile.ts
```diff
@@ -0,0 +1,90 @@
...
+            `gcloud secrets versions access latest --secret ${secretName} --project cody-agent-tokens --quiet`,
+            { encoding: 'utf-8' }
+        )
+        withMissing.set(exportedName, token.trim())
+    }
+    return withMissing
+}
+/**
+ * Updates env file
```
dominiccooney, line 59: This comment adds nothing that the name doesn't already have.

If it is worth documenting, then document it. Otherwise let's drop this repetitive comment.

vscode/e2e/utils/vscody/credentials-envfile.ts
```diff
@@ -0,0 +1,90 @@
...
+    const content = dedent(`
+        # This file is automatically generated by \`scripts/export-credentials.ts\`
+
+        ${Array.from(values.entries())
+            .map(([name, value]) => `${name}=${value}`)
+            .sort()
+            .join('\n')}
+    `)
+
+    fs.writeFileSync(CREDENTIALS_ENVFILE_PATH, content)
```
dominiccooney, line 87: Tell me more about how the isolation works?

vscode/e2e/utils/vscody/extension/main.js
```diff
@@ -0,0 +1,84 @@
...
+        await flushEventStack()
+        return res
+    })
+
+    context.subscriptions.push(disposable)
+}
+
+function deactivate() {}
+
+// create a command for running macros by name. Normally you would just call the macro directly, but this is useful for testing
```
dominiccooney, line 34: ??? what does this comment relate to?

vscode/e2e/utils/vscody/extension/main.js
```diff
@@ -0,0 +1,84 @@
...
+
+function flushEventStack() {
+    // this is a sleep timer for 0 seconds, which sounds dumb
+    // the reason it's useful is because it puts a function on the BOTTOM of the javascript event stack
+    // and then we wait for it to occur
+    // this means runs all of the already-scheduled things to occur
+    // which is ideal because it makes pop ups and other events happen in a more sequential/timely order
+    return new Promise(r => setTimeout(r, 0))
```
dominiccooney, line 42: You can instead just `Promise.resolve(undefined)`, `then`-ing off that will happen at microtask timing.

dominiccooney (dismissed):
Some incremental feedback inline.

keegancsmith:
> Would love to find some time to pair and migrate a few more together?

Lets pair monday morning?

vscode/e2e/utils/vscody/extension/main.js
RXminuS: Yeah, this was originally copied from another library and then incrementally stripped until it seems only some trivially ugly snippets remain 😅 I'll tidy it up.

vscode/e2e/utils/vscody/credentials-envfile.ts
RXminuS: It really only runs once on a dev machine and then it's set. I probably should add some expiration, but performance wise it maybe adds a few seconds the first time your ran some tests.

vscode/e2e/utils/credentials.setup.ts
RXminuS: I'll instead add exclusion rules to Biome config. This is an artifact of how Playwright mandates these functions to be written as they use it to infer a dependency graph.

dominiccooney: The ignores are fine, I just think `<explanation>` is lame. Just write Playwright needs it that way etc.

vscode/e2e/utils/vscody/extension/main.js
RXminuS: Actually digging into it they're not 100% equivalent. This bumps things down to the bottom of the Event Loop, whereas Promise.resolve(undefined) only bumps it to the microtask queue tick. We atually need the event loop here.

sqs (approved):
Approving to unblock if you want

vscode/e2e/features/command-custom.test.ts
```diff
@@ -0,0 +1,9 @@
+import { fixture as test } from '../utils/vscody'
+
+test.use({
+    templateWorkspaceDir: 'test/fixtures/legacy-polyglot-template',
+})
+
+test.describe('custom command', {}, () => {
+    // this has entirely changed since the last time this test-suite was created
```
dominiccooney, line 8: This should fail with an expected failure, or have a TODO or something here to follow up on.

biome.jsonc
```diff
@@ -0,0 +1,120 @@
...
+                }
+            }
+        },
+        {
+            "include": ["vscode/e2e/utils/**/*.ts"],
+            "linter": {
+                "rules": {
+                    // These rules interfere with how Playwright needs test functions to be written.
+                    "complexity": {
+                        "noBannedTypes": "off"
```
dominiccooney, line 106: 🎉 this is great, this was a pain in the past

lib/shared/src/telemetry-v2/TelemetryRecorderProvider.ts
```diff
@@ -85,7 +85,7 @@ export class TelemetryRecorderProvider extends BaseTelemetryRecorderProvider<
             ],
             {
                 ...defaultEventRecordingOptions,
-                bufferTimeMs: 0, // disable buffering for now
+                bufferTimeMs: 0, // disable buffering for now. If this is enabled make sure that it can be disabled during tests.
```
dominiccooney, line 88: This is a bit mysterious, let's add a pointer to more details, at least the name of a test fixture that you'd need to consult.

lib/shared/src/telemetry-v2/events/internal.ts
```diff
@@ -0,0 +1,121 @@
...
+
+type ObjWithFallback<T extends Record<string, number>> = T & {
+    [fallbackValue]?: number
+}
+type KeyOfOmitFallback<T> = T extends ObjWithFallback<infer U> ? keyof U : never
+
+type HasFallback<M extends MapperInput> = M extends { [fallbackValue]: infer V }
+    ? V extends number
+        ? true
+        : false
```
dominiccooney, line 21: Do these have to be boolean literals or  can we

```
? V extends number : false
```

or simpler

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+and connect a standard Playwright browser to it. Despite what you might be
+thinking right now...
+
+> 🔑 VSCode Server + Browser ≠ VSCode Web
+
+VSCode Web runs the extension host (where Cody is executed) in the browser,
+which limits access to Node.js APIs. In contrast, VSCode Server runs actual
+for-realsies VSCode with only the UI displayed in a normal browser. This allows
+Cody to run in a proper Node.js extension host with full API access. This
+approach is functionally equivalen to running the VSCode electron application.
```
dominiccooney, line 203:
```suggestion
approach is functionally equivalent to running the VSCode Electron application.
```

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+But with significantly less overhead and the magic ability to run in headless
```
dominiccooney, line 204: Might be clearer to reverse these clauses to make it clear you're not suggesting "less magic"
```suggestion
But with the magic ability to run in headless mode and significantly less overhead. :fire:
```

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+// <div class="chat-view">
+//    <button data-testid="tab-user">Me</button>
+//    <button data-testid="tab-account">Account Settings</button>
+//    ...
+// </div>
+await page.locator('div.chat-view button').nth(2) // ❌
+await sidebar.content.getByTestId('tab-account').click() // ✅
+```
+
+However that only works if there is a single unique semantic element of that
```
dominiccooney, line 388: This caveat applies to all kinds of locators though. If you want to encourage IDs, it is probably worth relaxing this a bit and making your first scoped example use IDs to illustrate scoping with IDs.

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+and potential for timing issues.
+
+One good rule of thumb is to follow every click with a expect statement that
+validates the click was successful. This is because `expect` statements halt
+execution until the assertion is met (or times out). This makes them much more
+reliable and performant than `sleep` statements.
+
+Locators come with auto waiting and retry-ability. Auto waiting means that
+Playwright performs a range of actionability checks on the elements, such as
+ensuring the element is visible and enabled before it performs the click.
```
dominiccooney, line 438: Do you think it is worth recommending (or adding an abstraction that) checks the post-click assertion doesn't pass pre-click? That has been a source of errors I have observed. There's a natural bias towards assertions that pass and sometimes people water them down to the point that they're not actually reacting to a click.

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+
+```ts
+await cody.statusBar.click() // 1. click
+await expect(session.QuickPick.items({ hasText: 'Cody is disabled in this file' })).toBeVisible() // 2. assert success with an expect
+await cody.statusBar.filter({hasText: 'Cody is disabled'}).click() // 3. Clicks can themselves also serve as an assertion.
+```
+
+When a test fails you can always look at the trace recording to get a
```
dominiccooney, line 446: Let's link to instructions about how to do that so people don't have to scroll around or guess.

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+visualization of where the locator was pointing.
+
+It is also best to reseve clicks when part of the actual behaviour being tested.
```
dominiccooney, line 449: Be direct.

```suggestion
Limit clicks to the behaviour being tested.
```

(Nit for spelling, `reserve`, if you don't take this suggestion.)

vscode/e2e/README.md
```diff
@@ -1,6 +1,935 @@
...
+                // placeholder that the MitM server will swap out for us.
+                'cody.override.authToken': MITM_AUTH_TOKEN_PLACEHOLDER,
+                'cody.override.serverEndpoint': mitmProxy.sourcegraph.dotcom.endpoint,
+            }),
+            { workspaceDir }
+        )
+
+        await vscodeSession.start()
+        await codyExtension.waitUntilReady()
+    })
```
dominiccooney, line 607: This seems verbose given how common it will be.

dominiccooney (commented):
Some minor feedback inline. I can't review this patch, GitHub is letting me type about 1 character per second.

High level the demo looked great and I can't wait to have it. I see @sqs has already approved it, let's land it and work through the details incrementally.

Next time a more incremental approach could have been less stressful. For example we could have switched everything to headless, we could have switched networking to PitM proxying, those could have been done in sequence... We would have got the earned value of better tests while developing the rest of the framework, it would have been easier to split up the work, it would have got higher quality review attention because the review tools would handle it, etc etc. 

Should I expect this to work on Windows? I patched it and tried it...

Playwright (whether UI, command line, VSCode extension) does not seem to be enumerating the tests. The tests I am finding are legacy ones and they're opening a VSCode instance.

Other things did not work:

```
PS C:\Users\DominicCooney\dev\cody\vscode> pnpm run test:e2e e2e/example.test.ts

> cody-ai@1.34.3 test:e2e C:\Users\DominicCooney\dev\cody\vscode
> playwright install && tsc --build && node dist/tsc/test/e2e/install-deps.js && pnpm run -s build:dev:desktop && pnpm run -s test:e2e:run "e2e/example.test.ts"

[0] All files were successful downloaded, check resources/wasm directory
[0]
  dist\post-uninstall.js      745.7kb
[0]   dist\post-uninstall.js.map    1.6mb
[0]
[0] ⚡ Done in 167ms
[0]
  dist\extension.node.js      12.8mb ⚠️
[0]   dist\extension.node.js.map  21.3mb
[0]
[0] ⚡ Done in 526ms
[0] pnpm run -s _build:esbuild:desktop exited with code 0
[1] pnpm run -s _build:webviews --mode development exited with code 0
Error: No tests found.
Make sure that arguments are regular expressions matching test files.
You may need to escape symbols like "$" or "*" and quote the arguments.
Error: ENOENT: no such file or directory, scandir 'C:\Users\DominicCooney\dev\cody\playwright'
```

similarly for `--grep example`.

Could you add the Playwright extension to the recommended extensions? I installed it manually but had none of the test running that appeared in your talk. I installed Playwright but the set of tests it found were very spare:

image

I tried running one of these but it failed:

```
 1) [utils] › e2e\utils\uix.test.ts:12:9 › Sidebar selectors › It works ───────────────────────────

    Error: ENOENT: no such file or directory, mkdtemp 'c:\Users\DominicCooney\dev\cody\.test\runs\01J941Z9W9KMEP6WZY10SKD5FW\test-vscode-server-XXXXXX'

    attachment #1: trace (application/zip) ─────────────────────────────────────────────────────────
    vscode\test-results\uix-Sidebar-selectors-It-works-utils\trace.zip
    Usage:

        npx playwright show-trace vscode\test-results\uix-Sidebar-selectors-It-works-utils\trace.zip
```

vscode/e2e/README.md
RXminuS: Like the idea!

vscode/e2e/README.md
RXminuS: Agreed, I'll abstract it. 

Just wanted to wait until I'd migrated the normal auth tests to make sure I wasn't shooting myself in the foot. I've done those now...turns out I was 😅 But with that clear now I'll wrap it in a helper

PR #5281, opened 2024-08-22
Codegen: add C# support for Agent bindings

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3412/agent-protocol-bindings-for-c

This PR attempts to add C# support to codegen in order to facilitate smooth development of Cody and Agent in Visual Studio.

- Updated Codegen to convert TypeScript interfaces/types into C# for use in Visual Studio
- Updated the agent/scripts/generate-agent-kotlin-bindings.sh script to support generating bindings with different arguments for different clients
- Renamed agent/scripts/generate-agent-kotlin-bindings.sh to agent/scripts/generate-agent-bindings.sh

I tried not to touch the existing logic for Kotlin and Java in the codegen code by adding additional if-else statements for c#, but if you spot any issues with my change please let me know 😄 

Update: updated to use switch cases as requested per review

## Test plan

Run the current command for generating Agent bindings for kotlin is still working correctly with no diff:

```
 pnpm generate-agent-kotlin-bindings
```

Run the cli command to generate bindings in C#:

```
pnpm exec ts-node agent/src/cli/scip-codegen/command.ts --output "agent/bindings/csharp" --language "csharp" --kotlin-package "Cody.Core.Agent.Protocol"
```

Example output:

image

See full results @ https://github.com/sourcegraph/cody/compare/bee/codegen-csharp-result#files_bucket


dominiccooney:
@abeatrix the WebviewMessagesEnum Object-encoded, String-encoded names aren't valid identifiers because of the hyphens.

Here's the relevant part of the C# spec: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers

agent/src/cli/scip-codegen/JvmCodegen.ts
```diff
@@ -88,23 +89,31 @@ export class JvmCodegen extends BaseCodegen {
         p.line()
         if (this.language === JvmLanguage.Kotlin) {
             p.line('object ProtocolTypeAdapters {')
-        } else {
```
dominiccooney, line 91: Use

```ts
switch (this.language) {
case JvmLanguage.Kotlin:
  ...
  break;
case ...
```

etc. Switch statements were made for this.

agent/src/cli/scip-codegen/JvmCodegen.ts
```diff
@@ -23,6 +23,7 @@ interface DocumentContext {
 export enum JvmLanguage {
```
dominiccooney, line 23: Let's rename this Language or TargetLanguage or something. CSharp isn't a JvmLanguage so the name has become confusing.

dominiccooney (changes requested):
If this is the last language, probably fine.

However there's another way to do this: Have ONE switch statement that creates an "emitter", and the emitters have a method for each case (for example `startTypeDecl`, `startEnumDecl`, `writeEnumMember`, etc. ...) That way the generator code is very succinct, it just unpacks whatever those need... the name of the enum for example... and calls `emitter.startEnumDecl(name)` instead of having switch statements everywhere.

abeatrix:
> @abeatrix the WebviewMessagesEnum Object-encoded, String-encoded names aren't valid identifiers because of the hyphens.
> 
> 
> 
> Here's the relevant part of the C# spec: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#643-identifiers

This enum was the [exact same issue](https://sourcegraph.slack.com/archives/C073U3KE5DK/p1723656474638159) we ran into last week that motivated me to work on this 😂 

will work on the feedback tomorrow, thank you!

agent/scripts/generate-agent-bindings.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -eux
+INDEXER_DIR=${SCIP_TYPESCRIPT_DIR:-../scip-typescript-cody-bindings}
+
+# Use OUT_LANG if it's not empty, or fall back to "kotlin". Example:
+# OUT_LANG=csharp OUT_DIR=agent/bindings/csharp NAMESPACE=Cody.Core.Agent.Protocol pnpm generate-agent-kotlin-bindings
+OUT_LANG=${OUT_LANG:-"kotlin"}
+OUT_DIR=${OUT_DIR:-"agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated"}
+NAMESPACE=${NAMESPACE:-"com.sourcegraph.cody.agent.protocol_generated"}
```
olafurpg, line 9: I don't mind merging this, but I recommend calling the CLI tool directly with relevant flags. This is what we do for Eclipse https://github.com/sourcegraph/eclipse/blob/main/scripts/build-java-bindings.sh#L23

This bash script was already on the upper limit of complexity that I tolerate for bash scripts and I'd prefer to keep it unchanged. It's cleaner to just have a separate bash script in the VS repo.

agent/src/cli/scip-codegen/JvmCodegen.ts
olafurpg: Agreed, I thought I was generalizing by JvmLanguage but my attempt to abstract over languages has failed so far, it's simpler to just keep adding more if/else for now (I hope to clean it up eventually, but not when I'll get to it realistically)

agent/src/cli/scip-codegen/JvmCodegen.ts
olafurpg: It's subjective, I prefer if/else here and use switch only for early returns. JS scoping rules are too weird around switch statements to make them practical inside other blocks. It's a different story for switch expressions in other languages.

olafurpg (approved):
LGTM 👍🏻 non-blocking comments on bash script and `JvmLanguage`

olafurpg:
Thank you for doing this! 🙏🏻

tomaszgolebiowski:
Some comments on the generated code:
We use Json.Net as the serialization library. System.Text.Json is newer, but we have to use what the Visual Studio SDK provides.

Technically, we don't need to use `JsonPropertyName` because serializer follows the camel case naming convention. If we want to add an attribute, let's use `JsonProperty` instead:
`[JsonPropertyName("FooBar")] -> [JsonProperty(PropertyName = "FooBar")]`

Return type can't be Task<void> please use Task
`Task<Void> -> Task`

same goes to parameters
`Task<AuthStatus> extensionConfigurationstatus(Void params) -> Task<AuthStatus> extensionConfigurationstatus()`

for methods definitions please use pascal case:
`void autocompletecompletionAccepted(CompletionItemParams params); -> void AutocompleteCompletionAccepted(CompletionItemParams params);`

for enums please use 'EnumMember'
`[JsonPropertyName("workspace")] -> [EnumMember(Value="workspace")]`

I found several places where the name was incorrectly translated into C#. We need to remove special characters such as spaces and hyphens. The code should look something like this:
'string.Join("", zm.Split(' ','-').Select((x => char.ToUpper(x[0]) + x.Substring(1))));'

abeatrix:
Thanks for the reviews ya'll! I've updated the PR based on all the helpful feedback. You can find the latest results in https://github.com/sourcegraph/cody/compare/bee/codegen-csharp-result

Please let me know if you spot anything unusual @tomaszgolebiowski !

Here is an example after the change

```cs
using System.Text.Json.Serialization;

namespace Cody.Core.Agent.Protocol
{
  public class CompletionItemInfo
  {
    [JsonProperty(PropertyName = "parseErrorCount")]
    public int ParseErrorCount { get; set; }
    [JsonProperty(PropertyName = "lineTruncatedCount")]
    public int LineTruncatedCount { get; set; }
    [JsonProperty(PropertyName = "truncatedWith")]
    public string TruncatedWith { get; set; } // Oneof: tree-sitter, indentation
    [JsonProperty(PropertyName = "nodeTypes")]
    public NodeTypesParams NodeTypes { get; set; }
    [JsonProperty(PropertyName = "nodeTypesWithCompletion")]
    public NodeTypesWithCompletionParams NodeTypesWithCompletion { get; set; }
    [JsonProperty(PropertyName = "lineCount")]
    public int LineCount { get; set; }
    [JsonProperty(PropertyName = "charCount")]
    public int CharCount { get; set; }
    [JsonProperty(PropertyName = "insertText")]
    public string InsertText { get; set; }
    [JsonProperty(PropertyName = "stopReason")]
    public string StopReason { get; set; }
  }
}
```

P.S: I'll be OOO next week so will have to pick it up when I'm back if there is any issues with the new changes

abeatrix:
Merging so that we can start integrating this into Visual Studio, and will make improvements as we go.

PR #5282, opened 2024-08-22
Autocomplete: reduce `createProviderConfig` duplication

valerybugakov:
- Merged case statements from the `createProviderConfig` function with the switch statement in the `createProviderConfigHelper` function to keep one source of truth and simplify the setup.
- No functional changes.

## Test plan

CI

### DIFF
--- a/lib/shared/src/models/index.ts
+++ b/lib/shared/src/models/index.ts
@@ -350,7 +350,9 @@ export class ModelsService {
         }
         const endpoint = this.authStatus?.endpoint
         if (!endpoint) {
-            logError('ModelsService::preferences', 'No auth status set')
+            if (!process.env.VITEST) {
+                logError('ModelsService::preferences', 'No auth status set')
+            }
             return empty
         }
         // If global cache is missing, try loading from storage
--- a/vscode/src/completions/providers/create-provider.ts
+++ b/vscode/src/completions/providers/create-provider.ts
@@ -36,18 +36,31 @@ import { createProviderConfig as createUnstableOpenAIProviderConfig } from './un
 interface CreateConfigHelperParams {
     client: CodeCompletionsClient
     authStatus: AuthStatus
-    model: string | undefined
+    modelId: string | undefined
     provider: string
     config: ConfigurationWithAccessToken
+    model?: Model
 }
 
-export async function createProviderHelper(
+export async function createProviderConfigHelper(
     params: CreateConfigHelperParams
 ): Promise<ProviderConfig | null> {
-    const { client, authStatus, model, provider, config } = params
+    const { client, authStatus, modelId, model, provider, config } = params
 
     switch (provider) {
-        case 'azure-openai':
+        case 'openai': {
+            return createUnstableOpenAIProviderConfig({
+                client,
+                model: modelId,
+            })
+        }
+        case 'azure-openai': {
+            return createUnstableOpenAIProviderConfig({
+                client,
+                // Model name for azure openai provider is a deployment name. It shouldn't appear in logs.
+                model: modelId ? '' : modelId,
+            })
+        }
         case 'unstable-openai': {
             return createUnstableOpenAIProviderConfig({
                 client,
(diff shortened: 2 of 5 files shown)


lib/shared/src/models/index.ts
```diff
@@ -350,7 +350,9 @@ export class ModelsService {
         }
         const endpoint = this.authStatus?.endpoint
         if (!endpoint) {
-            logError('ModelsService::preferences', 'No auth status set')
+            if (!process.env.VITEST) {
+                logError('ModelsService::preferences', 'No auth status set')
```
valerybugakov, line 354: Uncluttering unit-tests output.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -36,18 +36,31 @@ import { createProviderConfig as createUnstableOpenAIProviderConfig } from './un
...
+            return createUnstableOpenAIProviderConfig({
+                client,
+                model: modelId,
+            })
+        }
+        case 'azure-openai': {
+            return createUnstableOpenAIProviderConfig({
+                client,
+                // Model name for azure openai provider is a deployment name. It shouldn't appear in logs.
+                model: modelId ? '' : modelId,
```
dominiccooney, line 61: Why not just hardcode this as `''`? Do you care about transmitting `null` and `undefined` here? Seems weird.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -57,34 +70,68 @@ export async function createProviderHelper(
...
                 authStatus,
                 config,
             })
         }
+        case 'openaicompatible': {
+            if (model) {
+                return createOpenAICompatibleProviderConfig({
+                    client,
+                    timeouts: config.autocompleteTimeouts,
+                    model: model,
```
dominiccooney, line 96:
```suggestion
                    model,
```

vscode/src/completions/providers/create-provider.ts
```diff
@@ -57,34 +70,68 @@ export async function createProviderHelper(
...
+                        ? authStatus.configOverwrites.completionModel
+                        : authStatus.isDotCom
+                          ? DEFAULT_PLG_ANTHROPIC_MODEL
+                          : undefined,
+            })
+        }
+        case 'google': {
+            if (authStatus.configOverwrites?.completionModel?.includes('claude')) {
+                return createAnthropicProviderConfig({
+                    client, // Model name for google provider is a deployment name. It shouldn't appear in logs.
```
dominiccooney, line 124: This comment belongs above or trailing the `model: undefined` line, not here.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -57,34 +70,68 @@ export async function createProviderHelper(
...
+                    model: undefined,
+                })
+            }
+            // Gemini models
+            return createGeminiProviderConfig({ client, model: modelId })
+        }
+        case 'gemini':
+        case 'unstable-gemini': {
+            return createGeminiProviderConfig({ client, model: modelId })
+        }
```
dominiccooney, line 134: Readability of this is not great... 'google', so Gemini, unless we said 'gemini' or 'unstable-gemini' which is also Gemini... why is this like this?

dominiccooney (approved):
Some questions inline.

Please test JetBrains, in case there's some old way it specifies autocomplete models or something.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -57,34 +70,68 @@ export async function createProviderHelper(
             const { anonymousUserID } = await localStorage.anonymousUserID()
             return createFireworksProviderConfig({
                 client,
-                model: config.autocompleteAdvancedModel ?? model ?? null,
+                model: modelId ?? null,
```
slimsag, line 73: is this a behavior change? would prevent users from using that config option, right?

vscode/src/completions/providers/create-provider.ts
valerybugakov: I'm simplifying/improving this part step by step. It's weird to me, and I want to fix this, but I also want to have many focused revertable PRs for doing that. 

In this PR, I only merge two switch statements into one, preserving all the execution paths. Will look into provider implementations in one of the follow-ups.

vscode/src/completions/providers/create-provider.ts
valerybugakov: Completely agree: https://github.com/sourcegraph/cody/pull/5282#discussion_r1726344128

vscode/src/completions/providers/create-provider.ts
valerybugakov: This case is handled here, so the removed code is redundant: https://github.com/sourcegraph/cody/blob/f9af436e1336914db28c6e580d88e70f3df4ef15/vscode/src/completions/providers/create-provider.ts#L274-L279

PR #5283, opened 2024-08-22
Autocomplete: remove starcoder2

valerybugakov:
- Removed the unused `starcoder2` support and feature flag from the Fireworks provider. 
- No functional changes.

## Test plan

CI


dominiccooney (approved):
What's the thinking with removing models.test.ts? It was testing more than just starcoder2...

slimsag (approved):

valerybugakov:
Nice catch, Dom! The manual invocation was nested under the `starcoder2` describe call, so I missed it. Will keep it around.

PR #5284, opened 2024-08-22
Autocomplete: extract the fast-path client

valerybugakov:
- Extracted the Fireworks `fast-path` client into a separate file. 
- No functional changes.

## Test plan

CI


dominiccooney (approved):
Only skimmed assuming this is minimal changes to effect the separation into a separate file.

slimsag (approved):
Awesome! I really wonder how much value this fast-path client actually gives us, though. It seems like we need a larger architecture change here.. I'll bring it up to prime team.

Because if PLG needs Client -> Gateway -> LLM provider to be reasonable..... then Enterprise users having to go Client -> Enterprise instance -> Gateway -> LLM provider must be pretty bad?

hitesh-1997 (approved):
Nice, Thanks for making the changes !!

valerybugakov:
> Awesome! I really wonder how much value this fast-path client actually gives us, though. It seems like we need a larger architecture change here.. I'll bring it up to prime team.

@slimsag, see these issues for the context:
- https://linear.app/sourcegraph/issue/CODY-2359/[decision]-should-we-do-a-fast-path-for-autocomplete-for-enterprises
- https://linear.app/sourcegraph/issue/CODY-1372/[autocomplete-latency]-implement-fast-path-support-for-enterprise

I'd love to unify this but [the last conclusion](https://linear.app/sourcegraph/issue/CODY-1372/[autocomplete-latency]-implement-fast-path-support-for-enterprise#comment-f3236904) was: "_this feels like a bad move architecturely, in conflict with our general Cody brains on the Cloud_".

slimsag:
@valerybugakov Ah, thanks for the links! The `last conclusion` you referenced seems to be 'bringing fast path to enterprise` but I meant the other way around! :D I am saying maybe we can get to a point where we _remove_ fast path alltogether and have all traffic go through sourcegraph.com instance because it is fast enough?

Anyway, I'll think about it some more and raise a discussion elsewhere

valerybugakov:
@slimsag, ah, I see! 

Technically, it would be awesome if we could do it this way without increasing latency. However, this would also increase the setup complexity for the autocomplete pipeline, making it harder to analyze latency and reason about slight deviations. Tens of milliseconds matter here. 

Currently, we struggle to analyze latency end-to-end, even without the SG instance in the middle (the distributed tracing setup still needs a lot of improvements). We would need to keep this in mind because it was a big part of the Faster Autocomplete project we worked on recently.

PR #5290, opened 2024-08-22
Smart Apply: Use system prompt to encourage code blocks being created

umpox:
## Description

This PR improves the smart apply prompt tweak so that:
- We use the system prompt when available
- We use the same desired Markdown syntax when _we_ include codebase context, which helps steer the LLM onto the right path.

I have ran an eval on this here: https://github.com/sourcegraph/cody-leaderboard/pull/16

Quite happy with the results so far

## Test plan

- [x] Tested creating code blocks on all major models

### DIFF
--- a/agent/src/__snapshots__/custom-commands.test.ts.snap
+++ b/agent/src/__snapshots__/custom-commands.test.ts.snap
@@ -9,12 +9,14 @@ exports[`Custom Commands > commands/custom, chat command, open tabs context 1`] 
 `;
 
 exports[`Custom Commands > commands/custom, edit command, edit mode 1`] = `
-"export interface Animal {
+"/* SELECTION_START */
+export interface Animal {
     name: string
     makeAnimalSound(): string
     isMammal: boolean
     logName(): void
 }
+/* SELECTION_END */
 "
 `;
 
--- a/agent/src/__snapshots__/custom-commands.test.ts.snap
+++ b/agent/src/__snapshots__/custom-commands.test.ts.snap
@@ -1,7 +1,7 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Custom Commands > commands/custom, chat command, open tabs context 1`] = `
-"Here are the names of the files you've shared so far, formatted as a bulleted list:
+"Here are the names of the files you've shared with me so far, formatted as a bulleted list:
 
 • example3.ts
 • example2.ts
--- a/agent/src/__snapshots__/custom-commands.test.ts.snap
+++ b/agent/src/__snapshots__/custom-commands.test.ts.snap
@@ -1,7 +1,7 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Custom Commands > commands/custom, chat command, open tabs context 1`] = `
-"Here are the names of the files you've shared with me so far, formatted as a bulleted list:
+"Here are the names of the files you have shared with me so far, formatted as a bulleted list:
 
 • example3.ts
 • example2.ts
@@ -9,14 +9,12 @@ exports[`Custom Commands > commands/custom, chat command, open tabs context 1`] 
 `;
 
 exports[`Custom Commands > commands/custom, edit command, edit mode 1`] = `
-"/* SELECTION_START */
-export interface Animal {
+"export interface Animal {
     name: string
     makeAnimalSound(): string
     isMammal: boolean
     logName(): void
 }
-/* SELECTION_END */
 "
 `;
 
--- a/agent/src/__snapshots__/custom-commands.test.ts.snap
+++ b/agent/src/__snapshots__/custom-commands.test.ts.snap
@@ -1,7 +1,7 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
 exports[`Custom Commands > commands/custom, chat command, open tabs context 1`] = `
-"Here are the names of the files you have shared with me so far, formatted as a bulleted list:
+"Here's a bulleted list of the filenames you've shared with me so far:
 
 • example3.ts
 • example2.ts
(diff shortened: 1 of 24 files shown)


umpox:
@abeatrix Seems just as reliable without the weird occasional behaviour in chat... Running evals here https://github.com/sourcegraph/cody-leaderboard/pull/16

agent/src/index.test.ts
```diff
@@ -874,45 +876,105 @@ describe('Agent', () => {
...
-
-                  test('Test animal implementation isMammal flag', () => {
-                      const dog = new Dog()
-                      expect(dog.isMammal).toBe(true)
-                  })
-
-                  test('Test animal implementation name property', () => {
-                      const dog = new Dog()
-                      expect(dog.name).toEqual('Dog')
+                  // src/animal.test.ts
```
abeatrix, line 885: This is added as a comment instead of markdown format we provided 👀 

(It wasn't doing it before the prompt change though so not a regression but an edge case?)

lib/shared/src/chat/preamble.ts
```diff
@@ -22,7 +22,7 @@ export function getSimplePreamble(
     preInstruction?: PromptString
 ): Message[] {
     const preamble = type === 'Chat' ? CHAT_PREAMBLE : DEFAULT_PREAMBLE
-    const intro = ps`${preamble} ${preInstruction ?? ''}`.trim()
+    const intro = ps`${preamble} ${preInstruction ? ps`\n\n${preInstruction}` : ''}`.trim()
```
abeatrix, line 25:
```suggestion
    const intro = ps`${preamble}\n\n${preInstruction.toString() ?? ''}`.trim()
```
would this work?

agent/src/__snapshots__/custom-commands.test.ts.snap
```diff
@@ -1,20 +1,22 @@
...
+"Here are the names of the files you've shared with me so far, formatted as a bulleted list:
 
 • example3.ts
 • example2.ts
 • example1.ts"
 `;
 
 exports[`Custom Commands > commands/custom, edit command, edit mode 1`] = `
-"export interface Animal {
+"/* SELECTION_START */
```
abeatrix, line 12: I think this is a regression because it shouldn't include the "selection" pointers from the file?

lib/shared/src/chat/preamble.ts
```diff
@@ -1,14 +1,28 @@
...
 import type { Message } from '../sourcegraph-api'
 
+const DEFAULT_PREAMBLE = ps`You are Cody, an AI coding assistant from Sourcegraph.`
+
+/**
+ * For chat, we add an additional preamble to encourage the model to
+ * produce code blocks that we can associate with existing file paths.
+ * We want to read these file paths to support applying code directly to files from chat.
+ */
+const CHAT_PREAMBLE = psDedent`
```
abeatrix, line 11:
```suggestion
const SMART_APPLY_PREAMBLE = ps`If your answer contains fenced code blocks in Markdown, include the full file path in the code block tag using this structure: \`\`\`$LANGUAGE:$FILEPATH\n\`\`\`.`
const CHAT_PREAMBLE = DEFAULT_PREAMBLE.join(SMART_APPLY_PREAMBLE)
```

I think we don't need to say that this is an additional rules in system prompt. We can probably update it to something like this? ( Haven't tested it though!)

abeatrix (commented):
im seeing a minor regression where it's including the file pointers  in the response.

cc @jtibshirani on the response change

agent/src/__snapshots__/custom-commands.test.ts.snap
umpox: I think this snapshot is flaky, the "SELECTION" pointer was removed in an unrelated PR: https://github.com/sourcegraph/cody/commit/00be23943b972a66bee1d4d3e8694972fdb55e68#diff-319fdb363fed27945ee845694011c34c3703e4ddc603a2f3ba0fc70f98d79ea2

I'll look at if we can just make it always exclude it

lib/shared/src/chat/preamble.ts
umpox: Happy to try anything, will run an eval on this change and we can see!

agent/src/__snapshots__/custom-commands.test.ts.snap
umpox: Not included again now

lib/shared/src/chat/preamble.ts
umpox: @abeatrix Applied this change and made a couple of more tweaks, looks pretty good!

Eval is here: https://github.com/sourcegraph/cody-leaderboard/pull/16

My comments here are still relevant: https://github.com/sourcegraph/cody-leaderboard/pull/16#issuecomment-2304581865

I think we can go ahead and merge this, we can always tweak later but let's get a patch out today as we may be both OOO next week

umpox:
Have a failing test because the demo instance is replying with

```
{
  model: 'anthropic/claude-3-opus-20240229',
  speaker: 'assistant',
  error: {
    isChatErrorGuard: 'isChatErrorGuard',
    status: 400,
    traceId: '024f1e65bdeb9024d5dd7ed41efbe4ed',
    message: 'Request to https://demo.sourcegraph.com/.api/completions/stream?client-name=enterpriseclient&client-version=v1 failed with 400 : unsupported chat model "anthropic/claude-3-opus-20240229" (default "anthropic::2023-06-01::claude-3-opus")\n',
    name: 'Error'
  }
}
```

It's likely because the demo config has changed since that test was last updated... have requested access, will skip if I can't get access before this PR is merged, as it's a broken test right now

abeatrix (approved):
Great! I haven't encountered any prompt leaks anymore, and I don't see any noticeable regression from the test results, so I think this would be a good change to patch the regression we are seeing. Thanks for fixing this!

PR #5291, opened 2024-08-22
Fix getCodyContext fetching

thenamankumar:
Call to `getCodyContext` API is broken in Cody due to the `filePatterns` argument. Passing null just fails the graphql query. 

This PR fixes it. 

## Test plan

- Open Cody and connect to s2
- Test Remote Repository & Remote Directories

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -1021,7 +1021,7 @@ export class SourcegraphGraphQLAPIClient {
                 query,
                 codeResultsCount: 15,
                 textResultsCount: 5,
-                ...(isValidVersion ? { filePatterns } : {}),
+                ...(isValidVersion ? { filePatterns: filePatterns || [] } : {}),
             },
             signal
         ).then(response =>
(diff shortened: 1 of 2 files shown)


vovakulikov (approved):
LGTM. the only concern I have is "is it safe to pass an empty patter, can it somehow change set of context results from the backend?"

I think it's fine so approve it.

thenamankumar:
> LGTM. the only concern I have is "is it safe to pass an empty patter, can it somehow change set of context results from the backend?"
> 
> I think it's fine so approve it.

skipped passing the arg altogether is empty.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1021,7 +1021,7 @@ export class SourcegraphGraphQLAPIClient {
                 query,
                 codeResultsCount: 15,
                 textResultsCount: 5,
-                ...(isValidVersion ? { filePatterns } : {}),
+                ...(isValidVersion && filePatterns?.length ? { filePatterns } : {}),
```
vovakulikov, line 1024: I think the problem with not passing the file pattern is still there in this case, no? If we take a look at gql query 

```gql
query GetCodyContext($repos: [ID!]!, $query: String!, $codeResultsCount: Int!, $textResultsCount: Int!, $filePatterns: [String!]!) {
```

FilePatterns marked as required non-null argument. I think we should fallback on valid version and provide empty filePattern always we need to make sure that backend is ok with empty filePattern and it doesn't empty result set of context files

vovakulikov (changes requested):
Blocking due this comment https://github.com/sourcegraph/cody/pull/5291/files#r1726671336

vovakulikov (approved):

PR #5293, opened 2024-08-22
Cody Web: Fix cody web file/symbol mention search

vovakulikov:
Fixes CODY-3375
Fixes CODY-3376

This PR makes symbol and file mention search respect the initial context repository, so we will look for files and symbols only within this repository. You can still select files from different repositories by selecting the "Files" category directly. 

## Test plan
- Check that you can select files in the cody web demo by typing `@<search term>` string and you see files only within sourcegraph/cody repository (which is set as a default context for the cody web demo)

### DIFF
--- a/vscode/src/editor/utils/editor-context.ts
+++ b/vscode/src/editor/utils/editor-context.ts
@@ -86,7 +86,7 @@ export async function getFileContextFiles(options: FileContextItemsOptions): Pro
             source: ContextItemSource.User,
             remoteRepositoryName: item.repository.name,
             isIgnored: contextFiltersProvider.isRepoNameIgnored(item.repository.name),
-            uri: URI.file(item.repository.name + item.file.path),
+            uri: URI.file(`${item.repository.name}/${item.file.path}`),
         }))
     }
 
@@ -192,7 +192,7 @@ export async function getSymbolContextFiles(
             item.symbols.map(symbol => ({
                 type: 'symbol',
                 remoteRepositoryName: item.repository.name,
-                uri: URI.file(item.repository.name + symbol.location.resource.path),
+                uri: URI.file(`${item.repository.name}/${symbol.location.resource.path}`),
                 isIgnored: contextFiltersProvider.isRepoNameIgnored(item.repository.name),
                 source: ContextItemSource.User,
                 symbolName: symbol.name,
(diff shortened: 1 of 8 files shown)


web/lib/components/CodyWebPanelProvider.tsx
```diff
@@ -114,6 +114,24 @@ export const CodyWebPanelProvider: FunctionComponent<PropsWithChildren<CodyWebPa
         })()
     }, [accessToken, serverEndpoint, createAgentWorker, customHeaders, telemetryClientName])
 
+    useEffect(() => {
```
vovakulikov, line 117: Move this to the useEffect to call this `'context/choose-remote-search-repo'` last, cause during chat initialization there are fired some events that reset remote search repo (new session id and restore chat event)

web/lib/components/CodyWebPanelProvider.tsx
```diff
@@ -114,6 +114,24 @@ export const CodyWebPanelProvider: FunctionComponent<PropsWithChildren<CodyWebPa
...
+        }
+
+        // Set initial context after we restore history so context won't be
+        // overridden by the previous chat session context
+        if (initialContext?.repository) {
+            void client.rpc.sendRequest('webview/receiveMessage', {
+                id: activeWebviewPanelID,
+                message: {
+                    command: 'context/choose-remote-search-repo',
+                    explicitRepos: [initialContext?.repository],
```
thenamankumar, line 129: `initialContext?.repository ? [initialContext.repository] : []`

vscode/src/editor/utils/editor-context.ts
```diff
@@ -192,7 +192,7 @@ export async function getSymbolContextFiles(
             item.symbols.map(symbol => ({
                 type: 'symbol',
                 remoteRepositoryName: item.repository.name,
-                uri: URI.file(item.repository.name + symbol.location.resource.path),
+                uri: URI.file(`${item.repository.name}/${symbol.location.resource.path}`),
```
thenamankumar, line 195: Does `URI.file` correctly handles `//` if they occur in the path?

thenamankumar (approved):

web/lib/components/CodyWebPanelProvider.tsx
vovakulikov: Just reilized that there is no need for ? operator here since above is if guard statement

kalanchan:
yeahhhhhhhh vova!!!

PR #5297, opened 2024-08-22
Cody Web: Support directory as initial context

vovakulikov:
Part of SRCH-940

See the original issue; in the Sourcegraph repo page, it's a common case when you open Cody Web chat from the directory page, so it makes sense to support directories as an initial mention for the chat. 

## Test plan
- Check that Cody Web demo resolves proper context with default directory mention (demo already has directory mention by default)

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@sourcegraph/cody-web",
-  "version": "0.6.0",
+  "version": "0.6.1",
   "description": "Cody standalone web app",
   "license": "Apache-2.0",
   "repository": {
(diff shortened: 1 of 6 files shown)


web/vite.config.ts
```diff
@@ -104,7 +104,7 @@ export default defineProjectWithDefaults(__dirname, {
         assetsDir: '.',
         reportCompressedSize: true,
         lib: {
-            formats: ['cjs'],
+            formats: ['es'],
```
vovakulikov, line 107: This change is needed to unblock the svelte Cody Web update.

web/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "main": "dist/index.js",
   "types": "dist/lib/index.d.ts",
-  "sideEffects": false,
+  "sideEffects": true,
```
thenamankumar, line 13: what is this about?

vovakulikov: This is needed for the proper bundle process on the Sourcegrpah side. When we target to es modules when we bundle/build Cody Web library it's important to specify this that compiler and minifier on Sourcegraph side won't remove unused es imports in Cody Web web-worker.

PR #5304, opened 2024-08-23
added(agent): `chat/import` to import historical chats

jamesmcnamara:
Adds a new `chat/import` request that allows editors that had custom chat implementations (JetBrains) to import their chat history to the agent so that it can be used in the webview.

## Test plan
Manually tested with JetBrains. Need to add some agent tests.

## Changelog
Added ability to import historical chats through the agent.


mkondratek:
I tried these changes with JetBrains. I understand that with this (and with the JB part) we will be able to import the chats to the agent's storage. However, we still have no mechanism to retrieve the historical chats from the storage. Is that right?

I ask because I tested it like this:
1. started two conversations with cody 
2. reviewed the chats in the history tab
3. closed project
4. opened project
5. reviewed the chats in the history tab

Actual: The chat history is empty
Expected: The persisted chats are available in the history

jamesmcnamara:
@mkondratek I'm not sure if I completely follow the steps. To test this, I did the following:
1. Checkout this branch in Cody
2. In the sibling JetBrains repo, I checked that there was a ./.info/cody_history.xml file with a past history of chats.
3. Started up [this branch](https://github.com/sourcegraph/jetbrains/pull/2108) of JetBrains with `CODY_DIR=<path to cody> ./gradlew -PforceBuild=true runIde`
4. Opened the webview chat UI and checked out the history tab and verified my JetBrains chat history was present.

pkukielka:
I tried it locally (also with a debugger) and:

1. Migration code is ran
2. I see history being exported/imported

But:

1. It only happens once, and looks like state is not really persisted. After a second restart it does not work anymore
2. Even during the first start when import succeeded, when I click on one of the historic chats to import them it does not work (it switches beck to the chat view but old chat is not loaded).

jamesmcnamara:
@pkukielka Good catches! those were fixed by:
1. [Specifying Server-Managed global state](https://github.com/sourcegraph/jetbrains/pull/2108/commits/836d7baa3868e5d2e1b26eb3fe6ac81108f5cf2b#diff-1590630a1388e52e0e1ba6c7d2018523b9edda98c759fb605c0c88d1d672addbR129)
2. [Keying history by ID instead of date](https://github.com/sourcegraph/jetbrains/pull/2108/commits/427369d0015349ce3cbb9d274c477f4fd9f95267). I must have been looking at a bad sample file that used dates as keys.

mkondratek:
I tested it locally. Works fine now 🚀

mkondratek (approved):
I am not an expert here but lgtm and works with JetBrains

pkukielka:
I tested it as well and indeed it works now, but I found one bug.
When restarting a past conversation date in history was not set to today but `NaN`:

image

pkukielka:
On the tem call we decided to merge it to be able to cut nightly release which can be run through QA.
Fixes can be done in a followup PRs.

jamesmcnamara:
@pkukielka I tried running it with the history you sent me, but I was unable to reproduce. The only change I made to the cody_history.xml was to sub in my account ID for yours, and the resulting cody-local-chatHistory-v2 was the same as yours 
except for the account name swap. 

Perhaps it has something to do with having run it with the in-progress PR when it was using dates as IDs instead of the actual chat ID?

image

pkukielka:
@jamesmcnamara 

> When restarting a past conversation date in history was not set to today but NaN:

It appears not after export, but after you open past conversation and try to continue it, and then check history tab again.

PR #5307, opened 2024-08-23
Autocomplete: refactor the Fireworks provider

valerybugakov:
- Extracted all model-specific code from the Fireworks provider into a separate module.
- Merged FIM helpers with Ollama model helpers to have one place to define model-specific logic.
- No functional changes.

## TODO

In follow-up PRs, we should be able to generalize the model helpers' methods further, making them usable by more than just Ollama and Fireworks. For now, we are keeping things as is to reduce the blast radius and the PR diff.

## Test plan

1. CI
2. Locally tested the Fireworks provider
3. (TODO) locally tested the Ollama provider


vscode/src/completions/model-helpers/deepseek.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    formatIntroSnippets(params: FormatIntroSnippetsParams): PromptString {
+        // These model families take code from the context files without comments.
+        return ps`${PromptString.join(params.intro, ps`\n\n`)}\n`
+    }
+
+    fileSnippetToPromptString(snippet: AutocompleteFileContextSnippet): PromptString {
+        const { content } = PromptString.fromAutocompleteContextSnippet(snippet)
+        return ps`#${PromptString.fromDisplayPath(snippet.uri)}\n${content}`
```
valerybugakov, line 54: @hitesh-1997, is there a place we can link here to explain the `#` char for the display path?

vscode/src/completions/model-helpers/deepseek.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    formatFireworksPrompt(params: FormatFireworksPromptParams): PromptString {
+        const { intro, prefix, suffix, repoName, fileName } = params
+
+        // Deepseek paper: https://arxiv.org/pdf/2401.14196
+        const prompt = ps`${intro}\n#${fileName}\n<｜fim▁begin｜>${prefix}<｜fim▁hole｜>${suffix}<｜fim▁end｜>`
```
valerybugakov, line 61: @hitesh-1997, is there a place we can link here to explain the `#` char for the `fileName`? I searched for it in the paper and tokenizer but couldn't find the relevant part quickly. 

Also, AFAIU, we link the deepseek-v1 paper here. Is there anything new in the v2 that's worth linking there?

vscode/src/completions/providers/fireworks.ts
```diff
@@ -448,19 +328,3 @@ export function createProviderConfig({
...
-    return model.startsWith('llama-code')
-}
-
-function isDeepSeekModelFamily(model: string): boolean {
-    return [
-        DEEPSEEK_CODER_V2_LITE_BASE,
-        FIREWORKS_DEEPSEEK_7B_LANG_STACK_FINETUNED,
-        FIREWORKS_DEEPSEEK_7B_LANG_LOG_FINETUNED,
-    ].includes(model)
-}
```
valerybugakov, line 466: @taras-yemets, we discussed it on the call. This is fixed now.

slimsag (approved):
Interesting approach, generally LGTM - this is quite helpful in being able to more easily see what model-specific logic we have going on here.

I will be thinking a bit about how we might lift some of this logic out of typescript and into a config format later - but this is a major step in the right direction. Thank you @valerybugakov !

vscode/src/completions/model-helpers/deepseek.ts
hitesh-1997: @valerybugakov Sorry I didn't get the question, are we concerned about number of character in fileName being too long ?

vscode/src/completions/model-helpers/deepseek.ts
hitesh-1997: DeepSeek trains with `#` appended before the file name, followed by the code.
[Adding link to DS-v1](https://huggingface.co/deepseek-ai/deepseek-coder-6.7b-base#3%EF%BC%89repository-level-code-completion). I could not find the similar example for DeepSeek-v2, but they do mention in the paper somewhere that there prompt is same as DS-v1 model.

vscode/src/completions/model-helpers/deepseek.ts
hitesh-1997: DeepSeek trains with # appended before the file name, followed by the code.
[Adding link to DS-v1](https://huggingface.co/deepseek-ai/deepseek-coder-6.7b-base#3%EF%BC%89repository-level-code-completion). I could not find the similar example for DeepSeek-v2, 
but they do mention in the paper somewhere that there prompt is same as DS-v1 model.

Also, The prompt is exactly same as DS-v1.

hitesh-1997 (approved):
nice, this looks really clean now !!

vscode/src/completions/model-helpers/deepseek.ts
valerybugakov: @hitesh-1997, could it be a comment symbol that is different for different languages? If so, we must use `//` in Typescript for better quality.

hitesh-1997: @valerybugakov It is `#` for all the languages, I also had this doubt when I first saw it, offline evaluation did not make much different, replacing `#` with `//`

PR #5310, opened 2024-08-23
Chat: don't add selection to chat on option+l shortcut

olafurpg:
Previously, the option+l shortcut triggered two commands:

- Toggle chat
- Add selection to chat

This was problematic because the selection is already an initial context in chat so the chat input ended up having duplicate chips (one "initial"
    and one "user").

This PR fixes the problem by removing the option+l shortcut for "Add selection to chat" so that it only toggles the visibility of the sidebar. Users can still manually add the selection to the chat input with the option+/ shortcut.

## Test plan

* Local build
* Select text
* Trigger option+l multiple times
* Confirm that it doesn't keep adding the selection to the chat

## Changelog

* Fix bug where continuously triggering alt+l (option+l for macOS) would add duplicate context items to the chat input. Use the alt+/ shortcut to explicitly add the selection to the chat input.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -648,11 +648,6 @@
         "when": "cody.activated && editorTextFocus && editorHasSelection"
       },
       {
-        "command": "cody.mention.selection",
-        "key": "alt+l",
-        "when": "cody.activated && editorTextFocus && editorHasSelection"
-      },
-      {
         "command": "cody.tutorial.chat",
         "key": "alt+l",
         "when": "cody.activated && cody.tutorialActive"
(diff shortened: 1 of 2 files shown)


vscode/package.json
```diff
@@ -647,11 +647,6 @@
         "key": "alt+/",
         "when": "cody.activated && editorTextFocus && editorHasSelection"
       },
-      {
-        "command": "cody.mention.selection",
```
olafurpg, line 651: Duplicate to here https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody@main/-/blob/vscode/package.json?L646

umpox (approved):

sqs (approved):

PR #5311, opened 2024-08-23
Chat: default to sidebar location for chat instead of sticky

olafurpg:
Previously, we used the setting `'sticky'` by default to determine the location of the chat UI. This was problematic because the implementation of `'sticky'` can feel buggy sometimes, it was easy to reproduce scenarios where triggering alt+l in a row would cycle between the sidebar, editor, and code. This PR works around this issue by defaulting to the sidebar instead, making the alt+l shortcut have more consistent behavior.

We can discuss changing the default to `'sticky'` again when the implementation has more predictable behavior.

## Test plan

N/A

## Changelog

* The default value for the `'cody.chat.defaultLocation'` setting is now `'sidebar'` instead of `'sticky'`. This means that the alt+l shortcut (option+l on macOS) always toggles between the chat sidebar. Change this setting to `'sticky'` to get the old behavior, or `'editor'` to always use toggle to a chat editor.

### DIFF
--- a/vscode/src/chat/chat-view/ChatsController.ts
+++ b/vscode/src/chat/chat-view/ChatsController.ts
@@ -542,10 +542,9 @@ export class ChatsController implements vscode.Disposable {
 }
 
 function getNewChatLocation(): ChatLocation {
-    const chatDefaultLocation =
-        vscode.workspace
-            .getConfiguration()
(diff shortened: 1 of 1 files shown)


olafurpg:
Closing this as canceled for now since I'm unable to reproduce the surprising behavior. Will reopen in the future if I can reproduce again.

PR #5325, opened 2024-08-25
Agent: authentication capability - enable url handler and auth redirections

abeatrix:
Allows agent to set up HTTP server to listen for authentication redirections when agent capabilities for authentication are enabled.

TODO: 
- [x] authentication url handler
- [x] agent capability configuration 
- [x] work with secret storage (added by https://github.com/sourcegraph/cody/pull/5348)

Follow-ups:
- we should close the opened browser to refocus the webview if possible.

## Test plan

Example of this feature working in Visual Studio:

https://github.com/user-attachments/assets/d21c8c2f-2667-426a-9c4d-991b7645d2e5

Updated onboarding view for non VS Code editors to support login with browser for enterprise:

image


agent/src/AgentAuthHandler.ts
```diff
@@ -0,0 +1,105 @@
+import type { IncomingMessage, Server, ServerResponse } from 'node:http'
+import http from 'node:http'
+import { type CodyIDE, getCodyAuthReferralCode, logDebug } from '@sourcegraph/cody-shared'
+import open from 'open'
+import { URI } from 'vscode-uri'
+
+type CallbackHandler = (url: URI, token?: string) => void
+
+export class AgentAuthHandler {
+    private readonly port = 43452
```
dominiccooney, line 10: Could we use a dynamic port? Does the web frontend support redirecting to other ports?

agent/src/AgentAuthHandler.ts
```diff
@@ -0,0 +1,105 @@
...
+            if (req.url?.startsWith('/api/sourcegraph/token')) {
+                const url = new URL(req.url, `http://localhost:${this.port}`)
+                const token = url.searchParams.get('token')
+
+                if (token) {
+                    for (const handler of this.tokenCallbackHandlers) {
+                        handler(URI.parse(req.url), token)
+                    }
+                    res.writeHead(200, { 'Content-Type': 'text/plain' })
+                    res.end('Token received. You can now close this window.')
```
dominiccooney, line 40: We could consider opening a vscode:// redirect, or dropping a link, here to pop people back into the editor. (For editors that support it.)

For other editors which can request focus (...probably everything except vscode? ...) we could send the client a notification to request focus.

agent/src/AgentAuthHandler.ts
```diff
@@ -0,0 +1,105 @@
...
+                res.writeHead(404, { 'Content-Type': 'text/plain' })
+                res.end('Not found')
+            }
+        })
+
+        open(callbackUri)
+
+        this.server.listen(this.port, () => {
+            logDebug('AgentAuthHandler', `Server listening on port ${this.port}`)
+            setTimeout(() => this.closeServer(), 3 * 60 * 1000)
```
dominiccooney, line 56: Shall we make this 5 or 10 minutes? Someone might need to go get their 2-factor keyring or whatever... 3 minutes sounds short.

dominiccooney (changes requested):
Love the direction. Can we polish this a bit...

- Bind a dynamic port so we don't have port conflicts, assuming the site side supports different port numbers.
- Bind the loopback interface, some firewalls may be more lenient about this.
- Would be cool if the client requested focus when it has the token, although I guess that's a separate, per-client patch.
- I guess we can't remove the old app link method until a certain Sourcegraph server support window passes. Can you put TODOs or something mentioning specific versions being sunset when we can remove that? Can we turn off the app link method for dotcom and any compatible Sourcegraph version?

dominiccooney:
Ah... one more thing we need to consider. When the extension is running remotely, like SSH hosts, we should take that opportunity to have a login method that works (or at least helpfully fails... link to docs, way to manually set a token, etc.)

agent/src/AgentAuthHandler.ts
abeatrix: updated to use dynamic port

agent/src/AgentAuthHandler.ts
abeatrix: updated to 6 minutes and added a comment for the reason

agent/src/AgentAuthHandler.ts
abeatrix: > We could consider opening a vscode:// redirect, or dropping a link, here to pop people back into the editor. (For editors that support it.) 

I will add a TODO comment for this. For now, I have updated the code to send a notification to request focus via "window/focusSidebar"

dominiccooney:
> For the last item, which old app link method are you referring to?

The ones that read the querystring and redirect to URLs like [this](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/client/web/src/user/settings/accessTokens/UserSettingsCreateAccessTokenCallbackPage.tsx?L74)... vscode://sourcegraph.cody-ai etc.

High level question is, should we migrate to all http://localhost redirect logins? Or should we keep supporting the vscode:// etc. method forever. Simpler is better and localhost redirects mean we can ship new clients without updating old Sourcegraph instances.

dominiccooney (approved):
Great!

abeatrix:
> High level question is, should we migrate to all http://localhost/ redirect logins? Or should we keep supporting the vscode:// etc. method forever. Simpler is better and localhost redirects mean we can ship new clients without updating old Sourcegraph instances.

I agree that simpler is better. Right now, adding new IDE redirection support through the SG instance requires enterprise users to upgrade to the latest version of SG. Since that change would require a server-side change that is not covered by this PR, I will create a linear issue for it. Thanks for the review, Dom!

PR #5330, opened 2024-08-25
misc refactors to help make config more reactive

sqs:
A small step along the way to making our configuration more reactive throughout Cody.

## Test plan

CI

### DIFF
--- a/vscode/src/services/LocalStorageProvider.ts
+++ b/vscode/src/services/LocalStorageProvider.ts
@@ -194,17 +194,15 @@ class LocalStorage {
      * Return the anonymous user ID stored in local storage or create one if none exists (which
      * occurs on a fresh installation).
      */
-    public async anonymousUserID(): Promise<{ anonymousUserID: string; created: boolean }> {
+    public anonymousUserID(): { anonymousUserID: string; created: boolean } {
         let id = this.storage.get<string>(this.ANONYMOUS_USER_ID_KEY)
         let created = false
         if (!id) {
             created = true
             id = uuid.v4()
-            try {
-                await this.storage.update(this.ANONYMOUS_USER_ID_KEY, id)
-            } catch (error) {
-                console.error(error)
-            }
+            Promise.resolve(this.storage.update(this.ANONYMOUS_USER_ID_KEY, id).then(undefined)).catch(
+                error => console.error(error)
+            )
         }
         return { anonymousUserID: id, created }
     }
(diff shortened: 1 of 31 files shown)


varungandhi-src (approved):
This PR seems a bit too complex compared to the other PRs we typically use stampers for. It looks OK, but it took me a while to grok the changes, as someone who is not familiar with the surrounding code.

sqs:
@varungandhi-src This is a PR that combines a bunch of independent minor commits, and I got @RXminuS's and @olafurpg's high-level feedback on this direction and on a PR that had these at https://sourcegraph.slack.com/archives/C05AGQYD528/p1723791822404399. In the interest of moving forward with this big code quality improvement, I'm going to merge, but happy to chat about this if you disagree in general or think I should break this into 8 different PRs.

varungandhi-src:
My point isn't that this PR needs to be broken up into 8 PRs, it's that there is not enough context for stampers to actually stamp the PR. Here is what our [handbook](https://www.notion.so/sourcegraph/Pull-Requests-65ca2cf90370449485fa747fbad1c05c?pvs=4#b793d61ad98b4d79b59dc3118d213ab6) says:

> "Quick stamp" is a term used when requesting a review for a pull request that is likely to be trivial and easy to approve. However, it is essential to understand that it is not a request for a blind approval. A "quick stamp" should never mean bypassing your review standards or blindly approving a PR without understanding the changes.

This PR is not trivial, it's not obvious why some of the changes are being made (you did not link that Slack thread in the PR description, and that Slack thread has 10+ comments anyways). I understand that for people _with that context_, it might be a trivial PR, but not everyone on stampers has that.

valerybugakov (approved):

vscode/src/services/LocalStorageProvider.ts
```diff
@@ -194,17 +194,15 @@ class LocalStorage {
...
             created = true
             id = uuid.v4()
-            try {
-                await this.storage.update(this.ANONYMOUS_USER_ID_KEY, id)
-            } catch (error) {
-                console.error(error)
-            }
+            this.storage
+                .update(this.ANONYMOUS_USER_ID_KEY, id)
+                .then(undefined, error => console.error(error))
```
vovakulikov, line 205: Super minor. Can we use `.catch()` here instead?

olafurpg (approved):

vscode/src/services/LocalStorageProvider.ts
sqs: Yeah. It's actually a `Thenable` from the VS Code API, not a `Promise`, but `Promise.resolve` wrapping it should be fine.

PR #5335, opened 2024-08-25
Refactor Cody Web Chat wrapper

vovakulikov:
Partially fixes https://linear.app/sourcegraph/issue/SRCH-862/make-cody-web-load-faster

Kind of tech PR with Cody Web refactoring, cleans up styles and component structure since the new Cody UI is based on tabs and includes history and prompts UI there are no reasons to have component sepration like it was when we had to care about chat history on consumer level. 

It also fixes 
- Brings back skeleton UI (it doesn't solve the problem with big bundle size and time to load but improves UI feedback a bit in the Sourcegraph client) 
- Fixes problem with undefined revision in the tooltip text over context files that don't specify revision 

## Test plan
- Manual checks over Cody's Web functionality 
- If you want to check it with Sourcegraph client together you have to link it first with this PR build locally here https://github.com/sourcegraph/sourcegraph/pull/144


web/lib/components/CodyWebChat.tsx
```diff
@@ -213,39 +269,33 @@ export const CodyWebPanel: FC<CodyWebPanelProps> = props => {
         [vscodeAPI, telemetryRecorder, chatModelContext, clientState, config, envVars]
```
sqs, line 269: This is a great simplification. Any other big outstanding things you need done (or will do) in the rest of the Cody codebase to permanently ease the Cody Web build target?

web/lib/components/skeleton/ChatSkeleton.module.css
```diff
@@ -0,0 +1,117 @@
...
+        height: 0.5rem;
+    }
+
+    &-short-text {
+
+        max-width: 3rem;
+    }
+
+    &-text {
+
```
sqs, line 102: Tailwind CSS is preferred going forward but not a big deal

web/lib/components/skeleton/ChatSkeleton.tsx
```diff
@@ -0,0 +1,78 @@
+import type { FC } from 'react'
+
+import classNames from 'classnames'
+import styles from './ChatSkeleton.module.css'
+
+interface ChatSkeletonProps {
+    className?: string
+}
+
+export const ChatSkeleton: FC<ChatSkeletonProps> = props => {
```
sqs, line 10: Explain what it is in a docstring + explain why this is only needed in web. Is it because it’s only needed when loading, and on non-web there is no loading?

web/lib/components/use-cody-agent.ts
```diff
@@ -0,0 +1,180 @@
...
+    useMemo,
+    useRef,
+    useState,
+} from 'react'
+import type { MessageConnection } from 'vscode-jsonrpc/browser'
+import { URI } from 'vscode-uri'
+import { createAgentClient } from '../agent/agent.client'
+import type { InitialContext } from '../types'
+
+// Usually the CodyWebPanelProvider VSCode API wrapper listens only to messages from the Extension host
```
sqs, line 18: Use a docstring in jsdoc style so hovers show this comment on the symbol

web/lib/components/use-cody-agent.ts
```diff
@@ -0,0 +1,180 @@
...
+
+        await agent.rpc.sendRequest('webview/receiveMessage', {
+            id: activeWebviewPanelIDRef.current,
+            message: { chatID: chatId, command: 'restoreHistory' },
+        })
+    }, [])
+
+    const vscodeAPI = useVSCodeAPI({ activeWebviewPanelIDRef, createNewChat, client })
+
+    // Always create new chat when Cody Web is opened for the first time
```
sqs, line 93: Consistently end sentence comments with a period or not. Next one ends with a period. (Minor!)

sqs (approved):

web/lib/components/CodyWebChat.tsx
vovakulikov: I think the biggest further changes should be done on the DX side of this project, automatic releases and e2e tests. I have it on my board for post-GA release. 

If we're talking about the codebase itself, I don't have much at the moment. I wanted to change how RPC API works but I saw that you've already migrated it and it should be no problem with long-awaited events anymore (I fixed it partially for Cody Web for some events by having req/res IDs but your general approach is much better) 

Other topic here is to have a way to override UI parts of the Cody like confirmation modals when you click on "Delete all history items". I have it for the GA release, so it will be in my next PR or so. 

If you have anything you want me to change or improve let me know.

web/lib/components/skeleton/ChatSkeleton.tsx
vovakulikov: This is correct, it is also because it takes noticeable time to init Cody Web through rpc and cody agent API. So, this skeleton UI will probably be useful for other IDE integrations where we use Cody Agent and unified Cody Webview UI. 

I left a docstring comment there as you suggested

web/lib/components/use-cody-agent.ts
vovakulikov: I'm with you here (I'm in the docstring team), I just noticed in other parts of the cody repository that we use just plain comments usually like in a few places in the agent codebase [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/vscode-shim.ts?L193-197), [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/vscode-shim.ts?L219-220), [here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/AgentTextDocument.ts?L10-17)  

I changed plain comments to docstring in Cody Web codebase, thanks for calling it out.

web/lib/components/CodyWebChat.tsx
vovakulikov: I try to keep our Cody Web GA board as updated as possible here https://linear.app/sourcegraph/project/cody-web-ga-c4287f433a8c/view/2a8cc3b2-8196-41fc-aa4e-60f5193e1e8c

So If you have any questions about what is going on with Cody Web or what's the plan, it should help with an initial understanding 😉

PR #5336, opened 2024-08-25
agent: add internal context command

beyang:
Adds a `cody internal context` command for running batch evaluations against the Sourcegraph context API.

Here's an example schema of the input file: 

repoNames | query | essentialContext
-- | -- | --
github.com/sourcegraph/conc | show me how to use WaitGroup | conc waitgroup.go:22-52conc waitgroup_test.go:13-40
github.com/sourcegraph/concgithub.com/golang/go | show how to use WaitGroup to read a batch of files in parallel | conc waitgroup.go:22-52go src/sync/waitgroup:13-30go src/os/file.go:717-73

The output file contains the context fetched and evaluation stats (currently variants of essential recall against the input list of essential context).

## Test plan

`cody internal context --src-endpoint 'https://sourcegraph.com' --src-access-token $ACCESS_TOKEN --input-file inputs.csv  --output-file output.csv`


sqs (approved):

janhartman:
What's the point of this if we have the [context-eval framework](https://github.com/sourcegraph/context-eval/tree/main)?

PR #5341, opened 2024-08-26
Commands ux improvements

arafatkatze:
[Linear Issue
](https://linear.app/sourcegraph/issue/CODY-3234/prefer-visual-indication-of-running-status-after-clicking-a-command#comment-459cd279)

This PR addresses two key issues with the Generate Tests and Document Code commands:

	1.	Lack of User Feedback: Previously, when users clicked on Generate Tests or Document Code, there was no immediate visual feedback, leaving users uncertain whether the commands were executed. This often led to users repeatedly clicking the buttons, as they couldn’t see any change or indication that something was happening.
	2.	User Experience Improvement: To mitigate this issue, I’ve implemented a change where the cursor now jumps to the appropriate location when either command is executed. This provides instant feedback, informing users that the command is in progress and reducing the likelihood of repeated clicks.

For a more detailed explanation of the problem, please refer to the [accompanying video](https://www.loom.com/share/f39b15824fd5409eae873ad8032e209a).

[Additional comments
](https://linear.app/sourcegraph/issue/CODY-3234/prefer-visual-indication-of-running-status-after-clicking-a-command#comment-41bef8b9)

## Test plan
Tested locally with a vs code extension debugger. ITs fairly easy to see that cursor jumps to codelens for both test and doc

[See Video
](https://www.loom.com/share/6c4f2914597747b4a5b8daa19b656971)

### DIFF
--- a/agent/src/AgentTextEditor.ts
+++ b/agent/src/AgentTextEditor.ts
@@ -4,23 +4,28 @@ import type { AgentTextDocument } from './AgentTextDocument'
 import type { EditFunction } from './AgentWorkspaceDocuments'
 
 export class AgentTextEditor implements vscode.TextEditor {
+    private _selection: vscode.Selection
     constructor(
         private readonly agentDocument: AgentTextDocument,
         private readonly params?: { edit?: EditFunction }
-    ) {}
+    ) {
+        this._selection = new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
+    }
     get document(): AgentTextDocument {
         return this.agentDocument
     }
     get selection(): vscode.Selection {
         const protocolSelection = this.agentDocument.protocolDocument.selection
-        const selection: vscode.Selection = protocolSelection
-            ? new vscode.Selection(
-                  new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
-                  new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
-              )
-            : // Default to putting the cursor at the start of the file.
-              new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
-        return selection
+        if (protocolSelection) {
+            this._selection = new vscode.Selection(
+                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+            )
+        }
+        return this._selection
+    }
+    set selection(newSelection: vscode.Selection) {
+        this._selection = newSelection
     }
     get selections(): readonly vscode.Selection[] {
         return [this.selection]
--- a/agent/src/AgentTextEditor.ts
+++ b/agent/src/AgentTextEditor.ts
@@ -22,8 +22,7 @@ export class AgentTextEditor implements vscode.TextEditor {
               new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
         return selection
     }
-    set selection(newSelection: vscode.Selection) {
-    }
+    set selection(newSelection: vscode.Selection) {}
 
     get selections(): readonly vscode.Selection[] {
         return [this.selection]
--- a/agent/src/AgentTextEditor.ts
+++ b/agent/src/AgentTextEditor.ts
@@ -10,6 +10,13 @@ export class AgentTextEditor implements vscode.TextEditor {
         private readonly params?: { edit?: EditFunction }
     ) {
         this._selection = new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
+        const protocolSelection = this.agentDocument.protocolDocument.selection
+        if (protocolSelection) {
+            this._selection = new vscode.Selection(
+                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+            )
+        }
     }
     get document(): AgentTextDocument {
         return this.agentDocument
--- a/agent/src/AgentTextEditor.ts
+++ b/agent/src/AgentTextEditor.ts
@@ -4,36 +4,27 @@ import type { AgentTextDocument } from './AgentTextDocument'
 import type { EditFunction } from './AgentWorkspaceDocuments'
 
 export class AgentTextEditor implements vscode.TextEditor {
-    private _selection: vscode.Selection
     constructor(
         private readonly agentDocument: AgentTextDocument,
         private readonly params?: { edit?: EditFunction }
-    ) {
-        this._selection = new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
-        const protocolSelection = this.agentDocument.protocolDocument.selection
-        if (protocolSelection) {
-            this._selection = new vscode.Selection(
-                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
-                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
-            )
-        }
-    }
+    ) {}
     get document(): AgentTextDocument {
         return this.agentDocument
     }
     get selection(): vscode.Selection {
         const protocolSelection = this.agentDocument.protocolDocument.selection
-        if (protocolSelection) {
-            this._selection = new vscode.Selection(
-                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
-                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
-            )
-        }
-        return this._selection
+        const selection: vscode.Selection = protocolSelection
+            ? new vscode.Selection(
+                  new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                  new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+              )
+            : // Default to putting the cursor at the start of the file.
+              new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
+        return selection
     }
     set selection(newSelection: vscode.Selection) {
-        this._selection = newSelection
     }
+
     get selections(): readonly vscode.Selection[] {
         return [this.selection]
     }
--- a/vscode/src/edit/manager.ts
+++ b/vscode/src/edit/manager.ts
@@ -218,12 +218,16 @@ export class EditManager implements vscode.Disposable {
                 model: task.model,
             },
         })
+        /**
+         * Updates the editor's selection and view for 'doc' or 'test' intents, causing the cursor to
+         * move to the beginning of the selection range considered for the edit.
+         * 
+         */
         if (editor.active && (intent === 'doc' || intent === 'test')) {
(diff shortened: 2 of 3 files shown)


agent/src/AgentTextEditor.ts
```diff
@@ -4,23 +4,28 @@ import type { AgentTextDocument } from './AgentTextDocument'
...
-        return selection
+        if (protocolSelection) {
+            this._selection = new vscode.Selection(
+                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+            )
+        }
+        return this._selection
+    }
+    set selection(newSelection: vscode.Selection) {
```
abeatrix, line 27: i dont see any code change related to this new set method, is it actually being used / needed?

agent/src/AgentTextEditor.ts
```diff
@@ -4,23 +4,28 @@ import type { AgentTextDocument } from './AgentTextDocument'
 import type { EditFunction } from './AgentWorkspaceDocuments'
 
 export class AgentTextEditor implements vscode.TextEditor {
+    private _selection: vscode.Selection
```
abeatrix, line 7: what's the different between `this._selection` and `this.selection`?

agent/src/AgentTextEditor.ts
```diff
@@ -4,23 +4,28 @@ import type { AgentTextDocument } from './AgentTextDocument'
...
     constructor(
         private readonly agentDocument: AgentTextDocument,
         private readonly params?: { edit?: EditFunction }
-    ) {}
+    ) {
+        this._selection = new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
```
abeatrix, line 12: i'm not following why we need to set the _selection to 0 instead of using the values from `this.agentDocument.protocolDocument.selection`?  Maybe a comment would be helpful here!

abeatrix (commented):
The change on the vscode/src/edit/manager.ts file makes sense to me, but i'm not sure about the changes on the Agent side--are those changes needed on the Agent side for it to work in other clients? Maybe some comments to clarifying the changes on the agent side would be helpful too!

agent/src/AgentTextEditor.ts
arafatkatze: The setter method is being used here 
image

without the setter this code won't work

arafatkatze: Yes its being used, there are some tests relating to agents [here](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=commands-ux-improvements&file=agent%2Fsrc%2Fdocument-code.test.ts&editor=VSCode&version=2.2.17&start_row=24&start_col=21&end_row=34&end_col=6) and if I don't do add a setter method here those tests would fail with 

```

 FAIL  |agent| src/document-code.test.ts > Document Code > commands/document (nested test case)
Error: Cannot set property selection of #<AgentTextEditor> which has only a getter

TypeError: Cannot set property selection of #<AgentTextEditor> which has only a getter
    at EditManager.executeEdit (/Users/arafatkhan/Desktop/cody/vscode/src/edit/manager.ts:228:27)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at process.processImmediate (node:internal/timers:449:9)
    at process.topLevelDomainCallback (node:domain:161:15)
    at process.callbackTrampoline (node:internal/async_hooks:128:24)
    at <anonymous> (/Users/arafatkhan/Desktop/cody/vscode/src/commands/execute/doc.ts:168:19)
    at Agent.createEditTask (/Users/arafatkhan/Desktop/cody/agent/src/agent.ts:1770:25)
    at <anonymous> (/Users/arafatkhan/Desktop/cody/vscode/src/jsonrpc/jsonrpc.ts:95:21)
 ❯ EditManager.executeEdit ../vscode/src/edit/manager.ts:228:27
    226|         if (editor.active && (intent === 'doc' || intent === 'test')) {
    227|             const newPosition = proposedRange.start
    228|             editor.active.selection = new vscode.Selection(newPosition, newPosition)
       |                           ^
    229|             editor.active.revealRange(
    230|                 new vscode.Range(newPosition, newPosition),
 ❯ <anonymous> ../vscode/src/commands/execute/doc.ts:168:19
 ❯ Agent.createEditTask src/agent.ts:1770:25
 ❯ <anonymous> ../vscode/src/jsonrpc/jsonrpc.ts:95:21
 ❯ handleResponse ../node_modules/.pnpm/vscode-jsonrpc@8.2.0/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48
 ❯ handleMessage ../node_modules/.pnpm/vscode-jsonrpc@8.2.0/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13
 ❯ processMessageQueue ../node_modules/.pnpm/vscode-jsonrpc@8.2.0/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17
 ❯ Immediate.<anonymous> ../node_modules/.pnpm/vscode-jsonrpc@8.2.0/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13

```

The setter method is responsible for doing 

`            editor.active.selection = new vscode.Selection(newPosition, newPosition)`

for the case of when the editor is the agentTextEditor

agent/src/AgentTextEditor.ts
arafatkatze: Fair point I added the condition to fix that in the constructor

agent/src/AgentTextEditor.ts
arafatkatze: 1. _selection is a private property of the class, initialized in the constructor with a default selection at position (0, 0).

2. selection is a getter method that:

Retrieves the current selection from the protocolDocument
Updates the _selection property with the retrieved selection
Returns the updated _selection


Since this is confusing I change _selection to _lastknownselection

agent/src/AgentTextEditor.ts
arafatkatze: Its just the default initialization value every instance of actual use of `_selection` or `_lastknownSelection` would entail a getter method which would override the value with `this.agentDocument.protocolDocument.selection` if it exists.

agent/src/AgentTextEditor.ts
abeatrix: Thanks for clarifying! I think this is not necessary because we should always use the selection values returned from this.agentDocument.protocolDocument.selection, so I'd remove this along with the rest of the suggestions I make below to see if that makes sense to you:

```suggestion
```

agent/src/AgentTextEditor.ts
```diff
@@ -4,23 +4,35 @@ import type { AgentTextDocument } from './AgentTextDocument'
...
+        const protocolSelection = this.agentDocument.protocolDocument.selection
+        if (protocolSelection) {
+            this._selection = new vscode.Selection(
+                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+            )
+        }
+    }
```
abeatrix, line 20: I'd remove this since this is just a copy of the get selection() code

```suggestion
```

agent/src/AgentTextEditor.ts
```diff
@@ -4,23 +4,35 @@ import type { AgentTextDocument } from './AgentTextDocument'
...
-            : // Default to putting the cursor at the start of the file.
-              new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
-        return selection
+        if (protocolSelection) {
+            this._selection = new vscode.Selection(
+                new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
+                new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
+            )
+        }
+        return this._selection
```
abeatrix, line 32:
```suggestion
     const selection: vscode.Selection = protocolSelection
            ? new vscode.Selection(
                  new vscode.Position(protocolSelection.start.line, protocolSelection.start.character),
                  new vscode.Position(protocolSelection.end.line, protocolSelection.end.character)
              )
            : // Default to putting the cursor at the start of the file.
              new vscode.Selection(new vscode.Position(0, 0), new vscode.Position(0, 0))
        return selection
```

I'd revert this change since we should always use the selection from the agentDocument. If there is no selection in the current AgentDocument, this is returning the last selection we have stored in the TextEditor, which is going to be incorrect.

agent/src/AgentTextEditor.ts
```diff
@@ -4,23 +4,35 @@ import type { AgentTextDocument } from './AgentTextDocument'
...
+    }
+    set selection(newSelection: vscode.Selection) {
+        this._selection = newSelection
```
abeatrix, line 35:
```suggestion
        // Do nothing as selection is set by agentDocument
```

I think we didn't implement the set selection() because it's not needed, so we can just add a no-op method to bypass the error `Error: Cannot set property selection of #<AgentTextEditor> which has only a getter` as we don't need to set the selection twice in TextEditor and AgentDocument

abeatrix (changes requested):
@arafatkatze thanks for clarifying! with more info i think i understand why you made the changes on the agent side, which I don't think is needed. I've made some new suggestions on the PR, does this make sense to you? Or am I misunderstood the intent for this change?

agent/src/AgentTextEditor.ts
abeatrix: The test is still passing after making all the suggested changes: 
image

vscode/src/edit/manager.ts
```diff
@@ -218,7 +218,19 @@ export class EditManager implements vscode.Disposable {
...
+         *
+         */
+        if (editor.active && (intent === 'doc' || intent === 'test')) {
+            const newPosition = proposedRange.start
+            editor.active.selection = new vscode.Selection(newPosition, newPosition)
+            editor.active.revealRange(
+                new vscode.Range(newPosition, newPosition),
+                vscode.TextEditorRevealType.InCenter
+            )
+        }
```
abeatrix, line 233: Hmmm i might have missed some context, but this is going to replace user's current selection which is not what we want to do, is that correct? we could just revealRange to the `range` value, which is created above using the proposedRange?

```suggestion
        if (editor.active && (isDocCommand || isUnitTestCommand)) {
            editor.active.revealRange(range, vscode.TextEditorRevealType.InCenter)
        }
```

arafatkatze:
@abeatrix that makes sense. Thanks for clarifying the logic around the selection overriding. I cleaned up the code.

vscode/src/edit/manager.ts
arafatkatze: So the reason to not make this change is because having the cursor jump also is an added pointer that makes the user think that this change actually "does something" and so rather than just shift the view of the screen upwards making the cursor jump gives people a tangible evidence that clicking this does something instantaneously.

abeatrix (approved):
thanks for making the changes! approving since it sounds like you have discussed moving user's selection to the top of selection with the product team.

PR #5343, opened 2024-08-27
Autocomplete: use model helpers in `openaicompatible` provider

valerybugakov:
- `openaicompatible` provider now uses the same model helpers as the `fireworks` provider. 
- Added unit tests for each model helper for easy prompt reviews.
- Made adjustments to the intro snippets formatter, which reduces the number of empty lines and duplicate spaces after comment symbols.

## Test plan

1. CI
2. Manually tested with sg04 and starcoder-7b


hitesh-1997 (approved):
Thanks for the change @valerybugakov. This really helps have a common setup across provider :pray: 
I mostly looked at the new apis for the model and snapshot and looks good to me. Approving to unblock from my end.

vscode/src/completions/model-helpers/index.ts
```diff
@@ -15,10 +16,14 @@ export function getModelHelpers(model: string): DefaultModel {
         return new DeepseekCoder()
     }
 
-    if (model.includes('starcoder')) {
+    if (model.includes('starcoder') || model.includes('starchat')) {
         return new StarCoder()
```
slimsag, line 20: Is the assumption here that StarCoder is close enough in terms of prompting to behave similarly as StarChat?

I think both support FIM (don't quote me on it) so maybe it works out that way.

Maybe we should rename `new StarCoder()` to `new StarCoderOrChat()` or something to remove confusion though?

valerybugakov: Added a comment to address [this](https://github.com/sourcegraph/cody/pull/5343/commits/1b318ffd474c5a3e6a94fc271aed5c11bb39b5cf)

PR #5345, opened 2024-08-27
Smart Apply: support writing to empty files

olafurpg:
Fixes CODY-3421

Previously, smart apply didn't work for empty files, it would create a new file in the root directory instead. This was problematic because users could create an empty file in a subdirectory, ask chat to write code, press smart "Apply" and Cody would open a different file with the same name (but in the root directory).

This PR fixes the problem by adding empty files to the context if those files are explicitly added as context items in the chat input. Previously, we skipped empty files, presumably because they add limited value when pulled implicitly (via embeddings or keyword search).

## Test plan

* Create empty file in subdirectory lib/foo.ts
* Ask chat question with empty file added as context: "@foo.ts write fibonacci"
* Click "Apply"
* Expectation: it adds the code to lib/foo.ts
  * Actual, before this PR: it adds the code to foo.ts

## Changelog

* Fixed a bug when smart apply would not work with empty files

### DIFF
--- a/lib/shared/src/prompt/prompt-string.ts
+++ b/lib/shared/src/prompt/prompt-string.ts
@@ -350,12 +350,17 @@ export class PromptString {
         }
     }
 
-    public static fromContextItem(contextItem: ContextItem) {
+    public static fromContextItem(contextItem: ContextItem): {
+        content: PromptString | undefined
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

PR #5346, opened 2024-08-27
Guardrails: fix DDoS bug

olafurpg:
Fixes SRCH-942

Previously, if an enterprise user had a chat window open with a code block containing more than 10 lines of code, then Cody would send a request to our Guardrails server on every single editor event. This was caused by a `useEffect` block that has a large number of dependencies where one of the dependencies (`humanMessage`) was not memoized.

This PR fixes the bug by wrapping `humanMessage` in a `useMemo` block. Ideally, the code should be refactored a bit to have more narrow `useEffect` blocks with fewer dependencies.

## Test plan

* Sign into enterprise account (S2)
* Open chat in an editor
* Ask a question that generates a code block with >10 loc
* Open webview developer tools, network tab
* Select text in the editor
* Old behavior: the guardrails icon was spinning on every selection event.
* Expected: confirm that we don't send a guardrails network request on every editor selection. The icon no longer spins.

CleanShot 2024-08-27 at 12 00 32@2x

## Changelog

* Fixed a bug where the guardrails icon was spinning on every editor selection event

### DIFF
--- a/vscode/webviews/chat/ChatMessageContent/ChatMessageContent.tsx
+++ b/vscode/webviews/chat/ChatMessageContent/ChatMessageContent.tsx
@@ -87,6 +87,10 @@ export const ChatMessageContent: React.FunctionComponent<ChatMessageContentProps
         }, [])
     )
 
+    // See SRCH-942: this `useEffect` is very large and any update to the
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

PR #5348, opened 2024-08-27
Agent: add new `secrets` capability to implement secret storage

olafurpg:
Previously, the agent only supported stateless secret storage, where the agent server stored secrets in a temporary hashmap that was lost whenever the agent started. Now, clients can optionally declare that they're able to store secrets using the `secrets: 'client-managed'` capability. With this new capability, client can
store/retrieve/delete/change secrets using the new JSON-RPC methods:

* `secrets/get`
* `secrets/store`
* `secrets/delete`
* `secrets/didChange`

Here's the PR moving to client-managed secrets for the Eclipse plugin, which allowed us to delete 600 lines of native UI code 😮 https://github.com/sourcegraph/eclipse/pull/54

## Test plan

This PR changes `TestClient` to use client-managed secrets, so we're stressing this new code path in all the integration tests by default.


abeatrix (approved):
Lgtm this would be super helpful with the authentication capabilities!

jamesmcnamara (approved):
Nice! Agent code is finally starting to look straightforward to me.

PR #5350, opened 2024-08-27
Adding trigger delay to the code

arafatkatze:
[Solves Linear Issue
](https://linear.app/sourcegraph/issue/CODY-3331/introduce-option-to-control-suggestion-pause-time-in-vscode) to introduce configurable trigger delays before making autocomplete suggestions for people who might want to slow down autocomplete suggestions for different reasons
 
## Test plan
[Added a loom video](https://www.loom.com/share/30266970e3c640e4a3cd2accce8e1503) showing trigger delay in different cases with the debugger. 
The way I tested this was first setting the trigger delay to really high and saving it in the settings of Vscode(as shown in the video) and then the slow manifestation of completion suggestions was obvious and then I lowered the trigger delay to 0 in the settings and the completion suggestions were much much faster thereby proving that the setting for trigger delay can be useful for people who find the immediate trigger suggestions to be too cumbersome.

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -38,6 +38,7 @@ Chat: Fixed feedback buttons not working in chat. [pull/5509](https://github.com
 
 - Edit: Implemented cursor feedback for Generate Tests and Document Code commands to improve user experience by indicating command execution. [pull/5341](https://github.com/sourcegraph/cody/pull/5341)
 - Dev: Added support for configurable "options" field in locally configured LLM providers, available behind the `cody.dev.models` setting. [pull/5467](https://github.com/sourcegraph/cody/pull/5467)
+- Autocomplete Trigger Delay: Introduced a configurable setting to add a delay before returning autocomplete results, enhancing user control over completion suggestion timing. [pull/5350](https://github.com/sourcegraph/cody/pull/5350)
 
 ## 1.32.5
 
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1025,6 +1025,12 @@
           "markdownDescription": "Enables code autocompletions.",
           "default": true
         },
+        "cody.autocomplete.triggerDelay": {
+          "order": 5,
+          "type": "number",
+          "markdownDescription": "Adds a delay in milliseconds before triggering autocomplete. This is useful for avoiding accidental autocomplete triggers when typing. Restart VS Code to apply changes.",
+          "default": 0
+        },
         "cody.autocomplete.languages": {
           "order": 5,
           "type": "object",
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1028,7 +1028,7 @@
         "cody.autocomplete.triggerDelay": {
           "order": 5,
           "type": "number",
-          "markdownDescription": "Adds a delay in milliseconds before triggering autocomplete. This is useful for avoiding accidental autocomplete triggers when typing. Restart VS Code to apply changes.",
+          "markdownDescription": "The trigger delay ensures a minimum wait time before showing autocomplete suggestions to avoid accidental autocomplete triggers. Restart VS Code to apply changes.",
           "default": 0
         },
         "cody.autocomplete.languages": {
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1028,7 +1028,7 @@
         "cody.autocomplete.triggerDelay": {
           "order": 5,
           "type": "number",
-          "markdownDescription": "The trigger delay ensures a minimum wait time before showing autocomplete suggestions to avoid accidental autocomplete triggers. Restart VS Code to apply changes.",
+          "markdownDescription": "The trigger delay ensures a minimum wait time before showing autocomplete suggestions to avoid accidental autocomplete triggers.",
           "default": 0
         },
         "cody.autocomplete.languages": {
--- a/vscode/src/completions/inline-completion-item-provider.test.ts
+++ b/vscode/src/completions/inline-completion-item-provider.test.ts
@@ -42,6 +42,7 @@ class MockableInlineCompletionItemProvider extends InlineCompletionItemProvider 
     ) {
         super({
             completeSuggestWidgetSelection: true,
+            triggerDelay: 0,
             // Most of these are just passed directly to `getInlineCompletions`, which we've mocked, so
             // we can just make them `null`.
             statusBar: null as any,
@@ -105,6 +106,56 @@ describe('InlineCompletionItemProvider', () => {
             },
           ]
         `)
+    })
+
+    it('should record telemetry event after completion is visible and respects the trigger delay', async () => {
+        vi.useRealTimers()
+        const spy = vi.spyOn(telemetryRecorder, 'recordEvent')
+        const startTime = Date.now()
+        const triggerDelay = 30
+
+        const completionParams = params('const foo = █', [completion`bar`])
+        vi.spyOn(vscode.window, 'activeTextEditor', 'get').mockReturnValue({
+            ...vscode.window.activeTextEditor,
+            document: completionParams.document,
+            selection: { active: completionParams.position },
+        } as any)
+
+        const provider = new MockableInlineCompletionItemProvider(
+            () => getInlineCompletionsFullResponse(completionParams),
+            { triggerDelay }
+        )
+
+        // Provide completion items
+        await provider.provideInlineCompletionItems(
+            completionParams.document,
+            completionParams.position,
+            DUMMY_CONTEXT
+        )
+
+        // Check if the trigger delay is respected
+        const elapsedTime = Date.now() - startTime
+        expect(elapsedTime).toBeGreaterThanOrEqual(triggerDelay)
+        // We only check for greater than because a less than would vary depending on the CI machine
+
+        // Switch to fake timers for precise control
+        vi.useFakeTimers()
+
+        // Advance time, but not enough for the completion to be considered visible
+        vi.advanceTimersByTime(500)
+        expect(spy).toHaveBeenCalledTimes(0)
+
+        // Advance time to make the completion visible (total 750ms)
+        vi.advanceTimersByTime(250)
+
+        // Check if telemetry event is recorded
+        CompletionLogger.logSuggestionEvents(true)
+        expect(spy).toHaveBeenCalledTimes(1)
+        expect(spy).toHaveBeenCalledWith(
+            'cody.completion',
+            'suggested',
+            expect.objectContaining({ metadata: expect.objectContaining({ read: 0 }) })
+        )
     })
 
     it('prevents completions inside comments', async () => {
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -129,6 +129,7 @@ export class InlineCompletionItemProvider
 
     constructor({
         completeSuggestWidgetSelection = true,
+        triggerDelay = 0,
         formatOnAccept = true,
         disableInsideComments = false,
         tracer = null,
@@ -140,6 +141,7 @@ export class InlineCompletionItemProvider
         InlineCompletionItemProviderConfigSingleton.set({
             ...config,
             completeSuggestWidgetSelection,
+            triggerDelay,
             formatOnAccept,
             disableInsideComments,
             tracer,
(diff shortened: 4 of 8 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -619,6 +619,22 @@ export class InlineCompletionItemProvider
...
+                // 2. Allowing time for the user to continue typing, which may result in more accurate suggestions.
+                // 3. Mimicking a more natural, non-instantaneous autocomplete behavior if desired.
+                const triggerDelay = vscode.workspace
+                    .getConfiguration()
+                    .get<string>('cody.autocomplete.triggerDelay')
+
+                // Use a default value if triggerDelay is undefined
+                const delay = triggerDelay ? Number.parseInt(triggerDelay, 0) : 0
+
+                if (delay > 0) {
```
valerybugakov, line 634: We should use this logic [after the cancellation token check](https://github.com/sourcegraph/cody/blob/9dde3e60d6b80ed835330607c22e6d42315283f8/vscode/src/completions/inline-completion-item-provider.ts#L369-L374) to delay the start of the completion generation pipeline.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -619,6 +619,22 @@ export class InlineCompletionItemProvider
...
+                    await new Promise(resolve => setTimeout(resolve, delay))
```
valerybugakov, line 635: To avoid doing redundant work if a user continues typing after the delay expires, we should return `null` here to exit early. [Similar](https://github.com/sourcegraph/cody/blob/9dde3e60d6b80ed835330607c22e6d42315283f8/vscode/src/completions/get-inline-completions.ts#L425-L430) to `waitInterval`.

vscode/package.json
```diff
@@ -1014,6 +1014,13 @@
           "markdownDescription": "Enables code autocompletions.",
           "default": true
         },
+
```
valerybugakov, line 1017:
```suggestion
```

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -619,6 +619,22 @@ export class InlineCompletionItemProvider
...
                 recordExposedExperimentsToSpan(span)
 
+                // Trigger delay is used to introduce a configurable delay before returning autocomplete results.
+                // This can be useful for:
+                // 1. Reducing the frequency of autocomplete requests, potentially saving resources.
+                // 2. Allowing time for the user to continue typing, which may result in more accurate suggestions.
+                // 3. Mimicking a more natural, non-instantaneous autocomplete behavior if desired.
+                const triggerDelay = vscode.workspace
+                    .getConfiguration()
+                    .get<string>('cody.autocomplete.triggerDelay')
```
valerybugakov, line 629: Let's add a test case for this logic to `vscode/src/completions/inline-completion-item-provider-e2e.test.ts`

vscode/package.json
```diff
@@ -1014,6 +1014,13 @@
...
+        "cody.autocomplete.triggerDelay": {
```
valerybugakov, line 1018: Let's add it to the changelog.

dominiccooney:
Feedback on the description:

Say "Fixes CODY-3331" to have good things happen, we don't need that verbose link.

The test plan needs work. So vague. What settings. Where's the instructions to connect the debugger? What actions do you take? What are the expected results?

arafatkatze:
@dominiccooney Apologies for the imperfect description. I will make a video and share after I add the tests as suggested by Valerie.

vscode/src/completions/inline-completion-item-provider.ts
arafatkatze: @valerybugakov I have spent a lot of time trying to figure this out but I can't seem to figure out how to write a test that actually goes through this code in a proper way. Firstly all the end to end tests [were skipped beforehand](https://sourcegraph.com/github.com/sourcegraph/cody@triggerDelay/-/blob/vscode/src/completions/inline-completion-item-provider-e2e.test.ts?L192:39-200:38) so it was slightly harder to take inspiration.
 
So here's a couple of my attempts at writing the tests that didn't work out:

1. I first wrote this test 
```


describe('MainInlineCompletionItemProvider E2E', () => {
    describe('smart throttle in-flight requests', () => {
        let getCompletionProviderSpy: MockInstance

        beforeAll(async () => {
            await initCompletionProviderConfig({
                autocompleteExperimentalHotStreakAndSmartThrottle: true,
            })
            localStorage.setStorage({
                get: () => null,
                update: () => {},
            } as any as vscode.Memento)
        })

        beforeEach(() => {
            vi.spyOn(contextFiltersProvider, 'isUriIgnored').mockResolvedValue(false)
            getCompletionProviderSpy = vi.spyOn(CompletionProvider, 'getCompletionProvider')
        })

        it('handles multiple parallel requests, by marking the old one as stale and only suggesting one of the remaining ones', async () => {
            vi.useFakeTimers()
            const logSpy: MockInstance = vi.spyOn(telemetryRecorder, 'recordEvent')
            // const mockGet = vi.fn().mockReturnValue('500')
            // const mockGetConfiguration = vi.fn().mockReturnValue({ get: mockGet })
            // vi.spyOn(vsCodeMocks.workspace, 'getConfiguration').mockImplementation(mockGetConfiguration)

            const provider = getInlineCompletionProvider()

            const { mockRequestProvider: provider1, resolve: resolve1 } = createCompletion(
                'console.█',
                provider
            )

            getCompletionProviderSpy.mockReturnValueOnce(provider1)

            const [result1] = await Promise.all([
                // The first completion will be triggered immediately, but takes a while to resolve
                resolve1("log('hello')", {
                    delay: 0,
                    duration: 800,
                }),
                vi.advanceTimersByTimeAsync(2000), // Enough for all to be shown
            ])

            // Result 1 is marked as stale
            expect(result1).toBeDefined()

            
            // Enough for completion events to be logged
            vi.advanceTimersByTime(1000)
            CompletionLogger.logSuggestionEvents(true)

            expect(getAnalyticEventCalls(logSpy)).toMatchInlineSnapshot(`
              [
                [
                  "cody.completion",
                  "suggested",
                  {
                    "read": false,
                  },
                ],
              ]
            `)
        })
    })
})

```

And this test works but I needed to make a proper test for the triggerdelay too.  I couldn't use the VscodeMocks to properly raise the trigger delay. I would set the triggerdelay [here](https://sourcegraph.com/github.com/sourcegraph/cody@triggerDelay/-/blob/vscode/src/testutils/mocks.ts?L827:20-841:11) in the codebase with a diff of 

```
                  switch (key) {
                        case 'cody.debug.filter':
                            return '.*'
                        case 'cody.autocomplete.triggerDelay':
                            return "2900"
                        default:
                            return defaultValue
                    }
```
And that would add a global trigger delay to the whole function but then the function would just timeout(rather than the promise resolving and me getting to prove that the result is delayed). 

2. The other thing I tried doing was something like this
```
vi.useFakeTimers()
const logSpy: MockInstance = vi.spyOn(telemetryRecorder, 'recordEvent')
const getConfigurationSpy = vi.spyOn(vsCodeMocks.workspace, 'getConfiguration')
getConfigurationSpy.mockReturnValue({
    get: vi.fn().mockImplementation((key) => {
        if (key === 'cody.autocomplete.triggerDelay') {
            return '0'  // Changed from '2000' to '0'
        }
        return undefined
    })
} as any)
const provider = getInlineCompletionProvider()

// ... rest of the test remains the same
```

But even then I would never see the `getConfigurationSpy` actually being called when I would set the spy on it. MY goal was to set up a test such that there are 2 cases one where the triggerdelay is so long the result gets aborted before it even comes and the other case where you wait long enough for the result to come. 

3.  I didn't see much of a point in making a totally fake test that doesn't actually touch the triggerdelay part of the codebase and only wanted to do something that was more end to end so I figured dependence injection in the function might be a good idea.


So at this point I have spent  many hours on this and I can probably learn more by trying a few other things so I have a few suggestions
1. Can you recommend a few simple things that I haven't tried that might make this work for me?
2. IF you have time would you be interested in pairing on this?

arafatkatze:
@dominiccooney Update the test instructions and the PR description and added a video too.

vscode/CHANGELOG.md
```diff
@@ -9,7 +9,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 ### Fixed
 
 ### Changed
-
+- Autocomplete Trigger Delay: Introduced a configurable setting to add a delay before returning autocomplete results, enhancing user control over completion suggestion timing. [pull/5350](https://github.com/sourcegraph/cody/pull/5350)
```
valerybugakov, line 12: 🙌

vscode/src/completions/inline-completion-item-provider.ts
valerybugakov: Let's pair! I sent you a DM to find a time slot for that.

arafatkatze: I figured out how to make the tests for this thanks to the learnings from your call.

vscode/src/completions/inline-completion-item-provider.test.ts
```diff
@@ -120,6 +121,51 @@ describe('InlineCompletionItemProvider', () => {
...
+        // Switch to fake timers for precise control
+        vi.useFakeTimers()
+
+        // Advance time, but not enough for the completion to be considered visible
+        vi.advanceTimersByTime(500)
+        expect(spy).toHaveBeenCalledTimes(0)
+
+        // Advance time to make the completion visible (total 750ms)
+        vi.advanceTimersByTime(250)
+        CompletionLogger.logSuggestionEvents(true)
```
dominiccooney, line 163: Do you need to call this here, and if so, is your `expect(spy).toHaveBeenCalledTimes(0)` on line 159 *really* testing anything?

dominiccooney (commented):
This looks good to me. One quick question about the unit test inline.

I have another philosophical question... There's another way we could do this, which is to kick off the autocomplete request but put the *display* behind the time gate. If we did it that way, not only would we not display completions too fast, we're more likely to display completions at after the delay. The way this is structured, you have the delay + random variance of the rest of the completion machinery. Just blocking displaying the candidate would cut off the left hand side of the random variance and you'd have more predictable timing... The user in CODY-3331 is worried about predictability, not network traffic.

@valerybugakov, @arafatkatze, @aramaraju what do you think?

aramaraju:
> This looks good to me. One quick question about the unit test inline.
> 
> I have another philosophical question... There's another way we could do this, which is to kick off the autocomplete request but put the _display_ behind the time gate. If we did it that way, not only would we not display completions too fast, we're more likely to display completions "right after" the delay. The way this is structured, you have the delay + random variance of the rest of the completion machinery. Just blocking displaying the candidate would cut off the left hand side of the random variance and you'd have more predictable timing... The user in [CODY-3331](https://linear.app/sourcegraph/issue/CODY-3331) is worried about predictability, not network traffic.
> 
> @valerybugakov, @arafatkatze, @aramaraju what do you think?

I prefer Dom's approach to this, so we don't add another subsequent delay on top of whatever value is configured by the user. I am also still in general not bought on the idea of adding another setting to the Cody client settings. It's far too convoluted there already, if this is simply for this user, we should "hide" it somehow.

vscode/src/completions/inline-completion-item-provider.test.ts
arafatkatze: I morphed the test to have these lines 

     ```
   // Check if telemetry event is recorded
        CompletionLogger.logSuggestionEvents(true)
        expect(spy).toHaveBeenCalledTimes(1)
        expect(spy).toHaveBeenCalledWith(
            'cody.completion',
            'suggested',
            expect.objectContaining({ metadata: expect.objectContaining({ read: 0 }) })
        )
```

So that the spy is also done on the logger(just to be more thorough).


` expect(spy).toHaveBeenCalledTimes(0)` is just to assert that it takes a while for the codebase to hit that point(because of the completion logic) so it would be nice to show a before and after as you advance the fake timer. 


Hopefully I am understanding the intent of your question correctly but feel free to ask more clarification if you like,

arafatkatze:
@dominiccooney @aramaraju Yes, I agree with the suggestion made so I added the trigger delay such that the trigger delay ensures a minimum wait time before showing autocomplete suggestions.  

> It's far too convoluted there already, if this is simply for this user, we should "hide" it somehow.
Lemme look into it.

valerybugakov:
> Just blocking displaying the candidate would cut off the left hand side of the random variance and you'd have more predictable timing...

Agreed that this is the ideal behavior.

valerybugakov (approved):
Great work, Ara! Unit tests figured out ✅.

PR #5363, opened 2024-08-28
Chat: fix bug where the "new chat" could crash the agent

olafurpg:
Fixes CODY-3510

While testing the Eclipse plugin, I was able to get the webview into a permanently broken state with the reproduction:

- Send message `@Hello.java explain`
- Press "New Chat" while it's streaming the reply

The root cause was that we threw an exception when trying to store the chat history with an error message. This PR works around the issue by not storing messages with errors. However, I'm not 100% satisfied with the fix because

- The process shouldn't permanently error if an exception is thrown
- I wasn't able to reproduce this in VS Code, only Eclipse and JetBrains. This indicates we're not 100% reproducing the VCS environment.

Either way, this should be a low-risk fix to merge.

## Test plan

Run reproduction steps in Eclipse.

### DIFF
--- a/vscode/src/chat/chat-view/ChatHistoryManager.ts
+++ b/vscode/src/chat/chat-view/ChatHistoryManager.ts
@@ -34,9 +34,12 @@ export class ChatHistoryManager implements vscode.Disposable {
 
     public async saveChat(
         authStatus: AuthStatus,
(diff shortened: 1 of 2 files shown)


pkukielka (approved):
LGTM

PR #5367, opened 2024-08-28
Improve Tabs UI layout for mid-size and small container width

vovakulikov:
Part of https://linear.app/sourcegraph/issue/SRCH-810/make-updates-to-cody-chat-ui-as-per-taiyabs-designs

This is the first PR of polishing Chat UI before the Sep Cody Web release; changes come from @taiyab's suggestions to make UI more generic and adaptive for Cody Web and other consumers who use Chat UI as a sharable webview.

This PR 
- Fixes mention icon for the remote files openctx (it was probably by mistake folder, now it's a plain file icon)
- Improves Tabs UI layout for different container widths (prior it was using media query which didn't solve problems with layout for Cody Web and JetBrains integration webview) now it renders tabs and their sub-actions in one row if there is enough space, then two rows if there is not enough space and one row with no labels for small container width 
- Fixes spacing for the welcome message collapsable panels 

## Test plan
- Manual testing on Cody Chat UI 
- I recorded a quick demo video with the new tabs' ui layout [here](https://www.loom.com/share/637425cd65e142e696fc5d8f19ea71c9?sid=5f6ce8df-b837-4bda-80c8-31157409291c)

### DIFF
--- a/lib/prompt-editor/src/mentions/mentionMenu/MentionMenuItem.tsx
+++ b/lib/prompt-editor/src/mentions/mentionMenu/MentionMenuItem.tsx
@@ -143,7 +143,7 @@ export const iconForProvider: Record<
     'https://openctx.org/npm/@openctx/provider-sourcegraph-search': SourcegraphLogo,
     'internal-linear-issues': LinearLogo, // Can't import LinearIssuesProvider due to transitive dep on vscode.
     [REMOTE_REPOSITORY_PROVIDER_URI]: FolderGitIcon,
-    [REMOTE_FILE_PROVIDER_URI]: FolderGitIcon,
+    [REMOTE_FILE_PROVIDER_URI]: FileIcon,
     [REMOTE_DIRECTORY_PROVIDER_URI]: FolderGitIcon,
     [CURRENT_REPOSITORY_DIRECTORY_PROVIDER_URI]: FolderGitIcon,
     [WEB_PROVIDER_URI]: LinkIcon,
--- a/vscode/webviews/chat/components/WelcomeMessage.tsx
+++ b/vscode/webviews/chat/components/WelcomeMessage.tsx
@@ -53,11 +53,11 @@ export const WelcomeMessage: FunctionComponent<{ IDE: CodyIDE; setView: (view: V
     const dispatchClientAction = useClientActionDispatcher()
 
     return (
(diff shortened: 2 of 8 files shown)


taiyab:
Thanks for this @vovakulikov!

Trying this out locally:

The New Chat button is a top-level item, hence it should have different behaviour to the sub-options of history.

- For the home tab: I'd rather it all always stay on one row, so hide the labels earlier and prioritise always showing the `New Chat` button as prominently as possible
- We should always drop the `Export History and `Clear Chat History` options to the next row
- Hide the label for "Open in Editor" always, only show a tooltip
- The tooltip for the "New Chat" button doesn't seem to work at all
- We should bump up the [toolbar text size](https://github.com/user-attachments/assets/4ec64c87-3286-4c9d-b258-8239509b23bd)
- We should [truncate prompt titles passed one line](https://github.com/user-attachments/assets/2bf5092f-b00b-48ee-9171-4a21cc794a31) 

The thing that will make the biggest impact here is splitting out the commands window and restyling it a bit, who do we need to get involved to help do that?

taiyab (commented):
Design-wise: 👍 some good steps forward, I've left a comment above for the next PRs.

vovakulikov:
@taiyab 

> For the home tab: I'd rather it all always stay on one row, so hide the labels earlier and prioritize always showing the New Chat button as prominently as possible

> We should always drop the Export History and Clear Chat History options to the next row

For now, I think to avoid inconsistent behavior between tabs I would like to make the logic the same for all tabs, but I hear what you're saying I think we will change this logic in the future anyway to make it by content width (if there is enough space one row if not two rows) 

> The tooltip for the "New Chat" button doesn't seem to work at all

Hmm it works for me, interesting why this is different for your build 🤔 
Screenshot 2024-08-28 at 19 59 56

> We should bump up the [toolbar text size](https://github.com/user-attachments/assets/4ec64c87-3286-4c9d-b258-8239509b23bd)
> We should [truncate prompt titles passed one line](https://github.com/user-attachments/assets/2bf5092f-b00b-48ee-9171-4a21cc794a31)

Yes but it will be implemented in a separate PR, I have it on my list

jamesmcnamara (approved):
This looks good to me but I'm no expert on the web UI.

PR #5371, opened 2024-08-28
Make textbox partially sticky to the bottom

vovakulikov:
Part of SRCH-810

This PR improves the search box position based on @taiyab suggestion. The logic is 
- When you submit your very first message the next textbox sticks to the bottom of the container 
- And it stays there if the content doesn't push it below 
- If there is enough content it pushes the textbox further 
- So the last textbox is always at the bottom of the container 

Note that this logic isn't standard sticky position; you still can scroll the whole container, and the textbox will scroll along with the content. 

The reason for this is that it is less jumpy at the beginning with your first message is resolving (textbox is in one place) but it doesn't take additional space from the actual content because it scrolls with the content

| Before | After |
| ------ | ------ | 
| <video src="https://github.com/user-attachments/assets/f683e2d0-30e8-49aa-a4e0-c98fdab40c09" />  | <video src="https://github.com/user-attachments/assets/fc89108c-76c6-4179-a6e9-850f6efb9c54" /> |

## Test plan
- VSCode Cody 
- Check that after you submit the message the next search box is sticky to the bottom of the chat container
- But this textbox is still scrollable if the content makes the whole block scroll

### DIFF
--- a/vscode/webviews/chat/cells/messageCell/human/editor/toolbar/Toolbar.tsx
+++ b/vscode/webviews/chat/cells/messageCell/human/editor/toolbar/Toolbar.tsx
@@ -90,7 +90,7 @@ export const Toolbar: FunctionComponent<{
                     className="tw-mr-1"
                 />
             </div>
-            <div className="tw-flex-1 tw-text-right">
+            <div className="tw-flex tw-justify-end tw-flex-1 tw-text-right">
                 <SubmitButton
                     onClick={onSubmitClick}
                     isEditorFocused={isEditorFocused}
--- a/vscode/webviews/components/shadcn/ui/toolbar.module.css
+++ b/vscode/webviews/components/shadcn/ui/toolbar.module.css
@@ -1,5 +1,5 @@
 :root {
-    --toolbar-button-font-size: 11px;
+    --toolbar-button-font-size: 12px;
 }
 
 .button {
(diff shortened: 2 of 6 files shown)


vscode/webviews/components/shadcn/ui/toolbar.module.css
```diff
@@ -1,5 +1,5 @@
 :root {
-    --toolbar-button-font-size: 11px;
+    --toolbar-button-font-size: 12px;
```
vovakulikov, line 2: Increase base font-size for toolbar also by suggestion from @taiyab

vscode/webviews/chat/cells/messageCell/human/editor/toolbar/Toolbar.tsx
```diff
@@ -90,7 +90,7 @@ export const Toolbar: FunctionComponent<{
                     className="tw-mr-1"
                 />
             </div>
-            <div className="tw-flex-1 tw-text-right">
+            <div className="tw-flex tw-justify-end tw-flex-1 tw-text-right">
```
vovakulikov, line 93: This fixes a problem with jumping loading the submit button, we didn't notice this before because the whole textbox was jumping but there was a problem with loading-submit button spreading the toolbar in the textbox a little bit

sqs (approved):

PR #5378, opened 2024-08-28
repro the hotstreak issue

hitesh-1997:
## Context
- When the generation is same as the suffix, the hot-streak logic trims the first correct. Adding the reproducer of the issue in the unit test.
- Closes https://linear.app/sourcegraph/issue/CODY-3542/autocomplete-suggestion-post-processing-check-for-additional-trimming

## Test plan
1. Added unit test to reproduce the issue

### DIFF
--- a/vscode/src/completions/providers/fetch-and-process-completions.ts
+++ b/vscode/src/completions/providers/fetch-and-process-completions.ts
@@ -75,12 +75,16 @@ export async function* fetchAndProcessDynamicMultilineCompletions(
             },
         }
 
-        hotStreakExtractor = createHotStreakExtractor({
-            completedCompletion,
-            ...params,
-        })
+        if (completedCompletion.insertText.length > 0) {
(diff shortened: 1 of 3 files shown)


vscode/src/completions/providers/fetch-and-process-completions.ts
```diff
@@ -75,12 +75,16 @@ export async function* fetchAndProcessDynamicMultilineCompletions(
...
+            hotStreakExtractor = createHotStreakExtractor({
+                completedCompletion,
+                ...params,
+            })
 
-        yield* hotStreakExtractor.extract(rawCompletion, isFullResponse)
+            yield* hotStreakExtractor.extract(rawCompletion, isFullResponse)
+        } else {
+            abortController.abort()
+        }
```
valerybugakov, line 87: Here's the fix part

valerybugakov (approved):

PR #5380, opened 2024-08-29
Autocomplete: fix Anthropic model for PLG users

valerybugakov:
- This fixes the issue of using the `codeCompletion` model from site configuration with the Anthropic provider for PLG users. In the current state of things, we should always default to the `DEFAULT_PLG_ANTHROPIC_MODEL` on DotCom.
- Closes https://github.com/sourcegraph/cody/issues/4639
- Follow up: we should deprecate `cody.autocomplete.advanced.*` settings to avoid such cases and ensure that Cody users always use the recommended setup. 

## Test plan

Updated unit tests to account for this case.


dominiccooney (approved):
LGTM

PR #5382, opened 2024-08-29
remove unused code with knip

sqs:
`pnpm exec knip --fix`

No behavior change, just a dead code removal.

## Test plan

CI


sqs:
Stamp pls!

varungandhi-src (approved):

varungandhi-src:
@sqs should this be added as a CI check so that people don't accidentally commit dead code?

sqs:
@varungandhi-src Perhaps in the future. I added most of the manual exceptions (where we want to keep some dead code around) to the knip.jsonc file, but it's not totally automatable yet. And there wasn't too much dead code around, so it's not a big problem.

PR #5385, opened 2024-08-29
Add tests for provider.ts

mkondratek:
## Test plan
- green ci

## Changelog
Extract `deriveNewFileUri` from `handleFileCreationResponse`.


vscode/src/edit/provider.test.ts
```diff
@@ -0,0 +1,75 @@
...
+        joinPath: vi
+            .fn()
+            .mockImplementation((base: Uri, path: string) => ({ path: `${base.path}/${path}` })),
+    },
+}))
+
+describe('deriveNewFileUri', () => {
+    it('should derive new file URI with workspace URI', () => {
+        const currentFileUri = Uri.parse('/workspace/project/file.txt')
+        const text = '<filename>newfile.txt</filename>'
```
pkukielka, line 26: How is that supposed to work  when filename tag is in fact `TESTFILE7041`?
I think let's just use `PROMPT_TOPICS.FILENAME` (you can always assign it to local variable if you want to avoid too much verbosity).

vscode/src/edit/provider.ts
```diff
@@ -323,8 +315,28 @@ export class EditProvider {
...
+
+    // remove open and close tags from text
+    const newFilePath = text.trim().replaceAll(new RegExp(`${opentag}(.*)${closetag}`, 'g'), '$1')
+
+    // Get workspace uri using the current file uri
+    const workspaceUri = workspace.getWorkspaceFolder(currentFileUri)?.uri
+    const currentDirUri = Utils.joinPath(currentFileUri, '..')
+
+    // Create a new file uri by replacing the file name of the currentFileUri with fileName
+    return Utils.joinPath(workspaceUri ?? currentDirUri, newFilePath)
```
pkukielka, line 341: This is clearly a potential bug.
It can work correctly only as long as `newFilePath` is relative path... but we never check that.

vscode/src/edit/provider.test.ts
mkondratek: fixed

vscode/src/edit/provider.ts
pkukielka: Cool, let's add a test for this case and fix it :)

mkondratek:
closing in favor of https://github.com/sourcegraph/cody/pull/5392

PR #5386, opened 2024-08-29
Focus textbox after we submit a follow-up message

vovakulikov:
Part of SRCH-810

This PR updates scroll-down button styles and makes that after you send a follow-up message we focus and scroll to view the last textbox element

Visual change has done by @taiyab designs 

| Before | After |
| ----- | ------ |
| Screenshot 2024-08-29 at 15 36 39  | Screenshot 2024-08-29 at 15 36 30 |

## Test plan
- Manual checks on Cody Chat UI

### DIFF
--- a/vscode/webviews/components/ScrollDown.tsx
+++ b/vscode/webviews/components/ScrollDown.tsx
@@ -5,7 +5,7 @@ import { Button } from './shadcn/ui/button'
 const MARGIN = 200 /* px */
 
 interface Scroller {
-    root: HTMLElement | Window
+    root: HTMLElement
     getObserveElement: () => Element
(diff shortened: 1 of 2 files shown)


vscode/webviews/components/ScrollDown.tsx
```diff
@@ -5,7 +5,7 @@ import { Button } from './shadcn/ui/button'
 const MARGIN = 200 /* px */
 
 interface Scroller {
-    root: HTMLElement | Window
```
vovakulikov, line 8: Window is never a case anymore since we render everything within tab UI container

thenamankumar (approved):

PR #5389, opened 2024-08-29
Cody Web: Fix Cody Web wrapper in dev mode

vovakulikov:
I've noticed that Cody Web wrappers over cody agent work with race conditions sometimes, this happens because in dev mode react fires useEffect twice and this breaks code agent internal states and handling order (twice init actions case clears active panel IDs for example)

This PR adds init flags to avoid this problem when you run pnpm dev; it should have no effect on the production build, though. 

## Test plan
- Run Cody Web demo and check that you still can submit the message

### DIFF
--- a/web/lib/components/use-cody-agent.ts
+++ b/web/lib/components/use-cody-agent.ts
@@ -2,10 +2,11 @@ import { hydrateAfterPostMessage, isErrorLike } from '@sourcegraph/cody-shared'
 import type { ExtensionMessage } from 'cody-ai/src/chat/protocol'
 import { type VSCodeWrapper, setVSCodeWrapper } from 'cody-ai/webviews/utils/VSCodeApi'
 import {
+    type DependencyList,
+    type EffectCallback,
     type MutableRefObject,
     useCallback,
     useEffect,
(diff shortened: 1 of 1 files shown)


web/lib/components/use-cody-agent.ts
```diff
@@ -114,21 +115,22 @@ export function useCodyWebAgent(input: UseCodyWebAgentInput): UseCodyWebAgentRes
         [initialContext]
     )
 
+    const isInitRef = useRef(false)
     const vscodeAPI = useVSCodeAPI({ activeWebviewPanelIDRef, createNewChat, client })
 
     // Always create new chat when Cody Web is opened for the first time
     useEffect(() => {
+        // Skip panel creation if it already happened before
```
thenamankumar, line 123: why not use useEffectOnce here as well?

thenamankumar (approved):

web/lib/components/use-cody-agent.ts
vovakulikov: good question, this is because in this particular case, we should not count runs when the client is null or error and set init to true only when we get to running `createNewChat` function

PR #5390, opened 2024-08-29
Add zero state for history tab

vovakulikov:
Part of SRCH-810

This PR adds zero state for the history tab by @taiyab designs listed [here](https://www.figma.com/design/f078wFMKsIOaEwj7Iwj5xy/⬜%EF%B8%8F--Unified-Cody?node-id=5081-155750&node-type=FRAME&t=E311cWtgz19OTYip-0) 

Note that colors might be different compared to what we have in Figma since Figma designs operate with updated design components (we will update our component implementation after the release), for now, I used the standard UI that we have in the current cody chat implementation

Screenshot 2024-08-29 at 17 10 45

## Test plan
- Go to the history tab
- Remove all chats
- Check that zero state look good and start a new chat button works as expected

### DIFF
--- a/vscode/webviews/CodyPanel.tsx
+++ b/vscode/webviews/CodyPanel.tsx
@@ -95,7 +95,15 @@ export const CodyPanel: FunctionComponent<
                         setView={setView}
                     />
                 )}
-                {view === View.History && <HistoryTab userHistory={userHistory} />}
+                {view === View.History && (
+                    <HistoryTab
+                        IDE={config.agentIDE || CodyIDE.VSCode}
+                        setView={setView}
+                        webviewType={config.webviewType}
+                        multipleWebviewsEnabled={config.multipleWebviewsEnabled}
+                        userHistory={userHistory}
+                    />
+                )}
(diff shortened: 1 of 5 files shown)


vscode/webviews/CodyPanel.tsx
```diff
@@ -95,7 +95,15 @@ export const CodyPanel: FunctionComponent<
                         setView={setView}
                     />
                 )}
-                {view === View.History && <HistoryTab userHistory={userHistory} />}
+                {view === View.History && (
+                    <HistoryTab
+                        IDE={config.agentIDE || CodyIDE.VSCode}
```
thenamankumar, line 100: Can we remove all this complexity from the HistoryTab and make it receive a `onStartNewChat` argument. and then  also extract 
``` getVSCodeAPI().postMessage({
            command: 'command',
            id: getCreateNewChatCommand({ IDE, webviewType, multipleWebviewsEnabled }),
        })
```
 into `useStartNewChat` hook?

thenamankumar (approved):

PR #5399, opened 2024-08-30
Autocomplete: simplify provider creation

valerybugakov:
- Simplified the way we create autocomplete providers:
    - Specified a list of supported autocomplete provider IDs with comments explaining the context behind each of them.
    - Reduced code duplication by extracting common logic from all autocomplete providers into an abstract class.
    - Streamlined provider instantiation by eliminating the intermediate provider configuration structure.
    - Standardized the interface and format across all provider factory functions, ensuring consistency.
    - Converted autocomplete HTTP client provider into a singleton to avoid propagating it to each autocomplete provider.
    - Renamed several properties to be consistent and self-explanatory across the codebase.
- No functional changes.

## Test plan

1. CI
2. Manually tested singleline and multiline autocomplete works on DotCom, sg02, and sg04.


vscode/src/completions/providers/create-provider.ts
```diff
@@ -71,119 +59,107 @@ export async function createProviderConfig(
...
+            anonymousUserID,
+            provider,
+        })
     }
+
+    return null
+}
+
+interface GetProviderCreatorParams {
+    provider: string
```
taras-yemets, line 93: What do you think about using a string union type here? It’ll make the supported providers clearer.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -71,119 +59,107 @@ export async function createProviderConfig(
...
+    authStatus: AuthStatus
+}
+
+function getProviderCreator(params: GetProviderCreatorParams): ProviderFactory | null {
```
taras-yemets, line 97: Great call on extracting the provider creator into a separate function! The comments you added are super helpful too. 

One suggestion: switching to a `switch` statement with `provider` as a string union might give us a `never` type at the end, which helps ensure all known providers have a matching `createProvider` function.

vscode/src/completions/providers/create-provider.ts
```diff
@@ -1,35 +1,25 @@
...
 import { logError } from '../../log'
 import { localStorage } from '../../services/LocalStorageProvider'
-import {
-    DEFAULT_PLG_ANTHROPIC_MODEL,
-    createProviderConfig as createAnthropicProviderConfig,
-} from './anthropic'
-import { createProviderConfig as createExperimentalOllamaProviderConfig } from './experimental-ollama'
-import { createProviderConfig as createExperimentalOpenAICompatibleProviderConfig } from './expopenaicompatible'
-import { createProviderConfig as createFireworksProviderConfig } from './fireworks'
+import { createProvider as createAnthropicProviderConfig } from './anthropic'
```
taras-yemets, line 5:
```suggestion
import { createProvider as createAnthropicProvider } from './anthropic'
```

vscode/src/completions/providers/unstable-openai.ts
```diff
@@ -162,55 +142,55 @@ ${OPENING_CODE_TAG}${infillBlock}`
...
+            }
+
+            // Remove bad symbols from the start of the completion string.
+            completion = fixBadCompletionStart(completion)
+
+            return completion
         }
+}
+
+export function createProvider(params: ProviderFactoryParams): Provider {
```
taras-yemets, line 176: Great job simplifying the provider creation!

I’m a fan of early returns since it puts the provider definition right after the `if` clause. Just a personal preference, so no worries if you prefer otherwise.

One suggestion: what do you think about making `ProviderFactoryParams` generic? This would allow `createProvider` to handle all the cases for each provider type. For example:
```
type Provider = 'openai' | 'unstable-openai' | 'azure-openai' | 'google' // ...rest from getProviderCreator

export type ProviderFactoryParams<T extends Provider> = {
  provider: T
  // ...other fields
}

// vscode/src/completions/providers/unstable-openai.ts
// cover openai cases from getProviderCreator
export function createProvider(params: ProviderFactoryParams<'openai' | 'unstable-openai' | 'azure-openai'>): Provider {
  // implementation
}
```

vscode/src/completions/get-inline-completions.ts
```diff
@@ -63,6 +68,7 @@ export interface InlineCompletionsParams {
     tracer?: (data: Partial<ProvideInlineCompletionsItemTraceData>) => void
     artificialDelay?: number
     firstCompletionTimeout: number
+    n?: number
```
taras-yemets, line 71: Does it represent the number of completions to generate? A comment or a more descriptive name would be helpful.

vscode/src/completions/providers/expopenaicompatible.ts
```diff
@@ -415,3 +339,28 @@ function promptString(prompt: Prompt, infill: boolean, model: string): PromptStr
 
     return context.concat(currentFileNameComment, prompt.prefix)
 }
+
+function getClientModel(isDotCom: boolean, model?: string): OpenAICompatibleModel {
```
taras-yemets, line 343: `isDotCom` param is not used. Can we remove it?

vscode/src/completions/providers/google.ts
```diff
@@ -193,35 +166,17 @@ Your response should contains only the code required to connect the gap, and the
...
-    model,
-    maxContextTokens = 2048,
-    ...otherOptions
-}: GoogleGeminiOptions & { model?: string }): ProviderConfig {
-    if (!model) {
-        model = DEFAULT_GEMINI_MODEL
-    }
+export function createProvider(params: ProviderFactoryParams): Provider {
+    const { legacyModel, anonymousUserID } = params
+    const clientModel = legacyModel ?? DEFAULT_GEMINI_MODEL
```
taras-yemets, line 171: What format of `clientModel` do we expect here? Is it "provider/model" or just "model"?

vscode/src/completions/providers/provider.ts
```diff
@@ -84,12 +52,101 @@ export interface ProviderOptions {
...
+    legacyModel: string
+}
+
+export type ProviderOptions = (ProviderModelOptions | ProviderLegacyModelOptions) & {
+    id: string
+    anonymousUserID: string
+    /**
+     * Defaults to `DEFAULT_MAX_CONTEXT_TOKENS`
+     */
+    maxContextTokens?: number
```
taras-yemets, line 84: Very helpful comment! Thank you!

vscode/src/completions/providers/provider.ts
```diff
@@ -84,12 +52,101 @@ export interface ProviderOptions {
...
+export type ProviderFactory = (params: ProviderFactoryParams) => Provider
+
 export abstract class Provider {
-    constructor(public readonly options: Readonly<ProviderOptions>) {}
+    /**
+     * A unique and descriptive identifier for the provider.
+     */
+    public id: string
+    public model?: Model
+    public legacyModel: string
```
taras-yemets, line 107: Adding some examples in comment would be very hepful!

taras-yemets (approved):
Great work! This PR really streamlines provider creation and boosts readability. Thanks a lot!

vscode/src/completions/providers/create-provider.ts
valerybugakov: I kept the `switch` statement initial and switched to if statements to support this case:

```ts
(provider === AUTOCOMPLETE_PROVIDER_ID.google &&
            authStatus.configOverwrites?.completionModel?.includes('claude'))
```

I want to have one return statement for each provider creator for clarity. With the switch statement, we had to duplicate it for `createAnthropicProviderConfig`.

I updated the `getProviderCreator` to use a string union, which results in a `never` type at the end of the function.

vscode/src/completions/providers/unstable-openai.ts
valerybugakov: I like the idea! I will look into it in one of the follow-ups because it affects many call sites if we want to utilize this generic properly.

vscode/src/completions/get-inline-completions.ts
valerybugakov: Update the property name 👍

vscode/src/completions/providers/google.ts
valerybugakov: Based on the default value, we expect `provider/model`. I added a TODO to unify the `legacyModel` format to one syntax across providers.

hitesh-1997 (approved):
Nice !!

PR #5400, opened 2024-08-30
Fix floating ui elements (tooltips and popovers)

vovakulikov:
Prior to this PR tabs UI tooltips were using just media query which doesn't work in the case of sharable webview where the client usually has a bigger viewport than actual cody chat UI sizes. 

This PR basically uses a container query to decide when we should show tooltips for tabs UI.

## Test plan
- Manual checks on Cody Chat UI

### DIFF
--- a/lib/prompt-editor/src/nodes/MentionComponent.tsx
+++ b/lib/prompt-editor/src/nodes/MentionComponent.tsx
@@ -1,6 +1,7 @@
 import { useLexicalComposerContext } from '@lexical/react/LexicalComposerContext'
 import { useLexicalNodeSelection } from '@lexical/react/useLexicalNodeSelection'
 import { mergeRegister } from '@lexical/utils'
+import { clsx } from 'clsx'
 import {
     $getNodeByKey,
     $getSelection,
@@ -196,7 +197,7 @@ export const MentionComponent: FunctionComponent<{
 
     const content = (
         <span ref={ref} className={composedClassNames} title={tooltipComponents ? undefined : tooltip}>
-            {Icon && <Icon size={14} strokeWidth={2} className={iconClassName} />}
+            {Icon && <Icon size={14} strokeWidth={2} className={clsx(iconClassName, 'tw-shrink-0')} />}
             <span>{text}</span>
         </span>
     )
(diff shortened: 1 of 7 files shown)


lib/prompt-editor/src/nodes/MentionComponent.tsx
```diff
@@ -196,7 +197,7 @@ export const MentionComponent: FunctionComponent<{
 
     const content = (
         <span ref={ref} className={composedClassNames} title={tooltipComponents ? undefined : tooltip}>
-            {Icon && <Icon size={14} strokeWidth={2} className={iconClassName} />}
+            {Icon && <Icon size={14} strokeWidth={2} className={clsx(iconClassName, 'tw-shrink-0')} />}
```
vovakulikov, line 200: Unrelated to this PR problem, if the mention is shrinking we shouldn't shrink the icon (don't want to create a separate PR for such small fix)

vscode/webviews/chat/cells/Cell.tsx
```diff
@@ -1,36 +1,43 @@
...
-        role="row"
-        aria-current={ariaCurrent}
-        aria-disabled={ariaDisabled}
-        data-testid={dataTestID}
-    >
-        <header className="tw-flex tw-gap-4 tw-items-center [&_>_*]:tw-flex-shrink-0">{header}</header>
-        <div className={clsx('tw-flex-1 tw-overflow-hidden', contentClassName)}>{children}</div>
-    </div>
-)
+export const Cell = forwardRef<HTMLDivElement, PropsWithChildren<CellProps>>((props, ref) => {
```
vovakulikov, line 17: Pure dev mode change, with no ref it throws a react warning because slot clone can't handle this element properly

vscode/webviews/App.module.css
```diff
@@ -5,7 +5,6 @@
     box-sizing: border-box;
     height: 100%;
     overflow: hidden;
-    isolation: isolate;
```
vovakulikov, line 8: This is no needed anymore and it breaks floating UI within chat UI dom tree

abeatrix (approved):

PR #5407, opened 2024-08-31
fixed(chat): Added a migration to local storage to update chat history (CODY-3538)

jamesmcnamara:
There was a bug in https://github.com/sourcegraph/jetbrains/pull/2108 where we imported chat's with ID's but the app actually expected them to be UTC date strings. This caused some errors where it tried to do time grouping on the UUID's and if those conversations were continued, they would actually lose all of their timestamps.
image

This adds a migration that will be run once at startup to try to detect and patch any corrupted conversations. Additionally this PR should fix the migration for newer users.

## Test plan
Unit Tests and manually ran in JB

## Changelog
Fixed a bug in continuing conversations from JB in the webview that caused them to be displayed as NaN weeks ago


pkukielka (approved):
LGTM

PR #5412, opened 2024-08-31
use higher-level API for chat models in webview

sqs:
Previously, the webview and extension host communicated about the models with the following API methods:

- `get-chat-models`: webview asking the exthost to refresh the chat models (by sending a `chatModels` message)
- `chatModel`: webview telling the exthost that the user selected a chat model in the dropdown
- `chatModels`: exthost telling the webview the list of available models

Now, there are 2 API methods that can return errors and stream results:

- `models`: webview asking the exthost to observe the list of chat models (i.e., it streams in new results when the list changes)
- `setChatModel`: webview telling the exthost that the user selected a chat model in the dropdown

These 2 APIs use the same new Observable RPC layer that is used for `@`-mention items, prompts, and feature flags.

Also removes some usage of `webview/receiveMessage` in the agent API and adds a higher-level `chat/setModel`.

## Test plan

CI and e2e


lib/shared/src/misc/rpc/webviewAPI.ts
```diff
@@ -24,6 +26,28 @@ export interface WebviewToExtensionAPI {
      * the Prompt Library).
      */
     prompts(query: string): Observable<PromptsResult>
+
+    /**
+     * Observe the list of available models.
+     */
+    models(): Observable<Model[]>
```
olafurpg, line 33: So nice <3

olafurpg (approved):
Great cleanup! It's probably time to move the agent tests and CLI to some higher level chat API so we can stop using the webview/* notifications

PR #5418, opened 2024-09-01
Unify prompts and commands in the chat UI

sqs:
WIP implementing https://www.figma.com/design/f078wFMKsIOaEwj7Iwj5xy/%E2%AC%9C%EF%B8%8F--Unified-Cody?node-id=4426-93729&node-type=CANVAS&t=yxXwA2teMRN6rFbz-0 to unify prompts and commands

image


sqs:
@vovakulikov merged this in https://github.com/sourcegraph/cody/pull/5438.

@vovakulikov was there anything in this PR that you didn't adopt that we should still merge in the future?

vovakulikov:
> @vovakulikov was there anything in this PR that you didn't adopt that we should still merge in the future?
@sqs I don't think so, but my PR was reverted so it's still on the table to make this change. But I think I will handle it after the release and rever my reverted PR and add missing things like empty zero states

PR #5428, opened 2024-09-02
Fix account switching and chat history issue

mkondratek:
Fixes https://github.com/sourcegraph/jetbrains/issues/2173.

## Test plan
See the issue: https://github.com/sourcegraph/jetbrains/issues/2173

1. log in to an account 
2. write something in chat
3. switch account
4. review the chat history
expected:
The chat from the first account should not appear in the account switched into.

### DIFF
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -1511,7 +1511,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
 
     private async saveSession(authStatus?: AuthStatus): Promise<void> {
         const allHistory = await chatHistory.saveChat(
-            authStatus ? authStatus : this.authProvider.getAuthStatus(),
+            authStatus || this.authProvider.getAuthStatus(),
             this.chatModel.toSerializedChatTranscript()
         )
         if (allHistory) {
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -588,7 +588,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
     // #region top-level view action handlers
     // =======================================================================
 
-    public setAuthStatus(status: AuthStatus): void {
+    public async setAuthStatus(status: AuthStatus): Promise<void> {
         // Run this async because this method may be called during initialization
         // and awaiting on this.postMessage may result in a deadlock
         void this.sendConfig()
@@ -596,6 +596,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
         // Get the latest model list available to the current user to update the ChatModel.
         if (status.isLoggedIn) {
             this.handleSetChatModel(getDefaultModelID())
+            return await this.saveSession()
         }
     }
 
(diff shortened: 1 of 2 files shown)


vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -500,6 +494,7 @@ export class ChatsController implements vscode.Disposable {
             extensionClient: this.extensionClient,
         })
     }
+    play
```
valerybugakov, line 497: Redundant? 
```suggestion
```

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -1508,9 +1509,9 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
         this.postViewTranscript()
     }
 
-    private async saveSession(): Promise<void> {
+    private async saveSession(authStatus?: AuthStatus): Promise<void> {
         const allHistory = await chatHistory.saveChat(
-            this.authProvider.getAuthStatus(),
+            authStatus ? authStatus : this.authProvider.getAuthStatus(),
```
valerybugakov, line 1514:
```suggestion
            authStatus || this.authProvider.getAuthStatus(),
```

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -588,14 +588,15 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
     // #region top-level view action handlers
     // =======================================================================
 
-    public setAuthStatus(status: AuthStatus): void {
+    public async setAuthStatus(status: AuthStatus): Promise<void> {
         // Run this async because this method may be called during initialization
         // and awaiting on this.postMessage may result in a deadlock
         void this.sendConfig()
```
valerybugakov, line 594: Should we worry about a deadlock mentioned in the comment while we add `await` to this function?

valerybugakov (approved):
Verified that the fix works based on the test plan. 

Would it be more resilient to keep one source of truth for the current `authStatus` by disposing all chats _before_ changing the value of `this.authProvider.getAuthStatus()`? If we do all the cleanup tasks before setting the new value, all the code paths can still rely exclusively on `this.authProvider.getAuthStatus()`, which makes the implementation more robust. 

I might be missing something obvious here, so I approve the fix.

vscode/src/chat/chat-view/ChatController.ts
mkondratek: I am not 100% sure about it but I think the comment is about `void this.sendConfig()` call.

mkondratek: Looks like some old leftovers: https://github.com/sourcegraph/cody/blob/eabbdcf735592796438e44ba1fdcb3777f5098c3/vscode/src/chat/chat-view/SimpleChatPanelProvider.ts#L518

mkondratek: The comment applies to `postMessage({type: 'config', ...)` - I do not change that. I uphold that my changes should not break anything (at least not in this area 😅)

mkondratek:
> Verified that the fix works based on the test plan.
> 
> Would it be more resilient to keep one source of truth for the current `authStatus` by disposing all chats _before_ changing the value of `this.authProvider.getAuthStatus()`? If we do all the cleanup tasks before setting the new value, all the code paths can still rely exclusively on `this.authProvider.getAuthStatus()`, which makes the implementation more robust.
> 
> I might be missing something obvious here, so I approve the fix.

I agree with you. Presently, `ChatsController.setAuthStatus` does two things - unsets the old auth status (disposes chats) and sets the new auth status (loads chats). It looks like we should split these two operations somehow. However, I do not feel like having enough proficiency in TS to apply the ideal solution. 

Also note that some of our problem are going to be solved once we switch to webview account management (which is our plan afaik).

mkondratek:
Thanks for your input and review @valerybugakov !

pkukielka (approved):
LGTM

PR #5432, opened 2024-09-02
Improve tabs UI for different consumers

vovakulikov:
This PR updates tabs UI based on the latest suggestions from @taiyab 

| Before | After |
| ----- | ------ |
| <video src="https://github.com/user-attachments/assets/f8fdc326-f109-4943-8b93-5da46d87ed7d" /> | <video src="https://github.com/user-attachments/assets/700f417f-6c7e-4d97-b7a6-f4e1a3b0f064" /> |

- Show the "New chat" button unconditionally as global actions on the right side 
- Always show subactions in separate rows below the main tabs row
- Don't hide labels for sub-actions 
- Change label text for sub-actions like export and delete

We still rely on containers query for showing/hiding labels for tab items, but this PR makes breakpoints for Cody Web different (since Cody Web has different tab configurations it makes sense to hide labels with a bit smaller screens)

Also, I've noticed that the container query root element creates some sort of isolation layer that restricts position-fixed elements from visually leaving this container (and as a result, they are completely visually cut off). To avoid this problem, tooltips for tabs UI are rendered outside of the DOM tree with the portal in case of Cody Web review.

## Test plan
- Check tabs UI in VSCode 
- Check tabs UI in Cody Web (demo or full instance check)

### DIFF
--- a/vscode/webviews/chat/Transcript.test.tsx
+++ b/vscode/webviews/chat/Transcript.test.tsx
@@ -33,7 +33,7 @@ function render(element: JSX.Element): ReturnType<typeof render_> {
     return render_(element, { wrapper: AppWrapperForTest })
 }
 
-describe('Transcript', () => {
+describe.only('Transcript', () => {
     test('empty', () => {
         render(<Transcript {...PROPS} transcript={[]} />)
         expectCells([{ message: '' }])
@@ -203,7 +203,7 @@ describe('Transcript', () => {
                 ]}
             />
         )
-        expectCells([{ message: 'Foo' }, { message: 'Model\nRequest Failed: some error' }])
+        expectCells([{ message: 'Foo' }, { message: 'Model\n\nRequest Failed: some error' }])
         expect(screen.queryByText('Try again with different context')).toBeNull()
     })
 
--- a/vscode/webviews/chat/Transcript.test.tsx
+++ b/vscode/webviews/chat/Transcript.test.tsx
@@ -33,7 +33,7 @@ function render(element: JSX.Element): ReturnType<typeof render_> {
     return render_(element, { wrapper: AppWrapperForTest })
 }
 
-describe.only('Transcript', () => {
+describe('Transcript', () => {
     test('empty', () => {
         render(<Transcript {...PROPS} transcript={[]} />)
         expectCells([{ message: '' }])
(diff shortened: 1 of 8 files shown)


vscode/webviews/tabs/TabsBar.module.css
```diff
@@ -48,61 +54,25 @@
...
+            display: none;
+        }
     }
+}
 
-    .tooltip {
-        display: block !important;
+/* Special override for Cody Web tabs */
+@container tabs-container (width < 375px) {
+    .tabs-root--cody-web .tab-action-label {
```
vovakulikov, line 74: container query API doesn't support any dynamic conditions, so I had to go with this specific to cody web class selector here

vscode/webviews/tabs/TabsBar.tsx
```diff
@@ -261,12 +297,12 @@ const TabButton = forwardRef<HTMLButtonElement, TabButtonProps>(
                     <span className={alwaysShowTitle ? '' : styles.tabActionLabel}>{title}</span>
                 </button>
             </TooltipTrigger>
-            <TooltipContent className={styles.tooltip}>
+            <TooltipContent portal={IDE === CodyIDE.Web}>
```
vovakulikov, line 300: Render tooltip here in portal when it Cody Web, to avoid problem with cut off elements in sidebar blob UI panel in Sourcegraph Web app

vscode/webviews/chat/Transcript.test.tsx
```diff
@@ -33,7 +33,7 @@ function render(element: JSX.Element): ReturnType<typeof render_> {
     return render_(element, { wrapper: AppWrapperForTest })
 }
 
-describe('Transcript', () => {
+describe.only('Transcript', () => {
```
thenamankumar, line 36: remove only.

vscode/webviews/tabs/TabsBar.module.css
thenamankumar: If this is cody web specific, should we do it in cody web then?

vovakulikov: @thenamankumar too many moves to do this, and in the end, it would expose too many layout details to the Sourcegraph Web app. 

I will improve this after release to avoid having any container queries here and add dynamic logic for hiding/showing labels, so this cody web specific condition will go away pretty soon

thenamankumar (approved):

PR #5433, opened 2024-09-02
Include initial context from OpenCtx providers. 

thenamankumar:
context: https://linear.app/sourcegraph/issue/PROD-264/one-pager-for-how-to-get-this-set-up-using-openctx

We have [added](https://github.com/sourcegraph/openctx/pull/199) `meta.mentions.autoInclude` property for OpenCtx providers. The providers which set that to `true` will be used to include the mentions they return as initial context items. We are also passing the active file uri and codebase names to the `mentions` function. 

## Test plan
Set the following as the vs code setting and see the hello world mention item chip appearing as the default context item. 
```json
"openctx.providers": {
        "https://openctx.org/npm/@openctx/provider-hello-world": true
}
```

## Changelog

Include initial context from OpenCtx providers.

### DIFF
--- a/vscode/src/chat/clientStateBroadcaster.ts
+++ b/vscode/src/chat/clientStateBroadcaster.ts
@@ -1,5 +1,6 @@
 import {
     type ContextItem,
+    type ContextItemOpenCtx,
     ContextItemSource,
     type ContextItemTree,
     REMOTE_REPOSITORY_PROVIDER_URI,
@@ -7,15 +8,20 @@ import {
     displayLineRange,
     displayPathBasename,
     expandToLineRange,
+    openCtx,
     subscriptionDisposable,
 } from '@sourcegraph/cody-shared'
 import * as vscode from 'vscode'
+import { URI } from 'vscode-uri'
 import { getSelectionOrFileContext } from '../commands/context/selection'
 import { createRepositoryMention } from '../context/openctx/common/get-repository-mentions'
 import { workspaceReposMonitor } from '../repository/repo-metadata-from-git-api'
 import { authProvider } from '../services/AuthProvider'
 import type { ChatModel } from './chat-view/ChatModel'
-import { contextItemMentionFromOpenCtxItem } from './context/chatContext'
+import {
+    contextItemMentionFromOpenCtxItem,
+    getActiveEditorContextForOpenCtxMentions,
+} from './context/chatContext'
 import type { ExtensionMessage } from './protocol'
 
 type PostMessage = (message: Extract<ExtensionMessage, { type: 'clientState' }>) => void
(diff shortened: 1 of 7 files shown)


vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -153,7 +159,30 @@ export async function getCorpusContextItemsForEditorState(useRemote: boolean): P
...
+        return items
+    }
+
+    const activeEditorContext = await getActiveEditorContextForOpenCtxMentions()
+
+    const openctxMentions = (
+        await Promise.all(
+            providers.map(async (provider): Promise<ContextItemOpenCtx[]> => {
+                const mentions =
+                    (await openCtx?.controller?.mentions(activeEditorContext, provider)) || []
```
vovakulikov, line 173: I suspect the `mention` method API has been changed to support the initial context, can you link it in this PR?. I wonder if it requires us to migrate open ctx providers that we wrote in the cody repository, like a directory or repo or files providers?

vovakulikov (approved):
consumer code looks good to me, only wonder about openctx provider change that had to be done for this

vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -153,7 +159,30 @@ export async function getCorpusContextItemsForEditorState(useRemote: boolean): P
         }
     }
 
-    return items
+    const providers = (await openCtx.controller?.meta({}))?.filter(meta => meta.mentions?.autoInclude)
+    if (!providers) {
+        return items
+    }
+
+    const activeEditorContext = await getActiveEditorContextForOpenCtxMentions()
```
vovakulikov, line 167: I think you had to type cast editor context since mensionParams is slightly different with undefined types. Do you see TS errors locally?

thenamankumar: All the props in mentionsParams are optional so no typecasting required. There are only 3: query, uri, and codebase.

vscode/src/chat/clientStateBroadcaster.ts
thenamankumar: Here is the PR: https://github.com/sourcegraph/openctx/pull/199

No, this is an optional feature we are implementing for a customer to facilitate default prompt powered by OpenCtx. 

We do not need to update any of the providers.

vscode/src/chat/clientStateBroadcaster.ts
vovakulikov: Maybe it's my IDE 
Screenshot 2024-09-02 at 14 32 40

thenamankumar: Try doing `pnpm i` and reload. Because the openctx version is upgraded.

vovakulikov: I did, and I see the types from the new version. But I checked it in typescript playground and it looks fine, so i think it's my IDE problem

vscode/src/chat/clientStateBroadcaster.ts
```diff
@@ -153,7 +159,30 @@ export async function getCorpusContextItemsForEditorState(useRemote: boolean): P
         }
     }
 
-    return items
+    const providers = (await openCtx.controller?.meta({}))?.filter(meta => meta.mentions?.autoInclude)
```
sqs, line 162: This API design makes it so that a provider can't supply some mentions that are auto-included and some that aren't. Maybe that's best, but I am not sure. It's fine to unblock, but please document this in code and on the docs (https://openctx.org) as an experimental API.

sqs (approved):

vscode/src/chat/clientStateBroadcaster.ts
thenamankumar: Hmm, the only distinction would be the query will be undefined in this case, but that is not a good enough distinction. 

Let me mark it experimental for now on OpenCtx, then figure out how better we can represent it.

thenamankumar: How about passing the `autoInclude` boolean to the `mentions` func as well?

PR #5434, opened 2024-09-02
Adjust mistral model context window

taras-yemets:
This change adjusts the context window size for models using the Mistral tokenizer to address an issue where the OpenAI tokenizer, currently used by default for all models, undercounts tokens in English by about 13%. To prevent completion requests from failing due to context window overflow, the window is reduced by 15% (using a 0.85 multiplier) as a safety buffer.

The proper fix would be to implement model-specific tokenizers, but this requires significant refactoring, which isn’t feasible before the upcoming release. This adjustment serves as a temporary workaround to maintain stability until we can make that larger change.

For more details, please check out [this thread](https://sourcegraph.slack.com/archives/C075G57B5PH/p1724964291794849).

## Test plan
- Tested manually
- CI
- TODO: add unit test

### DIFF
--- a/vscode/src/models/sync.ts
+++ b/vscode/src/models/sync.ts
@@ -10,7 +10,7 @@ import {
     isDotCom,
     modelsService,
 } from '@sourcegraph/cody-shared'
-import type { ServerModelConfiguration } from '@sourcegraph/cody-shared/src/models'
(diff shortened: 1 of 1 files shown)


valerybugakov (approved):
Nice!

slimsag:
fantastic, thank you!

PR #5441, opened 2024-09-03
strongly typed {Authenticated,Unauthenticated}AuthState

sqs:
Most places in code assume that AuthState represents an authenticated user. When unauthenticated, values for the AuthStatus fields were misleading, such as `username: ''` and `endpoint: ''`. Now, those fields are not even available on the type if `authenticated: false`.

(No behavior change. Just a code refactor.)

## Test plan

CI


dominiccooney (approved):
Nice cleanup.

PR #5443, opened 2024-09-03
CLI: use client name "cody-cli" instead of "jetbrains"

olafurpg:
Fixes CODY-3597

Previously, Cody CLI used the client name "jetbrains" to work around an issue where old SG instances would reject LLM requests for unknown client names. This issue was fixed in July but it will take some months until we can assume all users have upgraded their instances.

Now, we can use the client name "cody-cli" for telemetry and only "jetbrains" for the LLM requests.

## Test plan
n/a

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1516,6 +1516,10 @@ export class Agent extends MessageHandler implements ExtensionClient {
         return this.clientInfo?.name.toLowerCase() || 'uninitialized-agent'
     }
 
+    get httpClientNameForLegacyReasons(): string | undefined {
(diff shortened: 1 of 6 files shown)


valerybugakov (approved):

olafurpg:
FYI @jamesmcnamara turns out that the legacy name for HTTP clients wasn't working correctly because we were setting the name to `clientName` in main.ts. This PR fixes the issue so that we use the legacy name in main.ts

olafurpg:
Thank you CLI tests and HTTP recordings for catching this!

PR #5444, opened 2024-09-03
Chat: make "insert code at cursor" replace selection

olafurpg:
Previously, the "Insert code at cursor" button always added the code at the start of the cursor or text selection. This was problematic because when you have a text selection, you intuitively expect it to replace the whole selection, not add the code. This PR fixes the issue by replacing when there is a text selection, and adding when there is a cursor.

## Test plan

Manually tested this with a text selection and with a cursor.

## Changelog

* The "Insert code at cursor" action now replaces the selected text instead of adding it to the start of the selection. The behavior is unchanged when there is a cursor (no text selection).


valerybugakov (approved):
Makes sense!

PR #5448, opened 2024-09-03
Telemetry: Update naming for smart-apply

akalia25:
This PR refactors the telemetry event for 'cody.smart-apply...' event to use a consistent name for the various `selectionType` and update the action to be a verb as it's intended.

Changes:
- Standardized the event name to `cody.smart-apply:selected` for all selection types.
- Updated the `action` to `selected` to indicate the action being recorded.
- Included a metadata object to capture the `selectionType` with a value of 1.

image

## Test plan
Tested locally

### DIFF
--- a/vscode/src/edit/manager.ts
+++ b/vscode/src/edit/manager.ts
@@ -319,7 +319,11 @@ export class EditManager implements vscode.Disposable {
             return
         }
 
-        telemetryRecorder.recordEvent('cody.smart-apply.selection', selection.type)
+        telemetryRecorder.recordEvent('cody.smart-apply', 'selected', {
+            metadata: {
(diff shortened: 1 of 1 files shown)


valerybugakov (approved):

valerybugakov (commented):
Is there a way to update https://sourcegraph.looker.com/looks/2017 in a backward compatible way?

akalia25:
> Is there a way to update https://sourcegraph.looker.com/looks/2017 in a backward compatible way?

Thanks @valerybugakov for calling this out. We can re-create this look with the new data, if its important to have the historical data alongside the new data, let us know and we can try to work through that too :). going to go ahead and merge this change.

Issue #4863, opened 2024-07-14
bug: `Error: Failed to get git output.` when using commit generation


A2va:
### Version

v1.26.2

### Describe the bug

After clicking on the commit generation button, the text entry for the commit message is filled with `Error: Failed to get git output.`
This happens regardless of whether some files are staged or not.
![image](https://github.com/user-attachments/assets/7b67ec75-8ac7-46c3-b003-3dcb74a25c30)

### Expected behavior

The commit message should be generated.

### Additional context

Git output from the vscode session:

```
2024-07-14 21:24:48.285 [info] > git config --local commit.template [85ms]
2024-07-14 21:24:48.285 [warning] git config failed: Failed to execute git
2024-07-14 21:24:48.297 [info] > git config --global commit.template [90ms]
2024-07-14 21:24:48.297 [warning] git config failed: Failed to execute git
2024-07-14 21:24:48.412 [info] > git diff --name-status -z --diff-filter=ADMR --cached [107ms]
2024-07-14 21:24:48.420 [info] > git diff --name-status -z --diff-filter=ADMR [108ms]
2024-07-14 21:24:48.432 [info] > git log --format=%H%n%aN%n%aE%n%at%n%ct%n%P%n%D%n%B -z -n5 [113ms]
2024-07-14 21:24:48.509 [info] > git diff --cached [82ms]
```
OS: Windows 11

ichim-david:
Same issue on Mac Sonoma

ichim-david:
I think it has todo with more than one source of git repos.
In my case I have a react project which has a github repository and inside a src/addons I have several add-ons that have their own repo.
I am trying to create the commit message usually on these add-ons but even on the root it fails.
In order for VS Code to see those git remotes I had to make a workspace for the repo project and then add to the worspace the add-ons folder.

For another project where it's just the main repo I was able to finally get the commit message functionality to create a message.

Do be determined if this feedback is as any use :)

ichim-david:
@jdorfman @sqs can this issue get some love? It seems that Cody tries todo the git commit from the workspace path.
have a  look at this screenshot
![cody-git](https://github.com/user-attachments/assets/0d35c538-6c4e-4457-a6b5-a502860087d4)

from my project https://github.com/eea/eea-website-frontend
It's a react project with some packages inside src/addons folder.
Each folder is another git clone, for example https://github.com/eea/volto-eea-website-policy/

[...]

PR #5449, opened 2024-09-03
fix: handle git diff path parsing on Windows

abeatrix:
CLOSE  https://linear.app/sourcegraph/issue/CODY-3601/bug-error-failed-to-get-git-output-when-using-commit-generation-on https://github.com/sourcegraph/cody/issues/4863

This commit fixes an issue where the `Generate Commit Message` command would fail on Windows due to incorrect parsing of the git diff output. The change ensures that the file path is correctly identified even when the diff output does not contain the full file path specifically on Windows

## Test plan

Manually tested the `Generate Commit Message` command on both Windows and Mac to ensure it works as expected.

We also have an e2e test covering this command on all platforms, will look into it separately on why this error didn't happen during the e2e tests

#### After

Screenshot 2024-09-03 083325

#### Before

image

### Mac

Confirmed the command still works correctly on Mac:

image

## Changelog

Command: Fixed an issue where the experimental `Generate Commit Message` command would fail on Windows due to incorrect parsing of the git diff output.

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -8,12 +8,14 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Fixed
 
-### Changed
+Command: Fixed an issue where the experimental `Generate Commit Message` command would fail on Windows due to incorrect parsing of the git diff output.
 
+### Changed
 
 ## 1.32.5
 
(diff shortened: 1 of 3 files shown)


arafatkatze (approved):
This change doesn't affect Mac and it has E2E tests(being added) so this is good to go

PR #5452, opened 2024-09-03
prevent invalid ranges from borking context

beyang:
## Test plan

Test locally with the following query against s2: `@sourcegraph/cody where do we check for content attribution?`

image

s2 returns a result with a negative end line, which causes the rest of the context items to be discarded. This should not happen with this change.

### DIFF
--- a/vscode/src/chat/chat-view/context.ts
+++ b/vscode/src/chat/chat-view/context.ts
@@ -134,7 +134,10 @@ async function searchRemote(
         if (!remoteSearch) {
             return []
         }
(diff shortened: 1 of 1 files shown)


camdencheek (approved):
Approving to fix the error, but specifically pinging @sourcegraph/search-platform since I think this should be fixed on the backend. We should never be returning a negative end line.

keegancsmith:
@beyang is it possible the context came from sources that are not our normal search backend? I know you have a bunch of feature flags around stuff the AI team is working on. So just trying to triage if this is something for our team to look into or the AI team.

PR #5453, opened 2024-09-03
add suggestions diff tracker for online context and nes dataset

hitesh-1997:
## Context
This PR adds a persistence telemetry event which logs the git diff of the file for each suggestion after 30 seconds **only for dotCom users and public github Repos** in the private metadata field.
The utility of this data is:
1. For the completions with `negative CAR` values, this event can give the actual code typed by the user after rejecting cody suggestions. These are typically called `negative examples` which can be used to improve the completions quality.
2. The `git diff` at different location than the suggestion one can give us an initial dataset for `next-edit-suggestion`. This data can be used to draw insights into `next-edits` (for the same file).
3. All the online datasets typically uses BM25/Embeddings as the context sampler but ignores the user activity like recent edits. Having this information along with context items logging can give us a ground truth sampled from online user queries for context improvements. 

## Test plan
1. Manual debugging and checking the events logged on vscode output channel.
2. The telemetry output is of the following format (Note the `field` `diff` in `privateMetadata`)

```
█ telemetry-v2: recordEvent: cody.completion.suggestion/documentDiff  {
  "parameters": {
    "version": 0,
    "interactionID": "14bd590d-e7cc-4d2a-a67c-e3900d8a0cf5",
    "metadata": [
      {
        "key": "afterSec",
        "value": 10
      },
      {
        "key": "difference",
        "value": 0.014861213461066078
      },
      {
        "key": "lineCount",
        "value": 665
      },
      {
        "key": "charCount",
        "value": 24426
      }
    ],
    "privateMetadata": {
      "id": "14bd590d-e7cc-4d2a-a67c-e3900d8a0cf5",
      "diff": "--- a/file:///Users/hiteshsagtani/dev/Compiler-Project-BITS-Pilani/symboltable.c\n+++ b/file:///Users/hiteshsagtani/dev/Compiler-Project-BITS-Pilani/symboltable.c\n@@ -523,8 +523,14 @@\n             if(strcmp(gettype(head->Lex->tk->value, scope), \"INTEGER\")!=0)\n                 fprintf(fp,\"ERROR 6: Array index should be of type integer but here, %s has type %s, line number %d\\n\", head->Lex->tk->value, gettype(head->Lex->tk->value, scope), head->Lex->tk->line);\n             }\n         \n+        if(head->parent->value==24){\n+            //assert(0);\n+            if(strcmp(gettype(head->Lex->tk->value, scope), \"INTEGER\")!=0)\n+                fprintf(fp,\"ERROR 6: Array index should be of type integer but here, %s has type %s, line number %d\\n\", head->Lex->tk->value, gettype(head->Lex->tk->value, scope), head->Lex->tk->line);\n+            }\n+        }\n         \n \n \n           if(head->parent->rule == 41 && head->right->rule==44)\n"
    }
  },
  "timestamp": "2024-09-03T23:10:32.184Z"
}
```

### DIFF
--- a/vscode/src/common/persistence-tracker/index.ts
+++ b/vscode/src/common/persistence-tracker/index.ts
@@ -22,14 +22,29 @@ interface TrackedInsertion<T = string> {
     metadata?: PersistenceEventMetadata
 }
 
-function getDefaultMeasureTimeouts(): number[] {
-    return [
-        30 * 1000, // 30 seconds
-        120 * 1000, // 2 minutes
-        300 * 1000, // 5 minutes
-        600 * 1000, // 10 minutes
-    ]
+type Logger<T> = {
+    onRemoved: (event: PersistenceRemovedEventPayload<T>) => void
+    onPresent: (event: PersistencePresentEventPayload<T>) => void
 }
+
+interface TrackingPayload<T = string> {
+    id: T
+    insertedAt: number
+    insertText: string
+    insertRange: vscode.Range
+    document: vscode.TextDocument
+    metadata?: PersistenceEventMetadata
+    logger: Logger<T>
+    persistenceTimeoutList?: number[]
+    shouldComputeCodeDiff?: boolean
+}
+
+const DEFAULT_MEASURE_TIMEOUTS = [
+    30 * 1000, // 30 seconds
+    120 * 1000, // 2 minutes
+    300 * 1000, // 5 minutes
+    600 * 1000, // 10 minutes
+]
 
 export class PersistenceTracker<T = string> implements vscode.Disposable {
     private disposables: vscode.Disposable[] = []
@@ -42,12 +57,7 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
         workspace: Pick<
             typeof vscode.workspace,
             'onDidChangeTextDocument' | 'onDidRenameFiles' | 'onDidDeleteFiles'
-        >,
-        public logger: {
-            onRemoved: (event: PersistenceRemovedEventPayload<T>) => void
-            onPresent: (event: PersistencePresentEventPayload<T>) => void
-        },
-        private readonly shouldComputeCodeDiff = false
+        >
     ) {
         this.disposables.push(workspace.onDidChangeTextDocument(this.onDidChangeTextDocument.bind(this)))
         this.disposables.push(workspace.onDidRenameFiles(this.onDidRenameFiles.bind(this)))
@@ -61,20 +71,13 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
         insertRange,
         document,
         metadata,
-        persistenceTimeoutList = getDefaultMeasureTimeouts(),
-    }: {
-        id: T
-        insertedAt: number
-        insertText: string
-        insertRange: vscode.Range
-        document: vscode.TextDocument
-        metadata?: PersistenceEventMetadata
-        persistenceTimeoutList?: number[]
-    }): void {
+        logger,
+        persistenceTimeoutList = DEFAULT_MEASURE_TIMEOUTS,
+        shouldComputeCodeDiff = false,
+    }: TrackingPayload<T>): void {
         if (insertText.length === 0) {
             return
         }
-
         const trackedInsertion = {
             insertText,
             insertRange,
@@ -94,18 +97,33 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
 
         documentInsertions.add(trackedInsertion)
         const firstTimeoutIndex = 0
-        this.enqueueMeasure(trackedInsertion, firstTimeoutIndex, persistenceTimeoutList)
+        this.enqueueMeasure(
+            trackedInsertion,
+            firstTimeoutIndex,
+            persistenceTimeoutList,
+            shouldComputeCodeDiff,
+            logger
+        )
     }
+
     private enqueueMeasure(
         trackedInsertion: TrackedInsertion<T>,
         nextTimeoutIndex: number,
-        persistenceTimeoutList: number[]
+        persistenceTimeoutList: number[],
+        shouldComputeCodeDiff: boolean,
+        logger: Logger<T>
     ): void {
         const timeout =
             trackedInsertion.insertedAt + persistenceTimeoutList[nextTimeoutIndex] - Date.now()
         const timeoutId = setTimeout(() => {
             this.managedTimeouts.delete(timeoutId)
-            this.measure(trackedInsertion, nextTimeoutIndex, persistenceTimeoutList)
+            this.measure(
+                trackedInsertion,
+                nextTimeoutIndex,
+                persistenceTimeoutList,
+                shouldComputeCodeDiff,
+                logger
+            )
         }, timeout)
         this.managedTimeouts.add(timeoutId)
     }
@@ -114,7 +132,9 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
         trackedInsertion: TrackedInsertion<T>,
         // The index in the MEASURE_TIMEOUTS array
         measureTimeoutsIndex: number,
-        persistenceTimeoutList: number[]
+        persistenceTimeoutList: number[],
+        shouldComputeCodeDiff: boolean,
+        logger: Logger<T>
     ): void {
         const isStillTracked = this.trackedInsertions
             .get(trackedInsertion.uri.toString())
@@ -128,7 +148,7 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
 
         if (latestText.length === 0) {
             // Text was fully deleted
-            this.logger.onRemoved({
+            logger.onRemoved({
                 id: trackedInsertion.id,
                 difference: 1,
                 metadata: trackedInsertion.metadata,
@@ -137,11 +157,11 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
             const maxLength = Math.max(initialText.length, latestText.length)
             const editOperations = levenshtein(initialText, latestText)
             const difference = editOperations / maxLength
-            const diff = this.shouldComputeCodeDiff
+            const diff = shouldComputeCodeDiff
                 ? createGitDiff(trackedInsertion.uri.toString(), initialText, latestText)
                 : undefined
 
-            this.logger.onPresent({
+            logger.onPresent({
                 id: trackedInsertion.id,
                 afterSec: persistenceTimeoutList[measureTimeoutsIndex] / 1000,
                 difference,
@@ -154,7 +174,13 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
 
             // If the text is not deleted yet and there are more timeouts, schedule a new run.
             if (measureTimeoutsIndex < persistenceTimeoutList.length - 1) {
-                this.enqueueMeasure(trackedInsertion, measureTimeoutsIndex + 1, persistenceTimeoutList)
+                this.enqueueMeasure(
+                    trackedInsertion,
+                    measureTimeoutsIndex + 1,
+                    persistenceTimeoutList,
+                    shouldComputeCodeDiff,
+                    logger
+                )
                 return
             }
         }
--- a/vscode/src/common/persistence-tracker/index.ts
+++ b/vscode/src/common/persistence-tracker/index.ts
@@ -1,6 +1,6 @@
 import levenshtein from 'js-levenshtein'
 import type * as vscode from 'vscode'
-
+import { createGitDiff } from '../../../../lib/shared/src/editor/create-git-diff'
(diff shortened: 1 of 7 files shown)


vscode/src/completions/logger.ts
```diff
@@ -315,6 +315,22 @@ function logCompletionPartiallyAcceptedEvent(params: PartiallyAcceptedEventPaylo
         params
     )
 }
+
+function logSuggestionsDocumentDiffEvent(params: PersistencePresentEventPayload): void {
+    // Use automatic splitting for now - make this manual as needed
+    const { metadata, privateMetadata } = splitSafeMetadata(params)
```
hitesh-1997, line 321: @akalia25 Hi Aditya, Does the addition of this new field require any backend changes. I checked the data is only going for telemetry-v2 and `dotCom` and `public` github repos.

vscode/src/completions/logger.ts
```diff
@@ -718,6 +734,47 @@ export function loaded(params: LoadedParams): void {
...
+        document.lineAt(document.lineCount - 1).range.end
+    )
+    const documentText = document.getText(insertRange)
+
+    const isValidCodeFile = document.uri.scheme === 'file'
+    const isValidSize = documentText.length < 256 * 1024
+    if (isValidCodeFile && isValidSize) {
+        const persistenceTimeoutList = [
+            30 * 1000, // 30 seconds
+        ]
```
hitesh-1997, line 769: @valerybugakov Do you think we need additional timeouts here eg: track diff after 1 minute etc.
Also do you think we should limit the datapoints further (Although currently this only triggers for ~10% of suggestion requests which are public and dotCom)

vscode/src/completions/logger.ts
```diff
@@ -718,6 +734,47 @@ export function loaded(params: LoadedParams): void {
...
+        )
+    }
+    const insertRange = new vscode.Range(
+        new vscode.Position(0, 0),
+        document.lineAt(document.lineCount - 1).range.end
+    )
+    const documentText = document.getText(insertRange)
+
+    const isValidCodeFile = document.uri.scheme === 'file'
+    const isValidSize = documentText.length < 256 * 1024
```
hitesh-1997, line 765: @valerybugakov I added this 2 checks to ensure the data is not big for which we compute diff. Do you think we need any additional checks or correction in these checks ?

vscode/src/completions/logger.ts
akalia25: Hi @hitesh-1997, thanks for checking in, yes we would need to add the new `privateMetadata` key `diff` to our backend infra to handle this sensitive data correctly. Happy to handle this for you, and can update this thread once it's added. Curious, have you checked in with legal on this data, wondering if we should give them a heads-up.

Question about how often this telemetry event will get logged, is it only **once**, 30 seconds after a completion suggestion, or is it continuously logging **every** 30 seconds after a suggestion is shown? I worry that this might be a high-volume event, and checking if we need to reduce the frequency a bit.

Nit: I would also love to update the `action` name of this event, we try to typically think of this as the verb of the event, maybe `generateDiff` could be a better action name?

vscode/src/completions/logger.ts
```diff
@@ -315,6 +315,22 @@ function logCompletionPartiallyAcceptedEvent(params: PartiallyAcceptedEventPaylo
         params
     )
 }
+
+function logSuggestionsDocumentDiffEvent(params: PersistencePresentEventPayload): void {
+    // Use automatic splitting for now - make this manual as needed
+    const { metadata, privateMetadata } = splitSafeMetadata(params)
+    writeCompletionEvent(
+        'suggestion',
```
valerybugakov, line 323: Let's use the `persistence` name here, similar to other events from the persistence tracker.

vscode/src/completions/logger.ts
```diff
@@ -580,6 +596,7 @@ const completionIdsMarkedAsSuggested = new LRUCache<CompletionAnalyticsID, true>
 })
 
 let persistenceTracker: PersistenceTracker<CompletionAnalyticsID> | null = null
+let suggestionTracker: PersistenceTracker<CompletionAnalyticsID> | null = null
```
valerybugakov, line 599: Let's add comments explaining the need for two instances of the same tracker. I would also suggest renaming `suggestionTracker` to something that matches the goal of this new instance.

valerybugakov: Thinking more about it, I realize the two trackers are not that different. 

- We call the first one only for accepted completions without generating file diffs. 
- For the second one, we call for all suggested completions (including accepted ones) and add the file diff to the event.

Can we merge them into one tracker that would:

- Add the diff to every first event.
- Create additional events only for accepted completions.

It would simplify the setup and clarify the distinctions between the events. We would need only one tracker instance and one call of the `track` method.

vscode/src/common/persistence-tracker/index.ts
```diff
@@ -127,20 +137,24 @@ export class PersistenceTracker<T = string> implements vscode.Disposable {
...
             this.logger.onPresent({
                 id: trackedInsertion.id,
-                afterSec: MEASURE_TIMEOUTS[measureTimeoutsIndex] / 1000,
+                afterSec: persistenceTimeoutList[measureTimeoutsIndex] / 1000,
                 difference,
                 lineCount:
                     trackedInsertion.latestRange.end.line - trackedInsertion.latestRange.start.line + 1,
                 charCount: latestText.length,
                 metadata: trackedInsertion.metadata,
+                diff,
```
valerybugakov, line 152: @akalia25, should we be concerned about the event's size here?

vscode/src/common/persistence-tracker/types.ts
```diff
@@ -13,6 +13,8 @@ export interface PersistencePresentEventPayload<T = string> {
     charCount: number
     /** Attached metadata to the insertion */
     metadata?: PersistenceEventMetadata
+
+    diff?: string
```
valerybugakov, line 17: Let's add a doc string here. Later it will be cryptic for people trying to understand the structure of the analytics events.

vscode/src/completions/logger.ts
valerybugakov: We can start with one event here and adjust it later based on the results. Probably it should be enough to get negative examples.

vscode/src/completions/logger.ts
hitesh-1997: Thanks for the pointer @akalia25
 
> We also need to add recordsPrivateMetadataTranscript:1
Sorry I forgot about this one, Added this field in the metadata :)

> Curious, have you checked in with legal on this data, wondering if we should give them a heads-up.
I think this should have the same severity as the `prefix` and `suffix` of the file (since this only records the diff of the file and we already log `prefix` and `suffix` of the file). Just in case, I asked @dadlerj on [slack here](https://sourcegraph.slack.com/archives/CN4FC7XT4/p1725454159946369?thread_ts=1716253514.127349&cid=CN4FC7XT4) for approval. Please let me know if we should tag anyone else here ? (Just FYI: this is just for `dotCom` users with `Public` github repos)
Also cc: @rishabhmehrotra @beyang for approval

> Question about how often this telemetry event will get logged, is it only once, 30 seconds after a completion suggestion

It only logs the data once, i.e. 30 secs after completion suggestion. Since if logs the data for public and dotCom users, its frequency is same as previous added `inlineContextItems` in [this PR](https://github.com/sourcegraph/cody/pull/4501) (i.e. ~10% if completions suggestion events)

I would also love to update the action name of this event, we try to typically think of t

> his as the verb of the event, maybe `generateDiff` could be a better action name?
Nice, I love this name :) changed to `generateDiff`

vscode/src/completions/logger.ts
hitesh-1997: nice suggestion !! Added offset in the range to track 128k chars before and after the current cursor position (for comparison the `logger.ts` file is `49k` characters), please let me know if this looks sufficient.

vscode/src/completions/logger.ts
```diff
@@ -718,6 +734,47 @@ export function loaded(params: LoadedParams): void {
...
+            vscode.workspace,
+            // Only Log the diff if the document was not deleted.
+            {
+                onPresent: logSuggestionsDocumentDiffEvent,
+                onRemoved: () => {},
+            },
+            true
+        )
+    }
+    const insertRange = new vscode.Range(
```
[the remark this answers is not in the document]
hitesh-1997: Since we moved the tracking 128k chars, renamed it to `trackingRange`

vscode/src/common/persistence-tracker/index.ts
```diff
@@ -9,13 +9,6 @@ import type {
     PersistenceRemovedEventPayload,
 } from './types'
 
-const MEASURE_TIMEOUTS = [
```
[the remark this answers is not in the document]
hitesh-1997: done

vscode/src/completions/logger.ts
```diff
@@ -718,6 +734,47 @@ export function loaded(params: LoadedParams): void {
...
+        new vscode.Position(0, 0),
+        document.lineAt(document.lineCount - 1).range.end
+    )
+    const documentText = document.getText(insertRange)
+
+    const isValidCodeFile = document.uri.scheme === 'file'
+    const isValidSize = documentText.length < 256 * 1024
+    if (isValidCodeFile && isValidSize) {
```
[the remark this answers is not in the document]
hitesh-1997: moved the checks to the `beginning` of the function

vscode/src/common/persistence-tracker/index.ts
hitesh-1997: Added a check, if the diff is larger than 256kb, we dont send the event in the backend (also makes sense, since the diff would be too large to be useful for analysis).

vscode/src/common/persistence-tracker/types.ts
hitesh-1997: done

vscode/src/completions/logger.ts
hitesh-1997: Nice, I really like the suggestion, since both are conceptually the same.
I moved some methods like `logger` which were happening at service initialization to `track` function and make them common.
I think we can go one step further to make it a singleton service so that `fixupController` and `completions` can use the same instance, but not going into too much details, I limited the current change to use a single `persistenceTracker` for completions.

vscode/src/completions/logger.ts
akalia25: Thanks @hitesh-1997, I've added the key `diff` to our backend infra, this will handle `diff` the same way we are with transcripts (`promptText`, `responseText`, etc). 

Thanks for the clarification on event volume, maximum data storage of 256KB, and updating the name. 🚀

vscode/src/common/persistence-tracker/index.ts
akalia25: Thanks for checking in on this @valerybugakov, the limit of 256KB sounds good! Just for context, GCP's limit is 10mb of data on each event, and we try to keep our events much lower, with our heavier (transcript) events coming in around 3-5Mb.

vscode/src/completions/logger.ts
hitesh-1997: done

vscode/src/completions/logger.ts
```diff
@@ -718,6 +734,47 @@ export function loaded(params: LoadedParams): void {
                 filePath: snippet.uri.fsPath,
             })),
         }
+        const document = requestParams.document
+        suggestionDocumentDiffTracker(event.params.id, document)
```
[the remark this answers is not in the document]
hitesh-1997: sorry I missed this, `markAsRead` should be the ideal place for this logging
Changed to calling this from the `markAsRead` callback

vscode/src/completions/logger.ts
```diff
@@ -720,6 +748,49 @@ export function loaded(params: LoadedParams): void {
...
+    const offsetBytes = 1024 * 128
+    const startPosition = document.positionAt(Math.max(0, document.offsetAt(position) - offsetBytes))
+    const endPosition = document.positionAt(
+        Math.min(document.getText().length, document.offsetAt(position) + offsetBytes)
+    )
+    const trackingRange = new vscode.Range(startPosition, endPosition)
+    const documentText = document.getText(trackingRange)
+
+    const persistenceTimeoutList = [
+        60 * 1000, // 60 seconds
```
hitesh-1997, line 773: @valerybugakov While playing locally I felt that 30 secs might be too less to generate meaningful diffs so increased it to 60 secs. Please let me know if you want me to change it to some other value

valerybugakov: LGTM. It would be best to adjust it based on the first days of analytics data we get from this.

valerybugakov (approved):
THIS IS CLEAN! 🧼

PR #5454, opened 2024-09-03
feat(chat): improve context item display in client state broadcaster

ykdojo:
Closes SRCH-1013.

This change enhances the display of context items in the client state broadcaster by:

- Always adding the current file item, with the file name as the description

https://linear.app/sourcegraph/issue/SRCH-1013/always-show-current-file-mention-option-even-if-there-is-a-selection

## Test plan

Manual testing.

### DIFF
--- a/vscode/src/chat/clientStateBroadcaster.ts
+++ b/vscode/src/chat/clientStateBroadcaster.ts
@@ -50,22 +50,33 @@ export function startClientStateBroadcaster({
         signal?.throwIfAborted()
         if (contextFile) {
             const range = contextFile.range ? expandToLineRange(contextFile.range) : undefined
-            const item = {
+
+            // Always add the current file item
+            items.push({
                 ...contextFile,
                 type: 'file',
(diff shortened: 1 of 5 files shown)


abeatrix (changes requested):
This change doesn't work as described because even though the @-mention file and selection are showing up as @-mention chips in the input box, the selection context would get removed from the context list by the unique context filter:

image

We will need to add `ContextItemSource.Initial` to this list to exclude initial context from being filtered out: 

https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/prompt-builder/utils.ts?L67-75

Update it to :

```ts
export function getContextItemTokenUsageType(item: ContextItem): ContextTokenUsageType {
    switch (item.source) {
        case 'user':
        case 'initial':
        case 'selection':
            return 'user'
        default:
            return 'corpus'
    }
}
```

With the above change:

image

abeatrix (approved):

PR #5455, opened 2024-09-04
Agent: clean-up assertion of telemetry events within integration tests

akalia25:
Previously, the assertion of telemetry events for agent integration tests was happening within an `afterEach`, and this was not the cleanest approach, as the error messages would not point to the relevant code causing the error. In addition, this PR also adds `parameters` as a property that gets sent with the testing telemetry events.

Now, we manually add the assertion at the end of test cases where we want to assert what events got exported. As @olafurpg mentioned, the benefits of this approach are:

> The benefit of this approach is that you get clearer error messages pointing to a relevant line of code in the tests. With the afterEach() approach, you need to hunt down the test name and reason about this currentClient logic. Also, this afterEach() block is only available in index.test.ts, it's not straightforward how to reuse this logic in other test files.

Addresses feedback [here](https://github.com/sourcegraph/cody/pull/4923#discussion_r1722970638) 

## Test plan
CI, specifically `pnpm run test src/index.test.ts`


agent/src/index.test.ts
```diff
@@ -192,6 +184,12 @@ describe('Agent', () => {
         //
         // If you don't have access to this private file then you need to ask
         expect(valid?.username).toStrictEqual('sourcegraphbot9k-fnwmu')
+
+        // telemetry assertion, to validate the expected events fired during the test run
+        // Do not remove this assertion, and instead update the expectedEvents list above
+        expect(await returnTelemetryEvents(client)).toEqual(
+            expect.arrayContaining(client.expectedEvents)
```
olafurpg, line 191: We should remove the mutable `TestClient.expectedEvents` property and inline the values into the assertions instead. This is the only blocking comment for the PR

olafurpg (changes requested):
I'm sorry, I had published pending comments a long time ago but forgot to "comment" the reviwe.

This is a nice cleanup! Only one blocking comment, otherwise LGTM

agent/src/index.test.ts
```diff
@@ -30,6 +30,28 @@ const workspace = new TestWorkspace(path.join(__dirname, '__tests__', 'example-t
 const mayRecord =
     process.env.CODY_RECORDING_MODE === 'record' || process.env.CODY_RECORD_IF_MISSING === 'true'
 
+async function returnTelemetryEvents(currentClient: TestClient) {
```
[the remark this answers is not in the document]
akalia25: Updated, thank you for the suggestion!

agent/src/index.test.ts
akalia25: thank you for the comment, updated it so that the expectedEvents are inline values to the assertion :) ~ Saves a bunch of lines, great suggestion!

olafurpg (commented):
LGTM 👍🏻 Thank you!

olafurpg (approved):
LGTM 👍🏻 Thank you!

PR #5469, opened 2024-09-04
Chat: show correct model icons and deduplicate local models

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3628/bug-icon-display-incorrectly-for-anthropic-and-openai-models

PR to fix issues where local models are showing up twice in the model list, and the icons are not displayed correctly for openAI and Anthropic models.

- Updates the `ChatModelIcon` component to display the correct icon based on the model provider instead of the model ID.
- Fixes an issue where local models were being duplicated in the model list by filtering out any models that are already in the cache when adding new models.
- The `ChatModelIcon` component now expects the model provider instead of the model ID as the input.

## Test plan

1. Icons are not displayed correctly for openAI and Anthropic models: 

#### After

image

#### Before

image

2. Local Models are duplicated:

#### After

image

#### Before

image
## Changelog

Fixed:
Chat: Model list now shows the correct icon based on the model provider. 
Chat: Fixed an issue where local models were duplicated in the model list.

### DIFF
--- a/lib/shared/src/models/index.ts
+++ b/lib/shared/src/models/index.ts
@@ -445,14 +445,18 @@ export class ModelsService {
     }
 
     /**
-     * Add new models for use.
+     * Add new models in addition to the primary models for use.
+     * NOTE: use setModels for a complete replacement of the primary models.
      */
     public addModels(models: Model[]): void {
-        const set = new Set(this.primaryModels)
-        for (const provider of models) {
-            set.add(provider)
-        }
-        this.primaryModels = Array.from(set)
+        const existingIds = new Set(
+            this.primaryModels.filter(m => m.tags.includes(ModelTag.BYOK)).map(m => m.id)
+        )
+        // Filter out any models that are already in the cache.
+        this.primaryModels = [
+            ...this.primaryModels,
+            ...models.filter(model => !existingIds.has(model.id)),
+        ]
     }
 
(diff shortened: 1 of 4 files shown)


kalanchan (approved):

lib/shared/src/models/index.ts
```diff
@@ -445,14 +445,18 @@ export class ModelsService {
...
-        const set = new Set(this.primaryModels)
-        for (const provider of models) {
-            set.add(provider)
-        }
-        this.primaryModels = Array.from(set)
+        const existingIds = new Set(
+            this.primaryModels.filter(m => m.tags.includes(ModelTag.BYOK)).map(m => m.id)
+        )
+        // Filter out any models that are already in the cache.
+        this.primaryModels = [
```
vovakulikov, line 456: I think you could just 
```js
Array.from(
  new Set([
... this.primaryModels.filter(m => m.tags.includes(ModelTag.BYOK)).map(m => m.id)
...models
])
)
           
```

abeatrix: Ohhh let me update it, thanks!

vovakulikov (approved):

lib/shared/src/models/index.ts
vovakulikov: Oh, right, I thought it was all just list of IDs. feel free to ignore, my mistake

PR #5476, opened 2024-09-05
Agent: add platform-specific secret storage implementation

abeatrix:
This change adds a new `AgentServerManagedSecretStorage` class that implements the `vscode.SecretStorage` interface and uses the system keychain or equivalent to store secrets based on the current platform (macOS, Windows, or Linux). 

The new implementation supports the following operations: 
- `get(key)`: Retrieves the secret value for the given key. 
- `store(key, value)`: Stores the secret value for the given key. 
- `delete(key)`: Deletes the secret value for the given key. 

The `AgentClientManagedSecretStorage` implementation is retained for cases where the client is responsible for managing the secrets, while the `AgentServerManagedSecretStorage` implementation is for cases where the client does not have access to secret storage API.

NOTE: The  `AgentClientManagedSecretStorage` was implemented using the current secret storage implementation in Cody CLI as reference: https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/cli/command-auth/secrets.ts. The powershell commands were however updated after testing in a native Windows env.

## Test plan

- [x] Test it on Windows with Eclipse/Visual Studio to confirm the secret storage is working as expected
- [ ] Test it on Mac with JetBrains to confirm the secret storage is working as expected

I don't have access to Linux machine to confirm if this would work on Linux, but right now this is using the existing implementation we have for CLI so if it is working there it should work here?

### DIFF
--- a/agent/src/AgentSecretStorage.ts
+++ b/agent/src/AgentSecretStorage.ts
@@ -1,3 +1,4 @@
+import { spawn } from 'node:child_process'
 import type * as vscode from 'vscode'
 import { emptyEvent } from '../../vscode/src/testutils/emptyEvent'
 import type { MessageHandler } from './jsonrpc-alias'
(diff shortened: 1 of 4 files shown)


agent/src/AgentSecretStorage.ts
```diff
@@ -34,3 +36,181 @@ export class AgentClientManagedSecretStorage implements vscode.SecretStorage {
         await this.agent.request('secrets/delete', { key })
     }
 }
+
+export class AgentServerManagedSecretStorage implements vscode.SecretStorage {
```
abeatrix, line 40: @olafurpg i followed your implementation in https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/agent/src/cli/command-auth/secrets.ts to create this. I tested it with Visual Studio on Windows and confirm it's working there, so it should also work on Eclipse. Does this look ok to you or do you think we should hold off from implementing server-managed secret storage for now?

olafurpg: If we go this route then we should reuse the code from the CLI tool, you can just import the functions in that file since it's also in the agent project.

However, I'm not convinced we should rely on server-managed secret storage. It's always going to be safer to rely on client-managed secret storage. Did you look into the APIs Tomasz shared here? https://sourcegraph.slack.com/archives/C073U3KE5DK/p1725367417308519

abeatrix: We did! We have an issue created for implementing it this sprint.

As noted in the description this is intended for clients that do not have access to secret storage api.

I do have a question about the current implementation of client managed secret storage: is there any security concern about passing the token through jsonRPC, like having the token showing up in logs unexpectedly?

olafurpg: The JSON-RPC logs should never be written to disk by default for that reason. Eclipse and JB are using client-managed secrets and I think it's best if we do that for VS as well. I can imagine terminal-based editors like Neovim/Emacs may like server-managed secrets, but they're not a priority rn

abeatrix: Since this is not a priority i'll close this for now, thanks!

PR #5478, opened 2024-09-05
react to config/feature-flag changes using Observables in autocomplete

sqs:
Make it so our autocomplete logic reacts to config and feature flag changes using Observables, instead of the current situation where there are many different ad-hoc ways configuration and feature flag values are passed through. This simplifies (or at least standardizes) the code, reduces bugs, and (once this kind of change is applied everywhere in our code) eliminates the need for users to reload the editor to pick up new config.

## Test plan

e2e tests. Also try changing Cody autocomplete configs and signing in/out and ensure that the right config is loaded.

### DIFF
--- a/lib/shared/src/misc/observable.test.ts
+++ b/lib/shared/src/misc/observable.test.ts
@@ -5,8 +5,10 @@ import {
     type ObservableValue,
     allValuesFrom,
     combineLatest,
+    createDisposables,
     distinctUntilChanged,
     firstValueFrom,
+    fromLateSetSource,
     fromVSCodeEvent,
     memoizeLastValue,
     observableOfSequence,
@@ -14,6 +16,7 @@ import {
     promiseFactoryToObservable,
     readValuesFrom,
     shareReplay,
+    take,
 } from './observable'
 
 describe('firstValueFrom', () => {
@@ -54,6 +57,48 @@ describe('promiseFactoryToObservable', () => {
         await done
         expect(values).toEqual([])
         vi.useRealTimers()
+    })
+})
+
+describe('fromLateSetSource', () => {
+    test('emits values after source is set', async () => {
+        const { observable, setSource } = fromLateSetSource<number>()
+        const { values, done } = readValuesFrom(observable)
+
+        setSource(observableOfSequence(1, 2, 3))
+
+        await done
+        expect(values).toEqual([1, 2, 3])
+    })
+
+    test('handles multiple observers', async () => {
+        const { observable, setSource } = fromLateSetSource<string>()
+        const observer1 = readValuesFrom(observable)
+        const observer2 = readValuesFrom(observable)
+
+        setSource(observableOfSequence('a', 'b', 'c'))
+
+        await observer1.done
+        await observer2.done
+        expect(observer1.values).toEqual(['a', 'b', 'c'])
+        expect(observer2.values).toEqual(['a', 'b', 'c'])
+    })
+
+    test('throws error when trying to set source multiple times', () => {
+        const { setSource } = fromLateSetSource<number>()
+        setSource(observableOfSequence(1, 2, 3))
+        expect(() => setSource(observableOfSequence(4, 5, 6))).toThrow('source is already set')
+    })
+
+    test('unsubscribes correctly', async () => {
+        const { observable, setSource } = fromLateSetSource<number>()
+        const { values, done, unsubscribe } = readValuesFrom(observable)
+
+        unsubscribe()
+        setSource(observableOfSequence(1, 2, 3))
+
+        await done
+        expect(values).toEqual([])
     })
 })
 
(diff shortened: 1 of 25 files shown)


vscode/src/completions/context/context-strategy.ts
```diff
@@ -21,79 +28,86 @@ export type ContextStrategy =
...
-    public getStrategy(document: vscode.TextDocument): {
+    public async getStrategy(document: vscode.TextDocument): Promise<{
         name: ContextStrategy
         retrievers: ContextRetriever[]
-    } {
+    }> {
         const retrievers: ContextRetriever[] = []
-
-        switch (this.contextStrategy) {
+        const contextStrategy = await firstValueFrom(this.contextStrategy)
```
valerybugakov, line 109: Is there a potential for it to fall out of sync with the context strategy value computed from the refreshed feature flags? 

Feature flags are refreshed -> `this.contextStrategy` is updated -> `strategyFactory.getStrategy` is called, and it uses the outdated (first) `this.contextStrategy` value. I might be misreading one of the intermediate steps here.

vscode/src/completions/context/retrievers/jaccard-similarity/history.ts
```diff
@@ -63,7 +62,17 @@ export class VSCodeDocumentHistory implements DocumentHistory, vscode.Disposable
...
      * Returns the last n items of history in reverse chronological order (latest item at the front)
      */
-    public lastN(n: number, baseLanguageId: string, ignoreUris?: vscode.Uri[]): HistoryItem[] {
+    public async lastN(
+        n: number,
+        baseLanguageId: string,
+        ignoreUris?: vscode.Uri[]
+    ): Promise<HistoryItem[]> {
+        const enableExtendedLanguagePool = Boolean(
+            await featureFlagProvider.instance!.evaluateFeatureFlag(
```
valerybugakov, line 71: Ideally, we should have the resolved values before running this function. All the downstream getters were kept synchronous to avoid misuse and accidental slowdowns in the autocomplete pipeline.

vscode/src/completions/logger.ts
```diff
@@ -800,7 +798,8 @@ export type SuggestionMarkReadParam = {
 // suggested completion count as soon as we count it as such.
 export function prepareSuggestionEvent(
     id: CompletionLogID,
-    span?: Span
+    span: Span | undefined,
+    shouldSample?: boolean
```
valerybugakov, line 802: Converting the args list to an object would make tests more readable. For example:

```ts
CompletionLogger.prepareSuggestionEvent(id3, undefined)
```

vscode/src/main.ts
```diff
@@ -661,20 +661,14 @@ function registerAutocomplete(
...
-                    dispose: autocompleteFeatureFlagChangeSubscriber,
-                })
                 autocompleteDisposables.push(
-                    await createInlineCompletionItemProvider({
-                        config,
-                        statusBar,
-                        createBfgRetriever: platform.createBfgRetriever,
-                    })
+                    subscriptionDisposable(
+                        createInlineCompletionItemProvider({
```
valerybugakov, line 666: It might be unrelated to the PR, but I noticed that this code path is called twice on VS Code reload.

valerybugakov (approved):
Great work, thank you, Quinn! 

Manually tested with DotCom, sg02, and sg04 ✅. We should consider maintaining guardrails in the code to ensure that all autocomplete-related feature flags are ready before the first completion is triggered.

lib/shared/src/misc/observable.test.ts
```diff
@@ -57,6 +60,48 @@ describe('promiseFactoryToObservable', () => {
     })
 })
 
+describe('fromLateSetSource', () => {
```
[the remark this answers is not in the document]
sqs: ah yeah

vscode/src/completions/completion-provider-config.ts
```diff
@@ -35,85 +27,87 @@ class CompletionProviderConfig {
      */
     public async init(config: ClientConfiguration): Promise<void> {
         this._config = config
-
-        await Promise.all(
-            this.flagsToResolve.map(flag => featureFlagProvider.instance!.evaluateFeatureFlag(flag))
-        )
```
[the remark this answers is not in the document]
sqs: I see. I'll add back the code that pre-fetches these so they are cached.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -192,18 +212,28 @@ export class InlineCompletionItemProvider
...
-                preloadDebounceInterval
-            )
-
-            this.disposables.push(
-                vscode.window.onDidChangeTextEditorSelection(this.onSelectionChangeDebounced)
+        this.disposables.push(
+            subscriptionDisposable(
+                completionProviderConfig.autocompletePreloadDebounceInterval
+                    .pipe(
+                        createDisposables(preloadDebounceInterval => {
```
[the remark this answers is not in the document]
sqs: The helper *is* `createDisposables`. It might be nice to have a helper that can dispose Disposables and Unsubscribables, but that can follow later. Did you have another idea for a helper here?

vscode/src/completions/context/context-strategy.ts
sqs: Yes, that is absolutely still a problem. This PR does not make everything fully reactive, but it does make stuff *more* reactive. I have a lot more commits pending here (`sqs/ui-state` branch).

vscode/src/main.ts
sqs: Yeah. I have another PR to come soon that coalesces updates of config soon.

vscode/src/completions/inline-completion-item-provider.ts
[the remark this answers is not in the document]
valerybugakov: You're right, nvm. I was thinking about a set of operations required to convert an observable to a disposable and had an impression that `createDisposables` is always a part of it.

```ts
disposables.push(subscriptionDisposable(YOUR_OBSERVABLE.pipe(createDisposables(...).subscribe({})))
```

PR #5480, opened 2024-09-05
Testing the Removal of Artificial Latency behind a feature flag

arafatkatze:
This PR aims to remove the artificial delay introduced for low-performance languages and specific AST nodes, as outlined in th [issue](https://linear.app/sourcegraph/issue/CODY-3315/delete-the-artificial-latency). The artificial delay was originally implemented for low performance languages so that users are not spammed with unhelpful suggestions. , but it has become less relevant over time, and it may now impact the predictability of autocomplete latency and user experience (UX).

# Changes:

## Introduce a feature flag:
A new feature flag has been added to toggle the artificial delay.
- Objective: The main objective is to achieve predictable latency and a consistent autocomplete UX for all languages.

- After merging this PR, the feature flag will be enabled for controlled testing, and data will be collected to assess the impact on CAR and latency for all languages
- Depending on the results, the team will decide whether to remove the artificial delay logic entirely.

## Test Plan:
I tested this PR locally by enabling and disabling the feature flag on my local Sourcegraph instance(I added the feature flag on it and then set the true/false values). I also used the debugger to run VS Code Extension on Desktop

### DIFF
--- a/lib/shared/src/experimentation/FeatureFlagProvider.ts
+++ b/lib/shared/src/experimentation/FeatureFlagProvider.ts
@@ -40,7 +40,7 @@ export enum FeatureFlag {
     CodyAutocompleteFIMModelExperimentVariant2 = 'cody-autocomplete-fim-model-experiment-variant-2-v2',
     CodyAutocompleteFIMModelExperimentVariant3 = 'cody-autocomplete-fim-model-experiment-variant-3-v2',
     CodyAutocompleteFIMModelExperimentVariant4 = 'cody-autocomplete-fim-model-experiment-variant-4-v2',
-
+    CodyAutocompleteDisableLowPerfLangDelay = 'cody-autocomplete-disable-low-perf-lang-delay',
     // Enables Claude 3 if the user is in our holdout group
     CodyAutocompleteClaude3 = 'cody-autocomplete-claude-3',
     // Enable latency adjustments based on accept/reject streaks
--- a/vscode/src/completions/artificial-delay.test.ts
+++ b/vscode/src/completions/artificial-delay.test.ts
@@ -1,7 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { getArtificialDelay, lowPerformanceLanguageIds, resetArtificialDelay } from './artificial-delay'
-
 const featureFlags = {
     user: true,
 }
@@ -10,9 +9,26 @@ describe('getArtificialDelay', () => {
     beforeEach(() => {
         vi.useFakeTimers()
     })
+
     afterEach(() => {
         vi.restoreAllMocks()
         resetArtificialDelay()
+    })
+
+    it('returns no artificial delay when codyAutocompleteDisableLowPerfLangDelay is true for a low performance language', () => {
+        const uri = 'file://foo/bar/test.css'
+        // css is a low performance language
+        const languageId = 'css'
+        expect(lowPerformanceLanguageIds.has(languageId)).toBe(true)
+        const params = {
+            featureFlags,
+            uri,
+            languageId,
+            codyAutocompleteDisableLowPerfLangDelay: true,
+        }
+
+        // expect no artificial delay when codyAutocompleteDisableLowPerfLangDelay is true
+        expect(getArtificialDelay(params)).toBe(0)
     })
 
     it('returns gradually increasing latency up to max for CSS when suggestions are rejected', () => {
@@ -21,51 +37,57 @@ describe('getArtificialDelay', () => {
         // css is a low performance language
         const languageId = 'css'
         expect(lowPerformanceLanguageIds.has(languageId)).toBe(true)
-
+        const codyAutocompleteDisableLowPerfLangDelay = false
+        const params = {
+            featureFlags,
+            uri,
+            languageId,
+            codyAutocompleteDisableLowPerfLangDelay,
+        }
         // start with default high latency for low performance lang with default user latency added
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
         // start at default, but gradually increasing latency after 5 rejected suggestions
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1050)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1100)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1150)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1200)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1250)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1300)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1350)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1400)
+        expect(getArtificialDelay(params)).toBe(1050)
+        expect(getArtificialDelay(params)).toBe(1100)
+        expect(getArtificialDelay(params)).toBe(1150)
+        expect(getArtificialDelay(params)).toBe(1200)
+        expect(getArtificialDelay(params)).toBe(1250)
+        expect(getArtificialDelay(params)).toBe(1300)
+        expect(getArtificialDelay(params)).toBe(1350)
+        expect(getArtificialDelay(params)).toBe(1400)
         // max latency at 1400
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1400)
+        expect(getArtificialDelay(params)).toBe(1400)
         // after the suggestion was accepted, user latency resets to 0, using baseline only
         resetArtificialDelay()
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
         resetArtificialDelay()
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
         // gradually increasing latency after 5 rejected suggestions
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1050)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1100)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1150)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1200)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1250)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1300)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1350)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1400)
+        expect(getArtificialDelay(params)).toBe(1050)
+        expect(getArtificialDelay(params)).toBe(1100)
+        expect(getArtificialDelay(params)).toBe(1150)
+        expect(getArtificialDelay(params)).toBe(1200)
+        expect(getArtificialDelay(params)).toBe(1250)
+        expect(getArtificialDelay(params)).toBe(1300)
+        expect(getArtificialDelay(params)).toBe(1350)
+        expect(getArtificialDelay(params)).toBe(1400)
         // Latency will not reset before 5 minutes
         vi.advanceTimersByTime(3 * 60 * 1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1400)
+        expect(getArtificialDelay(params)).toBe(1400)
         // Latency will be reset after 5 minutes
         vi.advanceTimersByTime(5 * 60 * 1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
         // reset latency on accepted suggestion
         resetArtificialDelay()
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
     })
 
     it('returns increasing latency after rejecting suggestions', () => {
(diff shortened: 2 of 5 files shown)


vscode/src/completions/artificial-delay.test.ts
```diff
@@ -1,163 +1,191 @@
...
     })
 
-    it('returns gradually increasing latency up to max for CSS when suggestions are rejected', () => {
+    it('returns no artificial delay when CodyAutocompleteTracing feature flag is enabled', async () => {
+        const uri = 'file://foo/bar/test.css'
+        enabledFeatureFlags.add(FeatureFlag.CodyDisableArtificialDelay)
+
+        // css is a low performance language
+        const languageId = 'css'
+        expect(lowPerformanceLanguageIds.has(languageId)).toBe(true)
```
dominiccooney, line 39: This unit testing style where you're asserting your assumptions is beautiful. Nice work.

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -1,163 +1,191 @@
...
+
     beforeEach(() => {
         vi.useFakeTimers()
+        originalFeatureFlagProviderInstance = featureFlagProvider.instance
+        enabledFeatureFlags.clear()
+        featureFlagProvider.instance = {
+            evaluateFeatureFlag: (flag: FeatureFlag) => Promise.resolve(enabledFeatureFlags.has(flag)),
+            refresh: () => {},
+            getFromCache: (flagName: FeatureFlag) => {
+                return false
```
dominiccooney, line 22: Oooh this looks subtle or wrong. Is this used? If not, let's throw here to fail noisily. If so, why would we consult `enabledFeatureFlags` the first time, but not when we're checking the cache?

dominiccooney (approved):
Is getting feature flags really fast, when they're cached? I'm just wondering if making `getFeatureFlag` async means you're going to introduce another kind of random delay that depends on how busy the main thread is.

Could we drill down on the point mentioned in the test?

vscode/src/completions/artificial-delay.test.ts
arafatkatze: Yeah it was used elsewhere but not here so I removed it.

arafatkatze: Full disclosure I was writing some other tests which entailed mocking the feature flags and those tests would hit [this part of the codebase](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=artificial-latency-fix&file=vscode%2Fsrc%2Fcompletions%2Fcompletion-provider-config.ts&editor=VSCode&version=2.2.17&start_row=48&start_col=26&end_row=48&end_col=26)  where you see getFromCache method being called so I had to specifically mock out that thing(NOTE: It tries to get the feature flag value from cache and that behavior is slightly different from evaluateFeatureFlag which also calls [getFromCache](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=artificial-latency-fix&file=lib%2Fshared%2Fsrc%2Fexperimentation%2FFeatureFlagProvider.ts&editor=VSCode&version=2.2.17&start_row=127&start_col=32&end_row=127&end_col=32) here). 

I kept that thing here because I thought it was necessary but turns out that its not so I got rid of that and started failing loudly.

lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -30,7 +30,8 @@ export enum FeatureFlag {
     CodyAutocompleteFIMModelExperimentVariant2 = 'cody-autocomplete-fim-model-experiment-variant-2-v2',
     CodyAutocompleteFIMModelExperimentVariant3 = 'cody-autocomplete-fim-model-experiment-variant-3-v2',
     CodyAutocompleteFIMModelExperimentVariant4 = 'cody-autocomplete-fim-model-experiment-variant-4-v2',
-
+    CodyInLineSuffixAutocomplete = 'cody-in-line-suffix-autocomplete',
```
valerybugakov, line 33: This should live in a separate PR, right?

lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -30,7 +30,8 @@ export enum FeatureFlag {
...
+    CodyDisableArtificialDelay = 'cody-disable-artificial-delay',
```
valerybugakov, line 34: Let's use the same `cody-autocomplete-x-y-z` pattern for feature flag names. Makes it easier to find all active autocomplete flags on DotCom in the site-admin.

vscode/src/completions/artificial-delay.ts
```diff
@@ -42,12 +42,19 @@ let userMetrics = {
...
-export function getArtificialDelay(
+export async function getArtificialDelay(
     featureFlags: LatencyFeatureFlags,
     uri: string,
     languageId: string,
     completionIntent?: CompletionIntent
-): number {
+): Promise<number> {
+    const codyDisableArtificialDelay = await featureFlagProvider.instance!.evaluateFeatureFlag(
+        FeatureFlag.CodyDisableArtificialDelay
```
valerybugakov, line 52: As Dominic pointed out, we should prefetch this feature using `vscode/src/completions/completion-provider-config.ts` and access the value here in a sync way.

lib/shared/src/experimentation/FeatureFlagProvider.ts
arafatkatze: Sorry with a dozen rebases things get mixed up

lib/shared/src/experimentation/FeatureFlagProvider.ts
arafatkatze: Done

vscode/src/completions/completion-provider-config.ts
```diff
@@ -161,6 +162,10 @@ class CompletionProviderConfig {
                 return 0
         }
     }
+
+    public get autocompleteEnableAritificialDelay(): boolean {
+        return this.getPrefetchedFlag(FeatureFlag.CodyAutocompleteDisableArtificialDelay)
+    }
```
valerybugakov, line 168: Not used, can be deleted?

vscode/src/completions/artificial-delay.ts
```diff
@@ -46,8 +45,12 @@ export function getArtificialDelay(
     featureFlags: LatencyFeatureFlags,
     uri: string,
     languageId: string,
+    CodyAutocompleteDisableArtificialDelay: boolean,
     completionIntent?: CompletionIntent
 ): number {
+    if (CodyAutocompleteDisableArtificialDelay) {
```
valerybugakov, line 51: See related logic below:
```ts
// Increase latency linearly up to max after 5 rejected suggestions
```

It's reasonable to keep using it. We just want to get rid of the baseline increase for low-perf languages and AST nodes.

vscode/src/completions/artificial-delay.ts
```diff
@@ -46,8 +45,12 @@ export function getArtificialDelay(
     featureFlags: LatencyFeatureFlags,
     uri: string,
     languageId: string,
+    CodyAutocompleteDisableArtificialDelay: boolean,
```
valerybugakov, line 48: Usually all the arguments use camelCase for naming.

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -196,21 +227,21 @@ describe('getArtificialDelay', () => {
...
+        expect(getArtificialDelay(featureFlagsNoUser, uri, languageId, false)).toBe(1000)
+        expect(getArtificialDelay(featureFlagsNoUser, uri, languageId, false)).toBe(1000)
 
         // switch to a non-low-performance language - go is not a low performance language
         const goLanguageId = 'go'
         const goUri = 'foo/bar/test.go'
         expect(lowPerformanceLanguageIds.has(goLanguageId)).toBe(false)
         // reset to provider latency because language latency is ignored for non-low-performance languages
-        expect(getArtificialDelay(featureFlagsNoUser, goUri, goLanguageId)).toBe(0)
+        expect(getArtificialDelay(featureFlagsNoUser, goUri, goLanguageId, false)).toBe(0)
```
valerybugakov, line 245: Let's convert the argument list of an object to make it clear what `false` means.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -414,11 +414,14 @@ export class InlineCompletionItemProvider
                     FeatureFlag.CodyAutocompleteUserLatency
                 ),
             }
-
+            const CodyAutocompleteDisableArtificialDelay = completionProviderConfig.getPrefetchedFlag(
```
valerybugakov, line 417: Nice!

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -1,163 +1,194 @@
...
     })
+
     afterEach(() => {
+        featureFlagProvider.instance = originalFeatureFlagProviderInstance
         vi.restoreAllMocks()
         resetArtificialDelay()
     })
 
-    it('returns gradually increasing latency up to max for CSS when suggestions are rejected', () => {
+    it('returns no artificial delay when CodyAutocompleteTracing feature flag is enabled', async () => {
```
valerybugakov, line 36: `CodyAutocompleteTracing` — wrong flag name?

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -1,163 +1,194 @@
...
 
 describe('getArtificialDelay', () => {
+    const enabledFeatureFlags = new Set<FeatureFlag>()
+    let originalFeatureFlagProviderInstance: FeatureFlagProvider | null
+
     beforeEach(() => {
         vi.useFakeTimers()
+        originalFeatureFlagProviderInstance = featureFlagProvider.instance
+        enabledFeatureFlags.clear()
+        featureFlagProvider.instance = {
```
valerybugakov, line 18: We can use `vi.spyOn` here to spy on relevant `featureFlagProvider` methods and replace their return values with `spy.mockReturnValue()`. Or we can mock the whole import by the path. With that in place, `vi.restoreAllMocks()` would make `featureFlagProvider.instance = originalFeatureFlagProviderInstance` obsolete.

Also, we can spy on the graphql client, which resolves feature flag values. That would test the feature flag provider logic, too.

arafatkatze:
@valerybugakov Apologies I was still in the middle of fixing the PR when you saw the new changes. It wasn't in the right state(I will try to be mindful about communicating that the PR is not in the right state)

vscode/src/completions/artificial-delay.test.ts
arafatkatze: @valerybugakov I got something better. I adopted the new logic to get the value of the feature flag [by doing a subscription](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=artificial-latency-fix&file=vscode%2Fsrc%2Fcompletions%2Fcompletion-provider-config.ts&editor=VSCode&version=2.2.17&start_row=197&start_col=14&end_row=204&end_col=5)
 and then changed the function to have the args 
```

export function getArtificialDelay(params: {
    featureFlags: LatencyFeatureFlags
    uri: string
    languageId: string
    codyCompletionDisableLowPerfLangDelay: boolean
    completionIntent?: CompletionIntent
}): number {

```
This made the testing of the function much easier as I only have to tweak the params `codyCompletionDisableLowPerfLangDelay`. 


NOTE: 

1. The logic of subscription to a feature flag + GQL CLient is also tested [here](https://sourcegraph.com/-/editor?remote_url=https%3A%2F%2Fgithub.com%2Fsourcegraph%2Fcody.git&branch=artificial-latency-fix&file=lib%2Fshared%2Fsrc%2Fexperimentation%2FFeatureFlagProvider.test.ts&editor=VSCode&version=2.2.17&start_row=203&start_col=0&end_row=216&end_col=14)(but that's a separate thing).

vscode/src/completions/artificial-delay.ts
arafatkatze: Thanks for the clarification I made the. changes.

vscode/src/completions/completion-provider-config.ts
arafatkatze: Fixed the logic in the new PR

vscode/src/completions/artificial-delay.ts
arafatkatze: Fixed

vscode/src/completions/artificial-delay.test.ts
arafatkatze: Done

arafatkatze:
@valerybugakov I updated the PR again with all the recommendations that you suggested and now its a lot lot cleaner. 

With the latest changes from Quinn I have the subscription/update logic for feature flags also(Read[ this comment](https://github.com/sourcegraph/cody/pull/5480#discussion_r1760381822))

As for testing I have also incorporated the changes that you recommended so its easy to understand the tests too. 

Thanks again for such a thorough review, I would like another one now.

vscode/src/completions/artificial-delay.test.ts
valerybugakov: This is great!

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -435,13 +453,14 @@ export class InlineCompletionItemProvider
...
-                completionIntent
-            )
+            const params = {
+                featureFlags: latencyFeatureFlags,
+                uri: document.uri.toString(),
+                languageId: document.languageId,
+                codyCompletionDisableLowPerfLangDelay: this.disableLowPerfLangDelay,
+                completionIntent: completionIntent,
+            }
+            const artificialDelay = getArtificialDelay({ ...params })
```
valerybugakov, line 463: This way, the created object is type-checked on the declaration. Also, we don't need to create a new object from `params`.

```suggestion
            const artificialDelay = getArtificialDelay({
                featureFlags: latencyFeatureFlags,
                uri: document.uri.toString(),
                languageId: document.languageId,
                codyCompletionDisableLowPerfLangDelay: this.disableLowPerfLangDelay,
                completionIntent,
            })
```

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -435,13 +453,14 @@ export class InlineCompletionItemProvider
...
             }
-
-            const artificialDelay = getArtificialDelay(
-                latencyFeatureFlags,
-                document.uri.toString(),
-                document.languageId,
-                completionIntent
-            )
+            const params = {
+                featureFlags: latencyFeatureFlags,
```
valerybugakov, line 457: We can remove the `FeatureFlag.CodyAutocompleteUserLatency` feature flag and all the related logic. It's not actively used, and we do not plan to conduct new experiments with it. This can be done in a follow-up PR or here.

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -179,7 +181,23 @@ export class InlineCompletionItemProvider
...
+        // Subscribe to changes in disableLowPerfLangDelay and update the value accordingly
+        this.disposables.push(
+            subscriptionDisposable(
+                completionProviderConfig.completionDisableLowPerfLangDelay.subscribe(
+                    disableLowPerfLangDelay => {
+                        this.disableLowPerfLangDelay = disableLowPerfLangDelay
+                    }
+                )
+            )
+        )
```
valerybugakov, line 200: Do we need both of these? There's an example of using `CodyAutocompleteTracing` in the `InlineCompletionItemProvider` constructor, which relies only on one call sequence to initialize and keep the local value up to date.

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -10,158 +9,218 @@ describe('getArtificialDelay', () => {
...
     })
 
-    it('returns default latency for low performance language only when only language flag is enabled', () => {
+    it('returns default latency for low performance language only when only language flag is enabled', async () => {
         const uri = 'file://foo/bar/test.css'
+        const codyCompletionDisableLowPerfLangDelay = false
 
         const featureFlagsLangOnly = {
-            user: false,
+            user: codyCompletionDisableLowPerfLangDelay,
```
valerybugakov, line 223: Is this a property for a different feature flag? `CodyAutocompleteUserLatency`

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -171,46 +230,66 @@ describe('getArtificialDelay', () => {
...
+        expect(getArtificialDelay({ ...params, languageId: lowPerformLanguageId })).toBe(1000)
+        expect(getArtificialDelay({ ...params, languageId: lowPerformLanguageId })).toBe(1000)
+        expect(getArtificialDelay({ ...params, languageId: lowPerformLanguageId })).toBe(1000)
         // latency back to 0 when language is no longer low-performance
-        expect(getArtificialDelay(featureFlagsLangOnly, goUri, languageId)).toBe(0)
+        expect(getArtificialDelay({ ...params, uri: goUri, languageId })).toBe(0)
     })
 
-    it('returns latency based on language only when user flag is disabled', () => {
+    it('returns latency based on language only when user flag is disabled', async () => {
```
valerybugakov, line 252: Do we still need `async` callbacks here?

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -10,158 +9,218 @@ describe('getArtificialDelay', () => {
...
+            languageId,
+            codyCompletionDisableLowPerfLangDelay,
+        }
         // start with default high latency for low performance lang with default user latency added
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
-        expect(getArtificialDelay(featureFlags, uri, languageId)).toBe(1000)
+        expect(getArtificialDelay(params)).toBe(1000)
```
valerybugakov, line 50: Nice!

vscode/src/completions/artificial-delay.test.ts
arafatkatze: Good catch I removed them

vscode/src/completions/inline-completion-item-provider.ts
arafatkatze: Will do it in a followup PR for sure.

vscode/src/completions/inline-completion-item-provider.ts
arafatkatze: Both of these are technically NOT needed but I wanted to maintain the logical consistency with "initializing" something and then subscribing to its value but yeah since its redundant I don't see the reason to actually keep it for anything other than having a reasonably understandable logic.

vscode/src/completions/artificial-delay.test.ts
```diff
@@ -171,46 +228,66 @@ describe('getArtificialDelay', () => {
...
         const uri = 'file://foo/bar/test.css'
 
         // css is a low performance language
         const languageId = 'css'
+        const codyAutocompleteDisableLowPerfLangDelay = false
         expect(lowPerformanceLanguageIds.has(languageId)).toBe(true)
 
         const featureFlagsNoUser = {
-            user: false,
+            user: codyAutocompleteDisableLowPerfLangDelay,
```
valerybugakov, line 259: We need to update the variable name here because it doesn't match the underlying feature flag.

valerybugakov (approved):
Thank you for addressing all the comments! 💜

vscode/src/completions/artificial-delay.test.ts
arafatkatze: Done

PR #5481, opened 2024-09-05
Add option to disable ctx providers

pkukielka:
## Test plan

N/A, unittests should pass.

When it will get adopted in JetBrains test plan will be as follow: 
https://github.com/sourcegraph/cody/pull/5384

### DIFF
--- a/lib/shared/src/mentions/api.test.ts
+++ b/lib/shared/src/mentions/api.test.ts
@@ -1,28 +1,19 @@
-import type { Observable } from 'observable-fns'
 import { describe, expect, it } from 'vitest'
+import { firstValueFrom } from '../misc/observable'
 import {
     FILE_CONTEXT_MENTION_PROVIDER,
     SYMBOL_CONTEXT_MENTION_PROVIDER,
     mentionProvidersMetadata,
 } from './api'
 
-function waitForObservable<T>(observable: Observable<T>): Promise<T> {
-    return new Promise<T>(resolve => {
-        const subscription = observable.subscribe(value => {
-            resolve(value)
-            subscription.unsubscribe()
-        })
-    })
-}
-
 describe('mentionProvidersMetadata', () => {
     it('should return all providers when no options are provided', async () => {
-        const providers = await waitForObservable(mentionProvidersMetadata())
+        const providers = await firstValueFrom(mentionProvidersMetadata())
         expect(providers.length).toBeGreaterThanOrEqual(2)
     })
     it('should filter out disabled providers', async () => {
         const options = { disableProviders: [FILE_CONTEXT_MENTION_PROVIDER.id] }
-        const providers = await waitForObservable(mentionProvidersMetadata(options))
+        const providers = await firstValueFrom(mentionProvidersMetadata(options))
         expect(providers.length).toBeGreaterThanOrEqual(1)
         expect(providers).toContain(SYMBOL_CONTEXT_MENTION_PROVIDER)
         expect(providers).not.toContain(FILE_CONTEXT_MENTION_PROVIDER)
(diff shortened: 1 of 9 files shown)


lib/shared/src/mentions/api.ts
```diff
@@ -51,21 +51,20 @@ export const SYMBOL_CONTEXT_MENTION_PROVIDER: ContextMentionProviderMetadata & {
     emptyLabel: 'No symbols found',
 }
 
-/** Metadata for all registered {@link ContextMentionProvider}s. */
-export function allMentionProvidersMetadata(): Observable<ContextMentionProviderMetadata[]> {
+const availableMentionProviders = [FILE_CONTEXT_MENTION_PROVIDER, SYMBOL_CONTEXT_MENTION_PROVIDER]
+export type OptionalProviderId = (typeof availableMentionProviders)[number]['id']
+
+export function mentionProvidersMetadata(
+    disabledMentionsProviders?: OptionalProviderId[] | undefined | null
```
dominiccooney, line 58: Making this an object so the providers are labeled with `disable` will be more readable now and evolvable in future, something like:

```
mentionProvidersMetadata({disable: [SYMBOL_CONTEXT_MENTION_PROVIDER]})
```

lib/shared/src/mentions/api.ts
```diff
@@ -51,21 +51,20 @@ export const SYMBOL_CONTEXT_MENTION_PROVIDER: ContextMentionProviderMetadata & {
...
+): Observable<ContextMentionProviderMetadata[]> {
     return openCtxMentionProviders().map(providers => [
-        FILE_CONTEXT_MENTION_PROVIDER,
-        SYMBOL_CONTEXT_MENTION_PROVIDER,
+        ...availableMentionProviders.filter(
+            provider => !disabledMentionsProviders?.includes(provider.id)
+        ),
         ...providers,
```
dominiccooney, line 64: Why only filter the `availableMentionProviders`? Why not others?

dominiccooney (changes requested):
A nice generalization of what web and JetBrains needs.

Let's filter all the providers, not just the two built-in ones? (Or make the API design a bit clearer...)

Let's add a unit test.

lib/shared/src/mentions/api.ts
pkukielka: Fixed

lib/shared/src/mentions/api.ts
pkukielka: Done

lib/shared/src/mentions/api.test.ts
```diff
@@ -0,0 +1,38 @@
+import type { Observable } from 'observable-fns'
+import { describe, expect, it } from 'vitest'
+import {
+    FILE_CONTEXT_MENTION_PROVIDER,
+    SYMBOL_CONTEXT_MENTION_PROVIDER,
+    mentionProvidersMetadata,
+} from './api'
+
+function waitForObservable<T>(observable: Observable<T>): Promise<T> {
```
dominiccooney, line 9: This already exists, it is `firstValueFrom`

lib/shared/src/mentions/api.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        const subscription = observable.subscribe(value => {
+            resolve(value)
+            subscription.unsubscribe()
+        })
+    })
+}
+
+describe('mentionProvidersMetadata', () => {
+    it('should return all providers when no options are provided', async () => {
+        const providers = await waitForObservable(mentionProvidersMetadata())
```
dominiccooney, line 20:
```suggestion
        const providers = await firstValueFrom(mentionProvidersMetadata())
```

dominiccooney (approved):
Thank you for writing the test.

The test helper function can be de-duped with `firstValueFrom`, PTAL.

PR #5483, opened 2024-09-05
fix(Generate Commit Message): handle file paths in git diff on Windows

abeatrix:
FIX https://github.com/sourcegraph/cody/issues/4863

This PR fixes an issue where the generate commit message doesn't work on Windows:

image

Root cause: the displayPath function returns file path native to the OS system, while the file path returned by the diff output is following the unix one, causing the file match logic to always be false.

image

This PR fixed this by always replace all \\ used by Windows with '/' before doing matching.

Summary:

- Fixed an issue where the `Generate Commit Message` command would fail on Windows due to file path handling in the `getContextFilesFromGitDiff` function.
- Normalized file paths to use forward slashes instead of backslashes to ensure consistent handling across platforms.
- Improved the logic for matching the git diff output to the actual file URIs to handle nested workspaces and other edge cases.

## Test plan

Added new unit tests to cover this feature.

Tested in Window:

image

## Changelog

Command: Fixed the `Generate Commit Message` command on Windows caused by file path. [pull/5483](https://github.com/sourcegraph/cody/pull/5483) ?

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -9,6 +9,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 ### Fixed
 
 - Cody Ignore: Fixed an issue where Cody would treat Notebook cells as ignored files when .cody/ignore is enabled. [pull/5473](https://github.com/sourcegraph/cody/pull/5473)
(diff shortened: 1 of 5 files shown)


kalanchan (approved):
niceeee!! 😍 

can you add a description under the changelog section in the PR description? james and i will be introducing a new process soon!

PR #5490, opened 2024-09-06
fix(dev): fix params used by groqChatClient

abeatrix:
Fix: https://linear.app/sourcegraph/issue/CODY-3672

This PR fixes an issue where the `chatParams` was being passed as a parameter object when using the local GROQ model provider, causing all network requests to fail when using the local config models.

The fix is to properly destructure the `chatParams` object before passing it to the fetch request.

## Changes

- Destructure the `chatParams` object before passing it to the fetch request in the `groqChatClient` function.

## Test plan

- Verified that the local GROQ model provider now works correctly when using the `chatParams` object.

## Changelog

Dev: Fixed an issue where incorrect request parameters caused stream requests to fail when using local OpenAI-compatible model provider through the groqChatClients.

### DIFF
--- a/lib/shared/src/llm-providers/groq/chat-client.ts
+++ b/lib/shared/src/llm-providers/groq/chat-client.ts
@@ -92,7 +92,7 @@ export async function groqChatClient({
     // Stream requests
     fetch(config?.endpoint ?? GROQ_CHAT_API_URL, {
         method: 'POST',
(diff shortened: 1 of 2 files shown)


kalanchan:
HEY BUG BASH HASN'T STARTED YET 😆

vovakulikov (approved):

PR #5493, opened 2024-09-06
telemetry: add billing categorization to cody events 1

dadlerj:
In accordance with [RFC 978](https://docs.google.com/document/d/1EvyH1kaU-fsn59h-DyoaO2Qy4C2TYLmDGjG73bdb2V4/edit#heading=h.trqab8y0kufp)

## Test plan

Low impact change—testing here is only ensuring that this passes CI and builds correctly without errors.

### DIFF
--- a/lib/shared/src/telemetry-v2/index.ts
+++ b/lib/shared/src/telemetry-v2/index.ts
@@ -2,10 +2,10 @@
  * Events accept billing metadata for ease of categorization in analytics
  * pipelines - this type enumerates known categories.
  */
-export type BillingCategory = 'exampleBillingCategory'
+export type BillingCategory = 'exampleBillingCategory' | 'core' | 'billable'
 
 /**
  * Events accept billing metadata for ease of categorization in analytics
  * pipelines - this type enumerates known products.
  */
-export type BillingProduct = 'exampleBillingProduct'
+export type BillingProduct = 'exampleBillingProduct' | 'cody'
--- a/lib/shared/src/telemetry-v2/index.ts
+++ b/lib/shared/src/telemetry-v2/index.ts
@@ -2,10 +2,10 @@
  * Events accept billing metadata for ease of categorization in analytics
  * pipelines - this type enumerates known categories.
  */
-export type BillingCategory = 'exampleBillingCategory' | 'core' | 'billable'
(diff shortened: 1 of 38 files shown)


dadlerj:
I tagged a few people for reviews, but none of you have to do a detailed, line-by-line review here. I just wanted to get a few opinions before merging:

* @bobheadxi confirmation that this looks right to you considering the v2 telemetry architecture
* @abeatrix confirmation that this change "looks" fine in the context of the Cody codebase
* @akalia25 just tagged for awareness

More such PRs will be coming for our other codebases, and I've likely missed a few in sourcegraph/cody as well, so this is only an early proof of concept. A bigger announcement about this will come once our _existing_ events have this set up.

lib/shared/src/telemetry-v2/index.ts
```diff
@@ -2,10 +2,10 @@
...
  */
-export type BillingCategory = 'exampleBillingCategory'
+export type BillingCategory = 'exampleBillingCategory' | 'core' | 'billable'
 
 /**
  * Events accept billing metadata for ease of categorization in analytics
  * pipelines - this type enumerates known products.
  */
-export type BillingProduct = 'exampleBillingProduct'
+export type BillingProduct = 'exampleBillingProduct' | 'cody'
```
bobheadxi, line 11: I think now that we have real values, we can remove the example ones!

vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -83,6 +83,11 @@ function setLastStoredCode(
             source,
             op: operation,
         },
+        billingMetadata: {
+            product: 'cody',
+            // 🚨 ensure that any new event names added qualify as core events, or update this parameter.
```
bobheadxi, line 88: I think this docstring might be good to have in the switch case above too, otherwise someone editing the code above might miss it

bobheadxi (approved):
Nice, LGTM! Excited to see this getting used 😁

lib/shared/src/telemetry-v2/index.ts
akalia25: Similarly, to `BillingCategory` would it make sense to list out all the product options we have:
`cody | search | codeInsights | batchChanges | other` ?

PR #5501, opened 2024-09-09
Add a test for smart apply

mkondratek:
## Test plan

- green CI

### DIFF
--- a/agent/src/edit.test.ts
+++ b/agent/src/edit.test.ts
@@ -1,34 +1,20 @@
-import fspromises from 'node:fs/promises'
-import os from 'node:os'
 import path from 'node:path'
 import {
-    ChatClient,
-    type GraphQLAPIClientConfig,
     contextFiltersProvider,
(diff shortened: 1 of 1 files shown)


agent/src/edit.test.ts
```diff
@@ -49,6 +62,44 @@ describe('Edit', () => {
...
     })
 
+    it('editCommands/code (having one file active apply to the other file)', async () => {
+        const initiallyActiveFileUri = workspace.file('src', 'sum.ts')
+        await client.openFile(initiallyActiveFileUri)
+
+        const fileToEditUri = workspace.file('src', 'a-new-file.ts')
+        const newDocument = { uri: fileToEditUri } as vscode.TextDocument
+
+        const fixupTask = await executeSmartApply({
```
mkondratek, line 63: For some reason the code from `EditManager.smartApplyEdit` is not being executed. Can we use commands from tests?

PR #5502, opened 2024-09-09
improve config-watching and singletons

sqs:
- Expose the full resolved config value as a global `resolvedConfig: Observable<ResolvedConfig>` and make a bunch of places use this
- Do the same for `authStatus`
- Make featureFlagProvider, authStatus, resolvedConfig, repoNameResolver, and contextFiltersProvider simple global singletons

This helps us get to a better and more consistent way of reading and reacting to config changes throughout Cody.

## Test plan

CI


agent/src/agent.ts
```diff
@@ -407,7 +407,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
                 )
             }
             if (clientInfo.capabilities?.ignore === 'enabled') {
-                contextFiltersProvider.instance!.onContextFiltersChanged(() => {
+                contextFiltersProvider.onContextFiltersChanged(() => {
```
valerybugakov, line 410: 🤌

vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -76,7 +76,6 @@ export function createInlineCompletionItemProvider({
                     if (provider) {
                         const authStatus = authProvider.instance!.statusAuthed
                         const completionsProvider = new InlineCompletionItemProvider({
-                            authStatus,
```
valerybugakov, line 79: 💜

valerybugakov (approved):

sqs:
@olafurpg I am getting 2 weird kinds of errors. Could you please help me out here?

### Polly recording errors

```
There exists a recording with exactly the same request body, but for some reason the recordings did not match.
This only really happens in exceptional cases like
- There is a bug in how Polly computes HTTP request identifiers
- Somebody manually edited the HTTP recording file
Possible ways to fix the problem:
- Confirm tests run in passthrough mode: CODY_RECORDING_MODE=passthrough pnpm test agent/src/index.test.ts
- Reset recordings and re-record everything: rm -rf agent/recordings && pnpm update-agent-recordings
```

I see this for the `Repository` and `SiteProductVersion` queries in almost all HAR files, and for many completions requests as well. I was only able to fix some of them by doing manual surgery on the HAR files and adding duplicate entries (since the requests needed to remain in the HAR files with the old IDs as well or the same issue would occur). 

Repro on this branch with `pnpm run test agent/src/document-code.test.ts -t 'commands/document'`.

### SCIP codegen errors

`pnpm run generate-agent-kotlin-bindings` gives this error, even though these types haven't changed on this branch:

```
error: /Users/sqs/src/github.com/sourcegraph/cody/lib/shared/src/auth/types.ts:53:4:
   1|     authenticated: false
    |     ^^^^^^^^^^^^^
Incompatible signatures. For discriminated unions, each property name must map to a unique type.
For example, it's not allowed to have a property named 'result', which is a string for one type in the
discrimated union and a number for another type in the union. To fix this problem, give one of the
following properties a unique name and try running the code generator again.
error: /Users/sqs/src/github.com/sourcegraph/cody/lib/shared/src/auth/types.ts:16:4: conflict here
   1|     authenticated: true
    |     ^^^^^^^^^^^^^

error: /Users/sqs/src/github.com/sourcegraph/cody/lib/shared/src/auth/types.ts:53:4:
   2|     authenticated: false
    |     ^^^^^^^^^^^^^
Incompatible signatures. For discriminated unions, each property name must map to a unique type.
For example, it's not allowed to have a property named 'result', which is a string for one type in the
discrimated union and a number for another type in the union. To fix this problem, give one of the
following properties a unique name and try running the code generator again.
error: /Users/sqs/src/github.com/sourcegraph/cody/lib/shared/src/auth/types.ts:16:4: conflict here
   2|     authenticated: true
    |     ^^^^^^^^^^^^^

error: /Users/sqs/src/github.com/sourcegraph/cody/lib/shared/src/auth/types.ts:53:4:
   3|     authenticated: false
    |     ^^^^^^^^^^^^^
Incompatible signatures. For discriminated unions, each property name must map to a unique type.
For example, it's not allowed to have a property named 'result', which is a string for one type in the
discrimated union and a number for another type in the union. To fix this problem, give one of the
following properties a unique name and try running the code generator again.
error: /Users/sqs/src/github.com/sourcegraph/cody/lib/shared/src/auth/types.ts:16:4: conflict here
   3|     authenticated: true
    |     ^^^^^^^^^^^^^

3 errors
```

But the discriminated union's property name (`authenticated`) does map to unique types. And, again, the type did not change on this branch.

valerybugakov:
@sqs, I took a quick look into it. Shared my finding here: https://github.com/sourcegraph/cody/pull/5519. Curious if you can repro the same locally.

valerybugakov:
UPD: fixed document command tests here: https://github.com/sourcegraph/cody/pull/5519/commits/c3cd6d39131dfa39eef5a017319041cca5016890

olafurpg:
I pushed a commit adding widening of boolean/true/false types so that they're not treated a conflicting types in discriminated unions. The `AuthStatus` type becomes a flat data class in Java/Kotlin/C#, which requires that each property has a "compatible" type in the union

```
data class AuthStatus(
  val endpoint: String,
  val authenticated: Boolean, // <- it was struggling to widen true/false into Boolean here
  val showNetworkError: Boolean? = null,
  val showInvalidAccessTokenError: Boolean? = null,
  val username: String,
  val isFireworksTracingEnabled: Boolean? = null,
  val hasVerifiedEmail: Boolean? = null,
  val requiresVerifiedEmail: Boolean? = null,
  val siteVersion: String,
  val codyApiVersion: Long,
  val configOverwrites: CodyLLMSiteConfiguration? = null,
  val primaryEmail: String? = null,
  val displayName: String? = null,
  val avatarURL: String? = null,
  val userCanUpgrade: Boolean? = null,
  val isOfflineMode: Boolean? = null,
)
```

PR #5506, opened 2024-09-09
Fix lenses provider

pkukielka:
Fixes https://github.com/sourcegraph/jetbrains/issues/2240

## Test plan

**Steps to reproduce:**
Open a source code file and run Generate Unit Tests command for any method

**Observed behavior:**
Focus will not be shifted to newly created unit test file

**Expected behavior:**
Focus should be shifted to unit test file


vscode/src/non-stop/codelenses/provider.ts
```diff
@@ -171,7 +173,7 @@ export class FixupCodeLenses implements vscode.CodeLensProvider, FixupControlApp
     private removeLensesFor(task: FixupTask): void {
         if (this.taskLenses.delete(task)) {
             // TODO: Clean up the fixup file when there are no remaining code lenses
-            this.notifyCodeLensesChanged(task.document.uri, [])
+            this.notifyCodeLensesChanged(task.fixupFile.uri, [])
```
mkondratek, line 176: Wow, we probably should have some better names for these. What do they mean actually and why cam they be different?

mkondratek (approved):
Lgtm

mkondratek (approved):
Tested with scenario from: https://github.com/sourcegraph/jetbrains/issues/2197.

Works 🎉

PR #5508, opened 2024-09-09
Chat: waitlist for OpenAI-o1 & OpenAI-o1 mini

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3681/improve-support-for-previewexperimental-models

Related to https://github.com/sourcegraph/sourcegraph/pull/323

- Added `OpenAI o1` and `OpenAI o1 mini` to dropdown list where users can join the waitlist
- Clicking on `Join Waitlist` will open the blog post, fire a telemetry event called `cody.joinLlmWaitlist` and update the `Join Waitlist` label to `On Waitlist` 
- Added a new `ModelTag.StreamDisabled` tag to indicate models that do not support streaming
- Updated the `AssistantMessageCell` and `HumanMessageEditor` components to handle models without streaming support
  - Displayed a message to the user when a non-streaming model is used
  - Filtered out the initial codebase context when using a non-streaming model to avoid longer processing times
  - Display `Try again with a different model` instead of `Try again with different context` below each assistant response
- Updated the `ModelSelectField` component to display an "Early Access" badge for models with the `ModelTag.Preview` tag
- Expanded the `ModelRef` interface and `ModelsService` to better handle model ID matching
- Adds the ability to duplicate the current chat session in the Cody chat interface. 
  - When the "Try again with a different model" button is clicked, a new chat session is created that is a copy of the current session, allowing the user to resubmit their request with a different model.
  - The changes include:
    - Adding a new `chatSession` command to the webview message protocol, with `duplicate` and `new` actions
    - Implementing the `duplicateSession` method in the `ChatController` to create a new session with a unique ID based on the current session
    - Updating the `ContextFocusActions` component to use the new `chatSession` command when the "Try again with a different model" button is clicked
- Models with the `ModelTag.StreamDisabled` tag will no longer display the initial codebase context to avoid longer processing times.
- Updated the current `OpenAI o1` and `OpenAI o1 mini`  models to use the stable version instead of `-latest` versions
- Fix the model provider logo not matching issue

## Test plan

image

Build from this branch and verify the following in your debug modes:

1. `OpenAI o1` & `OpenAI o1 mini` are added to the model dropdown list
2. These models have the `Join Waitlist` label
3. Clicking on one of the model will open a webpage in browser
4. The `Join Waitlist` label now turned to `On Waitlist`

image

When using these models:
1. When submitting a question using one of these models, you will see `Model without streaming support takes longer to response.` when waiting for the LLM response
1. You should not find `Try again with different context` underneath the LLM response when using these models.
1. You should see `Try again with a different model` instead
1. Verify that when the "Try again with a different model" button is clicked, a new chat session is created that is a copy of the current session, and the user is able to resubmit their request with a different model.

## Changelog

feat(chat): add support for preview models `Gemini 1.5 Pro Latest` & `Gemini 1.5 Flash Latest` 
feat(chat): Added ability to duplicate chat sessions


vscode/src/completions/nodeClient.ts
```diff
@@ -200,6 +200,17 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
                         bufferText += str
                         bufferBin = buf
 
+                        // HACK: Handles non-stream request.
+                        // TODO: Implement a function to make and process non-stream requests.
+                        if (params.stream === false) {
```
abeatrix, line 205: Don't have time to implement a proper function to process non-streaming requests so will require a follow-up.

vscode/webviews/components/modelSelectField/ModelSelectField.tsx
```diff
@@ -83,11 +84,12 @@ export const ModelSelectField: React.FunctionComponent<{
                 })
                 return
             }
-            getVSCodeAPI().postMessage({
-                command: 'event',
```
abeatrix, line 87: deprecated

vscode/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 ### Added
 
+- Chat: The new `Gemini 1.5 Pro Latest` and `Gemini 1.5 Flash Latest` models are now available for Cody Pro users. [pull/5508](https://github.com/sourcegraph/cody/pull/5508)
```
slimsag, line 9: is this correct

slimsag (approved):
Really impressive work @abeatrix !

vscode/CHANGELOG.md
abeatrix: Good catch! 🦅

PR #5518, opened 2024-09-09
Chat: shorten remote repository titles in mention menu

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3499/increase-width-of-remote-repo-selector-in-vs-code-sidebar

image

The repository titles in the mention menu were displaying the full repository path, which could make the menu items too long to be displayed in the dropdown menu.

This change shortens the repository titles to only show the repository name (the part after the last slash) rather than including the codebase as display name since this open more spaces to display the actual repository name, and secondly, it is not common for our customers to host the same repository in two different codehost, and then have both of them imported to their Sourcegraph instance.

## Test plan

Manually verify that repository titles in the mention menu are shortened as expected.

1. Connect to S2

2. @-mention `Remote Repositories` 

image

3. Search for `sourcegraph/sourcegraph` and verify you can see the names for most of the repositories:

image

Hovering over the repository name should also show you the full path:

image

Before: most of the repo names are cut off:

image

4. For repositories that have long names, you can hover over the repository name to see the name being trucated from the start:

image

image

## Changelog

Enterprise: Remote Repository items in the mention menu now display only the org/repo part of the title, omitting the code host name to prevent repository names from being truncated in the UI.

### DIFF
--- a/lib/prompt-editor/src/mentions/mentionMenu/MentionMenuItem.tsx
+++ b/lib/prompt-editor/src/mentions/mentionMenu/MentionMenuItem.tsx
@@ -61,6 +61,20 @@ function getDescription(item: ContextItem, query: MentionQuery): string {
     }
 }
 
+export function getMentionItemTitleAndDisplayName(item: ContextItem): {
+    title: string
+    displayName: string
+} {
+    const isSymbol = item.type === 'symbol'
+    const isRepo = item.type === 'repository'
+    const title = item.title ?? (isSymbol ? item.symbolName : displayPathBasename(item.uri))
+    // For repository items, display only the org/repo part of the title
+    // to prevent long hostnames from causing repo names to be truncated in the UI.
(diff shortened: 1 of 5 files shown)


lib/prompt-editor/src/mentions/mentionMenu/MentionMenuItem.tsx
```diff
@@ -89,7 +89,7 @@ export const MentionMenuContextItemContent: FunctionComponent<{
             <div className={styles.row}>
                 {icon && <i className={`codicon codicon-${icon}`} title={isSymbol ? item.kind : ''} />}
                 <span className={clsx(styles.title, warning && styles.titleWithWarning)} title={title}>
-                    {title}
+                    {item.type === 'repository' ? title?.split('/').slice(1).join('/') : title}
```
valerybugakov, line 92: Are there cases where we won't have two or more `/`?

valerybugakov (approved):

PR #5522, opened 2024-09-10
Pass title for a dialog

mkondratek:
~Let's include the change from the release branch (https://github.com/sourcegraph/cody/pull/5405).~

Fixed in JB: https://github.com/sourcegraph/jetbrains/pull/2282

## Test plan
- chat export works

### DIFF
--- a/vscode/webviews/tabs/TabsBar.tsx
+++ b/vscode/webviews/tabs/TabsBar.tsx
@@ -335,7 +335,7 @@ function useTabs(input: Pick<TabsBarProps, 'IDE' | 'onDownloadChatClick'>): TabC
                                 title: 'Export',
                                 Icon: DownloadIcon,
                                 command: 'cody.chat.history.export',
-                                callback: onDownloadChatClick,
+                                callback: IDE === CodyIDE.JetBrains ? undefined : onDownloadChatClick,
                             },
                             {
                                 title: 'Delete all',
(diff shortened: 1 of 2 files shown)


vscode/webviews/tabs/TabsBar.tsx
```diff
@@ -335,7 +335,7 @@ function useTabs(input: Pick<TabsBarProps, 'IDE' | 'onDownloadChatClick'>): TabC
                                 title: 'Export',
                                 Icon: DownloadIcon,
                                 command: 'cody.chat.history.export',
-                                callback: onDownloadChatClick,
+                                callback: IDE === CodyIDE.JetBrains ? undefined : onDownloadChatClick,
```
pkukielka, line 338: Could you please explain details of this change?
We should not hardcode anything for JetBrains.
How are we different from other IDEs herE?

mkondratek: image

`a.click();` does not work for some reason. I am investigating it.

pkukielka: Right. I think we need the same fix for all the non-vsc IDEs.
CC @jamesmcnamara @abeatrix

abeatrix: Hmmm it works fine in Visual Studio. I think we just need to open the link in a separate window instead of the current window

abeatrix: follow up on our slack discussion i think the current approach doesn't work in JetBrains so i approved this PR to unblock for now until we could figure out the issue on JB

PR #5535, opened 2024-09-10
Chat: update chat history export button

abeatrix:
This fixes an issue where clicking on the "Export" chat history button in Eclipse would open the chat history as json in the current webview, instead of showing up as a download dialogue.

- Open the blob link in a new window instead of current window
- Add timestamp to the downloaded file name to make it unique
- Update Content Security Policy to allow more connections for the webview

## Test plan

Run this branch in Eclipse to confirm the Chat History export button works. Here is a demo when running this branch from the latest eclipse main branch:

https://github.com/user-attachments/assets/4ef85c4d-0cb5-4130-9d57-c4e5c78183db


jamesmcnamara (approved):
Thanks!

PR #5541, opened 2024-09-10
onboarding: add endpoint selection UI for non-VSCode IDEs

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3701/switch-account-button-does-not-work-in-visual-studio

This change adds a new UI flow for non-VSCode IDEs (e.g. JetBrains IDEs) that allows users to select a Sourcegraph instance from their previously used endpoints.

The key changes are:
- Add a dropdown to select the Sourcegraph instance from the user's endpoint history
- Add a button to sign in to the selected instance
- Add a button to add a new Sourcegraph instance if the desired one is not in the history

The endpoint history is stored and retrieved using the `getEndpointHistory()` and `handleEndpointAuthRequest()` functions added in the `auth.ts` file.

This change ensures a smooth onboarding experience for users of non-VSCode IDEs by providing a familiar way to select and authenticate with their Sourcegraph instance.

The `ConfigurationSubsetForWebview` interface in `protocol.ts` has been updated to include the `endpointHistory` field.

### Updated

Auth

image

Account Tab

image

### Old

Verify the new UI flow doesn't show up in VS Code.

Start Eclipse / JetBrains / Visual Studio with this branch:
- Manually verify the new UI flow works as expected for non-VSCode IDEs
- Ensure the existing VSCode IDE onboarding flow is not affected

image
image

You can "Sign In" to the selected endpoint, or click "Add New Account" to switch to the regular Sign in page
image

image

## Changelog

- Add endpoint selection UI for non-VSCode IDEs


vscode/webviews/CodyPanel.tsx
```diff
@@ -106,7 +106,9 @@ export const CodyPanel: FunctionComponent<
                         />
                     )}
                     {view === View.Prompts && <PromptsTab setView={setView} />}
-                    {view === View.Account && <AccountTab setView={setView} />}
+                    {view === View.Account && !clientCapabilities.isVSCode && (
```
abeatrix, line 109:
```suggestion
                    {view === View.Chat && clientCapabilities.isVSCode && (
```

If you change the code to this, and then change the `view === View.Chat` to `view === View.Account` above, you can then debug it in VS Code  :D

vscode/webviews/AuthPage.tsx
```diff
@@ -237,10 +268,7 @@ const ClientSignInForm: React.FC<ClientSignInFormProps> = ({ className, authStat
                 <GlobeIcon size={16} /> Sign In with Browser
             </Button>
 
-            <FormField
-                name="accessToken"
-                serverInvalid={authStatus && !authStatus.authenticated && authStatus.showNetworkError}
-            >
+            <FormField name="accessToken" serverInvalid={serverInvalid}>
```
pkukielka, line 271: Out of curiosity how does that `serverInvalid` field work?
I searched codebase for that name and it does not seem to be used/read anywhere.

abeatrix: Good question! It's actually part of the component from the library we are using for UI:  https://www.radix-ui.com/primitives/docs/components/form#field

pkukielka (approved):
LGTM, I tested that it works fine with JetBrains now.

vscode/webviews/AuthPage.tsx
```diff
@@ -267,3 +301,57 @@ const ClientSignInForm: React.FC<ClientSignInFormProps> = ({ className, authStat
...
+                            <SelectItem key={endpoint} value={endpoint} className="tw-w-full tw-p-3">
+                                {endpoint}
+                            </SelectItem>
+                        ))}
+                    </SelectGroup>
+                </SelectContent>
+            </Select>
+            {!authStatus.authenticated && authStatus.endpoint === selectedEndpoint && (
+                <p className="tw-mt-2 tw-mb-0 tw-text-red-500">
+                    Sign in failed. Try re-authenticate again with the forms above.
```
dominiccooney, line 352: Grammar: Try *to* re-authenticate ...

Also you're introducing a new word here, "authenticate", when above it is "Sign-in sign-in sign-in." So we should say "sign in again" so people know don't have uncertainty about whether re-authenticating means sign-in.

Also... we have an endpoint selected and we know we need to sign in... can we just give them a button to fix the problem, instead of making them click buttons above? You *know* that there's a 75%+ chance they click on one of the Free/Pro sign-in buttons, because the instance sign-in button didn't work...

PR #5543, opened 2024-09-10
Chat: add support to send terminal text to Cody chat

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3703/add-terminal-output-as-chat-context

The `cody.mention.selection` command now handles `selection` argument that enables users to send selected text from terminal to Cody chat.

- Update  `cody.mention.selection` command to send terminal text to Cody chat
- Update mention selection command to handle text selection
- Update package.json to enable terminal text mention command

Note: Keybinding doesn't seem to work when terminal is focused?

## Test plan

Select some text in your terminal and right click > `Add Selection to Cody Chat`:

https://github.com/user-attachments/assets/89c7f12a-398a-46df-8daa-32d4184a3c39

## Changelog

Chat: add support to send selected text from terminal to Cody chat

### DIFF
--- a/vscode/src/chat/chat-view/ChatsController.ts
+++ b/vscode/src/chat/chat-view/ChatsController.ts
@@ -200,8 +200,8 @@ export class ChatsController implements vscode.Disposable {
             ),
 
             // Mention selection/file commands
-            vscode.commands.registerCommand('cody.mention.selection', uri =>
-                this.sendEditorContextToChat(uri)
+            vscode.commands.registerCommand('cody.mention.selection', args =>
+                this.sendEditorContextToChat(args?.selection ?? args)
             ),
             vscode.commands.registerCommand('cody.mention.file', uri =>
                 this.sendEditorContextToChat(uri)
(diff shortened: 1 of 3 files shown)


jamesmcnamara (approved):
That's cool!

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -243,18 +243,27 @@ export class ChatsController implements vscode.Disposable {
         await vscode.commands.executeCommand('cody.chat.focus')
     }
 
-    private async sendEditorContextToChat(uri?: URI): Promise<void> {
-        telemetryRecorder.recordEvent('cody.addChatContext', 'clicked', {
+    private async sendEditorContextToChat(payload?: URI | string): Promise<void> {
```
dominiccooney, line 246: This API makes me nervous because we have been so burned by slinging URIs and file paths as URIs or strings etc.

Let's make this an object with properties like `payload: { editorUri: URI } | { selection: string }`

Why do we just want the string without a filename for context? Like @foo.ts:1-10 or whatever (we'd need to support character offsets...)

vscode/src/chat/chat-view/ChatsController.ts
```diff
@@ -243,18 +243,27 @@ export class ChatsController implements vscode.Disposable {
...
         })
+
         const provider = await this.getActiveChatController()
         if (provider === this.panel) {
             await vscode.commands.executeCommand('cody.chat.focus')
         }
-        await provider.handleGetUserEditorContext(uri)
+
+        if (isText) {
+            await provider.handleTextToUserInput(payload as string)
```
dominiccooney, line 263: Using the object as suggested above you can avoid the `as` here.

dominiccooney (changes requested):
Let's design the types to be easier to use.

Does this change the way sending context from files to chat works? Will that have `args.selection` set or is that a totally different code path? (If so... why?)

I think the context could be richer that this is coming from a terminal. Like, why not treat that as a first class thing, maybe we want to pull in surrounding context later, etc.

It is worth keeping in mind that `when` doesn't filter explicit invocations of commands so you should be defensive... This may not always be called with a terminal selection.

Please also do JetBrains, CC @taylorsperry

kalanchan:
@beyang @chenkc805 could you also take a quick look if this is the right implementation for https://linear.app/sourcegraph/issue/CODY-3703/add-terminal-output-as-chat-context? 

@abeatrix would it be possible to add a top level shortcut instead of going into the Cody dropdown when you right click? I feel like that would make it easier to use

abeatrix:
> would it be possible to add a top level shortcut 

just found a hacky way to do it. will continue tmr.

PR #5544, opened 2024-09-11
added(logout): Set view to chat after logout (CODY-3606)

jamesmcnamara:
[CODY-3606](https://linear.app/sourcegraph/issue/CODY-3606/first-time-login-redirects-to-profile-page-instead-of-chat-page)

If the user signs out and then signs back in, the view is still cached as the settings page. This updates the view to the chat page before signout.

I tried to implement this in the ChatController but could not get it to work, and sank an embarrassing amount of time before finding [this sneaky little special case](https://sourcegraph.com/github.com/sourcegraph/jetbrains/-/blob/src/main/kotlin/com/sourcegraph/cody/ui/web/WebUIHost.kt?L71%3A1-74%3A1). 

Also added a proxy to log messages from the webview.

## Test plan
Manually tested. I could also add an integration test if we think it's a good idea.


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -1512,15 +1517,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
...
 
-    private async setWebviewView(view: View): Promise<void> {
-        if (view !== 'chat') {
-            // Only chat view is supported in the webview panel.
-            // When a different view is requested,
-            // Set context to notify the webview panel to close.
-            // This should close the webview panel and open the login view in the sidebar.
-            await vscode.commands.executeCommand('setContext', 'cody.activated', false)
-            return
-        }
```
jamesmcnamara, line 1523: Because of this, this function was never used for anything but setting the view to 'chat'. And because the only place we called this with another argument, we also called signout, which executes this same command, we didn't need this either.

abeatrix: Yea this is a leftover from the previous set up where we only display the chat view when users is already log in, so I agree we should remove this 👍

abeatrix (commented):
In VS Code, I tried switching to the "Command" tab, and then clicked on the "Account" tab to "Switch Account" / "Sign Out", but I will be taken back to the "Command" tab after signing back in 🤔

vscode/webviews/tabs/AccountTab.tsx
```diff
@@ -52,7 +61,12 @@ export const AccountTab: React.FC = () => {
     })
     actions.push({
         text: 'Sign Out',
-        onClick: () => getVSCodeAPI().postMessage({ command: 'auth', authKind: 'signout' }),
+        onClick: () => {
+            getVSCodeAPI().postMessage({ command: 'auth', authKind: 'signout' })
+            // Set the view to the Chat tab so that if the user signs back in, they will be
+            // automatically redirected to the Chat tab, rather than the accounts tab.
+            setView(View.Chat)
```
abeatrix, line 68:
```suggestion
```

Instead of setting this here (which only works for this action but not for the Switch Account action, or the same action in other places), i think we can just add the following

```ts
                    case 'config':
                        setConfig(message)
                        updateDisplayPathEnvInfoForWebview(message.workspaceFolderUris)
                        // Reset Chat to the default view for unauthenticated users.
                        if (view && view !== View.Chat && !message.authStatus?.authenticated) {
                            setView(View.Chat)
                        }
                        break
```

to 

https://github.com/sourcegraph/cody/blob/003dfbd91fa81f6bd44f576764de3bb88da783f8/vscode/webviews/App.tsx#L89-L92

This change should automatically reset the view back to Chat every time the user has signed in and then sign out again. Do you think this makes sense?

jamesmcnamara: @abeatrix The issue with this is that for JB, the config never updates because the signout call is hijacked. I made the above change and [updated JB](https://github.com/sourcegraph/jetbrains/compare/jsm/cody-3606-login-to-chat?expand=1) to dispatch the message to the webview if it was a signout, but it causes a different problem because now the webview lags behind the native UI so we see two login screens briefly.


https://github.com/user-attachments/assets/494a98bf-736b-4536-9a4c-c8eac0b3b6e1

One option is to keep both setView calls with a comment that the one in Sign out should be removed once JB moves to agent auth, but it's messy.

abeatrix: > @abeatrix The issue with this is that for JB, the config never updates because the signout call is hijacked. I made the above change and [updated JB](https://github.com/sourcegraph/jetbrains/compare/jsm/cody-3606-login-to-chat?expand=1) to dispatch the message to the webview if it was a signout, but it causes a different problem because now the webview lags behind the native UI so we see two login screens briefly.
> 
>  double-login-flash.mov 
> One option is to keep both setView calls with a comment that the one in Sign out should be removed once JB moves to agent auth, but it's messy.

Ahhh is JetBrains planning to switch to unified webview for authentication or will they be sticking with the native webview? For now, can we maybe add a few seconds timeout before it switches to the unified webview as a workaround 😅

jamesmcnamara: I don't know but I think it would be for the best. I don't love the sleep before redirect because it still has the possibility to go wrong and it also makes the app feel sluggish :/. I added a commit here that does both setViews with a comment explaining that the signout one is just for JB and can be removed. What do you think? It means no changes are required for JB and it feels snappier. What do you think?

abeatrix: Oh I might have misunderstood your comment. I was referring to the issue where  `the webview lags behind the native UI so we see two login screens briefly`, and your comment was addressing the issue where the webview View was not reset back to Chat after user has signed out?

PR #5550, opened 2024-09-11
make some services more reactive to config

sqs:
Partial application of https://github.com/sourcegraph/cody/pull/5221 to some services: FeatureFlagProvider, ClientStateBroadcaster, UpstreamHealthProvider, OpenTelemetryService, SentryService, etc.

The idea is to make services listen to config instead of checking it once and not updating themselves when config changes, which is bad because users need to reload the editor sometimes when they change config.

## Test plan

CI

### DIFF
--- a/lib/shared/src/experimentation/FeatureFlagProvider.ts
+++ b/lib/shared/src/experimentation/FeatureFlagProvider.ts
@@ -1,8 +1,18 @@
-import { Observable } from 'observable-fns'
-import type { Event } from 'vscode'
-import { currentResolvedConfig, resolvedConfig } from '../configuration/resolver'
-import { logDebug } from '../logger'
-import { type Unsubscribable, distinctUntilChanged, fromVSCodeEvent, pluck } from '../misc/observable'
+import { Observable, Subject, interval, map } from 'observable-fns'
+import { authStatus } from '../auth/authStatus'
+import type { AuthStatus, AuthenticatedAuthStatus } from '../auth/types'
+import { logError } from '../logger'
+import {
+    combineLatest,
+    concat,
+    debounceTime,
+    distinctUntilChanged,
+    firstValueFrom,
+    mergeMap,
+    promiseFactoryToObservable,
+    shareReplay,
+    startWith,
+} from '../misc/observable'
 import { graphqlClient } from '../sourcegraph-api/graphql'
 import { wrapInActiveSpan } from '../tracing'
 import { isError } from '../utils'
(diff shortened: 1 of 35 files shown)


lib/shared/src/experimentation/FeatureFlagProvider.ts
```diff
@@ -194,7 +216,9 @@ export class FeatureFlagProvider {
 
     private async refreshFeatureFlags(): Promise<void> {
         return wrapInActiveSpan('FeatureFlagProvider.refreshFeatureFlags', async () => {
-            const endpoint = graphqlClient.endpoint
+            const {
+                auth: { serverEndpoint: endpoint },
+            } = await currentResolvedConfig()
```
valerybugakov, line 221: Thoughts on having a similar to `authStatus` sync-store for the configuration? 

```ts
const { serverEndpoint: endpoint } = currentConfig().auth

File: vscode/src/main.ts [hunk 2]
```diff
@@ -316,7 +316,6 @@ async function initializeSingletons(
                     graphqlClient.setConfig(config)
                     void featureFlagProvider.refresh()
```
valerybugakov, line 317: Can we drop this since we're now subscribed to the config? 

```suggestion
```

valerybugakov: To ensure the correct order of operations, should we subscribe `graphqlClient` to config before `featureFlagProvider`? Otherwise, `featureFlagProvider`'s internal subscription would be triggered before the update of `graphqlClient`, and the `graphqlClient` would use the outdated server URL that is out of sync with the latest config values. 

Is this a realistic scenario? I didn't test it locally. Just thinking about edge cases based on the diff here.

vscode/src/services/UpstreamHealthProvider.ts
```diff
@@ -23,49 +26,38 @@ const INITIAL_PING_DELAY_MS = 10 * 1000 // 10 seconds
...
-        ClientConfigurationWithAccessToken,
-        'serverEndpoint' | 'customHeaders' | 'accessToken'
-    > | null = null
     private nextTimeoutId: NodeJS.Timeout | null = null
 
+    private configSubscription: Unsubscribable
+
+    constructor() {
+        // Refresh when auth (endpoint or token) changes.
+        this.configSubscription = resolvedConfig.pipe(distinctUntilChanged()).subscribe(() => {
```
valerybugakov, line 39: Would this be triggered on any config property change?

valerybugakov (approved):
Love to see it all come together!

lib/shared/src/experimentation/FeatureFlagProvider.ts
sqs: Yep, will do.

sqs: I will probably wait on this and leave currentResolvedConfig() async as I try to eliminate more calls to it and replace them with Observable. Then I'll audit the remaining call sites and see if we can/should make it sync.

vscode/src/main.ts
sqs: Even if so, SourcegraphGraphQLAPIClient should get the latest config values when it is called. I will confirm.

vscode/src/services/UpstreamHealthProvider.ts
sqs: Yes. That is inefficient, and we can restrict it with `pluck` or `map`, but that IMO is only safe enough to do when you are directly using that value and not just using it to trigger a future refresh as it is here. So this should be refactored to work that way. Otherwise you could have a problem where you think it only needs to refresh when certain config keys change, but actually the code is using other config keys and so you get staleness.

sqs:
`test-integration (Windows)` is failing, but it's also failing on main: https://github.com/sourcegraph/cody/commits/main. Will merge this and chat with people here in Montreal who might be overhauling these this week.

PR #5552, opened 2024-09-11
Prepare one box UI for Cody Web usage

vovakulikov:
This PR 
- Fixes styles problem with Cody Web (the whole time vs code dark theme was included in the main lib bundle, which shouldn't be used anywhere outside of storybook demos)
- Prepare one box UI (render code block only for search intent results and only for one box UI)

## Test plan
- Check Cody Web UI (with one box functionality)
- Check that vscode one box doesn't get any regressions

### DIFF
--- a/vscode/webviews/chat/Transcript.tsx
+++ b/vscode/webviews/chat/Transcript.tsx
@@ -289,7 +289,7 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
                 <InfoMessage>
                     {humanMessage.intent === 'search' ? (
                         <div className="tw-flex tw-justify-between tw-gap-4">
-                            <p>Intent detection selected a code search response.</p>
+                            <span>Intent detection selected a code search response.</span>
                             <div>
                                 <Button
                                     size="sm"
@@ -304,7 +304,7 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
                         </div>
                     ) : (
                         <div className="tw-flex tw-justify-between tw-gap-4">
-                            <p>Intent detection selected an LLM response.</p>
+                            <span>Intent detection selected an LLM response.</span>
                             <div>
                                 <Button
                                     size="sm"
--- a/vscode/webviews/chat/Transcript.tsx
+++ b/vscode/webviews/chat/Transcript.tsx
@@ -328,8 +328,8 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
                     contextAlternatives={humanMessage.contextAlternatives}
                     model={assistantMessage?.model}
                     isForFirstMessage={humanMessage.index === 0}
+                    showSnippets={experimentalOneBoxEnabled && humanMessage.intent === 'search'}
                     defaultOpen={experimentalOneBoxEnabled && humanMessage.intent === 'search'}
-                    showSnippets={experimentalOneBoxEnabled}
                 />
             )}
             {assistantMessage && !isContextLoading && (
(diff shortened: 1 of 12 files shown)


vscode/webviews/chat/Transcript.tsx
```diff
@@ -328,8 +328,8 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
                     contextAlternatives={humanMessage.contextAlternatives}
                     model={assistantMessage?.model}
                     isForFirstMessage={humanMessage.index === 0}
+                    showSnippets={experimentalOneBoxEnabled && humanMessage.intent === 'search'}
```
vovakulikov, line 331: Show snippets only for search intent results (for now)

vscode/webviews/components/codeSnippet/CodeSnippet.module.css
```diff
@@ -46,7 +46,7 @@
     z-index: 1;
 
     border-bottom: 1px solid var(--cody-chat-code-border-color);
-    background-color: var(--vscode-titleBar-inactiveBackground);
+    background-color: var(--cody-chat-code-header-background);
```
vovakulikov, line 49: My mistake, forgot to use a more specific design token for code blocks header

vscode/webviews/components/codeSnippet/components/CodeExcerpt.tsx
```diff
@@ -103,7 +103,7 @@ export const CodeExcerpt: FC<Props> = props => {
     }, [highlightRanges, startLine, endLine, tableContainerElement, table])
 
     return (
-        <Code className={clsx(styles.codeExcerpt, className)}>
+        <Code className={clsx(styles.codeExcerpt, className)} data-code-excerpt={true}>
```
vovakulikov, line 106: This is needed to target it from Sourcegraph App via CSS selector (to avoid some web app styling for code tags)

vscode/webviews/components/shadcn/ui/accordion.tsx
```diff
@@ -40,7 +40,7 @@ const AccordionContent = React.forwardRef<
 >(({ className, children, ...props }, ref) => (
     <AccordionPrimitive.Content
         ref={ref}
-        className="tw-overflow-hidden tw-transition-all data-[state=closed]:tw-animate-accordion-up data-[state=open]:tw-animate-accordion-down"
```
vovakulikov, line 43: Search Code blocks should have sticky headers but with overflowed container sticky position simply doesn't work, I don't know the exact reasons why we need this overflow, I checked and animation and accordion UI still works without it, will be happy to revert if someone has any concerns about it

web/demo/App.module.css
```diff
@@ -1,3 +1,5 @@
+@import url('../../vscode/.storybook/static/vscode-themes/dark-modern.css');
```
vovakulikov, line 1: Include dark theme styles only for Cody Web demo

vscode/webviews/chat/cells/contextCell/ContextCell.tsx
```diff
@@ -171,7 +171,6 @@ export const ContextCell: FunctionComponent<{
                                                 <FileContextItem
                                                     item={item}
                                                     showSnippets={showSnippets}
-                                                    defaultOpen={defaultOpen}
```
thenamankumar, line 174: I would still keep both the args. Both do not every time we show snippets, we need them to be opened by default, and we might experiment with this UX in the near future. 

But its fine, we can add it later if required. Non blocking.

PR #5553, opened 2024-09-11
Chat Telemetry: Restores `prompt` and `response` text recording

akalia25:
Previously, chat telemetry events were not capturing `promptText` and `responseText` due to the asynchronous nature of the `truncatePromptString` function, which was modified in a recent pull request(https://github.com/sourcegraph/cody/pull/5231). 

Now, we have added the proper `await` calls when using `truncatePromptString`, re-enabling the collection of this data for chat telemetry events.

Key changes:
1. Added `await` to `ChatController` where `truncatePromptString` is being used
2. Made `addBotMessage` an async func

## Test plan
Tested locally and verified `promptText` and `responseText` are being recorded again.

https://github.com/user-attachments/assets/b3952819-0945-4162-baf7-d98704aec21d

### DIFF
--- a/lib/shared/src/prompt/truncation.ts
+++ b/lib/shared/src/prompt/truncation.ts
@@ -3,16 +3,15 @@ import { getTokenCounterUtils } from '../token/counter'
 
 import type { PromptString } from './prompt-string'
 
-import { Tiktoken } from 'js-tiktoken/lite'
-import ranks from 'js-tiktoken/ranks/cl100k_base'
-
-export function truncatePromptString(text: PromptString, maxTokens: number): PromptString {
-    const encoder = new Tiktoken(ranks)
-    const encoded = encoder.encode(text.toString())
-
+export async function truncatePromptString(
+    text: PromptString,
+    maxTokens: number
+): Promise<PromptString> {
+    const tokenCounterUtils = await getTokenCounterUtils()
+    const encoded = tokenCounterUtils.encode(text.toString())
     return encoded.length <= maxTokens
         ? text
-        : text.slice(0, encoder.decode(encoded.slice(0, maxTokens))?.length).trim()
+        : text.slice(0, tokenCounterUtils.decode(encoded.slice(0, maxTokens))?.length).trim()
 }
 
 /**
--- a/lib/shared/src/prompt/truncation.ts
+++ b/lib/shared/src/prompt/truncation.ts
@@ -3,15 +3,16 @@ import { getTokenCounterUtils } from '../token/counter'
 
 import type { PromptString } from './prompt-string'
 
-export async function truncatePromptString(
-    text: PromptString,
-    maxTokens: number
-): Promise<PromptString> {
-    const tokenCounterUtils = await getTokenCounterUtils()
-    const encoded = tokenCounterUtils.encode(text.toString())
+import { Tiktoken } from 'js-tiktoken/lite'
+import ranks from 'js-tiktoken/ranks/cl100k_base'
+
+export function truncatePromptString(text: PromptString, maxTokens: number): PromptString {
+    const encoder = new Tiktoken(ranks)
+    const encoded = encoder.encode(text.toString())
+
     return encoded.length <= maxTokens
         ? text
-        : text.slice(0, tokenCounterUtils.decode(encoded.slice(0, maxTokens))?.length).trim()
+        : text.slice(0, encoder.decode(encoded.slice(0, maxTokens))?.length).trim()
 }
 
 /**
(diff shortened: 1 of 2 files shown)


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -1377,7 +1378,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
     /**
      * Finalizes adding a bot message to the chat model and triggers an update to the view.
      */
-    private addBotMessage(requestID: string, rawResponse: PromptString): void {
+    private async addBotMessage(requestID: string, rawResponse: PromptString): Promise<void> {
```
jtibshirani, line 1381: Do we need to `await` on `addBotMessage` calls now?

jtibshirani (commented):
Makes sense to me. It'd be great it this were covered by a unit test. (After a bug, I always try to think "how could we prevent this from regressing again?" 🤔 )

vscode/src/chat/chat-view/ChatController.ts
akalia25: hmm yeah good point, we would need to. Do you think it would make more sense to update the calls to `addBotMessage` or change `truncatePromptString` to be synchronous again?

jtibshirani: From my reading of the code, `truncatePromptString` is always 'awaited' immediately. So I think we can just make it synchronous again.

akalia25: Made these changes! I ended up using the synchronous`Tiktoken` class to do the encode and decode.

lib/shared/src/prompt/truncation.ts
```diff
@@ -3,15 +3,16 @@ import { getTokenCounterUtils } from '../token/counter'
...
-export async function truncatePromptString(
-    text: PromptString,
-    maxTokens: number
-): Promise<PromptString> {
-    const tokenCounterUtils = await getTokenCounterUtils()
-    const encoded = tokenCounterUtils.encode(text.toString())
+import { Tiktoken } from 'js-tiktoken/lite'
+import ranks from 'js-tiktoken/ranks/cl100k_base'
+
+export function truncatePromptString(text: PromptString, maxTokens: number): PromptString {
```
jtibshirani, line 9: Unfortunately this now undoes the original motivation for the change: https://github.com/sourcegraph/cody/pull/5231. Apologies for the bad advice before, I missed this important aspect of why the method was `async`.

vscode/src/chat/chat-view/ChatController.ts
jtibshirani: I left a comment here: https://github.com/sourcegraph/cody/pull/5553#discussion_r1757016770, we should stick with async after all! So let's go with that. I looked and I don't think we need to await `addBotMessage` everywhere. To be extra confident, it'd be good to do a quick manual test of Cody Chat locally.

akalia25: Thanks @jtibshirani, I'll revert this PR to the original logic. Agreed on the extra testing, i'll record a video showing chat is working as expected.

lib/shared/src/prompt/truncation.ts
akalia25: No Problem! :)

jtibshirani (approved):
Looks good to me! For future work, it'd be great to have a couple unit tests for telemetry. That way, we can prevent similar regressions from happening again. And... it makes things easier on the data team, since the unit test will block a PR from being merged, and that developer needs to fix the telemetry integration!

akalia25:
> Looks good to me! For future work, it'd be great to have a couple unit tests for telemetry. That way, we can prevent similar regressions from happening again. And... it makes things easier on the data team, since the unit test will block a PR from being merged, and that developer needs to fix the telemetry integration!

Thanks @jtibshirani, totally agree. We have built-out telemetry unit tests within the Cody VSCode E2E Tests ([example here](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/test/e2e/chat-input.test.ts?L15-24)) that validate the correct eventNames (`feature`/`action`) are being recorded. In this case, the issue was within the metadata of the event not being recorded (one layer deeper) which is why it got through. We plan to address this gap with @RXminuS's new E2E Testing framework )([thread](https://sourcegraph.slack.com/archives/C05EVRLQEUR/p1726152511789379)), which will allow us to test this granularity within CI.

Thanks for the feedback and review on this PR!

PR #5555, opened 2024-09-11
Update Remote Directory label

simoncwaterer:
I found the label/prompt to select a remote directory confusing when using @-mention remote directories. It was not clear to me that it was possible to continue typing to search for matching directories within the repo.

## Test plan

Open Cody
Select Remote Directories
Check for updated label once repo selected

## Changelog

Update Remote Directory label.


sqs (approved):
Thanks @simoncwaterer. LGTM as an improvement.

(@taiyab the overall flow here is still confusing; see https://sourcegraph.slack.com/archives/C04MSD3DP5L/p1725982101380229. `Select` is not the right verb and it should be Title Case for consistency.)

PR #5557, opened 2024-09-11
fix(commands): remove duplicated commands from Cody Commands menu

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3698/remove-duplicate-command-buttons-in-ui-to-avoid-confusion

This PR fixes an issue where duplicated default commands are listed as custom commands in the Cody Commands QuickPick menu.

Root cause: All commands (including default ones) were being sent to the quick pick menu during the build step used to append to the menu.

Fix: Send only custom commands to the menu, preventing default commands from being listed as custom.

- Update showCommandMenu function to use only custom commands
- Modify CommandsProvider to pass only custom commands for menu display
- Update E2E tests to verify default commands are not shown as custom
- Refactor command menu opening functions for consistency
- Update CHANGELOG.md to reflect the fix

This change ensures that default Cody commands are not incorrectly listed as custom commands in the Cody Commands menu, improving the user experience and reducing confusion.

## Test plan

Green CI with the updated E2E test.

Manual Test:

1. Click on the Cody icon in your editor to bring up the Cody Commands menu
2. Verify the default commands are only showing up once

#### After

image

#### Before:

Default commands are showing up twice, once at the top and once at the bottom of the menu:

image

### DIFF
--- a/vscode/CHANGELOG.md
+++ b/vscode/CHANGELOG.md
@@ -9,6 +9,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 ### Fixed
 
 Chat: Fixed feedback buttons not working in chat. [pull/5509](https://github.com/sourcegraph/cody/pull/5509)
+Command: Removed duplicated default commands from the Cody Commands menu that were incorrectly listed as custom commands.
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


kalanchan (approved):
thank you! this is nice clean up

PR #5563, opened 2024-09-12
Chat: rename the 'OpenAI o1' model to 'OpenAI o1-preview

abeatrix:
fix: update model selection to only allow available models
- Only update the model ID if the new model is available to the user
- Rename the 'OpenAI o1' model to 'OpenAI o1-preview' to better indicate it is a preview model

## Test plan

Switch to OpenAI o1-preview model to confirm it opens the waitlist blog
Then switch to another model to confirm you can submit chat without issue

## Changelog

Rename the 'OpenAI o1' model to 'OpenAI o1-preview'


jamesmcnamara (approved):
LGTM

PR #5565, opened 2024-09-13
fix(chat): support non-streaming chat completion requests

abeatrix:
Follow up on https://github.com/sourcegraph/cody/pull/5508
CLOSE https://linear.app/sourcegraph/issue/CODY-3741/oai-o1-hits-parsing-error

This change adds support for non-streaming requests, particularly for new OpenAI models that do not support streaming. It also fixes a "Parse Error: JS exception" issue that occurred when handling non-streaming responses in the streaming client.

- Remove HACK for handling non-streaming requests in the stream method in nodeClient.ts that's resulting in parsing error
- Implement a new `_fetchWithCallbacks` method in the `SourcegraphCompletionsClient` class to handle non-streaming requests
- Update the `SourcegraphNodeCompletionsClient` and `SourcegraphBrowserCompletionsClient` classes to use the new `_fetchWithCallbacks` method when `params.stream` is `false`
- This change ensures that non-streaming requests, such as those for the new OpenAI models that do not support streaming, are handled correctly and avoid issues like "Parse Error: JS exception" that can occur when the response is not parsed correctly in the streaming client

The `SourcegraphCompletionsClient` class now has a new abstract method `_fetchWithCallbacks` that must be implemented by subclasses

## Test plan

Verify that chat requests using non-streaming models (e.g., OpenAI O1, OpenAI O1-mini) work correctly without any parsing errors. Example: ask Cody about cody repo with "cody what are the latest updates in this repo?"

### After

First try

image

Second try

image

### Before

First try

image

## Changelog

feat(chat): support non-streaming requests

### DIFF
--- a/lib/shared/src/sourcegraph-api/completions/browserClient.ts
+++ b/lib/shared/src/sourcegraph-api/completions/browserClient.ts
@@ -115,6 +115,67 @@ export class SourcegraphBrowserCompletionsClient extends SourcegraphCompletionsC
             console.error(error)
         })
     }
+
+    protected async _fetchWithCallbacks(
+        params: CompletionParameters,
+        requestParams: CompletionRequestParameters,
+        cb: CompletionCallbacks,
+        signal?: AbortSignal
+    ): Promise<void> {
+        const { apiVersion } = requestParams
+        const serializedParams = await getSerializedParams(params)
+
+        const url = new URL(this.completionsEndpoint)
+        if (apiVersion >= 1) {
+            url.searchParams.append('api-version', '' + apiVersion)
+        }
+        addClientInfoParams(url.searchParams)
+
+        const headersInstance = new Headers({
+            ...this.config.customHeaders,
+            ...requestParams.customHeaders,
+        } as HeadersInit)
+        addCustomUserAgent(headersInstance)
+        headersInstance.set('Content-Type', 'application/json; charset=utf-8')
+        if (this.config.accessToken) {
+            headersInstance.set('Authorization', `token ${this.config.accessToken}`)
+        }
+
+        const parameters = new URLSearchParams(globalThis.location.search)
+        const trace = parameters.get('trace')
+        if (trace) {
+            headersInstance.set('X-Sourcegraph-Should-Trace', 'true')
+        }
+
+        try {
+            const response = await fetch(url.toString(), {
+                method: 'POST',
+                headers: headersInstance,
+                body: JSON.stringify(serializedParams),
+                signal,
+            })
+
+            if (!response.ok) {
+                const errorMessage = await response.text()
+                throw new Error(
+                    errorMessage.length === 0
+                        ? `Request failed with status code ${response.status}`
+                        : errorMessage
+                )
+            }
+
+            const data = await response.json()
+            if (data?.completion) {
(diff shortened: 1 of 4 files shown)


lib/shared/src/sourcegraph-api/completions/browserClient.ts
```diff
@@ -115,6 +115,67 @@ export class SourcegraphBrowserCompletionsClient extends SourcegraphCompletionsC
...
+            if (!response.ok) {
+                const errorMessage = await response.text()
+                throw new Error(
+                    errorMessage.length === 0
+                        ? `Request failed with status code ${response.status}`
+                        : errorMessage
+                )
+            }
+
+            const data = await response.json()
```
valerybugakov, line 153: What data structure do we expect here? Can we use a corresponding type to state our expectations explicitly explicitly?

vscode/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This is a log of all notable changes to Cody for VS Code. [Unreleased] changes a
 
 - The [new OpenAI models (OpenAI O1 & OpenAI O1-mini)](https://sourcegraph.com/blog/openai-o1-for-cody) are now available to selected Cody Pro users for early access. [pull/5508](https://github.com/sourcegraph/cody/pull/5508)
 - Cody Pro users can join the waitlist for the new models by clicking the "Join Waitlist" button. [pull/5508](https://github.com/sourcegraph/cody/pull/5508)
+- Chat: Support non-streaming requests.
```
valerybugakov, line 11: PR link 👀

vscode/src/completions/nodeClient.ts
```diff
@@ -286,6 +273,82 @@ export class SourcegraphNodeCompletionsClient extends SourcegraphCompletionsClie
...
+                const json = await response.json()
+                if (typeof json?.completion === 'string') {
+                    cb.onChange(json.completion)
+                    cb.onComplete()
+                    return
+                }
+
+                throw new Error('Unexpected response format')
+            } catch (error) {
+                const errorObject = error instanceof Error ? error : new Error(`${error}`)
```
valerybugakov, line 331: Do we want to do the same in the browser client?

lib/shared/src/sourcegraph-api/completions/client.ts
```diff
@@ -88,6 +90,27 @@ export abstract class SourcegraphCompletionsClient {
         }
     }
 
+    protected async prepareRequest(
```
valerybugakov, line 93: Nice!

lib/shared/src/sourcegraph-api/completions/client.ts
```diff
@@ -88,6 +90,27 @@ export abstract class SourcegraphCompletionsClient {
...
+        params: CompletionParameters,
+        requestParams: CompletionRequestParameters
+    ): Promise<{ url: URL; serializedParams: any }> {
```
valerybugakov, line 96:
```suggestion
    ): Promise<{ url: URL; serializedParams: SerializedCompletionParameters }> {
```

valerybugakov (approved):
Nice! Tested locally ✅

PR #5571, opened 2024-09-13
Fix typo around longer response time with non-streaming models

martinamps:
## Test plan

Verify when using o1 models that it shows "This model may take longer to respond."


martinamps:
ah nevermind https://github.com/sourcegraph/cody/pull/5568 is better

PR #5573, opened 2024-09-13
feat(eclipse): Add support for linux (CODY-3536)

jamesmcnamara:
Adds support for linux for Cody Eclipse. Replacement for https://github.com/sourcegraph/cody/pull/5354.

There is a bug in the logic for processing the webview index.html file that was allowing the eclipse plugin to work accidentally, because the MacOS and Windows integrated browsers were effectively  ignoring the content security policy. 

This patches that logic and allows a setting for injecting scripts and removing the CSP (required for inline-scripts) and allows skipping converting all embedded links into file:// paths (the eclipse browser cannot handle custom protocols).

Also fixes two bugs around Java code generation.

## Test plan
Unit tests added.


olafurpg (approved):
I have not tested this locallly but LGTM

PR #5574, opened 2024-09-13
fix the diff for recent edits by replacing psDedent with ps

hitesh-1997:
## Context
`psDedent` removed the spaces making diff invalid. Replaced with `ps`

## Test plan
1. Manual testing by checking the diff validity


abeatrix (approved):
not sure how to test this but the removal looks fine to me.

PR #5577, opened 2024-09-13
Chat: show context excluded reason in UI

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3725/if-a-user-specifically-mentions-a-file-but-we-dont-use-it-as-context

Right now, users might not be aware of excluded items unless they open the context list, which can cause a lot of confusion when context items were expected:

image

This PR adds a new UI component below the Context item list to display more information about the excluded context to address this issue:

image

## Test plan

Updated storybook for the new UI component:

image

### After

User can know about the skipped context without opening the context list.

image

### Before

User will not know about the skipped context unless they have the context list opened (at the bottom of the list):

image

## Changelog

Chat: display a warning in UI when at-mention items were excluded.


chenkc805:
Thanks Bee! 

Can we tweak the copy of the one where user exceeds token limits? Here's what I recommend, assuming we can hyperlink:
- 2 context items were retrieved but not used because they exceed the token limit. Learn more about token limits [here](https://sourcegraph.com/docs/cody/core-concepts/token-limits).

valerybugakov (approved):

chenkc805 (changes requested):
Changes in my above comment. Thanks @abeatrix

abeatrix:
> * were retrieved but not used because they exceed the token limit. Learn more about token limits

This looks ok?

image

abeatrix:
> * were retrieved but not used because they exceed the token limit. Learn more about token limits

@chenkc805 This looks ok?

image

chenkc805:
Look great! Approved

chenkc805 (approved):

PR #5590, opened 2024-09-16
Autocomplete: fix test labels

valerybugakov:
No functional changes. Fixing incorrect test labels found by @sqs.

## Test plan

Check that the updated test labels make sense.


sqs (approved):
TY!!

PR #5594, opened 2024-09-16
make ExternalServices and autocomplete more reactive

sqs:
This is another PR on the way to making it so a user never needs to reload their VS Code editor again to pick up new config or auth state changes.

See individual commits:

- autocomplete providers observe config
- remove ClientConfiguration types with access token and endpoint
- logGlobalStateEmissions helper
- register misc services reactively
- create external services reactively
- make local embeddings observe config and auth state
- document and rename to make fireworks custom headers clearer

## Test plan

e2e tests

### DIFF
--- a/vscode/src/completions/create-inline-completion-item-provider.ts
+++ b/vscode/src/completions/create-inline-completion-item-provider.ts
@@ -1,10 +1,9 @@
 import {
+    type AuthenticatedAuthStatus,
     NEVER,
-    type ResolvedConfiguration,
+    type PickResolvedConfiguration,
+    type UnauthenticatedAuthStatus,
     createDisposables,
-    currentAuthStatus,
-    currentAuthStatusAuthed,
-    isDotCom,
     mergeMap,
     promiseFactoryToObservable,
     vscodeResource,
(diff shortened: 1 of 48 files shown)


vscode/src/completions/providers/shared/create-provider.ts
```diff
@@ -23,84 +25,96 @@ import { getDotComExperimentModel } from './get-experiment-model'
...
+                                source: 'server-side-model-config',
+                            })
+                        )
+                    }
+
+                    // Fallback to site-config Cody LLM configuration.
+                    return authStatus.map(authStatus => {
+                        if (!authStatus.authenticated) {
+                            return null
+                        }
```
valerybugakov, line 78: What do you think about keeping the return type equal to `Provider | Error` and raising an error here? We should never reach this point in the current implementation because we exit early from `createInlineCompletionItemProvider` call if we're not authenticated.

vscode/src/completions/create-inline-completion-item-provider.ts
```diff
@@ -43,85 +41,92 @@ class NoopCompletionItemProvider implements vscode.InlineCompletionItemProvider
...
+                                        `Can't register completion provider because \`createProvider\` returned an error (${providerOrError.message}). To fix this problem, debug why createProvider returned an error. To further debug this problem, here is the configuration:\n${configString}`
+                                    )
+                                }
+
+                                vscode.window.showErrorMessage(providerOrError.message)
+                                return []
+                            }
+                            if (!providerOrError) {
+                                return []
+                            }
```
valerybugakov, line 91: We won't need this if we keep the narrow return type. It would make it easier to debug issues because if ever reach this code path it indicated a problem but doesn't log anything about it.

valerybugakov (approved):
This is awesome!

vscode/src/completions/create-inline-completion-item-provider.ts
valerybugakov: If there is a good reason for having `null` as part of the union, let's add some `logDebug` statement here.

vscode/src/completions/providers/shared/create-provider.ts
sqs: Ah, yes, I see. So `createProvider` should actually accept the `authStatus`. I will change that.

sqs: Yeah, this simplifies the code a lot. Thanks for pointing this out.

PR #5559, opened 2024-09-12
Autocomplete: add unit tests for server-side models config

valerybugakov:
- No functional changes
- Added new unit tests for the server-side model config <-> autocomplete integration.
- Added debug output to help us understand how the autocomplete provider was picked:
    - sg02: `AutocompleteProvider:initialized: using "server-side-model-config": "anthropic::model_configured_by_site_config"`
    - sg04: `AutocompleteProvider:initialized: using "server-side-models-config": "openaicompatible::starcoder-7b"`
    - DotCom: `AutocompleteProvider:initialized: using "dotcom-feature-flags": "fireworks::deepseek-coder-v2-lite-base-direct-route"`
- Part of https://linear.app/sourcegraph/issue/CODY-3409/self-hosted-models-openaicompatible-autocomplete-provider-supports

## Test plan

CI

## TODO

Create follow-up PRs with changes based on [this Slack discussion](https://sourcegraph.slack.com/archives/C075G57B5PH/p1726135223580289):
- Remove the model ID from HTTP requests for all BYOK customers.
- Remove special cases from all autocomplete provider factories based on the previous change. 
- Keep the model ID for client logging for debugging purposes. Use `Model.displayName` if it's available.
- Add more server-side model config tests to cover these changes.
- Add more server-side model config tests to cover all supported providers.

## Test plan

CI

PR #5595, opened 2024-09-16
Autocomplete: `unstable-openai` provider `legacyModel` conditions

valerybugakov:
- The diff in this PR affects only Cody debug output logs because the model ID is not used by the `unstable-openai` autocomplete provider. I suggest keeping whatever model name clients receive in the Cody output channel for debugging purposes. Open to suggestions 🙂 (cc @slimsag because of your comment [here](https://github.com/sourcegraph/cody/pull/5559#issuecomment-2349595371)).
     - Yes, there might be a deployment name that does not clearly translate to the actual model behind it, but this is still a debugging signal that the client correctly resolved the site configuration. 
     - It makes the code simpler by removing special cases. 
     -  I plan to go over other providers and remove similar conditions there.
- No functional changes.

## Test plan

CI

### DIFF
--- a/agent/recordings/cody-chat_103640681/recording.har.yaml
+++ b/agent/recordings/cody-chat_103640681/recording.har.yaml
@@ -88,7 +88,7 @@ log:
         send: 0
         ssl: -1
         wait: 0
-    - _id: 709c04331ef053b321decb3b9eb8e0f0
+    - _id: 319763e777ba1797e749b0e6468a49a8
       _order: 0
       cache: {}
       request:
@@ -103,7 +103,7 @@ log:
             value: token
               REDACTED_964f5256e709a8c5c151a63d8696d5c7ac81604d179405864d88ff48a9232364
           - name: user-agent
-            value: jetbrains / 6.0.0-SNAPSHOT
+            value: cody-cli / 6.0.0-SNAPSHOT
           - name: connection
             value: keep-alive
           - name: host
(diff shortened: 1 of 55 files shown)


valerybugakov:
Going to update failing tests tomorrow 🙂

arafatkatze (approved):
Valerie Said

>Yes, there might be a deployment name that does not clearly translate to the actual model behind it, but this is still a debugging signal that the client correctly resolved the site configuration.
 

I think that's helpful.

Take this Azure log for example 
```

ChatController: updated authStatus {"authStatus":{"id":"VXNlcjox","hasVerifiedEmail":false,"displayName":null,
"username":"arafatkatze","avatarURL":null,
"primaryEmail":"ara.khan@sourcegraph.com","organizations":{"nodes":[]},

"endpoint":"https://sourcegraph.test:3443/",
"siteVersion":"0.0.0+dev","configOverwrites":{"chatModel":"gpt-4-test","chatModelMaxTokens":7000,

"fastChatModel":"gpt-4-test","fastChatModelMaxTokens":7000,"completionModel":"gpt-35-turbo-instruct-test","completionModelMaxTokens":7000,
"provider":"azure-openai","smartContextWindow":true},"authenticated":true,
"userCanUpgrade":false,"requiresVerifiedEmail":false,"codyApiVersion":2,
"isFireworksTracingEnabled":false},"defaultModelID":"gpt-4-test","currentModelID":"gpt-4-test"}`

```

It has the deployment names and that's better than nothing.

NOTE:  the PR doesn't make any functional changes so I am approving this(I am aware that Valerie needs to fix some testing)

valerybugakov:
TY @arafatkatze. I'll fix the tests and merge this one as is. 

I have a follow-up PR that will add tests for what matters — autocomplete request parameters. Right now, we assert the model ID assigned to the provider instance, but often, it's not used at all to make requests, which means this assertion is irrelevant to the actual functionality.

valerybugakov:
We can also iterate on output logs in follow-up PRs based on the feedback and further testing.

valerybugakov:
Here's the follow up in-progress PR: https://github.com/sourcegraph/cody/pull/5604

PR #5602, opened 2024-09-17
Autocomplete: use prefetched flag in the jaccard-similarity retriever

valerybugakov:
Closes https://linear.app/sourcegraph/issue/CODY-3760/cody-prerelease-%E2%96%88-jaccardsimilarityretriever-cannot-add-dependency

## Test plan

1. Local VS Code build
2. Open multiple files in the editor
3. Generate a completion in one of them
4. Observe that `jaccardsimilarityretriever-cannot-add-dependency` is not logged in the Cody output channel

### DIFF
--- a/vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
+++ b/vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
@@ -126,11 +126,12 @@ export class JaccardSimilarityRetriever extends CachedRetriever implements Conte
 
         const curLang = currentDocument.languageId
 
-        const enableExtendedLanguagePool = Boolean(
-            await featureFlagProvider.evaluateFeatureFlag(
-                FeatureFlag.CodyAutocompleteContextExtendLanguagePool
-            )
-        )
+        let enableExtendedLanguagePool = false
+        featureFlagProvider
+            .evaluatedFeatureFlag(FeatureFlag.CodyAutocompleteContextExtendLanguagePool)
+            .subscribe(resolvedFlag => {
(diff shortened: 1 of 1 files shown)


vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
```diff
@@ -126,11 +126,12 @@ export class JaccardSimilarityRetriever extends CachedRetriever implements Conte
 
         const curLang = currentDocument.languageId
 
-        const enableExtendedLanguagePool = Boolean(
-            await featureFlagProvider.evaluateFeatureFlag(
-                FeatureFlag.CodyAutocompleteContextExtendLanguagePool
-            )
-        )
+        let enableExtendedLanguagePool = false
```
sqs, line 129: I think this will result in `enableExtendedLanguagePool` always being `false` for the operation, since the `subscribe` callback in the added portion doesn't run synchronously. So, it suppresses the log error message but doesn't have the right behavior. Unless the important thing is to run synchronously until a certain point in this func.

sqs (approved):
approving to unblock, but see comment

vscode/src/completions/context/retrievers/jaccard-similarity/jaccard-similarity-retriever.ts
valerybugakov: Got it, will follow up!

valerybugakov: https://github.com/sourcegraph/cody/pull/5603

PR #5607, opened 2024-09-17
Use `ps` for PromptString literals.

dominiccooney:
Do not use PromptString.unsafe_fromLLMResponse for anything other than LLM responses. In this case, the string is a literal and should just be constructed with `ps`.

Remove the mention of VSCode. This string shows up in all IDEs, so "VSCode settings" is confusing in JetBrains, Visual Studio, Eclipse.

## Test plan

CI


willdollman (approved):

olafurpg (approved):
Much better fix

thenamankumar (approved):

thenamankumar (approved):
I had changed the message text already. Updating it here and merging.

PR #5613, opened 2024-09-18
Revoke access token

arafatkatze:
Solves [Linear issue](https://linear.app/sourcegraph/issue/CODY-350/bug-revoke-the-access-token-when-a-user-signs-out-of-jetbrains-and)

## Test plan
Tested locally on both a local sourcegraph instance and with dotcom instance using Vscode cody.

### DIFF
--- a/vscode/src/auth/auth.ts
+++ b/vscode/src/auth/auth.ts
@@ -358,10 +358,10 @@ export async function showSignOutMenu(): Promise<void> {
  * Log user out of the selected endpoint (remove token from secret).
  */
 async function signOut(endpoint: string): Promise<void> {
-    const theToken = await secretStorage.getToken(endpoint)
-    if (theToken) {
-        const response= await graphqlClient.DeleteAccessToken(theToken)
-        console.log("the response", response)
+    const token = await secretStorage.getToken(endpoint)
+    // Delete the access token from the Sourcegraph instance
+    if (token) {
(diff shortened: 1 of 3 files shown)


vscode/src/auth/auth.ts
```diff
@@ -177,7 +178,7 @@ async function showAccessTokenInputBox(endpoint: string): Promise<string | undef
     })
 
     if (typeof result === 'string') {
-        return result.trim()
+        return 'MANUAL_' + result.trim()
```
abeatrix, line 181: Would this mean token that were input manually before this change would be treated as "automatically" imported and will be removed on sign out? 

We will remove the quick pick for sign in and move into a unified webview for auth. When we do, this solution will probably not going to work?

abeatrix (commented):
I don't think we will need to implement this for Cody Web since Cody Web doesn't need manual sign in with token (Vova please correct me if I'm wrong 😀 ). However I don't think the current implementation would work as intended because all the tokens that were imported manually before this change will now be treated as "AUTO" source and therefore be removed. This also will not treat the token input through the UI as "MANUAL".

It might make more sense to perform the deletion for all tokens regardless of their source in this case, Wdyt?

abeatrix:
not sure if possible, but maybe you can try to match the note of the token to see if it is generated by IDE redirections: 
image

PR #5620, opened 2024-09-18
linkify filenames

sqs:
Linkify filenames, so that you can click on filenames in Cody's response and open the mentioned file.

Still lots TODO, but the basic clicking works.

This was from a live code competition on stage with @beyang. :)


vscode/webviews/components/MarkdownFromCody.tsx
```diff
@@ -201,3 +181,38 @@ const LANGUAGES = [
...
+                        return ''
+                    })
+                    .join('')
+                console.log('XX', textContent)
+                if (textContent) {
+                    const linkNode = {
+                        type: 'element',
+                        tagName: 'a',
+                        properties: {
+                            href: `command:vscode.open?${encodeURIComponent(
```
vovakulikov, line 203: I know this is just a draft, but `vscode.open` should be not used here unconditionally because it won't work in Cody Web or JetBrains integration

PR #5621, opened 2024-09-18
Cody Web: Fix cody web server endpoint configuration

vovakulikov:
Fixes https://linear.app/sourcegraph/issue/CODY-3782/cody-web-chat-fails-on-s2-with-not-authenticated-error

A few facts:
- Prior to this PR  we attempted to run auth even though we didn't have all settings propagated in activate extension from Cody Agent. 
- It worked previously because we had a slightly different flow of setting observation (recent PRs about singletons DI and auth setting changed this)
- We didn't notice this problem during those PR tests because we fell back on the dotcom instance in the logic, and Cody Web was conveniently connected to the dotcom instance by default. 
- So this PR seeds all vital information for Cody Web (auth token and importantly server endpoint before we hit auth flow) 

## Test plan
- Run Cody Web demo with s2 remote instance 
- Open the browser application dev tool tab and clear your index db storage 
- Refresh the page
- You should see no cors problems over standard gql requests in the network tab
- Chat should work properly

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -141,8 +141,7 @@ export async function initializeVscodeExtension(
     extensionActivate: ExtensionActivate,
     extensionClient: ExtensionClient,
     globalState: AgentGlobalState,
-    secrets: vscode.SecretStorage,
-    extensionConfiguration?: ExtensionConfiguration
+    secrets: vscode.SecretStorage
 ): Promise<void> {
     const paths = codyPaths()
     const extensionPath = paths.config
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -49,6 +49,7 @@ import levenshtein from 'js-levenshtein'
 import * as uuid from 'uuid'
 import type { MessageConnection } from 'vscode-jsonrpc'
 import type { CommandResult } from '../../vscode/src/CommandResult'
+import { formatURL } from '../../vscode/src/auth/auth'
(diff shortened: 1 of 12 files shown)


agent/src/agent.ts
```diff
@@ -424,7 +426,10 @@ export class Agent extends MessageHandler implements ExtensionClient {
                 const secrets =
                     clientInfo.capabilities?.secrets === 'client-managed'
                         ? new AgentClientManagedSecretStorage(this, this.secretsDidChange.event)
-                        : new AgentStatelessSecretStorage()
+                        : new AgentStatelessSecretStorage({
```
vovakulikov, line 429: I think I could use client-managed API here to avoid this seeding, but I don't want to handle it on the client either, I just want to seed it once and don't think about store methods rpc API. 

@olafurpg btw do we have any consumers of `client-managed` API? Wanted to look on use-cases for this

web/demo/App.tsx
```diff
@@ -12,7 +11,7 @@ const CREATE_AGENT_WORKER = (): Worker => new AgentWorker() as Worker
 import '../../vscode/webviews/utils/highlight.css'
 import styles from './App.module.css'
 
-const DOTCOM_SERVER_ENDPOINT = 'https://sourcegraph.com'
+const DOTCOM_SERVER_ENDPOINT = 'https://sourcegraph.sourcegraph.com'
```
vovakulikov, line 14: Update Cody Web demo to s2 in order to avoid super happy path testing against dotcom

PR #5623, opened 2024-09-18
Generate and source map for agent cli

beyang:
Generate source map so stack traces are more informative.

How do we feel about doing this for prod? It would make bug reports easier to read, but adds 28MB :(

|Before|After|
| --- | --- |
| | image |

## Test plan

Test locally


olafurpg (approved):
I've been using the Node --enable-source-maps flag to get stack traces. LGTM if this effectively does the same without the flag. I'm not sure what the TS config option does but just important to make sure we don't bloat our released production assets with source maps we only use for development.

PR #5628, opened 2024-09-18
Chat: prevent Claude 3.5 Sonnet model from apologizing constantly

abeatrix:
This change adds a prompt that prevents the Claude 3.5 Sonnet model from constantly apologizing. Currently we are adding this prompt to model ID that matches claude-3-5-sonnet; However, the server side config on S2 is using claude-3.5-sonnet as the model ID, causing this check to fail.

The prompt now applies to model when the model ID contains '3.5-sonnet' or '3-5-sonnet'.

## Test plan

Connect to S2 and open the Output channel to confirm the chat Model id is showing up as `sourcegraph/claude-3.5-sonnet`:

```
█ CodyLLMConfiguration: {"chatModel":"sourcegraph/claude-3.5-sonnet"}
```

Ask Cody a question and confirm the heading prompt is attached to your question when checking the transcript in the Output channel.

### DIFF
--- a/lib/shared/src/prompt/prompt-mixin.ts
+++ b/lib/shared/src/prompt/prompt-mixin.ts
@@ -29,7 +29,8 @@ export class PromptMixin {
             ps`\n\n`
         )
 
-        if (modelID.includes('claude-3-5-sonnet')) {
+        // Add prompt that prevents Claude 3.5 Sonnet from apologizing constantly.
+        if (modelID.includes('3-5-sonnet') || modelID.includes('3.5-sonnet')) {
             mixins = mixins.concat(HEDGES_PREVENTION)
         }
 


jtibshirani (approved):
It's surprising this didn't require a test change 🤔 Is Claude 3.5 sonnet covered in our LLM chat recording tests?

I'm also curious how we could prevent this sort of thing from happening in the future. Is there a way to better standardize model IDs? (Or maybe this was just a weird one-off S2 issue?)

abeatrix:
> It's surprising this didn't require a test change 🤔 Is Claude 3.5 sonnet covered in our LLM chat recording tests?
> 
> I'm also curious how we could prevent this sort of thing from happening in the future. Is there a way to better standardize model IDs? (Or maybe this was just a weird one-off S2 issue?)

yea this is currently a s2 specific issue because of the new model config change which will eventually roll out to all instances!

ErikaRS (commented):
Thank you!

PR #5629, opened 2024-09-19
Agent: update access token for integration tests

kalanchan:
Update agent recordings and tokens.

Also had to move the multiline autocomplete test to the block before single line to make it work

## Test plan
N/A updating recordings


sqs (approved):

agent/src/autocomplete.test.ts
```diff
@@ -23,87 +23,86 @@ describe('Autocomplete', () => {
         await client.afterAll()
     })
 
-    // TODO: use `vi.useFakeTimers()` instead of `sleep()` once it's supported by the agent tests.
-    it('autocomplete/execute (non-empty result)', async () => {
-        const uri = workspace.file('src', 'sum.ts')
+     // TODO: use `vi.useFakeTimers()` instead of `sleep()` once it's supported by the agent tests.
```
olafurpg, line 26: Let's disable all the tests that sleep. Agent tests are not supposed to sleep, period, no exceptions.

I am surprised we even have this in the first place. We are never going to be able to use `vi.useFakeTimers` for the agent because the agent server is running in a separate Node process. The real fix is to change the telemetry logic to be more test friendly.

valerybugakov: Hey @olafurpg, do you have any thoughts on how to make the telemetry logic more test-friendly? This test asserts the correct behavior of our agent-specific logic, which is critical for calculating CAR. I also do not like having `sleep` calls in tests, but because the functionality depends on the milliseconds elapsed after a particular function call and the fact that we don't have access to fake timers here, the test was implemented this way to cover the significant source of telemetry bugs for us.

> We are never going to be able to use vi.useFakeTimers for the agent because the agent server is running in a separate Node process.

Technically, it's possible to propagate fake timers from one process to another with custom additional logic. I had this implementation in mind while adding the comment.

PR #5638, opened 2024-09-19
fix auth issues

sqs:
## Test plan

CI and e2e, try server-side model config on s2, ensure no "autocomplete provider" alert shows, ensure SiteProductVersion GraphQL request spam is not happening


sqs:
Server-side model config appears to be working, but I am seeing this still:
image

Still, this PR has other good fixes and is good to merge. It doesn't fix that alert but it doesn't regress it either.

dominiccooney (approved):
As I mentioned elsewhere, using lodash/isEqual instead of JSON serialization looks good.

Retreating from observing for these flags is interesting, evidently we need some container where inside the container you have stable, non-async values but it would be nice to force the containers to respond to changes in the observables at somepoint, otherwise we are back to where we started.

But that needs more thought, so LGTM.

sqs:
@dominiccooney Thanks! BTW it's not retreating from observing, it's retreating from calling them anew each time an operation occurs.

PR #5640, opened 2024-09-19
When there are 0 items in the context, instead of saying "0 items", it should say "fetching..."

ykdojo:
## Test plan

Test manually.

image


ichim-david:
@ykdojo nice, I noticed this bug myself where I had the repo as context and it said 0 items while thinking and I clicked to show what was added as context since I knew for a fact I passed the repo context. Upon expansion, several files were mentioned and the number changed eventually. As such it's better to have a loading info than 0 as you propose.

arafatkatze (approved):

PR #4073, opened 2024-05-07
Agent - adding support for proxies(Repeat PR)

arafatkatze:
### Overall issue description

- Original Issue →  https://github.com/sourcegraph/jetbrains/issues/1127
- [Google sheet](https://docs.google.com/spreadsheets/d/1He8ECpzJftbBVMqRbQ_jzEOMeT9XcfvXtjycFRtxbVg/edit#gid=0) for all the related edge cases that is the target of the proxy issues

### Previous attempts relating to this and what went wrong

- Original proxy support PR → Add support for Codegen provider → https://github.com/sourcegraph/cody/pull/836

This was the first PR to add any kind of proxy support and it was cool to add socks proxy. This was an effort last year by the great Cezary. NOTE: That pr just focused on adding socks proxy for Agent for Unstable-codegen provider

- Eventually we removed Codegen provider but didn’t totally get rid of that code → https://github.com/sourcegraph/cody/pull/1364/

I am actually morphing some of the code of that PR so that we can “pull” the vscode proxy settings to be able to use it while spinning up the network connection. 

- Initial Http/https support https://github.com/sourcegraph/cody/pull/3695

Very useful PR from Piotr but unfortunately the [[keep alive stuff was interfering](https://github.com/sourcegraph/cody/pull/3695#discussion_r1553064771)](https://github.com/sourcegraph/cody/pull/3695#discussion_r1553064771)  Keep alive stuff makes sure that we don’t gain latency time by making new connections every time. . interestingly enough this PR can also have this issue in the sense that I am still using https-proxy-agent and http-proxy-agent as the new packages that I added with this PR but I am keeping httpAgent and httpsAgent from before for the non proxy usage so that means that atleast the keep alive would work for the non proxy usage. 

I trust Piotrs testing with squid from before but I also see that 

- Chris PR Review https://github.com/sourcegraph/cody/pull/3769

This PR was a followup to the PR above with a solution and a test for the [[keep-alive](https://github.com/sourcegraph/cody/pull/3769#discussion_r1561065702)](https://github.com/sourcegraph/cody/pull/3769#discussion_r1561065702) stuff its a great PR and we tested it with Charles for http and Socks proxy locally but NOT for HTTPS proxy so we can’t be sure if that worked perfectly and also this PR had test failures locally and with agent. 

[Stevey Agent Testing Issue](https://sourcegraph.slack.com/archives/C05AGQYD528/p1713320886954569)

[...]

PR #5650, opened 2024-09-20
Enable extension-specific proxy settings

peterguy:
Support extension-specific proxy settings (don't rely on `http(s)_proxy`).

Support using a unix domain socket as a proxy.

Add a CA cert setting to enable using proxies with self-signed or not-public certificates.

Closes CODY-3092

## Test plan

To test locally:

0. Install `socat` and `mitmproxy` (`brew install socat mitmproxy`)
1. Switch to this branch: `git switch peterguy/vscode-add-proxy-settings`
2. In one terminal, start up `socat` to listen on a UDS and send TCP connections to `mitm`
```
socat -d -d unix-listen:${HOME}/mitm-proxy.sock,fork tcp:localhost:8080
```
3. In another terminal, launch `mitmproxy`. I like to use `mitmweb` because it opens a web view of the traffic that is friendly.
```
mitmweb
```
4. Add the following to VSCode's `settings.json` (It's at `~/Library/Application Support/Code/User/settings.json` for me; you can also get to it by opening the Command Palette (Cmd-Shift-P), typing "settings" to filter, then selecting "Open User Settings (JSON)".
```
"http.proxySupport": "on",
"cody.proxy.path": "~/mitm-proxy.sock",
"cody.proxy.cacert": "~/.mitmproxy/mitmproxy-ca-cert.pem"
```
In addition to specifying proxy settings for the Cody extension, it also toggles VSCode's proxy support from `override` to `on`. If it's in the default `override` mode, it won't use the proxy agents we set up. I'd like to avoid this requirement, but until we figure out how to do that (probably with more "monkey patching", to quote @RXminuS 😄 ), we have to toggle the proxy support to either `on` or `fallback`.
5. Launch the Cody extension via the Run And Debug pane (Cmd-Shift-D), by clicking the green "play" icon at the top of the pane, when "Launch VSCode Extension (Desktop, recommended)" is selected (it's the default).
6. Note the outputs of `socat` and `mitmproxy` in the two terminals, and if you launched `mitmweb`, see the requests in the browser. If you are using `mitmweb` you can use the web interface to see the types of requests being made.
7. Execute a completion. I used Document Code as a simple triggerable one. If you launched `mitmweb` and are using a non-local LLM model (not Ollama), you should see calls to the completions endpoint in the mitm web page.
8. Optional: change the proxy settings to use `mitmproxy` directly. Remove the `proxyPath` setting and add `proxyHost` and `proxyPort`:
```
"cody.proxy.host": "localhost",
"cody.proxy.port": 8080
```
Then verify again that the extension is still operative. You should still see activity in the terminal that's running `mitmproxy` (or `mitmweb`), and the `mitmweb` web page should show more traffic, but there should be no activity in the `socat` terminal.

### DIFF
--- a/vscode/src/configuration.ts
+++ b/vscode/src/configuration.ts
@@ -1,4 +1,3 @@
-import * as fs from 'node:fs'
 import * as os from 'node:os'
 import * as path from 'node:path'
 import * as vscode from 'vscode'
@@ -20,15 +19,6 @@ import { localStorage } from './services/LocalStorageProvider'
 interface ConfigGetter {
     get<T>(section: (typeof CONFIG_KEY)[ConfigKeys], defaultValue?: T): T
 }
-
-let lastProxyConfig: {
-    host?: string
-    port?: number
-    path?: string
-    cacert?: string
-} | null = null
-let cachedProxyPath: string | undefined
-let cachedProxyCACert: string | undefined
 
 /**
  * All configuration values, with some sanitization performed.
@@ -60,98 +50,28 @@ export function getConfiguration(
         debugRegex = /.*/
     }
 
-    function resolveHomedir(filePath?: string): string | undefined {
+    const vsCodeConfig = vscode.workspace.getConfiguration()
+
+    const proxyConfig = config.get<{
+        server?: string
+        path?: string
+        cacert?: string
+    } | null>(CONFIG_KEY.proxy, null)
+
+    function resolveHomedir(filePath: string | null | undefined): string | undefined {
         for (const homeDir of ['~/', '%USERPROFILE%\\']) {
             if (filePath?.startsWith(homeDir)) {
                 return `${os.homedir()}${path.sep}${filePath.slice(homeDir.length)}`
             }
         }
-        return filePath
+        return filePath ? filePath : undefined
     }
-
-    function readProxyPath(filePath: string | undefined): string | undefined {
-        if (filePath !== lastProxyConfig?.path) {
-            const path = resolveHomedir(filePath)
-            cachedProxyPath = undefined
-            if (path) {
-                try {
-                    if (!fs.statSync(path).isSocket()) {
-                        throw new Error('Not a socket')
-                    }
-                    fs.accessSync(path, fs.constants.R_OK | fs.constants.W_OK)
-                    cachedProxyPath = path
-                } catch (error) {
-                    // `logError` caused repeated calls of this code
-                    console.error(`Cannot verify ${CONFIG_KEY.proxy}.path: ${path}: ${error}`)
-                    void vscode.window.showErrorMessage(
-                        `Cannot verify ${CONFIG_KEY.proxy}.path: ${path}:\n${error}`
-                    )
-                }
-            }
-        }
-        return cachedProxyPath
-    }
-
-    function readProxyCACert(filePath: string | undefined): string | undefined {
-        if (filePath !== lastProxyConfig?.cacert) {
-            const path = resolveHomedir(filePath)
-            cachedProxyCACert = undefined
-            if (path) {
-                // support directly embedding a CA cert in the settings
-                if (path.startsWith('-----BEGIN CERTIFICATE-----')) {
-                    cachedProxyCACert = path
-                } else {
-                    try {
-                        cachedProxyCACert = fs.readFileSync(path, { encoding: 'utf-8' })
-                    } catch (error) {
-                        // `logError` caused repeated calls of this code
-                        console.error(`Cannot read ${CONFIG_KEY.proxy}.cacert: ${path}:\n${error}`)
-                        void vscode.window.showErrorMessage(
-                            `Error reading ${CONFIG_KEY.proxy}.cacert from ${path}:\n${error}`
-                        )
-                    }
-                }
-            }
-        }
-        return cachedProxyCACert
-    }
-
-    const vsCodeConfig = vscode.workspace.getConfiguration()
-
-    const proxyConfig = config.get<{
-        host?: string
-        port?: number
-        path?: string
-        cacert?: string
-    } | null>(CONFIG_KEY.proxy, null)
-
-    let proxyHost = proxyConfig?.host
-    let proxyPort = proxyConfig?.port
-
-    if (
-        (lastProxyConfig?.host !== proxyHost || lastProxyConfig?.port !== proxyPort) &&
-        ((proxyHost && !proxyPort) || (!proxyHost && proxyPort))
-    ) {
-        // `logError` caused repeated calls of this code
-        console.error(`${CONFIG_KEY.proxy}.host and ${CONFIG_KEY.proxy}.port must be set together`)
-        void vscode.window.showErrorMessage(
-            `${CONFIG_KEY.proxy}.host and ${CONFIG_KEY.proxy}.port must be set together`
-        )
-        proxyHost = undefined
-        proxyPort = undefined
-    }
-
-    const proxyPath = readProxyPath(proxyConfig?.path)
-    const proxyCACert = readProxyCACert(proxyConfig?.cacert)
-
-    lastProxyConfig = proxyConfig
 
     return {
         proxy: vsCodeConfig.get<string>('http.proxy'),
-        proxyHost: proxyHost,
-        proxyPort: proxyPort,
-        proxyPath: proxyPath,
-        proxyCACert: proxyCACert,
+        proxyServer: proxyConfig?.server,
+        proxyPath: resolveHomedir(proxyConfig?.path),
+        proxyCACert: resolveHomedir(proxyConfig?.cacert),
         codebase: sanitizeCodebase(config.get(CONFIG_KEY.codebase)),
         serverEndpoint: config.get<string>(CONFIG_KEY.serverEndpoint),
         customHeaders: config.get<Record<string, string>>(CONFIG_KEY.customHeaders),
--- a/vscode/src/configuration.ts
+++ b/vscode/src/configuration.ts
@@ -1,4 +1,6 @@
 import * as fs from 'node:fs'
+import * as os from 'node:os'
+import * as path from 'node:path'
 import * as vscode from 'vscode'
 
 import {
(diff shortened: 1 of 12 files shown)


vscode/src/configuration.ts
```diff
@@ -48,10 +49,60 @@ export function getConfiguration(
...
 
+    function expandTilde(path?: string): string | undefined {
+        if (path?.startsWith('~/')) {
+            const homeDir = process.env.HOME || process.env.USERPROFILE
+            if (homeDir) {
+                return `${homeDir}/${path.slice(2)}`
+            }
+        }
+        return path
+    }
```
RXminuS, line 60: Can't this just be a path.resolve?

vscode/src/configuration.ts
```diff
@@ -48,10 +49,60 @@ export function getConfiguration(
...
+
+    function readProxyPath(): string | undefined {
+        const path = expandTilde(config.get<string>(CONFIG_KEY.proxyPath))
+        if (path) {
+            try {
+                if (!fs.statSync(path).isSocket()) {
+                    throw new Error('Not a socket')
+                }
+                fs.accessSync(path, fs.constants.R_OK | fs.constants.W_OK)
```
RXminuS, line 69: I'd be very careful with this and double check it platform specific docs. For instance on windows there's all sorts of EACCESS edge cases if the file or folder is busy.

From memory we are already importing extra-fs which handles this or I think the fs/promises avoids some of it.

vscode/src/configuration.ts
```diff
@@ -48,10 +49,60 @@ export function getConfiguration(
...
+                void vscode.window.showErrorMessage(
+                    `Cannot verify ${CONFIG_KEY.proxyPath}: ${path}:\n${error}`
+                )
+            }
+        }
+        return undefined
+    }
+
+    function readProxyCACert(): string | undefined {
+        const path = expandTilde(config.get<string>(CONFIG_KEY.proxyCacert))
```
RXminuS, line 81: We should be moving to the new observable config so it automatically re-calculates on change without reload. 

Could be some some initialization timing problems that might come up with reading config as fixed value.

vscode/src/configuration.ts
```diff
@@ -48,10 +49,60 @@ export function getConfiguration(
...
+        if (path) {
+            // support directly embedding a CA cert in the settings
+            if (path.startsWith('-----BEGIN CERTIFICATE-----')) {
+                return path
+            }
+            try {
+                return fs.readFileSync(path, { encoding: 'utf-8' })
+            } catch (error) {
```
RXminuS, line 89: Can you make sure to add a logerror too? Notifications don't show in logs if someone asks for support

vscode/src/configuration.ts
```diff
@@ -48,10 +49,60 @@ export function getConfiguration(
...
+            }
+        }
+        return undefined
+    }
+
     const vsCodeConfig = vscode.workspace.getConfiguration()
 
     return {
         proxy: vsCodeConfig.get<string>('http.proxy'),
+        proxyHost: config.get<string>(CONFIG_KEY.proxyHost),
```
RXminuS, line 102: Should we validate what it says in description that values need to be set together?

Alternatively, the config is full json schema compatible and it still gets flattened down. so you can instead of separate strings ask for a union of different config objects so that they are always set together.

vscode/src/fetch.node.ts
```diff
@@ -28,49 +29,78 @@ let httpsProxyAgent: HttpsProxyAgent<string>
...
-        if (!httpsProxyAgent) {
-            httpsProxyAgent = new HttpsProxyAgent(proxyURL, opts)
+        if (proxyHost || proxyPort || proxyPath) {
+            const agent = new ProxyAgent({
+                protocol: protocol || 'https:',
+                ...(proxyHost ? { host: proxyHost } : null),
+                ...(proxyPort ? { port: proxyPort } : null),
+                ...(proxyPath ? { socketPath: proxyPath } : null),
+                keepAlive: true,
+                keepAliveMsecs: 60000,
```
RXminuS, line 96: 🙈 ... oh keepAlive; the "golden bullet" that keeps claiming friendly fire. But that's for a later day 😅

Just for now, I don't think Keep Alive works this way with socks proxy agent. So there might be significant latency overhead. 

Also I think EventSource has problems with this and reconnection / server closed connections. Which given how socks proxy agent operates might give issues. Not quite sure But just so you're aware.

vscode/src/configuration.ts
peterguy: `path.resolve` doesn't understand `~/`, but I could probably use `os.homedir()` instead of relying on the environment variables, and maybe also handle `%USERPROFILE%\` as well for Windows.

vscode/src/fetch.node.ts
peterguy: ok; I was just making sure to copy the settings that are being used in the proxy agents if the global proxy url is set.

vscode/src/configuration.ts
peterguy: Good point. Switched to using `fs-extra`, and looked into issues with Windows and file locks. Looks like `stat` and `access` aren't affected by busy files.

arafatkatze:
@peterguy @RXminuS This is an interesting new addition to the proxy space but I wanted to add that its entirely possible certain edge cases be not covered by this. 

We would also test all the cases with [this google sheet](https://docs.google.com/spreadsheets/d/1He8ECpzJftbBVMqRbQ_jzEOMeT9XcfvXtjycFRtxbVg/edit#gid=0) just to extra sure that this actually works. I think using Fiddler Mitm etc is fine but we have [sourcegraph based proxies](https://devproxy.sgdev.org/) too so using those could also be helpful just to be extra certain. 

Aside from what rik already pointed out about the keep alive stuff.  We also had a long long history of changes around this as mentioned [in this PR ](https://github.com/sourcegraph/cody/pull/4073#issue-2283290996). We did use [proxy agent ](https://github.com/sourcegraph/cody/pull/3769)before but then caused [more issues](https://sourcegraph.slack.com/archives/C05B7C6FBPX/p1713281669129329) so we reverted it although perhaps it might be okay to use it just for your case coz `pnpm update-agent-recordings`  worked fine.

You can read the whole [history here ](https://github.com/sourcegraph/cody/pull/4073#issue-2283290996) and if the PR still concedes I am happy to do a deep dive by manually running all the cases etc. There's other issues also with the proxy stuff like the ability to load local windows certs and [self signed certs loading being specifically an issue](https://linear.app/sourcegraph/issue/CODY-2626/network-errors-when-using-self-signed-certificates-on-windows)(You are very likely fully aware of all of these).

I am sure you have probably covered many to almost all of these but maybe not ALL of these and that's my concern(Unless you do in-fact cover all of the cases atleast for now). 

So a question for you both(Rik and Peter): 
- If we were to use ProxyAgent library mainly for the proxy stuff then what if we face the same [issues as before](https://github.com/sourcegraph/cody/pull/3769)? 
- Would it be feasible to wait for [Rik's solution here](https://sourcegraph.slack.com/archives/C05AGQYD528/p1727369824273939?thread_ts=1727281142.234439&cid=C05AGQYD528) for the stripe thing also?

vscode/src/configuration.ts
peterguy: Good point! Added them.

vscode/src/configuration.ts
peterguy: Good idea. I re-worked the schema to enforce setting them together. Doesn't really enforce anything, but it does show the user when a value is missing. I added an error message display when the config is read to help out.

vscode/src/configuration.ts
peterguy: ... and took them away again. Adding `logError` calls caused the code to be called repeatedly. Did I do something wrong? [Posted in Slack](https://sourcegraph.slack.com/archives/C05AGQYD528/p1727908103740819).

vscode/src/fetch.node.ts
peterguy: Should we not include keepAlive in this proxy agent? They're included in the existing http(s) proxy agents, so I thought they should be included in this one.

peterguy:
@arafatkatze this affects VSCode. Do you have a setup where you can test on Windows? I'm working on getting a dev environment set up in Windows, but if you have a setup already, it'll be faster to test.

arafatkatze:
@peterguy When I tested proxy stuff on Windows I would just make the VSIX File(in my macbook environment) by 
```

cd vscode 
CODY_RELEASE_TYPE=stable pnpm release:dry-run
```

This would make a `cody.vsix `file in `cody/vscode/dist` directory and then I would take the file to windows and install cody from that vsix file in Windows to test stuff. So basically I didn't have to deal with any of the dev setup in windows.  

- I assume you do already have a windows machine so that should help out with the rest and you probably don't have to worry about the licensing stuff etc?
- Also as for the real testing I only just used `[this website](https://devproxy.sgdev.org/)` for proxies so that would cover the cases? 


IS that something that would work for you? Or are you perhaps looking at a more complex setup with some sort of packet capture etc?

vscode/src/configuration.ts
```diff
@@ -50,8 +52,26 @@ export function getConfiguration(
...
+
+    function resolveHomedir(filePath: string | null | undefined): string | undefined {
+        for (const homeDir of ['~/', '%USERPROFILE%\\']) {
+            if (filePath?.startsWith(homeDir)) {
+                return `${os.homedir()}${path.sep}${filePath.slice(homeDir.length)}`
+            }
+        }
+        return filePath ? filePath : undefined
+    }
+
```
peterguy, line 69: Left this in here because it seems to comply with "some sanitization". Can move to `validateProxySettings` if there's too much clutter in here.

vscode/src/extension.node.ts
```diff
@@ -52,9 +53,7 @@ export function activate(
             : undefined,
         startTokenReceiver: (...args) => startTokenReceiver(...args),
         otherInitialization: () => {
-            return subscriptionDisposable(
-                resolvedConfig.subscribe(config => setCustomAgent(config.configuration))
-            )
+            return subscriptionDisposable(proxySettings.subscribe(setCustomAgent))
```
peterguy, line 56: Need to use the validated proxy settings here, so need to use `proxySettings`.

vscode/src/fetch.node.ts
```diff
@@ -104,7 +135,13 @@ export function initializeNetworkAgent(context: Pick<vscode.ExtensionContext, 'e
...
     })
 
-    const customAgent = setCustomAgent(getConfiguration())
+    const customAgent = setCustomAgent(
+        validateProxySettings({
+            configuration: getConfiguration(),
+            auth: {} as AuthCredentials,
+            clientState: {} as ClientState,
+        })
+    )
```
peterguy, line 144: Had to manually validate the proxy settings, because I don't know how to get around using `getConfiguration()` here. `initializeNetworkAgent` is called when the extension is starting up, from `vscode/src/extension.node.ts => activate()`. Would like to use a subscription, but not sure how to do that, or if it can be done.

vscode/src/main.ts
```diff
@@ -157,6 +161,9 @@ export async function start(
             )
         )
     )
+
+    proxySettings.subscribe(setCustomAgent)
+
```
peterguy, line 166: Is this how we envision using the config subscription? I'm not sure that `getCustomAgent` really conforms to the subscription approach, but I'm not familiar enough with either to make major changes.

`proxySettings` is exported from the new `configuration-proxy.ts` file because it's used in multiple places.

peterguy:
Moved to #5883

PR #5652, opened 2024-09-20
single-flighting and unit test for ClientConfigSingleton

sqs:
This reduces the number of duplicate requests (in particular SiteProductVersion) sent by ClientConfigSingleton if a lot of things call it at the same time in response to a state change. It also adds a fairly comprehensive unit test.

## Test plan

CI


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1613,13 +1632,21 @@ export class ClientConfigSingleton {
...
+                map(
+                    authStatus =>
+                        ({
+                            authenticated: authStatus.authenticated,
+                            endpoint: authStatus.endpoint,
+                        }) satisfies Pick<AuthStatus, 'authenticated' | 'endpoint'>
+                ),
+                distinctUntilChanged(),
                 abortableOperation(async (authStatus, signal) => {
-                    this.isSignedIn = !!authStatus.authenticated
```
sqs, line 1617: Note: Overall I didn't want to completely rewrite this class to use observables or make any big changes, just wanted to add single-flighting to reduce SiteProductVersion calls and to add a test. (The reason I changed the cache code is that vitest fake timers only work with `setTimeout` and not `Date.now()`. The latter is a bit more work to make testable.)

beyang (approved):

olafurpg (approved):
This is a nice code cleanup. I was reviewing ClientConfigSingleton last week and itching to refactor, than you for this! Also lovely to have tests

PR #5668, opened 2024-09-23
prettier rendering of context items

beyang:
Improve the rendering of context items in Cody
* Render the basename first
* Followed by line numbers in gray text
* Repo and dirname after in gray text and smaller font

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>

<tr>

<td>

</td>

<td>

</td>

</tr>

</table>

## Test plan

Test locally and on web.


vovakulikov (approved):
Changes look good, noticed one small thing that three dot in case when context item is truncated has link color 

Screenshot 2024-09-24 at 13 09 46

This happens because overflow text is set on link block, but in this case I think truncation should be on both parts elements, maybe PrettyPrintedContextItem actually should handle it itself

PR #5671, opened 2024-09-24
Autocomplete: add `anthropic` provider tests

valerybugakov:
The problem: currently, we [assert](https://github.com/sourcegraph/cody/blob/d10b88f3bcaad1c140352088a15f641f4697c80f/vscode/src/completions/providers/shared/create-provider.test.ts#L218-L219) the model ID assigned to the provider instance, but often, it's not used at all to make requests, which means this assertion is irrelevant to the actual functionality where we bump into bugs.

- This PR adds tests for autocomplete request parameters in all supported instance configurations: dotcom/enterprise X local-settings/server-side-model-config/site-config for the `anthropic` provider.
- Moves Claude-specific code to a model helper.
- Changes the model helper interface to provider messages instead of only the prompt string.
- Enforces the consistent `requestParams.model` format — `providerId/modelName` and consistent `provider.legacyModel` format — `modelName`.
- Follow up for https://github.com/sourcegraph/cody/pull/5604.
- No functional changes.
- Part of https://linear.app/sourcegraph/issue/CODY-3778/add-autocomplete-provider-tests-to-assert-request-parameters

## Note

The only remaining provider with few tests is Fireworks. I'm working on adding request params tests for it rn. But if you open different autocomplete providers now, you'll see they are 90% identical. I hope we can merge most of them soon and rely primarily on model helpers. Then, it will be much easier to work towards the future, where models are defined in the site-config, and the client has only one autocomplete provider that can handle all the cases. Also, unifying the `requestParms.model` format also brings us closer to migrating to server-side model config usage.

## Test plan

1. CI and new unit tests
2. For manual testing, add "cody.autocomplete.advanced.provider": "anthropic" to your local settings and see if autocomplete works on DotCom.
3. For enterprise testing, use one of the instances with the configured Anthropic model (sg02 with Haiku), set the Anthropic model as the default `codeCompletion` in the site-config under `modelConfiguration.defaultModels`, and generate completions.

Beware that s2, sg02, and sg04 have server-side model config disabled [currently](https://sourcegraph.slack.com/archives/C05AGQYD528/p1727322294844429).

### DIFF
--- a/vscode/src/completions/model-helpers/__tests__/claude.test.ts
+++ b/vscode/src/completions/model-helpers/__tests__/claude.test.ts
@@ -2,14 +2,13 @@ import { describe, expect, it } from 'vitest'
 
 import { isWindows } from '@sourcegraph/cody-shared'
 
+import { Claude } from '../claude'
 import { completionParams, contextSnippets } from './test-data'
 
-import { Mistral } from '../mistral'
-
-describe('Mistral', () => {
+describe('Claude ', () => {
     describe.skipIf(isWindows())('getMessages', () => {
         it('returns the prompt with the correct intro snippets', () => {
-            const model = new Mistral()
+            const model = new Claude()
             const { docContext, document, provider } = completionParams
 
             const result = model.getMessages({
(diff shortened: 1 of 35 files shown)


vscode/src/completions/model-helpers/__tests__/claude.test.ts
```diff
@@ -0,0 +1,201 @@
+import { describe, expect, it } from 'vitest'
+
+import { isWindows } from '@sourcegraph/cody-shared'
+
+import { Claude } from '../claude'
+import { completionParams, contextSnippets } from './test-data'
+
+describe('Claude ', () => {
+    describe.skipIf(isWindows())('getMessages', () => {
```
abeatrix, line 9: Would be nice to explain why we are skipping the tests on windows. Are we expecting this to not work for users on windows?

vscode/src/completions/providers/anthropic.test.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    const claudeInstantAssertion = {
+        providerId: 'anthropic',
+        legacyModel: 'claude-instant-1.2',
+        requestParams: {
+            maxTokensToSample: 256,
+            temperature: 0.5,
+            timeoutMs: 7000,
+            topK: 0,
+            model: 'anthropic/claude-instant-1.2',
```
abeatrix, line 162: Since we are moving to the server side model config which uses the new format, should the tests be using the new format instead of the current one that we are trying to move away from?

abeatrix (approved):
Nice to have more text coverage! Left some comments in line regarding  the skipped tests on Windows, and the tests for using the new server side model id.

vscode/src/completions/providers/anthropic.test.ts
valerybugakov: We test all possible configurations. The server-side model config case is covered here: https://github.com/sourcegraph/cody/blob/8d6990fca423ed290e864d115773dcbc6b142bb3/vscode/src/completions/providers/anthropic.test.ts#L90-L101

vscode/src/completions/model-helpers/__tests__/claude.test.ts
valerybugakov: Will add a comment about it in a follow-up 👍 . I decided not to test on Windows because file paths would differ, it wouldn't work with inline snapshots, and it's not part of the prompt to focus on here.

PR #5672, opened 2024-09-24
make unified prompts feature flag reactive

sqs:
Fixes an issue where the auth was not yet available during init, so `evaluateFeatureFlag` could not be called in Cody Web (which supplies creds after initialization). Also makes it so that an editor reload is not necessary to pick up an updated value for this feature flag. (I will remove the `evaluateFeatureFlag` method because it is hard to think of a situation where you only want to read a feature flag's value at a given instant and not watch its value.)

## Test plan

n/a

### DIFF
--- a/vscode/src/main.ts
+++ b/vscode/src/main.ts
@@ -15,6 +15,7 @@ import {
     catchError,
     combineLatest,
     contextFiltersProvider,
+    createDisposables,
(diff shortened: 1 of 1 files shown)


vovakulikov (approved):
Thanks for fixing this.

PR #5676, opened 2024-09-24
watch feature flags instead of calling evaluateFeatureFlag (which is non-reactive)

sqs:
- Updates some places in code to watch feature flags instead of evaluating them once. This makes our application immediately reactive to changes in feature flag values, which means users can get the desired functionality without needing to restart their editor.
- Renames the `evaluateFeatureFlag` method to `evaluateFeatureFlagEphemerally`, deprecates it, documents when it may be safely used, and makes it cache the result.

This also reduces the latency of autocomplete after #5221 was merged because it makes `evaluateFeatureFlag` calls not hit the network each time (and each autocomplete calls that 1-2 times).
    
Fixes https://linear.app/sourcegraph/issue/CODY-3846/remove-most-uses-of-evaluatefeatureflag-method.

## Test plan

CI. Check that the unified prompts feature flag works correctly.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1387,7 +1387,7 @@ export class Agent extends MessageHandler implements ExtensionClient {
         })
 
         this.registerAuthenticatedRequest('featureFlags/getFeatureFlag', async ({ flagName }) => {
-            return featureFlagProvider.evaluateFeatureFlag(
+            return featureFlagProvider.evaluateFeatureFlagEphemerally(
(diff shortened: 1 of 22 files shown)


abeatrix (approved):
Manually Tested:
Logged into DotCom and the models that required feature flags are showing up:
image

Switched to S2 and back to DotCom as Cody Free User, and the Model List is showing up correctly: 
image

And then switched back to S2 (because sign out doesn't work, will file an issue), and signed back in to DotCom with Cody Pro, and the feature flag are working as expected 👍

PR #5681, opened 2024-09-24
Bump Cody Web version 0.9.0

vovakulikov:
This is just a Cody Web bump PR in order to publish new version of cody chat with fixes from https://github.com/sourcegraph/cody/pull/5221

## Test plan
- CI is green

### DIFF
--- a/web/CHANGELOG.md
+++ b/web/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.9.0
+- Big refactoring change around authorization (see [#5221](https://github.com/sourcegraph/cody/pull/5221))
+
 ## 0.8.3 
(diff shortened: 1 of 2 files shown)


camdencheek (approved):

vovakulikov:
@sqs We had a chat about useFeatureFlag in Slack, I wanted to do it here (use the useFeatureFlag hook for the unified prompts feature), but I didn't because I think having a feature flag with observables adds some flashes in the UI. Currently, we wait until the configuration is fetched, and this is how we avoid flashes in the UI with observables we fetch as we render, and there is a short time window when we already render UI but don't have feature flag values. 

I leave it for now, but I think we should have some sort of loader function for loading essential configuration on auth change and only then resolve rendering in cody chat.

PR #5683, opened 2024-09-24
PLG: use server-side models

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3800

- Add support to new fields added by https://github.com/sourcegraph/sourcegraph/pull/643
  - This ensures the server-side config for dotcom will be rendered as the client-side models currently we are currently using.
- Replace current client-side models with server-side models
- Remove DEFAULT_DOT_COM_MODELS and getDotComDefaultModels, or turn them into MOCK for testing
  - Separated into a different PR so that it's easier for review: https://github.com/sourcegraph/cody/pull/5725
- Fix an issue where we are using model ID with the current modelID without the provider, which is not recognized by the server. To fix this, i've updated all model id field to use the ModelRef when converted from the server-side mode config.

Here is the error that we get in current clients before the change:

image

## Test plan

Updated all the current tests to use the new config and added tests for supporting old configs.

Manual Testing: verify all the features are working as expected after the change.
- [ ] chat
- [ ] edit
- [ ] completion

Chat: 

confirm the models are showing up correctly in the dropdown:

image

confirm the request was submitted with the new model ref and that it works

image

Edit: Confirm the models are showing up in the model list and inline edit still work as expected1

image

image

Completions

image

## Changelog

Use server side model config


lib/shared/src/models/modelsService.ts
```diff
@@ -37,16 +37,17 @@ export interface ModelRef {
...
     | ModelTag.Experimental
     | ModelTag.EarlyAccess
     | ModelTag.OnWaitlist
     | ModelTag.Waitlist
     | 'stable'
     | ModelTag.Deprecated
+    | ModelTag.Internal
 export type ModelTier = ModelTag.Free | ModelTag.Pro | ModelTag.Enterprise
-export type ModelCapability = 'chat' | 'autocomplete'
+export type ModelCapability = 'chat' | 'autocomplete' | 'edit' | 'streamDisabled' | 'vision'
```
olafurpg, line 50: woo hoo vision 🐙 :excited:

abeatrix:
@jamesmcnamara the tests are failing after the last commit due to this error `Error: No default edit model found. Please set one.`
looks like it's having issue finding the default model for edit?

jamesmcnamara (approved):

abeatrix:
@valerybugakov would like your eyes on the changes made to autocomplete 🙇‍♀️

lib/shared/src/models/model.ts
```diff
@@ -180,3 +198,47 @@ export function parseModelRef(ref: ModelRefStr): ModelRef {
...
     }
 }
+
+/**
+ * Converts a model reference or ID to its legacy model ID format.
+ * If the input is a model reference, it extracts the modelId.
+ * If the input is already a legacy ID, it returns it unchanged.
+ *
+ * @param modelRefOrID - The model reference string or legacy model ID
+ * @returns The legacy model ID
```
valerybugakov, line 208: Could you add an example here for clarity? E.g., `fireworks/starcoder-7b -> starcoder-7b`

valerybugakov (approved):
LGMT! TY for the ping

lib/shared/src/models/model.ts
abeatrix: Rip I just saw this. 
Will add it in a follow-up PR tomorrow 😃

PR #5684, opened 2024-09-24
Chat: add ability to execute terminal commands from chat

abeatrix:
## Do not merge until reviewed by the security team - done

CLOSE https://linear.app/sourcegraph/issue/CODY-3811/support-for-vscode

image

- Adds support for executing terminal commands generated by Cody directly from the chat interface on button click
- Updates the preamble to include instructions for formatting terminal commands in Markdown code blocks
- Modifies the ChatMessageContent component to detect and handle terminal commands in the generated code blocks
- ~~New "Cody Command Line" feature for executing shell commands using natural language. (removed)~~

## Test plan

Execute button:

1. Ask Cody to "give me a command to list all the files in the current directory"
2. Verify Cody has responded with a command with an `Execute` button underneath the command
3. The command is not executed until you hit the button
4. Clicking on the `Execute` button will open a dialogue asking if you want to run the command in your terminal
5. Click "Run" to run the command in your terminal 
6. Verify the command that runs in your terminal is the same one you click `Execute` from

image

## Changelog

Chat: add ability to execute terminal commands from chat interface.


vscode/src/commands/services/command-line.ts
```diff
@@ -0,0 +1,189 @@
...
+            isTransient: true,
+        })
+        this.disposables.push(terminal)
+        return terminal
+    }
+
+    private async open(): Promise<void> {
+        telemetryRecorder.recordEvent('cody.command.commit', 'executed')
+        const models = await firstResultFromOperation(modelsService.getModels(ModelUsage.Chat))
+        const model = models.find(model => model.id === 'haiku')?.id ?? models[0]?.id ?? ''
```
ykdojo, line 64: Does this mean haiku is the default model?

If so, why not use 3.5 Sonnet instead?

vscode/webviews/chat/ChatMessageContent/ChatMessageContent.tsx
```diff
@@ -114,7 +114,16 @@ export const ChatMessageContent: React.FunctionComponent<ChatMessageContentProps
                 // Extract the <code> element and attached `data-file-path` if present.
                 // This allows us to intelligently apply code to the suitable file.
                 const codeElement = preElement.querySelectorAll('code')?.[0]
-                const fileName = codeElement?.getAttribute('data-file-path') || undefined
+                let fileName = codeElement?.getAttribute('data-file-path') || undefined
```
ykdojo, line 117: It might be helpful to change the variable name here to clarify what it's for.

Maybe something like:

fileNameOrCommand

or maybe separate it into two variables like:

smartApplyTargetFileName

and

isBashCommand

(or something similar)

vscode/webviews/chat/ChatMessageContent/create-buttons.ts
```diff
@@ -268,9 +269,12 @@ function createApplyButton(
 ): HTMLElement {
     const button = document.createElement('button')
     button.className = styles.button
-
+    const isShellCommand = fileName === 'command'
```
ykdojo, line 272: Same thing as my earlier comment - will be helpful to have a more descriptive fileName

vscode/webviews/chat/ChatMessageContent/create-buttons.ts
```diff
@@ -283,7 +287,7 @@ function createApplyButton(
             break
         }
         default: {
-            button.innerHTML = 'Apply'
+            button.innerHTML = isShellCommand ? 'Execute' : 'Apply'
```
ykdojo, line 290: Nice. For a future PR, it'd be nice to change the icon, too

ykdojo (changes requested):

ykdojo:
It works well, but a few comments:

1. It opens up a "Run command in terminal" dialogue every time, which I don't think is the ideal UX.

Can we simply have it send the command to the current terminal without opening up a specific "Cody Command Line" terminal?

image

As a developer, chances are, I'd be working in a terminal already. So why not send it to that?

2. It'd be great to clean up the "Ask Cody to explain" command so it simply sends the selected part of the terminal to the current chat instead of opening up a new chat as we discussed at the Merge - perhaps as a separate PR.

With these changes combined together, I think we're going to have really smooth DX - by letting the dev seamlessly jump / pass context between Cody and the terminal.

vscode/src/commands/services/command-line.ts
abeatrix: The user input should be small for this command, so any of the faster and cheaper model should be able to handle this type of request. We can change it if we notice any quality issue though. 

Note: this is only for the Cody Command Line that's hidden behind the internal feature flag:

image

ykdojo: Sounds good. Or maybe 4o? I'll test it more and I'll let you know what seems to work the best - but for now I trust your judgement!

abeatrix:
> Can we simply have it send the command to the current terminal without opening up a specific "Cody Command Line" terminal?

This is a limitation of the VS Code API iirc, unfortunately we can only send to the terminal we created 😞 

> It opens up a "Run command in terminal" dialogue every time, which I don't think is the ideal UX.

Let's wait for the security team's review to see what options we have here!

ykdojo:
> This is a limitation of the VS Code API iirc, unfortunately we can only send to the terminal we created 😞

Looks like we can send a command to an existing terminal

https://github.com/user-attachments/assets/527ae67f-324a-4995-aa47-7f442ebe2f11

Code I used: https://github.com/sourcegraph/cody/commit/7d8239800180fc189d5a3e782f14b83fd69cc243

abeatrix:
> > This is a limitation of the VS Code API iirc, unfortunately we can only send to the terminal we created 😞
> 
> 
> 
> Looks like we can send a command to an existing terminal
> 
> 
> 
> https://github.com/user-attachments/assets/527ae67f-324a-4995-aa47-7f442ebe2f11
> 
> 
> 
> Code I used: https://github.com/sourcegraph/cody/commit/7d8239800180fc189d5a3e782f14b83fd69cc243

Ohhh nice found! Will update, thank you!

ykdojo:
Love the new icon! Looks like the "Do not ask again" button is not working, though


https://github.com/user-attachments/assets/30f99f66-7b76-4888-8d91-cbf4615c8097

ykdojo:
Also I think we should go with bash, not shell. It looks like we only get syntax highlighting with bash, at least on the web.
image

image

abeatrix:
> Love the new icon! Looks like the "Do not ask again" button is not working, though
> 
>  do.not.ask.again.not.working.mov

@ykdojo typo... was matching "Do not ask" against "Do not show" ... Thanks for catching this!

aramaraju:
> Also I think we should go with bash, not shell. It looks like we only get syntax highlighting with bash, at least on the web. image
> 
> image

Isn't bash a type of shell? It feels like "shell" is the right term here?

vscode/src/commands/services/command-line.ts
aramaraju: Haiku or 4o-mini size should work decently well here IMO. It's okay starting with it

aramaraju:
> Introduces a new "Cody Command Line" experimental command that allows users to ask Cody to generate terminal commands using natural language
This is only available when cody.internal.unstable is enable

This feels like non-ideal UX, and I vote that we hold off on this piece. With things like Warp and GH integrated into Terminal, I think we should do it properly and integrate Cody directly into Terminal instead of this roundabout approach.

vscode/src/commands/services/command-line.ts
abeatrix: resoled since the Cody Command Line is removed

vovakulikov (approved):
TECH review is ok. I would like to change how we render our actions button (now it looks like a very big workaround and hack), but it's out of the scope of this PR, I guess.

ykdojo (approved):

ykdojo:
I was trying to dismiss my previous review but approved it by accident - just FYI!

lib/shared/src/chat/preamble.ts
```diff
@@ -9,7 +9,8 @@ const DEFAULT_PREAMBLE = ps`You are Cody, an AI coding assistant from Sourcegrap
  * produce code blocks that we can associate with existing file paths.
  * We want to read these file paths to support applying code directly to files from chat.
  */
-const SMART_APPLY_PREAMBLE = ps`If your answer contains fenced code blocks in Markdown, include the relevant full file path in the code block tag using this structure: \`\`\`$LANGUAGE:$FILEPATH\`\`\`.`
+const SMART_APPLY_PREAMBLE = ps`If your answer contains fenced code blocks in Markdown, include the relevant full file path in the code block tag using this structure: \`\`\`$LANGUAGE:$FILEPATH\`\`\`
+For executable terminal commands: enclose in a Markdown code block with the language tag "bash".`
```
PriNova, line 13: If Cody provides multiple commands then there are sometimes docstrings included:
Screenshot 2024-09-27 131843

This will not affect the execution of the commands, but could lead to potential issues:
```suggestion
For executable terminal commands: enclose in a Markdown code block with the language tag "bash". Avoid including documentation to the executable commands.`
```

evict (approved):

lib/shared/src/chat/preamble.ts
abeatrix: Good point! I've updated the prompt to include this rule with minor adjustment that hopefully would work with most LLMs 🙏

PR #5688, opened 2024-09-25
fix/context: Resolve project folders to remote repositories in JetBrains

dominiccooney:
Agent does not implement the vscode.git extension. The workspace repo mapper turned off if this extension was absent, causing Enterprise users of JetBrains, etc. to not get remote context search results for their project's current repository. This caused very poor quality, generic chat responses.

This changes the extension to always create the workspace repo mapper. When the vscode.git extension is not available there's a fallback which executes git directly.

Fixes CODY-3839.

## Test plan

Tested manually:

1. Build JetBrains, see sourcegraph/jetbrains CONTRIBUTING.md with `CODY_DIR`, `-PforceAgentBuild=true -PforceCodyBuild=true`.
2. Log into s2 and open a project with a git repository indexed by the remote. Verify that the current repository context chip is a remote one, for example it shows up as `sourcegraph/cody`, and LLM queries produce remote search context.
3. Open a project in a git repository that is not indexed on the remote instance and verify it appears as a local repo, for example, `foo` without a `prefix/`. Note, local repo context appears to be broken in JetBrains (may be Windows-specific) but that is unrelated to this change.
4. Log into a dotcom account and verify that the repository context chip is a local one, for example it shows up as `cody`.

image

## Changelog

- Uses `git` to determine the remote repository from a workspace folder in JetBrains and other non-VSCode IDEs.
- Enterprise JetBrains and other users now get remote context results for their current workspace if it is indexed on the remote.

### DIFF
--- a/agent/recordings/defaultClient_631904893/recording.har.yaml
+++ b/agent/recordings/defaultClient_631904893/recording.har.yaml
@@ -4353,6 +4353,318 @@ log:
         send: 0
         ssl: -1
         wait: 0
+    - _id: 1710cf66797fb7b3457b1f7cf0c46325
+      _order: 0
+      cache: {}
+      request:
+        bodySize: 247
+        cookies: []
+        headers:
+          - _fromType: array
+            name: authorization
+            value: token
+              REDACTED_3dd704711f82a44ff6aba261b53b61a03fb8edba658774639148630d838c2d1d
+          - _fromType: array
+            name: content-type
+            value: application/json; charset=utf-8
+          - _fromType: array
+            name: user-agent
+            value: defaultClient / v1
+          - _fromType: array
+            name: accept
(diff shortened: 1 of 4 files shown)


dominiccooney:
Hmm, this passes locally so I'm unsure what the problem with the recordings is. Will try to update them on macOS tomorrow.

dominiccooney:
Currently blocked by s2 being broken, I can't update recordings on macOS either.

abeatrix (approved):
Thank you! The VS team was just talking about this yesterday so this will be super helpful! 

Will test it with VS once this is merged and report back!

dominiccooney:
I'm going to close this, @sqs has a nice simplification that deletes the workspaceRepoMapper entirely, let's land that.

PR #5689, opened 2024-09-25
add auth and account switching agent tests

sqs:
This did not find as many issues as I expected, but it did surface a few things:

- We do not return an error if `setChatModel` is called with an invalid model. We should!
- Webviews in the agent can enter a dispose infinite cycle, which is fixed now.

These tests are useful standalone, but I'll also continue to investigate the chat webview bugs after switching accounts (and see why they aren't occurring on the agent).

Fixes https://linear.app/sourcegraph/issue/CODY-3848/add-agent-tests-for-switching-accounts.

## Test plan

CI


agent/src/auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+                    !preAuthStatus ||
+                    !preAuthStatus.authenticated ||
+                    preAuthStatus.endpoint !== INITIAL_CREDENTIALS.serverEndpoint
+                ) {
+                    expect.fail('invalid state to run this test (likely due to prior test failure)')
+                }
+            }
+
+            // Start a new chat *before* our authentication becomes invalid, so we can test that it is
+            // not usable.
```
dominiccooney, line 97: Nit, but worth elaborating... not usable *after our authentication becomes invalid.*

agent/src/auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+                ...client.info.extensionConfiguration,
+                serverEndpoint: INVALID_CREDENTIALS.serverEndpoint,
+                accessToken: INVALID_CREDENTIALS.token ?? INVALID_CREDENTIALS.redactedToken,
+                customHeaders: {},
+            })
+            expect(authStatus?.authenticated).toBeFalsy()
+
+            // Test things that should fail when not authenticated.
+
+            // The chat we started before our invalid authentication should not be usable.
```
dominiccooney, line 114: This test would be more persuasive if it first did a chat that *did* work while the authentication was valid.

agent/src/auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            await expect(
+                client.request('chat/setModel', {
+                    id: preChatID,
+                    model: FIXTURE_MODELS.differentFromDotcomAndS2DefaultChatModel[1],
+                })
+            ).rejects.toThrowError(`No panel with ID ${preChatID}`)
```
dominiccooney, line 120: This is fine and it is a nice test, but it highlights the reality that VSCode and JetBrains have more parallelism and will exhibit all kinds of different orderings compared to Agent.

For real webviews, if the webview held onto MORE state (not less) and made fewer coarse requests that could be checked for consistency I wonder if we would have a more robust system.

There's something analogous with JetBrains where it started relatively lightweight reporting document opens and edits, but the state has expanded over time as we found, in practice, that Agent and JetBrains were out of sync. I expect due to the concurrency inherent in the JetBrains plugin.

agent/src/auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            const { models } = await client.request('chat/models', { modelUsage: ModelUsage.Chat })
+            expect(models).toStrictEqual<typeof models>([])
+
+            testInvalidCredentialsRan = true
+        })
+
+        let testReauthenticateRan = false
+        it('re-authenticates using valid initial credentials', async ({ task }) => {
+            if (!testInvalidCredentialsRan || (task.repeats && task.repeats > 0)) {
+                // If we're repeating, re-authenticate again using invalid credentials first to undo
```
dominiccooney, line 138: Why not just `beforeEach`, get back to a good state, even up to restarting agent, and be done with it? Setting and checking these flags is tedious (what if we want to reorder tests, insert different ones, etc.) and looks error prone.

agent/src/auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+            // Listing models should work.
+            const { models } = await client.request('chat/models', { modelUsage: ModelUsage.Chat })
+            expect(models.length).toBeGreaterThanOrEqual(2)
+            expect(models.map(model => model.id)).toContain('openai/gpt-4o') // arbitrary model that we expect to be included
+
+            testReauthenticateRan = true
+        })
+
+        it('switches to a different account', async ({ task }) => {
+            if (!testReauthenticateRan || (task.repeats && task.repeats > 0)) {
```
dominiccooney, line 190: There's a lot of stuttering of this `task.repeats && task.repeats > 0`, and in general this conditional "before" with similar auth parameters, would be nice to have something more succinct and fluent.

Yes, copying and pasting with the LLM is cheap, reading it and maintaining it may be less so.

dominiccooney (commented):
Some comments inline. Kudos for adding tests.

agent/src/auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+    type TestingCredentials,
+} from '../../vscode/src/testutils/testing-credentials'
+import { TestClient } from './TestClient'
+import { TestWorkspace } from './TestWorkspace'
+
+describe(
+    'Auth',
+    {
+        timeout: 5000,
+        // Repeat to find race conditions. Set to 0 when recording for faster execution.
```
olafurpg, line 15: TIL `repeats`, very cool

olafurpg (approved):
I did not review the diff in detail but I love seeing more tests to cover these cases. Thank you @sqs !

agent/src/auth.test.ts
sqs: It does that right above. The above `chat/setModel` call before the invalid-auth tests this.

agent/src/auth.test.ts
sqs: Haha, the LLM didn't actually write any of this code (I wrote it on an airplane). :)

agent/src/auth.test.ts
sqs: Yeah, I'll just remove this and find a better way to achieve `repeats`.

PR #5692, opened 2024-09-25
remove unused BFG retriever

sqs:
This is not used anymore, except by a very small number of clients (who can continue to use it if they choose to remain on an older version). It has never been enabled by default and was only ever available for self-serve users, not enterprise. See https://sourcegraph.slack.com/archives/C05AGQYD528/p1727274693847419 for more context.

Removing it simplifies the way that `cody-engine` is used in our code and removes ~550 LOC.

## Test plan

CI


olafurpg (approved):

dominiccooney (approved):
👍

PR #5693, opened 2024-09-25
simplify cody-engine spawning and error handling

sqs:
Previously, we had a non-standard promise usage pattern to initialize cody-engine, which meant that an error could occur during spawning that would not be reported back as a promise rejection. To reliably detect if cody-engine spawned and became available, we now run `bfg/initialize` and ensure the request is handled successfully by cody-engine. It is also possible to dispose/kill and then restart the child process again later if the configuration changes.

## Test plan

TODO


sqs:
closing since we removed bfg/cody-engine

PR #5694, opened 2024-09-25
remove "enhanced context" concept

sqs:
The concept of "enhanced context" was removed in VS Code in May 2024 and in JetBrains in early Sep 2024 when we unified the chat UI across all editors. Now, the chat message explicitly @-mentions repositories that should be searched for context. There were still many remnants of "enhanced context" in our internal and agent APIs that were no longer being used. Removing them simplifies the code.

## Test plan

CI & e2e tests


beyang (approved):

olafurpg (approved):
Nice! Great simplification

PR #5695, opened 2024-09-25
treat ChatBuilder (nee ChatModel) as having minimal state

sqs:
Previously, the ChatModel was a source of truth and a cache. It cached the model that would be used for the next assistant message, and ChatController had to update that whenever the default chat model changed. Now, ChatBuilder (renamed from ChatModel to be less confusing) only stores the chat model *if* the user explicitly selected one; otherwise, its `model` is `undefined` and it is resolved to the default chat model as late as possible, when the chat is actually submitted. The actual chat model used is then stored in the message transcript.
    
This is better because ChatController does not need to worry about stale state being present in the ChatBuilder, except if the user's explicitly chosen model is no longer available on the endpoint, which is an error case that needs to be handled interactively (and isn't an error case that is just present because of our internal logic).    

Closes https://linear.app/sourcegraph/issue/CODY-3845/fix-chatcontroller-chipmunk-state.

Fixes https://linear.app/sourcegraph/issue/CODY-3835/chat-model-used-can-be-inconsistent-with-what-is-shown.

## Test plan

1. Sign into s2 and send a chat message with Claude 3 Haiku. Confirm it is sent using that model.
2. Switch accounts to dotcom and send a message with Gemini 1.5 Flash. Confirm it is sent using that model.
3. Switch back to s2 and confirm that the chat model selector defaults to Claude 3 Haiku. Without using the chat model selector, send another chat message and ensure it is sent using Claude 3 Haiku and not Gemini 1.5 Flash.

(Note: You can choose any models here as long as they are different on s2 and dotcom so you can see that it's properly resetting to the default.)

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -37,8 +37,8 @@ import {
     setUserAgent,
 } from '@sourcegraph/cody-shared'
 
+import { ChatBuilder } from '../../vscode/src/chat/chat-view/ChatBuilder'
 import { chatHistory } from '../../vscode/src/chat/chat-view/ChatHistoryManager'
-import { ChatModel } from '../../vscode/src/chat/chat-view/ChatModel'
 import type { ExtensionMessage, WebviewMessage } from '../../vscode/src/chat/protocol'
 import { ProtocolTextDocumentWithUri } from '../../vscode/src/jsonrpc/TextDocumentWithUri'
 import type * as agent_protocol from '../../vscode/src/jsonrpc/agent-protocol'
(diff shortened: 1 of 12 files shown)


sqs:
(There are no new or changed uses of PromptString, just moved around code.)

abeatrix (approved):
Manually tested:
- [x] DotCom Cody Pro - Chat
- [x] DotCom Cody Pro - Smart Apply
- [x] DotCom Cody Free - Chat
- [x] DotCom Cody Free - Smart Apply
- [x] S2 - Chat
- [x] S2 - Smart Apply

Did run into this error when connected to S2 but looks like an issue on the server side:
image

PR #5699, opened 2024-09-25
New Prompt list UI for IDE Extensions

jasonhawkharris:
This is a WIP and is in no way ready for review. Reference: [SRCH-1103](https://linear.app/sourcegraph/issue/SRCH-1103/update-prompts-ui-welcome-screen)

Figma: https://www.figma.com/design/f078wFMKsIOaEwj7Iwj5xy/Unified-Cody?node-id=4845-26998&node-type=symbol&t=pQoTRV7aO8GPmlIB-0


vovakulikov:
From the sync Oct 1st

- We should always show the first N (where N is 4) items on the welcome screen
  - If we don't have custom commands or custom prompts we should show the first 4 default commands/prompts
  - As we get more custom prompts/commands we shift this list and show fewer default prompts/commands
- Edit Code command item should be always on top of the list 
- We should add modal/combobox UI for the case when you click "Search Prompts"
- All prompts button leads to the Prompts Tab
- Prompts Tab lists all available commands and prompts (custom and defaults) with the search box on top so you can search over them. 

cc @jasonhawkharris

jasonhawkharris:
Closing since we merged https://github.com/sourcegraph/cody/pull/5772

PR #5702, opened 2024-09-25
make clientStateBroadcaster use observables

sqs:
Previously, clientStateBroadcaster used ad-hoc debouncing and observing. Now, it uses observables, which makes it cleaner. It is also exposed through the webview API in a simple way instead of having its own `clientState` context and messages.

## Test plan

Select a file and a range within a file and ensure that the chat message field updates with that context. Add a `tapLog('initialContext')` to the `observeInitialContext` observable chain and ensure it does not emit too often.


vscode/src/chat/initialContext.ts
```diff
@@ -0,0 +1,262 @@
...
+const activeTextEditor = fromVSCodeEvent(
+    vscode.window.onDidChangeActiveTextEditor,
+    () => vscode.window.activeTextEditor
+).pipe(shareReplay())
+
+function getCurrentFileOrSelection({
+    chatBuilder,
+}: { chatBuilder: Observable<ChatBuilder> }): Observable<ContextItem[] | typeof pendingOperation> {
+    // TODO!(sqs): debounce this
+    const selectionOrFileChanged = activeTextEditor.pipe(
```
dominiccooney, line 74: In VSCode, I don't think you need `activeTextEditor` here, I believe VSCode fires `onDidChangeTextEditorSelection` when the active editor changes, too.

However I am unsure about JetBrains, VS, etc.

vscode/src/chat/initialContext.ts
```diff
@@ -0,0 +1,262 @@
...
+        switchMap(() =>
+            fromVSCodeEvent(vscode.window.onDidChangeTextEditorSelection).pipe(
```
dominiccooney, line 76: You're really relying on this active text editor, onDidChangeTextEditorSelection event ordering. How do we work out that our editors conform to this in practice--and keep it that way--and that new editors will conform to it?

vscode/src/chat/initialContext.ts
```diff
@@ -0,0 +1,262 @@
...
+                debounceTime(200),
+                startWith(undefined),
+                map(() => vscode.window.activeTextEditor?.selection)
```
dominiccooney, line 79: Let me think through this... this is not flaky because if the active text editor changes during the debounce time the line 75 switchMap will come to life and cancel this event subscription before it has a chance to deliver a selection for a different activeTextEditor? Is that right?

dominiccooney (approved):
Very nice although I'm interested in your thoughts on the inline comment about other editors conforming to the event ordering we depend on here.

vscode/src/chat/initialContext.ts
sqs: The behavior here is a bit nuanced, and I will document it.

- If the active text editor changes, I want the observable to immediately emit.
- If ONLY the active selection changes, I want to debounce emissions by 200ms.

The thinking is that users are OK if there is a slight lag to seeing the updated selection range in the initial context chip, but they expect switching to another file to immediately update it.

vscode/src/chat/initialContext.ts
sqs: Yes, that is exactly right.

vscode/src/chat/initialContext.ts
sqs: If an editor has a different ordering, such as: "first fire the selection change then the active editor change", then:

1. We would immediately start a 200ms debounce timer for the selection change
2. But it would be immediately canceled by the active editor change, which would pipe to immediately emitting the active selection (because `startWith(undefined), map(() => vscode.window.activeTextEditor?.selection)` means it emits immediately when the outer observable (`activeTextEditor`) emits.

The only negative thing is that we incur an unnecessary debounce, but it's immediately canceled and has no effects on the actual emissions from the overall observable.

dominiccooney: Yeah but won't you emit `undefined` instead of the selection?

PR #5703, opened 2024-09-25
do not update ClientConfig when the editor is in the background

sqs:
The ClientConfigSingleton was changed last week to fetch new client config every 60s. Previously it would trigger an async refetch if the data was more than 60s old and was requested due to a user action. This means that it would now fetch in the background, which reintroduces the problem https://linear.app/sourcegraph/issue/CODY-3745/codys-background-periodic-network-access-causes-2fa. This change fixes that again in this new implementation.

Fixes https://linear.app/sourcegraph/issue/CODY-3844/make-clientconfigsingleton-not-update-when-the-editor-is-in-the.

## Test plan

Open the Cody output channel and put VS Code in the background. Ensure that after 60 seconds it does not update client config. Bring it to the foreground. Ensure it does.


abeatrix (approved):
I dont think Agent has `vscode.window.state.focused` set up so this probably won't work on other clients at the moment. Would using idle time a good alternative (for future follow-up)?

sqs:
@abeatrix That’s a good idea, yep.

PR #5704, opened 2024-09-25
reset repo name->ID mapping when switching accounts

sqs:
Previously, the repo name->ID mapping was not cleared when switching accounts. This meant that if you were on one instance and switched to another, and if you had a chat that was open but had not sent any messages on it, then after switching accounts, the chat would return an error `Error retrieving context, no search context was used: Error: could not find repo with id ...`. The repo ID was the ID from the prior instance before the account switch. Now, it properly clears the mapping when you switch accounts.

This code is due for some improvements, but this fixes the immediate issue.

Fixes https://linear.app/sourcegraph/issue/CODY-3838/error-retrieving-context-could-not-find-repo-id-error-when-chatting.

## Test plan

As described above. Confirm there is no error.

### DIFF
--- a/vscode/src/context/workspace-repo-mapper.ts
+++ b/vscode/src/context/workspace-repo-mapper.ts
@@ -1,8 +1,9 @@
 import {
+    NEVER,
     abortableOperation,
     authStatus,
+    catchError,
     combineLatest,
-    debounceTime,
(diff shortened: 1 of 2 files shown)


dominiccooney (approved):
Splendid.

PR #5709, opened 2024-09-26
Fix switching accounts auth problem

vovakulikov:
Currently there is a problem that you can't switch account with new auth logic (after we make auth state reactive), my guess this is because we fire config event too late and in this method we wait for the last event which never will be emited because by the time we run send config we already missed this. 

This PR just tries to take the last auth value, I'm not sure that maybe I should use `currentAuthStatusOrNotReadyYet` instead, but I think it should be not possible to run into config event without non init auth status observable. 

## Test plan
- Try to switch accounts 
- Try to log out completely and sign in in multiple accounts

### DIFF
--- a/vscode/src/chat/chat-view/ChatController.ts
+++ b/vscode/src/chat/chat-view/ChatController.ts
@@ -38,7 +38,6 @@ import {
     currentAuthStatusAuthed,
     currentResolvedConfig,
     featureFlagProvider,
-    firstValueFrom,
     getContextForChatMessage,
     graphqlClient,
     hydrateAfterPostMessage,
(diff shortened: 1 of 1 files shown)


dominiccooney (approved):
Looks fine but @sqs knows this best.

sqs (approved):
This change is probably fine. `sendConfig` is fired far more often than is needed anyway, so any inconsistency is likely to be temporary. (ChatController should be updated to not sprinkle state updates and instead be intentional about it.)

I am not seeing the problem you report, though. Switching accounts does work for me. What is the problem that you are seeing?

Feel free to merge, approving to unblock.

vovakulikov:
> I am not seeing the problem you report, though. Switching accounts does work for me. What is the problem that you are seeing?

It looks like it's not always reproducable, I've seen this problem yesterday and now I can't reproduce this for some reason. @kalanchan also has seen this so this is not my local machine problem. 

The problem itself is when you try to switch account, UI doesn't pick up this change even though I see toast about that I have signed in on the instance.

vovakulikov:
I'm going to merge this one just to make sure that this problem won't happen again (I actually think that some PR from yesterday slightly changed ordering of auth events streams, I saw that for auth status was added skip operator) so probably that's why I don't this problem with auth anymore.

PR #5710, opened 2024-09-26
Fix auth flashes

vovakulikov:
This PR is on top of another fix for auth https://github.com/sourcegraph/cody/pull/5709

So, currently, you can see the log-out UI state when we're verifying the auth state. This PR tries to fix it on the extension side (so, in theory, we should fix this problem for all consumers). 

The fix is simple but I'm not sure we shouldn't send the configuration if we're verifying the auth state, maybe this prevent rendering logic should be on the client side and we just need to expose verify status information

## Test plan
- Try to run the VSCode extension and check you don't see log out screen if you're already signed in

### DIFF
--- a/lib/shared/src/auth/types.ts
+++ b/lib/shared/src/auth/types.ts
@@ -5,7 +5,7 @@ import type { CodyLLMSiteConfiguration } from '../sourcegraph-api/graphql/client
  * The authentication status, which includes representing the state when authentication failed or
  * has not yet been attempted.
  */
-export type AuthStatus = UnauthenticatedAuthStatus | AuthenticatedAuthStatus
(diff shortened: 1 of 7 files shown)


dominiccooney (changes requested):
src/services/AuthProvider.test.ts needs updating.

sqs (approved):
Cool, I wanted to add something like this.

1. I think it should be called `pendingValidation` to reuse the term `validation` that it uses elsewhere (eg for `validateCredentials`).
2. If `'verifying' in authStatus` should check if the value is true or false, to guard against future type changes where it is set to false on some valid-typed values.
3. The UI should probably have some kind of indication that auth is occurring and prevent the user from taking actions, because any in-progress work could be wiped out by the impending auth change. That is fine to do in a followup.

dominiccooney (approved):
Fantastic.

PR #5715, opened 2024-09-26
fix explicitly selected model on new chat is not reset on account switch

sqs:
If you are on endpoint E1 and choose a model M1 in a new chat that is NOT sent, and then switch accounts to endpoint E2, the model selector will show your default model for E2 but your unsent chat will, when sent, use the model M1, which is likely not valid on E2. Now, we reset the new chat when your endpoint changes, which eliminates this problem.

Fixes https://linear.app/sourcegraph/issue/CODY-3898/explicitly-selected-model-in-new-chat-is-not-overwritten-on-account.

## Test plan

As described above.

### DIFF
--- a/vscode/src/chat/chat-view/ChatBuilder.ts
+++ b/vscode/src/chat/chat-view/ChatBuilder.ts
@@ -94,7 +94,7 @@ export class ChatBuilder {
          * one, or else `undefined` to use the default chat model on the current endpoint at the
          * time the chat is sent.
          */
-        public selectedModel: ChatModel | undefined,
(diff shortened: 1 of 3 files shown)


abeatrix (approved):
Manually tested:
1. As Cody Pro user, submitted a message with a model (m1) that doesn't exist on S2
2. Switched to S2 to confirm m1 is not selected and my last selected model (m2) on S2 was selected instead. I was able to submit a new message without switching model
3. Switched back to DotCom Cody Pro account and confirmed model m1 is selected
4. Submitted a message to confirm m1 was used and not m2

PR #5716, opened 2024-09-26
StateDebugOverlay debug helper for seeing state values in the webview

sqs:
Adds a component that displays the current state (configuration, auth status, models, etc.) at the top of the CodyPanel to help when debugging. To enable, set the `cody.internal.debug.state` user setting to `true`.

image

## Test plan

n/a


abeatrix (approved):
Nice! It shows up instantly right after I saved my config too!

image

PR #5717, opened 2024-09-26
simplify repo name->ID mapping

sqs:
There was a lot of complexity in the following classes:

* RepoNameResolver: maps FS paths and git remote URLs to repo names
* WorkspaceRepoMapper: maps repo names to Sourcegraph instance GraphQL repo IDs
* WorkspaceReposMonitor: combines the above 2 and exposes a list of all (GraphQL repo ID, commit) values for all workspace folders

It turns out we only really needed 2 things:

* `getRepoNamesContainingUri(uri: vscode.Uri)`
* `remoteReposForAllWorkspaceFolders: Observable<RemoteRepo[]>`

There were a lot of convoluted transformations happening in the code, such as converting a document URI to a workspace root folder to a Git remote URL and back to one of those, and then to a repo name and then repo ID. This simplifies the code.

It also makes it so that all caches are properly evicted when the user switches accounts. Previously, the caches were not always being cleared, and old GraphQL repo IDs for different Sourcegraph endpoints would be stored. Now, caches are cleared correctly, and there is only one cache and it's basically a memoized `graphqlClient.getRepoName` (previously there were 4 different and redundant caches).

## Test plan

Test switching between s2 and dotcom while trying:

- Initial context: the repository initial context @-mention should show up.
- OpenCtx providers: selecting one triggers a call to get the current repo, and that should not fail
- Remote repository @-mentions on s2

Test autocomplete telemetry: check the output channel logs. Ensure that the GitHub API or GraphQL API are not being called per keystroke or file switch (once the values are cached).

Test chat telemetry. Ensure that for public repos, the context is being logged in greater detail.


vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -896,7 +893,6 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
                 detectedIntent,
                 userSpecifiedIntent,
                 properties,
-                privateContextSummary: privateContextSummary,
```
dominiccooney, line 899: Yes it is simpler, but it is also less Telemetry. Is the context team happy with that @chenkc805 & co?

vscode/src/chat/context/chatContext.ts
```diff
@@ -200,20 +204,46 @@ export async function getChatContextItemsForMention(
     }
 }
 
-export async function getActiveEditorContextForOpenCtxMentions(): Promise<{
+const activeTextEditor: Observable<vscode.TextEditor | undefined> = fromVSCodeEvent(
+    vscode.window.onDidChangeActiveTextEditor
+).pipe(
+    startWith(undefined),
```
dominiccooney, line 210: Why not start with `vscode.window.activeTextEditor`?

vscode/src/chat/context/chatContext.ts
```diff
@@ -200,20 +204,46 @@ export async function getChatContextItemsForMention(
...
+    ContextForOpenCtxMentions | typeof pendingOperation | Error
+> = activeTextEditor.pipe(
+    switchMapReplayOperation(
+        (textEditor): Observable<ContextForOpenCtxMentions | typeof pendingOperation> => {
+            const uri = textEditor?.document.uri
+            if (!uri) {
+                return Observable.of({ uri: undefined, codebase: undefined })
+            }
+
+            return repoNameResolver.getRepoNamesContainingUri(uri).pipe(
```
dominiccooney, line 228: This means if the account changes, but the active text editor doesn't change, we will update the repo ID at that point? That is quite neat.

vscode/src/context/openctx/common/get-repository-mentions.ts
```diff
@@ -43,14 +44,20 @@ export async function getRepositoryMentions(
...
+    } catch (error) {}
 
     return await Promise.all(
         fzf.find(cleanRegex(query)).map(repository =>
             createRepositoryMention(
                 {
                     ...repository.item,
-                    current: !!localRepos.find(({ repoName }) => repoName === repository.item.name),
+                    current: !!localRepos.find(
+                        ({ name: repoName }) => repoName === repository.item.name
```
dominiccooney, line 59: Nit,

```suggestion
                        ({ name }) => name === repository.item.name
```

vscode/src/repository/git-metadata-for-editor.ts
```diff
@@ -34,30 +23,31 @@ class GitMetadataForCurrentEditor {
...
-            (currentFile
-                ? (await repoNameResolver.getRepoRemoteUrlsFromWorkspaceUri(currentFile))[0]
-                : fakeGitURLFromCodebase(config.codebase))
-
-        const gitUrl = remoteGitUrl
-            ? convertGitCloneURLToCodebaseName(remoteGitUrl) || undefined
-            : undefined
         if (currentFile) {
+            const repoName =
+                config.codebase ??
```
dominiccooney, line 38: Removing `fakeGitURLFromCodebase` has the potential to break people/things who are setting `config.codebase`.

This includes JetBrains which pushes a git remote into config.codebase as files are opened. (VSCode used to support subdirectories which are repos and followed the current repository as you interacted with specific files, but I am unsure if that behavior still exists.)

Maybe you removed some downstream parsing that depended on that specific format though? Could you share some details here? What was JetBrains behavior like? Is there cleanup in JetBrains this change unlocks?

vscode/src/repository/remote-urls-from-parent-dirs.test.ts
```diff
@@ -1,11 +1,20 @@
-import { describe, expect, it, vi } from 'vitest'
+import { beforeAll, describe, expect, it, vi } from 'vitest'
```
dominiccooney, line 1: While you're here, should we strengthen this test to have a gitlab HTTPS remote, and a random HTTP remote, so we can check that we're not erroneously equating https = github.com at some point?

vscode/src/repository/remoteRepos.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    readValuesFrom,
+} from '@sourcegraph/cody-shared'
+import { Observable, Subject } from 'observable-fns'
+import { describe, expect, test, vi } from 'vitest'
+import * as vscode from 'vscode'
+import { EventEmitter } from '../testutils/mocks'
+import { remoteReposForAllWorkspaceFolders } from './remoteRepos'
+import { repoNameResolver } from './repo-name-resolver'
+
+// vi.mock('vscode')
```
dominiccooney, line 17: Nit, clean up.

vscode/src/repository/remoteRepos.ts
```diff
@@ -0,0 +1,84 @@
...
+    /** The GraphQL ID of the repository on the Sourcegraph instance. */
+    id: string
+}
+
+const MAX_REPO_COUNT = 10
+
+const workspaceFolders: Observable<readonly vscode.WorkspaceFolder[] | undefined> = fromVSCodeEvent(
+    vscode.workspace.onDidChangeWorkspaceFolders
+).pipe(
+    startWith(undefined),
```
dominiccooney, line 30: Does vscode fire this event if we're installed with a workspace already open? Why not start with `vscode.workspace.workspaceFolders`?

vscode/src/repository/remoteRepos.ts
```diff
@@ -0,0 +1,84 @@
...
+    ),
+    authStatus,
+]).pipe(
+    switchMapReplayOperation(
+        ([workspaceFolders]): Observable<RemoteRepo[] | typeof pendingOperation> => {
+            if (!workspaceFolders) {
+                return Observable.of([])
+            }
+            return combineLatest(
+                workspaceFolders.map(folder => repoNameResolver.getRepoNamesContainingUri(folder.uri))
```
dominiccooney, line 52: Just curious but given `repoNameResolver` depends on auth status, and this depends on auth status, what in the Observables model ensures that `repoNameResolver` has reacted to auth status changing by the time we get here?

dominiccooney (approved):
Looks good and demonstrates the power of observables... Some nits and questions inline.

In particular I'm concerned that JetBrains stuffs config.codebase and this touches how that is interpreted. Also curious about the guarantees of the Observables evaluation model wrt auth status changes propagating through the observables graph.

vscode/src/chat/chat-view/ChatController.ts
sqs: Will add it back

vscode/src/chat/context/chatContext.ts
sqs: It gets mapped to that value right after. It would just be more letters and a redundant property access.

vscode/src/chat/context/chatContext.ts
sqs: Yes!

vscode/src/chat/context/chatContext.ts
dominiccooney: Right after that event fires, right?

vscode/src/repository/git-metadata-for-editor.ts
sqs: Ah. We're basically doing `convertGitCloneURLToCodebaseName(fakeGitURLFromCodebase('github.com/foo/bar'))`, which `== 'github.com/foo/bar'`. We *did* then *undid* our string transformation.

But if the input string is a remote URL like `git@github.com:foo/bar`, then I can't remove these. Thanks for the catch.

vscode/src/repository/remote-urls-from-parent-dirs.test.ts
sqs: Done

vscode/src/repository/remoteRepos.ts
sqs: It's mapped to that value in the next line. vscode does *not* fire the event when a new event listener is set up. The `fromVSCodeEvent` helper takes a 2nd arg that specifies an optional initial emission, but I can't use it since the `vscode.workspace.onDidChangeWorkspaceFolders` event does not call the event listeners with the `vscode.workspace.workspaceFolders` value, but rather some object that tells you the delta, so I just ignore the value passed to the event listeners.

dominiccooney: OK, but if you're sitting in VSCode and install Cody, then the workspace and its folders aren't changing. So won't the first run experience be stuck with `undefined`?

vscode/src/repository/remoteRepos.ts
sqs: Good question. I actually believe this is an underappreciated risk of Observables that most people paper over.

I created the `switchMapReplayOperation` for this purpose.

Suppose you have 2 observable chains that both depend on `authStatus` and then perform some async operation based on it. And then you want to combine those 2 and ensure you are seeing a consistent view of the world based on the same `authStatus` value.

The key is that the async operations must use `switchMap(() => someOperation().pipe(startWith(pendingOperation)))`[1]. This ensures that when their outer observable emits (e.g., when `authStatus` emits a new auth status), they *immediately* and synchronously emit a `pendingOperation` value (which is just a sentinel `Symbol.for('@@pendingOperation')`). This means that, synchronously, the 2 observable chains emit `pendingOperation` as well, so their "value" is immediately invalidated.

When you combine the 2 observable chains, you should check if either's last emission is `pendingOperation`, and if so, just emit `pendingOperation`; you are unable to actually perform your computation (which is a good thing! avoids footguns) because you don't have the data you'd need, only the sentinel value `pendingOperation`.

But finally when both async operations that are downstream of `authStatus` finish, they will emit something other than `pendingOperation` and you can run the computation. You know that you have a consistent state of the world.

The keys are:

- Synchronous and immediate emission of `pendingOperation` (which is poison in a good way because you have no choice but to propagate it since you can't perform the computation because you don't have the value you'd need, just that dumb sentinel value)
- Making sure that `combineLatest` uses the `AsyncSerialScheduler`, which basically emits the combined values after a nextTick so that all synchronous emissions (chiefly `pendingOperation`) have been propagated

What makes me a bit uneasy is that I haven't seen anyone else come up with this pattern (as far as I can tell), but it's a pattern I found critical in both OpenCtx and now Cody when using observables.

[1] Note that this is wrapped in the helper `switchMapReplayOperation`, which also adds a `shareReplay()` to "cache" the latest emission for any subsequent subscribers so long as any other subscriber is still listening to it. The `shareReplay()` is not necessary for consistency, it's just an optimization.

vscode/src/chat/context/chatContext.ts
sqs: Yep

vscode/src/repository/remoteRepos.ts
sqs: No, `undefined` is not emitted due to `startWith(undefined)`, because in the next line `map(() => vscode.workspace.workspaceFolders)`, it emits `vscode.workspace.workspaceFolders` regardless of whatever the outer observable emitted. If the `startWith` was placed *after* the `map`, then yes, it would emit `undefined`. I can see it's confusing because `vscode.workspace.workspaceFolders` itself can return `undefined`, though.

I added chat telemetry testing to the test plan, which would cover this.

vscode/src/repository/remoteRepos.ts
dominiccooney: If I understand correctly, the flip to `pendingOperation` is effectively synchronous and the join delays until both are no longer pending.

And the two streams of pending -> valid -> pending -> valid flip-flops always align... if observable A and B have a common dependency but observable A is "slow", you will never mix a "valid at step 1" from observable A and "valid at step 2" from observable B because the switch will cancel A's step 1?

So I guess that leaves DoS or livelock type situations. But at least that shows up in the observables graph at the switchMapReplayOperation...

Thanks for taking the time to explain this, it is a bit to get your head around at first.

sqs: > if observable A and B have a common dependency but observable A is "slow", you will never mix a "valid at step 1" from observable A and "valid at step 2" from observable B because the switch will cancel A's step 1?

Yes, that's correct, so long as both emit a `pendingOperation`-like sentinel value when their dependency emits (with `startWith(pendingOperation)`).

Yes, you could have a cycle where the 2 observables in `combineLatest([input1, input2])` cause the other one to become `pendingOperation` when the other one emits.

sqs:
@sourcegraph/jetbrains-plugin FYI, I posted [https://github.com/sourcegraph/cody/pull/5717 with som](https://linear.app/sourcegraph/issue/CODY-3906/agent-allow-use-of-existing-fallback-that-looks-at-gitconfig-to-get) with some non-urgent followups to unify agent/JetBrains behavior with VS Code.

PR #5721, opened 2024-09-26
remove unused code

sqs:
`pnpm exec knip --fix` and then some manual reverts of code that is unused but that we want to keep.

## Test plan

CI


camdencheek (approved):

camdencheek:
TIL about `knip` -- seems useful!

sqs:
@camdencheek Yes, knip is amazing!!

PR #5722, opened 2024-09-26
Improve chat context eval runner

beyang:
Previously, `cody internal bench` when running the `chat-context` strategy would emit an `output.csv` file. This was fine for a standalone eval run, but there was no way to easily compare results across different eval runs.

Update the `chat-context` strategy to do the following:
* Emit an output CSV whose name incorporates the following:
  * The input file name
  * The product version of the Sourcegraph instance run against
* Emits an output YAML file with the following metadata:
  * timestamp
  * Sourcegraph instance URL
  * Sourcegraph username and id of the access token used
  * evaluated feature flags returned for that user on that instance

Outputs from multiple runs will be displayed in the next version of cody-leaderboard-private.

## Test plan

Run

```
pnpm -C agent agent:skip-root-build internal bench --evaluation-config path/to/cody-leaderboard-private/chat-context-bench-v2.json --src-endpoint https://sourcegraph.sourcegraph.com --src-access-token $ACCESS_TOKEN
```


sqs (approved):

beyang:

This was okay, because the unsafe reference is taken from the `cody internal bench` command.

PR #5726, opened 2024-09-27
Autocomplete: add `fireworks` provider tests

valerybugakov:
The problem: currently, we [assert](https://github.com/sourcegraph/cody/blob/d10b88f3bcaad1c140352088a15f641f4697c80f/vscode/src/completions/providers/shared/create-provider.test.ts#L218-L219) the model ID assigned to the provider instance, but often, it's not used at all to make requests, which means this assertion is irrelevant to the actual functionality where we bump into bugs.

- This PR adds tests for autocomplete request parameters in all supported instance configurations: dotcom/enterprise X local-settings/server-side-model-config/site-config for the `fireworks` provider.
- Follow up for https://github.com/sourcegraph/cody/pull/5604.
- No functional changes.
- Part of https://linear.app/sourcegraph/issue/CODY-3778/add-autocomplete-provider-tests-to-assert-request-parameters

## Test plan

1. CI and new unit tests
2. For manual testing, check if autocomplete still works on DotCom.
3. For enterprise testing, use one of the instances with the configured Fireworks model, set the Fireworks model as the default `codeCompletion` in the site-config under `modelConfiguration.defaultModels`, and generate completions.


abeatrix (approved):
Verified autocompletion still works with this branch! Thanks for adding all the tests!

PR #5728, opened 2024-09-27
Edit: use consistent model provider icon mapping

abeatrix:
- Refactor `getModelProviderIcon` function to use a constant mapping object `MODEL_PROVIDER_ICONS` instead of a switch statement
- This makes the icon mapping more explicit and easier to maintain
- This also fixes an issue where OpenAI models are not using the correct logo because the capitalized "AI" in the previous switch map

## Test plan

Check the Edit model dropdown list to confirm openAI models are showing the correct logo:

image

### Currently on main

OpenAI model (GPT-4o) is showing up with Cody logo instead of OpenAI logo.

image


dominiccooney (approved):
👍

PR #5730, opened 2024-09-27
add tests to detect observable subscription leaks

sqs:
Adds a leak detector to ensure that in the observable tests, subscriptions get cleaned up.

## Test plan

n/a


dominiccooney (approved):
That is persuasive.

sqs:
@dominiccooney I added a subscription leak detector. No leaks found (which kinda surprised me). It runs during the tests now, so if you have a hunch, add a test case and see what it says.

dominiccooney:
I guess we need to make a distinction between leaking these objects and them being alive. I am observing the objects (and their closures etc.) leak.

PR #5731, opened 2024-09-27
chore: Remove unused queries from 969 prototype and rename client

janhartman:
This functionality was introduced by Rafal a few months ago but we never got around to actually using any of it.

## Test plan
CI + local test if chat still works as expected.


vscode/src/chat/context/chatIntentAPIClient.ts
```diff
@@ -0,0 +1,33 @@
...
+        return this.apiClient.chatIntent(interactionID, query)
+    }
+
+    private isIntentDetectionAPIEnabled(): boolean {
+        if (vscode.workspace.getConfiguration().get<boolean>('cody.internal.intentDetectionAPI')) {
+            return true
+        }
+        return !!this.featureCodyIntentDetectionAPI.value.last
+    }
+}
```
janhartman, line 33: I renamed this file to better reflect the current functionality, which is only intent detection at the moment.

camdencheek (approved):

PR #5734, opened 2024-09-27
allow agent to initialize when no access token is provided

sqs:
The agent can now initialize when no access token is initially provided. Previously it would exit with a fatal error:

```
Cody Agent: failed to initialize VSCode extension at workspace root path 'file:///Users/pkukielka/Work/jetbrains-test2/': Error: AuthStatus is not initialized
```

Fixes https://linear.app/sourcegraph/issue/CODY-3915/agent-fails-to-start-up-if-no-access-token-is-provided.

## Test plan

Added agent test.

### DIFF
--- a/agent/src/vscode-shim.ts
+++ b/agent/src/vscode-shim.ts
@@ -134,10 +134,6 @@ export function isAuthenticationChange(newConfig: ExtensionConfiguration): boole
         return true
     }
 
-    if (!newConfig.accessToken || !newConfig.serverEndpoint) {
-        return false
-    }
-
     return (
         extensionConfiguration.accessToken !== newConfig.accessToken ||
         extensionConfiguration.serverEndpoint !== newConfig.serverEndpoint
--- a/agent/recordings/unauthed_2245427793/recording.har.yaml
+++ b/agent/recordings/unauthed_2245427793/recording.har.yaml
@@ -1,0 +1,1208 @@
+log:
+  _recordingName: unauthed
+  creator:
+    comment: persister:fs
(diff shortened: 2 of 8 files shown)


agent/src/vscode-shim.ts
```diff
@@ -134,10 +134,6 @@ export function isAuthenticationChange(newConfig: ExtensionConfiguration): boole
         return true
     }
 
-    if (!newConfig.accessToken || !newConfig.serverEndpoint) {
```
sqs, line 137: @sourcegraph/jetbrains-plugin This makes it so that if partial config is passed, the agent will unauth itself because it got no accessToken. It is correct per the type system since ExtensionConfiguration is not a `Partial<ExtensionConfiguration>`, but will this break JetBrains?

dominiccooney: Maybe yes: https://sourcegraph.slack.com/archives/C06R69BC8UW/p1727478192421909?thread_ts=1727469741.848119&cid=C06R69BC8UW

But let's try to switch JetBrains to native sidebar sign-in buttons early next week.

sqs: @dominiccooney You want to approve/merge this PR then? Or wait until the week?

pkukielka: @dominiccooney I'm fine with merging this and then fixing JetBrains where necessary on my PR.
Question is what about other editors.

pkukielka (approved):
LGTM, it does not make situation worse than it is now, and we can do necessary fixes on the client side.

PR #5736, opened 2024-09-27
improved StateDebugView

sqs:
- more minimal UI so you can keep it enabled while devving
- shows in Cody Web demo

image

## Test plan

n/a

### DIFF
--- a/web/lib/agent/agent.client.ts
+++ b/web/lib/agent/agent.client.ts
@@ -85,6 +85,7 @@ export async function createAgentClient({
                 'cody.autocomplete.enabled': false,
                 'cody.experimental.urlContext': true,
                 'cody.web': true,
+                'cody.internal.debug.state': !!process.env.CODY_WEB_DEMO,
             },
         },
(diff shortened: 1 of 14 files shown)


web/lib/agent/agent.client.ts
```diff
@@ -85,6 +85,7 @@ export async function createAgentClient({
                 'cody.autocomplete.enabled': false,
                 'cody.experimental.urlContext': true,
                 'cody.web': true,
+                'cody.internal.debug.state': true,
```
dominiccooney, line 88: Would this mean web is seeing this debug state by default? CC @vovakulikov

sqs: Only in the internal `pnpm -C web dev` subproject. Not in the Cody Web product.

sqs: Actually, nevermind...it would. Will change.

pkukielka (approved):
LGTM

PR #5737, opened 2024-09-27
Auth: Update separator label for endpoint history

abeatrix:
Update the accounts separator label from "current" to "account history" for cases where the top endpoint is not always the current one in the Quickpick for switching accounts.

I also did some minor code cleanup for readability.

Closes https://linear.app/sourcegraph/issue/CODY-3896/switch-account-quickpick-incorrectly-marks-current

## Test plan

Open the Switch Account QuickPick to confirm the "current" separator has been updated to "account history"

image

### Before 

image


kalanchan (approved):
nice!

PR #5739, opened 2024-09-28
fix Cody Web issue with @-mention menu appearing behind tabs bar

sqs:
The removal of `z-index: 1` is what was needed, and I don't see why other elements would be on top of this tab bar or its popovers.

The other diffs are from stylelint.

atmention-zindex-issue

## Test plan

Test that @-mentions are not clipped by the tabs bar in `pnpm -C web dev`.


vovakulikov (approved):
Thanks, it was needed for Safari tooltips overlaps, but since then layout has changed a bit and it's no longer needed (Checked manually in Safari within Sourcegraph web)

PR #5741, opened 2024-09-28
pass chat history to webview using webviewAPI, not old postMessage protocol

sqs:
Previously, chat history was synced to the webview using the old postMessage protocol. Now, the ChatHistoryManager exposes an observable and that is read directly (using the standard webviewAPI) by the webview. This is a code refactor with no user-facing behavior change.

## Test plan

In Cody Web and in VS Code, try downloading chat history. Ensure the JSON file is downloaded and is a JSON array of chat history.

### DIFF
--- a/agent/src/agent.ts
+++ b/agent/src/agent.ts
@@ -1232,7 +1232,10 @@ export class Agent extends MessageHandler implements ExtensionClient {
             modelID ??= (await firstResultFromOperation(modelsService.getDefaultChatModel())) ?? ''
             const chatMessages = messages?.map(PromptString.unsafe_deserializeChatMessage) ?? []
             const chatBuilder = new ChatBuilder(modelID, chatID, chatMessages)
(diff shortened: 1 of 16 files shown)


vovakulikov (commented):
Tested manually in Cody Web demo and linked in Sourcegraph Web 
Found a few problems (nothing big but just UI/UX regressions)

- It looks like there is no way to reach zero empty state that we've added here https://github.com/sourcegraph/cody/pull/5390 anymore
- Chat deleting doesn't work right if we try to delete the chat that we're currently in. Not sure this is a PR problem, the real issue here is how we work with active panel IDs (once we delete chat we delete the panel id but without the panel id we can't send anything through the postMessage channel, feel free to ignore it)

sqs:
@vovakulikov Cool, I fixed the empty state problem and added a test to verify the fix (after confirming it manually).

sqs:
Stamp because @vovakulikov reviewed it and I fixed the one outstanding issue

burmudar (approved):

PR #5746, opened 2024-09-29
consolidate sign-in forms and use standard tailwind styles

sqs:
- Consolidates the sign-in forms (previously there was a misnomer OnboardingExperiment and then a ClientSignInForm in 2 separate files)
- Uses Tailwind styles instead of the old-style and more verbose CSS
- Fixes the display of the "invalid URL" warning (FormMessage was not triggering the matcher because `<input type="url">` meant it failed at an even earlier stage)
- Updates the storybook

No behavior change, slight look-and-feel change but no change in wording or layout.

### VS Code

image

### JetBrains

image

## Test plan

Try signing in using VS Code.


dominiccooney:
@abeatrix , @olafurpg does VS/Eclipse use this alternative login form? AFAIK it currently does not work in JetBrains and is only displayed in error, although @pkukielka is working on fixing one of them.

sqs:
@dominiccooney:

> @abeatrix , @olafurpg does VS/Eclipse use this alternative login form? AFAIK it currently does not work in JetBrains and is only displayed in error, although @pkukielka is working on fixing one of them.

Based on the poor state of the client name detection, I would be surprised if this was working in any other client. I did try to not break the way it reads config or calls commands, though, so it should work. (I have a separate PR where I will change that more and actually I now have it working with `pnpm -C web dev` so you can sign in there through this sign-in form.)

olafurpg:
FWIW Eclipse is using the login form from the agent server but browser redirect doesn't work, only paste access token.

sqs:
@olafurpg Cool, that works with this and it looks nicer now.

vscode/webviews/AuthPage.tsx
```diff
@@ -0,0 +1,269 @@
...
+    vscodeAPI,
+    codyIDE,
+}) => {
+    const authStatus = useConfig().authStatus
+    const telemetryRecorder = useTelemetryRecorder()
+    const otherSignInClick = (): void => {
+        vscodeAPI.postMessage({ command: 'auth', authKind: 'signin' })
+    }
+    const isNonVSCodeIDE = codyIDE !== CodyIDE.Web && codyIDE !== CodyIDE.VSCode
+    const isCodyWebUI = (uiKindIsWeb || codyIDE === CodyIDE.Web) && !isNonVSCodeIDE
```
pkukielka, line 33: Minor nit: This conditions are a bit hard to parse. Aren't they in fact evaluating to:

```const isCodyWebUI = codyIDE === CodyIDE.Web || (uiKindIsWeb && codyIDE === CodyIDE.VSCode)```

?

pkukielka (approved):
LGTM

pkukielka:
> FWIW Eclipse is using the login form from the agent server but browser redirect doesn't work, only paste access token.

I reported redirect not working on Friday, we have separate fixes for that @olafurpg

vscode/webviews/AuthPage.tsx
sqs: Yeah, this was how the code was. I have another PR to clean this up, but I wanted to make this PR have no logic changes to reduce risk of breakage.

abeatrix:
> @abeatrix , @olafurpg does VS/Eclipse use this alternative login form? AFAIK it currently does not work in JetBrains and is only displayed in error, although @pkukielka is working on fixing one of them.

VS is using this alternative login form and I believe Eclipse as well (@jamesmcnamara can confirm). Also worked with @pkukielka last week to fix the redirect for JetBrains so hopefully all the editors will be using the unified login form when it's ready. 

Thanks @sqs for standardizing the styles!

PR #5749, opened 2024-09-30
rm unneeded ChatEnvironmentContext

sqs:
This just reads the `agentIDE` value, and it can do so directly without a separate context provider.

## Test plan

CI


dominiccooney (approved):

dominiccooney:
Looks good, CI says some markdown tests need updating.

PR #5750, opened 2024-09-30
centralize accessing the agentIDE and related values

sqs:
Previously, we were doing 2 problematic things:

1. Determining extension/agent behavior based on the identity of the editor (`agentIDE`), instead of based on the client's reported capabilities. This is brittle and it is not clear *what* about (say) VS Code makes this code path work and this one not.
2. Taking the editor's identity from the agent (but falling back to VS Code implicitly), which meant that we were inconsistently doing `agentIDE ?? CodyIDE.VSCode` and similar for the extension version in many places. Of course, it's easy to make a mistake here.

This doesn't solve all of the problems, but it helps:

1. All places where we read the `agentIDE` are now centralized in a global `clientCapabilities()`. This will make future refactoring easier, solves a few problems (such as *explicitly* defaulting to `CodyIDE.VSCode`, which is not a great idea but at least codifies our current behavior with the type system), and makes it clear that this value can't change throughout the lifetime of the extension.
2. The gotchas when reading `agentIDE` and related values are documented, and the symbols are jsdoc'd with `@deprecated` to make them noticeable.

There is no intentional user-facing behavior change.

I found this code tricky when improving the sign-in page in https://github.com/sourcegraph/cody/pull/5746 because it was not clear why something was gated for VS Code. For example, the JetBrains UI in the storybook was configured to claim that it was running on the web, which made no sense but might be because it needed a UI more similar to what Cody Web used, and someone just reused that flag. We want to get away from that!

## Test plan

Agent tests and e2e tests are pretty good here. Also inspect telemetry and ensure that the right client name is being sent.

### DIFF
--- a/lib/shared/src/configuration/clientCapabilities.ts
+++ b/lib/shared/src/configuration/clientCapabilities.ts
@@ -1,0 +1,128 @@
+import { type ClientConfiguration, CodyIDE } from '../configuration'
+
+/**
+ * The capabilities of the client, such as the editor that Cody is being used with (directly for VS
+ * Code or via the agent for other editors) or the CLI.
+ */
+export interface ClientCapabilities {
+    /**
+     * The `agentIDE` value, which is the editor that Cody is being used with. If not set, it
+     * defaults to {@link CodyIDE.VSCode} to match the previous behavior.
+     *
+     * @deprecated Use one of the other {@link ClientCapabilities} fields instead to assert based on
+     * the client's actual capabilities at runtime and not your assumptions about their current
+     * capabilities, and to support future clients that are not in the {@link CodyIDE} enum. If you
+     * are truly interested in whether the editor is VS Code, such as to use the right link to
+     * documentation or describe the VS Code user interface, then it is OK to use this field.
+     */
+    agentIDE: CodyIDE
+
+    /**
+     * @deprecated For the same reason as {@link ClientCapabilities.agentIDE}.
+     */
+    isVSCode: boolean
+
+    /**
+     * @deprecated For the same reason as {@link ClientCapabilities.agentIDE}.
+     */
+    isCodyWeb: boolean
+
+    /**
+     * The `agentExtensionVersion` value, which is the version of Cody that is being used in the
+     * agent (if set).
+     *
+     * @deprecated Be careful when using this. It is NOT set if using VS Code, and you need to use
+     * the `/vscode/src/version.ts` module's `version` export. Soon this {@link ClientConfiguration}
+     * value will expose that.
+     */
+    agentExtensionVersion?: string
+
+    /**
+     * The `agentIDEVersion` value, which is the version of the client editor (if set).
+     */
+    agentIDEVersion?: string
+
+    /**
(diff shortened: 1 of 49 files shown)


lib/shared/src/configuration/clientCapabilities.ts
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * The capabilities of the client, such as the editor that Cody is being used with (directly for VS
+ * Code or via the agent for other editors) or the CLI.
+ */
+export interface ClientCapabilities {
+    /**
+     * The `agentIDE` value, which is the editor that Cody is being used with. If not set, it
+     * defaults to {@link CodyIDE.VSCode} to match the previous behavior.
+     *
+     * @deprecated Use one of the other {@link ClientCapabilities} fields instead to assert based on
```
dominiccooney, line 12: I appreciate the direction, we should be doing more feature detection. But does deprecation go too far? In a few places in the UI we direct to VSCode/Marketplace links, refer to the product (and I could see us referring to parts of the IDE like side bar vs tool window), echo the version number, etc.

I think having the `agentIDE` for those is fine, especially if we write exhaustive switch statements... It would be tedious to pass all that in like a localization bundle or something.

lib/shared/src/configuration/clientCapabilities.ts
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * The return value does not change. However, it is only available after the configuration has been
+ * set, and this function throws if it's not available. This means that it can't be used at
+ * initialization time.
+ */
+export function clientCapabilities(): ClientCapabilities {
+    if (_mockValue) {
+        return _mockValue
+    }
+    if (!_configuration) {
```
dominiccooney, line 66: Is it worth making this observable to reduce the likelihood we have problems with someone casually reading this too early?

lib/shared/src/configuration/clientCapabilities.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+let _mockValue: ClientCapabilities | undefined
+
+/**
+ * Mock the {@link clientCapabilities} result.
+ *
+ * For use in tests only.
+ */
+export function mockClientCapabilities(value: ClientCapabilities | null): void {
```
dominiccooney, line 116: Why not simply make the type `ClientCapabilities | undefined` and be done with it?

dominiccooney (approved):
Only skimmed this and some comments inline, but looks OK.

lib/shared/src/configuration/clientCapabilities.ts
sqs: I did that originally, but there are some places where we read this value in a constructor called synchronously from the extension's `activate` function, which means it would not be ready by then.

We could make this observable, but that would be another refactor, and I wanted to minimize this diff.

lib/shared/src/configuration/clientCapabilities.ts
sqs: Yeah, that is fine. I'll add a comment to that effect. And we can remove the `@deprecated` once we have a good centralized way of doing this.

PR #5751, opened 2024-09-30
pass config to webview through new Observable-based webview API

sqs:
The webview protocol message `config` had 2 problems:

1. It was sent (in `ChatController.sendConfig`) whenever ChatController thought it might have changed, which meant a bunch of unnecessary calls to `sendConfig` sprinkled throughout ChatController.
2. It is a different schema from the other configuration, with `configFeatures` and other things having custom logic to construct them in ChatController.

This solves problem 1. Now, the config is sent through the `legacyConfig(): LegacyWebviewConfig` observable webview API method to the webview, which means it gets an up-to-date view of the config. It uses the same type, so this does not yet address problem 2, but it makes it much easier to address it.

There are several other fixes and changes to account for how the value is sent immediately when it changes and not on the previously more brittle schedule.

Removes the old `config` webview protocol message.

There is no user-facing behavior change.

## Test plan

CI and e2e


dominiccooney (changes requested):
Ad-hoc testing by @pkukielka indicates this regresses login redirection for free/pro users in JetBrains, and exacerbates a problem where theme styles sometimes do not apply.

Frankly, the rate of these changes is too high given the latency of the JetBrains QA and release cycle. We need authors to do more manual testing on JetBrains and contribute JetBrains tests if we want to keep producing releases.

sqs:
@dominiccooney I updated it and believe I have fixed the theme issue. I'll check the redirect issue as well.

sqs:
@dominiccooney @pkukielka I reviewed this branch on latest `sourcegraph/jetbrains`. See https://gist.github.com/sqs/3329c890ce45d7c07b14c5b4e22fa173?permalink_comment_id=5222610#gistcomment-5222610 for a list of issues.

I believe they are all issues that will be fixed by @pkukielka's work on making Cody for JetBrains use the webviews for auth. The "problem" seems to be that it starts to route users to the webview-based auth routes, which don't fully work in JetBrains yet. We can either:

1. Delay merging this PR *until* we make the webviews instead route users to the JetBrains plugin settings instead of the webview UI --OR--
2. Merge this PR and have @pkukielka follow up with those changes

I would prefer the former but defer to you, @dominiccooney.

pkukielka:
@sqs let's delay merging this if we can, at least for a couple of days.

My PR still have some rough edges which I would like to correct but I also need to focus on some gateway/remote fixes for Bridgewater and for the release at the end of this week. I think merging it all now could impact release date if we would discover something does not work as expected. So unless this PR blocks something critical in Cody itself I think it will be much safer this way.

sqs:
@pkukielka Sounds good, will do.

sqs:
@pkukielka @dominiccooney Good to review/merge now, or still want to wait?

dominiccooney:
@sqs JetBrains FYI is red, let's not break the JetBrains build.

PR #5756, opened 2024-09-30
Remove automatically generated Tokens to reduce redundant tokens

arafatkatze:
Solves [Linear issue](https://linear.app/sourcegraph/issue/CODY-350/bug-revoke-the-access-token-when-a-user-signs-out-of-jetbrains-and)

## Problem Statement
When signing in to Sourcegraph (and potentially JetBrains) using certain sign-in providers, access tokens are automatically generated. These tokens are subject to deletion under specific conditions, primarily upon logging out. However, there are nuances depending on the method of sign-in, the access token source, and whether multiple instances are involved.

Details:
1. Automatic Token Generation:
•Tokens can be automatically generated when signing in using one of the following providers:
•GitHub
•GitLab
•Google
•Sourcegraph instance (via  sign-in without token for v5.1 and above)
2. Token Deletion Criteria:
•An automatically generated token will only be deleted upon logging out of the instance it was created for.
•If you sign out, the token will be deleted.
•However, if you switch instances or change access tokens while signed in using an automatically generated token, the token will not be deleted.
•Token deletion only occurs when the user logs out from the instance where the token was generated, and only if the token was created using one of the automatic sign-in methods listed above.
3. Instance and Token Behavior:
•If the user switches to a different instance or uses a different token during a session, the original token remains active until explicitly logged out.
•If the user returns to the original instance or token, no deletion occurs unless the logout process is followed.
4. Manual Token Override:
•If the user manually inputs an access token from the instance (rather than using automatic generation via sign-in), the token is not subject to automatic deletion on logout.

[See video Proof
](https://www.loom.com/share/2a7bea8e2bbe4dcabc85cf7429a65c9d)

Thanks to Peter Guy, Dominic and Erik for the strategy on this.

## Test plan
Tested locally on both a local sourcegraph instance and with dotcom instance using Vscode cody.

### DIFF
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -5,12 +5,20 @@ import type { PromptString } from './prompt/prompt-string'
 import type { ReadonlyDeep } from './utils'
 
 /**
+ * Represents the source of an authentication token generation, either a redirect or non-redirect flow.
+ * A redirect flow is initiated by the user clicking a link in the browser, while a non-redirect flow is initiated by the user
+ * manually entering the access from into the VsCode App.
+ */
+export type TokenSource = 'redirect' | 'nonredirect'
+
+/**
  * The user's authentication credentials, which are stored separately from the rest of the
  * configuration.
  */
 export interface AuthCredentials {
     serverEndpoint: string
     accessToken: string | null
+    tokenSource?: TokenSource | undefined
 }
 
 interface RawClientConfiguration {
--- a/lib/shared/src/configuration.ts
+++ b/lib/shared/src/configuration.ts
@@ -5,11 +5,11 @@ import type { PromptString } from './prompt/prompt-string'
 import type { ReadonlyDeep } from './utils'
 
 /**
- * Represents the source of an authentication token generation, either a redirect or non-redirect flow.
- * A redirect flow is initiated by the user clicking a link in the browser, while a non-redirect flow is initiated by the user
+ * Represents the source of an authentication token generation, either a redirect or paste flow.
+ * A redirect flow is initiated by the user clicking a link in the browser, while a paste flow is initiated by the user
  * manually entering the access from into the VsCode App.
  */
-export type TokenSource = 'redirect' | 'nonredirect'
+export type TokenSource = 'redirect' | 'paste'
 
 /**
  * The user's authentication credentials, which are stored separately from the rest of the
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -32,6 +32,7 @@ import {
     CURRENT_USER_CODY_SUBSCRIPTION_QUERY,
     CURRENT_USER_ID_QUERY,
     CURRENT_USER_INFO_QUERY,
+    DELETE_ACCESS_TOKEN_MUTATION,
     EVALUATE_FEATURE_FLAG_QUERY,
     FILE_CONTENTS_QUERY,
     FILE_MATCH_SEARCH_QUERY,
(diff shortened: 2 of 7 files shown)


arafatkatze:
@dominiccooney @peterguy @eseliger 

After exploring various options and considering the feedback from our in-person discussions, this approach is a compromise between simplicity and functionality. The key advantage here is that it keeps everything client-side, relying on local storage, which reduces complexity significantly compared to a server-side solution. This approach ensures that only the necessary tokens are deleted while maintaining minimal overhead on the client side.

I’ve tested this method using the debugger and uploaded a recording too. However, I recognize that this solution lacks formal automated tests. Based on Rik's work, I believe it would be helpful to integrate some of his testing methodology to further validate this implementation. 

Let me know your thoughts on moving forward with this approach.

lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1249,7 +1250,16 @@ export class SourcegraphGraphQLAPIClient {
...
+    // Deletes an access token, if it exists on the server
+    public async DeleteAccessToken(token: string): Promise<unknown | Error> {
+        const initialResponse = await this.fetchSourcegraphAPI<APIResponse<unknown>>(
+            DELETE_ACCESS_TOKEN_MUTATION,
+            {
+                token,
+            }
+        )
+        return extractDataOrError(initialResponse, data => data)
+    }
```
dominiccooney, line 1262:
```suggestion
    }
    
```

Let's separate these methods with one blank line for consistency.

vscode/src/auth/auth.ts
```diff
@@ -79,9 +80,10 @@ export async function showSignInMenu(
             // Auto log user if token for the selected instance was found in secret
             const selectedEndpoint = item.uri
             const token = await secretStorage.getToken(selectedEndpoint)
+            const tokenSource = await secretStorage.getTokenSource(selectedEndpoint)
             let { authStatus } = token
                 ? await authProvider.validateAndStoreCredentials(
-                      { serverEndpoint: selectedEndpoint, accessToken: token },
+                      { serverEndpoint: selectedEndpoint, accessToken: token, tokenSource: tokenSource },
```
dominiccooney, line 86: When the property name and the variable name are the same, you can just name the variable.
```suggestion
                      { serverEndpoint: selectedEndpoint, accessToken: token, tokenSource },
```

lib/shared/src/configuration.ts
```diff
@@ -4,13 +4,21 @@ import type { ChatModelProviderConfig } from './models/sync'
 import type { PromptString } from './prompt/prompt-string'
 import type { ReadonlyDeep } from './utils'
 
+/**
+ * Represents the source of an authentication token generation, either a redirect or non-redirect flow.
+ * A redirect flow is initiated by the user clicking a link in the browser, while a non-redirect flow is initiated by the user
+ * manually entering the access from into the VsCode App.
+ */
+export type TokenSource = 'redirect' | 'nonredirect'
```
dominiccooney, line 12: Can we make this a positive statement of what it is, like `'redirect' | 'paste'`? Because in future we may have other token sources.

Alternatively, you could just name this what it does: `export type LogoutBehavior = 'delete' | undefined` something like that.

I prefer recording the source because we might find that useful for other things later...

vscode/src/auth/auth.ts
```diff
@@ -361,6 +367,14 @@ export async function showSignOutMenu(): Promise<void> {
  * Log user out of the selected endpoint (remove token from secret).
  */
 async function signOut(endpoint: string): Promise<void> {
+    const token = await secretStorage.getToken(endpoint)
+    const tokenSource = await secretStorage.getTokenSource(endpoint)
+    // Delete the access token from the Sourcegraph instance on signout if it was created
```
dominiccooney, line 372: Very nice and clear.

dominiccooney (approved):
Looks good. Some minor feedback inline.

Slightly larger: I think we should store related properties together, especially when they are small, instead of doing multiple reads and writes like this. The multiple reads open up the possibility of torn/overlapping reads and writes. However reading and writing a single property across different versions of the clients is hard, so you don't need to worry about it in this PR.

lib/shared/src/configuration.ts
arafatkatze: > Can we make this a positive statement of what it is, like 'redirect' | 'paste'? Because in future we may have other token sources.


This looks a lot more intuitive so I changed to this.

PR #5758, opened 2024-09-30
Pass intent scores to telemetry

thenamankumar:
closes: https://linear.app/sourcegraph/issue/SRCH-1028/log-telemetry-of-the-output-of-the-intent-detection-model-with-each

Passes intent scores to `chat-question:executed` event.

## Test plan

- execute chat and check if the intent scores are logged.
- execute chat with manually triggering search intent, that should not log the intent scores. 
- execute chat with manually triggering chat intent, that should not log the intent scores.

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -41,6 +41,7 @@ import {
     GET_REMOTE_FILE_QUERY,
     GET_URL_CONTENT_QUERY,
     HIGHLIGHTED_FILE_QUERY,
+    LEGACY_CHAT_INTENT_QUERY,
     LEGACY_CONTEXT_SEARCH_QUERY,
     LOG_EVENT_MUTATION,
     LOG_EVENT_MUTATION_DEPRECATED,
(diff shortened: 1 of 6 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -935,8 +944,13 @@ export class SourcegraphGraphQLAPIClient {
 
     /** Experimental API */
     public async chatIntent(interactionID: string, query: string): Promise<ChatIntentResult | Error> {
+        const hasAllScoresField = await this.isValidSiteVersion({
```
vovakulikov, line 947: It's not this PR problem, but I've noticed that this version compatibility check can become very big and complex. 
In my last change with supporting match ranges, I encountered this. 

Curios maybe we should ensure our politics about how many versions behind we support in the latest VSCode release.

vovakulikov (approved):

lib/shared/src/sourcegraph-api/graphql/client.ts
thenamankumar: hmm, I'll check with @chenkc805 if we have any stats on what versions the users are on. But we can't basically break it ever, so we will need a banner or something to force users to upgrade their instance.

PR #5760, opened 2024-09-30
add(chat): Alt+l adds current selection to chat

jamesmcnamara:
Adds the current behavior of "Alt+/" to "Alt+l" which is that if you are continuing a conversation and re-open the chat, the current text selection is added to the chat. 

One bug that this fixes for both is that we shouldn't add the same selection multiple times, so this instead only adds new selections if they aren't the same as older ones and merges two if they overlapping. 

## Test plan
Added an E2E test and unit tests

## Changelog
"Alt-L" will add the current text selection to a continued conversation


vscode/package.json
```diff
@@ -641,6 +641,11 @@
           "editorFocus": false
         }
       },
+      {
+        "command": "cody.mention.selection",
+        "key": "alt+l",
```
abeatrix, line 646: I know it's the current behavior, but curious to know if there is a reason we want to keep two keybindings for the same commands.

abeatrix (approved):
Manually tested and verified alt+L and alt+/ works as described.

Unblocking comment: maybe it's just me, but sometimes I would re-add my current selection to chat to make it easier for the LLM  (and myself lol) to understand which part of the code I'm referring to in my prompt, so not being able to add an exisiting selection as @-mention chip again makes it look like a bug for me 😅

vscode/package.json
jamesmcnamara: Mostly so that if anyone is using it and has built up muscle memory around it we won't break it for them.

jamesmcnamara:
@abeatrix the deduplication only happens on the current message and for the current input source (i.e. context coming from user selection). So you can still re-add past mentions to a new message, it's only if you have multiple overlapping references in the same message that they'll be de-duplicated.

abeatrix:
> @abeatrix the deduplication only happens on the current message and for the current input source (i.e. context coming from user selection). So you can still re-add past mentions to a new message, it's only if you have multiple overlapping references in the same message that they'll be de-duplicated.

sounds good to me! It's a good fix in general anyway, thanks for fixing it!

PR #5763, opened 2024-09-30
Telemetry: Record installation events for all non-Web clients

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-49/bug-v2-telemetry-for-cody-installations-not-being-fired

I think this might have already been fixed when we update the `clientCapabilities().isCodyWeb` check to returns false when agentIDE is not set to CodyIDE.Web. Before that change, `clientCapabilities().isCodyWeb` would always returns true for all IDEs.

We can confirm by enabling E2E tests for logging the `installed` event and installing Cody in a new machine.

## Test plan

1. Green CI to confirm the `cody.extension:installed` events are now being logged.
2. In a new machine, confirm the `cody.extension:installed` events is logged:

image

```
█ telemetry-v2: recordEvent: cody.extension/installed  {
  "parameters": {
    "version": 0,
    "billingMetadata": {
      "product": "cody",
      "category": "billable"
    }
  },
  "timestamp": "2024-10-01T16:35:14.262Z"
}
```

NOTE: The installed event does not fire for users who uninstalled then re-install Cody.

### DIFF
--- a/vscode/src/services/telemetry-v2.ts
+++ b/vscode/src/services/telemetry-v2.ts
@@ -84,9 +84,11 @@ export function createOrUpdateTelemetryRecorderProvider(
              * On first initialization, also record some initial events.
              * Skip any init events for Cody Web use case.
              */
-            const newAnonymousUser = localStorage.checkIfCreatedAnonymousUserID()
+            // Makes sure anonymousUserID is set before recording events.
+            const anonymousUserID = localStorage.anonymousUserID()
+            const newAnonymousUserCreated = localStorage.checkIfCreatedAnonymousUserID()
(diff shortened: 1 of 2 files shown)


akalia25:
Thanks @abeatrix! Just a note, that on the VSCode E2E tests, `cody.extension.installed` has been firing successfully; its being validated on each test's expectedEvents list ([ref](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/test/e2e/auth.test.ts?L10)), that comment on `vscode/test/e2e/helpers.ts` is an old one from me, sorry for not updating that!

abeatrix:
looks like this is already fixed and no longer an issue

PR #5767, opened 2024-10-01
use the accurate docContext and persist repo visibility using local storage

hitesh-1997:
The PR improves the following:
1. When calculating the `diff` over time for suggestion, we were using `document.getText`. Analyzing offline telemetry we see in ~20% we get some extra chars if user types between calculating `docContext` and `loaded` call. We leverage already computed `completePrefix` and `completeSuffix` to compute the text, which is accurate.
2. We were using in-memory cache for calculating `repoVisibility` before. But without passing token, we frequently hit rate limit for querying github api and cached tokens are evicted from in-memory cache when user refresh. This PR stores the tokens in `localStorage` instead of in-memory store to cache the repo visibility.
3. Adds the `code qwen` and `deepseek` model optimized for long prompts.

## Test plan
1. Added unit tests.
2. Manual Testing: Adding debugger and refreshing to see if the token value is persisted/ deleted from local storage.

### DIFF
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -848,8 +848,8 @@ export class InlineCompletionItemProvider
                 suggestionEvent.markAsRead({
                     document: invokedDocument,
                     position: invokedPosition,
-                    docPrefix: completion.requestParams.docContext.completePrefix,
-                    docSuffix: completion.requestParams.docContext.completeSuffix,
+                    completePrefix: completion.requestParams.docContext.completePrefix,
+                    completeSuffix: completion.requestParams.docContext.completeSuffix,
                 })
             }
         }, this.COMPLETION_VISIBLE_DELAY_MS)
--- a/vscode/src/completions/inline-completion-item-provider.ts
+++ b/vscode/src/completions/inline-completion-item-provider.ts
@@ -843,10 +843,13 @@ export class InlineCompletionItemProvider
                 takeSuggestWidgetSelectionIntoAccount,
                 undefined
             )
+            completion.requestParams.docContext.position
             if (isStillVisible) {
                 suggestionEvent.markAsRead({
                     document: invokedDocument,
                     position: invokedPosition,
+                    docPrefix: completion.requestParams.docContext.completePrefix,
+                    docSuffix: completion.requestParams.docContext.completeSuffix,
                 })
             }
         }, this.COMPLETION_VISIBLE_DELAY_MS)
--- a/vscode/src/completions/logger.ts
+++ b/vscode/src/completions/logger.ts
@@ -7,6 +7,7 @@ import {
     type BillingCategory,
     type BillingProduct,
     currentAuthStatusAuthed,
+    displayPathWithoutWorkspaceFolderPrefix,
     isDotCom,
     isNetworkError,
     telemetryRecorder,
@@ -767,7 +768,7 @@ function getInlineContextItemContext(
             content,
             startLine,
             endLine,
-            filePath: uri.fsPath,
+            filePath: displayPathWithoutWorkspaceFolderPrefix(uri),
         })),
     }
 }
@@ -776,8 +777,8 @@ function suggestionDocumentDiffTracker(
     interactionId: CompletionAnalyticsID,
     document: vscode.TextDocument,
     position: vscode.Position,
-    docPrefix: string,
-    docSuffix: string
+    completePrefix: string,
+    completeSuffix: string
 ): void {
     // If user is not in the same document, we don't track the diff.
     if (document.uri.scheme !== 'file') {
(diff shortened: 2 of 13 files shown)


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -843,10 +843,13 @@ export class InlineCompletionItemProvider
                 takeSuggestWidgetSelectionIntoAccount,
                 undefined
             )
+            completion.requestParams.docContext.position
             if (isStillVisible) {
                 suggestionEvent.markAsRead({
                     document: invokedDocument,
                     position: invokedPosition,
+                    docPrefix: completion.requestParams.docContext.completePrefix,
+                    docSuffix: completion.requestParams.docContext.completeSuffix,
```
valerybugakov, line 852: Let's avoid renaming things as much as possible to make it easier to trace the data flow. It's unclear if the `markAsRead` implementation expects `completePrefix` or `prefix` based on its function parameters. 

```suggestion
                    completePrefix: completion.requestParams.docContext.completePrefix,
                    completeSuffix: completion.requestParams.docContext.completeSuffix,
```

vscode/src/completions/logger.ts
```diff
@@ -775,7 +775,9 @@ function getInlineContextItemContext(
 function suggestionDocumentDiffTracker(
     interactionId: CompletionAnalyticsID,
     document: vscode.TextDocument,
-    position: vscode.Position
+    position: vscode.Position,
+    docPrefix: string,
+    docSuffix: string
```
valerybugakov, line 780:
```suggestion
    completePrefix: string,
    completeSuffix: string
```

vscode/src/completions/model-helpers/codeqwen.ts
```diff
@@ -0,0 +1,82 @@
...
+    getOllamaPrompt(promptContext: GetOllamaPromptParams): PromptString {
+        const { context, currentFileNameComment, prefix, suffix } = promptContext
+
+        const infillPrefix = context.concat(currentFileNameComment, prefix)
+
+        return ps`<|fim_prefix|>${infillPrefix}<|fim_suffix|>${suffix}<|fim_middle|>`
+    }
+
+    getOllamaRequestOptions(isMultiline: boolean): OllamaGenerateParameters {
+        const params = {
```
valerybugakov, line 35: Did you have a chance to test it locally with Ollama?

vscode/src/completions/model-helpers/codeqwen.ts
```diff
@@ -0,0 +1,82 @@
...
+        return ps`<|file_sep|>${PromptString.fromDisplayPath(snippet.uri)}\n${content}`
+    }
+
+    formatPrompt(params: FormatPromptParams): PromptString {
+        // Prompt format for CodeQwen in technical report: https://arxiv.org/pdf/2409.12186
+        const { intro, prefix, suffix, repoName, fileName } = params
+        let introPrefix = ps``
+        if (intro.length > 0) {
+            introPrefix = ps`${intro}\n`
+        }
```
valerybugakov, line 75: Is this already handled in `formatIntroSnippets`?

valerybugakov (approved):

vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -843,10 +843,13 @@ export class InlineCompletionItemProvider
                 takeSuggestWidgetSelectionIntoAccount,
                 undefined
             )
+            completion.requestParams.docContext.position
```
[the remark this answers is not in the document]
hitesh-1997: `position` is being used in the `persistenceTracker` to calculate range around trigger point.

vscode/src/completions/inline-completion-item-provider.ts
hitesh-1997: makes sense, changed as per suggestion

vscode/src/completions/logger.ts
hitesh-1997: done

vscode/src/completions/model-helpers/codeqwen.ts
hitesh-1997: ah, sorry didn't test with Ollama
for now, removed the ollama options from the file, as we want to A/B test the model itself.

vscode/src/completions/model-helpers/codeqwen.ts
hitesh-1997: yes, removed from here

vscode/src/completions/inline-completion-item-provider.ts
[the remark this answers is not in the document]
valerybugakov: @hitesh-1997, does this line do anything? It looks like a reference to the existing field on the `docContext` object and this line itself doesn't affect its presence.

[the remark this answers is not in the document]
hitesh-1997: Ah yes, sorry I looked at some other file.
raised this PR to clean this: https://github.com/sourcegraph/cody/pull/5793/files

PR #5558, opened 2024-09-11
Telemetry: At Mentions

RXminuS:
Before the `cody.at-mention` event fired as soon as the chat view was rendered leading to an overcount of DAUs. Duplicate mention events were also fired for other mention providers such as when typing a file name where each character would fire an event.

Now events only fire when actually triggered. We also clean up and add additional public metadata that can help evaluate the actual effectiveness of different mention providers.

Fixes CODY-3405

## Test plan
- Added new Telemetry testing framework. Run `cd vscode & pnpm run test:e2e2 --grep "telemetry`.
- See the snapshots to view the included data

PR #5770, opened 2024-10-01
Rename onebox telemetry metadata fields

thenamankumar:
This PR renames a bunch of OneBox telemetry events metadata fields and make them camelcase.

~~This PR also removes the check `isDotCom` from the `promptText` private metadata field passed with the `chat-question:executed` event. None of the private metadata fields are sent to SG through the telemetery pipeline, so that data is still private, similar to the other present privateMetadata field. The only change is that the `promptText` will be recorded in the event_logs on that sourcegraph instance.~~

We need to remove the check so that we can selectively collect that field for the S2 instance, configured by a env var here: https://github.com/sourcegraph/cloud/pull/15143

## Test plan

Make sure all the updated events are still recorded correctly.

### DIFF
--- a/vscode/webviews/chat/Transcript.tsx
+++ b/vscode/webviews/chat/Transcript.tsx
@@ -274,9 +274,9 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
                         recordsPrivateMetadataTranscript: 1,
                     },
                     privateMetadata: {
-                        initial_intent: humanMessage.intent,
-                        user_selected_intent: intent,
-                        query: editorState.text,
+                        initialIntent: humanMessage.intent,
+                        userSpecifiedIntent: intent,
+                        promptText: editorState.text,
                     },
                 })
             }
--- a/vscode/webviews/components/FileLink.tsx
+++ b/vscode/webviews/components/FileLink.tsx
@@ -120,8 +120,8 @@ export const FileLink: React.FunctionComponent<
         const external = uri.scheme === 'http' || uri.scheme === 'https'
         telemetryRecorder.recordEvent('onebox.searchResult', 'clicked', {
             metadata: {
-                is_local: external ? 0 : 1,
-                is_remote: external ? 1 : 0,
+                isLocal: external ? 0 : 1,
+                isRemote: external ? 1 : 0,
             },
             privateMetadata: {
                 filename: displayPath(uri),


camdencheek (approved):

akalia25:
Hey @thenamankumar I thought we were only interested in collecting `query` (or now `promptText`) for the event `onebox.intentCorrection`, curious why the scope has changed to now include `cody.chat-question`?

vscode/src/chat/chat-view/ChatController.ts
```diff
@@ -896,12 +896,7 @@ export class ChatController implements vscode.Disposable, vscode.WebviewViewProv
                 userSpecifiedIntent,
                 properties,
                 privateContextSummary,
-                // 🚨 SECURITY: chat transcripts are to be included only for DotCom users AND for V2 telemetry
-                // V2 telemetry exports privateMetadata only for DotCom users
-                // the condition below is an additional safeguard measure
-                promptText:
-                    isDotCom(authStatus) &&
-                    (await truncatePromptString(inputText, CHAT_INPUT_TOKEN_BUDGET)),
+                promptText: await truncatePromptString(inputText, CHAT_INPUT_TOKEN_BUDGET),
```
akalia25, line 899: We have been _really_ strict on how we collect sensitive data like this, to ensure, it is never collected from customers. I do understand that `privateMetadata` is only collected for non dotcom instances, but to relax this additional dotcom check for transcripts, I will need to confirm with the team we are okay with this.

akalia25 (dismissed):

vscode/src/chat/chat-view/ChatController.ts
thenamankumar: Yes @akalia25 , I have tagged the @sourcegraph/security-code-review to get approvals. We can do according to the suggestions. 

To confirm again, @chenkc805 do we really need promptText for the `chat-question:executed` events on S2? Along with the detectedIntent, userSpecifiedIntent and intentScores? 

context: https://linear.app/sourcegraph/issue/SRCH-1028/log-telemetry-of-the-output-of-the-intent-detection-model-with-each

akalia25: Hey @thenamankumar could we add an `isS2()` clause (similar to `isDotCom()`), that would allow for transcripts `promptText`/`responseText` to be stored for both dotcom instance and S2.

This would help avoid collecting this data on customer's local PG instance and keeping it extra secure.

akalia25: something like this: `isDotCom(authStatus) || isS2(authStatus))`

RXminuS:
@thenamankumar we should maybe sync on this tomorrow because I just removed the field all together after was told it was unused. https://github.com/sourcegraph/cody/pull/5558 Also there's a new way of testing these now

vscode/src/chat/chat-view/ChatController.ts
thenamankumar: I have removed this change for now from this PR.

PR #5772, opened 2024-10-02
Update Prompts UI (enhanced with client last used action sorting)

vovakulikov:
Part of https://linear.app/sourcegraph/issue/SRCH-1103/update-prompts-ui-welcome-screen

This PR does a few things (TL;DR I suggest checking the [loom](https://www.loom.com/share/3922d57405fc4db2b328a317c211231e?sid=28e903cd-aa63-4c7c-93cb-262b1de6ef7d))
 
- Improve prompts and commands list UI (you can find the latest designs for it [here](https://www.figma.com/design/f078wFMKsIOaEwj7Iwj5xy/Unified-Cody?node-id=4845-20933&node-type=symbol&t=Ao2Y68OYj5lFV3rA-0))
- Adds smart last-used prompts/commands sorting

Screenshot 2024-10-02 at 02 52 01

So this PR is another attempt to bring commands and prompts closer and smooth transition from commands to prompts, so in this PR
- There is no separation between commands and prompts it is basically one list
- on the welcome screen we show only the last 4 used actions (action here means prompts or commands)
- By default, we show out-of-the-box actions 
- If we have custom commands or custom prompts we shift the list to show the last created prompts or commands
- If you used some prompts or commands it will show up on top of the list on the welcome screen 

This is a crucial improvement for customers with a possible high count of prompts because it helps with discoverability and we don't bombard the welcome screen with a big set of items in the suggestion list

## Future follow-ups
- In designs, we have a Search Prompt button below the action list on the welcome screen, for simplicity now it's just a link to the prompt tab but in future, we should open in place modal UI with a search prompt UI 
The last used action is currently implemented with local storage, later we should track usage on the backend  

## Test plan
- Check that you can see only 4 items max on the welcome screen 
- Check that you can find prompts that you have on prompts tab or prompts toolbar 
- Check that last used prompts/commands can be found on the welcome screen right after you used them

### DIFF
--- a/lib/shared/src/misc/rpc/webviewAPI.ts
+++ b/lib/shared/src/misc/rpc/webviewAPI.ts
@@ -114,27 +114,42 @@ export interface MentionMenuData {
     error?: string
 }
 
+export interface PromptAction extends Prompt {
+    actionType: 'prompt'
+}
+
+export interface CommandAction extends CodyCommand {
+    actionType: 'command'
+}
+
+export type Action = PromptAction | CommandAction
+
 export interface PromptsResult {
-    /**
-     * `undefined` means the Sourcegraph endpoint is an older Sourcegraph version that doesn't
-     * support the Prompt Library.
-     */
-    prompts:
-        | { type: 'results'; results: Prompt[] }
-        | { type: 'error'; error: string }
-        | { type: 'unsupported' }
+    // /**
+    //  * `undefined` means the Sourcegraph endpoint is an older Sourcegraph version that doesn't
+    //  * support the Prompt Library.
+    //  */
+    // prompts:
+    //     | { type: 'results'; results: Prompt[] }
+    //     | { type: 'error'; error: string }
+    //     | { type: 'unsupported' }
+    //
+    // /**
+    //  * Provides previously built-in commands which became prompt-like actions (explain code,
+    //  * generate unit tests, document symbol, etc.) Currently, is used behind feature flag.
+    //  */
+    // standardPrompts?: CodyCommand[]
+    //
+    // /**
+    //  * `undefined` means that commands should not be shown at all (not even as an empty
+    //  * list). Builtin and custom commands are deprecated in favor of the Prompt Library.
+    //  */
+    // commands: CodyCommand[]
 
-    /**
-     * Provides previously built-in commands which became prompt-like actions (explain code,
-     * generate unit tests, document symbol, etc.) Currently, is used behind feature flag.
-     */
-    standardPrompts?: CodyCommand[]
+    isPromptsSupported: boolean
 
-    /**
-     * `undefined` means that commands should not be shown at all (not even as an empty
-     * list). Builtin and custom commands are deprecated in favor of the Prompt Library.
-     */
-    commands: CodyCommand[]
+    /** List of all available actions (prompts and/or commands) */
+    actions: Action[]
 
     /** The original query used to fetch this result. */
     query: string
--- a/lib/shared/src/misc/rpc/webviewAPI.ts
+++ b/lib/shared/src/misc/rpc/webviewAPI.ts
@@ -125,28 +125,8 @@ export interface CommandAction extends CodyCommand {
 export type Action = PromptAction | CommandAction
 
 export interface PromptsResult {
-    // /**
-    //  * `undefined` means the Sourcegraph endpoint is an older Sourcegraph version that doesn't
-    //  * support the Prompt Library.
-    //  */
-    // prompts:
-    //     | { type: 'results'; results: Prompt[] }
-    //     | { type: 'error'; error: string }
-    //     | { type: 'unsupported' }
-    //
-    // /**
-    //  * Provides previously built-in commands which became prompt-like actions (explain code,
-    //  * generate unit tests, document symbol, etc.) Currently, is used behind feature flag.
-    //  */
-    // standardPrompts?: CodyCommand[]
-    //
-    // /**
-    //  * `undefined` means that commands should not be shown at all (not even as an empty
-    //  * list). Builtin and custom commands are deprecated in favor of the Prompt Library.
-    //  */
-    // commands: CodyCommand[]
 
-    isPromptsSupported: boolean
+    arePromptsSupported: boolean
 
     /** List of all available actions (prompts and/or commands) */
     actions: Action[]
--- a/lib/shared/src/misc/rpc/webviewAPI.ts
+++ b/lib/shared/src/misc/rpc/webviewAPI.ts
@@ -125,7 +125,6 @@ export interface CommandAction extends CodyCommand {
 export type Action = PromptAction | CommandAction
 
 export interface PromptsResult {
-
     arePromptsSupported: boolean
 
     /** List of all available actions (prompts and/or commands) */
--- a/vscode/src/prompts/prompts.ts
+++ b/vscode/src/prompts/prompts.ts
@@ -55,7 +55,7 @@ const STANDARD_PROMPTS_LIKE_COMMAND: CommandAction[] = [
 ]
 
 /**
- * Observe results of querying the prompts from the Prompt Library, (deprecated) built-in commands,
+ * Merges results  of querying the prompts from the Prompt Library, (deprecated) built-in commands,
  * and (deprecated) custom commands. Commands are deprecated in favor of prompts in the Prompt
  * Library.
  */
@@ -79,7 +79,7 @@ export async function mergedPromptsAndLegacyCommands(
                   ...c,
                   actionType: 'command',
               }))
-        : // Ignore any commands for Cody Web since none commands are supported there
+        : // Ignore any commands for Cody Web since no commands are supported
           []
 
     const matchingCommands = allCommands.filter(
@@ -90,12 +90,12 @@ export async function mergedPromptsAndLegacyCommands(
     )
 
     const actions =
-        customPrompts !== 'unsupported' ? [...customPrompts, ...matchingCommands] : matchingCommands
+        customPrompts === 'unsupported' ? matchingCommands : [...customPrompts, ...matchingCommands]
 
     return {
         query,
         actions,
-        isPromptsSupported: customPrompts !== 'unsupported',
+        arePromptsSupported: customPrompts !== 'unsupported',
     }
 }
 
--- a/vscode/src/prompts/prompts.ts
+++ b/vscode/src/prompts/prompts.ts
@@ -1,6 +1,7 @@
 import {
-    type CodyCommand,
+    type CommandAction,
     FeatureFlag,
+    type PromptAction,
     type PromptsResult,
     clientCapabilities,
     featureFlagProvider,
@@ -11,6 +12,48 @@ import {
 import { FIXTURE_COMMANDS } from '../../webviews/components/promptList/fixtures'
 import { getCodyCommandList } from '../commands/CommandsController'
 
+/** For testing only. */
+const USE_CUSTOM_COMMANDS_FIXTURE = false
+
+const STANDARD_PROMPTS_LIKE_COMMAND: CommandAction[] = [
+    {
+        key: 'doc',
+        type: 'default',
+        mode: 'ask',
+        description: 'Document Code',
+        slashCommand: 'cody.command.prompt-document-code',
+        prompt: '',
+        actionType: 'command',
+    },
+    {
+        key: 'explain',
+        type: 'default',
+        mode: 'ask',
+        description: 'Explain Code',
+        slashCommand: 'cody.command.explain-code',
+        prompt: '',
+        actionType: 'command',
+    },
+    {
+        key: 'test',
+        type: 'default',
+        mode: 'ask',
+        description: 'Generate Unit Tests',
+        slashCommand: 'cody.command.unit-tests',
+        prompt: '',
+        actionType: 'command',
+    },
+    {
+        key: 'smell',
+        type: 'default',
+        mode: 'ask',
+        description: 'Find Code Smells',
+        slashCommand: 'cody.command.smell-code',
+        prompt: '',
+        actionType: 'command',
+    },
+]
+
 /**
  * Observe results of querying the prompts from the Prompt Library, (deprecated) built-in commands,
  * and (deprecated) custom commands. Commands are deprecated in favor of prompts in the Prompt
@@ -18,78 +61,26 @@ import { getCodyCommandList } from '../commands/CommandsController'
  */
 export async function mergedPromptsAndLegacyCommands(
     query: string,
-    signal?: AbortSignal
+    signal: AbortSignal
 ): Promise<PromptsResult> {
-    let promptsValue: PromptsResult['prompts']
-    try {
-        const prompts = await graphqlClient.queryPrompts(query, signal)
-        promptsValue = { type: 'results', results: prompts }
-    } catch (error) {
-        if (isAbortError(error)) {
-            throw error
-        }
-        const errorMessage = isErrorLike(error) ? error.message : String(error)
-        if (errorMessage.startsWith(`Cannot query field "prompts"`)) {
-            // Server does not yet support Prompt Library.
-            promptsValue = { type: 'unsupported' }
-        } else {
-            promptsValue = { type: 'error', error: errorMessage }
-        }
-    }
-
     const queryLower = query.toLowerCase()
-    const isUnifiedPromptsEnabled = await featureFlagProvider.evaluateFeatureFlagEphemerally(
-        FeatureFlag.CodyUnifiedPrompts
-    )
+    const [customPrompts, isUnifiedPromptsEnabled] = await Promise.all([
+        fetchCustomPrompts(queryLower, signal),
+        featureFlagProvider.evaluateFeatureFlagEphemerally(FeatureFlag.CodyUnifiedPrompts),
+    ])
 
-    // Ignore commands since with unified prompts vital commands will be replaced by out-of-box
-    // commands, see main.ts register cody commands for unified prompts
-    if (isUnifiedPromptsEnabled && !clientCapabilities().isCodyWeb) {
-        return {
-            query,
-            commands: [],
-            prompts: promptsValue,
-            standardPrompts: [
-                {
-                    key: 'doc',
-                    description: 'Document Code',
-                    prompt: '',
-                    slashCommand: 'cody.command.prompt-document-code',
-                    mode: 'ask',
-                    type: 'default',
-                },
-                {
-                    key: 'explain',
-                    description: 'Explain Code',
-                    prompt: '',
-                    slashCommand: 'cody.command.explain-code',
-                    mode: 'ask',
-                    type: 'default',
-                },
-                {
-                    key: 'test',
-                    description: 'Generate Unit Tests',
-                    prompt: '',
-                    slashCommand: 'cody.command.unit-tests',
-                    mode: 'ask',
-                    type: 'default',
-                },
-                {
-                    key: 'smell',
-                    description: 'Find Code Smells',
-                    slashCommand: 'cody.command.smell-code',
-                    prompt: '',
-                    mode: 'ask',
-                    type: 'default',
-                },
-            ] satisfies CodyCommand[],
-        }
-    }
-
-    const allCommands = [
-        ...getCodyCommandList(),
-        ...(USE_CUSTOM_COMMANDS_FIXTURE ? FIXTURE_COMMANDS : []),
-    ].filter(command => (isUnifiedPromptsEnabled ? { ...command, mode: 'ask' } : command))
+    const codyCommands = getCodyCommandList()
+    const allCommands: CommandAction[] = !clientCapabilities().isCodyWeb
+        ? // Ignore commands since with unified prompts vital commands will be replaced by out-of-box
+          // prompts, see main.ts register cody commands for unified prompts
+          isUnifiedPromptsEnabled
+            ? STANDARD_PROMPTS_LIKE_COMMAND
+            : [...codyCommands, ...(USE_CUSTOM_COMMANDS_FIXTURE ? FIXTURE_COMMANDS : [])].map(c => ({
+                  ...c,
+                  actionType: 'command',
+              }))
+        : // Ignore any commands for Cody Web since none commands are supported there
+          []
 
     const matchingCommands = allCommands.filter(
         c =>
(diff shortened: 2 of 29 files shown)


lib/shared/src/misc/rpc/webviewAPI.ts
```diff
@@ -114,27 +114,42 @@ export interface MentionMenuData {
...
+    //  */
+    // standardPrompts?: CodyCommand[]
+    //
+    // /**
+    //  * `undefined` means that commands should not be shown at all (not even as an empty
+    //  * list). Builtin and custom commands are deprecated in favor of the Prompt Library.
+    //  */
+    // commands: CodyCommand[]
+
+    isPromptsSupported: boolean
```
dominiccooney, line 149: This is an awkward name because "is" is singular but "prompts" is plural. Consider `supportsPrompts`, or `arePromptsSupported`, or if we really want "is" then `isPromptSupporting` I guess...

vscode/src/prompts/prompts.ts
```diff
@@ -11,85 +12,75 @@ import {
 import { FIXTURE_COMMANDS } from '../../webviews/components/promptList/fixtures'
 import { getCodyCommandList } from '../commands/CommandsController'
 
+/** For testing only. */
```
dominiccooney, line 15: Is this for *testing* per se, or development?

vscode/src/prompts/prompts.ts
```diff
@@ -11,85 +12,75 @@ import {
...
+        mode: 'ask',
+        description: 'Find Code Smells',
+        slashCommand: 'cody.command.smell-code',
+        prompt: '',
+        actionType: 'command',
+    },
+]
+
 /**
  * Observe results of querying the prompts from the Prompt Library, (deprecated) built-in commands,
```
dominiccooney, line 58: Given the recent moves to observables, "observe" isn't a great verb choice. How about "Merges results of ..."

vscode/src/prompts/prompts.ts
```diff
@@ -11,85 +12,75 @@ import {
...
-        }
-    }
-
     const queryLower = query.toLowerCase()
-    const isUnifiedPromptsEnabled = await featureFlagProvider.evaluateFeatureFlagEphemerally(
-        FeatureFlag.CodyUnifiedPrompts
-    )
+    const [customPrompts, isUnifiedPromptsEnabled] = await Promise.all([
+        fetchCustomPrompts(queryLower, signal),
+        featureFlagProvider.evaluateFeatureFlagEphemerally(FeatureFlag.CodyUnifiedPrompts),
```
dominiccooney, line 69: Should this in fact *be* an observable and be polling the custom prompt fetch and observing the feature flag?

vscode/src/prompts/prompts.ts
```diff
@@ -11,85 +12,75 @@ import {
...
-        }
-    }
-
-    const allCommands = [
-        ...getCodyCommandList(),
-        ...(USE_CUSTOM_COMMANDS_FIXTURE ? FIXTURE_COMMANDS : []),
-    ].filter(command => (isUnifiedPromptsEnabled ? { ...command, mode: 'ask' } : command))
+    const codyCommands = getCodyCommandList()
+    const allCommands: CommandAction[] = !clientCapabilities().isCodyWeb
+        ? // Ignore commands since with unified prompts vital commands will be replaced by out-of-box
```
dominiccooney, line 74: Word choice, "vital" -> "essential"

Something weird with "unified prompts" too, is that possessive like "unified prompts' essential commands"?

vscode/src/prompts/prompts.ts
```diff
@@ -11,85 +12,75 @@ import {
...
+          // prompts, see main.ts register cody commands for unified prompts
+          isUnifiedPromptsEnabled
+            ? STANDARD_PROMPTS_LIKE_COMMAND
+            : [...codyCommands, ...(USE_CUSTOM_COMMANDS_FIXTURE ? FIXTURE_COMMANDS : [])].map(c => ({
+                  ...c,
+                  actionType: 'command',
+              }))
+        : // Ignore any commands for Cody Web since none commands are supported there
```
dominiccooney, line 82: "none commands are supported there" -> "no commands are supported there", or, "commands are not supported there"

vscode/src/prompts/prompts.ts
```diff
@@ -98,20 +89,51 @@ export async function mergedPromptsAndLegacyCommands(
             matchesQuery(queryLower, c.prompt)
     )
 
+    const actions =
+        customPrompts !== 'unsupported' ? [...customPrompts, ...matchingCommands] : matchingCommands
```
dominiccooney, line 93: Simplify `a !== b ? c : d` to `a === b ? d : c`.

Consider eliminating the stuttering of `matchingCommands` by floating the ternary inside the splice with `[...(x ? y : []), ...z]`

taiyab:
This is looking great! Good job @vovakulikov and @jasonhawkharris.

- Chat help should use the updated design from Jason's PR.
- Prompt/command labels should be slightly less bold
- We can knock down the icon sizes for prompts/commands down a bit (if they're 20x20, make them 16x16)
- If we're able to do some sort of sorting by last used, we should use that on the dropdown list from the prompts menu, and not the welcome area prompts (which should be sorted by last created by org (and published, not drafted)), if possible.
- I should not be able to see other peoples' drafted prompts anywhere, at all
- I should be able to see my drafted prompts in the library and last used dropdown only
- Edit Code needs to always exist on the upfront welcome area

vscode/src/prompts/prompts.ts
```diff
@@ -98,20 +89,51 @@ export async function mergedPromptsAndLegacyCommands(
...
+    try {
+        const prompts = await graphqlClient.queryPrompts(query, signal)
+        return prompts.map(prompt => ({ ...prompt, actionType: 'prompt' }))
+    } catch (error) {
+        if (isAbortError(error)) {
+            throw error
+        }
+
+        const errorMessage = isErrorLike(error) ? error.message : String(error)
+        if (errorMessage.startsWith(`Cannot query field "prompts"`)) {
```
dominiccooney, line 132: FYI when adding context filters we found some old Sourcegraph instances returned an empty string in some error cases. If this 'unsupported' state is important, you may want to test against old instances.

vscode/webviews/components/hooks.ts
```diff
@@ -0,0 +1,22 @@
+import { type Dispatch, type SetStateAction, useCallback, useState } from 'react'
+
+export function useLocalStorage<T>(
+    key: string,
+    defaultValue?: T
```
dominiccooney, line 5: Would be neat if there was a way callers providing a default were not punished with the `T | undefined` type.

vscode/webviews/components/promptList/ActionItem.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    <strong className={styles.promptName}>{prompt.name}</strong>
+                    {prompt.draft && (
+                        <Badge variant="secondary" className="tw-text-xxs tw-mt-0.5">
+                            Draft
+                        </Badge>
+                    )}
+                </div>
+
+                <span className={styles.promptDescription}>
+                    {prompt.description ?? 'No description provided'}
```
dominiccooney, line 69: Personally I dislike defaults like this because it is so much visual junk if people aren't providing descriptions... But if this is the design, so be it.

vscode/webviews/components/promptList/PromptList.module.css
```diff
@@ -0,0 +1,40 @@
+
+.list {
+    outline: none;
+    background-color: transparent !important;
```
dominiccooney, line 4: What's going on here that we need to reach for `!important`?

vscode/webviews/components/promptList/PromptList.module.css
```diff
@@ -0,0 +1,40 @@
...
+
+    &--input {
+        margin: 0.5rem;
+        padding: 0.5rem;
+        border-radius: 3px;
+        background-color: var(--vscode-input-background);
+        color: var(--vscode-input-foreground);
+
+        &:focus {
+            //border-color: var(--vscode-focusBorder);
```
dominiccooney, line 15: I guess //-style comments are a feature of our CSS processor or does this mean something? If not let's remove it.

vscode/webviews/components/promptList/PromptList.tsx
```diff
@@ -66,68 +69,50 @@ export const PromptList: React.FunctionComponent<{
...
-                standardPromptCommand ??
-                result?.commands?.find(c => commandRowValue({ type: 'command', value: c }) === rowValue)
-
-            const entry: PromptOrDeprecatedCommand | undefined = prompt
-                ? { type: 'prompt', value: prompt }
-                : codyCommand
-                  ? { type: 'command', value: codyCommand }
-                  : undefined
-            if (!entry) {
+            if (!action || !result) {
```
dominiccooney, line 74: Simplify `!a || !b` -> `!(a && b)` which is one less operator and more emphatic that both are truthy.

vscode/webviews/components/promptList/PromptList.tsx
```diff
@@ -66,68 +69,50 @@ export const PromptList: React.FunctionComponent<{
...
+            const isCommand = action.actionType === 'command'
+            const isBuiltInCommand = isCommand && action.type === 'default'
+
             telemetryRecorder.recordEvent('cody.promptList', 'select', {
                 metadata: {
-                    isPrompt: prompt ? 1 : 0,
-                    isCommand: codyCommand ? 1 : 0,
-                    isCommandBuiltin: codyCommand?.type === 'default' ? 1 : 0,
-                    isCommandCustom: codyCommand?.type !== 'default' ? 1 : 0,
+                    isPrompt: isPrompt ? 1 : 0,
```
dominiccooney, line 84: Consider `Number(isPrompt)` unless tsc yells at you.

vscode/webviews/components/promptList/PromptList.tsx
```diff
@@ -66,68 +69,50 @@ export const PromptList: React.FunctionComponent<{
...
+                    isCommand: isCommand ? 1 : 0,
+                    isCommandBuiltin: isBuiltInCommand ? 1 : 0,
+                    isCommandCustom: !isBuiltInCommand ? 1 : 0,
                     ...telemetryPublicMetadata,
                 },
                 privateMetadata: {
-                    nameWithOwner: prompt ? prompt.nameWithOwner : undefined,
+                    nameWithOwner: isPrompt ? action.nameWithOwner : undefined,
```
dominiccooney, line 91: Would it be better to omit the `nameWithOwner` field entirely if it is not a prompt?

vscode/webviews/components/promptList/PromptList.tsx
```diff
@@ -139,166 +124,83 @@ export const PromptList: React.FunctionComponent<{
...
-        : result?.commands
+    const allActions = showOnlyPromptInsertableCommands
+        ? result?.actions.filter(action => action.actionType === 'prompt' || action.mode === 'ask') ?? []
+        : result?.actions ?? []
+
+    const sortedActions = getSortedActions(allActions, lastUsedActions)
+    const actions = showFirstNItems ? sortedActions.slice(0, showFirstNItems) : sortedActions
+
+    const inputPaddingClass =
+        paddingLevels !== 'none' ? (paddingLevels === 'middle' ? '!tw-p-2' : '!tw-px-4 !tw-py-4') : ''
```
dominiccooney, line 136: As mentioned above, I'd use === and reverse the conditions so the reader doesn't have to think through negations. Ditto below.

How much does the important, and burying spacing in the code here, hurt us later? Can we make our own tailwind class for these and apply something logical here instead?

Why do you `!tw-px-4 !tw-py-4` and not simply `!tw-p-4`?

vscode/webviews/components/promptList/PromptList.tsx
```diff
@@ -309,123 +211,13 @@ export const PromptList: React.FunctionComponent<{
...
 
-function hasCustomCommands(commands: CodyCommand[]): boolean {
-    return commands.some(
-        command =>
-            command.type === CustomCommandType.Workspace || command.type === CustomCommandType.User
-    )
-}
+function getSortedActions(actions: Action[], lastUsedActions: Record<string, number>): Action[] {
+    return [...actions].sort((action1, action2) => {
+        const action1Key = action1.actionType === 'prompt' ? action1.id : action1.key
```
dominiccooney, line 216: Is there a lodash sortBy or something we could use here to cut down the duplication?

vscode/webviews/components/promptList/fixtures.ts
```diff
@@ -72,15 +77,17 @@ export function makePromptsAPIWithData(
...
-                prompts:
-                    data.prompts.type === 'results'
-                        ? {
-                              type: 'results',
-                              results: data.prompts.results.filter(prompt => matchQuery(prompt.name)),
-                          }
-                        : data.prompts,
-                commands: data.commands?.filter(c => matchQuery(c.key)),
                 query,
+                isPromptsSupported: isPromptsSupported,
```
dominiccooney, line 81: Just `isPromptsSupported` when the property name and variable name line up.

vscode/webviews/components/promptList/utils.ts
```diff
@@ -0,0 +1,5 @@
+import type { Action } from '@sourcegraph/cody-shared'
+
+export function commandRowValue(row: Action): string {
+    return row.actionType === 'prompt' ? `prompt-${row.id}` : `command-${row.key}`
```
dominiccooney, line 4: There's a few places where you do this switch on `actionType` and consult `id` or `key`. Why not write a helper function and remove all the conditionals from those places?

vscode/webviews/prompts/PromptsTab.tsx
```diff
@@ -1,73 +1,76 @@
...
+export function useActionSelect() {
+    const dispatchClientAction = useClientActionDispatcher()
+    const [lastUsedActions = {}, persistValue] = useLocalStorage<Record<string, number>>(
+        'last-used-actions',
+        {}
+    )
+
+    return (action: Action, setView: (view: View) => void) => {
+        try {
+            const actionKey = action.actionType === 'prompt' ? action.id : action.key
```
dominiccooney, line 37: ...places which could call that helper and be more readable++

dominiccooney (approved):
Looks good; many comments to simplify the code inline. I have generally commented the first time something occurs.

I am surprised/alarmed by the amount of `!` in the CSS, how does maintenance go over time with that?

vovakulikov:
> Chat help should use the updated design from Jason's PR.

Not a part of this PR will do as follow-up, or @jasonhawkharris will merge it as part of his PR

> Prompt/command labels should be slightly less bold, even tone of the color a bit like from Jason's PR

Set it to 500 now, let me know if it look ok.

> We can knock down the icon/avatar sizes for prompts/commands down a bit (if they're 20x20, make them 16x16)

16px seemed a bit off IMO. I reduced size down to 20 and 22 on the icon, should be ok now, also let me know

> If we're able to do some sort of sorting by last used, we should use that on the dropdown list from the prompts menu, and not the welcome area prompts (which should be sorted by last created by org (and published, not drafted)), if possible.

Ok, I turned off smart sorting on the prompt tab, but the welcome page seems like a good spot to have this functionality. This is based on what @rrhyne highly suggested last week and I agree with him here, any concerns about welcome page with sorting on? 

> I should not be able to see other peoples' drafted prompts anywhere, at all (@aramaraju only exception to this might be an admin?)

The problem here is that there is no API to turn off drafts from other users and keep your created drafts. I included drafts from everyone for now, but would be happy to turn off drafts completely if it's better (while we don't have behaviour about viewing only your drafts) 

> Edit Code needs to always exist on the upfront welcome area

Fixed, wonder if we need to show 5 items on the welcome area, because edit command will always be taking one spot from the list

> We can change the "Create Prompt' and "Prompt Library" links on the Prompts tab submenu to: "Create" and "Manage".

Would stick for now with prompt suffix, because we still have commands in place, Create and manage are too generic at the moment

> The "Find Code Smells" prompt should use the text-search icon from Lucide 

Fixed

vscode/src/prompts/prompts.ts
vovakulikov: I actually have no idea, this code was here before me from the very beginning of commands/prompts functionality

vscode/src/prompts/prompts.ts
vovakulikov: True, this is, in fact, a problem when you switch accounts, but your list of prompts is still the same until you refresh the prompt tab. 

I don't want to fix it in this PR because I expect some other thing may be broken. But good follow up

vscode/src/prompts/prompts.ts
vovakulikov: I don't think this is very important; worst case scenario, we don't show warning text that you're on an old instance. I checked it with 5.5 locally and it worked, but yeah I've seen before that errors sometimes are intercepted by something else on our backend

vscode/webviews/components/hooks.ts
vovakulikov: I have as I think (hehe maybe not) a better implementation of this, will bring it in follow up PR.

vscode/webviews/components/promptList/ActionItem.tsx
vovakulikov: cc @taiyab it's your call

vscode/webviews/components/promptList/PromptList.module.css
vovakulikov: Just missed this, removed

vscode/webviews/components/promptList/PromptList.module.css
vovakulikov: Just a left over, thanks for spotting

vscode/webviews/components/promptList/PromptList.tsx
vovakulikov: disagree about this one, maybe it's a habit but I found that negative and or operator is better whey you try to unwrap this, you basically read it from left to right when with `!(a && b)` you have to keep in mind `!` operator when you read a && b

vscode/webviews/components/promptList/PromptList.tsx
vovakulikov: > Why do you !tw-px-4 !tw-py-4 and not simply !tw-p-4?

No reasons just a mistake, fixed it here (originally I think it was `!tw-px-4 !tw-py-6`

> How much does the important, and burying spacing in the code here, hurt us later? Can we make our own tailwind class for these and apply something logical here instead?

Yeah I will re-write this anyway this week, so will skip this comment for now, later it will be just a specific css class, yeah

vscode/webviews/components/promptList/PromptList.tsx
vovakulikov: different keys on union type makes lodash `sortBy` a bit more tricky when I want it to be.

taiyab:
> > Chat help should use the updated design from Jason's PR.
> 
> Not a part of this PR will do as follow-up, or @jasonhawkharris will merge it as part of his PR
> 
Cool.

> > Prompt/command labels should be slightly less bold, even tone of the color a bit like from Jason's PR
> 
> Set it to 500 now, let me know if it look ok.
> 

Looks much better 👍 

> > We can knock down the icon/avatar sizes for prompts/commands down a bit (if they're 20x20, make them 16x16)
> 
> 16px seemed a bit off IMO. I reduced size down to 20 and 22 on the icon, should be ok now, also let me know
> 

Feel free to knock down the avatars too. They should all be the same size and aligned.

> > If we're able to do some sort of sorting by last used, we should use that on the dropdown list from the prompts menu, and not the welcome area prompts (which should be sorted by last created by org (and published, not drafted)), if possible.
> 
> Ok, I turned off smart sorting on the prompt tab, but the welcome page seems like a good spot to have this functionality. This is based on what @rrhyne highly suggested last week and I agree with him here, any concerns about welcome page with sorting on?
> 

If we can only have ordered by recently used in the dropdown, that would be ideal. Welcome area needs to be reserved for discoverability of new prompts (not accessing ones that you're already using).

> > I should not be able to see other peoples' drafted prompts anywhere, at all (@aramaraju only exception to this might be an admin?)
> 
> The problem here is that there is no API to turn off drafts from other users and keep your created drafts. I included drafts from everyone for now, but would be happy to turn off drafts completely if it's better (while we don't have behaviour about viewing only your drafts)
> 

I'm happy to switch them all off in terms of visibility for now -- I assume we can keep them visible on the web to compensate, and so I can access drafted prompts whilst I'm drafting them via the web?

> > Edit Code needs to always exist on the upfront welcome area
> 
> Fixed, wonder if we need to show 5 items on the welcome area, because edit command will always be taking one spot from the list
> 

That would be OK for now. FYI, I'm not seeing Edit Code option when checking out your branch locally and authed to S2.

> > We can change the "Create Prompt' and "Prompt Library" links on the Prompts tab submenu to: "Create" and "Manage".
> 
> Would stick for now with prompt suffix, because we still have commands in place, Create and manage are too generic at the moment
> 

That's fine.

> > The "Find Code Smells" prompt should use the text-search icon from Lucide
> 
> Fixed

Sweet.

taiyab:
cc @danielmarquespt @taylorsperry for visibility

vscode/src/prompts/prompts.ts
dominiccooney: Consider putting a TODO to that effect, and linking to a Linear issue, so we have something to de-dup the QA bug reports against.

vscode/src/prompts/prompts.ts
dominiccooney: If it is not very important, do we need this code at all? Put a TODO on it to remove it when the support window for 5.5 closes?

vscode/webviews/components/promptList/PromptList.tsx
dominiccooney: You could just `...isPrompt ? { nameWithOwner: ... } : {}` or something.

Really depends what the Telemetry consumer wants to see. Just seems weird to log the unrelated property with `undefined`.

vscode/webviews/components/promptList/PromptList.tsx
dominiccooney: Isn't there a `sortBy` that takes a function that "projects" the value you want to sort on?

taiyab:
We can change the "Prompt Library" link at the bottom of the welcome area list to "All prompts"

vscode/webviews/tabs/TabsBar.tsx
```diff
@@ -285,15 +291,20 @@ const TabButton = forwardRef<HTMLButtonElement, TabButtonProps>((props, ref) =>
         'data-testid': dataTestId,
     } = props
 
+    const Component = uri ? 'a' : 'button'
```
thenamankumar, line 294: better just render ahref and button separately with only 1 `uri ?` condition at top. Rather than merging them into a single component and then have multiple `uri?` in props.

thenamankumar (approved):

vscode/webviews/components/promptList/PromptList.tsx
vovakulikov: Yeah, but in this case if we use function it's not any different from native sort implementation like here

PR #5778, opened 2024-10-02
Reduce padding of container around search input.

taiyab:
Reduces the padding of the container outside the prompt search to make it slightly more dense.

CleanShot 2024-10-02 at 16 46 06@2x

## Test Plan

Tested manually.


thenamankumar (approved):

vovakulikov (commented):
@taiyab I think this PR should change only paddings for prompts popover layout and shouldn't change paddings on the prompt list in the prompts tab. I will fix it here https://github.com/sourcegraph/cody/pull/6056

PR #5784, opened 2024-10-03
Autocomplete: re-enable agent tests

valerybugakov:
- No functional changes. 
- 90% of the diff are recordings and bindings updates 🙂
- Re-enabled existing autocomplete integration tests and improved the stability of recoding updates. Initially, they were disabled here: https://github.com/sourcegraph/cody/pull/5629
- Replaced most of the `sleep` calls with new agent `testing/X` requests to await for a timeout in question ([Slack thread](https://sourcegraph.slack.com/archives/C0729T2PBV2/p1727834492679739)).
- Removed the `NoopCompletionItemProvider` registration for unauthenticated users to fix the flaky behavior where the real autocomplete provider was not registered before the first `autocomplete/execute` call in tests ([Slack thread](https://sourcegraph.slack.com/archives/C05AGQYD528/p1727836036656949)). Autocomplete is currently available only for authenticated requests, so this change should not have any functional side effects.
- Updated Kotlin bindings (some are not related to this PR, but AFAIK, this is caused by the main branch not being updated).
- Part of https://linear.app/sourcegraph/issue/CODY-3910/add-autocomplete-integation-tests-for-server-side-model-config

## Test plan

CI and  `pnpm vitest agent/src/autocomplete.test.ts`


agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
```diff
@@ -17,5 +47,11 @@ data class AuthStatus(
   val displayName: String? = null,
   val avatarURL: String? = null,
   val userCanUpgrade: Boolean? = null,
-)
+  val pendingValidation: Boolean,
+) : AuthStatus() {
+
+  enum class EndpointEnum {
```
sqs, line 53: These changes are bad and are why the bindings codegen is disabled (https://linear.app/sourcegraph/issue/CODY-3809/reenable-kotlin-bindings-in-ci). It doesn't handle something about the union discriminator. You should keep the codegen disabled in CI and revert the changes to this file (but the others look good).

sqs (approved):
Very nice, great to get these enabled again. Thanks!

PR #5785, opened 2024-10-03
never minify agent

sqs:
The benefits (smaller bundle size) are not worth the additional debugging pain, esp. since this build was only ever used for desktop releases and not web.

## Test plan

n/a


varungandhi-src (approved):

evict (approved):

vovakulikov:
Just FYI in the web for example we minify it with Sourcegraph repository bundler (esbuild in react version and with vite in the svelte version) 

There is some problem with the minification Cody Web agent and rpc lib (so in the react version we don't minify the web-worker part), In svelte we use a non-default minifyer  which doesn't corrupt web-worker part.

Issue #5783, opened 2024-10-03
bug: I am being repeatedly logged out.


warm--tape:
Type: <b>Bug</b>

I am logged in. Often when I reopen a new vscode window, I need to reauthenticate with Cody. This has only started happening today.

## Extension Information

- Cody Version: 1.36.0
- VS Code Version: 1.93.1
- Extension Host: desktop

##  Steps to Reproduce
1. Log in
2. Working, opening different windows etc
3. Cody logs me out

## Expected Behaviour
Stay logged in

## Logs
█ telemetry-v2: recordEvent: cody.extension/savedLogin  {
  "parameters": {
    "version": 0,
    "metadata": []
  },
  "timestamp": "2024-10-03T00:26:06.592Z"
}
█ ModelsService: User model preferences changed {"defaults":{"chat":"anthropic::2023-06-01::claude-3.5-sonnet","edit":"anthropic::2023-06-01::claude-3.5-sonnet","autocomplete":"fireworks::v1::deepseek-coder-v2-lite-base"},"selected":{}}
█ ChatsController:constructor: init
█ AutocompleteProvider:notSignedIn: You are not signed in.
█ GraphQLTelemetryExporter: telemetry: failed to evaluate server version: Error: ETIMEDOUT: Request timed out after 6000ms (https://sourcegraph.com/.api/graphql?SiteProductVersion)
█ CodyLLMConfiguration: {"siteHasCodyEnabled":false,"siteVersion":"unknown","codyLLMConfiguration":{},"userInfo":{}}
█ telemetry-v2: recordEvent: cody.auth/failed  {
  "parameters": {
    "version": 0,
    "metadata": [
      {
        "key": "tier"
      }
    ],
    "billingMetadata": {
      "product": "cody",
      "category": "billable"
    }
  },
  "timestamp": "2024-10-03T00:26:15.363Z"
}
█ AutocompleteProvider:notSignedIn: You are not signed in.
█ GraphQLTelemetryExporter: Error exporting telemetry events: Error: ETIMEDOUT: Request timed out after 6000ms (https://sourcegraph.com/.api/graphql?RecordTelemetryEvents)
█ GraphQLTelemetryExporter: telemetry: failed to evaluate server version: Error: ETIMEDOUT: Request timed out after 6000ms (https://sourcegraph.com/.api/graphql?SiteProductVersion)
█ GraphQLTelemetryExporter: Error exporting telemetry events: Error: Variable "value" has invalid value null.
[...]

PR #5789, opened 2024-10-03
higher default timeout than 6s

sqs:
The 6000ms timeout may be too short for some users on high-latency network connections. I have experienced this as a problem once or twice recently when on airplane (satellite) wifi. This may also help with https://github.com/sourcegraph/cody/issues/5783, but I am not certain of that since this timeout has been in place for a while and the reports of that issue are new.

Why 20s and not higher/lower? This makes it so the user does get feedback about network problems without waiting 60s (default) or longer. Why not lower (like 15s)? Arbitrary decision.

## Test plan

n/a

### DIFF
--- a/lib/shared/src/sourcegraph-api/graphql/client.ts
+++ b/lib/shared/src/sourcegraph-api/graphql/client.ts
@@ -581,6 +581,8 @@ export function setUserAgent(newUseragent: string): void {
 
 const QUERY_TO_NAME_REGEXP = /^\s*(?:query|mutation)\s+(\w+)/m
 
+const DEFAULT_TIMEOUT_MSEC = 20000
+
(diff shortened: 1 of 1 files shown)


lib/shared/src/sourcegraph-api/graphql/client.ts
```diff
@@ -1427,7 +1429,7 @@ export class SourcegraphGraphQLAPIClient {
         })
 
         // Default timeout of 6 seconds.
```
dominiccooney, line 1431: We could move this comment to the variable name and update it, or delete it ... the variable name is clear.
```suggestion
```

dominiccooney (approved):
Looks good, assuming there aren't callers passing in the `signalOrTimeout`.

PR #5792, opened 2024-10-03
remove ChatSubmitType user-newchat

sqs:
This controlled whether a chat would reuse the same window or create a new one. Replies and edits reuse the same chat window, but everything else (except for custom commands, optionally) starts a new chat.

The only user-facing behavior change is for custom commands (described below).

The `user` `ChatSubmitType` was used in 2 places:

- For edits in ChatController (not needed, can refactor more simply internally)
- For custom commands

Since custom commands are being removed and the "reuse existing chat window" of custom commands is not important to preserve even for now, we can simplify the code by removing this mode and `submitType` throughout.

## Test plan

CI & e2e

### DIFF
--- a/agent/src/TestClient.ts
+++ b/agent/src/TestClient.ts
@@ -805,7 +805,7 @@ export class TestClient extends MessageHandler {
                 message: {
                     command: 'submit',
                     text,
-                    submitType: 'user',
+                    submitType: 'user-newchat',
                     contextItems: params?.contextFiles,
                 },
             })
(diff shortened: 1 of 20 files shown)


abeatrix (approved):
LGTM.

Q: YK has a PR that I forgot to merge for him before his PTO. This looks like we can still do the behavior Yk is expecting in his PR with this change?
https://github.com/sourcegraph/cody/pull/5698

sqs:
@abeatrix Thanks! For that PR, it looks like your suggestion is to use `user-newchat`, which will work because that is now the *only* behavior.

PR #5793, opened 2024-10-03
clean redudant line

hitesh-1997:
Clean redundant line as per: https://github.com/sourcegraph/cody/pull/5767#discussion_r1783712351

## Test plan
No functional change


hitesh-1997:
Fixed in https://github.com/sourcegraph/cody/pull/5798

PR #5773, opened 2024-10-02
Add a CI workflow that builds the JetBrains plugin as an FYI

dominiccooney:
Adds a workflow that builds JetBrains main with a Cody commit, as an FYI.

## Test plan

Look at output of the JetBrains FYI job.

PR #5794, opened 2024-10-03
Fix exported types conflicting with agent protocol

pkukielka:
## Test plan

Compile JetBrains repo with CODY_DIR set to checked out code from this branch.
There should be no errors.

### DIFF
--- a/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
+++ b/agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
@@ -1,11 +1,41 @@
 @file:Suppress("FunctionName", "ClassName", "unused", "EnumEntryName", "UnusedImport")
 package com.sourcegraph.cody.agent.protocol_generated;
 
-data class AuthStatus(
-  val endpoint: String,
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.Gson;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import java.lang.reflect.Type;
+
+sealed class AuthStatus {
+  companion object {
+    val deserializer: JsonDeserializer<AuthStatus> =
+      JsonDeserializer { element: JsonElement, _: Type, context: JsonDeserializationContext ->
+        when (element.getAsJsonObject().get("endpoint").getAsString()) {
+          "https://example.com" -> context.deserialize<UnauthenticatedAuthStatus>(element, UnauthenticatedAuthStatus::class.java)
+          else -> throw Exception("Unknown discriminator ${element}")
+        }
+      }
+  }
+}
+
+data class UnauthenticatedAuthStatus(
+  val endpoint: EndpointEnum, // Oneof: https://example.com
   val authenticated: Boolean,
   val showNetworkError: Boolean? = null,
   val showInvalidAccessTokenError: Boolean? = null,
+  val pendingValidation: Boolean,
+) : AuthStatus() {
+
+  enum class EndpointEnum {
+    @SerializedName("https://example.com") `Https-example-com`,
+  }
+}
+
+data class AuthenticatedAuthStatus(
+  val endpoint: EndpointEnum, // Oneof: https://example.com
+  val authenticated: Boolean,
   val username: String,
   val isFireworksTracingEnabled: Boolean? = null,
   val hasVerifiedEmail: Boolean? = null,
@@ -17,5 +47,11 @@ data class AuthStatus(
   val displayName: String? = null,
   val avatarURL: String? = null,
   val userCanUpgrade: Boolean? = null,
-)
+  val pendingValidation: Boolean,
+) : AuthStatus() {
+
+  enum class EndpointEnum {
+    @SerializedName("https://example.com") `Https-example-com`,
+  }
+}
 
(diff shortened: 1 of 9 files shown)


agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
```diff
@@ -1,11 +1,41 @@
...
-data class AuthStatus(
-  val endpoint: String,
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.Gson;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import java.lang.reflect.Type;
+
+sealed class AuthStatus {
```
pkukielka, line 11: Deserialisation code is incorrect, we need to fix this before merging.

pkukielka: I patched it manually. It is not perfect but should be enough for a temp solution.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
```diff
@@ -1,21 +1,59 @@
+/*
```
dominiccooney, line 1: You could review https://github.com/sourcegraph/cody/pull/5773 so people are aware when they are breaking JetBrains (...eventual goal to make that land blocking.)

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
```diff
@@ -1,21 +1,59 @@
...
-  val userCanUpgrade: Boolean? = null,
-)
+import com.google.gson.annotations.SerializedName;
+import com.google.gson.Gson;
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonDeserializer;
+import com.google.gson.JsonElement;
+import java.lang.reflect.Type;
 
+// TODO: This was edited manually due to definiencies in the codegen.
```
dominiccooney, line 21: Spelling, deficiencies

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
dominiccooney: Is it incorrect because it is just out of date, or is it more fundamental?

If there's a bigger issue here someone fixing CODY-3809 needs to work on, could you drop a sketch of the issue in the comment, maybe make it TODO(CODY-3809): ...

dominiccooney (approved):
Approving to unblock because this is part of a multi-sided patch to unblock JetBrains release.

However we are compounding tech debt here, which is not great. I'm concerned that there is no test plan, so how will we ensure this doesn't break when CODY-3089 is fixed? Could you:

- Add a test plan
- Mention TODO(CODY-3089) in your comment, and add some details about what is hand-edited here
- Drop a note in CODY-3089 pointing to this PR and noting that we need to run the test plan after switching back to the generated bindings.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
pkukielka: It is a bigger issue, I already discussed it with @olafurpg and @sqs on slack.
I will add comment in the linear.

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
pkukielka: Fixed.

pkukielka:
> Approving to unblock because this is part of a multi-sided patch to unblock JetBrains release.
> 
> However we are compounding tech debt here, which is not great. I'm concerned that there is no test plan, so how will we ensure this doesn't break when [CODY-3089](https://linear.app/sourcegraph/issue/CODY-3089) is fixed? Could you:
> 
> * Add a test plan
> * Mention TODO([CODY-3089](https://linear.app/sourcegraph/issue/CODY-3089)) in your comment, and add some details about what is hand-edited here
> * Drop a note in [CODY-3089](https://linear.app/sourcegraph/issue/CODY-3089) pointing to this PR and noting that we need to run the test plan after switching back to the generated bindings.

1. I'm not sure what testplan it should be.
TBH this would and should be caught by automatic binding regeneration.
We ran into this issue because it was disabled. 
2. I already mentioned CODY-3089 in my comment before but I made it more explicit now.
3. I added a note in linear (and changed it title to be more truthful) 👍

agent/bindings/kotlin/lib/src/main/kotlin/com/sourcegraph/cody/agent/protocol_generated/AuthStatus.kt
dominiccooney: Still not right: definiencies -> deficiencies

dominiccooney:
Wow, JetBrains FYI did its thing: https://github.com/sourcegraph/cody/actions/runs/11175844736/job/31067972094?pr=5794

pkukielka:
> Wow, JetBrains FYI did its thing: https://github.com/sourcegraph/cody/actions/runs/11175844736/job/31067972094?pr=5794

This requires fix on the jetbrains side, I will merge this PR and soon raise a fix with a cody commit bump.

PR #5795, opened 2024-10-03
Add globalStorage/didChange agent endpoint

pkukielka:
## Test plan

N/A. It will be needed for https://github.com/sourcegraph/jetbrains/pull/2382

Protocol files need to be regenerated but [this have to be fixed and merged first.](https://github.com/sourcegraph/cody/pull/5794)

### DIFF
--- a/agent/src/global-state/AgentGlobalState.ts
+++ b/agent/src/global-state/AgentGlobalState.ts
@@ -6,14 +6,15 @@ import * as vscode_shim from '../vscode-shim'
 import path from 'node:path'
 import { localStorage } from '../../../vscode/src/services/LocalStorageProvider'
 import migrate from './migrations/migrate'
+import type {MessageHandler} from "../../../vscode/src/jsonrpc/jsonrpc";
 
 type GlobalStateManager = 'client' | 'server'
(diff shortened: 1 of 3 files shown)


agent/src/global-state/AgentGlobalState.ts
```diff
@@ -40,8 +42,11 @@ export class AgentGlobalState implements vscode.Memento {
         }
     }
 
-    private set(key: string, value: any): void {
+    private set(key: string, value: any) {
         this.db.set(key, value)
+        if (this.agent) {
```
dominiccooney, line 47: Let's not do it this way. VSCode `globalState` is shared, and there's no notifications when other agents are editing it. We should build the notification we need at the semantic level in the extension and use that, instead of exciting some action every time global state changes.

What specifically are we using this for on the JetBrains side?

dominiccooney (changes requested):
I think we need to build the notification we need at the semantic level, instead of adding a notification on global state writes. Global state doesn't have this kind of notification for other concurrent editors, and I doubt we need notice every time someone writes anything into global state.

PR #5799, opened 2024-10-03
do not block on CodyLLMSiteConfiguration (configOverwrites) fetch in initial auth

sqs:
To authenticate a user and start showing the UI, we actually don't need to
   fetch the CodyLLMSiteConfiguration (`configOverwrites`). This reduces the
   number of HTTP requests (albeit they were parallelized somewhat) needed
   before the user can be authenticated and start seeing the Cody UI.

   This uncovered some possible "bugs" in the `anthropic.test.ts` autocomplete
   tests, where they were asserting that the `model` would be set but it
   actually would not be in real life. In `Provider.maybeFilterOutModel` it
   sees that `this.configOverwrites?.provider === 'sourcegraph'` in the test,
   but that should not be true in real life because it’s being passed an
   `authStatus` that is non-dotcom. The reason the test is bad is because that
   method gets the auth status from `currentAuthStatus()`, which is dotcom in
   testing only because of a prior test's mock value.

- do not block on fetching the user's Cody Pro subscription status when authing

## Test plan

CI & e2e. Sign in and ensure that operation that need the site version (such as autocomplete) still function without an error.

## Changelog

- Made authentication faster and less prone to network instability by reducing the number of HTTP requests needed for authentication.

### DIFF
--- a/vscode/src/completions/providers/anthropic.test.ts
+++ b/vscode/src/completions/providers/anthropic.test.ts
@@ -141,11 +141,7 @@ describe('anthropic autocomplete provider', () => {
 
         expect(provider.id).toBe(providerId)
         expect(provider.legacyModel).toBe(legacyModel)
-        expect(getRequestParamsWithoutMessages(provider)).toStrictEqual({
-            ...requestParams,
-            // The model ID is ignored by BYOK clients
-            model: undefined,
-        })
+        expect(getRequestParamsWithoutMessages(provider)).toStrictEqual(requestParams)
     })
 
     it('[dotcom] site-config-cody-llm-configuration special case for google hosted models', async () => {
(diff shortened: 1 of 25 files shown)


sqs:
cc @RXminuS

RXminuS:
Just having a look at another failing test where UI seems to render before it responds to clicks to account tab

sqs:
@valerybugakov Can you PTAL at the changes I needed to make to the `anthropic.test.ts` autocomplete provider tests? I think some of the expected values were still wrong as I understood the actual expected behavior, but I am not sure. The purpose of this PR is to extract `configOverwrites` into a separate observable and value, so it's possible that the new code has a bug or the old code has a bug in reading the wrong `configOverwrites`.

vscode/src/completions/providers/anthropic.test.ts
```diff
@@ -141,11 +141,7 @@ describe('anthropic autocomplete provider', () => {
 
         expect(provider.id).toBe(providerId)
         expect(provider.legacyModel).toBe(legacyModel)
-        expect(getRequestParamsWithoutMessages(provider)).toStrictEqual({
-            ...requestParams,
-            // The model ID is ignored by BYOK clients
-            model: undefined,
-        })
+        expect(getRequestParamsWithoutMessages(provider)).toStrictEqual(requestParams)
```
valerybugakov, line 144: In the previous version, the test used `mockAuthStatus(AUTH_STATUS_FIXTURE_AUTHED)`,  which didn't have `configOverwrites`. This was inconsistent with provider creation arguments and implicitly made it a BYOK provider. Thanks for fixing this!

valerybugakov (approved):

PR #5804, opened 2024-10-04
Send info if model is available together with the model

pkukielka:
## Changes

This change will allow clients to know if model is useable for them with  of manually fetching account tier and computing model availability.
This makes client code simpler, ensures the same conditions are used by both Cody and clients, and is less error prone.

## Test plan

This is only used in JetBrains currently.

Testplan on the JetBrains side:
https://github.com/sourcegraph/jetbrains/pull/2386

### DIFF
--- a/lib/shared/src/models/modelsService.ts
+++ b/lib/shared/src/models/modelsService.ts
@@ -207,6 +207,11 @@ export interface LocalStorageForModelPreferences {
     setModelPreferences(preferences: PerSitePreferences): Promise<void>
 }
 
+export interface ModelAvailabilityStatus {
+    model: Model
+    isModelAvailable: boolean
+}
+
 /**
  * ModelsService is the component responsible for keeping track of which models
  * are supported on the backend, which ones are available based on the user's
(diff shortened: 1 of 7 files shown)


lib/shared/src/models/modelsService.ts
```diff
@@ -207,6 +207,11 @@ export interface LocalStorageForModelPreferences {
     setModelPreferences(preferences: PerSitePreferences): Promise<void>
 }
 
+export interface ModelAvailabilityStatus {
+    model: Model
+    isModelAvailable: boolean
+}
```
mkondratek, line 213: It is OK but...

image

We have `ModelTag`s to provide the model's meta-data. And a few values are called "statuses" in the comment.

mkondratek (approved):
I think the `ModelTag` approach would be slightly cleaner but anyway lgtm

lib/shared/src/models/modelsService.ts
pkukielka: We already have model tags in the model.
But that does not solve our issue.
It's ok to know if model is available for `free` or `pro` user, but that in itself does not answer the question if current user is `pro` or `free`. That is something outside of the model itself.

PR #5808, opened 2024-10-04
Chat: Fix hanging issue in repository name resolution for workspaces

abeatrix:
CLOSE: https://linear.app/sourcegraph/issue/CODY-3971/investigate-submit-button-issue-in-chat-for-enterprise-users

The `remoteReposForAllWorkspaceFolders` function was incorrectly handling non-array results, such as errors or pending operations, which could cause the function to hang indefinitely. This change filters out these non-array results and returns an empty array instead, ensuring the function behaves correctly.

Description:

This PR addresses an issue where Cody would hang when encountering a workspace folder without a remote origin URL. The problem was caused by incorrect error handling in the repository name resolution process, which led to unresolved promises and pending operations that cause the function to hang indefinitely.

Root Cause:

- The RepoNameResolver class was catching errors in getUniqueRemoteUrlsCached and getRepoNameCached methods but returning empty arrays or null values instead of properly propagating the errors.
- These non-error, non-array results were being interpreted as pending operations in the remoteReposForAllWorkspaceFolders observable, causing the process to hang.

Solution:

1. Modified getUniqueRemoteUrlsCached to return null instead of an empty array on error.
2. Updated getRepoNameCached to handle null cases explicitly and ensure consistent return values.
3. Adjusted getRepoNamesContainingUri to filter out null values before returning results.
4. Removed the premature pendingOperation check in the abortableOperation within remoteReposForAllWorkspaceFolders.

These changes ensure that errors and missing data are handled consistently throughout the resolver process, preventing the hanging behavior and allowing the extension to continue functioning even when some workspace folders lack remote URLs or other errors that block the extension from working.

## Test plan

Manual Testing:

1. Verified that the extension no longer hangs when processing workspaces without remote URLs.
2. Confirmed that repository name resolution works correctly for valid remote URLs.
3. Tested with mixed scenarios of workspaces with and without remote URLs.

Before: if you open a repository that’s working with Cody on S2, and then run  git remote remove origin to remove the origin from the repo locally, chat will stop working after reloading the Window.

## Changelog

Chat: Fix an issue in repository name resolution for workspaces that caused Chat to hang.

### DIFF
--- a/vscode/src/repository/remoteRepos.ts
+++ b/vscode/src/repository/remoteRepos.ts
@@ -6,7 +6,7 @@ import {
     fromVSCodeEvent,
     graphqlClient,
     isError,
-    pendingOperation,
+    type pendingOperation,
     startWith,
     switchMapReplayOperation,
 } from '@sourcegraph/cody-shared'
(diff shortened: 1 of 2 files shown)


vscode/src/repository/remoteRepos.ts
```diff
@@ -61,22 +61,27 @@ export const remoteReposForAllWorkspaceFolders: Observable<
...
-                    if (repoNames === pendingOperation) {
-                        return pendingOperation
-                    }
                     if (repoNames.length === 0) {
                         // If we pass an empty repoNames array to getRepoIds, we would
                         // fetch the first 10 repos from the Sourcegraph instance,
                         // because it would think that argument is not set.
                         return []
                     }
+                    // Proess the validated results without checking for pendingOperation that
```
hitesh-1997, line 83: nit: process

hitesh-1997 (approved):
Amazing !! Thanks for fixing, Tried the repro steps in the test plan, chat no longer hangs for me for workspace without remote urls 🚀🚀

arafatkatze (approved):
Thanks Bee 🎉 🎉 
This works perfectly for me 🙇

PR #5810, opened 2024-10-04
show notice to remind Sourcegraphers to dogfood s2 not dotcom

sqs:
## Test plan

Sign into s2 and confirm it doesn't show. Sign into another internal instance and confirm it doesn't show. Sign into dotcom and ensure it does show and is dismissible.


camdencheek (approved):

abeatrix (approved):
Love the icons at the top. We should add a donate button for each bug we catch with switching to S2 🤣

PR #5811, opened 2024-10-04
Fix styling of avatar component when no avatarURL is present

jasonhawkharris:
This is a stop-gap commit that fixes the users profile picture so that it doesn't break the UI. The previous behavior can be viewed [in this issue](https://linear.app/sourcegraph/issue/SRCH-1139/fix-chat-ui-issues-caused-by-profile-pictures-in-cody).

Ultimately, we should follow up with another PR with a more elegant solution, but this fixes the immediate problem. 

## Test plan
- Pass a broken avatar link into the UserAvatar component and make sure it doesn't break the UI.
- Passing CI

## Changelog

- Fixes a bug where avatar's were stretching to fit the content of the alt text instead of remaining a static size.

### DIFF
--- a/vscode/webviews/components/UserAvatar.module.css
+++ b/vscode/webviews/components/UserAvatar.module.css
@@ -6,7 +6,7 @@
     color: var(--vscode-inputOption-activeForeground);
     align-items: center;
     justify-content: center;
-    height: fit-content;
+    height: 100%;
 }
 
 .sourcegraph-gradient-border {


vscode/webviews/components/UserAvatar.module.css
```diff
@@ -6,7 +6,7 @@
     color: var(--vscode-inputOption-activeForeground);
     align-items: center;
     justify-content: center;
-    height: fit-content;
+    height: 100%;
```
camdencheek, line 9: Does this still size the avatar correctly for a non-broken avatar? `height:100%` looks a little fishy to me 🤔

slimsag:
I am guessing we'll need to update the copy of this package in the search product later, as well?

vscode/webviews/components/UserAvatar.module.css
jasonhawkharris: Yes, it does.

jasonhawkharris: Will add that I checked that as well.

jasonhawkharris:
@slimsag This isn't an imported component, it lives in the Cody code base and is different from the avatar component that code search uses.

vscode/webviews/components/UserAvatar.module.css
jasonhawkharris: Proof: 
Screenshot 2024-10-04 at 1 06 28 PM

PR #5815, opened 2024-10-04
DeepCody: clean up and fix stream

abeatrix:
FIX: https://sourcegraph.slack.com/archives/C07CNJTGB3Q/p1727983740644479

Updated to wait for streaming is done before starting the context retrieval process.

Also included some minor refactoring on namings

## Test plan

Check the output channel to verify each tool is only run once after the streaming is completed.


vscode/src/chat/agentic/DeepCody.ts
```diff
@@ -1,150 +1,161 @@
...
-        const agenticContext = await this.review(chatAbortSignal)
-        // Run review again if the last requested context contains codebase search results.
-        if (agenticContext?.some(c => c.type === 'file')) {
-            this.currentContext.push(...agenticContext)
-            const additionalContext = await this.review(chatAbortSignal)
-            agenticContext.push(...additionalContext)
+    public async getContext(
+        chatAbortSignal: AbortSignal,
+        loop = 2,
+        maxNewItems = 20
```
hitesh-1997, line 82: nit: can we put a comment why we decide these numbers and see if they can be constant somewhere.

hitesh-1997 (approved):
LGTM !!

PR #5819, opened 2024-10-05
feat(uninstall): Delete cached credentials during re-install (CODY-1043)

jamesmcnamara:
This is the VSCode half of [CODY-1043](https://linear.app/sourcegraph/issue/CODY-1043/bug-account-credentials-are-cached-between-installs).

Unfortunately, we don't have access to any VSCode APIs during the uninstall script; it is just a raw node process that VSCode kicks off the next time VSCode is started after an uninstall (could be an arbitrary time after uninstalling). So the way we collected telemetry was to create files in the deactivation subroutine (which is run anytime the extension is stopped, i.e. when it is deactivated, VSCode is closed, the extension is upgraded or uninstalled) that stored all the config needed to boot telemetry and then we read those resources in the uninstall script.

Here we do basically the opposite. The uninstall script is run only when the user explicitly uninstalls the extension and closes and re-opens VSCode (importantly it does not run when the extension is upgraded). We create a marker file in the uninstallation process which we can check when the extension is initializing. When present, we delete the file and perform any cleanup from the previous installation (currently just deleting auth credentials and previous endpoints). 

This also fixes some existing errors in the uninstall script (it isn't working today because the client capabilities aren't initialized), adds a check to the bundling step for the uninstaller that no vscode dependencies are transitively included (similar to the vscode-shim) and adds a re-install telemetry event.

## Test plan
Finally added an E2E test for the uninstall flow. It took forever, but seeing as how it was broken for gosh knows how long and we didn't know indicates that it was necessary.

To manually test it:
1. Build the VSIX bundle with `pnpm -C vscode _build:vsix_for_test`
2. Start up VSCode and uninstall the Cody extension (or use the alternate editor, for instance I use VSCode Insiders day to day so I do this testing in VSCode).
3. From the command palette run "Extension: Install from VSIX" and select `vscode/dist/cody.e2e.vsix`. Log into Cody normally.
4. Close and reopen code. Verify that your credentials are cached.
5. Uninstall Cody (click on the gear icon on the extension view and select uninstall).
6. Fully close Code (not just the open window but the whole process)
7. Re-open Code and re-install Cody from VSIX.
8. Verify that credentials are cleared and you are presented with a fresh login screen.

## Changelog
deletes auth credentials and endpoint history when the extension is reinstalled


vscode/src/services/LocalStorageProvider.ts
```diff
@@ -140,6 +140,13 @@ class LocalStorage implements LocalStorageForModelPreferences {
         await this.set(this.LAST_USED_ENDPOINT, null)
     }
 
+    // Deletes and returns the endpoint history
+    public async deleteEndpointHistory(): Promise<string[]> {
+        const history = this.getEndpointHistory()
+        await Promise.all([this.deleteEndpoint(), this.set(this.CODY_ENDPOINT_HISTORY, null)])
```
abeatrix, line 146: We might also want to remove the stored tokens from the secretStorage for each endpoint?

vscode/src/extension.node.ts
```diff
@@ -66,8 +69,17 @@ export function activate(
...
     const authStatus = currentAuthStatus()
-    serializeConfigSnapshot({
+    let clientCapabilities: ClientCapabilities | undefined
+    try {
+        clientCapabilities = getClientCapabilities()
+    } catch {
+        // If client capabilities cannot be retrieved, we will just synthesize
+        // them from defaults in the post-uninstall script.
+    }
+    await serializeConfigSnapshot({
```
abeatrix, line 79: we only have a short window when `deactivate` is called. 
I wonder if it would be easier to always take this snapshot one sign-in instead so that it'd still work if users sign out before uninstalling?

vscode/src/services/LocalStorageProvider.ts
jamesmcnamara: We do, it's [here](https://github.com/sourcegraph/cody/pull/5819/files#diff-c0148ab15a21c10f2d174f2e2bd6b4dba61c243909f3439415fa2d7855b64ddbR171-R182)

vscode/src/extension.node.ts
jamesmcnamara: Updated!

agent/src/global-state/AgentGlobalState.ts
```diff
@@ -45,12 +45,10 @@ export class AgentGlobalState implements vscode.Memento {
     }
 
     public async reset(): Promise<void> {
-        if (this.db instanceof InMemoryDB) {
-            this.db.clear()
+        this.db.clear()
```
abeatrix, line 48:
```suggestion
        if (this.db instanceof InMemoryDB) {
            this.db.clear()

            // HACK(sqs): Force `localStorage` to fire a change event.
            await localStorage.delete('')
        }
```

Reverting this change fixed the test for me locally 🤔 

image

abeatrix: never mind - if i restart it could fail

vscode/src/services/AuthProvider.ts
```diff
@@ -212,6 +218,33 @@ class AuthProvider implements vscode.Disposable {
...
     }
+
+    // When the auth status is updated, we serialize the current configuration to disk,
+    // so that it can be sent with Telemetry when the post-uninstall script runs.
+    // we only write on auth change as that is the only significantly important factor
+    // and we don't want to write too frequently (so we don't react to config changes)
+    // The vscode API is not available in the post-uninstall script.
+    // Public so that it can be mocked for testing
+    public async serializeUninstallerInfo(authStatus: AuthStatus): Promise<void> {
+        let clientCapabilities: ClientCapabilitiesWithLegacyFields | undefined
```
abeatrix, line 229:
```suggestion
        if (!authStatus.authenticated) return
        let clientCapabilities: ClientCapabilitiesWithLegacyFields | undefined
```

abeatrix (approved):
I didn't test the uninstall step since we now have a test that covers it, but i've started the build locally and everything works as expected!

PR #5833, opened 2024-10-07
Autocomplete: assert provider config and request params on S2

valerybugakov:
- Adds autocomplete config and requests params assertions for the S2 clients in agent tests.
- Switches autocomplete agent tests to use `triggerKind: 'Automatic'`, which generates fewer API requests (fewer Polly recordings) and is a more popular logic path we want to test.  
    - We have a confusing type for the `triggerKind` parameter which is different from the types we use in VS Code, where `invoked` means automatic and `manual` means `manual`.
- Created on top of https://github.com/sourcegraph/cody/pull/5831
- Part of https://linear.app/sourcegraph/issue/CODY-3910/add-autocomplete-integation-tests-for-server-side-model-config

## Test plan

CI and `pnpm vitest agent/src/enterprise-s2.test.ts`


vscode/src/completions/inline-completion-item-provider.ts
```diff
@@ -899,6 +899,7 @@ export class InlineCompletionItemProvider
     }
 
     public async manuallyTriggerCompletion(): Promise<void> {
+        console.log('manuallyTriggerCompletion')
```
sqs, line 902:
```suggestion
        console.log('manuallyTriggerCompletion')
```

sqs (approved):
LGTM

PR #5835, opened 2024-10-07
Record promptText for S2

thenamankumar:
Issue: https://linear.app/sourcegraph/issue/AI-337/create-way-to-query-s2-query-privatemetadata-data

We need to record `promptText` on S2 instance for learning from dog fooding. 

## Test plan

- Submit a chat and make sure that the prompt Text is being recorded.

### DIFF
--- a/lib/shared/src/telemetry-v2/events/chat-question.ts
+++ b/lib/shared/src/telemetry-v2/events/chat-question.ts
@@ -14,7 +14,7 @@ import {
 
 import type { PromptString } from '../../prompt/prompt-string'
 import { truncatePromptString } from '../../prompt/truncation'
-import { isDotCom } from '../../sourcegraph-api/environments'
+import { isDotCom, isS2 } from '../../sourcegraph-api/environments'
 import { CHAT_INPUT_TOKEN_BUDGET } from '../../token/constants'
 import type { TokenCounterUtils } from '../../token/counter'
 import { telemetryRecorder } from '../singleton'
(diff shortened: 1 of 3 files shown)


lib/shared/src/telemetry-v2/events/chat-question.ts
```diff
@@ -135,6 +136,7 @@ export const events = [
                             ? map.intent(params.detectedIntent)
                             : undefined,
                         ...metadata,
+                        recordsPrivateMetadataTranscript: isS2(params.authStatus) ? 1 : 0,
```
akalia25, line 139: Could you broaden this to: `isS2 OR isDotCom` we can collect from both.

lib/shared/src/telemetry-v2/events/chat-question.ts
```diff
@@ -150,6 +152,16 @@ export const events = [
                         userSpecifiedIntent: params.userSpecifiedIntent,
                         traceId: spans.current.spanContext().traceId,
                         gitMetadata,
+                        // 🚨 SECURITY: Chat transcripts are to be included only for S2 users AND for V2 telemetry.
+                        // V2 telemetry exports privateMetadata only for S2 users. The condition below is an additional safeguard measure.
+                        // Check `SRC_TELEMETRY_SENSITIVEMETADATA_ADDITIONAL_ALLOWED_EVENT_TYPES` env to learn more.
+                        promptText: isS2(params.authStatus)
```
akalia25, line 158: same here "Could you broaden this to: isS2 OR isDotCom we can collect from both."

akalia25 (approved):
Like the `isS2` approach! Approved with one comment to broaden for `isS2` OR `isDotCom`

lib/shared/src/telemetry-v2/events/chat-question.ts
thenamankumar: I didn't extent to isDotCom because it was present for DotCom until last week and then it was removed. I'll add back.

PR #5839, opened 2024-10-07
Chat: update logic to select Deep Cody as default model when available

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-3988/move-enrollment-out-of-models-getter that addresses comment in https://github.com/sourcegraph/cody/pull/5687/files#r1788676940 

Main change

Eemove the logic that mutate user-selected model from the "getter" in ChatController on models changes & add new logic handles setting Deep Cody as the default on first sync:

- Set Deep Cody as default chat model on first sync when it is available.
- Add logic to set the user's selected chat model to Deep Cody during the first sync.

This also fixes the issue where Deep Cody is showing up in the Edit model dropdown as we updated to set `data.preferences!.defaults.edit` and `data.preferences!.defaults.chat` separately:

image

#### Before

image

## Test plan

Manual Test Plan:

1. Log into S2 and make sure you can see `Deep Cody` is selected as the default model for you in the Chat model dropdown list.
2. Verify you can switch chat model
3. Log into DotCom to ensure everything works as expected 

Added new unit test for syncing Deep Cody model

### DIFF
--- a/lib/shared/src/models/sync.ts
+++ b/lib/shared/src/models/sync.ts
@@ -38,7 +38,6 @@ import { ModelUsage } from './types'
 import { getEnterpriseContextWindow } from './utils'
 
 const EMPTY_PREFERENCES: DefaultsAndUserPreferencesForEndpoint = { defaults: {}, selected: {} }
-const DEEPCODY_MODEL = getExperimentalClientModelByFeatureFlag(FeatureFlag.DeepCody)!
 
 /**
  * Observe the list of all available models.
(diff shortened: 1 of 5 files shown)


lib/shared/src/models/sync.ts
```diff
@@ -199,17 +199,30 @@ export function syncModels({
...
+                                                    ...maybeAdjustContextWindows([DEEPCODY_MODEL]).map(
                                                         createModelFromServerModel
                                                     )
                                                 )
-                                                data.preferences!.defaults.chat = DeepCody.modelRef
+                                                // Update model preferences for chat to DEEP CODY once on first sync.
+                                                data.preferences!.defaults.edit =
+                                                    data.preferences!.defaults.chat
+                                                data.preferences!.defaults.chat = DEEPCODY_MODEL.modelRef
+                                                return userModelPreferences.pipe(
```
abeatrix, line 218: this is the main change where we will update user's selected preference to Deep Cody once on sync. This allows users to switch to a different model after without needing to change the site default chat model

arafatkatze:
I'm not able to fully grok the complexities of observables and the different edge cases, but I do have the rudimentary understanding of the code. The way I tested this PR is by checking if it logically does what it claims to do:
- When Deep Cody is available, then it's set as the default chat model on the first sync ->  I logged into S2.
And I got deep cody in my chat and In the edit flow of S2, it did not show DeepCody as a model(expected behaviour)

- For dotcom, I don't see DeepCody as a model, both in chat and in edit, which I believe is expected behavior. 


So, approving the PR because it does what it claims to do.

arafatkatze (approved):

PR #5844, opened 2024-10-08
feat(ui-ModelSelectorField): Add max height and scrolling to model selector popover

PriNova:
This PR enhances the user experience of the model selector popover by adding a maximum height and enabling vertical scrolling. This change ensures that the popover remains within a reasonable size on the screen, even when there are many model options available.

Before:
Before_ModelList

After:

https://github.com/user-attachments/assets/21679d7c-1cfb-494e-81a1-190f8664994c

### Changes:

- Added a maximum height of 80% of the viewport (`tw-max-h-[80vh]`) to the `CommandList` component.
- Enabled vertical scrolling with `tw-overflow-y-auto` for the `CommandList`.
- These classes are added using the `cn` function, which allows for conditional class naming.

### Impact:

This change improves the usability of the model selector, especially for users with many model options or on smaller screens. It prevents the popover from potentially extending beyond the visible area of the screen, making all options accessible through scrolling.

## Test plan

Please test this change with various numbers of model options and on different screen sizes to ensure the scrolling behavior works as expected and the popover remains within the viewport.


PriNova:
@vovakulikov Please review this changes for enhanced UX.
Thank you

vscode/webviews/components/modelSelectField/ModelSelectField.tsx
```diff
@@ -190,7 +190,7 @@ export const ModelSelectField: React.FunctionComponent<{
                     data-testid="chat-model-popover"
                 >
                     <CommandList
-                        className="model-selector-popover"
+                        className={cn('model-selector-popover', 'tw-max-h-[80vh] tw-overflow-y-auto')}
```
abeatrix, line 193:
```suggestion
                        className="model-selector-popover tw-max-h-[80vh] tw-overflow-y-auto"
```

I think we can just use the string directly here since they are all strings?

abeatrix (approved):
Nice one!

vscode/webviews/components/modelSelectField/ModelSelectField.tsx
PriNova: ty

abeatrix:
verified the model selector is working as expected with improved UI:

image

Before

image

PR #5849, opened 2024-10-08
fix bug in cody bench

beyang:
Call `setClientCapabilitiesFromConfiguration(getConfiguration())` in the right place so `cody internal bench` doesn't throw an exception.

## Test plan

Only affects `cody internal bench`


camdencheek (approved):

olafurpg (approved):
LGTM It is a bit strange to me that this is needed because the command-bench.ts file runs on a Node.js process that orchestrates Cody agent processes, this process doesn't actually activate the VSC extension. Either way, this is a low-risk change

Issue #5296, opened 2024-08-22
bug: THe Cody chat window panel shows nothing but a gray background.


OCIO-ricky:
Type: <b>Bug</b>

## Extension Information

- Cody Version: 1.32.0
- VS Code Version: 1.92.2
- Extension Host: desktop

##  Steps to Reproduce

1.
2.
3.

## Logs

Extension version: 1.32.0
VS Code version: Code 1.92.2 (fee1edb8d6d72a0ddff41e5f71a671c23ed924b9, 2024-08-14T17:29:30.058Z)
OS version: Windows_NT x64 10.0.22631
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz (8 x 2592)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>skia_graphite: disabled_off<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled<br>webnn: disabled_off|
|Load (avg)|undefined|
|Memory (System)|31.83GB (16.24GB free)|
|Process Argv|--crash-reporter-id 0cd76725-8175-4795-9208-d6bca1b6fc07|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
vspor879:30202332
vspor708:30202333
vspor363:30204092
vswsl492:30256859
vscod805cf:30301675
binariesv615:30325510
vsaa593cf:30376535
py29gd2263:31024239
c4g48928:30535728
azure-dev_surveyone:30548225
962ge761:30959799
pythongtdpath:30769146
welcomedialog:30910333
pythonnoceb:30805159
asynctok:30898717
pythonregdiag2:30936856
pythonmypyd1:30879173
2e7ec940:31000449
pythontbext0:30879054
accentitlementsc:30995553
dsvsc016:30899300
dsvsc017:30899301
dsvsc018:30899302
cppperfnew:31000557
dsvsc020:30976470
pythonait:31006305
dsvsc021:30996838
[...]

PR #5850, opened 2024-10-08
Fix Cody Chat UI rendering reconciliation

vovakulikov:
Fixes part of https://linear.app/sourcegraph/issue/CODY-3943/optimize-performance-for-large-chat-windows-in-cody

It looks like since https://github.com/sourcegraph/cody/pull/5036 we've lost memoization in a few places. As a result, our chat UI doesn't perform well with more than 4-5 messages (depending on how long these chat messages are)

This PR tries to improve it by fixing memoization on expensive components (Transcript, HumanMessage, ...etc)

| Before | After |
| ------- | ------- |
| <video src="https://github.com/user-attachments/assets/79e7b93c-ac82-4a8c-9b2e-6c712af3fb2c" /> | <video src="https://github.com/user-attachments/assets/223b3513-0f30-47cc-b24c-c54f4861329c" /> | 

This PR also fixes a problem with one-box intention detection; previously, it produced a state where we didn't have a follow-up message UI during intent resolution. 

## Test plan
- Check with the React profile that we don't render components during message response in places where we don't need to re-render anything
- Check that intent detection in one-box doesn't produce UI flashes

### DIFF
--- a/vscode/webviews/chat/Transcript.tsx
+++ b/vscode/webviews/chat/Transcript.tsx
@@ -11,7 +11,7 @@ import { clsx } from 'clsx'
 import debounce from 'lodash/debounce'
 import isEqual from 'lodash/isEqual'
 import { Search } from 'lucide-react'
-import { type FC, memo, useCallback, useMemo, useRef, useState } from 'react'
+import { type FC, type MutableRefObject, memo, useCallback, useMemo, useRef } from 'react'
 import type { UserAccountInfo } from '../Chat'
 import type { ApiPostMessage } from '../Chat'
 import { Button } from '../components/shadcn/ui/button'
@@ -180,7 +180,7 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
         smartApplyEnabled,
     } = props
 
-    const [intentResults, setIntentResults] = useState<
+    const [intentResults, setIntentResults] = useMutatedValue<
         | { intent: ChatMessage['intent']; allScores?: { intent: string; score: number }[] }
         | undefined
         | null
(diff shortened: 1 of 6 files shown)


abeatrix (approved):
Tested locally and confirmed the states were not re-render.

I was looking into a similar issue for the editor state. Would appreciate your review on it as well since I think there are a few things I can clean up on my PR after yours is merged 😄  https://github.com/sourcegraph/cody/pull/5847

abeatrix:
(Not a blocker) would be great to have a test covering transcript rendering for regression if possible 😄

thenamankumar (approved):

vscode/webviews/utils/debugger.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useRef } from 'react'
+
+export function useTracePropsUpdate(props: any) {
+    const prev = useRef(props)
+
+    useEffect(() => {
+        const changedProps = Object.entries(props).reduce(
+            (ps, [k, v]) => {
```
dominiccooney, line 8: Our lints for PromptString rely on only using `ps` to refer to the PromptString formatter. This ensures we never alias `ps`, which can be used to defeat the static checks for correct PromptString usage.

We could write a more complicated check that understands JavaScript scoping to allow the use you have here, or only lint files which import "the" `ps`, but until we do that, can you rename these variables to anything else?

dominiccooney (changes requested):
We need a trivial renaming to keep the PromptString lint results accurate.

vscode/webviews/utils/debugger.ts
vovakulikov: Sure, renamed it in my last commit

vscode/webviews/utils/debugger.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useRef } from 'react'
+
+export function useTracePropsUpdate(props: any) {
```
dominiccooney, line 3: Let's add a comment here that this is for debugging with advice about how to use it.

From time to team people do mass-deletions based on use. You probably intend for this to stick around.

vscode/webviews/chat/Transcript.tsx
```diff
@@ -389,6 +389,17 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
     )
 }, isEqual)
 
+function useMutatedValue<T>(value?: T): [MutableRefObject<T | undefined>, setValue: (value: T) => void] {
```
dominiccooney, line 392: You should document that this should not be used for rendering and won't cause re-rendering.

vscode/webviews/chat/Transcript.tsx
```diff
@@ -180,7 +180,7 @@ const TranscriptInteraction: FC<TranscriptInteractionProps> = memo(props => {
         smartApplyEnabled,
     } = props
 
-    const [intentResults, setIntentResults] = useState<
+    const [intentResults, setIntentResults] = useMutatedValue<
```
dominiccooney, line 183: Why isn't simply `useRef` enough here?

dominiccooney (approved):
Approving for the `ps` fixes, thank you. I have not reviewed the rest in detail but I have some questions and suggestions inline.

I think the message output stuttering has a different cause, see https://github.com/sourcegraph/cody/issues/5296#issuecomment-2402410782

PR #5851, opened 2024-10-09
Agent: re-enable bindings generation CI check

olafurpg:
Fixes CODY-3809

Previously, we didn't automatically generate Kotlin bindings when making changes to the agent protocol. This was problematic because it meant that JetBrains, Visual Studio, and Eclipse teams need to manually write the bindings to stay updated with the latest Cody.

Now, we re-enable the CI check. In addition, the codegen tool also reports much more helpful errors now to increase the odds that a developer can work around codegen issues in the future when making changes to protocol types.

We have made a mistake in the past of leaking too many internal types into the protocol. This is problematic because internal types tend to lean heavily on advanced TypeScript typing techniques that are complicated to translate to other languages. Instead, we should start using more actively `Protocol*` named types, which mirror internal types but don't rely on advanced type machinery. This is a bit extra work on the agent side, but increases the odds we can keep the bindings generation enabled in CI. Internal types are regulary updated with features we don't need to care about in other IDE clients so this technique should also lower the burden on VSC changes to constantly run `pnpm generate-agent-kotlin-bindings`.

## Test plan
Green CI.


vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -70,7 +69,10 @@ export type ClientRequests = {
 
     // history is Map of {endpoint}-{username} to chat transcripts by date
     'chat/import': [
-        { history: Record<string, Record<string, SerializedChatTranscript>>; merge: boolean },
+        {
```
olafurpg, line 72: I have the biome extension installed on my computer, but for some reason I was getting tons of formatting changes. Probably due to a different version or something 😢

vscode/src/jsonrpc/agent-protocol.ts
```diff
@@ -736,6 +738,52 @@ export interface Range {
...
 
+// Equivalent to our internal `AuthStatus` type but using a string discriminator
+// instead of a boolean discriminator. Boolean discriminators complicate
+// deserializing in other languages. We have custom codegen for string
+// discriminators but not boolean ones.
+// It's good practice to be more intentional about the Agent protocol types
+// anyways.  As a rule of thumb, we should try to avoid leaking internal types
+// that are constantly making tiny changes that are irrelevant for the other
+// clients anyways.
+export type ProtocolAuthStatus = ProtocolAuthenticatedAuthStatus | ProtocolUnauthenticatedAuthStatus
```
olafurpg, line 749: The docstring here is important documentation. We need to start using these kinds of types more actively to keep these bindings alive. The other alternative is to move entirely to something like TypeSpec/Protobuf so 100% of the types in the protocol don't come anywhere from internal code.

agent/src/cli/scip-codegen/Codegen.ts
```diff
@@ -516,7 +536,14 @@ export class Codegen extends BaseCodegen {
                     memberTypeSyntax = enumTypeName + this.f.nullableSyntax(memberType)
                     enums.push({ name: enumTypeName, members: constants })
                 } else {
-                    this.queueClassLikeType(memberType, member, 'parameter')
+                    try {
+                        this.queueClassLikeType(memberType, member, 'parameter')
+                    } catch (error) {
+                        const stack = error instanceof Error ? '\n' + error.stack : ''
+                        const errorMessage = `error handling member: ${member.symbol}. To fix this problem, you may want to ignore it from code generation by adding the symbol name to the "ignoredProperties" in the Formatter.ts file.\n${error}${stack}`
```
olafurpg, line 543: These new error messages make it 100x easier to resolve some common errors. I didn't spend super much time on formulating the message so any feedback is appreciated 🙏🏻 The short explanation is that we can often fix errors by ignoring properties with weird types because they're not relevant for the protocol anyways.

agent/src/cli/scip-codegen/BaseCodegen.ts
```diff
@@ -290,8 +296,6 @@ export abstract class BaseCodegen {
         // literals. If you're hitting on this error with types like string
         // literals it means you are not guarding against it higher up in the
         // call stack.
-        // throw new TypeError(`type has no properties: ${this.debug(type)}`)
-        this.reporter.error('', `type has no properties: ${this.debug(type)}`)
-        return []
+        throw new TypeError(`type has no properties: ${this.debug(type)}`)
```
olafurpg, line 299: This change was OK because we now report much more helpful error messages

olafurpg:
Closing this as superseded by #5854

PR #5854, opened 2024-10-10
fix(codegen): Adds special cases for `AuthStatus` to re-enable codegen

jamesmcnamara:
Adds a special case for the AuthStatus to fix the code generation issues. Also refactors the code generation into separate classes for each language type leaves all SCIP logic in Codegen.

## Test plan
Regenerated all three language types and verified that code changes were just due to updates in the underlying types since disabling.


agent/src/cli/scip-codegen/Codegen.ts
```diff
@@ -1025,9 +525,22 @@ export class Codegen extends BaseCodegen {
...
+        if (info.display_name === 'AuthStatus') {
+            return {
+                symbol: info.symbol,
+                discriminatorDisplayName: 'authenticated',
+                members: unionTypes.map(type => ({
+                    type,
+                    value: type.type_ref.symbol.includes('Authenticated'),
+                })),
+            }
+        }
```
jamesmcnamara, line 541: This is the actual hack for AuthStatus

agent/src/cli/scip-codegen/Codegen.ts
```diff
@@ -1025,9 +525,22 @@ export class Codegen extends BaseCodegen {
         if (!type.has_union_type || type.union_type.types.length === 0) {
             return undefined
         }
+        const unionTypes = this.unionTypes(type)
+        // HACK: We have to special case the `AuthStatus` type because it's
+        // SCIP doesn't recognize the `authenticated` member as a constant type
+        // and thus picks the wrong discriminator
+        if (info.display_name === 'AuthStatus') {
```
olafurpg, line 532: This hack is OK for now. I think the cleanest fix for this is to introduce `ProtocolAuthStatus` with a string discriminator like in this PR here https://github.com/sourcegraph/cody/pull/5851

There's a limit how much we should reuse internal types in the protocol, and I think it's OK to draw the line at string-typed discriminators. It's not a lot of work to add dedicated protocol types so that we're more intentional about the design of the protocol. It's undesirable that small internal refactorings leak into diffs in the protocol that are irrelevant to agent clients.

olafurpg (approved):
I am in love with the changes in this PR. Thank you @jamesmcnamara 🙏🏻 🙏🏻 

The new language-specific classes clean up the code significantly. I'll share a bit more context via Slack

agent/src/cli/scip-codegen/Codegen.ts
jamesmcnamara: I like it! I incorporated your changes from that PR (I think I got all of them) and removed this hack. But I also left the code generation in place for generating deserializers based on booleans and integers. We'll probably never be able to use it, but if we do it it should just work 🤞

olafurpg: Agreed on keeping Boolean and numeric discriminator since you e already done the hard work. The codegen logic in this tool is reaching a level of sophistication that's quite far beyond what most off-the-shelf tools support. Would be cool to turn it into a standalone OSS project one day :)

jamesmcnamara: Haha it's so true. I thought the codegen tool I wrote at my last company was pretty nifty but it's a toy compared to what this can do.

olafurpg: Now we just need to convince #team-graph to accept adding `scip.Signature` https://github.com/sourcegraph/scip/pull/231 and merging https://github.com/sourcegraph/scip-typescript/pull/322 so we can stop running a fork of scip-typescript.

Then we need a nice API to read SCIP data from Sourcegraph and we're off to the races to generates all kinds of cool tooling/products 😎

PR #5856, opened 2024-10-10
Include draft prompts for viewer

thenamankumar:
closes: https://linear.app/sourcegraph/issue/SRCH-1138/make-your-drafts-be-available-in-the-editor
closes: https://linear.app/sourcegraph/issue/SRCH-1173/prompts-have-the-option-to-be-set-to-automatically-run-without-the

Blocked on: https://github.com/sourcegraph/sourcegraph/pull/1003

This PR includes the new `includeViewerDrafts: true` argument in prompts query, added as part of the backend PR. This will allow to list user's owned draft prompts along with all the non draft prompts accessible to the user.

This PR also integrates the new `autoSubmit` field on prompts, added as part of the backend PR. The prompts with autoSubmit set as true will be automatically executed in one click when the user selected them. For other only the input box will be updated with the prompt text as currently. 

First the backend change PR should be merged and deployed on both dotcom and s2 for this PR to be merged. 

## Test plan

- Visit the prompts screen and make sure the user's draft prompts are being listed. 

https://www.loom.com/share/4d74ba90ba6643a29d0798ad977a5c39

## Changelog

- List viewer's draft prompts in the Prompts Library.

### DIFF
--- a/lib/prompt-editor/src/PromptEditor.tsx
+++ b/lib/prompt-editor/src/PromptEditor.tsx
@@ -51,10 +51,10 @@ interface Props extends KeyboardEventPluginProps {
 
 export interface PromptEditorRefAPI {
     getSerializedValue(): SerializedPromptEditorValue
-    setFocus(focus: boolean, options?: { moveCursorToEnd?: boolean }): void
-    appendText(text: string, ensureWhitespaceBefore?: boolean): void
-    addMentions(items: ContextItem[]): void
-    setInitialContextMentions(items: ContextItem[]): void
+    setFocus(focus: boolean, options?: { moveCursorToEnd?: boolean }, cb?: () => void): void
+    appendText(text: string, cb?: () => void): void
+    addMentions(items: ContextItem[], cb?: () => void): void
+    setInitialContextMentions(items: ContextItem[], cb?: () => void): void
     setEditorState(state: SerializedPromptEditorState): void
 }
 
@@ -92,7 +92,8 @@ export const PromptEditor: FunctionComponent<Props> = ({
                 }
                 return toSerializedPromptEditorValue(editorRef.current)
             },
-            setFocus(focus, { moveCursorToEnd } = {}): void {
+            // biome-ignore lint/style/useDefaultParameterLast:
+            setFocus(focus, { moveCursorToEnd } = {}, cb): void {
                 const editor = editorRef.current
                 if (editor) {
                     if (focus) {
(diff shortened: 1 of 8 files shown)


vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
```diff
@@ -256,42 +256,51 @@ export const HumanMessageEditor: FunctionComponent<{
...
-                        return
+            ({ addContextItemsToLastHumanInput, appendTextToLastPromptEditor, submitHumanInput }) => {
+                // Add new context to chat from the "Cody Add Selection to Cody Chat"
+                // command, etc. Only add to the last human input field.
+                if (isSent) {
+                    return
+                }
+
+                const updates: Promise<unknown>[] = []
+                const awaitUpdate = () => {
```
vovakulikov, line 267: I'm just curious about what will happen if we don't have any updates; I assume that all editor-state updates happen in sync mode (it's rendering, which would be async, I guess). So, the submit message should have the most updated state even without awaiting updates. no? (probably not but just checking)

vovakulikov (approved):
Approve to unblock, still have this question about `awaitUpdates` but I don't think this is a blocker here

vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
thenamankumar: No, interestingly editorState updates are all async using a callback. Purely because of how the LexicalEditor API is implemented. 

If we do not await the updates, and submit, the submitted text won't have the appended prompt text, tests it thoroughly.

Thus implemented `awaitUpdated`.Passed the CB to all the editor interface functions which are performed async.

lib/prompt-editor/src/PromptEditor.tsx
```diff
@@ -51,10 +51,10 @@ interface Props extends KeyboardEventPluginProps {
 
 export interface PromptEditorRefAPI {
     getSerializedValue(): SerializedPromptEditorValue
-    setFocus(focus: boolean, options?: { moveCursorToEnd?: boolean }): void
-    appendText(text: string, ensureWhitespaceBefore?: boolean): void
-    addMentions(items: ContextItem[]): void
-    setInitialContextMentions(items: ContextItem[]): void
+    setFocus(focus: boolean, options?: { moveCursorToEnd?: boolean }, cb?: () => void): void
```
sqs, line 54: I don't see if this came up in code review, but I don't think a callback param is a good pattern here. Why not use a promise instead, which has clearer semantics and supports use in a caller that is itself async?

thenamankumar: Missed the comment. Will refactor this.

PR #5859, opened 2024-10-10
Chat: fix waitlist check on model sync

abeatrix:
## Issue

In the current implementation of `syncModels`, the waitlist check is not applied to users who have Deep Cody enabled. This is because the Deep Cody logic returns early from the observable pipeline, preventing the execution of the waitlist check for these users.

## Fix

This PR restructures the logic in `syncModels` to ensure that the waitlist check is applied to all users before handling the Deep Cody-specific logic.

## Changes

- Modified the `remoteModelsData` observable in `lib/shared/src/models/sync.ts` to perform the waitlist check before the Deep Cody check.
- Ensured that the `data` object is consistently updated with waitlist information for all users.
- Maintained the existing Deep Cody functionality for eligible users.

## Test plan

image

Sign in as DotCom user:

- [ ] Verified that the waitlist check is applied to users without Deep Cody enabled.
- [ ] Confirmed that the waitlist check is now also applied to users with Deep Cody enabled.
- [ ] Ensured that Deep Cody functionality remains intact for eligible users.
- [ ] Tested with various combinations of waitlist status and Deep Cody eligibility to ensure correct behavior.

### Before

o1 models showing up with "Join Waitlist" label for approved users:

image

### DIFF
--- a/lib/shared/src/models/sync.ts
+++ b/lib/shared/src/models/sync.ts
@@ -199,6 +199,25 @@ export function syncModels({
                                         featureFlagProvider.evaluatedFeatureFlag(FeatureFlag.DeepCody)
                                     ).pipe(
                                         switchMap(([hasEarlyAccess, deepCodyEnabled]) => {
+                                            // TODO(sqs): remove waitlist from localStorage when user has access
+                                            const isOnWaitlist = config.clientState.waitlist_o1
+                                            if (isDotComUser && (hasEarlyAccess || isOnWaitlist)) {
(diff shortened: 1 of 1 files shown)


lib/shared/src/models/sync.ts
```diff
@@ -225,25 +244,6 @@ export function syncModels({
                                                 )
                                             }
 
-                                            // TODO(sqs): remove waitlist from localStorage when user has access
```
abeatrix, line 228: just moving this beofore the deep cody check

valerybugakov (approved):

PR #5860, opened 2024-10-10
Chat: update range and source fields for priority context item

abeatrix:
CLOSE https://linear.app/sourcegraph/issue/CODY-4002/investigate-recent-changes-in-context-deduplication-logic

Issue: priority items are currently not deduplicated properly and are showing up twice in context list:

image

Root cause: Looks like  “open file ”  triggers the getVisibleEditorContext() which returns a [context item without the range field](https://sourcegraph.com/github.com/sourcegraph/cody/-/blob/vscode/src/chat/chat-view/context.ts?L116-120) and is being treated as context explicitly added by user which doesn’t get deduplicated.

Fix: Add range and new context source for Priotiy items that would be handled by the duplicate logic properly.

## Test plan

1. Ask Cody a question with codebase at-mentioned, and includes words that would trigger the priority context item, e.g. "open file" 
2. Verify the context files used are deduplicated correctly

## Changelog

Chat: handle duplicated priority context properly.


vscode/src/prompt-builder/unique-context.test.ts
```diff
@@ -305,6 +305,88 @@ describe('Unique Context Items', () => {
             expect(getUniqueContextItems([inner, noRange])).toStrictEqual([noRange])
             expect(getUniqueContextItems([inner, noRange, inner, inner])).toStrictEqual([noRange])
         })
+
+        it('should keep priority context item when it does not overlap with user-added item', () => {
```
valerybugakov, line 309: 🔥

valerybugakov (approved):

PR #5862, opened 2024-10-10
Adding more checks for absolute paths while generating new files with smart Apply

arafatkatze:
[Loom Video
](https://www.loom.com/share/bea3a44d28764ebd91170858915a5139)

[Linear Issue
](https://linear.app/sourcegraph/issue/QA-82/vsc-continuous-loading-is-shown-on-clicking-apply-button-when-source)

## Test plan
Tested locally and attached a video showing the solution

### DIFF
--- a/vscode/src/services/utils/codeblock-action-tracker.ts
+++ b/vscode/src/services/utils/codeblock-action-tracker.ts
@@ -170,7 +170,7 @@ export async function handleSmartApply(
 
     const uri =
         fileUri && workspaceUri
-            ? path.isAbsolute(fileUri)
+            ? path.isAbsolute(fileUri) && (await doesFileExist(vscode.Uri.file(fileUri || '')))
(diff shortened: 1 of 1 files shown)


vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -170,7 +171,7 @@ export async function handleSmartApply(
 
     const uri =
         fileUri && workspaceUri
-            ? path.isAbsolute(fileUri)
+            ? path.isAbsolute(fileUri) && fs.existsSync(fileUri)
```
abeatrix, line 174: We have a "doesFileExist" function in vscode/src/commands/utils/workspace-files.ts that would work in all VS code environments, let's use that one instead of the node:fs 😃

abeatrix (approved):
Let's use the vs code file system api instead of node:fs to make sure it work across all vs code environments :)

PR #5866, opened 2024-10-10
Chat: move chatModels observable creation to top level

abeatrix:
Description: This PR aims to improve the performance and reduce potential memory leaks in our chat component by lifting the chatModels observable creation to a higher-level component and passing it down as props. This change eliminates multiple subscriptions created for each chat message.

Key changes:

- Moved the chatModels observable creation from individual components to the CodyPanel component.
- Added a 'models' prop to relevant components to pass down the chat models.
- Updated the Transcript, HumanMessageCell, HumanMessageEditor, and Toolbar components to use the passed models prop instead of creating their own observable.
- Updated related test files and storybook stories to accommodate the new 'models' prop.

## Test plan

- Verify that the chat functionality works as expected with the new changes.
- Ensure that model selection still functions correctly in the chat interface.
- Check that all chat messages display and interact properly with the selected model.
- Monitor memory usage before and after these changes to verify reduced memory consumption.
- Use browser developer tools to confirm that fewer subscriptions are being created.

After

image

Before 

image

## Changelog

Chat: tbc

### DIFF
--- a/vscode/webviews/Chat.story.tsx
+++ b/vscode/webviews/Chat.story.tsx
@@ -1,3 +1,4 @@
+import { getMockedDotComClientModels } from '@sourcegraph/cody-shared'
 import { ExtensionAPIProviderForTestsOnly, MOCK_API } from '@sourcegraph/prompt-editor'
 import type { Meta, StoryObj } from '@storybook/react'
 import { Observable } from 'observable-fns'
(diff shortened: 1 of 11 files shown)


vovakulikov (approved):
Review, ok, tested manually found no regression, thanks

PR #5867, opened 2024-10-10
Chat: use memoized models props in HumanMessageEditor

abeatrix:
Follow up on https://github.com/sourcegraph/cody/pull/5866

- Use `useMemo` to get the first current chat model instead of `useObservable` to avoid unnecessary re-renders
- Update `Transcript` to use the `humanMessage.index` as the key for `TranscriptInteraction` components

## Test plan

Green CI

Same test plan as https://github.com/sourcegraph/cody/pull/5866

### After 

image

### Before

image

## Changelog

Chat: Fix performance issue with Chat webview.

### DIFF
--- a/vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
+++ b/vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
@@ -10,9 +10,7 @@ import {
 import {
     PromptEditor,
     type PromptEditorRefAPI,
-    useExtensionAPI,
     useInitialContextForChat,
-    useObservable,
 } from '@sourcegraph/prompt-editor'
 import clsx from 'clsx'
 import {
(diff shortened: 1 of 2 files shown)


vovakulikov (approved):

vscode/webviews/chat/cells/messageCell/human/editor/HumanMessageEditor.tsx
```diff
@@ -376,8 +374,3 @@ export const HumanMessageEditor: FunctionComponent<{
         </div>
     )
 }
-
-function useCurrentChatModel(): Model | undefined {
```
beyang, line 380: @abeatrix Drive-by comment/question on code structure here... Is there a way we can express still this as a custom hook that doesn't produce a bunch of Observables every time this hook is called? E.g., wrap `useObservable` in a `useMemo` call or something like that? Reason I ask is there are advantages in being able to provide a custom hook for this, and without something like that, it's entirely possible for devs to produce the same issue here (too many Observables) within the rendering code of some component.

PR #5870, opened 2024-10-11
Chat: use memoized chat models in AssistantMessageCell

abeatrix:
- Use the `models` prop passed down from the parent component instead of creating a new observable for the chat models for each assistant message cell
- Update `useChatModelByID` to use the passed `chatModels` array instead of creating a new observable
- Update `Transcript` to pass the `models` prop to `AssistantMessageCell`

## Test plan

No behavior changes. Everything including the tests are expected to work the same.


dominiccooney (approved):
Splendid.

PR #5313, opened 2024-08-23
Revert: rehype-highlight bump

RXminuS:
There has been a (long ongoing) issue with `rehype-highlight` / `highight-js` causing OOM issues on `>v6.0.0`. This reverts a recent version bump that might have [re-introduced this issue](https://github.com/sourcegraph/cody/pull/4459) in Cody chat causing grey-screens and crashes.

> Important: I wasn't able to re-produce the issue myself for some reason. So this is a bit of a "shot-in-the-dark" fix. Hopefully we can have those who reported the issue validate if it actually worked. Lowering the version shouldn't hurt any other changes though as it was merely done to reduce the bundle size somewhat.

## Test plan
CI

Issue #5187, opened 2024-08-13
bug: Chat: Python formatting does not show correct colors.


jeronimd:
### Version

1.92.0

### Describe the bug

Using chat.
When the bot tries to format code blocks:

<img src="https://github.com/user-attachments/assets/90376649-7ceb-4ba5-8233-b624b817d714" alt="codeblock" width="400"/>

It does not show the Python syntax highlighting fully.
I'm using Ollama, but it seems that all models have this problem.

### Expected behavior

What I expected are these colors:
<img src="https://github.com/user-attachments/assets/d10f8a5c-edbf-4158-a73d-4cb2185b164f" alt="codeblock" width="400"/>

ichim-david:
@jeronimd I am not working for Sourcegraph and I am just a user of Cody but here is my explanation of why this is happening.

The reason you are seeing this difference is because the chat side panel is a web application  which is like a shared asset used by each editor that they support, think Jetbrains, VS Studio, VS Code, Eclipse.

The syntax highlighting is done by this library and right now it uses the common option meaning that only 37 langauges have syntax highlighting https://github.com/wooorm/lowlight#common
It provides it's own coloring independent of what you have set in your editor.

This is why you see the difference and I don't know how easy it would be to make the syntax highlight take into consideration the syntax set by each editor.

PR #5874, opened 2024-10-11
Chat: brought back syntax highlighting for most common languages

ichim-david:
## Test plan
Ask chat to provide an example program for dart language. 
You should now see the dart snippets having syntax highlighting which started missing after
the work done in this pull request:
https://github.com/sourcegraph/cody/pull/5232

Example screenshot with the fix in action:
dart-syntax

## Changelog
Chat: brought back syntax highlighting for most common languages


h4rumon:
+1

ichim-david:
@sqs in https://github.com/sourcegraph/cody/pull/5232 you went from all to the common syntax bundle.
This meant that ppl now complain about missing dart syntax on GitHub and community forum
https://github.com/sourcegraph/cody/issues?q=is%3Aissue+is%3Aopen+syntax

I decided to give it a shot and I've added all of the languages that were part of the old "LANGUAGES" constant
that were missing from the common bunch.
These were:
    clojure,
    dart,
    dockerfile,
    elixir,
    fortran,
    groovy,
    haskell,
    html,
    http,
    jsonc,
    matlab,
    nix,
    ocaml,
    scala,
    verilog,
    vhdl,

Lowlight COMMON brings:
    'arduino',
    'bash',
    'c',
    'cpp',
    'csharp',
    'css',
    'diff',
    'go',
    'graphql',
    'ini',
    'java',
    'javascript',
    'json',
    'kotlin',
    'less',
    'lua',
    'makefile',
    'markdown',
    'objectivec',
    'perl',
    'php',
    'php-template',
    'plaintext',
    'python',
    'python-repl',
    'r',
    'ruby',
    'rust',
    'scss',
    'shell',
    'sql',
    'swift',
    'typescript',
    'vbnet',
    'wasm',
    'xml',
    'yaml'

in total there are 53 languages supported.

There was also a request for Delphi that I could add an import for as there is someone in the forum asking for it. Depending on how many people would need this, support for Delphi syntax could also be added.
https://community.sourcegraph.com/t/syntax-highlighting/236/2

I believe that this repo could also forgo the loading of the common selection from Lowlight and opt to import exactly what it thinks it should import based on whatever analytics you guys have on what languages are requested with Cody Chat.
Just like they do in common we can also have in utils/highlights.ts.
https://github.com/wooorm/lowlight/blob/main/lib/common.js

Once you decide if the current crop I have made is what you want to support this should also be mentioned in the documentation so that the end users know which languages Cody Chat supports for syntax highlighting.

ichim-david:
In the future either an alternative library should be used or extra works need to be done to see if you can read the color scheme of the editor and see if there isn't something that matches.

There is already an issue that mentions the difference in coloring from the chat to the actual editor color
https://github.com/sourcegraph/cody/issues/5187

vscode/package.json
```diff
@@ -1376,7 +1377,7 @@
     "postcss-scss": "^4.0.9",
     "react-markdown": "^9.0.1",
     "react-visibility-sensor": "^5.1.1",
-    "rehype-highlight": "^6.0.0",
+    "rehype-highlight": "^7.0.0",
```
vovakulikov, line 1380: Is there a reason to update rehype-highlight lib? We've seen some performance problems before with > 6.0.0 versions  (see https://github.com/sourcegraph/cody/pull/5313)

ichim-david: @vovakulikov I can switch back if you want, the reason was written in the commit message
- Updated rehype-highlight to ^7.0.0 in orde to get rid of lowlight
   2.9.0 and load only version 3.1.0

ichim-david: I have read the issue and given your find I will revert the version bump. Waiting for any other feedback so that I get them all in one go.

ichim-david: @vovakulikov pinned lowlight 2.9.0 and 6.0.0 this way we don't update to the version that has performance issues.

ichim-david:
@vovakulikov @dominiccooney I am aware now that the build fails after pinning lowlight to version 2.9.0 since they made the common export the way I've imported from 3.0.
I will fix this issue, initially my pull request was fine however due to me not knowing of issues with  "rehype-highlight": "^7.0.0" performance problem upon receiving feedback I've went back to 6.0.0 and wanted to get rid of the extra dependency of lowlight 3.1.
This is not easily seen for a contributor outside of the sourcegraph organisation such as myself where I won't even get any tests running until it's triggered by one of you guys.
Perhaps in the future this can be improved, for now that I know it's my fault with the latest code change  I will fix it.

vscode/package.json
```diff
@@ -56,7 +56,14 @@
     "version-bump:patch": "RELEASE_TYPE=patch ts-node-transpile-only ./scripts/version-bump.ts",
     "version-bump:dry-run": "RELEASE_TYPE=prerelease ts-node-transpile-only ./scripts/version-bump.ts"
   },
-  "categories": ["AI", "Chat", "Programming Languages", "Machine Learning", "Snippets", "Education"],
+  "categories": [
```
ichim-david, line 59: I didn't choose todo this format. Every time I save package.json within vscode I get this autoformat and this time it slipped my guard when I removed lowlight pin.

I don't know how you guys do not get package.json to re-format, looking at the biome config package.json is not an ignored file
https://github.com/sourcegraph/cody/blob/main/biome.jsonc#L82

ichim-david:
> @vovakulikov @dominiccooney I am aware now that the build fails after pinning lowlight to version 2.9.0 since they made the common export the way I've imported from 3.0. I will fix this issue, initially my pull request was fine however due to me not knowing of issues with "rehype-highlight": "^7.0.0" performance problem upon receiving feedback I've went back to 6.0.0 and wanted to get rid of the extra dependency of lowlight 3.1. This is not easily seen for a contributor outside of the sourcegraph organisation such as myself where I won't even get any tests running until it's triggered by one of you guys. Perhaps in the future this can be improved, for now that I know it's my fault with the latest code change I will fix it.

I have removed lowlight dependency opting to import directly the languages from highlight.js which is what lowlight was doing anyway.
This way we don't have to deal with lowlight 2.9.0 which is a dependency of rehype-highlight 6.0.

the list of languages are:
    arduino,
    bash,
    c,
    clojure,
    cpp,
    csharp,
    css,
    dart,
    delphi, (I added this myself in previous commit, the rest of languages are common plus your previous list)
    diff,
    dockerfile,
    elixir,
    fortran,
    go,
    graphql,
    groovy,
    haskell,
    html,
    http,
    ini,
    java,
    javascript,
    json,
    jsonc,
    kotlin,
    less,
    lua,
    makefile,
    markdown,
    matlab,
    nix,
    objectivec,
    ocaml,
    perl,
    php,
    phpTemplate,
    plaintext,
    python,
    pythonRepl,
    r,
    ruby,
    rust,
    scala,
    scss,
    shell,
    sql,
    swift,
    typescript,
    vbnet,
    verilog,
    vhdl,
    wasm,
    xml,
    yaml,

At this point it is up to you guys on which languages you want to support and which not, if you are concerned with bundle size and you have some analytics on what people use as the most common languages then by all means let me know and I can trim this list even further.

vscode/package.json
dominiccooney: I think this arises because your IDE is formatting it with some different rule... if in VSCode, open that file, cmd-shift-P and Format Document with... and confirm that Biome is the default.

dominiccooney (approved):
This looks good to me.

How does this affect the plugin size?

It would be great to have a test that quickly confirmed some colors appear for a few major and minor kinds of files in this list.

vscode/package.json
ichim-david: @dominiccooney indeed I didn't have biome set as default which is a bit strange considering that the biome extension is installed and the biome config is within the Cody repo. At the very least VS Code should have given me the warning that biome isn't set as default for json files popup and ask me if I want to set it as default.
Nevertheless, I've modified the default so this shouldn't be a problem in the future.

ichim-david:
> This looks good to me.
> 
> How does this affect the plugin size?
> 
> It would be great to have a test that quickly confirmed some colors appear for a few major and minor kinds of files in this list.

@dominiccooney I'll run the vite-analyzer and check. I am still waiting to hear from you guys on which syntaxes you think it actually make sense to support. Besides the 17 new ones loaded I question the usefulness of a bunch of languages that were previously loaded from the common bundle. If you can give me a list of things you know you want to have syntax support for I can modify the pull request to only load those.
Thank you for your feedback!

dominiccooney:
Patched this locally and tried it, it's great. I have pushed it here to run CI: https://github.com/sourcegraph/cody/pull/5953

ichim-david:
> Patched this locally and tried it, it's great. I have pushed it here to run CI: #5953

@dominiccooney you've asked about the bundle difference. For Cody web the syntaxes are saved and loaded within dist/index.js
With these extra languages loaded:
dist/index.js                   2,605.76 kB │ gzip:   581.65 kB 
Without:
dist/index.js                   2,542.84 kB │ gzip:   564.06 kB
So a 17.59kb increase.

ichim-david:
> @ichim-david regarding syntaxes to support, I don't have data about generated code languages handy, but here's the top 50 languages for autocomplete suggestions.
> 
> Obviously some of these don't make sense, like we're not going to syntax highlight "txt" output.
> 
> Regarding the existing languages being dubious, which ones specifically? I would be reticent to take a language away unless the highlighting quality is poor or it is super expensive for some reason.
> 
> python typescriptreact typescript javascript php java javascriptreact go vue dart html markdown css csharp cpp kotlin yaml rust c scss json ruby plaintext sql blade svelte shellscript xml lua scala properties terraform elixir powershell dockerfile astro dockercompose apex jsonc prisma latex perl groovy luau hcl erb zig ini txt twig

@dominiccooney to finish this logic I've looked at your list and found some languages that there are syntaxes available and some that aren't.

Syntax not found
apex
typescriptreact (there is typescript and js)
javascriptreact
vue
blade
svelte
shellscript (we load shell)
terraform
astro
dockercompose
apex
prisma
hcl
zig

Languages with support from highlight.js (https://github.com/highlightjs/highlight.js/tree/main/src/languages)
and not loaded but could be if desired:
julia
powershell
latex
erb
properties
twig

beyang:
Merged in https://github.com/sourcegraph/cody/pull/5953.

Thank you for the contribution and apologies for the delay @ichim-david!

PR #5875, opened 2024-10-11
fix/chat: Improve performance in streaming responses in long chats

dominiccooney:
Webviews' RPC system registers "a lot" (in a long chat, hundreds) of 'message' event handlers. Each of these typically causes the following work:

- Deliver the message, including creating MessageEvent.data. For chat transcripts this can be a large object.
- Walk the whole event data, testing whether each value is a URI, and if so, instantiating a URI for it.
- Examine a field, like `type` or RPC stream ID, and decide to discard the message because it is for a different subscriber.

This work is dispatched synchronously to the hundreds of listeners on the webview main thread, causing us to have long pauses where we don't produce a frame. In extreme cases we can hang the renderer enough for VSCode to trigger a webview reload, (...but our webview reloading may not be robust because we do not use `retainContextWhenHidden`) leaving users with a blank sidebar/chat panel.

This change makes the hydrate message-check one property-discard steps extremely cheap by using a Proxy to lazily do the hydration. The typical case will allocate two proxies, cache one in a WeakMap, and return the primitive string value for the message type or RPC ID by reading through the proxy to the underlying message... effectively O(1) instead of O(message size). All of the objects in this fast path are short lived so should be ~free to collect.

Using Proxies has a caveat: Proxies cannot be cloned by the structured clone algorithm, so they can't be passed back via postMessage. This makes a symmetrical change to onMessage, which lazily hydrates, to postMessage, so that it forces hydration. However the caller must remember to do this if they pass these objects to other structured cloners.

This change does not address the root cause of too many event listeners. That is recommended follow-up work. Then we can remove lazy hydration.

Before and after videos demonstrating the effect of this simple change:

https://www.loom.com/share/dd37421120114f8f805a97ff0de7f366

https://www.loom.com/share/3ca61e9ab493400884ba44e49e55dbf0

## Test plan

This is a performance change, there should be no functional difference in behavior.

To verify performance improved:

1. Run the extension
2. Load your longest chat... dozens of messages and type a new chat query with a long response like `write a Java bytecode verifier in Rust`
3. Cmd-Shift-P, Developer: Open Webview Developer Tools
4. Performance tab, start recording
5. Back in VSCode, submit the chat. Verify that the response streams smoothly.
6. Back in Webview developer tools, stop the profiler. Verify that there are no (or few) dropped frames.
7. Lastly, verify with a heap snapshot that the heap has not regressed.

Profile before this change: Note the chat response takes ~52s with ~46s of script time and there are many slow (red) tasks in excess of 100-200ms each:

image

Profile after this change: Note that the total time is about 13s with 7s of script time and we only had four slow tasks (red marks) of about 50ms each. The extra rendering time is good thing, reflecting we produced a lot more frames much more quickly = smoothness. Note the CPU utilization (top line) is no longer pegged and spilling onto a second core.

image

Note, with a long enough chat and slow enough machine, this will lag again.

## Changelog

- Chat response streaming is much smoother, especially for long chats.

### DIFF
--- a/lib/shared/src/editor/hydrateAfterPostMessage.test.ts
+++ b/lib/shared/src/editor/hydrateAfterPostMessage.test.ts
@@ -7,36 +7,23 @@ const mockHydrateUri = (value: unknown) => {
     return { hydrated: true, originalValue: value }
 }
 
-describe('lazyHydrateAfterPostMessage', () => {
-    test('handles non-object values', () => {
-        expect(hydrateAfterPostMessage(7, mockHydrateUri)).toEqual(7)
+describe('hydrateAfterPostMessage', () => {
+    test('re-hydrates a dehydrated URI object', () => {
+        const dehydratedUri = () => ({ $mid: 1, path: '/path/to/resource', scheme: 'file' })
+        expect(isDehydratedUri(dehydratedUri())).toBe(true)
+        expect(hydrateAfterPostMessage(dehydratedUri(), mockHydrateUri)).toEqual(
+            mockHydrateUri(dehydratedUri())
+        )
     })
 
-    test('handles shallow, non-URI values', () => {
-        expect(hydrateAfterPostMessage({ foo: 7, bar: 'hello' }, mockHydrateUri)).toEqual({
-            foo: 7,
-            bar: 'hello',
-        })
+    test('re-hydrates a short array of dehydrated URIs', () => {
+        const dehydratedUris = () => [{ $mid: 1, path: '/path/to/resource', scheme: 'file' }]
+        expect(hydrateAfterPostMessage(dehydratedUris(), mockHydrateUri)).toEqual(
+            dehydratedUris().map(mockHydrateUri)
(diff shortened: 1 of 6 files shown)


dominiccooney:
It looks like this breaks @-mentions with line ranges. I changed the behavior of the hydrate function a bit, so I will look into that... I may need to update a caller.

dominiccooney:
This is because the lazy Proxies are not cloneable, so if they're posted back then they cause an exception.

dominiccooney:
Expect CI to cycle green, please take a look.

If this is too complicated, another simpler solution is to make onMessage take a fast filter and a slow callback. It makes like tedious for callers because they need to duplicate the set of objects they handle in their callback's `switch (message.type) ...`, and it might be slower for big messages which we only use subsets of. But with some API craft we can probably make something palatable, if this is too complex.

abeatrix (approved):
Tested locally without issues in a long convo 🤘

PR #5876, opened 2024-10-11
Simplify the logic in handle smart apply for better readability

arafatkatze:
@pkukielka Pointed out that this code was quite complex and had some redundancies so I made the double usage of ? operator much simpler by using the good old if/else statements. 

## Test plan
Tested locally with the edge cases

### DIFF
--- a/vscode/src/services/utils/codeblock-action-tracker.ts
+++ b/vscode/src/services/utils/codeblock-action-tracker.ts
@@ -168,15 +168,20 @@ export async function handleSmartApply(
     const activeEditor = getEditor()?.active
     const workspaceUri = vscode.workspace.workspaceFolders?.[0].uri
 
-    const uri =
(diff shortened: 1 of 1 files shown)


vscode/src/services/utils/codeblock-action-tracker.ts
```diff
@@ -168,15 +168,20 @@ export async function handleSmartApply(
...
+        } else {
+            uri = smartJoinPath(workspaceUri, fileUri)
+        }
+    } else {
+        uri = activeEditor?.document.uri
+    }
 
     const isNewFile = uri && !(await doesFileExist(uri))
-    if (isNewFile) {
+    if (isNewFile && uri) {
```
arafatkatze, line 184: Had to do this extra thing for typechecking stuff

PR #5877, opened 2024-10-11
Improve edit task and smart apply destination file path computation

pkukielka:
Fixes https://linear.app/sourcegraph/issue/QA-116

## Test plan

It's hard to reproduce actual issue as it depends on the LLM response.
Test scenario which sometimes can reproduce it:

**For "Generate Unit Test":**

1. Opens some workspace, and then file from outside of that workspace
2. Run "Generate Unit Test" for that file
3. Make sure test file is generated properly and the location of the file looks sensible

**For "Smart Apply":**

Case1:

1. Open some workspace
2. Ask Cody to generate some code sample in totally unrelated to the existing workspace (e.g. "quicksort in haskell")
3. Click Smart Apply
4. Make sure Cody created new file for the given snipped and it's name and location looks sensible

Case 2: 

1. Open some workspace, and then unrelated file from outside of that workspace
2. Ask Cody to generate some code related to the opened file
3. Click Smart Apply
4. Make sure Cody appended code to the opened file correctly

Test on both MacOs and Windows.

### DIFF
--- a/vscode/src/services/utils/edit-create-file.ts
+++ b/vscode/src/services/utils/edit-create-file.ts
@@ -1,0 +1,54 @@
+import path from 'node:path'
+import isEqual from 'lodash/isEqual'
+import * as vscode from 'vscode'
+import { doesFileExist } from '../../commands/utils/workspace-files'
+
+const separatorsRegex = /[/\\]/g
+
+/**
+ * Resolves URI to a workspace or absolute URI, handling various edge cases.
+ *
+ * If a URI is provided, it checks if the root directory of the URI exists. If it does, the URI is returned as-is.
+ * If the root directory does not exist, it attempts to join the URI with the base dir URI.
+ * If no URI is provided or no workspace folder is available, it falls back document URI is returned.
+ *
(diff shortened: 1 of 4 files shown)


vscode/src/services/utils/edit-create-file.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    if (!baseDirUri) {
+        return fallbackUri
+    }
+
+    return smartJoinPath(baseDirUri, uri)
+}
+
+export function smartJoinPath(baseDirUri: vscode.Uri, relativeFileUri: string): vscode.Uri {
```
arafatkatze, line 43: No need to export this as you don't call it anywhere?

pkukielka: It is needed for tests

arafatkatze (approved):
I tested the PR on mac and it worked fine for all the edge cases of smart apply and edit(that I am aware of). 

Piotr tested it on Windows also(I didn't do that myself  though)

PR #5881, opened 2024-10-11
move enterprise, improve svg

rrhyne:
Changes the order of the login options to present the enterprise login first.  Note, there were almost no code changes other than the order of those panels. 

Also fixes the Cody logo image size. 

Updated layout:
image

Current layout:
image

## Test plan

Manually testing the authentication methods to ensure they work after reordering the sections. Ensure the graphic displays as required.


abeatrix (approved):
LOOKS GREAT

PR #5883, opened 2024-10-12
Network: More Reliable Cody Networking & Proxy Support

RXminuS:
### Before

Firstly, networking was largely dictated by VSCode's patching of the built-in `http`, `https`, `net` and `tls` libraries. Thus preventing direct access to the original implementation of these methods. Being limited by VSCode's proxy options meant some of our customers also couldn't specify a pre-existing socket path.

Additionally if there were any problems in the network configuration it could be very hard to debug as there was not clear error log or status indication in the extension. This was made worse by that fact that changes to the settings would require a full app/extension reload.

Finally, a previous attempt at bypassing these patches by overriding `connect` on VSCode's PacProxyAgent so that `keep-alive` headers could be passed along to the library:

a. Didn't actually patch the agent anymore in VSCode 1.93+
b. Due to a module loading issue not all modules were using the same patched `PacProxyAgent`. For instance `completions-client` was using an unpatched `PacProxyAgent` whereas the graphql api client did use a patched version.
c. Didn't actually create connection re-use since VSCode creates a new Agent / Socket for every request regardless
d. Would only provide an injection point if specific VSCode settings were used. For instance using `proxySupport: off` would not work the way you'd expect where the `agent` set on a request level doesn't actually get used.
e. VSCode has in the last couple of months enabled and reversed the passing along of `keep-alive` multiple times as it continues to create issues. For example https://github.com/microsoft/vscode/issues/228872

Thus `keep-alive` can't be considered a stable mitigation for lowering TLS handshake overhead. What's more is that our current use of `http`/`https` would not work work correctly if `keep-alive` actually did function correctly. For example we use `req.on('close')` in the completions client which doesn't fire if a socket is kept alive. There's additional issues with proxies where `keep-alive` requires careful consideration to prevent resource exhaustion or even locking out other applications from being able to use the proxy. The use of `node-fetch`, other request wrappers and VSCode's involvement makes it hard to fully untangle and carefully implement this correctly.

### Now

Instead of overriding PacProxyAgent we re-patch the `http` and `https` agents to check if an agent is set on the request. If this agent provides the special symbol `bypassVSCodeSymbol` (which is a function that returns a boolean), then it has the opportunity to request bypassing VSCode's default network stack. If requested, this is done by ProxyWrapping the agent in such a way that `agent instanceof vscode.PacProxyAgent` is `true`. This check is done in [here in microsoft/vscode-proxy-agent](https://sourcegraph.sourcegraph.com/github.com/microsoft/vscode-proxy-agent@e2f71125da6ae1b726c3177fd26819de9183e034/-/blob/src/index.ts?L348), allowing us to bypass the subsequent confusing boolean checks and preventing further VSCode interference with the request.

This means that by default we rely on VSCode to provide networking and for most normal use cases their proxy settings should suffice. When running inside of the Cody Agent the http(s) agent set on the request is already used as per usual.

A new `DelegatingAgent` now observes `cody.net.proxy` variables and if they are configured correctly, it will notify that it wants to bypass VSCode. This means we can now also support non-standard proxy configurations such as directly using a socket path.

A new `NetworkDiagnostics` service now observes errors in the `DelegatingAgent` and provides a status bar indication if there are validation issues. If there were network issues while authenticating the user (during startup of the extension) it will also automatically re-try the auth on any `cody.net` configuration changes. Finally it also adds a new `Cody: Network` output panel that, depending on the verbosity set in VSCode, will display errors in the configuration / requests more clearly.

CleanShot 2024-10-15 at 10 15 08@2x

#### Walkthrough of the Cody network settings

<a href="https://www.loom.com/share/4170d47038264433b932f22915bfac49?sid=5e2ef979-99b2-482c-8c51-43dc52ce03a9">Walkthrough thumbnail</a>

#### Other Changes

- (Bug) Loading of CA certs on the agent incorrectly abandoned all built-in TLS certs
- (Pendatic) Removing `keep-alive` references throughout. It sets false expectations of this actually working.
- (Bug) Ensure agent/http patching is placed in modules without dependencies (other than types) to ensure they are loaded before `@sourcegraph/lib/shared`. This fixes a bug where importing of `sentry` from `lib-shared` would make it so that certain parts of the code were using an unpatched `PacProxyAgent` (and would have used an unpatched `http`/`https` agent)

### Future

- VSCode is moving towards use of Electron's network stack. This is currently behind an experimental flag. Further significant changes to the proxy/network configuration are very much expected.
- Using an http(s) proxy via `cody.net.proxy.endpoint` could require configuration of `cody.net.proxy.cacert` if the proxy uses self-signed certificates. Alternatively you can specify `cody.net.proxy.skipCertValidation` to not validate the proxy server's CA certs.
- We should be moving to HTTP2/3 instead of trying to patch HTTP1.1 with `keep-alive`. There's a  PoC native node-library (will comment on this PR when PR for that is open) that leverages `libcurl` to not only provide HTTP2/3 to the extension but also provide much more robust automatic loading of proxy config & ca-certs without the risk of differences between VSCode and Agent Extensions.
- Not only is `keep-alive` with use of EventSource problematic, EventSource provides many additional caveats when used in a proxy. There is an experimental `EventSource` coming to NodeJS but we're 2 major version behind where this can be enabled. Thus a more robust EventSource implementation is highly desired too.

## Test plan

- E2E tests are created to verify the UI around mis-configuration and network errors. However due to a recent change in VSCode moving from AMD to ESM the E2Ev2 framework is currently not working. Tests will be merged in a soon to come PR.
- Manually verified together with @peterguy that all scenarios for custom proxy configuration work. 
- Manually verified together with @peterguy that behavior under error conditions presents useful signals to the user (error messages, status displays, etc...)
- Manually verified that VSCode provides a default implementation for networking.
- Manually benchmarked (non-existent) performance difference for previous implementation between `keep-alive` flag provided and removed.


RXminuS:
* **#5883** <a href="https://app.graphite.dev/github/pr/sourcegraph/cody/5883?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/shared/src/configuration/environment.ts
```diff
@@ -1,15 +1,39 @@
...
 
 /**
- * This should be the only location we read process.env variables from. It's
- * just a small layer on top of raw process.env variables, so should be kept to
- * static or a get method at most.
+ * In general config values should be preferred over environment variables. The
+ * exeption roughly being:
+ * 1. Settings that shouldn't be used by users
+ * 2. Settings that have deep implications for the execution environment and
+ *    would require a reload of the extension/platform if changed. might cause
```
peterguy, line 10: > might cause ...
Don't keep us in suspense! ;-)

lib/shared/src/configuration/environment.ts
```diff
@@ -1,15 +1,39 @@
...
+ * 3. Security sensitive settings that might expose sensitive information or
+ *    untested/unsafe functionality if accidentally/maliciously enabled in the
+ *    config file.
+ *
+ * So if you do need an invorment variable this should be the only location we
```
peterguy, line 15:
```suggestion
 * So if you do need an environment variable this should be the only location we
```

lib/shared/src/configuration/environment.ts
```diff
@@ -1,15 +1,39 @@
...
 import type { UIKind } from 'vscode' // types are ok
+import { logError as _logError } from '../logger'
 
 /**
- * This should be the only location we read process.env variables from. It's
- * just a small layer on top of raw process.env variables, so should be kept to
- * static or a get method at most.
+ * In general config values should be preferred over environment variables. The
+ * exeption roughly being:
+ * 1. Settings that shouldn't be used by users
```
peterguy, line 8: > Settings that shouldn't be used by users

Can we elaborate here? Debug/Test/Dev settings? Any examples?

lib/shared/src/configuration/environment.ts
```diff
@@ -1,15 +1,39 @@
...
+ * process.env variables, so should be kept to static or a get method at most.
  *
  * We provide them as getters as some parts of code might rely on being able to
  * modify and re-read `process.env`.
+ *
  */
 
 export const cenv = defineEnvBuilder({
+    /**
+     * Mirrors the HTTP_PROXY and HTTPS_PROXY environment varilables but can be
```
peterguy, line 26:
```suggestion
     * Mirrors the HTTP_PROXY and HTTPS_PROXY environment variables but can be
```

lib/shared/src/configuration/environment.ts
```diff
@@ -103,3 +161,39 @@ function uiKind(uiKind: string | undefined): UIKind | undefined {
...
+    const noProxy = bool(noProxyRaw) ?? str(noProxyRaw)
+    // TODO: We probably want to have a look through this
+    // for now we just flag a warning
+    const httpsProxy = str(getEnv('HTTPS_PROXY'))
+    const httpProxy = str(getEnv('HTTP_PROXY'))
+
+    // we assume we always want to prioritize https requests by default
+    if (httpsProxy || httpProxy) {
+        if (noProxy) {
+            logErrorOnce(`Ignoring NO_PROXY=${noProxy}`)
```
peterguy, line 193:
```suggestion
            logErrorOnce(`HTTPS_PROXY or HTTP_PROXY set; ignoring NO_PROXY=${noProxy}`)
```

lib/shared/src/fetch.patch.ts
```diff
@@ -0,0 +1,62 @@
...
+ * control the network requests being made.
+ *
+ * To do this, we have a mutable agent variable that can be set externally. For
+ * this to work we need to ensure no other dependencies have loaded yet. That's
+ * why this module is seperate from the fetch.ts file which simply re-exports
+ * this variable. This way the patching code can simply import this specific
+ * file and gain early acceessss to the agentRef
+ *
+ * 🚨 Do not import this file unless you're specifically patching the variable.
+ * For read only access simply import the fetch.ts file
```
peterguy, line 12:
```suggestion
 * For read-only access simply import the fetch.ts file
```

vscode/src/completions/fast-path-client.ts
```diff
@@ -97,7 +97,7 @@ export function createFastPathClient(
         const headers = new Headers(fireworksCustomHeaders)
         // Force HTTP connection reuse to reduce latency.
         // c.f. https://github.com/microsoft/vscode/issues/173861
-        headers.set('Connection', 'keep-alive')
+        // headers.set('Connection', 'keep-alive')
```
olafurpg, line 100: Intentional? Fastpath is disabled for Enterprise so this may not be needed in the first place

olafurpg: Alternatively, is this because we're enabling keep-alive elsewhere? I see we removed this in the default client too, and doing so may introduce perf regressions.

lib/shared/src/configuration/environment.ts
```diff
@@ -103,3 +161,39 @@ function uiKind(uiKind: string | undefined): UIKind | undefined {
             return undefined
     }
 }
+
+function codyProxy(v: string | undefined, k: string): string | undefined {
```
olafurpg, line 165: What does this function do? Does it return a proxy URL?
```suggestion
function codyProxyURL(v: string | undefined, k: string): string | undefined {
```

olafurpg (changes requested):
I only did a quick skim over the diff. My main recommendations:

- This is a large PR. Are there low-risk big-diff changes we can land separately to minimize the blast radius?
- Let's update the PR description following https://www.notion.so/sourcegraph/Write-a-good-pull-request-description-610a7fd3e613496eb76f450db5a49b6e TLDR; "Previously, PROBLEM. Now, SOLUTION". It does not have to be long. One important question to answer in the description: what is the risk of merging this change?

vscode/src/completions/fast-path-client.ts
RXminuS: No this is because keep-alive doesn't actually work correctly with our implementation and the only reason it "seemed" to work was because VSCode was broken so it didn't actually re-use connections (thus there was no actual perf gain).

lib/shared/src/configuration/environment.ts
RXminuS: It's just a combination of NODE_HTTPS and NODE_HTTP proxy. It's needed for proxies to work on Agent

vscode/src/completions/fast-path-client.ts
RXminuS: It's disabled on an agent level so even if you'd try to set it, it wouldn't work

lib/shared/src/configuration/environment.ts
olafurpg: A better function name or docstring would be helpful. The `k` parameter also seems unused? It's unclear what `v` is or where it comes from.
```suggestion
/** Returns the HTTPS_PROXY or HTTP_PROXY environment variables with optional ability to disable these settings  we respect these settings */
function nodeHttpProxyURL(settingValue: string | undefined): string | undefined {
```

olafurpg: I generally don't nitpick so much about these kinds of things, and I generally don't write many docstrings, but this function here stood out as a weird one

RXminuS: Yeah fair enough 😆

RXminuS: I'll rename it to something like `CODY_NODE_PROXY_DEFAULT` to indicate it's not really a "us" feature

arafatkatze:
image
image

Is this some sort of a redundancy or are you folks setting different proxies for using the SG Vscode app and the Cody Vscode app?

peterguy:
@olafurpg 

> Are there low-risk big-diff changes we can land separately to minimize the blast radius?

We could break out the new proxy settings. They were the inspiration for the network stack changes. I'm not eager to do that because of the added overhead and delays to landing them 😅, but definitely can-do. 

> what is the risk of merging this change?

It's gated behind the setting `cody.net.mode`, so other than some hooking into VSCode startup, normal usage won't be impacted.

olafurpg:
As discussed over Slack, I'm on board with landing this ASAP. Will do a last round of review today and approve.

peterguy:
> As discussed over Slack

I see that convo now; thanks.

> Will do a last round of review today and approve.

Great! I'm wrapping up merging conflicts and mollifying tests.

lib/shared/src/configuration/environment.ts
```diff
@@ -103,3 +160,32 @@ function uiKind(uiKind: string | undefined): UIKind | undefined {
             return undefined
     }
 }
+
+function codyProxy(envValue: string | undefined): string | undefined {
```
olafurpg, line 164: My comments on about this function missing clarity are still unaddressed.

olafurpg: Not a blocking comment. However, if I'm asked to review a PR I'd expect a resolution on the comment. The original thread is here https://github.com/sourcegraph/cody/pull/5883#discussion_r1800740537

lib/shared/src/fetch.patch.ts
```diff
@@ -0,0 +1,62 @@
...
+ * In node environments, it might be necessary to set up a custom agent to
+ * control the network requests being made.
+ *
+ * To do this, we have a mutable agent variable that can be set externally. For
+ * this to work we need to ensure no other dependencies have loaded yet. That's
+ * why this module is seperate from the fetch.ts file which simply re-exports
+ * this variable. This way the patching code can simply import this specific
+ * file and gain early access to the agentRef
+ *
+ * 🚨 Do not import this file unless you're specifically patching the variable.
```
olafurpg, line 11: Does this module actually do patching on import? Looking at the file, there seem to be no side-effects during module loading. It's unclear where the patching actually happens

vscode/src/net/net.patch.ts
```diff
@@ -0,0 +1,217 @@
...
+
+// This should be a function that returns true/false wether this agent wants to
+// bypass vscode
+export const bypassVSCodeSymbol = Symbol('bypassVSCodeSymbol')
+let patched = false
+patchNetworkStack()
+// This needs to happen after so that we don't break import order
+globalAgentRef.blockEarlyAccess = true
+
+function patchNetworkStack(): void {
```
olafurpg, line 17: IIUC, this module unconditionally patches stuff during module initialization in `extension.node.ts`. This comment [here](https://github.com/sourcegraph/cody/pull/5883#issuecomment-2415478043) from @peterguy indicates the functionality is gated by `cody.net.mode`. What is the risk of this function crashing making the extension fail to load even when everything is OK without network patching?

vscode/src/net/net.patch.ts
```diff
@@ -0,0 +1,217 @@
+// 🚨 This patch file mast have NO additional dependencies as import order
```
olafurpg, line 1:
```suggestion
// 🚨 This patch file must have NO additional dependencies as import order
```

mkondratek (approved):

olafurpg (approved):
Let's merge this and be very careful with the rollout of this change as discussed [here](https://sourcegraph.slack.com/archives/C05AGQYD528/p1728982786019919?thread_ts=1728981853.764659&cid=C05AGQYD528).

I'm assuming people hesitated to approve this PR because it's doing a lot of advanced stuff with seemingly large accompanying risk, and the PR description and code implementation do not adequately address those concerns. I'm approving to unblock progress but expect you @RXminuS and @peterguy to closely follow through on the rollout of this change and make sure we rapidly rollback the PR if there are reported issues.

If the functionality is properly gated behind a setting, then we can consider fixing forward. An important goal that we should not compromise on is: let's make sure the functionality continues to work as before for users who can already successfully use Cody. We've shipped several regressions over the past weeks and we can't afford to ship more.

lib/shared/src/configuration/environment.ts
RXminuS: Yeah sorry, this was a bit messy because Peter and I kept pushing to the same branch but lost some things when I did a graphite sync

PR #5884, opened 2024-10-12
support forceHydration on PromptString

sqs:
PromptString instances can't be cloned because their reference value is used as the key in the WeakMap that implements immutability and encapsulation for PromptString.

(This broke something on a separate branch, so I extracted the fix and test case to merge separately here.)

## Test plan

Added test case


beyang (approved):

dominiccooney:
I'm surprised this happened, because:

```js
let x = {}
let y = new Proxy(x, {})
x.foo = 'bar'
y.foo // evalutes to 'bar'
x === y // is false
let z = new WeakMap()
z[x] = 7
z[y] // evaluates to 7
```

So unless PromptString is doing something with `===` then keying the weakmap should work...

PR #5890, opened 2024-10-12
different agentIDE for Cody Web vs. this repo's standalone web/

sqs:
`CodyIDE.Web` previously referred to both (1) Cody Web, i.e., the Cody sidebar in Sourcegraph Code Search and the Cody chat page, and (2) this repo's `web/` demo app. They are different and have different needs. It is generally bad to change behavior based on the identity (as opposed to capabilities) of the client, which will be addressed in https://github.com/sourcegraph/cody/pull/5889 and beyond, but for now, it's helpful for them to report different client identities.

## Test plan

CI


valerybugakov (approved):

vovakulikov:
> They are different and have different needs
@sqs, can you expand on this? Have you seen something that caused some troubles with the unified Cody Web name for this client on the web/ and the client in the Sourcegraph App? 

Currently, Cody Web build is broken because of this change (no worries I can fix it) just curious why you decided to separate them

